PR #443, opened 2019-05-23
Revert "http: delete conn parameter in readRequest (#430)"

ry:
Fixes https://github.com/denoland/deno_std/issues/441

We'll have to wait to remove the conn parameter until ws is working.

This reverts commit 209183e24812095a40e48e60484f80a5a254b1c3.

PR #550, opened 2019-07-30
Revert "http: delete conn parameter in readRequest (#430)"

### LINKED ISSUE #441: ws module broken
https://github.com/denoland/deno_std/pull/430 removed the `conn` property from `ServerRequest` "because it's not used anymore". However, this property is still used in `acceptWebSocket` (see: https://github.com/denoland/deno_std/blob/master/ws/mod.ts#L368). So the `ws` module is broken now.

bartlomieju:
Opening new PR instead of #443 cause I can't push there.

Fixes #441

Merged master and fixed other conflicts.


bartlomieju:
CC @ry

PR #554, opened 2019-08-08
print out the failed tests after the summary

oygen87:
We have 575 tests and it is a problem to scroll up the terminal to see which tests failed.
So I added a summary of all the failed tests down under with a red label next to them.

These changes will only affect cases when we have failed tests (If all tests pass, it will look like before.) 

Also some refactoring to use existing `TestResult` model and populate it with the test results. We only initialized it before when going Serial testing and could not extract the details from each test afterwards.

proposed look :
deno-test-runner2

### DIFF
--- a/testing/mod.ts
+++ b/testing/mod.ts
@@ -171,9 +171,7 @@ function printFailedSummary(results: TestResults): void {
     (v): void => {
       if (!v.ok) {
         console.error(`${RED_BG_FAIL} ${red(v.name)}`);
-        if (Deno.args.includes("--verbose")) {
-          console.error(v.error);
-        }
+        console.error(v.error);
       }
     }
   );
--- a/testing/mod.ts
+++ b/testing/mod.ts
@@ -104,8 +104,6 @@ const RED_FAILED = red("FAILED");
 const GREEN_OK = green("OK");
 const RED_BG_FAIL = bgRed(" FAIL ");
 
-const VERBOSE = true;
-
 interface TestStats {
   filtered: number;
   ignored: number;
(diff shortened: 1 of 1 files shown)


testing/mod.ts
```diff
@@ -102,6 +102,9 @@ export function test(t: TestDefinition | TestFunction): void {
 
 const RED_FAILED = red("FAILED");
 const GREEN_OK = green("OK");
+const RED_BG_FAIL = bgRed(" FAIL ");
+
+const VERBOSE = true;
```
zekth: possible to consider adding an arg like the silent mode? example `--verbose` ?

bartlomieju: What's the purpose of this variable? It seems superfluous since it's always true

oygen87: @bartlomieju the purpose is to lhave an option in the code for devs to change it later on, if we know which approach to go for, then we dont need the variable.

oygen87: @zekth Yes! this is best alternative. ill add this for sure and get rid of the variable @bartlomieju

testing/mod.ts
```diff
@@ -165,6 +166,19 @@ function report(result: TestResult): void {
   result.printed = true;
 }
 
+function printFailedSummary(results: TestResults): void {
+  results.cases.forEach(
+    (v): void => {
+      if (!v.ok) {
+        console.error(`${RED_BG_FAIL} ${red(v.name)}`);
+        if (Deno.args.includes("--verbose")) {
```
bartlomieju: I'm currently writing test runner which will abstract away most of the options for testing framework. Could you remove this conditional for now and always print error?

oygen87: okay, i see! yes ive fixed it now

ry:
Cool - looks good. I will wait for https://github.com/denoland/deno_std/pull/556 before landing this one.

ry:
@bartlomieju PTAL

PR #556, opened 2019-08-09
ci: bump to v0.15.0

bartlomieju:


nayeem-rahman:
@bartlomieju Is it supposed to be checking the format of the transpiler output JS? Because it is.

bartlomieju:
@ry should be landable. I had to pin `@typescript-eslint` deps.

Issue #555, opened 2019-08-09
[testing] assertEquals() throws a new Error() instead of a new AssertionError()


tedstoychev:
In [deno_std/testing/asserts.ts][1] in [the declaration of assertEquals()][2] a `new Error()` is thrown instead of a `new AssertionError()`.

### Possible fix:
Throw a `new AssertionError()` instead of a `new Error()` in [deno_std/testing/pretty.ts#L83][3].

[1]: https://github.com/denoland/deno_std/blob/master/testing/asserts.ts
[2]: https://github.com/denoland/deno_std/blob/master/testing/asserts.ts#L75-L81
[3]: https://github.com/denoland/deno_std/blob/3e17815c7e61cfd2e4c46a12649811543509d563/testing/pretty.ts#L83

PR #560, opened 2019-08-12
fix: Use AssertionError in assertEqual

bartlomieju:
Closes https://github.com/denoland/deno_std/issues/555

I merged `testing/asserts.ts` and `testing/pretty.ts` - both files were short so there's really no need to split them.

~~Note that this PR is blocked by #556~~


bartlomieju:
@ry should be landable

bartlomieju:
@ry rebased and added commit message: https://github.com/denoland/deno_std/pull/560/commits/9cd711fa72c174088cd65a4f8224354af7fd8b6b

PR #563, opened 2019-08-14
feat: add overloaded form of unit test declaration

### LINKED ISSUE #562: Add overloaded form of unit test declaration: test(name, fn)
There are currently two ways to declare a unit test
```javascript
// original form
test(function foo() { assert(true); });
// expanded form
test({ name: "foo", fn: () => { assert(true); });
```
It turns out that many of the tests in deno_std are using the expanded form. We could shorten this significantly by also allowing
```js
test( "foo", () => { assert(true); });
```

Someone has suggested this before and noted that a certain existing test framework uses this form as well.  (I can't find a reference.) So that's another reason to change.

Edit: reportedly Ava.js, Mocha and Jest all support this.

bartlomieju:
Closes #562

### DIFF
--- a/testing/mod.ts
+++ b/testing/mod.ts
@@ -87,8 +87,8 @@ function filter(name: string): boolean {
 }
 
 export function test(t: TestDefinition): void;
-export function test(t: TestFunction): void;
-export function test(name: string, t: TestFunction): void;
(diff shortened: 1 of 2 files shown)


bartlomieju:
CC @ry

PR #566, opened 2019-08-15
sprintf implementation / golang fmt.

a2800276:
I've attached code for an sprintf implementation that is (loosely) based on the golang fmt package. I've been told there may be some interest to include it to std.

[I created a post request a couple of hours ago but was not aware of all the linter CI going on, sorry about that.]

It's still in and early form, but fairly full featured.

A number of semantic issues are still unclear to me, mostly concerning what the default representation of arbitray objects should be, what flags/variants would be useful, etc. I've outlined a number of those issues in the README and would be interested in receiving input into those points as well.


j-f1:
A useful feature that Python’s format strings have is the ability to pass an object to the format function instead of a list of values. You can use `%(key)s` in the string to pull keys off the object.

a2800276:
I like the idea and have thought about it previously, I'm not sure what the implementation would look like. Currenlty, positional params are indicated with:

    %[<index>]s

where <index> refers to the index'th paramter. Sematics of the `sprintf` function already implies that there are a number of different parameters. To reference keys, it becomes necessary to indicate both the parameter passed to `sprintf` as well as the key or index of the value withing the parameter, i.e:

     %[<index>][<key>]s

or

    %[<index>.<key>]s

Alternatively, the <key> based value selection _could_ imply only passing a single parameter.

OR there could be a different formatting function for this entirely.

I'd be super happy to hear your thoughts of what the interface / format string syntax for this could look like, because I can't really think of anything reasonable myself.

What I would like to avoid is changing the semantics of `printf` entirely, the function should be able to perform "normal" formatting tasks in the way someone acquainted with posix would expect.

PR #568, opened 2019-08-18
fix: add global test registry

bartlomieju:
After adding `deno test` command a new problem appeared.

If you try running `deno test` inside this repo test runner will find 80 test files but won't run any tests! This is caused by fact that `deno test` uses tagged version of standard library which causes `test` function available inside to repo to be different function that `test` available in standard lib used by `deno test`.

```
~/src/deno_std> deno test -A
Found 79 matching test files.
Hello world!
running 0 tests

test result: OK. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (0.34ms)
```

A simple fix appears to be to add a global registry of all tests. It causes to register every test (even using `test` function from different version of standard lib) in the same global array that can be picked up by test runner.

That's just first iteration and we'll probably arrive at better solution before landing this PR.

CC @ry

### DIFF
--- a/testing/mod.ts
+++ b/testing/mod.ts
@@ -69,8 +69,19 @@ function print(txt: string, newline: boolean = true): void {
   Deno.stdout.writeSync(encoder.encode(`${txt}`));
 }
 
+declare global {
+  interface Window {
+    __DENO_TEST_REGISTRY: TestDefinition[];
+  }
+}
+
+let candidates: TestDefinition[] = [];
+if (window["__DENO_TEST_REGISTRY"]) {
+  candidates = window.__DENO_TEST_REGISTRY as TestDefinition[];
+} else {
+  window["__DENO_TEST_REGISTRY"] = candidates;
+}
 let filterRegExp: RegExp | null;
-const candidates: TestDefinition[] = [];
 let filtered = 0;
 
 // Must be called before any test() that needs to be filtered.
--- a/testing/mod.ts
+++ b/testing/mod.ts
@@ -71,6 +71,16 @@ function print(txt: string, newline: boolean = true): void {
 
 declare global {
   interface Window {
+    /**
+     * A global property to collect all registered test cases.
+     *
+     * It is required because user's code can import multiple versions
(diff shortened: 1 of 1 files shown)


testing/mod.ts
```diff
@@ -69,8 +69,19 @@ function print(txt: string, newline: boolean = true): void {
   Deno.stdout.writeSync(encoder.encode(`${txt}`));
 }
 
+declare global {
+  interface Window {
+    __DENO_TEST_REGISTRY: TestDefinition[];
```
ry: How about sticking it on the Deno object - like Deno.testRegistry ?

bartlomieju: Yes, that's fine for me :+1: thanks

ry: Probably there needs to be an explanatory comment about it's purpose - I think it will be confusing to readers

bartlomieju: @ry I think it won't be possible - firstly I would need to overwrite `Deno` namespace and additionally `window.Deno` is frozen - we can't add any new properties to it

bartlomieju:
@ry do you want me to change anything else in this PR?

testing/mod.ts
```diff
@@ -69,8 +69,29 @@ function print(txt: string, newline: boolean = true): void {
   Deno.stdout.writeSync(encoder.encode(`${txt}`));
 }
 
+declare global {
+  interface Window {
```
ry: Is it really necessary to declare this before hand?

bartlomieju: Without it TS complains that `__DENO_TEST_REGISTRY` doesn't exist on `Window`:
```
error TS2339: Property '__DENO_TEST_REGISTRY' does not exist on type 'Window'.

► file:///Users/biwanczuk/dev/deno_std/testing/mod.ts:90:23

90   candidates = window.__DENO_TEST_REGISTRY as TestDefinition[];
```

I can add `@ts-ignore` but this solution is more idiomatic

ry: Ok sounds good.

PR #591, opened 2019-09-10
feat: Add types to prettier

bartlomieju:
This PR adds types to `prettier`.

WARNING: types are for `1.18` version, while `prettier/` uses `1.17.1` version. I couldn't find types `1.18`, so it's a good idea to update prettier anyway.

This was suggested by @MarkTiedemann in denoland/deno#2490 and providing types significantly shortens time to compile prettier`

```
Before:
$ time deno --reload --no-prompt ../deno_std/prettier/main.ts
real	0m36.094s
user	0m43.474s
sys	0m1.148s

After:
$ time deno_dev --reload --no-prompt ../deno_std/prettier/main.ts
real	0m6.710s
user	0m10.205s
sys	0m0.232s
```
(Note that `after` is debug build, while `before` is `v0.17.0` release)

To get this to work we need to land denoland/deno#2910 first

### DIFF
--- a/prettier/prettier.ts
+++ b/prettier/prettier.ts
@@ -1,7 +1,11 @@
 // Copyright 2018-2019 the Deno authors. All rights reserved. MIT license.
+// @deno-types="./vendor/standalone.d.ts"
 import "./vendor/standalone.js";
+// @deno-types="./vendor/parser_typescript.d.ts"
 import "./vendor/parser_typescript.js";
+// @deno-types="./vendor/parser_babylon.d.ts"
 import "./vendor/parser_babylon.js";
+// @deno-types="./vendor/parser_markdown.d.ts"
 import "./vendor/parser_markdown.js";
 
 // TODO: provide decent type declarions for these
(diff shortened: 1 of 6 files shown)


bartlomieju:
@axetroy can you upgrade `prettier` to `1.18`?

prettier/prettier.ts
```diff
@@ -1,7 +1,11 @@
 // Copyright 2018-2019 the Deno authors. All rights reserved. MIT license.
+// @deno-types="./vendor/standalone.d.ts"
```
ry: Can you leave a comment explaining why you're giving types here? so we remember this issue in a few months?

ry:
Nice!

prettier/prettier.ts
bartlomieju: Done!

bartlomieju:
It works fine for me with denoland/deno#2910 and it's just ignored without it.

I can update prettier tomorrow to match version of types.

PR #592, opened 2019-09-11
bump prettier to 1.18.2

axetroy:
refs: #591


bartlomieju:
Should there be so many changes after the upgrade of `prettier`? I'd expect not to reformat any files... 🤔

axetroy:
@bartlomieju 
Prettier does need to reformat the file after the upgrade.

But this upgrade got a strange thing, I can run the test locally, but it does not pass in CI.

prettier/vendor/standalone.js
```diff
@@ -26228,6 +26335,130 @@ function isSimpleFlowType(node) {
...
+    });
+    parts.push("`", indent$6(concat$11([hardline$8, join$7(hardline$8, table.map(function (row) {
+      return join$7(" | ", row.cells.map(function (cell, index) {
+        return row.hasLineBreak ? cell : cell + " ".repeat(maxColumnWidths[index] - getStringWidth$2(cell));
+      }));
+    }))])), hardline$8, "`");
+    return concat$11(parts);
+  }
+}
+/** @param node {import("estree").TemplateLiteral} */
```
j-f1: Could that error be caused by this line?

axetroy: Resolved, Looks like this is the problem.

Thanks

PR #596, opened 2019-09-12
Fix issue when HTTP when no body is sent for POST

### LINKED ISSUE #538: [HTTP] Body's Promise is blocked when no body is present
- Code to replicate:

```ts
import { serve } from "https://deno.land/std@v0.11.0/http/server.ts";
const s = serve("0.0.0.0:8000");

async function main() {
  for await (const req of s) {
    const body = await req.body()
    console.log('Request body:', body)
    req.respond({ body: new TextEncoder().encode("Hello World\n") });
  }
}

main();
```
- Send GET: http://127.0.0.1:8000/ OK

image

- Send POST (with some body): http://127.0.0.1:8000/ OK

image

- Send POST (without body): http://127.0.0.1:8000/ BLOCKS

image

Mark1626:
Fixes #538

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -112,7 +112,7 @@ export class ServerRequest {
   public async *bodyStream(): AsyncIterableIterator<Uint8Array> {
     if (this.headers.has("content-length")) {
       const len = +this.headers.get("content-length")!;
-      if (Number.isNaN(len)) {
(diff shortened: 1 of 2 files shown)


M4rk9696:
\cc @ry

Issue #428, opened 2019-05-21
import typescript into deno_std


ry:
Yes, it's built into deno, but it's difficult to access from the special snapshot where it's stored.

It would be nice to have access to typescript in 3rd party code. So the easy solution is to just import it in here.

kitsonk:
@ry and talked about this... it makes sense... largely we need a CompilerHost that works well under Deno, which would be similar to what we use in the built in Deno compiler (possibly to become _the_ host).

I am going to be taking a look at it, but anyone else wanting to help would be glad to talk about what needs to be done.

ry:
As a very first pass, I think just having a test that loads TS and checked the version (or some other basic thing) would be sufficient.

kitsonk:
For anyone interested, the "trick" is that `typescript.js` loads in two ways, either as CommonJS or as global (see: microsoft/TypeScript#32949).  Because it does not detect what it needs for CommonJS it creates a global variable `ts` when loaded.

So we need to combine the `@deno_types` directive and also import `typescript.js` and scope the type definitions in the global scope (similar to what we do in `deno/deno_typescript/globals.d.ts` or `deno/js/ts_global.d.ts`).

So something like this:

```ts
// @deno-types="./vendor/typescript.d.ts"
import "./vendor/typescript.js";

import * as ts_ from "./vendor/typescript.d.ts";

declare global {
  namespace ts {
    export = ts_;
  }
}
```

kitsonk:
@ry I have it mostly working with the following (with the couple recent patches I made to deno master):

```ts
// @deno-types="./vendor/typescript.d.ts"
import "./vendor/typescript.js";

import * as ts_ from "./vendor/typescript.d.ts";

declare global {
    // eslint-disable-next-line @typescript-eslint/no-namespace
    namespace ts {
        // @ts-ignore
        export = ts_;
    }
}
```

But the problem is that we load everything as a module into the isolate at runtime, which means that the global `ts` variable is created only in the scope of the module.  Unlike prettier, which properly detects that it isn't being loaded as a CommonJS or AMD module and defaults to exposing itself as a global variable, TypeScript only detects CommonJS otherwise just keeps itself scoped.  I already opened microsoft/TypeScript#32949 which would fix this properly, but it seems like the only real choice I have is to postpend to `vendor/typescript.js` something that properly lifts it into the global scope.

PR #597, opened 2019-09-16
Add TypeScript (3.5.1)

kitsonk:
Refs #428 

This is an MVP for supporting TypeScript as part of the std library.

It is a copy of the `typescript/lib` in `typescript/vendor` with and addition to the `./vendor/typescript.js` so that it properly exposes itself globally when loaded in Deno.  It also requires https://github.com/denoland/deno/commit/c30decab77c4ebeb34ebd1d73b0f45bb26944c79 or later to work properly.


brandonros:
Any idea why CI/CD would be failing?

kitsonk:
> Any idea why CI/CD would be failing?

Yeah, as mentioned, it needs a new version of Deno to be updated in CI, then I need to rebase the PR.

hayd:
Can this be a git submodule? Then when deno_std is migrated to deno it could be a symlink.

zekth:
> Can this be a git submodule? Then when deno_std is migrated to deno it could be a symlink.

As the move is coming in a short amount of time, this patch is not really usefull i think.

kitsonk:
> As the move is coming in a short amount of time, this patch is not really usefull i think.

As it stands no. It was more to show how we would solve #428.  @ry asked me specifically to look at it.  I got it working, but I was leaving Ry to comment on it to any degree he felt he wanted to.

PR #598, opened 2019-09-16
Changed tests for HTTP versions to a Regex

jmisiti42:
I removed the `switch` and conditions there was inside of the `default` to one unique condition using regexp to test if `vers` is formatted as an http version.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -229,8 +229,7 @@ function fixLength(req: ServerRequest): void {
 }
 
 // ParseHTTPVersion parses a HTTP version string.
-// "HTTP/1.0" returns (1, 0, true).
-// Ported from https://github.com/golang/go/blob/f5c43b9/src/net/http/request.go#L766-L792
+// "HTTP/1.0" returns (1, 0).
 export function parseHTTPVersion(vers: string): [number, number] {
   const httpReg = /^HTTP\/([0-9])\.([0-9])$/gm;
   const match = httpReg.exec(vers);
--- a/http/server.ts
+++ b/http/server.ts
@@ -231,7 +231,7 @@ function fixLength(req: ServerRequest): void {
 // ParseHTTPVersion parses a HTTP version string.
 // "HTTP/1.0" returns (1, 0).
 export function parseHTTPVersion(vers: string): [number, number] {
-  const httpReg = /^HTTP\/([0-9]{0,2})\.([0-9]{0,2})$/gm;
+  const httpReg = /^HTTP\/([0-9]+)\.([0-9]+)$/gm;
   const match = httpReg.exec(vers);
 
   if (match == null) {
--- a/http/server.ts
+++ b/http/server.ts
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
 // "HTTP/1.0" returns (1, 0, true).
 // Ported from https://github.com/golang/go/blob/f5c43b9/src/net/http/request.go#L766-L792
 export function parseHTTPVersion(vers: string): [number, number] {
-  switch (vers) {
-    case "HTTP/1.1":
-      return [1, 1];
+  const httpReg = /^HTTP\/[0-9]\.[0-9]$/gm; // test if string is formatted as an HTTP version
+  let major: number;
+  let minor: number;
 
-    case "HTTP/1.0":
-      return [1, 0];
-
-    default: {
-      const Big = 1000000; // arbitrary upper bound
-      const digitReg = /^\d+$/; // test if string is only digit
-      let major: number;
-      let minor: number;
-
-      if (!vers.startsWith("HTTP/")) {
-        break;
-      }
-
-      const dot = vers.indexOf(".");
-      if (dot < 0) {
-        break;
-      }
-
-      let majorStr = vers.substring(vers.indexOf("/") + 1, dot);
-      major = parseInt(majorStr);
-      if (
-        !digitReg.test(majorStr) ||
-        isNaN(major) ||
-        major < 0 ||
-        major > Big
-      ) {
-        break;
-      }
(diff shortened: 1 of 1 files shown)


http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
 // "HTTP/1.0" returns (1, 0, true).
 // Ported from https://github.com/golang/go/blob/f5c43b9/src/net/http/request.go#L766-L792
 export function parseHTTPVersion(vers: string): [number, number] {
-  switch (vers) {
-    case "HTTP/1.1":
-      return [1, 1];
+  const httpReg = /^HTTP\/[0-9]\.[0-9]$/gm; // test if string is formatted as an HTTP version
```
nayeemrmn:
```js
const httpReg = /^HTTP\/([0-9])\.([0-9])$/gm;
```

http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
...
-
-      if (!vers.startsWith("HTTP/")) {
-        break;
-      }
-
-      const dot = vers.indexOf(".");
-      if (dot < 0) {
-        break;
-      }
+  if (!httpReg.test(vers)) {
```
nayeemrmn:
```js
const match = httpReg.exec(vers);
if (match == null) {
```

http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
...
-      major = parseInt(majorStr);
-      if (
-        !digitReg.test(majorStr) ||
-        isNaN(major) ||
-        major < 0 ||
-        major > Big
-      ) {
-        break;
-      }
+  const dot = vers.indexOf(".");
```
nayeemrmn: Not needed.

http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
...
-      if (
-        !digitReg.test(minorStr) ||
-        isNaN(minor) ||
-        minor < 0 ||
-        minor > Big
-      ) {
-        break;
-      }
+  let majorStr = vers.substring(vers.indexOf("/") + 1, dot);
+  let minorStr = vers.substring(dot + 1);
```
nayeemrmn: Forgot this one.
```js
const [, majorStr, minorStr] = match;
```

http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
...
 
-      return [major, minor];
-    }
-  }
+  major = parseInt(majorStr);
+  minor = parseInt(minorStr);
```
nayeemrmn:
```js
const major = parseInt(majorStr);
const minor = parseInt(minorStr);
```

http/server.ts
```diff
@@ -232,55 +232,23 @@ function fixLength(req: ServerRequest): void {
 // "HTTP/1.0" returns (1, 0, true).
 // Ported from https://github.com/golang/go/blob/f5c43b9/src/net/http/request.go#L766-L792
```
nayeemrmn: These don't apply anymore.

http/server.ts
```diff
@@ -232,55 +232,19 @@ function fixLength(req: ServerRequest): void {
...
 export function parseHTTPVersion(vers: string): [number, number] {
-  switch (vers) {
-    case "HTTP/1.1":
-      return [1, 1];
+  const httpReg = /^HTTP\/([0-9])\.([0-9])$/gm;
```
nayeemrmn: Ah, the versions can be more than one digit. Try:
```js
const httpReg = /^HTTP\/([0-9]+)\.([0-9]+)$/gm;
```

http/server.ts
```diff
@@ -229,58 +229,21 @@ function fixLength(req: ServerRequest): void {
...
 
 // ParseHTTPVersion parses a HTTP version string.
-// "HTTP/1.0" returns (1, 0, true).
-// Ported from https://github.com/golang/go/blob/f5c43b9/src/net/http/request.go#L766-L792
+// "HTTP/1.0" returns (1, 0).
 export function parseHTTPVersion(vers: string): [number, number] {
-  switch (vers) {
-    case "HTTP/1.1":
-      return [1, 1];
+  const httpReg = /^HTTP\/([0-9]{0,2})\.([0-9]{0,2})$/gm;
```
nayeemrmn: This isn't correct. The digit minimum should be 1.

You might know something I don't but also why is it capped at 2? 🤔

jmisiti42: You'r totally true ! I will use your previous answer !

jmisiti42: And i was capping at 2 because of the actual version wish cap at 1000000 wish was arbitrary
But 2 was even more arbitrary.

nayeemrmn:
Does anyone know what the significance of the golang's arbitrary upper bound `Big = 1000000` was? AFAICT not having it here just introduces the possibility of returning unsafe integers or `Infinity`, but I think passing this on to callers is fine.

zekth:
> Does anyone know what the significance of the golang's arbitrary upper bound `Big = 1000000` was? AFAICT not having it here just introduces the possibility of returning unsafe integers or `Infinity`, but I think passing this on to callers is fine.

There is not so much info about the reason of this number. IMO we should keep this guard untill we know the real reason of it.

nayeemrmn:
@jmisiti42 Can you run `deno fmt http/server.ts`? I ran it and the diff is just:
```diff
-  if (major > 1000000 || minor > 1000000) {
+  if (major > 1000000 || minor > 1000000) {
     throw new Error(`malformed HTTP version ${vers}`);
   }
```
Anyway it should work after that 🤷‍♂️

PR #599, opened 2019-09-16
Fix "deno test" on windows

### LINKED ISSUE #593: [testing] deno test does not work on windows
The `deno test` command does not work on windows. It always outputs `No matching test files found.` [Here](https://dev.azure.com/denoland/deno_std/_build/results?buildId=2230&view=logs&jobId=2d2b3007-3c5c-5840-9bb0-2b1ea49925f3&taskId=ddeebf02-4eb3-5b9c-39ee-cfd42d31a3cf&lineStart=17&lineEnd=18&colStart=1&colEnd=1) is an example from the `deno_std` CI. I think the file walker/matcher does not work on windows.

lucacasonato:
Fixes #593 

Use backslashes on windows for the default matcher globs.

Some tests don't run correctly on windows because they expect `\n` to result in a line break or `/` to be the file separator. These should probably be fixed in different PRs to keep this one simple.

### DIFF
--- a/.ci/check_source_file_changes.ts
+++ b/.ci/check_source_file_changes.ts
@@ -1,7 +1,6 @@
 // Copyright 2018-2019 the Deno authors. All rights reserved. MIT license.
 import { xrun } from "../prettier/util.ts";
 import { red, green } from "../fmt/colors.ts";
-import { EOL } from "../fs/path/constants.ts";
 
 /**
  * Checks whether any source file is changed since the given start time.
--- a/.ci/check_source_file_changes.ts
+++ b/.ci/check_source_file_changes.ts
@@ -12,7 +12,7 @@ async function main(startTime: number): Promise<void> {
   const changed = new TextDecoder()
     .decode(await xrun({ args: ["git", "ls-files"], stdout: "piped" }).output())
     .trim()
-    .split(EOL)
+    .split("\n")
     .filter(file => {
       const stat = Deno.lstatSync(file);
       if (stat != null) {
--- a/.ci/check_source_file_changes.ts
+++ b/.ci/check_source_file_changes.ts
@@ -1,6 +1,7 @@
 // Copyright 2018-2019 the Deno authors. All rights reserved. MIT license.
 import { xrun } from "../prettier/util.ts";
 import { red, green } from "../fmt/colors.ts";
+import { EOL } from "../fs/path/constants.ts";
 
 /**
  * Checks whether any source file is changed since the given start time.
@@ -11,7 +12,7 @@ async function main(startTime: number): Promise<void> {
   const changed = new TextDecoder()
     .decode(await xrun({ args: ["git", "ls-files"], stdout: "piped" }).output())
     .trim()
-    .split("\n")
+    .split(EOL)
     .filter(file => {
       const stat = Deno.lstatSync(file);
       if (stat != null) {
--- a/testing/runner.ts
+++ b/testing/runner.ts
@@ -5,12 +5,18 @@ import { glob, isGlob, walk } from "../fs/mod.ts";
 import { runTests } from "./mod.ts";
 const { args, cwd } = Deno;
 
-const DEFAULT_GLOBS = [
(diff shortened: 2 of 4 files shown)


bartlomieju:
It seems that the file matching is working fine after all ([example build](https://dev.azure.com/denoland/deno_std/_build/results?buildId=2273)). Maybe let's put off this change until we land updated test runner. Tests will fail if there are no matching files found.

lucacasonato:
@bartlomieju The test succeedes, but it doesn't actually test anything. The tests itself aren't actually run. Look at the windows part of that link you sent. It finishes after just under 2 minutes while both Linux and MacOS take significantly longer. If you look at where 'deno test' is actually we excuted you will see that it does not find any tests to run. That means the tests on windows are broken. This is caused by the globs that are designed for Unix not working for windows because it uses a different path delimiter. I think this should go in as it fixes the issue.

lucacasonato:
> It seems that the file matching is working fine after all ([example build](https://dev.azure.com/denoland/deno_std/_build/results?buildId=2273)). Maybe let's put off this change until we land updated test runner. Tests will fail if there are no matching files found.

Oh and tests do not fail if no files are found.

bartlomieju:
> @bartlomieju The test succeedes, but it doesn't actually test anything. The tests itself aren't actually run. Look at the windows part of that link you sent. It finishes after just under 2 minutes while both Linux and MacOS take significantly longer. If you look at where 'deno test' is actually we excuted you will see that it does not find any tests to run. That means the tests on windows are broken. This is caused by the globs that are designed for Unix not working for windows because it uses a different path delimiter. I think this should go in as it fixes the issue.

Ha! I could swear that I had different result when I pasted the link, dunno... Then the issue stands.



> Oh and tests do not fail if no files are found.

Yes, they don't right now, I forgot to tag the PR which will introduce this behavior: #604

testing/runner.ts
```diff
@@ -3,13 +3,14 @@
...
 
 const DEFAULT_GLOBS = [
-  "**/*_test.ts",
-  "**/*_test.js",
-  "**/test.ts",
-  "**/test.js"
+  join("**", "*_test.ts"),
+  join("**", "*_test.js"),
+  join("**", "test.ts"),
+  join("**", "test.ts")
```
nayeemrmn: Should be `join("**", "test.js")`

nayeemrmn:
@lucacasonato You'll have to fix the issues to get this green.

For `getMatchingUrlsLocal()` just use `join()` instead of `/` like you've done here.

For `xevalCliReplvar()` replace `\n` with EOL from `//fs/path/constants.ts`.

testing/runner_test.ts
```diff
@@ -22,7 +22,7 @@ test(async function getMatchingUrlsRemote(): Promise<void> {
 
 test(async function getMatchingUrlsLocal(): Promise<void> {
   const urls = await getMatchingUrls(
-    ["fmt/*_test.ts"],
+    [join("fmt", "*_test.ts")],
```
nayeemrmn: Doh! `fmt` is specified in `TEST_ROOT_PATH`, change this to just `["*_test.ts"],`.

There's a reason this works on unix and I've found lots of other issues. Let's just get this green and I'll see what I can improve in #604.

.ci/check_source_file_changes.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018-2019 the Deno authors. All rights reserved. MIT license.
 import { xrun } from "../prettier/util.ts";
 import { red, green } from "../fmt/colors.ts";
+import { EOL } from "../fs/path/constants.ts";
```
nayeemrmn: Unused import

nayeemrmn:
@lucacasonato On second thoughts, I've just applied these fixes in #604 (https://github.com/denoland/deno_std/pull/604/commits/91f1eafd1a5ac2baa35a2c4bab0bc24fb40c5d12).

It would probably be easier to close this 🙏

lucacasonato:
No Problem

PR #601, opened 2019-09-18
Add includeDirs to WalkOptions

nayeemrmn:
Towards API improvements in the test runner (touched upon here https://github.com/denoland/deno/issues/2948#issuecomment-531471956).
Also just a useful feature.

Fixes outdated documentation. cc @bartlomieju.

### DIFF
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -1,13 +1,14 @@
 // Documentation and interface for walk were adapted from Go
 // https://golang.org/pkg/path/filepath/#Walk
 // Copyright 2009 The Go Authors. All rights reserved. BSD license.
-const { readDir, readDirSync } = Deno;
+const { readDir, readDirSync, stat, statSync } = Deno;
 type FileInfo = Deno.FileInfo;
 import { unimplemented } from "../testing/asserts.ts";
 import { join } from "./path/mod.ts";
 
 export interface WalkOptions {
   maxDepth?: number;
+  includeDirs?: boolean;
   exts?: string[];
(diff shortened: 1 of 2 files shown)


bartlomieju:
So I guess with the `includeDirs` option you would want to use like this: `deno test src/`. Then check if walk found any files/dirs, if one of found entries is a directory you'd apply default globs again. Am I understanding it correctly?

nayeemrmn:
> So I guess with the `includeDirs` option you would want to use like this: `deno test src/`. Then check if walk found any files/dirs, if one of found entries is a directory you'd apply default globs again. Am I understanding it correctly?

@bartlomieju That's exactly correct! A nice consequence of this is that `deno test` could simply default to `deno test .` which would expand into `deno test <DEFAULT GLOBS ROOTED AT .>`, as it would with any matched directory. Another one is that `deno test src` would do the same as `cd src; deno test .`!

I haven't thought the implementation through too deeply, but I think it would be really inefficient using any existing tools (though not in the most common usages), even with this change. There's a good chance `walk` wouldn't be used for it at all at some point, but this'll allow for an MVP.

bartlomieju:
> I haven't thought the implementation through too deeply, but I think it would be really inefficient using any existing tools (though not in the most common usages), even with this change. There's a good chance `walk` wouldn't be used for it at all at some point, but this'll allow for an MVP.

It sounds fine for me, but we need to test it experimentally. Could you work on MVP?

nayeemrmn:
@ry Ready for review.

PR #604, opened 2019-09-18
Test runner v2

nayeemrmn:
Currently, a directory argument will match every file in that directory which would never be useful (and the way it's implemented is the cause of https://github.com/denoland/deno/issues/2948). We only get to benefit from the default globs when no arguments are given, so they have to be written out manually unless you want the exact set of files matched by just `deno test`. See https://github.com/denoland/deno/issues/2948#issuecomment-531471956.

This change will make it so that any directory given as - or expanded from - an argument is interpreted as a root on which to apply the default globs. Such an API would be more ergonomic and introduce some desirable properties. See https://github.com/denoland/deno_std/pull/601#issuecomment-532662357.

The test runner will fail by default if no tests are found. Override this with a flag.

Other API changes:

`//testing/runner.ts`:
Add `runTestModules()` and `RunTestModulesOptions`, rename `getMatchingUrls()` to `findTestModules()`.

`//fs/mod.ts`:
Rename `RunOptions` to `RunTestsOptions`.

`//fs/glob.ts`:
Add `expandGlob()`, `expandGlobSync()` and `ExpandGlobOptions`. These are prototypes extracted from the test runner.

`//fs/path/constants.ts`:
Add `SEP`.


bartlomieju:
@nayeemrmn can you update the tests to present the new behavior of match? Also, I'm still convinced we should fail test run if not matching files are found (surely, a flag to disable this behavior can be added, but the default should fail), so if you could add that option as well.

testing/runner.ts
```diff
@@ -12,26 +12,24 @@ const DEFAULT_GLOBS = [
...
 ARGS:
-  [FILES...]  List of file names to run. Defaults to: ${DEFAULT_GLOBS.join(
-    ","
-  )} 
-`);
+  [FILES...]  List of files to run.
+              A directory <dir> will expand to:
+                ${DEFAULT_GLOBS.map((s: string): string => `<dir>/${s}`).join(`
+                `)}
+              Defaults to "." when none are provided.`);
```
bartlomieju: s/"."/current directory

testing/runner.ts
```diff
@@ -12,26 +12,24 @@ const DEFAULT_GLOBS = [
...
+  -q, --quiet               Don't show output from test cases
+  -f, --failfast            Stop running tests on first error
+  -e, --exclude <FILES...>  List of comma-separated files to exclude from run
+
 ARGS:
-  [FILES...]  List of file names to run. Defaults to: ${DEFAULT_GLOBS.join(
-    ","
-  )} 
-`);
+  [FILES...]  List of files to run.
```
bartlomieju: Okay now I'm not sure, @nayeemrmn what will happen if I run test runner like that:
```
1. deno test src/
2. deno test src/my_test.ts
3. deno test foo/ bar/
4. deno test foo/test1.ts bar/test2.ts
5. deno test \*\*\/*_test.ts
```

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
 
   return matchingLocalUrls.concat(matchingRemoteUrls);
 }
+
+export interface RunTestModulesOptions extends RunOptions {
```
bartlomieju: Can you rename `RunOptions` to `RunTestsOptions`, it's the same name as options for `Deno.run`

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
...
+ *
+ *       runTestModules(["."]);
  */
-// TODO: change return type to `Promise<void>` once, `runTests` is updated
-// to return boolean instead of exiting
-export async function main(root: string = cwd()): Promise<void> {
+// TODO: Change return type to `Promise<void>` once, `runTests` is updated
+// to return boolean instead of exiting.
+export async function runTestModules(
+  include: string[],
```
bartlomieju: -1 for having `include` as first arg and the rest in options. I think it makes sense to put them all inside `options`:
```
runTestModules({include: ["."], exclude: ["dist/"]})
```

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
...
+    exclude = [],
+    parallel = false,
+    exitOnFail = false,
+    only = /[^\s]/,
+    skip = /^\s*$/,
+    disableLog = false
+  }: RunTestModulesOptions = {}
+): Promise<void> {
+  // TODO: Remove the concept of a root, support parent paths.
+  const root = cwd();
```
bartlomieju: `root` can be moved to `options` object as well

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
...
+// to return boolean instead of exiting.
+export async function runTestModules(
+  include: string[],
+  {
+    exclude = [],
+    parallel = false,
+    exitOnFail = false,
+    only = /[^\s]/,
+    skip = /^\s*$/,
+    disableLog = false
```
bartlomieju: Add `failIfNoMatchingFiles` option (or something along the lines)

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
...
+): Promise<void> {
+  // TODO: Remove the concept of a root, support parent paths.
+  const root = cwd();
+  const testModuleUrls = await getMatchingUrls(include, exclude, root);
+
+  if (testModuleUrls.length == 0) {
+    if (!disableLog) {
+      console.error("No matching test modules found.");
+    }
+    return;
```
bartlomieju: :+1:

testing/runner.ts
```diff
@@ -95,74 +93,110 @@ export async function getMatchingUrls(
...
+  if (!disableLog) {
+    console.log(`Found ${testModuleUrls.length} matching test modules.`);
+  }
+
+  for (const url of testModuleUrls) {
+    await import(url);
+  }
+
+  await runTests({
+    parallel,
```
bartlomieju: I think it makes sense to leave out `parallel` option for now. It doesn't work as expected know anyway. Tests are not run in parallel (by separate workers) but rather concurrently (all tests are as promises by single thread and await at the end).

nayeemrmn: This is just for completeness and consistency with `RunOptions`, the default value given to it at the top is the same as for `runTests()`. I'm just propagating it here since we're extending that.

testing/runner.ts
nayeemrmn: Outdated, but to be clear I think root should be gotten rid of. I've contained it to `getMatchingUrls()` and left that as a TODO.

testing/runner.ts
nayeemrmn: Yes, WIP.

testing/runner.ts
nayeemrmn: Hmm, I want to separate out the optional arguments into `RunTestModuleOptions`. This function doesn't make sense without includes, whereas the excludes wouldn't be used for most well structured source trees. Also it's analogous to the CLI.

testing/runner.ts
nayeemrmn: Not a problem, but I was purposefully trying to be verbatim here i.e. `deno test` => `deno test .`.  For me "current directory" would take an extra second to parse. Are you sure?

testing/runner.ts
nayeemrmn: 2, 4 and 5 would be the same as they are now,
1 would expand into `deno test src/**/test.ts src/**/*_test.ts` (also s/ts/js)
4 would expand into `deno test foo/**/test.ts foo/**/*_test.ts bar/**/test.ts bar/**/*_test.ts` (also s/ts/js ones)

It helps to execute the help message and see how that code evaluates.

testing/runner.ts
bartlomieju: I'm not sure... If I run `deno test` it should work like now: walk `root` and find all files matching default globs. So you should be able to run just `runTestModules()` without any arguments.

testing/runner.ts
nayeemrmn: 👍 That extends the scope a bit so I'll save it for the end.

testing/runner.ts
bartlomieju: Okay, that's good. Then I suggest to update help message to better reflect that (possibly adding presented cases as examples).

`[FILES...]  List of files to run.` suggest that only file paths can be given but that's not true.

nayeemrmn: > `[FILES...] List of files to run.` suggest that only file paths can be given but that's not true.

Yeah, that was bothering me. Forgot to change it with the first commit.

testing/runner.ts
nayeemrmn: Ah okay, you want to "." default to apply to the exported API and not just the CLI.

I think an optional array argument that default to a singleton array isn't a very intuitive design. You have `runTestModules()` and `runTestModules({ include: [] })` as two situations where the default could (possibly?) be applied... it just complicates the idea of a default. The API ends up being kind of hard to reason about and the implementation would be ugly as well.

testing/runner.ts
nayeemrmn: I'll go with "List of test modules to run" like I have been in the code. "module" captures files, URLs, and directories according to Deno's nomenclature.

testing/runner.ts
bartlomieju: > I think an optional array argument that default to a singleton array isn't a very intuitive design. You have `runTestModules()` and `runTestModules({ include: [] })` as two situations where the default could (possibly?) be applied... it just complicates the idea of a default. The API ends up being kind of hard to reason about and the implementation would be ugly as well.

I'm not sure I follow...

If you defined `runTestModules` as such:
```
function runTestModules(options: {
  include?: string[],
  exclude?: string[]
} = {}) {
...
options = {...DEFAULT_OPTIONS, ...options};
}
```
Then the API shouldn't be that hard.

bartlomieju: Ah! Now I see your point.

Then maybe we should split options for globs and directories to find?

If we keep options like now, then we'd have something alike:
```ts
// I'd expect to find all files matching default globs inside `src/` directory relative to current working directory.
runTestModules({ root: "src/" });

// I'd expect to find all files matching provided globs relative to current working directory.
runTestModules({ include: ["**/*.test.ts", "**/*.test.js"] });

// I'd expect to run only that single file without any walking of directory tree.
runTestModules({ include: ["https://deno.land/std/http/file_server_test.ts"] });

// I'd expect it would find files matching default globs 
runTestModules({ include: [] })
```

But then `include` (and `exclude` as well) would sometimes be used to specify globs, sometimes directory path (still applying globs) and sometimes remote files.

🤔

bartlomieju: It's high time to check how other runners (Ava, Mocha, Jest) handle that. The problem is AFAIK neither of them can run remote files.

nayeemrmn: Yes, glob expansion is done post-partition. They won't matter for this change, hopefully. I think URLs are only supported here for consistency with Deno's definition of a module.

nayeemrmn:
@bartlomieju @lucacasonato
Well, I introduced it. Didn't realise CI was using the live version of the test runner 😮 (Probably from when the runner was a 'tool' here and not part of the library. It should just be using `deno test` now IMO... that's for another issue). Now this is red obviously.

I suggest landing #599 first.

testing/runner.ts
nayeemrmn: @bartlomieju 😅 I've come around to your original suggestion.

My thought is that this function is not really a good logical unit and will pretty much only be used as an exact stand-in for the CLI version. It should be as analogous to the CLI as possible since that's the only value it has. Ignore my rambling above :p

nayeemrmn:
@bartlomieju I think this is ready. Could I get another review?

cc @ry

sholladay:
Is there a specific reason for having `--allow-none`? What's the use case? Seems crufty to me.

nayeemrmn:
> Is there a specific reason for having `--allow-none`? What's the use case? Seems crufty to me.

@sholladay When the number of tests is dynamic somehow, or someone wants their infrastructure up before they have any tests? Maybe that's contrived 😅

While it makes sense practically, I just think that failing when no tests are run is 'illogical' enough that there should be a way of at least overriding it.

testing/runner.ts
```diff
@@ -58,111 +55,180 @@ function partition(
...
+          return fileGlob.split(",");
+        })
+      : ["."];
+  const exclude =
+    parsedArgs.exclude != null ? (parsedArgs.exclude as string).split(",") : [];
+  const allowNone = parsedArgs["allow-none"];
+  const exitOnFail = parsedArgs.failfast;
+  const disableLog = parsedArgs.quiet;
+
+  await runTestModules({
```
bartlomieju: why not `try/catch`?

fs/glob.ts
```diff
@@ -76,3 +79,61 @@ export function isGlob(str: string): boolean {
...
+}
+
+/**
+ * Expand the glob string from the specified `root` directory and yield each
+ * result as a `WalkInfo` object.
+ */
+// TODO: Use a proper glob expansion algorithm.
+// This is a very incomplete solution. The whole directory tree from `root` is
+// walked and parent paths are not supported.
+export async function* expandGlob(
```
bartlomieju: These methods need unit tests

testing/runner.ts
```diff
@@ -1,44 +1,41 @@
...
-  [FILES...]  List of file names to run. Defaults to: ${DEFAULT_GLOBS.join(
-    ","
-  )} 
-`);
-}
-/* eslint-enable max-len */
+  -q, --quiet                 Don't show output from test cases
+  -f, --failfast              Stop running tests on first error
+  -e, --exclude <MODULES...>  List of comma-separated modules to exclude
+  --allow-none                Exit with status 0 even when no test modules are
```
bartlomieju: :+1: IMHO this is better for CI experience. Given our own incident where there were no test files found on Windows I believe that failing in that situation is a safe default.

testing/runner.ts
```diff
@@ -58,111 +55,180 @@ function partition(
...
- * list of file URLs that should be imported for test runner.
+ * Given a list of globs or URLs to include and exclude and a root directory
+ * from which to expand relative globs, return a list of URLs
+ * (file: or remote) that should be imported for the test runner.
  */
-export async function getMatchingUrls(
-  matchPaths: string[],
-  excludePaths: string[],
-  root: string
+export async function findTestModules(
```
bartlomieju: Nice :+1: I need more time to go through it in detail, but I suggest writing some more unit tests for this method

testing/runner.ts
nayeemrmn: Done 👍

testing/runner.ts
nayeemrmn: Done. The test module looks a lot better: https://github.com/denoland/deno_std/blob/09e47c704a64b09be10d1681aef2dfd02ebf3392/testing/runner_test.ts.

fs/glob.ts
nayeemrmn: Done (finally!)

Is this ready now?

bartlomieju:
@ry PTAL

xeval/test.ts
```diff
@@ -33,7 +34,7 @@ test(async function xevalCliReplvar(): Promise<void> {
   await p.stdin!.write(encode("hello"));
   await p.stdin!.close();
   assertEquals(await p.status(), { code: 0, success: true });
-  assertEquals(decode(await p.output()), "hello\n");
+  assertEquals(decode(await p.output()), `hello${EOL}`);
```
ry: How about using trim rather than importing EOL?

testing/runner_test.ts
```diff
@@ -1,30 +1,77 @@
...
 
-/**
- * IMPORTANT: This file assumes it is run from root of repository.
- */
-const cwd = Deno.cwd();
-const TEST_ROOT_PATH = join(cwd, "fmt");
+const TEST_DATA_URL = new URL("testdata", import.meta.url);
+const TEST_DATA_PATH = TEST_DATA_URL.pathname
+  .slice(isWindows ? 1 : 0)
+  .replace(/\//g, SEP);
```
ry: What's happening here? This looks brittle. Ideally this operation be abstracted into a function with a comment.

ry:
Look like a nice clean up. Just a couple comments...

I'll land when @bartlomieju approves.

xeval/test.ts
nayeemrmn: 👍

testing/runner_test.ts
nayeemrmn: The `slice()` is because `new URL('file:///C:/a').pathname` is `/C:/a`, the leading slash is bad. The backslashes are necessary I think because of `globrex()`'s implementation... but they shouldn't be. I'll tinker.

nayeemrmn: 👍

nayeemrmn:
@bartlomieju Ready.

testing/runner.ts
```diff
@@ -1,44 +1,42 @@
...
+ARGS:
+  [MODULES...]  List of test modules to run.
+                A directory <dir> will expand to:
+                  ${DIR_GLOBS.map((s: string): string => `${join("<dir>", s)}`)
+                    .join(`
+                  `)}
+                Defaults to "." when none are provided.
+
+Note that modules can refer to file paths or URLs and globs are supported for
+the former.`);
```
bartlomieju:
```
Deno test runner

USAGE:
  deno -A https://deno.land/std/testing/runner.ts [OPTIONS] [MODULES...]

OPTIONS:
  -q, --quiet                 Don't show output from test cases
  -f, --failfast              Stop running tests on first error
  -e, --exclude <MODULES...>  List of comma-separated modules to exclude
  --allow-none                Exit with status 0 even when no test modules are
                              found

ARGS:
  [MODULES...]  List of test modules to run.
                A directory <dir> will expand to:
                  <dir>/**/?(*_)test.{js,ts}
                Defaults to "." when none are provided.

Note that modules can refer to file paths or URLs and globs are supported for
the former.
```

Nice :+1: 

Maybe the last sentence could be changed to `Note that modules can refer to file paths or URLs. File paths support glob expansion.`

bartlomieju: Also I wouldn't mind if there were 1-2 examples at the end of help message

nayeemrmn: 👍

PR #607, opened 2019-09-21
Implement readString

### LINKED ISSUE #606: Implement `BufReader.readString(delim)`
Noticed that this is [not yet implemented](https://github.com/denoland/deno_std/blob/a8f6cf7b4fa4030878bb702a65f28f469ee1f87d/io/bufio.ts#L215). Should be a very useful utility (also since both `readSlice` and `readLine` are implemented but in the documentation pointing to prefer using `readString`).

Also may be a good task for new contributors

sh7dm:
Closes #606

### DIFF
--- a/io/bufio.ts
+++ b/io/bufio.ts
@@ -213,7 +213,7 @@ export class BufReader implements Reader {
    */
   async readString(delim: string): Promise<string | Deno.EOF> {
     const buffer = await this.readSlice(delim.charCodeAt(0));
-    return new TextDecoder().decode(buffer);
+    return new TextDecoder().decode(buffer || undefined);
   }
 
   /** `readLine()` is a low-level line-reading primitive. Most callers should
--- a/io/bufio.ts
+++ b/io/bufio.ts
@@ -212,6 +212,8 @@ export class BufReader implements Reader {
    * For simple uses, a Scanner may be more convenient.
    */
   async readString(delim: string): Promise<string | Deno.EOF> {
+    if (delim.length !== 1)
+      throw new Error("Delimiter should be a single character");
     const buffer = await this.readSlice(delim.charCodeAt(0));
     return new TextDecoder().decode(buffer || undefined);
   }
--- a/io/bufio.ts
+++ b/io/bufio.ts
@@ -212,7 +212,8 @@ export class BufReader implements Reader {
    * For simple uses, a Scanner may be more convenient.
    */
   async readString(_delim: string): Promise<string | Deno.EOF> {
-    throw new Error("Not implemented");
+    const buffer = await this.readSlice(_delim.charCodeAt(0));
+    return new TextDecoder().decode(buffer);
   }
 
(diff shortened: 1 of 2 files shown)


io/bufio.ts
```diff
@@ -212,7 +212,8 @@ export class BufReader implements Reader {
    * For simple uses, a Scanner may be more convenient.
    */
   async readString(_delim: string): Promise<string | Deno.EOF> {
```
kitsonk: The leading `_` should be removed from this now... it was a signal to allow unused arguments in the signature.

sh7dm: @kitsonk done, thanks.

io/bufio.ts
```diff
@@ -211,8 +211,9 @@ export class BufReader implements Reader {
    * delim.
    * For simple uses, a Scanner may be more convenient.
    */
-  async readString(_delim: string): Promise<string | Deno.EOF> {
-    throw new Error("Not implemented");
+  async readString(delim: string): Promise<string | Deno.EOF> {
+    const buffer = await this.readSlice(delim.charCodeAt(0));
```
ry: Maybe throw if delim.length != 1

kevinkassimo:
Hmm I am also thinking if we could provide a version that possibly allow multi-char delimiter?

zekth:
> Hmm I am also thinking if we should also provide a version that possibly allow multi-char delimiter?

IMO we need multichar delimiter

kevinkassimo:
> > Hmm I am also thinking if we should also provide a version that possibly allow multi-char delimiter?
> 
> IMO we need multichar delimiter

On second thought I noticed that since BufReader is implemented in a way strictly compliant with Go’s bufio counterparts, maybe single char restriction is fine here. I do hope though if we can provide another utility for unlimited buffered reader that automatically resize/reallocate the internal buffer which would allow implementation of multi-char delim easier (maybe an extension of `deno.Buffer`. We have something similar for xeval that recently got moved here)

io/bufio.ts
sh7dm: Keeping in mind comments :point_down:, I'll check if `delim.lenght > 0`

io/bufio.ts
```diff
@@ -211,8 +211,11 @@ export class BufReader implements Reader {
    * delim.
    * For simple uses, a Scanner may be more convenient.
    */
-  async readString(_delim: string): Promise<string | Deno.EOF> {
-    throw new Error("Not implemented");
+  async readString(delim: string): Promise<string | Deno.EOF> {
+    if (delim.length !== 1)
+      throw new Error("Delimiter should be a single character");
+    const buffer = await this.readSlice(delim.charCodeAt(0));
+    return new TextDecoder().decode(buffer || undefined);
```
sh7dm: should we use `strings` module here?

ry: No this is fine

Issue #610, opened 2019-09-23
Support generics in fs/read_json.ts


wkirk01:
It would be nice to support generic types in [fs/read_json.ts](https://github.com/denoland/deno_std/blob/master/fs/read_json.ts).

```ts
import { SomeType } from "./interfaces/SomeType.ts";
import { readJson, readJsonSync } from "https://deno.land/std/fs/mod.ts";

const f = await readJson<SomeType>("./foo.json");
const foo = readJsonSync<SomeType>("./foo.json");
```

kitsonk:
When you don't use the generic in the arguments or implementation, the generic is meaningless.  Look at the type definitions in TypeScript for [`JSON.parse`](https://github.com/microsoft/TypeScript/blob/b963e1a2a7e3b1056ee2552927fa08fcc51e4c7d/lib/lib.es5.d.ts#L1059).  It doesn't support a generic because they are unneeded.  And for historic reasons, it returns `any` versus what is more appropriate which is `unknown`.  This does not restrict assigning the value to something that is more narrowly type.

PR #611, opened 2019-09-23
Add generics to fs/read_json

wkirk01:
This is in response to issue #610


kitsonk:
As mentioned in issue #610, I don't believe we should land this.  Generics where only effect the return type are meaningless and give people a false impression of how the type system works.  The corollary is how `JSON.parse()` is typed, which uses `any` as a return type, we predated the introduction of `unknown` which is the correct type for this function.

wkirk01:
@kitsonk That makes sense. Closing this PR.

PR #612, opened 2019-09-23
fix: [toml] adapted to TOML's key encoding spec

keroxp:
- Dumper is emitting invalid TOML format.
- Table names or keys should be wrapped with `""` if title contains characters except `A-Za-z0-9_-`


keroxp:
- Parser should also be fixed. Quoted titles like `["https://deno.land"]` can't be parsed properly.

PR #615, opened 2019-09-28
feat: [ws] wss support with dialTLS

keroxp:
- feat: `wss` support with dialTLS
- fixed: Default port check by url.protocol was not working

### DIFF
--- a/ws/mod.ts
+++ b/ws/mod.ts
@@ -480,15 +480,17 @@ export async function connectWebSocket(
   headers: Headers = new Headers()
 ): Promise<WebSocket> {
   const url = new URL(endpoint);
-  let { hostname, port } = url;
-  if (!port) {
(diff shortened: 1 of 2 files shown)


ws/mod.ts
```diff
@@ -480,15 +480,17 @@ export async function connectWebSocket(
...
-      throw new Error("currently https/wss is not supported");
-    }
+  let { hostname } = url;
+  let conn: Conn;
+  if (url.protocol === "http:" || url.protocol === "ws:") {
+    const port = parseInt(url.port || "80");
+    conn = await Deno.dial({ hostname, port });
+  } else if (url.protocol === "https:" || url.protocol === "wss:") {
+    const port = parseInt(url.port || "443");
+    conn = await Deno.dialTLS({ hostname, port });
```
ry: Probably we should make it so dialTLS also accepts port as a string...

PR #617, opened 2019-09-28
Implement expandGlob()

### LINKED ISSUE #576: Test runner file matcher should walk as little as possible
If I run `deno -A https://deno.land/std/testing/runner.ts projects/deno_std/fmt/sprintf_test.ts`, it walks the whole directory tree from CWD and takes a really long time. Obviously no walking should be done in this case - sound easy enough.

Ideally though, `deno test projects/deno_std/fmt/*_test.ts` would only walk `projects/deno_std/fmt`. At that point you'd want to extract a magical glob expansion utility in `//fs`.

nayeemrmn:
...properly. Fixes #576.

This had to be done from scratch unless we could utilise `globrex` somehow but I think it can be done by splitting paths and flat-mapping on each segment, applying `globrex` on individual segments. I also want to give it an `exclude` option so excluded directories won't be traversed - callers all want to exclude by path prefix anyway, this is currently done with a hacky `g` flag.

### DIFF
--- a/fs/glob.ts
+++ b/fs/glob.ts
@@ -94,12 +94,13 @@ interface SplitPath {
 
 // TODO: Maybe make this public somewhere if it can be fixed for Windows.
 function split(path: string): SplitPath {
+  const s = SEP_PATTERN.source;
   return {
     segments: path
-      .replace(new RegExp(`^${SEP_PATTERN.source}|${SEP_PATTERN.source}$`), "")
+      .replace(new RegExp(`^${s}|${s}$`, "g"), "")
       .split(SEP_PATTERN),
     isAbsolute: isAbsolute(path),
-    hasTrailingSep: !!path.match(new RegExp(`${SEP_PATTERN.source}$`))
+    hasTrailingSep: !!path.match(new RegExp(`${s}$`))
   };
 }
 
--- a/fs/glob.ts
+++ b/fs/glob.ts
@@ -4,18 +4,12 @@ import { WalkInfo, walk, walkSync } from "./walk.ts";
 const { cwd } = Deno;
 
 export interface GlobOptions {
-  // Allow ExtGlob features
   extended?: boolean;
-  // When globstar is true, '/foo/**' is equivelant
-  // to '/foo/*' when globstar is false.
-  // Having globstar set to true is the same usage as
-  // using wildcards in bash
   globstar?: boolean;
-  // be laissez faire about mutiple slashes
   strict?: boolean;
-  // Parse as filepath for extra path related features
-  filepath?: boolean;
-  // Flag to use in the generated RegExp
+}
+
+export interface GlobToRegExpOptions extends GlobOptions {
   flags?: string;
 }
 
@@ -27,12 +21,12 @@ export interface GlobOptions {
  *
  *     Looking for all the `ts` files:
  *     walkSync(".", {
- *       match: [glob("*.ts")]
+ *       match: [globToRegExp("*.ts")]
  *     })
  *
  *     Looking for all the `.json` files in any subfolder:
  *     walkSync(".", {
- *       match: [glob(join("a", "**", "*.json"),{
+ *       match: [globToRegExp(join("a", "**", "*.json"),{
  *         flags: "g",
  *         extended: true,
  *         globstar: true
@@ -43,12 +37,12 @@ export interface GlobOptions {
  * @param options - Specific options for the glob pattern
  * @returns A RegExp for the glob pattern
  */
-export function glob(glob: string, options: GlobOptions = {}): RegExp {
-  const result = globrex(glob, options);
-  if (options.filepath) {
-    return result.path!.regex;
-  }
-  return result.regex;
+export function globToRegExp(
+  glob: string,
+  options: GlobToRegExpOptions = {}
+): RegExp {
+  const result = globrex(glob, { ...options, filepath: true });
+  return result.path!.regex;
 }
 
 /** Test whether the given string is a glob */
@@ -103,17 +97,14 @@ export async function* expandGlob(
     includeDirs = true,
     extended = false,
     globstar = false,
-    strict = false,
-    filepath = true,
-    flags = ""
+    strict = false
   }: ExpandGlobOptions = {}
 ): AsyncIterableIterator<WalkInfo> {
   const absoluteGlob = isAbsolute(globString)
     ? globString
     : join(root, globString);
-  const globOptions = { extended, globstar, strict, filepath, flags };
   yield* walk(root, {
-    match: [glob(absoluteGlob, globOptions)],
+    match: [globToRegExp(absoluteGlob, { extended, globstar, strict })],
     includeDirs
   });
 }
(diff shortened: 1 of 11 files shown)


nayeemrmn:
@ry @bartlomieju Ready for review!

I've checked that `expandGlobSync()` on any direct path with a `root` provided uses just one op! This affects the test runner and prettier. File-system walking is more or less minimised.

fs/walk.ts
```diff
@@ -97,27 +105,31 @@ export async function* walk(
...
-        yield* walk(filename, options);
-      }
+      yield* walk(filename, { ...options, maxDepth: maxDepth - 1 });
     }
   }
 }
 
 /** Same as walk() but uses synchronous ops */
 export function* walkSync(
-  root: string = ".",
```
zekth: why?

nayeemrmn: `walk()` didn't have this so something was wrong, I chose to change this one for no particular reason. I guess they could both be changed to have "." as a default but IMO root would have to go in the options object then so this was just the smaller change.

ry:
@kt3k a review from you would be appreciated if you have the time.

fs/glob.ts
```diff
@@ -86,9 +82,29 @@ export function isGlob(str: string): boolean {
 
 export interface ExpandGlobOptions extends GlobOptions {
   root?: string;
+  exclude?: string[];
   includeDirs?: boolean;
 }
 
+const SEP_PATTERN = isWindows ? /[\\/]+/ : /\/+/;
```
kt3k: Maybe we should have this constant in "./path/constants.ts"?

fs/glob.ts
```diff
@@ -100,22 +116,93 @@ export async function* expandGlob(
...
-    includeDirs
-  });
+  const resolveFromRoot = (path: string): string =>
+    isAbsolute(path) ? normalize(path) : join(root, path);
+  const globOptions: GlobOptions = { extended, globstar, strict };
+  const excludePatterns = exclude
+    .map(resolveFromRoot)
+    .map((s: string): RegExp => globToRegExp(s, globOptions));
+  const shouldInclude = ({ filename }: WalkInfo): boolean =>
+    !excludePatterns.some((p: RegExp): boolean => !!filename.match(p));
```
kt3k: If you're not interested in matched object, you can use regexp.test() method.

`!!filename.match(p)` => `p.test(filename)`

fs/glob.ts
```diff
@@ -86,9 +82,29 @@ export function isGlob(str: string): boolean {
...
+  segments: string[];
+  isAbsolute: boolean;
+  hasTrailingSep: boolean;
+}
+
+// TODO: Maybe make this public somewhere if it can be fixed for Windows.
+function split(path: string): SplitPath {
+  return {
+    segments: path
+      .replace(new RegExp(`^${SEP_PATTERN.source}|${SEP_PATTERN.source}$`), "")
```
kt3k: if you want to replace both leading and trailing separators, you probably need "g" modifier in regexp.

```
> "/aaa/".replace(/^\/|\/$/, '')
aaa/
> "/aaa/".replace(/^\/|\/$/g, '')
aaa
```

fs/glob.ts
```diff
@@ -86,9 +82,29 @@ export function isGlob(str: string): boolean {
...
   root?: string;
+  exclude?: string[];
   includeDirs?: boolean;
 }
 
+const SEP_PATTERN = isWindows ? /[\\/]+/ : /\/+/;
+
+interface SplitPath {
+  segments: string[];
+  isAbsolute: boolean;
```
kt3k: Is this property used somewhere?

fs/glob.ts
nayeemrmn: `String::match()` doesn't modify `RegExp::lastIndex`. Honestly it doesn't matter here because the exclude patterns don't have `g` but I think this'll be more stable.

fs/glob.ts
nayeemrmn: No... the paths I happen to give `split()` are necessarily absolute. I just wanted this to be a lossless path representation.

fs/glob.ts
nayeemrmn: 👍

ry:
@nayeemrmn In your description, it says "DON'T MERGE". Please update if it is no longer applicable.

fs/glob.ts
nayeemrmn: 👍

fs/glob.ts
```diff
@@ -97,47 +159,187 @@ export interface ExpandGlobOptions extends GlobOptions {
...
+  let fixedRoot = winRoot != undefined ? winRoot : "/";
+  while (segments.length > 0 && !isGlob(segments[0])) {
+    fixedRoot = join(fixedRoot, segments.shift()!);
+  }
+
+  let fixedRootInfo: WalkInfo;
+  try {
+    fixedRootInfo = { filename: fixedRoot, info: await stat(fixedRoot) };
+  } catch {
+    return;
```
kt3k: Is this swallowing of the error intentional? I think it's better to throw here.

nayeemrmn: It's intentional. If the fixed root doesn't exist we simply know in advance that there are no matches. I think that's the expected behaviour of glob expansion.

nayeemrmn:
@ry Fixed! Ready again.

kt3k:
//fs/README has `glob` entry. I think we need to update that to globToRegExp as well.

fs/glob.ts
```diff
@@ -84,11 +79,78 @@ export function isGlob(str: string): boolean {
...
+export function joinGlobs(
+  globs: string[],
+  { extended = false, globstar = false }: GlobOptions = {}
+): string {
+  if (!globstar || globs.length == 0) {
+    return join(...globs);
+  }
+  if (globs.length === 0) return ".";
+  let joined: string | undefined;
+  for (let i = 0, len = globs.length; i < len; ++i) {
```
kt3k: how about using for-of loop like `for (const path of globs) {`?

fs/walk.ts
```diff
@@ -55,6 +56,7 @@ export interface WalkInfo {
  *
  * Options:
  * - maxDepth?: number;
+ * - includeFiles?: boolean;
```
kt3k: I thinks it's nice to have test cases for `includeFiles: false`.

fs/glob.ts
nayeemrmn: 👍 (This was copy-pasted from `//fs/path/posix.ts`)

fs/walk.ts
nayeemrmn: 👍

fs/walk_test.ts
```diff
@@ -131,6 +131,23 @@ testWalk(
   }
 );
 
+testWalk(
+  async (d: string): Promise<void> => {
+    await touch(d + "/a");
+    await mkdir(d + "/b");
+    await touch(d + "/b/c");
+  },
+  async function includeFiles(): Promise<void> {
```
kt3k: 👍

ry:
@nayeemrmn Can you draft a commit message so I can squash this all up?

nayeemrmn:
@ry Added an empty commit with the message.

PR #624, opened 2019-10-06
Move fs/path to the top-level

### LINKED ISSUE #623: fs export structure
To access everything in `fs`, you have to import from:
- `fs/mod.ts`
- `fs/path.ts` (`fs/path/mod.ts` + `fs/path/interface.ts`)
- `fs/path/constants.ts`

These are all used from outside of `fs`.

`fs/path/mod.ts` should re-export everything public in `fs/path/constants.ts`.

`fs/mod.ts` should also have to re-export `fs/path.ts`, but I favour making `fs/path` top-level.

That would leave:
- `fs/mod.ts`
- `path/mod.ts`

Some exports should be moved from `path` to `fs`, namely `EOL` and `isWindows`... if dependency cycles are okay.

nayeemrmn:
Closes #623.

- Move `fs/path` to the top-level.
- Remove `fs/path.ts`.
- Extract `fs/expand_glob.ts` from `path/glob.ts`.
- Move `EOL` and `isWindows` from `path/constants.ts` to `fs/constants.ts`, rename `EOL`  to `EOL_` (name conflict).


ry:
Aren't paths only involved with fs stuff? It seems reasonable to me that it be a submodule of fs.

nayeemrmn:
> Aren't paths only involved with fs stuff? It seems reasonable to me that it be a submodule of fs.

In terms of the functionality provided by `deno_std`? I'd consider them parallels above Deno core.

It's mostly the export structure. As a submodule of `fs`, it should be exposed through `fs/mod.ts` like any other. But at that point the names become insufficient: `import { normalize } from "./fs/mod.ts"`. There's just too much in `fs` considering a meaningful chunk of it is strictly path stuff.

PR #625, opened 2019-10-07
fix: Update echo_server to new listen API

bartlomieju:
Fixes denoland/deno#3083


examples/echo_server.ts
```diff
@@ -2,8 +2,9 @@
 const { listen, copy } = Deno;
 
 (async (): Promise<void> => {
```
ry: I think you can remove this anonymous function too

bartlomieju: Done

PR #681, opened 2021-02-01
build: install Deno using bash installer

lucacasonato:


bartlomieju (approved):

.github/workflows/ci.yml
```diff
@@ -23,10 +23,21 @@ jobs:
...
+        if: |
+          !startsWith(matrix.os, 'windows')
+        run: |-
+          curl -fsSL https://deno.land/x/install/install.sh | sh
+          echo "$HOME/.deno/bin" >> $GITHUB_PATH
+      - name: Install Deno (Windows)
+        if: startsWith(matrix.os, 'windows')
+        run: |-
+          curl -fsSL https://deno.land/x/install/install.sh | sh
+          echo "$HOME/.deno/bin" >> $env:GITHUB_PATH
```
caspervonb, line 36: Can't we just set `bash` as the default shell globally instead of doing all this branching?

lucacasonato: Doesn't work. We tried on denoland/deno repo already.

PR #685, opened 2021-02-01
Migrating path.resolve to fs.resolvePath, path.relative to fs.relativePath and adding a modified version of path.resolve

whosmayank:
This commit introduces three major changes 

1. Migrates path.resolve to fs.resolvePath
path.resolve() uses Deno.cwd() to add cwd as first parameter so that the resolved path is absolute. However, apis under std/path shouldn't be using Deno.cwd() implicitly as that would require read access to cwd.
A lot of code which depended on it has been modified to suit the change.

2. Migrating path.relative to fs.relativePath #8559
Same reason as 1

3. Add path.resolve() which does not use Deno.cwd() implicitly to std/path
1 removes path.resolve() from std/path. It'd be worth having path.resolve() as a counterpart to fs.resolvePath which mandates returning an absolute path


http/testdata/tls
```diff
@@ -0,0 +1 @@
+../../../cli/tests/tls
```
bartlomieju, line 1: This should be reverted; alongside whole `http/testdata/tls` directory

bartlomieju (changes requested):

http/testdata/tls
whosmayank: Fixed it. Copy-paste never yielded good results haha. Thanks for pointing it out :)

ry:
Hi @mayankagarwals, Thanks for the patch. I think we're generally in agreement that we want to have a resolve function that doesn't depend on Deno.cwd. How exactly that should be done is up for debate still. I'm going to close this patch because it's quite out of sync with main. I'm sorry we didn't get fasters reviews for this back in February.  I've opened https://github.com/denoland/deno_std/issues/1136 to discuss further but I'm closing this PR.

PR #689, opened 2021-02-01
build: remove matrix checks for lint job

caspervonb:
This removes some dead branches in the lint job introduced in #681.
Lint doesn't actually run on matrix.os, only ubuntu-latest.


caspervonb:
Unless you intended to run lint/fmt on the test matrix @lucacasonato?

ry (approved):
LGTM - thanks

PR #690, opened 2021-02-02
feat(http/file_server): support do not show dotfiles

zhmushan:
default:

with `--no-dotfiles`:


ry (approved):
LGTM - thanks @zhmushan

PR #691, opened 2021-02-02
feat(http/file_server): show .. if it makes sense & make dir end with /

zhmushan:
- If the current dir is not root, show "../".
- Make dir end with "/"
- `Deno.stat()` no longer needs to catch errors. (The "mode" prop may not exist before, see https://github.com/denoland/deno_std/pull/15/files#diff-106086daad2313990285a9c61085eeb5995d479127db151e613f84ea709450c8R118


lucacasonato (approved):
LGTM. Thanks.

Issue #666, opened 2021-02-01
std/node: add net module


EvanHahn:
For completeness, we should add a polyfill for [Node's net module](https://nodejs.org/api/net.html).

pavel-kolesnikov:
The `net` module is needed for `rethinkdb` clients (all of them, for reference https://github.com/rethinkdb/rethinkdb-ts/issues/89).

Current implementations are tied to `net` module; `npm` to ES module providers like `jspm`, `esm.sh`, etc can not or not interested to support Deno specifics.

cmorten:
Started up on this - wip available at https://github.com/cmorten/deno_std/pull/2. Shout if already working on it and will avoid duplicating effort.

Started from scratch, but using https://github.com/denoland/deno_std/pull/694 to inform implementation / types as I go.

bartlomieju:
@cmorten I'm pretty sure no one is working on it, glad that you picked it up - it's will be a major achievement to get compatiblity with `net` module.

ebebbington:
> Started up on this - wip available at [cmorten#2](https://github.com/cmorten/deno_std/pull/2). Shout if already working on it and will stop to avoid duplicating effort!
> 
> Started from scratch, but using #694 to inform implementation / types as I go.

Nice!:) I pretty much just got as far as porting the us code for Socket and Server to TS I believe, only primordial usage and tests were left to address

jcs224:
Don't see it linked here but

![image](https://user-images.githubusercontent.com/6226188/136901333-98eefa8e-9fa6-4b69-b692-3f55ceda3e49.png)

https://github.com/denoland/deno_std/pull/1375

PR #694, opened 2021-02-02
[WIP] feat: net.Socket

ebebbington:
Partially addresses https://github.com/denoland/deno_std/issues/666

Moved from `https://github.com/denoland/deno`, see [here](https://github.com/denoland/deno/pull/8949) for original PR

- [ ] Tests
- [x] All code required copied over
- [ ] Port bindings
- [x] Port requires
- [ ] Port primordials
- [ ] Port JS to TS (largely relies on bindings)


ebebbington:
Regarding the latest comments in the original PR, myself and @bartlomieju discussed a few bits on discord:

image
image
image
image

ebebbington:
@bartlomieju How would i go about "emulating" some of the internal bindings? Is it just a matter of finding the source code and converting it from c++ into typescript?

bartlomieju:
> @bartlomieju How would i go about "emulating" some of the internal bindings? Is it just a matter of finding the source code and converting it from c++ into typescript?

@ebebbington yes, pretty much

ebebbington:
I believe there's only 1 `require` statement i need to address, and the rest is porting bindings and adding types

ebebbington:
Stopping work on this PR for the time being, until either bindings are implemented (im unable to do them unfortunately) or when https://github.com/denoland/deno_std/pull/695 is merged (then ill get the tests sorted)

jopemachine:
As far as I know, Deno uses tokio instead of libuv to handle Async I/O. 

Is it possible to emulate these internal bindings (c++ code like `tcp_wrap.cc`) without libuv?

Should it be implemented through tokio?

ry:
@ebebbington Thanks for the work on this patch - I'm going to close it tho because it's stale and not close to being mergeable. We very much would like to support std/node/net so hopefully we can come back and reference this patch when someone attempts to undertake it.

ebebbington:
> @ebebbington Thanks for the work on this patch - I'm going to close it tho because it's stale and not close to being mergeable. We very much would like to support std/node/net so hopefully we can come back and reference this patch when someone attempts to undertake it.

Fair enough :) unfortunately I probably won’t have enough time in the future to carry on with it so best of luck to whoever may wish to take it on in the future!

PR #695, opened 2021-02-02
fix(node): Enable native node tests

Soremwar:


Soremwar:
This will implement https://github.com/Soremwar/deno_node_tests

This is gonna be a big refactor, many tests won't be enabled as of now due to possible incompatibilities between the node lib and /std/node but luckily I designed the test script to be very configurable allowing to us to choose what tests to bring in

ry:
Sorry I'm a bit confused about the purpose of this. Can you explain more what you had in mind?

Soremwar:
This should allow us to test std/node with the actual test suite the node folks use in their repo. This should help catch any implementation bugs or lack of features that the current library might have (I know for a fact this kind of inconsistencies are present in the Buffer implementation)

node/module_test.ts
```diff
@@ -15,7 +15,7 @@ const require = createRequire(import.meta.url);
 
 Deno.test("requireSuccess", function () {
   // Relative to import.meta.url
-  const result = require("./tests/cjs/cjs_a.js");
+  const result = require("./_module/cjs/cjs_a.js");
```
caspervonb, line 18: Name this testdata instead for consistency?

Soremwar: I named it that way since that folder contains the tests for `module.ts`, so I think it's more consistent this way (`stream` has `_stream`, `crypto` has `_crypto`, etc)

Soremwar:
I'm done for the night, I need to actually find some tests we can enable to start using this but it should be good to go for review

Soremwar:
BTW How should we handle the node license here? I can append the MIT license to the top of each file but IDK if that's enough

ry:
Regarding license, you do not need to include the full text of the license at the top of each file, just a single line 
```
// Copyright Joyent and Node contributors. All rights reserved. MIT license.
```

Soremwar:
Test suite functional and configurable, well documented and with the proper copyright headers.

Big plus, I just set up the first test and catched the first implementation inconsistency (a static method inside EventEmitter)

This is ready for review

.github/workflows/ci.yml
```diff
@@ -42,6 +42,12 @@ jobs:
       - name: Test
         run: deno test --unstable --allow-all
 
+      - name: Generate Node tests
```
caspervonb, line 45: Do this ahead of time and commit the results.
We shouldn't be adding module specific tests to the CI.

.github/workflows/ci.yml
```diff
@@ -42,6 +42,12 @@ jobs:
...
+        run: deno run --allow-all node/testdata/setup.ts
+
+      - name: Test Node
+        run: deno run --allow-run --allow-read node/testdata/runner.ts
```
caspervonb, line 49: This could just be run as `node/test.ts` no?

node/testdata/runner.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Usage: `deno run --allow-run --allow-read runner.ts`
+ */
+
+const dir = walk(fromFileUrl(new URL(config.suitesFolder, import.meta.url)), {
+  includeDirs: false,
+  match: testList,
+});
+
+for await (const file of dir) {
+  const process = Deno.run({
```
caspervonb, line 20:
```suggestion
  Deno.test(file, async function() {
    const process = Deno.run({
```

node/testdata/runner.ts
```diff
@@ -0,0 +1,39 @@
...
+      "run",
+      "-A",
+      "--unstable",
+      "require.ts",
+      file.path,
+    ],
+  });
+
+  const { code } = await process.status();
+
```
caspervonb, line 33: Just assert that code == 0 and let our test runner act as a test runner.

caspervonb (changes requested):
Would prefer if this was more like regular tests and didn't require special steps to test.
WASI does this too with upstream tests but re-uses Deno.test.

.github/workflows/ci.yml
Soremwar: I don't do this for a simple reason, I don't want reviewers to be checking node test files every time one is enabled.

This also ensures that the final result will be in fact a Node test and not some modified version of one (that's the whole reason why there is an ignore system in those cases)

node/README.md
```diff
@@ -85,6 +85,21 @@ const leftPad = require("left-pad");
...
+This library contains automated tests pulled directly from the Node repo in
+order ensure compatibility.
+
+Setting up the test runner is as simple as running the `node/testdata/setup.ts`
+file, this will pull the configured tests in and then add them to the test
+workflow.
+
+To enable new tests, simply add a new entry inside `node/testdata/config.json`
+under the `tests` property. The structure this entries must have has to resemble
+a path inside `https://github.com/nodejs/node/tree/master/test`.
```
ry, line 99: Have you considered including a git submodule? Is it too big?

Soremwar: A git submodule could suffice, but it makes it harder to patch some files that are needed for this to work, namely `common/index.js`.

That file is used in pretty much every single file, however it uses things like child_process that are not needed in most cases, so instead I just patch it and all the tests that are enabled continue to use it without any trouble

Not to mention this is more user friendly than submodules :)

kt3k: How about mocking `common/index.js` by injecting a fake module into `require.cache`, using a technique described [here](https://glebbahmutov.com/blog/put-mock-data-into-node-require-cache/)?

node/testdata/test.ts
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+const dir = walk(fromFileUrl(new URL(config.suitesFolder, import.meta.url)), {
+  includeDirs: false,
+  match: testList,
+});
+
+for await (const file of dir) {
+  Deno.test({
+    name: `Node test runner: ${basename(file.path)}`,
```
caspervonb, line 20: Last nit; maybe just let this be the whole path, or at-least the relative path from the node dir so that it's obvious (once we have a lot of tests) which test suite it belongs to.

Soremwar: Done

node/README.md
Soremwar: That might be more complicated than the current solution though @kt3k

node/testdata/setup.ts
```diff
@@ -0,0 +1,177 @@
...
+      truncate: true,
+      write: true,
+    });
+    // This will allow CI to pass without checking linting and formatting
+    // on the test suite files, removing the need to mantain that as well
+    await Deno.writeAll(
+      file,
+      new TextEncoder().encode(
+        "// deno-fmt-ignore-file\n// deno-lint-ignore-file\n" +
+          "\n// Copyright Joyent and Node contributors. All rights reserved. MIT license.\n\n",
```
kt3k, line 140: @Soremwar How about adding version numbers of Node.js around these lines to prevent the confusion about versions? like:

```js
// Copied from Node.js v15.5.1
```

I also think it would be good to add note that it's a generated file.

```js
// This files is automatically generated by ./node/testdata/setup.ts. Do not modify this file manually.
```

kt3k:
Tested locally. I was able to add test cases by adding items in config.json and running `setup.ts` 👍 

`setup.ts` looks good to me. It requires the internet access, but that's only in the contributors' machine who want to change the node compatibility test. It doesn't affect CI and other contributors. `setup.ts` is also good in the sense it uses the fixed version of node.js and clear all the previously downloaded cases. So the result is deterministic and idempotent. I think there will be less (or no) confusions about updating these test cases.

BTW the directory name for these tools look a bit strange to me. I think we usually use `testdata` for just data for testing (like fixtures). Because `setup.ts` is the entrypoint for the developers, how about naming like `/node/tools/setup.ts`?

ry:
I also tested it locally and it works. I'm in favor of this change. Nice work @Soremwar! 

> BTW the directory name for these tools look a bit strange to me. I think we usually use testdata for just data for testing (like fixtures). Because setup.ts is the entrypoint for the developers, how about naming like /node/tools/setup.ts? 

This also stood out to me. Non-fixture stuff should not be in testdata.

node/testdata/setup.ts
Soremwar: I like this. Just added it

Soremwar:
`node/tools` sounds good to me, just updated the test folder

lucacasonato:
Maybe node/_tools because it is internal?

Soremwar:
`_tools` it is

kt3k (approved):
LGTM

caspervonb (approved):
LGTM 😁

kt3k:
@Soremwar Thank you for your contribution! Merging.

PR #698, opened 2021-02-03
 refactor(md5): throw `TypeError` for wrong type

hong4rc:
Use [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) is a good option for `a value is not of the expected type`


ry (approved):
Thanks

nayeemrmn:
I don't see why we have runtime type checking here at all, though.

PR #701, opened 2021-02-03
test(node): Prevent fixture from being run as a test

caspervonb:
The process exit test fixture is being run as a test, it uses console.log which adds to the noise produced by running deno test.

This moves it into testdata folder and drops the test suffix so that it doesn't run as a test.


caspervonb:
Just for clarity, it removes this noise.

Screenshot 2021-02-03 at 9 33 14 PM

ry (approved):
LGTM

Issue #659, opened 2020-06-16
Uncaught BrokenPipe: Broken pipe (os error 32)  


ITS-NIGHTOWL:
After running my ws client for 11 hours, I got this error (for the first time):
```
error: Uncaught BrokenPipe: Broken pipe (os error 32)   
    at unwrapResponse ($deno$/ops/dispatch_minimal.ts:63:11)
    at Object.sendAsyncMinimal ($deno$/ops/dispatch_minimal.ts:106:10)
    at async Object.write ($deno$/ops/io.ts:65:18)
    at async BufWriter.flush (https://deno.land/std@0.54.0/io/bufio.ts:479:25)
    at async writeFrame (https://deno.land/std@0.54.0/ws/mod.ts:144:3)
```
Any reasons as to why?

lideming:
@timonson 
> at async ServerRequest.respond (server.ts:84:7)

Do you use other library or call `ServerRequest.respond` directly? The caller have to catch and handle the error.

timonson:
Hi @lideming , thanks for coming back to this issue!

I call `ServerRequest.respond` directly. In fact, I just tried to reproduce this error with the following code and I was able to do it several times:

```typescript
import { serve } from "https://deno.land/std@0.86.0/http/server.ts";
const server = serve({ port: 8080 });
console.log("http://localhost:8080/");
for await (const req of server) {
  req.respond({ body: await Deno.readFile("./cat.jpeg") });
}
```

> error: Uncaught (in promise) BrokenPipe: Broken pipe (os error 32)
      throw new ErrorClass(res.err.message);
            ^
    at unwrapResponse (deno:runtime/js/10_dispatch_minimal.js:60:13)
    at sendAsync (deno:runtime/js/10_dispatch_minimal.js:99:12)
    at async write (deno:runtime/js/12_io.js:118:20)
    at async BufWriter.write (bufio.ts:499:29)
    at async writeResponse (_io.ts:273:15)
    at async ServerRequest.respond (server.ts:89:7)

The second most annoying thing is that even if I tried to catch the error, it would not work and the server breaks.

The most annoying thing is that it is not deterministically reproducible. 

But here is what has worked for me:

1. Open three different browsers (switching between browsers seems to be the key factor).
2. Refresh the browsers in no specific order.
3. Don't give up if it doesn't break immediately.

Please let me know if you are able to reproduce it. Otherwise I could make a video showing what is happening. But there is not really more magic to it than what I already explained.

[...]

PR #703, opened 2021-02-03
feat(std/http): add HTTP server "error" event

lideming:
This PR allows the user to handle/log the error from the listener and the request reader.
(Moved from denoland/deno#8554)

Before:
The HTTP server rethrows the error from the listener if the error type is unknown, and causes an uncaught error. Also, it ignores all errors while reading requests.

After:
The `Server` class is now inherited from `EventTarget`. It will pass the error to the `error` event. The user can handle it after adding an event listener to the server. By default, the server keeps running after encountering an error.

Related issue: denoland/deno#8499 (not closing it because I don't know where did the ENOTCONN come from)

Usage:
```ts
const server = serve({ port: 8000 });
server.addEventListener('error', (evt) => {
  if (evt.origin == "request") {
    console.warn("Error reading request", evt.connection, evt.error);
    // The server will send "400 Bad Request" and close the connection
    // unless `evt.preventDefault()` is called.
  } else if (evt.origin == "listener") {
    console.warn("Error accepting connection", evt.error);
    // The server will continue to accept connections
    // unless `evt.preventDefault()` is called.
  }
})
  req.respond({ body: "Hello World\n" });
}
```

This PR also make `file_server` print errors from the http server.

Something to discuss:
1. Is it okay to have `Server extends EventTarget` to implement events? Should we implement it in another way? While I am guessing it will hardly break something, should it be considered a BREAKING change?
2. `server.addEventListener("error", (evt) => { ... })` then check `evt.origin == "request"`, or just `server.addEventListener("request-error", ...)`?

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -9,10 +9,11 @@
 import { extname, posix } from "../path/mod.ts";
 import {
   HTTPSOptions,
-  listenAndServe,
-  listenAndServeTLS,
+  serve,
+  Server,
   Response,
   ServerRequest,
+  serveTLS,
 } from "./server.ts";
 import { parse } from "../flags/mod.ts";
 import { assert } from "../_util/assert.ts";
@@ -374,6 +375,16 @@ function normalizeURL(url: string): string {
     : normalizedUrl;
 }
 
+function addrToString(addr: Deno.Addr) {
+  assert(addr.transport == "tcp");
+  if (addr.hostname.indexOf(":") >= 0 && addr.hostname[0] != "[") {
+    // Print "[ip]:port" for IPv6 addresses.
+    return `[${addr.hostname}]:${addr.port}`;
+  } else {
+    return `${addr.hostname}:${addr.port}`;
+  }
+}
+
 function main(): void {
   const CORSEnabled = serverArgs.cors ? true : false;
   const port = serverArgs.port ?? serverArgs.p ?? 4507;
@@ -450,15 +461,36 @@ function main(): void {
     }
   };
 
+  let server: Server;
   let proto = "http";
   if (tlsOpts.keyFile || tlsOpts.certFile) {
     proto += "s";
     tlsOpts.hostname = host;
     tlsOpts.port = port;
-    listenAndServeTLS(tlsOpts, handler);
+    server = serveTLS(tlsOpts);
   } else {
-    listenAndServe(addr, handler);
+    server = serve(addr);
   }
+
+  server.addEventListener("error", (evt) => {
+    if (evt.origin == "request") {
+      console.error(
+        `Error occurred while reading the request from ${
+          addrToString(evt.connection!.remoteAddr)
+        }:`,
+        evt.error,
+      );
+    } else {
+      console.error(evt.error);
+    }
+  });
+
+  (async () => {
+    for await (const req of server) {
+      handler(req);
+    }
+  })();
+
   console.log(`${proto.toUpperCase()} server listening on ${proto}://${addr}/`);
 }
 
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -9,9 +9,9 @@
 import { extname, posix } from "../path/mod.ts";
 import {
   HTTPSOptions,
+  Response,
   serve,
   Server,
-  Response,
   ServerRequest,
   serveTLS,
 } from "./server.ts";
--- a/http/server.ts
+++ b/http/server.ts
@@ -117,11 +117,57 @@ export class ServerRequest {
   }
 }
 
-export class Server implements AsyncIterable<ServerRequest> {
+export type ServerErrorEventListenerOrEventListenerObject =
+  | ServerErrorListener
+  | ServerErrorListenerObject;
+
+export interface ServerErrorListener {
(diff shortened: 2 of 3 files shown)


timonson:
Hi @lideming would you mind telling me if this PR would fix this [issue](https://github.com/denoland/deno_std/issues/659)?

lideming:
@timonson this PR catches errors from the listener and request reader, but not from `ServerRequest.respond`.

bartlomieju:
@cmorten do you have an opinion on this PR?

cmorten:
It's a curious one - certainly we should look to re-invest future efforts in the new "native" abstraction over the old JS Server implementation.

There is no precedent for error eventing (nor error callback or sim. etc.) in the golang std lib from which the RFC / new impl. was inspired - it instead opts for an `ErrorLog` logger.

```golang
// ErrorLog specifies an optional logger for errors accepting
// connections, unexpected behavior from handlers, and
// underlying FileSystem errors.
// If nil, logging is done via the log package's standard logger.
ErrorLog *log.Logger
```

REFs:
- https://pkg.go.dev/net/http#Server
- https://github.com/golang/go/blob/master/src/net/http/server.go#L3165

Looking to JS frameworks, there is prior art in Oak (REF: https://github.com/oakserver/oak) for successful incorporation of eventing into the server (application), indeed it extends `EventTarget`, just as this PR suggests, supporting the likes of `listen`, `error`, and `abort` events.

Equally, the likes of Express (REF: https://github.com/expressjs/express) also support events via the Node `events` built-in `EventEmitter` (though has limited use), and Node's `net.Server` also extends `EventEmitter` with a plethera of events such as `listening` and `error`.

Because we have adopted the golang style in most cases to gracefully handle the majority of errors (closing connections as required etc.), there is little to nothing that a consumer can "hook" into if they do want act upon some exception. Though I suspect the majority of error handling can be done in the consumer's `handler` itself, I could foresee them wanting to know when say accepting a connection has failed for one reason or another etc. - things like APM / instrumentation come to mind, and certainly if there have been a cascade of tls handshake issues it would be good to allow the consumer to gets eyes on that if they desire for debug.

Using the Web `EventTarget` makes sense (over say Node's rolled `EventEmitter`) in the aim to utilise where can Web APIs (certainly should one wish to mirror https://stackblitz.com/ for Deno, I'd hope the usage of Web APIs would make the effort somewhat less than with Node!). An obvious alternative would be to provide an `onError` like handler to the server on instantiation, but I feel eventing is far superior in this instance with the flexibility to add and remove listeners as the server is running.

R.E.

> server.addEventListener("error", (evt) => { ... }) then check evt.origin == "request", or just server.addEventListener("request-error", ...)?

One benefit of an `EventTarget` is the use of event namespacing to be declarative in style when events are broadcasted to listeners - I would suggest we look to avoid overloading a single event namespace where there is a clear distinction as it is generally more scalable to have multiple well scope listeners over a single listener with a large switch statement. We should utilise the fact that `EventTarget` can save consumers on branching logic. That said, we needn't over-segment unnecessarily - for guidance I believe Node's `http.Server` has only two forms of "error" event: `error` and `clientError`.

REFs:
- https://nodejs.org/api/net.html#net_event_error
- https://nodejs.org/api/http.html#http_event_clienterror

A server's behaviour is generally divided into (1) listening/accepting connections, and (2) HTTP over said connections, so in agreement with this PR, perhaps we could consider `connError` and `requestError` (or sim. nomenclature) and divide errors between those events.

One thing would be good to ensure is that with the greater machinary involved (`EventTarget` or sim. + firing errors) that we don't degrade/regress the golden-path performance of the server too much. Certainly errors will be slower, but exceptions always are - finding / keeping the balance between developer experience / tooling and perf _feels_ like an important factor in this part of the std lib. Despite discarding it earlier, simply introducing a callback handler (which I guess is similar in feel to golang's `ErrorLogger`, only not _specifically_ a logger) would potentially be lighter touch and have less of a performance impact? Only PoC + benchmarking could tell 😄

bartlomieju:
@cmorten thanks for in-depth write up. Especially the last paragraph about performance impact is standing out; that said I want to know what @kitsonk thinks about it. If `oak` successfully uses standard DOM event listeners maybe it might be worth exploring providing that functionality in `std/http`? 

Additionally I believe it would nicely integrate with Deploy.

cmorten:
> @cmorten thanks for in-depth write up. Especially the last paragraph about performance impact is standing out; that said I want to know what @kitsonk thinks about it. If `oak` successfully uses standard DOM event listeners maybe it might be worth exploring providing that functionality in `std/http`?

I suspect even with a perf hit it is worth doing - W.r.t Oak, I’d feel like we’d have got std/http right when the low level of something like oak could just be swapped out for the std lib, framework authors can then spend their time adding value at higher levels over dealing with the nitty gritty around the Deno APIs. Indeed there is already quite some overlap between the std/http server and Oak (especially as used as inspiration, and latter contributed the close logic back!). Be great to hear @kitsonk’s thoughts.

> Additionally I believe it would nicely integrate with Deploy.

Agreed, now there is support (REF: https://deno.com/blog/deploy-beta2#deno.listen-and-deno.servehttp) it would be great to keep deploy in mind 😄

PR #705, opened 2021-02-04
WIP fix(node): Fix assert implementation and enable assert test suite

Soremwar:
Based on #695

### DIFF
--- a/node/_tools/config.json
+++ b/node/_tools/config.json
@@ -3,10 +3,14 @@
   "ignore": {
     "common": [
       "index.js"
+    ],
+    "parallel": [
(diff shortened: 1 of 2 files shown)


caspervonb:
We're adding tests for assert here but the actual assertions are based on testing/assert.ts behavior.
Do you plan on porting over over Node's current implementation https://github.com/nodejs/node/blob/master/lib/assert.js here @Soremwar?

Soremwar:
Superseded by #769 . Awesome work @kt3k

kt3k:
@Soremwar Thank you for your initial work!

Issue #654, opened 2020-08-30
Please document the incredibly useful bufio.ts


webdev3301:
I simply cannot find ANY documentation on the following extremely useful imports:

```
import { BufReader, BufWriter } from 'https://deno.land/std@0.67.0/io/bufio.ts';
import { TextProtoReader }      from 'https://deno.land/std@0.67.0/textproto/mod.ts';
```

Can these please be documented. I'm a huge fan of Deno but right now I am literally copying & pasting code knowing that's going to break and hours later it eventually works but I have no idea how. It's like coding in the dark.

bartlomieju:
The module was updated and contains doc strings for all exported APIs.

PR #706, opened 2021-02-04
docs(io/ioutil.ts): add jsdoc

grian32:
I wasn't sure what to put for the `r` and `dest` parameters for `copyN` so i just put Reader and Writer.

Didn't add any description for `buf` parameters because it's very self explanatory.


kt3k (approved):
LGTM

@grian32 You might also be interested in #654

PR #712, opened 2021-02-06
feat(log): Added ability to configure a fallback configuration

Absebo:
You can specify a configuration for a `default` logger but this just adds a logging configuration for unnamed loggers.
I added the ability to add a `fallback` logger to be able to configure loggers that have a name but no configuration.


ry:
@Absebo Sorry for the slow review. It seems like rather than adding extra logic to 'fallback' on a default logger, the system should just throw an exception. Closing without merge.

PR #718, opened 2021-02-10
fix(http/file_server): svg media type

fonsp:
If you use `file_server` to host a site with local svg files...
```html
<img src="./image.svg">
```
...then the image does not show...
image
...because the browser assumes the XML media type, not the svg+xml type.

**This PR adds the correct `Content-Type` header to svg files, which fixes the problem.**

---

_(I also checked png and jpg, they work already because the browser can deduce the media type from the file headers)_


kitsonk (approved):
LGTM

PR #719, opened 2021-02-10
rewrite log (PR #8460)

timreichen:
original: https://github.com/denoland/deno/pull/8460

- remove global state
- class based Loggers
- class based LogLevels
- custom LogLevels
- default LogLevels
  - trace
  - debug
  - info
  - warn
  - error


ry:
Hi @timreichen - very sorry you didn't get a review on this. This patch removes a lot of code, which seems desirable.. but does it also change the interface? If so we need to discuss the changes more exactly. Because this is so old, I think it's best to close this one and allow you to reopen a new PR with the desired changes, if they are still necessary. I will insure they get a prompt review.

timreichen:
@ry Yes, it does change the interface.
- replace global state based logger with class based logger to avoid intransparent state manipulation and more flexibility
- adds `trace` log level and removes `critical` log level to reflect `console` names
- log levels as class instances to avoid translation between strings and codes
- allow custom log levels 
I still think the changes would be appropriate to have more generic and simpler logger mod with mostly the same functionality. I opened a new PR: https://github.com/denoland/deno_std/pull/1137

PR #722, opened 2021-02-11
BREAKING(mime/multipart): return array for multiple values with same form name

### LINKED ISSUE #716: MultipartReader.readForm().entries() does not contain all values for same name
I was writing a test for one of my modules when I got into this issue.
If data is an array (appending to form data with same name multiple times), I cannot retrieve all values for the same name.
I have made a [repo for reproducing this issue](https://github.com/jd1378/deno-multipart-incomplete-repro).
I also made an extra test (go_test.ts) for doing the same test with a [simple go multipart echo server](https://github.com/jd1378/go-simple-multipart-echo) to make sure that the fetch is sending the data (it does).

still, I'm not sure if I'm doing it wrong or it is really a bug.

jd1378:
also adds the related tests

closes #716


mime/multipart.ts
```diff
@@ -245,13 +245,13 @@ function skipLWSPChar(u: Uint8Array): Uint8Array {
 }
 
 export interface MultipartFormData {
-  file(key: string): FormFile | FormFile[] | undefined;
-  value(key: string): string | undefined;
+  files(key: string): FormFile[] | undefined;
+  values(key: string): string[] | undefined;
```
kt3k, line 249: This is going to be a breaking change...

How about aligning `value()` method to the current `file()` method like `value(): string | string[] | undefined`?

And in that case, `entries` would be:
```ts
function* entries(): IterableIterator<
  [string, string | string[] | FormFile | FormFile[] | undefined]
> { ... }
```

caspervonb: That is what the original PR did in the first commit and over conversations in Discord I suggested going this route.

I don't think overloading the return type is a good idea, actually let me word that more strongly, it's terrible and I'm okay with making this a breaking change.

jd1378: I did exactly that in `89229de`, but I agree with caspervonb on breaking this to make it better. everyone should know that std is brittle right now anyway and imports are versioned. So overall I don't think that should be an issue.

kt3k: OK. I'm convinced. Maybe it's good time to break things, and existing users should be protected by versioning.

kt3k (approved):
LGTM

caspervonb (approved):
LGTM.

Friendly remember to re-tag this as BREAKING(mime/multipart) when landing 🙃

PR #728, opened 2021-02-14
BREAKING(encoding): remove module utf8.ts

zhangyuannie:
This PR removed `utf8.ts` as per the std-wg. I'm not sure if we should add `@deprecated` for some time before completely removing through.


bartlomieju (approved):
LGTM! Thanks @zhangyuannie

PR #735, opened 2021-02-17
BREAKING feat(io/streams): Strengthen iterator to readable stream conversion

### LINKED ISSUE #734: Change readableStreamFromAsyncIterator to readableStreamFromIterable
```ts
export function readableStreamFromIterable<T>(
  iterable: Iterable<T> | AsyncIterable<T>,
): ReadableStream<T>;
```
This would be a non-breaking enhancement, if the name change wasn't required.

This helper should be as flexible as possible. The proposed signature would allow passing anything from an array to a generator function iife. Matches `pooledMap()` for instance.

nayeemrmn:
Closes #734.

### DIFF
--- a/io/streams.ts
+++ b/io/streams.ts
@@ -46,10 +46,24 @@ export function writableStreamFromWriter(
   });
 }
 
-/** Create a `ReadableStream` from an `AsyncIterator`. */
-export function readableStreamFromAsyncIterator<T>(
-  iterator: AsyncIterableIterator<T>,
+/** Create a `ReadableStream` from any kind of iterable.
+ *
+ *      const r1 = readableStreamFromIterable(["foo, bar, baz"]);
+ *      const r2 = readableStreamFromIterable((async function* () {
+ *        await new Promise(((r) => setTimeout(r, 1000)));
+ *        yield "foo";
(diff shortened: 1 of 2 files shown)


ry (commented):
I'm in favor of this change - looks clean and useful. 

Are there any objections? This is a breaking API.

lucacasonato:
Looks good to me.

caspervonb:
Seems like a more general purpose improvement to me.

io/streams.ts
```diff
@@ -46,10 +46,24 @@ export function writableStreamFromWriter(
...
+ *        await new Promise(((r) => setTimeout(r, 1000)));
+ *        yield "baz";
+ *      })());
+*/
+export function readableStreamFromIterable<T>(
+  iterable: Iterable<T> | AsyncIterable<T>,
 ): ReadableStream<T> {
+  const iterator: Iterator<T> | AsyncIterator<T> =
+    (iterable as Iterable<T>)[Symbol.iterator]?.() ??
+      (iterable as AsyncIterable<T>)[Symbol.asyncIterator]?.();
```
nayeemrmn, line 66: Need to switch these around to match `for await`.

ry (approved):
LGTM

PR #738, opened 2021-02-18
feat(node): add tty module

kt3k:
This PR implements `isatty` function of [tty](https://nodejs.org/api/tty.html) module of Node.js API.

A heavily dependent npm module [supports-color](https://www.npmjs.com/package/supports-color) uses this method. (ref: https://github.com/chalk/supports-color/blob/main/index.js#L150). This PR enables requiring `supports-color` in deno.

### DIFF
--- a/node/tty.ts
+++ b/node/tty.ts
@@ -4,7 +4,7 @@
 // ref: https://nodejs.org/api/tty.html
 
 // Returns true when the given numeric fd is associated with a TTY and false otherwise.
-function isatty(fd: unknown) {
+function isatty(fd: number) {
   if (typeof fd !== "number") {
     return false;
   }
--- a/node/tty.ts
+++ b/node/tty.ts
@@ -1,0 +1,22 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+// This module implements 'tty' module of Node.JS API.
+// ref: https://nodejs.org/api/tty.html
+
+// Returns true when the given numeric fd is associated with a TTY and false otherwise.
+function isatty(fd: unknown) {
+  if (typeof fd !== "number") {
+    return false;
+  }
+  try {
+    return Deno.isatty(fd);
+  } catch (_) {
+    return false;
+  }
+}
+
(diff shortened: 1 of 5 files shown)


node/tty.ts
```diff
@@ -0,0 +1,22 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+// This module implements 'tty' module of Node.JS API.
+// ref: https://nodejs.org/api/tty.html
+
+// Returns true when the given numeric fd is associated with a TTY and false otherwise.
+function isatty(fd: unknown) {
```
satyarohith, line 7:
```suggestion
function isatty(fd: number) {
```
I think we can make this a number since the [node docs](https://nodejs.org/api/tty.html#tty_tty_isatty_fd) also state it as a number. Thoughts?

node/tty_test.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { assert } from "../testing/asserts.ts";
+import { isatty } from "./tty.ts";
+
+Deno.test("[node/tty isatty] returns true when fd is a tty, false otherwise", () => {
+  assert(Deno.isatty(Deno.stdin.rid) === isatty(Deno.stdin.rid));
```
caspervonb, line 6: I don't find these test cases to be particularly meaningful, but I'm not sure if we can do better here in std as that would involve spawning pty processes 🤔

node/tty.ts
kt3k: I checked DefinitelyTyped as well and they also type this as `number`.  ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/cd10dd7/types/node/tty.d.ts#L8

It seems better to make it `number`. Thank you for you suggestion!

caspervonb (approved):
LGTM

node/tty_test.ts
ry: agree

ry (approved):
LGTM

PR #742, opened 2021-02-19
0.88.0

ry:


.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         run: deno upgrade --canary
 
       - name: Format
-        run: deno fmt --check
+        run: deno fmt --check --ignore=fs/testdata/json_invalid.json
```
bartlomieju, line 63: Why? We should strive not to add any ignores and we got away with adding any until now

ry: -> https://github.com/denoland/deno_std/pull/743

bartlomieju (approved):

PR #743, opened 2021-02-19
Fix JSON formatting

ry:
Since https://github.com/denoland/deno/commit/d9b1f96897239bf7de8cdfd11d40e3f99bb29a4a landed in CLI, CI is broken here.


.github/workflows/ci.yml
```diff
@@ -60,7 +60,7 @@ jobs:
         run: deno upgrade --canary
 
       - name: Format
-        run: deno fmt --check
+        run: deno fmt --check --ignore=fs/testdata/json_invalid.json
```
bartlomieju, line 63: Again, this should be removed and inline ignore should be added to the file

ry: How does that work?

bartlomieju: Add `// deno-fmt-ignore-file` in the appropriate file

ry: but it's json..

bartlomieju: Right... This is unlucky, we could fix that by changing it to JSONC

ry:
-> #744

PR #744, opened 2021-02-19
Remove unused testdata files

ry:


caspervonb (approved):
Right, these are probably left over from when I removed the json functions a few months back.

LGTM

bartlomieju (approved):

PR #745, opened 2021-02-19
fix(node): native modules should be extensible

kt3k:
This PR changes the native commonjs modules (in our node.js compat later) extensible, i.e. this enables the code like the below:

```js
const fs = require("fs");
Object.defineProperty(fs, "key", { value: "value" });
```

The above code throws at `Object.defineProperty` call with the current implementation because `fs` is not an usual object, but a [module object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#creating_a_module_object) which is not extensible.

---

Note: Some modules, such as (heavily depended) [graceful-fs](https://www.npmjs.com/package/graceful-fs), [extend](https://github.com/isaacs/node-graceful-fs/blob/195c934/graceful-fs.js#L25) the native modules. So this feature is required for supporting those modules.


ry (approved):
LGTM

PR #747, opened 2021-02-22
feat(node): add constants module

kt3k:
This PR adds `constants` module in node.js compatibility library.

`constants` is deprecated in documentation since v6.x, but the module still exists in source code, and a hugely depended module like `graceful-fs` still [depends on it](https://github.com/isaacs/node-graceful-fs/blob/v4.2.6/polyfills.js#L1). So we need to polyfill `constants` module to support many tools which depends on graceful-fs.

Examples of the modules which depends on graceful-fs include: webpack, npm, fs-extra, vinyl-fs(part of gulp), karma, etc..


ry (approved):
LGTM

PR #748, opened 2021-02-23
fix(node/util): export promisify & callbackify

kt3k:
`callbackify`, `promisify`, and `types` of `util` module wasn't exported as part of default export. This PR fixes it.


bnoordhuis (approved):

kt3k:
@bnoordhuis Thanks! :)

PR #749, opened 2021-02-23
fix(node/process): make process.argv an array

kt3k:
This PR makes process.argv simply an array.

The current implementation simulates an array by using Proxy, but it doesn't support array methods and it causes problem in executing cli tools in npm.

In node.js, process.argv is just an array (ref 1: https://github.com/nodejs/node/blob/eed3c72/src/node_process_object.cc#L158, ref 2: https://github.com/nodejs/node/blob/eed3c72/src/env-inl.h#L601). So this PR change it to an array.

### DIFF
--- a/node/process.ts
+++ b/node/process.ts
@@ -27,9 +27,21 @@ const notImplementedEvents = [
 /** https://nodejs.org/api/process.html#process_process_arch */
 export const arch = Deno.build.arch;
 
-// TODO(kt3k): The first item should be the entrypoint of
-// node compatibility library.
-const argv = [Deno.execPath(), fromFileUrl(Deno.mainModule), ...Deno.args];
+// The first 2 items are placeholders.
+// They will be overwritten by the below Object.defineProperty calls.
+const argv = ["", "", ...Deno.args];
+// Overwrites the 1st item with getter.
+Object.defineProperty(argv, "0", {
+  get() {
+    return Deno.execPath();
+  },
+});
+// Overwrites the 2nd item with getter.
+Object.defineProperty(argv, "1", {
+  get() {
+    return fromFileUrl(Deno.mainModule);
+  },
+});
 
 /** https://nodejs.org/api/process.html#process_process_chdir_directory */
 export const chdir = Deno.chdir;
--- a/node/process.ts
+++ b/node/process.ts
@@ -27,40 +27,9 @@ const notImplementedEvents = [
 /** https://nodejs.org/api/process.html#process_process_arch */
 export const arch = Deno.build.arch;
 
-function getArguments() {
-  return [Deno.execPath(), fromFileUrl(Deno.mainModule), ...Deno.args];
-}
-
-//deno-lint-ignore ban-ts-comment
-//@ts-ignore
-const _argv: {
-  [Deno.customInspect]: () => string;
-  [key: number]: string;
-} = [];
-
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

Soremwar:
This is a stepback from https://github.com/denoland/deno/pull/8940, making read permissions mandatory just to use node globals.

An ideal solution would improve over the Proxy solution to make all array methods available, not remove it all together

caspervonb:
> This is a stepback from [denoland/deno#8940](https://github.com/denoland/deno/pull/8940), making read permissions mandatory just to use node globals.
> 
> An ideal solution would improve over the Proxy solution to make all array methods available, not remove it all together

Not sure why we need a proxy for accessing `Deno.args` as that is an array that is initialized once and has no permission requirements?

Soremwar:
No, we need read access to use `Deno.execPath`, which is the first item of process.argv

caspervonb:
> No, we need read access to use `Deno.execPath`, which is the first item of process.argv

In that case, can't we just wrap the call to Deno.execPath in a try/catch?

Soremwar:
Swallowing a permission error doesn't sound like the best way to approach this if I'm being honest

kt3k:
@Soremwar 
I set getters for the keys "0" and "1" of argv. This seems requiring the permissions lazily. It requires permissions only when accessing argv[0] or argv[1]. Please take a look when you have time.

bnoordhuis (approved):
LGTM, but I have a question: is there any downside from a security perspective if we'd remove the read access restriction from `Deno.execPath()`?

This PR is as good as it's going to get with the current restriction but, ideally, the node compat layer wouldn't throw access exceptions at runtime.

kt3k:
@bnoordhuis In terms of Compatibility, I personally think it's ok to remove the permission restrictions as Node.js doesn't have such security layer originally, but I'm not sure we can disable security only in part of the program.

bnoordhuis:
What I mean is: would removing the read access check from `Deno.execPath()` be okay? The location of the binary doesn't seem like particularly sensitive information to me - that you know where it is doesn't mean you can read/write/execute it.

The counterargument is that it exposes information about the file system layout to an attacker but is knowing that deno resides in `/usr/local/bin` helpful when staging an attack? 🤷

At any rate, feel free to merge this. :-)

kt3k:
> The counterargument is that it exposes information about the file system layout to an attacker but is knowing that deno resides in /usr/local/bin helpful when staging an attack? 🤷

If deno is invoked with only `--allow-write` permission (and deno is placed in strange random path), then attacker may replace the deno executable with their own executable by knowing Deno.execPath(). Probably that is the reasoning behind this restriction. (BTW [this issue](https://github.com/denoland/deno/issues/5627) seems the first motivation of this restriction, though I don't find detailed discussion about possible attacks.)

caspervonb:
So might be due for discussion for v2.0? I don't think we should be changing permissions for stable features.
To me it seems like the permissions for `Deno.execPath` should probably either be removed or be changed to `--allow-env` as that's what it implies: environmental information.

To compare it with another sandbox that's relevant these days, WASI does not currently put any restrictions on cwd (cwd is not intrinsic and is derived from $CWD) or argv[0].

kt3k:
Yes, let's continue in denoland/deno.

@Soremwar If you have further concern on this implementation, please open an issue.

PR #757, opened 2021-02-25
feat(node): add crypto.createHash

kt3k:
This PR adds crypto.createHash() method and related functions.

Many items are still missing, but this partial implementation covers the use case of [object-hash](https://www.npmjs.com/package/object-hash) npm module. object-hash is heavily depended by many modules, which include eslint-loader, nestjs, tailwindcss, [etc](https://www.npmjs.com/browse/depended/object-hash).

### DIFF
--- a/node/crypto.ts
+++ b/node/crypto.ts
@@ -1,6 +1,100 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright Joyent, Inc. and Node.js contributors. All rights reserved. MIT license.
 import { default as randomBytes } from "./_crypto/randomBytes.ts";
+import {
+  createHash as stdCreateHash,
+  Hasher,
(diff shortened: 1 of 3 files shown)


node/crypto.ts
```diff
@@ -1,6 +1,103 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright Joyent, Inc. and Node.js contributors. All rights reserved. MIT license.
```
bnoordhuis, line 2: What specifically requires the copyright attribution? I don't see anything that looks like it's lifted straight from node.

bnoordhuis (approved):

node/crypto.ts
kt3k: I didn't copy the source code, but I coped some of the API documentation such as `class Hash`, `update`, `digest`, etc.

PR #758, opened 2021-02-25
fix(node): fix export items of events

kt3k:
In Node.js, EventEmitter is exported as both `require("events")` and `require("events").EventEmitter`, but latter is not exported in the current implementation. This PR fixes it.


caspervonb:
Would it maybe make more sense to just fix the default exports in that module instead of special casting the shim rules for it?

caspervonb (approved):
LGTM

ry (approved):

PR #761, opened 2021-02-26
fix(http/server): Swallow NotConnected errors from listener.accept()

nayeemrmn:
Apparently another error kind that can be thrown from `listener.accept()` with a quirky client but shouldn't terminate the server. Experienced by https://github.com/denoland/deno_std/issues/659#issuecomment-786386671, discussed on Discord. No repro or test, but it makes sense.


nuts-n-bits:
Yes so fwiw the error stack is this
```
error: Uncaught (in promise) NotConnected: Transport endpoint is not connected (os error 107)
    throw new ErrorClass(res.err.message);
          ^
    at processResponse (deno:core/core.js:213:11)
    at Object.jsonOpAsync (deno:core/core.js:231:12)
    at async Listener.accept (deno:runtime/js/30_net.js:98:19)
    at async Server.acceptConnAndIterateHttpRequests (https://deno.land/std@0.88.0/http/server.ts:226:14)
    at async MuxAsyncIterator.callIteratorNext (https://deno.land/std@0.88.0/async/mux_async_iterator.ts:30:31)
```

bartlomieju (approved):
LGTM

Issue #741, opened 2021-02-18
[std/testing] Function signature of assertObjectEquals() does not accept interfaces


Seally:
The type `Record<PropertyKey, unknown>` used in `assertObjectEquals()` does not accept interface types as an argument. This is due to an intentional quirk in how TypeScript operates with interfaces. With that said, I do not believe this assertion function should disallow inputs that uses types declared with the `interface` keyword.

```ts
import { assertObjectMatch } from "https://deno.land/std@0.87.0/testing/asserts.ts";

interface NumberToken {
    type: "number";
    value: number;
}

const token: NumberToken = {
    type: "number",
    value: 15
};

assertObjectMatch(token, { type: "number" }); // doesn't work

type TNumberToken = {
    type: "number";
    value: number;
};

const typeToken: TNumberToken = {
    type: "number",
    value: 15
};

assertObjectMatch(typeToken, { type: "number" }); // works
```

I'm not completely sure what to suggest changing this to, but the most minimal change is to use `Record<PropertyKey, any>` instead, which seems to work for this case and confirmed to not be a bug from the TypeScript team.

Here are some test inputs for the `Record<PropertyKey, any>` signature:

```ts
// Continued from previous block
function foo(input: Record<PropertyKey, any>) {}

foo(token); // pass
foo(typeToken); // pass
foo(foo); // pass
foo(""); // fail
foo(15); // fail
```

Relevant issues:
* https://github.com/microsoft/TypeScript/issues/42825
* https://github.com/microsoft/TypeScript/issues/41518

cola119:
I'm facing this issue too. May I make the p-r to change the props type `Record<PropertyKey, unknown>` to `Record<PropertyKey, any>`? I think this change is trivial and there don't seem to be any unexpected problems.

getspooky:
I fixed https://github.com/denoland/deno_std/pull/763

wperron:
Fixed in #763

PR #763, opened 2021-02-26
fix(testing): Function signature of assertObjectEquals() does not accept interfaces #741

getspooky:
The type `Record<PropertyKey, unknown>` used in `assertObjectEquals()` does not accept interface types as an argument.

```ts
import { assertObjectMatch } from "https://deno.land/std@0.87.0/testing/asserts.ts";

interface NumberToken {
    type: "number";
    value: number;
}

const token: NumberToken = {
    type: "number",
    value: 15
};

assertObjectMatch(token, { type: "number" }); // doesn't work

type TNumberToken = {
    type: "number";
    value: number;
};

const typeToken: TNumberToken = {
    type: "number",
    value: 15
};

assertObjectMatch(typeToken, { type: "number" }); // works
```
The solution is to change `assertObjectMatch` first param from `unknown` to `any`.

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -477,6 +477,7 @@ export function assertNotMatch(
  * If not, then throw.
  */
 export function assertObjectMatch(
+  // deno-lint-ignore no-explicit-any
   actual: Record<PropertyKey, any>,
   expected: Record<PropertyKey, unknown>,
 ): void {
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -72,7 +72,9 @@ function buildMessage(diffResult: ReadonlyArray<DiffResult<string>>): string[] {
   messages.push("");
   messages.push(
     `    ${gray(bold("[Diff]"))} ${red(bold("Actual"))} / ${
-      green(bold("Expected"))
+      green(
+        bold("Expected"),
+      )
     }`,
   );
   messages.push("");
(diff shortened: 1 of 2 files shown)


testing/asserts.ts
```diff
@@ -469,15 +472,15 @@ export function assertNotMatch(
  * If not, then throw.
  */
 export function assertObjectMatch(
-  actual: Record<PropertyKey, unknown>,
+  actual: Record<PropertyKey, any>,
```
lowlighter, line 475: I think you need to add `// deno-lint-ignore no-explicit-any` to pass the linter check.
Should the `expected` variable keep the typing `Record<PropertyKey, unknown>` or should it also use `any` instead of `unknown`?

getspooky: This is due to an intentional quirk in how `TypeScript` operates with `interfaces` 

- microsoft/TypeScript#42825
- microsoft/TypeScript#41518

lowlighter: Actually I was wondering why just `actual` is being changed while `expected` is left as is.
Shouldn't both be modified?

getspooky: The problem occurs for the first `params` `actual` because it would be `interface` or `types` there is no need to change the second param to `any`.

getspooky:
@JJ it would be good to fixe this issue.

wperron (changes requested):
Change looks good, but please revert the formatting changes as they're not pertinent to the change.

getspooky:
@wperron done.

wperron (approved):
LGTM thanks

PR #768, opened 2021-02-28
feat: add event dispatch for server response

keithamus:
Hey deno folk!

### The Sales Pitch

I think it is critically important to provide middleware as a base in `std/http` because it gives a foundation for frameworks like Oak to build off of, but in a way to which other frameworks can benefit; "a rising tide lifts all boats" and all that. One area of contention in the Node ecosystem is that various frameworks exist, and yet all have to effectively re-implement the same set of middlewares over and over, as there is no centralised mechanism for middleware. Compare this to a language like Golang where middleware is an intrinsic part of the std library and we see frameworks building atop that, but also smaller libraries integrating with it in powerful ways which gives user's more choice and library/framework authors more power & expresivity.

### The Big Picture Idea

Okay enough sales pitch. I have an idea around the `std/http` server, and how it can be a foundation for middleware capable applications, in a similar-ish way to how Golang's `http.Handler` works. Just a quick reminder on Go's `http.Handler`, and how Middleware can be defined in Go:

```golang
func Middleware(next http.Handler) http.Handler {
  return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    // Do something
    next.ServeHTTP(w, r)
    // Cleanup
  })
}
```

The basic pattern is a middleware is a `type (next: Handler) => Handler` to use TS parlance. The beauty is in the simplicity; a middleware gets to chose when to call the next handler, if ever, and as such it can perform actions before and/or after, and act as flow control - determining what to do with the next handler.

I believe we have something pretty darn close with the existing Deno server:

```typescript
type Middleware = (incoming: AsyncIterable<ServerRequest>): AsyncIterable<ServerRequest>
```

With this type signature it is possible to consume incoming requests, handle logic before, and enact flow control, conditionally yielding to the next middleware. For example we could write a middleware which logs incoming requests:

```typescript
export function logger(): Middleware {
  return async function* (server: AsyncIterable<ServerRequest>): AsyncIterable<ServerRequest> {
      console.log("->", req.method, req.url)
      yield req
    }
  }
}

logger(serve({ port: 8080 }))
```

We can also conditionally handle some requests:

```typescript
export function randomlyReject(): Middleware {
  return async function* (server: AsyncIterable<ServerRequest>): AsyncIterable<ServerRequest> {
      if (Math.random() > 0.5) {
        req.respond({ body: 'Bad luck' })
      } else {
        yield req
      }
    }
  }
}

randomlyReject(serve({ port: 8080 }))
```

### The Problem in the Way

Where this pattern falls down a little, is handling logic when a response happens. Right now there is no way to determine what happens to a `ServerRequest` once it gets yielded. A `ServerRequest` is currently _unobservable_.

### The Solution?

One way to solve this is to have `ServerRequest` be an `EventTarget`, and to have it dispatch events. This PR is a sketch for that. I've changed `ServerRequest` to extend `EventTarget` and changed the `respond` function to dispatch a `before-respond` event, and an `after-respond` event. Now our logger middleware can get a bit smarter:

```typescript
export function logger(): Middleware {
  return async function* (server: AsyncIterable<ServerRequest>): AsyncIterable<ServerRequest> {
      const start = Date.now()
      console.log("->", req.method, req.url)
      req.addEventListener('after-respond', () => {
        console.log("<-", req.method, req.url, req.status, `(${time}ms)`);
      })
      yield req
    }
  }
}
```

I think this change, as well as the broader concept of middleware within `std/http` will open up all kinds of possibilities. I'm interested to see what y'all think about this? Happy to steer the specifics of this PR in a different direction - I just wanted to get something a little bit tangible in front of y'all.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -10,7 +10,7 @@ import {
   writeResponse,
 } from "./_io.ts";
 
-export class ServerRequest {
+export class ServerRequest extends EventTarget {
   url!: string;
   method!: string;
   proto!: string;
@@ -83,6 +83,8 @@ export class ServerRequest {
 
   async respond(r: Response): Promise<void> {
     let err: Error | undefined;
+    const cancelled = !this.dispatchEvent(new CustomEvent('before-respond', { cancelable: true }))
+    if (cancelled) return
     try {
       // Write our response!
       await writeResponse(this.w, r);
@@ -95,6 +97,7 @@ export class ServerRequest {
       }
       err = e;
     }
+    this.dispatchEvent(new CustomEvent('after-respond'))
     // Signal that this request has been processed and the next pipelined
     // request on the same connection can be accepted.
     this.#done.resolve(err);


bartlomieju:
@kitsonk I would love to get your input on this PR as it would have an effect on `oak`. Also we need to keep in mind that there's a plan to depracate `std/http` server and instead stabilize `Deno.serveHttp` API in the near future.

keithamus:
I'll close this in favour of https://github.com/denoland/deno_std/pull/1283

PR #769, opened 2021-02-28
feat(node): fix assert module, enable test cases

kt3k:
Continued from #705.

This PR implements the following 16 assertion APIs from Node.js [assert](https://nodejs.org/api/assert.html) module.

- `AssertionError` - done
- `ok` - done
- `fail` - done
- `strict` - done
- `throws` - done
- `doesNotThrow` - done
- `equal` - done except error message
- `notEqual` - done except error message
- `match` - done except error message
- `doesNotMatch` - done except error message
- `strictEqual` - done except error message
- `notStrictEqual` - done except error message
- `deepStrictEqual` - done except error message
- `notDeepStrictEqual` - done except error message
- `deepEqual` - skipped implementation because `test-assert.js` doesn't have test case
- `notDeepEqual` - skipped implementation because `test-assert.js` doesn't have test case

This branch currently passes [test/parallel/test-assert.js](https://github.com/nodejs/node/blob/master/test/parallel/test-assert.js) except assertion message validation part.

I skipped `deepEqual` and `notDeepEqual` because these 2 don't have test cases in `test-assert.js`.

I skipped the assertions of error messages of 8 equality methods (equal/notEqual/match/doesNotMatch/strictEqual/notStrictEqual/deepStrictEqual/notDeepStrictEqual) because these uses the testing/asserts.ts under the hood, and it requires much more work to align them to node.js.

(I left 64 TODO comments. Most of them are about validation of error messages.)


kt3k:
I skipped the checks of the assertion error messages because this PR uses `std/testing/asserts.ts` under the hood. I'd like to fix that in follow up RPs.

This PR leaves many TODO comments in test-assert.js, but most of them are about validation of error messages. Most APIs are functional at this point and usable for implementing other test cases. So I'd like this to get reviewed at this point.

node/_errors.ts
```diff
@@ -83,15 +83,47 @@ export class ERR_INVALID_ARG_TYPE extends NodeTypeError {
   constructor(a1: string, a2: string | string[], a3: unknown) {
     super(
       "ERR_INVALID_ARG_TYPE",
+      // TODO(kt3k): Expression for a2 is wrong.
```
bnoordhuis, line 86: Wrong how? Do you mean it differs from node's? If so, please mention that explicitly. :-)

node/_errors.ts
```diff
@@ -83,15 +83,47 @@ export class ERR_INVALID_ARG_TYPE extends NodeTypeError {
...
         typeof a2 === "string"
           ? a2.toLocaleLowerCase()
           : a2.map((x) => x.toLocaleLowerCase()).join(", ")
-      }. Received ${typeof a3} (${a3})`,
+      }.${invalidArgTypeHelper(a3)}`,
     );
   }
 }
 
+export class ERR_INVALID_ARG_VALUE extends NodeTypeError {
```
bnoordhuis, line 96: Why `extends NodeTypeError`? Node's is a `TypeError`.

bnoordhuis (approved):
Nice work, Yoshiya.

node/_errors.ts
kt3k: I'm not sure about the intention of `NodeTypeError`, but it seems introduced in this PR https://github.com/denoland/deno/pull/7934/files, and it seems being used instead of TypeError in our code base. So I don't think it's good idea to change it for only this one, but rather I suggest we should address them together with others in later PRs.

bnoordhuis: Fair enough. I'll mark this resolved.

PR #775, opened 2021-03-02
test(path): update test cases for canary

### LINKED ISSUE #774: 3 test cases fail with canary (2021-03-02)
These 3 test cases of `std/path` module seem failing with deno canary.

- [path] fromFileUrl (win32)
- [path] toFileUrl
- [path] toFileUrl (win32)

kt3k:
This PR disables some test assertions of `std/path` which started failing with canary today.

related to: https://github.com/denoland/deno/pull/9276

closes #774

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,8 @@
 comment: false
 codecov:
   require_ci_to_pass: true
+coverage:
+  status:
+    project:
+      default:
+        informational: true
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,3 @@
+comment: false
+codecov:
+  require_ci_to_pass: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,8 +39,19 @@ jobs:
         if: matrix.deno == 'canary'
         run: deno upgrade --canary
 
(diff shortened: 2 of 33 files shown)


kitsonk:
Shouldn't we be fixing the code instead of disabling tests?

kt3k:
@kitsonk I first thought of that, but the CI runs with both 1.7.5 and canary until 1.8 released. So I think it's difficult to update it to pass with both versions.

lucacasonato:
> @kitsonk I first thought of that, but the CI runs with both 1.7.5 and canary until 1.8 released. So I think it's difficult to update it to pass with both versions.

I think we should disable this test in stable CI then and fix it in canary. It working on canary is more important than it working on stable IMO.

kt3k:
> I think we should disable this test in stable CI then and fix it in canary. It working on canary is more important than it working on stable IMO.

hmm... OK, I'll try it.

kt3k:
Introduced isCanary flag (`const isCanary = /\d\.\d\.\d\+[0-9a-f]{7}/.test(Deno.version.deno)`), and also merged #776, which is necessary to make CI green.

lucacasonato (approved):
LGTM - lets revert this right after 1.8 is released.

PR #776, opened 2021-03-02
chore(build): disable codecov status

kt3k:
This PR disables the commit status from codecov.io.

Codecov sends failed commit status to the branch when the coverage rate is decreased. Higher coverage rate is generally a good thing, but marking the commit as `failure` because of the decreased coverage doesn't make sense to our use case.

ref: https://docs.codecov.io/docs/commit-status#disabling-a-status

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,7 @@
 comment: false
 codecov:
   require_ci_to_pass: true
+coverage:
+  status:
+    project: off
+    patch: off
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -3,5 +3,6 @@ codecov:
   require_ci_to_pass: true
 coverage:
   status:
-    project: off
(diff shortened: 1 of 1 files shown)


caspervonb:
Right so it shows up in the checks for the main branch which was not intentional.

Maybe we should set `informational: true` if we don't care about thresholds? (https://docs.codecov.io/docs/commit-status#informational
> Use Codecov in informational mode. Default is false. If true is specified the resulting status will pass no matter what the coverage is or what other settings are specified. Informational mode is great to use if you want to expose codecov information to other developers in your pull request without necessarily gating PRs on that information.

kt3k:
Informational mode sounds good 👍 Updated the PR.

caspervonb (approved):
LGTM

ry (approved):

kt3k:
This is landed as part of #775

PR #777, opened 2021-03-02
feat(testing): Adding `assertTruthy` to ensure a value is true in a boolean context

getspooky:
Use `assertTruthy` when  you don't care what a value is and you want to ensure a value is true in a boolean context.
In Javascript The following values are **always falsy**:

- `false`
- `0`(zero)
- `''` or `""` (empty string)
- `null`
- `undefined`
- `NaN`


caspervonb:
Why is this preferable over `assert`?

PR #778, opened 2021-03-02
property into method (converted type)

nurmohammed840:
Now typescript would infer type properly.

### DIFF
--- a/async/deferred.ts
+++ b/async/deferred.ts
@@ -5,9 +5,9 @@
 // See https://github.com/Microsoft/TypeScript/issues/15202
 // At the time of writing, the github issue is closed but the problem remains.
 export interface Deferred<T> extends Promise<T> {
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - thanks!

nurmohammed840:
```ts
// TODO(ry) It'd be better to make Deferred a class that inherits from
// Promise, rather than an interface. 
```
I also agree...

Some noticeable traits are:
-  `Deferred` class can infer its type itself.
-  Check defer object, if it's an instance of `Deferred` class.

```js
Deferred extends Promise {
    constructor() {
        let methods;
        super((resolve, reject) => {
            methods = { resolve, reject };
        });
        Object.assign(this, methods);
    }
}

var defer = new Deferred();
console.log(defer instanceof Deferred); // not that important feature though, But nice to have
```

PR #781, opened 2021-03-02
added overloading (async/delay.ts)

nurmohammed840:
This is a nice feature to have.
```ts
Promise.all([
  delay(1000),
  delay(1500, "Hello, World!"),
  delay(2000, () => "function"),
  delay(2500, Promise.resolve("resolved")),
  delay(2000, async () => "async function"),
]).then(console.log);
```

### DIFF
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -3,6 +3,7 @@
 export function delay<T>(ms: number): Promise<T>;
 export function delay<T>(ms: number, fn: () => T): Promise<T>;
 export function delay<T>(ms: number, value: T): Promise<T>;
+// deno-lint-ignore no-explicit-any
 export function delay<T>(ms: number, arg?: any): Promise<T> {
   return typeof arg == "function"
     ? new Promise((res) => setTimeout(() => res(arg()), ms))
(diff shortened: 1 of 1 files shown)


kt3k:
I think this feature isn't such useful. The example above can be rewritten like the below without having this feature:

```ts
Promise.all([
  delay(1000),
  delay(1500).then(() => "Hello, World!"),
  delay(2000).then(() => "function"),
  delay(2500).then(() => Promise.resolve("resolved")),
  delay(2000).then(async () => "async function"),
]).then(console.log);
```

nurmohammed840:
You are right, So closing...

nurmohammed840:
Sir,
Should we need timeout?

```ts
/** Reject after the given number of milliseconds, If not Resolved */
export function timeout<T>(ms: number, fn: () => T): Promise<T>;
export function timeout<T>(ms: number, value: T): Promise<T>;
export function timeout<T>(ms: number, arg: any, msg = ""): Promise<T> {
    return Promise.race([
        typeof arg == "function" ? arg() : arg,
        new Promise((_, reject) => setTimeout(reject, ms, new Error(`Timeout: ${msg}`))),
    ]);
}
```

PR #782, opened 2021-03-02
implement overloaded test. (async/delay.ts)

nurmohammed840:


kt3k:
delay is not overloaded? https://github.com/denoland/deno_std/blob/main/async/delay.ts

nurmohammed840:
Overloaded proposal dismissed.so closing...

PR #785, opened 2021-03-04
feat(node): Basic support for child_process.spawn

uki00a:
## Summary

This PR adds basic support for `child_process.spawn` based on the `Deno.run` API. 

Related to https://github.com/denoland/deno/issues/5376

## Limitation

Currently, support for the `shell` option in Windows is quite limited due to escaping issues. For more details, see the following issues:

* https://github.com/rust-lang/rust/issues/29494
* https://github.com/denoland/deno/issues/8852


node/child_process_test.ts
```diff
@@ -0,0 +1,236 @@
...
+    env: { NO_COLOR: "true" },
+  });
+  try {
+    let exitCode = null;
+    childProcess.on("exit", (code) => {
+      promise.resolve();
+      exitCode = code;
+    });
+    await promise;
+    assertStrictEquals(exitCode, 0);
```
kt3k, line 47: 👍

kt3k:
@uki00a What do you think is the remaining task for the first pass?

node/module.ts
```diff
@@ -607,6 +608,10 @@ nativeModulePolyfill.set(
   "constants",
   createNativeModule("constants", nodeConstants),
 );
+nativeModulePolyfill.set(
+  "child_process",
+  createNativeModule("child_process", nodeChildProcess),
+);
```
kt3k, line 614: Could you update the corresponding test case of this change in module_test.ts for consistency?

uki00a:
@kt3k I'm still having one problem and not sure if that problem is caused by this PR's code or by the `Deno.run` code. :thinking:

## What is the problem?

There is a `test.ts` with the following contents:

```typescript
import { spawn } from "./node/child_process.ts"

const deno = spawn("deno", ["--version"]);
let out = "";
deno.stdout!.on("data", (chunk) => {
  out += chunk;
});
deno.on("exit", () => {
  console.log(out);
});
```

When running this file, it works as intended:

```shell
$ deno run -A --unstable test.ts

deno 1.8.3 (release, x86_64-unknown-linux-gnu)
v8 9.0.257.3
typescript 4.2.2
```

Then, change the `test.ts` as follows:

```typescript
import { spawn } from "./node/child_process.ts"

const git = spawn("git", ["--version"]);
let out = "";
git.stdout!.on("data", (chunk) => {
  out += chunk;
});
git.stderr!.on("data", (err) => console.error("hello"));
git.on("exit", () => {
  console.log(out);
});
```

When I run the `test.ts`, I get the following errors:

```shell
$ deno run -A --unstable test.ts

Uncaught Interrupted: operation canceled
Uncaught Interrupted: operation canceled
```

node/module.ts
uki00a: Done!

uki00a:
The problem causing the "operation canceled" error has been resolved by 9e38473. However, the canary build seems to be failing... :thinking: :cry:

kt3k:
Canary test is failing because of #860, and probably will be fixed by #861. No need of actions for now.

kt3k (approved):
@uki00a Nice work! Thanks!

LGTM

uki00a:
Thanks!

PR #789, opened 2021-03-06
feat(encoding/yaml): Add support for JS types and user types

lowlighter:
This extends current YAML support with an `EXTENDED_SCHEMA` which supports `!!js/regexp`, `!!js/function` and `!!js/undefined` to remove current [Deno limitation](https://github.com/denoland/deno_std/tree/main/encoding#warning-limitations) (heavily based on [nodeca/js-yaml-js-types](https://github.com/nodeca/js-yaml-js-types) but reworked a bit to be more succint).

This also exposes the `Type` function (which is already implemented but not available to end-users) to allow the creation of custom types with `!tags` as defined in [YAML spec](https://yaml.org/spec/1.2/spec.html#tag/).

### DIFF
--- a/encoding/_yaml/stringify_test.ts
+++ b/encoding/_yaml/stringify_test.ts
@@ -3,8 +3,11 @@
 // Copyright 2011-2015 by Vitaly Puzrin. All rights reserved. MIT license.
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-import { assertEquals } from "../../testing/asserts.ts";
+import { assertEquals, assertThrows } from "../../testing/asserts.ts";
 import { stringify } from "./stringify.ts";
+import { YAMLError } from "./error.ts";
+import { DEFAULT_SCHEMA, EXTENDED_SCHEMA } from "./schema/mod.ts";
+import { Type } from "./type.ts";
 
 Deno.test({
   name: "stringified correctly",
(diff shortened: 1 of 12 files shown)


encoding/_yaml/type/function.ts
```diff
@@ -0,0 +1,41 @@
+// Ported and adapted from js-yaml-js-types v1.0.0:
+// https://github.com/nodeca/js-yaml-js-types/tree/ac537e7bbdd3c2cbbd9882ca3919c520c2dc022b
+// Copyright 2011-2015 by Vitaly Puzrin. All rights reserved. MIT license.
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { Type } from "../type.ts";
+import type { Any } from "../utils.ts";
+
+// Note: original implementation used Esprima to handle functions
+// To avoid depencies, we'll just try to check if we can construct a function from given string
```
kt3k, line 10: `depencies` -> `dependencies`?

encoding/_yaml/type/function.ts
```diff
@@ -0,0 +1,41 @@
...
+function reconstructFunction(code: string) {
+  const func = new Function(`return ${code}`)();
+  if (!(func instanceof Function)) {
+    throw new TypeError(`Expected function but got ${typeof func}`);
```
kt3k, line 14:
```suggestion
    throw new TypeError(`Expected function but got ${typeof func}: ${code}`);
```

It would be more user friendly if the error message included the given code.

lowlighter: Yes, I applied suggestion 👍🏼

encoding/_yaml/stringify_test.ts
```diff
@@ -39,3 +42,93 @@ test: foobar
...
+    const object = {
+      regexp: {
+        simple: /foobar/,
+        modifiers: /foobar/im,
+      },
+      undefined: undefined,
+    };
+
+    const expected = `regexp:
+  simple: !<tag:yaml.org,2002:js/regexp> /foobar/
```
lowlighter, line 72: so unfortunately, in js-yaml 3 (ported version of Deno), custom tags were dumped with `!<tag:yaml.orf,2022:..>`, and it has been fixed in v4 (https://github.com/nodeca/js-yaml/commit/a0d0caa5aa0f5354fefa9c637cfb7c4c17ef7d02) 

it does works though, it's just the display which is a bit more verbose and less pretty

encoding/_yaml/stringify_test.ts
```diff
@@ -39,3 +42,93 @@ test: foobar
...
+      () => stringify(object),
+      YAMLError,
+      "unacceptable kind of an object to dump",
+    );
+  },
+});
+
+Deno.test({
+  name:
+    "`!!js/*` yaml types are correctly handled with extended schema while stringifying",
```
kt3k, line 61: nice test!

kt3k (approved):
Tests look very well (It has 81% diff coverage ref: https://github.com/denoland/deno_std/pull/789/checks?check_run_id=2079608904)

LGTM

Issue #788, opened 2021-03-06
Serve std/88 2x slower then std/87


irustm:
deno --version
```
deno 1.8.0 (release, x86_64-apple-darwin)
v8 9.0.257.3
typescript 4.2.2
```

**87:**

```ts
import { serve } from "https://deno.land/std@0.87.0/http/server.ts";
const s = serve({ port: 8000 });
for await (const req of s) {
    req.respond({ body: "Hello World\n" });
}


```
test:
`npx autocannon -c 100 http://localhost:8000`

<img width="732" alt="Снимок экрана 2021-03-07 в 00 21 30" src="https://user-images.githubusercontent.com/16316579/110221078-19c1c900-7edb-11eb-8ca7-55f3e000e6d0.png">

**88:**

```ts
import { serve } from "https://deno.land/std@0.88.0/http/server.ts";
const s = serve({ port: 8000 });
for await (const req of s) {
    req.respond({ body: "Hello World\n" });
}
```

test:
`npx autocannon -c 100 http://localhost:8000`

<img width="664" alt="Снимок экрана 2021-03-07 в 00 25 56" src="https://user-images.githubusercontent.com/16316579/110221170-b1bfb280-7edb-11eb-9e95-e359e562a1d1.png">

irustm:
perhaps because a new TextEncoder is created every time

https://github.com/denoland/deno_std/pull/728

kitsonk:
Yeah, that isn't desirable.

PR #790, opened 2021-03-06
fix: encoder/decoder one instance

irustm:
https://github.com/denoland/deno_std/issues/788


kitsonk (approved):
LGTM

We need to be more careful about PR reviews it seems.

PR #793, opened 2021-03-07
refactor(node/fs): update fs import

ije:
this pr is to avoid the `node/fs` to import unnecessary fs modules.


kt3k (approved):
LGTM. Thanks @ije !

PR #799, opened 2021-03-11
docs(deno_std): Add Commit Message Format section

getspooky:


kt3k:
Thank you for the suggestion, but we don't have these rules for commit messages and actually contributors and even core members don't follow them. So I think enforcing such rules doesn't make sense for our case.

Closing without merge.

PR #807, opened 2021-03-15
fix(multipart): support useDefineForClassFields

lucacasonato:
Broke as a result of https://github.com/denoland/deno/commit/c4709834b37640fd3c9d492123e6add904546573.


ry (approved):
LGTM - tho the extra boilerplate is rather unfortunate.

PR #808, opened 2021-03-15
feat: add io/buffer and io/util module

lucacasonato:
This moves the `Deno.Buffer` and `Deno.readAll` / `Deno.readAllSync` /
`Deno.writeAll` / `Deno.writeAllSync` implementations to std, so we can
deprecate them in Deno 1.9, and remove them in Deno 2.0.

### DIFF
--- a/archive/README.md
+++ b/archive/README.md
@@ -4,11 +4,12 @@
 
 ```ts
 import { Tar } from "https://deno.land/std@$STD_VERSION/archive/tar.ts";
+import { Buffer } from "https://deno.land/std@$STD_VERSION/io/buffer.ts";
 
 const tar = new Tar();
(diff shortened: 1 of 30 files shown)


nayeemrmn:
The exports of `io/buffer.ts` and `io/util.ts` can all go in `io/bufio.ts` (which I prefer was named `io/buffer.ts`, but breaking). They all provide some kind of buffering for Deno IO. Not sure what `io/buffer` vs `io/bufio` would be.

lucacasonato:
I think that makes sense. We can combine all three into `io/buffer.ts`. I don't think the breaking change matters as std is still pre 1.0.

ry (approved):
Nice work! LGTM

Issue #803, opened 2021-03-14
bufio.readLines is pathologically slow for very long lines


sswam:
I found that bufio.readLines becomes pathologically slow when given large lines, such as a single 10MB line. Deno took 58 seconds to read and write this line, compared to under 0.1 seconds for Go, Python and Perl. Those other languages are able to read and write a 100MB line in under 0.3 seconds on my laptop, whereas I was not patient enough to wait for Deno bufio to finish that task.

I'm guessing that bufio is reallocing its buffer in small steps, which results in quadratic complexity, rather than the normal method which is to double the capacity of the buffer each time it needs to be expanded, which gives linear complexity. I can have a go at fixing this, but would like to get your go ahead first.

## Create a file containing a single 10MB line. I used Perl:

```
perl -e 'for (1..10000000) { print "X"; } print "\n";' > bigline-10M.txt
```

## readLine.ts: TypeScript Deno code to read and write the line

```
import { readLines } from "https://deno.land/std/io/bufio.ts";

const lines = readLines(Deno.stdin);

for await (const l of lines) {
        console.log(l);
        break;
}
```

## readLine.py: Python code to read and write the line

```
import sys
print(sys.stdin.readline(), end='')
```

## readLine.pl: Perl code to read and write the line

```
print scalar <>;
```

## readLine.go: Go code to read and write the line

```
package main

import "bufio"
import "os"
import "fmt"
import "log"

func main() {
        var buf []byte = make([]byte, 64*1024)
        scanner := bufio.NewScanner(os.Stdin)
        scanner.Buffer(buf, 1e12)
        if scanner.Scan() {
                fmt.Println(scanner.Text())
        }
        if err := scanner.Err(); err != nil {
                log.Fatal(err)
        }
}
```

## test.sh: script to time execution for each language

```
time() { command time -f '%e %U %S %C' "$@"; }
time deno run readLine.ts < bigline-10M.txt > /dev/null
time python readLine.py < bigline-10M.txt > /dev/null
time perl readLine.pl < bigline-10M.txt > /dev/null
time go build readLine.go
time ./readLine < bigline-10M.txt > /dev/null
```

## results

```
58.67 56.13 12.43 deno run readLine.ts
0.05 0.03 0.02 python readLine.py
0.01 0.00 0.00 perl readLine.pl
0.07 0.07 0.04 go build readLine.go
0.02 0.00 0.02 ./readLine
```

sam-watkins-sr:
Thank-you for working on this.

[...]

PR #812, opened 2021-03-22
readDelim fixes and optimizations

kfdf:
Replacing `Deno.Buffer` with `concat` made `readDelim` about 2.5 times faster. Also, now it shouldn't have problems with unusual input, like the one described in #803.
And a bug (typo?) in `createLPS` caused it to fail for many non-trivial inputs.


kt3k:
Closing in favor of #867 and #877. Thank you for your contributions anyway.

PR #815, opened 2021-03-25
feat(testing): assertMatch function support string as param

getspooky:
`assertMatch` accepts a string, which it will try to match:
```ts
assertMatch("foobar@deno.com", "deno.com");
```


wperron:
I'm not sure this is necessary, creating a RegExp from a string is hardly a complicated step to do before using `assertMatch`

lucacasonato:
@getspooky How is this different from `assertIncludesString`?

getspooky:
@lucacasonato `assertMatch` uses include in order to match `string`, The purpose of this PR is to give more option.

wperron:
Respectfully closing; constructing a `RegExp` from a string is already trivial.

PR #818, opened 2021-03-25
ref(encoding): Promise<void> return types are unnecessary boilerplate.

getspooky:
We should strive for concise code. `void` and `Promoise<void>` return types are unnecessary boilerplate.
Discussion https://github.com/denoland/deno/issues/8832

### DIFF
--- a/encoding/csv_test.ts
+++ b/encoding/csv_test.ts
@@ -377,8 +377,8 @@ x,,,
    */
   {
     Name: "HugeLines",
-    Input:
-      "#ignore\n".repeat(10000) + "@".repeat(5000) + "," + "*".repeat(5000),
+    Input: "#ignore\n".repeat(10000) + "@".repeat(5000) + "," +
+      "*".repeat(5000),
     Output: [["@".repeat(5000), "*".repeat(5000)]],
     Comment: "#",
   },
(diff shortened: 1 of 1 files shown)


encoding/csv_test.ts
```diff
@@ -473,16 +473,13 @@ for (const t of testCases) {
...
-            new BufReader(new StringReader(t.Input ?? "")),
-            {
-              separator,
-              comment: comment,
-              trimLeadingSpace: trim,
-              fieldsPerRecord: fieldsPerRec,
-              lazyQuotes: lazyquote,
-            },
-          );
+          await readMatrix(new BufReader(new StringReader(t.Input ?? "")), {
```
wperron, line 476: We don't need the formatting change, you can revert this

wperron (changes requested):
Thanks, just a small comment on the formatting change.

encoding/csv_test.ts
getspooky: @wperron fixed

wperron: Thanks, but I'm really just interested in the changes on line 452 and 624

getspooky: @wperron done

wperron (approved):
LGTM, thanks

PR #819, opened 2021-03-26
remove `Promise<void>` return type annotation

wperron:
Follow-up of #818

There's a couple of instances where `Promise<void>` is still there,
those are the exceptions, for example functions that explicitely return
a `Promise` object or functions that return the promise from another
function call without any `await` statement.

used this one-liner to automatically remove everything:

```bash
grep -rli 'promise<void>' | xargs sed -i 's/: Promise<void>//g
```

### DIFF
--- a/archive/tar.ts
+++ b/archive/tar.ts
@@ -318,7 +318,7 @@ export class Tar {
    *                 e.g., test.txt; use slash for directory separators
    * @param opts options
    */
-  async append(fn: string, opts: TarOptions): Promise<void> {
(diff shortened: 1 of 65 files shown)


kt3k (approved):
LGTM 👍

kt3k:
(Memo: We used to be against this (ref: https://github.com/denoland/deno/issues/4409), but the newer issue https://github.com/denoland/deno/issues/8832 suggests we should follow this direction. So this change makes sense to me.)

Issue #648, opened 2021-02-01
std/node/fs missing a lot of functionality


AliBasicCoder:
so, the problem is std/node/fs is missing a lot of functions:

important ones are highlighted

1. `fs.createReadStream`
2. `fs.createWriteStream`
5. fs.fdatasync
6. fs.fdatasyncSync
7. `fs.fstat`
8. `fs.fstatSync`
9. fs.fsync
10. fs.fsyncSync
11. fs.ftruncate
12. fs.ftruncateSync
13. fs.futimes
14. fs.futimesSync
15. `fs.lchmod`
16. `fs.lchmodSync`
17. fs.lutimes
18. fs.lutimesSync
19. `fs.link`
20. `fs.linkSync`
21. `fs.lstat`
22. `fs.lstatSync`
23. fs.mkdtemp
24. fs.mkdtempSync
25. fs.open
26. fs.opendir
27. fs.opendirSync
28. fs.read (with it's to overwrites)
29. `fs.readdir`
30. `fs.readdirSync`
31. fs.readSync (with it's to overwrites)
32. fs.readv
33. fs.readvSync
34. fs.realpath
35. fs.realpath.native
36. fs.realpathSync
37. fs.realpathSync.native
38. `fs.rename`
39. `fs.renameSync`
40. `fs.rmdir`
41. `fs.rmdirSync`
42. `fs.stat`
43. `fs.statSync`
44. `fs.symlink`
45. `fs.symlinkSync`
46. fs.truncate
47. fs.truncateSync
48. `fs.unlink`
49. `fs.unlinkSync`
50. `fs.unwatchFile`
51. fs.utimes
52. fs.utimesSync
53. `fs.watch`
54. `fs.watchFile`
55. fs.write (with it's to overwrites)
56. fs.writeSync (with it's to overwrites)
57. fs.writev
58. fs.writevSync

majidsajadi:
@kt3k I see you updated this [comment](https://github.com/denoland/deno_std/issues/648#issuecomment-734691245). i guess `link` and `linkSync` are implemented too. 

I will send some PR`s for:

- [ ] fs.fdatasync
- [ ] fs.fdatasyncSync
- [ ] fs.fstat
- [ ] fs.fstatSync
- [ ] fs.fsync
- [ ] fs.fsyncSync
- [ ] fs.ftruncate
- [ ] fs.ftruncateSync
- [ ] fs.futimes
- [ ] fs.futimesSync

[...]

PR #825, opened 2021-03-29
Feat/add node symlink and symlink sync

majidsajadi:
Implemented symlink and symlinkSync in node/fs. #648


kt3k (approved):
@majidsajadi LGTM. Thank you for the contribution!

PR #826, opened 2021-03-31
Bugfix / Inconsistency: Respect Uint8Array offsets

trgwii:
Okay, so found a potential bug in std:
```ts
import { varnum } from "https://deno.land/std@0.91.0/encoding/binary.ts";

const main = new Uint8Array([1, 2, 3, 4, 5, 6, 7, 8]);
const sub = main.subarray(4, 8);
console.log(varnum(sub, { dataType: "uint8" })); // logs 1
```

This PR should only affect varnum() and friends operating on Uint8Array subarrays where the offset is > 0.

I replaced all hardcoded 0 indices in the relevant functions with b.byteOffset to get the desired behavior of logging 5 in the above example.

I also added a test for my change (note that this only tests varnum, and not all the relevant functions).

### DIFF
--- a/encoding/binary.ts
+++ b/encoding/binary.ts
@@ -65,21 +65,21 @@ export function varnum(b: Uint8Array, o: VarnumOptions = {}): number | null {
   const view = new DataView(b.buffer);
   switch (o.dataType) {
     case "int8":
-      return view.getInt8(0);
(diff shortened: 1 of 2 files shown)


lucacasonato (approved):
LGTM, thanks!

PR #836, opened 2021-04-02
ref:  Promise<void> return types are unnecessary boilerplate

getspooky:
We should strive for concise code. `void` and `Promoise<void>` return types are unnecessary boilerplate.
Discussion denoland/deno#8832.
This changes includes folders:
- [x] encoding
- [x] io
- [x] http
- [x] text-proto


wperron:
Thanks, but there's already #819 opened which addresses all of the modules at once, so I'll close in favor of that one

PR #841, opened 2021-04-04
Feat/add node fdatasync and fdatasync sync

majidsajadi:
Implemented `fdatasync` and `fdatasyncSync` in `node/fs`. #648


kt3k (approved):
LGTM. Thanks!

PR #843, opened 2021-04-05
feat: add iter and iterSync to io/util

lucacasonato:
This moves the `Deno.iter` and `Deno.iterSync` implementations to std,
so we can deprecate them in Deno 1.9, and remove them in Deno 2.0.

### DIFF
--- a/http/_io.ts
+++ b/http/_io.ts
@@ -4,6 +4,7 @@ import { TextProtoReader } from "../textproto/mod.ts";
 import { assert } from "../_util/assert.ts";
 import { Response, ServerRequest } from "./server.ts";
 import { STATUS_TEXT } from "./http_status.ts";
+import { iter } from "../io/util.ts";
 
 const encoder = new TextEncoder();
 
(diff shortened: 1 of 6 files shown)


nayeemrmn:
I think the opportunity should be taken to give it a more explicit name, since it's not built-in anyway. Maybe `readIter()`, being the basis of `readLines()` and `readDelim()`.

kt3k:
> I think the opportunity should be taken to give it a more explicit name

How about `iterableFromReader` following the convention of `/io/streams.ts`?

lowlighter:
Seems that https://github.com/denoland/deno_lint/pull/651#issue-608334650 has been prodded recently which introduce the `no-deprecated-deno-api` rule, but since #843 (this PR) has not been prodded yet, it's making builds from all new PRs failing  ([example](https://github.com/denoland/deno_std/pull/849/checks?check_run_id=2313065288#step:6)):

```
 (no-deprecated-deno-api) `iter` is deprecated and scheduled for removal in Deno 2.0
  for await (const chunk of Deno.iter(r)) {
                            ^^^^^^^^^
    at /home/runner/work/deno_std/deno_std/http/_io.ts:178:28

    hint: Use `iter` from https://deno.land/std/io/util.ts instead

(no-deprecated-deno-api) `iter` is deprecated and scheduled for removal in Deno 2.0
    for await (const chunk of Deno.iter(conn)) {
                              ^^^^^^^^^
    at /home/runner/work/deno_std/deno_std/http/file_server_test.ts:107:30

    hint: Use `iter` from https://deno.land/std/io/util.ts instead
```

It's not blocking per se, but I think it's worth mentioning 🙂

lucacasonato:
Thanks for the heads up @lowlighter. I'll address it shortly.

ry (approved):
LGTM

PR #849, opened 2021-04-10
fix(std/testing): equals does not differentiate undefined/absent keys

### LINKED ISSUE #842: std/testing assertEquals handling of undefined properties
I have the latest Deno version installed. I wanted assertEquals to ignore fields on the actual object that were undefined when doing the comparison but found that assertEquals will fail if keys were set to undefined on the actual but not on the expected object. If it worked in the way I desired, none of the assertEquals calls in my example below would throw.

If assertEquals is meant to fail due to keys not matching, then I found a bug with that behavior. In the below example, all of the assertions pass. I added a comment after the assertion that should fail. I'm guessing that it is checking the number of keys, then comparing the values for each of those keys.

```ts
import { assertEquals, assertThrows } from "https://deno.land/std@0.92.0/testing/asserts.ts";

Deno.test("example", () => {
  const obj = { a: undefined, b: undefined };
  assertThrows(() => assertEquals(obj, {}));
  assertThrows(() => assertEquals(obj, { a: undefined }));
  assertEquals(obj, { a: undefined, b: undefined });
  assertEquals(obj, { a: undefined, c: undefined }); // should fail but doesn't
  assertThrows(() => assertEquals(obj, { a: undefined, b: undefined, c: undefined }));
});
```

```sh
deno test example.ts
running 1 tests
test example ... ok (6ms)

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (6ms)
```

lowlighter:
Fix #842

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -161,6 +161,9 @@ export function equal(c: unknown, d: unknown): boolean {
         if (!compare(a && a[key as Key], b && b[key as Key])) {
           return false;
         }
+        if (((key in a) && (!(key in b))) || ((key in b) && (!(key in a)))) {
+          return false;
+        }
       }
       seen.set(a, b);
       return true;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks! @lowlighter

PR #852, opened 2021-04-14
feat(io/streams): add readableStreamFromReader

kitsonk:
This adds `readableStreamFromReader` to `io/streams`.  This is quite useful when dealing with files and the new native HTTP bindings.  For example to "stream" a file from disk in a response you would do something like this:

```ts
import { readableStreamFromReader } from "https://deno.land/std/io/streams.ts";

for await (const conn of Deno.listen({ port: 8000 })) {
  (async () => {
    const httpConn = Deno.serveHttp(conn);
    for await (const { request, respondWith } of httpConn) {
      const file = await Deno.open("./big_file.zip", { read: true });
      respondWith(new Response(readableStreamFromReader(file)));
    }
  })();
}
```

### DIFF
--- a/io/streams.ts
+++ b/io/streams.ts
@@ -3,6 +3,14 @@
 import { Buffer } from "./buffer.ts";
 import { writeAll } from "./util.ts";
 
+const DEFAULT_CHUNK_SIZE = 16_640;
+
+function isCloser(value: unknown): value is Deno.Closer {
+  return typeof value === "object" && value != null && "close" in value &&
+    // deno-lint-ignore no-explicit-any
+    typeof (value as Record<string, any>)["close"] === "function";
+}
+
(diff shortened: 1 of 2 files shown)


crowlKats:
We already had this at a point, but it was decided to replace it with readableStreamFromIterable, as you can just use Deno.iter()

kitsonk:
> We already had this at a point, but it was decided to replace it with readableStreamFromIterable, as you can just use Deno.iter()

Ok, though `Deno.iter()`/`io/util::iter()` plus `readableStreamFromIterable()` don't handle IO errors gracefully with the stream, and they double up the number of promises needed to read the stream, as there is the promise to read from the reader and the yielded promise from the async iterator. It also doesn't manage the closing of the resource when it is done (and doesn't allowing setting of the queuing strategy, but that is a minor point)

lucacasonato (approved):
LGTM!

PR #853, opened 2021-04-14
ref(io/buffer): use std/bytes copy

### LINKED ISSUE #851: io/buffer should use std/bytes copy
Currently `io/buffer.ts` uses a private `copyBytes` function, which should be replaced with an import of `copy()` from `std/bytes/copy.ts`.

majidsajadi:
CLoses #851


kt3k (approved):
LGTM. Thanks!

PR #854, opened 2021-04-14
feat(path/glob): add caseInsensitive option

### LINKED ISSUE #806: [std/fs] `expandGlob` needs a `caseSensitive` option
Given that almost all of the Windows' file systems in use are case-insensitive, the fs module's `expandGlob` and `expandGlobSync` functions should take an option such as `caseSensitive` to allow for case-sensitive or case-insensitive glob searches. It would default to `true` except for Windows (or other usually case-insensitive) platforms where it would default to `false`. That seems to be the usual practice for various other languages.

It is actually more complicated (and, yes, NTFS can be configured to be case-sensitive), but this would cover 99% of cases correctly and allow user modification in the other small number of cases which don't quite fit the default.

In looking at the code, commits, and discussion (#1856), `flags` (for regex flags such as "i") was a component of the options, but was removed. I couldn't find any discussion on why that occurred, but the current behavior is very posix/unix specific and could use the flexibility to minimize surprise for users on platforms with case-insensitive file systems (most commonly Windows obviously).

This can be currently hacked by replacing any non-special character in the glob with a glob character set including both the lower and upper case character forms. But, I do think that's a "hack" and direct support would be a better option.

ref: [Windows ~ File System Comparison](https://docs.microsoft.com/en-us/windows/win32/fileio/filesystem-functionality-comparison) @@ <https://archive.is/l6dPi>

lowlighter:
Closes #806

This add a `caseInsensitive` option to `globToRegExp`.

When `os` option (which defaults to current os) is set to `"windows"`, this option will be enabled by default as file system is case insensitive on it

### DIFF
--- a/fs/expand_glob.ts
+++ b/fs/expand_glob.ts
@@ -78,7 +78,7 @@ export async function* expandGlob(
     includeDirs = true,
     extended = false,
     globstar = false,
-    caseInsensitive = Deno.build.os === "windows",
+    caseInsensitive,
   }: ExpandGlobOptions = {},
 ): AsyncIterableIterator<WalkEntry> {
   const globOptions: GlobOptions = { extended, globstar, caseInsensitive };
@@ -185,7 +185,7 @@ export function* expandGlobSync(
     includeDirs = true,
     extended = false,
     globstar = false,
-    caseInsensitive = Deno.build.os === "windows",
+    caseInsensitive,
   }: ExpandGlobOptions = {},
 ): IterableIterator<WalkEntry> {
   const globOptions: GlobOptions = { extended, globstar, caseInsensitive };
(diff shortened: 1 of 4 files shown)


piscisaureus:
I'm not convinced 'caseInsensitive' should be enabled on windows by default. 

* Users might be tempted to think that `glob` will do case folding exactly like windows, which this doesn't. 
* With the advent of WSL there are now case-sensitive directories on my windows box. 
* On mac, we don't perform unicode normalization either.

I think it'd be better to make the default 'case sensitive' everywhere, and make windows users think about it a little harder.

lowlighter:
> I think it'd be better to make the default 'case sensitive' everywhere, and make windows users think about it a little harder.

Yes. I changed it to default it to `false` on all platforms

kt3k (approved):
LGTM

PR #855, opened 2021-04-14
feat(http): allow custom http codes along with custom response text

### LINKED ISSUE #664: [bug]: allow custom http codes
e.g. returning status 555 gives this:

```
error: Uncaught InvalidData: Bad status code
    throw new Deno.errors.InvalidData("Bad status code");
```

lowlighter:
Closes #664

As specified in the [rfc2616 spec](https://tools.ietf.org/html/rfc2616):
> HTTP status codes are extensible. HTTP applications are not required to understand the meaning of all registered status codes, though such understanding is obviously desirable. However, applications MUST understand the class of any status code, as indicated by the first digit, and treat any unrecognized response as being equivalent to the x00 status code of that class, with the exception that an unrecognized response MUST NOT be cached. For example, if an unrecognized status code of 431 is received by the client, it can safely assume that there was something wrong with its request and treat the response as if it had received a 400 status code. In such cases, user agents SHOULD present to the user the entity returned with the response, since that entity is likely to include human- readable information which will explain the unusual status. 

Since this PR allows custom codes which may have not default response text defined, this also add `statusText` so custom error codes can be sent with human readable messages.

### DIFF
--- a/http/_io.ts
+++ b/http/_io.ts
@@ -237,11 +237,8 @@ export async function writeResponse(
   const protoMajor = 1;
   const protoMinor = 1;
   const statusCode = r.status || 200;
-  const statusText = STATUS_TEXT.get(statusCode);
+  const statusText = r.statusText ?? STATUS_TEXT.get(statusCode) ?? "";
   const writer = BufWriter.create(w);
-  if (!statusText) {
-    throw new Deno.errors.InvalidData("Bad status code");
-  }
   if (!r.body) {
     r.body = new Uint8Array();
   }
--- a/http/_io.ts
+++ b/http/_io.ts
@@ -240,7 +240,9 @@ export async function writeResponse(
   const statusText = r.statusText ?? STATUS_TEXT.get(statusCode) ?? null;
   const writer = BufWriter.create(w);
   if (statusText === null) {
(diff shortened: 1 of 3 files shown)


http/_io.ts
```diff
@@ -237,11 +237,8 @@ export async function writeResponse(
   const protoMajor = 1;
   const protoMinor = 1;
   const statusCode = r.status || 200;
-  const statusText = STATUS_TEXT.get(statusCode);
+  const statusText = r.statusText ?? STATUS_TEXT.get(statusCode) ?? "";
```
bnoordhuis, line 240: An empty status text might confuse some HTTP parsers so maybe don't allow that. The PR looks fine otherwise.

lowlighter: So maybe something like below would be better?
```diff
if (!statusText) {
-  throw new Deno.errors.InvalidData("Bad status code");
+  throw new Deno.errors.InvalidData("Missing status text");
}
```

bnoordhuis: Yes, I'd say so.  Allowing `{ statusCode: 777, statusText: "" }` (statusText deliberately empty but not omitted) is something I'd be supportive of, too. That way people can deliberately opt into it if they wanted to.

PR #859, opened 2021-04-16
Improved textproto/mod.ts readability and maintenability

SmashingQuasar:
Hey!

I made a few changes to this file to make it simpler and avoid common pitfalls.

I removed repeated calls to `charCode` by simply defining constants at the beginning of the file.
`charCode` became unused so I removed it.
`str` could be simplified to one line using a simple ternary so I did it.
`readLine` could also be simplified so it became a ternary.
`skipSpace` got improved readability and does not repeatedly calls `charCode`. Comparison has also been factorized.
`readLineSlice` was entirely rewrote to avoid `while(true)`. It made it shorter and more comfortable to read and understand. It's purpose should be clearer. I let `skipSpace` exist for now and added a more comprehensive comment about it. I intend to review it's use later to avoid such a hack if possible.

**_These changes pass the provided tests but I also wrote my own to make sure_** `readLineSlice` **_provides a similar behavior.
I did not include those tests to the pull request. I can indeed include them if necessary._**

Cheers!


SmashingQuasar:
I fixed the lint issues. It seems that tests are failing for Canary. I am unsure what Canary is in the development life of this application.
It also seems the tests fail for totally unrelated reasons. I can fix those errors provided that I know how to run the Canary tests.

wperron:
> I fixed the lint issues. It seems that tests are failing for Canary. I am unsure what Canary is in the development life of this application.
> It also seems the tests fail for totally unrelated reasons. I can fix those errors provided that I know how to run the Canary tests.

This is an unrelated issue caused by a change in core (ref #860 ) that I'm fixing at the moment. Just for info, you can upgrade to the canary version with `deno upgrade --canary`.

SmashingQuasar:
@wperron thanks for the answer!

If you are already working on this issue, I will not look into it to avoid doing the same work twice.

I believe my PR is fixed then so I will not make any change to this PR for now.

👍

textproto/mod.ts
```diff
@@ -126,39 +122,29 @@ export class TextProtoReader {
...
   skipSpace(l: Uint8Array): number {
+    const charCodes: Array<number> = [CHAR_SPACE, CHAR_TAB];
     let n = 0;
-    for (let i = 0; i < l.length; i++) {
-      if (l[i] === charCode(" ") || l[i] === charCode("\t")) {
-        continue;
+
+    for (const val of l) {
+      if (!charCodes.includes(val)) {
+        ++n;
```
wperron, line 144: Is there a reason for not using `n++` here?

wperron (changes requested):
I've got some minor nitpicks but overall seems nice :+1: If you have the test case(s) I would gladly take them as well

textproto/mod.ts
```diff
@@ -31,9 +28,8 @@ export class TextProtoReader {
    * eliding the final \n or \r\n from the returned string.
    */
   async readLine(): Promise<string | null> {
-    const s = await this.readLineSlice();
-    if (s === null) return null;
-    return str(s);
+    const s: Uint8Array | null = await this.readLineSlice();
```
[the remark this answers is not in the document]
SmashingQuasar: I can change it if this is part of this repository's requirements. As a general rule I never use type inferring as it makes code difficult to read and understand for people unfamiliar with the code base. This is especially true in this case where the return type isn't obvious when reading the method's name. I would agree with you for the other cases of inferred types, but not this one, or at least not with this method name.

textproto/mod.ts
```diff
@@ -126,39 +122,29 @@ export class TextProtoReader {
...
-      }
-      line = line ? concat(line, l) : l;
-      if (!more) {
-        break;
+    let line: Uint8Array = new Uint8Array(0);
+    let r: ReadLineResult | null = null;
+
+    do {
+      r = await this.r.readLine();
+      if (r !== null && this.skipSpace(r.line) !== 0) { //TODO: Kept skipSpace to preserve behavior but it should be looked into to check if it makes sense when this is used.
```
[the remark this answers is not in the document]
SmashingQuasar: Sure will change, if you have any insight about the original TODO I would also be interested since it was relatively vague.
👍

textproto/mod.ts
```diff
@@ -126,39 +122,29 @@ export class TextProtoReader {
...
-        // comes from nor how to fix it.
-        if (this.skipSpace(l) === 0) {
-          return new Uint8Array(0);
-        }
-        return l;
-      }
-      line = line ? concat(line, l) : l;
-      if (!more) {
-        break;
+    let line: Uint8Array = new Uint8Array(0);
```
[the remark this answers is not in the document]
SmashingQuasar: Will change, no problem.

textproto/mod.ts
```diff
@@ -126,39 +122,29 @@ export class TextProtoReader {
...
       }
-    }
-    return line;
+    } while (r !== null && r.more);
+
+    return r === null ? null : line;
   }
 
   skipSpace(l: Uint8Array): number {
+    const charCodes: Array<number> = [CHAR_SPACE, CHAR_TAB];
```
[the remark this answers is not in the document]
SmashingQuasar: Will change, no problem.

textproto/mod.ts
SmashingQuasar: Personal discipline and preference really. Using pre-increment tends to lead to a safer code base than post increment. In this case it makes no difference. If the norm of this repository is to use post increment I have no problem changing it.

textproto/mod.ts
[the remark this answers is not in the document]
wperron: I don't have the context on the original comment so I'm not sure, @ry could provide more context

textproto/mod.ts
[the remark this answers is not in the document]
wperron: We typically let the type checker infer the types unless there's a specific reason not to, which isn't the case here

textproto/mod.ts
wperron: The post-increment form is widely used, and most people would consider it the "default"; using the pre-increment form here makes me ask myself "why?" "is there something I'm not seeing here?". I also have my doubts about it being safer, but that's secondary to the first point.

textproto/mod.ts
```diff
@@ -64,14 +60,14 @@ export class TextProtoReader {
     let buf = await this.r.peek(1);
     if (buf === null) {
       return null;
-    } else if (buf[0] == charCode(" ") || buf[0] == charCode("\t")) {
```
Liamolucko, line 67: Shouldn't this be `buf[0] == CHAR_SPACE || buf[0] == CHAR_TAB`?

Ditto for lines 74 and 109

textproto/mod.ts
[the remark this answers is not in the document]
SmashingQuasar: I changed it.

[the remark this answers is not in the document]
SmashingQuasar: Regarding tests, they are purely written in TypeScript, they do not use Deno.test as I am unfamiliar with it. I can submit them in another PR as it may take time to convert them.

textproto/mod.ts
SmashingQuasar: Excellent catch, I even improved this part by creating a constant as this expression is used repeatedly.

textproto/mod.ts
SmashingQuasar: I changed it to `n++` as it seems to bother you (and thus may bother other people).
A PR comment section does not seem to be the proper place to have a debate about post-increment and pre-increment but I would be interested in exchanging opinions on the subject. You can DM me on Reddit or e-mail me about it. 👍

SmashingQuasar:
I believe I addressed all requested changes. If there is anything else to change, please let me know.

textproto/test.ts
```diff
@@ -189,3 +189,66 @@ Deno.test({
     assertEquals(line, input);
   },
 });
+
+/*
+Test: 1 null
```
wperron, line 194: How do these work? What are the expected/actual values? I'm assuming these are valuable tests, so I'm interested in having them run on `Deno.test` like the rest of the test suite. As per your other comment, I understand you're unfamiliar with Deno's test runner; Feel free to contact me directly by email or on Discord, or ask around on our community server at https://discord.gg/deno

SmashingQuasar: This is a comment that I added accidentally, my bad. Basically @Zamralik designed a test to compare the output of the old code to the output of the new code to ensure behavior was the same. I am adapting those test to verify that the output stays the same. This means I will input some determined values and compare the output to a list of expected determined values.
Those take time to write, if it should be part of this PR, then I will need some time to adapt them.

SmashingQuasar:
I added unit tests for this PR, courtesy of @Zamralik for helping me transpile them into Deno compliant tests.
I believe this PR is done, except if anyone request any other change.
If all lights are green, maybe it could be merged?

textproto/test.ts
wperron: You can remove these comments now the tests have been added.

SmashingQuasar: Good catch, just did it.

wperron (approved):
LGTM thanks

PR #861, opened 2021-04-16
fix(http,ws): conn typings on internal utils

wperron:
Update the return type from `Deno.Conn` to `Deno.Conn<Deno.NetAddr>` on
two of the internal functions used to mock conn objects. The issue was
introduced in core at `9c7c9a35`.

In the case of `http/_mock_conn.ts@mockConn` I also had to add an
explicit type hint on the return object because of the `base` parameter
that was causing the type checker to not like the `transport` property
on `localAddr` and `remoteAddr`.

### DIFF
--- a/http/_mock_conn.ts
+++ b/http/_mock_conn.ts
@@ -1,7 +1,9 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 /** Create dummy Deno.Conn object with given base properties */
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM 👍

lucacasonato (changes requested):
I don't want to land this. This being broken points to the upstream patch in CLI to be majorly breaking. It will break all old std versions. It should be reverted. cc @kitsonk

PR #862, opened 2021-04-18
fix(node/fs): Actually export everything

Liamolucko:
There were a lot of `fs` apis which had been implemented, but not actually exported in the main `node/fs.ts`.

I also sorted everything, so it was easier to read the file tree and check everything was there.


kt3k (approved):
@Liamolucko LGTM. Thanks!

Will merge when main branch is fixed.

PR #864, opened 2021-04-20
feat(_util): Validate `Cookie` name , path and value

getspooky:
Moving `validateCookiePath` , `validateCookieValue` and  `validateCookieName` from `http/cookie` to `_util/rfc_6265`


getspooky:
@wperron it would be good idea to move rfc_6265 to `_utils`

wperron:
Respectfully disagree; the RFC 6265 cookie implementation is only used inside `http/cookie.ts` and I think it makes sense to keep it there, plus the new name of the file is quite obscure unless one already knows by heart that RFC 6265 is the RFC for HTTP state management.

PR #867, opened 2021-04-21
fix(io): Improve readDelims() performance

### LINKED ISSUE #803: bufio.readLines is pathologically slow for very long lines

keroxp:
- fix #803
- After short research, I found that problem comes from `Buffer` itself.
- Stopped using `Buffer` and introduced `BytesList` that manage multiple Uint8Array like one single Uint8Array.
- This change has significant performance improvements against main branch especially for a big file.

## Improvements

v2 (de049d6)

| Size | main (sec) | this PR (sec) |
| --- | --- | --- |
| 1kb | 0.001 | 0.0006 |
| 10kb | 0.002 | 0.002 |
| 100kb | 0.008 | 0.005 |
| 1mb | 0.467 | 0.062 |
|10mb | 46.962 | 0.330 |
| 100mb | N/A (over 45min) | 3.407 |

v1 (59b3c76)

| Size | main (sec) | this PR (sec) |
| --- | --- | --- |
| 1kb | 0.001 | 0.002 |
| 10kb | 0.002 | 0.004 |
| 100kb | 0.008 | 0.01 |
| 1mb | 0.467 | 0.1 |
|10mb | 46.962 | 1.295 |
| 100mb | N/A (over 45min) | 15.791 |


bytes/bytes_list_test.ts
```diff
@@ -0,0 +1,137 @@
...
+    bytes.equals(
+      arr.concat(),
+      new Uint8Array([
+        2,
+      ]),
+    ),
+    true,
+  );
+});
+Deno.test("[bytes] BytesList.shift 3", () => {
```
ry, line 131: shift 100?

keroxp: It's simply ordinal number.

keroxp:
@ry Got more improvements by using iterator.

ry (approved):
LGTM

PR #873, opened 2021-04-24
feat(std/node): Added implementation of os.homedir()

arwtyxouymz:
I found the implementation of `os.homedir()` was lack.
This is related to https://github.com/denoland/deno/issues/3802


node/os.ts
```diff
@@ -136,7 +136,15 @@ export function getPriority(pid = 0): number {
 
 /** Returns the string path of the current user's home directory. */
 export function homedir(): string | null {
-  notImplemented(SEE_GITHUB_ISSUE);
+  switch (Deno.build.os) {
+    case "windows":
+      return Deno.env.get("HOMEPATH") || null;
```
kt3k, line 141: I think 'USERPROFILE' is more reliable env var for pointing the home directory in windows. See this discussion for example https://github.com/denoland/deno/pull/3462

Node.js documentation also says it uses 'USERPROFILE' https://nodejs.org/api/os.html#os_os_homedir

kt3k (commented):
Thank you for your contribution. I left one comment.

arwtyxouymz:
@kt3k Thank you for your review. I updated to use `USERPROFILE` instead of `HOMEPATH`.

kt3k (approved):
LGTM 👍

Issue #872, opened 2021-04-24
std/node: assert.rejects is missing


uki00a:
`assert.rejects` is used in some [native node tests](https://github.com/denoland/deno_std/pull/695).

uki00a:
I'll work on this. :slightly_smiling_face:

PR #874, opened 2021-04-24
fix(node/assert): Enable test-assert-fail.js and align fail to it

uki00a:
While working on #872, I noticed that `assert.fail` is not implemented properly. So this PR implements `assert.fail` so that it passes `test-assert-fail.js`.


uki00a:
Blocked by #879 :slightly_smiling_face:

kt3k (approved):
@uki00a Thanks! LGTM.

PR #876, opened 2021-04-24
fix(node/child_process): Try to fix flaky tests

uki00a:
Some tests for `node/child_process` fails at 7498d9da5505011e31c7eea00eeebf7266fd9927. :scream: This PR tries to fix those flaky tests.


ry (approved):
LGTM - let's try

PR #877, opened 2021-04-24
fix(io): readDelim returns wrong result if delim strides over chunks

keroxp:
- This is bug from #867 
- @ry Sorry, would you see this?

### DIFF
--- a/io/bufio.ts
+++ b/io/bufio.ts
@@ -653,12 +653,11 @@ export async function* readDelim(
       return;
     }
     chunks.add(inspectArr, 0, result);
-    const itr = chunks.iterator(inspectIndex);
-    let curr = itr.next().value;
+    let localIndex = 0;
     while (inspectIndex < chunks.size()) {
-      if (curr === delim[matchIndex]) {
-        curr = itr.next().value;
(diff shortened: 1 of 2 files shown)


kfdf:
Nice to see the issue #803 is being handled, but as I mentiond before, the `createLPS` function is still bugged. This causes an infinite loop:
```
import { readStringDelim } from 'https://deno.land/std@0.95.0/io/mod.ts'
readStringDelim(null, '1213').next()
```
As may the use of very long delimiters (longer than 256 bytes).

keroxp:
@kfdf Oh, that's true! Also fixed on 82b327a

kfdf:
Wait, maybe I'm missing something, because the previous PR was benchmarked, but this doesn't actually fix #803?
```
import { readLines, StringReader } from 'https://deno.land/std@0.95.0/io/mod.ts'
let reader = new StringReader('X'.repeat(10_000_000))
for await (let _ of readLines(reader)) ;
```
This takes about 1.5 minutes for me, and the time grows quadraticaly with size.

kt3k (approved):
@keroxp Thank you for the fix! LGTM

@kfdf Thank you for your review!

PR #883, opened 2021-04-27
ci: disable codecov's patch status

kt3k:
codecov often reports failure status to the commit which has low coverage rate in diff, but we don't have such policy to reject such PRs. This PR disables that check from codecov.

ref: https://docs.codecov.io/docs/commit-status#disabling-a-status

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -3,8 +3,7 @@ codecov:
   require_ci_to_pass: true
 coverage:
   status:
-    project:
-      default:
-        informational: true
+    project: off
+    patch: off
 ignore:
   - examples


wperron (approved):
LGTM -- maybe we should start thinking about having a code coverage policy in the future? :thinking:

kt3k:
Maybe we can have some policy in the future, but the current coverage report doesn't seem 100% accurate for covered/non-covered lines. I think we need to fix it first.

caspervonb:
> but the current coverage report doesn't seem 100% accurate for covered/non-covered lines. I think we need to fix it first.

Yeah think my last fix was a regression as-well as a fix.

PR #884, opened 2021-04-28
feat(io): add readRange, readRangeSync

kitsonk:
I found I needed to read a range of a seekable file and felt that `readRange` and `readRangeSync` are good compliments to `readAll` and `realAllSync`, so I wanted to contribute them back to `std`.

This PR also cleans up the `Buffer()` init types to accept what is truly needed to init the buffer which eliminates the need for arbitrary type asserts.

### DIFF
--- a/io/util.ts
+++ b/io/util.ts
@@ -1,4 +1,6 @@
 import { Buffer } from "./buffer.ts";
+import { copy } from "../bytes/mod.ts";
+import { assert } from "../testing/asserts.ts";
 
 const DEFAULT_BUFFER_SIZE = 32 * 1024;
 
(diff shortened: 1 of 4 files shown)


kt3k (approved):
LGTM. Nice feature to have 👍

io/util.ts
```diff
@@ -53,6 +55,84 @@ export function readAllSync(r: Deno.ReaderSync): Uint8Array {
...
+  assert(length > 0, "Invalid byte range was passed.");
+  await r.seek(range.start, Deno.SeekMode.Start);
+  const result = new Uint8Array(length);
+  let off = 0;
+  while (length) {
+    const p = new Uint8Array(Math.min(length, DEFAULT_BUFFER_SIZE));
+    const nread = await r.read(p);
+    assert(nread !== null, "Unexpected EOF reach while reading a range.");
+    assert(nread > 0, "Unexpected read of 0 bytes while reading a range.");
+    copy(p, result, off);
```
jabr, line 93: Why create and a copy from a new buffer here rather than just read into the `result` buffer directly?

```ts
const nread = await r.read(result.subarray(off))
assert(...)
assert(...)
```

PR #887, opened 2021-04-29
feat(fmt/colors): allow formatters to be used as tagged templates

### LINKED ISSUE #886: [Feature Request] Support Tagged Template Literals for `std/fmt/colors.ts`
Currently, we can customize the color of text via the various functions exported from [https://deno.land/std@0.95.0/fmt/colors.ts](https://deno.land/std@0.95.0/fmt/colors.ts) like in the following example:

```ts
import { blue } from "https://deno.land/std@0.95.0/fmt/colors.ts";
console.log(blue("I. am. blue."));
```

However, it would be nice if we were able to just use tagged template literals also:

```ts
import { blue } from "https://deno.land/std@0.95.0/fmt/colors.ts";
console.log(blue`I. am. blue.`));
```

~This would require us to include a separate file which mirrors the `colors.ts`'s API that is compatible with tagged template literals, though.~

sno2:
Closes #886

I know I made it b4 anyone else talked on the issue, but I made the higher-order function and I just couldn't stop myself from going further...

### DIFF
--- a/fmt/colors.ts
+++ b/fmt/colors.ts
@@ -428,7 +428,10 @@ export function rgb24(str: string, color: number | Rgb): string {
   if (typeof color === "number") {
     return run(
       str,
-      code([38, 2, (color >> 16) & 0xff, (color >> 8) & 0xff, color & 0xff], 39)
+      code(
+        [38, 2, (color >> 16) & 0xff, (color >> 8) & 0xff, color & 0xff],
+        39,
+      ),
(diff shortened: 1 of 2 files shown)


nayeemrmn:
This is not what tagged template functions are for... it's not just so you can skip the parentheses on any function that accepts a string.

sno2:
See #886 for more info.

PR #888, opened 2021-04-29
feat(std/node): add writeBuffer of internal binding fs module

kt3k:
This PR implements `writeBuffer` function (sync part) in node.js internal binding fs module (`internalBinding("fs").writeBuffer(...)`.

This function is used in the `net.Socket` implementation ([ref](https://github.com/nodejs/node/blob/0869b82/lib/internal/net.js#L53), related #694). `net.Socket` is the parent class of `tty.WriteStream` and process.stdout is an instance of `tty.WriteStream`.


bartlomieju (approved):
LGTM 👍

PR #892, opened 2021-04-30
feat(node): Add console module

Liamolucko:
This just exports the `console` object.

I tried enabling the Node console tests, but all of them either don't work or don't really apply.

Note that Deno's console doesn't have all the same methods as Node's, so this isn't a perfect polyfill. Node has these extra properties:
- `Console`
- `context`
- `profile`
- `profileEnd`
- `timeStamp`

I've marked it as partly implemented for now, although I don't know whether those APIs are worth polyfilling.

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -12,7 +12,7 @@ deno standard library as it's a compatibility module.
 - [x] buffer
 - [x] child_process _partly_
 - [ ] cluster
-- [ ] console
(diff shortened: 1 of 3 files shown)


kt3k (approved):
@Liamolucko LGTM. Thanks!

I didn't know this is a module in node.js.

PR #895, opened 2021-05-01
feat(multipart): add options to readForm

### LINKED ISSUE #672: Make location of tempFiles created by MultipartReader configurable
Quoting.

> Related to (big) files in multipartrequest, I have noticed that tempFiles are created at ./, hardcoded.

```
const { file, filepath } = await tempFile(".", {
    prefix: "multipart-",
    postfix: ext,
});

```
> Wouldn't it be worth it to add a parameter for this in the readForm method (and default to `"."` for bw compat if wanted)? Possibly even an option for the prefix.

@npup I think it's worth adding an options argument, but definitely not on this PR. If you open an issue I can work on it after this get merged :)

_Originally posted by @marcosc90 in https://github.com/denoland/deno/pull/4865#issuecomment-621241394_

zebp:
Closes #672 

As mentioned in #672 there is current no way to where and how temporary files created by the reader are stored. In some cases storing temporary files in the current non-configurable location (the current working directory) is not possible or unacceptable for other reasons.

### DIFF
--- a/mime/multipart.ts
+++ b/mime/multipart.ts
@@ -259,6 +259,17 @@ export interface MultipartFormData {
   removeAll(): Promise<void>;
 }
 
+/**
+ * options for reading forms.
+ * @property maxMemory - maximum memory size to store file in memory. bytes.
+ * @default 10485760 (10MB)
+ * @property dir - directory where files that don't fit into maxMemory will be
+ * stored.
+ * @property prefix - a prefix that will be used for all files created if
+ * maxMemory is exceeded.
+ * @property suffix - a suffix that will be used for all files created if
+ * maxMemory is exceeded, defaults to the fole extension
+ */
 export interface ReadFormOptions {
   maxMemory?: number;
   dir?: string;
--- a/mime/multipart.ts
+++ b/mime/multipart.ts
@@ -294,6 +294,14 @@ export class MultipartReader {
   }
 
   /** Read all form data from stream.
+   * If total size of stored data in memory exceed maxMemory,
+   * overflowed file data will be written to temporal files.
+   * String field values are never written to files.
+   * null value means parsing or writing to file was failed in some reason.
+   * @param maxMemory maximum memory size to store file in memory. bytes. @default 10485760 (10MB)
+   *  */
+  async readForm(maxMemory?: number): Promise<MultipartFormData>;
+  /** Read all form data from stream.
    * If total size of stored data in memory exceed options.maxMemory,
    * overflowed file data will be written to temporal files.
    * String field values are never written to files.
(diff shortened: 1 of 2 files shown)


mime/multipart.ts
```diff
@@ -259,6 +259,13 @@ export interface MultipartFormData {
   removeAll(): Promise<void>;
 }
 
+export interface ReadFormOptions {
+  maxMemory?: number;
+  dir?: string;
+  prefix?: string;
+  suffix?: string;
+}
```
kt3k, line 267: It might be great if these properties are documented using this syntax https://github.com/denoland/deno_std/blob/main/encoding/ascii85.ts#L5-L9

kt3k:
`dir: "."` is added here https://github.com/denoland/deno/pull/8032/files#diff-6d8e556628e1016bd0b5b2b4ae879a717cb1401d0bf15c0b1e916319ae66b8d6R313
but it wasn't mentioned in the PR description. This might have been a mistake? cc @caspervonb

mime/multipart.ts
```diff
@@ -276,13 +283,15 @@ export class MultipartReader {
...
+   * If total size of stored data in memory exceed options.maxMemory,
    * overflowed file data will be written to temporal files.
    * String field values are never written to files.
    * null value means parsing or writing to file was failed in some reason.
-   * @param maxMemory maximum memory size to store file in memory. bytes. @default 10485760 (10MB)
+   * @param options options to configure the behavior of storing
+   * overflow file data in temporal files.
    *  */
-  async readForm(maxMemory = 10 << 20): Promise<MultipartFormData> {
+  async readForm(options?: ReadFormOptions): Promise<MultipartFormData> {
```
kt3k, line 293: To protect the current usage of this method, we'd like to support the current signature as well. I think we can use overload feature like the below to support both the current and suggested interface:

```ts
  async readForm(maxMemory: number): Promise<MultipartFormData>;
  async readForm(options?: ReadFormOptions): Promise<MultipartFormData>;
  async readForm(maxMemoryOrOptions: number | ReadFormOptions): Promise<MultipartFormData> {
    ...
  }
```

zebp: How should the documentation reflect this? Should `@param` last function definition?

mime/multipart.ts
zebp: [This commit](https://github.com/denoland/deno_std/pull/895/commits/5614f5f986d62000d99694de60a4f4486eace8d8) should fix this, but I don't know if this documentation is still subpar.

kt3k (approved):
@zebp Thanks for updating! LGTM.

PR #899, opened 2021-05-04
chore(io/buffer): Buffer should explicitly implement Reader & Writer

uki00a:
This PR adds `implements` declaration to `Buffer` class.

Originally, [`Deno.Buffer` class implemented `Deno.Reader` and `Deno.Writer`](https://github.com/denoland/deno/blob/v1.2.0/cli/js/buffer.ts). It seems that this `implements` declaration have been lost in migration from TypeScript to JavaScript. (https://github.com/denoland/deno/blob/v1.3.0/cli/rt/13_buffer.js#L29)

### DIFF
--- a/io/buffer.ts
+++ b/io/buffer.ts
@@ -23,7 +23,7 @@ const MAX_SIZE = 2 ** 32 - 2;
  *
  * Based on [Go Buffer](https://golang.org/pkg/bytes/#Buffer). */
 
-export class Buffer {
+export class Buffer implements Deno.Reader, Deno.Writer {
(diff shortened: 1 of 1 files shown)


lucacasonato:
Thanks for the PR, but this is intentional. We want the implementation to be usable in the DOM without the Deno namespace.

lucacasonato:
The implementation is implicit isn't it? This has no effect on usage.

uki00a:
> Thanks for the PR, but this is intentional.

Oh, I did not understand it :cry: 

> We want the implementation to be usable in the DOM without the Deno namespace.

It certainly makes sense :+1: So I'll close this PR. Thanks for the review!

PR #900, opened 2021-05-04
encoding/TOML: delete useless replace

zekth:
Deleting double use of `string.replace` speeding up the parsing.

### DIFF
--- a/encoding/testdata/cargo.toml
+++ b/encoding/testdata/cargo.toml
@@ -49,3 +49,6 @@ url = "1.7.2"
 
 [target.'cfg(windows)'.dependencies]
 winapi = "0.3.6"
+
+[target."cfg(linux)".dependencies]
+winapi = "0.3.9"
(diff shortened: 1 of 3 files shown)


kt3k:
Tried the benchmark like the below, parsing of a toml which contains 1000-line multiline string literal.

```js
import { parse } from "./encoding/toml.ts";

const toml = "data = '''" + [...Array(1000)].map(() => 'foo').join('\n') + "'''"

const timeStarted = Date.now();
for (const _ of Array(1000)) {
  parse(toml);
}
const timeEnded = Date.now();

console.log(`elapsed time: ${timeEnded - timeStarted}ms`);
```

`520~530ms` in main branch, `470~480ms` in this branch consistently. I confirmed this is actually an improvement of parsing of large string literals.

kt3k (approved):
@zekth Thank you for looking into this. LGTM

PR #901, opened 2021-05-05
fix(node/stream): Make `Stream` the default export

Liamolucko:
In Node, `require("stream")` returns `Stream` itself with all the other kinds of streams as properties. This maps to the default export for ESM, so I've changed the default export of `stream.ts` to be `Stream` and added all the other streams as properties.


node/stream_test.ts
```diff
@@ -131,3 +140,7 @@ Deno.test("Readable can be piped through Transform", async () => {
   await flushExecution;
   await readableExecution;
 });
+
+Deno.test("The default export is Stream", () => {
+  assertStrictEquals(streamDefault, Stream);
+});
```
kt3k, line 146: 👍

kt3k (approved):
LGTM. Thanks, @Liamolucko

PR #906, opened 2021-05-08
fix(testing): support array values in assertObjectMatch

lowlighter:
The filtered subset in `assertObjectMatch` is built with the following:

https://github.com/denoland/deno_std/blob/69183a05261b58870870b48e4191660b49212039/testing/asserts.ts#L502-L503

But because `typeof [] === "object"`, array values match this condition and we end up with the following:
```diff
    {
-     foo: {
-       "0": 1,
-       "1": 2,
-       "2": 3,
-       length: 3,
-     },
+     foo: [
+       1,
+       2,
+       3,
+     ],
    }
```

This add a skip for arrays to get the wanted behaviour

### DIFF
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -505,6 +505,19 @@ Deno.test("testingAssertObjectMatching", function (): void {
     let didThrow;
     try {
       assertObjectMatch(i, {
+        foo: [1, 2, 3, 4],
+      });
+      didThrow = false;
+    } catch (e) {
+      assert(e instanceof AssertionError);
+      didThrow = true;
+    }
+    assertEquals(didThrow, true);
+  }
+  {
+    let didThrow;
+    try {
+      assertObjectMatch(i, {
         foo: [{ bar: true }, { foo: false }],
       });
       didThrow = false;
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -322,6 +322,7 @@ Deno.test("testingAssertObjectMatching", function (): void {
   }
   const g: r = { foo: true, bar: false };
   const h = { foo: [1, 2, 3], bar: true };
+  const i = { foo: [a, e], bar: true };
 
   // Simple subset
   assertObjectMatch(a, {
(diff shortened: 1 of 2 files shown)


testing/asserts_test.ts
```diff
@@ -487,6 +490,20 @@ Deno.test("testingAssertObjectMatching", function (): void {
     }
     assertEquals(didThrow, true);
   }
+  // Subset with array inside but doesn't match key subset
+  {
+    let didThrow;
+    try {
+      assertObjectMatch(h, {
+        foo: [],
+      });
```
kt3k, line 499: I feel this might better to pass the assertion because everything in `{ foo: [] }` is contained in h (`{ foo: [1, 2, 3], bar: true }`). Maybe we should do some recursive check for array as well?

lowlighter: Yes it makes sense, will make the change 👍

lowlighter: I changed it so `{ foo: [1, 2, 3], bar: true }` matches `{ foo: [] }` as suggested
When specifying complex arrays like `{ foo: [ { bar: true, baz: false } ] }` it'll match both `{ foo: [ { bar:true } ] }` and `{ foo: [ { baz: false } ] }`

testing/asserts_test.ts
```diff
@@ -367,6 +369,17 @@ Deno.test("testingAssertObjectMatching", function (): void {
...
   });
+  // Subset with array inside
+  assertObjectMatch(h, { foo: [] });
+  assertObjectMatch(h, { foo: [1, 2] });
+  assertObjectMatch(h, { foo: [1, 2, 3] });
+  assertObjectMatch(i, { foo: [{ bar: false }] });
+  assertObjectMatch(i, {
+    foo: [
+      { bar: false },
+      { bar: { bar: { bar: { foo: true } } } },
```
kt3k, line 380: 👏

kt3k (approved):
@lowlighter Great! Thanks! LGTM

PR #908, opened 2021-05-10
fix(http/file_server): keep url and name consistent

### LINKED ISSUE #784: file_server directory view leaves trailing forward slash off directory URLs
When browsing the auto index created by file_server, directory entries are displayed with a trailing forward slash but the generated URLs do not include the trailing slash.

While this is probably an opinion rather than a bug, the behaviour is surprising and inconsistent with (at least) the automatic indexes generated by:
- Apache
- NGINX
- Python's SimpleHTTPServer
- mod_autoindex for Express
- IIS
- Caddy

The main impact of this is that this breaks relative URLs when the directory contains an index.html.

The fix appears to be relatively trivial, replacing https://deno.land/std@0.89.0/http/file_server.ts#L182 with something similar to the treatment given to the name on the previous line - that is, adding a forward slash to the URL if it's a directory.

zhmushan:
fix #784

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -179,7 +179,7 @@ async function serveDir(
       mode: modeToString(entry.isDirectory, fileInfo.mode),
       size: entry.isFile ? fileLenToString(fileInfo.size ?? 0) : "",
       name: `${entry.name}${entry.isDirectory ? "/" : ""}`,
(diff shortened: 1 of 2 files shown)


wperron (approved):
LGTM thanks :slightly_smiling_face:

PR #909, opened 2021-05-10
feat(http): Validate Cookie Domain rfc6265#section-4.1.2.3

getspooky:
Validate cookie domain to match [ rfc6265#section-4.1.2.3](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.3)

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -84,6 +84,7 @@ function toString(cookie: Cookie): string {
 /**
  * Validate Cookie Name.
  * @param name Cookie name.
+ * @returns {void}
  */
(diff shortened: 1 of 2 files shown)


getspooky:
@kt3k i just add unit testing for cookie domain validation.

kt3k:
@getspooky Sorry for the delay in the review. I think the RFC defines these values are ignored by the user agents (=browsers) but it doesn't forbid servers setting these values in responses. So throwing here doesn't look correct to me.

Closing without merge, but feel free to reopen it if I miss something or you have another point to add.

PR #910, opened 2021-05-10
feat(http): Gets the body raw of the message

getspooky:
Extract the entire body portion of an incoming request stream and exposes it `readRaw` method.


bartlomieju:
@getspooky what's the use case for this API?

getspooky:
I think this PR is not useful for deno version `1.13.1`.

PR #912, opened 2021-05-11
fix(io/buffer): make Buffer compatible with Deploy

### LINKED ISSUE #907: Make `encoding/yaml.ts` type-compatible with Deno Deploy
Currently yaml can't be used in Deploy because of the following errors:

```shellsession
$ deployctl run test.ts 
Check file:///Users/kt3k/tmp/deploy-test/$deno$eval.ts
error: TS2694 [ERROR]: Namespace 'Deno' has no exported member 'Reader'.
  async readFrom(r: Deno.Reader): Promise<number> {
                         ~~~~~~
    at https://deno.land/std@0.95.0/io/buffer.ts:196:26

TS2694 [ERROR]: Namespace 'Deno' has no exported member 'ReaderSync'.
  readFromSync(r: Deno.ReaderSync): number {
                       ~~~~~~~~~~
    at https://deno.land/std@0.95.0/io/buffer.ts:226:24

Found 2 errors.
error: Process exited with code 1
```

kt3k:
This PR fixes type error of `io/buffer` in Deploy environment by inlining `Reader` / `ReaderSync` interafaces.

This enables deploy script using, for example, `encoding/yaml` module without type error with `deployctl run`. https://gist.github.com/kt3k/43e25efffc6e1c360fc64274ea0381eb

closes #907

### DIFF
--- a/io/buffer.ts
+++ b/io/buffer.ts
@@ -1,13 +1,7 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import { assert } from "../_util/assert.ts";
 import { copy } from "../bytes/mod.ts";
-
-interface Reader {
-  read(p: Uint8Array): Promise<number | null>;
-}
-
-interface ReaderSync {
-  readSync(p: Uint8Array): number | null;
-}
+import type { Reader, ReaderSync } from "./types.d.ts";
 
 // MIN_READ is the minimum ArrayBuffer size passed to a read call by
 // buffer.ReadFrom. As long as the Buffer has at least MIN_READ bytes beyond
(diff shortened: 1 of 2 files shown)


io/buffer.ts
```diff
@@ -1,6 +1,14 @@
 import { assert } from "../_util/assert.ts";
 import { copy } from "../bytes/mod.ts";
 
+interface Reader {
```
kitsonk, line 4: does it make sense to have a `io/types.d.ts` and put the `Reader*`/`Writer*`, etc and then `import type` from there?  It would solve part of https://github.com/denoland/deno/issues/9795.  I wouldn't necessarily refactor all of `io` to use them at the moment, but it would be better than just burying them in `buffer.ts` while all of `io` and other `std` modules should move over to these.

kt3k: That would make more sense and sounds natural. Updated.

kitsonk (approved):
LGTM

PR #914, opened 2021-05-11
chore: add coverage directories to gitignore

bartlomieju:
After collecting coverage following directories are produced:
```
	cov/
	examples/cov/
	fs/testdata/cov/
	node/_tools/cov/
	node/cov/
```

Added them to `.gitignore`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ deno.d.ts
 package.json
 package-lock.json
 .vscode/settings.json
+**/cov/


caspervonb:
Its because they're changing cwd and the collection path is relative. Have a patch for it already just pending adding a test case.

caspervonb:
See https://github.com/denoland/deno/pull/10364

kt3k (approved):
Thanks! LGTM

Issue #911, opened 2021-05-10
test(node): Enable more native node tests


uki00a:
In #695, the native node test runner has been introduced. Currently, only a very limited amount of tests is enabled. For better compatibility, I think more tests should be enabled.

The following list shows the modules for which native node tests should be enabled:

- [x] assert (_partly_)
- [ ] buffer
- [ ] child_process (_partly_)
- [ ] console
- [ ] constants
- [ ] crypto
- [x] events (_partly_)
- [x] fs (_wip_)
- [ ] module
- [ ] os
- [x] path
- [ ] process
- [x] querystring (_partly_)
- [x] stream
- [ ] string_decoder
- [ ] timers
- [ ] tty
- [x] url (_partly_)
- [x] util (_partly_)

uki00a:
I'm working on `events` :raising_hand_man:

wafuwafu13:
I'm working on `url` 🙋‍♂️

wafuwafu13:
`url` partly done

- [ ]  `test-url-domain-ascii-unicode.js`: TODO because https://github.com/denoland/deno_std/pull/1365#discussion_r724875257
- [x] `test-url-fileurltopath.js`
- [ ] `test-format-invalid-input.js`: Legacy https://nodejs.org/api/url.html#url_url_format_urlobject
- [ ] `test-format-whatwg.js`: Legacy
- [ ] `test-url-format.js`: Legacy
- [ ] `test-url-parse-format.js`: Legacy https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost
- [ ] `test-url-parse-invalid-input.js`: Legacy
- [ ] `test-url-parse-query.js`: Legacy
- [x] `test-url-pathtofileurl.js`
- [ ] `test-url-relative.js`: Legacy https://nodejs.org/api/url.html#url_url_resolve_from_to
- [ ] `test-url-urltooptions.js`: TODO because Add in v15.7.0 https://nodejs.org/api/url.html#url_url_urltohttpoptions_url

wafuwafu13:
`querystring` partly done and working on `util`

F3n67u:
I am working on `fs`.

wafuwafu13:
`util` partly done and working on `path`.

wafuwafu13:
`path` done and working on `fs`

PolarETech:
Hi,
I am working on `fs.ReadStream` (#2653) and `child_process.exec`.

[...]

PR #916, opened 2021-05-12
fix(node/events): Accept only functions as listener arguments

uki00a:
Part of #911

This PR enables `test-event-emitter-invalid-listener.js` and aligns `node/events` to it.


kt3k (approved):
@uki00a Thanks! LGTM

PR #917, opened 2021-05-13
chore(node/_tools): make setup.ts more efficient

uki00a:
Related to #911

Currently, `node/_tools/setup.ts` does the following things:

1. Download a node.js source archive (This step is skippable)
2. Extract the test files from the archive (This step is **NOT** skippable)

The step 2 above takes a certain amount of time...:cry: Therefore, this PR changes the script as follows:

1. Download a node.js source archive (This step is skippable)
2. Extract the test files from the archive and cache them to the local disk  (This step is skippable)
3. Copy the cached test files to `/node/_tools/suites` directory (This step is not skippable)


kt3k (approved):
@uki00a Thanks! Nice improvement. LGTM

PR #919, opened 2021-05-16
feat(async): add async/tee

kt3k:
This PR adds the utility (`tee`) for branching the given async iterable into n copies.

This has been originally developed in https://github.com/kt3k/tee_async_iterable as part of file stream processing tool (which is not published yet).

### DIFF
--- a/async/README.md
+++ b/async/README.md
@@ -83,3 +83,31 @@ for await (const value of results) {
   // ...
 }
 ```
+
+## tee
+
+Branches the given async iterable into the n branches.
+
+```typescript
+import { tee } from "https://deno.land/std/async/tee.ts";
+
+const gen = async function* gen() {
+  yield 1;
+  yield 2;
+  yield 3;
+};
+
+const [branch1, branch2] = tee(gen());
+
+(async () => {
+  for await (const n of branch1) {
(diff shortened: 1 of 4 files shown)


ry (approved):
LGTM

exe-dealer:
This code is very fragile and buggy to be in std. It violates async iterator protocol

.next method should be able to queue concurent calls

```js
import { tee } from 'https://deno.land/std@0.106.0/async/tee.ts';

async function * gen() {
  yield 1;
  yield 2;
}

const [left, right] = tee(gen());

// sequential calls works fine
console.log([await right.next(), await right.next()]);
// [ { value: 1, done: false }, { value: 2, done: false } ]

// should be able to enqueue concurent .next calls
console.log(await Promise.all([ left.next(), left.next() ]));
// [ { value: 1, done: false }, { value: 1, done: false } ]
// but it returns the same result twice

```




guys are you serious?

https://github.com/denoland/deno_std/blob/f519c159f23a3b2b2b14aa205c7d1af9ae0aacca/async/tee.ts#L98-L100

.return() should be called on source iterator to terminate iteration when all brunches are .return()'ed

and if error occurs during source.next() then error should be propagated to branches

Issue #920, opened 2021-05-16
std/io/bufio - `readLines` and `readStringDelim` should allow to specify the character encoding


carragom:
Hi,
This is an improvement proposal. It should somehow be possible to tell [readLines](https://deno.land/std@0.96.0/io/bufio.ts#L703) and [readStringDelim](https://deno.land/std@0.96.0/io/bufio.ts#L691) which encoding the `Reader` data actually have. That way we could easily read files by line from any supported encoding.
The idea would be to somehow parameterize both functions in order to affect the creation of the `TextDecoder` [here](https://deno.land/std@0.96.0/io/bufio.ts#L696).

What do you think?

PR #921, opened 2021-05-16
std/io/bufio - Parameterize input TextEncoder in `readLines` and `readStringDelim`

carragom:
Here is an attempt to implement what is proposed on #920, please let me know what you think.


kt3k:
@carragom Thank you for your contribution! I think this is a nice addtion to the API.

Can you run `deno fmt` command to the changed files?

carragom:
Hi, thanks for the review, I formatted the files. But the reason the tests are failing is because the new test requires read access to be able to open the file with the encoded data. Not sure where to change that.

kt3k (approved):
@carragom Thanks! LGTM

kt3k:
Note: mdn doesn't seem documenting `ignoreBOM` option? ref: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder . But it's documented in whatwg spec https://encoding.spec.whatwg.org/#interface-textdecoder , and deno actually implements it.

PR #923, opened 2021-05-17
feat(async/mux): takes AsyncIterable as source iterator

kt3k:
Currently MuxAsyncIterator takes `AsyncIterableIterator<T>` as source of iterator, but `AsyncIterable<T>` is a more general form of async iterable data. This PR changes the signature of `add` method of MuxAsyncIterator from `add(iterator: AsyncIterableIterator<T>)` to `add(iterable: AsyncIterable<T>)` and makes it work with `AsyncIterable`s.


ry (approved):
LGTM

Issue #925, opened 2021-05-17
std/datetime/mod `dayOfyear` is incorrect for years with daylight savings


DurableMicron:
Tried to run tests for the first time and noticed that this test fails. https://github.com/denoland/deno_std/blob/241d2d2783607ffd83cc459118f565fcfdd945f3/datetime/test.ts#L167

```
failures:

[std/datetime] dayOfYear
AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   90
+   91
```

the source code uses the time offset at the start of the year compared to the actual date and there was a daylight savings change in my timezone between the start of the year and the source code ( i believe) is trying to address that

```    (yearStart.getTimezoneOffset() - date.getTimezoneOffset()) * 60 * 1000;
```
however, since my timezone is negative (according to `getTimezoneOffset()`) this seems to have the affect of breaking the functionality.

kt3k:
@zekth Do you have any idea about these `.getTimezoneOffset()` handlings?

PR #926, opened 2021-05-17
fix(datetime) dayOfYear now works for negative timezones

DurableMicron:
see issue in #925 

Added a math.abs so that negative timezones work like positive ones. Unsure if this has any other consequences as removing this line entirely also caused the tests to pass. (but im unsure if theres timezone consequences to this, maybe a test should be added for a specific timezone boundary).

### DIFF
--- a/datetime/mod.ts
+++ b/datetime/mod.ts
@@ -56,8 +56,7 @@ export function dayOfYear(date: Date): number {
   yearStart.setUTCFullYear(date.getUTCFullYear(), 0, 0);
   const diff = date.getTime() -
     yearStart.getTime() +
-    (yearStart.getTimezoneOffset() - date.getTimezoneOffset()) * 60 * 1000;
-
+    (Math.abs(yearStart.getTimezoneOffset()) - Math.abs(date.getTimezoneOffset())) * 60 * 1000;
   return Math.floor(diff / DAY);
 }
 /**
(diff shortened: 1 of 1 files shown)


zekth:
My wondering in here is more about the test rather than the function itself. The test doesn't use UTC time, so it can change per environment it's running.
Could you describe more the case you're facing with the values you're getting for timezone offset? Timezones and such.
eg my values (Paris Time)
```javascript
new Date().getTimezoneOffset(); // outputs: -120
```

DurableMicron:
> My wondering in here is more about the test rather than the function itself. The test doesn't use UTC time, so it can change per environment it's running. (we had similar issue in the past)
> Could you describe more the case you're facing with the values you're getting for timezone offset? Timezones and such.
> eg my values (Paris Time)
> 
> ```js
> new Date().getTimezoneOffset(); // outputs: -120
> ```

The issue is that if you have a negative timezone, and you try and apply the day of year function during a year in which your timezone offset changed (ie daylight savings) then the calculation is a day off.

wperron:
Testing locally, it passes the tests when running in GMT+4 but fails in GMT-1, specifically on the assertion on line 161 (in the `dayOfYear` test)

DurableMicron:
> Testing locally, it passes the tests when running in GMT-4 but fails in GMT+1, specifically on the assertion on line 161 (in the `dayOfYear` test), but both passes when I'm checkout out on main. I have a feeling the `dayOfYear` calculation is actually correct.

The original fails in GMT + 8:00 (Perth), ill try that test case out

bartlomieju:
@DurableMicron thanks you for the PR, the code looks okay to me, but we really should add some tests that exercise this change.

PR #932, opened 2021-05-22
fix(encoding/binary): allow getNBytes to read until EOF

### LINKED ISSUE #651: `getNBytes` throws EOF when it should wait for more input
The `encoding/binary` function [`getNBytes`](https://deno.land/std@0.68.0/encoding/binary.ts#L45-L56) gives up too easily — EOF is not the only case when [`Deno.Reader#read`](https://doc.deno.land/builtin/v1.3.3#Deno.Reader) can read less than `n` bytes:

> If some data is available but not `p.byteLength` bytes, `read()` conventionally resolves to what is available instead of waiting for more.

It is possible to expose this issue through higher-level `encoding/binary` functions like `readVarnum` as well:

```console
$ cat binary_bug.ts
import { readVarnum } from "https://deno.land/std@0.68.0/encoding/binary.ts"
// Reads a 4-byte int32 varnum by default
console.log(await readVarnum(Deno.stdin));

$ echo -ne '\x01\x02\x03\x04' | xxd
00000000: 0102 0304                                ....

$ (echo -ne '\x01\x02'; sleep 1; echo -ne '\x03\x04') | xxd
00000000: 0102 0304                                ....

$ echo -ne '\x01\x02\x03\x04' | deno run binary_bug.ts
16909060

$ (echo -ne '\x01\x02'; sleep 1; echo -ne '\x03\x04') | deno run binary_bug.ts
error: Uncaught UnexpectedEof:
  if (nRead === null || nRead < n) throw new Deno.errors.UnexpectedEof();
                                         ^
    at getNBytes (binary.ts:54:42)
    at async readVarnum (binary.ts:204:19)
    at async binary_bug.ts:3:13
```

The [original Go code](https://golang.org/src/encoding/binary/varint.go#L106) avoids this problem by manually reading one byte at a time, but we could handle it more cleanly with a new `Deno.readExact` function inspired by Go's [`readFull`](https://golang.org/pkg/io/#ReadFull) (and/or `readAtLeast`) and Rust's [`read_exact`](https://github.com/rust-lang/rust/blob/5a6b426e3471382c0385c11b09c2d6b37f70ac49/library/std/src/io/mod.rs#L770-L787)<sup>1</sup>.

(I suggest `readExact` as the name because `readFull` sounds too much like a synonym for the already existing `Deno.readAll`.)

Then, `getNBytes` could be reimplemented as...

```ts
export async function getNBytes(
  r: Deno.Reader,
  n: number,
): Promise<Uint8Array> {
  const scratch = new Uint8Array(n);
  await Deno.readExact(r, scratch);
  return scratch;
}
```

...while other potential users could take advantage of the buffer re-use that `readExact` enables.

---

<sup>1</sup> There is even a [very detailed RFC](https://github.com/rust-lang/rfcs/blob/master/text/0980-read-exact.md) justifying the precise API and behaviour of `read_exact` in Rust.

hermesespinola:
## Changes

- Write `readExact` function for `encoding/binary`:
  - Read from `reader` until the `Uint8Array` array is filled.
  - If the reader reaches EOF, throws a `UnexpectedEof` error.
- Rewrite `getNBytes` to use `readExact`.

closes #651

### DIFF
--- a/encoding/binary.ts
+++ b/encoding/binary.ts
@@ -51,7 +51,7 @@ export async function readExact(
 ): Promise<void> {
   let totalRead = 0;
   do {
-    let tmp = new Uint8Array(b.length - totalRead);
+    const tmp = new Uint8Array(b.length - totalRead);
     const nRead = await r.read(tmp);
     if (nRead === null) throw new Deno.errors.UnexpectedEof();
     b.set(tmp, totalRead);
--- a/encoding/binary.ts
+++ b/encoding/binary.ts
@@ -42,7 +42,24 @@ export function sizeof(dataType: DataType): number {
   return rawTypeSizes[dataType];
 }
 
-/** Reads `n` bytes from `r`.
+/** Reads the exact number of bytes from `r` required to fill `b`.
+ *
+ * Throws `Deno.errors.UnexpectedEof` if `n` bytes cannot be read. */
+export async function readExact(
+  r: Deno.Reader,
+  b: Uint8Array,
+): Promise<void> {
+  let totalRead = 0;
+  do {
+    let tmp = new Uint8Array(b.length - totalRead);
+    const nRead = await r.read(tmp);
+    if (nRead === null) throw new Deno.errors.UnexpectedEof();
+    b.set(tmp, totalRead);
+    totalRead += nRead;
+  } while (totalRead < b.length);
+}
+
+/** Reads exactly `n` bytes from `r`.
  *
  * Resolves it in a `Uint8Array`, or throws `Deno.errors.UnexpectedEof` if `n` bytes cannot be read. */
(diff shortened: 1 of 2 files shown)


kt3k:
@hermesespinola Thanks for contributing! I have 2 comments.

1. Please run lint check to pass the CI
2. Can you create a test case in which it reads from a slowly readable reader to reproduce the case of #651 ?

I guess something like `readerFromIterable((async function *() { yield ...; await delay(); ... })())` would be usable for 2.

hermesespinola:
Hello @kt3k I fixed the linter warning and added the suggested test 😄.

kt3k (approved):
@hermesespinola Thanks for updating! LGTM. Nice fix 👍

PR #936, opened 2021-05-24
feat(testing): Adding beforeEach , afterEach, withHooks helper function

getspooky:
See discussion here denoland/deno#10741

```ts
import { beforeEach, afterEach, withHooks } from "http://deno.land/std/testing/hooks.ts";

beforeEach(async function() {
});

afterEach(async function() {
});

Deno.test("", withHooks(async function() {
});
```

### DIFF
--- a/testing/hooks.ts
+++ b/testing/hooks.ts
@@ -14,12 +14,12 @@ const hooks: Partial<Record<HookType, Function>> = {};
 function _addHook(fn: Function, hookType: HookType): void {
   if (typeof fn !== "function") {
     throw new TypeError(
-      "Invalid first argument. It must be a callback function."
+      "Invalid first argument. It must be a callback function.",
     );
   }
   if (typeof hookType !== "string") {
     throw new TypeError(
-      `Invalid second argument, ${hookType}. It must be a string`
+      `Invalid second argument, ${hookType}. It must be a string`,
     );
   }
   hooks[hookType] = fn;
@@ -28,21 +28,22 @@ function _addHook(fn: Function, hookType: HookType): void {
 /**
  * Helper function to execute registred hooks.
  */
-export const withHooks = (fn: () => void | Promise<void>) => () => {
-  if (typeof fn !== "function") {
-    throw new TypeError(
-      "Invalid first argument. It must be a callback function."
-    );
-  }
-  if (hooks.beforeEach) {
-    hooks.beforeEach();
-  }
-  // execute the callback function
-  fn();
-  if (hooks.afterEach) {
-    hooks.afterEach();
-  }
-};
+export const withHooks = (fn: () => void | Promise<void>) =>
+  () => {
+    if (typeof fn !== "function") {
+      throw new TypeError(
+        "Invalid first argument. It must be a callback function.",
+      );
+    }
+    if (hooks.beforeEach) {
+      hooks.beforeEach();
+    }
+    // execute the callback function
+    fn();
+    if (hooks.afterEach) {
+      hooks.afterEach();
+    }
+  };
 
 /**
  * INFO:
--- a/testing/hooks.ts
+++ b/testing/hooks.ts
@@ -24,3 +24,23 @@ function _addHook(fn: Function, hookType: HookType): void {
   }
   hooks[hookType] = fn;
 }
+
+/**
+ * Helper function to execute registred hooks.
+ */
+export const withHooks = (fn: () => void | Promise<void>) => () => {
+  if (typeof fn !== "function") {
+    throw new TypeError(
+      "Invalid first argument. It must be a callback function."
+    );
+  }
+  if (hooks.beforeEach) {
+    hooks.beforeEach();
+  }
+  // execute the callback function
+  fn();
+  if (hooks.afterEach) {
+    hooks.afterEach();
+  }
+};
+
--- a/testing/hooks.ts
+++ b/testing/hooks.ts
@@ -44,3 +44,17 @@ export const withHooks = (fn: () => void | Promise<void>) => () => {
   }
 };
 
+/**
+ * INFO:
+ * Runs a function before each of the tests.
+ * If the function returns a promise or is a generator,
+ * Deno aits for that promise to resolve before running the test.
+ * `fn` can be async if required.
+ */
+export const beforeEach: THook = (fn) => _addHook(fn, "beforeEach");
+/**
+ * INFO:
+ * Runs a function after each of the tests.
+ * `fn` can be async if required.
+ */
+export const afterEach: THook = (fn) => _addHook(fn, "afterEach");
--- a/testing/hooks.ts
+++ b/testing/hooks.ts
@@ -1,0 +1,26 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+type THook = (fn: Function) => void | Promise<void>;
+type HookType = "beforeEach" | "afterEach";
+
+const hooks: Partial<Record<HookType, Function>> = {};
+
+/**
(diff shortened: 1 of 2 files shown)


testing/hooks_test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+Deno.test("beforeEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  beforeEach(function () {
+    didInvoke = true;
+    assertEquals(/deno/, /deno/);
+  });
+  assertEquals(didInvoke, false);
+  withHooks(function () {
+    assertEquals(/cat/, /cat/);
```
kt3k, line 13: I think we should assert `assertEquals(didInvoke, true)` here to check `beforeEach` hook is actually called before the main function.

testing/hooks_test.ts
```diff
@@ -0,0 +1,33 @@
...
+
+Deno.test("afterEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  afterEach(function () {
+    try {
+      assertEquals(/deno/, /cat/);
+    } catch (e) {
+      assert(e instanceof AssertionError);
+      didInvoke = true;
+    }
```
kt3k, line 26: This try-catch structure looks meaningless to me. It can be replaced by simply `didInvoke = true`. And we should check didInvoke value after `withHooks(...)` block to check wether it's actually called 'after' the main function.

getspooky: yes, i will remove try-catch structure

testing/hooks_test.ts
getspooky: Yes , I will `assertEquals(didInvoke, true)` inside withHooks

testing/hooks_test.ts
```diff
@@ -0,0 +1,27 @@
...
+Deno.test("afterEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  afterEach(function () {
+    didInvoke = true;
+    assertEquals(didInvoke, true);
+  });
+  assertEquals(didInvoke, false);
+  withHooks(function () {
+    assertEquals(/cat/, /cat/);
+  })();
```
kt3k, line 26: I think we should check didInvoke === false inside the main function, and check didInvoke === true after the main, to ensure the the hook is actually called after the main func. It might look something like the below

```suggestion
  afterEach(function () {
    didInvoke = true;
  });
  assertEquals(didInvoke, false);
  withHooks(function () {
    assertEquals(didInvoke, false);
  })();
  assertEquals(didInvoke, true);
```

getspooky: It will return false unless we remove `async` and `await` from `withHooks` or adding await before 
```ts
await withHooks(function () {   
 assertEquals(didInvoke, false);  })();
}
```

testing/hooks.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+type THookFn = () => void | Promise<void>;
+type THook = (fn: THookFn) => void | Promise<void>;
+type HookType = "beforeEach" | "afterEach";
+
+const hooks: Partial<Record<HookType, THookFn>> = {};
```
caspervonb, line 7: This is needlessly complicated.
Just define an object with well known members to contain the state.
Also should probably be arrays.

```suggestion
const hooks = {
  beforeEach: [],
  afterEach: [],
};
```

testing/hooks.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Register new testing hooks.
+ *
+ * @param {Function} fn
+ * @param {HookType} hookType
+ */
+function _addHook(fn: THookFn, hookType: HookType): void {
```
caspervonb, line 15: Inline this

testing/hooks.ts
```diff
@@ -0,0 +1,63 @@
...
+  }
+  hooks[hookType] = fn;
+}
+
+/**
+ * Helper function to execute registred hooks.
+ */
+export const withHooks = (fn: THookFn) =>
+  async () => {
+    if (typeof fn !== "function") {
```
caspervonb, line 34: Type check before creating the promise.

testing/hooks_test.ts
```diff
@@ -0,0 +1,27 @@
...
+});
+
+Deno.test("afterEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  afterEach(function () {
+    didInvoke = true;
+    assertEquals(didInvoke, true);
+  });
+  assertEquals(didInvoke, false);
+  withHooks(function () {
```
caspervonb, line 24: Await this, this leaks.

```suggestion
  withHooks(function () {
```

testing/hooks_test.ts
```diff
@@ -0,0 +1,27 @@
...
+import { assertEquals } from "./asserts.ts";
+
+Deno.test("beforeEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  beforeEach(function () {
+    didInvoke = true;
+    assertEquals(/deno/, /deno/);
+  });
+  assertEquals(didInvoke, false);
+  withHooks(function () {
```
caspervonb, line 12: Await this, leaking.

testing/hooks_test.ts
```diff
@@ -0,0 +1,27 @@
...
+
+Deno.test("afterEach is exectued correctly", function (): void {
+  let didInvoke = false;
+  afterEach(function () {
+    didInvoke = true;
+    assertEquals(didInvoke, true);
+  });
+  assertEquals(didInvoke, false);
+  withHooks(function () {
+    assertEquals(/cat/, /cat/);
```
caspervonb, line 25:
```suggestion
```

testing/hooks.ts
getspooky: Array means that user can add more beforeEach and afterEach

testing/hooks.ts
getspooky: why we will check before creating a promise ?

testing/hooks_test.ts
kt3k: > adding await before withHook

This sounds good to me. BTW I think we should check `assertEquals(didInvoke, false)` inside the main function to ensure it's actually called after it.

getspooky: yes, it's good solution to check that

ry (commented):
I'm not sure if the before/after hooks should be global.

Also does it make sense to have a wrapper for `Deno.test` ?

Something like this?

```js
import { setup } from "https://deno.land/std/tests/hooks.ts";

const test = setup({
  before: () => {
    console.log("before each test");
  },
});

test("my test", () => {
  console.log("during test");
});
```

nayeemrmn:
Are any of these designs worth saving the two extra `setup(); teardown();` lines per-test? If you look at things like https://jestjs.io/docs/setup-teardown, this kind of API is built-in and usually lets you passively add hooks at the top without having to adapt the `test()` invocations from what they would normally look like. Global states and complicating the built-in API are things that we reasonably don't want, but having to a) bring in a dependency and b) reference the hooks at each test site, adds friction which more than offsets whatever advantage there is in the first place, IMO.

getspooky:
I think providing utils like `beforeEach`, `afterEach`, and `withHooks` without having a wrapper for `Deno.test` could be good, any suggestion ?

ry:
I think we need to give this feature more thought in light of https://github.com/denoland/deno/discussions/10771

bartlomieju:
@getspooky thank you for the PR, but I think we should focus on the proposal in denoland/deno#10771 and https://github.com/denoland/deno/discussions/11235 and support it natively in `deno` without changes to the `deno_std`. Respectfully closing without a merge.

PR #938, opened 2021-05-27
chore: update std for TypeScript 4.3 changes

kitsonk:
This will require canary once denoland/deno#9960 is merged until Deno 1.11 is released.

PR #940, opened 2021-05-29
chore: add underscore prefix to unused variables

magurotuna:
The current published version of deno_lint doesn't claim, but the next version is going to check if variables declared in `for-of` and `for-in` are used or not.
https://github.com/denoland/deno_lint/issues/697
So I added underscore prefix to such variables, in order to tell the linter that it's intentional to declare unused ones.

As a side note: we should remove ESLint disable directives because the linter doesn't support them since https://github.com/denoland/deno/issues/8226. ~~I'm working on this.~~ Done in #941


magurotuna:
Maybe blocked by #938

wperron (approved):
Indeed blocked by #938, otherwise looks good

kt3k (approved):
@magurotuna Thanks!

PR #942, opened 2021-05-29
docs(path): remove inexistant option

lowlighter:
The default example on [`path/README.md`](https://github.com/denoland/deno_std/tree/main/path) is incorrect:
```ts
import { globToRegExp } from "https://deno.land/std@$STD_VERSION/path/glob.ts";

globToRegExp("foo/**/*.json", {
  flags: "g",
  extended: true,
  globstar: true,
  caseInsensitive: false,
}); // returns the regex to find all .json files in the folder foo.
```

It throws an error because [`flags`](https://github.com/denoland/deno_std/blob/ac2f40e7e19f471d71ce787f7ea1d7ebe36f616f/path/glob.ts#L12-L25) is not defined on interface:
```
error: TS2345 [ERROR]: Argument of type '{ flags: string; extended: true; globstar: true; caseInsensitive: false; }' is not assignable to parameter of type 'GlobOptions'.
  Object literal may only specify known properties, and 'flags' does not exist in type 'GlobOptions'.
  flags: "g",
  ~~~~~~~~~~
    at ~/test.ts:4:3
```

It should be removed since it creates confusion


wperron (approved):
LGTM, Let's just rebase when #938 lands to fix the type errors on canary

kt3k (approved):
@lowlighter Thanks!

PR #946, opened 2021-06-01
docs(permissions): Add Permission documentation

getspooky:


wperron:
Thanks, I agree that the `permissions` directory could use a README, but I don't think a copy/paste of the manual page really helps to understand what this module does.

PR #947, opened 2021-06-01
fix(node/fs): fix type error in fs.watch impl

kt3k:
This PR fixes the type error in fs.watch impl of std/node. This error is caused by the change of builtin API in https://github.com/denoland/deno/pull/10798


wperron (approved):
LGTM

PR #948, opened 2021-06-02
feat(testing): improved strings diff

### LINKED ISSUE #929: [feature request] assertEquals multiline string diffs
currently `assertEquals` will pretty print js object differences 
screenshot(616)

but will show multiline string diffs as a single blob 
screenshot(617)
The actual diff here is pretty much unreadible so I often end up writing the changes to a file and using `git diff` or the `diff` command to see the actual changes.

It would be great if assertEquals could show multiline string diffs in a similar manner to `git diff`

lowlighter:
This add a specialized tokenizer that is called when both operands are string before entering `diff()`, making changes in large strings assertions more visible.
 
A word diff has been implemented (similar to `git diff --word-diff`):
image

As the above is not practical when a lot of words have been changed, a multiline diff has also been implemented:
image

**Todo**
- [x] word diff
- [x] multiline diff
- [x] combine word and multiline diffs 
- [x] link new code with `diff()` and `buildMessage()` 
- [x] tests 
  - [x] `_diff_test`
  - [x] `asserts_test`
- [ ] find good colors
- [x] merge word diff token if a in-between token is a space 

Closes #929

### DIFF
--- a/testing/_diff.ts
+++ b/testing/_diff.ts
@@ -274,15 +274,19 @@ export function diffstr(A: string, B: string) {
       return tokens;
     }
   }
-  //Try to autodetect the most suitable string rendering, starting by words-diff
+
+  // Try to autodetect the most suitable string rendering, starting by words-diff
   let wordDiff = true
   let diffResult = diff(tokenize(A, {wordDiff}), tokenize(B, {wordDiff}));
   const common = diffResult.filter(({type, value}) => type === DiffType.common && value.trim()).length
   const edited = diffResult.filter(({type}) => type === DiffType.added || type === DiffType.removed).length
-  //If edited ratio is too high switch to multi-line diff instead
-  if (edited/(common+edited) > 0.9**Math.max(1, A.search(/\n\r?/g), B.search(/\n\r?/g))) {
+
+  // If edited ratio is too high switch to multi-line diff instead
+  // We add a trailing newline to ensure that last tokens are displayed correctly
+  if (edited/(common+edited) > 0.85**Math.max(1, A.match(/\n/g)?.length ?? 0, B.match(/\n/g)?.length ?? 0)) {
     wordDiff = false
-    diffResult = diff(tokenize(A, {wordDiff}), tokenize(B, {wordDiff}));
+    diffResult = diff(tokenize(`${A}\n`, {wordDiff}), tokenize(`${B}\n`, {wordDiff}));
   }
+
   return {diffResult, wordDiff}
 }
--- a/testing/_diff.ts
+++ b/testing/_diff.ts
@@ -226,3 +226,51 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
     ),
   ];
 }
+
+/**
+ * Renders the differences between the actual and expected strings
+ * Partially inspired from https://github.com/kpdecker/jsdiff
+ * @param A Actual string
+ * @param B Expected string
+ * @param wordDiff If enabled, will tokenize words instead of lines
+ */
+export function diffstr(A:string, B:string, {wordDiff = false} = {}) {
+  function tokenize(string:string):string[] {
+    if (wordDiff) {
+      // Split string on whitespace symbols
+      const tokens = string.split(/([^\S\r\n]+|[()[\]{}'"\r\n]|\b)/);
+      // Extended Latin character set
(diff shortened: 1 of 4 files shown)


testing/_diff.ts
```diff
@@ -226,3 +226,80 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
...
+    diffResult.filter(({ type }) =>
+      type === DiffType.added || type === DiffType.removed
+    ).length;
+
+  // If edited ratio is too high switch to multi-line diff instead
+  // We add a trailing newline to ensure that last tokens are displayed correctly
+  if (
+    edited / (common + edited) >
+      0.85 **
+        Math.max(1, A.match(/\n/g)?.length ?? 0, B.match(/\n/g)?.length ?? 0)
```
lowlighter, line 295: Not sure whether it's the best way to make this but here's the reasoning:

- `edited / (common + edited)` give a ratio of "changes performed"
- `Math.max(1, A.match(/\n/g)?.length ?? 0, B.match(/\n/g)?.length ?? 0)` tells how many lines input has
- `0.85 ** x` (where `x` is above value) is the threshold from switching to word-diff to multiline-diff
  
Thus the more lines you have, the more likely you'll end up with multiline-diff unless there are only a few changes (like a word or two) whereas when there are less changes present the word-diff will be prefered

wperron:
I created a small gist just to get a sense of how the new diff looks visually -- it seems to be working _ok_ when both sides of the assertion are single-line string. Things get a little weird when the part of the string that differs contains a mix of newlines and spaces though.

Screenshot from 2021-06-03 10-44-17
Screenshot from 2021-06-03 10-42-15

When the strings are completely different, or when the string is inside an object or an array, it reverts back to the old diff. In the case where the string is nested in an object or array, I guess it's ok, but having to different diffs for strings seems a bit weird to me.

Screenshot from 2021-06-03 10-45-59
Screenshot from 2021-06-03 10-45-51
Screenshot from 2021-06-03 10-45-40

Here's the gist I used https://gist.github.com/wperron/77ba7b170799cc7b925c1734b38624e1

lowlighter:
Thanks for linking the gist

Yes I wasn't too sure how to handle both multilines and word diffs. I was actually waiting for some input about it. I can see multiple paths to takes:
- Only keep multiline diff and scrap word diff
  - 
- Combine them both to achieve something like this
  - 
- Or else I can check if it's possible to group alternating tokens together (like the ones you reported) but I think one of the above solution would be better
  - 

Currently it's an "autodetect" which choose between word-diff/multiline-diff based on the changes ratio and the number of lines, so that's why a full line diff revert back to old diff system but it's kind of hack, I'm not very convinced myself it 😅 

If you have others suggestions please let me know 👍 

___

For nested strings, I chose to not support them because I think it would quite a lof of work because of how diffs are computed. From my understanding, the generic `diff()` method actually takes from `Deno.inspect()`. So to support it, it would requires to catch strings properties from both `actual` and `expected` objects, and anyway I think a mixed diff would probably be more confusing I guess.

Current behaviour where both root operands must be strings seems fine to me, in the sense of when you're asserting two strings together you'd probably like to see a detailed diff between them, while if you assert whole objects/arrays you're probably more insterested in a more global view. If someone wants detailed string diff within a complex structure, one can always assert it explicitely by checking members

lowlighter:
Here's what it looks like now:

image

Not sure about the colors though, I think it has some accessibility issues 🤔

testing/asserts.ts
```diff
@@ -309,12 +332,13 @@ export function assertStrictEquals(
...
-        );
-        const diffMsg = buildMessage(diffResult).join("\n");
-        message = `Values are not strictly equal:\n${diffMsg}`;
+        const stringDiff = (typeof actual === "string") &&
+          (typeof expected === "string");
+        const diffResult = stringDiff
+          ? diffstr(actual as string, expected as string)
+          : diff(actualString.split("\n"), expectedString.split("\n"));
+        const diffMsg = buildMessage(diffResult, { stringDiff }).join("\n");
+        message = `Values are not equal:\n${diffMsg}`;
```
lowlighter, line 341:
```suggestion
        message = `Values are not strictly equal:\n${diffMsg}`;
```

kt3k:
Looks very useful 👍 Does anyone have any concern about introducing this?

kt3k (approved):
@lowlighter LGTM. Great feature!

PR #951, opened 2021-06-03
fix(equal): account for Weak* objects

mfulton26:
- `WeakMap` and `WeakSet` cannot be deeply compared (no `size`, `Symbol.iterator`, etc.)
- `WeakRef` can be dereferenced

Refs #949


kt3k (approved):
@mfulton26 LGTM. Thanks for your contribution!

maxxturing:
Hi @mfulton26 @kt3k I think this might have broken something?

I'm getting an error due to 'WeakRef'

https://github.com/denoland/deno/issues/10910

PR #952, opened 2021-06-05
docs(path): Rewrite README to fill examples

lambdalisue:
First of all, thanks for the project and maintainers for this great system and libraries. I really love it.

I've copied the description of each method and add some example codes to complete the README of the `path` module while it's a bit difficult to find correct documentation of `path` module while it switches the implementation by running the platform. 

This is my first contribution and I could not find any style guide for README so let me know if there is something I should fix 👍

### DIFF
--- a/path/README.md
+++ b/path/README.md
@@ -1,9 +1,196 @@
-# Deno Path Manipulation Libraries
+# path
 
-Usage:
+path module is made to provide helpers to manipulate the path.
+
+## Usage
 
 ```ts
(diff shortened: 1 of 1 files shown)


lambdalisue:
I haven't tested the example codes I wrote. I'll change the status to "Ready for review" once I've complete manual tests.

lambdalisue:
I've tested all example codes in README works. Now please review it.

kt3k (approved):
@lambdalisue Thanks! This is very helpful. LGTM.

PR #957, opened 2021-06-08
fix(node/process): make process.env iterable

cvng:
This PR add `getOwnPropertyDescriptor()` to `process.env` in order to make it iterable.


kt3k (approved):
@cvng LGTM. Thanks!

PR #958, opened 2021-06-09
feat(node/util): add inherits

cvng:
Port inherits from https://github.com/nodejs/node/blob/master/lib/util.js#L231

### DIFF
--- a/node/util.ts
+++ b/node/util.ts
@@ -265,6 +265,38 @@ export function format(input: string, ...args: unknown[]) {
   return result;
 }
 
+/**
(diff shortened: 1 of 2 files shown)


kt3k (approved):
@cvng Thanks! LGTM

kt3k:
note: definitelytyped provides looser declaration. ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a07e9cfb005682fb6be0a2e85113eac131c3006f/types/node/util.d.ts#L36

PR #959, opened 2021-06-09
fix(node/util): fix util.inherits

kt3k:
Found an error in `util.inherits` of node module https://github.com/denoland/deno_std/pull/958 (`inherits` util doesn't actually inherits the class). This PR fixes it, and enables the test cases from node repository (`parallel/test-util-inherits.js`)


node/_tools/config.json
```diff
@@ -17,7 +17,8 @@
       "test-event-emitter-invalid-listener.js",
       "test-event-emitter-listener-count.js",
       "test-event-emitter-listeners.js",
-      "test-event-emitter-once.js"
+      "test-event-emitter-once.js",
+      "test-util-inherits.js"
```
ry, line 21: 👍

ry (approved):
LGTM

PR #960, opened 2021-06-10
feat(mime): Add Custom content-type for MultipartWriter

getspooky:
It's good idea to specify custom headers: 
```ts
const buf = new Buffer();
const mw = new MultipartWriter(buf);
await mw.writeField("foo", "foo");
await mw.writeField("bar", "bar");
const f = await Deno.open(path.join(testdataDir, "sample.txt"), {
    read: true,
 });
const header = new Headers();
header.set("deno", "deno");
await mw.writeFile("file", "sample.txt", f, header);
await mw.close();
f.close();
```

### DIFF
--- a/mime/multipart.ts
+++ b/mime/multipart.ts
@@ -592,7 +592,7 @@ export class MultipartWriter {
       this.writer,
       this.boundary,
       headers,
-      !this.lastPart
+      !this.lastPart,
     );
     this.lastPart = part;
     return part;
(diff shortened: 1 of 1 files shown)


ry (commented):
This is what the "createPart" method is for?

getspooky:
`createPart` is private method. i don't there is a way to pass custom header ?

ry:
@getspooky Have a look at the Go version of this module, from which this is based: https://golang.org/pkg/mime/multipart/#Writer.CreatePart
https://golang.org/pkg/mime/multipart/#Writer.WriteField
The signatures here should match those in Go. So it seems like createPart should be public.

codehz (approved):

kt3k (approved):
LGTM

PR #962, opened 2021-06-11
fix: Apply node test suite 'ignore' configuration when running tests

Soremwar:
Currently the 'ignore' option in the configuration for the test runner is being ignored, causing edited test files to be regenerated when running the setup, deleting any manual changes and making it hard to track what files have been manually edited.

Additionally, the documentation about what this option actually did was not that clear, so that has been enhanced as well


kt3k (approved):
@Soremwar Thanks! LGTM

PR #968, opened 2021-06-13
fix(node/events): Enable test-event-emitter-remove-all-listeners.js and fix EventEmitter#removeAllListeners

uki00a:
Part of #911

## Changes

* Enabled `test-event-emitter-remove-all-listeners.js`.
* Fixed `EventEmitter#removeAllListeners` to remove listeners in LIFO order.


kt3k (approved):
@uki00a Thanks! LGTM

Issue #878, opened 2021-04-24
Assertion library custom message should still allow for inclusion of values


51ngul4r1ty:
This may just be a case of poor documentation, but I've been trying to figure out how I could get "1" and "2" to be included in the msg arg that's output when the test fails.  See code example provided in docs below:

```
Deno.test("Test Assert Equal Fail Custom Message", () => {
  assertEquals(1, 2, "Values Don't Match!");
});
```

It would be nice if it was possible to do something like this:

```
Deno.test("Test Assert Equal Fail Custom Message", () => {
  assertEquals(1, 2, "Values $1 and $2 don't match!");
});
```

I've used placeholders $1 and $2 because they're used in VS Code when doing regex replacement, but any other placeholders would be acceptable.

The reason this is useful is that you really want the message to provide the "why" information for the test result expected, but you still want to know what the value of the result actually was and how it differed from the expected value.  You could always duplicate the expected value in the message, but then there's two places to change it if you need to change it at some point in future.  However, it is essential to be able to output the actual value in the message because there's no way ahead of time that you'll know what that is.

kitsonk:
This would have to do with the std library assertion functions, not the `Deno.test()` API, so moving.

Nautigsam:
@sno2 I think a function is the way to go, but we could pass an object like `{actual: any, expected: any}` instead of an array. This could make the template string clearer:
```
Deno.test("Test Assert Equal Fail Custom Message", () => {
  assertEquals(1, 2, ({expected, actual}) => `Values ${expected} and ${actual} don't match!`);
});
```

grian32:
Will be PR-ing this in the coming days

jsejcksn:
Does a template literal address this?

```ts
Deno.test("Test Assert Equal Fail Custom Message", () => {
  const actual = 1;
  const expected = 2;
  assertEquals(actual, expected, `Values ${actual} and ${expected} don't match!`);
});
```

sno2:
> Does a template literal address this?
> 
> ```ts
> Deno.test("Test Assert Equal Fail Custom Message", () => {
>   const actual = 1;
>   const expected = 2;
>   assertEquals(actual, expected, `Values ${actual} and ${expected} don't match!`);
> });
> ```

Yes, perhaps it would be better to accept `(actual: T, expected: T) => string` or `string` (templates valid then).  The reason I was thinking about using callbacks is because I often include many asserts like this:
```ts
assertEquals(await foo.getBar(), "foo");
```
[...]

PR #982, opened 2021-06-23
feat(testing/asserts.ts): allows custom error message that includes actual and expected values.

grian32:
Implements #878 which allows this syntax:
```ts
Deno.test("some test", () => {
  assertEquals(1, 2, (actual, expected) => `actual: ${actual}, expected: ${expected}`);
});
```

Wasn't sure how to implement this for `assertThrows()` and `assertThrowsAsync()`, I would like some help with that.

Would also like to know if I should typealias `msg?: string | ((first: typeof actual, second: typeof expected) => string)` or leave it as is.

Will also be commiting a docs change that mentions this later today/tommrow, but wanted feedback on the actual implementation before i did that.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -53,6 +53,20 @@ Deno.test({
 });
 ```
 
+`assertEquals` and other functions that allow custom messages also accept this syntax:
+```ts
+import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+
+Deno.test({
+  name: "testing example",
+  fn(): void {
+    assertEquals(dynamicValue1(), dynamicValue2(), ({message, actual, expected}) => {
+      return `${actual} didnt match ${expected}, here's the diff: ${message}`
+    });
+  },
+});
+```
+
 Short syntax (named function instead of object):
 
 ```ts
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -204,13 +204,13 @@ export function assert(expr: unknown, msg = ""): asserts expr {
 export function assertEquals(
   actual: unknown,
   expected: unknown,
-  msg?: string,
+  msg?: string | ((first: typeof actual, second: typeof expected) => string),
 ): void;
-export function assertEquals<T>(actual: T, expected: T, msg?: string): void;
+export function assertEquals<T>(actual: T, expected: T, msg?: string | ((first: typeof actual, second: typeof expected) => string)): void;
 export function assertEquals(
   actual: unknown,
   expected: unknown,
-  msg?: string,
+  msg?: string | ((first: typeof actual, second: typeof expected) => string), // uses first and second for the callback because ts complains when using actual/expected
 ): void {
   if (equal(actual, expected)) {
     return;
@@ -229,7 +229,11 @@ export function assertEquals(
     message = `\n${red(CAN_NOT_DISPLAY)} + \n\n`;
   }
   if (msg) {
-    message = msg;
+    if (typeof msg === "string") {
+      message = msg;
+    } else if (typeof msg === "function") {
+      message = msg(actual, expected);
+    }
   }
   throw new AssertionError(message);
 }
@@ -247,19 +251,20 @@ export function assertEquals(
 export function assertNotEquals(
   actual: unknown,
   expected: unknown,
-  msg?: string,
+  msg?: string | ((first: typeof actual, second: typeof expected) => string),
 ): void;
-export function assertNotEquals<T>(actual: T, expected: T, msg?: string): void;
+export function assertNotEquals<T>(actual: T, expected: T, msg?: string | ((first: typeof actual, second: typeof expected) => string)): void;
 export function assertNotEquals(
   actual: unknown,
   expected: unknown,
-  msg?: string,
+  msg?: string | ((first: typeof actual, second: typeof expected) => string),
 ): void {
   if (!equal(actual, expected)) {
     return;
   }
   let actualString: string;
   let expectedString: string;
+  let message = "";
   try {
     actualString = String(actual);
   } catch {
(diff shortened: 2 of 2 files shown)


wperron:
Before reviewing in more details, is there a use-case where using a template literal instead would not work?

grian32:
> Before reviewing in more details, is there a use-case where using a template literal instead would not work?

If you're talking about this when you say template literal:
```ts
Deno.test("some test", () => {
  const actual = 1;
  const expected = 2;
  assertEquals(actual, expected, `actual: ${actual}, expected: ${expected}`);
});
```

It's just cleaner and doesn't require you to assign constant/small(i.e if actual was `arr.slice(0, 4)` or something) values to a variable.

caspervonb:
I'm not sure `actual` and `expected` are useful, the original message however could be useful to add more context.

```
assertEquals(a, b, (message) => { `This specific state failed after doing thing: ${message}` });
```

grian32:
> I'm not sure `actual` and `expected` are useful

It's mostly useful if you want to include dynamic values into your message. There's also the explanation by the original suggestion maker for this in #878

caspervonb:
Some prior art comes to mind https://github.com/srittau/python-asserts (pip asserts)

grian32:
> Some prior art comes to mind https://github.com/srittau/python-asserts (pip asserts).
> What if we did this with restructuring rather than positional arguments so you can drop whatever you don't need, for example:
> 
> ```ts
> assertEquals(actual, expected, ({ message, actual, expected }) => `...`);
> ```

That seems like the best way to do it yeah, I'll look at implementing that sometime soon.

kt3k:
> Some prior art comes to mind https://github.com/srittau/python-asserts (pip asserts).
> What if we did this with restructuring rather than positional arguments so you can drop whatever you don't need, for example:
> 
> ```ts
> assertEquals(actual, expected, ({ message, actual, expected }) => `...`);
> ```

I agree with this design. Only passing actual and expected is not that useful as they are already available in the same line.

And we also need test cases for checking this overloading works correctly.

wperron:
The more I think about it, the more I think we shouldn't land this; a template literal already works just fine and we don't need to add test cases to check that the function overload works correctly. The only downside is that in the simple cases, it requires putting `actual` and `expected` in consts, imo not a big enough issue to warrant a change in the API

Aplet123:
I think the biggest issue right now is that supplying a custom message loses a lot of the information in the original message (in the case of `assertEquals`, it would be the diff output). I think that it'd be helpful to optionally allow a function that takes in the diff (among other things) and can format a custom but informational message. I don't see why changing the API would be a huge issue, since it should only expand the API and not result in any breaking changes (hopefully people weren't passing functions as the message parameter and expecting it to implicitly cast to a string).

wperron:
> I think the biggest issue right now is that supplying a custom message loses a lot of the information in the original message (in the case of `assertEquals`, it would be the diff output). I think that it'd be helpful to optionally allow a function that takes in the diff (among other things) and can format a custom but informational message. I don't see why changing the API would be a huge issue, since it should only expand the API and not result in any breaking changes (hopefully people weren't passing functions as the message parameter and expecting it to implicitly cast to a string).

The proposed change wouldn't add back the missing context, it's just a wrapper around the custom string message, the diff would still be taken away. Having the diff show in `assertEquals` with or without a custom message would be a separate issue.

Aplet123:
> The proposed change wouldn't add back the missing context, it's just a wrapper around the custom string message, the diff would still be taken away.

While I definitely don't agree with the implementation in *this* specific pull request (maybe this discussion should be moved out of the PR and into an issue?), I think that it would be very helpful to have a change that does address the things that have been discussed here (such as taking the diff as an argument, passing in values as an object instead of parameters, etc.).

bartlomieju:
> Some prior art comes to mind https://github.com/srittau/python-asserts (pip asserts).
> What if we did this with restructuring rather than positional arguments so you can drop whatever you don't need, for example:
> 
> ```ts
> assertEquals(actual, expected, ({ message, actual, expected }) => `...`);
> ```

I agree with this statement - using destructuring will be the least error prone way as well as most ergonomic (don't need to lookup the order of arguments).

I think we should land it in that form. @grian32 can you please update your PR to match that signature?

wperron:
at that point what's the difference between `(any) => string` and just using a template literal? the latest proposal still requires putting all the parameters in variables first, or construct an object with explicit keys, which was the only downside mentioned to using a template literal. I don't think this is worth making the function signature more complex and less clear when there's already a solution in the language for templating.

bartlomieju:
> at that point what's the difference between `(any) => string` and just using a template literal? the latest proposal still requires putting all the parameters in variables first, or construct an object with explicit keys, which was the only downside mentioned to using a template literal. I don't think this is worth making the function signature more complex and less clear when there's already a solution in the language for templating.

@wperron I'm not sure I understand your comment. If we were to go with the template literal you will lose the default message produced by assertion function.

wperron:
> > at that point what's the difference between `(any) => string` and just using a template literal? the latest proposal still requires putting all the parameters in variables first, or construct an object with explicit keys, which was the only downside mentioned to using a template literal. I don't think this is worth making the function signature more complex and less clear when there's already a solution in the language for templating.
> 
> @wperron I'm not sure I understand your comment. If we were to go with the template literal you will lose the default message produced by assertion function.

Passing a third parameter for a custom message already obfuscate the original assertion message, passing a function that returns a string won't change that (we could open a separate issue for that though). The proposed change here changes the signature of that third parameter from `string` to `() => string` to allow passing `actual` and `expected` to the custom message, but this is already possible by using a template literal instead of a regular string, and it already supports passing an arbitrary number variables, with arbitrary names; I don't think it's necessary to wrap it in a function at all.

grian32:
> changes the signature of that third parameter from string to () => string to allow passing actual
you can still pass in a `string` with this PR, any code that existed before this wouldn't break

grian32:
> > Some prior art comes to mind https://github.com/srittau/python-asserts (pip asserts).
> > What if we did this with restructuring rather than positional arguments so you can drop whatever you don't need, for example:
> > ```ts
> > assertEquals(actual, expected, ({ message, actual, expected }) => `...`);
> > ```
> 
> I agree with this statement - using destructuring will be the least error prone way as well as most ergonomic (don't need to lookup the order of arguments).
> 
> I think we should land it in that form. @grian32 can you please update your PR to match that signature?

Sure, will update it in the next few days

grian32:
@bartlomieju I've updated the signature and I will start writing tests, how am I expected to do tests for `assertEquals` and other asserts with diffs(since they contain colors)? My first instinct would be to have test files and `Deno.run` them with the `NO_COLOR` env var set and pipe stdout, but I'm not sure if this is wanted.

lino-levan:
@grian32 Wondering why you closed this?

PR #991, opened 2021-06-30
feat(encoding/toml): fix bad string format. Improve coverage

zekth:
- feat: delete unecessary strip comment
- feat: fix bad string format

This PR fixes a bad string format case and improve the code coverage of the module.

### DIFF
--- a/encoding/toml_test.ts
+++ b/encoding/toml_test.ts
@@ -1,5 +1,5 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
-import { assertEquals } from "../testing/asserts.ts";
+import { assertEquals, assertThrows } from "../testing/asserts.ts";
 import { existsSync } from "../fs/exists.ts";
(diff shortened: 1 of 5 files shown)


encoding/toml_test.ts
```diff
@@ -480,3 +484,29 @@ Deno.test({
     assertEquals(actual, expected);
   },
 });
+
+Deno.test({
+  name: "[TOML] Single-line string comment error",
```
wperron, line 489: This test passes when running against the previous version of the `encoding/toml.ts` code -- is that expected?

zekth: This test was added to add coverage for this line : https://github.com/denoland/deno_std/blob/635da0f56c8c7b2294b0448763bda1d055dcfe44/encoding/toml.ts#L101

wperron (approved):
LGTM -- validated the new test cases locally, thanks for the contribution

PR #992, opened 2021-06-30
fix: `setCookie` with `maxAge` of `0`

### LINKED ISSUE #989: std/http/cookie: unable to call `setCookie()` with `maxAge` of `0`
**Describe the bug** It is currently not possible to set a cookie with a `Max-Age` of `0` using the `http` std lib `setCookie()` method.

```ts
// cookieMaxAgeError.ts
import { setCookie } from "https://deno.land/std@0.100.0/http/cookie.ts";

setCookie({}, { name: "deno", value: "example", maxAge: 0 });
```

```console
$ deno run cookieMaxAgeError.ts 
Check ./cookieMaxAgeError.ts
error: Uncaught DenoStdInternalError: Max-Age must be an integer superior to 0
    throw new DenoStdInternalError(msg);
          ^
    at assert (https://deno.land/std@0.100.0/_util/assert.ts:13:11)
    at toString (https://deno.land/std@0.100.0/http/cookie.ts:60:5)
    at setCookie (https://deno.land/std@0.100.0/http/cookie.ts:180:13)
    at ./cookieMaxAgeError.ts:3:1
```

**Expected behavior** A cookie instance with a `Max-Age` of `0` should be created, which subsequently follows the behaviour of https://www.ietf.org/rfc/rfc2109.txt

> Max-Age=delta-seconds
>      Optional.  The Max-Age attribute defines the lifetime of the
>      cookie, in seconds.  The delta-seconds value is a decimal non-
>      negative integer.  After delta-seconds seconds elapse, the client
>      should discard the cookie.  A value of zero means the cookie
>      should be discarded immediately.

**Desktop (please complete the following information):**

- OS: MacOS 11 ( Big Sur )

```console
deno 1.11.3 (release, x86_64-apple-darwin)
v8 9.1.269.35
typescript 4.3.2
```

**Additional context** Raising following https://github.com/asos-craigmorten/opine/issues/125

SimonRask:
This PR changes the `toString` function for `Cookie` in order for `setCookie` to allow for `maxAge` with a value of `0`. This fixes #989

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -57,7 +57,10 @@ function toString(cookie: Cookie): string {
     out.push("HttpOnly");
   }
   if (typeof cookie.maxAge === "number" && Number.isInteger(cookie.maxAge)) {
-    assert(cookie.maxAge >= 0, "Max-Age must be an integer superior or equal to 0");
+    assert(
+      cookie.maxAge >= 0,
+      "Max-Age must be an integer superior or equal to 0",
+    );
     out.push(`Max-Age=${cookie.maxAge}`);
   }
   if (cookie.domain) {
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -11,7 +11,7 @@ export interface Cookie {
   value: string;
   /** Expiration date of the cookie. */
   expires?: Date;
-  /** Max-Age of the Cookie. Must be integer superior to 0. */
+  /** Max-Age of the Cookie. Must be a non-negative integer. */
   maxAge?: number;
   /** Specifies those hosts to which the cookie will be sent. */
   domain?: string;
@@ -57,7 +57,7 @@ function toString(cookie: Cookie): string {
     out.push("HttpOnly");
   }
   if (typeof cookie.maxAge === "number" && Number.isInteger(cookie.maxAge)) {
-    assert(cookie.maxAge > 0, "Max-Age must be an integer superior to 0");
+    assert(cookie.maxAge >= 0, "Max-Age must be a non-negative integer");
     out.push(`Max-Age=${cookie.maxAge}`);
   }
   if (cookie.domain) {
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -11,7 +11,7 @@ export interface Cookie {
   value: string;
   /** Expiration date of the cookie. */
   expires?: Date;
-  /** Max-Age of the Cookie. Must be a non-negative integer. */
+  /** Max-Age of the Cookie. Max-Age must be an integer superior or equal to 0. */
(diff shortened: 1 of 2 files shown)


http/cookie_test.ts
```diff
@@ -177,20 +177,14 @@ Deno.test({
...
-      error = true;
-    }
-    assert(error);
+    setCookie(res, {
+      name: "Space",
+      value: "Cat",
+      httpOnly: true,
+      secure: true,
+      maxAge: 0,
+    });
```
getspooky, line 188:
```ts

let error = false;
    res.headers = new Headers();
    try {
      setCookie(res, {
        name: "Space",
        value: "Cat",
        httpOnly: true,
        secure: true,
        maxAge: -1,
      });
    } catch {
      error = true;
      assert(error);
    }

```

getspooky, line 188: I think we should also add this test 
```suggestion

let error = false;
    res.headers = new Headers();
    try {
      setCookie(res, {
        name: "Space",
        value: "Cat",
        httpOnly: true,
        secure: true,
        maxAge: -1,
      });
    } catch {
      error = true;
      assert(error);
    }

```

http/cookie.ts
```diff
@@ -57,7 +57,7 @@ function toString(cookie: Cookie): string {
     out.push("HttpOnly");
   }
   if (typeof cookie.maxAge === "number" && Number.isInteger(cookie.maxAge)) {
-    assert(cookie.maxAge > 0, "Max-Age must be an integer superior to 0");
+    assert(cookie.maxAge >= 0, "Max-Age must be a non-negative integer");
```
wperron, line 60: Minor nit on the error message:
```suggestion
    assert(cookie.maxAge >= 0, "Max-Age must be an integer superior or equal to 0");
```

http/cookie_test.ts
wperron: Agreed, though the formatting on the suggested patch will need to be fixed.

wperron (changes requested):

http/cookie_test.ts
```diff
@@ -177,6 +177,15 @@ Deno.test({
...
     );
 
+    res.headers = new Headers();
+    setCookie(res, {
+      name: "Space",
+      value: "Cat",
+      httpOnly: true,
+      secure: true,
+      maxAge: 0,
+    });
```
wperron, line 187: There should probably be an assertion here, either check for the value of the header or that it doesn't throw

wperron (changes requested):

getspooky (approved):

wperron (approved):
LGTM, Thanks!

PR #993, opened 2021-06-30
Proposal: std/collections - first part

kt3k:
This proposal suggests adding 17 APIs (11 Array utils, 6 [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeystype) utils) as `std/collections` module. (These are extracted from the original proposal #978 by @LionC. We keep discussion on other APIs on the original PR.)

Array utils (11)
- distinct
- distinctBy
- intersect
- union
- findLast
- zip
- unzip
- groupBy
- partition
- chunked
- permutations

Record utils (6)
- filterEntries
- filterKeys
- filterValues
- mapEntries
- mapKeys
- mapValues

In the original issue, nobody was against adding these 17. So this can be a minimal set of APIs which we could land as a first pass of `std/collections`.

I'd like to land these items first if there's no strong objections to some specific items or the addition of `std/collections` entirely.


ry (commented):
This looks like a good set of APIs to start with.

wperron:
I started implementing `union` and `intersect` one think I'm noticing is that we could change the input parameters from `a: Array<T>, b: Array<T>` to `...a: Array<T>` so that it can be called on an arbitrary number of arrays at once.

collections/union.ts
```diff
@@ -0,0 +1,25 @@
...
+ * const saladIngredients = [ 'Carrots', 'Radicchio', 'Pepper' ]
+ * const shoppingList = union(soupIngredients, saladIngredients)
+ *
+ * console.assert(shoppingList === [ 'Pepper', 'Carrots', 'Leek', 'Radicchio' ])
+ * ```
+ */
+export function union<T>(a: Array<T>, b: Array<T>): Array<T> {
+  const s = new Set<T>();
+  while (a.length > 0) {
+    s.add(a.shift() as T);
```
Zzzen, line 19: Why is the input mutated here?

kt3k: Right, this shouldn't change the input.

wperron: Brain fart on my end, I'll update the implementation and add a test that checks the original arrays are intact

LionC: We should add that test to all functions probably.

LionC:
I like the varargs for union and intersect.

I am back from vacation now and will continue implementing :-)

collections/union.ts
LionC: > Brain fart on my end, I'll update the implementation and add a test that checks the original arrays are intact

If you are not shifting, this looks like a good case for a simple forEach call :-)

wperron: > This looks like a good case for a simple forEach call :-)

I updated to a `for...of`

collections/distinct.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * Returns all distinct elements in the given array, preserving order by first occurence
+ *
+ * Example:
+ *
+ * ```typescript
+ * const numbers = [ 3, 2, 5, 2, 5 ]
+ * const distinctNumbers = distinct(numbers)
+ *
+ * console.assert(distinctNumbers === [ 3, 2, 5 ])
```
Zzzen, line 12: nitpick: `console.assert` will always fail because their identities are different. `assert.deepStrictEqual(distinctNumbers, [3,2,5])` in nodejs or `expect(distinctNumbers).toEqual([3,2,5])` in jest might be a better choice.

LionC: You are absolutely right - we should make sure that those examples actually work. I just totally missed that. I will put it on my to do list to do a sweep of that in the end.

LionC:
I am finished with implementations and tests for now - obviously pending review. Will move on to refining types and docs now.

sebastienfilion:
Amazing work! 

I was wondering if there’d be a lot of opposition to sticking to intransitive verbs for the functions. I am confused by chunked and permutations — why isn’t it chunk and permute?
Those two specifically led me off their true intent.

 Then there’s union which definitely is more common… although it could be named concat, or concatUnique since it seems to remove duplicates… but I don’t have a strong opinion about that last one.

LionC:
> Amazing work!
> 
> I was wondering if there’d be a lot of opposition to sticking to intransitive verbs for the functions. I am confused by chunked and permutations — why isn’t it chunk and permute?
> Those two specifically led me off their true intent.
> 
> Then there’s union which definitely is more common… although it could be named concat, or concatUnique since it seems to remove duplicates… but I don’t have a strong opinion about that last one.

To be honest, names are mostly stolen from `kotlin/collections` and `lodash`, which seem to have a very similiar idea about this. As they are both used a lot and I personally never had an issue with the names, I did not modify them too much.

getspooky:
Amazing work 👍,
I think it would be helpful to add this collections of 17 APIs (11 Array utils, 6 Record utils) into [Third Party Modules
GitHub](https://deno.land/x) instead of Standard Library because it's not related to deno but to typescript.

LionC:
> Amazing work 👍,
> 
> I think it would be helpful to add this collections of 17 APIs (11 Array utils, 6 Record utils) into [Third Party Modules
> 
> GitHub](https://deno.land/x) instead of Standard Library because it's not related to deno but to typescript.
> 
> 

As I understand it, the std lib wBts to battle the dependency hell of needing third party code for everything, which is why we decided to do it this way. There has been a lot of discussion around this module on Discord and in the linked Discussions (following the links in the initial post of this PR).

collections/map_values.ts
```diff
@@ -0,0 +1,38 @@
...
+ * ```
+ */
+export function mapValues<T, O>(
+  record: Record<string, T>,
+  transformer: Selector<T, O>,
+): Record<string, O> {
+  const ret: Record<string, O> = {};
+  const entries = Object.entries(record);
+
+  for (const [key, value] of entries) {
```
wperron, line 31: Same here, `Object.entries` can be inlined

collections/unzip.ts
```diff
@@ -0,0 +1,34 @@
...
+ * ```
+ */
+export function unzip<T, U>(pairs: Array<[T, U]>): [Array<T>, Array<U>] {
+  const { length } = pairs;
+  const ret: [Array<T>, Array<U>] = [
+    new Array(length),
+    new Array(length),
+  ];
+
+  pairs.forEach(([first, second], index) => {
```
wperron, line 28: Feels a bit weird to have `.forEach` here when we're using `for...of` everywhere else, is there a reason to use it specifically here?

collections/_utils.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { Predicate } from "./types.ts";
+
+/**
+ * Filters the given array, removing all elements that do not match the given predicate
+ * **in place. This means `array` will be modified!**.
+ */
+export function filterInPlace<T>(
```
wperron, line 9: Since this function is used to implement some of the other functions it would be nice to have some tests for it.

wperron (changes requested):
I have some comments, most of them are minor nitpick, and obviously there needs to be some more docs. Also make sure that the block comments are formatted to stick within 80 characters and fix the linting errors.

Other than that, great work, looking forward to landing this!

collections/intersect.ts
```diff
@@ -0,0 +1,32 @@
...
+    return [];
+  }
+
+  const [originalHead, ...tail] = arrays;
+  const head = [...originalHead];
+  const tailSets = tail.map((it) => new Set(it));
+
+  for (const set of tailSets) {
+    filterInPlace(head, (it) => set.has(it));
+  }
```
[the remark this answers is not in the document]
LionC: The reason I did it this (admittedly more complicated) way is that it reduces the complexity from O(N \* L^2) to O(N \* L), which seemed significant to me. Lets benchmark maybe?

collections/unzip.ts
LionC: Because I need the index and the elements, so forEach seemed most pragmatic. Could use entries and for of, but that creates an unnecessary iteration within entries.

collections/_utils.ts
LionC: As long as it is internal I personally do not agree - testing implementation details instead of the stable contract leads to maintenance effort that does not create value in my experience (see [this talk](https://youtu.be/EZ05e7EMOLM) for a way better delivery of that point).

My opinion here is not strong though, so I am open to add tests if that is what is needed to bring it live^^

collections/find_last.ts
```diff
@@ -0,0 +1,30 @@
...
+ * const lastEvenNumber = findLast(numbers, it => it % 2 === 0)
+ *
+ * console.assert(lastEvenNumber === 2)
+ * ```
+ */
+export function findLast<T>(
+  array: Array<T>,
+  predicate: Predicate<T>,
+): T | undefined {
+  for (let i = array.length - 1; i >= 0; i -= 1) {
```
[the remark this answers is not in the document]
LionC: You will find that I have not used i++ anywhere here^^ I am not the biggest fan of the syntax, it feels very three-star-y to me and does a lot of implicit things as an expression, while + 1 is a very simple and intuitive real life concept. So I just followed that here as well as in all the other functions.

Again, no strong opinion though, just explaining why it is that way.

collections/filter_entries_test.ts
```diff
@@ -0,0 +1,107 @@
...
+      },
+    );
+    filterEntriesTest(
+      [
+        {
+          "A": true,
+          "b": "foo",
+          "C": 5,
+          "d": -2,
+          "": false,
```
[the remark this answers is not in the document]
LionC: I agree - I cannot add it on vacation, feel free to donit or I will do it once I return :-)

collections/distinct_by_test.ts
```diff
@@ -0,0 +1,126 @@
...
+    distinctByTest(
+      [],
+      () => {},
+      [],
+    );
+  },
+});
+
+Deno.test({
+  name: "[collections/distinctBy] gets head on noop selector",
```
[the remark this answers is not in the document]
LionC: As `() => {}` is equivalent to `it => undefined` here, it would feel like unexpected magic behavior to me if it would explicitly ignore equality (distinctity? is that a word?) for `undefined`. What about `null`? `""`? What if `undefined` is a legit value in my model?

I wrote those tests before implementing the function, so to mee it seemed like the intuitive behavior to just distinct all values by their produced value - no matter what it is.

collections/distinct_by_test.ts
```diff
@@ -0,0 +1,126 @@
...
+      [true, "asdf", 4, "asdf", true],
+      (it) => it,
+      [true, "asdf", 4],
+    );
+    distinctByTest(
+      [null, undefined, null, "foo", undefined],
+      (it) => it,
+      [null, undefined, "foo"],
+    );
+    distinctByTest(
```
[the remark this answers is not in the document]
LionC: Yep 🙈 I probably copied the test three times and forgot to adapt the third - it was a long copy-paste evening^^

I cannot change it ln vacation. Feel free to delete the duplicat test or change it to a new one. I will donit when I return otherwise :-)

ry (approved):
LGTM - thank you @LionC !

Good enough for a first pass, we can clean up in the future as necessary

collections/distinct.ts
```diff
@@ -0,0 +1,19 @@
...
+ * ```
+ */
+export function distinct<T>(array: Array<T>): Array<T> {
+  const set = new Set(array);
+
+  return Array.from(set);
```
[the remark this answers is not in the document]
timreichen: @wperron https://github.com/denoland/deno_std/pull/1099

Issue #658, opened 2020-07-07
Remove JavaScript versions of std/hash


caspervonb:
Right now we're doubling up on a lot of hashing algorithms.

The WebAssembly implementation is a good first pass but if we're keeping single algorithm imports (e.g std/hash/sha1.ts) then we should probably generate WebAssembly backed classes for these as-well.

iuioiua:
@bartlomieju, is this able to be closed? `std/hash` is deprecated.

PR #994, opened 2021-06-30
BLAKE3 hash support

jeremyBanks:
[BLAKE3](https://github.com/BLAKE3-team/BLAKE3/) is a performant and versatile hash function. It's the successor to the well-studied BLAKE2 family of hashes. I'd like to use for some projects, and I think could make a nice addition to the standard library. 

If there is interest, here is a PR which adds it to std/hash via the official Rust implementation. (Per #658 it looks like we wouldn't want to also include a TypeScript implementation.)

I'm not sure if this is redundant with the ongoing work on WebCrypto. However, the WebCrypto specification seems to only include SHA-1 and SHA-2-family digests, so perhaps this implementation would still be useful until/unless a `"blake3"` option is added to that specification.

---

Adds support for the BLAKE3 hash function to std/hash.

This uses the official blake3 Rust crate, via the hash/_wasm/wasm.js bundle.
No TypeScript implementation is included.

This includes only the primary BLAKE3 function with its default 256-bit output length.
This does NOT expose its secondary modes or support for variable-length output.

Test cases were generated using the official b3sum CLI.
For example, for testSetBase64.blake3.millionAs:

    printf "%1000000s" | tr " " "a" | b3sum --no-names --raw - | base64 --wrap=0


kt3k (approved):
Great! LGTM

Can you also update 'supported algorithms' section in `hash/RAEDME.md`?

jeremyBanks:
@kt3k Thanks, done.

kt3k:
@jeremyBanks Thank you for your contribution!

PR #995, opened 2021-07-01
fix: Improve type safety for browser-compatible modules

jsejcksn:
This PR fixes the current compiler errors emitted when using browser-compatible TypeScript modules without including the Deno namespace library in `compilerOptions.lib`.

---

There is a link to [contributing guidelines](https://github.com/denoland/deno/blob/de6e44794b7a95b6895a9aab907e01788820d18c/docs/contributing/style_guide.md#document-and-maintain-browser-compatibility) toward the end of the [Deno repo README](https://github.com/denoland/deno/tree/de6e44794b7a95b6895a9aab907e01788820d18c#contributing), which includes this:

> #### Document and maintain browser compatibility.
> 
> If a module is browser compatible, include the following in the JSDoc at the top of the module:
> 
> ```ts
> // This module is browser compatible.
> ```
> 
> Maintain browser compatibility for such a module by either not using the global `Deno` namespace or feature-testing for it. Make sure any new dependencies are also browser compatible.

This statement is not explicit about type compatibility, but I think it is natural for the type safety of browser-compatible modules to be aligned in the same way.

I searched for all files including the mentioned text using this method:
```
# cd to repo directory
grep -Rl "This module is browser compatible." .
```

For each file, I checked the diagnostic output of the compiler for warnings/errors using a browser-compatible TSConfig with Deno. ([The TSConfig is presently included in the PR](https://github.com/denoland/deno_std/blob/5d06b505e0535211a4d01f650ef5674ce271b2a0/browser-compat.tsconfig.json), but can be removed if desired.) If diagnostic messages were emitted for a file, I addressed them in the source. Most of the changes involve modifications to conditional expressions used to determine whether or not a Deno method or property exists.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,9 @@ jobs:
       - name: Run tests
         run: deno test --unstable --allow-all
 
+      - name: Type check browser compatible modules
+        run: grep -Rl --exclude-dir=./.github "// This module is browser compatible." . | xargs deno cache --config browser-compat.tsconfig.json
+
       - name: Generate lcov
         run: deno coverage --unstable --lcov ./cov > cov.lcov
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,9 +32,6 @@ jobs:
       - name: Run tests
         run: deno test --unstable --allow-all
 
-      - name: Type check browser compatible modules
-        run: grep -Rl "// This module is browser compatible." . | xargs deno cache --config browser-compat.tsconfig.json
-
       - name: Generate lcov
         run: deno coverage --unstable --lcov ./cov > cov.lcov
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,11 @@ jobs:
       - name: Run tests
         run: deno test --unstable --allow-all
 
+      - name: Type check browser compatible modules
+        shell: bash
+        run: |
+          git grep --name-only "// This module is browser compatible." | grep -v ".github/workflows" | xargs deno cache --config browser-compat.tsconfig.json
+
       - name: Generate lcov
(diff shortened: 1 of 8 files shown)


kt3k:
> For each file, I checked the diagnostic output of the compiler for warnings/errors using a browser-compatible TSConfig with Deno.

I think we need to add a new step in `.github/workflows/ci.yaml` to check these.

something like:
```
deno cache --config browser-compat.tsconfig.json _util/os.ts
deno cache --config browser-compat.tsconfig.json fmt/colors.ts
...
```

jsejcksn:
> I think we need to add a new step in `.github/workflows/ci.yaml` to check these.

@kt3k This makes sense. Have any ideas on how to programmatically generate the list of matching modules and insert into the workflow file so that manual maintenance is not needed?

kt3k:
@jsejcksn I guess something like the below might work

```sh
grep "This module is browser compatible" . -Rl | xargs deno cache --config browser-compat.tsconfig.json
```

jsejcksn:
@kt3k I tried adding a step for type checking, but got an error like this:

```
error: Expected ';', '}' or <eof> at file:///home/runner/work/deno_std/deno_std/.github/workflows/ci.yml:11:11
```

At first, I thought it was because the YAML file was matching, too, but even after applying a variation to ignore the `./.github` directory, the error persisted.

Feel free to make a code change suggestion or add a commit to this PR if you think you can make it work. This was my last addition before reverting:

```yaml
- name: Type check browser compatible modules
  run: grep -Rl --exclude-dir=./.github "// This module is browser compatible." . | xargs deno cache --config browser-compat.tsconfig.json
```

kt3k (approved):
@jsejcksn Thank you for your contribution! LGTM

jsejcksn:
@kt3k Thanks for fixing the workflow file

Issue #786, opened 2021-03-05
createHash memory leak


shamblesides:
Discovered this because pbkdf2 from node/crypto was leaking memory, but the hash module seems to be the source. This script increases in memory at a rate of 50MB/s on my machine, indefinitely (I stopped it at 1GB):

```js
import { createHash } from "https://deno.land/std@0.89.0/hash/mod.ts"

while (1) {
  createHash('sha512')
}
```

It appears to happen with any hash, not just sha512.

Verison info

```
$ deno --version
deno 1.8.0 (release, x86_64-unknown-linux-gnu)
v8 9.0.257.3
typescript 4.2.2
```

heyheyhello:
I think it's because it's instantiated in WASM rather than in JS.
 https://www.github.com/denoland/deno_std/tree/main/hash%2F_wasm%2Fwasm.js

It looks like you can call `free()` manually which should help.

shamblesides:
For my particular use case, I guess that means there will need to be a PR that adds the free() call somewhere around here: https://github.com/denoland/deno_std/blob/main/node/_crypto/pbkdf2.ts#L33

heyheyhello:
Hmm well createHasher returns an anonymous function, so you could add free as a property to the function so it would be available to manually call.

shamblesides:
Actually it looks like this won't work without changes to the hash module itself.

The public module function, createHash:

https://github.com/denoland/deno_std/blob/main/hash/mod.ts#L34

returns a Hash object:

https://github.com/denoland/deno_std/blob/main/hash/_wasm/hash.ts#L19

and the Hash object has a private property called `#hash` which is from a completely different function called createHash which is generated by wasm:

https://github.com/denoland/deno_std/blob/main/hash/_wasm/hash.ts#L24

so if I want to call free() I have to access private properties on a class defined within the _wasm folder, which seems like a hack to me.

caspervonb:
`digest` from `crypto.subtle` should superceede this by 1.9

shamblesides:
Sounds good, then I will wait for `crypto.subtle`

Do you think `deriveKey` will also be available? That's the interface I'm really wanting to use, for PBKDF2

caspervonb:
Probably https://github.com/denoland/deno/pull/9614

PR #998, opened 2021-07-02
fix(hash): Partially mitigate WASM memory leak #786 using FinalizationRegistry

jeremyBanks:
As described in #786, the `createHash` function currently leaks memory every time it's used: the Hasher instance allocated in the WASM heap is never freed.

This change uses `FinalizationRegistry` (which is working correctly on deno@main as of today 🎉) to .free() the Hasher instances in WASM when their corresponding `Hash` JavaScript objects are garbage collected. This does **not** entirely resolve the memory leak: the WASM heap will never shrink and release the used memory, but it can significantly mitigate it. In a test on my machine, endlessly calling `createHash('sha256')`, the WASM heap grew more slowly, and stopped growing around ~70MB, as the allocator inside the WASM was able to reuse the heap space we freed. Without this change, the memory use continued to grow without bound; I stopped it after 1GB.

This won't be effective in all cases: The garbage collector may not get a chance to run if you're repeatedly creating hashes in a hot sync loop. If you're passing huge chunks of data into WASM, the heap will still need to be expanded to hold them ([discussed in Discord](https://discord.com/channels/684898665143206084/775393009981849600/860494502475988992)). But I think it's still a significant improvement.

It appears that `wasm-bindgen` does have an option that would allow it to do this automatically internally, but [it hasn't been exposed through `wasm-pack` that we're using here, and it's not clear that it will be soon](https://github.com/rustwasm/wasm-pack/issues/930).

---

Test script:

```typescript
#!/usr/bin/env -S deno run --unstable --allow-all
import { createHash } from "./mod.ts";
import { Hash, _wasm } from "./_wasm/hash.ts";


for (let i = 0; i < 1_000_000_000; i++) {
    createHash("sha512");

    if (i % 10_000 === 0) {
      // Yield thread, making it more likely for the GC to run.
      await new Promise((resolve) => setTimeout(resolve, 1));
    }

    if (i % 100_000 === 0) {
      console.log(`
        ${i} hashes created
        ${_wasm.memory.buffer.byteLength} bytes used by WASM`)
    }
  }
```

with

```typescript
export const _wasm = await init(source);
```

added in `hash.ts` to let us check the memory use.

### DIFF
--- a/hash/_wasm/hash.ts
+++ b/hash/_wasm/hash.ts
@@ -17,12 +17,19 @@ await init(source);
 const TYPE_ERROR_MSG = "hash: `data` is invalid type";
 
 export class Hash implements Hasher {
+  static #finalizationRegistry = new FinalizationRegistry(
+    (wasmHash: DenoHash) => {
+      wasmHash.free();
+    },
+  );
+
   #hash: DenoHash;
   #digested: boolean;
 
   constructor(algorithm: string) {
     this.#hash = createHash(algorithm);
     this.#digested = false;
+    Hash.#finalizationRegistry.register(this, this.#hash);
   }
 
   /**
--- a/hash/_wasm/hash.ts
+++ b/hash/_wasm/hash.ts
@@ -16,20 +16,20 @@ await init(source);
 
 const TYPE_ERROR_MSG = "hash: `data` is invalid type";
 
-export class Hash implements Hasher {
-  static #finalizationRegistry = new FinalizationRegistry(
-    (wasmHash: DenoHash) => {
-      wasmHash.free();
-    },
-  );
+const finalizationRegistry = new FinalizationRegistry(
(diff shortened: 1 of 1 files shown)


jeremyBanks:
This seems to be affecting the exit status of a test subprocess in an unexpected way. Converting to draft since this obviously isn't suitable for merge as-is.

jeremyBanks:
It looks like the type error was caused by my using a `#private` `static` property, which TypeScript and SWC supported differently. Switched to a module variable and it seems to be working fine, with tests passing and stable memory use in the latest Deno canary.

jeremyBanks:
@caspervonb suggested we might want to just switch away from `wasm-pack`. That would make this redundant with `wasm-bindgen`'s automatic capability to insert these. Closing this in favour of exploring that.

PR #999, opened 2021-07-03
build(hash): switch from wasm-pack to using wasm-bindgen directly

jeremyBanks:
wasm-pack limits our ability to configure wasm-bindgen and is no longer necessary as we can use wasm-bindgen-cli directly (as @caspervonb pointed out). This updates some rust dependencies and removes our use of wasm-pack. The logic to rebuild the WASM is added to the our GitHub Actions `ci.yaml`.

This PR also modifies the inlined WASM data to insert line breaks every 78 characters, so that it's split across lines instead of being on a single line. Putting it all on a single line can cause performance issues for some debugging or source control tools. It also can give us more useful diffs in some cases: you can see in cb3d9fba193d717dd2600d2dcf12066de80f2a05 that there are some changes to the file when regenerated it, but they are small and don't change the binary size.

Using wasm-bindgen directly allows us to enable `--weakrefs`, which partially mitigates #786 by using a FinalizationRegistry to free Hasher instances in the WASM heap when their corresponding `Hash` JavaScript objects are garbage collected. This does **not** entirely resolve the memory leak: the WASM heap will never shrink and release the used memory, but it can significantly mitigate it. In a test on my machine, endlessly calling `createHash('sha256')`, the WASM heap grew more slowly, and stopped growing around ~30MB, as the allocator inside the WASM was able to reuse the heap space we freed. Without this change, the memory use continued to grow without bound; I stopped it after 1GB. (The test script I used is posted in #998.)

This mitigation won't be effective in all cases: The garbage collector may not get a chance to run if you're repeatedly creating hashes in a hot sync loop. If you're passing huge chunks of data into WASM, the heap will still need to be expanded to hold them ([discussed in Discord](https://discord.com/channels/684898665143206084/775393009981849600/860494502475988992)). But it's still a significant improvement.


.github/workflows/wasm.yml
```diff
@@ -0,0 +1,62 @@
...
+            declare author_email="$(git log -1 --pretty=format:'%ae')"
+            declare current_action_url="https://github.com/${GITHUB_REPOSITORY}/actions/runs/${{ github.run_id }}"
+
+            export GIT_AUTHOR_DATE="${date}"
+            export GIT_AUTHOR_NAME="${author_name}"
+            export GIT_AUTHOR_EMAIL="${author_email}"
+
+            export GIT_COMMITTER_DATE="${date}"
+            export GIT_COMMITTER_NAME="github-actions"
+            export GIT_COMMITTER_EMAIL="41898282+github-actions[bot]@users.noreply.github.com"
```
wperron, line 54: Not super familiar with using github actions to commit changes to the PR -- what is this email address and why does it work?

hash/_wasm/build.ts
```diff
@@ -1,48 +1,78 @@
...
+if (!bindgenStatus.success) {
+  console.error(`Failed to generated wasm bindings: ${bindgenStatus.code}`);
+  Deno.exit(1);
 }
 
-const OUTPUT_DIR = "./out";
+const generatedScript = await Deno.readTextFile("./out/deno_hash.js");
+const generatedWasm = await Deno.readFile("./out/deno_hash_bg.wasm");
+
+// Replace the lines loading the WASM from an external file with our inlined
```
wperron, line 51: :+1: :fire:

wperron (changes requested):

wperron:
I'm also unsure about the github action to rebuild the wasm binary, on the one hand it's nice to have it automated and standardized but on the other hand I'm sure we could simply have a build script and have one of the core maintainers do it manually, after all we don't get that many PRs on the hash module. @bartlomieju wdyt?

.github/workflows/wasm.yml
jeremyBanks: This is just a cosmetic choice and can be dropped if you prefer. It's the email address of the internal GitHub Actions bot account, so by specifying it here the commit gets a GitHub actions logo as its committer, making it stand out more clearly as being generated.

bartlomieju:
> I'm also unsure about the github action to rebuild the wasm binary, on the one hand it's nice to have it automated and standardized but on the other hand I'm sure we could simply have a build script and have one of the core maintainers do it manually, after all we don't get that many PRs on the hash module. @bartlomieju wdyt?

I think it'd be better to have that script integrated into CI otherwise it might quickly go stale and don't work when upgrade time comes up.

wperron:
> I think it'd be better to have that script integrated into CI otherwise it might quickly go stale and don't work when upgrade time comes up.

So a job that runs when the PR is opened so we don't have to think about it as we can be sure it's there when we merge? that's a good idea, I like that :+1:

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,62 @@
...
+  workflow_dispatch:
+
+jobs:
+  rebuild-wasm:
+    runs-on: ubuntu-latest
+    steps:
+        - uses: actions/checkout@v2
+        - name: install dependencies
+          run: |-
+            curl -fsSL https://deno.land/x/install/install.sh | sh
```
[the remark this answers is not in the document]
jeremyBanks: Done, copied the Rust and Deno installation approach from that repo.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,62 @@
+name: "rebuild wasm"
+
+on:
+  workflow_dispatch:
```
wperron, line 4: As per @bartlomieju we can change the trigger to be on PR instead of workflow dispatch, with an if statement in the job to only run when the `hash/` path is modified, I think we have something solid here.

jeremyBanks:
I've merged the WASM Action into the existing `ci.yaml` as @wperron suggested. The logic is now:

1. Check if there were any changes to `hash/` in the latest commit (per `git diff HEAD~ -- ./hash/_wasm`). If not, stop.
2. Install the WASM toolchain dependencies.
3. Rebuild the WASM, bindings, etc.
4. Check whether the rebuilt code has any changes. If not, stop.
5. Run the `wasm/` tests again (only on this Linux build server, not the whole matrix) to ensure that the new build hasn't broken them. If they fail, stop.
6. Push the changes up to the branch.

Commits that are pushed from GitHub actions do not themselves trigger GitHub actions, hence the need to re-test them ourselves (`#5` above).

I've also eliminated the non-determinism we were seeing in the builds (it was a build path in an error string), at least while building on Linux, so this shouldn't create commits unnecessarily.

hash/_wasm/build.ts
```diff
@@ -1,48 +1,88 @@
...
-  if (!status.success) {
-    console.error(`Failed to build wasm: ${status.code}`);
-    Deno.exit(1);
-  }
-}
+    "--target",
+    "wasm32-unknown-unknown",
+  ],
+  env: {
+    // eliminate some potential sources of non-determinism
```
wperron, line 27: I don't think this mitigation works, I still get changes locally after building the latest version of this PR

jeremyBanks: https://github.com/denoland/deno_std/pull/999/files/fa119991d79f7e432111b0546dda24fb06105f88..bae5ded329ea07f4bc062552e6d02d4c10b7253f

Sorry, typo. Fixed.

jeremyBanks:
:facepalm: It looks like there isn't any (reasonable) way to push changes back to a source branch that's in a different repository (i.e., most PRs). Oops. It had seemed to be working because when testing I was running it on `push` within my fork.

But if the build is reproducible and stable, it should be enough to rebuild and verify on GitHub, to make sure that the output matches what was committed by the developer (sort-of like `deno fmt --check`), so I've updated the action to just do that instead.

hash/_wasm/src/lib.rs
```diff
@@ -5,48 +5,48 @@ use wasm_bindgen::prelude::*;
 
 #[wasm_bindgen]
 pub struct DenoHash {
-  inner: Box<dyn DynDigest>,
+    inner: Box<dyn DynDigest>,
```
jeremyBanks, line 8: These changes are the result of `cargo fmt`

wperron: we don't use `cargo fmt` in CI for this repo, I'd just leave it as is for now -- we can address formatting of rust in a separate PR (we typically indent using 2-character wide & spaces)

jeremyBanks: reverted

wperron (approved):
LGTM thanks!

PR #1000, opened 2021-07-03
fix(testing/equals): cater for different class constructor functions

gtramontina:
While writing a polymorphic `Either`, I stumbled upon this weird comparison behavior: I have two classes that are structurally equal `class Left<L> { private value: L }` and `class Right<R> { private value: R }`. When comparing them with `assertEquals` from `https://deno.land/std@0.99.0/testing/asserts.ts`, I've received what feels like a false positive.

**Example**

```typescript
class A {
  private val = 1;
}
class B {
  private val = 1;
}

assertEquals(new A(), new B()); // I'd expect this to fail, but it passes
```

I'm not sure if this is the intended original behavior, but it sure is unexpected. This fix compares whether the object constructors are the same at the very end of the `equals` function.


gtramontina:
By the way, I forgot to mention that I couldn't run the full suite of tests locally, so I was hoping the the pipeline ☝️ would catch some more failures, which it did. 

From what I could tell, it seems that the majority are indeed type differences. Some examples:

```diff
-   Uint8Array(4) [
+   [
```

```diff
-   Buffer(4) [
+   Uint8Array(4) [
```

To me they look like legit/expected failures given the type difference. Keen to hear your thoughts. 💭

wperron:
If you're having issues running the test suite, don't hesitate to reach out, either on the [Discord server](https://discord.gg/deno) or to someone directly (my email is on my github profile, my inbox is always open)

As for the fix itself, I'm not sure I entirely agree that the current behavior is unexpected; JavaScript's type system is based on duck typing, so if an object has property `foo` and has function `bar()` then it's probably the type `Foobar`. It's a subtlety of JavaScript, but the current implementation of `assertEquals` doesn't prevent checking for `assertEquals(a.constructor, b.constructor)`

It'd be interesting to see how other existing JavaScript test frameworks handle this before going forward with this fix.

gtramontina:
Thanks for making yourself available! I've managed to go further with the tests, though. For context, I'm running like the workflow describes, in a docker container – if I run into more issues I'll definitely reach out.
https://github.com/denoland/deno_std/blob/1914407fc82311efd04ab1fc702c239cf6ac70fd/.github/workflows/ci.yml#L32-L33

In relation to other JS test libs, Node's own [assert](https://nodejs.org/api/assert.htm) (both [`.equal`](https://nodejs.org/api/assert.html#assert_assert_equal_actual_expected_message) and [`.strictEqual`](https://nodejs.org/api/assert.html#assert_assert_strictequal_actual_expected_message)) behave the way I described. So does [Chai](https://www.chaijs.com/).

Drawing from experience in other dynamically typed languages such as Ruby and Python where duck-typing is prevalent, equality assertions always fail on different types.

_(opinion notice, take it with a grain of salt 😄)_ I wouldn't rely on duck typing in tests especially, where we want to ascertain behavior, and being of a certain type conveys that. I wouldn't want to work with "probably" in my tests; I would want to be sure. (please note that I put the word "probably" in quotes to emphasize it, not to mock what you said).

bartlomieju:
Thanks for opening the PR @gtramontina! I agree with you that instances of different classes shouldn't be equal to each other - even if they have effectively the same interface (or shape). I think we should go forward with your PR and it's good that it actually caught these discrepancies. Would you mind fixing currently failing tests?

gtramontina:
I'll fix the failing tests when I get some down time and we can take it from there. 👍 Cheers!

testing/asserts.ts
```diff
@@ -202,6 +202,9 @@ export function equal(c: unknown, d: unknown): boolean {
         if (!(a instanceof WeakRef && b instanceof WeakRef)) return false;
         return compare(a.deref(), b.deref());
       }
+      if (a && b && (a.constructor !== b.constructor)) {
```
kitsonk, line 205: Feels like this should go a lot earlier in the check, as it would short circuit much quicker.

caspervonb:
Won't this break a lot of loose comparisons? isn't this why we have `assertStrictEquals`?

testing/asserts.ts
gtramontina: I agree. I tried adding it closer to top, but wasn't being too successful (was getting unexpected failures). So I put it at the end as it was the least disruptive place I could find and make the new test pass. This way I could at least get something out and get feedback. Once I get the rest of the tests fixed, I'll try moving it earlier.

gtramontina:
> Won't this break a lot of loose comparisons? feels like goes against duck typing too far into strict typing.

It didn't break that many… Majority of them were related to buffers and arrays (vide https://github.com/denoland/deno_std/pull/1000/commits/09484d47ed23eb6c3ea77a3053567e3577b347b5).

Regarding duck typing, I've presented my opinion earlier (https://github.com/denoland/deno_std/pull/1000#issuecomment-873694840). Happy to chat more about it…

gtramontina:
Regarding the two remaining failures:

The first one seems to be a real issue with [`_util/deep_assign.ts`](https://github.com/denoland/deno_std/blob/a23eb709ebd4cd8df4b29b0cf6103d4ae530c141/_util/deep_assign.ts) revealed by comparing types. The [failure](https://github.com/denoland/deno_std/pull/1000/checks?check_run_id=3092691582#step:4:3091) presents the following diff:

```diff
-     reg: {},
+     reg: /DENOWOWO/,
```

After a _preliminary_ test, it seems that `deep_assign` might have problems with copying prototypes or something along these lines. I'll see if I can dig a bit deeper here.

The [second failure](https://github.com/denoland/deno_std/pull/1000/checks?check_run_id=3092691582#step:4:3111) is on a test that seems to be asserting on an internal state; potentially on a structure that was meant to remain unexported?

```diff
-   TarEntry {
+   {
```

Any thoughts on these? 💭

bartlomieju:
> Regarding the two remaining failures:
> 
> The first one seems to be a real issue with [`_util/deep_assign.ts`](https://github.com/denoland/deno_std/blob/a23eb709ebd4cd8df4b29b0cf6103d4ae530c141/_util/deep_assign.ts) revealed by comparing types. The [failure](https://github.com/denoland/deno_std/pull/1000/checks?check_run_id=3092691582#step:4:3091) presents the following diff:
> 
> ```diff
> -     reg: {},
> +     reg: /DENOWOWO/,
> ```
> 
> After a _preliminary_ test, it seems that `deep_assign` might have problems with copying prototypes or something along these lines. I'll see if I can dig a bit deeper here.
> 
> The [second failure](https://github.com/denoland/deno_std/pull/1000/checks?check_run_id=3092691582#step:4:3111) is on a test that seems to be asserting on an internal state; potentially on a structure that was meant to remain unexported?
> 
> ```diff
> -   TarEntry {
> +   {
> ```
> 
> Any thoughts on these? 💭

@gtramontina thanks for digging into that, indeed the first failure seems like an actual problem with `deep_assign.ts` that needs to be addressed. As for the second failure, I'm fine with changing the test.

gtramontina:
I've literally just pushed some updates… 😅 I'll add some thoughts as inline comments. I've also adjusted the commit message so it is tagged as a breaking change as pointed out by @caspervonb.

_util/deep_assign.ts
```diff
@@ -30,6 +30,10 @@ object | undefined {
         target[key] = new Date(value);
         return;
       }
+      if (value instanceof RegExp) {
+        target[key] = new RegExp(value);
+        return;
+      }
```
gtramontina, line 36: I'm not quite happy with the fix as it seems `deep_assign` is still flawed. I'm sure sure about its usecases, given it lives in a `_util` directory… Looks like it is supposed to be something private?

Anyway, as fixing this implementation in particular is not the goal of this PR, I guess we could come back to it later.

archive/tar_test.ts
```diff
@@ -397,7 +397,7 @@ Deno.test("untarLinuxGeneratedTar", async function () {
     const content = expected.content;
     delete expected.content;
 
-    assertEquals(entry, expected);
+    assertEquals({ ...entry }, expected);
```
gtramontina, line 400: This seems a bit hacky, but given that `TarEntry` is a private type, downcasting it to object in order to keep the assertion behaving as before seems to be an ok trade-off (?).

kt3k (approved):
@gtramontina Nice work! LGTM

PR #1001, opened 2021-07-03
fix(encoding/toml): serializes mixed array

### LINKED ISSUE #983: TOML incorrectly serializes mixed-type and nested arrays
In the stdlib TOML dumper, arrays whose first element is considered "trivially serializable" (numbers, strings, etc.) are considered to be "trivially serializable", and are thus subject to JSON serialization instead of TOML-specific array serialization. However, this incorrectly assumes that the entire array is "trivially serializable," which it doesn't have to be since the TOML spec dictates that mixed-type arrays are allowed. This can lead to invalid TOML being emitted, such as the following code:
```typescript
import * as TOML from "https://deno.land/std@0.99.0/encoding/toml.ts";

const obj = {a: [1, {b: 2}]};
console.log(TOML.stringify(obj));
```
Which will print:
```
a = [1,{"b":2}]
```
which is invalid TOML. This maybe isn't a huge issue, since the stdlib TOML parser is lenient enough to handle it, but it would be nice to conform to the spec and not just the implementation.

Zzzen:
fixes #983

Previously, arrays are assumed to be homogenous, but that is not the case (or it is changed recently?).

>arrays can contain values of the same data types as allowed in key/value pairs. Values of different types may be mixed.

https://github.com/toml-lang/toml/blob/master/toml.md#user-content-array

### DIFF
--- a/encoding/toml.ts
+++ b/encoding/toml.ts
@@ -559,7 +559,7 @@ class Dumper {
   maxPad = 0;
   srcObject: Record<string, unknown>;
   output: string[] = [];
-  _arrayTypeCache = new Map<unknown[], ArrayType>();
+  #arrayTypeCache = new Map<unknown[], ArrayType>();
   constructor(srcObjc: Record<string, unknown>) {
     this.srcObject = srcObjc;
   }
@@ -630,11 +630,11 @@ class Dumper {
       ["string", "number", "boolean"].includes(typeof value);
   }
   _getTypeOfArray(arr: unknown[]): ArrayType {
-    if (this._arrayTypeCache.has(arr)) {
-      return this._arrayTypeCache.get(arr)!;
+    if (this.#arrayTypeCache.has(arr)) {
+      return this.#arrayTypeCache.get(arr)!;
     }
     const type = this._doGetTypeOfArray(arr);
-    this._arrayTypeCache.set(arr, type);
+    this.#arrayTypeCache.set(arr, type);
     return type;
   }
   _doGetTypeOfArray(arr: unknown[]): ArrayType {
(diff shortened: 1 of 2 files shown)


Aplet123:
I think your unit tests fail to account for the fact that object keys aren't necessarily ordered :)

encoding/toml.ts
```diff
@@ -549,39 +549,41 @@ function joinKeys(keys: string[]): string {
...
+  ONLY_PRIMITIVE,
+  ONLY_OBJECT_EXCLUDING_ARRAY,
+  MIXED,
+}
+
 class Dumper {
   maxPad = 0;
   srcObject: Record<string, unknown>;
   output: string[] = [];
+  _arrayTypeCache = new Map<unknown[], ArrayType>();
```
bartlomieju, line 562: I suggest to use private field (prefixed with `#` instead of `_`)

encoding/toml_test.ts
```diff
@@ -424,6 +424,33 @@ the     = "array"
...
+      mixedArray1: [1, { b: 2 }],
+      mixedArray2: [{ b: 2 }, 1],
+      nestedArray1: [[{ b: 1 }]],
+      nestedArray2: [[[{ b: 1 }]]],
+      deepNested: {
+        a: {
+          b: [1, { c: 2, d: [{ e: 3 }, true] }],
+        },
+      },
+    };
```
bartlomieju, line 440: Make sure that this test exercises all variants from `ArrayType` enum

encoding/toml.ts
Zzzen: 👌 Do you think we should update methods starting with `_` too? Maybe for the whole project?

kt3k: I think we should use `#` for methods too.

kt3k (approved):
LGTM

PR #1003, opened 2021-07-05
Turn off canary builds, turn on stable builds.

ry:


lucacasonato (approved):
Not in favor at all (stable builds don't catch incompatibilities in std like canary does). This has been an issue ever since we enabled canary on std, and I don't think keeping it enabled here for a few more hours/days until I can get it fixed in CLI is much of an issue.

ry:
Screen Shot 2021-07-05 at 3 44 44 PM

We cannot enable canary - even trying to do it now creates a red status here. People contributing patches to deno_std should not encounter our technical debt elsewhere - they don't have the context to understand that it's not their problem.

PR #1005, opened 2021-07-06
feat(http): Allow Passing the exact same path and domain attributes while removing cookies

getspooky:


http/cookie_test.ts
```diff
@@ -3,6 +3,13 @@ import { Response, ServerRequest } from "./server.ts";
 import { deleteCookie, getCookies, setCookie } from "./cookie.ts";
 import { assert, assertEquals, assertThrows } from "../testing/asserts.ts";
 
+Deno.test({
+  name: "Cookie response",
+  fn(): void {
+    //
+  },
+});
+
```
bartlomieju, line 12: Seems like left by mistake?

bartlomieju (commented):
Thanks for the PR @getspooky, does it fix some concrete problem?

http/cookie_test.ts
getspooky: yes, fixed

getspooky:
@bartlomieju  let's say
 ```ts
setCookie({ headers }, { name: "deno", value: "cat"});
setCookie({ headers }, { name: "deno", value: "cat", path: "/demo" });
```
if we run 

```ts
deleteCookie({ headers }, "deno");
```
it will not remove both cookies , so the puporse of this PR is to provide a way to remove cookie by specify path or domain.

for example 

```ts
deleteCookie({ headers }, "deno", { path: "/demo" });
```

bartlomieju (approved):

PR #1006, opened 2021-07-06
feat(async): add debounce method to async module

c4spar:
Resolves: #1008 

This is a proposal for adding a debounce function to the async module.

I always use this function together with the `Deno.watchFs` method and I have seen it in another project as well.
So i think this is a usefull function that could be added to the async module.

It is just a first version of how the module could look like. I'm open for any suggestions and changes.

Example:
```ts
import { debounce } from "https://deno.land/std/async/debounce.ts";

const log = debounce((param: string) => console.log(param), 200);

log("foo");
log("bar");
log("baz");
// wait 200ms ...
// output: baz
```

### DIFF
--- a/async/README.md
+++ b/async/README.md
@@ -6,6 +6,21 @@ async is a module to provide help with asynchronous tasks.
 
 The following functions and class are exposed in `mod.ts`:
 
+## debounce
+
+Debounces a given function by a given time.
+
+```typescript
+import { debounce } from "https://deno.land/std/async/mod.ts";
+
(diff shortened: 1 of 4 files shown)


c4spar:
I'm not sure if the `async` module is the right place for this method, since the `debounce` method doesn't return a promise, but it was the best place I found at the moment.

wperron:
Can you open an issue so we can discuss if it's something we want to include and std and where to put it?

c4spar:
@wperron i have opened an issue #1008

kt3k (approved):
LGTM

PR #1009, opened 2021-07-07
feat(http): Add Cookie domain validation

getspooky:
Cookie domain should not start with `-` and end with `.` or `-`.
https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.3


wperron (approved):
Seems reasonable, but needs more tests

getspooky:
@wperron thanks for feedback, i will add more tests.

PR #1010, opened 2021-07-07
perf(hash): Chunk message input in JavaScript to reduce WASM heap size

jeremyBanks:
It is generally not possible to reduce the size of a WASM module's heap, only grow it (pointed out by 
@evanwashere in Discord, see also https://github.com/WebAssembly/design/issues/1300). There are some possible work-arounds, but they would add complexity and might not be possible with our current wasm-bindgen/wasm-pack build approach.

When you `.update(...)` our WASM Hasher with a message, that data needs to be copied over to the WASM heap, which may require it to grow if it doesn't have enough available space. So if you pass a message that's 10MB, or 2GB, the WASM heap and memory usage will increase to at least that much and never come down.

This change instead splits messages into chunks of up to 64KB each and feeds them to WASM individually, keeping the total WASM heap size under a few megabytes regardless of the message size.

I benchmarked this to ensure it wouldn't adversely affect throughput, and found that as long as the piece size isn't too small, throughput is very-slightly improved relative to the existing code. I used 64KB chunks because they seemed the smallest fastest chunk size in my testing, by a small margin.


kt3k:
@jeremyBanks This seems like a decent fix to me, but is there any way to check this actually reduces the resulted heap size? Ideally we'd like to have a test case for checking it.

jeremyBanks:
@kt3k Yes, we can check wasm.memory.buffer.byteLength. I had an ad hoc test script but it wasn't a clean unit test so I didn't commit it. I'll see if I can fix it up later (I guess I'll need to run the tests in subprocesses to reset the heal between them.)

jeremyBanks:
@kt3k I have added a test for memory use, ensuring that it stays under 2MB when provided with a 64MB input. Without the change, it would fail with this message:

```
AssertionError: WASM heap was too large after large input: 65.2 MB
```

kt3k (approved):
@jeremyBanks Great! LGTM. Thanks!

PR #1012, opened 2021-07-07
fix(hash): Fix broken handling of non-byte-sized TypedArray views

jeremyBanks:
Currently, hash.update() corrupts input data if it comes from a TypedArray with items larger than a byte. The main source of trouble is:

```ts
      if (data instanceof ArrayBuffer || ArrayBuffer.isView(data)) {
        msg = new Uint8Array(data);
```

If the input is an ArrayBuffer or a Uint8Array, this is correct. However, if it's an Int16Array, this will cause each 16-bit value to be cast to an 8-bit value, and those values will be hashed, as you could see in an example like:

```ts
console.log(new Uint8Array(new Int16Array([1000])));
// Uint8Array [232]
```

Instead, it seems like the intended behaviour should be to hash the underlying bytes that the Int16Array is a view over, instead of a lossy coercion.

This commit updates our implementation to do that, and adds several test cases to make sure that various views of the same data produce the same result. It also avoids creating unnecessary copies of input data when we can instead just use another view of the same data (we consume the input synchronously, so we don't need to worry about it being changed under us).

Co-authored-by: @evanwashere <evan@tsu.sh>


jeremyBanks:
Marked as draft because this implementation is on top of the unmerged changes in https://github.com/denoland/deno_std/pull/1010, but if that isn't good to merge I could update this to go first instead.

kt3k (approved):
@jeremyBanks Thanks! Nice fix! LGTM

Issue #823, opened 2021-03-28
TOML parser's inline/nested tables misaligned with real TOML


danopia:
I'm comparing this TOML parse with http://toml-online-parser.ovonick.com

First opened here because I'm dumb: https://github.com/denoland/deno/issues/9919

## case 1

This is proper TOML (AFAIK) that parses correctly elsewhere:

```toml
annotation_filter = { "kubernetes.io/ingress.class" = "nginx" }
```

... but Deno's parser tries reading as JSON:

```
> TOML.parse(`annotation_filter = { "kubernetes.io/ingress.class" = "nginx" }`);
error: Uncaught SyntaxError: Unexpected string in JSON at position 31
      return JSON.parse(dataString);
                  ^
    at JSON.parse (<anonymous>)
    at Parser._parseInlineTableOrArray (https://deno.land/std@0.91.0/encoding/toml.ts:322:19)
    at Parser._parseData (https://deno.land/std@0.91.0/encoding/toml.ts:275:21)
    at Parser._processDeclaration (https://deno.land/std@0.91.0/encoding/toml.ts:264:24)
    at Parser._parseLines (https://deno.land/std@0.91.0/encoding/toml.ts:492:25)
    at Parser.parse (https://deno.land/std@0.91.0/encoding/toml.ts:541:10)
    at Module.parse (https://deno.land/std@0.91.0/encoding/toml.ts:735:33)
```

## case 2

As a work around, I tried this non-inline syntax instead:

```toml
[[aaa]]
hi = "hi"
[aaa.bbb]
asdf = "asdf"
```

... but the key just silently gets lost:

```
> TOML.parse(`[[aaa]]\nhi = "hi"\n[aaa.bbb]\nasdf = "asdf"`)
{ aaa: [ { hi: "hi" } ] }
```

I was expecting:

```json
{ "aaa": [ {
            "hi": "hi",
            "bbb": {
                "asdf": "asdf"
            }
} ] }
```

## case 3

This actually works with Deno as of std@0.91.0:
```toml
[[source]]
annotation_filter = { "\"kubernetes.io/ingress.class\"": "nginx" }
 # also this comment is read as a key because of leading spaces = "..."
```

That last line seems to be included because of some parsing bug:

```
> TOML.parse(`[[source]]\nannotation_filter = { "\\"kubernetes.io/ingress.class\\"": "nginx" }\n # also this comment is read as a key because of leading spaces = "..."`)
{
  source: [
    {
[...]

PR #1017, opened 2021-07-10
fix(encoding/toml): comment line starting with whitespaces

FujiHaruka:
Fix case 3b of https://github.com/denoland/deno_std/issues/823#issuecomment-814367934

As @danopia clearly showed an example, TOML Parser parses a comment line starting with whitespaces

```
 # comment = "..."
```

as a key-value:

```json
{
  "# comment": "..."
}
```

This PR fixes it.


kt3k (approved):
Thanks! LGTM

PR #1021, opened 2021-07-10
test(bytes): improve tests for equals, startsWith and endsWith

grian32:
This pr does 3 things to all tests for those:
 - Adds tests to check that it fails when it should
 - Adds tests to check that it also fails when array sizes are different
 - Changes `assertEquals(v, true)` to `assert(v)`


wperron (approved):
LGTM, thanks!

PR #1022, opened 2021-07-10
feat(async): add `timeout` to async module

lambdalisue:
Hi, I'm not sure if you like the idea but I added a `timeout` function to the `async` module while simple code with `delay(100).then(() => new Error("Timeout"))` would cause the following error

```
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 0
  - completed: 0
After:
  - dispatched: 4
  - completed: 3

Make sure to await all promises returned from Deno APIs before
```
Added `timeout` function call `clearTimeout` to avoid such kind of error.

Please close this PR if it does not suit the std requirements.

### DIFF
--- a/async/README.md
+++ b/async/README.md
@@ -130,3 +130,16 @@ const [branch1, branch2] = tee(gen());
   }
 })();
 ```
+
+## deadline
+
+Create a promise which will be rejected with `DeadlineError` when a given delay
+is exceeded.
+
+```typescript
+import { deadline } from "https://deno.land/std/async/mod.ts";
+
+const delayedPromise = delay(1000);
+// Below throws `DeadlineError` after 10 ms
+const result = await deadline(delayedPromise, 10);
+```
--- a/async/deadline.ts
+++ b/async/deadline.ts
@@ -1,0 +1,18 @@
+import { deferred } from "./deferred.ts";
+
+export class DeadlineError extends Error {
+  constructor() {
+    super("Deadline");
+    this.name = "DeadlineError";
+  }
+}
+
+/**
+ * Create a promise which will be rejected with DeadlineError when a given delay is exceeded.
+ */
+export function deadline<T>(p: Promise<T>, delay: number): Promise<T> {
+  const d = deferred<never>();
+  const t = setTimeout(() => d.reject(new DeadlineError()), delay);
+  p.finally(() => clearTimeout(t));
+  return Promise.race([p, d]);
+}
--- a/async/deadline_test.ts
+++ b/async/deadline_test.ts
@@ -1,0 +1,20 @@
+import { assertEquals, assertThrowsAsync } from "../testing/asserts.ts";
+import { deferred } from "./deferred.ts";
+import { deadline, DeadlineError } from "./deadline.ts";
+
+Deno.test("[async] deadline: return fulfilled promise", async () => {
+  const p = deferred();
+  const t = setTimeout(() => p.resolve("Hello"), 100);
+  const result = await deadline(p, 1000);
+  assertEquals(result, "Hello");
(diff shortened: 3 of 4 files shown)


async/timeout.ts
```diff
@@ -0,0 +1,18 @@
...
+  constructor() {
+    super("Timeout");
+    this.name = "TimeoutError";
+  }
+}
+
+/**
+ * Create a promise which will be rejected with TimeoutError when a given delay is exceeded.
+ */
+export function timeout<T>(p: Promise<T>, delay: number): Promise<T> {
```
wperron, line 13: I don't really like the name, I think it's too close to `setTimeout` and I expect would cause confusion. Maybe `race` or `deadline` would be a better name for this function.

async/timeout.ts
```diff
@@ -0,0 +1,18 @@
+import { deferred } from "./deferred.ts";
+
+export class TimeoutError extends Error {
```
wperron, line 3: I don't think we need a concrete error type here, a simple `Error` would do just fine.

wperron (changes requested):

async/README.md
```diff
@@ -130,3 +130,16 @@ const [branch1, branch2] = tee(gen());
...
+
+Create a promise which will be rejected with `TimeoutError` when a given delay
+is exceeded.
+
+```typescript
+import { timeout } from "https://deno.land/std/async/mod.ts";
+
+const delayedPromise = delay(1000);
+const result = await timeout(delayedPromise, 10);
+// Above throws `TimeoutError` after 10 ms
```
wperron, line 144: Thanks for updating the docs :+1: 

This comment might be better placed above the related line rather than below, as would be the case in source code anyway.

async/timeout.ts
lambdalisue: OK. I'll use `deadline` 👍

async/timeout.ts
lambdalisue: Well... Then how can we distinguish if an error occurred from an original promise or `timeout` itself?
I think the `timeout` adds extra context which is not a part of the core concern of the original promise so users would want to distinguish the error.

async/README.md
lambdalisue: OK. I'll fix it.

lambdalisue:
I'll squash commits once reviews are completed.

async/README.md
```diff
@@ -130,3 +130,16 @@ const [branch1, branch2] = tee(gen());
...
+## deadline
+
+Create a promise which will be rejected with `DeadlineError` when a given delay
+is exceeded.
+
+```typescript
+import { deadline } from "https://deno.land/std/async/mod.ts";
+
+const delayedPromise = delay(1000);
+// Above throws `DeadlineError` after 10 ms
```
wperron, line 143: Except it's not above anymore :wink:

async/deadline.ts
```diff
@@ -0,0 +1,18 @@
+import { deferred } from "./deferred.ts";
+
+export class DeadlineError extends Error {
```
wperron, line 3: I double checked the rest of the std lib, it seems fine, let's keep it :+1:

wperron (commented):
There's just a small thing to fix in the docs, otherwise LGTM

wperron:
Also, just make sure you rebase, but you don't _have_ to squash, we can do it on merge ourselves

async/README.md
lambdalisue: Oops. I'll reword.

lambdalisue:
Rebase & force pushed 👍

wperron (approved):
LGTM, thanks a lot for the contribution!

PR #1024, opened 2021-07-11
fix(path): Add question mark as a glob indicator

### LINKED ISSUE #1023: expandGlob with `?` as only placeholder never matches.
The [`expandGlob`](https://deno.land/std@0.100.0/fs/expand_glob.ts#L61) function does not match patterns that only include `?`

I assume you have a `/tmp` folder.
1. Run `deno repl`
2. import expandGlob `const {expandGlob} = await import("https://deno.land/std@0.100.0/fs/expand_glob.ts");`
3. run `for await (const {path} of expandGlob('/tmp')) console.log(path);` which prints `/tmp`
4. run `for await (const {path} of expandGlob('/t?p')) console.log(path);` which prints nothing
5. run `for await (const {path} of expandGlob('/t?p*')) console.log(path);` which prints `/tmp` again

If you use your local bash/zsh, you can counter check
1. `echo /tmp` obviously prints `/tmp`
2. `echo /t?p` still prints `/tmp`
3. `echo /t?p*` also prints `/tmp`

Mac OS 11.4 (20F71)

```
deno 1.11.5 (release, x86_64-apple-darwin)
v8 9.1.269.35
typescript 4.3.2
```

The documentation of [`globToRegExp`](https://doc.deno.land/https/deno.land/std@0.100.0/path/glob.ts#globToRegExp) does not mention `?`, but it [is implemented there](https://deno.land/std@0.100.0/path/glob.ts#L231).

nayeemrmn:
Fixes #1023.

### DIFF
--- a/path/glob.ts
+++ b/path/glob.ts
@@ -36,6 +36,7 @@ const rangeEscapeChars = ["-", "\\", "]"];
  *
  * Basic glob syntax:
  * - `*` - Matches everything without leaving the path segment.
+ * - `?` - Matches any single character.
  * - `{foo,bar}` - Matches `foo` or `bar`.
  * - `[abcd]` - Matches `a`, `b`, `c` or `d`.
  * - `[a-d]` - Matches `a`, `b`, `c` or `d`.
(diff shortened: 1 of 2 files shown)


wperron (approved):
LGTM, Thanks @nayeemrmn !

Issue #652, opened 2020-09-01
Review the std/hash Hasher interface


caspervonb:
The Hasher interface defined in std/hash might warrant a review.

First thing I noticed was that Hasher.digest returns an ArrayBuffer, This is more or less the only interface we have that doesn't return an Uint8Array view of the generated ArrayBuffer.

Other concerns, @kitsonk ?

jeremyBanks:
I have proposed an expanded interface:

- https://github.com/denoland/deno_std/discussions/1014

PR #1025, opened 2021-07-12
feat(crypto): add std/crypto wrapping and extending runtime WebCrypto

jeremyBanks:
## Background

The current `std/hash` implementation (inspired by Node's `crypto.Hash` interface) is largely redundant with the standard WebCrypto `crypto.subtle.digest` function that the runtime will soon support. Deno tries to aim for interoperability, so we probably want to eliminate our redundant interface eventually.

However, WebCrypto only supports a very minimal set of hash digest algorithms, and our users may need others. We don't want to add these to our runtime's WebCrypto implementation, as this is discouraged by the standard and could harm interoperability.

See discussion at https://github.com/denoland/deno_std/pull/1025#issuecomment-881747166.

## Description

Attempting to balance these concerns, this PR adds  `std/crypto`, which exports a single object extending the standard Crypto/SubtleCrypto interface. For most operations, it will simply re-export the methods from the runtime's WebCrypto. However, for digest operations using an algorithm that WebCrypto doesn't support, it will instead use our bundled Rust/WASM implementation. This allows users to adopt it with minimal code changes and deviation from the standard interface, while maintaining support for other runtimes (such as browsers).

```ts
// using runtime implementation directly:
const digestA = await crypto.subtle.digest("SHA-384", new ArrayBuffer(0));

// throws an error as BLAKE3 is not supported in the WebCrypto standard:
const digestB = await crypto.subtle.digest("BLAKE3", new ArrayBuffer(0));
```

```ts
import crypto from "https://deno.land/std/crypto/mod.ts";

// using the runtime implementation transparently through our wrapper:
const digestA = await crypto.subtle.digest("SHA-384", new ArrayBuffer(0));

// using our bundled WASM implementation of BLAKE3:
const digestB = await crypto.subtle.digest("BLAKE3", new ArrayBuffer(0));
```

This is intended to replace the existing `std/hash` eventually, but we leave it untouched for now (despite the redundancy) as [that removal should wait for future release](https://github.com/denoland/deno_std/pull/1025#issuecomment-883830855).

The `std/node/crypto` compatibility module has features (incremental `.update()`s, `.copy()`able hash state) that aren't part of the standard WebCrypto interface. So instead of depending on `std/hash` or `std/crypto`, it will now depend directly on the WASM implementation, which we move to a common location, `std/_wasm_crypto`.

## Features

Capabilities that `std/crypto` adds on top of the runtime's `crypto`:

- Support for additional hash digest algorithms:
  - `BLAKE2B-256`
  - `BLAKE2B-384`
  - `BLAKE2B`
  - `BLAKE2S`
  - `BLAKE3`
  - `KECCAK224`
  - `KECCAK256`
  - `KECCAK384`
  - `KECCAK512`
  - `MD5`
  - `RIPEMD-160`
  - `SHA-224`
  - `SHA3-224`
  - `SHA3-256`
  - `SHA3-384`
  - `SHA3-512`
  - `SHAKE128`
  - `SHAKE256`
- Adds `{ length?: number }` parameter for extendable-output algorithms ("XOFs": `BLAKE3`, `SHAKE128`, `SHAKE256`). For example, `{ name: "BLAKE3", length: 20 }`. This is consistent with how WebCrypto parameterizes other algorithms.
- Adds `crypto.subtle.digestSync`, a synchronous version of `.digest` with the same arguments.
- Adds support for `AsyncIterable` and `Iterable` inputs for `.digest()` (and `Iterable` for `.digestSync()`), to allow incremental digesting of large inputs. (This is a superset of the proposed addition of ReadableStream support to the spec: https://github.com/w3c/webcrypto/issues/73.)

## Performance

Benchmarking script included. This is currently faster than the runtime's WebCrypto implementation ([see related discussion in Discord](https://discord.com/channels/684898665143206084/775393009981849600/867096350801920040)), so I think it's fast enough for now. There is probably room to improve both going forward.

## Related Issues

- resolves https://github.com/denoland/deno_std/issues/652
- resolves https://github.com/denoland/deno_std/issues/786

---

<details>
<summary>
Original Description: <b>BREAKING(hash): Expand hashing interface in line with other languages</b>
</summary>

<br>

The WASM-based Hasher implementation in `hash/` currently exposes an [initial](https://github.com/denoland/deno/pull/6292#pullrequestreview-431895698) minimum-viable interface. It requires a new Hasher instance to be created for each message, and once `.digest()` or `.toString()` is called for the first time, the instance is invalidated and any further method calls will throw. This interface is usable (modulo implementation issues, see #999, #1010, #1012), but it lacks some of the conveniences that exist in some other languages' libraries. (I've been looking at [Python](https://docs.python.org/3/library/hashlib.html), [Rust](https://docs.rs/digest/0.9.0/digest/), [Go](https://golang.org/pkg/hash/), and [WebCrypto](https://www.w3.org/TR/WebCryptoAPI/#subtlecrypto-interface-methods).) Needing to create new instances for every input can also be a significant source of overhead in some cases (such as performing a lot of hashes in a hot loop, where the GC may not have a chance to clean them up very often), and the interface doesn't provide any way to avoid that if you need to.

I propose this update to the hashing interface to bring it more in line with what other languages offer.

## Aside re: WebCrypto

I'm not sure whether or not this is redundant with the ongoing work on WebCrypto. I think it might still be useful, as this implementation provides several more algorithms than WebCrypto's conservative specification. This implementation is synchronous, while WebCrypto can only be called async. This implementation should be browser-compatible, and can be used on non-secure origins where WebCrypto is unavailable. However, this implementation will presumably be a bit slower than WebCrypto, particularly for algorithms that are able to exploit parallelism.

This is up to the team to decide, but I would just say that if we are going to keep `std/hash`, then it deserves an overhaul before 1.0.

## Interface Changes

- BREAKING CHANGE: `.digest()` returns a `Uint8Array` instead of an `ArrayBuffer`, as suggested at https://github.com/denoland/deno_std/issues/652#issue-798302526.
- `.digest()` and `.toString()` can be called multiple times without side-effects.
- `.reset()` and `.digestAndReset()` methods are added to allow instance reuse.
- A `.clone()` method is added to allow more efficient calculation of messages from a common prefix.
- Support is added for the `shake128`, `shake256`, `blake2b`, `blake2b-256`, `blake2b-384`, and `blake2s` algorithms, all of which are also supported in current versions of Python's hashlib or Go's crypto package, and have implementations available from the RustCrypto project we already depend on.
- `.digest()` now takes an optional `options` object with a `length` property.
  - `length` is intended for use with algorithms that support variable-length digest output, such as `blake3` and `shake128`. Specifying a `length` for an algorithm with a fixed-length output will throw an error unless the specified size matches that fixed size.
- `.toString()` now takes an options object with the property `encoding` to replace the previous positional `formatting` argument. This option object can also contain digest options (`length`).
  - The previous function signature is still also supported for compatibility.
- A`"unicode"` `encoding` option is added to `.toString()`; this encodes byte values as unicode code points, in the same manner as the `bota()`/`atob()` functions.
- A module-level `digest(algorithm, message, options?)` function was added, which internally reuses a hasher instance for calls with the same algorithm, providing users with an efficient option for the most common simple use cases.
  - with this option and previous fixes, I think we can mark https://github.com/denoland/deno_std/issues/786 as resolved.
- BREAKING CHANGE: Renamed `createHash` to `createHasher` since the type is called `Hasher`, not `Hash`.

## Implementation Changes

- Errors thrown from Rust are now proper `Error` object with stack traces (they were previously just strings without stack traces).
- The generated bindings and generated WASM binary are split into separate files, as the bindings are now getting larger and more significant.
- Removes the only remaining internal reference to the JavaScript hash implementations, in case we want to remove them as suggested at https://github.com/denoland/deno_std/issues/658.

## Documentation Changes

- Added more detailed docstrings to the public interface.
- README: Replaced use of MD5 in examples with SHA-384 to help guide inexperienced users towards a safer option.
- README: Added links to canonical references for each supported hash algorithm.

## Out of Scope

Although these could all make reasonable inclusions in the module, this PR does not include:

- a way to read the output of an extendable hash digest (XOF) as an infinite stream.
- keyed-hashing/HMAC support.
- incremental validation capabilities supported by hashes like BLAKE3 or KangarooTwelve.
- 
</details>


jeremyBanks:
The CI lint failure appears to be a pre-existing violation -- maybe it's being flagged because I made a change in dependent code? I don't think it's related to my changes, so I'm ignoring it.

hash/_wasm/src/lib.rs
```diff
@@ -1,52 +1,262 @@
...
+/// A Hasher for any supported cryptographic hash algorithm, wrapping them in
+/// a common interface for our WASM use.
 #[wasm_bindgen]
-pub struct DenoHash {
-  inner: Box<dyn DynDigest>,
+#[derive(Clone)]
+pub struct Hasher {
+    algorithm: &'static str,
+    inner: HashBox,
 }
```
bnoordhuis, line 13: Can you run `cargo fmt`? In particular, deno uses two spaces for indentation, not four.

bnoordhuis (commented):
I can't make the call on whether this should go in or not but I don't see any glaring technical issues.

Caveat emptor, I only did a real quick review. I'll do a second pass if there's consensus it's eligible for inclusion.

wperron:
I don't have any issues to the changes to `digest` and `toString` nor the
addition of `reset` -- Maybe `resetAndDigest` can be ommited? `clone` seems a
bit weird to me but I don't really have strong reason to not include it.

The addition of the option bags on `digest` and `toString` seem a bit superflous
on the surface -- Do they fix an actual issue, or does it make the API easier to
use in a significant way? What would be downside of leaving those out?

As for the renaming of `createHash` to `createHasher` I'm not sure; I agree that
it's annoying to have the function name and interface name differ slightly, but
does that really require a breaking change?

jeremyBanks:
I'm ready to make the suggested changes, but just to lay out my thinking.

It's true that these are backwards-incompatible changes. But given the previous state of the API - badly leaking memory, erroneous on certain data types - I doubt it has much adoption, and if we're making any breaking changes, I'm inclined to go ahead and clean up while we're at it. (Mainly re: the createHasher rename). 

Regarding the option bags, I agree they seemed like overkill for the current options. However, the Deno style guide suggests no more than two positional arguments, and the digest function requires at least three arguments. It was also a forward compatibility consideration, as if we decide to later add support for signed hashing modes or whatever, those options could join them, rather than having a mix of positional and keyword arguments.

Regarding clone: I agree that the use case is a bit obscure, but it's been supported in most of the other hashing libraries I've looked at, and I don't think it adds much bloat.

Regarding digestAndReset: this one is a stretch. I added it because for the underlying Rust implementations, this is faster than the separate calls, because this allows it to destructively finalize the existing state as its going to be reset, instead of saving a copy of the existing state as digest on its own would do. This is a micro-optimization, maybe unnecessary, but the Rust API suggested it to me.

Regarding formatting: I'll see if I can copy in the standard Debo rustfmt config, since my editor keeps auto formatting incorrectly.

wperron:
We're still in version 0.X of the standard lib so we do have some leeway to make breaking changes here. Unfortunately we don't yet have data on downloads for the standard lib, so exactly how much usage it gets in the wild is a bit of a shot in the dark at the moment.

Pending any more comments from @bnoordhuis I'd be ok with landing as is.

kitsonk:
I am of the opinion that we should take a breaking change in std/hash, but:

- It should line exactly to the API of WebCrypto.
- It should only include hashes that WebCrypto doesn't support.  We want people to use the web standards and we don't want to duplicate APIs or functionalities of the web standards.
- We should get rid of the JavaScript only interfaces.  The main reason they were remaining was that the web assembly versions didn't provide HMAC key generation, which is pretty critical for a lot of workloads.

Anything over and above that isn't `std` library, it is a community package, IMO.

hash/_wasm/src/lib.rs
jeremyBanks: Copied in deno core's `.rustfmt.toml` and applied it.

jeremyBanks:
I defer to the maintainers on the subject of scope (I do not love WebCrypto, but I do love Deno's interoperability goals), but for what it's worth: it looks like Node's built-in `crypto` module provides [synchronous hashing interfaces](https://nodejs.org/api/crypto.html#crypto_class_hash), which can't be implemented on top of WebCrypto's async APIs, so the Deno standard library may need to include its own implementation for that purpose if nothing else. (Although I suppose that doesn't necessarily mean the implementation needs to be exposed anywhere else.)

jeremyBanks:
I'm working on updating this to align with the WebCrypto interface. Currently I'm imagining the following:

```typescript
import stdCrypto from "https://deno.land/std@$STD_VERSION/crypto/mod.ts";

// Use the built-in `crypto` interface for WebCrypto-supported algorithms:
console.log(await crypto.subtle.digest("SHA-384", "hello world"));

// Use std/crypto for other supported algorithms:
console.log(await stdCrypto.subtle.digest("BLAKE3", "hello world"));
```

If they are decided to be in-scope, sync implementations of supported algorithms could go under `stdCrypto.subtle.digestSync`.

```ts
console.log(await stdCrypto.subtle.digestSync("BLAKE3", "hello world"));
```

If they are decided to be in-scope, variable-length digests of supported algorithms could use an added optional added `length` argument (Deno's style guide would suggest an options object, but WebCrypto only uses positional arguments):

```ts
console.log(await stdCrypto.subtle.digest("BLAKE3", "hello world", 64));
```

littledivy:
Having two different implementations of the same interface doesn't feel right to me. The WebCrypto spec allows introducing additional algorithms so why can't we add to it? 

From https://w3c.github.io/webcrypto/#algorithms-section-overview -
> ...Further, this specification defines a process to allow additional specifications to introduce additional cryptographic algorithms. 

Node.js did a similar thing with it's `NODE-ED25519`

jeremyBanks:
@littledivy The spec allows for that, but it's "strongly discouraged":

> ### [3.1 Extensibility](https://w3c.github.io/webcrypto/#extensibility)
> 
> Vendor-specific proprietary extensions to this specification are strongly discouraged. Authors must not use such extensions, as doing so reduces interoperability and fragments the user base, allowing only users of specific user agents to access the content in question.
>
> If vendor-specific extensions are needed, the members should be prefixed by vendor-specific strings to prevent clashes with future versions of this specification. Extensions must be defined so that the use of extensions neither contradicts nor causes the non-conformance of functionality defined in the specification.

Adding these algorithms to WebCrypto seems like the less interoperable option, because code that depends on it will fail on non-Deno environments, while the WASM implementations allows us to author code that's browser-compatible.

kitsonk:
I agree that it makes it difficult to sniff and therefore difficult to make isomorphic by adding them to WebCrypto. Having it part of std opens up the possibility of people using it on other runtimes.

Just thinking about usability, I would personally be fine if std offloaded to Web Crypto for hashes that are part of the Web standard. I just want to avoid duplicate implementations. It is bad for users and it is bad trying to chase down bugs.

jeremyBanks:
I've updated the implementation and description based on the discussion above. It still needs a bit more polish and updated tests, but it's mostly there.

kitsonk:
@jeremyBanks thanks a lot for the work... one thing I just discussed with @kt3k is that to make this land-able, we probably need to add the new `/crypto` modules and land the removal of existing modules separately, at a later date.  I think we need to give people a Deno CLI release cycle to adjust.  I know of a few projects that depend on these modules and will have either update to use only the web crypto APIs in Deno or use these modules.

My opinion is we can add the new modules before Deno 1.13, but we shouldn't remove the existing ones until the `std` release that comes with Deno 1.14.

We might want to consider adding some sort of "deprecated" message to the existing APIs.

jeremyBanks:
Reverted all changes to std/hash.

Added a test for `digest()`ing a stream (https://github.com/w3c/webcrypto/issues/73), since that's likely to be officially supported in the future per the new WebCrypto spec editor.

_wasm_crypto/algorithms.ts
```diff
@@ -0,0 +1,77 @@
...
+] as const;
+
+/** A cryptographic hash/digest algorithm supported by std/_wasm_crypto. */
+export type DigestAlgorithm = typeof digestAlgorithms[number];
+
+/**
+ * Other platforms may identify hash algorithms by slightly different names,
+ * so we use this map of some common variations to the names we support.
+ */
+export const digestAliases: Record<string, DigestAlgorithm> = {
```
kitsonk, line 45: I think there needs to be a good reason why we would alias them. I think we should provide a canonical name.  Allowing aliases leads to accidents and we don't really want accidents with security related software. If the canonical algorithm name isn't available, we should just throw, maybe with a list of supported alogrithms.

_wasm_crypto/crypto.js
```diff
@@ -0,0 +1,344 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// deno-lint-ignore-file
```
kitsonk, line 2: why? If it is because this is essentially a code-gen file (which I believe it is) then we should put a note here to that effect.

_wasm_crypto/crypto.js
```diff
@@ -0,0 +1,344 @@
...
+    return v1;
+  } finally {
+    wasm.__wbindgen_add_to_stack_pointer(16);
+  }
+}
+
+const DigestContextFinalization = new FinalizationRegistry((ptr) =>
+  wasm.__wbg_digestcontext_free(ptr)
+);
+/**
```
kitsonk, line 157: I am assuming this is codegen as well, which indicates we are missing some sort of comment in the source? We should have at least something there as these flow through to intellisense.

_wasm_crypto/src/lib.rs
```diff
@@ -0,0 +1,110 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+use derive_more::{From, Into};
+use js_sys::Uint8Array;
+use wasm_bindgen::prelude::*;
+
+mod digest;
+
+#[wasm_bindgen]
+#[derive(Clone, Into, From)]
+pub struct DigestContext(digest::Context);
```
kitsonk, line 10: In the gen code, this is missing the JSDoc, we should have rust doc blocks for publish structures like this as they will flow through to the intellisense for the library.

_wasm_crypto/src/lib.rs
```diff
@@ -0,0 +1,110 @@
...
+  pub fn new(algorithm: &str) -> Result<DigestContext, JsValue> {
+    Ok(
+      digest::Context::new(algorithm)
+        .map_err(|message| JsValue::from(js_sys::TypeError::new(message)))?
+        .into(),
+    )
+  }
+
+  #[wasm_bindgen]
+  pub fn update(&mut self, data: js_sys::Uint8Array) -> Result<(), JsValue> {
```
kitsonk, line 36: and on these public methods, we should probably have doc blocks for these

crypto/mod.ts
```diff
@@ -0,0 +1,247 @@
...
+  }
+  return bytes;
+};
+
+/**
+ * An wrapper for WebCrypto adding support for additional non-standard
+ * algorithms, but delegating to the runtime WebCrypto implementation whenever
+ * possible.
+ */
+const stdCrypto = (<T extends WebCrypto>(x: T) => x)({
```
kitsonk, line 134: My original objection to dropping the JavaScript/TypeScript implementations in lieu of the existing web assembly ones was that the existing web assembly ones didn't support signing specifically with HMAC, while the JavaScript implementations did, and I have a project (oak) that uses signing.  Of course now that we have WebCrypto, for the 4 web crypto algorithms, we would still have signing with HMAC (and the other types supported by web crypto) but it seems like we should try to support at least signing as part of this.  It doesn't have to be "now" but I think it should be before we drop any of the JavaScript/TypeScript implementations that are not covered by Web Crypto.

crypto/mod.ts
```diff
@@ -0,0 +1,247 @@
...
+  ...webCrypto,
+  subtle: {
+    ...webCrypto.subtle,
+
+    async digest(
```
kitsonk, line 139: need some JSDoc here

crypto/mod.ts
```diff
@@ -0,0 +1,247 @@
...
+        return await webCrypto.subtle.digest(
+          algorithm,
+          data as unknown as Uint8Array,
+        );
+      } else {
+        throw new TypeError(`unsupported digest algorithm: ${algorithm}`);
+      }
+    },
+
+    digestSync(
```
kitsonk, line 197: Some JSDoc here

crypto/mod.ts
```diff
@@ -0,0 +1,247 @@
...
+
+type DigestAlgorithm = DigestAlgorithmName | DigestAlgorithmObject;
+
+const normalizeAlgorithm = (algorithm: DigestAlgorithm) =>
+  ((typeof algorithm === "string") ? { name: algorithm.toUpperCase() } : {
+    ...algorithm,
+    name: algorithm.name.toUpperCase(),
+  }) as DigestAlgorithmObject;
+
+export default stdCrypto;
```
kitsonk, line 247: We shouldn't use `export default`. We only use it in `/node` for API compatibility reasons.

kitsonk (commented):
Looks good, but a few thoughts.

crypto/mod.ts
jeremyBanks: That sounds like a good plan. I don't want to try to take on the signing as part of this PR, but it would be nice to get it done soon.

_wasm_crypto/algorithms.ts
jeremyBanks: I've moved this into `node/crypto` and pared it down to differently-formatted hash identifiers I saw from Node's `.getHashes()`.

_wasm_crypto/_build.ts
```diff
@@ -0,0 +1,147 @@
...
+}
+
+// for testing/debugging
+export const _wasm = wasm;
+export const _wasmModule = wasmModule;
+export const _wasmInstance = wasmInstance;
+export const _wasmBytes = wasmBytes;
+`;
+
+await Deno.writeTextFile("./crypto.wasm.ts", wasmJs);
```
kt3k, line 129: I think this file should be JavaScript instead of TypeScript. The file only exports data, and they are consumed internally. So there seem no benefit of making it TypeScript, but it causes type check overhead at the start up.

jeremyBanks: makes sense, done.

_wasm_crypto/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+[package]
```
littledivy, line 1:
```suggestion
# Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
[package]
```

crypto/mod.ts
```diff
@@ -0,0 +1,275 @@
+import {
```
littledivy, line 1:
```suggestion
// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
import {
```

littledivy (changes requested):
Minor nits :smile:

crypto/mod.ts
jeremyBanks: done

_wasm_crypto/Cargo.toml
jeremyBanks: done

jeremyBanks:
Marked as draft to do a bit more testing and clean-up.

_wasm_crypto/README.md
```diff
@@ -0,0 +1,26 @@
...
+cargo install -f wasm-bindgen-cli --version 0.2.74
+```
+
+### Build
+
+```sh
+deno run --allow-all ./_build.ts
+```
+
+This will regenerate `./wasm_file.ts` and `./wasm_bindings.ts` from the Rust
```
kt3k, line 25: nit: `crypto.js` and `crypto.wasm.js`?

_wasm_crypto/mod.ts
```diff
@@ -0,0 +1,40 @@
...
+  "BLAKE2B-256",
+  "BLAKE2B-384",
+  "BLAKE2B",
+  "BLAKE2S",
+  "BLAKE3",
+  "KECCAK-224",
+  "KECCAK-256",
+  "KECCAK-384",
+  "KECCAK-512",
+  "SHA-384",
```
kt3k, line 22: wikipedia SHA-2 page seems putting SHA-384 in the same insecure category as SHA-{224,256,512} for length extendable attacks ref: https://en.wikipedia.org/wiki/SHA-2 Do you think SHA-384 is secure?

node/crypto.ts
```diff
@@ -58,34 +101,50 @@ export class Hash extends Transform {
    *
    * If encoding is provided a string will be returned; otherwise a Buffer is returned.
    *
-   * Supported encoding is currently 'hex' only. 'binary', 'base64' will be supported in the future versions.
+   * Supported encoding is currently 'hex', 'binary', 'base64'.
```
kt3k, line 104: Thanks!

kt3k (approved):
LGTM

_wasm_crypto/README.md
jeremyBanks: fixed thanks

_wasm_crypto/mod.ts
jeremyBanks: That depends on your definition. Wikipedia's table points out that SHA-384 only provides 128 bits of collision resistance, which is less than the 384 bits that we should want from its digest size. However, in practice this is still only a theoretical concern, as 128 bits is still considered "secure" by most definitions, even if it's not efficient.

I asked a couple of friends who are cryptography professionals, and they both said they considered SHA-384 a safe recommendation, even in 2021, with the benefit of also being extremely widely-supported (such as by native WebCrypto), and the cost of efficiency.

(But if this is a point of disagreement I'm happy to just delete the comments since they're not really important.)

kt3k: Thank you for the input! Ok. Let's keep this as is.

ry (approved):
LGTM - thank you @jeremyBanks, I think many people will find this useful!

PR #1028, opened 2021-07-12
Add range request and etag support to `file_server.ts`

pseudosavant:
First pass at adding support for range requests
- Basic etag support
- Server name header
- Date and Last-Modified headers
- Added ~200 additional mimetypes


kitsonk:
Personal opinion is that this is really getting into the range of what a Web server framework does. file_server is intended to be a minimalist example application. Also it doesn't have tests, but even with exhaustive tests I would feel the same way, that it isn't part of the std library.

pseudosavant:
> Personal opinion is that this is really getting into the range of what a Web server framework does. file_server is intended to be a minimalist example application. Also it doesn't have tests, but even with exhaustive tests I would feel the same way, that it isn't part of the std library.

Agreed on the tests. I'm working on them. I wanted to get some feedback on this first. Is `file_server.ts` not actually part of the std library even though it is in the std repo?

I'm not sure how supporting a basic HTTP 1.1 feature makes it a 'framework'? Without range support, it is basically an HTTP 1.0 server. I know `file_server.ts` is made for basic non-prod work, but even for a basic 'built-in' web server, it is very limited. If it is to remain this limited, it should advertise itself as HTTP 1.0.

I have a video player app that I work on that does thumbnailing for any video URL. Lack of range support caused it to either: 1) only ever show the very first frame no matter the requested timestamp, or 2) it wouldn't work at all, but also wasn't throwing a visible error in the browser. The issue affected playback too. The browser video engine wouldn't seek (even fully downloaded files) or would have to download the entire file because the MP4 index was at the end.

pseudosavant:
I have added and fixed all the tests in `file_server_ts.ts`, and fixed issues in `file_server.ts` discovered from tests.

pseudosavant:
Of course all of the tests passed on my machine, but not on the CI. The file metadata (length, date) is being returned different, probably due to platform differences. I'm changing them to dynamically pull the metadata instead of it being hard coded.

http/file_server_test.ts
```diff
@@ -498,3 +503,296 @@ Deno.test("file_server should show .. if it makes sense", async function (): Pro
...
+  },
+);
+
+Deno.test(
+  "file_server sets `Last-Modified` header correctly",
+  async () => {
+    await startFileServer();
+    try {
+      const res = await fetch("http://localhost:4507/testdata/test%20file.txt");
+      assertEquals(res.headers.get("last-modified"), "Mon, 12 Jul 2021 23:31:46 GMT");
```
kt3k, line 724: git doesn't store mtime. ref https://softwareengineering.stackexchange.com/questions/350403/why-doesnt-git-set-the-file-time

You might need to get mtime from the actual file.

The same problem also seems affecting etag and if-none-match assertions

http/file_server.ts
```diff
@@ -127,20 +235,91 @@ export async function serveFile(
...
+  }
+  
+  // Get and parse the "range" header
+  const range = req.headers.get("range") as string;
+  const rangeRe = /bytes=(\d+)-(\d+)?/;
+  const parsed = rangeRe.exec(range);
+  
+  // Use the parsed value if available, fallback to the start and end of the entire file
+  const start = parsed && parsed[1] ? +parsed[1] : 0;
+  const end   = parsed && parsed[2] ? +parsed[2] : Math.max(0, fileInfo.size - 1);
```
kt3k, line 286: Please run `deno fmt` to pass CI lint check

kt3k:
I'm slightly in favor of including this change because according to the file's [history](https://github.com/denoland/deno_std/commits/b850d01991ee2c6e498182b6a758e8bc8b938307/http/file_server.ts) we've been accepting the PRs which simply add the usability as a file server, (which therefore reduce the minimality as an example). I think this is one of that types of changes. But I'd like to hear other collaborators opinions as well.

http/file_server_test.ts
pseudosavant: I have changed it to pull the values dynamically at runtime. They all pass on my machine. We'll see about the CI.

pseudosavant:
Fixed a couple of hardcoded values I missed in 2 tests.

ry (commented):
This falls in the scope of the file_server.ts and looks well-implemented with tests. Thank you for this PR @pseudosavant !

I'm a little concerned that the file_server now depends on createHash. Can [SubtleCrypto.digest](digest) crypto be used instead?

pseudosavant:
> I'm a little concerned that the file_server now depends on createHash. Can [crypto.subtle.digest](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) be used instead?

Sure, that was a straight-forward change. I just pushed a commit for it.

http/file_server_test.ts
```diff
@@ -498,3 +503,383 @@ Deno.test("file_server should show .. if it makes sense", async function (): Pro
...
+  // see: https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest
+  const hashType = "SHA-1"; // Faster, and this isn't a security senitive cryptographic use case
+  const msgUint8 = new TextEncoder().encode(message);
+  const hashBuffer = await crypto.subtle.digest(hashType, msgUint8);
+  const hashArray = Array.from(new Uint8Array(hashBuffer));
+  const hashHex = hashArray.map((b) => b.toString(16).padStart(2, "0")).join(
+    "",
+  );
+  return hashHex;
+};
```
ry, line 589: This code is duplicated... Seems better if you export it.

Also the `.join("",)` seems odd...

pseudosavant: I can move that out into a module. Is there guidance on where I should put that in the repo?

As for the `join`. Agreed `.join("",)` looks odd. But do you mean the trailing comma or joining the array into strings using `join`? The comma and line break are being added by `deno fmt`. I don't understand why it is doing that though. That method only accepts 1 argument.

pseudosavant: I pushed a commit to make `createEtagHash` clearer, and format better. It doesn't address moving it out into a module yet.

kt3k:
LGTM except CLA check error. It seems 2 email addresses are used in the commits and one of them causing the error... @pseudosavant Can you fix this situation?

pseudosavant:
> LGTM except CLA check error. It seems 2 email addresses are used in the commits and one of them causing the error... @pseudosavant Can you fix this situation?

I've been trying to get it corrected. I didn't realize my git config was slightly different on one of my other machines I used. I followed the CLA assistant's directions and added and verified the other email address in my GitHub account. The CLA assistant doesn't seem to pick it up though. And when I try doing the assistant again it says I've signed it already.

Do you know of another way I could fix this?

http/file_server_test.ts
pseudosavant: @ry Any guidance on where I should put the module? Should I put it in something like `_createEtagHash.ts`?

ry: That would work. But I question whether the test needs the ability to create the etag header generally - how about just hardcoding some expected etag values into the test?

pseudosavant: My first tests did used hard coded values but it didn't work. I learned that the last modified time, and the system reported file size, aren't consistent across environments. Those are the two inputs, concatenated, for `createEtagHash`.

kt3k:
Only the last commit had a different email address. I squashed it to the one before it.

kt3k (approved):
@pseudosavant Thank you for your contribution!

pseudosavant:
> @pseudosavant Thank you for your contribution!

Thanks for the help getting it merged in! Deno is a great project. 🙏

PR #1032, opened 2021-07-13
fix(testing/_diff) Don't merge diff when a token that is surrounded by word-diffs isn't a space

FujiHaruka:
Thanks to #948 by @lowlighter's great contribution, strings diff of assertion is now very readable.

But I found some edge case that strings diff displays like unwelcome.

It's a feature of merging word diffs if a in-between token is a space. Actually word diffs are merged even when the in-between token is not a space.

For example, this is as expected. Fine.

foo bar vs yoo yoo

However, this is not as expected.

2 71 vs 3 14

And this is also not as expected.

It's vs Don't

That's the issue. So this PR fixes it not to merge if the in-between token is not a space.

After fixed, above 2 bad examples will be improved as below.

2 71 vs 3 14 (expected)
It's vs Don't (expected)

### DIFF
--- a/testing/_diff.ts
+++ b/testing/_diff.ts
@@ -287,7 +287,7 @@ export function diffstr(A: string, B: string) {
     ).map((result, i, t) => {
       if (
         (result.type === DiffType.common) && (t[i - 1]) &&
-        (t[i - 1]?.type === t[i + 1]?.type)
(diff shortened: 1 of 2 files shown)


wperron (approved):
LGTM, thanks!

PR #1035, opened 2021-07-14
fix: fix lint errors in deno 1.12

kt3k:
`Deno.customInspect` has been deprecated since Deno v1.12 https://github.com/denoland/deno/pull/10035 and the usage of it now causes lint errors. This PR fixes them.


satyarohith (approved):
LGTM.

PR #1040, opened 2021-07-16
fix(encoding/base64url): throw TypeError in case of invalid base64url string

### LINKED ISSUE #1011: base64url decoder allows decoding base64
The encoding/base64url.ts decoder allows decoding base64 (not base64url) encoded strings.

image

The third case should error with the same error as the first case.

std@0.100.0 & deno 1.11.2 (release, x86_64-unknown-linux-gnu)

TotallyNotChase:
fixes #1011 

Validate input in base64url decoder.

Add tests for making sure invalid base64url strings are rejected.


encoding/base64url.ts
```diff
@@ -17,6 +17,10 @@ export function addPaddingToBase64url(base64url: string): string {
 }
 
 function convertBase64urlToBase64(b64url: string): string {
+  if (!/^[-_A-Z0-9]*?={0,2}$/i.test(b64url)) {
+    // Contains characters not part of base64url spec.
+    throw new TypeError("Invalid base64url character in string!");
```
kt3k, line 22: Maybe we should align the error message with `base64.decode` like:

```ts
throw new TypeError("Failed to decode base64url: invalid character);
```

---
BTW @lucacasonato do you think the error here should be DOMException?

kt3k:
The test case looks good to me

encoding/base64url.ts
TotallyNotChase: Changed, for the time being. Awaiting input on whether or not this should be a `DOMException`.

bartlomieju (approved):
LGTM

PR #1041, opened 2021-07-18
BREAKING(http): cookie headers as params

timreichen:
Change cookie functions `getCookies`, `setCookie` and `deleteCookie` params to `Headers` instance instead of object with (optional) `Headers` instance property.
The former param was so a response can directly be passed. Since the functions only require a `Headers` instance to manipulate, that should be passed directly.

**before**
```ts
setCookie(response,  { name: "foo", value: "bar" })
```
**after**
```ts
setCookie(response.headers,  { name: "foo", value: "bar" })
```
If any of the functions wanted to be used on a `Headers` instance directly, it had to be wrapped:

**before**
```ts
const headers = new Headers()
setCookie({ headers },  { name: "foo", value: "bar" })
```
**after**
```ts
const headers = new Headers()
setCookie(headers,  { name: "foo", value: "bar" })
```
Since the `headers` property in `setCookie` was optional, this allowed to pass an empty object in which a `headers` property would be initiated, manipulated, which seems avoidable.

**before**
```ts
cookie.setCookie({}, { name: "foo", value: "bar" }) // weird
```
**after**
not possible


http/cookie.ts
```diff
@@ -164,8 +164,8 @@ function validateDomain(domain: string): void {
  * Parse the cookies of the Server Request
  * @param req An object which has a `headers` property
```
cmorten, line 165:
```suggestion
 * @param {Headers} headers The headers instance to get cookies from.
```

Need to take care to keep the jsdocs in line with changes 🙂

cmorten: By way of "improve what you touch", consider also adding the `@return {Object} …` jsdoc for these methods as well to add clarity to what is returned for users of the Deno docs.

http/cookie.ts
```diff
@@ -195,16 +196,13 @@ export function getCookies(req: { headers: Headers }): Record<string, string> {
  *   httpOnly: true, secure: true, maxAge: 2, domain: "deno.land" });
```
cmorten, line 196: See also the `@param` definitions for this method.

http/cookie.ts
```diff
@@ -219,11 +217,11 @@ export function setCookie(res: { headers?: Headers }, cookie: Cookie): void {
  *     deleteCookie(res,'foo', {domain:'deno.land'});
```
cmorten, line 217: Sim.

timreichen:
@cmorten
I added a `@return` and updated the `@param`. But I think the jsdocs probably need an overdo for each function in this file. Is there a style guide on how to do jsdocs?
- Dot at the end of a single phrase (both present in std)
- `@return`or `@returns` (both present in std)
- examples with `import` statement or not (both present in std)

Probably best to open a separate PR for this.

cmorten:
> Is there a style guide on how to do jsdocs?

Only what is currently in https://deno.land/manual/contributing/style_guide#use-jsdoc-for-exported-symbols, though not sure it’s been kept to strictly.
 
> * `@return`or `@returns` (both present in std)

I _think_ the singular simply because that is what doc.deno seems to support with syntax highlighting.

> * examples with `import` statement or not (both present in std)

_I believe_ `deno test —doc` requires the example code block to be fully functional/self-contained. This might require use of an import statement.

> Probably best to open a separate PR for this.

We should make sure the jsdocs are at least consistent with the implementation, anything more perhaps could be raised as a separate issue and PR.

timreichen:
@cmorten I updated readme and jsdoc and tests pass now.

cmorten (approved):

kt3k (approved):
I think this now makes sense because we switch to the new http/server module.

@timreichen Thank your for your suggestion and contribution!

Thanks @cmorten for reviewing.

PR #1042, opened 2021-07-18
fix(encoding/toml): fix inline table and nested array

### LINKED ISSUE #823: TOML parser's inline/nested tables misaligned with real TOML

FujiHaruka:
close #823

Sorry for a large change. I almost overall rewrite TOML parser based on recursive descent parsing because current parser includes many edge case bugs. This will make the parser more stable and maintainable.

Of course all tests passed, by which we can believe the parser doesn't get worse. I also added some test cases the previous parser doesn't pass.

This PR fixes case 1, 2, 3, and 4 of https://github.com/denoland/deno_std/issues/823#issuecomment-814367934.

In addition, if the parser fails to parse, it outputs the position the error occurred like this:

> Parse error on line 1, column 34: Single-line string cannot contain EOL


encoding/_toml/parser.ts
```diff
@@ -0,0 +1,897 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { deepAssign } from "https://deno.land/std@0.101.0/_util/deep_assign.ts";
```
kt3k, line 2: Please use relative paths instead of versioned std module url (to prevent multiple versions of std mixed in the future)

encoding/_toml/parser_test.ts
```diff
@@ -0,0 +1,441 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import {
+  assertEquals,
+  assertThrows,
+} from "https://deno.land/std@0.101.0/testing/asserts.ts";
```
kt3k, line 5: This also needs to be relative path dependency.

kt3k:
I think this is well designed rewrite of TOML parser. The unit tests of each parser component look great to me. Those make the maintenance of this module significantly easier in my view. Thank you for the contribution.

encoding/_toml/parser.ts
FujiHaruka: Oh, I didn't notice that. Thank you. I fixed!

encoding/_toml/parser_test.ts
FujiHaruka: And fixed too!

ry (approved):
LGTM - huge patch but seems quite clean and well implemented. Thank you!

PR #1047, opened 2021-07-20
feat(async): add status to deferred promises

lowlighter:
This edits a bit `deferred<T>` in order to be able to track promise status, which is not exposed directly by default.
It's mostly useful to know whether the deferred promise is still `pending` or not (as they're manually resolved)

Personnaly I sometimes use it in unit testing so I figured out it may be useful for others too, though I could understand it may not be that revelant 🙂

### DIFF
--- a/async/deferred.ts
+++ b/async/deferred.ts
@@ -22,7 +22,8 @@ export function deferred<T>(): Deferred<T> {
   let methods;
   const promise = new Promise<T>((resolve, reject): void => {
     methods = {
-      resolve(value: T | PromiseLike<T>) {
+      async resolve(value: T | PromiseLike<T>) {
+        await value;
         Object.assign(promise, { status: "fulfilled" });
         resolve(value);
       },
--- a/async/deferred.ts
+++ b/async/deferred.ts
@@ -5,6 +5,7 @@
 // See https://github.com/Microsoft/TypeScript/issues/15202
 // At the time of writing, the github issue is closed but the problem remains.
 export interface Deferred<T> extends Promise<T> {
+  status: "pending" | "fulfilled" | "rejected";
   resolve(value?: T | PromiseLike<T>): void;
   // deno-lint-ignore no-explicit-any
   reject(reason?: any): void;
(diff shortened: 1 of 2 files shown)


kt3k:
I like the idea. Because `deferred` is mostly used for testing purpose, this addition makes sense to me.

async/deferred.ts
```diff
@@ -20,7 +21,17 @@ export interface Deferred<T> extends Promise<T> {
 export function deferred<T>(): Deferred<T> {
   let methods;
   const promise = new Promise<T>((resolve, reject): void => {
-    methods = { resolve, reject };
+    methods = {
+      resolve(value: T | PromiseLike<T>) {
+        Object.assign(promise, { status: "fulfilled" });
```
kt3k, line 26: The timing of setting "fulfilled" looks too early to me when the value is actually a promise.

I think these lines should be something like:
```ts
await value;
Object.assign(promise, { status: "fulfilled" });
resolve(value);
```

---
note: when I checked the following expression in deno repl:

```ts
let a = new Promise((resolve) => resolve(new Promise(() => {})));
```

then console.log(a) prints `Promise { <pending> }` forever. So a promise is fulfilled when the resolved value is fulfilled.

lowlighter: Nice catch! 
Indeed this case was not handled properly, I applied your suggested changes and added another test to cover it 👍

async/deferred_test.ts
```diff
@@ -4,14 +4,29 @@ import { deferred } from "./deferred.ts";
...
   });
+  assertEquals(d.status, "rejected");
+});
+
+Deno.test("[async] deferred: status with promised value", async function () {
+  const d = deferred<string>();
+  const e = deferred<string>();
+  assertEquals(d.status, "pending");
+  d.resolve(e);
+  assertEquals(d.status, "pending");
```
kt3k, line 28: nice 👍

kt3k (approved):
@lowlighter LGTM. Thank you for the contribution!

async/deferred.ts
```diff
@@ -5,6 +5,7 @@
 // See https://github.com/Microsoft/TypeScript/issues/15202
 // At the time of writing, the github issue is closed but the problem remains.
 export interface Deferred<T> extends Promise<T> {
+  status: "pending" | "fulfilled" | "rejected";
```
kitsonk, line 8: In the ECMAScript specification, the internal field is called `[[PromiseState]]`, so I think `state` would be better name.

kitsonk:
Also, just to mention, why wouldn't we want to use `Object.defineProperty()` and make the value read only?  No big deal though since this is already merged.

PR #1048, opened 2021-07-21
fix(encoding/yaml): fix `parseAll` type definition by using overloads

### LINKED ISSUE #679: [bug] yaml types -- ignores schema
https://github.com/denoland/deno/blob/9de8178c9b9bc21cc880ea81f163118408cc6c46/std/encoding/yaml/parse.ts#L28

Following the types for yaml.parseAll(), the caller expects this to function:
```ts
yaml.parseAll(result, undefined, { schema: yaml.JSON_SCHEMA })
```

However, parseAll types hide what is really going on:
```ts
export function loadAll<T extends CbFunction | LoaderStateOptions>(
  input: string,
  iteratorOrOption?: T,
  options?: LoaderStateOptions
): T extends CbFunction ? void : unknown[] {
  if (!isCbFunction(iteratorOrOption)) {
    return loadDocuments(input, iteratorOrOption as LoaderStateOptions) as Any;
```
undefined is not a callback function and so the third parameter is ignored.

The way to call it is like so, which the types do not allow:
```ts
//@ts-ignore -- this is a bug in YAML library
yaml.parseAll(result, { schema: yaml.JSON_SCHEMA })
```

FujiHaruka:
fix #679

This fixes type definition of `parseAll` in yaml module.

Current:
```ts
// ok
parseAll(yaml)
// ok
parseAll(yaml, callback)
// ok
parseAll(yaml, callback, options)
// typing is ok, but options are ignored
parseAll(yaml, undefined, options)
// type error, but options are accepted
parseAll(yaml, options)
```

Modified:
```ts
// ok
parseAll(yaml)
// ok
parseAll(yaml, callback)
// ok
parseAll(yaml, callback, options)
// type error
parseAll(yaml, undefined, options)
// ok
parseAll(yaml, options)
```

### DIFF
--- a/encoding/_yaml/parse.ts
+++ b/encoding/_yaml/parse.ts
@@ -23,9 +23,17 @@ export function parse(content: string, options?: ParseOptions): unknown {
  * Same as `parse()`, but understands multi-document sources.
  * Applies iterator to each document if specified, or returns array of documents.
  */
+export function parseAll(content: string): unknown;
+export function parseAll(content: string, iterator: CbFunction): void;
 export function parseAll(
   content: string,
-  iterator?: CbFunction,
+  iterator: CbFunction,
+  options: ParseOptions,
+): void;
+export function parseAll(content: string, options: ParseOptions): unknown;
+export function parseAll(
+  content: string,
+  iterator?: CbFunction | ParseOptions,
   options?: ParseOptions,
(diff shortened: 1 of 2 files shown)


kt3k (approved):
@FujiHaruka LGTM. Thanks!

PR #1050, opened 2021-07-21
chore: enable doc type checking

### LINKED ISSUE #1038: CI should type check examples in JSDocs
Deno now supports type checking examples in the JSDocs (via `deno test --doc`).

I'm pretty sure we'll find a lot of failing examples so I think this issue can be considered an "epic" and
we need to fix them over multiple PRs.

kt3k:
This PR enables the type checking of examples in jsdocs.

closes #1038

### DIFF
--- a/collections/chunked.ts
+++ b/collections/chunked.ts
@@ -6,6 +6,8 @@
  * Example:
  *
  * ```typescript
+ * import { chunked } from "https://deno.land/std/collections/chunked.ts";
+ *
  * const words = [ 'lorem', 'ipsum', 'dolor', 'sit', 'amet', 'consetetur', 'sadipscing' ]
  * const chunks = chunked(words, 3)
  *
--- a/collections/chunked.ts
+++ b/collections/chunked.ts
@@ -6,7 +6,7 @@
  * Example:
  *
  * ```typescript
- * import { chunked } from "https://deno.land/std/collections/chunked.ts";
+ * import { chunked } from "./chunked.ts";
  *
  * const words = [ 'lorem', 'ipsum', 'dolor', 'sit', 'amet', 'consetetur', 'sadipscing' ]
  * const chunks = chunked(words, 3)
--- a/collections/chunked.ts
+++ b/collections/chunked.ts
@@ -5,7 +5,7 @@
  *
  * Example:
  *
- * ```typescript
+ * ```ts
  * import { chunked } from "./chunked.ts";
  *
  * const words = [ 'lorem', 'ipsum', 'dolor', 'sit', 'amet', 'consetetur', 'sadipscing' ]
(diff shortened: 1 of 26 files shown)


collections/chunked.ts
```diff
@@ -6,6 +6,8 @@
  * Example:
  *
  * ```typescript
+ * import { chunked } from "https://deno.land/std/collections/chunked.ts";
```
bartlomieju, line 9: Using import like this will cause to download latest `main` branch - it means that if in a PR you change an API the doctests might fail because `main` contains older version. 

I think we should use `import { chunked } from "./chunked.ts";` (though it's not so ergonomic, because if you want to copy-paste the example, you'd have to adjust the import statement.

@caspervonb could you double check and suggest the best solution?

caspervonb: Should definitively be using _relative_ module specifiers, The source itself doesn't live on deno.land after all. If I haven't already documented this in the manual as the recommended way to import "self" I definitively intended to do so!

We can rewrite it to absolute urls tagged to the correct version when we render in deno.land's source view component.

collections/distinct.ts
```diff
@@ -6,6 +6,8 @@
  * Example:
  *
  * ```typescript
```
caspervonb, line 8: Can we do a quick normalisation of the language attributes as-well? `s/typescript/ts/g`

collections/chunked.ts
kt3k: Relative specifiers make sense especially from the maintainer's viewpoint. Updated.

kt3k:
I skipped 'Generate lcov' step in ubuntu runner because I see https://github.com/denoland/deno/issues/10420 happening again with ubuntu CI 😕

bartlomieju (approved):
LGTM, nice work, surprising how few changes were required.

caspervonb:
Looking at the run log, we're downloading a lot of files from deno.land/std

bartlomieju:
> Looking at the run log, we're downloading a lot of files from deno.land/std

This shouldn't happen because Yoshiya switched to relative specifier?

caspervonb:
> > Looking at the run log, we're downloading a lot of files from deno.land/std
> 
> This shouldn't happen because Yoshiya switched to relative specifier?

No it shouldn't, I am confused too as to where this is coming from 😅

kt3k:
Casper was right! I overlooked three imports from `deno.land/std` :facepalm:

caspervonb:
Still see 3 stubborn ones remaining in the logs:

```
Download https://deno.land/std@0.95.0/testing/asserts.ts
Download https://deno.land/std@0.95.0/testing/_diff.ts
Download https://deno.land/std@0.95.0/fmt/colors.ts
```

kt3k:
> Still see 3 stubborn ones remaining in the logs:

Thanks! That was hard coded test dependencies in `async/tee_test.ts` (also my mistake..)

PR #1052, opened 2021-07-21
fix(testing): change `assertThrows` and `assertThrowsAsync` return type to `void` and `Promise<void>`

lowlighter:
This add a side note in documentation about `assertThrowsAsync` to mention that it is actually **required** to `await` them.

Examples are correct, but I feel like this should be highlighted as this is pretty hard to debug. 

Since it's the only assertion that is async it's not really intuitive.
Failing to do so can result in the whole runner to be interrupted... (see below where `test 3` is not even executed) 

```ts
import { assert, assertThrowsAsync } from "https://deno.land/std@0.102.0/testing/asserts.ts";
Deno.test("1", () => { throw new Error() });
Deno.test("2", () => void assertThrowsAsync(async () => {}, Error));
Deno.test("3", () => {});
```

```
running 3 tests from file:///deno_std/test.ts
test 1 ... FAILED (25ms)
test 2 ...
test result: FAILED. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (129ms)

error: Uncaught (in promise) AssertionError: Expected function to throw.
          ^
    at assertThrowsAsync (https://deno.land/std@0.102.0/testing/asserts.ts:664:11)
```

What is confusing is that the assertion does have the expected result and looks fine looking at logs, and throwing inside `Deno.test` is fairly common but in this specific case it makes the test runner fails

I don't have another easy reproduction snippet, but sometimes it seems it could also result in having a mismatch between dispatched and fulfilled promises which will make the test case fails too.

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -33,7 +33,9 @@ pretty-printed diff of failing assertion.
   string.
 - `assertThrowsAsync()` - Expects the passed `fn` to be async and throw (or
   return a `Promise` that rejects). If the `fn` does not throw or reject, this
-  function will throw asynchronously. Also compares any errors thrown to an
+  function will throw asynchronously _(⚠️ assertion should be awaited or be the
+  return value of the test function to avoid any uncaught rejections which could
+  result in unexpected process exit)_. Also compares any errors thrown to an
   optional expected `Error` class and checks that the error `.message` includes
   an optional string.
 - `unimplemented()` - Use this to stub out methods that will throw when invoked.
@@ -107,10 +109,6 @@ Deno.test("fails", function (): void {
 ```
 
 Using `assertThrowsAsync()`:
-
-> ⚠️ Note that it is _required_ to `await` for `assertThrowsAsync` to ensure
-> that all promises get resolved within the test `Deno.test` callback. Failing
-> to do so may result in early exit or triggering the Op sanitizer
 
 ```ts
 Deno.test("doesThrow", async function () {
--- a/testing/README.md
+++ b/testing/README.md
@@ -108,6 +108,10 @@ Deno.test("fails", function (): void {
 
 Using `assertThrowsAsync()`:
 
+> ⚠️ Note that it is _required_ to `await` for `assertThrowsAsync` to ensure
+> that all promises get resolved within the test `Deno.test` callback. Failing
+> to do so may result in early exit or triggering the Op sanitizer
+
 ```ts
 Deno.test("doesThrow", async function () {
   await assertThrowsAsync(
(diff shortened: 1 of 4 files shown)


testing/README.md
```diff
@@ -108,6 +108,10 @@ Deno.test("fails", function (): void {
 
 Using `assertThrowsAsync()`:
 
+> ⚠️ Note that it is _required_ to `await` for `assertThrowsAsync` to ensure
```
kitsonk, line 111: This technically isn't true. The promise returned from `assertThrowsAsync` can just be returned as well.

It would be more accurate to state that `assertThrowsAsync` returns a promise that will be resolved if the assertion passes, or rejected if it fails. This promise should be be awaited or be the return value from the test function.

Also, it would be far better to put it in the inline documentation for the API.

lowlighter: > This technically isn't true. The promise returned from `assertThrowsAsync` can just be returned as well.
> This promise should be be awaited or be the return value from the test function.

This is actually not possible unless you `void` it. Trying to return the promise directly will make TypeScript complain because the test function expects `void | Promise<void>`:
```
error: TS2322 [ERROR]: Type 'Promise<Error>' is not assignable to type 'void | Promise<void>'.
  Type 'Promise<Error>' is not assignable to type 'Promise<void>'.
    Type 'Error' is not assignable to type 'void'.
Deno.test("2", () => assertThrowsAsync(async () => {}, Error));
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at file:///deno_std/test.ts:6:22

    The expected type comes from the return type of this signature.
      export function test(name: string, fn: () => void | Promise<void>): void;
                                             ~~~~~~~~~~~~~~~~~~~~~~~~~~
        at asset:///lib.deno.ns.d.ts:188:42
```

> It would be more accurate to state that assertThrowsAsync returns a promise that will be resolved if the assertion passes, or rejected if it fails. This promise should be be awaited or be the return value from the test function.

I'll change the wording to reflect the first part of your suggestion and update the inline documentation at the top.
The last part about *being the return value from the test function* is not valid though (at least currently)

kitsonk: `assertThrowsAsync` shouldn't resolve with an error. I didn't realise that.  Assertions shouldn't have return values.

lowlighter: Seems that both `assertThrows` and  `assertThrowsAsync` actually had `Promise<Error>` instead of `Promise<void>` as return type so I patched them.

> Assertions shouldn't have return values.

I agree with that statement, but it does feel awkward to retrieve error class from within the test function.
At least it makes them consistent with other assertions return type and make it possible to avoid brackets use from inline arrow functions

kt3k (approved):
LGTM

kt3k:
@lowlighter Thank you for the contribution! nice work!

PR #1055, opened 2021-07-22
fix: rename deferred.status -> state

kt3k:
As pointed in https://github.com/denoland/deno_std/pull/1047#discussion_r674598436 by @kitsonk, ecma262 spec calls this property of promise [[PromiseState]] (ref: https://tc39.es/ecma262/#sec-fulfillpromise). This PR follows that convention to reduce confusions in the future, and renames the property to '.state'

This PR also makes the property 'readonly' (both in typings and runtime)

cc @lowlighter


caspervonb:
The V8 API also refers to this as the promise state so SGTM.

kitsonk (approved):
LGTM

PR #1057, opened 2021-07-22
feat(fs): add lookPath function

### LINKED ISSUE #1056: Add `lookPath` function to find the path of an executable
Golang's exec package has this neat [`LookPath` function](https://pkg.go.dev/os/exec#LookPath) that allows finding the absolute path to an executable by it's shortname that I think would be useful in the Deno standard lib. Here's an example usage in Go:

```go
package main

import (
	"fmt"
	"log"
	"os/exec"
)

func main() {
	path, err := exec.LookPath("fortune")
	if err != nil {
		log.Fatal("installing fortune is in your future")
	}
	fmt.Printf("fortune is available at %s\n", path)
}
```

Since we don't have an `exec` package, I propose we put this function in the `fs` package, though `path` would also be a good choice.

wperron:
Port of the [lookPath][1] function from Go's exec package. Fixes #1056.

[1]https://pkg.go.dev/os/exec#example-LookPath

### DIFF
--- a/fs/_lp_windows.ts
+++ b/fs/_lp_windows.ts
@@ -1,0 +1,71 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { splitList } from "../path/separator.ts";
+import { join } from "../path/win32.ts";
+
+async function chkStat(file: string) {
+  const s = await Deno.stat(file);
+  if (s.isDirectory) {
+    throw new Deno.errors.PermissionDenied(file);
+  }
+  return;
+}
+
+function hasExt(file: string): boolean {
+  const i = file.lastIndexOf(".");
+  if (i < 0) {
+    return false;
+  }
+  return file.lastIndexOf(":") < i &&
+    file.lastIndexOf("\\") < i &&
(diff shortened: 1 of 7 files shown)


wperron:
Still need to add some tests, especially for Windows

nhrones:
@wperron Let me know if I can help validate anything on Windows.

fs/_lp_windows.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { splitList } from "../path/separator.ts";
+import { join } from "../path/win32.ts";
+
+async function chkStat(file: string) {
+  const s = await Deno.stat(file);
+  if (s.isDirectory) {
+    throw new Deno.errors.PermissionDenied(file);
```
kt3k, line 8: Maybe better to give more descriptive error message? somthing like `${file} is a directory`

bartlomieju:
This is stale and not close to landing. I'm going to close it for now, if anyone feels like picking it up, feel free to open a new PR.

Issue #1058, opened 2021-07-23
'file_server sets `Date` header correctly' test case is flaky


kt3k:

getspooky:
@kt3k I confirm this issue , I run mutilple times `file_server_test.ts` ,the 3 first tests has been successfully passed but the last has failed.
```sh
failures:

file_server sets `Date` header correctly
AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   903909
+   903910

```

kt3k:
fixed by #1059

uki00a:
This test still seems flaky :scream: (https://github.com/denoland/deno_std/pull/1094/checks?check_run_id=3242259868)

PR #1059, opened 2021-07-23
fix(http): 'file_server sets Date header correctly' test case is flaky

getspooky:
Issue: https://github.com/denoland/deno_std/issues/1058


kt3k (approved):
@getspooky Thank you for looking into this! LGTM

PR #1062, opened 2021-07-26
feat(collections): Returns the index of the last element in the given array matching the given predicate

### LINKED ISSUE #1060: [Feature Request] Add `findLastIndex` function to std/collections
`Array` has finding functions, `find()` and `findIndex()`.
`find()` returns the element, and `findIndex()` returns the index.

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findindex

Now, we have `findLast()` in std/collections, this returns the element.

- https://deno.land/std/collections/find_last.ts

So, how about add `findLastIndex()` ?
It is almost same as `findLast()`, but returns the index of the last found element.

```ts
import { Predicate } from "./types.ts";

export function findLastIndex<T>(
  array: Array<T>,
  predicate: Predicate<T>,
): number {
  for (let i = array.length - 1; i >= 0; i -= 1) {
    const element = array[i];

    if (predicate(element)) {
      return i;
    }
  }

  return -1;
}
```

getspooky:
Please see https://github.com/denoland/deno_std/issues/1060

closes #1060

### DIFF
--- a/collections/find_last_index.ts
+++ b/collections/find_last_index.ts
@@ -1,0 +1,32 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { Predicate } from "./types.ts";
+
+/**
+ * Returns the index of the last element in the given array matching the given predicate
+ *
+ * Example:
+ *
+ * ```ts
+ * import { findLastIndex } from "./find_last_index.ts";
+ *
+ * const numbers = [ 4, 2, 7 ]
+ * const lastIndexNumber = findLastIndex(numbers, it => it % 2 === 0)
+ *
+ * console.assert(lastIndexNumber === 1)
+ * ```
+ */
+export function findLastIndex<T>(
+  array: Array<T>,
+  predicate: Predicate<T>,
+): number {
+  for (let i = array.length - 1; i >= 0; i -= 1) {
+    const element = array[i];
+
+    if (predicate(element)) {
+      return i;
+    }
+  }
+
+  return -1;
+}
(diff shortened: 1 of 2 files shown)


kt3k:
I think we need a test case of multiple matches like `findLastIndexTest([[0, 1, 2, 3, 4, 5, 6], (it) => it % 2 === 0], 6)` to confirm the function actually finds the last index.

getspooky:
@kt3k sure, I added multiple matches.

kt3k (approved):
@getspooky Thank you for updating! LGTM

LionC:
I **personally** would use `undefined` for no index - I am aware that the native `.indexOf` uses -1, but all other functions within `collections` use `undefined` to communicate that nothing was found, which also has the huge benefit of forcing the consumer of the function to actually handle that case (if they are using typescript), while the current solution looks like it is fine (`-1` is a number) but might blow up during runtime.

getspooky:
Thanks for feedback 👍🏼 , 
As you might know `Array.prototype.findIndex()` returns `-1`  indicating that no element passed the test. So i just followed the same convention provided by API.

LionC:
I see the inspiration, however, `collections` is its own module and tries to be as Typescript "native" as possible, while the ES spec functions have a very historical API that cannot be changed anymore.

The same applies to `filter` / `map` functions, which have just one parameter within `collections` compared to their ES counterparts which have several optional parameters like index.

Issue #1065, opened 2021-07-27
[Umbrella] std/collections


kt3k:
We landed the initial version of `std/collections` at #993

This issue is the umbrella issue for the remaining 28 APIs, which still need discussion. More detailed type definitions / docs / examples are available in [the original proposal PR](https://github.com/denoland/deno_std/pull/978) by @LionC.

#### ARRAYS

Transformations to Array

- [ ] mapNotNullish
- [ ] takeFirstWhile
- [ ] takeLastWhile
- [ ] dropFirstWhile
- [ ] dropLastWhile
- [ ] filterNotNullish
- [ ] sortBy #1061
- [ ] runningReduce
- [ ] without
- [ ] withoutAll

Transformations to value

- [ ] findLastIndex #1060
- [ ] single
- [ ] firstNotNullishOf
- [ ] sumOf
- [ ] maxBy
- [ ] minBy
- [ ] maxWith
- [ ] minWith
- [ ] maxOf
- [ ] minOf

Number array transformations

- [ ] average

Transformations to other formats

- [ ] associateBy
- [ ] windowed

#### RECORDS

Predicates

- [ ] includesValue

Transformations to Records

- [ ] filterValuesNotNullish
- [ ] deepMerge

#### Groups

- [ ] reduceGroups
- [ ] aggregateGroups

If you're interested in some specific items in the above, please create a separate issue/PR and keep discussion there.

getspooky:
sounds good to me 👍🏼

lowlighter:
I'm tempted to implement `deepMerge` but I saw some concerns about it in the mentioned discussion. 
Is it still revelant ?

grian32:
I would like to attempt implementing `sumOf` is there any issues with that?

LionC:
> I would like to attempt implementing `sumOf` is there any issues with that particular function or am i good to go?

I would say just open the PR and if there are comments they will be posted there. Look at the ither functions for implementation and test reference, a lot can be copied as a template :-)

majidsajadi:
do we really need average function?  we can achieve this with a one-liner. 
```javascript
array.reduce((a, b) => a + b) / array.length
```

LionC:
> do we really need average function?  we can achieve this with a one-liner. 
> 
> ```javascript
> 
> array.reduce((a, b) => a + b) / array.length
> 
> ```

There is a PR for a separate math mofule that contains average together with other functions.

[...]

PR #1069, opened 2021-07-27
Implement collections/sortBy ( closes #1061 )

### LINKED ISSUE #1061: Feature request: `sortBy` function in `std/collections`

`Array.prototype.sort` sorts an array using a callback that takes two elements of the array and must return a number, whose sign will determine their sort order. This is not very intuitive, especially when you just want to compare the elements by the natural ordering of one of their fields, for example.

A related issue is that, if you don't pass a callback to `Array.prototype.sort`, the default behavior is to stringify the array elements, which is very counterintuitive in cases like `[10, 20, 110].sort()`.

Add a `sortBy` utility function in `std/collections`, whose callback parameter maps an array element to a value which will then be compared with the mapped values for other array elements, much like the `key` callback parameter to Python's [`list.sort`](https://docs.python.org/3/library/stdtypes.html#list.sort). The mapped values would be compared using their natural ordering (a.k.a. by using the regular comparison operators).

And since (as per above), `Array.prototype.sort` without a callback also has problems, the callback attribute to `sortBy` should also be optional and default to the identity function.

An alternative would be to have a function that would take a mapping callback and return a callback that can be used with `Array.prototype.sort`. This would incur additional complexity that probably isn't needed.

**Additional context**

N/A

LionC:
This PR adds a `sortBy` function to the `collections` module.

`sortBy` allows you to sort an array without mutating it by an arbitrary selected value like this:

```ts
const people = [
    { name: 'Anna', age: 34 },
    { name: 'Kim', age: 42 },
    { name: 'John', age: 23 },
]

const sortedByAge = sortBy(people, it => it.age)
```

`sortBy` was part of the original proposal (see #1065 and its links) and in the discussions following the merge of the initial set of functions, seemed to have consensus, which is why I went forward and implemented it.

Notes on the implementation: I chose to limit the selector function to common sortable types. I looked into different options and they all seemed hacky or inconvenient for users of the function. The approach in the PR seemed like the best compromise between not caring about return types at all (and thus not saving people from hard to track bugs) and specifying every possible case.

This closes #1061


collections/sort_by.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Example:
+ *
+ * ```typescript
+ * const people = [
+ *     { name: 'Anna', age: 34 },
+ *     { name: 'Kim', age: 42 },
+ *     { name: 'John', age: 23 },
+ * ]
+ * const sortedByAge = sortBy(people, it => it.age)
```
kt3k, line 14: We set up doc example type checking in #1050 . So this jsdoc example causes an error in CI saying "sortBy is not found". Please add `import { sortBy } from "./sort_by.ts";` at the beginning of the example block.

collections/sort_by.ts
```diff
@@ -0,0 +1,48 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Returns all elements in the given collection, sorted by their result using the given selector
+ *
+ * Example:
+ *
+ * ```typescript
```
kt3k, line 8: We prefer `ts` tag for typescript example block. ref: https://github.com/denoland/deno_std/pull/1050#discussion_r673884120

collections/sort_by.ts
```diff
@@ -0,0 +1,48 @@
...
+ * ```
+ */
+export function sortBy<T>(
+  array: Array<T>,
+  selector:
+      | ((el: T) => number)
+      | ((el: T) => string)
+      | ((el: T) => bigint)
+      | ((el: T) => boolean)
+      | ((el: T) => Date),
```
kt3k, line 30: I like this idea of restricting the return value :+1: This should reduce the confusion in the user's code base.

But I'm not sure if we should include boolean as returned value. Do you think there is a use case for that? If the purpose of sorting by boolean is to separate items in 2 groups, how about encouraging the use of `partition`?

kt3k:
The test cases look good to me.

collections/sort_by.ts
LionC: I have no strong opinion on this - I just included the builtins / primitives that get sorted "correctly" by the spec. I agree that `boolean` might be less useful. If no one objects, I will remove it.

collections/sort_by.ts
LionC: Oh the examples are not evaluated in the scope of the current file? I was not aware. That makes the examples kind of verbose doesn't it?

collections/sort_by.ts
LionC: Fixed

collections/sort_by.ts
LionC: I added the import - but it still feels like a lot of clutter for an example that is meant to be displayed in an LSP hover :-S

collections/sort_by.ts
LionC: I removed it

kt3k (approved):
@LionC LGTM. Nice work!

PR #1071, opened 2021-07-27
Implement collections/associateBy

LionC:
This adds an `associateBy` function to the `collections` module.

`associateBy` transforms an array into a `Record`, often used to conveniently access it by key instead of iterating via e.g. `.find()` afterwards:

```ts
const users = [
    { id: 'a2e', userName: 'Anna' },
    { id: '5f8', userName: 'Arnold' },
    { id: 'd2c', userName: 'Kim' },
]

const usersById = associateBy(users, it => it.id)
const someId = 'a23'
const user = usersById[someId]
```

`associateBy` was part of the original proposal (see #1065 and its links). In the discussions around it and following the merge of the initial set of functions, it seemed that there are no strong opinions against it, so I posted this PR.


kt3k (approved):
@LionC Nice work! LGTM

PR #1072, opened 2021-07-27
change (collections): Refactor find_last_index to return undefined on no index found

LionC:
**As `collections` has not been released yet, we are still free to do braking changes.**

This changes the return (type) of `findLastIndex` when no matching element was found from `-1` to `undefined`.

While ES `.indexOf` returns `-1`, all function in the `collections` proposal return `undefined` to communicate "no result". This has the huge benefit that, if typescript is used, the user of the function actually needs to handle that case conciously instead of `-1` blowing up during runtime (but compiling fine because it is a `number` but not a valid index).

Examples of runtime problems uncaught by TS with `-1`:

```ts
const arr = [ 1, 3 ]
const evenNumberIndex = findIndex(arr, it => it %2 == = 0)

const evenNumber = arr[evenNumberIndex] // will blow up later because `evenNumber` is `undefined`
const evenNumberWorse = arr.at(evenNumberIndex) // this is even worse, as it will return 3 - a number that does not match the initial predicate. That bug will cause some headache...
```

TS will not allow the accesses below if the type of `evenNumberIndex` is `number | undefined`, forcing the developer to handle that case before shipping the code by adding a default with `?? 0`, throwing or doing whatever makes sense to recover.


kt3k:
I personally feel this is too opinionated design because (as @getspooky says) array.findIndex returns -1, and also [_.findLastIndex](https://lodash.com/docs/4.17.15#findLastIndex) and es-proposal [findLastIndex](https://github.com/tc39/proposal-array-find-from-last) both return -1 for no matching case.

But I'm open to including this if there's enough supporters of this design.

dsherret:
I agree with @kt3k for the reasons he stated (consistency). What would be nice is for the documentation to be updated to describe that it returns `-1` in this case.

grian32:
I think this is a good choice because the new `.at()` method introduced in V8 9.2 returns the last element of the array when given `-1`, and as such might cause confusion, its safest to return `undefined`.

kt3k (approved):
Ok. Let's try this design. LGTM

PR #1074, opened 2021-07-28
feat(collections): Returns an array excluding single given value

getspooky:
Returns an array excluding all given values

### Example: 

```ts
without([2, 1, 2, 3, 2], 1,2);
// 3
```

### Discussion

Please see https://github.com/denoland/deno_std/issues/1065

### DIFF
--- a/collections/without.ts
+++ b/collections/without.ts
@@ -1,0 +1,31 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/** Used as the size to enable large array optimizations. */
+const LARGE_ARRAY_SIZE = 200;
+
+/**
+ * Returns an array excluding all given values using SameValueZero for equality comparisons.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { without } from "./without.ts";
+ *
+ * const withoutList = without([2, 1, 2, 3], 1, 2);
+ *
+ * console.assert(withoutList === [ 1, 2 ])
+ * ```
+ */
+export function without<T>(
+  arrays: Array<T>,
+  ...values: Array<T>
+): T[] {
+  if (values.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+    for (const element of arrays) {
+      set.add(element);
+    }
+    arrays = Array.from(set);
+  }
+  return arrays.filter((value) => values.indexOf(value) === -1);
+}
--- a/collections/without.ts
+++ b/collections/without.ts
@@ -17,17 +17,18 @@ const LARGE_ARRAY_SIZE = 200;
  * ```
  */
 export function without<T>(
-  arrays: Array<T>,
+  array: Array<T>,
   ...values: Array<T>
 ): Array<T> {
-  if (values.length >= LARGE_ARRAY_SIZE) {
+
+  if (array.length >= LARGE_ARRAY_SIZE) {
     const set = new Set<T>();
 
-    for (const element of arrays) {
+    for (const element of array) {
       set.add(element);
     }
 
-    arrays = Array.from(set);
+    array = Array.from(set);
   }
-  return arrays.filter((value) => values.indexOf(value) === -1);
+  return array.filter((value) => values.indexOf(value) === -1);
 }
--- a/collections/without.ts
+++ b/collections/without.ts
@@ -1,33 +1,21 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-/** Used as the size to enable large array optimizations. */
-const LARGE_ARRAY_SIZE = 200;
-
 /**
- * Returns an array excluding all given values.
+ * Returns an array excluding single given value.
  *
  * Example:
  *
  * ```ts
  * import { without } from "./without.ts";
  *
- * const withoutList = without([2, 1, 2, 3], 1, 2);
+ * const withoutList = without([1,2,3], 1);
  *
- * console.assert(withoutList === [ 3 ])
+ * console.assert(withoutList.length === 2)
  * ```
  */
 export function without<T>(
   array: Array<T>,
-  ...values: Array<T>
+  value: T,
 ): Array<T> {
-  if (array.length >= LARGE_ARRAY_SIZE) {
-    const set = new Set<T>();
-
-    for (const element of array) {
-      set.add(element);
-    }
-
-    array = Array.from(set);
-  }
-  return array.filter((value) => values.indexOf(value) === -1);
+  return array.filter((element) => value !== element);
 }
--- a/collections/without.ts
+++ b/collections/without.ts
@@ -20,7 +20,6 @@ export function without<T>(
   array: Array<T>,
   ...values: Array<T>
 ): Array<T> {
-
   if (array.length >= LARGE_ARRAY_SIZE) {
     const set = new Set<T>();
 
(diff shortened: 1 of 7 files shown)


kt3k:
Please add test cases when `excluded` has more than LARGE_ARRAY_SIZE items.

kt3k:
BTW this covers both `without` `withoutAll` cases because this `without` accepts any number or arguments. @LionC Do you agree with this design?

collections/without.ts
```diff
@@ -0,0 +1,31 @@
...
+export function without<T>(
+  arrays: Array<T>,
+  ...values: Array<T>
+): T[] {
+  if (values.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+    for (const element of arrays) {
+      set.add(element);
+    }
+    arrays = Array.from(set);
```
kt3k, line 28: This makes the array items distinct. Is this really what you wanted to do?

LionC:
> BTW this covers both `without` `withoutAll` cases because this `without` accepts any number or arguments. @LionC Do you agree with this design?

I would personally prefer to distingush `without` and `withoutAll`. The current design forces spreading for all `withoutAll` cases, which does not feel too nice to me and may have performance implications as well.

@getspooky would it be ok for you to split this into two functions?

collections/without.ts
LionC: This looks wrong to me. I would also suggest renaming the argument to `array`, as it is actually just one array and maybe create a new variable, as shadowing arguments can makes things harder to follow.

LionC:
It might make sense to build a `Set` for the values to exclude, as that reduces the runtime from `N*V` to `N + V` at the cost of  a bit of memory for the `Set`.

Something like:

```ts
const valuesToExclude = new Set(values)

return array.filter(it => !valuesToExclude.has(it))
```

collections/without.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * const withoutList = without([2, 1, 2, 3], 1, 2);
+ *
+ * console.assert(withoutList === [ 3 ])
+ * ```
+ */
+export function without<T>(
+  arrays: Array<T>,
+  ...values: Array<T>
+): T[] {
```
LionC, line 22: Nitpick: You are using `T[]` and `Array<T>` here - I would suggest being consistent about it

```suggestion
): Array<T> {
```

collections/without.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (values.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+    for (const element of arrays) {
+      set.add(element);
+    }
+    arrays = Array.from(set);
+  }
+  return arrays.filter((value) => values.indexOf(value) === -1);
```
LionC, line 30: Nitpick: Some whitespace makes control flow more scannable:

```suggestion
    const set = new Set<T>();
    
    for (const element of arrays) {
      set.add(element);
    }
    
    arrays = Array.from(set);
  }
  
  return arrays.filter((value) => values.indexOf(value) === -1);
```

collections/without.ts
getspooky: fixed

getspooky:
Thanks @kt3k and @LionC for feedback,
I don't see the purpose of making tow functions `without` and `withoutAll` that do the same task.
it would be more easy and efficient to make one function `without` that take n of args.

LionC:
> Thanks @kt3k and @LionC for feedback,
> 
> I don't see the purpose of making tow functions `without` and `withoutAll` that do the same task.
> 
> it would be more easy and efficient to make one function `without` that take n of args.
> 
> 

While they do the same task on an abstract level, a simple `without` can be optimized differently, while a `withoutAll` can take an `Array` instead of varargs, which saves the user the need to `...spread` every time they use the functions, saving some performance as well (`...spread`ing is not free).

As all the functions in this module are pretty low level and attract usage in other algorithms, we should emphasize even small-ish performance gains. We have done that for all other functions in this module to this point.

collections/without.ts
```diff
@@ -0,0 +1,33 @@
...
+  ...values: Array<T>
+): Array<T> {
+  if (array.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+
+    for (const element of array) {
+      set.add(element);
+    }
+
+    array = Array.from(set);
```
kt3k, line 30: I think this removes the duplicated elements in the original array, and that is not correct...

For example, `without(Array(200).fill(1), 0)` returns `[1]`. I think this should return `Array(200).fill(1)`.

getspooky:
@kt3k I just fix `without` to accept one value, i will open `withoutAll` in separated PR.

collections/without.ts
```diff
@@ -0,0 +1,21 @@
...
+ * import { without } from "./without.ts";
+ *
+ * const withoutList = without([1,2,3], 1);
+ *
+ * console.assert(withoutList.length === 2)
+ * ```
+ */
+export function without<T>(
+  array: Array<T>,
+  ...values: Array<T>
```
kt3k, line 18: still accepting multiple values?

getspooky: i fixed

kt3k:
@getspooky Thanks for updating! The implementation looks good to me now.

@LionC 
After seeing this one liner implementation:
```ts
return array.filter((element) => value !== element);
```
I'm now skeptical about the benefit of providing this function as part of `std/collections`. The above is straightforward way to do this and I don't know why the users want to use this `without` function over simple `array.filter((e) => e !== v)`.

timreichen:
> I'm now skeptical about the benefit of providing this function as part of `std/collections`. The above is straightforward way to do this and I don't know why the users want to use this `without` function over simple `array.filter((e) => e !== v)`.

I would recommend against implementing this function. An abstraction of a one-liner is not necessary imo. Using `array.filter` directly is more flexible.

LionC:
> > I'm now skeptical about the benefit of providing this function as part of `std/collections`. The above is straightforward way to do this and I don't know why the users want to use this `without` function over simple `array.filter((e) => e !== v)`.
> 
> 
> 
> I would recommend against implementing this function. An abstraction of a one-liner is not necessary imo. Using `array.filter` directly is more flexible.

`reduce` is more flexible than `filter`. `for` is more flexible than `reduce`.

As stated several times, I do not see the harm in having such a function, while there is gain in it for some people, even if that does not include everyone. Other libraries have choosen to include it even though they have `filter` as well and people us it - because it states their intent clearly, is easily scannable and short.

I feel like discussions like this boil down to coding style (how much abstraction / renaming do you like) a lot. We should keep in mind that `deno` does not try to make decisions like that for its users.

LionC:
@kt3k I think we should try to get a call on this, otherwise this PR will stand still forever.

I personally do not see the harm in having a clearly named function doing a common useful thing with a concise API.

Also, [this](https://stackoverflow.com/questions/5767325)(general) or [this](https://stackoverflow.com/questions/6824916)(specific) SO question and the (sometimes horrible) accepted/upvoted answers make me think that there is value in having a clearly named function with an optimal implementation...

kt3k:
@LionC 
From my experience of collaborating with many people with various levels, even very beginners can write this filtering by themselves. So I highly doubt this is useful for somebody.

And many collaborators (Luca, Nayeem, Divy,...) are strongly against adding 'too trivial' items into std/collections (and we already retracted many items in earlier iteration because of the triviality of them). I'm afraid this one typically falls into that category.

kt3k:
Talked a little bit with @LionC, but we're closing this particular one (mainly because of the triviality).

@getspooky Thank you for working on this anyways.

PR #1075, opened 2021-07-28
feat(collections): add `deepMerge`

lowlighter:
This implements `deepMerge`, which merges deeply two records together into the first one.

```ts
deepMerge({           // {
  foo: {              //   foo: {
    bar: true,        //     bar: true,
  },                  //     baz: true,
}, {                  //     quux: {},
  foo: {              //   },
    baz: true,        //   qux: true,
    quux: {},         // }
  },
  qux: true,
})
```

For convenience, a few options are proposed to handle edge cases:
- `arrays`, `sets` and `maps` can change their merging strategy from `"replace"` to `"concat"`. In the latter mode, values from both records will be kept
- `includeNonEnumerable` can be used to include properties and symbols which have an hidden visibility

Additional implementation notes:
- All `PropertyKey` types are supported (including `symbol`)
- Built-ins (like `RegExp`, `Date`, ...) and user classes won't be treated as objects, so the recursive merging won't apply on them (instead their reference is just replaced in target record) 

Ref:  #1065

____

- [x] Typings
  - [x] Basic typing
  - [x] Improved maps, arrays, sets typing
  - [x] Probably missing a recursive typing somewhere
  - [x] Handle merging strategy to reflect correct typing
- [x] Add signature for `Partial<T> : T` 
- [x] Change default to `merge` instead
- [x] Add example to comment and `README.md`  
- [x] Rename typing in PascalCase instead and export the options one
- [x] First clone the target object to avoid mutating in place
- [x] Improve prototype check

### DIFF
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -1,9 +1,11 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 /**
- * Merges the two given Records, recursively merging any nested Records with the second collection overriding the first in case of conflict
+ * Merges the two given Records, recursively merging any nested Records with
+ * the second collection overriding the first in case of conflict
  *
- * For arrays, maps and sets, a merging strategy can be specified to either "replace" values, or "concat" them instead.
+ * For arrays, maps and sets, a merging strategy can be specified to either
+ * "replace" values, or "concat" them instead.
  * Use "includeNonEnumerable" option to include non enumerable properties too.
  */
 export function deepMerge<
@@ -19,17 +21,17 @@ export function deepMerge<
     includeNonEnumerable = false,
   }: deepMergeOptions = {},
 ): T & U {
-  //Extract property and symbols
+  // Extract property and symbols
   const keys = [
     ...Object.getOwnPropertyNames(other),
     ...Object.getOwnPropertySymbols(other),
   ].filter((key) => includeNonEnumerable || other.propertyIsEnumerable(key));
 
-  //Iterate through each key of other object and use correct merging strategy
+  // Iterate through each key of other object and use correct merging strategy
   for (const key of keys as propertyKey[]) {
     const a = collection[key], b = other[key];
 
-    //Handle arrays
+    // Handle arrays
     if ((Array.isArray(a)) && (Array.isArray(b))) {
       if (arrays === "concat") {
         collection[key] = a.concat(...b);
@@ -39,7 +41,7 @@ export function deepMerge<
       continue;
     }
 
-    //Handle maps
+    // Handle maps
     if ((a instanceof Map) && (b instanceof Map)) {
       if (maps === "concat") {
         for (const [k, v] of b.entries()) {
@@ -51,7 +53,7 @@ export function deepMerge<
       continue;
     }
 
-    //Handle sets
+    // Handle sets
     if ((a instanceof Set) && (b instanceof Set)) {
       if (sets === "concat") {
         for (const v of b.values()) {
@@ -63,13 +65,13 @@ export function deepMerge<
       continue;
     }
 
-    //Recursively merge mergeable objects
+    // Recursively merge mergeable objects
     if (isMergeable<T | U>(a) && isMergeable<T | U>(b)) {
       collection[key] = deepMerge(a ?? {}, b);
       continue;
     }
 
-    //Override value
+    // Override value
     collection[key] = b;
   }
 
@@ -81,18 +83,18 @@ export function deepMerge<
  * Builtins object like, null and user classes are not considered mergeable
  */
 function isMergeable<T>(value: unknown): value is T {
-  //Ignore null
+  // Ignore null
   if (value === null) {
     return false;
   }
-  //Ignore builtins
+  // Ignore builtins
   if (
     (value instanceof RegExp) || (value instanceof Date) ||
     (value instanceof Array)
   ) {
     return false;
   }
-  //Ignore classes
+  // Ignore classes
   if ((typeof value === "object") && ("constructor" in value!)) {
     return !/^class /.test(value.constructor.toString());
   }
@@ -111,6 +113,7 @@ interface deepMergeOptions {
   includeNonEnumerable?: boolean;
 }
 
-//TypeScript does not support 'symbol' as index type currently though it's perfectly valid
-//deno-lint-ignore no-explicit-any
+// TypeScript does not support 'symbol' as index type currently though
+// it's perfectly valid
+// deno-lint-ignore no-explicit-any
 type propertyKey = any;
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -23,7 +23,7 @@
 export function deepMerge<
   T extends Record<PropertyKey, unknown>,
 >(
-  object: Partial<T>,
+  record: Partial<T>,
   other: Partial<T>,
   options?: DeepMergeOptions,
 ): T;
@@ -32,7 +32,7 @@ export function deepMerge<
   T extends Record<PropertyKey, unknown>,
   U extends Record<PropertyKey, unknown>,
 >(
-  object: T,
+  record: T,
   other: U,
   options?: DeepMergeOptions,
 ): T;
@@ -41,7 +41,7 @@ export function deepMerge<
   T extends Record<PropertyKey, unknown>,
   U extends Record<PropertyKey, unknown>,
 >(
-  object: T,
+  record: T,
   other: U,
   {
     arrays = "merge",
@@ -50,7 +50,7 @@ export function deepMerge<
     includeNonEnumerable = false,
   }: DeepMergeOptions = {},
 ): T & U {
-  const result = object;
+  const result = clone(record, { includeNonEnumerable });
 
   // Extract property and symbols
   const keys = [
@@ -65,7 +65,7 @@ export function deepMerge<
     // Handle arrays
     if ((Array.isArray(a)) && (Array.isArray(b))) {
       if (arrays === "merge") {
-        result[key] = a.concat(...b);
+        (result[key] as typeof a).push(...b);
       } else {
         result[key] = b;
       }
@@ -107,6 +107,47 @@ export function deepMerge<
   }
 
   return result as T & U;
+}
+
+/**
+ * Clone a record
+ * Arrays, maps, sets and objects are cloned so references doesn't point
+ * anymore to those of cloned record
+ */
+function clone<T extends Record<PropertyKey, unknown>>(
+  record: T,
+  { includeNonEnumerable = false } = {},
+) {
+  // Extract property and symbols
+  const keys = [
+    ...Object.getOwnPropertyNames(record),
+    ...Object.getOwnPropertySymbols(record),
+  ].filter((key) => includeNonEnumerable || record.propertyIsEnumerable(key));
+
+  // Build cloned record
+  const cloned = {} as T;
+  for (const key of keys as PropertyKeys) {
+    const v = record[key];
+    if (Array.isArray(v)) {
+      cloned[key] = [...v];
+      continue;
+    }
+    if (v instanceof Map) {
+      cloned[key] = new Map(v);
+      continue;
+    }
+    if (v instanceof Set) {
+      cloned[key] = new Set(v);
+      continue;
+    }
+    if (isMergeable<Record<PropertyKey, unknown>>(v)) {
+      cloned[key] = clone(v);
+      continue;
+    }
+    cloned[key] = v;
+  }
+
+  return cloned;
 }
 
 /**
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -47,7 +47,7 @@ export function deepMerge<
   other: U,
   options?: Options,
 ): DeepMerge<T, U, Options> {
-  //
+  // Extract options
   const {
     arrays = "merge",
     maps = "merge",
@@ -57,7 +57,7 @@ export function deepMerge<
 
   // Clone left operand to avoid performing mutations in-place
   type Result = DeepMerge<T, U, Options>;
-  const result = clone(record as Result, { includeNonEnumerable });
+  const result = clone(record, { includeNonEnumerable }) as Result;
 
   // Extract property and symbols
   const keys = [
@@ -121,7 +121,7 @@ export function deepMerge<
 /**
  * Clone a record
  * Arrays, maps, sets and objects are cloned so references doesn't point
- * anymore to those of cloned record
+ * anymore to those of target record
  */
 function clone<T extends Record<PropertyKey, unknown>>(
   record: T,
@@ -136,7 +136,7 @@ function clone<T extends Record<PropertyKey, unknown>>(
     includeNonEnumerable || record.propertyIsEnumerable(key)
   ) as Array<keyof T>;
 
-  // Build cloned record
+  // Build cloned record, creating new data structure when needed
   for (const key of keys) {
     type SameType = T[typeof key];
     const v = record[key];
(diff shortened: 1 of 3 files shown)


collections/deep_merge.ts
```diff
@@ -0,0 +1,116 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Merges the two given Records, recursively merging any nested Records with the second collection overriding the first in case of conflict
+ *
+ * For arrays, maps and sets, a merging strategy can be specified to either "replace" values, or "concat" them instead.
+ * Use "includeNonEnumerable" option to include non enumerable properties too.
```
ry, line 7: Please wrap at 80 columns.

(I'm surprised `deno fmt --check` didn't catch this)

lowlighter: it hasn't been caught since `deno fmt` didn't wrap them so I assume it's the same for `--check`

kt3k:
Tried on my end. Even typing seems working recursively with many cases. Great work!

But I found some cases where the typing doesn't match to the result, for example:
```
const a = { a: 1 };
const b = { a: "a" };
const c = deepMerge(a, b);
```
here c.a is "a", but it has `never` type. Can this be addressed somehow?

LionC:
Thanks for implementing this! I am **really*8 excited about a standardized `deepMerge` :-)

Some notes:

1. I **think** you could define recursive types yourself to actually have an outcome where the type of the right side overrides the one on the left. I think we should do this, as it would solve the problem pointed out by @kt3k . We should also be aware that `concat` will change the type of the resulting `Array` / `Map` etc.
2. The node `deep-merge` module offers an overloaded signature with a type parameter, allowing to specify the output and requiring the arguments to be partials `deepMerge<T extends SomeRecord>(a: Partial<T>, b: Partial<T>): T`. I would propose adding this, as this allows people to tell the compiler about their intent and get it to give them errors if they make a mistake.
3. I am not sure that `"replace"` is the default (meaning is more often used), in my experience IF you have collections on the same property name, you want to have all of them. Might be personal though.
4. Some more code notes but I will add them in a review

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Merges the two given Records, recursively merging any nested Records with
+ * the second collection overriding the first in case of conflict
+ *
+ * For arrays, maps and sets, a merging strategy can be specified to either
+ * "replace" values, or "concat" them instead.
+ * Use "includeNonEnumerable" option to include non enumerable properties too.
+ */
```
LionC, line 10: I think we should include an example, as that really helps to understand the basic idea of the function. See all other functions in the module for examples for...examples.

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+  // Ignore classes
+  if ((typeof value === "object") && ("constructor" in value!)) {
+    return !/^class /.test(value.constructor.toString());
+  }
+  return typeof value === "object";
+}
+
+/** Deep merge options */
+interface deepMergeOptions {
```
LionC, line 105: I think types should be capitalized, not sure if the styleguide says so explicitly though. It is also good practice to export option types, so people can explicitly use them in their code when building options e.g. y user input:

```suggestion
export interface DeepMergeOptions {
```

Also minor point: it could be an improvement to use a type alias instead here, as it (prevents people from adding things to it)[https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces] (accidentally):

```suggestion
export type DeepMergeOptions = {
```

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * For arrays, maps and sets, a merging strategy can be specified to either
+ * "replace" values, or "concat" them instead.
+ * Use "includeNonEnumerable" option to include non enumerable properties too.
+ */
+export function deepMerge<
+  T extends Record<PropertyKey, unknown>,
+  U extends Record<PropertyKey, unknown>,
+>(
+  collection: T,
```
LionC, line 15: We have changed the names of the arguments since the original proposal to simple things like `array` etc, maybe we could just do something like `obj` or `subject` here?

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+  ].filter((key) => includeNonEnumerable || other.propertyIsEnumerable(key));
+
+  // Iterate through each key of other object and use correct merging strategy
+  for (const key of keys as propertyKey[]) {
+    const a = collection[key], b = other[key];
+
+    // Handle arrays
+    if ((Array.isArray(a)) && (Array.isArray(b))) {
+      if (arrays === "concat") {
+        collection[key] = a.concat(...b);
```
LionC, line 37: You create an intermediate array here which needs to be created and collected - I would suggest doing this in place (mutating) instead:

```suggestion
        collection[key] = a.push(...b);
```

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+    if (isMergeable<T | U>(a) && isMergeable<T | U>(b)) {
+      collection[key] = deepMerge(a ?? {}, b);
+      continue;
+    }
+
+    // Override value
+    collection[key] = b;
+  }
+
+  return collection as T & U;
```
LionC, line 78: You are currently mutating the given object. This is very much not in line with the rest of the module, which always returns a new object immutable-style. I think we should be consistent here and create a new object to return instead.

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+  // Ignore builtins
+  if (
+    (value instanceof RegExp) || (value instanceof Date) ||
+    (value instanceof Array)
+  ) {
+    return false;
+  }
+  // Ignore classes
+  if ((typeof value === "object") && ("constructor" in value!)) {
+    return !/^class /.test(value.constructor.toString());
```
LionC, line 99: Is this the fastest way to do this? It feels weird to have a regex check run against everything to check if it is a class. Also see my comment above for an alternative idea.

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+ * Builtins object like, null and user classes are not considered mergeable
+ */
+function isMergeable<T>(value: unknown): value is T {
+  // Ignore null
+  if (value === null) {
+    return false;
+  }
+  // Ignore builtins
+  if (
+    (value instanceof RegExp) || (value instanceof Date) ||
```
LionC, line 92: Could we check the values `prototype` against `Object.prototype` instead or am I missing corner cases here?

collections/deep_merge.ts
```diff
@@ -0,0 +1,119 @@
...
+  /** Merging strategy for Sets */
+  sets?: "replace" | "concat";
+  /** Whether to include non enumerable properties */
+  includeNonEnumerable?: boolean;
+}
+
+// TypeScript does not support 'symbol' as index type currently though
+// it's perfectly valid
+// deno-lint-ignore no-explicit-any
+type propertyKey = any;
```
LionC, line 119: I think types should be capitalized

```suggestion
type FixedPropertyKey = any;
```

collections/deep_merge.ts
lowlighter: Indeed, I didn't know it was possible to do this 🙂  
Thanks for pointing this out

collections/deep_merge.ts
lowlighter: So I included an example, but used `assertEquals` from the testing instead.

What bugs me with the other examples is that some are `console.assert([] === [])` or `console.assert({} === {})` which will always be false since they do not refers to same reference. I know it's for examples, but I think in the future they should be changed as it could be misleading (especially if we later do `deno test --doc`

collections/deep_merge.ts
lowlighter: I added a cloner to avoid this

To keep along in this philosophy, the cloner also copies arrays, sets and maps, since it would be weird to not mutate the object but mutate stuff inside it

lowlighter:
> But I found some cases where the typing doesn't match to the result, for example:
>
> ```
> const a = { a: 1 };
> const b = { a: "a" };
> const c = deepMerge(a, b);
> ```
> here c.a is "a", but it has never type. Can this be addressed somehow? If not, maybe better to note something about it in the document
> (@kt3k )

> I think you could define recursive types yourself to actually have an outcome where the type of the right side overrides the one on the left. I think we should do this, as it would solve the problem pointed out by @kt3k . We should also be aware that concat will change the type of the resulting Array / Map etc.
> (@LionC)

I'll try to investigate to solve these typing issues if possible 👍 

It's possible that it comes from `if (isMergeable<T | U>(a) && isMergeable<T | U>(b))` and we should just let TypeScript infers the typing but didn't test yet

But if you have any leads on how to fix this, I'll take them 🙂 

Thanks for taking time to review and advise on this feature

collections/deep_merge.ts
LionC: Yes you are absolutely correct, that is still on my todo list and will not dtay that way.

collections/deep_merge.ts
```diff
@@ -0,0 +1,184 @@
...
+): T;
+
+export function deepMerge<
+  T extends Record<PropertyKey, unknown>,
+  U extends Record<PropertyKey, unknown>,
+>(
+  record: T,
+  other: U,
+  options?: DeepMergeOptions,
+): T;
```
lowlighter, line 38:
```suggestion
): : T & U;
```

lowlighter:
I think I'm not too far from a working typing, it seems mostly working.
I still need to check for recursive typing, along with the correct type overriding when changing merging strategy 👍 


Below are some examples:
____

Simple case:
image

Overriding from right:
image

Sets:
image

Arrays:
image

Maps:
image

lowlighter:
`deepMerge` typing is now fully functionning and should cover most cases from what I tested (see examples below).
It also handles the merging strategy, so if set to  `merge`, you'll get a data structure with union typing inside, and if set to `replace` you'll get the right-operand data structure typing 👍 

<table>
<tr>
<th>User code</th>
<th>Typing (infered)</th>
</tr>
<tr>
<td>

```ts
const a = deepMerge({
  foo:true,
  baz:{
    qux:NaN
  },
  quux:{
    grault:"test"
  }
}, {
  bar:"string",
  baz:{
    corge:true
  },
  quux:"ok"
})
```

</td>
<td>

```ts
const a: {
    foo: boolean;
    bar: string;
    quux: string;
    baz: {
        qux: number;
        corge: boolean;
    };
}
```

</td>
</tr>
<tr>
<td>

```ts
const a = deepMerge({
  foo:{
    bar:new Set([1, 2, 3]),
    baz:[1, 2, 3],
    qux:new Map([["a", 1], ["b", 2]])
  }
}, {
  foo:{
    bar:new Set(["a", "b", "c"]),
    baz:["a", "b", "c"],
    qux:new Map([["c", true]])
  }
})
```

</td>
<td>

```ts
const a: {
    foo: {
        bar: Set<string | number>;
        baz: (string | number)[];
        qux: Map<string, number | boolean>;
    };
}
```

</td>
</tr>
<tr>
<td>

```ts
const a = deepMerge({
  foo:{
    bar:new Set([1, 2, 3]),
    baz:[1, 2, 3],
    qux:new Map([["a", 1], ["b", 2]])
  }
}, {
  foo:{
    bar:new Set(["a", "b", "c"]),
    baz:["a", "b", "c"],
    qux:new Map([["c", true]])
  }
}, {sets:"replace", arrays:"replace", maps:"replace"})
```

</td>
<td>

```ts
const a: {
    foo: {
        bar: Set<string>;
        baz: string[];
        qux: Map<string, boolean>;
    };
}
```

</td>
</tr>

</table>

There are almost more typing logic than the actual algorithm but result seems worth it 🙂 

I think this is ready for review, I don't have anything to add and this seems goods to me for a first iteration

LionC:
> `deepMerge` typing is now fully functionning and should cover most cases from what I tested (see examples below).
> 
> It also handles the merging strategy, so if set to  `merge`, you'll get a data structure with union typing inside, and if set to `replace` you'll get the right-operand data structure typing 👍 
>
> There are almost more typing logic than the actual algorithm but result seems worth it 🙂 
> 
> 
> 
> I think this is ready for review, I don't have anything to add and this seems goods to me for a first iteration

Awesome! I will take a look tonight. Looking forward to it :-)

collections/deep_merge.ts
```diff
@@ -0,0 +1,385 @@
...
+  /** Merging strategy for Maps */
+  maps?: MergingStrategy;
+  /** Merging strategy for Sets */
+  sets?: MergingStrategy;
+  /** Whether to include non enumerable properties */
+  includeNonEnumerable?: boolean;
+};
+
+/**
+ * How does recursive typing works ?
```
kt3k, line 200: Great explanation! :clap:

LionC:
Me and @lowlighter are in contact ok Discord to finish this @kt3k , this will likely be ready for final review tomorrow.

LionC:
@kt3k we have decided to remove `nonEnumerable` props and made the code a bit more efficient regarding runtime, memory and readability. This should be good enough for a first releasable version for now - we thought about adding more generalized merging options, but decided to move them to a PR in the future.

TL;DR: This is ready to be merged^^

timreichen:
Instead of naming it `deepMerge`, how about `deepAssign` as in [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)?

lowlighter:
> Instead of naming it `deepMerge`, how about `deepAssign` as in [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign)?

It doesn't mutate the left operand like assign though. A new object is created instead

timreichen:
@lowlighter I think it would be a good idea to have it function more like the native shallow version aka assign, in a sense that it should mutate a target object and merge n objects instead of just two. That makes the function more flexible.
```ts
const mutatedA = deepAssign(a, b, c)
```
To create a new object
```ts
const d = deepAssign({}, a, b, c)
```

lowlighter:
Merging inplace was the original the design, but I've been told that std/collections' philosophy is to never mutate inplace and instead return a new object instead.

As for variable arguments I'm not against but it would make it awkward to pass options I think

kt3k:
@timreichen I think `deepAssign` is worth another proposal. Let's continue discussion in #1065.

kt3k (approved):
LGTM. Awesome work! @lowlighter @LionC

LionC:
> @lowlighter I think it would be a good idea to have it function more like the native shallow version aka assign, in a sense that it should mutate a target object and merge n objects instead of just two. That makes the function more flexible.
> 
> ```ts
> const mutatedA = deepAssign(a, b, c)
> ```
> 
> To create a new object
> 
> ```ts
> const d = deepAssign({}, a, b, c)
> ```

All functions in `collections` are pure right now and there is currently no plan to change that philosophy. I definitely see value in providing mutating functions (like `filterInPlace` which is used internally right now), especially for use in high-performance cases, but I think we would need to clearly distinguish them e.g. with their own subfolder.

timreichen:
I am sorry to write this after it has already been merged. But I think `deepMerge` is misnamed, once because functions should start with a verb (made the mistake in `deepAssign`, sorry) and secondly because I think it can be generalized withput loosing functionality.
Since it takes options why not call the function just `merge` and add an option for `objects` as well which will be set to `merge` by default as the others are to make it more general without loosing functionality?

LionC:
> I am sorry to write this after it has already been merged. But I think `deepMerge` is misnamed, once because functions should start with a verb (made the mistake as well naming it `deepAssign`, sorry) and secondly because I think it can be generalized without loosing functionality.
> Since it takes options why not call the function just `merge` and add an option for `objects` as well which will be set to `merge` by default as the others are to make it more general without loosing functionality?

Because `deepMerge` is a **very** widely used package  in the node space. While I might agree with your points in a vacuum, I think the concept of `deepMerge` is already so familiar for a lot of JS/TS devs, that it feels pragmatic to mirror that.

timreichen:
Yes, but even there the function name  of the API is `merge` not `deepMerge`. `deepmerge` is just the package name.

PR #1078, opened 2021-07-30
Return code 304 based on If-Modified-Since Header

timonson:
The If-Modified-Since request HTTP header makes the request conditional: the server will send back the requested resource, with a 200 status, only if it has been last modified after the given date. If the resource has not been modified since, the response will be a 304 without any body; the Last-Modified response header of a previous request will contain the date of last modification. Unlike If-Unmodified-Since, If-Modified-Since can only be used with a GET or HEAD.

When used in combination with If-None-Match, it is ignored, unless the server doesn't support If-None-Match.

The most common use case is to update a cached entity that has no associated ETag.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Modified-Since

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -286,8 +286,14 @@ export async function serveFile(
     headers.set("etag", simpleEtag);
 
     // If a `if-node-match` header is present and the value matches the tag return 304
+    // If a `if-modified-since` header is present and the value is bigger than
+    // the access timestamp value return 304
     const ifNoneMatch = req.headers.get("if-none-match");
-    if (ifNoneMatch && ifNoneMatch === simpleEtag) {
+    const ifModifiedSince = req.headers.get("if-modified-since");
+    if (
+      (ifNoneMatch && ifNoneMatch === simpleEtag) || (ifModifiedSince &&
+        fileInfo.mtime.getTime() < (new Date(ifModifiedSince).getTime() + 1000))
+    ) {
       response.status = 304;
       response.statusText = "Not Modified";
       return response;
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -286,8 +286,14 @@ export async function serveFile(
...
 
     // If a `if-node-match` header is present and the value matches the tag return 304
+    // If a `if-modified-since` header is present and the value is bigger than
+    // the access timestamp value return 304
     const ifNoneMatch = req.headers.get("if-none-match");
-    if (ifNoneMatch && ifNoneMatch === simpleEtag) {
+    const ifModifiedSince = req.headers.get("if-modified-since");
+    if (
+      (ifNoneMatch && ifNoneMatch === simpleEtag) || (ifModifiedSince &&
+        fileInfo.mtime.getTime() < (new Date(ifModifiedSince).getTime() + 1000))
```
kt3k, line 295: why `+ 1000`?

kt3k:
Please add test cases in `http/file_server_test.ts`.

http/file_server.ts
timonson: @kt3k because the `If-Modified-Since` has this format: `<day-name>, <day> <month> <year> <hour>:<minute>:<second> GMT` - meaning it doesn't break down to milliseconds. Therefore, the converted value (number of milliseconds* since the Unix Epoch) from `If-Modified-Since` will always be 1-1000 milliseconds lower.

kt3k: Thanks for the explanation. Makes sense.

kt3k (approved):
@timonson Thanks! LGTM

PR #1079, opened 2021-07-30
temporarily remove collections in order to correct author attribution

ry:
It will be returned immediately after landing with correct author attribution. I mistakenly had landed collections without attributing @LionC.

https://github.com/denoland/deno_std/commit/a23eb709ebd4cd8df4b29b0cf6103d4ae530c141
https://github.com/denoland/deno_std/pull/993


dsherret (approved):

ry:
Actually now that I think about it - this doesn't need to be done in two PRs.

https://github.com/denoland/deno_std/pull/1080 should suffice (as long as it's not squashed with landed)

PR #1076, opened 2021-07-30
chore: upgrade Rust to 1.54.0

magurotuna:
Upgrades Rust to 1.54.0.

PR #1081, opened 2021-07-31
build(_wasm_crypto): Upload rebuilt crypto.{js,wasm.js} if they don't match

jeremyBanks:
@magurotuna in #1076 has attempted to rebuild `_wasm_crypto` with the new version of Rust, but is unable to match the WASM binary expected on GitHub actions because they're using an M1 Mac, where the toolchain seems to produce *very slightly* different results.

For cases like that, this updates the associated GitHub Action to upload its rebuilt copies of `crypto.js` and `crypto.wasm.js` as artifacts, if they don't match, so that the developer can download and commit them regardless of the platform they're using.

You can see an example of what this looks like at the bottom of this page for a commit in this PR: https://github.com/denoland/deno_std/actions/runs/1084329739#artifacts

The Action was also mistakenly still referring to a mix of `std/hash/_wasm` and `std/crypto`. I've updated it to consistently refer to `std/crypto`, as that's the one that will be modified going forward.


kt3k:
@jeremyBanks Thank you for looking into this! This new workflow looks good to me, but probably it's hard for a new contributor to understand how this is working. Could you add some instructions in `_wasm_crypto/README.md` about what to do when `Verify WASM hasn't changed` step failed in the CI?

kt3k:
Let's merge this first as it's working already. Thanks for your contribution! @jeremyBanks

kt3k (approved):
LGTM

PR #1083, opened 2021-08-01
feat: fs/promises full implementation

guybedford:
This adds a full implementation for `fs/promises`.


guybedford:
Ahh of course there's already a partial `promises/mod.ts` but that seems like a lot of work... Let me know if you want to still proceed with this approach or not. Separate files seems like overkill to me personally though.

guybedford:
I've updated this PR to instead replace the previous promises implementation.

bartlomieju (approved):
LGTM

PR #1085, opened 2021-08-01
feat: remaining Node.js builtin aliases

guybedford:
This adds the remaining Node.js aliases:

* path/posix
* path/win32
* stream/promises
* timers/promises
* util/types

Which combined with https://github.com/denoland/deno_std/pull/1083 and https://github.com/denoland/deno_std/pull/1084 covers all the aliases.

PR #1086, opened 2021-08-01
docs: update Node.js builtin list, alias sys to util

guybedford:
This updates the Node.js builtin list assuming #1083, #1084, #1085 land.

It also aliases sys to util and extends the list to include the newest Node.js builtins.


ry (approved):
LGTM - thanks Guy

PR #1091, opened 2021-08-04
chore(collections): Clean up sort_by code and add BigInt selector tests.

mdx97:
This changeset updates the new `sort_by` function to be slightly more compact / idiomatic (imo). I also added tests for functions that return `bigint`s being provided for the `selector` parameter.

### DIFF
--- a/collections/sort_by.ts
+++ b/collections/sort_by.ts
@@ -24,26 +24,16 @@
  */
 export function sortBy<T>(
   array: Array<T>,
-  selector:
-    | ((el: T) => number)
-    | ((el: T) => string)
-    | ((el: T) => bigint)
-    | ((el: T) => Date),
+  selector: ((el: T) => number | string | bigint | Date),
 ): Array<T> {
-  const ret = Array.from(array);
-
-  return ret.sort((a, b) => {
+  return Array.from(array).sort((a, b) => {
     const selectedA = selector(a);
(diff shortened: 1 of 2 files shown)


collections/sort_by.ts
```diff
@@ -24,26 +24,16 @@
  */
 export function sortBy<T>(
   array: Array<T>,
-  selector:
-    | ((el: T) => number)
-    | ((el: T) => string)
-    | ((el: T) => bigint)
-    | ((el: T) => Date),
+  selector: ((el: T) => number | string | bigint | Date),
```
kt3k, line 27: By this change, the selector's return value can be mixed types. I don't think that is better design than the current one. (For example, we can provide a selector like `(item) => cond(item) ? "foo" : 1)`, which doesn't make much sense.

kt3k:
Thank you for the contribution. The test case looks good to me.

collections/sort_by.ts
mdx97: Ah, fair enough. I can revert that. Thanks for the tip!

Ayfri: Why don't you use type Selector ?

mdx97: @Ayfri good suggestion. Made that change as well.

kt3k (approved):
@mdx97 LGTM. Thanks!

PR #1094, opened 2021-08-04
chore(node/events): addListener and removeListener methods should have the correct names

uki00a:
Part of #911

This PR enables `test-event-emitter-method-names.js` and fixes `EventEmitter` to align with it.


kt3k (approved):
@uki00a LGTM. Thanks!

PR #1095, opened 2021-08-04
fix(http/file_server): Fix flaky 'file_server sets Date header correctly' test case

uki00a:
See #1058


kt3k (approved):
@uki00a Thanks! LGTM

PR #1098, opened 2021-08-05
BREAKING(std/collections): rename functions

timreichen:
rename `chunked` to `chunk`, since functions should start with verb.


grian32:
I don't think its a good idea to break stuff for a simple naming change especially when the naming follows what it was originally based on.

The naming is based off of Kotlin as was the function, as Kotlin uses adjective for functions that return a new array(`chunked`), and verb for functions that mutate it in place(`chunk`)

timreichen:
I don't think it is an issue on what it is based on. Std needs consistency across. Breaking is not a problem since std is not on v1 yet.

grian32:
> I don't think it is an issue on what it is based on. Std needs consistency across. Breaking is not a problem since std is not on v1 yet.

I don't really think consistency is as big of an issue as people make it out to be but yes it is not consistent with most of std, and a breaking change is a breaking change doesnt need to be v1 to break stuff lol

kt3k:
lodash also calls this function chunk, ref: https://lodash.com/docs/4.17.15#chunk . I'm slightly in favor of this change because of the consistency of the naming.

> as Kotlin uses adjective for functions that return a new array(chunked), and verb for functions that mutate it in place(chunk)

That seems nice rule for kotlin, but we don't have that rule in std/collections (most function names start with verbs, but create new arrays/objects). So there might be no reason to follow that rule here?

Does anyone have opinion on this?

timreichen:
I think that the philosophy of std/collections is to [always return a new instance](https://github.com/denoland/deno_std/pull/1075#issuecomment-893384370)

> The naming is based off of Kotlin as was the function, as Kotlin uses adjective for functions that return a new array(`chunked`), and verb for functions that mutate it in place(`chunk`)

so that distinction is either already inconsistent (`intersect` for example) or not necessary.

> Does anyone have opinion on this?

As far as I know std should prefer web api conventions over any other lib/framework for deno std in general.

LionC:
I do not have a strong opinion on this. Both names seem fine to me and the only reason the name is `chunked` right now is the `kotlin` inspiration.

I would prefer though that if we think rhere is a way to make this consistent, to make one PR for all renames and discuss naming consistency in there.

Would you do that @timreichen ?

timreichen:
@LionC can do. Would you prefer a new PR or adopting this one by changing the title?

LionC:
> @LionC can do. Would you prefer a new PR or adopting this one by changing the title?

Both works well I think, as long as the title reflects the content.

kt3k:
Is there any other candidate for renaming?

timreichen:
@kt3k 
`permutations` to `permute` and
`deepMerge` to `merge` [possibly](https://github.com/denoland/deno_std/pull/1075#issuecomment-894147267)

nayeemrmn:
> `permutations` to `permute`

This isn't correct, if you really wanted to verb-ify this it would have to be `getPermutations()`.

kt3k:
I don't think using noun for function names is necessarily bad.

See this stackoverflow for example https://stackoverflow.com/questions/1991324/function-naming-conventions Though the rule recommended there (verb for side effects, noun for non-side-effect) cannot be applied to std, naming function as noun looks common

timreichen:
> I don't think using noun for function names is necessarily bad.

> naming function as noun looks common

not bad per se, but it is not common in js. The goal here is just to make namings schemas as consistent as possible.
`getPermutations` is actually a good name compared to `window.getSelection` for example.

justinmchase:
> I don't think its a good idea to break stuff for a simple naming change...

Perhaps you could add:
```ts
/**
 * @deprecated Please use `chunk`. This function will be removed in a future version.
 */
export const chunked = chunk
```

LionC:
> > I don't think using noun for function names is necessarily bad.
> 
> 
> 
> > naming function as noun looks common
> 
> 
> 
> not bad per se, but it is not common in js. The goal here is just to make namings schemas as consistent as possible.
> 
> `getPermutations` is actually a good name compare to `window.getSelection` for example.

Very few things are "not common" in JS - it is a very style agnostic language and almost every style has a lot of usage.

LionC:
Having thought about this a bit more, I think the idea behind nouns or adjectives in kotlin (which inspired the names) is that it does not **do** anything to the passed `Array` (meaning it does not mutate it) but rather builds a different / derived thing of of it that it gives you. It is almost like a constructor (`new` is an adjective as well, at least in German^^).

With that in mind, I am not too sure that we need to derive from the names this was inspired by - they stem from a commonly used library that has very similiar standarda around naming.

timreichen:
@LionC I don't know of any example of a method or function that starts with an adjective in the native js landscape. Usually they are prefixed with a verb so the common js user probably interprets `chunked` as `isChunked`.
`new` is a keyword and not really comparable here.
 Functions/methods usually starts by verb or are in rare cases with a noun when it is part of a prototype (`Array.prototype.entries` for example). Since this is not the case here, I still would rename it to `chunk`.

kt3k:
Let's do `chunked` -> `chunk` rename first with this PR.

Please open new PRs for other renamings

kt3k (approved):

PR #1100, opened 2021-08-05
feat(collections): Add withoutAll that returns an array excluding all given values

getspooky:
Returns an array excluding all given values, For performance if given array is large size(bigger than `200`) , we store values in `Set` datastructure. 

### Example: 

```ts
const withoutList = withoutAll([1, 2, 3, 4, 6, 3], [1, 2], [3, 4, 6, 3]);
```

### Discussion

Please see https://github.com/denoland/deno_std/issues/1065

### DIFF
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -9,11 +9,10 @@ const LARGE_ARRAY_SIZE = 200;
  * Example:
  *
  * ```ts
- * import { withoutAll } from "./without.ts";
+ * import { withoutAll } from "./withoutAll.ts";
  *
  * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
  *
- * console.assert(withoutList === [ 3 ])
  * ```
  */
 export function withoutAll<T>(
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -1,0 +1,33 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/** Used as the size to enable large array optimizations. */
+const LARGE_ARRAY_SIZE = 200;
+
+/**
+ * Returns an array excluding all given values.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { withoutAll } from "./without.ts";
+ *
+ * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
+ *
+ * console.assert(withoutList === [ 3 ])
+ * ```
+ */
+export function withoutAll<T>(
+  array: Array<T>,
+  values: Array<T>,
+): Array<T> {
+  if (array.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+
+    for (const element of array) {
+      set.add(element);
+    }
+
+    array = Array.from(set);
+  }
+  return array.filter((value) => values.indexOf(value) === -1);
+}
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -6,7 +6,7 @@
  * Example:
  *
  * ```ts
- * import { withoutAll } from "./withoutAll.ts";
+ * import { withoutAll } from "./without_all.ts";
  *
  * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
  *
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -14,10 +14,7 @@
  * assertEquals(withoutList, [3]);
  * ```
  */
-export function withoutAll<T>(
-  array: Array<T>,
-  values: Array<T>,
-): Array<T> {
+export function withoutAll<T>(array: readonly T[], values: readonly T[]): T[] {
   const toExclude = new Set(values);
   return array.filter((it) => !toExclude.has(it));
 }
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -7,9 +7,11 @@
  *
  * ```ts
  * import { withoutAll } from "./without_all.ts";
+ * import { assertEquals } from "../testing/asserts.ts";
  *
  * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
  *
+ * assertEquals(withoutList, [3]);
  * ```
  */
 export function withoutAll<T>(
--- a/collections/without_all.ts
+++ b/collections/without_all.ts
@@ -1,8 +1,5 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-/** Used as the size to enable large array optimizations. */
-const LARGE_ARRAY_SIZE = 200;
-
 /**
  * Returns an array excluding all given values.
  *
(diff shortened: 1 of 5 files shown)


LionC:
I can take a first look tonight :-)

collections/without_all_test.ts
```diff
@@ -0,0 +1,53 @@
...
+function withoutAllTest<I>(
+  input: Array<I>,
+  excluded: Array<I>,
+  expected: Array<I>,
+  message?: string,
+) {
+  const actual = withoutAll(input, excluded);
+  assertEquals(actual, expected, message);
+}
+
```
LionC, line 15: A "no mutation" test making sure that the original array that gets passed in does not get changed would be nice - we have that in all other function test suites as well

collections/without_all.ts
```diff
@@ -0,0 +1,32 @@
...
+  values: Array<T>,
+): Array<T> {
+  if (array.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+
+    for (const element of array) {
+      set.add(element);
+    }
+
+    array = Array.from(set);
```
LionC, line 29: This seems wrong - you are `distinct`ing the original array. 

```ts
withoutAll(new Array(120).fill(3), [ 1 ])
```

will return `[3]` instead of an array of 120 `3`s. I think @kt3k already mentioned that in your last PR.

collections/without_all.ts
```diff
@@ -0,0 +1,32 @@
...
+export function withoutAll<T>(
+  array: Array<T>,
+  values: Array<T>,
+): Array<T> {
+  if (array.length >= LARGE_ARRAY_SIZE) {
+    const set = new Set<T>();
+
+    for (const element of array) {
+      set.add(element);
+    }
```
LionC, line 27: I do not think a `LARGE` check is necessary. Building a set **from the `values`** is better runtime in all cases where `values` is longer than 1, at the cost of one `Set` in memory - I would definitely do that.

I think this might be enough to implement the whole thing:

```suggestion
    const toExclude = new Set(values);
    
    return array.filter((it) => toExclude.has(it));
```

collections/without_all_test.ts
```diff
@@ -0,0 +1,53 @@
...
+Deno.test({
+  name: "[collections/withoutAll] empty input",
+  fn() {
+    withoutAllTest([], [], []);
+  },
+});
+
```
LionC, line 22: We should add a test to check against the bug mentioned above

```suggestion
Deno.test({
    name: "[collection/withoutAll] leaves duplicate elements",
    fn () {
        withoutAllTest(new Array(110).fill(3), [1], new Array(110).fill(3));
    },
});
```

LionC (changes requested):

collections/without_all.ts
kt3k: I'm ok with the above suggested implementation.

If we do care about the memory/speed trade off very much, I guess it could be something like this:
```ts
if (values.length >= LARGE) {
  const toExclude = new Set(values);
  return array.filter((it) => !toExclude.has(it));
}
return array.filter((it) => values.indexOf(it) === -1);
```

LionC: I mean the memory it takes scales with the input array. So I would argue that it is worth it at any length of `values`, because (with `V` as length of `values` and `N` as length of `array`) it always reduces the runtime from `N*V` to `N + V` (which is very significant) and only takes `V` memory (so not a lot for small `V`s).

kt3k: Sounds reasonable. Ok. Let's simply create a set always

collections/without_all.ts
```diff
@@ -0,0 +1,32 @@
...
+ * import { withoutAll } from "./withoutAll.ts";
+ *
+ * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
+ *
+ * ```
+ */
+export function withoutAll<T>(
+  array: Array<T>,
+  values: Array<T>,
+): Array<T> {
```
zhangyuannie, line 21: From the function signature and the description, it is unclear what this function actually does. Is this meant to always return `[]` since "an array excluding all given values" is returned?

Couple suggestions:
1. Tell the user what `withoutList` looks like after the function call with `assertEquals` or a comment.
2. Give parameters better names. `array` and `values` does not provide additional information to the user.
3. Maybe a better description would also be helpful.

collections/without_all_test.ts
getspooky: The `filter()` method creates `a new array` with all elements that pass the test implemented by the provided function. so if we want `no mutation` we must create own filter function in order to return the same array instance.

collections/without_all.ts
getspooky: Sure, You can read `Collections/README.md`, it contains clear description with example.

collections/without_all.ts
```diff
@@ -0,0 +1,21 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Returns an array excluding all given values.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { withoutAll } from "./withoutAll.ts";
```
kt3k, line 9: Looks like a typo in module path

```suggestion
 * import { withoutAll } from "./without_all.ts";
```

collections/without_all_test.ts
LionC: I meant to write a test that makes sure that the array you pass in does **not** get changed. This is true with the current implementation, I just say we should have a test to ensure that it stays so.

```suggestion
Deno.test({
  name: "[collections/withoutAll] no mutation",
  fn() {
    const array = [1, 2, 3, 4];
    withoutAll(array, [2, 3]);
    
    assertEquals(array, [1, 2, 3, 4]);
  },
});
```

collections/without_all.ts
```diff
@@ -0,0 +1,21 @@
...
+ * Returns an array excluding all given values.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { withoutAll } from "./without_all.ts";
+ *
+ * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
+ *
+ * ```
```
LionC, line 13: I would adapt the example to make it consistent with the others in the module:

```suggestion
 * ```ts
 * import { withoutAll } from "./without_all.ts";
 * import { assertEquals } from "../testing/asserts.ts";
 *
 * const withoutList = withoutAll([2, 1, 2, 3], [1, 2]);
 *
 * assertEquals(withoutList, [3]);
 * ```
```

LionC (changes requested):
Just the example left, the redt looks good to me

collections/without_all.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export function withoutAll<T>(
+  array: Array<T>,
+  values: Array<T>,
+): Array<T> {
```
zhangyuannie, line 18:
```suggestion
export function withoutAll<T>(array: readonly T[], values: readonly T[]): T[] {
```
See https://github.com/denoland/deno_std/pull/1119

LionC: Good point, I missed that. I am clearly too used to read `Array<T>` :-)

LionC (approved):
LGTM now, pinging @kt3k for re-review

zhangyuannie (approved):

kt3k (approved):
@getspooky LGTM. Nice work! Thanks!

Thanks @LionC @zhangyuannie for reviewing!

Issue #1073, opened 2021-07-27
Rename assertThrowsAsync to assertRejects


caspervonb:
The name `assertThrowsAsync` is overly long and verbose, yet it doesn't describe what it actually asserts which is that a promise rejects.

Rename `assertThrowsAsync` -> `assertRejects`.

getspooky:
Yes `assertRejects` is more meaningful.

PR #1101, opened 2021-08-05
BREAKING(testing): Renaming assertThrowsAsync to assertRejects

getspooky:
The name `assertThrowsAsync` is overly long and verbose, yet it doesn't describe what it actually asserts which is that a promise rejects. Rename `assertThrowsAsync` -> `assertRejects`.

Please see: https://github.com/denoland/deno_std/issues/1073


caspervonb (commented):
You'd have to apply this to all modules in the entire std library for the tests to pass.

But this hasn't been discussed yet, as you can see in the unlabelled issue you linked.

Probably the way we'd want to go forward is to keep exporting the old alias with deprecated annotations for a couple of releases as this is used a lot on the wild, including in CLI unit tests.

bartlomieju (changes requested):
I agree with Casper, also I'd suggest to replace all usages of `assertThrowsAsync` in the code in this PR

kt3k (approved):
LGTM

bartlomieju (approved):

caspervonb (approved):

kt3k:
@getspooky Thank you for your contribution!

@caspervonb Thank you for the original suggestion!

PR #1103, opened 2021-08-05
Add `mapNotNullish` to `collections`

LionC:
This adds `mapNotNullish<T, O>(array: Array<T>, transformer: (el: T) => O): Array<NonNullable<O>>` to `collections`. This method helps in common cases like extracting an optional property from objects or `map`ping and `filter`ing an array at the same time:

Example 1:

```ts
const processes = [
    { name: "Building", problem: null },
    { name: "Testing", problem: "Could not load test file" },
    { name: "Benchmarking", problem: "OutOfMemory" },
    { name: "Deploying", problem: undefined },
]

const problems = mapNotNullish(processes, it => it.problem)
// [ "Could not load test file", "OutOfMemory" ]
```

Example 2:

```ts
const numbers = [ 1, 2, 3, 4 ]

const evensSquared = mapNotNullish(numbers, it => it % 2 === 0 ? it*it : null)
// [4, 16]
```

`mapNotNullish(someArray, someFunction)` has the same result as

```ts
someArray
    .map(someFunction)
    .filter((it) => it !== undefined && it !== null)
```

but is more efficient regarding runtime and memory (only one iteration, no intermediate array etc), states the intent shorter and more specifically and gets the types right (which `filter` does not without explicit type guards).

Like a lot of the `collections` functions, `mapNotNullish()` is heavily inspired by `kotlin`s `.mapNotNull()`.


kt3k (approved):
@LionC LGTM.

As explained in the description, this is optimized version of `.map().filter()` idiom. I think this has decent use cases, especially when the source array is large.

collections/map_not_nullish.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Returns a new array, containing all elements in the given array transformed using the given transformer, except the ones
+ * that were transformed to `null` or `undefined`
+ *
+ * Example:
+ *
+ * ```typescript
```
kt3k, line 9: We prefer `ts` tag, but let's address this later.

caspervonb: Won't be typechecked, but I'll address that upstream in cli.

LionC: Oh sorry - I keep forgetting that. Should I post a separate PR?

PR #1104, opened 2021-08-05
Fix `NaN` handling in `collections/sortBy` ( closes #1092 )

### LINKED ISSUE #1092: [std/collections] sortBy returns strange result if selector returns NaN sometimes
Because NaN can't be compared correctly with other numbers, the output becomes broken when the selector returns NaN sometimes.

For example, the following outputs `[NaN, 1, NaN, 5, 7, NaN, 2,   6, 9]`
```ts
import { sortBy } from "./collections/sort_by.ts";

console.log(sortBy([NaN, 1, NaN, 7, NaN, 2, 6, 5, 9], (x) => x));
```

LionC:
This fixes `NaN` handling in `sortBy` as mentioned in #1092 by adding a simple explicit check. Also adds two tests for that behaviour.

### DIFF
--- a/collections/sort_by.ts
+++ b/collections/sort_by.ts
@@ -36,6 +36,16 @@ export function sortBy<T>(
     const selectedA = selector(a);
     const selectedB = selector(b);
 
+    if (typeof selectedA === "number") {
+      if (Number.isNaN(selectedA)) {
+        return 1;
+      }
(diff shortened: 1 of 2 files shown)


kt3k (approved):
@LionC LGTM. Thanks!

PR #1105, opened 2021-08-06
fix(async): let async exception of `deadline` can be caught

justjavac:
```ts
try {
  await deadline(Promise.reject("booom"), 20);
} catch (error) {
  console.log(`catch error: ${error}`);
}
```

will crash:

```plain
catch error: booom
error: Uncaught (in promise) booom
```


bartlomieju:
@justjavac looks good, could you add a test case that exercises provided example?

justjavac:
done

ry (approved):
LGTM - thanks @justjavac

PR #1108, opened 2021-08-07
feat(collections): add sumOf

grian32:
Not sure if there should be more tests, couldn't really think of anything else as its a pretty simple function.

Related: #1065


collections/sum_of_test.ts
```diff
@@ -0,0 +1,32 @@
...
+import { sumOf } from "./sum_of.ts";
+
+Deno.test("[collections/sumOf] On object properties", () => {
+  const object = [
+    { name: "Kyle", age: 34 },
+    { name: "John", age: 42 },
+    { name: "Anna", age: 23 },
+  ];
+
+  const actual = sumOf(object, (i) => i.age);
```
caspervonb, line 13: Sum of negatives and positives, multiple negatives, etc?

grian32: Added some more tests related to negatives, thanks!

collections/sum_of.ts
```diff
@@ -0,0 +1,35 @@
...
+ *     { name: 'John', age: 23 },
+ * ]
+ * const totalAge = sumOf(people, i => i.age)
+ *
+ * console.assert(totalAge === 99)
+ * ```
+ */
+export function sumOf<T>(
+  array: Array<T>,
+  selector: Selector<T, number>,
```
LionC, line 25: This was actually just in the proposal for internal clarity - I would suggest "flattening" the argument types, as the language server will otherwise just show `Selector`, which does not really help when you want to use the function quickly. I would suggest using a flat type in the args:

```suggestion
/**
 * Applies the given selector to all elements in the given collection and calculates the sum of the results
 *
 * Example:
 *
 * ```typescript
 * import { sumOf } from "./sum_of.ts"
 *
 * const people = [
 *     { name: 'Anna', age: 34 },
 *     { name: 'Kim', age: 42 },
 *     { name: 'John', age: 23 },
 * ]
 * const totalAge = sumOf(people, i => i.age)
 *
 * console.assert(totalAge === 99)
 * ```
 */
export function sumOf<T>(
  array: Array<T>,
  selector: (el: T) => number,
```

collections/sum_of.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { Selector } from "./types.ts";
+
+/**
+ * Applies the given selector to all elements in the given collection and calculates the sum of the results
+ *
+ * Example:
+ *
+ * ```typescript
```
LionC, line 10: Apparently we have to use the `ts` tag instead of `typescript` for `deno test` to pick it up - and the rest of `std` uses `ts` as well. That was a mistake on my part in the original proposal, sorry.

collections/sum_of.ts
```diff
@@ -0,0 +1,35 @@
...
+ * import { sumOf } from "./sum_of.ts"
+ *
+ * const people = [
+ *     { name: 'Anna', age: 34 },
+ *     { name: 'Kim', age: 42 },
+ *     { name: 'John', age: 23 },
+ * ]
+ * const totalAge = sumOf(people, i => i.age)
+ *
+ * console.assert(totalAge === 99)
```
LionC, line 20: We should start fixing that equality example thing. Others have used `assertEquals` now and I think that is the best for the moment:

```suggestion
 * import { sumOf } from "./sum_of.ts"
 * import { assertEquals } from "../testing/mod.ts"
 *
 * const people = [
 *     { name: 'Anna', age: 34 },
 *     { name: 'Kim', age: 42 },
 *     { name: 'John', age: 23 },
 * ]
 * const totalAge = sumOf(people, i => i.age)
 *
 * assertEquals(totalAge, 99)
```

collections/sum_of_test.ts
```diff
@@ -0,0 +1,88 @@
...
+import { assertEquals } from "../testing/asserts.ts";
+import { sumOf } from "./sum_of.ts";
+
+Deno.test("[collections/sumOf] On object properties", () => {
+  const object = [
+    { name: "Kyle", age: 34 },
+    { name: "John", age: 42 },
+    { name: "Anna", age: 23 },
+  ];
+
```
LionC, line 12: I would add test for

- the function not mutating the original array that was passed (see all other function test suites in the module)
- empty input array should result in a sum of 0
- `NaN` and `Infinity` behaviour

collections/sum_of.ts
```diff
@@ -0,0 +1,35 @@
...
+  selector: Selector<T, number>,
+): number {
+  const selected = array.map(selector);
+  let sum = 0;
+
+  for (const i of selected) {
+    sum += i;
+  }
+
+  return sum;
```
LionC, line 34: This is a bit more inefficient than it needs to, as `map` will create an intermediary array of the same size that eats memory and needs to be collected. You could instead apply the `selector` within your `for`. Minor nitpick: `I` is idiomatically used as an index when iterating but here it is the actual element. I would use `element` or something else.

```suggestion
  let sum = 0;

  for (const element of selected) {
    sum += selector(element);
  }

  return sum;
```

LionC (changes requested):

LionC:
Totally unrelated question now that I am seeing this implemented - should we maybe support `bigint`?

collections/sum_of.ts
grian32: I'll change to `ts` for consistency but im pretty sure it picked up on it, as I got ci errors from that

LionC: @kt3k mentioned it in my `mapNotNullish` PR, but it might already have been fixed

grian32:
> ~Totally unrelated question now that I am seeing this implemented - should we maybe support bigint?~
> 
> Edit: As I understand it, `bigint` supports `+` without weird caveats, so you would "just" need to add it to the typing of the function to support it. Sounds like an easy win?

I'd need some extra checks so `sum` could start off as bigint instead of number, not sure if that'd hurt performance but I'll push it eitherway.

caspervonb:
> Edit: Actually, I'm not really sure how to implement this. as I cant get typescript to really agree since the selector would be a union type and it doesn't like adding that to bigint or number even if i have typeguards

Would have to make it generic:

```
function sumOf<T>(array: T[]): T;
```

Keep in mind that you can't add a number and a bigint. You'll have to explicitly convert or get a runtime error.

LionC:
As the type is determined by the selector, you would probably use

```ts
S extends (el: T) => number | (el: T) => bigint
```

as the type for `selector` and `ReturnType<S>` as the return type for the whole function.

The only problem is the empty input case - it needs to return `0` but you cannot know if that is `bigint` 0 or number `0`...

grian32:
I don't think I'll be implementing bigint support in this PR as I don't know how to implement it(both type wise and the actual implementation)

kt3k (approved):
@grian32 LGTM. I think this is enough for the first pass. Thank you for your contribution!

kt3k:
With regard to bigint support, how about providing explicit option or alternative function?

```
sumOf(arr, func, { bigInt: true });
sumOfBigInt(arr, func);
```

LionC:
> With regard to bigint support, how about providing explicit option or alternative function?
> 
> 
> 
> ```
> 
> sumOf(arr, func, { bigInt: true });
> 
> sumOfBigInt(arr, func);
> 
> ```

I think this is actually the only way to do this with the current signature - I do not see how we can decide which `0` to return for an empty input array.

That is something for another day then :-) Thank you alot @grian32 , good work!

PR #1109, opened 2021-08-07
fix (collections): Add missing functions to mod.ts

LionC:
This adds functions to `collection`s `mod.ts` which I have forgotten to add when implementing them :see_no_evil:


kt3k (approved):
@LionC Thanks! LGTM

PR #1111, opened 2021-08-07
feat(statistics): Implement mean, median, and mode.

mdx97:
This PR is a proposal / proof-of-concept to add a `statistics` module to the standard library.

Whilst discussing with @LionC on Discord about the implementation of the std/collections `average` function, he suggested that it may make sense to break out a new statistics module for all the central tendency functions (mean, median, and mode).

I prefer this approach for the following (completely subjective) reasons:
- Including a function to calculate the arithmetic mean without functions for median and mode feels like an incomplete API. (if a user wants to write an application that uses primitive statistics functions, they can use the built in `average` function but would have to either roll their own median and mode functions or download a 3rd party library?)
- Location of the `average` function in the `collections` module seems a bit clunky because it operates strictly on arrays of `number`s, rather than generic collections like every other function does.

NOTE: mean and median are relatively straight forward implementations. However, how we implement mode should probably be discussed in a bit more detail. Do we want to return a Set here? An array? If an array, in what order? Or do we just want to return a single mode?

I have included the version which returns a Set in the PR. Here are some pros / cons off the top of my head:

Pros
- Supports multi-modes.
- Does not suggest any sort of order to the output.
- Internally, gives us performant (O(1)) look ups / insertions. (and returning directly means we don't have to copy to an Array)

Cons
- It is an awkward API to work with (in the case of a single mode, how would the user extract that single value?)

### DIFF
--- a/statistics/median.ts
+++ b/statistics/median.ts
@@ -6,6 +6,9 @@
  * Example:
  *
  * ```ts
+ * import { assertEquals } from "../testing/asserts.ts"
+ * import { median } from "./median.ts"
+ *
  * const numbers = [ 4, 2, 7 ]
  * assertEquals(median(numbers), 4)
  * ```
--- a/statistics/median.ts
+++ b/statistics/median.ts
@@ -1,0 +1,27 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Returns the arithmetic median of the numbers in the given collection
+ *
+ * Example:
+ *
+ * ```ts
+ * const numbers = [ 4, 2, 7 ]
+ * assertEquals(median(numbers), 4)
+ * ```
+ */
+export function median(collection: Array<number>): number | undefined {
+  if (collection.length === 0) {
+    return undefined;
+  }
+
+  const middle = Math.floor(collection.length / 2);
+  const values = Array.from(collection);
+  values.sort((a, b) => a - b);
+
+  if (values.length % 2 === 0) {
+    return (values[middle] + values[middle - 1]) / 2;
+  }
+
+  return values[middle];
+}
--- a/statistics/mode.ts
+++ b/statistics/mode.ts
@@ -6,6 +6,9 @@
  * Example:
  *
  * ```ts
+ * import { assertEquals } from "../testing/asserts.ts"
+ * import { mode } from "./mode.ts"
+ *
  * const numbers = [ 4, 2, 7, 4 ]
  * assertEquals(mode(numbers), 4)
  * ```
--- a/statistics/mode.ts
+++ b/statistics/mode.ts
@@ -1,0 +1,37 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Returns the arithmetic mode(s) of the numbers in the given collection
+ *
+ * Example:
+ *
+ * ```ts
+ * const numbers = [ 4, 2, 7, 4 ]
+ * assertEquals(mode(numbers), 4)
+ * ```
+ */
+export function mode(collection: Array<number>): Set<number> | undefined {
+  if (collection.length === 0) {
+    return undefined;
+  }
+
+  const counter: Record<number, number> = {};
+  counter[collection[0]] = 1;
+  const maxes = new Set([collection[0]]);
+  let maxCount = 1;
+
+  for (let i = 1; i < collection.length; i++) {
+    const number = collection[i];
+    counter[number] = (counter[number] || 0) + 1;
+
+    if (counter[number] === maxCount) {
+      maxes.add(number);
+    } else if (counter[number] > maxCount) {
+      maxes.clear();
+      maxes.add(number);
+      maxCount = counter[number];
+    }
+  }
+
+  return maxes;
+}
--- a/statistics/mode_test.ts
+++ b/statistics/mode_test.ts
@@ -33,6 +33,6 @@ Deno.test({
     assertEquals(mode([2, 4]), new Set([2, 4]));
     assertEquals(mode([1, 8, 1, 8]), new Set([1, 8]));
     assertEquals(mode([8, 20, 20, 9, 20, 8, 8, 9]), new Set([8, 20]));
-    assertEquals(mode([1, 3, 4, 5, 3, 1, 4]), new Set([1, 3, 4]))
+    assertEquals(mode([1, 3, 4, 5, 3, 1, 4]), new Set([1, 3, 4]));
   },
 });
--- a/statistics/mode_test.ts
+++ b/statistics/mode_test.ts
@@ -1,0 +1,38 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "../testing/asserts.ts";
+import { mode } from "./mode.ts";
+
+Deno.test({
+  name: "[statistics/mode] empty array",
+  fn() {
+    assertEquals(mode([]), undefined);
+  },
+});
+
+Deno.test({
+  name: "[statistics/mode] single element",
+  fn() {
+    assertEquals(mode([1]), new Set([1]));
+    assertEquals(mode([5]), new Set([5]));
+  },
+});
+
+Deno.test({
+  name: "[statistics/mode] one mode",
+  fn() {
+    assertEquals(mode([2, 2]), new Set([2]));
+    assertEquals(mode([1, 5, 7, 9, 7]), new Set([7]));
+    assertEquals(mode([8, 20, 20, 9, 20, 8]), new Set([20]));
+  },
+});
+
+Deno.test({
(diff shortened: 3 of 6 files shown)


kt3k:
I'm interested in `std/statistics` module. Statistics can be a good appealing use case of Deno in my view.

> However, how we implement mode should probably be discussed in a bit more detail.

Python's statistics module seems having both `mode` and `multimode`, and their `mode` returns the first one encountered in the source data. How about following this design?

LionC:
I can do a fiest review tonight if no one has done it by then.

I also think it would make sense to offer an optional mapping function argument to each statistics function. In most real world use cases you want to calculate based on some property / derived data and when statistics is involved, that data set might be big, which means manually `map`ping before is expensive in memory and potentially runtime vs inline `map`ping within the function.

With `collections/sortBy`, you also have a utility that you can easily pass that stuff do for things like percentiles etc.

statistics/mode.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Returns the arithmetic mode(s) of the numbers in the given collection
+ *
+ * Example:
+ *
+ * ```ts
+ * import { assertEquals } from "../testing/asserts.ts"
+ * import { mode } from "./mode.ts"
+ *
+ * const numbers = [ 4, 2, 7, 4 ]
+ * assertEquals(mode(numbers), 4)
```
kt3k, line 13: Doesn't seem matching to the implementation?

matthorner: You're right, I forgot to update the example when I changed the return value to be multimodal. I'll fix that when I get off work tonight 👍

mdx97: You're right, I'll fix when I'm off work.

mdx97:
> I'm interested in `std/statistics` module. Statistics can be a good appealing use case of Deno in my view.
> 
> > However, how we implement mode should probably be discussed in a bit more detail.
> 
> Python's statistics module seems having both `mode` and `multimode`, and their `mode` returns the first one encountered in the source data. How about following this design?

Yeah I was looking at Python's module as well. Implementing a new `multimode` function is definitely an option, and I like it from a user POV as well (if we only expect / care about a single mode, we don't want to have to worry about extracting that from a multi mode result set). It also gives a certain level of consistency to the three "primitive" functions mean, median, and mode (single return value).

statistics/mode_test.ts
```diff
@@ -0,0 +1,38 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "../testing/asserts.ts";
+import { mode } from "./mode.ts";
+
```
LionC, line 5: A no mutation test would fit in here as well.

LionC (changes requested):
Looked through it and added some comments, primarily related to the changes mentioned in the conversation comment.

ry:
Would it operate on typed arrays as well as arrays? I think it ought to follow the numpy API. But if it's following numpy, the question is why we wouldn't just use TFJS directly...

LionC:
> Would it operate on typed arrays as well as arrays? I think it ought to follow the numpy API. But if it's following numpy, the question is why we wouldn't just use TFJS directly...

@ry I suggested relying on `Iterable`s in general, as the return types of statistics functions are very predefined and do not really depend on the specific `Iterable` implementation input.

As to why to do it instead of relying on third party code - this discussion seems to arise every time we are talking about adding something new. I personally think that there is value in not relying on third party code for common use cases and if there are already people implementing and reviewing it in a good enough quality, it should not really conflict with other goals.

kt3k:
@ry 
> Would it operate on typed arrays as well as arrays? I think it ought to follow the numpy API. But if it's following numpy, the question is why we wouldn't just use TFJS directly...

Does TFJS support statistics features? I only found `tf.mean`, but not median, mode, variance, etc...

numpy seems having rich set of statistics routines. This seems a good inspiration for `std/statistics` to me, (though some APIs look strangely named, (ptp, nanmean, etc) https://numpy.org/doc/stable/reference/routines.statistics.html

ry:
Why call it statistics instead of math?

LionC:
> Why call it "statistics" instead of "math"? Seems limiting.

I agree with what you are saying at its core, but `math` might conflict with the `Math.` globals too much.

kt3k:
Talked a little bit internally with @ry, now I think `std/math` is a good collective name for math related stuff. `statistics` is probably too granular name for a small category of items.

> I agree with what you are saying at its core, but math might conflict with the Math. globals too much.

Because these two are in different places (global vs std/), I don't think these necessarily conflict.

LionC:
> Talked a little bit internally with @ry, now I think `std/math` is a good collective name for math related stuff. `statistics` is probably too granular name for a small category of items.
> 
> 
> 
> > I agree with what you are saying at its core, but math might conflict with the Math. globals too much.
> 
> 
> 
> Because these two are in different places (global vs std/), I don't think these necessarily conflict.

Yep it is not a big issue, might just be an inconvenience when doing `import *`. `math` is definitely a good name for the module!

ry:
> Does TFJS support statistics features? I only found tf.mean, but not median, mode, variance, etc...

Sure, it's a sophisticated statistics library (median doesn't have a standalone function but can be done)
https://js.tensorflow.org/api/latest/#moments

statistics/median.ts
```diff
@@ -0,0 +1,30 @@
...
+  if (collection.length === 0) {
+    return undefined;
+  }
+
+  const middle = Math.floor(collection.length / 2);
+  const values = Array.from(collection);
+  values.sort((a, b) => a - b);
+
+  if (values.length % 2 === 0) {
+    return (values[middle] + values[middle - 1]) / 2;
```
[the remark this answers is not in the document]
justinmchase: This was a little confusing to me since you're essentially getting the two values in reverse order that they appear in the set... I know it doesn't matter mathematically but it would have helped me at least with readability to reverse them like:

```ts
return (values[middle-1] + values[middle]) / 2
```

Or maybe:
```ts
return mean(values.slice(middle - 1, middle + 1))
```

statistics/mode.ts
```diff
@@ -0,0 +1,40 @@
...
+ * ```
+ */
+export function mode(collection: Array<number>): Set<number> | undefined {
+  if (collection.length === 0) {
+    return undefined;
+  }
+
+  const counter: Record<number, number> = {};
```
justinmchase, line 21: You should use `Map<number, number>`. A record will stringify the numbers in the keys technically.

```ts
> Object.entries({ 0: 0, 1: 1 })
[ [ "0", 0 ], [ "1", 1 ] ]
```

vs
```ts
> let m = new Map<number, number>()
undefined
> m.set(0, 0)
Map { 0 => 0 }
> m.set(1, 1)
Map { 0 => 0, 1 => 1 }
> [...m]
[ [ 0, 0 ], [ 1, 1 ] ]
```

statistics/mode.ts
```diff
@@ -0,0 +1,40 @@
...
+  const maxes = new Set([collection[0]]);
+  let maxCount = 1;
+
+  for (let i = 1; i < collection.length; i++) {
+    const number = collection[i];
+    counter[number] = (counter[number] || 0) + 1;
+
+    if (counter[number] === maxCount) {
+      maxes.add(number);
+    } else if (counter[number] > maxCount) {
```
justinmchase, line 32: I would try to reduce the calls to `counter[number]` to exactly one read and exactly one write, for performance reasons. You should be able to just read it into a variable, and reference that in the block below.

LionC:
What are we doing with this? I am working on a proposal for a rich default benchmark reporter and would actually need some of those functions, including percentiles.

I can copy the functions for now, but they should really be located in their own module. If no one wants to keep working on this, I would also be happy to push this to a first releasable version.

mdx97:
> What are we doing with this? I am working on a proposal for a rich default benchmark reporter and would actually need some of those functions, including percentiles.
> 
> I can copy the functions for now, but they should really be located in their own module. If no one wants to keep working on this, I would also be happy to push this to a first releasable version.

You are free to take over. I am a bit too busy to see this through at the moment. :)

LionC:
@mdx97 I will fetch your branch and continue work from there, so you should automatically be co-authored in the end. I think that should be a new PR, can you close this one?

@kt3k I will open a new PR once I am happy with the implementation

bartlomieju:
I think we should still pursue landing this PR as it seems like a valuable addition to the standard library. That said, at the moment this PR is not close to landing and became quite stale; so I'm going to close this PR for now. We need someone to champion the addition of this module.  @mdx97 thank you for your work, if you are still interested in getting this landed please open a new PR, @LionC are you still interested in working on the parts of it?

LionC:
> @LionC are you still interested in working on the parts of it?

I am and there is a branch on my machine with work on it, it just lost the priority battle to middleware and logs. If anyone wants to work on this right now, feel free to hmu and I will share what I have, otherwise I will continue on it as soon as my priority stack has reached it again :-)

PR #1113, opened 2021-08-08
fix(node/events): Optimize listener management

uki00a:
Part of #911

In this PR, `EventEmitter._events` has been changed as follows:

**before:**

```ts
  private _events: Map<
    string | symbol,
    Array<GenericFunction | WrappedFunction>
  >;
```

**after:**

```ts
  private _events: Record<
    string | symbol,
    (
      | (Array<GenericFunction | WrappedFunction>)
      | GenericFunction
      | WrappedFunction
    ) & { warned?: boolean }
  >;
```

This change is intended to avoid allocating unnecessary arrays and to pass some native node tests.

To verify the change, this PR also enables the following tests:

- `test-event-emitter-check-listener-leaks.js`
- `test-event-emitter-listeners-side-effects.js`
- `test-event-emitter-remove-listeners.js`
- `test-event-emitter-set-max-listeners-side-effects.js`
- `test-event-emitter-special-event-names.js`


node/_tools/suites/parallel/test-event-emitter-check-listener-leaks.js
```diff
@@ -0,0 +1,110 @@
...
+// default
+{
+  const e = new events.EventEmitter();
+
+  for (let i = 0; i < 10; i++) {
+    e.on('default', common.mustNotCall());
+  }
+  assert.ok(!e._events.default.hasOwnProperty('warned'));
+  e.on('default', common.mustNotCall());
+  assert.ok(e._events.default.warned);
```
kt3k, line 44: hmm node.js considers `._events` as a part of public API...

kt3k (approved):
@uki00a LGTM. Nice work! Thanks!

PR #1116, opened 2021-08-08
cleanup (collections): Clean up examples, types and file structure

LionC:
This does several tedious clean up tasks in the `collections` module:

- Unify all examples
    - Use `ts` tag for markdown code
    - Use `assertEquals` instead of `console.assert` to have actually working examples
- Move tests into their own subfolder, as the actual folder is being browsed by users and got
quite crowded
- Flatten argument types. Even though `Predicate` and `Selector` read nicely, their meaning
is not immediately clear and LSP hovers do not show their content. The types in the module are
mostly straightforward anyways

### DIFF
--- a/collections/group_by.ts
+++ b/collections/group_by.ts
@@ -10,6 +10,7 @@ import { Grouping, Selector } from "./types.ts";
  *
  * ```ts
  * import { groupBy } from "./group_by.ts";
+ * import { assertEquals } from "../testing/asserts.ts";
  *
  * type Person = {
  *   name: string;
@@ -22,7 +23,7 @@ import { Grouping, Selector } from "./types.ts";
  * ];
  * const peopleByFirstLetter = groupBy(people, it => it.name.charAt(0))
  *
- * console.assert(peopleByFirstLetter === {
+ * assertEquals(peopleByFirstLetter, {
  *     'A': [ { name: 'Anna' }, { name: 'Arnold' } ],
  *     'K': [ { name: 'Kim' } ],
  * })
--- a/collections/group_by.ts
+++ b/collections/group_by.ts
@@ -1,7 +1,5 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
-import { Grouping, Selector } from "./types.ts";
-
 /**
  * Applies the given selector to each element in the given array, returning a Record containing the results as keys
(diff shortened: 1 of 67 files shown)


kt3k:
@LionC Thank you for the clean-ups!

I mostly agree with these changes, but I'm not in favor of moving test files into their own folder. Other std modules put the test files next to the implementation, and in my view that flat structure is easier for the collaborators to work with because the corresponding implementations and tests are close to each other.

LionC:
> @LionC Thank you for the clean-ups!
> 
> 
> 
> I mostly agree with these changes, but I'm not in favor of moving test files into their own folder. Other std modules put the test files next to the implementation, and in my view that flat structure is easier for the collaborators to work with because the corresponding implementations and tests are close to each other.

The only problem I see with the current structure is that we explicitly allow `import`ing single functions from this folder, which means users of the module need to navigate it.

I work on the module a lot right now, and even I have a hard time to reorientate myself at a glance when opening the folder.ä, there is just a lot of files with similiar names directly next to each other, some being tests and some being implementations without a consistent ordering pattern (because of the default alphabetical sorting everywhere).

My opinion on this is not super steong, but I can say that working on the module and scanning the folders at a glance is already a lot nicer after moving the test files. I will leave final judgement to you and the others though :-)

caspervonb:
>  I will leave final judgement to you and the others though :-)

Test module next to the module being has been the way since the start and is the convention everywhere else so let's not randomly break from that for `collections`.

LionC:
Done, please re-review @kt3k

collections/group_by.ts
```diff
@@ -22,16 +21,16 @@ import { Grouping, Selector } from "./types.ts";
...
  *     'K': [ { name: 'Kim' } ],
  * })
  * ```
  */
 export function groupBy<T>(
   array: Array<T>,
-  selector: Selector<T, string>,
-): Grouping<T> {
+  selector: (el: T) => string,
+): { [key: string]: Array<T> } {
```
zhangyuannie, line 33: Minor nitpick, we might want to use `Record` here since that is what's used throughout `collections`.

LionC: You are right

LionC: Done.

kt3k (approved):
@LionC Thanks! LGTM

PR #1117, opened 2021-08-10
0.104.0

bartlomieju:


LionC:
Would it make sense to "squash" the `collections` related release notes into one? Some arbitrary functions having their own point and some not based on implementation order feels weird.

Also, is this fixed from now on or can things be merged today and tomorrow be included? There is a bigger clean up PR (affecting docs) that will likely get merged today for `collections`.

lucacasonato (approved):

bartlomieju:
> Would it make sense to "squash" the `collections` related release notes into one? Some arbitrary functions having their own point and some not based on implementation order feels weird.

I squashed "collections" PR into a single point.

> Also, is this fixed from now on or can things be merged today and tomorrow be included? There is a bigger clean up PR (affecting docs) that will likely get merged today for `collections`.

After this PR is merged all commits merged afterwards will end up in 0.105.0

ry (approved):
LGTM

PR #1118, opened 2021-08-10
feat(jose): partial jws and jwt implementation

timreichen:
Note that this is only a draft, jwe is missing and therefore jws encryption and decryption is not present atm. 
@panva Would be glad if you could take a look and give some feedback.

ref: https://github.com/denoland/deno/pull/7991#issuecomment-715598404


panva (commented):
I am not in a position to review the API design decision, method naming, or code structure. That being said here are a few points to focus on.

The Flattened and General serializations are incorrect. When it comes to General and Flattened JSON serializations the JWS algorithm may be sourced from either protected or unprotected header and the library must check the header parameter names in the two sources are disjoint.

bartlomieju:
@timreichen I think this is a cool addition to the standard library, but I have a few concerns:
- the code is quite stale at the moment, could you rebase it?
- are there any missing webcrypto APIs at the moment that would prevent it from working?
- the name is a bit misleading, if we were to add it, it should probably be named `jwt`

timreichen:
@bartlomieju sorry for the late answer.
> are there any missing webcrypto APIs at the moment that would prevent it from working?

Idk. I am not at all familiar with `jwe` which is the reason why that implementation is missing. I would need the help of someone who can champion that.

> the name is a bit misleading, if we were to add it, it should probably be named jwt

The reason why it is not called `jwt` is because jwt uses `jws` for signature or `jwe` for encryption which are two different specs and thus would make sense to have access to them separately imo. But instead of submodules I could imagine three different std modules (jws, jwe, jwt) where jwt is dependent on the other two.

However  I have not much time atm to take that on. But whoever wants to tackle that in my place is welcome.

panva:
With the release of Deno v1.18 https://deno.land/x/jose can now be considered stable and safe to use. I am aware it is not a `std` library but it is maintained, well tested and already put to use.

bartlomieju:
Thanks for the heads up @panva. @timreichen I think we should recommend using `jose` instead of trying to maintain it in tree. WDYT?

timreichen:
I still think it would make sense to have it in std which was the whole point of this pr since there are some 3rd party libs out there atm including jose.
But this isn't a high priority to me and I am not experienced enough on that topic to advance it currently. So for the time being we can recommend 3rd parties. But I would love to see it being part of std eventually.

bartlomieju:
I'm still not sure if we should add this module to standard library, especially with incoming improvements for discoverability in `deno.land/x/`. Having to periodically copy-paste code from `jose` seems less than ideal and poses significant maintenance burden.

PR #1119, opened 2021-08-11
feat(collections): compile time guarantee on pure functions

zhangyuannie:
#### Before :confused: 
```ts
const arr = Object.freeze([1, 2, 3, 4, 5]);
const a = findLast(arr, (e) => e === 1);
                   ^^^
// Argument of type 'readonly number[]' is not assignable to parameter of type 'number[]'.
//   The type 'readonly number[]' is 'readonly' and cannot be assigned to the mutable type 'number[]'.
```
#### After :smiley: 
```ts
const arr = Object.freeze([1, 2, 3, 4, 5]);
const a = findLast(arr, (e) => e === 1);
```
On top of that, the compiler will also help `std` maintainers to not mutate the parameters by mistake to a certain degree.
```ts
export function foo(array: readonly number[]) {
  array.push(1);
        ^^^^
  // Property 'push' does not exist on type 'readonly T[]'.

  array[2] = 3;
  ^^^^^^^^
  // Index signature in type 'readonly number[]' only permits reading.

  ...
}
```

### DIFF
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -27,8 +27,8 @@ import { filterInPlace } from "./_utils.ts";
 export function deepMerge<
   T extends Record<PropertyKey, unknown>,
 >(
-  record: Readonly<Partial<T>>,
-  other: Readonly<Partial<T>>,
+  record: Partial<Readonly<T>>,
+  other: Partial<Readonly<T>>,
   options?: Readonly<DeepMergeOptions>,
 ): T;
 
@@ -93,14 +93,14 @@ export function deepMerge<
 }
 
 function mergeObjects(
-  left: Readonly<NonNullable<unknown>>,
-  right: Readonly<NonNullable<unknown>>,
-  options: DeepMergeOptions = {
+  left: Readonly<NonNullable<object>>,
+  right: Readonly<NonNullable<object>>,
+  options: Readonly<DeepMergeOptions> = {
     arrays: "merge",
     sets: "merge",
     maps: "merge",
   },
-): NonNullable<unknown> {
+): Readonly<NonNullable<object>> {
   // Recursively merge mergeable objects
   if (isMergeable(left) && isMergeable(right)) {
     return deepMerge(left, right);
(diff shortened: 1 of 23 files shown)


LionC:
This has actually been on my to do list - thank you!

I will do a review later today.

collections/deep_merge.ts
```diff
@@ -93,14 +93,14 @@ export function deepMerge<
 }
 
 function mergeObjects(
-  left: NonNullable<object>,
-  right: NonNullable<object>,
+  left: Readonly<NonNullable<unknown>>,
```
LionC, line 96: Why has this been changed to `unknown`?

collections/deep_merge.ts
```diff
@@ -93,14 +93,14 @@ export function deepMerge<
...
+  right: Readonly<NonNullable<unknown>>,
   options: DeepMergeOptions = {
     arrays: "merge",
     sets: "merge",
     maps: "merge",
   },
-): NonNullable<object> {
+): NonNullable<unknown> {
```
LionC, line 103: Same here

collections/zip.ts
```diff
@@ -22,15 +22,15 @@
...
+  array: readonly T[],
+  withArray: readonly U[],
+): [T, U][] {
   const returnLength = Math.min(
     array.length,
     withArray.length,
   );
 
-  const ret: Array<[T, U]> = [];
+  const ret: [T, U][] = [];
```
LionC, line 33: We should use the calculated `returnLength` here (I missed that on my initial implementation):

```suggestion
  const ret = new Array<[T, U]>(returnLength);
```

LionC (commented):
Looks good except the one minor thing (which is not your fault but could be easily fixed here) and one question.

Thanks for doing the tedious work to go through this, much appreciated!

collections/deep_merge.ts
zhangyuannie: Good catch!
Because It was incompatible with the signature of `deepMerge` (`'object' is not assignable to parameter of type 'Readonly<Record<PropertyKey, unknown>>'`)
The typing issue is now fixed by making `isMergeable` a type guard of `Record<PropertyKey, unknown>` instead. This aligns better with what `isMergeable` is trying to accomplish and match the signature of `deepMerge`.

LionC:
LGTM! Pinging @kt3k for final approval :-)

kt3k (approved):
@zhangyuannie LGTM. Thanks! Nice improvement :+1: 

And thanks @LionC for reviewing!

PR #1121, opened 2021-08-11
fix: type check examples in README files

ry:
Fixes https://github.com/denoland/deno_std/issues/1120

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
           deno-version: ${{ matrix.deno }}
 
       - name: Run tests
-        run: deno test --doc --unstable --allow-all --coverage=./cov
+        run: deno test --doc --unstable --allow-all --coverage=./cov --import-map=test_import_map.json
 
       - name: Type check browser compatible modules
         shell: bash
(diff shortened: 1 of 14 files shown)


LionC:
Nitpick: You added imports to the examples in `collections/README.md`. I would personally prefer if we would import from the function file directly instead of `mod.ts` (e.g. `import { associateBy } from "./associate_by.ts"`). Examples often get copied and modified by users and I would like to encourage people to cherry pick modules instead of downloading everything. This would also be consistent with all JSDoc examples in `collections`.

It is a bit tedious to replace all that by hand - I would be happy to do it tonight if you want.

bartlomieju:
> Nitpick: You added imports to the examples in `collections/README.md`. I would personally prefer if we would import from the function file directly instead of `mod.ts` (e.g. `import { associateBy } from "./associate_by.ts"`). Examples often get copied and modified by users and thus influence how people use it. I would like to encourage users to cherry pick modules instead of downloading everything. This would also be consistent with all JSDoc examples in `collections`.
> 
> It is a bit tedious to replace all that by hand - I would be happy to do it tonight if you want.
> 
> Thank you for fixing the `README` either way!

If that's the case, what's the point of having `mod.ts` in the first place?

bartlomieju (approved):

PR #1123, opened 2021-08-11
Use os util on entire of codebase

behnammodi:
we had os util but in some where of code we didn't use it

### DIFF
--- a/_util/os.ts
+++ b/_util/os.ts
@@ -19,4 +19,4 @@ export const osType: OSType = (() => {
   return "linux";
 })();
 
-export const isWindows: boolean = osType === "windows";
+export const isWindows = osType === "windows";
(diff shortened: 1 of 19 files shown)


_util/os.ts
```diff
@@ -17,4 +19,4 @@ export const osType = (() => {
   return "linux";
 })();
 
-export const isWindows = osType === "windows";
+export const isWindows: boolean = osType === "windows";
```
kitsonk, line 22: This is a trivial type annotation and not required.

behnammodi: Exactly, I will revert this line

kt3k (approved):
LGTM

behnammodi:
Can I join as member to denoland?

behnammodi:
Can I join as a member in denoland? :)

bartlomieju:
> Can I join as a member in denoland? :)

Sorry, that's not possible, only employees of Denoland have "member" access.

PR #1124, opened 2021-08-11
fix(ws): add output type

behnammodi:
I think this is missed


bartlomieju:
Thanks for opening PR, but this is on purpose. Primitive types should be omitted and `Promise<void>` is one of them.

behnammodi:
@bartlomieju  All right.

PR #1125, opened 2021-08-11
fix(wasi): add output type 

behnammodi:


caspervonb (commented):
Thanks for the pull request! but...
These things are inferred just fine, nothing being explicitly annotated here is public or even exported so my question is, why?

behnammodi:
> Thanks for the pull request! but...
> These things are inferred just fine, nothing being explicitly annotated here is public or even exported so my question is, why?

Why not :) we specify output and input types everywhere of the codebase, so what is wrong with this?

kt3k:
See this issue, for example, https://github.com/denoland/deno/issues/8832

We generally prefer avoiding unnecessary type annotations when typescript infers them correctly.

behnammodi:
@kt3k I read your link, I convinced, so I'll close this PR if you agree

PR #1128, opened 2021-08-12
feat(http): implement native HTTP server APIs

cmorten:
Fixes #1034 

Implements native HTTP server APIs within `std/http`, following the proposal set out in the RFC with some minor deviations.

**References**

- [RFC: std/http built on the native HTTP server](https://github.com/denoland/deno_std/discussions/1034)
- [native_http userland module](https://github.com/cmorten/native_http)
- [prompt by @bartlomieju to raise the PR](https://discord.com/channels/684898665143206084/712010403302866974/875475301109477446)


http/native_server.ts
```diff
@@ -0,0 +1,369 @@
...
+export async function listenAndServeTls(
+  addr: AddrOptions,
+  certFile: string,
+  keyFile: string,
+  handler: HttpHandler,
+): Promise<void> {
+  let listenOptions;
+
+  if (typeof addr === "undefined" || addr === null) {
+    listenOptions = { port: 80 };
```
cmorten, line 351:
```suggestion
    listenOptions = { port: 443 };
```

http/native_server.ts
```diff
@@ -0,0 +1,369 @@
...
+  } else if (typeof addr === "number") {
+    listenOptions = { port: addr };
+  } else if (typeof addr === "string") {
+    listenOptions = _parseAddrFromStr(addr);
```
cmorten, line 355: This returns a default port of `80` whereas suspect better default for tls is `443`.

cmorten: c1ba097

http/native_server.ts
cmorten: c1ba097

http/native_server.ts
```diff
@@ -0,0 +1,388 @@
...
+      }
+    }
+
+    this.#httpConnections.clear();
+  }
+
+  /**
+   * Yields all HTTP requests on a single TCP connection.
+   *
+   * @param {HttpConn} httpConn The HTTP connection to yield requests from.
```
cmorten, line 124: Add jsdoc for conn param

http/native_server.ts
```diff
@@ -0,0 +1,388 @@
...
+    try {
+      conn = await this.listener.accept();
+    } catch (error) {
+      if (
+        // The listener is closed
+        error instanceof Deno.errors.BadResource ||
+        // TLS handshake errors
+        error instanceof Deno.errors.InvalidData ||
+        error instanceof Deno.errors.UnexpectedEof ||
+        error instanceof Deno.errors.ConnectionReset
```
cmorten, line 204: TODO: is this the full list?

http/native_server.ts
cmorten: ef987c4

http/native_server.ts
cmorten: b6b632a

http/native_server.ts
```diff
@@ -0,0 +1,360 @@
...
+  /**
+   * Get the connection info.
+   */
+  get connInfo(): ConnInfo {
+    return this.#connInfo;
+  }
+
+  /** Determine whether the response has completed. */
+  get done(): Promise<void> {
+    return this.#responsePromise;
```
cmorten, line 67:
```suggestion
    return this.#responsePromise.catch(() => {}):
```

http/native_server.ts
```diff
@@ -0,0 +1,360 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import type { HTTPOptions, HTTPSOptions } from "./server.ts";
+import { _parseAddrFromStr } from "./server.ts";
```
cmorten, line 3: Remove

cmorten: 24fac9c

cmorten:
While upping test coverage, a test copy pasta resulted in the following panic:

```console
$ RUST_BACKTRACE=1 deno test -A ./http/native_server_test.ts
running 21 tests from deno_std/http/native_server_test.ts
...
thread '<unnamed>' panicked at 'called `Option::unwrap()` on a `None` value', extensions/http/lib.rs:204:12
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: <futures_util::future::future::Map<Fut,F> as core::future::future::Future>::poll
   4: <futures_util::future::future::Inspect<Fut,F> as core::future::future::Future>::poll
   5: futures_util::stream::stream::StreamExt::poll_next_unpin
   6: deno_core::runtime::JsRuntime::poll_event_loop
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   8: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Any { .. }', cli/tools/test_runner.rs:601:26
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: tokio::runtime::task::raw::poll
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
test serve should construct a new HTTPS Server capable of handling requests and gracefully closing afterwards ...
test result: FAILED. 20 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (595ms)
```

Test to reproduce:

```ts
Deno.test(`serve should construct a new HTTPS Server capable of handling requests and gracefully closing afterwards`, async () => {
  const listenTlsOptions = {
    hostname: "localhost",
    port: 4505,
    certFile: join(testdataDir, "tls/localhost.crt"),
    keyFile: join(testdataDir, "tls/localhost.key"),
  };
  const listener = Deno.listenTls(listenTlsOptions);
  const url = `https://${listenTlsOptions.hostname}:${listenTlsOptions.port}`;
  const method = "GET";
  const expectedStatus = 418;
  const abortController = new AbortController();

  const servePromise = serve(listener, (request, connInfo) => {
    const remotePort = (connInfo.localAddr as Deno.NetAddr).port;
    return new Response(
      `${request.method}: ${remotePort} - Hello Deno from serve!`,
      {
        status: expectedStatus,
      },
    );
  }, { signal: abortController.signal });

  try {
    const response = await fetch(url, { method });
    const body = await response.text();

    assertEquals(
      body,
      `${method}: ${listenTlsOptions.port} - Hello Deno from serve!`,
    );
    assertEquals(response.status, expectedStatus);
  } catch (_) {
    unreachable();
  } finally {
    abortController.abort();
    await servePromise;
  }
});
```

AaronO:
@cmorten That bug was fixed by https://github.com/denoland/deno/pull/11674 (merged in `main`), should be shipped in a release Monday or Tuesday.

cmorten:
@AaronO fantastic 💯

http/native_server.ts
```diff
@@ -0,0 +1,358 @@
...
+  if (init?.signal) {
+    init.signal.onabort = () => {
+      server.close();
+    };
+  }
+
+  for await (const requestEvent of server) {
+    requestEvent.respondWith(
+      handler(requestEvent.request, requestEvent.connInfo),
+    );
```
cmorten, line 293: Main open question I have is are we ok with this potentially throwing and not catching (so iterator loop ends etc.)?

This is same for the existing server API implementations (REF: https://github.com/denoland/deno_std/blob/main/http/server.ts#L327) so perhaps is ok under the assumption that anyone wanting to have fuller control over their request loop would construct a server themselves

```ts
import { Server } from "https://deno.land/std@$STD_VERSION/http/native_server.ts";

// ... rest of code

const server = new Server(listener);

for await (const requestEvent of server) {
  // ... consumer handles the requestEvent with additional error handling etc.
}
```

Given the RFC proposal of an http handler passed in I don't think we have much of a choice but to be opionated on behalf of the consumer. Alternatives include us `.catch(() => {})` swallowing all errors, or perhaps extending the `ServerInit` or sim. to accept an `onRequestError(error: Error): Promise<void>`. Perhaps other options as well?

cmorten: If following precedent of https://pkg.go.dev/net/http#Handler then perhaps we should recover unless the error is of a certain instance

cmorten:
Having just read go’s http std lib interface/implementation I think I can do a better job here - mostly around the Server class interface and how it could be written for nice extensions in future. Going to put this back in draft.

Apologies if anyone has started reviewing - the core of the listener/connection handling will be unchanged (basically lifted from the current std/http implementation anyway).

cmorten:
Ok, think that's it...

http/native_server.ts
```diff
@@ -0,0 +1,579 @@
...
+    const listener = Deno.listen({
+      ...listenOptions,
+      transport: "tcp",
+    });
+
+    try {
+      return await this.serve(listener);
+    } finally {
+      try {
+        listener.close();
```
cmorten, line 216: Not required as done by serve

http/native_server.ts
```diff
@@ -0,0 +1,579 @@
...
+      transport: "tcp",
+      // ALPN protocol support not yet stable.
+      // alpnProtocols: ["h2", "http/1.1"],
+    });
+
+    try {
+      return await this.serve(listener);
+    } finally {
+      try {
+        listener.close();
```
cmorten, line 256: Not required as done by serve

http/native_server.ts
```diff
@@ -0,0 +1,579 @@
...
+      throw ERROR_SERVER_CLOSED;
+    }
+
+    this.trackListener(listener);
+
+    const mux = this.serveMux(listener);
+
+    try {
+      for await (const requestEvent of mux) {
+        requestEvent.respondWith(this.handler(
```
cmorten, line 173: Should we swallow errors from here as issues with a particular request, close that connection, but otherwise not kill the process? Seems to be what go's http lib does?

cmorten: Have gone for this - 2aac74b.

http/native_server.ts
cmorten: 2aac74b

http/native_server.ts
cmorten: 2aac74b

http/native_server.ts
```diff
@@ -0,0 +1,586 @@
...
+
+    try {
+      httpConn.close();
+    } catch (error) {
+      if (!(error instanceof Deno.errors.BadResource)) {
+        throw error;
+      }
+    }
+
+    if (requestEvent) {
```
asos-craigmorten, line 373: TODO: Should be fixed in Deno `1.13.1`

REF: https://github.com/denoland/deno/issues/11595#issuecomment-899414934

cmorten: Wasn’t resolved, raised https://github.com/denoland/deno/issues/11743

LionC:
As long as this does not block inteoducing some middleware-ish concept in the future without breaking compatibility, I am all for it :-)

cmorten:
> As long as this does not block inteoducing some middleware-ish concept in the future without breaking compatibility, I am all for it :-)

It’s mirroring go sufficiently that imagine should naturally extend to layering middleware over it. The RFC deviates from go in the `handler` provided to the `Server`, but I don’t think that should block middleware concepts.

Here we have a `handler` function expecting a `Request` and returning a `Response`, in go you have a `Handler` interface with a `ServeHTTP` method that fulfils the same purpose as our `handler` func.

How golang introduces middleware is through `ServeMux` which implements the `Handler` interface such that it’s `ServeHTTP` method calls into the routing / middlewares registered through additional `Handle` and `HandleFunc` methods.

In using a function as our `handler` we won’t be able to just start passing new `Middleware` classes in à la go, but can foresee that you would construct such a thing, register your route handlers, and then pass it's `handler` in the `Server` construction. E.g. some pseudocode…

```ts
// Probably a better name than this?
interface Middleware {
  handler: HttpHandler;
}

const middleware = new Middleware();
// register routes on the middleware class using various methods

const Server = new Server({ addr, handler: middleware.handler });
```

I reckon most JS folks (at least entry level) will get on with needing a `handler` function better than satisfying a `Hander` interface, so hopefully a reasonable design that we won’t need to change? How to integrate with future middleware classes (or otherwise) would be easily solved through docs + small examples.

LionC:
Yeah I think as long as there is a way to do it in `std/http` without breaking, we are fine and that seems to be the case here .I also think a `Request -> Response` function is a nice design and also works well for middleware (being able to "modify" on both ways).

With std having a logger, parsers and maybe validation in the future, not having std middleware at some point seems weird, as those are really common use cases. I am glad we are not ruling that out with the design :-)

cmorten:
One gotcha that’s occurred to me with this current implementation, and the deprecated of the legacy server, is that we have no websocket support equivalent.

I suspect the natural route is to simply follow the precedent set by @kitsonk in Oak (REF: https://github.com/oakserver/oak/blob/main/http_server_native.ts#L153), but we’d be delving into unstable APIs, and there appear to be other caveats as well (REF: https://github.com/denoland/deno/issues/11694)

It would be wrong to deprecate the old server APIs if we can’t support feature parity.

bartlomieju:
@kitsonk your review here would be highly appreciated

cmorten:
Sample of some performance testing results after e9f0fac. Found this std implementation typically has <0.5% performance drop c.w. the current Deno benchmark, sometimes outperforming it - effectively the difference appears to be less than the variation see for either implementation between runs.

**This PR**

Setup:

```console
$ deno run --allow-net --reload ./http/bench.ts
Check file:///Users/craigmorten/git/cmorten/deno_std/http/bench.ts
Server listening on 127.0.0.1:4500
```

Test Runs:

```console
$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.59ms    3.14ms 137.91ms   97.06%
    Req/Sec    14.31k     1.83k   18.97k    73.25%
  Latency Distribution
     50%    3.39ms
     75%    4.02ms
     90%    4.88ms
     99%    9.38ms
  569908 requests in 20.01s, 47.29MB read
Requests/sec:  28474.30
Transfer/sec:      2.36MB

$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.45ms    1.99ms  88.63ms   92.00%
    Req/Sec    14.61k     1.95k   18.83k    74.75%
  Latency Distribution
     50%    3.31ms
     75%    3.93ms
     90%    4.73ms
     99%    9.26ms
  582140 requests in 20.03s, 48.30MB read
Requests/sec:  29060.21
Transfer/sec:      2.41MB

$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.44ms    1.45ms  48.05ms   78.55%
    Req/Sec    14.55k     1.52k   18.07k    77.00%
  Latency Distribution
     50%    3.37ms
     75%    4.00ms
     90%    4.55ms
     99%    9.27ms
  579466 requests in 20.02s, 48.08MB read
Requests/sec:  28945.14
Transfer/sec:      2.40MB
```

**Deno Benchmark**

Setup:

```console
$ deno run --allow-net --reload https://raw.githubusercontent.com/denoland/deno/main/cli/bench/deno_http_native.js
Download https://raw.githubusercontent.com/denoland/deno/main/cli/bench/deno_http_native.js
Server listening on 127.0.0.1:4500
```

Test Runs:

```console
$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.66ms    2.96ms 134.83ms   96.84%
    Req/Sec    14.02k     1.95k   16.61k    75.25%
  Latency Distribution
     50%    3.18ms
     75%    3.73ms
     90%    4.75ms
     99%    9.59ms
  558346 requests in 20.02s, 46.33MB read
Requests/sec:  27894.85
Transfer/sec:      2.31MB

$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.45ms    1.36ms  68.27ms   91.62%
    Req/Sec    14.54k     1.51k   16.61k    82.50%
  Latency Distribution
     50%    3.14ms
     75%    3.50ms
     90%    4.52ms
     99%    9.06ms
  578582 requests in 20.01s, 48.00MB read
Requests/sec:  28918.82
Transfer/sec:      2.40MB

$ wrk --d 20s -t 2 -c 100 --latency http://127.0.0.1:4500
Running 20s test @ http://127.0.0.1:4500
  2 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.55ms    1.53ms  73.00ms   92.08%
    Req/Sec    14.17k     1.73k   16.37k    78.00%
  Latency Distribution
     50%    3.17ms
     75%    3.61ms
     90%    4.70ms
     99%    9.50ms
  563954 requests in 20.01s, 46.79MB read
Requests/sec:  28184.26
Transfer/sec:      2.34MB
```

ry (approved):
LGTM - nice work @cmorten thanks for seeing this through

PR #1130, opened 2021-08-13
feat(async): Add abort signal to delay

### LINKED ISSUE #1129: `delay` is not abortable
Once you begin a delay there is no way to abort it presently

**Describe the solution you'd like** A clear and concise description of what you
```sh
const abort = new AbortController()
await delay(30 * 1000, abort.signal)
```

The alternative is for each project to wrap the delay promise with your own abort logic and use `await Promise.race()` its ugly, it really should be abortable in the std library.

**Additional context** 
`N/A`

justinmchase:
This is a non-breaking change which adds an optional `signal?: AbortSignal` to the delay function.

3 new tests were added which test that:

1. The abort stops the timeout and immediately resolves the delay promise
2. Not aborting the signal behaves as normal
3. Aborting after the delay has already completed does not result in an error

Fixes #1129

### DIFF
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -1,6 +1,12 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+export interface DelayOptions {
+  signal?: AbortSignal;
+}
+
 /* Resolves after the given number of milliseconds. */
-export function delay(ms: number, signal?: AbortSignal): Promise<void> {
+export function delay(ms: number, options: DelayOptions = {}): Promise<void> {
+  const { signal } = options;
   return new Promise((resolve, reject): void => {
     const abort = () => {
       clearTimeout(i);
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -1,9 +1,13 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 /* Resolves after the given number of milliseconds. */
-export function delay(ms: number): Promise<void> {
-  return new Promise((res): number =>
-    setTimeout((): void => {
+export function delay(ms: number, signal?: AbortSignal): Promise<void> {
+  return new Promise((res): void => {
+    const done = () => {
+      clearTimeout(i);
+      signal?.removeEventListener("abort", done);
       res();
-    }, ms)
-  );
+    };
+    const i = setTimeout(done, ms);
+    signal?.addEventListener("abort", done, { once: true });
+  });
 }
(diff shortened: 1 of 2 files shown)


async/delay.ts
```diff
@@ -1,9 +1,13 @@
...
+  return new Promise((res): void => {
+    const done = () => {
+      clearTimeout(i);
+      signal?.removeEventListener("abort", done);
       res();
-    }, ms)
-  );
+    };
+    const i = setTimeout(done, ms);
+    signal?.addEventListener("abort", done, { once: true });
```
behnammodi, line 11: I think we new reject Promise when `abort` happening not call done

justinmchase: Ok makes sense, I see `fetch` throws an `AbortError` so it does seem like it should be consistent with that, thanks.

behnammodi: Glad to help

justinmchase: Strangely it seems like AbortError isn't an actual error kind but I see this in deno itself:

https://github.com/denoland/deno/blob/main/extensions/fetch/26_fetch.js#L473

So I altered it to throw like that.

async/delay_test.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import { delay } from "./delay.ts";
-import { assert } from "../testing/asserts.ts";
+import { assert, assertThrowsAsync } from "../testing/asserts.ts";
```
bartlomieju, line 3: Prefer `assertRejects` which was introduced in #1101

async/delay.ts
```diff
@@ -1,9 +1,16 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 /* Resolves after the given number of milliseconds. */
-export function delay(ms: number): Promise<void> {
-  return new Promise((res): number =>
-    setTimeout((): void => {
-      res();
-    }, ms)
-  );
+export function delay(ms: number, signal?: AbortSignal): Promise<void> {
```
kt3k, line 3: I think the 2nd argument should be an object of the shape of `options?: { signal?: AbortSignal }`. This design makes the future addition of other optional parameters hard. We have a style guideline for this rule https://deno.land/manual/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object

justinmchase: Good point, will do.

kt3k (approved):
@justinmchase Thank you for your contribution! LGTM

This should be useful when you want to quickly clear all the resources involved, especially during testing. Nice feature!

PR #1131, opened 2021-08-14
fix(collections): intersect does not handle duplicate values in head properly

zhangyuannie:
One would expect `intersection` to be commutative.


kt3k (approved):
@zhangyuannie LGTM. Thanks!

PR #1132, opened 2021-08-14
refactor(collections): permutations optimisation

zhangyuannie:
Remove function closure since it doesn't improve readability.

### DIFF
--- a/collections/permutations.ts
+++ b/collections/permutations.ts
@@ -39,13 +39,9 @@ export function permutations<T>(inputArray: readonly T[]): T[][] {
   while (i < k) {
     if (c[i] < i) {
       if (i % 2 === 0) {
-        const swapBuffer = array[0];
-        array[0] = array[i];
-        array[i] = swapBuffer;
+        [array[0], array[i]] = [array[i], array[0]];
       } else {
-        const swapBuffer = array[c[i]];
-        array[c[i]] = array[i];
-        array[i] = swapBuffer;
+        [array[c[i]], array[i]] = [array[i], array[c[i]]];
       }
 
       ret.push([...array]);
--- a/collections/permutations.ts
+++ b/collections/permutations.ts
@@ -20,45 +20,43 @@
  * ])
  * ```
  */
-export function permutations<T>(array: readonly T[]): T[][] {
+export function permutations<T>(inputArray: readonly T[]): T[][] {
   const ret: T[][] = [];
+  const k = inputArray.length;
 
-  if (array.length === 0) {
+  if (k === 0) {
     return ret;
   }
 
-  // Heap Algorithm
-  function heapPermutations(k: number, array: T[]) {
-    const c = new Array<number>(k).fill(0);
+  // Heap's Algorithm
+  const array = [...inputArray];
+  const c = new Array<number>(k).fill(0);
 
-    ret.push([...array]);
+  ret.push([...array]);
 
-    let i = 1;
+  let i = 1;
 
-    while (i < k) {
-      if (c[i] < i) {
-        if (i % 2 === 0) {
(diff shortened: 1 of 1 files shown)


LionC:
Thanks! I will take a look later :-)

collections/permutations.ts
```diff
@@ -20,45 +20,43 @@
...
-  if (array.length === 0) {
+  if (k === 0) {
     return ret;
   }
 
-  // Heap Algorithm
-  function heapPermutations(k: number, array: T[]) {
-    const c = new Array<number>(k).fill(0);
+  // Heap's Algorithm
+  const array = [...inputArray];
```
LionC, line 32: This should probably be

```suggestion
  const array = Array.from(inputArray);
```

instead - spreading is not free, so I try to avoid spreading arbitrarily large arrays.

collections/permutations.ts
```diff
@@ -20,45 +20,43 @@
...
+      if (i % 2 === 0) {
+        const swapBuffer = array[0];
+        array[0] = array[i];
+        array[i] = swapBuffer;
       } else {
-        c[i] = 0;
-        i += 1;
+        const swapBuffer = array[c[i]];
+        array[c[i]] = array[i];
+        array[i] = swapBuffer;
```
LionC, line 48: Missed this on the my implementation, but this could be

```suggestion
        [array[c[i]], array[i]] = [array[i], array[c[i]]];
```

instead

LionC (changes requested):
Looks good and makes sense to inline the whole thing. I just added two minor performance comments (with one being from the previous implementation).

collections/permutations.ts
zhangyuannie: Do you have a source for your claim?
1. AFAIK it is the reverse, spreading is almost always faster than `Array.from`. (and in general, things built into the syntax will always be faster than a prototype lookup if they do the same thing)
2. This is an O(n!) operation, you should consider another language/tool if you have more than a dozen elements. (it actually core dump here with 20 elements :sweat_smile:)
3. I am fine changing it to `Array.prototype.slice` through if we are into this kind of optimisation however `slice` preserves holes (:confused:) and works only on array.

collections/permutations.ts
zhangyuannie: I think we could keep the current implementation since we are into performance and memory optimisation.

collections/permutations.ts
LionC: That was a fun read indeed :D

The post suggests that the spread optimizations apply to Array.from as well. So I gueds it does bot really matter and I do not care either way^^

collections/permutations.ts
LionC: I thought the destructuring syntax gives v8 the chance to optimize (as the expression on the right can not be referenced outside of the statement), while it would need a very specific optimization for the buffer version to be faster.

I could not find details in any of the two with a quick search though...

zhangyuannie: I have updated with array destructuring.

My quick search suggests that it will likely be good enough nowadays, and my quick *likely flawed* benchmark with 10 elements suggest that the performance overhead is around 3% after a thousand iterations.

I am fine either way since `permutations` is already a slow O(n!) operation :)

Some fun readings: 
https://v8.dev/blog/v8-release-68#array-destructuring-improvements 
https://docs.google.com/document/d/1hWb-lQW4NSG9yRpyyiAA_9Ktytd5lypLnVLhPX9vamE/edit#heading=h.j2s8yv21fevz

kt3k (approved):
@zhangyuannie Thank you for the optimizations and researches! LGTM

And thank you @LionC for reviewing!

PR #1133, opened 2021-08-15
refactor(bytes): rename `contains` to `includes` with optional argument `fromIndex`

zhangyuannie:
- Rename `contains` to `includes` as discussed in std-wg.
- Add optional argument `fromIndex` to `includes` since `Array.prototype.includes` has it.

### DIFF
--- a/bytes/README.md
+++ b/bytes/README.md
@@ -106,20 +106,20 @@ concat(
 ); // => returns Uint8Array(8) [ 1, 2, 3, 4, 5, 6, 7, 8 ]
 ```
 
-## contains
+## includes
 
-Checks that a given binary array contains a sequence corresponding to a pattern
+Checks that a given binary array includes a sequence corresponding to a pattern
 array.
 
 ```typescript
-import { contains } from "https://deno.land/std@$STD_VERSION/bytes/mod.ts";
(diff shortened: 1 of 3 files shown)


kt3k:
@zhangyuannie Thank you for your contribution! Most part looks good to me.

Because `contains` existed for a relatively long time (ref https://github.com/denoland/deno/pull/4801), I think it would be safer to add an alias to it with `@deprecated` annotation like:

```
/** @deprecated */
export { includes as contains };
```

zhangyuannie:
> @zhangyuannie Thank you for your contribution! Most part looks good to me.
> 
> Because `contains` existed for a relatively long time (ref [denoland/deno#4801](https://github.com/denoland/deno/pull/4801)), I think it would be safer to add an alias to it with `@deprecated` annotation like:
> 
> ```
> /** @deprecated */
> export { includes as contains };
> ```

Sounds good!

kt3k (approved):
@zhangyuannie LGTM. Thanks!

kt3k:
For future reference, this change is derived from std-wg on 3 Dec 2020 https://github.com/denoland/deno_std/discussions/699#discussioncomment-231367

PR #1135, opened 2021-08-15
fix (crypto) Remove fixed version imports

LionC:
This removes two absolute version imports within the crypto bench, as std should always depend on sibling modules (and be sibling-compatible).

There is another fixed dependency within `node/_tools` on `x/compress` - this also seems a bit odd but harder to fix without some discussion. 

We had a short discussion on Discord about adding a CI step that checks if `std` does not contain any external dependencies. There are ways to do this with the current (unstable) `deno info` JSON output, but that might be another discussion altogether, as running / testing a deno module while allowing only local imports might be a more generally useful case.


kt3k (approved):
@LionC LGTM. Thanks!

kt3k:
`node/_tools` are the internal tools for developing `std/node` test cases, and not a part of std public API. So I think they should be allowed to depend on external modules when necessary.

For reference, there are also tools in `denoland/deno` which depends on external modules. https://github.com/denoland/deno/blob/main/tools/wpt/runner.ts

PR #1137, opened 2021-08-16
BREAKING(log): Class based loggers

timreichen:
coming from https://github.com/denoland/deno_std/pull/719#issuecomment-899361210
# Changes
- replace global state with class based Loggers
- add class based LogLevels
- remove `critical` log level
- add `trace` log level
- allow custom LogLevels
- new default LogLevels
  - `trace`
  - `debug`
  - `info`
  - `warn`
  - `error`


ry:
please rebase

Just glancing at this, but I don't understand the benefit of having the log levels be classes rather than strings?

timreichen:
@ry there are mainly three reasons
- the class contains the name and code, so that no conversion from name to code and vis versa needs to be done.
- log levels can be extended more easily that way. If code and name are separated as an enum and object, creating custom log levels is not really straight forward.
- an invalid log level name cannot be passed: Either log level names are restricted to throw a type error in case of an invalid name but then log level names cannot be extended without manipulating types. Or a separate check has to be done to verify that the name has a correspondent code. This is not necessary if the passed argument contains the code as well.

But that being said, the same could be achieved by having log level objects instead of class instances
```ts
const customLogLevel = { code: 50, name: "fatal" };
```
instead of
```ts
const customLogLevel = new LogLevel("fatal", 50);
```

log/README.md
```diff
@@ -3,342 +3,245 @@
...
-    },
-  },
-});
-
-let logger;
+import {
+  ConsoleHandler,
+  Logger,
+  logLevels,
+} from "https://deno.land/std@0.78.0/log/mod.ts";
```
kt3k, line 10: Please use `$STD_VERSION` for std version number

```suggestion
} from "https://deno.land/std@$STD_VERSION/log/mod.ts";
```

log/README.md
```diff
@@ -3,342 +3,245 @@
...
-Loggers are objects that you interact with. When you use a logger method it
-constructs a `LogRecord` and passes it down to its handlers for output. To
-create custom loggers, specify them in `loggers` when calling `log.setup`.
+```ts
+import {
+  ConsoleHandler,
+  FileHandler,
+  Logger,
+  logLevels,
+} from "https://deno.land/std@0.78.0/log/mod.ts";
```
kt3k, line 41:
```suggestion
} from "https://deno.land/std@$STD_VERSION/log/mod.ts";
```

kt3k:
some references for the context
- the original issue of this PR: https://github.com/denoland/deno/issues/7211
- previous PRs
  - https://github.com/denoland/deno_std/pull/719
  - https://github.com/denoland/deno/pull/8460
  - https://github.com/denoland/deno/pull/7293
- (the first version of log module: https://github.com/denoland/deno_std/pull/33 )

kt3k:
This PR changes quite many aspects of log module. And I feel we can't discuss well about each point with this large PR. Is it possible to split this into a few independent PRs? (like replace global state first, then change the default log levels, then introduce custom log levels, etc)

LionC:
> This PR changes quite many aspects of log module. And I feel we can't discuss well about each point with this large PR. Is it possible to split this into a few independent PRs? (like replace global state first, then change the default log levels, then introduce custom log levels, etc)

I agree - there is a lot of different discussion points and covering all of them in one review / comment will be very hard to follow quickly.

timreichen:
Closing in favour of https://github.com/denoland/deno_std/pull/1146

PR #1138, opened 2021-08-16
0.105.0

dsherret:
I'm going to do the release around 1:00pm ET because that's when I won't have any inturruptions.

### DIFF
--- a/Releases.md
+++ b/Releases.md
@@ -1,3 +1,12 @@
+### 0.105.0 / 2021.08.16
+
+- refactor(bytes): rename `contains` to `includes` with optional argument `fromIndex` (#1133)
+- fix(crypto): make crypto bench depend on sibling version of testing module (#1135)
+- fix(collections): intersect does not handle duplicate values in head properly (#1131)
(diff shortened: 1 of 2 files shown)


Releases.md
```diff
@@ -1,3 +1,15 @@
+### 0.105.0 / 2021.08.16
+
+- refactor(bytes): rename `contains` to `includes` with optional argument
+  `fromIndex` (#1133)
+- fix(crypto): make crypto bench depend on sibling version of testing module
+  (#1135)
+- fix(collections): intersect does not handle duplicate values in head properly
+  (#1131)
+- docs(collections): fix typo in collections docs (#1127)
```
bartlomieju, line 9: This could be dropped

Releases.md
```diff
@@ -1,3 +1,15 @@
+### 0.105.0 / 2021.08.16
+
+- refactor(bytes): rename `contains` to `includes` with optional argument
```
bartlomieju, line 3: this might be a breaking change so maybe `BREAKING(bytes)`?

bartlomieju (approved):

PR #1139, opened 2021-08-18
docs(testing): Add `Deprecated` tag for `assertThrowsSync` testing function

getspooky:


zhangyuannie:
Do you think it might be better to simply rename the function in README since `assertRejects` is missing form it?
```diff
- - `assertThrowsAsync()` - Expects the passed `fn` to be async and throw (or
+ - `assertRejects()` - Expects the passed `fn` to be async and throw (or
```

getspooky:
I think it would be good to keep the `assertThrowsAsync`.

kt3k:
I agree with @zhangyuannie . Keeping deprecated item as a doc entry doesn't make sense to me.

I think the entry for `assertThrowsSync` should look like:
```
- `assertThrowsAsync()` - Deprecated. Use `assertRejects()` instead.
```

caspervonb:
No point in documented deprecated things in the "manual".
Just document the new one, the old one still has a jsdoc entry until removed.

bartlomieju:
This PR is superseded by #1198, thanks @getspooky but closing without merge.

PR #1140, opened 2021-08-18
feat(node/buffer): add missing exports

Mesteery:


Mesteery:
How can I export atob and btoa?

kt3k (approved):
@Mesteery LGTM. Thank you for the contribution!

kt3k:
> How can I export atob and btoa?

Probably you can do something like `export const atob = globalThis.atob;`?

PR #1141, opened 2021-08-18
feat: streams utilities

crowlKats:


io/mod.ts
```diff
@@ -3,6 +3,5 @@ export * from "./buffer.ts";
 export * from "./bufio.ts";
 export * from "./ioutil.ts";
 export * from "./readers.ts";
-export * from "./streams.ts";
```
kt3k, line 6: Moving these to `std/streams` looks reasonable to me. But how about keeping alias to `std/streasm/conversion.ts` with `@deprecated` annotations? That makes the transitions easier for the users.

crowlKats: > Moving these to `std/streams` looks reasonable to me. But how about keeping alias to `std/streasm/conversion.ts` with `@deprecated` annotations? That makes the transitions easier for the users.

how would an alias have specific `@deprecated` annotation?

crowlKats:
One issue is that `readerFromIterable` which returns a `Deno.Reader` would be moved to streams folder as well. do we want to have `io/conversion.ts` for that?

nayeemrmn:
If you think about it almost everything in `std/io` revolves around streams, including utilities to convert them between formats (e.g. `readableStreamFromReader()`), collect them (e.g. `readAll()`), etc. Things like `Buffer` exist to hold data for the purpose of appending to / consuming from it with stream APIs. Once go-like stream interfaces are fully deprecated/removed from CLI, most of `std/io` should be nuked at that point and rebuilt fully around whatwg streams effectively becoming the one proposed here.

In fewer words, `std/io` is currently in a very messy state and we should avoid potentially spreading that outwards until 2.0 APIs drop and we have a clear picture of what to do with it. The new utilities you have are good and would survive CLI stream overhauls, so for now I would just throw them all in `io/streams.ts`.

crowlKats:
Not sure whats going wrong with CI in an unrelated module. testing locally works fine

kt3k:
Is it intentional to put `copy`, `readAll`, `readAllSync`, `writeAll`, `writeAllSync` in `streams/conversion.ts`? These don't seem involving Streams, so it seems a little strange to me. Maybe we should keep them somewhere in `io/`?

Otherwise this looks good to me.

kt3k:
> I believe it's streams as in general io streams, not specifically whatwg streams. So Reader and Writer should count.

Ok. Then it's fine

crowlKats:
> > Is it intentional to put `copy`, `readAll`, `readAllSync`, `writeAll`, `writeAllSync` in `streams/conversion.ts`? These don't seem involving Streams, so it seems a little strange to me. Maybe we should keep them somewhere in `io/`?
> 
> I believe it's streams as in general io streams, not specifically whatwg streams. So `Reader` and `Writer` should count.

it kinda is more targeted whatwg streams, but since deprecating go-like streams isnt that far off, i think its fine to have those functions there for now

PR #1143, opened 2021-08-19
feat(collentions/sortBy): Change the selector function to be called only once at the beginning

### LINKED ISSUE #1142: feat(collentions/sortBy): Change the selector function to be called only once at the beginning
The selector function (the function passed to the second argument of sortBy) has been called multiple times by Array.prototype.sort.

https://github.com/denoland/deno_std/blob/396445052d25b206e0adb00826c7365783fa578a/collections/sort_by.ts#L33-L37

It seems that we can speed up the execution of the `sortBy()` by changing it to call it only once at the beginning.

<details>
<summary>
benchmark
</summary>

```ts
import { sortBy as originalSortBy } from "https://deno.land/std@0.104.0/collections/sort_by.ts";

function sortBy1(array, selector) {
  const result = array.map((value) => ({ value, selected: selector(value) }))
    .sort((a, b) => {
      const selectedA = a.selected;
      const selectedB = b.selected;

      if (typeof selectedA === "number") {
        if (Number.isNaN(selectedA)) {
          return 1;
        }
        if (Number.isNaN(selectedB)) {
          return -1;
        }
      }
      if (selectedA > selectedB) {
        return 1;
      }
      if (selectedA < selectedB) {
        return -1;
      }
      return 0;
    });
  // This does the same thing as `result.map(v=>v.value)`, but speeds it up by overwriting the original array.
  for (let i = 0; i < result.length; i++) {
    result[i] = result[i].value;
  }
  return result;
}

// benchmark code

const shortArray = Array(10).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));
const longArray = Array(10000).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));

function average(array) {
  return array.reduce((p, c) => p + c, 0) / array.length;
}
const createSortFunc = (n) =>
  (it) => {
    for (let i = 0; i < n; i++); // something long task
    return it.age;
  };

// const targetArray = shortArray;
const targetArray = longArray;
for (let i = 0; i < 5000; i += 100) {
  const sortFunc = createSortFunc(i);
  // Measure the execution time of `sortFunc`
  const sortFuncResult = [];
  for (let i = 0; i < 1000000; i++) {
    const start = performance.now();
    sortFunc({ age: 1 });
    sortFuncResult.push(performance.now() - start);
  }
  // Measure the execution time of `originalSortBy`
  const originalSortByResult = [];
  for (let i = 0; i < 100; i++) {
    const start = performance.now();
    originalSortBy(targetArray, sortFunc);
    originalSortByResult.push(performance.now() - start);
  }
  // Measure the execution time of `sortBy1`
  const newSortByResult = [];
  for (let i = 0; i < 100; i++) {
    const start = performance.now();
    sortBy1(targetArray, sortFunc);
    newSortByResult.push(performance.now() - start);
  }
  console.log(
    `${average(sortFuncResult)},${average(originalSortByResult)},${average(newSortByResult)}`,
  );
}
```

</details>

ayame113:
close #1142

### DIFF
--- a/collections/sort_by.ts
+++ b/collections/sort_by.ts
@@ -31,28 +31,35 @@ export function sortBy<T>(
     | ((el: T) => bigint)
     | ((el: T) => Date),
 ): T[] {
-  return Array.from(array).sort((a, b) => {
-    const selectedA = selector(a);
-    const selectedB = selector(b);
+  const result = array.map((value) => ({ value, selected: selector(value) }))
+    .sort((a, b) => {
+      const selectedA = a.selected;
+      const selectedB = b.selected;
 
-    if (typeof selectedA === "number") {
-      if (Number.isNaN(selectedA)) {
+      if (typeof selectedA === "number") {
+        if (Number.isNaN(selectedA)) {
+          return 1;
+        }
+
+        if (Number.isNaN(selectedB)) {
+          return -1;
+        }
+      }
+
+      if (selectedA > selectedB) {
         return 1;
       }
 
-      if (Number.isNaN(selectedB)) {
+      if (selectedA < selectedB) {
         return -1;
       }
-    }
 
-    if (selectedA > selectedB) {
-      return 1;
-    }
+      return 0;
+    });
 
-    if (selectedA < selectedB) {
-      return -1;
-    }
-
-    return 0;
-  });
+  // This does the same thing as `result.map(v=>v.value)`, but speeds it up by overwriting the original array.
+  for (let i = 0; i < result.length; i++) {
+    (result as unknown as T[])[i] = result[i].value;
+  }
+  return result as unknown as T[];
 }
--- a/collections/sort_by_test.ts
+++ b/collections/sort_by_test.ts
@@ -29,6 +29,20 @@ Deno.test({
 });
 
 Deno.test({
+  name: "[collections/sortBy] calls the selector function once",
+  fn() {
+    let callCount = 0;
+    const array = [0, 1, 2];
+    sortBy(array, (it) => {
+      callCount++;
+      return it;
+    });
+
+    assertEquals(callCount, array.length);
+  },
+});
+
+Deno.test({
   name: "[collections/sortBy] empty input",
   fn() {
     sortByTest(


collections/sort_by.ts
```diff
@@ -31,28 +31,35 @@ export function sortBy<T>(
     | ((el: T) => bigint)
     | ((el: T) => Date),
 ): T[] {
-  return Array.from(array).sort((a, b) => {
-    const selectedA = selector(a);
-    const selectedB = selector(b);
+  const result = array.map((value) => ({ value, selected: selector(value) }))
+    .sort((a, b) => {
```
ayame113, line 35: We can also pass the index to the selector with the following changes:
```suggestion
  const result = array.map((value, i) => ({ value, selected: selector(value, i) }))
    .sort((a, b) => {
```

Use cases are like array_multisort () in php.
https://www.php.net/manual/ja/function.array-multisort.php

```js
const targetArray = [...]
const compareArray = [...]
sortBy(targetArray, (_val, i)=>compareArray[i])
```

collections/sort_by.ts
```diff
@@ -31,28 +31,35 @@ export function sortBy<T>(
...
+      return 0;
+    });
 
-    return 0;
-  });
+  // This does the same thing as `result.map(v=>v.value)`, but speeds it up by overwriting the original array.
+  for (let i = 0; i < result.length; i++) {
+    (result as unknown as T[])[i] = result[i].value;
+  }
+  return result as unknown as T[];
```
ayame113, line 64: Type definition is clunky, but did not come up with other methods.😓

LionC:
Thanks! I will do a review tonight.

One thing to keep in mind: there is a tradeoff between runtime optimization and memory optimization. Optimizing runtime by having more "caches" might not be desirable for every use case for large inputs.

That said this might not be the case here :-)

collections/sort_by.ts
zhangyuannie: I think it is better if we do a `map` here. Since:
- it will gives better typing. 
- we are not returning an optimised array back to the user. I.e. If the user gives an array of integer, instead of giving back `PACKED_SMI_ELEMENTS`, we are giving back un-optimised `PACKED_ELEMENTS`. [[1]](https://v8.dev/blog/elements-kinds)
- It is cleaner, `map` is very fast, we are only sacrificing a bit of memory here.

ayame113: Certainly, it seems that there is not much difference in execution time. I will change this to use `map`.

LionC:
So, looking through this, I am not too sure if we are making the right tradeoff here. We are **heavily** increasing memory consumption here, creating a completely new array of **objects** with the same length as the input and creating yet another array on top. All of that is memory scaling with input, and creating all the objects is not free performance wise either.

All that to have some fewer calls on a selector function, which in real world scenarios is something like `(it) => it.someField` most of the time - something really fast (most likely inline cached access, function will probably be inlined as well).

I think we would need to do heavy benchmarking on memory and performance for this to be an improvement that is worth it, right now I am not sure enough about it.

ayame113:
@LionC Even in the original implementation, the array is recreated using Array.from (), and I think that there is not much difference in memory usage.
https://github.com/denoland/deno_std/blob/475a982c28616975de1632a3af09b176b953a612/collections/sort_by.ts#L33-L35

If we care about memory usage,we can undo what I did in [this comment](#discussion_r692714944).

Anyway, I agree that a detailed benchmark is needed. :smile:

LionC:
> @LionC Even in the original implementation, the array is recreated using Array.from (), and I think that there is not much difference in memory usage.
> 
> https://github.com/denoland/deno_std/blob/475a982c28616975de1632a3af09b176b953a612/collections/sort_by.ts#L33-L35
> 
> If we care about memory usage,we can undo what I did in [this comment](#discussion_r692714944).
> 
> Anyway, I agree that a detailed benchmark is needed. 😄

But we are creating an additional array here, because we `map` twice. `Array.from(array)` is also just a copy of the exact set of references and heavily optimized (as discussed in another PR recently^^).

We are creating a completely new array with a new object for every single entry - that is a huge difference in memory and potentially even performance.

But let's get those benchmarks up, that should be the deciding factor anyways :-)

ayame113:
I took the benchmark about this again. First, I measured the execution speed.

The measurement targets are three functions.
- before: Original implementation
- after (using `for`):
```ts
function sortByWithFor(array, selector) {
  const result = array.map((value) => ({ value, selected: selector(value) }))
    .sort((a, b) => {
      const selectedA = a.selected;
      const selectedB = b.selected;

      if (typeof selectedA === "number") {
        if (Number.isNaN(selectedA)) {
          return 1;
        }
        if (Number.isNaN(selectedB)) {
          return -1;
        }
      }
      if (selectedA > selectedB) {
        return 1;
      }
      if (selectedA < selectedB) {
        return -1;
      }
      return 0;
    });
  // This does the same thing as `result.map(v=>v.value)`, but speeds it up by overwriting the original array.
  // https://github.com/denoland/deno_std/pull/1143/commits/8c9c7180c56d15a9c44e20df5d5066085b27dc03
  for (let i = 0; i < result.length; i++) {
    result[i] = result[i].value;
  }
  return result;
}
```
- after (using `map`)
```ts
function sortByWithMap(array, selector) {
  return array.map((value) => ({ value, selected: selector(value) }))
    .sort((a, b) => {
      const selectedA = a.selected;
      const selectedB = b.selected;

      if (typeof selectedA === "number") {
        if (Number.isNaN(selectedA)) {
          return 1;
        }
        if (Number.isNaN(selectedB)) {
          return -1;
        }
      }
      if (selectedA > selectedB) {
        return 1;
      }
      if (selectedA < selectedB) {
        return -1;
      }
      return 0;
    }).map((element) => element.value);
}
```

<details>
<summary>code</summary>

```ts
// test.js

async function runBenchmark(fileName, arg) {
  // Run the subprocess every time to avoid V8 optimization and measure execution time.
  const p = Deno.run({
    stdout: "piped",
    cmd: [
      "deno",
      "run",
      "--allow-hrtime",
      `${new URL(fileName, import.meta.url)}`,
      JSON.stringify(arg),
    ],
  });

  const [_status, stdout] = await Promise.all([
    p.status(),
    p.output(),
  ]);
  p.close();

  return Number(new TextDecoder().decode(stdout));
}

function average(array) {
  return array.reduce((p, c) => p + c, 0) / array.length;
}

for (let arrayLength = 1000; arrayLength < 10000; arrayLength += 1000) {
  const sortFuncLoop = 0;
  const runs = 1000;

  const resultsWithMap = [];
  for (let i = 0; i < runs; i++) {
    resultsWithMap.push(
      await runBenchmark("sort_by_with_map.js", { arrayLength, sortFuncLoop }),
    );
  }

  const resultsWithFor = [];
  for (let i = 0; i < runs; i++) {
    resultsWithFor.push(
      await runBenchmark("sort_by_with_for.js", { arrayLength, sortFuncLoop }),
    );
  }

  const resultsOriginal = [];
  for (let i = 0; i < runs; i++) {
    resultsOriginal.push(
      await runBenchmark("original_sort_by.js", { arrayLength, sortFuncLoop }),
    );
  }

  console.log(
    `${arrayLength},${
      [resultsOriginal, resultsWithFor, resultsWithMap]
        .map(average).join(",")
    }`,
  );
}

```


```ts
// original_sort_by.js
import { sortBy as originalSortBy } from "https://deno.land/std@0.104.0/collections/sort_by.ts";

const { arrayLength = 100, sortFuncLoop } = JSON.parse(Deno.args[0] || "{}");

/*const idToPrice = new Map(
  [...Array(100).keys()].map((i) => [i, Math.random()]),
);*/

const idToPrice = [...Array(100).keys()].map((id) => ({
  id,
  price: Math.random(),
}));

const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  id: Math.floor(Math.random() * 100),
}));

/*const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));*/

const start = performance.now();
originalSortBy(targetArray, (element) => {
  // for (let i = 0; i < sortFuncLoop; i++);
  // return element.age;
  // return idToPrice.get(element.id);
  return idToPrice.find((e) => e.id === element.id).price;
});
const time = performance.now() - start;
Deno.stdout.write(new TextEncoder().encode(`${time}`));

```

```ts
// sort_by_with_map.js
const { arrayLength, sortFuncLoop } = JSON.parse(Deno.args[0]);

function sortByWithMap(array, selector) {
  return array.map((value) => ({ value, selected: selector(value) }))
    .sort((a, b) => {
      const selectedA = a.selected;
      const selectedB = b.selected;

      if (typeof selectedA === "number") {
        if (Number.isNaN(selectedA)) {
          return 1;
        }
        if (Number.isNaN(selectedB)) {
          return -1;
        }
      }
      if (selectedA > selectedB) {
        return 1;
      }
      if (selectedA < selectedB) {
        return -1;
      }
      return 0;
    }).map((element) => element.value);
}

/*const idToPrice = new Map(
  [...Array(100).keys()].map((i) => [i, Math.random()]),
);*/

const idToPrice = [...Array(100).keys()].map((id) => ({
  id,
  price: Math.random(),
}));

const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  id: Math.floor(Math.random() * 100),
}));

/*const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));*/

const start = performance.now();
sortByWithMap(targetArray, (element) => {
  // for (let i = 0; i < sortFuncLoop; i++);
  // return element.age;
  // return idToPrice.get(element.id);
  return idToPrice.find((e) => e.id === element.id).price;
});
const time = performance.now() - start;
Deno.stdout.write(new TextEncoder().encode(`${time}`));

```

```ts
// sort_by_with_for.js
const { arrayLength, sortFuncLoop } = JSON.parse(Deno.args[0]);

function sortByWithFor(array, selector) {
  const result = array.map((value) => ({ value, selected: selector(value) }))
    .sort((a, b) => {
      const selectedA = a.selected;
      const selectedB = b.selected;

      if (typeof selectedA === "number") {
        if (Number.isNaN(selectedA)) {
          return 1;
        }
        if (Number.isNaN(selectedB)) {
          return -1;
        }
      }
      if (selectedA > selectedB) {
        return 1;
      }
      if (selectedA < selectedB) {
        return -1;
      }
      return 0;
    });
  for (let i = 0; i < result.length; i++) {
    result[i] = result[i].value;
  }
  return result;
}

/*const idToPrice = new Map(
  [...Array(100).keys()].map((i) => [i, Math.random()]),
);*/

const idToPrice = [...Array(100).keys()].map((id) => ({
  id,
  price: Math.random(),
}));

const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  id: Math.floor(Math.random() * 100),
}));

/*const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));*/

const start = performance.now();
sortByWithFor(targetArray, (element) => {
  // for (let i = 0; i < sortFuncLoop; i++);
  // return element.age;
  // return idToPrice.get(element.id);
  return idToPrice.find((e) => e.id === element.id).price;
});
const time = performance.now() - start;
Deno.stdout.write(new TextEncoder().encode(`${time}`));

```

</details>
-----

# results

As you say, for simple selector functions like `(it) => it.someField`, the original implementation is faster.
```ts
const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));

sortBy(targetArray, (element) => element.age});
```
image


However, the original implementation is slower when doing more complicated things. (Especially for long arrays)

```ts
const idToPrice = new Map(
  [...Array(100).keys()].map((i) => [i, Math.random()]),
);

const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  id: Math.floor(Math.random() * 100),
}));

sortBy(targetArray, (element) => idToPrice.get(element.id));
```
image


```ts
const idToPrice = [...Array(100).keys()].map((id) => ({
  id,
  price: Math.random(),
}));

const targetArray = Array(arrayLength).fill().map(() => ({
  name: "aaa",
  id: Math.floor(Math.random() * 100),
}));

sortBy(targetArray, (element) => idToPrice.find((e) => e.id === element.id).price);
```
image


And the slower the selector function, the slower the original implementation.

```ts
const targetArray = Array(100).fill().map(() => ({
  name: "aaa",
  age: Math.random(),
}));

sortBy(targetArray, (element) => {
  // Slow the selector function with an empty for loop
  for (let i = 0; i < selecorFuncLoop; i++);
  return element.age;
});
```
image


In my opinion, I would like to prioritize 0ms-5ms optimizations in other cases over 0ms-1ms optimizations in a particular case (like `(it) => it.someField`).

And I will take the benchmark of memory usage...

LionC:
While I encourage exploring this deeper, the fact that for the most common use cases, which is selectors being very simple functions, this performs worse or equal **and** will have a significant memory and thus GC (which is hard to measure performance) impact, I think I am against this change as it stands.

evanwashere:
the way you did benchmarks is slightly inaccurate, so i redid simple selector benchmark
benchmark

according to results (99th and 99.*th percentiles) as array size grows, so does garbage created by your implementations and that causes unwanted GC spikes

i reimplemented your idea with goal to do the same but reduce garbage and GC spikes as much as possible (in benchmarks it's called "my")

```js
export function sort(array, selector) {
  const len = array.length;
  const indexes = new Array(len);
  const selectors = new Array(len);

    indexes[i] = i;
    const s = selector(array[i]);
    selectors[i] = Number.isNaN(s) ? null : s;
  };

  indexes.sort((ai, bi) => {
    const a = selectors[ai];
    const b = selectors[bi];
    if (a === null) return 1;
    if (b === null) return -1;
    return a > b ? 1 : a < b ? -1 : 0;
  });

  for (let i = 0; i < len; i++) indexes[i] = array[indexes[i]];

  return indexes;
}
```

and i achieved exactly that, but like @LionC i believe most selectors are fast and simple, so this change will usually end up being negative

kt3k:
@LionC 
> the fact that for the most common use cases, which is selectors being very simple functions

Do we really assume this? I feel that assumption unnecessarily limits the usage of this function.

Something like the 3rd example of @ayame113's benchmarks seems very possible in actual applications to me, and it has about 2x difference. So I think this change is worth inclusion regarding the exec speed.

kt3k:
BTW lodash seems doing the similar to this change https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L929-L937

LionC:
> @LionC
> 
> > the fact that for the most common use cases, which is selectors being very simple functions
> 
> Do we really assume this? I feel that assumption unnecessarily limits the usage of this function.
> 
> Something like the 3rd example of @ayame113's benchmarks seems very possible in actual applications to me, and it has about 2x difference. So I think this change is worth inclusion regarding the exec speed.

I agree that we should not limit the function. However, the decision here is a tradeoff, not a straight improvement. As you can see in @evanwashere 's benchmarks, the version in the PR is **three times slower** on small selectors because of GC. It will also consume **a lot** more memory, because instead of holding just the selector results and the array in memory, it adds a whole new array on top with a newly allocated and kept object for each entry.

So the decision we are making is to make the function a lot worse for smal selectors while making it faster for big ones. But the same result can be achieved by memoizing the big selector as a user (and the performance will be even better emmory wise, because just the results are kept without an additional array and additional objects). So by changing this, we are keeping the very efficient implementation for small selectors away from people, while allowing them something that they can already do with the current implementation with better performance than in this change, albeit by wrapping their selector in a memoizer.

kt3k:
I read @evanwashere 's comment more carefully and it looks that his suggested implementation has less problem than this PR and also solves the original issue #1142. Maybe we should follow it? What do you think @ayame113 @LionC ?

ayame113:
I benchmarked memory usage.
This measures the difference between `Deno.memoryUsage().heapUsed`.
As you say, there is a small increase in memory.

image

-----

> i reimplemented your idea with goal to do the same

@evanwashere Thank you for your suggestion. Your code seems to have an advantage over my code in terms of execution speed and memory usage.

> I read @evanwashere 's comment more carefully and it looks that his suggested implementation has less problem than this PR and also solves the original issue #1142. Maybe we should follow it?

@kt3k It makes sense. I would like to update the PR using the code of @evanwashere .

ayame113:
So far, my opinion is as follows.
- Note that sorting using `(it) => it.someField` is fast from the beginning. For an array of length 100, I don't really care if 0.25ms goes to 0.28ms, but 2.1ms goes to 0.5ms, which is a big change.
- In the original implementation, I think the selector function will be called up to `2nlogn` times for an array of length n. I think changing this to n calls is a legitimate way to speed things up.
- The original implementation seems to be faster only for `(it) => it.someField` or similar patterns. Is this the only use case for sortBy? You can also consider introducing something like `sortByKey(array: T[], key: keyof T): T[]` if necessary.

I noticed that there was a trade-off in this in my first benchmark on #1142. I thought this was a change that might have disagreements, so I opened an issue and PR to hear people's opinions. I'm not going to stick to this. @LionC Thank you for your feedback on this!

ayame113:
> @ayame113 see https://v8.dev/blog/optimizing-v8-memory on how to benchmark memory - you need to do that from the "outside" I think.

I tried `--v8-flags=--track-gc-object-stats` but nothing was output. I think I made a mistake, but I'm not sure. sorry.

ayame113:
For the future, I think there will be a slight speedup with the following changes, but I'm okay with this as it has already been merged.

```diff
-  const indexes = new Array(len);
-  const selectors = new Array(len);

-  for (let i = 0; i < len; i++) {
-    indexes[i] = i;
-    const s = selector(array[i]);
-    selectors[i] = Number.isNaN(s) ? null : s;
-  }
+  const indexes = array.map((_, i) => i);
+  const selectors = array.map((element) => {
+    const s = selector(element);
+    return Number.isNaN(s) ? null : s;
+  });
```

Thanks to everyone who reviewed it!

PR #1144, opened 2021-08-19
feat(collections): add maxOf

Hakizu:
Suggestion how to implement a maxOf method

Related: #1065

### DIFF
--- a/collections/max_of.ts
+++ b/collections/max_of.ts
@@ -1,0 +1,39 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Applies the given selector to all elements in the given collection and returns the max of it
+ *
+ * Example:
+ *
+ * ```ts
+ * import { sumOf } from "./sum_of.ts"
+ * import { assertEquals } from "../testing/asserts.ts"
+ *
+ * const inventory = [
+ *      { name: "mustard", count: 2 },
+ *      { name: "soy", count: 4 },
+ *      { name: "tomato", count: 32 },
+ *  ];
+ * const maxItem = maxOf(inventory, (i) => i.count);
+ * 
+ * assertEquals(maxItem, 32);
+ * ```
+ */
+
+export function maxOf<T>(
+    array: readonly T[],
+    selector: (el: T) => number,
+): number {
+    let maxOf = -Infinity;
+    
+    for (const i of array) {
+        if (selector(i) > maxOf) {
+            maxOf = selector(i);
+        } 
+        else if (Number.isNaN(selector(i))) {
+            return maxOf = NaN
+        }
+    }
+    
+    return maxOf;
+}
--- a/collections/max_of.ts
+++ b/collections/max_of.ts
@@ -6,7 +6,7 @@
  * Example:
  *
  * ```ts
- * import { sumOf } from "./sum_of.ts"
+ * import { maxOf } from "./max_of.ts"
  * import { assertEquals } from "../testing/asserts.ts"
  *
  * const inventory = [
--- a/collections/max_of.ts
+++ b/collections/max_of.ts
@@ -1,7 +1,9 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
 /**
- * Applies the given selector to all elements of the given collection and returns the max value of all elements
+ * Applies the given selector to all elements of the given collection and
+ * returns the max value of all elements. If an empty array is provided the
+ * function will return undefined
  *
  * Example:
  *
--- a/collections/max_of.ts
+++ b/collections/max_of.ts
@@ -15,25 +15,24 @@
  *      { name: "tomato", count: 32 },
  *  ];
  * const maxItem = maxOf(inventory, (i) => i.count);
- * 
+ *
  * assertEquals(maxItem, 32);
  * ```
  */
 
 export function maxOf<T>(
-    array: readonly T[],
-    selector: (el: T) => number,
+  array: readonly T[],
+  selector: (el: T) => number,
 ): number {
-    let maxOf = -Infinity;
-    
-    for (const i of array) {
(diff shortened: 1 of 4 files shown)


LionC:
I will add a review :-)

collections/max_of.ts
```diff
@@ -0,0 +1,39 @@
...
+ * const maxItem = maxOf(inventory, (i) => i.count);
+ * 
+ * assertEquals(maxItem, 32);
+ * ```
+ */
+
+export function maxOf<T>(
+    array: readonly T[],
+    selector: (el: T) => number,
+): number {
```
LionC, line 26: I think we should support `bigint` here, with something like this:

```suggestion
export function maxOf<T, S extends ((el: T) => number) | ((el: T) => bigint))>(
    array: readonly T[],
    selector: S,
): ReturnType<S> | undefined {
```

collections/max_of.ts
```diff
@@ -0,0 +1,39 @@
...
+    let maxOf = -Infinity;
```
LionC, line 27: I think we should return `undefined` on an empty input. The maximum for an empty list is not really commonly defined, so `undefined` seems correct.

```suggestion
    let maxOf = undefined;
```

collections/max_of.ts
```diff
@@ -0,0 +1,39 @@
...
+    
+    for (const i of array) {
+        if (selector(i) > maxOf) {
+            maxOf = selector(i);
+        }
```
LionC, line 32: You are calling `selector` twice here, I think you could reasonably call it only once instead (that would also lead to only one thing happening per line).

collections/max_of.ts
```diff
@@ -0,0 +1,39 @@
...
+): number {
+    let maxOf = -Infinity;
+    
+    for (const i of array) {
+        if (selector(i) > maxOf) {
+            maxOf = selector(i);
+        } 
+        else if (Number.isNaN(selector(i))) {
+            return maxOf = NaN
+        }
```
LionC, line 35: `else` on a new line is not really common style (did you do `deno fmt` and `deno lint`?).

Extremely minor nitpick: I would argue that this is not really an `else` case "semantically`. I would personally `continue` in the first `if` and convert the second case to its own `if`.

Also why do you set `maxOf` in the second case? If you are `return`ing anyways, you can just `return Number.NaN` (or even better, your selector result, whatever `NaN` representation that is).

collections/max_of_tests.ts
```diff
@@ -0,0 +1,89 @@
...
+    maxOf(array, (i) => i + 2);
+  
+    assertEquals(array, [1, 2, 3, 4]);
+});
+
+Deno.test("[collections/maxOf] Empty array results in -Infinity", () => {
+    const array: number[] = []
+
+    const actual = maxOf(array, (i) => i)
+    assertEquals(actual, -Infinity)
```
LionC, line 61: See above

collections/max_of_tests.ts
```diff
@@ -0,0 +1,89 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+import { maxOf } from "./max_of.ts";
+import { assertEquals } from "../testing/asserts.ts"
+
+Deno.test("[collections/maxOf] Regular max", () => {
```
LionC, line 5: We should add `bigint` Tests if we support it

collections/max_of.ts
```diff
@@ -0,0 +1,39 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Applies the given selector to all elements in the given collection and returns the max of it
+ *
```
LionC, line 5: We should add a section to the `collections/README.md` as well

LionC (changes requested):
Thanks for this! I like that you included all the commonly missed things from the other PRs :-)

I think there is still some stuff to do, but this is a very good draft!

LionC (approved):
LGTM! Pinging @kt3k for final review :-)

collections/max_of_tests.ts
```diff
@@ -0,0 +1,102 @@
...
+
+Deno.test("[collections/maxOf] No mutation", () => {
+  const array = [1, 2, 3, 4];
+
+  maxOf(array, (i) => i + 2);
+
+  assertEquals(array, [1, 2, 3, 4]);
+});
+
+Deno.test("[collections/maxOf] Empty array results in -Infinity", () => {
```
kt3k, line 70:
```suggestion
Deno.test("[collections/maxOf] Empty array results in undefined", () => {
```

collections/max_of.ts
```diff
@@ -0,0 +1,52 @@
...
+ * ];
+ * const maxCount = maxOf(inventory, (i) => i.count);
+ *
+ * assertEquals(maxCount, 32);
+ * ```
+ */
+export function maxOf<T>(
+  array: readonly T[],
+  selector: (el: T) => number,
+): number;
```
kt3k, line 25: Is this return type correct? I think this can return undefined. So maybe we should type it `number | undefined`?

collections/max_of.ts
```diff
@@ -0,0 +1,52 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Applies the given selector to all elements of the given collection and returns the max value of all elements
```
kt3k, line 4: I think the document should mention about empty array case.

kt3k (commented):
@Hakizu Thank you for your contribution. Left a few comments.

collections/max_of.ts
LionC: Same applies for the `bigint` signature @Hakizu !

Missed that one, thanks for the catch @kt3k

Hakizu:
Thanks for the feedback from both of you. I updated the PR

kt3k (approved):
@Hakizu LGTM. Thank you for the contribution!

PR #1145, opened 2021-08-19
chore(collections): Test that `sortBy()` of NaN values is stable.

andreubotella:


LionC:
LGTM! Fits what we talked about and accurately tests for possible `NaN` sorting problems :-) Tests succeeds on my machine as well, so it seems like we do not need to adapt the current implementation?

Pinging @kt3k for final review

kt3k (approved):
@andreubotella Thanks! LGTM.

PR #1146, opened 2021-08-20
BREAKING(log): class based logger

timreichen:
This PR is the successor of https://github.com/denoland/deno_std/pull/1137 where it was discussed to split up the rewrite into multiple PRs. 

Content of this PR:
- remove global state
- implement class based logger
- combining log level names and codes to log level interface with `name` and `code` property
- reflected log levels from console
  `trace`
  `debug`
  `info`
  `warn`
  `error`

This topics must be discussed before another PR:
- custom log levels
- default logger instance
- default handlers


timreichen:
@kt3k @LionC ready for review.

timreichen:
@LionC sorry took a bit longer to review.
- `args`
I took the name from [here](https://console.spec.whatwg.org/#logger). But instead of `message` and `args` or `additionalData` having a single array named `data` as in  that contains all the values as used [here](https://console.spec.whatwg.org/#debug) probably would make more sense and be less confusing. `message` would then become `data[0]` which seems straight forward and makes more sense if a logger handles other data (not messages).
- `name`
Is obsolete if logger doesn't need handlers. 
- `handlers`
On a previous PR I had a discussion where it was argued to my dislike that a logger should not have default handlers. But is also obsolete if logger doesn't need handlers.
- `LogLevels`
`level` as the numeric value would describe it well I think.
- `logLevels`
`defaultLogLevels`, or what came to my mind: If Logger is a class, it can have a `static get logLevels() { return […] }`property to access them (`Logger.logLevels.DEBUG`  for example) like WebSocket does it with the [readyState](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/readyState). I think that would be an elegant solution as one only needs to import one class. LogLevels also would become easily extendable
```ts
import { Logger } "https://deno.land/std/log/mod.ts"

const customLogLevel =  { name: "Foo", level: 50 }

class CustomLogger extends Logger {
  static get logLevels() {
    return {
      ...super.logLevels,
      foo: customLogLevel
    }
  }
  foo(...data) {
    this.dispatch(customLogLevel, data)
  }
}
```
```ts
const logger = new CustomLogger(CustomLogger.logLevels.foo)
```
- `functions`
https://github.com/denoland/deno_std/pull/1146#issuecomment-906762105

I am not too big of a fan of having functions that return an object that is interacted in that way. It doesn't feel like normal js to me but more _frameworkish_. I think `Logger` should be a class, where `ConsoleLogger` and `FileLogger` etc. should be derived from.
I also do not agree that there should be generated methods according to the passed log level
`const myLogLevels = {  foo: 10 }` -> … -> logger.foo()`
To me this seems to do a lot of "magic" under the hood. Declared methods in a class are straight forward, a common concept ins js landscape and provide the same functionality with more flexibility.

LionC:
@timreichen Looking at the current docs, everything is exposed via `function`s anyways. The current class approach mostly adds a lot of boilerplate code without really offering any additional features to users. Generallt, `class`es add a lot of complex semantics and ways for users to shoot themselves in the foot - in this use case, without really adding a lot. What we are doing here, is letting people set a handler function and an (unchangeable) log level, I do not see how that needs classes and all the things that come with them. I would also challenge the notion that classes are "more JS" than anything else - JS is an agnostic language that allows for lots of different paradigms, object orientation is really just one of many.

I would argue that `console` and the current implementation both set the whole `logger.[logLevel](message)` idiom as a standard and it makes a lot of sense why. Because what makes loggers special when compared to just some JS function (which we would not need an abstraction for) is that they have levels, a level filter and accept a message. Standardizing that isn't really magic, it is just plain old JS (and safe TS coverage for it!) that saves a lot of boilerplate for all custom log level users.

I do not think putting `message` and some optional `additionalData` into one is a good idea - most users will want to pass a `string` message and have additional stuff sometimes, forcing them to wrap that in an array on every log call feels weird to me.

If the function approach is too far away for you from the idea in here, I would be happy to open an alternative PR, while still collaborating on a `class` approach with what I already suggested. More choice might mean a better result after all :-)

timreichen:
@LionC
# factory functions and classes
What I meant is that I don't know of any example in core js where factory functions in that way are used instead of classes. I know a lot of cases where classes are used though (`new Date(…)`, ` new Proxy(…)`, `new TextDecoder()`, etc.)
When an object has some sort of state (as logLevel, for FileLogger filepath) I think it is the practice in core js to have it as a class? I also would argue that these properties needs to be readable as well as changeable. There is no harm in that and makes the usage more flexible. No need to create a new logger when you just want to change one property.

# log level methods
Console does not always use the idiom `logger.[logLevel](message)`, since console also has different methods like `timeLog`, `group`, `debug` which use `log` as a log level.

# message, additionalData, ...data
I think every method can take an arbitrary number of arguments as console.log("foo", "bar", …), so I asked myself if there is a reason why to prefer one argument over any number where they are collected via `...data` in the method. ConsoleLogger not only wants to allow one string message but also other types which are formatted as console does, right?

I could see some kind of functional abstraction of the class as done in [std/datetime](https://github.com/denoland/deno_std/blob/47b915dc1d8ab650208a37957bb534fe1367bde6/datetime/mod.ts#L29) if it is somehow stateless. But we should avoid to have two ways to achieve the same result in std I think.

LionC:
To clarify: I am proposing to open an alternative PR for log with that approach - Ia m not suggesting to have two log modules in the end.

I will try to adress the rest later, but it sounds like an alternative PR might be a solution :-)

kt3k:
I'm thinking of just merging this as is.

@LionC Are you still planning on a new proposal?

LionC:
See #1404

kt3k:
@LionC Thanks!

timreichen:
I have made another simpler to use version of class based approach for std which is on that that branch: https://github.com/timreichen/deno_std/tree/class-based-log-v2/log. @LionC @bartlomieju before I update this PR or even a new one, maybe you could have a look at it? We probably should decide soon what way to go next because there are two different approaches for `std/log` atm.

LionC:
I think I never received a re-review on my log PR after the inital feedback by @bartlomieju (sorry if I am wrong here as I have not pushed that for some time), so I am not sure how tl reviw thos without knowing where core wants take this (just clarifying that because I was explicitly mentioned).

timreichen:
@LionC Yes, I don't know what is going on atm with the log rewrites. I just put this up to date and rebased it as I think class based is still the way to go and a bit of boilerplate is fine if it makes code more consistent and readable. @bartlomieju has there been any updates on how to proceed with std/log?

PR #1147, opened 2021-08-21
feat(node/buffer): export atob and btoa

Mesteery:


kt3k (approved):
@Mesteery Thank! LGTM

PR #1149, opened 2021-08-21
cleanup(collections/distinct): remove unnecessary set allocation

grian32:


collections/distinct.ts
```diff
@@ -16,7 +16,5 @@
  * ```
  */
 export function distinct<T>(array: readonly T[]): T[] {
-  const set = new Set(array);
-
-  return Array.from(set);
+  return Array.from(new Set(array));
```
LionC, line 19: Nitpick: I personally prefer doing one thing per line, especially if those things are impactful, like iterating a whole array:

```suggestion
  return Array.from(
    new Set(array),
  );
```

LionC (commented):
Did you run benchmarks against this? I do not have proof, but I would guess that v8 optimizes this, leading to this having no impact.

I really have to set up benchmarks to compare stuff like that - I will try to do that next week.

grian32:
> Did you run benchmarks against this? I do not have proof, but I would guess that v8 optimizes this, leading to this having no impact.
> 
> I really have to set up benchmarks to compare stuff like that - I will try to do that next week.

Not really sure how to benchmark things memory wise in JS, I doubt it makes a difference speed wise though.

LionC (approved):
Even though performance impact is unclear (and we should get memory and runtime benchmarks up for that), this change is still very readable and theoretically has the chance to be more optimizable, so I am in favour of including it.

Pinging @kt3k for final review

collections/distinct.ts
```diff
@@ -16,7 +16,7 @@
  * ```
  */
 export function distinct<T>(array: readonly T[]): T[] {
-  const set = new Set(array);
-
-  return Array.from(set);
+  return Array.from(
+    new Set(array),
+  );
```
bartlomieju, line 21: I don't think this removes an allocation for set, the only difference is that it's not assigned to a variable. I believe there won't be any performance difference, V8 should be smart enough to optimize it to this form anyway.

grian32: As @LionC said it theoretically has the chance to be more optimizable, while not changing readability etc, I don't really think it hurts anyone, we could wait for benchmarks though.

zhangyuannie: > Even though performance impact is unclear (and we should get memory and runtime benchmarks up for that), this change is still very readable and theoretically has the chance to be more optimizable

The performance impact is clear and this has no chance to be more optimisable. I think developers should know that Inlining an object allocation will not change anything in any sane programming language.
<details>
  <summary>For those curious, they generate the exact same bytecode in a different order</summary>
  
#### Before:
```
[generated bytecode for function: distinct (0x34cb01964439 <SharedFunctionInfo distinct>)]
Parameter count 2
Register count 3
Frame size 24
   17 E> 0x34cb01964c8e @    0 : a7                StackCheck 
   39 S> 0x34cb01964c8f @    1 : 13 00 00          LdaGlobal [0], [0]
         0x34cb01964c92 @    4 : 26 fa             Star r1
         0x34cb01964c94 @    6 : 25 fa             Ldar r1
   39 E> 0x34cb01964c96 @    8 : 65 fa 02 01 02    Construct r1, a0-a0, [2]
         0x34cb01964c9b @   13 : 26 fb             Star r0
   57 S> 0x34cb01964c9d @   15 : 13 01 04          LdaGlobal [1], [4]
         0x34cb01964ca0 @   18 : 26 f9             Star r2
   70 E> 0x34cb01964ca2 @   20 : 28 f9 02 06       LdaNamedProperty r2, [2], [6]
         0x34cb01964ca6 @   24 : 26 fa             Star r1
   70 E> 0x34cb01964ca8 @   26 : 59 fa f9 fb 08    CallProperty1 r1, r2, r0, [8]
   80 S> 0x34cb01964cad @   31 : ab                Return 
Constant pool (size = 3)
0x34cb01964c31: [FixedArray] in OldSpace
 - map: 0x278804a80791 <Map>
 - length: 3
           0: 0x278804a84411 <String[#3]: Set>
           1: 0x278804a83429 <String[#5]: Array>
           2: 0x278804a83ab9 <String[#4]: from>
Handler Table (size = 0)
[ 1 ]
```
#### After
```
[generated bytecode for function: distinct (0x02dba9824439 <SharedFunctionInfo distinct>)]
Parameter count 2
Register count 3
Frame size 24
   17 E> 0x2dba9824c8e @    0 : a7                StackCheck 
   29 S> 0x2dba9824c8f @    1 : 13 00 00          LdaGlobal [0], [0]
         0x2dba9824c92 @    4 : 26 fa             Star r1
   42 E> 0x2dba9824c94 @    6 : 28 fa 01 02       LdaNamedProperty r1, [1], [2]
         0x2dba9824c98 @   10 : 26 fb             Star r0
   56 E> 0x2dba9824c9a @   12 : 13 02 04          LdaGlobal [2], [4]
         0x2dba9824c9d @   15 : 26 f9             Star r2
         0x2dba9824c9f @   17 : 25 f9             Ldar r2
   52 E> 0x2dba9824ca1 @   19 : 65 f9 02 01 06    Construct r2, a0-a0, [6]
         0x2dba9824ca6 @   24 : 26 f9             Star r2
   42 E> 0x2dba9824ca8 @   26 : 59 fb fa f9 08    CallProperty1 r0, r1, r2, [8]
   72 S> 0x2dba9824cad @   31 : ab                Return 
Constant pool (size = 3)
0x2dba9824c31: [FixedArray] in OldSpace
 - map: 0x3064b0dc0791 <Map>
 - length: 3
           0: 0x3064b0dc3429 <String[#5]: Array>
           1: 0x3064b0dc3ab9 <String[#4]: from>
           2: 0x3064b0dc4411 <String[#3]: Set>
Handler Table (size = 0)
[ 1 ]
```
</details>

However if we are up to this kind of optimisation, what we can do instead is to use spread syntax instead since that will have an actual tiny performance impact  and protect against prototype pollution.
```suggestion
  return [...new Set(array)];
```

LionC: @zhangyuannie my thought process was that with the expression being inline, there is no way it can be referenced in e.g. nested closures, so maybe Turbofan is doing some Vodoo with it. I fully believe you that that is not the case though and will put that on my TIL list :-)

kt3k:
It seems that this doesn't have a good impact on performance. I believe the same generated bytecode proves it (Thanks @zhangyuannie for the research!)

Closing without merge. Thanks @grian32 for the contribution anyway

PR #1156, opened 2021-08-23
feat(collections): add maxBy

majidsajadi:
Implemented the `maxBy` method for the collection module. ref: https://github.com/denoland/deno_std/issues/1065

Types are declare based on [this comment](https://github.com/denoland/deno_std/pull/978#issuecomment-866126420)

Let me know if I should add any specific test cases.


collections/max_by.ts
```diff
@@ -0,0 +1,40 @@
...
+ *     { name: 'John', age: 23 },
+ * ];
+ *
+ * const personWithMaxAge = maxBy(people, i => i.age);
+ *
+ * assertEquals(personWithMaxAge, { name: 'Kim', age: 42 });
+ * ```
+ */
+export function maxBy<T>(
+  collection: Array<T>,
```
LionC, line 24: I think we have settled on `T[]` over `Array<T>` in the module now. Also mark as `readonly` to ensure no mutation happens:

```suggestion
  readonly collection: T[],
```

collections/max_by.ts
```diff
@@ -0,0 +1,40 @@
...
+  selector: ((el: T) => number | string),
```
LionC, line 25: I think this typing is too narrow - it allows a function that **sometimes** a `number` and **sometimes** returns a `string`.

You can do something like `((el: T) => number) | ((el: T) => string)` instead. I would add an overload to the actual exposed signature though, that is more readable and nicer to use.

collections/max_by.ts
```diff
@@ -0,0 +1,40 @@
...
+  for (const current of collection) {
+    const currentValue = selector(current);
+
+    if (currentValue > maxValue) {
+      max = current;
+      maxValue = currentValue;
+    }
+  }
+
+  return max;
```
LionC, line 39: You have not covered the `undefined` case - this will actually error on an empty input. You could do

```suggestion
  let max: T | undefined = undefined;
  let maxValue: ReturnType<typeof selector> = undefined;

  for (const current of collection) {
    const currentValue = selector(current);

    if (maxValue === undefined || currentValue > maxValue) {
      max = current;
      maxValue = currentValue;
    }
  }

  return max;
```

collections/max_by_test.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "../testing/asserts.ts";
+import { maxBy } from "./max_by.ts";
+
```
LionC, line 5: You should add tests for

- empty input (the current implementation actually has a bug there)
- `NaN` and `Infinity` behaviour
- no mutation (make sure the originally passed array does not get changed, see other functions in the module)

LionC (changes requested):

collections/max_by_test.ts
majidsajadi: > [collections/maxBy] of empty array

is a test case with an empty array as input. can you provide. can you please provide me the test case input that shows the bug?  @LionC

LionC: > is a test case with an empty array as input. can you provide. can you please provide me the test case input that shows the bug? @LionC

Yes, if you add this test:

```ts
Deno.test("[collections/maxBy] empty input", () => {
  const emptyArray: Array<{ age: number }> = [];
  const max = maxBy(emptyArray, (it) => it.age);

  assertEquals(max, undefined);
});
```

it will fail.

> there are two test cases with `Infinity` and `NaN` included in the test data. any other usage do you have in mind?

Nevermind, must have missed those, it was late^^ Sorry.

collections/max_by.ts
majidsajadi: if by `empty input` you mean `[]` i don't think there is any bug. since `max` variable initial value is `collection[0]` which in case of empty array is `undefined`. there is already a test case for empty array as input. let me know if I'm wrong.

LionC: See the test case I posted in the other comment - the problem is that you are passing `undefined` into a function that does not expect `undefined`. Typescript did not catch it for you because deno does not have `noUncheckedIndexAccess` turned on.

majidsajadi: Thanks for the provided test case. I fixed the issue. please review again and let me know if I missed anything. if the changes are approved I will update 'minBy' PR too. @LionC

kt3k (approved):
@majidsajadi LGTM. Nice work!

PR #1159, opened 2021-08-23
fix(testing): `assertEquals` now considers constructors equal if one is nullable and the other is `Object`

dsherret:
As discussed.


kitsonk (approved):
LGTM

PR #1167, opened 2021-08-26
feat(collections): add firstNotNullishOf

majidsajadi:
This PR adds `firstNotNullishOf` to the collection module.


collections/README.md
```diff
@@ -455,3 +455,23 @@ console.assert(
   ],
 );
 ```
+
+### firstNotNullishOf
+
+Applies the given selector to elements in the given array until a value is
+produced that is neither `null` nor `undefined` and returns that value Returns
```
kt3k, line 462: Seems like a period is missing?

```suggestion
produced that is neither `null` nor `undefined` and returns that value. Returns
```

kt3k (approved):
@majidsajadi Thank you for the work! LGTM

PR #1169, opened 2021-08-26
feat(collections): add minWith

majidsajadi:
This PR adds `minWith` to the `collections` module.


kt3k:
Side note: This should be useful when the comparator is very complex (like the below) and can't be rewritten with `minBy` easily.

```ts
minWith(array, (a, b) => {
  if (a.foo !== b.foo) {
    return a.foo - b.foo;
  }

  if (a.bar !== b.bar) {
    return a.bar.length - b.bar.length;
  }

  return a.baz - b.baz;
});
```

PR #1171, opened 2021-08-27
chore: updates related to TypeScript 4.4

kitsonk:
This PR does the following:

- Makes updates to `std` to make it compatible with TypeScript 4.4 as configured in Deno CLI.
- Changes a few `@ts-ignore` to `@ts-expect-error` as that is better, as several errors are now "fixed" under TS 4.4
- Make changes so the code base and tests are compatible with `"useUnknownInCatchVariables"` which is shipping as disabled in Deno CLI 1.14 but hopefully will be enabled for 1.15.

There was also a change in how Deno canary checks files for test and caught some errors in modules that weren't covered by tests.

Because of the improvements to TypeScript 4.4 and the ignored errors that are no longer ignored, the next release of `std` will not be compatible with Deno prior to 1.14.


kt3k:
`deno test --doc` seems executing all .ts files with canary. (Is that expected behavior? cc @caspervonb )

And therefore the test command executes `_wasm_crypto/_build.ts`, and that command fails because the command is not supposed to be run in that situation...

bartlomieju:
> `deno test --doc` seems executing all .ts files with canary. (Is that expected behavior? cc @caspervonb )
> 
> And therefore the test command executes `_wasm_crypto/_build.ts`, and that command fails because the command is not supposed to be run in that situation...

We should probably just exclude that file from testing

kt3k:
Ok. I'll update with appropriate `--ignore` options. It seems there are some more files which need to be excluded.

caspervonb:
> > `deno test --doc` seems executing all .ts files with canary. (Is that expected behavior? cc @caspervonb )
> > And therefore the test command executes `_wasm_crypto/_build.ts`, and that command fails because the command is not supposed to be run in that situation...
> 
> We should probably just exclude that file from testing

Should be scanned for documentation tests but the module should not imported.
Looking into it.

> Ok. I'll update with appropriate `--ignore` options. It seems there are some more files which need to be excluded.

Would gate executable scripts with `import.meta.main`.

kitsonk:
@kt3k I rebased and merged your commit... the commit history is a bit odd now in this PR, but hopefully CI passes.

kitsonk:
Hmmm... nope... still doing strange things.

kt3k:
@kitsonk `Run tests canary` step seems duplicated now. I think the first one should be removed.

kitsonk:
Ok, I just need someone to approve it! 😉

kt3k (approved):
LGTM!

Issue #1173, opened 2021-08-27
[Umbrella] `collections` APIs


LionC:
# `collections`

This is an umbrella issue for the `collections` module, which has landed in `std` with [this initial PR](https://github.com/denoland/deno_std/pull/993) and a lot of new contributions since. The goal of this issue is to give an overview about functions that have alreayd been implemented, are being worked on and have not been started yet.

For more details on the idea behind the module, check out its [Readme](https://github.com/denoland/deno_std/blob/main/collections/README.md) and [contribution guide](https://github.com/denoland/deno_std/blob/main/collections/CONTRIBUTING.md). For more details on the history of the module and its proposed signatures, see [the original proposal PR](https://github.com/denoland/deno_std/pull/978).

**If you want to implement one of the missing functions, please open a separate PR** for it and keep discussion in that PR. Please read the [contribution guide](https://github.com/denoland/deno_std/blob/main/collections/CONTRIBUTING.md) and make sure to keep the checklist in mind before opening your PR / marking it ready for review.

If you want to add onto or discuss the list, please do that here.

## Unstarted / Suggestions

None right now

## In Progress

* [ ] aggregateGroups #1553 
* ~filterNotNullish~ #1168
* ~filterValuesNotNullish~ #1186 
* ~without~ #1074

## Done

* [x]  associateBy #1071 
* [x]  associateWith #1213 
* [x]  chunk #993 
* [x]  deepMerge #1075 
* [x]  distinct #993
* [x]  distinctBy #993
* [x]  dropLastWhile #1193 
* [x]  dropWhile #1185 
* [x]  filterEntries #993
* [x]  filterKeys #993
* [x]  filterValues #993
* [x]  findLast #993
* [x]  findLastIndex #1062 
* [x]  firstNotNullishOf #1167
* [x]  groupBy #993
* [x]  includesValue #1184 
* [x]  intersect #993
* [x]  joinToString #1223 
* [x]  mapEntries #993
* [x]  mapKeys #993
* [x]  mapNotNullish #1103 
* [x]  mapValues #993
* [x]  maxBy #1156 
* [x]  maxOf #1144 
* [x]  maxWith #1170 
* [x]  minBy #1157 
* [x]  minOf #1164  
* [x]  minWith #1169 
* [x]  partition #993
* [x]  permutations #993
* [x]  reduceGroups #1187 
* [x]  runningReduce #1226 
* [x]  sample #1212 
* [x]  single #1166 
* [x]  sortBy #1069
[...]

PR #1175, opened 2021-08-27
feat(collections): add takeWhile

grian32:
I've currently made the array return a empty array if the collection given is an empty array, not sure that's the best approach.

Ref: #1173


grian32:
I've optimized this function based on a discussion in discord(read down from here: https://discord.com/channels/684898665143206084/775393009981849600/881533625491947541), credit to @evanwashere for the new implementation.

I'd like also like to bring up re-naming it to `takeWhile`, as that is what rust, the iterator-helpers proposal and kotlin(which this was based off of) uses.

LionC:
I like `takeWhile`. I can't remember which discussion lead to naming it `takeFirstWhile` in the proposal, but I definitely prefer `takeWhile`

LionC (approved):
This looks good to me now, pinging @kt3k

kt3k (approved):
@grian32 Thanks! LGTM

PR #1176, opened 2021-08-29
fix(encoding/toml): escape string key/value

FujiHaruka:
I found a bug in toml's stringify that's because of lack of escaping strings.
This PR fixes it by escaping strings by JSON.stringify().

For example when you run the following code:

```ts
import { stringify } from "https://deno.land/std@0.105.0/encoding/toml.ts";

console.log(stringify({
  '"': '"',
  "'": "'",
  " ": " ",
  "\\": "\\",
  "\n": "\n",
  "\t": "\t",
}));
```

the output is this, which is invalid TOML.

```
""" = """
"'" = "'"
" " = " "
"\" = "\"
"
" = "
"
"	" = "	"
```

The expected output is like this:

```toml
"\"" = "\""
"'"  = "'"
" "  = " "
"\\" = "\\"
"\n" = "\n"
"\t" = "\t"
```


kt3k (approved):
@FujiHaruka LGTM. Thanks!

PR #1177, opened 2021-08-29
refactor: Use native hasOwn

GrosSacASac:


MyIsaak:
If we're refactoring to hasOwn, theres also another usage you could checkout
https://github.com/denoland/deno_std/blob/d4c3330ac8c0c68ffe4d47e692dffa0e8adb2f5b/ws/mod.ts#L2

After that there shouldn't any use for the `hasOwnProperty` https://github.com/denoland/deno_std/blob/d4c3330ac8c0c68ffe4d47e692dffa0e8adb2f5b/_util/has_own_property.ts#L25

So we can remove it, or?

GrosSacASac:
Made another commit

kt3k:
There seem a few type errors. Can you address these? @GrosSacASac

GrosSacASac:
I think the errors are wrong

GrosSacASac:
```
TS2578 [ERROR]: Unused '@ts-expect-error' directive.
      // @ts-expect-error
      ~~~~~~~~~~~~~~~~~~~
```

not from me

lucacasonato:
Please rebase or merge with main.

GrosSacASac:
Rebased

GrosSacASac:
```
 TS2345 [ERROR]: Argument of type 'WebSocketEvent' is not assignable to parameter of type 'object'.
  Type 'string' is not assignable to type 'object'.
  return hasOwn(a, "code");
                ^
    at file:///D:/a/deno_std/deno_std/ws/mod.ts:39:17
```
I don't know how to get rid of this error

ws/mod.ts
```diff
@@ -36,7 +36,7 @@ export interface WebSocketCloseEvent {
 export function isWebSocketCloseEvent(
   a: WebSocketEvent,
 ): a is WebSocketCloseEvent {
-  return hasOwnProperty(a, "code");
+  return hasOwn(a, "code");
```
kt3k, line 39: How about `typeof a !== "string" && hasOwn(a, "code")`?

GrosSacASac: the previous code works the same as the new
```
 Object.prototype.hasOwnProperty.call("a", "b")
 false
 Object.hasOwn("a", "b")
 false
```

if it is a string it will return false like before

GrosSacASac: If think typescript is bugged in this scenario

kt3k: typescript seems supposing the first arg is object (not string) https://github.com/denoland/deno/blob/b518f5e1ba4c3281d4aee4c637f410db382ce751/cli/dts/lib.esnext.object.d.ts

Other solution might be type casting like `a as object` with `// deno-lint-ignore ban-types` lint directive

GrosSacASac: Is it good now ?

kt3k: It should be something like the below, I guess

```ts
// deno-lint-ignore ban-types
return hasOwn(a as object, "code");
```

You can check it with `deno run ws/mod.ts`.

GrosSacASac: I have other unrelated to this PR errors when I do this

GrosSacASac:
All tests seem to pass

kt3k (approved):
@GrosSacASac Thanks! LGTM. Nice work!

Issue #1165, opened 2021-08-26
addr defaults to "0.0.0.0"


nhrones:
**Describe the bug** 
In server.ts, the function serve() unconditionally returns addr "0.0.0.0"
This is incorrect for Windows and shows up in many Deno code examples and libs.

**Expected behavior** 
Should return "127.0.0.1" on Windows

OS: Windows 10

**Additional context** 

deno_std/http/server.ts  line# 265
```js
export function _parseAddrFromStr(addr: string): HTTPOptions {
  let url: URL;
  try {
    //HACK add this line to return a correct default addr
    const defaultAddr = (Deno.build.os === "windows") ? "127.0.0.1" : "0.0.0.0";
    const host = addr.startsWith(":") ? `${defaultAddr}${addr}` : addr;
    url = new URL(`http://${host}`);
  }
```

cmorten:
Guessing this was written following precedent set by Node (REF: https://github.com/nodejs/node/blob/master/lib/net.js#L126, https://nodejs.org/docs/latest-v10.x/api/net.html#net_server_listen_port_host_backlog_callback) and golang (_I_ believe this is what golang is doing, but find tracing through the stack of the net package tough going - happy to be corrected!)?

Be curious to know why the meta-address was chosen over the loopback address given windows caveats?

For consistency I wonder if opting for loopback only on windows, and all network interfaces otherwise is a good choice? Would it be better to use the same default across all OS (whichever default _is_ chosen).

nhrones:
@cmorten  I was under the impression that RY had a reason to default to '0.0.0.0' . That is why I suggested conditioning default based on build.os.   Also thanks for all the work on sever.ts.

cmorten:
Doesn’t seem to be much history to it - REF: https://github.com/denoland/deno/pull/6698.

Not a windows user - is the issue that serving from `0.0.0.0` doesn’t work at all, or that requesting the `0.0.0.0` meta-address from a client doesn’t resolve, but if hit `localhost` or `127.0.0.1` in the request it still works?

MyIsaak:
Hi guys. I've submitted a PR fixing this issue here: https://github.com/denoland/deno_std/pull/1178

nhrones:
[...]

PR #1178, opened 2021-08-29
fix(http/server): addr defaults to 127.0.0.1 on Windows, else 0.0.0.0 (#1165)

MyIsaak:
# Problem

When using `_parseAddrFromStr(addr: string)` the default hostname address defaults to `0.0.0.0`, which works on Mac/Linux, but not on Windows as described here: #1165. This problem affects the majority of developers (41.2%) that use Windows as their main OS, given the Stack Overflow Survey 2021. The current workaround for Window users is to manually edit their hosts file: `C:\Windows\System32\Drivers\etc\hosts`.

# Solution

This PR contains a fix for address defaults by adding a simple if statement checking whether the OS is windows and if so use `127.0.0.1` (which works on Windows) instead of `0.0.0.0`. This slightly improves the developer experience for Window users.

I've also updated the comments and the test for `_parseAddrFromStr(addr: string)`.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -59,7 +59,8 @@ export type Handler = (
 ) => Response | Promise<Response>;
 
 /**
- * Parse an address from a string.
+ * Parse an address from a string. When only the port is given, the hostname
+ * defaults to 127.0.0.1
  *
  * ```ts
  * import { _parseAddrFromStr } from "https://deno.land/std@$STD_VERSION/http/server.ts";
@@ -78,7 +79,7 @@ export function _parseAddrFromStr(
   addr: string,
   defaultPort = HTTP_PORT,
 ): Deno.ListenOptions {
-  const host = addr.startsWith(":") ? `0.0.0.0${addr}` : addr;
+  const host = addr.startsWith(":") ? `127.0.0.1${addr}` : addr;
 
   let url: URL;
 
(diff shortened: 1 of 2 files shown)


http/server.ts
```diff
@@ -255,17 +255,19 @@ export class Server implements AsyncIterable<ServerRequest> {
...
  *     parseAddrFromString(addr);
  *
- * @param addr Address string
+ * @param addr Address string.
  */
 export function _parseAddrFromStr(addr: string): HTTPOptions {
   let url: URL;
   try {
-    const host = addr.startsWith(":") ? `0.0.0.0${addr}` : addr;
+    const defaultAddr = (Deno.build.os === "windows") ? "127.0.0.1" : "0.0.0.0";
```
cmorten, line 269: Consider consistency across OS, e.g.

```suggestion
    const defaultAddr = "127.0.0.1";
```

Regardless of the default strategy used, it would be good to note this in the README.md for the http module, and probably also in the jsdocs (for Deno doc) for the public method(s) that call this private function 😊

MyIsaak: I like this idea more because it's simpler. I've described why in https://github.com/denoland/deno_std/issues/1165

MyIsaak:
Ok I've made another change, considering the review but also changing the doc and test.

cmorten:
@MyIsaak with the merge of the "native" http serve changes there are some merge conflicts. The parse address functionality is much the same, so the changes for this PR will be much the same, just move around a bit!

MyIsaak:
I've rebased my PR from the HEAD of the main branch to fix the merge conflicts. I still get some failed test results on my machine, but at least the tests in scope of this PR all pass.

cmorten:
Comment just to note that the OP's issue #1165 has been closed.

kt3k:
@MyIsaak Thank you for your contribution anyways!

We prefer to keep listenening 0.0.0.0, but I'm going to update the API docs of Deno CLI to improve the situation.

PR #1179, opened 2021-08-30
feat(collections): adds takeLastWhile

grian32:
Not sure if it's possible to reverse the original array/get the elements going from the end while not copying the array so it doesn't mutate.


LionC:
I dont think you should copy **and** reverse the whole array just to use the same imolementation `takeWhile` uses. That takes up more memory than needed.

I would just do a simple for loop starting from the end counting back, pushing until it hits a non-matching element as a baseline and go from there.

grian32:
> I dont think you should copy **and** reverse the whole array just to use the same imolementation `takeWhile` uses. That takes up more memory than needed.
> 
> I would just do a simple for loop starting from the end counting back, pushing until it hits a non-matching element as a baseline and go from there.

I mostly tried to use the same impl as `takeWhile` because it was pretty performant, I'll see what I can do though.

grian32:
I was able to to do this without copying and reversing the original array @LionC , I needed to use a more naive implementation though. Surprisingly it did not affect performance. Ready for review I think.

LionC:
Looking at the implementation now with not-as-tired eyes, I think this might actually be faster if you use `collections/findLastIndex` and then `slice`, as it will allocate the new array only once instead of reallocating on every push.

Sorry for the back and forth here :-)

grian32:
> Looking at the implementation now with not-as-tired eyes, I think this might actually be faster if you use `collections/findLastIndex` and then `slice`, as it will allocate the new array only once instead of reallocating on every push.

I don't think I fully understand what you mean but by toying around with an implementation using `collections/findLastIndex`, what tests I can get to pass(which is most of them) don't really have a visible perf improvement.

LionC:
> > Looking at the implementation now with not-as-tired eyes, I think this might actually be faster if you use `collections/findLastIndex` and then `slice`, as it will allocate the new array only once instead of reallocating on every push.
> 
> I don't think I fully understand what you mean but by toying around with an implementation using `collections/findLastIndex`, what tests I can get to pass(which is most of them) don't really have a visible perf difference(other than maybe being 1-2ms slower).

What I meant is that the current implementation `.push()`es into the array, causing it to reallocate several times to make space for new elements. It also needs to iterate the whole array by calling `.reverse()` in the end.

If you use `findLastIndex()`, you can leverage `.slice()` instead, which is optimized by v8 and can pretty much just copy the memory directly once. But that is obviously a question of benchmarks in the end :-)

grian32:
> What I meant is that the current implementation `.push()`es into the array, causing it to reallocate several times to make space for new elements. It also needs to iterate the whole array by calling `.reverse()` in the end.
> 
> If you use `findLastIndex()`, you can leverage `.slice()` instead, which is optimized by v8 and can pretty much just copy the memory directly once. But that is obviously a question of benchmarks in the end :-)

Yeah so you're saying to use `findLastIndex()` to get the indexes to slice at, I tried to do that and could get a somewhat working implementation although it failed horribly at returning empty array when predicate matched nothing/same array when predicate matched everything

collections/take_last_while.ts
```diff
@@ -0,0 +1,34 @@
...
+
+  for (let i = length - 1; i >= 0; i--) {
+    if (predicate(array[i])) {
+      newArray.push(array[i]);
+    } else {
+      break;
+    }
+  }
+
+  return newArray.reverse();
```
evanwashere, line 33:
```suggestion
  let offset = array.length;
  while (0 < offset && predicate(array[offset - 1])) offset--;

  return array.slice(offset, array.length);
```

evanwashere (changes requested):

collections/take_last_while.ts
grian32: Thanks for this!

LionC: Isn't

```ts
let offset = array.length;
while (0 < offset && predicate(array[offset - 1])) offset--;
```

pretty much `findLastIndex()`, which we already have implemented in the module?

LionC: Update:

Doing some spontanous benchmarking, I can not find any significant stable performance difference between

```ts
  const index = findLastIndex(array, (it) => !predicate(it))

  if (index === undefined) {
      return []
  }

  return array.slice(index + 1);
```

and

```ts
  let offset = array.length;
  while (0 < offset && predicate(array[offset - 1])) offset--;

  return array.slice(offset, array.length);
```

I guess either is fine, I just want to emphasize that if we can DRY code up without a significant impact, we probably should for maintenance reasons.

collections/README.md
```diff
@@ -527,6 +527,25 @@ assertEquals(
 );
 ```
 
+### takeLastWhile
+
+Returns all elements in the given collection until the first element that does
+not match the given predicate.
```
kt3k, line 533: I think we need to mention that this function collect elements 'from the back'

collections/README.md
```diff
@@ -527,6 +527,25 @@ assertEquals(
 );
 ```
 
+### takeLastWhile
+
+Starting from the last element, returns all elements in the given collection
```
LionC, line 532: This does sound like the order would be reversed, but it actually isnt

collections/take_last_while.ts
```diff
@@ -0,0 +1,27 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+/**
+ * Starting from the last element, returns all elements in the given collection
+ * until the first element that does not match the given predicate.
```
LionC, line 4: Same here obv

LionC (changes requested):

collections/README.md
grian32: What about `Returns all elements in the given collection until the last element that does not match the given predicate.`, its the equivalent of what I have in the `dropLastWhile` PR, @kt3k

LionC: But that is not correct - it returns all element in the given array (we should not use the word collection here) **after** the last element that does not match the given predicate.

But I dont know if that is the most elegant way to phrase it.

grian32: I don't really have any better ideas so I may just go with `Returns all elements in the given array after the last element that does not match the given predicate.`

collections/README.md
```diff
@@ -512,8 +512,8 @@ console.assert(
 
 ### takeWhile
 
-Returns all elements in the given collection until the first element that does
-not match the given predicate.
+Returns all elements in the given array after the last element that does not
```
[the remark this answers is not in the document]
grian32: Yeah looks like it, I think I've reverted it, what I get for commiting stuff without checking

LionC (approved):
LGTM, pinging @kt3k

kt3k (approved):
@grian32 LGTM. Thanks!

PR #1181, opened 2021-08-30
fix(node/util): format no longer adds `"` around formatted strings.

### LINKED ISSUE #1087: 🐛 `util.format(...)` ~ Deno (std@0.103.0) output does not match Node (v14)
```
# NOTE: same output in both Windows and Linux/WSL
$ node -v
v14.15.1
$ deno --version
deno 1.12.2 (release, ...) # both x86_64-pc-windows-msvc AND x86_64-unknown-linux-gnu
v8 9.2.230.14
typescript 4.3.5
$ node -e "const Util = require('util'); console.log(Util.format('test %s this', 'thing', 'suffix'))"
test thing this suffix
$ deno eval -T "import * as Util from 'https://deno.land/std@0.103.0/node/util.ts'; console.log(Util.format('test %s this', 'thing', 'suffix'))"
test "thing" this
```

Specifically, here, it is is wrongly quoting the string inserted into the interpolation and ignoring trailing arguments.

grian32:
This aligns that part of format with node, see this by running:

```bash
node -e "const Util = require('util'); console.log(Util.format('test %s this', 'thing'))"
```
and
```bash
deno eval -T "import * as Util from 'https://raw.githubusercontent.com/grian32/deno_std/util-fmt-fix/node/util.ts'; console.log(Util.format('test %s this', 'thing'))"
```

Partially fixes, but does not close #1087

### DIFF
--- a/node/util_test.ts
+++ b/node/util_test.ts
@@ -296,12 +296,12 @@ Deno.test("[util] format", () => {
     [10, 11],
   ], { hi: "hello" });
   const expected =
-    '[10, hi, null, undefined, NaN, Infinity, { hello: world }, [10, 11], hi: hello]';
+    "[10, hi, null, undefined, NaN, Infinity, { hello: world }, [10, 11], hi: hello]";
   assertEquals(util.format("%o", testData), expected);
   assertEquals(util.format("%O", testData), expected);
 
   const expected2 =
-    '[10, hi, null, undefined, NaN, Infinity, [Object], [Array], hi: hello]';
+    "[10, hi, null, undefined, NaN, Infinity, [Object], [Array], hi: hello]";
   assertEquals(util.format("%s", testData), expected2);
 
   assertEquals(
--- a/node/util_test.ts
+++ b/node/util_test.ts
@@ -296,12 +296,12 @@ Deno.test("[util] format", () => {
     [10, 11],
   ], { hi: "hello" });
   const expected =
-    '[10, "hi", null, undefined, NaN, Infinity, { hello: "world" }, [10, 11], hi: "hello"]';
+    '[10, hi, null, undefined, NaN, Infinity, { hello: world }, [10, 11], hi: hello]';
   assertEquals(util.format("%o", testData), expected);
   assertEquals(util.format("%O", testData), expected);
 
   const expected2 =
-    '[10, "hi", null, undefined, NaN, Infinity, [Object], [Array], hi: "hello"]';
+    '[10, hi, null, undefined, NaN, Infinity, [Object], [Array], hi: hello]';
   assertEquals(util.format("%s", testData), expected2);
 
   assertEquals(
--- a/node/util.ts
+++ b/node/util.ts
@@ -154,7 +154,7 @@ function circularRemover(): (key: string, value: unknown) => unknown {
 }
 
 function formatString(str: string) {
-  return `"${str.replace(/\\/, "\\\\").replace(/"/g, '\\"')}"`;
+  return `${str.replace(/\\/, "\\\\").replace(/"/g, '\\"')}`;
 }
 
 function thingToString(


node/util_test.ts
```diff
@@ -296,12 +296,12 @@ Deno.test("[util] format", () => {
...
   ], { hi: "hello" });
   const expected =
-    '[10, "hi", null, undefined, NaN, Infinity, { hello: "world" }, [10, 11], hi: "hello"]';
+    "[10, hi, null, undefined, NaN, Infinity, { hello: world }, [10, 11], hi: hello]";
   assertEquals(util.format("%o", testData), expected);
   assertEquals(util.format("%O", testData), expected);
 
   const expected2 =
-    '[10, "hi", null, undefined, NaN, Infinity, [Object], [Array], hi: "hello"]';
+    "[10, hi, null, undefined, NaN, Infinity, [Object], [Array], hi: hello]";
```
kt3k, line 304: I tested with the following Node.js script.
```js
const util = require("util");
const x = Object.assign([
  10,
  "hi",
  null,
  undefined,
  NaN,
  Infinity,
  { hello: "world" },
  [10, 11],
], { hi: "hello" });
console.log(util.format("%o", x));
```

and the result was:
```
[
  10,
  'hi',
  null,
  undefined,
  NaN,
  Infinity,
  { hello: 'world' },
  [ 10, 11, [length]: 2 ],
  [length]: 8,
  hi: 'hello'
]
```
(Node.js 16.6)

So in this case, these (`hi` and `hello`) should be quoted. (So the PR has the opposite effect)

kt3k (commented):
@grian32 Thank you for looking into this, but I found a problem in the test case.

node/util_test.ts
grian32: Oh that's true, I'll look into fixing it and adding a specific test case just for strings.

bartlomieju: @grian32 any chance we could include this fix for 0.107.0 next week?

grian32: Probably @bartlomieju , I've just been busy the past week or so getting a new PC toghether.

grian32:
There seem to be some more general issues with `deno_std`'s  implementation of `format`, along with some things being missing, although those are generally out of the scope of this PR, I will make an issue.

rivy:
@grian32 , thanks for taking some time to attempt a partial fix for #1087. If you're interested in trying to deliver a more comprehensive fix, I've generated some code that fully fixes #1087, is compatible with node's `format()` argument usage, and generates output that's a *very close* mimic to node's `format()` function.

https://github.com/rivy/deno.dxx/blob/6d52c233580aad138fc4f5e43726cfee12cae135/tests/%24shared.ts#L5-L91

I don't have the bandwidth to submit a PR at the moment. You're welcome to run with this code if you're interested in creating the full PR/tests/etc.

bartlomieju:
@grian32 do you want to apply @rivy's suggestion in this PR?

rivy:
@grian32 , here's my most updated version with some minor revisions to improve clarity/readability...

```typescript
// ToDO: [2021-09-16; rivy] * improved equivalency to NodeJS format/inspect string quoting requires changing the preference expressed [here](https://github.com/denoland/deno/blob/5d814a4c244d489b4ae51002a0cf1d3c2fe16058/ext/console/02_console.js#L648-L669)

// ref: <https://nodejs.org/api/util.html#util_util_format_format_args>
function toSpecFormat(specifier: string, value: unknown): string {
	if (specifier === '%s') {
		if (typeof value === 'string' || value instanceof String) {
			return value as string;
		} else return Deno.inspect(value, { depth: 2 });
	}
	if (specifier === '%d') {
		if (typeof value === 'bigint') {
			return value + 'n';
		}
		return Number(value).toString();
	}
	if (specifier === '%i') {
		if (typeof value === 'bigint') {
			return value + 'n';
		}
		return parseInt(value as string).toString();
	}
	if (specifier === '%f') {
		return parseFloat(value as string).toString();
	}
	if (specifier === '%j') {
		try {
			return JSON.stringify(value);
		} catch (e) {
			// nodeJS => 'cyclic object value' , deno => 'Converting circular structure to JSON ...'
			// ref: <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify>
			if (e instanceof TypeError && e.message.match(/cyclic|circular/)) {
				return '[Circular]';
			} else throw e;
		}
	}
	if (specifier === '%o') {
		return Deno.inspect(value, { showHidden: true, showProxy: true });
	}
	if (specifier === '%O') {
		return Deno.inspect(value);
	}
	if (specifier === '%c') {
		return '';
	}
	return '';
}

// ref: <https://nodejs.org/docs/latest-v16.x/api/console.html#console_console_log_data_args>
// ref: <https://nodejs.org/docs/latest-v16.x/api/util.html#util_util_format_format_args>
// modified from <https://deno.land/std@0.105.0/node/util.ts#L247-L266>
function format(...args: unknown[]) {
	const replacement: [number, string][] = [];
	const formatSpecifierRx = /%(s|d|i|f|j|o|O|c|%)/g;
	const hasFormatTemplate = args.length > 0 &&
		(typeof args[0] === 'string' || args[0] instanceof String);
	const formatTemplate = hasFormatTemplate ? (args[0] as string) : '';
	let i = hasFormatTemplate ? 1 : 0;
	let arr: RegExpExecArray | null = null;
	let done = false;
	while ((arr = formatSpecifierRx.exec(formatTemplate)) !== null && !done) {
		if (arr[0] === '%%') {
			replacement.push([arr['index'], '%']);
		} else if (i < args.length) {
			replacement.push([arr['index'], toSpecFormat(arr[0], args[i])]);
			i++;
		} else done = true;
	}
	const lastArgUsed = i;
	let result = '';
	let last = 0;
	for (let i = 0; i < replacement.length; i++) {
		const item = replacement[i];
		result += formatTemplate.slice(last, item[0]);
		result += item[1];
		last = item[0] + 2;
	}
	result += formatTemplate.slice(last);
	for (let i = lastArgUsed; i < args.length; i++) {
		if (i > 0) result += ' ';
		if (typeof args[i] === 'string') {
			result += args[i];
		} else result += Deno.inspect(args[i], { colors: true });
	}
	return result;
}
```

rivy:
@bartlomieju @grian32 , I've noted in the comment leading the code that to create better parity with the NodeJS `format()` function, the decision to use double-quotes in priority over single-quotes (when quoting strings for `format()` and `inspect()`) will need to be reversed within the [`deno` console code](https://github.com/denoland/deno/blob/5d814a4c244d489b4ae51002a0cf1d3c2fe16058/ext/console/02_console.js#L648-L669).

There's no explanation given at that point about why double quotes were chosen. The commentary just says that "we prefer double quotes as the default". Given the decreased visual noise of single quotes (especially when reading code), I don't understand the choice.

PR #1184, opened 2021-08-30
feat(collections): add `includesValue`

grian32:


collections/README.md
```diff
@@ -509,3 +509,19 @@ console.assert(
   ],
 );
 ```
+
+### includesValue
+
+Returns whether the given value is part of the collection.
```
LionC, line 515: We should change that to something meaningful. Imagine knowing nothing about this function and reading this for the first time.

collections/includes_value.ts
```diff
@@ -0,0 +1,22 @@
+/**
+ * Returns whether the given value is part of the collection.
```
LionC, line 2: Same here

collections/includes_value.ts
```diff
@@ -0,0 +1,22 @@
...
+ * const input = {
+ *   first: 33,
+ *   second: 34,
+ * };
+ *
+ * assert(includesValue(input, 34));
+ */
+
+export function includesValue<T>(
+  collection: Record<string, T>,
```
LionC, line 18: We should rename the argument to `record` or something more specific. Also marking as `readonly` here is good practice to signal that no mutation will happen.

collections/includes_value.ts
```diff
@@ -0,0 +1,22 @@
...
+ * ```ts
+ * import { includesValue } from "./includes_value.ts";
+ * import { assert } from "../testing/asserts.ts";
+ *
+ * const input = {
+ *   first: 33,
+ *   second: 34,
+ * };
+ *
+ * assert(includesValue(input, 34));
```
LionC, line 14: I get the idea, but I think sticking with `assertEquals` for consistency and to clearly show the returned value is preferable, especially because the current version works for all truthy values.

collections/includes_value.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+export function includesValue<T>(
+  collection: Record<string, T>,
+  value: T,
+): boolean {
+  return Object.values(collection).includes(value);
```
LionC, line 21: We should be better on memory and runtime than that, as `Object.values` needs to get **all** values and needs to allocate a whole array of all of them, just to do an `.includes` here. You should be able to do that by hand with `O(1)` memory and the same or better runtime.

collections/includes_value_test.ts
```diff
@@ -0,0 +1,44 @@
+import { includesValue } from "./includes_value.ts";
+import { assert, assertEquals } from "../testing/asserts.ts";
+
```
LionC, line 3: I would feel more comfortable with some more test cases, maybe around non-primitive types, non-enumerable props etc without claiming that we did that everywhere else. The test file is just really short right now.

LionC (changes requested):

collections/includes_value.ts
grian32: I don't believe we can do `readonly` for `Record`s, getting this:
`'readonly' type modifier is only permitted on array and tuple literal types.`

collections/README.md
grian32: I'm not the best at documentation in all honesty, how does `If the given value is part of the given object it returns true, otherwise it returns false.` sound?

LionC: You can do `Readonly<Record<>>`

collections/includes_value.ts
LionC: You could do `Readonly<Record<>>`

collections/README.md
LionC: Returns whether the given value is in the given record's values ?

grian32: I've pushed the above but will leave this unresolved for further discussion.

collections/includes_value.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * const input = {
+ *   first: 33,
+ *   second: 34,
+ * };
+ *
+ * assertEquals(includesValue(input, 34), true);
+ */
+
+export function includesValue<T>(
```
littledivy, line 17:
```typescript
includesValue({ k: NaN }, NaN); // gives false
```
Consider using `Object.is` instead? Or document this behavior :-)

littledivy (changes requested):

majidsajadi:
@LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?

grian32:
> @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?

It shouldn't work with non-primitive values, thats what the test checks for aswell.

collections/README.md
```diff
@@ -509,3 +509,20 @@ console.assert(
   ],
 );
 ```
+
+### includesValue
+
+If the given value is part of the given object it returns true, otherwise it
```
littledivy, line 515: Document non-primitive behavior?

littledivy (changes requested):

LionC:
> @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?

I think it is reasonable to mimic native `includes` behaviour, as we have done in all other similar functions in the module. So no magic deepequals stuff, just normal `===` checks. I just meant @grian32 should check for that in the tests.

majidsajadi:
> > @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?
> 
> It shouldn't work with non-primitive values, thats what the test checks for aswell. Leon can explain in more detail.

if it does not suppose to handle non-primitive values I guess the type definition must be more accurate.
`Record<string, T>` 

also, I'm not understanding why method `includesValue` should not check for non-primitive values. any reason for this decision? @LionC

collections/includes_value.ts
grian32: I've added a check for NaN values using `Number.isNan()` and a test for it.

grian32:
I've documented this behaviour in the JSDoc and the README.md  @majidsajadi

majidsajadi:
> > @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?
> 
> I think it is reasonable to mimic native `includes` behaviour, as we have done in all other similar functions in the module. So no magic deepequals stuff, just normal `===` checks. I just meant @grian32 should check for that in the tests.

if `includesValue` suppose to works exactly the same as `includes` what's the point of adding this one-liner? the reason we closed #1168 was not to implement trivial functions. right? @lucacasonato

```javascript
includesValue(input, someValue);

Object.values(input).includes(someValue);
```

grian32:
> > > @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?
> > 
> > 
> > I think it is reasonable to mimic native `includes` behaviour, as we have done in all other similar functions in the module. So no magic deepequals stuff, just normal `===` checks. I just meant @grian32 should check for that in the tests.
> 
> if `includesValue` suppose to works exactly the same as `includes` what's the point of adding this one-liner? the reason we closed #1168 was not to implement trivial functions. right? @lucacasonato
> 
> ```js
> includesValue(input, someValue);
> 
> Object.values(input).includes(someValue);
> ```

Please check the current implementation before responding,  it is more performant memory and speed wise than `Object.values(input).includes(value)`

collections/includes_value.ts
grian32: The typings seem to work out using `Readonly<Record<string, T>>`, I will push that.

LionC:
> > > @LionC @grian32 this implementation is not working with non-primitive values. and according to the type definition and the function description we need to support non-primitve values?
> > 
> > 
> > I think it is reasonable to mimic native `includes` behaviour, as we have done in all other similar functions in the module. So no magic deepequals stuff, just normal `===` checks. I just meant @grian32 should check for that in the tests.
> 
> if `includesValue` suppose to works exactly the same as `includes` what's the point of adding this one-liner? the reason we closed #1168 was not to implement trivial functions. right? @lucacasonato
> 
> ```js
> includesValue(input, someValue);
> 
> Object.values(input).includes(someValue);
> ```

`Object.values` needs to go through everything and allocate a whole new array for it, only to go through it again for `includes` (see one of the first comments in here were I said indeed that this should not be the implementation or otherwise we would not include it).

This implementation here has O(1) memory, which is a huge scaling difference.

collections/README.md
grian32: > Returns whether the given value is in the given record's values ?

Hmm I guess we could do that but imo what is currently rn is more detailed, i guess i could change `object` to `record` but it doesn't make much difference.

LionC (approved):
LGTM now, pinging @kt3k

kt3k (approved):
@grian32 LGTM. Thank you for your contribution!

PR #1185, opened 2021-08-30
feat(collections): add dropWhile

getspooky:
Related: https://github.com/denoland/deno_std/issues/1173


LionC:
Would you mind changing the title to a simple `feat(collections): add dropWhile`? That would help with overview, as there is lots of small PRs adding functions.

I also think this function should be named `dropWhile` to match up with `takeWhile` :-)

getspooky:
@LionC sure i will implement this changes , I think something went wrong with `CI/CD`

LionC:
`dropWhile` vs `dropLastWhile` is in line with `takeWhile` and `takeLastWhile`, which is in line with the current ES iterator proposal naming for `take` and with kotlin, which set most of the other names.

getspooky:
@LionC could you please review this PR

collections/README.md
```diff
@@ -112,6 +112,20 @@ const distinctNumbers = distinct(numbers);
 console.assert(distinctNumbers === [3, 2, 5]);
 ```
 
+### dropWhile
+
+Drop all elements in the given collection until the first element that does not
+match the given predicate.
```
LionC, line 118: This should make clear that it does not touch the original array. We should also not use the word "collection" - it just takes an Array right now :-)

collections/drop_while.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Drop all elements in the given collection until the first element that does not match the given predicate.
+ *
```
LionC, line 5: Same here obv

collections/drop_while.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Example:
+ *
+ * ```ts
+ * import { dropWhile } from "./drop_first_while.ts";
+ * import { assertEquals } from "../testing/asserts.ts";
+ *
+ * const numbers = [ 3, 2, 5, 2, 5 ]
+ * const dropWhileNumbers = dropWhile(numbers, i => i !== 2)
+ *
+ * assertEquals(dropWhileNumbers, [ 5, 2, 5 ])
```
LionC, line 15: That examples looks wrong. Shouldn't this be `[ 2, 5, 2, 5 ]`?

collections/drop_while.ts
```diff
@@ -0,0 +1,30 @@
...
+  predicate: (el: T) => boolean,
+): T[] {
+  let offset = 0;
+  const length = array.length;
+
+  while (length > offset && predicate(array[offset])) {
+    offset++;
+  }
+
+  return array.slice(offset + 1, length);
```
LionC, line 29: I think the `+ 1` is actually a bug isnt it?

I would expect `dropWhile` to drop everything that fulfils the predicate. This actually just drops one more element.

collections/drop_while_test.ts
```diff
@@ -0,0 +1,57 @@
...
+  assertEquals(actual, []);
+});
+
+Deno.test("[collections/dropWhile] Returns the same array when all elements match the predicate", () => {
+  const arr = [1, 2, 3, 4];
+
+  const actual = dropWhile(arr, (i) => i !== 400);
+
+  assertEquals(actual, []);
+});
```
LionC, line 57: The tests seem to test the buggy behaviour mentioned above.

LionC (changes requested):
I think the current implementation does not behave as expected -- see the comments

collections/drop_while.ts
getspooky: As i know we should not include predictable number,

LionC: I don't think I understand what you mean.

`dropWhile` should drop all elements from the start that match the given predicate, until it hits one that doesnt. The one that doesnt should not be dropped, as it does not match the predicate.

collections/drop_while.ts
getspooky: fixed

collections/drop_while.ts
getspooky: @LionC fixed

LionC (approved):
LGTM! Pinging @kt3k

kt3k (approved):
@getspooky LGTM. Thanks!

PR #1186, opened 2021-08-30
feat(collections): add filterValuesNotNullish

majidsajadi:


LionC:
I think we should use `filterValues` here - there is no additional overhead, the function mostly provides correct typing and a correct predicate. Also noting that this might be parallel to `filterNotNullish` which was refused.

LionC:
> I think we should use `filterValues` here - there is no additional overhead, the function mostly provides correct typing and a correct predicate. Also noting that this might be parallel to `filterNotNullish` which was refused.

Looking at this again, I do not think we can include this function right now, as it is really just

```ts
filterValues(array, it => it !== undefined && it !== null)
```

except the typings - but that goes for `filterNotNullish` as well, which was refused. I would vote for closing this for now @kt3k , but thank you for the contribution anyways @majidsajadi .

However, I think we should add an overload to `filterValues` to pass a type guard that adapts the resulting type, allowing users to do the above snippet in a type safe way like they would with `Array.filter()`.

kt3k:
I agree with @LionC. Let's close this one as filterValues already covers it.

@majidsajadi Thank you for your contribution anyway!

PR #1188, opened 2021-08-30
feat(collections): add aggregateGroups

majidsajadi:


majidsajadi:
@LionC Review this, please. if the implementation is okay I can add docs and tests.

collections/aggregate_groups.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * assetEquals(descriptions === {
+ *     'Curry': 'Curry is spicy and vegan',
+ *     'Omelette': 'Omelette is creamy and vegetarian',
+ * })
+ * ```
+ */
+export function aggregateGroups<T, A>(
+  record: Record<string, Array<T>>,
+  aggregator: (accumulator: A, current: T, key: string, first: boolean) => A,
```
LionC, line 32: The kotlin version has an optional parameter for current which is not set for the first element. I think trabslating that to

```ts
aggregator: (key: string, first: boolean, accumulator: A, current?: T) => A
```

might be appropiate.

collections/aggregate_groups.ts
```diff
@@ -0,0 +1,46 @@
...
+ * )
+ *
+ * assetEquals(descriptions === {
+ *     'Curry': 'Curry is spicy and vegan',
+ *     'Omelette': 'Omelette is creamy and vegetarian',
+ * })
+ * ```
+ */
+export function aggregateGroups<T, A>(
+  record: Record<string, Array<T>>,
```
LionC, line 31: We should mark this as read only

collections/aggregate_groups.ts
```diff
@@ -0,0 +1,46 @@
...
+  aggregator: (accumulator: A, current: T, key: string, first: boolean) => A,
+  initalValue: A,
```
LionC, line 33: I think we actually don't need this with the aggregator signature

collections/aggregate_groups.ts
```diff
@@ -0,0 +1,46 @@
...
+): Record<string, A> {
+  const result: Record<string, A> = {};
+
+  for (const [key, values] of Object.entries(record)) {
+    result[key] = values.reduce(
+      (accumulator, current, currentIndex) =>
+        aggregator(accumulator, current, key, currentIndex === 0),
+      initalValue,
+    );
+  }
```
LionC, line 43: Thos would beed to be adapted to the signature of course.

I also think you can use `mapEntries` here without any significant additional overhead.

collections/aggregate_groups.ts
```diff
@@ -0,0 +1,46 @@
...
+ *     'Curry': [ 'spicy', 'vegan' ],
+ *     'Omelette': [ 'creamy', 'vegetarian' ],
+ * }
+ * const descriptions = aggregateGroups(foodProperties,
+ *     (acc, current, key, first) => {
+ *         if (first)
+ *             return `${key} is ${current}`
+ *
+ *         return `${acc} and ${current}`
+ *     },
```
LionC, line 20: Would need ti be adapted too obv

LionC (changes requested):
I thibk we could actually use the kotlin signature directly - see my review

collections/aggregate_groups.ts
majidsajadi: what is the benefit of removing `current` for the first `element` of the `array`?

collections/aggregate_groups.ts
majidsajadi: I don't get it. if we don't want to accept the initial value as an argument what should we pass to the reducer as the initial value?

collections/aggregate_groups.ts
LionC: Oh sorry - I switched the parameters^^ The accumulator is not set for the first element, because nothing has been accumulated yet :-)

My bad

collections/aggregate_groups.ts
LionC: The aggregator gets the first element (and first = true) for the first call

bartlomieju:
I'm still interested in landing this PR; @majidsajadi @LionC do you think we could get the PR rebased and add the docs?

LionC:
Sure, I will make it happen myself or find someone to do it tomorrow^^

LionC:
@bartlomieju @majidsajadi I have check out and continued this branch, fixing the comments and adding tests. See the linked PR above.

bartlomieju:
Let's continue in https://github.com/denoland/deno_std/pull/1553

PR #1189, opened 2021-08-30
chore: update for canary

kt3k:
Continued from #1171.

This PR disables several test cases to pass the CI.

This also excludes some file in `deno test` command(, but this change probably can be reverted when https://github.com/denoland/deno/pull/11878 is merged)


kt3k:
closing in favor of #1171

PR #1191, opened 2021-08-30
feat(collections): add slidingWindows

ayame113:
for #1173

- [x] Did you support importing from mod.ts as well as your functions file?
- [x] Have you made sure to allocate as little memory and do as little steps in your algorithm as possible?
- [x] Did you add/adapt JSDoc comments, add/adapt an example, made sure that the example uses the ts markdown tag and assertEquals to show the result?
- [x] Did you add/adapt your functions documentation in README.md?
- [x] Did you make sure not to mutate the arguments passe to your function?
- [x] Did you add tests ensuring no mutation, empty input and corner cases?
- [x] Are your types flat, meaning there is no unnecessary alias in them that makes your users "go to type definition" twice to understand what to pass?


collections/windowed.ts
```diff
@@ -0,0 +1,75 @@
...
+     */
+    partial?: boolean;
+  } = {},
+): T[][] {
+  if (
+    typeof size !== "number" || typeof step !== "number" ||
+    size <= 0 || step <= 0 || Number.isNaN(size) || Number.isNaN(step)
+  ) {
+    throw new RangeError("Both size and step must be positive.");
+  }
```
LionC, line 66: I think you can replace all of this with a `Number.isInteger()` check for both values, which is shorter, probably more optimised and also checks that it is an integer.

collections/windowed.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /** length of the return array */
+  const length = Math.floor((array.length - (partial ? 1 : size)) / step + 1);
+
+  return Array.from(
+    { length },
+    (_, i) => array.slice(i * step, i * step + size),
+  );
```
LionC, line 74: I really like this - it is concise, understandable but also not trivial.

collections/windowed.ts
```diff
@@ -0,0 +1,75 @@
...
+  ) {
+    throw new RangeError("Both size and step must be positive.");
+  }
+
+  /** length of the return array */
+  const length = Math.floor((array.length - (partial ? 1 : size)) / step + 1);
+
+  return Array.from(
+    { length },
+    (_, i) => array.slice(i * step, i * step + size),
```
LionC, line 73: Personal nitpick: When too many things happen on one line, it starts to become hard to read. You could definitely split some value (like the ternary) into their own intermediary `const`. That is really just a nitpick though, so don't bother if you do not agree :-)

collections/windowed_test.ts
```diff
@@ -0,0 +1,220 @@
...
+      () => {
+        windowed([1, 2, 3, 4, 5], 3, { step: NaN });
+      },
+      RangeError,
+      "Both size and step must be positive.",
+    );
+    assertThrows(
+      () => {
+        // @ts-ignore: for test
+        windowed([1, 2, 3, 4, 5], "invalid");
```
LionC, line 160: I don't think we need to test for invalid TS calls - if we start doing that, you might as well check every parameter.

collections/windowed_test.ts
```diff
@@ -0,0 +1,220 @@
...
+    assertThrows(
+      () => {
+        windowed([1, 2, 3, 4, 5], NaN);
+      },
+      RangeError,
+      "Both size and step must be positive.",
+    );
+    assertThrows(
+      () => {
+        windowed([1, 2, 3, 4, 5], 3, { step: NaN });
```
LionC, line 152: You could add tests for floats (like `0.5`, `1.2`), `Number.POSITIVE_INFINITY` and `Number.NEGATIVE_INFINITY`

collections/windowed_test.ts
```diff
@@ -0,0 +1,220 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
```
LionC, line 2: Very thorough tests overall - thanks!

LionC (changes requested):
Only minor comments, well done overall :-)

ayame113:
I found it faster to use the for statement and array.push than to use Array.from, so I modified the code accordingly.

Benchmark for arrays of length 10000:
- before -> 1.107292060000005ms
    ```ts
    return Array.from(
      { length },
      (_, i) => array.slice(i * step, i * step + size),
    );
    ```

- after -> 0.36173926999999984ms
    ```ts
    const result = [];
    for (let i = 0; i < length; i++) {
      result.push(array.slice(i * step, i * step + size));
    }
    return result;
    ```

collections/windowed_test.ts
ayame113: If we forgot to throw an error here, the `length` will be `NaN` and we will get the unexpected behavior of returning an empty array. I'd like to leave this test alone to make sure that an error is thrown when passing a non-numeric value.

collections/windowed.ts
ayame113: For now, the result of `deno fmt` is within single line, so I'd like to leave it as it is.

collections/windowed.ts
```diff
@@ -0,0 +1,75 @@
...
+  if (
+    !Number.isInteger(size) || !Number.isInteger(step) || size <= 0 || step <= 0
+  ) {
+    throw new RangeError("Both size and step must be positive integer.");
+  }
+
+  /** length of the return array */
+  const length = Math.floor((array.length - (partial ? 1 : size)) / step + 1);
+
+  const result = [];
```
devcat, line 70:
```suggestion
  const result = new Array(length);
```

ayame113: Proposed Thank you.
Sure, `array.push` (0.36173926999999984ms) is a bit faster than` new Array(length)` (0.33261830000000053ms).
However, according to [this V8 blog post](https://v8.dev/blog/elements-kinds), it seems that V8 can optimize the code better with `array.push` than with` new Array(length)`. I would like to use `array.push` to improve the performance of subsequent code that uses the return value of this function.

kt3k:
@ayame113 Thank you for your contribution! Nice work!

The implementation and tests looks good to me.

But we try to avoid using `adjective` name for public API in general. ref: https://github.com/denoland/deno_std/pull/1098 Can you think of any alternative name for this function?

collections/windowed.ts
devcat: you are pushing arrays into array, primitive value optimization is impossible in this case

LionC: @devcat what @ayame113 is (I think correctly) getting at, is that when you build an Array using the constructor with a length value, you are saving some performance for reallocation when filling it, but that Array is and will forever be markes as `HOLEY_`, which prevents v8 from applying performance optimizations to further operations on it like `.map()`.

I think this is actually really interesting here **and** in all the other functions. We have been doing `new Array(length)` whenever possible, but thinking about it now, that means that we improve the performance of our function a bit, but `return` an unoptimizable `HOLEY_` Array to the user that will perform worse in their code. Specifically for bigger Arrays, I don't think that is desirable, as the reallocation impact is only `O(log n)` and only happens once, while the impact of the user carrying a `HOLEY_` Array through their code happens ok every single further operation on it.

I plan to do a benchmark suite for the whole module as soons as we are API complete, we should investigate that further once that lands.

ayame113:
@kt3k 

I found a discussion that took place when I introduced the windowed function to Kotlin.
https://github.com/Kotlin/KEEP/issues/11

According to it, other names used in other languages or suggested are:
- `slideWindow`
- `subLists` (`subArrays`)
- `eachSlice`
- `collat​​e`
- `partition`

I like `slideWindow`.

collections/windowed.ts
ayame113: @LionC 
Thanks for the explanation, that's what I wanted to say. 👍

LionC:
> @kt3k 
> 
> 
> 
> I found a discussion that took place when I introduced the windowed function to Kotlin.
> 
> https://github.com/Kotlin/KEEP/issues/11
> 
> 
> 
> According to it, other names used in other languages or suggested are:
> 
> - `slideWindow`
> 
> - `subLists` (`subArrays`)
> 
> - `eachSlice`
> 
> - `collat​​e`
> 
> - `partition`
> 
> 
> 
> I like `slideWindow`.

Something with "sliding" is very good I think. But why only one "window" when it returns multiple?

What about `slidingWindows`?

LionC:
I would vote strongly for `slidingWindows` over `slideWindow` - I get the "verb" idea, but trying to understand what I get when reading `slideWindow` is hard for a non-native speaker (or for someone who does not expect a verb). Either way, both would be fine and are mergeable IMO, @kt3k can you make a call?

kt3k:
Thank you @ayame113 for researching the past discussion in kotlin community.

Among 2 candidates, I slightly prefer `slidingWindows` because it exactly describes the returned value. `slideWindow` sounds a little bit obscure about what the returned value is.

ayame113:
> Among 2 candidates, I slightly prefer slidingWindows because it exactly describes the returned value. slideWindow sounds a little bit obscure about what the returned value is.

I updated to use the `slidingWindows`.

kt3k (approved):
@ayame113 Thank you for updating the name 🙏 

LGTM. Nice work!

PR #1192, opened 2021-08-31
feat(collections): add filterValuesNotNullish

grian32:
There are a few things I'd like to bring up regarding this:
- The name is kinda long/wordy, I don't really have any better ideas but if anyone has I'd appreciate if they could share it.
- In the current implementation I'm not sure how to get rid of the typecast, as even tho its as "type-guarded" as can be, TS still doesn't recognize it as NonNullable
- I've followed the nullish coalescing operator, which only considers(to my knowledge) `null` and `undefined` as nullish, I'm not sure if that's a problem/if that's what the function was intended to do originally.

### DIFF
--- a/collections/README.md
+++ b/collections/README.md
@@ -509,3 +509,24 @@ console.assert(
   ],
 );
 ```
+
+### filterValuesNotNullish
+
+Returns a new collection with all entries of the given collection except the
+ones that have a nullish value
+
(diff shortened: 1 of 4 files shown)


grian32:
duplicate of #1186 , wasn't noted on the umbrella issue so i didnt notice, closing.

PR #1195, opened 2021-08-31
chore(ci): Remove canary tests and specify Deno version

### LINKED ISSUE #870: CI should use a locked version of Deno

---

### LINKED ISSUE #1174: Exclude canary CI statuses from merge requirements
Currently we require canary CI statues passing as merge requirement. That means the contributors need to run the test with canary deno, but that would set a too high bar for contribution. I suggest we should exclude canary CI statuses from merge requirements.

nayeemrmn:
Closes #870.
Closes #1174.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,9 +6,10 @@ on:
   pull_request:
     branches: [main]
 
+env:
+  DENO_VERSION: 1.13.2
+
 jobs:
-  env:
-    DENO_VERSION: 1.13.2
   test:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,13 +7,15 @@ on:
     branches: [main]
 
 jobs:
+  env:
+    DENO_VERSION: 1.13.2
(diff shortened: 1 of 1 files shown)


kt3k:
Thanks Nayeem for the PR. I talked with Ryan about this, and we decided to keep running canary CI, but we don't require those CI checks as requirement for merge. I'm trying to clarify that in README (ref: #1209).

Closing without merge. Thank you for the work anyway

nayeemrmn:
@kt3k What about explicitly specifying the CLI version? Right now when we are bumping std, the release CLI version that gets used in CI is one behind the one which the new std version is supposed to be released alongside, since that normally hasn't published yet.

If we explicitly managed and bumped the CLI version used in CI, these problems would be exposed plus it would provide a deterministic point in the release cycle where we are actually supposed to adapt to breaking changes for `--unstable` bindings and TS.

Keeping canary tests is fine as long as we treat them as strictly informative, perhaps a scrappy way of finding unintended CLI breakages that need to be reverted in the CLI repo. But for intended breaking API changes which need to be fixed here, the only viable moment to do that is just after the next CLI version, when we are bumping CI to use that version, just before we release a new accompanying std version.

PR #1196, opened 2021-08-31
fix(ci): ignore `http/bench_legacy.ts` in canary runs

cmorten:
An oversight of #1128 was the introduction of `http/bench_legacy.ts` vs `http/bench.ts` which was missed from the canary test configuration's ignore list, resulting in the http bench server being run and hanging canary jobs in GH actions.

This PR adds the missing `http/bench_legacy.ts` entry.

Will be superseeded by #1195 if it lands first.


kt3k (approved):
@cmorten LGTM. Thanks!

PR #1197, opened 2021-09-01
fix(collections): Prevent prototype merge in collections/deep_merge.ts

petamoriken:
The use of the `in` operator to check for the existence of properties is causing unexpected merging in objects with the same property names as `Object.prototype`. This PR fixes it by changing it to use `Object.prototype.hasOwnProperty`.

Expected behavior:

```ts
assertEquals(
  deepMerge({
    constructor: undefined,
  }, {
    foo: true,
  }),
  {
    constructor: undefined,
    foo: true,
  }
);
```

Actual behavior:

```ts
assertEquals(
  deepMerge({
    constructor: undefined,
  }, {
    foo: true,
  }),
  {
    constructor: Object.prototype.constructor,
    foo: true,
  }
);
```

### DIFF
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -2,9 +2,9 @@
 
 // deno-lint-ignore-file ban-types
 
-import { hasOwnProperty } from "../_util/has_own_property.ts";
-
 import { filterInPlace } from "./_utils.ts";
+
+const { hasOwn } = Object;
 
 /**
  * Merges the two given Records, recursively merging any nested Records with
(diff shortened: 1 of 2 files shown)


collections/deep_merge.ts
```diff
@@ -71,7 +73,7 @@ export function deepMerge<
 
     const a = record[key] as ResultMember;
 
-    if (!(key in other)) {
+    if (!(hasOwnProperty(other, key))) {
```
kt3k, line 76: We're removing `_util/has_own_property.ts` in #1177. Can you use Object.hasOwn instead?

kt3k (commented):
@petamoriken Good catch! Thanks!

Left one comment.

collections/deep_merge.ts
petamoriken: Done!

kt3k (approved):
@petamoriken LGTM. Thank you for your contribution!

PR #1202, opened 2021-09-04
docs(http): adhere to deno style guide in `std/http` module

cmorten:
Following https://github.com/denoland/deno_std/pull/1128, this PR corrects the JSDoc syntax previously implemented to match the [style guide](https://deno.land/manual/contributing/style_guide) (apologies for not getting right first time!) and the supported syntax of <https://doc.deno.land/>.

In addition to correcting the docs in the `http/server.ts` module, I have also corrected/formatted docs in the rest of the `http` module for consistency. I haven't done the whole of std to keep the scope of this PR small and easily reviewable.

Compare [docs from this PR](https://doc.deno.land/https/raw.githubusercontent.com%2Fdenoland%2Fdeno_std%2Fd929a3a%2Fhttp%2Fmod.ts) with [current docs](https://doc.deno.land/https/raw.githubusercontent.com%2Fdenoland%2Fdeno_std%2Fmain%2Fhttp%2Fmod.ts).

### DIFF
--- a/http/_io.ts
+++ b/http/_io.ts
@@ -116,13 +116,15 @@ export function chunkedBodyReader(h: Headers, r: BufReader): Deno.Reader {
   return { read };
 }
 
-function isProhibidedForTrailer(key: string): boolean {
+function isProhibitedForTrailer(key: string): boolean {
   const s = new Set(["transfer-encoding", "content-length", "trailer"]);
   return s.has(key.toLowerCase());
 }
 
-/** Read trailer headers from reader and append values to headers. "trailer"
- * field will be deleted. */
+/**
+ * Read trailer headers from reader and append values to headers. "trailer"
+ * field will be deleted.
+ */
 export async function readTrailers(
   headers: Headers,
   r: BufReader,
@@ -163,7 +165,7 @@ function parseTrailer(field: string | null): Headers | undefined {
   if (trailerNames.length === 0) {
     throw new Deno.errors.InvalidData("Empty trailer header.");
   }
-  const prohibited = trailerNames.filter((k) => isProhibidedForTrailer(k));
+  const prohibited = trailerNames.filter((k) => isProhibitedForTrailer(k));
   if (prohibited.length > 0) {
     throw new Deno.errors.InvalidData(
       `Prohibited trailer names: ${Deno.inspect(prohibited)}.`,
@@ -190,8 +192,10 @@ export async function writeChunkedBody(
   await w.write(endChunk);
 }
 
-/** Write trailer headers to writer. It should mostly should be called after
- * `writeResponse()`. */
+/**
+ * Write trailer headers to writer. It should mostly should be called after
+ * `writeResponse()`.
+ */
 export async function writeTrailers(
   w: Deno.Writer,
   headers: Headers,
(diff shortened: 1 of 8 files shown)


http/_io.ts
```diff
@@ -116,13 +116,15 @@ export function chunkedBodyReader(h: Headers, r: BufReader): Deno.Reader {
...
   const s = new Set(["transfer-encoding", "content-length", "trailer"]);
   return s.has(key.toLowerCase());
 }
 
-/** Read trailer headers from reader and append values to headers. "trailer"
- * field will be deleted. */
+/**
+ * Read trailer headers from reader and append values to headers. "trailer"
+ * field will be deleted.
+ */
```
zhangyuannie, line 127: The style guide is a bit unclear, but it seems like this may actually not be we want.

See https://github.com/denoland/deno/pull/4801#pullrequestreview-408686580, which is posted after the current style guide is written.

Just a note, I'm fine either way :smiley:

cmorten: Ah interesting! Yeah there appears to be conflict between example 3 and 6 in https://deno.land/manual/contributing/style_guide#use-jsdoc-for-exported-symbols regarding placement of the first comment line, there is explicit text to that effect, and also no mention of the closing syntax trailing the bottom line.

If it’s something worthy of a comment then probably worth canonising in the style guide - it would be good if can confirm which style we want (or if multiple, under what conditions!) and happy to raise a PR against the manual @kitsonk / @bartlomieju (as reviewers of that referenced PR)

Option 1 (example 3 in style guide):

```ts
/**
 * Line 1
 * @param param description
 */
```

Option 2 (example 6 in style guide):

```ts
/** Line 1
 * @param param description
 */
```

Option 3 (@kitsonk’s suggestion):

```ts
/** Line 1
 * @param param description */
```

cmorten: Happy to conform either way, be great to standardise/be consistent on style where can.

cmorten: To get a handle on precedent, have done some basic regex counts to see what is currently in std:

```txt
994 JSDoc `/\*\*`

393 single line `/\*\*.*\*/`

140 multi-line, description on first line `/\*\*(?!\s*$).+(?<!\*/)\n`
459 multi-line, description on second line `/\*\*\n.*(?<!\*/)\n`

66 multi-line, closing syntax not on new line `\n\s+\* (?!\s*$).+\*/\n`
510 multi-line, closing syntax on new line `\n\s+\*\/\n`
```

(feel free to critique the regex 😉)

Assuming not completely off - the above hits would suggest that the inherent style of the codebase is to place the description on the second line of the JSDoc, and to placing the closing syntax on it's own line following the end of the tags/description of the JSDoc. This is contrary to the style guide / suggestion above, but makes sense given this is what many IDEs will do by default when adding JSDoc and I suspect is probably the most common way to do JSDoc universally (citation needed).

Regardless, be good to get a decision, bake into the style docs and then will happily update this PR and raise others to have us conform to that style 😄

kt3k (approved):
LGTM. Thanks! @cmorten

http/_io.ts
cmorten: @kt3k do you have any opinions on the above and how JSDoc style should be aligned?

kt3k: I think we don't have a rule when the jsdoc is multiline. Option 1, 2, and 3 look all fine to me.

kt3k: And personally I feel it's a little too strict to enforce one of these styles to contributors/collaborators. I would let is as is.

But if you have any strong opinion among those styles, please feel free to open an issue/suggestion to make a new guideline about it :)

cmorten: Amazing, thanks for your thoughts ☺️

Issue #1203, opened 2021-09-04
Allow Zip and Unzip to work on more than two arrays


Takashiidobe:
Previous discussion for this issue here: https://github.com/denoland/deno_std/discussions/970#discussioncomment-1272939

**Is your feature request related to a problem? Please describe.**
Currently, zip and unzip in std/collections follow this type signature, which means they operate on two arrays in the case of zip or an array of pairs, in the case of unzip.
```ts
function zip<T, U>(array: readonly T[], withArray: readonly U[]): [T, U][];
function unzip<T, U>(pairs: readonly [T, U][]): [T[], U[]]
```

**Describe the solution you'd like** 
I'd like it if zip could work on more than two arrays at a time, and if unzip could work on more than an array of pairs.

**Describe alternatives you've considered** 
There are two ways forward: Either use varargs to accept as many arrays as possible like this and zip them up:
```ts
function variadicZip<T = any>(...arrays: (readonly T[])[]): T[] {
    const arrLength = arrays.map(i => i.length);

    const returnLength = Math.min(...arrLength);

    const ret = new Array<any>(returnLength);

    for (let i = 0; i < returnLength; i += 1) {
        const arr = [];
        for (const array of arrays) {
            arr.push(array[i]);
        }
        ret[i] = arr;
    }

    return ret;
}
```

or to overload zip to take up to 10 arrays and return a zip of the arrays provided (implemented for 2 + 3 arrays). This has the advantage that it provides the caller with more type information (you get back an array of `[T, U][]` if you zip with two arrays, whereas with the previous implementation, you would get back an array of `[any][]`.

```ts
function zip<T, U>(
  array1: readonly T[], 
  array2: readonly U[]
): [T, U][] {
   const returnLength = Math.min(array1.length, array2.length);

  const ret = new Array<[T, U]>(returnLength);

  for (let i = 0; i < returnLength; i += 1) {
    ret[i] = [array1[i], array2[i]];
  }

  return ret;
}

function zip<T, U, V>(
  array1: readonly T[], 
  array2: readonly U[],
  array3: readonly V[]
): [T, U, V][] {
   const returnLength = Math.min(...[array1.length, array2.length, array3.length]);

  const ret = new Array<[T, U, V]>(returnLength);

  for (let i = 0; i < returnLength; i += 1) {
    ret[i] = [array1[i], array2[i], array3[i]];
  }

  return ret;
}
```

LionC:
[...]

PR #1204, opened 2021-09-04
feat(collections): allow zip to be variadic and typed

Takashiidobe:
~~WIP for now, adding tests and have to do the same for unzip.~~

Closes the zip part of #1203.

This implements zip variadically and in a type safe fashion and adds tests. I couldn't do the same for unzip, I will leave that to someone else.  

~~Here's a question: `unzip_test.ts` has this test currently. If given a nested empty array, `[[]]`, it returns two empty arrays `[[], []]`. I'm confused if this is desired. The old function signature was `export function unzip<T, U>(pairs: readonly [T, U][]): [T[], U[]]`, which means it must return two arrays. But if I do the same example in python, I get `[]`. I'd moreso favor getting back an empty array instead of a nested array with two empty arrays in it.~~

```ts
Deno.test({
  name: "[collections/unzip] empty input",
  fn() {
    unzipTest(
      [[]],
      [[], []],
    );
  },
});
```

~~Notes: `zip` appears to correctly infer the types, but it looks like `unzip` returns two `unknown[]`, which isn't correct. That needs to be fixed before merge, but I'm not sure how to do that.~~


bartlomieju:
@Takashiidobe I'm still interested in landing this PR! Please sign the CLA and rebase your branch so we can get to landing it.

CC @LionC can you please review it?

LionC:
Will do tomorrow

collections/README.md
```diff
@@ -879,8 +879,8 @@ assertEquals(withoutList, [3]);
 
 ### zip
 
-Builds 2-tuples of elements from the given array with matching indices, stopping
-when the smaller array's end is reached
+Builds N-tuples of elements from the given N arrays with matching indices,
+stopping when the smaller array's end is reached
```
LionC, line 883: Should maybe be

```suggestion
stopping when the smallest array's end is reached
```

now

collections/zip.ts
```diff
@@ -2,36 +2,41 @@
...
-): [T, U][] {
-  const returnLength = Math.min(array.length, withArray.length);
 
-  const ret = new Array<[T, U]>(returnLength);
+export function zip<T extends unknown[]>(
+  ...arrays: { [K in keyof T]: T[K][] }
+): T[] {
+  const arrLength = arrays.map((i) => i.length);
+
+  const returnLength = Math.min(...arrLength);
```
LionC, line 30: This seems like a good case for `minOf` from the same module, which should also be more memory efficient:

```suggestion
  const returnLength = minOf(arrays, (it) => it.length);
```

collections/zip.ts
```diff
@@ -2,36 +2,41 @@
...
+
+  const ret: T[] = new Array(returnLength);
 
   for (let i = 0; i < returnLength; i += 1) {
-    ret[i] = [array[i], withArray[i]];
+    const arr = [];
+    for (const array of arrays) {
+      arr.push(array[i]);
+    }
```
LionC, line 38: Isn't this equivalent to

```typescript
const arr = arrays.map((it) => it[i]);
```

? That is probably faster because it can predict the resulting length and use the optimized `.map()`?

LionC (changes requested):
Thanks, cool feature! Just some comments regarding code that could be shorter or more optimized.

collections/README.md
kt3k: I forgot to do that in https://github.com/denoland/deno_std/pull/1204/commits/0fefab0b2e0b6da4eb2677239136e9684ef935d2. Thanks for catching

collections/zip.ts
Takashiidobe: done

collections/zip.ts
Takashiidobe: done

kt3k (approved):
@Takashiidobe LGTM. Thank you for your contribution!

PR #1206, opened 2021-09-05
docs(collections): replace console.assert with assertEquals

majidsajadi:
In this PR i replace `console.assert` with `assertEquals` since `console.assert` will fail in case of non-primitive types.
In some cases `console.assert` was used correctly (for primitive values) but i replaced it with `assertEquals` anyway.
```javascript
console.assert({} === {}); // Assertion failed
```


kt3k (approved):
@majidsajadi Thanks! LGTM

This makes more sense!

PR #1207, opened 2021-09-06
docs: correct comment and code typos

cmorten:
Corrects outstanding typos across the std modules.


kt3k (approved):
We had so many typos...

Thank you for the fixes! @cmorten

PR #1208, opened 2021-09-06
fix(testing/asserts): export Constructor type

zhangyuannie:
It is used as parameter for `assertThrows` and `assertRejects`.


kt3k (approved):
@zhangyuannie Good catch. Thanks!

PR #1209, opened 2021-09-07
docs(README): add note about ci checks

### LINKED ISSUE #870: CI should use a locked version of Deno

### LINKED ISSUE #1174: Exclude canary CI statuses from merge requirements

kt3k:
I talked with @ry and we decided to keep running canary CI because we want to know the breakage as soon as possible. But we don't require them passing for the PR to be accepted. This PR tries to clarify that for the contributors.

closes #870 
closes #1174

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,8 @@ For contributions to the Node compatibility library please check the
 
 _About CI checks_:
 
-We currently have 9 checks on CI. You need to pass the following 6 checks for your PR to be accepted:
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM

PR #1210, opened 2021-09-07
docs(http): reduce friction for windows users r.e. `0.0.0.0`

### LINKED ISSUE #1165: addr defaults to "0.0.0.0"

cmorten:
Fixes #1165

Attempt to reduce friction for windows users r.e. `0.0.0.0` by opting to use `localhost` in logged statements.

Related to https://github.com/denoland/deno/pull/11938. As noted by @kt3k:

> Listening this address works for all platforms, but the browsers on windows don't work with the address 0.0.0.0, and that's been causing the confusions to Windows users of several tools (ref: denoland/deno_std#1165, denoland/deployctl#63, kt3k/packup#10)

### DIFF
--- a/examples/echo_server.ts
+++ b/examples/echo_server.ts
@@ -1,9 +1,12 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import { copy } from "../io/util.ts";
+
 const hostname = "0.0.0.0";
 const port = 8080;
 const listener = Deno.listen({ hostname, port });
-console.log(`Listening on ${hostname}:${port}`);
+
+console.log(`Listening on http://localhost:${port}`);
+
 for await (const conn of listener) {
   copy(conn, conn);
 }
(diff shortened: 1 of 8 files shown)


kt3k (approved):
@cmorten LGTM. Thanks!

PR #1211, opened 2021-09-07
fix: update to latest signal API changes

bartlomieju:
Updates to changes introduced in https://github.com/denoland/deno/pull/11909


bartlomieju:
This is gonna be tricky to land, because stable Deno still uses concrete types, while canary uses string literals.

kt3k:
Maybe we're going to land this during the next release process?

bartlomieju:
> Maybe we're going to land this during the next release process?

Yes I guess we'll have to, unlucky because we always cut `deno_std` before cutting last PR in `deno` but this won't work this time. I guess @nayeemrmn has merit in suggesting we should treat `deno_std` as a regular Deno project.

dsherret (approved):

PR #1212, opened 2021-09-08
feat(collections): add sample

majidsajadi:
This PR adds `sample` function to `collections` module. #1173


collections/_utils.ts
```diff
@@ -23,3 +23,10 @@ export function filterInPlace<T>(
 
   return array;
 }
+
+/**
+ * Produces a random number between the inclusive `lower` and `upper` bounds.
+ */
+export function randomNumber(lower: number, upper: number) {
```
kt3k, line 30: Because it uses `Math.floor`, I guess it's better to name this `randomIntegerNumber`

collections/sample.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Example:
+ *
+ * ```ts
+ * import { sample } from "./sample.ts"
+ * import { assert } from "../testing/asserts.ts";
+ *
+ * const numbers = [1, 2, 3, 4];
+ * const random = sample(numbers);
+ *
+ * assert(numbers.includes(random));
```
kt3k, line 17: this seems causing type error in doc testing. you can check that with the command
`deno test --doc collections/sample.ts`

collections/_utils.ts
LionC: Or `randomInteger`, the number part is kind of redundant there :-)

kt3k (approved):
@majidsajadi LGTM! Thank you for your contribution!

PR #1213, opened 2021-09-08
feat(collections): Add `associatewith`

getspooky:
Builds a new Record using the given array as keys and choosing a value for each key using the given selector.

```ts
import { associateWith } from "https://deno.land/std@$STD_VERSION/collections/mod.ts";
import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";

const names = ["Kim", "Lara", "Jonathan"];
const namesToLength = associateWith(names, (it) => it.length);

assertEquals(namesToLength, {
  "Kim": 3,
  "Lara": 4,
  "Jonathan": 8,
});
```

Related https://github.com/denoland/deno_std/issues/1173

### DIFF
--- a/collections/associate_with.ts
+++ b/collections/associate_with.ts
@@ -1,0 +1,37 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Builds a new Record using the given array as keys and choosing a value for each
+ * key using the given selector.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { associateWith } from "./associate_with.ts"
+ * import { assertEquals } from "../testing/asserts.ts";
+ *
+ * const names = [ 'Kim', 'Lara', 'Jonathan' ]
+ * const namesToLength = associateWith(names, it => it.length)
+ *
+ * assertEquals(namesToLength, {
+ *   'Kim': 3,
+ *   'Lara': 4,
+ *   'Jonathan': 8,
+ * })
+ *
+ * ```
+ */
+export function associateWith<T>(
+  array: readonly string[],
+  selector: (key: string) => T,
+): Record<string, T> {
+  const ret: Record<string, T> = {};
+
+  for (const element of array) {
+    const selectedValue = selector(element);
+
+    ret[element] = selectedValue;
+  }
+
+  return ret;
+}
--- a/collections/associate_with.ts
+++ b/collections/associate_with.ts
@@ -18,7 +18,6 @@
  *   'Lara': 4,
  *   'Jonathan': 8,
  * })
- *
  * ```
  */
 export function associateWith<T>(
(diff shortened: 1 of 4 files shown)


collections/associate_with.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * const names = [ 'Kim', 'Lara', 'Jonathan' ]
+ * const namesToLength = associateWith(names, it => it.length)
+ *
+ * assertEquals(namesToLength, {
+ *   'Kim': 3,
+ *   'Lara': 4,
+ *   'Jonathan': 8,
+ * })
+ *
```
LionC, line 21: That seems to be an accidental additional empty line

collections/associate_with.ts
```diff
@@ -0,0 +1,37 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Builds a new Record using the given array as keys and choosing a value for each
+ * key using the given selector.
+ *
```
LionC, line 6: We should document what happens if the same key is processed multiple times (e.g. with `["a", "b", "a"]` input).

collections/associate_with_test.ts
```diff
@@ -0,0 +1,81 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "../testing/asserts.ts";
+import { associateWith } from "./associate_with.ts";
+
```
LionC, line 5: We should test for consistent behaviour if the same key appears mumtiple times in the input (see my comment on docs)

LionC (changes requested):
Thanks for the implementation! I left some comments :-)

collections/associate_with.ts
getspooky: Fixed

collections/associate_with_test.ts
getspooky: Fixed, @LionC can you please review.

kt3k (approved):
@getspooky Nice work! LGTM

kt3k:
I'll wait a while before merging to see if there's any inputs, opinions on this because this item looks comparatively new in the proposed list.

getspooky:
I think something went wrong with test :
```sh
error: TS2551 [ERROR]: Property 'Signal' does not exist on type 'typeof Deno'. 'Deno.Signal' is an unstable API. Did you forget to run with the '--unstable' flag, or did you mean 'signal'?

```

kt3k:
@getspooky That CI error is going to be resolved in #1211, but we can't merge that fix right now because that will break stable CI. So let's just ignore canary CI for now.

LionC:
@kt3k can we merge this before the upcoming release?

kt3k:
@LionC Thank you for the ping! Let's merge it.

@getspooky Thanks again for the work!

PR #1219, opened 2021-09-09
feat(testing/asserts): Add assertThrows() overload to receive error

nayeemrmn:
Returning the error is useful.

### DIFF
--- a/async/delay_test.ts
+++ b/async/delay_test.ts
@@ -17,14 +17,13 @@ Deno.test("[async] delay with abort", async function () {
   const { signal } = abort;
   const delayedPromise = delay(100, { signal });
   setTimeout(() => abort.abort(), 0);
-  const result = await assertRejects(
+  await assertRejects(
     () => delayedPromise,
     DOMException,
     "Delay was aborted",
   );
 
   const diff = new Date().getTime() - start.getTime();
-  assert(result === undefined);
   assert(diff < 100);
 });
 
--- a/async/pool_test.ts
+++ b/async/pool_test.ts
@@ -32,20 +32,15 @@ Deno.test("[async] pooledMap errors", async function () {
     return n;
   }
   const mappedNumbers: number[] = [];
-  let error: unknown;
   await assertRejects(async () => {
-    try {
-      for await (const m of pooledMap(3, [1, 2, 3, 4], mapNumber)) {
-        mappedNumbers.push(m);
-      }
-    } catch (e) {
-      error = e;
-      throw e;
+    for await (const m of pooledMap(3, [1, 2, 3, 4], mapNumber)) {
+      mappedNumbers.push(m);
     }
-  }, AggregateError);
+  }, (error: Error) => {
+    assert(error instanceof AggregateError);
+    assertEquals(error.errors.length, 2);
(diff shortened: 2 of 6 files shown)


bartlomieju (approved):
I agree this is useful 👍  @kt3k any objections?

kt3k:
This might be practically useful, but I also agree with @kitsonk 's this comment https://github.com/denoland/deno_std/pull/1052#discussion_r674510375

> Assertions shouldn't have return values.

This makes sense to me.

Node.js does the overloading of the 2nd argument and it passes the error object to it if the 2nd arg is a function.
https://nodejs.org/api/assert.html#assert_assert_throws_fn_error_message
How about following this pattern?

nayeemrmn:
@kt3k I've implemented the error callback overload. The usage does look better. Please review.

kt3k (approved):
Thank you @nayeemrmn for updating! Looks very nice!

LGTM

kitsonk:
This PR shouldn't have been merged as it stands.  The failures in canary should have been addressed, as the code is not safe to use with `useUnknownInCatchVariables`.  Now that Deno 1.14 is release, we should promote that change to the main test configuration, so we ensure that code does not get merged.

Can we please fix the code here please?

Also, this is the risk of having "canary" be red for a long time, and just assuming it is normal to be red. 😢 

Generally canary being red or not is not useful for `std` CI, it is useful information if a PR in CLI "breaks" std.  Sometimes that is intentional or unintentional.

nayeemrmn:
@kitsonk Opened #1278.

> The failures in canary should have been addressed, as the code is not safe to use with `useUnknownInCatchVariables`

> Generally canary being red or not is not useful for `std` CI, it is useful information if a PR in CLI "breaks" std. Sometimes that is intentional or unintentional.

Correction,
1. The recently landed PR** where std formally changed its target CLI version from 1.13.2 to 1.14.0 should have also updated CI to use `strict-ts44.tsconfig.json` on the release tests, which we did in advance for canary tests. This was missed.
2. Before landing _this_ PR, it should have been rebased above that change. At this point the release tests here would have also gone red and it would have been fixed.
3. Canary tests are still useless. With normal procedure ^^ release tests would have sufficed to expose the `useUnknownInCatchVariables` issue, and canary tests have been red for the past 9 days because of a separate issue that could not have been resolved until release tests also went red. Perfectly illustrates the uselessness.

** Technically there is no PR where std formally changes its target CLI version because **we don't explicitly select a version in CI** and it depends on the external datetime CLI is released. But the version pivot was #1211 - `fix: update to latest signal API changes`. It's no wonder why we'd miss other chores that need to be done when bumping CLI versions, or forget to rebase upcoming changes above such an inconspicuous sounding 'fix'.

kt3k:
@kitsonk Thank you for pointing this. I wasn't aware there was an error related to `useUnknownInCatchVariables` in canary CI as canary CI was always red these days 😓 

And thank you @nayeemrmn for opening #1278 very quickly! And thank you for analyzing the situation in details..

Probably the lesson is I should've merged main to this branch before merging and that should've revealed this error..

kt3k:
I think we could've kept canary CI passing if we skip some of the test cases only on canary CI, but I didn't do that because of the complexity of the steps. But maybe we can try such conditional test skipping in a similar situation in the future.

PR #1221, opened 2021-09-10
doc: Adds an explanation about how to test markdown files

grian32:
This is good to document for modules such as `collections` who have a `README.md` with examples.


kt3k (approved):
@grian32 Thanks!

PR #1222, opened 2021-09-12
chore: wrap examples in fenced code blocks

uki00a:
Wraps all examples in fenced code blocks so that they can be tested with `deno test --doc`.


kt3k (approved):
@uki00a Thanks! LGTM

PR #1223, opened 2021-09-12
feat(collections): add joinToString

majidsajadi:


collections/join_to_string.ts
```diff
@@ -0,0 +1,59 @@
...
+
+  array.forEach((it, index) => {
+    if (index > 0) {
+      ret += separator;
+    }
+
+    ret += selector(it);
+  });
+
+  return (ret += suffix);
```
LionC, line 58: Minor nitpick: Inline modification expressions are easy to miss when glancing over, I think it is best to do one thing per line:

```suggestion
  ret += suffix;
  
  return ret;
```

collections/join_to_string_test.ts
```diff
@@ -0,0 +1,93 @@
...
+      { name: "Anna", age: 31 },
+      { name: "Tim", age: 58 },
+    ];
+    const out = joinToString(array, (it) => it.name, {
+      suffix: " are winners",
+    });
+
+    assertEquals(out, "Kim, Anna, Tim are winners");
+  },
+});
```
LionC, line 93: Maybe add one test that uses all options

LionC (changes requested):
Thanks for implementing this! I left some minor comments.

majidsajadi:
@LionC updated. let me know if anything else you have in mind.

collections/join_to_string.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Transforms the elements in the given array to strings using the given selector.
+ * Joins the produced strings into one using the given separator and applying the given prefix and suffix to the whole string afterwards.
+ * Returns the resulting string.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { joinToString } from "./join_to_string.ts";
```
LionC, line 20: See #1227 , I think we should use the same `import` as in the `README` here

majidsajadi: oh, nice. that was something that bothered me a lot. I didn't know somebody fixed this in another PR.

kt3k:
@majidsajadi @LionC How about adding `limit` and `truncated` options as well from kotlin's version? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html

The current proposed API looks a little too trivial to me to be included in std..

majidsajadi:
> @majidsajadi @LionC How about adding `limit` and `truncated` options as well from kotlin's version? https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/join-to-string.html
> 
> The current proposed API looks a little too trivial to me to be included in std..

I guess `limit` and `truncated` options are something that would be very useful. the current implemented version can be replaced by a one-liner. `array.map(selector).join(separator)`. adding more options makes the module valuable for real-life projects.

kt3k:
@majidsajadi 
> P.S. how about transform? that's something I guess we can use a lot?

I guess `transform` is already mapped to `selector` in our case

LionC:
I am fine with adding `limit` and `truncate`. I think in TS, having a subobject might make more sense for safer typing:

```ts
type JoinToStringOptions = {
  separator?: string;
  prefix?: string;
  suffix?: string;
  truncate?: {
    after: number;
    with?: string;
  };
};
```

That will ensure that if you truncate, you need to set a limit and the optional truncate suffix thing (in kotlin you do not really have objet literals, you need to call constructors, so they tend to flatten parameters).

We could also think abot adding a `lastSeparator` string option that sets a different separator for the very last join, wich allows for the classic `A, B, C and D` thing for e.g. log messages.

I want to note though @kt3k, that the current version *should* already be faster than `.map().join()`, which is a good reason in itself.

majidsajadi:
```ts
type JoinToStringOptions = {
  separator?: string;
  prefix?: string;
  suffix?: string;
  truncate?: {
    after: number;
    with?: string;
  };
};
```
we settle on this? should i update the MR? @kt3k @LionC

kt3k:
> we settle on this? should i update the MR? @kt3k @LionC

That design looks good to me. Let's settle on it. It'd be helpful if you could update it

collections/join_to_string.ts
```diff
@@ -0,0 +1,63 @@
...
+ * );
+ *
+ * assertEquals(message, "<Kim and Anna and Tim>");
+ * ```
+ */
+export function joinToString<T>(
+  array: readonly T[],
+  selector: (el: T) => string,
+  {
+    separator = ", ",
```
getspooky, line 45: I think space separator is more appropriate

```suggestion
separator = " ",
```

getspooky (changes requested):

collections/join_to_string.ts
LionC: `Array.join` and Kotlin's `joinToString` both have the comma in their default, why should we deviate?

zhangyuannie (commented):
Personally, I'm against this API as it stands right now since
```js
array.map(selector).join(separator)
```
is more idiomatic and faster.

I agree with @kt3k that we should include `limit` and `truncated` for this to be included in std.

> I want to note though @kt3k, that the current version should already be faster than .map().join(), which is a good reason in itself.

@LionC Do you have any source for your claim?
With an array of length `500` with `{ num: Math.random() * 100 }` as element and `obj => "" + Math.floor(obj.num)` as selector, current implementation takes 500ms while map-join takes 150ms.

collections/join_to_string.ts
majidsajadi: @zhangyuannie @LionC `Array.prototype.join` default separator is `","` but Kotlin's `joinToString` default separator is `", "`. 

I copied my default parameters from Kotlin's `joinToString` but I guess it's better to go with `Array.prototype.join` in this case. more familiar API for JS developers.

collections/join_to_string.ts
```diff
@@ -0,0 +1,63 @@
...
+    if (index > 0) {
+      ret += separator;
+    }
+
+    ret += selector(it);
+  });
+
+  ret += suffix;
+
+  return ret;
```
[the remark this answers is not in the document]
majidsajadi: > User might pass null for prefix and suffix, I don't think they want null literally in the string.

i guess passing `null` for `prefix` or `suffix`  throw `Type 'null' is not assignable to type 'string'.` error.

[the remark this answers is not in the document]
LionC: I am very strongly against introducing special code to protect JS consumers from passing the wrong types - that is what TS is for.

majidsajadi:
@kt3k review, please.

kt3k:
@majidsajadi Sorry we changed the CLA a little bit (changed the contact email address), and we need you to sign it again..

majidsajadi:
@kt3k Done.

kt3k:
Thanks!

PR #1225, opened 2021-09-12
fix (collections/includesValue): prevent enumerable prototype check

ayame113:
Add an `Object.hasOwn()` check to the `for...in` loop to prevent enumeration of keys that exist in prototype.

before:
```ts
Object.prototype.a = "hello";
includesValue({}, "hello"); // => true
```
after fix:
```ts
Object.prototype.a = "hello";
includesValue({}, "hello"); // => false
```

### DIFF
--- a/collections/includes_value.ts
+++ b/collections/includes_value.ts
@@ -21,7 +21,10 @@ export function includesValue<T>(
   value: T,
 ): boolean {
   for (const i in record) {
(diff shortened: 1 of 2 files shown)


LionC (approved):
LGTM - and good timing, as this is not breaking now (function has not been released yet)

Pinging @kt3k

kt3k (approved):
@ayame113 LGTM. Nice!

PR #1226, opened 2021-09-12
feat (collections): add `runningReduce`

ayame113:
last part of #1173 

- [x] Did you support importing from mod.ts as well as your functions file?
- [x] Have you made sure to allocate as little memory and do as little steps in your algorithm as possible?
- [x] Did you add/adapt JSDoc comments, add/adapt an example, made sure that the example uses the ts markdown tag and assertEquals to show the result?
- [x] Did you add/adapt your functions documentation in README.md?
- [x] Did you make sure not to mutate the arguments passe to your function?
- [x] Did you add tests ensuring no mutation, empty input and corner cases?
- [x] Are your types flat, meaning there is no unnecessary alias in them that makes your users "go to type definition" twice to understand what to pass?

### DIFF
--- a/collections/running_reduce.ts
+++ b/collections/running_reduce.ts
@@ -18,10 +18,10 @@
  * ```
  */
 export function runningReduce<T, A>(
-  collection: readonly T[],
+  array: readonly T[],
   reducer: (accumulator: A, current: T) => A,
   initialValue: A,
 ): A[] {
   let _initialValue = initialValue;
-  return collection.map((el) => _initialValue = reducer(_initialValue, el));
+  return array.map((el) => _initialValue = reducer(_initialValue, el));
 }
--- a/collections/running_reduce.ts
+++ b/collections/running_reduce.ts
@@ -1,0 +1,27 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Calls the given reducer on each element of the given collection, passing it's
+ * result as the accumulator to the next respective call, starting with the given
+ * initialValue. Returns all intermediate accumulator results.
+ *
+ * Example:
+ *
+ * ```ts
+ * import { runningReduce } from "https://deno.land/std@$STD_VERSION/collections/mod.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * const numbers = [1, 2, 3, 4, 5];
+ * const sumSteps = runningReduce(numbers, (sum, current) => sum + current, 0);
+ *
+ * assertEquals(sumSteps, [1, 3, 6, 10, 15]);
+ * ```
+ */
+export function runningReduce<T, A>(
+  collection: readonly T[],
+  reducer: (accumulator: A, current: T) => A,
+  initialValue: A,
+): A[] {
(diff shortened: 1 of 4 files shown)


collections/running_reduce.ts
```diff
@@ -0,0 +1,27 @@
...
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * const numbers = [1, 2, 3, 4, 5];
+ * const sumSteps = runningReduce(numbers, (sum, current) => sum + current, 0);
+ *
+ * assertEquals(sumSteps, [1, 3, 6, 10, 15]);
+ * ```
+ */
+export function runningReduce<T, A>(
+  collection: readonly T[],
```
LionC, line 21: We have settled on `array` instead of `collection` for this argument in all other functions, I think we should be consistent here :-)

collections/running_reduce.ts
```diff
@@ -0,0 +1,27 @@
...
+ * import { runningReduce } from "https://deno.land/std@$STD_VERSION/collections/mod.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * const numbers = [1, 2, 3, 4, 5];
+ * const sumSteps = runningReduce(numbers, (sum, current) => sum + current, 0);
+ *
+ * assertEquals(sumSteps, [1, 3, 6, 10, 15]);
+ * ```
+ */
+export function runningReduce<T, A>(
```
LionC, line 20: Very minor nitpick: If there is only one output type, it is called `O` in the other functions (I think consistently).

collections/running_reduce.ts
```diff
@@ -0,0 +1,27 @@
...
+  collection: readonly T[],
+  reducer: (accumulator: A, current: T) => A,
+  initialValue: A,
+): A[] {
+  let _initialValue = initialValue;
```
LionC, line 25: Minor nitpick: Underscore prefixed values have some historical meaning attached to them which does not really apply here. I would just choose some "normal" name (like `currentResult` or something?)

collections/running_reduce.ts
```diff
@@ -0,0 +1,27 @@
...
+  return collection.map((el) => _initialValue = reducer(_initialValue, el));
```
LionC, line 26: This is actually pretty smart!

I would personally put the function body on a newline, as a lot of things are happening here, the line is hard enough to read as it is and it hides a statements in there.

```suggestion
  return collection.map((el) =>
    _initialValue = reducer(_initialValue, el)
  );
```

LionC (changes requested):
Thanks for implementing this! Loos good, I just added naming and API comments for consistency with the module. I really like the implementation.

ayame113:
@LionC 
Thank you for your review, I have updated the PR.

collections/running_reduce.ts
ayame113: When I apply this change, it doesn't seem to pass the `deno fmt` check.

LionC (approved):
LGTM! Pinging @kt3k

kt3k (approved):
@ayame113 LGTM. Nice work!

@LionC Thank you for reviewing in detail

PR #1227, opened 2021-09-12
docs (collections): Avoid using relative paths in import statements in code examples

ayame113:
Currently, relative paths are used in the import statement in some code examples.
Change this to import from https.

### DIFF
--- a/collections/README.md
+++ b/collections/README.md
@@ -97,7 +97,7 @@ For arrays, maps and sets, a merging strategy can be specified to either
 include non enumerable properties too.
 
 ```ts
-import { deepMerge } from "./deep_merge.ts";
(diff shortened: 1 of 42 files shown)


LionC:
If that works with `deno test`, I am all for it

LionC:
Oh sorry, I thought you changed that in the actual function JSDocs. That might actually be a good idea - as they are what is being displayed on doc.deno.land

Would you consider refactoring that here as well?

ayame113:
I also changed the import statement in JSDoc.
Although it makes document testing difficult locally, I'm in favor of modifying JSDoc to improve the display of the document.

kt3k (approved):
I agree with this change. This makes more sense to the readers of the docs. LGTM

PR #1228, opened 2021-09-12
chore (collections): rename `_tests.ts` to `_test.ts`

ayame113:
Rename the file:
- `min_of_tests.ts` -> `min_of_test.ts`
- `max_of_tests.ts` -> `max_of_test.ts`

I suspect that these two have not been tested in CI until now.


LionC (approved):
Good catch! Pinging @kt3k

ry (approved):
thx @ayame113

PR #1229, opened 2021-09-12
feat(std/node): add `ParsedUrlQuery` to querystring

talentlessguy:
This PR adds the following type:

```ts
export interface ParsedUrlQuery {
  [key: string]: string | string[] | undefined;
}
```

this fixes https://github.com/alephjs/esm.sh/issues/115


kt3k:
@talentlessguy Thank you for your contribution.

ParsedUrlQuery is not strictly part of node.js API, but it's defined and used in [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/43dc8c40f145ad0aa33e2cc91e6cd00f002bf944/types/node/querystring.d.ts) (community type definition of node.js), and [tinyhttp](https://github.com/tinyhttp/tinyhttp) uses it. That's why this change is needed. Please correct me if I'm missing something.

I'm in favor of this change, but if we introduce this type, I think we should use this type as the returned type of `parse` to follow the typing of `@types/node`

talentlessguy:
ok I will use it as a return type for `parse`, will push the fix soon

talentlessguy:
@kt3k I added `ParsedUrlQuery` type as a return type for `parse`, please take a look :)

kt3k (approved):
@talentlessguy LGTM. Thank you for your contribution!

talentlessguy:
omg thx for merging!

Issue #1224, opened 2021-09-12
Should `findSingle` has the default predicate?


kt3k:
`findSingle` in `collections` module has the default ~~selector~~ predicate of `(_) => true`. Does this make sense? What's the use case?

context of this discussion: https://github.com/denoland/deno_std/pull/1166#issuecomment-917268388

getspooky:
@kt3k yes there is no need to add this default selector of `(_) => true` , it does not make sense.

LionC:
> @kt3k yes there is no need to add this default selector of `(_) => true` , it does not make sense. 
> 
> I will push a fix for this. 

That is a very opinionated statement without stating reasons^^

Keep in mind that the default selector has been dorectly copied from the inspired module - which seems to see a use case for it (one I have used amd encountered myself several times).

I do not see how rempving it adds any value right now - at least not without actually talking about why.

getspooky:
@LionC for feedback, let's take this example and discuss:

```ts
const bookings = [
  { month: 'January', active: false },
  { month: 'March', active: false },
  { month: 'June', active: true },
];

const dd = findSingle(bookings);
```

It will return `undefined` but there is no given condition passed. This think could be useful if we specify field like 

```
const dd = findSingle(bookings, 'active');
```

even if we pass one value like 

```ts
const bookings = ["January", "March", "June"];
const dd = findSingle(bookings);
```

it does not make sense

kt3k:
The default selector (predicate) can be used for getting the single element from a signleton array

```ts
const bookings = [{ month: 'January', active: false }];
const item = findSingle(bookings);
console.log(item);
// => { month: "January", active: false }
```

This might be useful sometimes in my view, but there was an argument that if we do this, the name `findSingle` doesn't work well. (by @timreichen) ref: https://github.com/denoland/deno_std/pull/1166#issuecomment-917268388

He suggests we should use `array.length === 1 ? array[0] : undefined` instead of `findSingle(array)`. I partially agree with it, but not 100% convinced.

majidsajadi:
isn't `predicate` required in type definition? also according to the description, `findSingle` finds a single element if matches the given condition. in my opinion, the return in the below example is not what I expect according to the description and type definition.

```ts
const bookings = [{ month: 'January', active: false }];
const item = findSingle(bookings);
console.log(item);
// => { month: "January", active: false }
```

same as native javascript `find`. the predicate must be provided.

if the method was something like this:

```ts
[...]

PR #1232, opened 2021-09-13
fix(collections): Remove default selector for `findSingle`

getspooky:
Related https://github.com/denoland/deno_std/issues/1224

### DIFF
--- a/collections/find_single.ts
+++ b/collections/find_single.ts
@@ -23,7 +23,7 @@
  */
 export function findSingle<T>(
   array: readonly T[],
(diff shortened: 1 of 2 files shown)


LionC:
See my comment here: https://github.com/denoland/deno_std/issues/1224#issuecomment-918268254

I do not see reasons posted yet why we should jusz remove this.

zhangyuannie (approved):
LGTM! I'm personally in favour of this change.

LionC:
The discussion in the issue is very much still ongoing, I think it makes sense to focus on that and open the PR based on a call there

PR #1274, opened 2021-09-13
feat(io/streams): Propagate cancel in readableStreamFromIterable()

nayeemrmn:
Passes readable stream cancellations to the underlying iterator when applicable.
```ts
const readable = readableStreamFromIterable(async function* () {
  try {
    yield "foo";
  } catch (error) {
    console.log(error); // Error: Cancelled by consumer.
  }
}());
const reader = readable.getReader();
console.log(await reader.read()); // { value: "foo", done: false }
await reader.cancel(new Error("Cancelled by consumer."));
```


kt3k (approved):
LGTM. Nice!

PR #1279, opened 2021-09-16
chore(.github): Specify Deno release version explicitly

### LINKED ISSUE #870: CI should use a locked version of Deno

nayeemrmn:
Also remove outdated `--ignore` entries from canary tests.

- This should be bumped just after CLI release publishes and just before cutting a new std release.
- Breaking changes in CLI (`--unstable` or TypeScript) should be adapted to here when bumping this.
- Accumulated `--ignore` entries in canary tests should be flushed at this time also.

Closes #870.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,15 @@ on:
   pull_request:
     branches: [main]
 
+# The CLI release version specified below should be bumped just after a new one
+# publishes and just before cutting a new std release.
+# When doing this, ensure to:
+# - Reference the commit by hash so we aren't blocked on it publishing.
+# - Adapt to any breaking changes in CLI (`--unstable` or TypeScript).
+# - Remove accumulated `--ignore` entries from the canary tests.
+env:
+  DENO_VERSION_HASH: f27902e749fb3a78ca35a0dc41c9fc53cd2806e1 # 1.14.1
+
 jobs:
   test:
     runs-on: ${{ matrix.os }}
@@ -13,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        deno: [1.x, canary]
+        deno: [release, canary]
         os: [macOS-latest, ubuntu-latest, windows-2019]
 
     steps:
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks for opening this Nayeem; we've been actually discussing going with your approach and completely decoupling `deno_std` releases from CLI after the recent mishap in 1.14. There are however some technical problems we'll need to address before that could happen, eg:

> This should be bumped just after CLI release publishes and just before cutting a new std release.

Right now the expectation is that they are released in tandem and https://github.com/denoland/deno_website2/blob/main/versions.json is updated - this file is used by actions to pull latest version of CLI. So when upgrading to 1.14 @dsherret and I actually had to specify `(1.14.0, 0.106.0)` pair to make 1.14 available in std, quickly land the fixes; release 0.107.0 of `deno_std` and fixup version pair to `(1.14.0, 0.107.0)`.

nayeemrmn:
> we've been actually discussing going with your approach and completely decoupling `deno_std` releases from CLI

For clarity I never quite suggested that. We can still release an std version to accompany every CLI version, and any lib should be able to do the same. We just need to fix the procedure and timing of things.

> Right now the expectation is that they are released in tandem and https://github.com/denoland/deno_website2/blob/main/versions.json is updated - this file is used by actions to pull latest version of CLI. So when upgrading to 1.14 @dsherret and I actually had to specify `(1.14.0, 0.106.0)` pair to make 1.14 available in std, quickly land the fixes; release 0.107.0 of `deno_std` and fixup version pair to `(1.14.0, 0.107.0)`.

Okay I can see the issue -- synchronised CLI and std version publishing is needed for rendering of the manual. A simple workaround is instead of writing 1.14.0 in CI here, we use specific canary syntax to reference the same commit:
```yml
- uses: denoland/setup-deno@v1
  with:
    # IMPORTANT: Reference the Deno release version by commit hash.
    # Current version: 1.14.0
    deno-version: f92cc66f0dbf8e9753f2e98d4469190f3b561eb0
```
It's the exact same thing but allows us to complete both releases before having to update https://github.com/denoland/deno_website2/blob/main/versions.json. I'll reflect that here.

kt3k:
Closing this for now. Revisit this after a few releases

Thank you @nayeemrmn for bringing this up

PR #1282, opened 2021-09-16
fix (collections/maxWith,minWith): improve handling of arrays containing undefined

ayame113:
In the case of a comparator that returns undefined as the maximum or minimum value, the value returned depends on the order of the array.
This is a pretty edge case, but I've fixed it.

```ts
const comparator = (a, b) => {
  if (a === undefined) {
    return 1;
  }
  if (b === undefined) {
    return -1;
  }
  return 0;
};
// current implementation:
maxWith([undefined, 1, undefined], comparator) //=> undefined (correct)
maxWith([undefined, undefined, 1], comparator) //=> 1 (wrong)
```

### DIFF
--- a/collections/max_with.ts
+++ b/collections/max_with.ts
@@ -25,10 +25,12 @@ export function maxWith<T>(
   comparator: (a: T, b: T) => number,
 ): T | undefined {
   let max: T | undefined = undefined;
+  let isFirst = true;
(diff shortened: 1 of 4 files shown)


kt3k (approved):
@ayame113 Good catch! LGTM

PR #1283, opened 2021-09-17
feat(http/server): create Middleware type

keithamus:
This PR creates a type of `Middleware` in the `http/server` module, which frameworks can consume as a standardized interface for creating Middleware applications within Deno.

See https://github.com/denoland/deno_std/pull/768#issue-818064027 for more context. I'd like to kick the discussion off again since between that PR and this one, the std `http/server` has had a ground up rewirte. To quote the original PR:

> ### The Sales Pitch
> 
> I think it is critically important to provide middleware as a base in `std/http` because it gives a foundation for frameworks like Oak to build off of, but in a way to which other frameworks can benefit; "a rising tide lifts all boats" and all that. One area of contention in the Node ecosystem is that various frameworks exist, and yet all have to effectively re-implement the same set of middlewares over and over, as there is no centralised mechanism for middleware. Compare this to a language like Golang where middleware is an intrinsic part of the std library and we see frameworks building atop that, but also smaller libraries integrating with it in powerful ways which gives user's more choice and library/framework authors more power & expresivity.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -41,6 +41,11 @@ export type Handler = (
 ) => Response | Promise<Response>;
 
 /**
+ * A definition for creating middleware
+ */
+export type Middleware = (next: Handler): Handler
+
+/**
  * Parse an address from a string.
  *
  * Throws a `TypeError` when the address is invalid.


LionC:
I have a draft (that I talked to @cmorten and some ithers about) that adds type safe middlewares. If it is ok for you to wait until the weekend, I would like to propose it.

MiddlewRe often needs to add things to the request "context" to be consumed by handlers. Habdling that in a typesafe way is a bit complicated but possible. It would be sad if we would establish a Middleware type that lets people run into runtime errors if we can prevent them from doing so.

keithamus:
I'm mostly interested in kicking off the discussion, I have no attachment to the PR nor the timeline :wink: 

FWIW [Golang handles this with `Context`](https://pkg.go.dev/context#WithValue).

LionC:
So does e.g. Koa - I am trying to find an approach that is compatible to the current handler though (successfully I think^^)

I very much agree that we should hve a middleware concept in std/http for it to be usable for actual applications - especiall because we already have some other std modules that could be used well for some middleware out of the box :-)

keithamus:
I would also like to port [ServeMux](https://pkg.go.dev/net/http#ServeMux), using the new `URLPattern` interface... but thats another PR :laughing:

cmorten:
> I would also like to port [ServeMux](https://pkg.go.dev/net/http#ServeMux), using the new `URLPattern` interface... but thats another PR 😆

I have a like for like in a branch, but agree be good to utilise URLPattern now that has dropped!

REF: https://github.com/cmorten/deno_std/pull/1

kitsonk:
For context, oak is not going to use std/http at all.  The std abstraction is great (thanks @cmorten), but now that there is a stable native HTTP server in Deno, oak will strip out std, and oak couldn't build off this type if it wanted to.

The middleware interface in oak looks like this:

```ts
export interface Middleware<
  S extends State = Record<string, any>,
  T extends Context = Context<S>,
> {
  (context: T, next: () => Promise<unknown>): Promise<unknown> | unknown;
}
```

And `@types/koa` looks like this:

```ts
namespace Application {
  type Middleware<StateT = DefaultState, ContextT = DefaultContext, ResponseBodyT = any> = compose.Middleware<
      ParameterizedContext<StateT, ContextT, ResponseBodyT>
  >;
}

namespace compose {
  type Middleware<T> = (context: T, next: Koa.Next) => any;
}
```

My opinion has always been std should avoid frameworky type abstracts.  `std/http` IMO is an abstraction that allows those who may have built something off of `std/http` and want a little help to move onto something else.  We have to draw a line where std ends and frameworks begin.

Once you start dealing with middleware as well, it is a lot more than a standardised interface, it is how the middleware is composed, what order it is called in, etc.

While Go lang has a `context` as part of the std library, it sort of doesn't map to Deno.  If anything, the fetch event of `{ request, respondWith() }` is the context and dealing with the `Request` and `Response` web platform APIs.  Additional abstractions on top of that become highly opinionated, and I don't think there is good consensus on those opinions.  Oak was inspired by koa, but I felt that the context API was a bit suspect, having lots of "convenience" references and it has evolved since.

Also, I think a type without an implementation is a problem, so there is nothing to build on top of.

Now, having said all that, if we abstract it back to the basics and discussed some primitives, there may make sense to do something around middleware, but we would need a few things, IMO:

- A generic middleware definition something more along the lines of the Koa compose one above, that has a single generic for the context, and a `next`.  I think we could narrow the return type though, to be `unknown` at least I think.
- Provide a compose implementation.  I would be glad to contribute the compose in oak.  I think middleware without composition is effectively useless, and `compose()` deals with all the things you need to produce a single function of which a `context` can be called.
- This isn't an `http` thing.  It should be either `std/middleware` or `std/compose`.

I haven't discussed it with the rest of the core team though, so at this point it is only my personal opinion.

LionC:
@kitsonk My draft (again, will post it in some rough POC format on the weekend) is "just" functional composition and some TS magic - so middleware is just a normal handler with an additional `next` parameter. It also contains a very very minimal function to compose middleware - bl magic again, in the end, it is just one call stack. I am looking forward to see if that adresses those concerns or not :-)

A note on middleware in std in general: If we want std to be usable to write actual applications (which at leaT in my mind is the purpose of a standard library), I don't think we can get around offering middleware or some other sinple pluggable way to do things like cors handling, Json accepting, logging and so on.

keithamus:
> Provide a compose implementation. I would be glad to contribute the compose in oak. I think middleware without composition is effectively useless, and compose() deals with all the things you need to produce a single function of which a context can be called.

I'm not sure a compose function is necessary if the signature of a middleware takes a next handler and returns a new handler. Function composition can be used, compose would just do the same thing without significant improvement in ergonomics:

```to
type Middleware = (next: Handler) => Handler

function logger...
function auth...
function router...


const app = logger(auth(router())

const compose = (...xs) => xs.reduce((f, x) => f(x))

compose(logger, auth, router)
```

LionC:
@keithamus @kitsonk @cmorten see the linked issue above for the promised concept proposal

bartlomieju:
@keithamus thank you for the PR, but I'm going to close it without a merge. We are still debating adding middleware support for `std/http`; with roughly 50/50 split among the core team. Current suggestion is to use similar middleware type as https://github.com/lucacasonato/fresh does.

PR #1286, opened 2021-09-17
collections: Use function overloading

majidsajadi:
This PR add typescript function overloading to 'maxBy', 'sortBy', 'minBy'


LionC:
Nice! Why don't we add the other natively comparable types to the functions they are missing from too while we are at it? Would you be fine with adding them?

majidsajadi:
> natively comparable types

@LionC yes. I would work on them also. which functions?

kt3k:
`sort_by_test.ts` seems causing type errors in CI. Can you address these as well? @majidsajadi

LionC:
> > natively comparable types
> 
> @LionC yes. I would work on them also. which functions?

All functions that do `max` or `min` should work with all the types `sortBy` works with I think.

majidsajadi:
@kt3k Updated

kt3k (approved):
@majidsajadi Thanks! Nice work!

LGTM

PR #1289, opened 2021-09-17
fix(testing): Fix nested array matching for assertObjectMatch

### LINKED ISSUE #1288: [std/testing] assertObjectMatch() doesn't correctly assert nested arrays
**Describe the bug**

The `assertObjectMatch` assertion function fails with tests that include nested arrays as values in objects.

The following assertion will fail:

```ts
assertObjectMatch({ positions: [[1, 2, 3]] }, { positions: [[1, 2, 3]] });
```

with the following `AssertionError`:

```
error: Uncaught AssertionError: Values are not equal:

    [Diff] Actual / Expected

    {
      positions: [
-       {
-         "0": 1,
-         "1": 2,
-         "2": 3,
-         length: 3,
-       },
+       [
+         1,
+         2,
+         3,
+       ],
      ],
    }

        ^
    at assertEquals (https://deno.land/std@0.107.0/testing/asserts.ts:271:9)
    at assertObjectMatch (https://deno.land/std@0.107.0/testing/asserts.ts:546:10)
    <snip>
```

**Expected behavior**

It should correctly match against nested arrays, similar to how objects are compared without coercing the values to "ArrayLike" object literals.

**Desktop (please complete the following information):**

- OS: macOS Mojave 10.14.6
- Version 0.107.0

```
deno 1.13.2 (release, x86_64-apple-darwin)
v8 9.3.345.11
typescript 4.3.5
```

fnky:
Fixes failing assertions for `assertObjectMatch` when using nested arrays as values.

Fixes #1288


kt3k:
@fnky 
Thank you for your contribution!

This might be a fix, but the suggested fix throws with the following assertion:

```ts
assertObjectMatch({ positions: [[1, 2, 3, 4]] }, { positions: [[1, 2, 3]] });
```

I'm not sure this is the expected behavior of this method as the 2nd arg looks like a `subset` of of the 1st arg..

Does anyone have any opinion about this?

@lowlighter What do you think about this approach?

fnky:
@kt3k You have a good point. Given my initial expected output, it should work with matching parts of the array.

In fact, the `assertObjectMatch` assertion could then also accept arrays as input, to match parts of arrays. What do you think?

kt3k:
> In fact, the assertObjectMatch assertion could then also accept arrays as input, to match parts of arrays.

I think that's probably what most users would expect to happen

fnky:
One edge case is matching an array with an ArrayLike (e.g. an object with `length` property). What do you think would be the optimal way to handle this:

- Arrays should match arrays and array-likes `[1, 2, 3] ≈ [1, 2, 3] ≈ { "0": 1, "1": 2, "2": 3, "length": 3 }`.
- Arrays should only match against other array `[1, 2, 3] ≈ [1, 2, 3] ≇ { "0": 1, "1": 2, "2": 3, "length": 3 }`.

kt3k:
@fnky 
The 2nd case would make more sense to me. But I'm even ok with the 1st case as this looks really an edge case. Maybe let's choose simpler one?

lowlighter:
I think `assertObjectMatch({ positions: [[1, 2, 3, 4]] }, { positions: [[1, 2, 3]] });` should pass so it matches what is currently done for 1d arrays.

As for handling array-like, I'd go with 2nd case too, especially since the `Deno.inspect` that is printed in assertions fails will send back different outputs anyway for both. Actually I think 1st cannot be done without patching `assertEquals` directly.

___

I have not digged in details, but it's possible that the fix should be located here instead:
https://github.com/denoland/deno_std/blob/9a9ff4d1d670c522eb945bac73db65dabf01d988/testing/asserts.ts#L570-L572

Basically, a bit above this code snippet, it checks that `a[key]` and `b[key]` are both arrays (else `assertEquals` will do the job) and if they are, an appropriate subset is extracted while making sure that all elements (especially objects) within will also be transformed in subsets too.

The issue is probably caused by `typeof array === "object"`, so nested arrays gets converted back into object-like which don't pass `assertEquals`

testing/asserts.ts
```diff
@@ -545,6 +545,10 @@ export function assertObjectMatch(
   const seen = new WeakMap();
   return assertEquals(
     (function filter(a: loose, b: loose): loose {
+      // If the actual value is an array, let assertEquals do the assertion.
+      if (Array.isArray(a)) {
+        return a;
+      }
       // Prevent infinite loop with circular references with same filter
```
getspooky, line 552:
```suggestion

```

getspooky, line 552:
```suggestion

```

testing/asserts.ts
```diff
@@ -558,6 +562,7 @@ export function assertObjectMatch(
       ]
         .filter((key) => key in b)
         .map((key) => [key, a[key as string]]) as Array<[string, unknown]>;
+
       for (const [key, value] of entries) {
         // On array references, build a filtered array and filter nested objects inside
```
getspooky, line 567:
```suggestion

```

getspooky (changes requested):
I think we should remove unuseful comment.

testing/asserts.ts
kt3k: Why do you think this comment is useless?

getspooky: Comments are typically seen as a good thing, they can improve readability, offer context as to what the code is trying to do but in some case the code don’t bring any other explanation rather than the code itself.

fnky: I don't get it. This code is complex enough that it warrants a comment that explains the purpose of it. The code reads as "make a weakmap to store things in it" but not for what purpose, the comment clarifies this.

caspervonb: Comments seem fine to me, they're providing context.

bartlomieju:
I really want to get this PR landed. @fnky can you please sign the CLA again and rebase the branch?

fnky:
@bartlomieju I'm unsure if this is the correct solution given that it doesn't match subsets of arrays.

kt3k (approved):
LGTM. Thanks for updating

We can still improve the case of right-side containing the subarray of left-side, but this change is already enough better than the current state. Thank you for working on this @fnky

PR #1292, opened 2021-09-19
BREAKING(feat): remove ws module

crowlKats:
Since denoland/deno#12024 merged, the ws module should be removed.


kt3k (approved):
LGTM

Does anyone have any concern about this?

zhangyuannie:
> LGTM
> 
> Does anyone have any concern about this?

It would be nice if we don't remove `examples/chat/*` but instead refactor it to utilise the new native HTTP and WebSocket API since a lot of blogs/guides/tutorials out there still use std. This could obviously also be tackled in another PR instead.

crowlKats:
> It would be nice if we don't remove `examples/chat/*` but instead refactor it to utilise the new native HTTP and WebSocket API since a lot of blogs/guides/tutorials out there still use std. This could obviously also be tackled in another PR instead.

Yes I was considering that, but was thinking maybe it would be better to have such an example again once `std/http` would have websocket support, but thinking about it again it might be better to have a bare-bones example that wouldnt use `std/http`.

kt3k:
@crowlKats Thank you for updating chat example.

Talked a little on discord dev-std channel, it seems more preferable to keep `ws` a few more std versions and remove them at the same time as `std/http/legacy_serever.ts`. What do you think?

crowlKats:
> @crowlKats Thank you for updating chat example.
> 
> Talked a little on discord dev-std channel, it seems more preferable to keep `ws` a few more std versions and remove them at the same time as `std/http/legacy_serever.ts`. What do you think?

That sounds good to move. Do you want me to open a PR for marking the `ws` things as deprecated?

kt3k:
@crowlKats 
> Do you want me to open a PR for marking the ws things as deprecated?

That's very helpful!

kt3k:
deprecation is done in #1306. I'm merging the example update part of this PR. Thank you again for this! @crowlKats

PR #1293, opened 2021-09-20
docs(collections): replace console.assert with assertEquals

kawarimidoll:
Fix document because `===` can't compare non-primitive types.
Same as #1206.


kt3k (approved):
@kawarimidoll LGTM. Thanks!

PR #1296, opened 2021-09-21
fix(node/buffer): fix toString with base64 encoding

### LINKED ISSUE #1290: Canary CI is failing with `readFileEncodeBase64Success` case

kt3k:
This PR fixes `toString` method of Buffer of `std/node`.

Currently `toString` with base64 encoding always uses underlying array buffer (b.buffer) as the source data, which means the method isn't actually slicing the data even when `start` and `end` are passed.

This PR fixes the above.

---
The change in canary accidentally revealed this bug as `Deno.readFile()` now returns Uint8Array with `arr.byteLength !== arr.buffer.byteLength`

closes #1290

### DIFF
--- a/node/buffer.ts
+++ b/node/buffer.ts
@@ -437,7 +437,7 @@ export class Buffer extends Uint8Array {
 
     const b = this.subarray(start, end);
     if (encoding === "hex") return new TextDecoder().decode(hex.encode(b));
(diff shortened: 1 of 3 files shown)


kt3k:
This is included in https://github.com/denoland/deno_std/commit/7a1db16dba9cc4fb1f01f0731ef35d80230ae256

PR #1297, opened 2021-09-21
feat(std/node/stream): add partial support for `stream/web`

talentlessguy:
this pull request adds partial support for the experimental `stream/web` API from Node.js which is basically re-exports browser APIs.

Some of the APIs aren't in Deno yet so I have added only the ones present in Deno.

Ref: https://nodejs.org/api/webstreams.html#webstreams_class_readablestreambyobreader

Closes https://github.com/alephjs/esm.sh/issues/163

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -12,7 +12,7 @@ Deno standard library as it's a compatibility module.
 - [x] assert/strict _partly_
 - [ ] async_hooks
 - [x] buffer
-- [x] child*process \_partly*
+- [x] child_process \_partly_
 - [ ] cluster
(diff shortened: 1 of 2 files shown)


kt3k (approved):
@talentlessguy LGTM. Thanks!

kt3k:
canary CI failures are already addressed in main

talentlessguy:
damn thanks! fyi I know it's an unstable API but I didn't add any custom implementations so it's technically fine

and it will make fetch-blob work (since they import stream/web)

PR #1299, opened 2021-09-21
fix(node): Use denoflate to decompress Node test folder

### LINKED ISSUE #1298: `node/_tools/setup.ts` doesn't work since Deno v1.13.0
The downloaded file has a wrong length with Deno >= v1.13.0, and it cannot be decompressed. The tool doesn't have a problem with Deno <= v1.12.2.

Soremwar:
Closes #1298 

`compress` seems to have been kinda abandoned, so I just opted to replace it with `denoflate`, solving some ugly TypeScript errors in the process

### DIFF
--- a/node/_tools/setup.ts
+++ b/node/_tools/setup.ts
@@ -1,4 +1,4 @@
-import { gunzip } from "https://deno.land/x/compress@v0.3.9/gzip/gzip.ts";
+import { gunzip } from "https://deno.land/x/denoflate@1.2.1/mod.ts";
 import { Untar } from "../../archive/tar.ts";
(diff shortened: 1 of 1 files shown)


kt3k:
@Soremwar Thank you for looking into this!

I confirmed this fixes the decompressing part of `setup.ts`, but the downloading part still doesn't look working on my machine (macOS, deno v1.14.0)

The actual file size of `node-v15.5.1.tar.gz` is `62304300`, but it becomes `62299180` with recent Deno. (might be a problem of deno cli itself)

kt3k (approved):
LGTM as this partially addresses #1298, but I'll keep the issue open

PR #1301, opened 2021-09-21
fix(std/node/module): treat .mjs files as ESM

nayeemrmn:


kt3k (approved):
@nayeemrmn LGTM. Thanks!

kt3k:
Canary failure is already addressed in main.

PR #1305, opened 2021-09-24
docs: change CI rule

kt3k:
We discussed this internally, and decided to try only run canary CI for a while.

The reasons of this change are:
- We had problems in some recent releases because canary CI wasn't passing at the time of release
- Merging with red status confuses people (and caused a mistake)
- The latest released Deno is compatible with the latest tag of std, and it doesn't need to compatible with std's main

The downside of this change is (probably) that we can't merge PRs in std when CLI canary accidentally broke something in std.

This is an experimental attempt. I think we'll revisit this after several releases.

related: #1279 #870 denoland/deno#9801

previous change: #1209

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        deno: [1.x, canary]
+        deno: [canary]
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

PR #1306, opened 2021-09-24
chore: deprecate ws module

crowlKats:


kt3k (approved):
@crowlKats LGTM. Thanks!

PR #1307, opened 2021-09-24
perf(crypto): reduce one mircotask

ahabhgk:
The `await` here will do a duplicate microtask, return the promise directly can reduce it


satyarohith (approved):
LGTM. THanks, @ahabhgk!

PR #1308, opened 2021-09-24
feat(std/node): Add back os.tmpdir() implementation

espindola:
This was lost when Dino.dir was removed. This implementation is in
typescript and should do the same thing as std::env::temp_dir was
doing in the previous implementation:

https://doc.rust-lang.org/std/env/fn.temp_dir.html

The previous implementation was in the pull request denoland/deno#4213

### DIFF
--- a/node/os.ts
+++ b/node/os.ts
@@ -201,7 +201,29 @@ export function setPriority(pid: number, priority?: number): void {
 
 /** Returns the operating system's default directory for temporary files as a string. */
 export function tmpdir(): string | null {
-  notImplemented(SEE_GITHUB_ISSUE);
+  /* This follows the node js implementation, but has a few
+     differences:
+     * On windows, if none of the environment variables are defined,
+       we return null.
+     * On unix we use a plain Deno.env.get, instead of safeGetenv,
+       which special cases setuid binaries.
+     * Node removes a single trailing / or \, we remove all.
+  */
+  if (isWindows) {
+    const temp = Deno.env.get("TEMP") || Deno.env.get("TMP");
+    if (temp) {
+      return temp.replace(/(?<!:)[/\\]*$/, "");
+    }
+    const base = Deno.env.get("SYSTEMROOT") || Deno.env.get("WINDIR");
+    if (base) {
+      return base + "\\temp";
+    }
+    return null;
+  } else { // !isWindows
+    const temp = Deno.env.get("TMPDIR") || Deno.env.get("TMP") ||
+      Deno.env.get("TEMP") || "/tmp";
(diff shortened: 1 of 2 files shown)


kt3k:
Thank you for your contribution. But this doesn't seem aligned to the Node.js implementation of os.tmpdir. I'd suggest we should simply follow the node.js implementation https://github.com/nodejs/node/blob/43291b98edaa682b9fa74f95e084ce7a01c85774/lib/os.js#L173-L189

espindola:
> Thank you for your contribution. But this doesn't seem aligned to the Node.js implementation of os.tmpdir. I'd suggest we should simply follow the node.js implementation https://github.com/nodejs/node/blob/43291b98edaa682b9fa74f95e084ce7a01c85774/lib/os.js#L173-L189

Done.

> And please sign the CLA

There should now be a corporate CLA in place.

node/os.ts
```diff
@@ -201,7 +201,26 @@ export function setPriority(pid: number, priority?: number): void {
...
+    const temp = Deno.env.get("TEMP") || Deno.env.get("TMP");
+    if (temp) {
+      return temp;
+    }
+    const base = Deno.env.get("SYSTEMROOT") || Deno.env.get("WINDIR");
+    if (base) {
+      return base + "\\temp";
+    }
+    return null;
+  }
```
piscisaureus, line 221: For readability purposes, use `else`.

piscisaureus:
It seems that node.js strips trailing slashes (and backslashes on windows) unless the temp dir is a root (`/` or `X:\`). 
Is there a reason this implementation doesn't?

espindola:
> It seems that node.js strips trailing slashes (and backslashes on windows) unless the temp dir is a root (`/` or `X:\`). Is there a reason this implementation doesn't?

Yes, sorry I meant to include that in the comment. The implementation in node removes only one trailing slash:

     $ TMP=/tmp/ node -e 'console.log(os.tmpdir())'
     /tmp
     $ TMP=/tmp// node -e 'console.log(os.tmpdir())'
     /tmp/

Which seems pretty arbitrary. I can do it if we want bug for bug compatibility, but it seems better to remove all trailing slashes or none, no?

piscisaureus:
> Which seems pretty arbitrary. I can do it if we want bug for bug compatibility, but it seems better to remove all trailing slashes or none, no?

It does seem arbitrary, we should remove all trailing slashes.
On windows we should remove both `/` and `\`.

kt3k (approved):
LGTM

kt3k:
@espindola Thank you for your contribution!

PR #1309, opened 2021-09-25
chore(node/url): Add tests for `fileURLToPath` and `pathToFileURL`

wafuwafu13:
Ref: https://github.com/nodejs/node/blob/8d83c470291d05f22010662380633bbddf0e7dcb/test/parallel/test-url-pathtofileurl.js#L96-L139

### DIFF
--- a/node/url_test.ts
+++ b/node/url_test.ts
@@ -1,6 +1,7 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import { assertEquals } from "../testing/asserts.ts";
 import * as url from "./url.ts";
+import { isWindows } from "../_util/os.ts";
 
 Deno.test({
   name: "[url] URL",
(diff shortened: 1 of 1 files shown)


node/url_test.ts
```diff
@@ -1,10 +1,114 @@
...
+import { isWindows } from "../_util/os.ts";
 
 Deno.test({
   name: "[url] URL",
   fn() {
     assertEquals(url.URL, URL);
   },
 });
+
+// todo(wafuwafu13) Add Windows and invalid case
```
kt3k, line 13: nit: We prefer the format of `TODO(username):`

kt3k (approved):
@wafuwafu13 LGTM. Thanks!

kt3k:
FYI we also have an infrastructure for running actual node.js test cases under `node/_tools`. (Modify `node/_tools/config.json` and run `node/_tools/setup.ts`, then it will enable the test cases listed in config.json)

PR #1311, opened 2021-09-25
feat(collections/running_reduce): Change `initialValue` to optional

wafuwafu13:
[kotlin.collections/running-reduce](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/running-reduce.html#runningreduce) has no optional specification of `initialValue` and 
 [JavaScript/Array/reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) can optionally specify `initialValue`.

### DIFF
--- a/collections/running_reduce.ts
+++ b/collections/running_reduce.ts
@@ -3,8 +3,8 @@
 
 /**
  * Calls the given reducer on each element of the given collection, passing it's
- * result as the accumulator to the next respective call, starting with the given
- * initialValue. Returns all intermediate accumulator results.
+ * result as the accumulator to the next respective call. If initialValue is given,
+ * starting with that value. Returns all intermediate accumulator results.
  *
  * Example:
  *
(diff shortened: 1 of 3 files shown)


collections/running_reduce.ts
```diff
@@ -3,26 +3,45 @@
...
+ * const secondNumbers = [1, 2, 3, 4, 5]
+ * const resultWithInitialValue = runningReduce(secondNumbers, (sum, current) => sum + current, 5);
+ *
+ * assertEquals(resultWithInitialValue, [6, 8, 11, 15, 20]);
  * ```
  */
-export function runningReduce<T, O>(
+export function runningReduce<T>(
   array: readonly T[],
-  reducer: (accumulator: O, current: T) => O,
```
wafuwafu13, line 23: [kotlin.collections/running-reduce](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/running-reduce.html#runningreduce) does not accept different types

```kotlin
fun main() {
    val strings = listOf(1, "foo", 2, "bar")
    println(strings.reduce { acc, current -> acc + current })
    // Type mismatch: inferred type is Unit but {Comparable<*> & java.io.Serializable} was expected
}
```

kt3k:
Thank you for the contribution. But this seems unnecessarily limiting the ability of this function. The example like the below doesn't work with this design anymore

```ts
import { runningReduce } from "https://deno.land/std@0.108.0/collections/mod.ts";
const users = [
  { name: "John", age: 20, },
  { name: "Jane", age: 21, }
];
console.log(runningReduce(users, (acc, current) => acc + current.age, 0));
```

LionC:
I will take a look as well tonight - just glancing at @kt3k s comment, we definitely do not want to remove functionality from the function to be more compliant with kotlin. Kotlin is just ab inspiration and a startig point, not a restriction for the final design :-)

Have not looked at the actual change yet though, just saw @kt3k s comment.

LionC:
Having looked at the code, I do not agree with removing support for `reduce`ing to another type - that just removes features for no reason I can see in the PR.

The other point is about having an optional `initialValue` - as I see it, that just makes sense if the `reducer` returns `T` - so we would need to overload the function to support that:

```typescript
export function runningReduce<T>(
  array: readonly T[],
  reducer: (accumulator: T, current: T) => T,
  initialValue?: T,
): T[]

export function runningReduce<T, O>(
  array: readonly T[],
  reducer: (accumulator: O, current: T) => O,
  initialValue: O,
): O[]
```

but that still does not answer what the semantics for the first version are. They are not obvious to me - is the resulting array one element shorter than with an `initialValue`? Or is it the same length but the first value is the first value of the input, meaning not every value is actually produced by the `reducer`?

All in all, **I think it is not worth it** to introduce that kind of complexity and ambiguity into the function and it's implementation - forcing an `initialValue` is clear, unambigous and covers the cases I can come up with well enough for me.

wafuwafu13:
@kt3k @LionC 

>this seems unnecessarily limiting the ability of this function
>we definitely do not want to remove functionality from the function to be more compliant with kotlin.

I agree with this opinion :)

I wanted to change `initialValue` to optional. Because:
- I think that `initialValue` is often used as `0` or `""`.
- https://github.com/denoland/deno_std/pull/1311#issue-1007070455

But I agree with @LionC
>All in all, I think it is not worth it to introduce that kind of complexity and ambiguity into the function and it's implementation - forcing an initialValue is clear, unambigous and covers the cases I can come up with well enough for me.

If there are no more comments, I'll close this PR soon and just add test case.
Thanks for reviewing!

PR #1315, opened 2021-09-26
fix(nodes/events): make `EventEmitter.call` compatible with es5

uki00a:
Part of #911

This PR patches `EventEmitter.call` so that `EventEmitter` can be inherited in the ES5 way.


node/events.ts
```diff
@@ -82,10 +82,27 @@ export class EventEmitter {
...
+      emitter._events === Object.getPrototypeOf(emitter)._events // If `emitter` does not own `_events` but the prototype does
+    ) {
+      emitter._events = Object.create(null);
+    }
+  }
+
+  /**
+   * Overrides `call` to mimic the es5 behavior with the es6 class.
+   */
+  // deno-lint-ignore no-explicit-any
```
uki00a, line 99: I think this is not ideal. However, I'm not sure if `EventEmitter` should be rewritten in ES5 style... :thinking:

kt3k (approved):
@uki00a LGTM. Nice work!

node/events.ts
kt3k: I think it's ok with either way as long as it passes the test cases

PR #1320, opened 2021-09-28
test(collections/running_reduce) Add test case

wafuwafu13:
Related: https://github.com/denoland/deno_std/pull/1311#issuecomment-927709291
>I'll close this PR soon and just add test case.

Add case that:
- `initialValue` is not `0` or `""`.
- array of objects


kt3k (approved):
@wafuwafu13 LGTM. Thanks!

Issue #1312, opened 2021-09-25
std/node/buffer: Add `Buffer.readUIntBE`


talentlessguy:
**Is your feature request related to a problem? Please describe.** 

I am in progress of porting Ganache to Deno and I found out `std/node/buffer` doesn't have `Buffer.readUIntBE` implemented.

https://github.com/trufflesuite/ganache/blob/92d1fb74452bff99e9d7d8940286dbd4586859a7/src/packages/utils/src/utils/buffer-to-bigint.ts#L19

**Describe the solution you'd like**

An implementation of `Buffer.readUIntBE` method

**Additional context** 

Node.js docs: https://nodejs.org/api/buffer.html#buffer_buf_readuintbe_offset_bytelength

PR #1321, opened 2021-09-28
feat(node): Support `Buffer.readUIntBE`

getspooky:
Related https://github.com/denoland/deno_std/issues/1312

### DIFF
--- a/node/buffer.ts
+++ b/node/buffer.ts
@@ -2,6 +2,7 @@
 import * as hex from "../encoding/hex.ts";
 import * as base64 from "../encoding/base64.ts";
 import { Encodings, normalizeEncoding, notImplemented } from "./_utils.ts";
+import { ERR_BUFFER_OUT_OF_BOUNDS, ERR_OUT_OF_RANGE } from "./_errors.ts";
 
 const notImplementedEncodings = [
   "ascii",
@@ -245,14 +246,28 @@ export class Buffer extends Uint8Array {
     );
   }
 
+  boundsError(value: number, length: number, type: string = "offset"): never {
+    if (Math.floor(value) !== value) {
+      throw new ERR_OUT_OF_RANGE(type, "an integer", value);
+    }
+    if (length < 0) throw new ERR_BUFFER_OUT_OF_BOUNDS();
+
+    throw new ERR_OUT_OF_RANGE(
+      type,
+      `>= ${type ? 1 : 0} and <= ${length}`,
+      value,
+    );
+  }
+
   /**
    * Reads byteLength number of bytes from buf at the specified offset and interprets
    * the result as an unsigned big-endian integer supporting up to 32 bits of accuracy.
    */
-  readUIntBE(offset = 0, byteLength: number) {
+  readUIntBE(offset = 0, byteLength: number): number {
     if (byteLength === 4) return this.readUInt32BE(offset);
     if (byteLength === 2) return this.readUInt16BE(offset);
     if (byteLength === 1) return this.readUInt8(offset);
+    this.boundsError(byteLength, 4, "byteLength");
   }
 
   /**
--- a/node/buffer.ts
+++ b/node/buffer.ts
@@ -246,7 +246,7 @@ export class Buffer extends Uint8Array {
     );
   }
 
-  boundsError(value: number, length: number, type: string): never {
+  private boundsError(value: number, length: number, type: string): never {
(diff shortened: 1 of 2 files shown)


node/buffer_test.ts
```diff
@@ -298,6 +298,27 @@ Deno.test({
...
+      0x04,
+      0x05,
+      0x06,
+      0x07,
+      0x08,
+    ]);
+    assertEquals(buffer.readUIntBE(0, 1), 0x01);
+    assertEquals(buffer.readUIntBE(0, 2), 0x0102);
+    assertEquals(buffer.readUIntBE(0, 4), 0x01020304);
+    assertEquals(buffer.readUIntBE(1, 6), undefined);
```
kt3k, line 317: Node.js version returns `0x020304050607` for this.

node/buffer_test.ts
```diff
@@ -298,6 +298,27 @@ Deno.test({
...
+    assertEquals(buffer.readUIntBE(2, 8), undefined);
```
kt3k, line 318: Node.js version throws with `RangeError [ERR_OUT_OF_RANGE]` in this case

node/buffer.ts
```diff
@@ -245,6 +245,16 @@ export class Buffer extends Uint8Array {
...
   }
 
+  /**
+   * Reads byteLength number of bytes from buf at the specified offset and interprets
+   * the result as an unsigned big-endian integer supporting up to 32 bits of accuracy.
+   */
+  readUIntBE(offset = 0, byteLength: number) {
+    if (byteLength === 4) return this.readUInt32BE(offset);
+    if (byteLength === 2) return this.readUInt16BE(offset);
+    if (byteLength === 1) return this.readUInt8(offset);
```
kt3k, line 255: Node.js version supports 1, 2, 3, 4, 5, 6 for byteLength, not only 1, 2, 4

Maybe let's use `DataView` and `getBigUint64()` method to get up to 6 bytes Uint, and then cast them to the number.

We also need to check the range of byteLength.

node/buffer_test.ts
getspooky: Yes because currently there are just support for `1, 2, 4` byteLength, we should also add support for `5, 6`

node/buffer_test.ts
getspooky: Yes , I just add `boundsError` in order to throw error if The value of "byteLength" is out of range.

node/buffer.ts
getspooky: @kt3k I just add boundsError in order to throw error if The value of "byteLength" is out of range. The probleme is that DataView does not support `getUint64` and if we use `getBigUint64` the result will be different from `node js ` result. 
I suggest to make another PR to support ` 5, 6` byteLength

node/buffer.ts
```diff
@@ -245,6 +246,30 @@ export class Buffer extends Uint8Array {
     );
   }
 
+  boundsError(value: number, length: number, type: string): never {
```
kt3k, line 249: This exposes `boundsError` in Buffer class, which doesn't exist in Node.js version. So I'd suggest we should inline these in `readUIntBE`

node/buffer_test.ts
```diff
@@ -298,6 +298,32 @@ Deno.test({
...
+    ]);
+    assertEquals(buffer.readUIntBE(0, 1), 0x01);
+    assertEquals(buffer.readUIntBE(0, 2), 0x0102);
+    assertEquals(buffer.readUIntBE(0, 4), 0x01020304);
+    assertThrows(
+      () => {
+        assertEquals(buffer.readUIntBE(0, 5), 0x01020304);
+      },
+      Error,
+      `The value of "byteLength" is out of range. It must be >= 1 and <= 4. Received 5`,
```
kt3k, line 322: I think the error message should say `The value of "byteLength" is out of range. It must be >= 1 and <= 6. Received ...` because it actually supports 1 to 6, and we should throw `notImplemented` errors in the case of byteLength is 3, 5, or 6

kt3k (commented):
I left a few more comments

node/buffer.ts
getspooky: I will make it private and not inline it on `readUIntBE`, because we will use it for `readUIntLE`.
I will Open PR for `readUIntLE` later.

kt3k (approved):
Thank you for updating!

LGTM

PR #1323, opened 2021-09-29
chore(node/tools/setup): Make sure filenames are in the correct order before setting up

wafuwafu13:
If file names are not in the correct order, it will be difficult to see which tests have already been added.


kt3k (approved):
This looks a good idea.

LGTM. Thanks!

PR #1324, opened 2021-09-30
fix (http/server): make websocket available with `listenAndServe()`

ayame113:
Currently, using websocket with `listenAndServe()` causes panic. (See https://github.com/denoland/deno/issues/12098#issuecomment-929742799)

Remove one `await` from the `#respond` function so that the following code works tentatively.

```ts
import { listenAndServe } from "./server.ts";

listenAndServe(":8080", (request) => {
  const { socket, response } = Deno.upgradeWebSocket(request);
  socket.onmessage = (event) => console.log(event.data);
  return response;
});

const ws = new WebSocket("ws://localhost:8080"); // Currently causes panic
```

(However, keep in mind that the following code that uses the async function still causes a panic.)
```ts
import { listenAndServe } from "./server.ts";

listenAndServe(":8080", async (request) => {
//                      ^ If the async function here, panic
  const { socket, response } = Deno.upgradeWebSocket(request);
  socket.onmessage = (event) => console.log(event.data);
  return response;
});

const ws = new WebSocket("ws://localhost:8080");
```

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -342,15 +342,19 @@ export class Server {
    */
   async #respond(
     requestEvent: Deno.RequestEvent,
-    httpCon: Deno.HttpConn,
+    httpConn: Deno.HttpConn,
     connInfo: ConnInfo,
   ): Promise<void> {
     try {
       // Handle the request event, generating a response.
-      const response = await this.#handler(
+      const response = this.#handler(
         requestEvent.request,
         connInfo,
       );
+
+      if (response instanceof Promise) {
+        response.catch((_error) => this.#closeHttpConn(httpConn));
+      }
 
       // Send the response.
       await requestEvent.respondWith(response);
@@ -361,7 +365,7 @@ export class Server {
       // Alternatively the connection has already been closed, or there is some
       // other error with responding on this connection that prompts us to
       // close it and open a new connection.
-      return this.#closeHttpConn(httpCon);
+      return this.#closeHttpConn(httpConn);
     }
   }
 
@@ -395,8 +399,6 @@ export class Server {
       // allow the connection to handle multiple requests in the case of h2.
       this.#respond(requestEvent, httpConn, connInfo);
     }
-
-    this.#closeHttpConn(httpConn);
   }
 
   /**
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -641,6 +641,34 @@ Deno.test(`listenAndServe should handle requests`, async () => {
   }
 });
 
+Deno.test(`listenAndServe should handle websocket requests`, async () => {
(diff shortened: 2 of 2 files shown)


http/server.ts
```diff
@@ -395,8 +399,6 @@ export class Server {
       // allow the connection to handle multiple requests in the case of h2.
       this.#respond(requestEvent, httpConn, connInfo);
     }
-
-    this.#closeHttpConn(httpConn);
   }
```
ayame113, line 402: If I do not remove `closeHttpConn` here, this will result in a "Bad Resource" error.
But I'm not familiar with http, so I'm not sure if it's okay to remove it.

bartlomieju:
@cmorten could you please take a look?

cmorten:
> @cmorten could you please take a look?

I’ll try and spare some time to have a look. From a quick scan, on one hand it feels like we’re trying to compensate here for a rust panic issue which should be fixed at source rather than worked around here… that said, if it’s a matter of rejigging some stuff with no impact to perf or functionality then perhaps good to allow this usecase while we fix the underlying issue…

The test case is certainly an excellent addition!

R.E. closing connections - we should always aim to tidy up due connections. In a scenario where there is no way to be sure whether something is always closed, we can use try catches to attempt to close a potentially already closed resource (and if really desire, check the error is BadResource and throw otherwise, but IIRC the code as stands doesn’t go that far).

It _feels_ curious to be using a mix of async/await and .then()/.catch(), I should expect to be able to consistent, though best tool for job should trump some contrived code style IMO - again wonder if just putting a plaster on a deeper issue! Need to read into the issue and this fox a bit more

ayame113:
> on one hand it feels like we’re trying to compensate here for a rust panic issue which should be fixed at source rather than worked around here

Yes, this PR only provides a temporary patch until the original issue is fixed.👍


> we can use try catches to attempt to close a potentially already closed resource (and if really desire, check the error is BadResource and throw otherwise

Here is some simple code to reproduce the "Bad Resource" error:

```ts
new WebSocket("ws://localhost:8080");

const server = Deno.listen({ port: 8080 });
for await (const conn of server) {
  handle(conn);
}
async function handle(conn: Deno.Conn) {
  const httpConn = Deno.serveHttp(conn);
  for await (const requestEvent of httpConn) {
    const { socket, response } = Deno.upgradeWebSocket(requestEvent.request);
    requestEvent.respondWith(response);
  }
  console.log("will close httpConn");
  httpConn.close(); // close httpConn here and get an error
}
```

I'm guessing that the alive (still valid) websocket connection is closed, resulting in an error.
So, I feel like we shouldn't close the connection here.

PR #1326, opened 2021-09-30
feat(node): Support `Buffer.readUIntLE`

getspooky:
Reads byteLength number of bytes from buf at the specified offset and interprets the result as an unsigned, little-endian integer supporting up to 48 bits of accuracy.

### DIFF
--- a/node/buffer.ts
+++ b/node/buffer.ts
@@ -81,11 +81,11 @@ export class Buffer extends Uint8Array {
   static alloc(
     size: number,
     fill?: number | string | Uint8Array | Buffer,
-    encoding = "utf8",
+    encoding = "utf8"
   ): Buffer {
     if (typeof size !== "number") {
       throw new TypeError(
-        `The "size" argument must be of type number. Received type ${typeof size}`,
+        `The "size" argument must be of type number. Received type ${typeof size}`
       );
     }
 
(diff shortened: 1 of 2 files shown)


talentlessguy:
related: #1326

node/buffer.ts
```diff
@@ -273,6 +273,20 @@ export class Buffer extends Uint8Array {
     this.boundsError(byteLength, 4, "byteLength");
   }
 
+  /**
+   * Reads byteLength number of bytes from buf at the specified offset and interprets
+   * the result as an unsigned, little-endian integer supporting up to 32 bits of accuracy.
```
kt3k, line 278: I think doc should say `supporting up to 48 bits of accuracy` ref: https://nodejs.org/api/buffer.html#buffer_buf_readuintle_offset_bytelength

The purpose of std/node is the compatibility to node.js, not fixing it or providing our own design.

getspooky: Yes, i will fixe that.

node/buffer_test.ts
```diff
@@ -299,7 +299,7 @@ Deno.test({
 });
 
 Deno.test({
-  name: "Buffer readUIntBE",
```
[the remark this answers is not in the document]
getspooky: Yes, i will add a separate test for `readUIntLE`.

kt3k (approved):
Thank you for updating! @getspooky 

LGTM

PR #1328, opened 2021-09-30
feat(node/crypto): Add createHmac

UltiRequiem:
Adds and exports `createHmac` function.
Reference: https://nodejs.org/api/crypto.html#crypto_class_hmac

 - [x] I have added tests


node/crypto_test.ts
```diff
@@ -86,6 +86,11 @@ Deno.test("[node/crypto.getHashes]", () => {
   }
 });
 
+Deno.test("[node/crypto.createHmac] basic usage - hex output", () => {
+  const d = createHmac("sha1", "abc").update("def").digest("hex");
+  assertEquals(d, "589c22335a381f122d129225f5c0ba3056ed5811");
```
kt3k, line 91: This doesn't seem matching to the node.js result?
```shellsession
$ node -e 'console.log(crypto.createHmac("sha1", "abc").update("def").digest("hex"))'
12554eabbaf7e8e12e4737020f987ca7901016e5
```

kt3k (commented):
@UltiRequiem Thank you for your contribution. Left a question

PR #1330, opened 2021-10-01
chore(node/_tools/test): Add `--only` flag to run specific tests

wafuwafu13:
As the number of tests in `config.json` increases, it becomes more difficult to remove untargeted tests when we want to run a specific test in a local environment. (We can't even comment it out in a json file.)

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -119,6 +119,23 @@ To run the tests you have set up, do the following:
 $ deno test --allow-read --allow-run node/_tools/test.ts
 ```
 
+If you want to run specific tests in a local environment, add `--only` flag to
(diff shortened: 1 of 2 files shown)


node/README.md
```diff
@@ -119,6 +119,23 @@ To run the tests you have set up, do the following:
 $ deno test --allow-read --allow-run node/_tools/test.ts
 ```
 
+If you want to run specific tests in a local environment, add `--only` flag to
```
kt3k, line 122: Nice doc 👍

kt3k (approved):
@wafuwafu13 LGTM. Thanks!

PR #1332, opened 2021-10-01
feat(std/node/crypto): Add `crypto.randomUUID`

talentlessguy:
This PR adds a function `crypto.randomUUID` from Web Crypto API to `std/node/crypto.ts`

### DIFF
--- a/node/crypto.ts
+++ b/node/crypto.ts
@@ -1,28 +1,32 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 // Copyright Joyent, Inc. and Node.js contributors. All rights reserved. MIT license.
-import { default as randomBytes } from './_crypto/randomBytes.ts'
-import { crypto as wasmCrypto, DigestAlgorithm, digestAlgorithms } from '../_wasm_crypto/mod.ts'
-import { pbkdf2, pbkdf2Sync } from './_crypto/pbkdf2.ts'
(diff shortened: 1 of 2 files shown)


bartlomieju (changes requested):
Looks good, but please add a test case exercising this API

talentlessguy:
ok will do, one sec

talentlessguy:
@bartlomieju done

bartlomieju:
@lucacasonato can you take a look, I believe it has something to do with how Web APIs work

node/crypto.ts
```diff
@@ -157,5 +150,15 @@ export function getHashes(): readonly string[] {
   return digestAlgorithms;
 }
 
-export default { Hash, createHash, getHashes, pbkdf2, pbkdf2Sync, randomBytes };
-export { pbkdf2, pbkdf2Sync, randomBytes };
+const randomUUID = globalThis.crypto.randomUUID;
```
littledivy, line 153:
```suggestion
const randomUUID = () => crypto.randomUUID();
```
This should fix the `Illegal invocation` error

bartlomieju: Ah perfect! Thanks Divy!

talentlessguy:
looks like the tests are passing now?

bartlomieju (approved):
LGTM

PR #1333, opened 2021-10-01
feat(node/crypto): add `timingSafeEqual`

talentlessguy:
Ref: https://nodejs.org/dist/latest-v16.x/docs/api/crypto.html#crypto_crypto_timingsafeequal_a_b

- [x] I have added tests
- [x] I have run `deno fmt` and `deno lint`


kt3k (approved):
@talentlessguy LGTM. Thanks!

Issue #1284, opened 2021-09-17
Make chai.js and SinonJS compatible with testing framework and add instructions how to use it


bartlomieju:
[`chai.js`](https://www.chaijs.com/) and [`Sinon.JS`](https://sinonjs.org/) are very popular testing libraries that run both in Node and browser.

Since they can be run in browser it should be a low hanging fruit to make them work in Deno and I believe it would be very valuable to add "first-class" instructions to `testing/README.md` how they could be used with Deno's built-in testing (`deno test`).

CC @keithamus

keithamus:
There's some slow progress to getting chai working in Deno. We're working on migrating the dependencies right now so that they work "natively" in Deno & Node. I hope to include TypeScript support also.

bartlomieju:
> There's some slow progress to getting chai working in Deno. We're working on migrating the dependencies right now so that they work "natively" in Deno & Node. I hope to include TypeScript support also.

Thanks for the heads up, is there anything I can assist with? Are you facing any challenges to target both runtimes?

bartlomieju:
> > Are you facing any challenges...
> 
> Mostly motivation 😂. Deno is pleasurable to work with although it can be prickly with regard to imports and the `.ts` extension which Node+TypeScript doesn't like, but Deno requires. I'm not sure if there is good tooling to help manage the difference in dependency APIs?

Keith I noticed that `chai` already has [browser support](https://www.chaijs.com/guide/installation/#browser); I'm willing to guess that it should just work in Deno without any changes. Maybe the only thing you want to do is link that output to https://deno.land/x/ to make it more discoverable for Deno users?

cmorten:
@bartlomieju I believe you are right, e.g.

```ts
// @deno-types="https://unpkg.com/@types/chai@4.2.22/index.d.ts"
import "https://chaijs.com/chai.js";

Deno.test("we can make chai assertions", () => {
  const assert = chai.assert;
  const foo = "bar";
  const beverages = { tea: ["chai", "matcha", "oolong"] };

  assert.typeOf(foo, "string"); // without optional message
  assert.typeOf(foo, "string", "foo is a string"); // with optional message
  assert.equal(foo, "bar", "foo equal `bar`");
  assert.lengthOf(foo, 3, "foo`s value has a length of 3");
  assert.lengthOf(beverages.tea, 3, "beverages has 3 types of tea");
});

Deno.test("we can make chai expectations", () => {
  const expect = chai.expect;
  const foo = "bar";
[...]

PR #1334, opened 2021-10-01
[#1284] Test examples using sinon and chai

cmorten:
Some simple examples (taken from the respective libs' docs) for Sinon and Chai.

REF:
- https://www.chaijs.com/
- https://sinonjs.org/

### DIFF
--- a/testing/chai_example.ts
+++ b/testing/chai_example.ts
@@ -1,0 +1,36 @@
+// @deno-types="https://unpkg.com/@types/chai@4.2.22/index.d.ts"
+import "https://chaijs.com/chai.js";
+
+Deno.test("we can make chai assertions", () => {
+  const assert = chai.assert;
+  const foo = "bar";
+  const beverages = { tea: ["chai", "matcha", "oolong"] };
+
+  assert.typeOf(foo, "string"); // without optional message
+  assert.typeOf(foo, "string", "foo is a string"); // with optional message
+  assert.equal(foo, "bar", "foo equal `bar`");
+  assert.lengthOf(foo, 3, "foo`s value has a length of 3");
+  assert.lengthOf(beverages.tea, 3, "beverages has 3 types of tea");
+});
+
+Deno.test("we can make chai expectations", () => {
+  const expect = chai.expect;
+  const foo = "bar";
+  const beverages = { tea: ["chai", "matcha", "oolong"] };
+
+  expect(foo).to.be.a("string");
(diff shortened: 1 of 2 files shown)


bartlomieju:
@cmorten looks awesome! Having to use `@deno-types` is unfortunate and something we should address to make it more painless so we could add it to https://deno.land/manual@v1.14.2/testing without having to rely on those directives.

CC @dsherret take a look at Mocha integration

cmorten:
@bartlomieju moving to skypack (guessing JSPM as well) removes the need for Deno type or triple slash directives (changes pushed).

testing/mocha_example.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (failures > 0) {
+    Deno.exit(1);
+  } else {
+    Deno.exit(0);
+  }
+}
+
+/**
+ * And finally we run our tests, passing the onCompleted function
+ * hook and setting some globals.
```
cmorten, line 48:
```suggestion
 * hook.
```

bartlomieju:
> @bartlomieju moving to skypack (guessing JSPM as well) removes the need for Deno type or triple slash directives (changes pushed).

Ah great! I think we can land this PR as is then, but we should also add these examples to the manual so they're easier to discover.

testing/mocha_example.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function add(a: number, b: number): number {
+  return a + b;
+}
+
+describe("add", () => {
+  it("should add two positive numbers correctly", () => {
+    chai.expect(add(2, 3)).to.equal(2);
+  });
+});
```
dsherret, line 17: This may work for a single file, but does it work if tests are scattered across many files? I don't see this being super useable just because someone would have to import all these files somehow. Maybe it would work with a dynamic import that walks the directory tree?

I think the mocha pollyfill Luca has (currently in a gist) would be better to clean up and then recommend: https://github.com/denoland/deno/pull/12190#issuecomment-933441089 -- I guess the problem though is that the polyfill is not 100%.

bartlomieju: I agree with David. @cmorten could you remove the mocha example for now? I think it's still valuable to land chai and sinon examples as well as add them to the Manual.

kt3k (approved):
Great examples! Thank you @cmorten !

LGTM

cmorten:
Thanks for the assist @bartlomieju

PR #1335, opened 2021-10-02
test(node/url): Enable `test-url-fileurltopath.js`

wafuwafu13:
Part of https://github.com/denoland/deno_std/issues/911

I've made `TODO` for the parts where the error occurred.


wafuwafu13:
Question: Do we need to add tests that are not native node tests if they overlap with native node tests?
ex) https://github.com/denoland/deno_std/pull/1309

kt3k (approved):
@wafuwafu13 LGTM. Nice work!

> Question: Do we need to add tests that are not native node tests if they overlap with native node tests?

Node.js test cases are usually more exhaustive than ours. So if native Node.js test cases are enabled, then we don't need our own version of test cases anymore.

PR #1340, opened 2021-10-03
feature(node/crypto): add randomFillSync an randomFill

UltiRequiem:
Reference: https://nodejs.org/api/crypto.html#crypto_crypto_randomfillsync_buffer_offset_size
Adds and exports `randomFillSync` and `randomFill` function.

- [x] I have added tests

### DIFF
--- a/node/_crypto/randomFill.ts
+++ b/node/_crypto/randomFill.ts
@@ -1,6 +1,49 @@
 import randomBytes from "./randomBytes.ts";
 import { Buffer } from "../buffer.ts";
 
+export default function randomFill(
+  buf: Buffer,
+  cb: (err: Error | null, buf?: Buffer) => void,
+): void;
+export default function randomFill(
+  buf: Buffer,
+  offset: number,
+  cb: (err: Error | null, buf?: Buffer) => void,
+): void;
+export default function randomFill(
+  buf: Buffer,
+  offset: number,
+  size: number,
+  cb: (err: Error | null, buf?: Buffer) => void,
+): void;
+
+export default function randomFill(
+  buf: Buffer,
+  // deno-lint-ignore no-explicit-any
+  offset?: any,
+  // deno-lint-ignore no-explicit-any
+  size?: any,
+  cb?: (err: Error | null, buf?: Buffer) => void,
+) {
+  if (size + offset > buf.length) {
+    throw new RangeError(
+      `The value of "size + offset" is out of range. It must be <= ${buf.length}. Received ${(size +
+        offset)}.`,
+    );
+  }
+
+  const buffer = buf as Buffer;
+  const callback = cb as (err: Error | null, buf?: Buffer) => void;
+  const trueOffset = 0 | offset;
+  const trueSize = buffer.length - trueOffset;
+
+  randomBytes(size, (err, buf) => {
+    if (err) return callback(err as Error);
+    buffer.copy(buf as Buffer, trueOffset, 0, trueSize);
+    callback(null, buffer);
+  });
+}
+
 export function randomFillSync(buffer: Buffer, offset = 0, size?: number) {
   if (size === undefined) {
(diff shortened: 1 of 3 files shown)


node/_crypto/randomFill.ts
```diff
@@ -0,0 +1,60 @@
...
+  size: number,
+  cb: (err: Error | null, buf?: Buffer) => void,
+): void;
+
+export default function randomFill(
+  buf: Buffer,
+  // deno-lint-ignore no-explicit-any
+  offset?: any,
+  // deno-lint-ignore no-explicit-any
+  size?: any,
```
bartlomieju, line 25: I believe these could be strongly typed without using `any`.

Would `number | (err: Error | null, buf?: Buffer) => void` work?

UltiRequiem: Now is better?

node/_crypto/randomFill.ts
```diff
@@ -0,0 +1,61 @@
...
+  cb: (err: Error | null, buf?: Buffer) => void,
+): void;
+
+export default function randomFill(
+  buf: Buffer,
+  offset?: number | ((err: Error | null, buf?: Buffer) => void),
+  size?: number | ((err: Error | null, buf?: Buffer) => void),
+  cb?: (err: Error | null, buf?: Buffer) => void,
+) {
+  const callback = cb as (err: Error | null, buf?: Buffer) => void;
```
kt3k, line 26: This doesn't seem supporting the 1st and 2nd overloads

If I run the code like the below
```ts
import randomFill from "./node/_crypto/randomFill.ts";                   
const u8 = new Uint8Array(10);  
randomFill(u8, () => {  
  console.log(u8);  
});
```

It throws with `error: Uncaught TypeError: callback is not a function`

I think you need to check values and types of `offset`, `size`, and `cb` and swap variables

UltiRequiem: Thanks! Working on that.

UltiRequiem: Could you check again please @kt3k ?

node/_crypto/randomFill_test.ts
```diff
@@ -0,0 +1,44 @@
+import { Buffer } from "../buffer.ts";
+import randomFill, { randomFillSync } from "./randomFill.ts";
+import { assertEquals, assertThrows } from "../../testing/asserts.ts";
+
+const validateNonZero = (buf: Buffer) => buf.some((ch) => ch > 0);
```
kt3k, line 5: This function returns a boolean, and it doesn't throw. So `validateNonZero(buf)` doesn't check whether buf actually includes non zero value or not.

node/_crypto/randomFill.ts
```diff
@@ -0,0 +1,81 @@
...
+  buf: Buffer,
+  offset: number,
+  cb: ((err: Error | null, buf?: Buffer) => void),
+): void;
+
+export default function randomFill(
+  buf: Buffer,
+  offset: number,
+  size: number,
+  cb: (err: Error | null, buf?: Buffer) => void,
```
kt3k, line 41: DefinitelyTyped gives this `buf` type as non-optional. ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9ddc52c6175556e3a9a0dce6234c1713fdcdccae/types/node/crypto.d.ts#L1911

I think that makes more sense because the buffer is already given as a required parameter. So it should be always available.

kt3k:
Other part looks good to me

node/_crypto/randomFill_test.ts
UltiRequiem: Updated.

kt3k (approved):
LGTM. Thank you for your contribution!

UltiRequiem:
> LGTM. Thank you for your contribution!

Thank, I hope I can contribute more in the future!

PR #1345, opened 2021-10-04
feat (node/crypto) add `randomInt`

UltiRequiem:
Reference: https://nodejs.org/api/crypto.html#crypto_crypto_randomint_min_max_callback
Adds and exports `randomInt`.

- [x] I have added tests


UltiRequiem:
I got confused and combine this pull request with #1340. I'll make a new one with only the changes related to `randomInt`.

PR #1348, opened 2021-10-04
feat(hash): use common `Sha` interface for Sha1, Sha256 and Sha512

talentlessguy:
each SHA algorithm in std/hash has their own class but since all of them have the same set of functions it'd be smart to have a common interface that all these classes implement.

This PR adds a `hash/types.ts` file which exports `Sha` and `Message` types for convinience.

Example use case:

```ts
import type { Sha } from 'https://deno.land/std/hash/types.ts'

const createHash = (algo: string, key: string): Sha => {
  // ...
}
```

this will also ease adding std/node/crypto `createHmac` function in the future


kt3k:
Sorry for not mentioning this in hash/README explicitly, but we are in the process of deprecating std/hash
ref: https://github.com/denoland/deno_std/pull/1025

> this will also ease adding std/node/crypto createHmac function in the future

This makes sense to me. Maybe we move these types somewhere in std/node?

talentlessguy:
oh alright then

I don't think it makes sense to have it in node cuz there's already `Hash` class

and also I was planning on using SHA from std/hash but if it's going to be replaced with `crypto.subtle` I probably should use it instead

PR #1349, opened 2021-10-05
feat(bytes): switch equals to simd for large byte arrays

Aplet123:
For large `Uint8Array`s, it's faster to make `Uint32Array`s from the underlying `ArrayBuffer`s and compare those instead of the original `Uint8Array`s. I've used the following benchmarking code and found the cutoff to be at around 1000 elements on my machine:
```ts
import {
  bench,
  runBenchmarks,
} from "https://deno.land/std@0.110.0/testing/bench.ts";

function equalsNaive(a: Uint8Array, b: Uint8Array): boolean {
  if (a.length !== b.length) return false;
  for (let i = 0; i < b.length; i++) {
    if (a[i] !== b[i]) return false;
  }
  return true;
}

function equalsSimd(a: Uint8Array, b: Uint8Array): boolean {
  if (a.length !== b.length) return false;
  const len = a.length;
  const compressable = Math.floor(len / 4);
  const compressedA = new Uint32Array(a.buffer, 0, compressable);
  const compressedB = new Uint32Array(b.buffer, 0, compressable);
  for (let i = compressable * 4; i < len; i++) {
    if (a[i] !== b[i]) return false;
  }
  for (let i = 0; i < compressedA.length; i++) {
    if (compressedA[i] !== compressedB[i]) return false;
  }
  return true;
}

function rand(n: number): number {
  return Math.floor(Math.random() * n);
}

function randArr(len: number): Uint8Array {
  let arr = new Uint8Array(len);
    arr[i] = rand(256);
  }
  return arr;
}

console.log("generating benchmarks...");
const testCases: [Uint8Array, Uint8Array][] = [];
// CHANGE THESE
const len = 2000;
const nCases = 10000;
for (let i = 0; i < nCases; i++) {
  const arr1 = randArr(len);
  const arr2 = randArr(len);
  testCases.push([arr1, arr1.slice(0)]);
  testCases.push([arr1, arr2]);
}

bench({
  name: "bench old equals",
  func(b): void {
    b.start();
    for (const [a, b] of testCases) {
      equalsNaive(a, b);
    }
    b.stop();
  },
});

bench({
  name: "bench simd equals",
  func(b): void {
    b.start();
    for (const [a, b] of testCases) {
      equalsSimd(a, b);
    }
    b.stop();
  },
});

runBenchmarks();
```
For reference, here's the results I get with 100k arrays of length 10k:
```
benchmark bench old equals ...
    1194ms
benchmark bench simd equals ...
    502ms
```

### DIFF
--- a/bytes/test.ts
+++ b/bytes/test.ts
@@ -119,6 +119,40 @@ Deno.test("[bytes] equals", () => {
   assert(!v3);
 });
 
+function rand(n: number): number {
+  return Math.floor(Math.random() * n);
+}
+
+function randArr(len: number): Uint8Array {
+  const arr = new Uint8Array(len);
+  for (let i = 0; i < len; i++) {
+    arr[i] = rand(256);
+  }
+  return arr;
+}
+
+Deno.test("[bytes] equals randomized testing", () => {
+  // run tests before and after cutoff
+  for (let len = 995; len <= 1005; len ++) {
+    const arr1 = randArr(len);
+    const arr2 = randArr(len);
+    const arr3 = arr1.slice(0);
+    // the chance of arr1 equaling arr2 is basically 0
+    // but introduce an inequality at the end just in case
+    arr2[arr2.length - 1] = arr1[arr1.length - 1] ^ 1;
+    // arr3 is arr1 but with an inequality in the very last element
+    // this is to test the equality check when length isn't a multiple of 4
+    arr3[arr3.length - 1] ^= 1;
+    // arrays with same underlying ArrayBuffer should be equal
+    assert(equals(arr1, arr1));
+    // equal arrays with different underlying ArrayBuffers should be equal
+    assert(equals(arr1, arr1.slice(0)));
+    // inequal arrays should be inequal
+    assert(!equals(arr1, arr2));
+    assert(!equals(arr1, arr3));
+  }
+});
+
 Deno.test("[bytes] startsWith", () => {
   const v = startsWith(new Uint8Array([0, 1, 2]), new Uint8Array([0, 1]));
   const v2 = startsWith(new Uint8Array([0, 1, 2]), new Uint8Array([0, 2]));
--- a/bytes/test.ts
+++ b/bytes/test.ts
@@ -133,7 +133,7 @@ function randArr(len: number): Uint8Array {
 
 Deno.test("[bytes] equals randomized testing", () => {
   // run tests before and after cutoff
-  for (let len = 995; len <= 1005; len ++) {
+  for (let len = 995; len <= 1005; len++) {
     const arr1 = randArr(len);
     const arr2 = randArr(len);
     const arr3 = arr1.slice(0);
--- a/bytes/test.ts
+++ b/bytes/test.ts
@@ -119,23 +119,11 @@ Deno.test("[bytes] equals", () => {
   assert(!v3);
 });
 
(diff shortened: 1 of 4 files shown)


kt3k:
Thank you for your contribution. This looks interesting

Can you also add your bench mark code as `bytes/equals_bench.ts`?

Can you also add test cases in `bytes/test.ts` which exercise `equalsSimd` function. (Currently we only have test cases for very small byte arrays)

tyler338:
this is not simd, you are just reading and writing 32 bits instead of 8 bits

names should be `equals8bits` `equals32bits`

Aplet123:
@kt3k I've added the benching and randomized tests, but it leads to a few things I'm a bit unhappy with:

1. `equalsNaive` and `equalsSimd` aren't exported (since they're helper functions), but the benchmarking code needs them, so I've just copy-pasted their implementations. However, this seems quite messy to me, and maybe it'd be nicer to have an internal `equals.ts` file or something that does export the functions?
2. I used a `randArr` helper for my randomized testing, however it seems like a random byte array generator would be a very useful feature for the stdlib. Maybe it's worth adding?

@tyler338 ...isn't reading and writing larger values literally what SIMD is? Besides, these functions aren't exposed as a public API so naming doesn't matter very much.

Aplet123:
@tyler338 Sure, specialized SIMD instructions perform SIMD, but the concepts of SIMD are separate from what instructions your CPU runs. Straight from [the wikipedia page](https://en.wikipedia.org/wiki/SIMD):
> SIMD can be internal (part of the hardware design) and it can be directly accessible through an instruction set architecture (ISA): it should not be confused with an ISA.

kt3k:
> However, this seems quite messy to me, and maybe it'd be nicer to have an internal equals.ts file or something that does export the functions

`equals.ts` sounds good to me 👍 It'd be also great to add `@private` jsdoc tag when exporting internal utils from the file

> I used a randArr helper for my randomized testing, however it seems like a random byte array generator would be a very useful feature for the stdlib. Maybe it's worth adding?

Deno has crypto.getRandomValues API (a part of Web Crypto API) and we can get random Uint8Array like `crypto.getRandomValues(new Uint8Array(n))`. In my view this is useful & ergonomic enough for that usage.

Aplet123:
> Deno has crypto.getRandomValues API (a part of Web Crypto API) and we can get random Uint8Array like `crypto.getRandomValues(new Uint8Array(n))`. In my view this is useful & ergonomic enough for that usage.
Ah, I didn't know about that. I'll change my benchmarking and tests to use that instead, since it'll probably also be faster.

bytes/test.ts
```diff
@@ -119,6 +119,28 @@ Deno.test("[bytes] equals", () => {
...
 });
 
+Deno.test("[bytes] equals randomized testing", () => {
+  // run tests before and after cutoff
+  for (let len = 995; len <= 1005; len++) {
+    const arr1 = crypto.getRandomValues(new Uint8Array(len));
+    const arr2 = crypto.getRandomValues(new Uint8Array(len));
+    const arr3 = arr1.slice(0);
+    // the chance of arr1 equaling arr2 is basically 0
+    // but introduce an inequality at the end just in case
```
kt3k, line 129: 👍

kt3k (approved):
LGTM. Thank you for your contribution! Nice work!

PR #1353, opened 2021-10-05
fix(path): remove =c: env var usage

### LINKED ISSUE #1341: Do not use Deno.env.get with `=${resolvedDevice}` in path/win32.ts
The code is doing the following (which throws an error because Deno doesn't allow `=` in env var names):

https://github.com/denoland/deno_std/blob/9436e51cf6d3c60b53329bb97efc1046f5436d83/path/win32.ts#L57

But we shouldn't do this because it doesn't work in powershell. For example, where this "works" in node, in command prompt:

```batch
C:\Users\david>node -e "console.log(process.env['=C:']);"
C:\Users\david
```

But then in powershell:

```powershell
>node -e "console.log(process.env['=C:']);"
undefined
```

This is because `=C:` is specific to command prompt (cmd.exe). Read more: https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133

Relates to https://github.com/denoland/deno/issues/12305

Edit: well, looks like either backticks within backticks doesn't work in github issues titles or I don't know what I'm doing.

kt3k:
closes #1341

As pointed in #1341, "=C:" special variable is the feature of cmd.exe, not an OS feature. So we can't safely rely on that for path resolution. (Also Deno.env.get doesn't allow `=` character in env var name. So this code path doesn't work anyway)

ref: https://devblogs.microsoft.com/oldnewthing/20100506-00/?p=14133

### DIFF
--- a/path/win32.ts
+++ b/path/win32.ts
@@ -49,12 +49,7 @@ export function resolve(...pathSegments: string[]): string {
       ) {
         throw new TypeError("Resolved a relative path without a CWD.");
       }
-      // Windows has the concept of drive-specific current working
-      // directories. If we've resolved a drive letter but not yet an
-      // absolute path, get cwd for that drive, or the process cwd if
-      // the drive cwd is not available. We're sure the device is not
-      // a UNC path at this points, because UNC paths are always absolute.
-      path = Deno.env.get(`=${resolvedDevice}`) || Deno.cwd();
+      path = Deno.cwd();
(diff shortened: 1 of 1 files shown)


UltiRequiem (approved):

bartlomieju (approved):
LGTM

dsherret (approved):
LGTM too.

PR #1370, opened 2021-10-09
perf(fs): Parallelize `expandGlob`

reosablo:
In my use case, this change makes `expandGlob` x1.9 faster.


reosablo:
I signed that. Thank you for reviewing.

Issue #1182, opened 2021-08-30
Add assertError to testing/asserts.ts


KyleJune:
**Is your feature request related to a problem? Please describe.**

assertThrows and assertThrowsAsync have duplicate code for checking if an error has the correct class and msg. Someone may want to test an error that was not thrown or that was caught.

**Describe the solution you'd like**

Move the common code from assertThrows and assertThrowsAsync to a function called assertError so that users can make assertions about errors without them having to be thrown. Have both of the assertThrows functions use assertError internally.

```ts
const error = new TypeError("wrong type");
assertError(error, TypeError, "wrong type");
```

**Describe alternatives you've considered**

If someone has an error they want to make an assertion about, they could currently use assertThrows by passing it a callback that throws the error.

```ts
const error = new TypeError("wrong type");
assertThrows(() => throw error, TypeError, "wrong type")
```

**Additional context**

I have a function that catches errors and passes them to another function to be handled. When I stub that function to capture the error it was called with, I have to use assertThrows and re-throw the error to be able to use the error checking that assertThrows and assertThrowsAsync already have built in.

kt3k:
It seems that the given example:

```ts
assertError(error, TypeError, "wrong type");
```

can be rewritten as:

```ts
assert(error instanceof TypeError);
assertEquals(error.message, "wrong type");
```

So I think providing assertError as public API is superfluous..

KyleJune:
> It seems that the given example:
> 
> ```ts
> assertError(error, TypeError, "wrong type");
> ```
> 
> can be rewritten as:
> 
> ```ts
> assert(error instanceof TypeError);
> assertEquals(error.message, "wrong type");
> ```
> 
> So I think providing assertError as public API is superfluous..

Maybe my example isn't the best. If you look at the actual assertThrows functions they have about 20 lines in common specifically for checking the error object. The second arg is actually msgIncludes and it strips out the color if there is any. The functions also provide nice assertion messages too.

PR #1376, opened 2021-10-10
feat(testing): Add assertIsError

KyleJune:
Closes https://github.com/denoland/deno_std/issues/1182

The assertThrows and assertRejects functions had common code for making assertions about errors that is useful on its own. This change moves the shared code to its own function called assertError. I updated assertThrows and assertRejects to use assertError.


KyleJune:
> If we want to factor out duplicate code for maintenance purposes, this should be private (by either not exporting it or calling it `_assertError`). Having it in the public API is confusing and it's very niche.

I don't think it would be confusing or too niche. Two other people have found my issue and liked it. This saves people from having to re-throw an error to be able to use the error assertions available in assertThrows and assertRejects.

With assertError, if I've already caught the error and would like to make an assertion about it, I could do the following.

```ts
assertError(error, CustomError, "fail");
```

Without assertError, I would need to create a function and make it re-throw the error like this.

```ts
assertThrows(
  () => {
    throw error;
  },
  CustomError,
  "fail",
);
```

nayeemrmn:
Any data structure benefits from having a specialised assertion function to check values with knowledge of the structure. We only have a utility for _thrown_ errors because that's a central case with special control flow behaviour and more boilerplate to hide. IMO the assertion utilities we have should strive to add more value than being able to express multiple expectations at once as positional arguments. This doesn't improve error messages much compared to @kt3k's example in https://github.com/denoland/deno_std/issues/1182#issuecomment-912279645.

kt3k:
I'm in favor of this addition. Some people prefer returning errors instead of throwing them. This utility supports that style of coding better.

wojpawlik:
:+1:, this is very useful for TypeScript, even (especially?) outside testing:

```ts
try {
  foo()
} catch (error: unknown /* default since TS 4.4 */) {
  assertError(error, Error)
  // error is now Error, not unknown 
}
```

PR #1384, opened 2021-10-12
fix(datetime): dayOfYear for the southern hemisphere

mathiasrw:
Because of the direction of time to/from daylight savings is
different from the northern and southern hemisphere the effect
of counting seconds between 1st of Jan to another date goes from
adding a few hours (not a problem when counting days) to
subtracting a few hours (a problem when you count days and end up
on the day before).

The solution proposed here is to disregard any timezone information
as the amount of days in UTC is the same as the amount of days in
any local timezone.


mathiasrw:
Oh. Hmm. Ill make sure it is the same email account - and ill rebase the branch so its only one commit.

PR #1404, opened 2021-10-15
Redo `log`

LionC:
This completely replaces the `log` module with a module trying to achieve the following goals:

- Have a very easy to set up global default logger
- Allow custom log levels with a minimal, clean API
- Adhere to the `console`-ish standard of `<logger>.<level>(message)` **also for custom log levels**
- Have a standard way for frameworks to log
- Allow complete customization but have sane preconfigured options out of the box

The `README` and `doc.deno.land` on the module should give an overview.

The design of this module is heavily influenced by my discussions with @timreichen and input by @caspervonb .

This is marked as draft because it is still missing tests - I prioritized finishing docs so the picture is complete, as I promised @kt3k to post this tonight ;-)


LionC:
Some notes: I typed most of this on the train and will not be able to fix the resulting typos (one of the broken Macbook keyboards -.-) or words I have missed while revisind until my return on Monday.

log/default_logger.ts
```diff
@@ -0,0 +1,118 @@
...
+ * ```
+ */
+export function addDefaultLogger(
+  logger: Logger<DefaultLogLevels, string, unknown>,
+) {
+  defaultLoggerConsumers.push(logger);
+}
+
+/**
+ * The default logger. Does nothing by default, as it is ane empty multi logger - use `addDefaultLogger` to add loggers to handle it's meessages.
```
bartlomieju, line 55:
```suggestion
 * The default logger. Does nothing by default, as it is an empty multi logger - use `addDefaultLogger` to add loggers to handle its messages.
```

log/default_logger.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * If you are writing an application, consider using this if you want to use th default log levels.
```
bartlomieju, line 57:
```suggestion
 * If you are writing an application, consider using this if you want to use the default log levels.
```

log/default_logger.ts
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * ```typescript
+ * import { log } from "https://deno.land/std/@STD_VERSION/log/mod.ts"
+ *
+ * log.info('Some messge')
+ * log.debug('A debug message')
+ * log.warn('Warning')
+ * ```
+ */
+export const log = defaultLogger;
```
bartlomieju, line 71: This is a bit surprising that it doesn't even use `console.log` under the hood; could you elaborate on the decision @LionC?

log/default_logger.ts
```diff
@@ -0,0 +1,118 @@
...
+
+/**
+ * Creates a file loggre with thee default log levels using the givn threshold.
```
bartlomieju, line 74:
```suggestion
 * Creates a file logger with the default log levels using the given threshold.
```

log/builtin_loggers.ts
```diff
@@ -0,0 +1,131 @@
...
+export type FileLogger<L extends LogLevels, M, A> =
+    & Logger<L, M, A>
+    & {
+        flush(): void
+        close(): void
+    }
+/**
+ * Creates a file logger that writes log messages into the given file. Will create the file
+ * if it does not exist yet and append to it if it does by default, but can be configured
+ * with a subset of `Deno.OpenOptions`.
```
bartlomieju, line 32:
```suggestion
 * Creates a file logger that writes log messages into the given file. By default a new file will be created if
 * one does not exist yet or append to already existing file. This behavior can be configured
 * with a subset of `Deno.OpenOptions`.
```

log/default_logger.ts
```diff
@@ -0,0 +1,118 @@
...
+  return buildFileLogger(
+    defaultLogLevels,
+    threshold,
+    filename,
+  );
+}
+
+/**
+ * Creates a console logger with the default levels using the given threshold. If you simply want to
+ * log to stdoout, this is probably what you want to use.
```
bartlomieju, line 89:
```suggestion
 * log to stdout, this is what you want to use.
```

log/logging.ts
```diff
@@ -0,0 +1,164 @@
...
+ */
+export type LogDispatcher<L extends LogLevels, M, A> = (
+  logLevels: L,
+  thresholdLevel: keyof L | null,
+  handler: LogHandler<L, M, A>,
+  ...handlerArgs: Parameters<typeof handler>
+) => void;
+
+/**
+ * Default log message builder useed by all builtin loggers. Converts message
```
bartlomieju, line 46:
```suggestion
 * Default log message builder used by all builtin loggers. Converts message
```

log/logging.ts
```diff
@@ -0,0 +1,164 @@
...
+export function buildDefaultLogMessage<L extends LogLevels, M, A>(
+  ...[logLevel, message, additionalData]: Parameters<LogHandler<L, M, A>>
+) {
+  return `[${logLevel}]\t[${new Date().toLocaleString()}]\t${
+    toLoggableString(message)
+  }${additionalData ? ` ${toLoggableString(additionalData)}` : ""}`;
+}
+
+/**
+ * Default string converter for arbitrary data to be logged
```
bartlomieju, line 59:
```suggestion
 * Default string converter for arbitrary data to be logged.
```

log/logging_test.ts
```diff
@@ -0,0 +1,181 @@
...
+    LogHandler,
+    toLoggableString,
+} from "./logging.ts";
+import { assert, assertEquals, assertMatch } from "../testing/asserts.ts";
+
+const testLevels = {
+  low: 1,
+  middle: 2,
+  high: 3,
+};
```
bartlomieju, line 14: Would be great to have some subprocess that tests handler using `console.log`

bartlomieju (commented):
@LionC very interesting. There's quite some concepts to digest here, would you mind adding a `README` that would present the popular use cases?

@timreichen what do you think about this proposal?

LionC:
> @LionC very interesting. There's quite some concepts to digest here, would you mind adding a `README` that would present the popular use cases? I like how it removes 1,5k lines of code.

Definitely! This is still in draft PR status because it is missing docs and tests. Working on those as we speak.

timreichen:
@bartlomieju sorry for the late response.
I still would favour a class based logger for std for the reasons I have mentioned before:
- factory functions in that case don't feel like native js code but *frameworkish*
- classes are extendable and therefore are more flexible imo
Although I see that this approach is simple to use, It is not really obvious what it is going on under the hood and one needs first to read the docs or the code itself. It is not the right approach for std imo.

LionC:
From the talks we had about this, I think Tim and I agree on the overall composition concepts (log = dispatch -> handle, if you want multiple independent handlers there is some form of multi-logger that is arbitrarily nestable). Also as we have read each others proposals back and forth now and incorporated each other ideas into the different versions, this will be a collaborative result and a definitive improvement over the current module no matter what^^ We just have the luxury to decide between two versions.

I added my (unnecessarily long, bikeshedding yay^^) perspective on the actual differences between the two proposals below - ultimately, the first point is the most important for me, because that one potentially affects all users who will ever log, not just logger authors.

## The API the **loggers** (not the module) expose

This proposal sticks to the `console`-ish standard no matter the log levels, so if you use custom log levels, say `audit`, `log`, `critical`, the API to the users of your **logger** is
    
```typescript
logger.audit("Things happened")
logger.critical("Something went deeply wrong")
logger.log("Someone did a thing that you should know about")
```
    
The goal here was to make the actual logging part as concise as possible, because loggers get called a lot and those calls are all over the place in a codebase, while also sticking to a web standard(-ish) that JS devs are very familiar to.

Tims proposal (at least when we discussed this and when I played around with his branch) did not do this, because, as far as I have explored, there is just no way to do that in a type safe manner with classes without having some wrapped API (which defeats the whole purpose of not having those).

## To class or not to class

For me, passing around callbacks is not a weird thing in JS - from `.setInterval()`, `.map()` etc. to event or request handlers, passing a function to something to give it behaviour has never been something that I saw JS devs struggle with. Classes on the other hand, especially if we talk about abstract classes and inheritance, have pretty complex semantics, verbose code and several pitfalls - which does not mean that one should never use them, but with JS being a style agnostic multi paradigm language, we have the luxury to choose the "right" tool for each job.

In my mind, a logger basically boils down to a callback. Someone logs a message, the callback handles it. All of a loggers behaviour is that callback. The only thing that differentiates a logger from any function, is that it has the concept of levels with a threshold filter (in this proposal, the `dispatcher`), which in almost all use cases will never be touched by consumers of the module (they will often use the default levels and almost always just set some static threshold and probably not be aware that a dispatcher exists).

Yes, that can be modelled as a class and in Java it would definitely be. But Classes expose a very large API surface - that of writing a class with all options and syntax and pitfalls and implications - to someone who wants to write a logger, while we know that they very very likely will just want to set one callback to handle messages.

I would personally also not call these functions factory functions - I do not think the term makes a lot of sense in the context of JS, as anonymous objects are first class citizens in JS (and have been around longer than "real" classes). Factory functions (as a JVM term) are a workaround for complex cases around classes, while functions that return new objects are a very common idiom in JS and not a workaround at all.

So in the end, for me both approaches are valid at their core, but the functions expose a more streamlined, smaller API and generate less code in the module and for consumers writing loggers.

## The rest

There are also some features that this proposal implements that are missing in the class based one, but those are not too relevant for the decision, as they can be implemented in both approaches - that would have to be done though.

log/logging_test.ts
LionC: Yes, I am working on how to do that elegantly

log/default_logger.ts
LionC: That is one of the bigger design decisions. Currently, the global default logger does not do anything - it is essentially a multi logger that references an (unexported) top level array which loggers can be pushed into via very strict functions.

That means that if you want to use the global default logger, for it to do stdout logging, you need to do

```typescript
import {
  addDefaultLogger,
  buildDefaultConsoleLogger,
} from "std/log/mod.ts";

addDefaultLogger(
  buildDefaultConsoleLogger("info")
)
```

to set it up (that could obviously be wrapped in an even shorter function, but that is not the important decision).

The alternative would be for that to be the default, so for the default logger to start out with a console logger in its array. But that would mean that we need a way for users to remove that if they don't want it and a way to change its threshold. That is both possible, we just need to do it and it is a different experience. It widens the amount of central state and the amount of ways that state can be modified from an "append only logger list" to "logger list that can be added to and removed from and some settings for a special value in that list". That felt more complex to me, but I also see the beauty in having a default logger that just...logs without any setup at all.

To adress the obvious alternative: I would not want to remove the ability to add loggers to the default logger set, as that enables users to set up a log file / some log reporting code which will receive all messages (especially from frameworks), which I think it is very important.

TL;DR: I have no strong feelings either way

LionC:
I think this is ready now. I gathered some more feedback and applied several changes:

- As suggested by @bartlomieju , the default logger now logs to the console, together with some simple functions to configure that behaviour
- Third party (library / framework) logging is now an actual feature, including configuration functions
- Finished tests, including subprocesses for console logging and more stability by excluding timestamps in them

I think the pipeline is failing because the current import mechanism in code snippets does not work if the token does not exist in the current release - is that intended or am I missing something?

iuioiua:
Gentle ping - what's the latest update on this PR?

iuioiua:
Gentle ping - what's the latest update on this PR?

ry:
closing because stale

PR #1412, opened 2021-10-16
feat(node): CJS-ESM integration

bartlomieju:
This commit ports most of `lib/internal/modules/esm/resolve.js`.
Alas, I ported all of that code to Rust in https://github.com/denoland/deno/pull/12424
but we have no way to call ESM resolution in Rust from TypeScript code.


kt3k:
@bartlomieju 
Is it possible to write some test cases to exercise the main code paths of this change?

Issue #1418, opened 2021-10-17
`assertObjectMatch` goes wrong when the expected value contains instances of class.


c0per:
**Describe the bug**

`assertObjectMatch` calls `assertEquals` to determine wether two objects are different. Before making the call, it pre-processes(using the function `filter`) the "actual" value to be an object of the intersection of "actual" and "expected", losing the `constructor` info in the "actual" value. Then, feed the `filter` value and raw `expected` value into `assertEquals`.

However, `assertEquals` checks the `constructor` field to determine wether two objects are different. If the "expected" value is an instance of class, it will cause a bug.

**Steps to Reproduce**

```
class A {
    a: number;
    constructor(a: number) {
        this.a = a;
    }
}

assertObjectMatch({ test: { a: 1 } }, { test: new A(1) }); // throws
assertObjectMatch({ test: new A(1) }, { test: new A(1) }); // throws
assertObjectMatch({ test: new A(1) }, { test: { a: 1 } }); // fine
```

It throws an AssertionError when the "expected" value contains a `constructor` field different than `Object`, but I think the test should pass.

**Expected behavior**

Maybe it's better not to depends on `constructor` when matching objects. Or don't ignore the constructor info when doing pre-processing in the `assertObjectMatch` function.

After all it's pretty common to have some instances in the expected value.

**Environment**

- OS: MacOS 11.6
- deno version: 1.15.1
- std version: v0.111.0

c0per:
I made a hot fix to strip `constructor` from `expected`. I think it works fine.

PR #1419, opened 2021-10-17
strip "constructor" info for "expected befor checking in assertObjectMatch

c0per:
I didn't change the `filter` function. I just call `filter` twice to strip `constructor` info off `expected`. #1418


kt3k:
Please add the example assertions in #1418 as test cases

kt3k:
Thank you for looking into this. Left a few comments @c0per

Issue #1194, opened 2021-08-31
Do not respond when requesting a large file in file_server


vi117:
**Describe the bug** In the current implementation, all data to be sent is put in the buffer before sending, so if the file is large, it tries to put all the data in the buffer and send it, so it could not response for long.

**To Reproduce** Steps to reproduce the behavior:

1. turn on file_server.
2. request a large video file.

**Expected behavior** Ceased or delayed response

To fix this, I propose two methods.

The first suggestion is to limit the size of the data to be sent. 
it can be implemented to modify the following code:

https://github.com/denoland/deno_std/blob/305b3647cd68e198998d90a125b68f63471febf6/http/file_server.ts#L311

replace it with the following code:
```ts
const MAX_READ_SIZE = 1024*1024;
const end = parsed && parsed[2] ? +parsed[2] : Math.min(Math.max(0, fileInfo.size - 1),MAX_READ_SIZE+start);
```
The second suggestion would be to change the file to be read whenever needed. It can be solved by adding a `FileSegment` class that implements `Deno.Reader`  and changing the code as follows: 

https://github.com/denoland/deno_std/blob/305b3647cd68e198998d90a125b68f63471febf6/http/file_server.ts#L336-L341
replace it with the following code:
```ts
// Read the selected range of the file
const segment = new FileSegment(file,start,end);
await segment.init();
// Set content length and response body
headers.set("content-length",(end - start + 1).toString());
response.body = segment;
```
The implementation of FileSegment class is followed:
```ts
class FileSegment implements Deno.Reader {
    private file: Deno.Reader & Deno.Seeker;
    private offset: number;
    private end_bytes: number;
    constructor(file: Deno.Reader & Deno.Seeker, start: number, end: number) {
        this.file = file;
        this.offset = start;
        this.end_bytes = end;
    }
    async init() {
        await this.file.seek(this.offset, Deno.SeekMode.Start);
    }
    async read(p: Uint8Array): Promise<number | null> {
        if (this.offset > this.end_bytes) return null;
        const bytes = await this.file.read(p);
        if (bytes === null) return null;
        if (this.offset + bytes <= this.end_bytes) {
            this.offset += bytes;
            return bytes;
        }
        else {
            const remainder = this.end_bytes - this.offset + 1
            this.offset += bytes;
            if (remainder === 0) return null;
            return remainder;
        }
    }
}
```

PR #1421, opened 2021-10-17
feat(http): add file streaming feature in file_server.ts (#1194)

vi117:
Related #1194 
In addition, fixed one bug related to URI. For example, when a file "%25A.txt" and a file "%A" are accessed from a link on the file server, both send the same file "%A".


vi117:
I will create another PR that implements streaming feature.

PR #1430, opened 2021-10-18
fix: node:stream/web wrapper formatting, declaration order

guybedford:
This updates the `stream/web` wrapper to match the Node.js ESM interface wrapper and also to include declarations in the same ordering for easier auditing.


lucacasonato:
Opened an issue for the typings issues here: https://github.com/denoland/deno/issues/12484

kt3k:
The type errors are rooted from the upstream issue in TypeScript repo, and might not be solved anytime soon

I think the alternative solution here could be something like:

```js
const {
  ReadableStream: ReadableStream_,
  ...
} = { 
  ReadableStream,
  ...
};

export {
  ReadableStream_ as ReadableStream,
  ...
};

export default {
  ReadableStream: ReadableStream_
  ...
};
```

bartlomieju:
@guybedford I covered them in https://github.com/denoland/deno/pull/12497

bartlomieju:
@guybedford please sign the CLA

PR #1432, opened 2021-10-19
chore(io): better deprecation message

### LINKED ISSUE #1391: Several `io/util` functions are "deprecated" but provide no information
We have several functions that are deprecated in `/io/util.ts` but there is no documentation on how to resolve the deprecation, making it confusing to user using intelligent IDEs.

In addition, things like `Deno.readAll` are deprecated, but the information with the deprecation points users as using the `/io/util.ts` which are also deprecated.  For example:

```md
@deprecated Use readAllSync from https://deno.land/std/io/util.ts instead. Deno.readAllSync will be removed in Deno 2.0.
```

This should all be cleared up.

kt3k:
This PR adds deprecation messages to deprecated items in std/io.

closes #1391

### DIFF
--- a/io/bufio.ts
+++ b/io/bufio.ts
@@ -4,21 +4,21 @@
 
 import * as buffer from "./buffer.ts";
 
-/** @deprecated */
+/** @deprecated Use BufferFullError from https://deno.land/std/io/buffer.ts instead. */
 export const BufferFullError = buffer.BufferFullError;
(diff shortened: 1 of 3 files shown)


kt3k:
`benching` test case (in `testing/bench_test.ts`) fails randomly in CI (when connecting https://deno.land). It doesn't reproduce locally

PR #1487, opened 2021-10-29
doc(testing): update fast-check examples

cdoremus:
This PR enhances examples that use the fast-check property-based testing library in Deno tests. It supercedes PR #1484 that had commit authorship issues 😳.


cdoremus:
I just realized that instead of using the lodash groupBy function on line 330, I can use the same function in the std collections module. I will submit a new commit for this tomorrow.

bartlomieju (approved):
LGTM, thank you Craig!

PR #1489, opened 2021-10-30
fix(node/events): enable remaining native node tests for EventEmitter

uki00a:
Part of #911

## Summary

- Enabled remaing native node tests for `EventEmitter`
  - **NOTE**: The following tests are added to `"ignored"` field in `node/_tools/config.json`:
      - `test-event-emitter-error-monitor.js` (This depends on the `domain` module that is not currently implemented yet)
      - `test-event-emitter-errors.js` (This depends on `util.inspect` that is not currently compatible with Node.js)
- Added `ERR_UNHANDLED_ERROR` to `node/_errors.ts`
  - `EventEmitter#emit(eventName, ...args)` now throws this error when no `"error"` listeners are registered and `args[0]` is not instance of `Error`


bartlomieju (approved):
LGTM, thank you @uki00a!

PR #1492, opened 2021-10-30
fix(node/_tools): Better error and output logging

Soremwar:
This patch pipes the output from the subprocess tests to `console.log` after the test has finished. That way the test output is marked as output of Deno.test and can be ignored by passing `--quiet` instead of going directly to stdout

Before and after with `--quiet`
image
image

Aditionally, the output of a failed test will be passed to a `fail` call for easier debugging

Before and after
image
image


bartlomieju (approved):
LGTM, thank you @Soremwar this is a huge improvement. Next step would be to get these tests running in `deno test --compat` so they don't need to spawn a subprocess for each file (which on my machine adds about 4s to each test). Let's iterate on that.

PR #1493, opened 2021-10-30
fix(node/_tools): Enable stream tests

Soremwar:
Enable as many tests for streams as possible for an upcoming refactor (RIP people who run this locally)


kt3k (approved):
Wow! Such large number of tests are just working! Great!

Thank you for doing this @Soremwar 

LGTM

PR #1497, opened 2021-10-31
node: add proper module.export for 'module'

bartlomieju:


node/module.ts
```diff
@@ -754,9 +754,20 @@ for (const key of Object.keys(nodeMods)) {
...
+    _initPaths: Module._initPaths,
+    _load: Module._load,
+    _nodeModulePaths: Module._nodeModulePaths,
+    _pathCache: Module._pathCache,
+    _preloadModules: Module._preloadModules,
+    _resolveFilename: Module._resolveFilename,
+    _resolveLookupPaths: Module._resolveLookupPaths,
     default: Module,
-    builtinModules: Module.builtinModules,
-    createRequire: Module.createRequire,
```
kt3k, line 759: What's the reason for deleting these two? I think these 2 are still necessary

bartlomieju: That was a mistake. Fixed now.

kt3k (approved):
LGTM

PR #1499, opened 2021-10-31
node: process.execArgv

bartlomieju:


node/process.ts
```diff
@@ -345,6 +345,9 @@ class Process extends EventEmitter {
    */
   argv = argv;
 
+  /** https://nodejs.org/api/process.html#process_process_execargv */
+  execArgv = [];
```
bartlomieju, line 349: We don't expose these arguments in Deno, so effectively this array will always be empty. I noticed some projects rely on it to sniff out if Node was run with `--inspect` flag. I guess we just have to roll with this solution for now

kt3k: I agree this is enough for now

kt3k (approved):
LGTM

PR #1501, opened 2021-10-31
node: mock 'vm' module

bartlomieju:


kt3k (approved):
Seems enough for a first pass. LGTM

PR #1504, opened 2021-11-01
fix(collections): Prevent calling `Object.prototype.__proto__` on non-Deno platforms in collections/deep_merge.ts

petamoriken:
On Deno, `Object.prototype.__proto__` has been removed, but not always on other platforms. As long as it is explicitly marked as browser compatible, it should be fixed.


kt3k (approved):
LGTM

kt3k:
@petamoriken Thank you for your contribution!

PR #1505, opened 2021-11-01
fix(node/querystring/stringify): Invalid surrogate pair throws URIError

wafuwafu13:

### DIFF
--- a/node/querystring.ts
+++ b/node/querystring.ts
@@ -251,9 +251,37 @@ interface StringifyOptions {
   encodeURIComponent: (string: string) => string;
 }
 
-export function encodeStr(
+/**
+ * These characters do not need escaping when generating query strings:
+ * ! - . _ ~
+ * ' ( ) *
+ * digits
+ * alpha (uppercase)
+ * alpha (lowercase)
+ */
+// deno-fmt-ignore
+const noEscape = new Int8Array([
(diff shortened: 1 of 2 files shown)


node/querystring.ts
```diff
@@ -371,38 +372,42 @@ export function stringify(
...
+          fields += convert(v, encode);
         }
-        fields += ks;
-        fields += convert(v, encode);
       }
+      return fields;
     }
-    return fields;
+  } catch (_) {
+    throw new ERR_INVALID_URI();
```
kt3k, line 410: Is it possible to narrow the kinds of error here? Do you know what is the exact error here when Invalid surrogate pair is given?

wafuwafu13: I needed to use `encodeStr` for exact error handling when an invalid surrogate pair is specified.
https://github.com/nodejs/node/blob/229a182823aa5e60ee5420170ec11e9a541f2163/lib/internal/querystring.js#L90-L97

and I made sure not to export `encodeStr` as it should only be used internally. https://github.com/denoland/deno_std/pull/1505/commits/35eec80f49e7f0f1769f22809fd13262e54f25c3

kt3k (approved):
@wafuwafu13 This looks nicer! LGTM.

PR #1506, opened 2021-11-01
BREAKING(http): update `serve` API

kt3k:
This PR supersedes #1478

This PR updates the APIs in `http/server.ts`.

`serve` now has the following signature:

```ts
async function serve(
  handler: Handler,
  options: ServeInit = {},
): Promise<void>;
```

And `ServeInit` has `addr`, `signal` (for aborting), `tls` options. The addr defaults to ":8000", so the example given in #1478 like the below works:

```ts
serve(req => new Response("Hello world"));
```

(http server starting with the port 8000 by default may look strange, but this has prior examples such as python's http.server, Django, etc. ref: https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers . Alternative default port might be 3000, 8008, 8080, etc)

This `serve` covers the use cases of `listenAndServe` and `listenAndServeTls`. So this PR also deprecates these APIs.

The original `serve` now becomes `serveListener`.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -526,10 +526,33 @@ export class Server {
   }
 }
 
+export interface TlsSettings {
+  /** The path to the file containing the TLS private key. */
+  keyFile: string;
+
+  /** The path to the file containing the TLS certificate */
+  certFile: string;
+}
+
 /** Additional serve options. */
 export interface ServeInit {
+  /**
+   * Optionally specifies the address to listen on, in the form
+   * "host:port". The default is ":8000".
+   *
+   * If the port is omitted, `:80` is used by default for HTTP when invoking
+   * without `tls` options, and `:443` is
+   * used by default for HTTPS when invoking with `tls` options.
+   *
+   * If the host is omitted, the non-routable meta-address `0.0.0.0` is used.
+   */
+  addr?: string;
+
   /** An AbortSignal to close the server and all connections. */
   signal?: AbortSignal;
+
+  /** Optional TLS settings. If specified the method serves with TLS connection */
+  tls?: TlsSettings;
 }
 
 /**
@@ -559,7 +582,7 @@ export interface ServeInit {
 export async function serveListener(
   listener: Deno.Listener,
   handler: Handler,
-  options?: ServeInit,
+  options?: Omit<ServeInit, "addr" | "tls">,
 ): Promise<void> {
   const server = new Server({ handler });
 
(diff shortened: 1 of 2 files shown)


lucacasonato (commented):
There should be a test for `serve`'s default port.

http/server.ts
```diff
@@ -570,6 +593,58 @@ export async function serve(
...
+ * ```ts
+ * import { serve } from "https://deno.land/std@$STD_VERSION/http/server.ts";
+ * const certFile = "/path/to/certFile.crt";
+ * const keyFile = "/path/to/keyFile.key";
+ * console.log("server is starting at https://localhost:8000");
+ * serve((_req) => new Response("Hello, world"), {
+ *   tls: { certFile, keyFile },
+ * });
+ * ```
+ */
```
cmorten, line 618: If only for consistency (but also richer documentation) it would be good to:
1. Include the description/defaults/caveats similar to the other methods, including details on default host and ports (which here are 0.0.0.0 and 8000 opposed to the other methods)
2. Add descriptions to each code block to point out what they show
3. Add the jsdoc `@param`

kt3k: Added docs!

kt3k:
> There should be a test for serve's default port.

Makes sense. Added a test case for it

http/server.ts
```diff
@@ -647,6 +738,9 @@ export async function listenAndServe(
  * @param keyFile The path to the file containing the TLS private key.
  * @param handler The handler for individual HTTP requests.
  * @param options Optional serve options.
+ *
+ * @deprecated Use `serve` instead. `serve` supports `tls` options, which starts the server with
+ * TLS connections.
```
ry, line 743: Rather than overloading server to handle both TCP and TLS, why not add a `serveTls` function? I would find that clearer...

kt3k: Ok. Let's do that. Updated

http/server.ts
```diff
@@ -528,6 +528,18 @@ export class Server {
 
 /** Additional serve options. */
 export interface ServeInit {
+  /**
+   * Optionally specifies the address to listen on, in the form
+   * "host:port". The default is ":8000".
+   *
+   * If the port is omitted, `:80` is used by default for HTTP when invoking
+   * without `tls` options, and `:443` is
+   * used by default for HTTPS when invoking with `tls` options.
```
vwkd, line 537: Nit: `serveTls` is a separate function now.

kt3k: Good catch! Thanks

ry (approved):
LGTM

http/server.ts
```diff
@@ -599,6 +697,8 @@ export async function serve(
  * @param addr The address to listen on.
  * @param handler The handler for individual HTTP requests.
  * @param options Optional serve options.
+ *
+ * @deprecated Use `serve` instead.
```
lucacasonato, line 701: Can you put these all the way at the top of the JSDOC?

kt3k: Sure. Done!

vwkd:
Is it on purpose that the methods on the `Server` class were not renamed?

kt3k:
> Is it on purpose that the methods on the Server class were not renamed?

I just didn't care much about it. Might be better to align those as well.

(BTW I also feel a little unsure about the use case of `Server` class (I guess `serve` and `serveListener` should cover the most of use cases.)

PR #1507, opened 2021-11-02
fix(node): Use "buffer" package instead of custom implementation

Soremwar:
The new module is a bundle of the https://github.com/feross/buffer package + a transformation of the @types/node package that should handle most of the missing functionality and inconsistencies with Node's implementation

UPDATE: So this escalated quickly into a mix between the buffer package and original NodeJS code so I guess we gotta credit...both? None? Also it's gonna be awful to mantain, so I hope the buffer implementation doesn't change much in the following years

Pending adding Node tests to ensure correctness


kt3k:
> So this escalated quickly into a mix between the buffer package and original NodeJS code so I guess we gotta credit...both? None?

I think that should be both

> Also it's gonna be awful to mantain, so I hope the buffer implementation doesn't change much in the following years

Sounds a little worrying.. it'll be hard to maintain because of the mixture of `feross/buffer` and Node.js buffer? If so, why can't we follow either one of those?

Soremwar:
What I'm gonna do is first make all the tests pass then start beautifying the code base

I don't think we are gonna be able to convert this to a class, since many applications expect Buffer to be a ES5 object, so pretty much everything is gonna be prototype based, that's what I mean by hard to mantain. However that doesn't mean it's gonna be unreadable, just way uglier than anything else on std

bartlomieju:
@Soremwar this patch is needed by @lucacasonato for his TLS work too, I think we should land it even if some tests are not enabled yet and do that in the follow ups.

Soremwar:
@bartlomieju Enabling the most possible will help us iron out any inconsistencies in this implementation, I have found and corrected several already and I'm sure there are still many I haven't discovered

We could skip the cleanup of the codebase as a requirement in the meanwhile though, if you want to land it as soon as possible

Soremwar:
@kt3k Ready for review. Most of the bulk of this PR is the tests added, changes are mostly concentrated on `_buffer.js` and `buffer.d.ts`

node/_errors.ts
```diff
@@ -447,6 +469,8 @@ export class ERR_INVALID_ARG_VALUE extends NodeTypeError {
       `The ${type} '${name}' ${reason}. Received ${inspected}`,
     );
   }
+
+  static RangeError = ERR_INVALID_ARG_VALUE_RANGE;
```
bartlomieju, line 473: Nice! I just found myself in the need of this error class.

kt3k (approved):
@Soremwar Thank you for the huge contribution! LGTM

PR #1508, opened 2021-11-02
fix(node/_tools): Remove unused tests

Soremwar:
Apparently, after #1493 I forgot to run "node/_tools/setup.ts" one time more to clean unused tests from disk :facepalm: 

This simply deletes tests commited to the repo that aren't part of the enabled tests


kt3k (approved):
LGTM. Thanks!

PR #1509, opened 2021-11-02
feat(node/_tools): Add cache and no cache flags for test setup

Soremwar:
Nice quality of live improvement so the test setup doesn't have to prompt us everytime time (:upside_down_face:) whether to use cache or not

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -113,6 +113,16 @@ workflow.
 $ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
 ```
 
+You can aditionally pass the `-Y`/`-N` flag to use test cache or generating
+tests from scratch instead of being prompted at the moment of running it.
+
+```zsh
+# Will use downloaded tests instead of prompting user
+$ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
+# Will not prompt but will download and extract the tests directly
+$ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
+```
+
 To run the tests you have set up, do the following:
 
 ```zsh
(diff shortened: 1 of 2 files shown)


node/_tools/setup.ts
```diff
@@ -22,8 +22,30 @@ import { downloadFile } from "../../_util/download_file.ts";
  * section of the configuration file
  *
  * Usage: `deno run --allow-read --allow-net --allow-write setup.ts`
+ *
+ * You can aditionally pass a flag to indicate if cache should be used for generating
+ * the tests, or to generate the tests from scratch (-Y/-N)
  */
 
+const args = Deno.args.map((arg) => arg.toUpperCase());
```
kt3k, line 30: case insensitive flags seem a little strange to me. How about just accept `-y` and `-n`?

node/README.md
```diff
@@ -113,6 +113,16 @@ workflow.
 $ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
 ```
 
+You can aditionally pass the `-Y`/`-N` flag to use test cache or generating
+tests from scratch instead of being prompted at the moment of running it.
+
+```zsh
+# Will use downloaded tests instead of prompting user
+$ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
```
kt3k, line 121: `-y` at the end seems missing

kt3k:
@Soremwar This seems useful. Thank you for your contribution! Left a few comments

Soremwar:
@kt3k Sorted it out

kt3k (approved):
@Soremwar LGTM. Nice!

PR #1510, opened 2021-11-02
fix(node/process): warn on not implemented event instead of throw

kt3k:
This PR changes the handling of not implemented events in process object. Currently it throws on notImplemented events but this PR changes it to just show the warnings about it.

This enables some of npm modules running in Deno compat mode (e.g. mocha)


bartlomieju (approved):
LGTM, this is desirable change for ESLint compat too

PR #1512, opened 2021-11-03
refactor(http/file_server): use default and alias features of flag parser

kt3k:
This PR refactors the default value and alias flag handling in http/file_server using `default` and `alias` options of `parse` function.


ry (commented):
Is this a breaking change?

kt3k:
No. This should be backward compatible

ry (approved):
LGTM

PR #1513, opened 2021-11-03
refactor(http/file_server): remove html template literal tag

kt3k:
`html` template literal tag in file_server is unnecessarily complex, but not useful at all. This PR removes it.


ry (approved):
LGTM

PR #1515, opened 2021-11-03
fix(crypto/digest): always return the underlying ArrayBuffer

sjx233:


kt3k (approved):
Right. These should return ArrayBuffers, not views.

LGTM

PR #1517, opened 2021-11-03
chore: add tools for lint and format

kt3k:


bartlomieju (commented):
This looks good to me, but why no script for running tests?

kt3k:
The `test` script strangely caused an issue (only on mac) in looking up test cases. It caused the infinite logging of the following pattern:

スクリーンショット 2021-11-09 19 47 33

I suspect something's wrong in the look up algorithm of doc testing, but I couldn't find out the cause nor workaround

bartlomieju:
> The `test` script strangely caused an issue (only on mac) in looking up test cases. It caused the infinite logging of the following pattern:
> 
> I suspect something's wrong in the look up algorithm of doc testing, but I couldn't find out the cause nor workaround

This is disconcerting. We should get to the bottom of it before landing this PR. I presume this might be because of different shell expansion?

bartlomieju (approved):
Actually, let's land it as is. @kt3k can you add a mention about the scripts in the README?

LionC:
Now that we have those unified into scripts, could we add a `Makefile` exposing calls to them and having a convenient local git hook setup command?

bartlomieju:
> Now that we have those unified into scripts, could we add a `Makefile` exposing calls to them and having a convenient local git hook setup command?

@LionC I'm not a fan of Makefiles as they are not portable, so I'm not keen on adding it. As for the git hooks, it's okay, but doesn't spark joy. When I'm developing std I'm running format/lint multiple times anyway, but I guess it might be useful for 1st time contributors (as long as these hooks would just work without additional commands to set them up)

PR #1520, opened 2021-11-04
fix(node/util): Deno.permissions is no longer called unless it exists.

### LINKED ISSUE #1518: https://deno.land/std@0.113.0/node/_util/_debuglog.ts does not work in Deploy
**Describe the bug**

https://deno.land/std@0.113.0/node/_util/_debuglog.ts uses `Deno.permissions`, an API not available in Deploy. Just importing this module fails with `TypeError: Cannot read properties of undefined (reading 'query')`.

**Steps to Reproduce**

https://dash.deno.com/playground/long-pheasant-80

**Expected behavior**

If the permissions API is not available, don't try to call it.

grian32:
fixes #1518

see: 
https://dash.deno.com/playground/loud-sheep-97

### DIFF
--- a/node/_util/_debuglog.ts
+++ b/node/_util/_debuglog.ts
@@ -104,10 +104,10 @@ function debuglog(
 const state = "";
 
 if (Deno.permissions) {
-  state = await Deno.permissions.query({
+  state = (await Deno.permissions.query({
     name: "env",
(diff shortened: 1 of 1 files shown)


kt3k (approved):
@grian32 Nice! Thanks!

PR #1529, opened 2021-11-06
chore(node/events): avoid directly emitting 'warning' event

uki00a:
This PR replaces `process.emit('warning', warning)` with `process.emitWarning(warning)`. (I had overlooked presence of `process.emitWarning`)


kt3k (approved):
@uki00a LGTM

PR #1531, opened 2021-11-07
docs(node/querystring): Add docs

wafuwafu13:
- Change functions order(No logical change)
- Add docs
- Remove unnecessary `hexTable` export
- Clean up deno test

### DIFF
--- a/node/querystring.ts
+++ b/node/querystring.ts
@@ -2,7 +2,114 @@
 import { Buffer } from "./buffer.ts";
 import { ERR_INVALID_URI } from "./_errors.ts";
 
-export interface ParsedUrlQuery {
+/******************** querystring.decode() ********************/
+
+/**
+ * Alias of querystring.parse()
+ * @deprecated
+ */
+export const decode = parse;
+
+/******************** querystring.encode() ********************/
+
+/**
+ * Alias of querystring.stringify()
+ * @deprecated
+ */
+export const encode = stringify;
+
+/******************** querystring.escape(str) ********************/
+
+const hexTable = new Array(256);
+for (let i = 0; i < 256; ++i) {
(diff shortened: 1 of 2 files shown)


node/querystring.ts
```diff
@@ -2,7 +2,114 @@
 import { Buffer } from "./buffer.ts";
 import { ERR_INVALID_URI } from "./_errors.ts";
 
-export interface ParsedUrlQuery {
+/******************** querystring.decode() ********************/
+
+/**
+ * Alias of querystring.parse()
+ * @deprecated
```
kt3k, line 9: querystring API are tagged as `Legacy` and it seems that `legacy` is different from `deprecated` in Node.js documentation (ref: https://github.com/nodejs/node/pull/37784 for example )

So maybe we should tag this as `@legacy`?

node/querystring.ts
```diff
@@ -2,7 +2,114 @@
 import { Buffer } from "./buffer.ts";
 import { ERR_INVALID_URI } from "./_errors.ts";
 
-export interface ParsedUrlQuery {
+/******************** querystring.decode() ********************/
```
kt3k, line 5: I'm afraid this kind of comments are just confusing to other contributors because we don't have rule to group functions & variable by the position in the source code

wafuwafu13: I reffered https://github.com/denoland/deno_std/blob/905408229b64a4b8b1b3886135fb7f1dae13417d/testing/fast_check_example.ts but this is just example so I deleted. https://github.com/denoland/deno_std/pull/1531/commits/7d04d5fd179fc24a4056b9a0515639a6ad2d52a9

node/querystring.ts
wafuwafu13: :+1: https://github.com/denoland/deno_std/pull/1531/commits/349b5f10310055c3c205b7f11fb2d560efb6d5a9

Issue #1525, opened 2021-11-05
Remove findLast and findLastIndex from std/collections


lucacasonato:
This feature is landing as a native JS feature in V8 9.7. https://v8.dev/blog/v8-release-97#findlast-and-findlastindex-array-methods

LionC:
Let's do that once that is released!

getspooky:
I agree with @LionC

lucacasonato:
@LionC Will be in 1.16 on Tuesday.

kt3k:
How about deprecating first before removing?

getspooky:
The function is still available using native JS feature in `V8 9.7`, I suggest that we add alert message in the documentation for example:
```txt
`findLastIndex` is available as native function in javascript. learn more https://v8.dev/blog/v8-release-97#findlast-and-findlastindex-array-methods
```

erfanium:
@lucacasonato i guess this issue should be closed

PR #1532, opened 2021-11-08
chore(collections): deprecate findLast

lucacasonato:
Towards #1525﻿

### DIFF
--- a/collections/find_last.ts
+++ b/collections/find_last.ts
@@ -2,6 +2,8 @@
 // This module is browser compatible.
 
 /**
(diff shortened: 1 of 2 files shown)


satyarohith (approved):
LGTM.

kt3k (approved):
LGTM. Thanks!

The commit tag might better to be `BREAKING` to get more attention. We did that for the deprecations of `fs.exists`, `std/hash` recently ref: https://github.com/denoland/deno_std/blob/main/Releases.md#01110--20211012

PR #1533, opened 2021-11-08
fix(node): @ts-ignore Error.captureStackTrace()

### LINKED ISSUE #1523: node/_errors.ts is not compatible with lib.dom
lib.dom.d.ts does not include typings for `Error.captureStackTrace`. This API is used in `node/_errors.ts` however. This API use should get a `// @ts-ignore` so it works in lib.dom.d.ts.

```
error: TS2339 [ERROR]: Property 'captureStackTrace' does not exist on type 'ErrorConstructor'.
    Error.captureStackTrace(err);
          ~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.113.0/node/_errors.ts:63:11
```

lucacasonato:
This call is not present in lib.dom.d.ts. As such, we should @ts-ignore
it.

Fixes #1523

### DIFF
--- a/node/_errors.ts
+++ b/node/_errors.ts
@@ -59,6 +59,7 @@ export function hideStackFrames(fn: GenericFunction) {
 
 const captureLargerStackTrace = hideStackFrames(
   function captureLargerStackTrace(err) {
+    // @ts-ignore this function is not available in lib.dom.d.ts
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

bartlomieju (approved):
LGTM too

PR #1534, opened 2021-11-08
feat(node): add wasi module

talentlessguy:
since [Node.js WASI](https://nodejs.org/api/wasi.html#wasiinitializeinstance) is 1:1 to Deno WASI API, we can just re-export `Context`

### DIFF
--- a/node/README.md
+++ b/node/README.md
@@ -53,7 +53,7 @@ Deno standard library as it's a compatibility module.
 - [x] util/types _partly_
 - [ ] v8
 - [ ] vm
-- [ ] wasi
+- [x] wasi
 - [ ] webcrypto
 - [ ] worker_threads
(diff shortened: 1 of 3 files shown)


bartlomieju:
@talentlessguy can you enable some of the existing Node tests?

kt3k:
Mostly looks good to me. @talentlessguy Can you also add the entry in `node/module_all.ts`?

talentlessguy:
ok I will set node tests locally and add this entry to `modal_all.ts` soon

kt3k:
Thanks. BTW I think it's ok to merge without native tests enabled as first pass if it requires too much work

kt3k (approved):
@talentlessguy LGTM. Thanks!

PR #1537, opened 2021-11-09
chore: skip 'serve should handle websocket requests'

kt3k:


kt3k:
Memo: The issue seems caused by https://github.com/denoland/deno/pull/12332. We'll resolve the issue in CLI

Issue #1485, opened 2021-10-28
Import of node modules in webworkers cause error


AndyDecker:
**Describe the bug**
When using libraries from node the std@0.106.0 is called to load the module. The std tries to call window.addEventListener, which isn't defined in a webworker.

backend_1       | error: Uncaught (in worker "") ReferenceError: window is not defined
backend_1       |     window.addEventListener("unload", () => {
backend_1       |     ^
backend_1       |     at new Process (https://deno.land/std@0.106.0/node/process.ts:133:5)
backend_1       |     at https://deno.land/std@0.106.0/node/process.ts:308:17
backend_1       | error: Uncaught (in promise) Error: Unhandled error event in child worker.
backend_1       |     at Worker.#pollControl (deno:runtime/js/11_workers.js:267:21)

**Steps to Reproduce**

**Expected behavior**

The webworker runs with the loaded script just like when running the code outside of a webworker.

**Environment**

- OS: Ubuntu 20.04
- deno version: 15.1
- std version: 106 (113 still uses the window call)

**Solution ideas**
use self instead of window (pretty new to deno, so don't know if that is applicable here)

PR #1543, opened 2021-11-10
fix(node/process): don't use window.addEventListener

bartlomieju:
Closes https://github.com/denoland/deno_std/issues/1485


kt3k (approved):
Nice test! LGTM

PR #1551, opened 2021-11-11
perf: use `Array.from` instead of `new Array`

jonahsnider:
This PR replaces instances of `new Array(n)` with `Array.from({ length: n })`.

This has 2 main benefits:

1. **Performance:**
   `new Array(n)` [**_always_** results in holey arrays](https://v8.dev/blog/elements-kinds#avoid-creating-holes).
   This results in very bad deoptimizations from the V8 engine.

   > Once the array is marked as holey, it remains holey forever — even if all its elements are present later!

   This can have serious performance implications for end-users. (see `chunk` and `unzip` for the most worrying instances of this)

2. **Type-safety:**
   `new Array()` returns `any[]` by default.
   This results in unsafe code which can lead to runtime errors.
   `Array.from` returns `unknown[]` by default, which ensures type-safety.

   While making this PR I discovered one bug in [`_diff.ts`](https://github.com/denoland/deno_std/blob/b15a2c285c6775373d4b5520610ec156d66cd73c/testing/_diff.ts#L82) in which a required property was missing from elements added on creation.
   All other usages of `new Array()` were implicitly `any[]`.

### DIFF
--- a/encoding/_yaml/dumper/dumper.ts
+++ b/encoding/_yaml/dumper/dumper.ts
@@ -872,7 +872,7 @@ function getDuplicateReferences(
   for (let index = 0; index < length; index += 1) {
     state.duplicates.push(objects[duplicatesIndexes[index]]);
   }
-  state.usedDuplicates = new Array(length);
+  state.usedDuplicates = Array.from({ length: length });
 }
 
 export function dump(input: Any, options?: DumperStateOptions): string {
--- a/encoding/_yaml/dumper/dumper.ts
+++ b/encoding/_yaml/dumper/dumper.ts
@@ -872,7 +872,7 @@ function getDuplicateReferences(
   for (let index = 0; index < length; index += 1) {
     state.duplicates.push(objects[duplicatesIndexes[index]]);
   }
-  state.usedDuplicates = Array.from({ length: length });
+  state.usedDuplicates = Array.from({ length });
 }
 
 export function dump(input: Any, options?: DumperStateOptions): string {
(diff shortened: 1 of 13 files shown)


kt3k:
Sounds reasonable to me. Thank you for bringing this up

Does anyone have any concern?

LionC:
There are some minot inconsistencies in the changes - I can submit a review later today.

encoding/_yaml/dumper/dumper.ts
```diff
@@ -872,7 +872,7 @@ function getDuplicateReferences(
   for (let index = 0; index < length; index += 1) {
     state.duplicates.push(objects[duplicatesIndexes[index]]);
   }
-  state.usedDuplicates = new Array(length);
+  state.usedDuplicates = Array.from({ length: length });
```
LionC, line 875: Should probably be

```suggestion
  state.usedDuplicates = Array.from({ length });
```

to be consistent

LionC (commented):
Code wise I was wrong (just found one small syntax point).

However, playing around with the v8 debug build on my machine, it seems that even if you supply an initalization function, v8 will always create a slack traced holey version (back pointing to a packed version) of the array, instead of the COW packed version an array literal would create. I am not sure (and did not find a way to actually test it with a debug flag) if that actually allows the optimizations on e.g. `.map()` that we would want to kick in.

All that said, the change is definitely not harmful :-)

bartlomieju (approved):
LGTM, thanks

PR #1552, opened 2021-11-11
feat(node/util): Add `util.isDate`

wafuwafu13:

[`types`](https://github.com/nodejs/node/blob/6e1629786fe16058e0f12bc1613fae7734bff9b0/lib/util.js#L367) is come from [internalBinding](https://github.com/nodejs/node/blob/6e1629786fe16058e0f12bc1613fae7734bff9b0/lib/internal/util/types.js#L58) but we actually don't touch V8 APIs directly in std/node so [simulating the behavior in typescript](https://github.com/denoland/deno_std/blob/e72289d02705b7ee75227c599f95c17ffe6aeb44/node/_util/_util_types.ts#L90).


kt3k (approved):
@wafuwafu13 LGTM. Thank you!

Issue #1482, opened 2021-10-28
std/node: "process.nextTick" is broken


bartlomieju:
Given this example:
```js
assert.strictEqual(readline.clearScreenDown(writable, common.mustCall()), true);
```

```js
export function clearScreenDown(stream, callback) {
  if (callback !== undefined) {
    validateCallback(callback);
  }

  if (stream === null || stream === undefined) {
    if (typeof callback === "function") process.nextTick(callback, null);
    return true;
  }

  console.log("stream.write", callback);
  const r = stream.write(kClearScreenDown, callback);
  console.log("after stream.write", callback);
  return r
}
```

I get following output:
```
stream.write [Function]
testwritable [Function: bound onwrite]
testwritable callback called [Function: bound onwrite]
after stream.write [Function]
calling mustcall
```
Somehow even though the callback is called in `TestWritable._write` the actual `common.mustCall()` is called the function returns. It suggest to me that `common.mustCall()` gets called on the next tick of event loop, <s>but I'm unable to spot where that happens</s>, actually I think it might be happening here:
https://github.com/denoland/deno_std/blob/e563b562d005a53dd194affe382d0828631bf3ba/node/_stream/writable_internal.ts#L383-L385

But why does it happen using `queueMicrotask`?

CC @kt3k any ideas?

bartlomieju:
The problem is definitely in `process.nextTick`; opened https://github.com/denoland/deno/issues/12735 which needs to be addressed to make it work properly.

PR #1554, opened 2021-11-11
feat(node/readline): Interface, createInterface

### LINKED ISSUE #1511: node: missing readline.createInterface API
```
$ export DENO_NODE_COMPAT_URL=file://dev/deno_std/
$ npm install eslint
$ deno run --compat --unstable --allow-read --allow-env ./node_modules/eslint/bin/eslint.js --init
...
Oops! Something went wrong! :(

ESLint: 8.0.1

TypeError: fn.emit is not a function
    at Enquirer.enquirer.emit (file:///Users/biwanczuk/dev/vite_test/node_modules/enquirer/index.js:215:12)
    at Enquirer.ask (file:///Users/biwanczuk/dev/vite_test/node_modules/enquirer/index.js:127:10)
    at Enquirer.prompt (file:///Users/biwanczuk/dev/vite_test/node_modules/enquirer/index.js:82:20)
    at Function.fn (file:///Users/biwanczuk/dev/vite_test/node_modules/enquirer/index.js:218:23)
    at promptUser (file:///Users/biwanczuk/dev/vite_test/node_modules/eslint/lib/init/config-initializer.js:472:21)
    at Object.initializeConfig (file:///Users/biwanczuk/dev/vite_test/node_modules/eslint/lib/init/config-initializer.js:705:16)
    at main (file:///Users/biwanczuk/dev/vite_test/node_modules/eslint/bin/eslint.js:126:57)
    at Object.<anonymous> (file:///Users/biwanczuk/dev/vite_test/node_modules/eslint/bin/eslint.js:135:2)
    at Module._compile (file:///Users/biwanczuk/dev/deno_std/node/module.ts:251:36)
    at Object.Module._extensions..js (file:///Users/biwanczuk/dev/deno_std/node/module.ts:1276:10)
```

This is function that errors:
```ts
  static get prompt() {
    const fn = (questions, ...rest) => {
      let enquirer = new this(...rest);
      let emit = enquirer.emit.bind(enquirer);
      enquirer.emit = (...args) => {
        fn.emit(...args); // error here
        return emit(...args);
      };
      return enquirer.prompt(questions);
    };
    utils.mixinEmitter(fn, new Events());
    return fn;
  }
```

Something is going wrong when `utils.mixinEmitter(fn, new Events());`.

In addition to that I'm pretty sure this will break on printing interactive prompts due to missing APIs in `readline` module (`node/readline.js`)

bartlomieju:
Closes #1511


node/events.ts
```diff
@@ -166,7 +166,8 @@ export class EventEmitter {
         try {
           listener.apply(this, args);
         } catch (err) {
-          this.emit("error", err);
+          // console.log("err", err);
+          // this.emit("error", err);
         }
```
bartlomieju, line 171: This bit causes stack overflow in case of error thrown from the listener

node/_tools/suites/parallel/test-readline.js
```diff
@@ -0,0 +1,158 @@
...
+    terminal: true,
+    input: input
+  });
+  
+  rl.on('line', common.mustCall((data) => {
+    assert.strictEqual(data, 'abc');
+  }));
+  
+  input.end('abc');
+}
```
bartlomieju, line 28: These tests are not reliable right, eg. if I change `assert.strictEqual(data, 'abc');` to `assert.strictEqual(data, 'abcdef');` no error is thrown and tests still pass. Most likely related to #1482

bartlomieju: Or maybe it doesn't fail because I disabled catching of error in `events.ts`

bartlomieju: Yeah, definitely related to commenting out `events.ts`

kt3k (approved):
LGTM. Great! 🚀

bartlomieju:
@kt3k thanks, but I'm still not satisfied with this PR. I tried using https://www.npmjs.com/package/enquirer and it still doesn't work.

bartlomieju:
So I wanted to try out what's the state of this PR, and decided to run the simplest possible example: https://nodejs.org/api/readline.html#example-tiny-cli And we're really close! It's mostly working with one caveat that after pressing a key nothing happens unless next key press is done. Ie. all "reads" from stdin are delayed by one key press. 

I'm really not sure what causes this situation, my current suspicion is that it might be related to:
https://github.com/denoland/deno_std/blob/6e60b49ba2b483346c577ebafb1fa563b66443fc/node/_stream/readable_internal.ts#L421-L426
where it uses `queueMicrotask` instead of `process.nextTick` (not really since `process.nextTick` is already broken: https://github.com/denoland/deno_std/issues/1482). That said I'm completely not sure if that's the case and it might be a bug in some other "streams" code. @Soremwar any ideas as it seems you were the one to add "streams" code.

To run the example locally:
1. get code of the example into `example.js`
2. `export DENO_NODE_COMPAT_URL=file:///path/to/local/deno_std/`
3. `deno run --unstable --compat -A --no-check ./example.js`

Soremwar:
The reason why it was originally wrote using queueMicrotask was due to a conversation I had during the process of porting streams originally with a Node dev (can't put my finger on who right now) who explained to me how queueMicrotask was an analog to nextTick in the browser. The assumption that they are the exact same thing is obviously wrong, however I am sure they can be polyfilled without exposing a V8 API

I was looking at https://github.com/defunctzombie/node-process trying to understand their rationale behind this concept and they seem to rely on `setTimeout` to achieve this. Since this module has to have been used hundreds of thousands of times in the browser without breaking bundled code I'd say it's a viable alternative

It's alarming that so much code depends on the abstract concept of event loop ticks :/

bartlomieju:
> The reason why it was originally wrote using queueMicrotask was due to a conversation I had during the process of porting streams originally with a Node dev (can't put my finger on who right now) who explained to me how queueMicrotask was an analog to nextTick in the browser. The assumption that they are the exact same thing is obviously wrong, however I am sure they can be polyfilled without exposing a V8 API
> 
> I was looking at https://github.com/defunctzombie/node-process trying to understand their rationale behind this concept and they seem to rely on `setTimeout` to achieve this. Since this module has to have been used hundreds of thousands of times in the browser without breaking bundled code I'd say it's a viable alternative
> 
> It's alarming that so much code depends on the abstract concept of event loop ticks :/ I wish we didn't have to support this

@Soremwar the `process.nextTick` is not a big deal, I will be working on the proper `Deno.core` bindings so we can polyfill it properly, but I will take a look at `node-process` first. I'm not entirely sure that the problem I'm seeing is related to `process.nextTick` I was asking in general do you have any ideas about the "delayed by one key press" situation. I added some console.logs everywhere and want to check what's the output of Node, but it takes some time to compile Node on my machine. Will report back tomorrow what I find.

Soremwar:
No idea, there are too many factors that can cause a race condition when considering streams

However, I'm looking into replacing the custom implementation of streams a la buffer/events right now. That should iron out any inconsistencies that are lingering in the current codebase, because realistically I don't think we can debug what is going on in there

bartlomieju:
> No idea, there are too many factors that can cause a race condition when considering streams
> 
> However, I'm looking into replacing the custom implementation of streams a la buffer/events right now. That should iron out any inconsistencies that are lingering in the current codebase, because realistically I don't think we can debug what is going on in there

Ah great to hear! I was supposed to do that myself after I found out that `process.stdin` is not actually modelled after `tty.Stream`, but rather implemented using `Readable; and requires bolting on several methods and fields with `_Readable` ad-hoc type. Looking forward to your PR!

Tangentially I grepped for usages of `queueMicrotask` across `deno_std/node` and there are several places where it's used in place of `process.nextTick` in Node; I will first look into changing that to `process.nextTick` and then I'll take a stab at fixing `nextTick` (looked at `node-process` and it seems to trivially implement `nextTick` so that's what I will try before adding `Deno.core` bindings).

bartlomieju:
Okay, managed to finally compile Node locally, and I can already see a difference in behavior:
```
deno run --unstable --compat -A --no-check ./foo.js
OHAI> pre read n 0 Error
    at Readable.read (file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:144:35)
    at resume_ (file:///Users/biwanczuk/dev/deno_std/node/_stream/readable_internal.ts:431:12)
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable_internal.ts:424:22
    at fire (deno:ext/timers/01_timers.js:477:7)
    at handleTimerMacrotask (deno:ext/timers/01_timers.js:338:7)
read n 0
pre first howMuchToRead 0
after first howMuchToRead 0
```

```
./out/Release/node ../enquirer/examples/foo.js
OHAI> pre read n 0 Error
    at ReadStream.Readable.read (node:internal/streams/readable:406:33)
    at ReadStream.Socket.read (node:net:629:39)
    at resume_ (node:internal/streams/readable:1017:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
read n 0
pre first howMuchToRead 0
after first howMuchToRead 0
pre read n undefined Error
    at ReadStream.Readable.read (node:internal/streams/readable:406:33)
    at ReadStream.Socket.read (node:net:629:39)
    at flow (node:internal/streams/readable:1041:34)
    at resume_ (node:internal/streams/readable:1022:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21)
read n NaN
pre first howMuchToRead NaN
after first howMuchToRead 0
```
Upon startup Node does **two** reads on the `stdin` while Deno does only one. I suspect this is either and error in the streams implementation or in `readline` port, but I'm leaning towards the former as most of readline tests pass fine.

bartlomieju:
I spent numerous hours debugging the problems with `stdin` I hit and I concur that all problems stem from buggy "streams" implementation. I've opened #1578 for that. I can land the PR as is right now, but I don't see much benefit to it if effectively it still doesn't work. So I guess we're blocked on #1578 on this one.

node/readline.ts
```diff
@@ -1,15 +0,0 @@
...
-  cursorTo,
-  moveCursor,
-} from "./internal/readline/callbacks.js";
-import { emitKeypressEvents } from "./internal/readline/emitKeypressEvents.js";
-
-export default {
-  clearLine,
-  clearScreenDown,
-  cursorTo,
-  moveCursor,
```
bartlomieju, line 13: This module should be kept and reorganized in similar fashion to `buffer.ts` and `events.ts`

bartlomieju:
There's still some more work needed for `readline` module, but I'm to land this PR as is as enabled tests are all passing.

node/readline.ts
bartlomieju: Done

Issue #1295, opened 2021-09-20
Concept Proposal: `std/http/middleware`


LionC:
There have been some discussions here and there (e.g. #1283 )  about middleware in `std/http`. I asked people for some days to post a concept idea and here it is :-)

# `std/http/middleware`

## Goals

- **Establish a middleware concept that enables `std/http` to be used for actual applications** directly in the future. Once a pattern is established, there are already some modules in `std` that could easily be wrapped into out-of-the-box middleware
- **Allow middleware and composed middleware stacks to just be a function** that takes some form of request and returns a response, optionally calling the next middleware. This ensures that we deal with normal call stacks, allows errors to bubble up as expected and reduces the amount of black magic happening at runtime
- **Be completely type safe, including middleware order and arbitrary middleware composition.** This means I want the type checker to stop me from registering a handler on the server that assumes that certain information is available from previous middlewares (e.g. auth info, parsed and validated bodies...), even though that middleware is not present in that handler's chain. Just having a global state that can be typed and is assumed to always be present in every function is not good enough - we are dealing with a chain of functions here, we should leverage Typescript and make sure that that chain actually works type-wise.
- The middleware signature should **be compatible to the `Server`s `Handler` signature**. Composing middleware should always just return a new middleware, so that compositions can be modularized and passed around opaquely

## POC

[Here is a branch](https://github.com/LionC/deno_std/tree/middleware-experiment/http) in which I have built a small dirty POC fullfiling the goals above. **This is just to show the idea**. It is not fleshed out, very rough around a lot of edges, has subpar ergonomics and several straight up bugs. All of them are solvable in several ways and their solution is not vital to the concept, so I left them as they are for the sake of starting a conversation.

I stopped writing as soon as I was sure enough that this can be done reasonably. There are many ways to do this basic concept and a lot of them are viable - I did not want to invest into one of them, just have something to start talking.

### API

The POC contains three components. Their actual runtime code is really small - most of the code around it (and most todos to fix the bugs / ergonomics issues) is just types.

The components are:

- A `Middleware` function type with two important generic type parameters:
  - What type the middleware expects (e.g. it needs a semantic auth field on top of the normal request)
  - Optionally, what information the middleware adds to the request "context" (e.g. validating the body to be a valid `Animal` and adding an `animal: Animal` property) It could be used like this (lots of abstracted functions in here to show the idea):

  ```typescript
  const validateFoo: Middleware<Request, { foo: Foo }> = async (req, con, next) => {
[...]

PR #1555, opened 2021-11-11
BREAKING: feat(http): Request API & Middleware

LionC:
# `std/http` Request API and Middleware

After delaying this again and again to polish it, here is a (evolved) review-ready version of POC in the linked issue #1295   .

Before I go into the ideas and changes here, I want to give some credit. Even though I wrote the code, I want to explicitly mention and thank:

- @keithamus for constant sparring and brainstorming during the process of writing this
- @happenslol for several feedback rounds on the API and inspiration from other languages
- @Hakizu , @andreubotella and @grian32 for feedback on the docs

## Goals

- Make the `std/http` API ergonomic and powerful enough to be a good starting point for actual applications at some point as well as offer smooth out of the box utilities for `deno deploy` HTTP cases
  - Similar to most http frameworks, there should be an ergonomic central `req => res` API offering all information needed regarding a specific request
  - Establish a minimal but complete middleware pattern that can be used to build out-of-the-box and third-party middleware in the future to cover very common use cases like CORS, compresison, deserialization, logging etc. Middleware should be...
    - ...just a function, there should be no magic and no additional IOC style runtime stuff, just a straight callstack
    - ...as composable and modularizable as possible
    - ...able to handle Errors with normal JS constructs
    - ...type(script) safe to use and combine, ensuring correct middleware context, order and termination. Middleware should not be less safe than combining normal functions

## Changes

I try to summarize the changes in the PR below, but I would highly encourage reading the docs - they are an important part of the module and should be good at explaining the concepts.

### Add `Method` enum to the `Status` module

Adds another enum with the HTTP methods which actually have a standardized representation, parallel to `Status`.

### `HttpRequest` Delegate Wrapper

Adds an `HttpRequest`-class (got a better name? please say so!) that wraps a `Request` object, delegating to it (and thus implementing the `Request` API interface itself) and adding convenience methods and properties to have a unified API:

- A `connInfo` property, holding the `connInfo` that was passed separately before
- A lazy `parsedUrl` property to access a `URL` object for the request url (as a first start for convenience APIs)
- A `context` object to hold (initially empty) request context with an `.addContext()` method to add onto it

Note that there should probably be a lot more convenience methods here - like a (`CookieStore` compatible?) lazy cookie reading API, maybe a way to access parsed content types lazily etc - I just wanted to create a space where that is possible and clean up the API without bloating the PR too much.

### **BREAKING**: `Handler` Signature

The `Handler` signature has been reduced to just `req: HttpRequest => Response`.

**This only breaks code that relies on the positional `connInfo` argument that was removed**. Code that just depends on the request will continue to work, as `HttpRequest implements Request`.

This cleans up the API, creating a place to add more request information onto in `HttpRequest` instead of adding more parameters while also enabling the established middleware signature in the change below.

### `Middleware` & `chain()`

Adds a `chain()` function that allows to chain `(req: HttpRequest, next: Handler) => Response` middleware together with very little actual runtime code in a type safe, order aware manner. Chains can be arbitrarily nested, understand when they are terminated and will generally stop you from running into runtime errors.

Terminated chains are actual `Handler`s and thus cann be passed to `serve`:

```typescript
const handler = chain(parseMiddleware)
  .add(validateMiddleware)
  .add(dataHandler);

await serve(handler);
```

The actual runtime code of `chain()` is very small - types omitted, the whole `middleware.ts` file is just:

```typescript
function composeMiddleware(first, second) {
  return (req, next) =>
    first(
      req
      (r) =>
        second(r, next),
    );
}

export function chain(middleware) {
  const copy = middleware.bind({})

  copy.add = (m) =>
    chain(
      composeMiddleware(
        middleware,
        m,
      ),
    );

  return copy;
}
```

But the type safety and resulting DX is made possible by typing code around the `Middleware` type, which is the TS way to write middleware, enforcing middleware authors to be type safe. The typing code mostly constraints what one can do to provide maximum safety and instant LSP feedback.

The typing code looks pretty big, which is mostly because TS does not have a good way to reuse type parameter sets (at least none that I know of).

### README Docs

Adds a table of contents to the `README`, and a section about Middleware, while integrating the other changes and lightly touching up some of the other docs.

Removes docs about the legacy server.

## Try it out

Under `http/middleware/example_server.ts` you find a server using the middleware examples from the docs.


LionC:
Adapting the server tests right now - some things have happened there since the original implementation.

keithamus:
I've validated this pattern with a couple of closed-source projects that currently use the bare net/http server, and it works very well. I'm excited by this change.

kt3k:
@LionC 
The chaining of types of necessary context and additional context looks great & useful 👍 I like the idea. But the signature of `Middleware` and `chain` utility looks a little magical to me. Isn't it possible to make `Middleware` something like `(f: Handler) => Handler` (with `Needs` and `Adds` type params)? And chaining those like `middlewareA(middlewareB(middlewareC(handler)))`. I guess that might be more natural and memorable

kt3k:
I think the custom `HttpRequest` is a little too opinionated. It introduces learning barrier to the very first example of server in Deno..

Isn't a similar thing possible with `Handler` of type `(req: Request, context: Context) => Response` where Request is Web standard request?

ry:
I'm concerned this is incompatible with how middleware works in Oak (cc @kitsonk)

kt3k:
Oak's Middleware depends on oak specific concepts like Context, Application, etc. So I don't think it's reasonable to provide oak-compatible middleware in `std/http`. Rather I suggest we should provide an adapter utility which converts `std/http`-Middleware to oak-Middleware

- oak refs:
  - Middleware: https://github.com/oakserver/oak/blob/main/middleware.ts
  - Context: https://github.com/oakserver/oak/blob/main/context.ts

LionC:
Updated to master, will fix updated tests soon (again^^)

bartlomieju:
@LionC thank you for the PR, but I'm going to close it without a merge. This PR sparked conversation among the core team to support middlewares and we are still debating adding support for `std/http`; with roughly 50/50 split among the team. Current suggestion is to use similar middleware type as https://github.com/lucacasonato/fresh does.

Issue #1496, opened 2021-10-31
node: add fs.read, fs.readSync, fs.write, fs.writeSync APIs


bartlomieju:
These APIs are needed by ESLint and TSC

- https://nodejs.org/api/fs.html#fsreadfd-buffer-offset-length-position-callback
- https://nodejs.org/api/fs.html#fsreadsyncfd-buffer-offset-length-position
- https://nodejs.org/api/fs.html#fswritefd-buffer-offset-length-position-callback
- https://nodejs.org/api/fs.html#fswritesyncfd-buffer-offset-length-position

getspooky:
I will start working on `fs.read`, `fs.readSync`, `fs.write`, `fs.writeSync` APIs

bartlomieju:
@getspooky great! Feel free to open separate PRs for each API if it will help.

talentlessguy:
All of the functions have been implemented in `std/node` so this can be closed

PR #1557, opened 2021-11-11
feat(node/_fs): Add `fs.read` API

getspooky:
Related https://github.com/denoland/deno_std/issues/1496

### DIFF
--- a/node/_fs/_fs_read.ts
+++ b/node/_fs/_fs_read.ts
@@ -49,12 +49,14 @@ export function read(
 
   if (typeof offsetOrCallback === "function") {
     cb = offsetOrCallback;
-  } else if (typeof optOrBuffer === "undefined") {
+  } else if (typeof optOrBuffer === "function") {
     cb = optOrBuffer;
   } else {
     offset = offsetOrCallback as number;
     cb = callback;
   }
+
+  if (!cb) throw new Error("No callback function supplied");
 
   if (optOrBuffer instanceof Buffer || optOrBuffer instanceof Uint8Array) {
     buffer = optOrBuffer;
@@ -65,10 +67,10 @@ export function read(
     position = null;
   } else {
     let opt = optOrBuffer as readOptions;
-    offset = opt.offset;
-    buffer = opt.buffer;
-    position = opt.position;
-    length = opt.length;
+    offset = opt.offset ?? 0;
+    buffer = opt.buffer ?? Buffer.alloc(16384);
+    length = opt.length ?? buffer.byteLength;
+    position = opt.position ?? null;
   }
 
   assert(offset >= 0, "offset should be greater or equal to 0");
@@ -100,13 +102,11 @@ export function read(
     err = error instanceof Error ? error : new Error("[non-error thrown]");
   }
 
-  if (cb) {
-    if (err) {
-      (callback as (err: Error) => void)(err);
-    } else {
-      const data = Buffer.from(buffer.buffer, offset, length);
-      cb(null, numberOfBytesRead, data);
-    }
+  if (err) {
+    (callback as (err: Error) => void)(err);
+  } else {
+    const data = Buffer.from(buffer.buffer, offset, length);
+    cb(null, numberOfBytesRead, data);
   }
 
   return;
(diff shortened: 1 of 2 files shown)


kt3k:
This (6-arguments version of fs.read) looks good to me, but `fs.read` also has 2-arguments and 3-arguments signatures (`fs.read(fd, [options,] callback)`). Can you also add support of those? ref: https://nodejs.org/api/fs.html#fsreadfd-options-callback

getspooky:
@kt3k if this implement looks good i will start implementing `readSync`.

node/_fs/_fs_read.ts
```diff
@@ -0,0 +1,104 @@
...
+    numberOfBytesRead = Deno.readSync(fd, buffer);
+  } catch (error) {
+    err = error instanceof Error ? error : new Error("[non-error thrown]");
+  }
+
+  if (cb) {
+    if (err) {
+      (callback as (err: Error) => void)(err);
+    } else {
+      const data = new Buffer(buffer.buffer, offset, length);
```
kt3k, line 98: This Buffer constructor (of 3 arguments) doesn't seem existing after #1507. So let's use `Buffer.from(arrayBuffer, offset, length)` instead

node/_fs/_fs_read.ts
```diff
@@ -0,0 +1,104 @@
...
+    buffer: Buffer | Uint8Array;
+
+  if (typeof offsetOrCallback === "function") {
+    cb = offsetOrCallback;
+  } else {
+    offset = offsetOrCallback;
+    cb = callback;
+  }
+
+  if (optOrBuffer instanceof Buffer) {
```
kt3k, line 52: We should also include the case of `optOrBuffer instanceof Uint8Array`

kt3k:
@getspooky Thank you for updating. This looks far better than before, but I see a few more issues.

We also need the case of `fs.read(fd, cb)` (2 arguments).

The test cases only call fs.read with 6 arguments. We also need test cases for `fs.read(fd, cb)` and `fs.read(fd, options, cb)` (No need of calling them with various combinations of buffer, length, offset, but at least call them once to confirm the argument swapping is correct.)

node/_fs/_fs_read.ts
```diff
@@ -0,0 +1,113 @@
...
+  let offset: number = 0,
+    buffer: Buffer | Uint8Array;
+
+  if (length == null) {
+    length = 0;
+  }
+
+  if (typeof offsetOrCallback === "function") {
+    cb = offsetOrCallback;
+  } else if (typeof optOrBuffer === "undefined") {
```
kt3k, line 52: I think this should be `typeof optOrBuffer === "function"`, and the test case doesn't seem catching this error

node/_fs/_fs_read_test.ts
```diff
@@ -0,0 +1,143 @@
...
+  name: "[std/node/fs] Read fs.read(fd, cb) signature",
+  async fn() {
+    const file = Deno.makeTempFileSync();
+    Deno.writeTextFileSync(file, "hi deno");
+    const fd = openSync(file, "r+");
+    await read(fd, (err, bytesRead, data) => {
+      assertEquals(err, null);
+      assertStrictEquals(bytesRead, 7);
+      assertStrictEquals(data?.byteLength, 16384);
+    });
```
kt3k, line 140: `read` doesn't return promise. So awaiting it doesn't make much sense. Because we want to confirm that the callback is actuall called, I think the these lines should be something like:
```js
await new Promise((resolve) => {
  read(fd, (err, bytesRead, data) => {
    assertEquals(err, null);
    assertStrictEquals(bytesRead, 7);
    assertStrictEquals(data?.byteLength, 16384);
    resolve();
  });
});
```

node/_fs/_fs_read.ts
```diff
@@ -0,0 +1,113 @@
...
+    offset = 0;
+    buffer = Buffer.alloc(16384);
+    length = buffer.byteLength;
+    position = null;
+  } else {
+    let opt = optOrBuffer as readOptions;
+    offset = opt.offset;
+    buffer = opt.buffer;
+    position = opt.position;
+    length = opt.length;
```
kt3k, line 71: these optional values should have default values such as `opt.offset ?? 0`, `opt.buffer ?? Buffer.alloc(16384)`, etc

kt3k:
@getspooky Thank you for updating! Looks much better, but left a few more comments!

kt3k (approved):
@getspooky LGTM. Thanks you for your contribution!

PR #1558, opened 2021-11-11
node: Use upstream source for "events" module

bartlomieju:


node/_tools/test.ts
```diff
@@ -47,6 +47,7 @@ for await (const file of dir) {
           "-A",
           "--quiet",
           "--unstable",
+          "--no-check",
```
bartlomieju, line 50: This change should be done in a separate PR - I think there's no need to type check on every run if we did it once using `deno cache` command

kt3k: I'm in favor of this change. This command checks the typing of `node/_tools/require.ts` hundreds of times, and that's redundant

kt3k:
hmm many test cases seem failing?

node/_tools/test.ts
bartlomieju: Correct, each subprocess takes at least 2s.

bartlomieju:
@kt3k there are three separate issues:
a) some failing tests due to missing `inspect` implementation (I will either add it or disable the tests for now)
b) I get this error, which suggests that some exports are not done correctly:
```
error: Uncaught TypeError: Class constructor Stream1 cannot be invoked without 'new'
  stream.Stream.call(this);
                ^
    at new Writable (file:///Users/biwanczuk/dev/deno_std/node/_tools/suites/parallel/test-event-emitter-prepend.js:35:17)
    at Object.<anonymous> (file:///Users/biwanczuk/dev/deno_std/node/_tools/suites/parallel/test-event-emitter-prepend.js:47:11)
    at Module._compile (file:///Users/biwanczuk/dev/deno_std/node/module.ts:251:36)
    at Object.Module._extensions..js (file:///Users/biwanczuk/dev/deno_std/node/module.ts:1276:10)
    at Module.load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:230:33)
    at Function._load (file:///Users/biwanczuk/dev/deno_std/node/module.ts:539:14)
    at Module.require (file:///Users/biwanczuk/dev/deno_std/node/module.ts:216:21)
    at require (file:///Users/biwanczuk/dev/deno_std/node/module.ts:1331:16)
    at file:///Users/biwanczuk/dev/deno_std/node/_tools/require.ts:23:1

```
c) there's about 45 type checking errors that need to be fixed, snippet:
```
running 226 tests from file:///Users/biwanczuk/dev/deno_std/node/_tools/test.ts
test suites/parallel/test-net-socket-end-before-connect.js ...error: TS7006 [ERROR]: Parameter 'er' implicitly has an 'any' type.
    this.once(kDestroy, (er) => {
                         ~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/writable_internal.ts:180:26

TS2345 [ERROR]: Argument of type 'Readable' is not assignable to parameter of type 'Stream'.
      destroyer(stream, err);
                ~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/async_iterator.ts:96:17

TS2345 [ERROR]: Argument of type 'IterableStreams' is not assignable to parameter of type 'Stream'.
  Type 'Readable' is not assignable to type 'Stream'.
    Types of property 'removeListener' are incompatible.
      Type '{ (event: "readable" | "close" | "end" | "pause" | "resume", listener: () => void): Readable; (event: "data", listener: (chunk: any) => void): Readable; (event: "error", listener: (err: Error) => void): Readable; (event: string | symbol, listener: (...args: any[]) => void): Readable; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
        Types of parameters 'event' and 'type' are incompatible.
          Type 'string | symbol' is not assignable to type '"readable" | "close" | "end" | "pause" | "resume"'.
            Type 'string' is not assignable to type '"readable" | "close" | "end" | "pause" | "resume"'.
    stream = new Readable({ objectMode: true }).wrap(src);
                                                     ~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/async_iterator.ts:208:54

TS2345 [ERROR]: Argument of type 'IterableStreams' is not assignable to parameter of type 'Stream'.
  Type 'Readable' is not assignable to type 'Stream'.
    finished(stream, (err) => destroyer(src, err));
                                        ~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/async_iterator.ts:209:41

TS2345 [ERROR]: Argument of type 'Streams' is not assignable to parameter of type 'Stream'.
  Type 'Duplex' is not assignable to type 'Stream'.
    Types of property 'removeListener' are incompatible.
      Type '{ (event: "readable" | "close" | "end" | "pause" | "resume", listener: () => void): Duplex; (event: "data", listener: (chunk: any) => void): Duplex; (event: "error", listener: (err: Error) => void): Duplex; (event: string | symbol, listener: (...args: any[]) => void): Duplex; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
        Types of parameters 'event' and 'type' are incompatible.
          Type 'string | symbol' is not assignable to type '"readable" | "close" | "end" | "pause" | "resume"'.
    implDestroyer(stream, err);
                  ~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/pipeline.ts:87:19

TS2345 [ERROR]: Argument of type 'Duplex | Writable' is not assignable to parameter of type 'EventEmitter'.
  Type 'Duplex' is not assignable to type 'EventEmitter'.
    Types of property 'removeListener' are incompatible.
      Type '{ (event: "readable" | "close" | "end" | "pause" | "resume", listener: () => void): Duplex; (event: "data", listener: (chunk: any) => void): Duplex; (event: "error", listener: (err: Error) => void): Duplex; (event: string | symbol, listener: (...args: any[]) => void): Duplex; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
        Types of parameters 'event' and 'type' are incompatible.
          Type 'string | symbol' is not assignable to type '"readable" | "close" | "end" | "pause" | "resume"'.
        await events.once(writable, "drain");
                          ~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/pipeline.ts:156:27

TS2345 [ERROR]: Argument of type 'Duplex | Writable' is not assignable to parameter of type 'EventEmitter'.
  Type 'Duplex' is not assignable to type 'EventEmitter'.
    prependListener(dest, "error", onerror);
                    ~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:365:21

TS2394 [ERROR]: This overload signature is not compatible with its implementation signature.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:462:3

TS2750 [ERROR]:     The implementation signature is declared here.
      removeListener(
      ~~~~~~~~~~~~~~
        at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:474:3

TS2394 [ERROR]: This overload signature is not compatible with its implementation signature.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:462:3

TS2750 [ERROR]:     The implementation signature is declared here.
      removeListener(
      ~~~~~~~~~~~~~~
        at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:474:3

TS2394 [ERROR]: This overload signature is not compatible with its implementation signature.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:462:3

TS2750 [ERROR]:     The implementation signature is declared here.
      removeListener(
      ~~~~~~~~~~~~~~
        at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:474:3

```

bartlomieju:
That said, this PR fixes all problems I encountered in the `readline.createInterface` PR (https://github.com/denoland/deno_std/pull/1554#discussion_r747412813)

bartlomieju:
> > this PR fixes all problems I encountered in the readline.createInterface PR (#1554 (comment))
> 
> That's great! Ok. Let's proceed with this direction
> 
> Regarding some Stream related error, we might also need to do the same about Node.js Streams (but the amount of work would be huge..

@kt3k I think we are out of options on this one - the error suggest that `Stream` should be written using `function` syntax:
```js
function Stream(opts) {
  EE.call(this, opts);
}
```
instead of ES class; I don't think we can do anything about it

bartlomieju:
I have trouble figuring out how to fix these type errors... Right now the types for `EventEmitter` are taken directly from [`DefinitelyTyped`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9b9cd671114a2a5178809798d8e7f4d8ca6c2671/types/node/events.d.ts), and similar for [`Readable`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts) however I'm getting a bunch of incompatible overloads:
```
deno run -A --unstable node/_tools/require.ts /Users/biwanczuk/dev/deno_std/node/_tools/suites/parallel/test-event-emitter-add-listeners.js
Check file:///Users/biwanczuk/dev/deno_std/node/_tools/require.ts
error: TS2416 [ERROR]: Property 'removeListener' in type 'Readable' is not assignable to the same property in base type 'Stream'.
  Type '{ (event: "resume" | "pause" | "readable" | "close" | "end", listener: () => void): this; (event: "data", listener: (chunk: any) => void): this; (event: "error", listener: (err: Error) => void): this; (event: string | symbol, listener: (...args: any[]) => void): this; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
    Types of parameters 'event' and 'type' are incompatible.
      Type 'string | symbol' is not assignable to type '"resume" | "pause" | "readable" | "close" | "end"'.
        Type 'string' is not assignable to type '"resume" | "pause" | "readable" | "close" | "end"'.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:462:3

TS2416 [ERROR]: Property 'removeListener' in type 'Readable' is not assignable to the same property in base type 'Stream'.
  Type '{ (event: "resume" | "pause" | "readable" | "close" | "end", listener: () => void): this; (event: "data", listener: (chunk: any) => void): this; (event: "error", listener: (err: Error) => void): this; (event: string | symbol, listener: (...args: any[]) => void): this; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
  removeListener(event: "data", listener: (chunk: any) => void): this;
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:467:3

TS2416 [ERROR]: Property 'removeListener' in type 'Readable' is not assignable to the same property in base type 'Stream'.
  Type '{ (event: "resume" | "pause" | "readable" | "close" | "end", listener: () => void): this; (event: "data", listener: (chunk: any) => void): this; (event: "error", listener: (err: Error) => void): this; (event: string | symbol, listener: (...args: any[]) => void): this; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
  removeListener(event: "error", listener: (err: Error) => void): this;
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:468:3

TS2416 [ERROR]: Property 'removeListener' in type 'Readable' is not assignable to the same property in base type 'Stream'.
  Type '{ (event: "resume" | "pause" | "readable" | "close" | "end", listener: () => void): this; (event: "data", listener: (chunk: any) => void): this; (event: "error", listener: (err: Error) => void): this; (event: string | symbol, listener: (...args: any[]) => void): this; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:469:3

TS2416 [ERROR]: Property 'removeListener' in type 'Readable' is not assignable to the same property in base type 'Stream'.
  Type '{ (event: "resume" | "pause" | "readable" | "close" | "end", listener: () => void): this; (event: "data", listener: (chunk: any) => void): this; (event: "error", listener: (err: Error) => void): this; (event: string | symbol, listener: (...args: any[]) => void): this; }' is not assignable to type '(type: string | symbol, listener: Function) => EventEmitter'.
  removeListener(
  ~~~~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:474:3

TS2322 [ERROR]: Type 'EventEmitter' is not assignable to type 'this'.
  'this' could be instantiated with an arbitrary type which could be unrelated to 'EventEmitter'.
    return res;
    ~~~~~~~~~~~
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/readable.ts:490:5
```

I'm not sure why it's problematic since the types on `Readable` are more concrete than the ones on `EventEmitter`. @kt3k @kitsonk any ideas?

Soremwar:
I don't think there is a clean solution to this. The current typings for Streams are in a way, much more strict than the ones that are being used in Node land, namely DefinitelyTyped.

DefinitelyTyped only cares about the function signatures, while the internal code does a lot of implicit casts that TypeScript has a hard time keeping up with (some would argue even things that should never be done). The simplest solution to me is to disable type checking for streams in this first patch, and rely on the Node tests to guarantee they are still working

Solving this for streams should be straight forward, since ReadableStream already has a JS only implementation that is fully compliant and mantained by the very NodeJS devs, so this won't be really a compromise if we are to replace the current implementation anyways

bartlomieju:
> I don't think there is a clean solution to this. The current typings for Streams are in a way, much more strict than the ones that are being used in Node land, namely DefinitelyTyped.
> 
> DefinitelyTyped only cares about the function signatures, while the internal code does a lot of implicit casts that TypeScript has a hard time keeping up with (some would argue even things that should never be done). The simplest solution to me is to disable type checking for streams in this first patch, and rely on the Node tests to guarantee they are still working
> 
> Solving this for streams should be straight forward, since ReadableStream already has a JS only implementation that is fully compliant and mantained by the very NodeJS devs, so this won't be really a compromise if we are to replace the current implementation anyways

Thanks for the response @Soremwar; I think we will have to replace our current `Stream` (and friends) implementation anyway for this reason: https://github.com/denoland/deno_std/pull/1558#issuecomment-967078180 

I'm also leaning towards just disabling/ignoring some of the errors encountered here and I think maintaining our own types for Node.js APIs is counter-productive (as is rewriting APIs from JS to TS instead of using JavaScript files directly from Node repo).

bartlomieju:
I had to add about 25 `// @ts-ignore` and a bunch of `as unknown as Stream` or `as unknown as EventEmitter` casts. Hopefully they could be removed once I'll change implementation of `streams` to use upstream code and typing from DefinitelyTyped.

node/events.d.ts
```diff
@@ -0,0 +1,664 @@
...
+ *
+ * const myEmitter = new MyEmitter();
+ * myEmitter.on('event', () => {
+ *   console.log('an event occurred!');
+ * });
+ * myEmitter.emit('event');
+ * ```
+ * @see [source](https://github.com/nodejs/node/blob/v16.9.0/lib/events.js)
+ */
+declare module "events" {
```
Soremwar, line 39: Not a fan of declare module here, IMO this should be an exported namespace

node/events.d.ts
```diff
@@ -0,0 +1,664 @@
...
+     * console.log(listenerCount(myEmitter, 'event'));
+     * // Prints: 2
+     * ```
+     * @since v0.9.12
+     * @deprecated Since v3.2.0 - Use `listenerCount` instead.
+     * @param emitter The emitter to query
+     * @param eventName The event name
+     */
+    static listenerCount(
+      emitter: NodeJS.EventEmitter,
```
Soremwar, line 249: This could be avoided by merging the NodeJS.EventEmitter declaration into the EventEmitter one

Soremwar (commented):
I don't like that there is a lot of TypeScript magic going around in `events.d.ts`, that could lead to some nasty type checking bugs (specially the global declaration for a NodeJS interface)

bartlomieju:
> I don't like that there is a lot of TypeScript magic going around in `events.d.ts`, that could lead to some nasty type checking bugs (specially the global declaration for a NodeJS interface)

@Soremwar I took these declarations directly from `DefinitelyTyped`. I'm also not a fan of what's going on here, but I think it might be easier to do it that way than having to maintain our own type declarations. Aren't there supposed to be the source of truth for Node APIs? (Btw, I think I already hit some of those nasty type checking bugs)

Soremwar:
I had to heavily modify the ones on Buffer to sort the problems on my side, sure we lose some mantainability but how often do these type declarations really change?

bartlomieju:
> I had to heavily modify the ones on Buffer to sort the problems on my side, sure we lose some mantainability but how often do these type declarations really change?

Okay, that's a valid point. I'll try to adjust the type declaration to your suggestion and get back to you.

bartlomieju:
@Soremwar PTAL

kt3k (approved):
LGTM from my perspective, but it would be nice if @Soremwar could take another look

node/events.d.ts
```diff
@@ -0,0 +1,644 @@
...
+  interface EventEmitterOptions {
+    /**
+     * Enables automatic capturing of promise rejection.
+     */
+    captureRejections?: boolean | undefined;
+  }
+  interface NodeEventTarget {
+    once(eventName: string | symbol, listener: (...args: any[]) => void): this;
+  }
+  interface DOMEventTarget {
```
Soremwar, line 51: Can't we just use the EventTarget signature?

node/events.js
```diff
@@ -0,0 +1,1042 @@
...
+// NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+// DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR
+// OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE
+// USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+"use strict";
+
+const kRejection = Symbol.for("nodejs.rejection");
+// TODO(bartlomieju): `inspect` is not ported yet
+// const { inspect } = require('internal/util/inspect');
```
Soremwar, line 28: Isn't this added in this PR?

Soremwar (commented):
Only nitpicks, LGTM

Curious as to why there aren't any Node tests added though, have we added all of them already?

node/events.d.ts
bartlomieju: Not sure what you mean here? Do `EventTarget` instead of `DOMEventTarget`?

bartlomieju:
> Only nitpicks, LGTM
> 
> Curious as to why there aren't any Node tests added though, have we added all of them already?

Good point, I will enable a few more test files.

node/events.d.ts
Soremwar: Yeah, since they seem to be the same thing in this context

Soremwar:
An important aspect of this patch was overlooked. For people using `node/events.ts` this will be a breaking change, I'll submit a fix for this

bartlomieju:
> An important aspect of this patch was overlooked. For people using `node/events.ts` this will be a breaking change, I'll submit a fix for this

Yikes, good point. Looking forward to the PR.

PR #1559, opened 2021-11-12
fix(fs/walk): dont throw when encountering a symlink to a file

### LINKED ISSUE #1358: `walk`/`walkSync` with `followSymlink: true` throws when encountering a symlink to a file (opposed to a folder)
The same probably happens for a symlink to another symlink if I am reading the code right.

ah-yu:
Close #1358


kt3k (approved):
@ah-yu LGTM. Thanks!

PR #1560, opened 2021-11-12
refactor(node/buffer): Remove `FastBuffer` Class

wafuwafu13:
Resolve: https://github.com/denoland/deno_std/blob/e19b170aa911aea0feb7a30bde641452a3d9efec/node/_buffer.ts#L22-L23

I ensured correctness of this change by `test-net-connect-buffer2.js`.

スクリーンショット 2021-11-12 18 29 08

スクリーンショット 2021-11-12 18 29 44

We may need `FastBuffer` Class in the future, but at least we don't want to have `_buffer.js` and `_buffer.ts` on the same hierarchy.


kt3k (approved):
FastBuffer seems like an internal utility for Node.js implementation (it's defined in lib/internal/buffer.js https://github.com/nodejs/node/blob/2d005bb961eb7c3d55684493b77468a3632e4a2a/lib/internal/buffer.js#L954

So I think it's ok to remove it if we don't need it in our implementation.

LGTM

PR #1561, opened 2021-11-12
chore(node): don't typecheck require module when running tests

bartlomieju:
This change should speed up CI runs.


kt3k (approved):
LGTM 👍

PR #1563, opened 2021-11-13
refactor(node/buffer): export `Buffer` instead of using `Buffer2`

wafuwafu13:
- Add space between definition
- export `Buffer` instead of using `Buffer2`
  - That's how it works in node.
  - https://github.com/nodejs/node/blob/f2170253b694c488f8ad2616dfc5c66b6a3c90a0/lib/buffer.js#L1244

### DIFF
--- a/node/_buffer.js
+++ b/node/_buffer.js
@@ -88,10 +88,10 @@ export const constants = {
   MAX_STRING_LENGTH: kStringMaxLength,
 };
 
-Buffer2.TYPED_ARRAY_SUPPORT = typedArraySupport();
+Buffer.TYPED_ARRAY_SUPPORT = typedArraySupport();
 
 if (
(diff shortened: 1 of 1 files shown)


bartlomieju:
@Soremwar PTAL, any reason it was `Buffer2` instead of `Buffer`?

Soremwar:
Since this was originally generated by esbuild and then modified based on Node's implementation, I assume esbuild didn't wanted the Buffer global to conflict if imported in NodeJS. No technical reason, just forgot to clean it up :)

bartlomieju:
Thanks @Soremwar, seems reasonable. I'll review and land tomorrow.

bartlomieju (approved):
LGTM, thank you @wafuwafu13

PR #1564, opened 2021-11-13
test(node/fs): enable `test-fs-write-file.js`

F3n67u:
part of: #911

PR #1568, opened 2021-11-14
feat(node/fs): implement `fs.rm` and `fs.rmSync`

F3n67u:
this is a part of https://github.com/denoland/deno_std/pull/1564;

`fs.rmSync` need by https://github.com/nodejs/node/blob/master/test/common/tmpdir.js#L8


kt3k (approved):
@F3n67u LGTM. Nice work!

PR #1570, opened 2021-11-14
fix(node): Align exports and declarations for Buffer and Events module

Soremwar:
This fixes a breaking change caused by #1558 were `events.ts` was replaced by `events.js` breaking compatibility

This additionally fixes the exports for the buffer and events module to align with NodeJS original ones. I had mistakenly assumed that declaring a namespace and using `export =` would have the same effect as exporting the items manually however this was not the case, hopefully this new structure for `d.ts` files can be used uniformally for new ports of Node modules

Added comments for missing functionality in Buffer


node/_buffer.js
```diff
@@ -1148,59 +1149,63 @@ Buffer.prototype.slice = function slice(start, end) {
   return newBuf;
 };
 
-Buffer.prototype.readUintLE = Buffer.prototype.readUIntLE =
```
Soremwar, line 1151: Don't even know why this wasn't formatted before

node/_buffer.js
```diff
@@ -2075,7 +2082,14 @@ function BufferBigIntNotDefined() {
   throw new Error("BigInt not supported");
 }
 
+export const atob = globalThis.atob;
```
Soremwar, line 2085: Buffer provides this globals as exports of the module

node/_events.d.ts
```diff
@@ -0,0 +1,788 @@
+// deno-lint-ignore-file no-explicit-any
```
Soremwar, line 1: I removed the general lint ignore for a more concrete ignore

node/_events.d.ts
```diff
@@ -0,0 +1,788 @@
...
+ * {
+ *   const et = new EventTarget();
+ *   const listener = () => console.log('Events are fun');
+ *   et.addEventListener('foo', listener);
+ *   getEventListeners(et, 'foo'); // [listener]
+ * }
+ * ```
+ * @since v15.2.0
+ */
+export function getEventListeners(
```
Soremwar, line 36: If you see duplicated comment blocks it's because VSCode doesn't seem to apply those when you do something like `x: typeof type_with_comments` so I figured it would make sense even if it's kinda ugly

node/_events.d.ts
```diff
@@ -0,0 +1,788 @@
...
+  static readonly captureRejectionSymbol: unique symbol;
+  /**
+   * Sets or gets the default captureRejection value for all emitters.
+   */
+  // TODO: These should be described using static getter/setter pairs:
+  static captureRejections: boolean;
+  static defaultMaxListeners: number;
+}
+
+export default EventEmitter;
```
Soremwar, line 788: The static properties and the named exports for EventEmitter are the same functions, however there is no clean way to represent such a thing in TypeScript without doing duplicate declarations

node/events.ts
```diff
@@ -0,0 +1,11 @@
+// @deno-types="./_events.d.ts"
```
Soremwar, line 1: Better to explicitly list them to prevent exports without declarations

node/events.ts
```diff
@@ -0,0 +1,11 @@
...
+export {
+  captureRejectionSymbol,
+  default,
+  defaultMaxListeners,
+  errorMonitor,
+  EventEmitter,
+  getEventListeners,
+  on,
+  once,
+} from "./_events.js";
```
Soremwar, line 11: Better to explicitly list them to prevent exports without types

bartlomieju (approved):
LGTM, thank you @Soremwar, I already hit some of these problems in https://github.com/denoland/deno_std/pull/1554, this should fix the issues.

bnoordhuis:
I'm getting the following lint errors as of this change:
```
$ deno run -A _tools/lint.js 
(ban-types) This provides no type safety because it represents all functions and classes
  listeners(eventName: string | symbol): Function[];
                                         ^^^^^^^^
    at /home/bnoordhuis/src/std/node/_events.d.ts:429:41

    hint: Define the function shape explicitly
    help: for further information visit https://lint.deno.land/#ban-types

(ban-types) This provides no type safety because it represents all functions and classes
  rawListeners(eventName: string | symbol): Function[];
                                            ^^^^^^^^
    at /home/bnoordhuis/src/std/node/_events.d.ts:459:44

    hint: Define the function shape explicitly
    help: for further information visit https://lint.deno.land/#ban-types

Found 2 problems
Checked 648 files
```

PR #1573, opened 2021-11-15
chore(node): Implement `internal/errors`

wafuwafu13:
Errors defined in [`nodejs/node/lib/internal/errors`](https://github.com/nodejs/node/blob/afe460ec9e30add944056760a01d1185c6aac276/lib/internal/errors.js#L390-L408) are implemented in [`node/_errors.ts`](https://github.com/denoland/deno_std/blob/main/node/_errors.ts) and work correctly, but can't import `internal/errors` in test file like `test-util.js`.


kt3k (approved):
LGTM. Thanks!

PR #1574, opened 2021-11-15
fix(node): improve http.Server.listen() compat

bnoordhuis:
Make patterns like `const server = http.createServer(go).listen()` work.

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -1,7 +1,4 @@
-import {
-  _normalizeArgs,
-  ListenOptions,
-} from "./net.ts";
+import { _normalizeArgs, ListenOptions } from "./net.ts";
 import { Status as STATUS_CODES } from "../http/http_status.ts";
 import { Buffer } from "./buffer.ts";
 import { EventEmitter } from "./events.ts";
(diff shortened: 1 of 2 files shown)


node/http.ts
```diff
@@ -217,14 +221,27 @@ export class Server extends EventEmitter {
   }
 
   // TODO(AaronO): support options object
```
AaronO, line 223: That comment should now be obsolete

kt3k (approved):
LGTM

node/http.ts
bnoordhuis, line 223:
```suggestion
```
(I have a distinct memory of removing it :shrug:)

node/http.ts
```diff
@@ -1,3 +1,7 @@
+import {
+  _normalizeArgs,
+  ListenOptions,
+} from "./net.ts";
```
bnoordhuis, line 4:
```suggestion
import { _normalizeArgs, ListenOptions } from "./net.ts";
```

bartlomieju (approved):
LGTM too

PR #1582, opened 2021-11-15
ci: split running tests into steps

### LINKED ISSUE #1580: ci: Change CI script to run "regular" tests and "node" tests in separate steps
It should make it easier to catch performance regressions caused by enabling more Node tests, and might be a good starting point to run "Node" tests conditionally only if something in `std/node` changes.

bartlomieju:
Closes #1580


kt3k (approved):
LGTM

It'd be great if you could also update README, mentioning the new CI checks

Issue #1539, opened 2021-11-09
std/fs/mod.ts depends on unstable APIs


ry:

ah-yu:
seems unstable api `Deno.utime(sync)` is used in `copyFile(sync)/copySymLink(sync)`

kt3k:
There is a plan for deprecating `utime` in favor of `utimes` (or `utimens`) https://github.com/denoland/deno/issues/8634. That's probably the reason why utime is not stabilized yet.

I think the correct steps to resolve this issue would be:
1. implement Deno.utimes (or Deno.utimens)
2. replace the `utime` usage with `utimes` in std/fs
3. stabilize Deno.utimes

ry:
I don't think `std/fs/mod.ts` is useful. It implies many dependencies. I would be in favor removing it entirely. 

The steps listed in https://github.com/denoland/deno_std/issues/1539#issuecomment-964803551 are good for updating `std/fs/copy.ts` to not use utime.

PR #1583, opened 2021-11-15
Remove fs/mod.ts

ry:
Fixes https://github.com/denoland/deno_std/issues/1539

### DIFF
--- a/fs/README.md
+++ b/fs/README.md
@@ -115,7 +115,7 @@ moveSync("./foo", "./existingFolder", { overwrite: true });
 copy a file or directory. Overwrites it if option provided.
 
 ```ts
-import { copy, copySync } from "https://deno.land/std@$STD_VERSION/fs/mod.ts";
+import { copy, copySync } from "https://deno.land/std@$STD_VERSION/fs/copy.ts";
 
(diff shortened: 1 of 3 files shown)


bartlomieju:
This will break an enormous amount of code in the wild and doesn't solve the issue of requiring unstable at all. We should at least deprecate this module for a few release before removing completely.

ry:
We make breaking changes in std often. I would just remove this...

LionC:
> We make breaking changes in std often. I would just remove this...

@ry does that mean that you plan to cover even more abstract FS related cases (e.g. `walk`) in the runtime API in the future? Or do you propose to straight up not cover them at all?

bartlomieju:
> We make breaking changes in std often. I would just remove this...

I'm still not in favor of this change. This make it more cumbersome for users to import APIs (now you have to add 5 imports instead of one) without an obvious benefit. The root issues is still not resolved with this change.

kt3k:
How about only removing `./copy.ts` from `fs/mod.ts`, which also resolves #1539 with less breaking changes.

ry:
@kt3k Ok I've made that change.

kt3k (approved):
LGTM

PR #1584, opened 2021-11-15
node: add nextTick helper module

bartlomieju:
Towards #1482

Added a helper module so `nextTick` can be used across `std/node` instead of using
`queueMicrotask` ad hoc. Still need to provide a proper polyfill.


bartlomieju:
CC @Soremwar your streams PR will be impacted by this change

kt3k (approved):
LGTM

PR #1585, opened 2021-11-15
fix(node): Use upstream implementation for streams

Soremwar:
As it stands now, this requires `--no-check` in order to work properly. Types will be added once the polyfill reaches the level of stability the previous one had


bartlomieju:
@Soremwar I made some progress on the `process.nextTick` if you build Deno locally off of https://github.com/denoland/deno/pull/12771 and stack https://github.com/denoland/deno_std/pull/1588 on top of your branch you should be able to check if that helps with your branch.

bartlomieju:
@Soremwar FYI a new polyfill for `process.nextTick` is now landed, after Deno 1.16.2 releases later today you will be able to use it (or use canary in the meantime). Let me know if there's some work we could split up on this PR

Soremwar:
Hitting the weirdest error here. `Writable` is being acused of not being initialized even when it's used multiple types by that point :/

bartlomieju:
@Soremwar I think I'm hitting similar error: https://github.com/denoland/deno_std/pull/1608#issuecomment-971736095

bartlomieju:
> @Soremwar I think I'm hitting similar error: [#1608 (comment)](https://github.com/denoland/deno_std/pull/1608#issuecomment-971736095)

I think we do have some circular imports going on that cause it

Soremwar:
So I discovered readable-stream is extremely out of sync and doesn't represent the current state of stream, plus a few quirks that were made to mimic it's behavior in browsers that I'm not sure will cause the expected behavior when ported to Deno

I'm gonna take a new approach here, building directly from the node repository (possible now that there is a shim for stream_common_base) in a similar fashion to the original implementation but changing as little as possible to prevent dealing with inconsistencies down the road

PR #1591, opened 2021-11-16
fix(node/test-util): Improve coverage

wafuwafu13:

### DIFF
--- a/node/_errors.ts
+++ b/node/_errors.ts
@@ -42,8 +42,6 @@ const kTypes = [
   "symbol",
 ];
 
-const nodeInternalPrefix = "__node_internal_";
-
 // Node uses an AbortError that isn't exactly the same as the DOMException
 // to make usage of the error in userland and readable-stream easier.
 // It is a regular error with `.code` and `.name`.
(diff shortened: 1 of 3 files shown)


node/_errors.ts
```diff
@@ -74,7 +72,7 @@ function addNumericalSeparator(val: string) {
 export function hideStackFrames(fn: GenericFunction) {
   // We rename the functions that will be hidden to cut off the stacktrace
   // at the outermost one.
-  const hidden = nodeInternalPrefix + fn.name;
+  const hidden = "__node_internal_" + fn.name;
```
wafuwafu13, line 75: Changed because following error occured

```zsh
AssertionError: Failed assertion: error: Uncaught ReferenceError: Cannot access 'nodeInternalPrefix' before initialization
  const hidden = nodeInternalPrefix + fn.name;
                 ^
    at hideStackFrames (file:///Users/tagawahirotaka/Desktop/deno_std/node/_errors.ts:77:18)
    at file:///Users/tagawahirotaka/Desktop/deno_std/node/_validators.ts:39:30
```

kt3k: I guess the initialization of that const might be blocked by top-level-await somewhere in the dependencies? Looks like a very strange situation...

Thank you for the fix!

kt3k (approved):
@wafuwafu13 LGTM. Thanks!

node/_errors.ts
wafuwafu13: >I guess the initialization of that const might be blocked by top-level-await somewhere in the dependencies? Looks like a very strange situation...

I cloudn't figure out... I'll look into it again.

PR #1592, opened 2021-11-16
feat(node/util): Add `util.inspect`

wafuwafu13:
https://nodejs.org/api/util.html#utilinspectobject-options

Some node tests are commented out, but the basic implementation is ready.


node/assertion_error.ts
```diff
@@ -89,7 +89,7 @@ export function inspectValue(val: unknown): string {
   return inspect(
     val,
     {
-      compact: false,
+      compact: true,
```
wafuwafu13, line 92: TODO

bartlomieju: I'm not sure what goes wrong here, @wafuwafu13 could you explain?

wafuwafu13: @bartlomieju 
`compact` was not specified correctly here (should be `!!opts[0].compat`).
https://github.com/denoland/deno_std/blob/f3f9baaf96466a86b9c62223900763e85de61837/node/util.ts#L43

I reimplemented `inspect` in this PR so that `compact: false` could be configured correctly and the AssertionError occurred.

bartlomieju: Ah perfect! Thanks

bartlomieju:
@wafuwafu13 could you please enumerate what is still missing in `inspect` API? If we can't inspect some stuff we should consider adding relevant bindings to `Deno.core` to make it possible.

wafuwafu13:
@bartlomieju 
>If we can't inspect some stuff we should consider adding relevant bindings to Deno.core to make it possible.

I still haven't figured out what `Deno.core` needs in order to complete `inspect`.
It's not just `inspect` issue, but at least we'll need to implement [`vm` and `v8`](https://github.com/nodejs/node/blob/a37b9c80ef837de5fc99d750c3597d32ad3814b0/test/parallel/test-util-inspect.js#L28-L29).
I will try to [shim `internalBindings`](https://github.com/denoland/deno_std/pull/1592/files#diff-d0cd544eddc5719fb79da9271eaac12a7745f99fc5a61b5f7499ecc60ed1ac39) more details and fix some bugs in another PR, so it will be later to enumerate what is missing in `inspect` API in detail.

kt3k (approved):
@wafuwafu13 LGTM

PR #1595, opened 2021-11-16
fix(node/http): improve http.Server compatibility

bnoordhuis:
The goal is to get http.Server in good enough shape to run Node's
benchmarks. Not quite there yet but getting closer.

This commit adds a few regression tests but those are expected to
eventually be subsumed by Node's own test suite.

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -1,9 +1,11 @@
-import { _normalizeArgs, ListenOptions } from "./net.ts";
-import { Status as STATUS_CODES } from "../http/http_status.ts";
-import { Buffer } from "./buffer.ts";
-import { EventEmitter } from "./events.ts";
 import NodeReadable from "./_stream/readable.ts";
 import NodeWritable from "./_stream/writable.ts";
+import { Buffer } from "./buffer.ts";
+import { EventEmitter } from "./events.ts";
+import { Status as STATUS_CODES } from "../http/http_status.ts";
+import { _normalizeArgs, ListenOptions } from "./net.ts";
+import { nextTick } from "./_next_tick.ts";
+import { validatePort } from "./_validators.ts";
 
 const METHODS = [
   "ACL",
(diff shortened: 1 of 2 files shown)


node/http.ts
```diff
@@ -247,7 +255,6 @@ export class Server extends EventEmitter {
           const req = new IncomingMessage(reqEvent.request);
           const res = new ServerResponse(reqEvent);
           this.emit("request", req, res);
-          this.#handler(req, res);
```
bnoordhuis, line 250: I infer this is an optimization but it's not quite right, it causes observable differences between node and deno.

bartlomieju (approved):

node/http_test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  {
+    const server = http.createServer(() => {});
+    assertEquals(1, EventEmitter.listenerCount(server, "request"));
+  }
+
+  {
+    const promise = deferred<void>();
+
+    http.createServer().listen(function () {
```
kt3k, line 22: It seems that we need `this: Server` annotation here to avoid typescript error

node/http.ts
```diff
@@ -208,13 +209,15 @@ export class IncomingMessage extends NodeReadable {
 type ServerHandler = (req: IncomingMessage, res: ServerResponse) => void;
 
 export class Server extends EventEmitter {
-  #handler: ServerHandler;
   #listener?: Deno.Listener;
   #listening = false;
 
-  constructor(handler: ServerHandler) {
+  constructor(handler?: ServerHandler) {
     super();
```
kt3k, line 216: Because EventEmitter is no longer ES6 class after #1558, it seems that we also need to call `EventEmitter.call(this)` around here to make Server actually work as EventEmitter

bartlomieju: That's surprising - the same thing happens in `Process` class and it seems to work fine (no broken tests)

kt3k: Tried again, and the above doesn't happen anymore.. I might have been missing something.. 🤕

bnoordhuis: Tangential: this probably needs to change anyway because in node you can do `const server = http.Server()` (plain function call, not a construct call.)

PR #1602, opened 2021-11-16
fix(node/http): improve http.Server#close() compat

bnoordhuis:
Test fails like this:
```
running 2 tests from file:///home/bnoordhuis/src/std/node/http_test.ts
test [node/http listen] ... ok (9ms)
test [node/http chunked response ... FAILED (12ms)

failures:

[node/http chunked response
AssertionError: Test case is leaking async ops.
Before:
  - dispatched: 9
  - completed: 9
After:
  - dispatched: 18
  - completed: 17
Ops:
  op_http_accept:
    Before:
      - dispatched: 0
      - completed: 0
    After:
      - dispatched: 2
      - completed: 1

Make sure to await all promises returned from Deno APIs before
finishing test case.
    at assert (deno:runtime/js/06_util.js:41:13)
    at asyncOpSanitizer (deno:runtime/js/40_testing.js:96:7)
    at async resourceSanitizer (deno:runtime/js/40_testing.js:112:7)
    at async Object.exitSanitizer [as fn] (deno:runtime/js/40_testing.js:144:9)
    at async runTest (deno:runtime/js/40_testing.js:349:7)
    at async Object.runTests (deno:runtime/js/40_testing.js:460:22)

failures:

        [node/http chunked response

test result: FAILED. 1 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (132ms)
```
I think it's because `server.close()` doesn't terminate the loop in `#listenLoop()` but I don't know how to fix that. Suggestions welcome.

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -1,6 +1,7 @@
 import NodeReadable from "./_stream/readable.ts";
 import NodeWritable from "./_stream/writable.ts";
 import { Buffer } from "./buffer.ts";
+import { ERR_SERVER_NOT_RUNNING } from "./_errors.ts";
 import { EventEmitter } from "./events.ts";
 import { Status as STATUS_CODES } from "../http/http_status.ts";
 import { _normalizeArgs, ListenOptions } from "./net.ts";
@@ -210,8 +211,8 @@ export class IncomingMessage extends NodeReadable {
 type ServerHandler = (req: IncomingMessage, res: ServerResponse) => void;
 
 export class Server extends EventEmitter {
+  #httpConnections: Set<Deno.HttpConn> = new Set();
   #listener?: Deno.Listener;
-  #listening = false;
 
   constructor(handler?: ServerHandler) {
     super();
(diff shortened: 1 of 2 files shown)


lucacasonato:
> I think it's because server.close() doesn't terminate the loop in #listenLoop() but I don't know how to fix that.

The inner loops or the outer loop/

lucacasonato:
The loop outer loop should terminate after the next round of the event loop. What if you `this.emit("close");` after the `for await` loop?

bnoordhuis:
@lucacasonato Doesn't change anything. Neither does moving `promise.resolve()` to a `"close"` event listener afterwards (not that that would be a good solution.)

kt3k:
`op_http_accept` seems dispatched twice but only completed once. I don't have no idea how to complete the 2nd op call though..

kt3k:
It seems that we need to keep track of `HttpConn`s from `Deno.serveHttp(conn)` and close those as well when the server closing as `std/http/server` does. ref: https://github.com/denoland/deno_std/blob/b4b4314c71feb3de4275e8bd5663c7caae511225/http/server.ts#L319-L321

Maybe the easier way to implement it is to use `std/http/server` in `node/http`

bnoordhuis:
@kt3k Tracking and manually closing connections sounds reasonable but it doesn't make the op_http_accept error go away, presumably because that's the server, not the client.

bnoordhuis:
Okay, I figured it out. This PR is now a http.Server#close() compat fix, PTAL. I'll tackle the chunked response bug in another PR.

lucacasonato (approved):

PR #1605, opened 2021-11-17
chore(node/util): Convert `internal/util/inspect.js` to TypeScript

wafuwafu13:
I want to define `util.inspect` written in TS at `internal/util/inspect`. ( https://github.com/denoland/deno_std/pull/1592 is in progress.)

### DIFF
--- a/node/internal/util/inspect.ts
+++ b/node/internal/util/inspect.ts
@@ -43,9 +43,9 @@ export function getStringWidth(str: string, removeControlChars = true): number {
   str = str.normalize("NFC");
   for (const char of str[Symbol.iterator]()) {
     const code = char.codePointAt(0);
-    if (isFullWidthCodePoint(code!)) {
+    if (code && isFullWidthCodePoint(code)) {
       width += 2;
-    } else if (!isZeroWidthCodePoint(code!)) {
+    } else if (code && !isZeroWidthCodePoint(code)) {
       width++;
     }
   }
--- a/node/internal/util/inspect.ts
+++ b/node/internal/util/inspect.ts
@@ -34,7 +34,7 @@ const ansi = new RegExp(ansiPattern, "g");
 /**
  * Returns the number of columns required to display the given string.
  */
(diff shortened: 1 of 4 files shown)


node/internal/util/inspect.ts
```diff
@@ -34,7 +34,7 @@ const ansi = new RegExp(ansiPattern, "g");
 /**
  * Returns the number of columns required to display the given string.
  */
-export function getStringWidth(str, removeControlChars = true) {
+export function getStringWidth(str: string, removeControlChars = true): number {
```
bartlomieju, line 37: I'm not in favor of changing this file to TypeScript - it was intentionally taken from the upstream as JS file so it's easier to maintain and there are no differences between Node's and our implementation. Why do you want to change it to TS?

wafuwafu13: I mistakenly thought that type inconsistency would occur when defining `util.inspect` for `internal/util/inspect.js` 💦 .

>there are no differences between Node's and our implementation

If I define [`util.inspect`](https://github.com/nodejs/node/blob/86e976f1f4f9a506385dccec4d83409b5f55964b/lib/internal/util/inspect.js#L286) in `deno_std/node/internal/util/inspect.js`, differences arise Node's and our implementation because of [primordials](https://github.com/nodejs/node/blob/86e976f1f4f9a506385dccec4d83409b5f55964b/lib/internal/util/inspect.js#L80) and [internalBinding](https://github.com/nodejs/node/blob/86e976f1f4f9a506385dccec4d83409b5f55964b/lib/internal/util/inspect.js#L95). is it ok?

bartlomieju: @wafuwafu13 it is expected that primordials will be changed, but you can shim `internalBindings` by adding a module to `std/node/internal_binding/` directory. If possible let's keep it JS with only required changes for primordials.

wafuwafu13: :+1:  I close this PR.

PR #1608, opened 2021-11-17
refactor(node): reorganize _next_tick.ts module

bartlomieju:
Follow up to #1588

PR #1610, opened 2021-11-17
fix(node): Split process to prevent circular dependency on streams

Soremwar:


node/_process/streams.ts
```diff
@@ -0,0 +1,111 @@
+import { Readable, Writable } from "../stream.ts";
+import { Buffer } from "../buffer.ts";
+import { nextTick } from "./process.ts";
```
bartlomieju, line 3: Pull it from `../_next_tick.ts`

Soremwar: My bad

bartlomieju: I fixed a few more imports in https://github.com/denoland/deno_std/pull/1608, feel free to apply them here or leave it for a follow up

bartlomieju (approved):
LGTM, thanks @Soremwar good and quick fix

PR #1612, opened 2021-11-17
fix(node): use old "process.nextTick" polyfill if Deno.core is not available

bartlomieju:


bartlomieju:
Screenshot 2021-11-17 at 21 18 05

node/_next_tick.ts
```diff
@@ -1,14 +1,13 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 
+// deno-lint-ignore-file no-inner-declarations
```
dsherret, line 3: Perhaps we should still keep the declarations at the top level, but just make certain statements conditional?

dsherret (approved):
LGTM

node/_next_tick.ts
bartlomieju: I'd prefer not - I want to save a reference to `Deno.core` in `core` constant and reuse that inside functions.

PR #1614, opened 2021-11-18
feat(http/cookies): Add cookies support for Request and Response directly

tjosepo:
Modified the `getCookies`, `setCookies` and `deleteCookies` functions to support passing the `Response` and `Request` objects to them directly. This removes the need to pass the `.headers` attribute around.

I think it's a good quality of life improvement.

```js
import { getCookies, setCookie, deleteCookie } from "https://deno.land/std/http/server.ts";

serve(req => {
  const res = new Response("Hello world");

  // getCookies
  const cookies = getCookies(req);

  // setCookie
  if (!cookies.session) {
    setCookie(res, { name: "session", value: "12345" });
  }

  // deleteCookie
  if (cookies.cookieToDelete) {
    deleteCookie(res, "cookieToDelete");
  }
  
  return res;
});
```

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -32,6 +32,14 @@ export interface Cookie {
 
 const FIELD_CONTENT_REGEXP = /^(?=[\x20-\x7E]*$)[^()@<>,;:\\"\[\]?={}\s]+$/;
 
+function getHeaders(object: Request | Response | Headers): Headers {
+  if (object instanceof Request || object instanceof Response) {
+    return object.headers;
+  } else {
+    return object;
+  }
+}
+
 function toString(cookie: Cookie): string {
   if (!cookie.name) {
     return "";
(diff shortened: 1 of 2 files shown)


kt3k (commented):
I'm not sure these overloads are desirable or worth it. Does anyone have any opinions on this?

LionC:
I personally think that the Cookies API needs work, but ideally some form of Request (and maybe even Response) wrapper should have a direct (`CookieStorage` standard compatible?) interface. I have implemented that already on top of the middleware / Request wrapping PR for reading cookies and I think doing something similar for writing cookies in the future might be a good approach.

As it stands though, I agree that this does not really change the underlying problem of the current API.

timreichen:
See this PR which I argued to do exactly the opposite: https://github.com/denoland/deno_std/pull/1041

PR #1621, opened 2021-11-21
feat(http): introduce onError option on ServerInit

### LINKED ISSUE #1618: http: no error logged when calling undefined function
https://dash.deno.com/playground/whats-the-error

image

### LINKED ISSUE #1629: Errors thrown inside http.serve magically disappear
**Describe the bug**
Errors thrown inside http.serve magically disappear.

https://gist.github.com/Vehmloewff/775e44186cb2732523f1b0ecafbc9788

The error is never written to the console.

**Steps to Reproduce**

Run the example...

```sh
deno run --unstable --allow-net https://gist.githubusercontent.com/Vehmloewff/775e44186cb2732523f1b0ecafbc9788/raw/44d92c2a9e0d60923aa9461024fd715478787e12/errors-disappear.ts
```

...then head to http://localhost:3000

**Expected behavior**

"before throw" to be written to the console, then the error to be printed

**Environment**

- OS: Zorin OS 15.3
- deno version: 1.16.2
- std version: 1.113.0

satyarohith:
Closes #1618
closes #1629

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -354,13 +354,15 @@ export class Server {
 
       // Send the response.
       await requestEvent.respondWith(response);
-    } catch {
+    } catch (error) {
       // If the handler throws then it is assumed that the impact of the error
-      // is isolated to the individual request, so we close the connection.
+      // is isolated to the individual request, so we log the error and close
+      // the connection.
       //
       // Alternatively the connection has already been closed, or there is some
       // other error with responding on this connection that prompts us to
       // close it and open a new connection.
+      console.error(error);
       return this.#closeHttpConn(httpConn);
     }
   }
--- a/http/server.ts
+++ b/http/server.ts
@@ -109,7 +109,7 @@ export interface ServerInit {
    *
    * The default error handler logs and returns the error in JSON format.
    */
-  onError?: (error: Error) => Response | Promise<Response>;
+  onError?: (error: unknown) => Response | Promise<Response>;
 }
 
 /** Used to construct an HTTP server. */
@@ -119,7 +119,7 @@ export class Server {
   #closed = false;
   #listeners: Set<Deno.Listener> = new Set();
   #httpConnections: Set<Deno.HttpConn> = new Set();
-  #onError: (error: Error) => Response | Promise<Response>;
+  #onError: (error: unknown) => Response | Promise<Response>;
 
   /**
    * Constructs a new HTTP Server instance.
@@ -144,18 +144,24 @@ export class Server {
   constructor(serverInit: ServerInit) {
     this.#addr = serverInit.addr;
     this.#handler = serverInit.handler;
-    this.#onError = serverInit.onError ?? function (error: Error) {
+    this.#onError = serverInit.onError ?? function (error: unknown) {
       console.error(error);
+      const message = error instanceof Error ? error.message : error;
       return new Response(
         JSON.stringify(
           {
             error: "Internal Server Error",
-            message: error.message,
+            message,
           },
           null,
           2,
         ),
-        { status: 500 },
+        {
+          status: 500,
+          headers: {
+            "content-type": "application/json",
+          },
+        },
       );
     };
   }
@@ -374,9 +380,9 @@ export class Server {
         requestEvent.request,
         connInfo,
       );
-    } catch (error) {
+    } catch (error: unknown) {
       // Invoke onError handler when request handler throws.
-      response = await this.#onError(error as Error);
+      response = await this.#onError(error);
     }
 
     try {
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

AaronO:
Would be nice to provide an option to silence this, it will likely be undesirable noise for many production systems since real-world servers will get client-aborted connections semi-frequently

kt3k (approved):
LGTM

kt3k:
This also closes #1629

satyarohith:
> Would be nice to provide an option to silence this, it will likely be undesirable noise for many production systems since real-world servers will get client-aborted connections semi-frequently

I agree. Does `ignoreHandlerErrors` sound good? Or `suppressHandlerErrors` or `logHandlerErrors`?

[)](https://api.gh-polls.com/poll/01FN868S6THW2HKV3EF1AVHBQM/ignoreHandlerErrors%20(default%3A%20false)/vote)
[)](https://api.gh-polls.com/poll/01FN868S6THW2HKV3EF1AVHBQM/suppressHandlerErrors%20(default%3A%20false)/vote)
[)](https://api.gh-polls.com/poll/01FN868S6THW2HKV3EF1AVHBQM/logHandlerErrors%20(default%3A%20true)/vote)

I'll add the option to `ServerInit`.

kt3k:
Another option might be `onError (default: (e) => console.log(e))`

satyarohith:
> Another option might be `onError (default: (e) => console.log(e))`

I like this. It will provide more flexibility. On the name: should it be `onError` or `onHandlerError`? The latter sounds better to me as it's more explicit.

@AaronO thoughts?

lucacasonato (changes requested):
Rather than logging all errors, could we log only the errors coming from the handler itself (not ones coming from `respondWith`)?

ry:
@satyarohith `onError` is fine.

http/server.ts
```diff
@@ -136,6 +144,20 @@ export class Server {
...
+        JSON.stringify(
+          {
+            error: "Internal Server Error",
+            message: error.message,
+          },
+          null,
+          2,
+        ),
+        { status: 500 },
+      );
```
ry, line 159: Nice!

Probably should set the content type header to "application/json"

http/server.ts
```diff
@@ -345,20 +367,23 @@ export class Server {
...
     try {
       // Handle the request event, generating a response.
-      const response = await this.#handler(
+      response = await this.#handler(
         requestEvent.request,
         connInfo,
       );
+    } catch (error) {
+      // Invoke onError handler when request handler throws.
+      response = await this.#onError(error as Error);
```
lucacasonato, line 379: What should we do if this throws? :upside_down_face:

http/server.ts
```diff
@@ -345,20 +367,23 @@ export class Server {
...
   ): Promise<void> {
+    let response: Response;
     try {
       // Handle the request event, generating a response.
-      const response = await this.#handler(
+      response = await this.#handler(
         requestEvent.request,
         connInfo,
       );
+    } catch (error) {
```
lucacasonato, line 377:
```suggestion
    } catch (error: unknown) {
```

http/server.ts
lucacasonato, line 379:
```suggestion
      response = await this.#onError(error);
```

http/server.ts
```diff
@@ -536,6 +561,9 @@ export interface ServeInit {
 
   /** An AbortSignal to close the server and all connections. */
   signal?: AbortSignal;
+
+  /** The handler to invoke when route handlers throw an error. */
+  onError?: (error: Error) => Response | Promise<Response>;
```
lucacasonato, line 566:
```suggestion
  onError?: (error: unknown) => Response | Promise<Response>;
```

http/server.ts
ry: then the server should crash

lucacasonato (commented):
Can we print just `Internal Server Error`, and not bother with this JSON stuff? Sending full error messages to the client could leak sensitive info and is probably not what the developer wants for production.

bartlomieju (approved):
LGTM

sant123:
This works for `serve` function only. Can this be implemented for both `serveTls` and `serveListener`? Just curious why it was not implemented for them.

satyarohith:
> Can this be implemented for both `serveTls` and `serveListener`?

Yes. Would you like to make a PR? :)

> Just curious why it was not implemented for them.

I forgot to update them. :)

sant123:
> 

Sure! Let me work on it 😉

satyarohith:
@sant123, the issue has been resolved. Thanks!

PR #1623, opened 2021-11-22
fix(fmt): print with rounding taken into account

vi117:
Currently `printf` behaves as seen in the following block.
```ts
import {printf} from "https://deno.land/std/fmt/printf.ts";
printf("%.3f",0.9999);
//it print 0.000
printf("%.1f\n",0.99);
//it print 0.0
printf("%.3f\n",1.9999);
//it print 1.000
```
The reason for this behavior is that `roundFractionToPrecision` ignores the carry.
so, I fixed it.

### DIFF
--- a/fmt/printf.ts
+++ b/fmt/printf.ts
@@ -502,20 +502,32 @@ class Printf {
    * Round fraction to precision
    * @param fractional
    * @param precision
+   * @returns tuple of fractional and round
+   * ```ts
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Good catch

bartlomieju (approved):
LGTM too

kt3k:
@vi117 Thank you for your contribution!

Issue #1627, opened 2021-11-23
node: Error when requiring package doesn't return 1 exit code


bartlomieju:
```
❯ deno run --no-check --compat --unstable --allow-read --allow-write=./ ./bin/eslint.js lib/ tests/
Not implemented: process.on("uncaughtException")
Not implemented: process.on("unhandledRejection")

Oops! Something went wrong! :(

ESLint: 8.2.0

PermissionDenied: Requires env access to "IGNORE_TEST_WIN32", run again with the --allow-env flag
    at Object.opSync (deno:core/01_core.js:142:12)
    at Object.getEnv [as get] (deno:runtime/js/30_os.js:56:17)
    at Object.get (https://deno.land/std@0.115.0/node/process.ts:63:21)
    at Object.<anonymous> (file:///Users/biwanczuk/dev/eslint/node_modules/ignore/index.js:445:32)
    at Module._compile (https://deno.land/std@0.115.0/node/module.ts:251:36)
    at Object.Module._extensions..js (https://deno.land/std@0.115.0/node/module.ts:1276:10)
    at Module.load (https://deno.land/std@0.115.0/node/module.ts:230:33)
    at Function._load (https://deno.land/std@0.115.0/node/module.ts:539:14)
    at Module.require (https://deno.land/std@0.115.0/node/module.ts:216:21)
    at require (https://deno.land/std@0.115.0/node/module.ts:1331:16)

❯ echo $status
0

```

This is a blocker for https://github.com/eslint/eslint/pull/15336

bartlomieju:
Probably related to this bit:
https://github.com/denoland/deno_std/blob/93ee52d5a8e2ee1d88153b51a4fda2efa452c9da/node/module.ts#L536-L554

If `module.load()` throws the error is never surfaced.

bartlomieju:
This is actually caused by the fact that `process.exitCode` is not respected when the event loop shuts down "cleanly". Ie. if `process.exitCode` is set anywhere in the code it is not forwarded to Rust in any way.

iuioiua:
@bartlomieju, is this still an issue?

bartlomieju:
To be honest I'm not sure - I haven't tried this example in a while. Any chance you could try to replicate it with `npm:` specifiers?

iuioiua:
```
asher@Ashers-MacBook-Pro-3 deno_std % deno run -A npm:eslint node  

Oops! Something went wrong! :(

ESLint: 8.28.0

[...]

PR #1628, opened 2021-11-23
fix(node): respect process.exitCode when exiting

bartlomieju:
Closes https://github.com/denoland/deno_std/issues/1627

### DIFF
--- a/node/process.ts
+++ b/node/process.ts
@@ -216,11 +216,10 @@ class Process extends EventEmitter {
   constructor() {
     super();
 
-    //This causes the exit event to be binded to the unload event
+    // This causes the exit event to be binded to the unload event
     globalThis.addEventListener("unload", () => {
-      //TODO(Soremwar)
-      //Get the exit code from the unload event
-      super.emit("exit", 0);
+      super.emit("exit", process.exitCode);
+      Deno.exit(process.exitCode || 0);
     });
   }
 
--- a/node/process_test.ts
+++ b/node/process_test.ts
@@ -432,6 +432,38 @@ Deno.test("process.exitCode", () => {
   assert(process.exitCode === 127);
 });
 
+Deno.test("process.exitCode when exiting", async () => {
+  const cwd = path.dirname(path.fromFileUrl(import.meta.url));
+
+  const p = Deno.run({
+    cmd: [
+      Deno.execPath(),
+      "run",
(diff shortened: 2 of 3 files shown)


node/process.ts
```diff
@@ -216,11 +216,10 @@ class Process extends EventEmitter {
...
     super();
 
-    //This causes the exit event to be binded to the unload event
+    // This causes the exit event to be binded to the unload event
     globalThis.addEventListener("unload", () => {
-      //TODO(Soremwar)
-      //Get the exit code from the unload event
-      super.emit("exit", 0);
+      super.emit("exit", process.exitCode);
+      Deno.exit(process.exitCode || 0);
```
bartlomieju, line 222: This is probably wrong - if there are more "unload" listeners attached they will never be run. Opened the PR to get discussion going.

bartlomieju: @bnoordhuis any ideas how we tackle that except adding functionality equivalent to `process.exitCode` directly into CLI?

kt3k (approved):
LGTM

node/process.ts
lucacasonato: > This is probably wrong - if there are more "unload" listeners attached they will never be run. Opened the PR to get discussion going.

I have pushed a fix for this to the PR.

lucacasonato (changes requested):
This PR is blocked on https://github.com/denoland/deno/issues/12888. This PR falsely makes it look like tests that fail actually succeed, because all `deno test` isolates exit with code 0 if they import `std/node/process.ts` anywhere now.

bartlomieju:
@bnoordhuis actually I was mistaken during the call, this PR is not yet ready.

bartlomieju:
This should now work properly, however we should probably wait until https://github.com/denoland/deno/pull/12938 is merged before merging this one.

iuioiua:
@bartlomieju, denoland/deno#12938 was cancelled. Is this PR still valid?

iuioiua:
Gently pinging @kt3k.

iuioiua:
@bartlomieju, this appears to no longer be an issue according to https://github.com/denoland/deno_std/issues/1627#issuecomment-1325900082

bartlomieju:
Thanks for checking @iuioiua, closing without a merge.

PR #1632, opened 2021-11-24
test(node/fs): enable `test-fs-rm.js`

F3n67u:
this is a part of #1564


node/internal/url.js
```diff
@@ -0,0 +1,111 @@
...
+
+export function toPathIfFileURL(fileURLOrPath) {
+  if (!isURLInstance(fileURLOrPath)) {
+    return fileURLOrPath;
+  }
+  return fileURLToPath(fileURLOrPath);
+}
+
+export default {
+  toPathIfFileURL,
```
wafuwafu13, line 110: Maybe you can use here.
https://github.com/denoland/deno_std/blob/main/node/url.ts

I've implemented here with reference to https://github.com/nodejs/node/blob/master/lib/internal/url.js

F3n67u: thanks for the review. I removed the `url.js` file and use `deno_std/node/url.ts` instead.

node/url.ts
```diff
@@ -325,6 +332,7 @@ export default {
   format,
   fileURLToPath,
   pathToFileURL,
+  toPathIfFileURL,
```
kt3k, line 335: Is this used somewhere?

node/_tools/config.json
```diff
@@ -129,6 +129,7 @@
       "test-event-emitter-symbols.js",
       "test-events-list.js",
       "test-events-once.js",
+      "test-fs-rm.js",
```
kt3k, line 132: nice 👍

kt3k (commented):
Mostly looks good. Left one question

node/url.ts
F3n67u: it's imported in `node/internal/fs/utils.js`

kt3k (approved):
@F3n67u LGTM. Thank you for your contribution!

PR #1636, opened 2021-11-27
fix(std/node): Add base64url encoding support, indexOf, lastIndexOf and includes to Buffer

Soremwar:


node/_tools/config.json
```diff
@@ -68,6 +68,7 @@
       "test-buffer-failed-alloc-typed-arrays.js",
       "test-buffer-fakes.js",
       "test-buffer-from.js",
+      "test-buffer-indexof.js",
```
kt3k, line 71: nice 👍

node/_events.js
```diff
@@ -37,7 +37,7 @@ import {
   validateAbortSignal,
   validateBoolean,
   validateFunction,
-} from "./_validators.ts";
+} from "./internal/validators.js";
```
Soremwar, line 40: All this renames might seem unnecessary, but this actually solves two problems:

1) _validators is actually a shim of internal/validators however it was left in that location in the early days of std/node, now it will be easier to find the required functions when porting modules
2) This solves a circular dependency created between _validators and internal/validators by moving the content of _validators

node/_tools/suites/common/dns.js
```diff
@@ -5,10 +5,10 @@
 // Taken from Node 16.13.0
 // This file is automatically generated by "node/_tools/setup.ts". Do not modify this file manually
 
-'use strict';
+"use strict";
```
Soremwar, line 8: I have no idea why this formatting wasn't being applied, all I did was run setup.ts and all this changes were applied

Soremwar:
Ready for review. Since this apparently also formatted some files unintentionally I gotta point out the relevant files:

- node/_buffer.js
- node/_tools/config.json
- node/_tools/suites/parallel/test-buffer-includes.js
- node/_tools/suites/parallel/test-buffer-indexof.js
- node/internal/buffer.js
- node/_validators.ts
- node/internal/validators.js
- node/internal_binding/_node.ts
- node/internal_binding/_utils.ts
- node/internal_binding/buffer.ts
- node/internal_binding/string_decoder.ts

Soremwar:
I have no idea what's going on :/

It doesn't seem to be `deno fmt` though, cause when I run it directly over the file after restoring it it doesn't affect it at all, but if I run setup.ts the changes resurface. Which makes no sense because the original format is different than the one we end up with

Soremwar:
Ok, I found out what was going on. Apparently I ran deno fmt and it formatted the cached decompressed versions of the file as well. Should be fine now

node/_events.js
bartlomieju: Thanks, it was bothering me too; good change.

bartlomieju (approved):
LGTM, thank you @Soremwar

PR #1641, opened 2021-11-28
http: serve() should log where it is listening

ry:
reduces boilerplate in examples https://deno.com/deploy/docs/examples


http/server.ts
```diff
@@ -604,13 +603,15 @@ export async function serve(
...
 ): Promise<void> {
-  const addr = options.addr ?? ":8000";
+  const addr = options.addr ?? "127.0.0.1:8000";
   const server = new Server({ addr, handler });
 
   if (options?.signal) {
     options.signal.onabort = () => server.close();
   }
 
+  console.log(`Listening on http://${addr}/`);
```
lucacasonato, line 613: This should probably use the resolved address, rather than the original address.

lucacasonato (commented):
There should be an option to disable this log message.

ry:
people can use listenAndServe if they don't want the log message. This is a convenience function.

kt3k:
@ry 
> people can use listenAndServe if they don't want the log message. This is a convenience function.

I like the idea. So in that case we cancel the deprecation of `listenAndServe` and recommend it for more customized use cases?

vwkd:
Is there a precedence for including logging in a std module?

 This change implicitly assumes that the console is the right place to log. Std has a whole logging framework std/log because that’s often not the case. Also what if you want to customize the log message, etc.? Baking the logging into the module makes this impossible.

It seems to me like a short sighted optimization for a very specific use case of making example applications one LOC less.

What if I don’t need logging? Using another function just to get rid of the logging? Sorry, this is ridiculous.

kt3k:
Having thought about this again, I now feel it's not worth providing 2 different APIs with such subtle difference.

If we have more features to add to the 'convenient' API, then that might worth it. However if the logging of that line is the only difference, I don't think it's worth it..

kt3k:
Reconsidered this again. `console.log("Listening ...")` is actually a boilerplate for almost 100% cases of the usages. I think this change is worth trying.

kt3k (approved):

mashaal:
> people can use listenAndServe if they don't want the log message. This is a convenience function.

@ry @kt3k - Is `listenAndServe` still the recommended way to get around this log? We have tested it's usage, and it seems to work as intended, but it's still marked as deprecated?

https://deno.land/std@0.137.0/http/server.ts#L630

kt3k:
@mashaal 
> @ry @kt3k - Is listenAndServe still the recommended way to get around this log?

Honestly I'm not sure. Canceling the deprecation of `listenAndServe` is one option. Adding some option for disabling logging might be another option. Do you have any opinion on this?

PR #1653, opened 2021-11-29
fix(node): use async read for stdin

bartlomieju:
This change makes [`enquirer`](https://github.com/enquirer/enquirer) work in compat mode.

### DIFF
--- a/node/_process/streams.ts
+++ b/node/_process/streams.ts
@@ -38,6 +38,8 @@ export const stdin = new Readable({
     const p = Buffer.alloc(size || 16 * 1024);
     Deno.stdin.read(p).then((length) => {
       this.push(length === null ? null : p.slice(0, length));
+    }, (error) => {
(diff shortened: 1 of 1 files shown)


node/_process/streams.ts
```diff
@@ -36,8 +36,9 @@ export const stdin = new Readable({
   emitClose: false,
   read(this: Readable, size: number) {
     const p = Buffer.alloc(size || 16 * 1024);
-    const length = Deno.stdin.readSync(p);
-    this.push(length === null ? null : p.slice(0, length));
+    Deno.stdin.read(p).then((length) => {
+      this.push(length === null ? null : p.slice(0, length));
+    });
```
bartlomieju, line 41: I guess there should be catch handler here, otherwise error in `Deno.stdin.read()` will result in unhandled promise rejection

bnoordhuis (approved):

Issue #1649, opened 2021-11-29
Port TS "Private" Variables to JavaScript Private Variables


sno2:
**Is your feature request related to a problem? Please describe.**

TypeScript's `private` class property modifier is only checked when you are using TypeScript and the encapsulation is lost through compilation.  Furthermore, we're assuming that someone didn't cast something as `any` along the line in which "private" variables are not private at all.  

**Describe the solution you'd like**

I think that it would be best if we migrate all instances of `private` class properties to using the ES6 Private Variable syntax (`#___`) to help mitigate the chances of users having unexpected bugs because they were using a property in their JavaScript (or `any`-typed TypeScript).  Also, this would help stop even more important mass code breakages from libraries using these private variables irresponsibly and causing issues for their unsuspecting users.

**Describe alternatives you've considered**

n/a

I am willing to work on this.

kitsonk:
Just to point out, private fields not "ES6".  They are technically not even part of the standard yet.  It is expected they will be part of ES2022.

Wholesale change/refactoring would be breaking changes where they exist, because even though they are "private" they are still accessible at runtime and people may have decided to access them.  They should be taken carefully on a case by case basis, not something that should be done wholesale across the code base.

sno2:
> Just to point out, private fields not "ES6".  They are technically not even part of the standard yet.  It is expected they will be part of ES2022.

Oh, thanks for the correction.

> Wholesale change/refactoring would be breaking changes where they exist, because even though they are "private" they are still accessible at runtime and people may have decided to access them.  They should be taken carefully on a case by case basis, not something that should be done wholesale across the code base.

Ok, do you know if there are any places where it would be ok to port to private variables to start off with? Or is the possibility of breaking code too large?

I guess also my issue is that we modify the types / names of these "private" variables  in updates so what is the difference between removing them as the code could have broken if we changed the property's name or type.

bartlomieju:
@sno2 feel free to start with code in `node/` and `io/buffer.ts`

jsejcksn:
> Wholesale change/refactoring would be breaking changes where they exist, because even though they are "private" they are still accessible at runtime and people may have decided to access them.

It seemed to me like this is exactly the misusage case that this issue intended to surface through breakage by preventing access.

@sno2 Is that correct? If not, why perform the refactor?

lino-levan:
I feel like this issue should be resolved as the `private` identifier no longer exists in std as far as I can tell.

PR #1655, opened 2021-11-30
chore(node/buffer): convert to js private variables

sno2:
See #1649 [please wait for CI]

### DIFF
--- a/io/buffer.ts
+++ b/io/buffer.ts
@@ -276,11 +276,11 @@ export interface ReadLineResult {
 
 /** BufReader implements buffering for a Reader object. */
 export class BufReader implements Reader {
-  private buf!: Uint8Array;
-  private rd!: Reader; // Reader provided by caller.
-  private r = 0; // buf read position.
-  private w = 0; // buf write position.
-  private eof = false;
+  #buf!: Uint8Array;
+  #rd!: Reader; // Reader provided by caller.
+  #r = 0; // buf read position.
+  #w = 0; // buf write position.
+  #eof = false;
   // private lastByte: number;
   // private lastCharSize: number;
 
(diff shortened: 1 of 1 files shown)


sno2:
I'm sorry @bartlomieju , but I can't seem to find out why the tests are hanging.  All I did was change the `BufferReader` abstract class to not unsafe anymore and rename all "private" variable / method instances to `#__`.

kt3k:
@sno2 
This seems crashing even without rust panic (maybe it crashes inside of V8?). The crash seems happening around `#reset` method of `BufReader`, but I couldn't figure out what exactly causes this.

If I renamed `#reset` (of BufReader) to `_reset`, then the issue disappeared on my local machine.

sno2:
> @sno2 This seems crashing even without rust panic (maybe it crashes inside of V8?). The crash seems happening around `#reset` method of `BufReader`, but I couldn't figure out what exactly causes this.
> 
> If I renamed `#reset` (of BufReader) to `_reset`, then the issue disappeared on my local machine.

Yeah, ummm when I changed `#reset` to an arrow function it passed.  I am so confused because I have no idea why it would behave differently.  We'll see what CI says, though.

io/buffer.ts
```diff
@@ -293,40 +293,40 @@ export class BufReader implements Reader {
...
   }
 
   buffered(): number {
-    return this.w - this.r;
+    return this.#w - this.#r;
   }
 
   // Reads a new chunk into the buffer.
-  private async _fill() {
+  async #fill() {
```
kt3k, line 309: Regarding the case of `#reset`, I suspect this might also cause the same issue..

kt3k:
Otherwise looks good to me

kt3k (approved):
LGTM. Nice work!

PR #1656, opened 2021-11-30
refactor(path): abstract functions by utility rather than os version

### LINKED ISSUE #687: path: Make os an option argument
```ts
export const posix = {
  basename,
  dirname,
  extname,
  ...
};

export const win32 = {
  basename,
  dirname,
  extname,
  ...
};

export {
  basename,
  dirname, 
  extname,
  ...,
  globToRegExp,
  normalizeGlob,
  SEP,
  SEP_PATTERN,
  ...
};
```

```ts
  ...,
  os?: typeof Deno.build.os;
}
```

lowlighter:
Closes #687

Basically this order path function by functionnality rather than by operating system.

When it's possible (and not too tedious) some code portions are factorized to reduce code size.

- [ ] basename
- [ ] delimiter
- [ ] dirname
- [x] extname
- [ ] format
- [x] fromFileUrl
- [x] isAbsolute
- [ ] join
- [ ] normalize
- [ ] parse
- [ ] relative
- [ ] resolve
- [ ] sep
- [x] toFileUrl
- [ ] toNamespacedPath


lowlighter:
Just didn't have time to work on this, idk if it's still revelant

PR #1659, opened 2021-12-01
refactor(node/url): Move `toPathIfFileURL` to internal

wafuwafu13:
[toPathIfFileURL](https://github.com/nodejs/node/blob/d81f328aa5b02397f7eac99389f28b951366c438/lib/internal/url.js#L1548) is not public API.


kt3k (approved):
@wafuwafu13 Good catch! Thanks!

PR #1662, opened 2021-12-01
feat(node): add process.binding()

### LINKED ISSUE #1630: node: process.binding(...)
Some npm modules depend on `process.binding(<module>)` ex. https://github.com/anseki/readline-sync/blob/202fe647300645c2727b4c4c9314d44703a67233/lib/readline-sync.js#L19

We need to shim these to run those npm modules.

kt3k:
This PR add `process.binding` API, which returns the internal binding module of the given name.

closes #1630


bartlomieju (approved):
LGTM

bartlomieju:
I can confirm that this PR moves us forward with efforts to run `npm`. Currently blocked by `https` module.

csgorham:
I made a bundle.js with Browserify  --- error is that process.binding is not supported.    Going through all of my dependency files to find process.binding.... should i replace with something else?   

For instance there is :  process.binding('tty_wrap') and process.binding('utils')  

I read somewhere these calls are only internal to node and depreciated, and browserify wont work.   

Any comment or direction? I am so new to javascript, this is my first project.  THANKS

Issue #1666, opened 2021-12-02
node: add url.parse


kt3k:

kt3k:
fixed by https://github.com/denoland/deno_std/pull/1667

PR #1667, opened 2021-12-02
feat(node): Add missing url.parse

getspooky:
Ref https://github.com/denoland/deno_std/issues/1666


node/url.ts
```diff
@@ -116,6 +175,296 @@ export function format(
...
+    } else if (end === -1 && lastPos < url.length) {
+      // We converted some backslashes and have only part of the entire string
+      rest += url.slice(lastPos);
+    } else if (end !== -1 && lastPos < end) {
+      // We converted some backslashes and have only part of the entire string
+      rest += url.slice(lastPos, end);
+    }
+  }
+
+  const self = new URL(url);
```
kt3k, line 291: I think `url.parse` should return url.Url object, not a global URL object

You can check it in Node.js repl like the below
```
> url.parse("https://example.com") instanceof url.Url
true
> url.parse("https://example.com") instanceof URL
false
```

node/url.ts
```diff
@@ -116,6 +175,296 @@ export function format(
...
+  } else if (firstIdx > 0) {
+    self.pathname = rest.slice(0, firstIdx);
+  }
+  if (slashedProtocol.has(lowerProto) && self.hostname && !self.pathname) {
+    self.pathname = "/";
+  }
+
+  // Finally, reconstruct the href based on what has been validated.
+  self.href = format(self);
+}
```
kt3k, line 466: You forgot returning `self`?

kt3k:
@getspooky 
Can you also enable some of native node.js test cases? The instruction for doing it is available in `node/README.md`

getspooky:
@kt3k i think we should create `WHATWG API` 
```js
// Original url.parse() API

function Url() {
  this.protocol = null;
  this.slashes = null;
  this.auth = null;
  this.host = null;
  this.port = null;
  this.hostname = null;
  this.hash = null;
  this.search = null;
  this.query = null;
  this.pathname = null;
  this.path = null;
  this.href = null;
}
```
in order to use `url.parse`

getspooky:
@kt3k fixed , Now parse return url.Url object, not a global URL object.

**Deno:**
```ts
test [url] URL ...Url {
  protocol: "http:",
  slashes: true,
  auth: null,
  host: "github.com",
  port: null,
  hostname: "github.com",
  hash: null,
  search: null,
  query: null,
  pathname: "/",
  path: "/",
  href: "http://github.com/"
}
```

**Node:**

```js
> url.parse("http://github.com")
Url {
  protocol: 'http:',
  slashes: true,
  auth: null,
  host: 'github.com',
  port: null,
  hostname: 'github.com',
  hash: null,
  search: null,
  query: null,
  pathname: '/',
  path: '/',
  href: 'http://github.com/'
}
```

kt3k:
@getspooky 
> fixed , Now parse return url.Url object, not a global URL object.

Thanks for updating! The change looks good to me, but the test file `test-url-parse-format.js` is not added to the source tree. Please remove `test-url-parse-format.js` from `ignore` part of `node/_tools/config.json` first, then run `node/_tools/setup.ts`, which should copy the actual test file under `node/_tools/suites/`

getspooky:
@kt3k done

wafuwafu13:
@getspooky 

We haven't implemented [hasIntl](https://github.com/nodejs/node/blob/781408f3772463e6d65386248030458901758f10/test/common/index.js#L38) yet, so it would be nice to comment it out like [this](https://github.com/denoland/deno_std/blob/d4a2ae275fdc78445a625d2bae06b3708caacdea/node/_tools/suites/parallel/test-url-format-whatwg.js#L12-L14).

node/_tools/suites/parallel/test-url-parse-format.js
```diff
@@ -8,8 +8,9 @@
 'use strict';
 const common = require('../common');
 
-if (!common.hasIntl)
-  common.skip('missing Intl');
+// TODO(wafuwafu): Implement `hasIntl`, `skip`
```
kt3k, line 11: Please use your account name in TODO comment
```suggestion
// TODO(getspooky): Implement `hasIntl`, `skip`
```

kt3k:
`test-url-parse-format.js` is not passing in mac and ubuntu (See the CI logs

getspooky:
@kt3k fixed

kt3k (approved):
@getspooky LGTM. Thank you for your contribution!

`url.format` tests are still remaining, but this adds very accurate compatibility of `url.parse`. Nice work!

PR #1669, opened 2021-12-03
chore(node): Update Node version to 16.13.1

wafuwafu13:
No logical change


bartlomieju:
That's a huge PR. I'm wondering if we should only update for minor versions. From quick skim it seems that nothing changed besides changing comments in all test files. @kt3k @bnoordhuis WDYT?

kt3k:
I don't have a strong opinion about this, but maybe let's skip this version as there's no significant change in it

wafuwafu13:
I agree :+1:

PR #1674, opened 2021-12-05
chore(node): show the command to repeat single test

kt3k:
With this change the test results show the command to repeat the failed single test file of node.js compat testing.

スクリーンショット 2021-12-05 15 20 45

This command is more efficient than `--only` feature of `node/_tools/test.ts` because of less overhead in the execution. So I now deleted `--only` feature of test runner to promote this method of running single test case.

cc @wafuwafu13


bartlomieju (approved):
LGTM

PR #1678, opened 2021-12-06
ref(node): Export encodeStr , hexTable and isHexTable

getspooky:
This is needed in https://github.com/denoland/deno_std/pull/1667


kt3k (approved):
@getspooky LGTM. Nice!

This reflects the Node.js directory structure better. Thanks!

PR #1676, opened 2021-12-06
fix(datetime): fix bug for parse at the end of the large month

NotFounds:
Resolve: https://github.com/denoland/deno_std/issues/1491

| before | after |
| ------ | ----- |
| image | image |

--- 

I have no idea how to mock `new Date()`.
This problem may reproduces only at the end of the large month.

PR #1680, opened 2021-12-07
chore: add faketime testing

kt3k:
This PR adds ability to run test with the specific date time. This is necessary for writing test for #1676

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,8 +28,19 @@ jobs:
         with:
           deno-version: ${{ matrix.deno }}
 
+      - name: Install fakedate
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          sudo apt install faketime
+
       - name: Run tests canary
-        run: deno test --doc --unstable --allow-all --coverage=./cov --import-map=test_import_map.json --config strict-ts44.tsconfig.json --ignore=node/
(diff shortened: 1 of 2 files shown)


bartlomieju:
Adding this means you won't be able to run tests locally anymore. Any idea how we could handle that gracefully?

kt3k:
@bartlomieju 
The faketime dependent file `datetime/test_2021_12_31.ts` has a filename which isn't auto discovered by `deno test`. So you can still run `deno test` without using `faketime` (that is what `Run tests canary` step does)

`Run test with the date 2021-12-31` step runs only `datetime/test_2021_12_31.ts` using `faketime`. You need to install `faketime` to run this test, but I think this only affects `datetime` module contributors

bartlomieju:
> @bartlomieju The faketime dependent file `datetime/test_2021_12_31.ts` has a filename which isn't auto discovered by `deno test`. So you can still run `deno test` without using `faketime` (that is what `Run tests canary` step does)
> 
> `Run test with the date 2021-12-31` step runs only `datetime/test_2021_12_31.ts` using `faketime`. You need to install `faketime` to run this test, but I think this only affects `datetime` module contributors

Okay, sounds good to me.

bartlomieju (approved):
LGTM

PR #1681, opened 2021-12-07
test(http): fix error message of unknown issuer

kt3k:
This PR fixes the error message expectations in https server tests. This should fix the CI failure in main

(The error is first observed in main 1 hour ago. It's probably caused by https://github.com/denoland/deno/pull/12488)

### DIFF
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -481,7 +481,7 @@ Deno.test(`Server.listenAndServeTls should handle requests`, async () => {
     await assertThrowsAsync(
       () => badConn.read(new Uint8Array(1)),
       Deno.errors.InvalidData,
-      "invalid certificate: UnknownIssuer",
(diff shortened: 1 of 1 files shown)


satyarohith (approved):

http/server_test.ts
```diff
@@ -553,7 +553,7 @@ Deno.test({
       await assertThrowsAsync(
         () => badConn.read(new Uint8Array(1)),
         Deno.errors.InvalidData,
-        "invalid certificate: UnknownIssuer",
+        "invalid peer certificate contents: invalid peer certificate: UnknownIssuer",
```
satyarohith, line 556: We should fix the error message to be less verbose in Deno. The old one seems to be on point.

bartlomieju: It will be hard as this error is surfaced from `rustls` and we'd have to manually rewrite it. I think it's not worth it.

Issue #1067, opened 2021-07-27
TOML formatting option around key alignment


kaleidawave:
**Is your feature request related to a problem? Please describe.** 

I am modifying TOMLs in Deno. Currently when decoding/parsing the following:

```toml
name = "myproject"
version = "0.1.0"
```

Encoding/to-stringing it space aligns the keys, resulting in:

```toml
name    = "myproject"
version = "0.1.0"
```

**Describe the solution you'd like**

I would like either by default or by optional setting to not add the alignment. Cargo's *standard* does not align them, I would like to retain the non key aligned formatting :)

ry:
cc @FujiHaruka (who has recently done a large refactor to std/encoding/toml)

FujiHaruka:
Here I am. I didn't change around `stringify` code, but I'm interested in your issue.

> Cargo's standard does not align them
Then, it may be better not to align. I'll also take a look at other toml parsers.

FujiHaruka:
@kaleidawave 

> Cargo's *standard* does not align them

By the way, do you mean Cargo's documentation recommends not to align? If so, can you show me the link?

kaleidawave:
The documentation for [Cargo's manifest format](https://doc.rust-lang.org/cargo/reference/manifest.html) does not mention anywhere about formatting. Formatting doesn't matter as its TOML parser handles the whitespace. What I meant is that a new cargo project's TOML configuration file does not start with its keys aligned.

For stylistic purposes a option to disable the padding alignment would be nice so the output matches the style in Cargo's docs, the crates ecosystem and https://toml.io/en/

FujiHaruka:
@kaleidawave Thank you for your explaining. It makes sense to me.
Then in my opinion it would be better if making keys not aligned is default and aligned is an optional.

FujiHaruka:
Oh I found a bug around key alignment format.

This code

```ts
import { stringify } from "https://deno.land/std@0.105.0/encoding/toml.ts";

console.log(stringify({
  "a": "a = 1",
  "helloooooooo": 1
}))

```

is expected to output like:

```toml
a            = "a = 1"
helloooooooo = 1
```

but actually outputs like this:

```toml
a = "a       = 1"
helloooooooo = 1
```

I think a possible fix of this will be simply disabling the format of key alignment.

PR #1682, opened 2021-12-07
fix(toml): parse declaration correctly

ah-yu:
This PR contains two parts.
1. origin regexp couldn't match the key correctly, see https://github.com/denoland/deno_std/issues/1067#issuecomment-907740319
2. fixes wrong test https://github.com/denoland/deno_std/blob/bfe8b128164d295bad9eb9f73250ced55b07cf4b/encoding/toml_test.ts#L513-L542


kt3k (approved):
@ah-yu LGTM. Thanks!

PR #1683, opened 2021-12-07
fix(datetime/tests): use expected date in assertion

meszarosdezso:
I think this should be the correct assertion given the test's title and the [parameter in the ci](https://github.com/denoland/deno_std/blob/main/.github/workflows/ci.yml#L43).


kt3k:
In Date constructor, `monthIndex` starts from `0`, and `11` means December here. See https://stackoverflow.com/questions/2552483/why-does-the-month-argument-range-from-0-to-11-in-javascripts-date-constructor for more discussion

meszarosdezso:
>  11 means December

Yes, I figured it out after seeing the test fail, JS's built in Date will always confuse me(: thanks for the clarification! Closing.

PR #1688, opened 2021-12-08
feat(node): add mock inspector module

kt3k:
This PR adds placeholder `inspector` module in std/node.

Some tools use `inspector` to check if there's active inspector automatically even when the user don't intend to use inspector (for example, ava does it https://github.com/avajs/ava/blob/c214512ddfd78530adf5f632974345396cc39e24/lib/cli.js#L121 ) This PR is useful for avoiding crashing the node compat mode in thoses cases.


bartlomieju (approved):
This is reasonable. LGTM

PR #1691, opened 2021-12-08
feat(node): Support resolveObject

getspooky:
Ref https://github.com/denoland/deno_std/pull/1667

### DIFF
--- a/node/url.ts
+++ b/node/url.ts
@@ -1050,6 +1050,11 @@ export function parse(
   return urlObject;
 }
 
+export function resolveObject(source: string | Url, relative: string) {
+  if (!source) return relative;
+  return parse(source, false, true).resolveObject(relative);
+}
+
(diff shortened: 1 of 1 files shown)


kt3k:
hmm this doesn't seem documented, but seems actually exported.. (also there seem some usages. ref: https://github.com/search?q=url.resolveObject&type=code )

I'm in favor of adding this because it contributes to the compatibility. Does anyone have any opinion on this?

kt3k (approved):

bartlomieju:
LGTM

kt3k:
I also found the usage of this API in the test case. https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/test/parallel/test-url-relative.js#L418

This API seems actually necessary

PR #1692, opened 2021-12-08
feat(node): Add resolve function

getspooky:
Ref https://github.com/denoland/deno_std/pull/1667


kt3k:
@getspooky Thanks for your contribution! Can you also enable `parallel/test-url-relative.js` test case? The test file exercises `url.resolve` and `url.resolveObject` APIs.

kt3k:
landed in #1851 

Thanks @getspooky for the initial work

PR #1695, opened 2021-12-09
feat(node): add child_process.fork

### LINKED ISSUE #1690: node: add child_process.fork
child_process.fork is used by ava. ref: https://github.com/avajs/ava/blob/77623a5f32c139fa4ef7d316a4bbf7e172cc5acb/lib/fork.js#L51

kt3k:
This PR implements child_process.fork in `std/node`

This implements at least the basic feature of `fork` and enabled the basic test cases (`child-process-spawn-node.js`).

This doesn't implement `ipc` features. So I commented out ipe-related assertion in the test cases.

The diff might look big, but that is because this PR moves `ChildProcess` class from `node/child_process` to `node/internal/chil_process` to follow the internal structure of Node.js code base.

closes #1690


kt3k:
The CI is blocked by #1694

bartlomieju (commented):
@kt3k is this ready? I don't see any new method in the `chil_process.ts` file.

kt3k:
@bartlomieju This should add `fork` method. Please see the last line of `node/child_process.ts`. BTW now we have two `child_process.ts` files in `node/` and `node/internal/`.

node/child_process.ts
```diff
@@ -2,299 +2,98 @@
...
+  ERR_INVALID_ARG_VALUE,
+} from "./_errors.ts";
+import { process } from "./process.ts";
+
+const denoCompatArgv = [
+  "run",
+  "--compat",
+  "--unstable",
+  "--no-check",
+  "--location=https://example.com",
```
bartlomieju, line 22: Why is `location` argument needed?

node/child_process.ts
```diff
@@ -2,299 +2,98 @@
...
+  "--allow-all",
```
bartlomieju, line 23: Yeah... this is problematic, ideally we'd use the same permissions as the parent process, but unfortunately we don't have access to these flags.

bartlomieju (commented):
Thanks @kt3k I was looking at the wrong file. This API is gonna be tricky due to flags, but this is good progress

node/child_process.ts
kt3k: Some npm module does something like `typeof location` to detect the environment and causes uncaught errors. So giving `--location` avoids those errors, but that probably should be solved by https://github.com/denoland/deno/issues/12969

Ok. I'll remove this option as a first pass

bartlomieju: Yeah, let's address it in Deno directly

kt3k:
Oh, because we use `deno run --compat` for spawning subcommand, the subprocess have to use `std@0.117.0` for std/node and which still includes `BINARY` notation in enum..

bartlomieju:
Unfortunately I don't see a good way to fix it at the moment, as `deno.land/std/` doesn't allow to link to `main` branch

.github/workflows/ci.yml
```diff
@@ -100,6 +100,8 @@ jobs:
 
       - name: Run Node tests canary
         run: deno test --unstable --allow-all --import-map=test_import_map.json --config strict-ts44.tsconfig.json node/_tools/test.ts
+        env:
+          DENO_NODE_COMPAT_URL: file:///${{ github.workspace }}/
```
bartlomieju, line 104: This is a good workaround for now

bartlomieju (approved):
LGTM

PR #1697, opened 2021-12-09
feat(node/util): Add util.deprecate

wafuwafu13:
Change `node/internal/util.ts` to `node/internal/util.js` because it makes maintenance easier. ref: https://github.com/denoland/deno_std/pull/1605#discussion_r751228984


node/_tools/config.json
```diff
@@ -35,6 +35,7 @@
       "test-stream-readable-event.js",
       "test-stream-writable-change-default-encoding.js",
       "test-url-format-whatwg.js",
+      "test-url-parse-format.js",
```
wafuwafu13, line 38: ref: https://github.com/denoland/deno_std/pull/1667/files#diff-2dc606fcc5aca07ba82c64a68167b04a1af3fbada738bd90d8cd499d81de580a

wafuwafu13:
https://github.com/denoland/deno_std/pull/1694

node/_tools/config.json
kt3k: This is correct. Thanks!

kt3k (approved):
LGTM. Nice!

node/internal/util.js
```diff
@@ -92,10 +93,56 @@ function createDeferredPromise() {
...
 
+// Keep a list of deprecation codes that have been warned on so we only warn on
+// each one once.
+const codesWarned = new Set();
+
+// Mark that a method should not be used.
+// Returns a modified function which warns once by default.
+// If --no-deprecation is set, then it is a no-op.
+export function deprecate(fn, msg, code) {
+  if (process.noDeprecation === true) {
```
kt3k, line 104: It seems we need to import `process`

kt3k (changes requested):
There was an error in CI

wafuwafu13:
in progress...

`error: Uncaught ReferenceError: Cannot access 'ERR_IPC_CHANNEL_CLOSED' before initialization` is caused by importing `process.ts`

node/internal/util.js
bartlomieju: This will most likely cause circular dependency; I suggest to move `doEmitWarning` and `emitWarning` functions from `process.ts` to `_process/process.ts` and import these function from that module, instead of importing `process.ts` directly.

wafuwafu13: `error: Uncaught ReferenceError: Cannot access 'ERR_IPC_CHANNEL_CLOSED' before initialization` does not cause after original test and `process` are deleted.https://github.com/denoland/deno_std/pull/1697/commits/fdf67e459415a7e81839ce9f32fd92d64704d8d3

`process` doesn't need to import for `parallel/test-util-deprecate.js` to pass. https://github.com/denoland/deno_std/pull/1697/commits/8635408a37706c6323f2c1d5234aa75ef4c784d2

kt3k:
> Change node/internal/util.ts to node/internal/util.js because it makes maintenance easier.

I'm in favor of this change.

kt3k (approved):
@wafuwafu13 LGTM. Nice work!

PR #1698, opened 2021-12-09
feat(node): add mock zlib module

kt3k:
This PR adds mock implementation of `zlib` module in std/node.

Some module depeneds on `zlib`'s interfaces even without using its actual features. (e.g. This rollup plugin depends on zlib's interface even when its function is actually not used. ref: https://github.com/btd/rollup-plugin-visualizer/blob/a56a9061da9f8902ec1a7cfcbfef85d6e075b706/plugin/compress.ts ). This addition is useful for those situations.


bartlomieju (approved):
LGTM

PR #1699, opened 2021-12-10
feat(node): add Timeout class

### LINKED ISSUE #1622: std/node: Node compatible setTimeout, setInterval
setTimeout in Node.js returns Timeout object

ref(Node.js):
- https://github.com/nodejs/node/blob/8ee4e672eca6826e7b6f2c1ad36ecbe6e570d412/lib/timers.js#L140
- https://github.com/nodejs/node/blob/ee9e2a2eb6143fa7b4b1454f7aed009a8703d4d7/lib/internal/timers.js#L167

We have `Timeout` class implemented in std/node, but doesn't seem implementing setTimeout correctly
- Timeout https://github.com/denoland/deno_std/blob/ce0ac6ec4b489ec3eab7bbf4991efb8a0c9fe827/node/_timers.ts#L31
- setTimeout https://github.com/denoland/deno_std/blob/ce0ac6ec4b489ec3eab7bbf4991efb8a0c9fe827/node/timers.ts

kt3k:
closes #1622


bnoordhuis (approved):

bartlomieju (dismissed):
I'm against merging this before tomorrow's meeting

bnoordhuis:
@bartlomieju Can you outline your objections?

bartlomieju:
@bnoordhuis importing `node/global.ts` will implicitly override existing timers APIs available in Deno. IIRC we tentatively agreed that we will change return types of global APIs instead of overriding them only in `node/global.ts`. so I'd prefer to discuss it once more before landing.

bartlomieju:
Okay, so we discussed this PR during the design meeting. 

The consensus was that we **shouldn't** mess with global types - we should keep Web compatible types and we shouldn't override `setTimeout` and `setInterval` when importing `deno_std/node` or running in compat mode. What we want to do is return "timer object" when you import these APIs like: `require("timers").setTimeout`. 

We searched using SourceGraph for packages using `ref()/unref()` for timers and we found there are about 500 of them. Most of them are rather popular packages (like `webpack`, `socket.io`, `mocha`) and we feel we should open PRs in these repos that change the usage from using global `setTimeout/setInterval` to one imported from `"timers"` module.

kt3k:
I reverted exposing global part of this PR for now.

> Most of them are rather popular packages (like webpack, socket.io, mocha) and we feel we should open PRs in these repos that change the usage from using global setTimeout/setInterval to one imported from "timers" module.

I feel that might be very difficult because they have small motivation to accept such PRs (in case of mocha, they probably even don't have time to start reviewing it. They seem very inactive.), but let's see how it works for a while.

bnoordhuis:
Seconded. I think you're being way too optimistic on the probability of such PRs getting accepted and merged/released in a timely fashion. To maintainers of those packages it's all downside and little or no upside.

We're aiming for end of January for v0. Being blocked on third parties is not an option.

kt3k:
Merged main.

This PR now doesn't include `replace globals` part. This only provides `timers` module with compatible APIs. So I think this PR is now landable. @bartlomieju

node/_http_client.js
```diff
@@ -0,0 +1,897 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
```
bnoordhuis, line 2: I suspect this file was accidentally resurrected during a rebase?

bnoordhuis (approved):

node/_http_client.js
kt3k: Oh, thanks! This was a mistake during resolving merge conflicts!

kt3k:
Merging this as it doesn't include the controversial part of timers compatibility issue.

I'm going to try to implement the idea of providing node-compatible timers via commonJS wrapper function

Issue #1703, opened 2021-12-13
node: add https.request


kt3k:
part of #1663 

- [x] Add `http.OutgoingMessage` https://github.com/denoland/deno_std/pull/1705
- [x] Add `http.Agent` https://github.com/denoland/deno_std/pull/1706
- [x] Add `http.ClientRequest` https://github.com/denoland/deno_std/pull/1712
- [x] Add `http.get`, `http.request` https://github.com/denoland/deno_std/pull/1712
- [ ] Add tls.connect
- [ ] Add `https.get`, `https.request`

PR #1706, opened 2021-12-14
feat(node): add http.Agent

kt3k:
part of #1703 
~~depends on #1705~~

This PR adds `http.Agent` class in std/node


bnoordhuis (approved):
RSLGTM. I recognize a lot of code from my Node days though. :-)

PR #1710, opened 2021-12-15
chore(node/_tools): enable node.js test filtering

kt3k:
This PR adds a function to filter the Node.js test cases. You can only execute tests which include the given specific patterns:

```sh
deno test -A node/_tools/test.ts -- <pattern-to-match>
```

This should improve the productivity of node.js compat testing


bartlomieju (approved):
Very nice!

PR #1712, opened 2021-12-15
feat(node): add http.request

kt3k:
part of #1703 
depends on #1706


node/internal/timers.ts
```diff
@@ -75,3 +75,10 @@ export function getTimerDuration(msecs: number, name: string) {
 
   return msecs;
 }
+
+export default {
+  kTimeout,
+  TIMEOUT_MAX,
+  setUnrefTimeout,
+  getTimerDuration,
+}
```
bnoordhuis, line 84:
```suggestion
};
```

node/internal/dtrace.ts
```diff
@@ -0,0 +1,23 @@
...
+} = (config.hasDtrace ? dtrace : {});
+
+export {
+  DTRACE_HTTP_CLIENT_REQUEST,
+  DTRACE_HTTP_CLIENT_RESPONSE,
+  DTRACE_HTTP_SERVER_REQUEST,
+  DTRACE_HTTP_SERVER_RESPONSE,
+  DTRACE_NET_SERVER_CONNECTION,
+  DTRACE_NET_STREAM_END,
+};
```
bnoordhuis, line 23: I assume you copied this from node but we're never going to have dtrace support so you might as well simplify this file to always export stubs.

node/internal_binding/dtrace.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+export function DTRACE_HTTP_SERVER_RESPONSE() {
+  notImplemented();
+}
+export function DTRACE_NET_SERVER_CONNECTION() {
+  notImplemented();
+}
+export function DTRACE_NET_STREAM_END() {
+  notImplemented();
+}
```
bnoordhuis, line 23: And remove this file (maybe config.ts too)

bnoordhuis (approved):
Nice work!

kt3k:
I'll continue working on a few more points:

- the server responds with an empty response with the code `res.write("foo"); res.end()`, which prevents some test cases passing.
- the client sometimes hangs with `socket hang up` error. e.g. in `test-http-client-get-url.js`, I still don't find the reason of this

bartlomieju (approved):
LGTM too, great work @kt3k

kt3k:
Thanks for your reviews!

Issue #638, opened 2021-01-21
[STD/FS] Synchronous copy method doesn't overwrite folder with such option enabled


ilovewine:
**ENV**: Linux

**Steps to reproduce:**

- create ./foo with a file.txt (with some data X)
- create ./folder with a file2.txt (with some data Y)
- create the script copy.ts:
```
import { copySync } from "https://deno.land/std@0.81.0/fs/mod.ts";
copySync("./foo", "./folder", {overwrite: true});
```
- run deno run --unstable --allow-read --allow-write copy.ts
**Expected result**: folder ./folder contains only file.txt with data X
**Actual result**: folder ./folder contains both file.txt with data X and file2.txt with data Y

q1qgames:
Seems like this is behaving as designed and should perhaps be a documentation addition with a reference to emptyDir (and the sync counterpart) if desired.

iuioiua:
@bartlomieju, this might be able to be closed, seeing as the PR was closed with good reason.

kt3k:
Yes. Let's close this. We recommend the use of `emptyDir` if you prefer to replace the entire directory contents. `overwrite: true` option means it overwrites the contents of a file if the file of the same name exists in the destination.

PR #1716, opened 2021-12-15
fix: fixing directory copy overwrite behavior

q1qgames:
This is a fix for #638 [STD/FS] Synchronous copy method doesn't overwrite folder with such option enabled.

I believe the current tests cover this case, do I still need to add another one?

First code commit to the project - so I'm sure I fumbled something but we all have to start someplace right?

### DIFF
--- a/fs/copy.ts
+++ b/fs/copy.ts
@@ -1,6 +1,7 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 import * as path from "../path/mod.ts";
 import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
+import { emptyDir, emptyDirSync } from "./empty_dir.ts"; 
 import { getFileInfoType, isSubdir } from "./_util.ts";
 import { assert } from "../_util/assert.ts";
 import { isWindows } from "../_util/os.ts";
@@ -171,6 +172,7 @@ async function copyDir(
     ...options,
     isFolder: true,
   });
+  
 
   if (!destStat) {
     await ensureDir(dest);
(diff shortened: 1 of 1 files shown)


q1qgames:
and I didn't include the await decorator for the async case - told you I'd fumble something.  Fixing.

kt3k:
Thank you for your contribution, but I personally think this behavior is not desired. `overwrite` option is used when the source file overwrites the existing file in the destination. Emptying the destination feels like more of that behavior.

I'd suggest we should introduce a new option (such as `delete`) for doing this.

What do you think?

> I believe the current tests cover this case, do I still need to add another one?

It seems this particular case (the destination has more files than the source directory) is not covered by the test cases.

q1qgames:
> I personally think this behavior is not desired. `overwrite` option is used when the source file overwrites the existing file in the destination. Emptying the destination feels like more of that behavior.

I agree that the extant behavior makes sense.  I implemented what was in the open issue given it was marked help wanted, I didn't venture into design considerations :). 

> 
> I'd suggest we should introduce a new option (such as `delete`) for doing this.
> 
> What do you think?

I don't know that adding further options is necessarily a good idea.

In case that you'd want to delete the contents of the directory prior to move I imagine you'd just use emptyDir before you did the copy operation.

If the current behavior is the desired, I think keeping everything else atomic is better.

bartlomieju:
@q1qgames please sign the CLA so we could merge this PR :)

kt3k:
@bartlomieju I don't think this change is a good idea.

As suggested by @q1qgames,
> In case that you'd want to delete the contents of the directory prior to move I imagine you'd just use emptyDir before you did the copy operation.

I think we should recommend using `emptyDir` on their own before using `copy`. WDYT?

bartlomieju:
> I think we should recommend using emptyDir on their own before using copy. WDYT?

@kt3k yes I agree, seems like a better solution.

q1qgames:
I'm going to close this given I concur with the conclusion.

On to the next one! 

Thanks for your time.

Issue #1707, opened 2021-12-14
Refactor: Move internals file in node folder to `/node/internal`


getspooky:
Move internals files like `_net.ts` , `_errors` , etc to `node/internal` .
Target: 
- [x] _net.ts
- [x] _async_hooks.ts #2051
- [x] _buffer.js #2643
- [x] _dtrace.ts
- [x] _errors_test.ts, _errors.ts
- [x] _fixed_queue.ts https://github.com/denoland/deno_std/pull/2064
- [x] _idna.ts
- [x] _options.ts #2065
- [x] _readline.js
- [x] _stream.js
- [x] _timers.ts #1699
- [x] `_util/_debuglog,ts` -> `internal/util/debuglog.ts`

kt3k:
@getspooky Thank you for creating this issue!

Right. We should keep the file structure as close as possible to Node.js structure. The above change make sense to me.

BTW `node/_core.js` is something we invented on our own. Maybe we should keep it somewhere else?

kt3k:
Also `deno_std/node/_buffer.js` seems corresponding to `node/lib/buffer.js`. So let's just rename it to `deno_std/node/buffer.js` in this case.

Please carefully compare node.js `lib/` directory and deno_std `node/` directory before starting working on each item

getspooky:
@kt3k thanks for feedback, i will work on it.

kt3k:
Also `_timers` will conflicts with #1699. Please avoid it at first.

kt3k:
I updated the description, and crossed out some items (`_core.ts` `_events.js` `_next_tick.ts` `_utils.ts`) which don't seem existing in `internal/`

iuioiua:
@kt3k, I'll do `_readline.mjs`. Is there any issue in essentially merging the `.mjs` and `.d.ts` files into a `.ts` file? I gather that `.mjs` is separated to make following the node version easier to follow.

kt3k:
@iuioiua I think there's no such issue. Do you think we should merge .mjs and .d.ts into .ts?

kt3k:
Looks these are all done now. Thanks for your efforts!

PR #1717, opened 2021-12-16
ref(node): Move `_dtrace` to `internal/dtrace`

getspooky:
Ref https://github.com/denoland/deno_std/issues/1707


bnoordhuis:
That happened earlier today in #1712 but, as mentioned in that PR, we're never going to support dtrace anyway and might as well remove it altogether.

kt3k:
Also this particular change is also included in #1712. Closing without merge. Thank you @getspooky for working on this anyways

PR #1719, opened 2021-12-17
chore(path): Defaults the extended argument to true in path's globToRegEx

### LINKED ISSUE #794: Default options for path/glob and fs/expand_glob differ
Not sure if this was done on purpose, but the defaults for the `extended` and `globstar` options in path/glob/`globToRegExpr`...

https://github.com/denoland/deno_std/blob/27dcc84ae1fdafa32eb77a9ab2caaa768ccea2f6/path/glob.ts#L83

...and fs/expand_glob/`expandGlob`:

https://github.com/denoland/deno_std/blob/27dcc84ae1fdafa32eb77a9ab2caaa768ccea2f6/fs/expand_glob.ts#L79

...are the opposite. Perhaps it would be better to use the same defaults for both?

mitch292:
Resolves  #794

Took the suggestion from the comment to default path's globToRegEx to `false` as `expandGlob`, `expandGlobSync` and `joinGlobs` all default to `false`


kt3k:
Thank you for your contribution.

I did some research about other major tools and I found that npm's de facto glob tool ([glob](https://www.npmjs.com/package/glob)) seems enabling 'extglob' by default. So I guess enabling this in all other APIs (`expandGlob`, `expandGlobSync` and `joinGlobs`) would be more sensible. What do you think?

mitch292:
> joinGlobs

Ah nice catch, I updated the PR!

mitch292:
> Thank you for your contribution.
> 
> I did some research about other major tools and I found that npm's de facto glob tool ([glob](https://www.npmjs.com/package/glob)) seems enabling 'extglob' by default. So I guess enabling this in all other APIs (`expandGlob`, `expandGlobSync` and `joinGlobs`) would be more sensible. What do you think?

Ah nice catch, I updated the PR!

kt3k (approved):
LGTM

kt3k:
Let's wait for a little while to see if there's any feedback/opinions/objections on this.

kt3k:
Thank you for working on this! @mitch292

kt3k:
@bartlomieju Technically this is a breaking change. Do you think we should wait to land this until the next minor CLI release (1.18)?

bartlomieju:
> @bartlomieju Technically this is a breaking change. Do you think we should wait to land this until the next minor CLI release (1.18)?

I don't have a strong opinion. If this change aligns us better with user expectations (due to how similar libs work in Node) I think we can land it now

kt3k:
Ok. Let's land now (assuming the current state as 'bug', and labeling this change as 'fix')

PR #1722, opened 2021-12-17
ref(node): Move _errors to `internal/error.ts`

getspooky:
Ref https://github.com/denoland/deno_std/issues/1707


kt3k:
CI seems blocked by the strange error: `error: TS5056 [ERROR]: Cannot write file 'deno://node/internal/errors.js' because it would be overwritten by multiple input files.`..

getspooky:
@kt3k  can you suggest solution ?

kt3k:
@getspooky Thank you for your work. We landed this at #1841

PR #1724, opened 2021-12-18
BREAKING(encoding/csv): add return type to csv's parse and remove a parse func from args

### LINKED ISSUE #881: std/encoding/csv.ts - Improve `parse` return type
Right now it is typed as `unknown[]` (https://github.com/denoland/deno/blob/d199e45ad5dd180a6969abcd528023e173a1e1ff/std/encoding/csv.ts#L392), so in order to figure out what the return type is, I have to either look at the implementation or run it once and check the output. There should be a way to type this properly.

mitch292:
Resolves #881

Took from a comment in the issue by @dsherret and removed the ability to pass a `parse` arg to the function which gave it the responsibility of parsing and, optionally, data transformation. Let me know wha you think!

### DIFF
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -365,11 +365,11 @@ export interface ParseOptions extends ReadOptions {
 
 /**
  * Csv parse helper to manipulate data.
- * Provides an auto/custom mapper for columns for columns and rows.
+ * Provides an auto/custom mapper for columns.
  * @param input Input to parse. Can be a string or BufReader.
  * @param opt options of the parser.
- * @returns If you don't provide `opt.skipFirstRow`, `opt.parse`, and `opt.columns`, it returns `string[][]`.
- *   If you provide `opt.skipFirstRow` or `opt.columns` but not `opt.parse`, it returns `object[]`.
+ * @returns If you don't provide `opt.skipFirstRow` and `opt.columns`, it returns `string[][]`.
+ *   If you provide `opt.skipFirstRow` or `opt.columns`, it returns `Record<string, unkown>[]`.
  */
 export async function parse(
   input: string | BufReader,
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -421,8 +421,7 @@ export async function parse(
       i++;
       const out: Record<string, unknown> = {};
       for (let j = 0; j < e.length; j++) {
-        const h = headers[j];
-        out[h.name] = e[j];
+        out[headers[j].name] = e[j];
       }
       return out;
     });
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -334,6 +334,8 @@ export async function readMatrix(
  * Parse the CSV string/buffer with the options provided.
  *
  * ColumnOptions provides the column definition
+ * and the parse function for each entry of the
+ * column.
  */
 export interface ColumnOptions {
   /**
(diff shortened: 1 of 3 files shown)


encoding/csv.ts
```diff
@@ -334,20 +334,12 @@ export async function readMatrix(
...
   /**
    * Name of the column to be used as property
    */
   name: string;
-  /**
-   * Parse function for the column.
-   * This is executed on each entry of the header.
-   * This can be combined with the Parse function of the rows.
-   */
-  parse?: (input: string) => unknown;
```
mitch292, line 350: I may be reading too much into the comment thread in #881. To keep the logic consistent I removed the parse function from columns as well. Should this be done? 

Now, if the parse is removed from column options maybe the function signature for `columns?: string[] | ColumnOptions[]` should just be simplified to `columns?: string[]` as below will evaluate to the same.

```js
parse(data, { columns: ['a', 'b', 'c']});

parse(data, { columns: [{ name: 'a' }, { name: 'b' }, { name: 'c' } ] });
```

kt3k: I'm in favor of removing this

encoding/csv.ts
```diff
@@ -361,44 +353,22 @@ export interface ParseOptions extends ReadOptions {
...
+ * @returns If you don't provide `opt.skipFirstRow` and `opt.columns`, it returns `string[][]`.
+ *   If you provide `opt.skipFirstRow` or `opt.columns`, it returns `Record<string, unkown>[]`.
  */
 export async function parse(
   input: string | BufReader,
   opt: ParseOptions = {
     skipFirstRow: false,
   },
-): Promise<unknown[]> {
+): Promise<string[][] | Record<string, unknown>[]> {
```
kt3k, line 371: I think if we do overloading like the below, it would be able to return more correct type depending of the options:

```ts
export async function parse(
  input: string | BufReader
): Promise<string[][]>;
export async function parse(
  input: string | BufReader,
  opt: Omit<ParseOptions, "columns" | "skipFirstRow">,
): Promise<string[][]>;
export async function parse(
  input: string | BufReader,
  opt: Omit<ParseOptions, "columns"> & {
    columns: string[] | ColumnOptions[],
  },
): Promise<Record<string, unknown>[]>;
export async function parse(
  input: string | BufReader,
  opt: Omit<ParseOptions, "skipFirstRow"> & {
    skipFirstRow: true,
  },
): Promise<Record<string, unknown>[]>;
export async function parse(
  input: string | BufReader,
  opt: ParseOptions = {
    skipFirstRow: false,
  },
): Promise<string[][] | Record<string, unknown>[]> { ... }
```

kt3k (commented):
@mitch292 
Thanks for working on this! Left a few comments. I think we can do a little further better typing

mitch292:
Added some function overloads and updated the readme as well!

dsherret (approved):
LGTM.

Yeah, people can just take the output and put it through their own parse function now in a way that works well with TypeScript. This is nicer than typing everything as `unknown`.

bx2:
Hi! This change renders the entire point of having column options pointless. We do allow to change column names (which is transforming the data) but we do not allow to parse the given column to, for instance, change the type of the field on the fly. I will give you an example - say you are parsing a file with monetary values passed in as micros:

```
YearsExperience,Salary
10, 23345213000
```

To get the decimal value you would divide the Salary field by 1000000. With the parse added per column, you can do all the basic transformations while loading the CSV, something like what pandas in python does (https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html). Currently, you cannot do that anymore. 

Also, having column options just to rename the fields but being required to always specify ALL the fields seems somewhat contrary what you are trying to accomplish.

Now, while I do understand that the basic parser should be just that and only parse, but then let's skip the entire concept of column options at once.

PR #1729, opened 2021-12-19
feat(hash): Add Tiger hash

TOGoS:
I have long longed to calculate tiger hashes from Deno, so that I may generate [bitprint URNs](http://www.nuke24.net/docs/2015/HashURNs.html).

This just uses the "[tiger](https://docs.rs/tiger/latest/tiger/)" Rust crate.  Adding it was far easier than I expected.

Let me know if you'd like more test cases.  I just copied a few from https://www.cs.technion.ac.il/~biham/Reports/Tiger/test-vectors-nessie-format.dat

Edit (reflected as force pushes): more test cases, add "tiger" to the algorithms list.


kt3k (approved):
Tiger seems like a common hash function in peer-to-peer network protocols (I'm not expert of it though) ref: https://en.wikipedia.org/wiki/Magnet_URI_scheme

I'm in favor of including this. LGTM

TOGoS:
> Tiger seems like a common hash function in peer-to-peer network protocols (I'm not expert of it though) ref: https://en.wikipedia.org/wiki/Magnet_URI_scheme

Yep, exactly.  Moreover, implementing it in pure JavaScript is an enormous pain due to it being defined in terms of 64-bit arithmetic.  I don't have much experience compiling Rust to Wasm but was pretty tickled at how easy it was to do as part of this library.  If I have to I'll fork off the parts I need in order to finally be able to get my Tiger hashes, but getting it into the standard library would definitely be preferable (back when I was doing a lot of PHP development I was pleased to find it built-in to PHP as "tiger192,3").

lucacasonato (approved):
LGTM!

kt3k:
We also landed Tiger hash in `crypto` module, which is more recommended way to use digest functions now. #1749

```ts
import { crypto } from "https://deno.land/std/crypto/mod.ts";
const digest = await crypto.subtle.digest("TIGER", data);
```

PR #1736, opened 2021-12-21
chore(node): enabled fixed test cases

kt3k:
This PR removes items from `ignore` section of node testing config. These cases now pass the original test cases without issues


bnoordhuis (approved):
Maybe s/enabled/enable/ in the commit log; past -> present tense and can't be mistaken for an adjective.

kt3k:
Oh, I usually use present tense there... Thanks for catching it!

PR #1737, opened 2021-12-21
chore(node): add stub modules

### LINKED ISSUE #1734: node: Add stubs for APIs in several modules
We should add stubs for following modules, so we can set them up automatically in REPL when running in `compat` mode:

- [ ] cluster
- [ ] domain
- [ ] https
- [ ] tls
- [ ] v8

The idea is to add top level export that would throw `not implemented` error instead of throwing the error when the module is loaded.

Ref https://github.com/denoland/deno/pull/13127

kt3k:
This PR adds stub implementations of missing node modules.

closes #1734

### DIFF
--- a/node/http2.ts
+++ b/node/http2.ts
@@ -46,7 +46,7 @@ export class Http2SecureServer {
 export function createServer() {}
 export function createSecureServer() {}
 export function connect() {}
-export const constants = undefined;
+export const constants = {};
 export function getDefaultSettings() {}
 export function getPackedSettings() {}
 export function getUnpackedSettings() {}
--- a/node/http2.ts
+++ b/node/http2.ts
@@ -50,7 +50,7 @@ export const constants = {};
 export function getDefaultSettings() {}
 export function getPackedSettings() {}
 export function getUnpackedSettings() {}
-export const sensitiveHeaders = undefined;
+export const sensitiveHeaders = Symbol("nodejs.http2.sensitiveHeaders");
 export class Http2ServerRequest {
   constructor() {
     notImplemented();
--- a/node/http2.ts
+++ b/node/http2.ts
@@ -1,0 +1,83 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
+
+import { notImplemented } from "./_utils.ts";
+
+export class Http2Session {
+  constructor() {
+    notImplemented();
+  }
+}
+export class ServerHttp2Session {
+  constructor() {
+    notImplemented();
+  }
+}
+export class ClientHttp2Session {
+  constructor() {
+    notImplemented();
+  }
+}
+export class Http2Stream {
+  constructor() {
+    notImplemented();
+  }
+}
+export class ClientHttp2Stream {
+  constructor() {
+    notImplemented();
+  }
+}
+export class ServerHttp2Stream {
+  constructor() {
+    notImplemented();
+  }
+}
+export class Http2Server {
+  constructor() {
+    notImplemented();
+  }
+}
+export class Http2SecureServer {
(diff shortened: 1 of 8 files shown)


node/http2.ts
```diff
@@ -0,0 +1,83 @@
...
+}
+export class Http2SecureServer {
+  constructor() {
+    notImplemented();
+  }
+}
+export function createServer() {}
+export function createSecureServer() {}
+export function connect() {}
+export const constants = undefined;
```
bnoordhuis, line 49:
```suggestion
export const constants = {};
```
(just a suggestion, `undefined` isn't really wrong either)

node/http2.ts
```diff
@@ -0,0 +1,83 @@
...
+export function getDefaultSettings() {}
+export function getPackedSettings() {}
+export function getUnpackedSettings() {}
+export const sensitiveHeaders = undefined;
```
bnoordhuis, line 53:
```suggestion
export const sensitiveHeaders = Symbol("nodejs.http2.sensitiveHeaders");
```

node/repl.ts
```diff
@@ -0,0 +1,19 @@
...
+// Copyright Joyent and Node contributors. All rights reserved. MIT license.
+
+import { notImplemented } from "./_utils.ts";
+
+class REPLServer {
+  constructor() {
+    notImplemented();
+  }
+}
+export const builtinModules = undefined;
```
bnoordhuis, line 11:
```suggestion
export const builtinModules = ["assert", "async_hooks", "buffer", "child_process", "cluster", "console", "constants", "crypto", "dgram", "diagnostics_channel", "dns", "domain", "events", "fs", "http", "http2", "https", "inspector", "module", "net", "os", "path", "perf_hooks", "process", "punycode", "querystring", "readline", "repl", "stream", "string_decoder", "sys", "timers", "tls", "trace_events", "tty", "url", "util", "v8", "vm", "wasi", "worker_threads", "zlib"];
```
Which, granted, is a bit of a lie.

node/tls.ts
```diff
@@ -1 +1,47 @@
...
+export function createSecurePair() {
+  notImplemented();
+}
+export function createServer() {
+  notImplemented();
+}
+export function getCiphers() {
+  notImplemented();
+}
+export const rootCertificates = undefined;
```
bnoordhuis, line 28: Side discussion: open question how to support this. Node uses a hard-coded list of root certificates, based off Mozilla's NSS certdata.txt.

We use `rustls_native_certs::load_native_certs()` but that produces different results on different platforms. For reasons of compatibility, we may want to use node's list instead.

bnoordhuis (approved):
👍

node/repl.ts
kt3k: Looks good to me. Thanks for the suggestion!

node/tls.ts
kt3k: maybe we should add an option to `Deno.ConnectTlsOptions` for completely swap the root certificates? https://doc.deno.land/deno/stable/~/Deno.ConnectTlsOptions

bnoordhuis: Yes, I agree. That'd be useful for more than just node compat too.

lucacasonato: @bnoordhuis `rustls_native_certs::load_native_certs()` no we don't. We use the same mozilla list by default. You can switch to your native cert list by setting `DENO_TLS_CA_STORE=system` or `DENO_TLS_CA_STORE=system,mozilla` to use both.

bnoordhuis: Oh, that's right - I believe I even reviewed the PR that added that. 🤦

Then perhaps we want to bypass `DENO_TLS_CA_STORE=system` and stick to the built-in list.

bartlomieju (approved):
LGTM!

PR #1720, opened 2021-12-17
feat(uuid): Add `validate` function to check UUID

getspooky:
Ref https://github.com/denoland/deno_std/issues/1715

PR #1745, opened 2021-12-22
style(uuid): add return type to `validate`

jsejcksn:
Follow up to https://github.com/denoland/deno_std/pull/1720

Side note: it seems like almost all code in this repo uses explicit return types, however, it's not mentioned in the [style guide](https://deno.land/manual@v1.17.0/contributing/style_guide). Should it be?

### DIFF
--- a/uuid/mod.ts
+++ b/uuid/mod.ts
@@ -33,7 +33,7 @@ export function isNil(id: string): boolean {
  * validate("6ec0bd7f-11c0-43da-975e-2a8ad9ebae0b") // true
  * ```
  */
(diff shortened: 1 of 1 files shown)


kt3k (approved):
@jsejcksn LGTM. Thanks

> Side note: it seems like almost all code in this repo uses explicit return types, however, it's not mentioned in the style guide. Should it be?

I think we don't have explicit rule about this. (Sometimes we even avoid them. ref: https://github.com/denoland/deno_std/pull/819/files)

Issue #1738, opened 2021-12-21
node: http.request is flaky


kt3k:
The following script doesn't show the response always (sometimes it shows the response though)
```js
import "./node/global.ts";
import * as http from "./node/http.ts";

const req = http.request("http://google.com", (res) => {
  let data = "";
  res.on("data", (chunk) => {
    data += chunk;
  });
  res.on("end", () => {
    console.log(data);
  });
});
req.end();
```

PR #1746, opened 2021-12-22
feat(node): add https.request

### LINKED ISSUE #1703: node: add https.request

kt3k:
closes #1703 

This PR implements `https.request` using `fetch` API.

I first tried to follow the node.js implementation more closely, but it seems too complicated to follow (wasted almost 2 days on it...). I also heard feedbacks in the design meetings and more people seem preferring implementation by `fetch`. So I switched to this version.

If this PR looks good, I'll align `http.request` to this implementation as it seems flaky ref: #1738

### DIFF
--- a/node/https.ts
+++ b/node/https.ts
@@ -1,7 +1,10 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
+// deno-lint-ignore-file no-explicit-any ban-types
 
 import { notImplemented } from "./_utils.ts";
+import { Readable, Writable } from "./stream.ts";
+import { urlToHttpOptions } from "./internal/url.ts";
 
 export class Agent {
   constructor() {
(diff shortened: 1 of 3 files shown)


node/https.ts
```diff
@@ -16,12 +18,195 @@ export class Server {
...
+export function get(...args: any[]) {
+  const req = request(args[0], args[1], args[2]);
+  req.end();
+  return req;
 }
+
 export const globalAgent = undefined;
-export function request() {
-  notImplemented();
+/** HttpsClientRequest class loosely follows http.ClientRequest class API. */
```
bartlomieju, line 78: Shouldn't this class match `http.ClientRequest` exactly?

bartlomieju (approved):
LGTM

node/https.ts
kt3k: Yup, this should be as close as possible to http.ClientRequest. The comment describes the current state. I'll continue work on that on follow-up PRs.

By the way, there're some props which don't make sense with this `fetch` implementation. For example, `req.socket` is a Duplex stream of underlying tcp connection, which we probably can't provide in our code.

PR #1748, opened 2021-12-23
feat(node): add process.execPath

kt3k:
This PR adds `process.execPath` in std/node.

This resolves one of the issues in running npm in compat mode. ref: https://github.com/denoland/deno/issues/13187


bartlomieju (approved):
LGTM, I'm still not convinced that we should require `--allow-read` flag to get this value

kt3k:
> I'm still not convinced that we should require --allow-read flag to get this value

There was also a similar question before. https://github.com/denoland/deno_std/pull/749#issuecomment-786534313

PR #1749, opened 2021-12-23
feat(crypto): Add Tiger hash

### LINKED ISSUE #1742: crypto: add tiger hash
Add Tiger hash to crypto module in a similar way as #1729

kevincennis:
Fixes #1742

[Edit] Apologies for the force push. CI failed on my fork because of a delta in the WASM, which I think might've been because I built it on an M1 Macbook. So I just grabbed the uploaded artifact from CI, committed that, and it seems happy now.


kt3k (approved):
@kevincennis Thanks! LGTM

PR #1751, opened 2021-12-23
refactor(node): use fetch in http.request impl

### LINKED ISSUE #1738: node: http.request is flaky

kt3k:
This PR aligns **http**.request implementation to **https**.request.

closes #1738 as a side effect


bartlomieju (approved):
LGTM, a little disconcerting that all these tests are being disabled.

kt3k:
> a little disconcerting that all these tests are being disabled.

I'll try enabling them in later PRs.

Issue #1752, opened 2021-12-23
node: 'exit' event is emitted twice when process.exit() is called


kt3k:
```
$ deno repl --unstable --compat
Deno 1.17.0+ac06797
exit using ctrl+d or close()
> process.on("exit", () => console.log("exit")); process.exit();
exit
exit
```

This causes `npm` error at exit

bartlomieju:
I think I know what's the problem. I'll look into it.

PR #1753, opened 2021-12-23
fix(node): don't emit exit twice

bartlomieju:
Fixes https://github.com/denoland/deno_std/issues/1752

Related to https://github.com/denoland/deno_std/pull/1628


kt3k (approved):
LGTM. Thanks!

PR #1757, opened 2021-12-24
test: do not require NODE_EXTRA_CA_CERTS env var

### LINKED ISSUE #1754: Incomplete documentation for running tests
**Describe the bug**

Right now, the README file says you should be able to run `deno test --unstable --allow-all`, but that actually fails with:

```
error: Uncaught TypeError: error sending request for url (https://localhost:4505/): error trying to connect: invalid peer certificate contents: invalid peer certificate: UnknownIssuer
    at async mainFetch (deno:ext/fetch/26_fetch.js:259:14)
```

I'm not 100% sure what the best option is here. Seems the docs could be updated to include the same environment variable as the `ci.yml` action:

```
NODE_EXTRA_CA_CERTS=http/testdata/tls/RootCA.pem deno test --unstable --allow-all
```

But that seems a bit cumbersome. 

**Steps to Reproduce**

```
deno test --unstable --allow-all
```

or, much more quickly

```
deno test --unstable --allow-all  node/https_test.ts
```

**Expected behavior**

Tests pass

**Environment**

- OS: MacOS 11.6
- deno version: 1.17.1
- std version: 0.119.0

kt3k:
closes #1754

As reported in #1754, now we require `NODE_EXTRA_CA_CERTS` env var set correctly to pass the test command, but that's too inconvenient. This PR fixes it.


node/https_test.ts
```diff
@@ -19,17 +18,30 @@ Deno.test("[node/https] request makes https request", async () => {
...
-
+  const p = Deno.run({
+    cmd: [
+      Deno.execPath(),
+      "run",
+      "--quiet",
+      "--unstable",
+      "--allow-all",
+      "--no-check",
+      "node/testdata/https-request.ts",
```
bartlomieju, line 29: Prefer underscore in file name?

node/https_test.ts
```diff
@@ -41,16 +53,29 @@ Deno.test("[node/https] get makes https GET request", async () => {
...
-
+  const p = Deno.run({
+    cmd: [
+      Deno.execPath(),
+      "run",
+      "--quiet",
+      "--unstable",
+      "--allow-all",
+      "--no-check",
+      "node/testdata/https-get.ts",
```
bartlomieju, line 64: Ditto

bartlomieju (approved):

node/https_test.ts
kt3k: Thanks. Makes sense

PR #1760, opened 2021-12-24
applyState has no 'connect'

clemens-tolboom:


kt3k (approved):
@clemens-tolboom LGTM. Thanks!

kt3k:
Note: This fixes the example chat client behavior. Currently the UI (buttons) don't change their states when the server shutdown. Now they got inactive when the server shutdown.

PR #1762, opened 2021-12-25
fix(node/child_process): Allow number and boolean env vars.

### LINKED ISSUE #1725: node: environment variables can contain non-string values
Node.js can accept environment variable values that are string, numeric, or boolean. On the other hand, Deno requires those values to be strings.

**Steps to Reproduce**

```js
// script.js

const child_process = require("child_process");

child_process.spawn("sh", [ "-c", "echo 'Hello'" ], {
    env: { N: 1 },
    stdio: [ "inherit", "inherit", "inherit" ],
});
```
```
> node script.js
Hello
```
```
> deno run --compat --unstable --allow-net --allow-read --allow-run script.js
Check https://deno.land/std@0.118.0/node/global.ts
Check https://deno.land/std@0.118.0/node/global.ts
Check https://deno.land/std@0.118.0/node/global.ts
Check https://deno.land/std@0.118.0/node/module.ts
error: Uncaught TypeError: Error parsing args: serde_v8 error: ExpectedString
      this.#process = Deno.run({
                           ^
    at Object.opSync (deno:core/01_core.js:149:12)
    at opRun (deno:runtime/js/40_process.js:27:17)
    at Object.run (deno:runtime/js/40_process.js:114:17)
    at new ChildProcess (https://deno.land/std@0.118.0/node/internal/child_process.ts:123:28)
```

**Expected behavior**

I expect the script to work in Deno when using the Node compatibility layer. The [process.env documentation](https://nodejs.org/api/process.html#processenv) says:

> Assigning a property on process.env will implicitly convert the value to a string. This behavior is deprecated. Future versions of Node.js may throw an error when the value is not a string, number, or boolean.

I think this means that Node.js in the future might no longer perform this implicit conversion and continue to allow string, number, or boolean values.

Could the compatibility layer coerce all of the values it receives to strings prior to passing them on to Deno? If this sounds reasonable, I can provide a PR.

**Environment**

- OS: Arch Linux
- deno version: 1.17.0
- std version: 0.118.0

tjni:
This is useful for additional compatibility with Node.JS, which allows
numbers and booleans to be values to environment variables. In Deno, we
convert those values to strings to match the signature of Deno#run.

This should resolve #1725.

### DIFF
--- a/node/child_process_test.ts
+++ b/node/child_process_test.ts
@@ -12,8 +12,8 @@ import { isWindows } from "../_util/os.ts";
 import * as path from "../path/mod.ts";
 import { Buffer } from "./buffer.ts";
 
-function withTimeout(timeoutInMS: number): Deferred<void> {
-  const promise = deferred<void>();
+function withTimeout<T>(timeoutInMS: number): Deferred<T> {
+  const promise = deferred<T>();
   const timer = setTimeout(() => {
     promise.reject("Timeout");
   }, timeoutInMS);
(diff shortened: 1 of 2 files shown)


kt3k (approved):
@tjni LGTM. Thanks!

PR #1764, opened 2021-12-26
Implement orderedPooledMap

### LINKED ISSUE #1763: async/pool: returned results may be out of order
pooledMap is helpful for many IO programs. However, IO is non-deterministic: newer requests may be resolved sooner than older requests. For pooledMap, this means that the sequence for returned results becomes less reliable as we increase the pool limit. This could lead to undesirable effects in certain applications that include updating the UI and writing to a database.

I would find it useful to have a version of pooledMap that provides some indication of the results's place in the original array.

lilnasy:
Closes #1763

### DIFF
--- a/async/mod.ts
+++ b/async/mod.ts
@@ -3,6 +3,7 @@ export * from "./debounce.ts";
 export * from "./deferred.ts";
 export * from "./delay.ts";
 export * from "./mux_async_iterator.ts";
+export * from "./ordered_pool.ts";
 export * from "./pool.ts";
 export * from "./tee.ts";
 export * from "./deadline.ts";
--- a/async/ordered_pool.ts
+++ b/async/ordered_pool.ts
@@ -12,50 +12,46 @@
  * @param iteratorFn The async function to call for every item of the array.
  */
 export function orderedPooledMap<T, R>(
-	poolLimit: number,
-	array: Iterable<T> | AsyncIterable<T>,
-	iteratorFn: (data: T) => Promise<R>,
+  poolLimit: number,
+  array: Iterable<T> | AsyncIterable<T>,
+  iteratorFn: (data: T) => Promise<R>,
 ) {
-	// Create the readable stream whose async iterable is returned from this function.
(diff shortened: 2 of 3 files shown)


kt3k:
Can I assume this is order-preserving version of `pooledMap`? I tried the below script and the order of items doesn't seem preserved:

```js
import { delay, orderedPooledMap } from "./async/mod.ts";

for await (
  const x of orderedPooledMap(
    3,
    Array.from({ length: 10 }, (_, i) => i),
    async (x) => {
      await delay(Math.random() * 1000);
      return x;
    },
  )
) {
  console.log(x);
}
```

Am I missing something?

lilnasy:
I should clarify a bit more in the comments, but basically the order is a value inside the returned value, rather than in the timing of its return.
```ts
import { delay, orderedPooledMap } from "./async/mod.ts";

for await (
  const [i,x] of orderedPooledMap(
    3,
    Array.from({ length: 10 }, (_, i) => i),
    async (x:any) => {
      await delay(Math.random() * 1000);
      return x;
    },
  )
) {
  console.log('\t'.repeat(i) + x)
}
```

lilnasy:
It's possible to instead make the returned values arrive in the original sequence, but this technique is a compromise because some iterations can be blocking: an iteration taking p99 time will make all following iterations take p99 time as well.

kt3k:
If you need to know the invoked order of each async operation, you can do it like the below:

```ts
import { delay, pooledMap } from "https://deno.land/std/async/mod.ts";
let c = 0;
for await (
  const [i,x] of pooledMap(
    3,  
    Array.from({ length: 10 }, (_, i) => i), 
    async (x) => {
      const index = c++;
      await delay(Math.random() * 1000);
      return [index, x]; 
    },  
  )
) {
  console.log('\t'.repeat(i) + x)
}
```

I think this isn't such complex construction. So I doubt this proposed API is worth it.

lilnasy:
yeah, i see what you mean

PR #1765, opened 2021-12-26
Implementation of deepStrictEqual

### LINKED ISSUE #662: Implement std/node/util isDeepStrictEqual
`isDeepStrictEqual` is not yet supported from Node's util library: https://nodejs.org/dist/latest-v14.x/docs/api/util.html#util_util_isdeepstrictequal_val1_val2

Implementation can be found here and should be added to the Deno std library + tests: https://github.com/nodejs/node/blob/master/lib/internal/util/comparisons.js

I would like to take this task upon me if adding this is fine.

standvpmnt:
Fixes #662 the test suite passes, there is one issue however, i think that is we should probably be having this check of getOwnPropertySymbols and getOwnPropertyNames somewhere earlier as presently the location where this is being done (inside the isBoxedPrimitive) seems rather prone to error as the check is only for length. 

Is there any other way we can implement this check to ensure that the code catches a scenario wherein prototype modifications are done on an object in a way so that the length of `getOwnPropertySymbols` and `getOwnPropertyNames` is same in both boxed primitive as well as the modified prototype object.

### DIFF
--- a/node/_tools/suites/parallel/test-util-isDeepStrictEqual.js
+++ b/node/_tools/suites/parallel/test-util-isDeepStrictEqual.js
@@ -94,7 +94,6 @@ function utilIsDeepStrict(a, b) {
   assert.strictEqual(util.isDeepStrictEqual(a, b), true);
   assert.strictEqual(util.isDeepStrictEqual(b, a), true);
 }
-
 function notUtilIsDeepStrict(a, b) {
   assert.strictEqual(util.isDeepStrictEqual(a, b), false);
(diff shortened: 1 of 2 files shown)


kt3k:
This finally passes all the native test cases! Nice work!

> there is one issue however, i think that is we should probably be having this check of getOwnPropertySymbols and getOwnPropertyNames somewhere earlier as presently the location where this is being done (inside the isBoxedPrimitive) seems rather prone to error as the check is only for length.

Can you leave TODO comments for these concerns?

kt3k (approved):
LGTM!

PR #1766, opened 2021-12-26
feat(uuid): add version function to detect RFC version of a UUID

erfanium:
refs:
https://github.com/uuidjs/uuid/blob/main/src/version.js
https://github.com/uuidjs/uuid/blob/main/test/unit/version.test.js


kt3k (approved):
@erfanium LGTM. Thanks!

bartlomieju (approved):
LGTM

PR #1773, opened 2021-12-30
feat(http/file_server): add 'quiet' flag

### LINKED ISSUE #1768: Make `file_server.ts` less verbose

When hosting a site (for development) that is making a lot of requests, the console is spammed with tons of messages for every request. I really don't need to know about every request, and if a request fails I can see that from the browser inspector.

Either disable logging for requests by default and add a `--verbose` option. Or leave it enabled by default and have an option for disabling it.

Wrapping the server with a `Deno.run()` call with `stdout: "null"`. But this also hides the "HTTP server listening on" message.

erfanium:
fixes #1768

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -38,6 +38,8 @@ interface FileServerArgs {
   key: string;
   // -h --help
   help: boolean;
+  // --quiet
+  quiet: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


bnoordhuis (approved):
Thanks, LGTM.

Issue #1445, opened 2021-10-22
Hashing string with sha-384 always returns an empty ArrayBuffer


herkulessi:
**Describe the bug**
When hashing something with sha-384, the await subtle.digest("sha-384", stringToHash) alway results in an empty ArrayBuffer.
**Steps to Reproduce**

1. Create a file with the example usage from [std@0.112.0/crypto](https://deno.land/std@0.112.0/crypto#usage)
2. Replace the `"hello world"` string with `new TextEncoder().encode("hello world")` (because the function does not take strings anymore)
3. Run the Script

**Expected behavior**

Something like
```
Check file:///PathToFile/cryptoTest.ts
Uint8Array(48) [
  253, 189, 142, 117, 166, 127,  41, 247,   1,
  164, 224,  64,  56,  94,  46,  35, 152,  99,
    3, 234,  16,  35, 146,  17, 175, 144, 127,
  203, 184,  53, 120, 179, 228,  23, 203, 113,
  206, 100, 110, 253,   8,  25, 221, 140,   8,
  141, 225, 189
]
Uint8Array(32) [
  215,  73, 129, 239, 167,  10,  12,
  136,  11, 141, 140,  25, 133, 208,
  117, 219, 203, 246, 121, 185, 154,
   95, 153,  20, 229, 170, 249, 107,
  131,  26, 158,  36
]
```
instead of 
```
Check file:///PathToFile/cryptoTest.ts
ArrayBuffer {}
Uint8Array(32) [
  215,  73, 129, 239, 167,  10,  12,
  136,  11, 141, 140,  25, 133, 208,
  117, 219, 203, 246, 121, 185, 154,
   95, 153,  20, 229, 170, 249, 107,
  131,  26, 158,  36
]
```
**Environment**

- OS: Windows 11 (AMD64) as well as Debian (aarch64)
- deno version: `1.15.2 (release, x86_64-pc-windows-msvc)` as well as `1.15.2 (debug, aarch64-unknown-linux-gnu)`
- std version: both 0.112.0 (Most recent one at  time of posting) and the main branch of this repo ([Last Commit before my testing](https://github.com/denoland/deno_std/commit/fb35d703472c8d38ea816a165e78f417cdf62dc9))

lmx-Hexagram:
I try to work on this issue and finally I find out that `console.log(sthArrayBufferLike)` will always display `ArrayBuffer {}`
[...]

PR #1775, opened 2022-01-01
doc(crypto): change example from print array to uint8Array

lmx-Hexagram:
First thing of #1445


kt3k (approved):
@lmx-Hexagram LGTM. Thanks!

PR #1781, opened 2022-01-02
chore(crypto): export interface for docs

lucacasonato:
This interface was previously not exposed making it invisible in deno
doc.


kt3k (approved):
LGTM

PR #1782, opened 2022-01-03
fix(log): flush to file when buffer is full

### LINKED ISSUE #1093: std/log buffer cut lines
Hi, I don't know if this is the correct behavior, but the log line is cut because the buffer size exceeds as you can see below.
```
{"ts":"2021-06-25T08:36:39.648Z","level":"ERROR","data":"{\"error\":true,\"msg\":\"error\"}"}
{"ts":"2021-06-25T08:36:40.147Z","level":"ERROR","data":"{\"error\":true,\"msg\":\"error\"}"}
{"ts":"2021-06-25T08:36:40.647Z","level":"ERROR","data
```
And when the buffer is full again here is the result. Previous line not completed and the new lines append.
```
{"ts":"2021-06-25T08:18:02.857Z","level":"ERROR","data":"{\"error\":true,\{"ts":"2021-06-25T08:18:17.500Z","level":"ERROR","data":"{\"error\":true,\"msg\":\"error0\"}"}
```

Deno code 
```ts
await log.setup({
  handlers: {
    console: new log.handlers.ConsoleHandler('INFO'),
    file: new log.handlers.RotatingFileHandler('INFO', {
      filename: `./logs/log.log`,
      maxBytes: 5 * 1024 * 1024,
      maxBackupCount: 5,
      formatter: rec => JSON.stringify({ ts: rec.datetime, level: rec.levelName, data: rec.msg }),
    }),
  },
  loggers: {
    default: {
      level: 'INFO',
      handlers: ['console', 'file'],
    },
  },
})

const Logger = log.getLogger()

setInterval(() => Logger.error({ error: true, msg: 'error' }), 500)

```

lmx-Hexagram:
fix #1093


lmx-Hexagram:
Reference to https://github.com/golang/glog/blob/master/glog.go#L815
By check before write to buffer can fix broken line when buffer is nearly full and cannot complete accommodate new log msg.

bartlomieju:
@lmx-Hexagram thanks for the PR, looks like a good fix. Could you add some test case exercising this change?

lmx-Hexagram:
Sorry for so many broken CI. But it is because an new issue.
The new issue is in previous code `addEventListener` and `removeEventListener` won't work as expect.

Here is the detail:
`addEventListener` and `removeEventListener` only work when you use exactly the same eventListener function.
But previous code write:
```ts
#unloadCallback() {
  this.destory()
}
...
addEventListener("unload", this.#unloadCallback.bind(this));
...
removeEventListener("unload",this.#unloadCallback)
```
Sadly, it breaks the rule.

lmx-Hexagram:
fix #1787

dgpgdev:
Thanks a lot for this fix 🙏🤘

lmx-Hexagram:
Hello, can someone review this PR?

log/handlers_test.ts
```diff
@@ -459,6 +498,7 @@ Deno.test({
 
     assertEquals((await Deno.stat(LOG_FILE)).size, 0);
     dispatchEvent(new Event("unload"));
+    dispatchEvent(new Event("unload"));
```
bartlomieju, line 501: Why is another `unload` event needed here?

log/handlers.ts
```diff
@@ -214,7 +217,8 @@ export class RotatingFileHandler extends FileHandler {
       this.#currentFileSize = 0;
     }
 
-    this._buf.writeSync(this._encoder.encode(msg + "\n"));
+    super.log(msg);
```
bartlomieju, line 220: 👍

log/handlers_test.ts
```diff
@@ -129,6 +129,45 @@ Deno.test("testFormatterAsFunction", function (): void {
...
+        super.flush();
+        const decoder = new TextDecoder("utf-8");
+        const data = Deno.readFileSync(LOG_FILE);
+        const text = decoder.decode(data);
+        assertEquals(text.slice(-1), "\n");
+      }
+
+      async destroy() {
+        await super.destroy();
+        Deno.removeSync(LOG_FILE);
```
bartlomieju, line 146: Maybe use `await Deno.remove()` here?

log/handlers_test.ts
lmx-Hexagram: For test #1787, before #1787 std add an event listener in wrong way. If call `dispatchEvent(new Event("unload"));` twice, this test will fail.
I reserve this as an improve this test.

bartlomieju (approved):
LGTM, thank you @lmx-Hexagram

PR #1792, opened 2022-01-07
fix(fs/walk): filter special files

### LINKED ISSUE #1789: walk() fails on socket
**Describe the bug**

Running `walk('/Users/clemens')` does not check for sockets.

> DEBUG RS - deno_runtime::ops::fs:1168 - op_read_dir_async /Users/clemens/.docker/run/docker.sock
> error: Uncaught (in promise) Error: Not a directory (os error 20), readdir '/Users/clemens/.docker/run/docker.sock' for path "/Users/clemens/.docker/run/docker.sock"

MacOS says
```
file /Users/clemens/.docker/run/docker.sock
/Users/clemens/.docker/run/docker.sock: socket
```

**Steps to Reproduce**

On MacOS run

```typescript
import { walk } from "https://deno.land/std@0.120.0/fs/walk.ts";

for await (const entry of walk('/Users/clemens')) {
    console.log(entry.path,entry.isFile);
}
```

The checks in https://github.com/denoland/deno/blob/d44011a69e0674acfa9c59bd7ad7f0523eb61d42/runtime/ops/fs.rs#L1178 seems logic. What is wrong?

**Expected behavior**

Happy scanner the whole tree

**Environment**

- MacOS 11.6.2
- deno 1.17.1 (release, x86_64-apple-darwin)
- v8 9.7.106.15
- typescript 4.5.2
- std@0.120.0

bnoordhuis:
Don't error on special files like UNIX domain sockets. Filter them out
instead.

Fixes #1789.

<hr>

I opted to filter them out because you can't really do anything except unlink them but maybe they should be included instead? Produces entries that look like this though:
```
{
  path: "/foo",
  name: "bar.sock",
  isFile: false,
  isDirectory: false,
  isSymlink: false,
}
```
Feedback appreciated.

### DIFF
--- a/fs/walk_test.ts
+++ b/fs/walk_test.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// deno-lint-ignore-file no-explicit-any
 import { walk, WalkEntry, WalkOptions, walkSync } from "./walk.ts";
 import {
   assert,
@@ -8,8 +9,8 @@ import {
 } from "../testing/asserts.ts";
 
 export function testWalk(
-  setup: (arg0: string) => void | Promise<void>,
-  t: () => void | Promise<void>,
+  setup: (arg0: string) => any | Promise<any>,
+  t: (context: any) => void | Promise<void>,
   ignore = false,
 ): void {
   const name = t.name;
@@ -18,8 +19,8 @@ export function testWalk(
     const d = await Deno.makeTempDir();
     Deno.chdir(d);
     try {
-      await setup(d);
-      await t();
+      const context = await setup(d);
+      await t(context);
     } finally {
       Deno.chdir(origCwd);
       await Deno.remove(d, { recursive: true });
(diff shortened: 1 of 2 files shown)


bnoordhuis:
The linter is failing on something that's definitely not my work. 🤦

lucacasonato:
I would vote for including them. If we don't include them, someone will want to add an option for including them in a few months. Let's just include them now.

bartlomieju:
> The linter is failing on something that's definitely not my work. 🤦

@bnoordhuis merge `main` and it will be fixed.

fs/walk_test.ts
```diff
@@ -283,6 +283,20 @@ testWalk(
   },
 );
 
+// https://github.com/denoland/deno_std/issues/1789
+testWalk(
+  async (d: string) => {
```
clemens-tolboom, line 288: VSCode says 'remove async'

fs/walk_test.ts
```diff
@@ -283,6 +283,20 @@ testWalk(
...
+    return Deno.listen({ path: d + "/a", transport: "unix" });
+  },
+  async function unixSocket(listener: any) {
```
clemens-tolboom, line 291: Let's type this `listener: Deno.Listener`

clemens-tolboom (commented):
This fix the socket.

I'll change my PR into a fix for `mkfifo`

bnoordhuis:
> I would vote for including them.

Done.

kt3k (approved):
LGTM!

PR #1801, opened 2022-01-08
feat(node): upstream caller-path package

bartlomieju:
Temporary workaround while working on getting `tap` to work.

This is required due to incompatible return types from `getFileName()` function on error stack:
- Node return a file path (`/Users/foo/bar.js`)
- Deno returns a URL (`file:///Users/foo/bar.js`)


bartlomieju:
Don't merge yet, this is wrong as "upstream" modules are gonna end up in "built-in Node modules" list and would be setup automatically in eg. REPL with compat mode.

lucacasonato (approved):

PR #1802, opened 2022-01-08
feat(node): mock more APIs

bartlomieju:


ry (approved):

bartlomieju:
Don't merge, this is actually wrong and needs to be reworked, as right now these "upstream" modules will end up in "built-in Node modules" list.

PR #1816, opened 2022-01-12
Change bind6 comment to IPv6

jcs224:
Correctly describe bind6 as connecting to an IPv6 address instead of IPv4.


kt3k (approved):
@jcs224 LGTM. Thanks!

Issue #1316, opened 2021-09-27
Tracking unstable APIs used in std


kt3k:
std/node
- Deno.utime
- Deno.utimeSync
- Deno.futime
- Deno.futimeSync
- Deno.consoleSize
- Deno.systemMemoryInfo
- Deno.loadavg
- Deno.osRelease

std/signal
- Deno.Signal
- Deno.signal
- Deno.kill (test)

bartlomieju:
@kt3k is this issue in any way actionable? It seems to already be stale due to changes in `Deno.Signal` and `Deno.kill` APIs.

kt3k:
Right, this isn't very much actionable. Closing

PR #1819, opened 2022-01-12
fix: Bypass TS errors for missing --unstable

nayeemrmn:
Approach suggested by @dsherret in https://github.com/denoland/deno/issues/12726#issuecomment-966355578. I didn't apply it to e.g. `signal` which glaringly builds on an unstable API (should I?), but modules included in blankets like `fs/copy.ts` and `node`.

Ref #1316, #1539.


nayeemrmn:
image
@bartlomieju A bit of a design flaw with #13324, the module referenced here is one that is created, run and deleted as part of a test.

_deno_unstable.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+export function umask(
+  ...args: Parameters<typeof Deno.umask>
+): ReturnType<typeof Deno.umask> {
+  if (typeof Deno.umask == "function") {
+    return Deno.umask(...args);
+  } else {
+    throw new TypeError("Requires --unstable");
```
lucacasonato, line 121: `Deno.umask requires --unstable` would be a better error message IMO

lucacasonato:
> image @bartlomieju A bit of a design flaw with [denoland/deno#13324](https://github.com/denoland/deno/pull/13324), the module referenced here is one that is created, run and deleted as part of a test.

Agreed. I think we should emit the source code / source maps for all modules we load into the coverage dir when collecting coverage. `deno coverage` should not have to touch DENO_DIR IMO.

bartlomieju:
> > image @bartlomieju A bit of a design flaw with [denoland/deno#13324](https://github.com/denoland/deno/pull/13324), the module referenced here is one that is created, run and deleted as part of a test.
> 
> Agreed. I think we should emit the source code / source maps for all modules we load into the coverage dir when collecting coverage. `deno coverage` should not have to touch DENO_DIR IMO.

That would entail quite a big rewrite to the whole coverage system. I have to give it more thought.

bartlomieju:
In the meantime can we just not delete the file in question?

_deno_unstable.ts
nayeemrmn: I'm not sure, the intended subject in this case is whatever API at the lib surface e.g. `copy()`. I figure as normal we should let the error stack relay this since it's not a static error.

bartlomieju:
@nayeemrmn @lucacasonato thinking about this problem more: after you delete a file its transpiled source and source map are still in `DENO_DIR` (because we never evict it), but the original source code is gone. We never cache local sources in `DENO_DIR` so copying all files (including original source code that is on local disk) would be quite exotic case if we take all `deno` subcommands into accout (as in, we don't copy local sources from their original location to `DENO_DIR`). In this case maybe we shouldn't error but just omit a coverage for that file?

kt3k (approved):
@nayeemrmn @bartlomieju Sorry for the delay in response. I'm in favor of landing this. This reduces the barrier to start using std in my view, and that should be nice for new users.

LGTM

bartlomieju (approved):

PR #1827, opened 2022-01-17
fix(node): work around |this| check in deno core

bnoordhuis:
Deno's built-in timer functions throw "Illegal invocation" exceptions
unless they are called with this == globalThis or this == null.

### DIFF
--- a/node/timers_test.ts
+++ b/node/timers_test.ts
@@ -1,0 +1,42 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+
+import timers from "./timers.ts";
+
+Deno.test("[node/timers setTimeout]", () => {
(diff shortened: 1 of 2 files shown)


node/timers_test.ts
```diff
@@ -0,0 +1,42 @@
...
+
+Deno.test("[node/timers setTimeout]", () => {
+  {
+    const { clearTimeout, setTimeout } = timers;
+    const id = setTimeout(() => {});
+    clearTimeout(id);
+  }
+
+  {
+    const id = timers.setTimeout(() => {});
```
kt3k, line 13: Confirmed this change fixes this case 👍

kt3k (approved):
LGTM

PR #1828, opened 2022-01-17
fix(node/net): setNoDelay and setKeepAlive no-ops

bnoordhuis:
These methods have no directly observable side effects and should
therefore be safe to turn into no-ops. Doing so unblocks a number
of third-party modules.

Can be implemented once https://github.com/denoland/deno/pull/13103
lands.


AaronO (dismissed):
These should return the socket (i.e: `this`), see: https://nodejs.org/api/net.html#socketsetnodelaynodelay

bnoordhuis:
> These should return the socket (i.e: this)

These are the internal handle methods, not the public `net.Socket` methods. Those already return `this`.

bartlomieju (approved):

PR #1817, opened 2022-01-12
feat(node/fs): fs.write()/fs.writeSync()

AaronO:

PR #1832, opened 2022-01-18
fix(node): fix fs.write/fs.writeAll

kt3k:
- Reverted the last commit of https://github.com/denoland/deno_std/pull/1817 as `writeAll` here is not compatible with what `streams/conversion.ts` provides.
- Enabled `test-fs-write-buffer.js` test case.
- Also added `offset` range check, which is tested in the above test suite.


bartlomieju (approved):
LGTM

PR #1833, opened 2022-01-18
chore: deprecate _tools scripts

### LINKED ISSUE #1830: Deprecate "_tools/" scripts
With https://github.com/denoland/deno/pull/13313 landed we no longer need to have these scripts as a config file will be auto discovered. However we shouldn't outright remove them, instead we should check Deno version used to execute these scripts and suggest users on `>=1.18.0` to directly use `deno fmt` and `deno lint`; while for users using `<1.18.0` we should suggest to upgrade to latest version.

kt3k:
closes #1830

### DIFF
--- a/_tools/format.js
+++ b/_tools/format.js
@@ -1,7 +1,14 @@
-#!/usr/bin/env -S deno run --no-check
+#!/usr/bin/env -S deno run --no-check --allow-read --allow-run
 import * as semver from "https://deno.land/x/semver@v1.0.0/mod.ts";
 if (semver.lt(Deno.version.deno, "1.18.0")) {
-  console.log("This script is deprecated. Upgrade to the latest Deno and run `deno fmt` command directly");
+  console.log(
+    "This script is deprecated. Upgrade to the latest Deno and run `deno fmt` command directly",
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Thank you :)

_tools/format.js
```diff
@@ -1,6 +1,7 @@
-#!/usr/bin/env -S deno run --allow-read --allow-run
-const p = Deno.run({
-  cmd: [Deno.execPath(), "fmt", "--config", "deno.json", ...Deno.args],
-});
-const status = await p.status();
-Deno.exit(status.code);
+#!/usr/bin/env -S deno run --no-check
+import * as semver from "https://deno.land/x/semver@v1.0.0/mod.ts";
+if (semver.lt(Deno.version.deno, "1.18.0")) {
+  console.log("This script is deprecated. Upgrade to the latest Deno and run `deno fmt` command directly");
```
bartlomieju, line 4: Although, maybe we should actually keep the script running instead of doing nothing here?

kt3k: Ah, ok. Let's do it

PR #1834, opened 2022-01-19
feat(node): provide node compatible timer APIs in commonjs wrapper

kt3k:
This PR provides node-compatible timer APIs in commonJS wrapper function. This works around the error `TypeError: this.peakMemoryInterval.unref is not a function` while executing `yarn` in node compatible mode. ref: https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/reporters/base-reporter.js#L169

part of denoland/deno#13420


node/module.ts
```diff
@@ -200,9 +206,10 @@ class Module {
   static _cache: { [key: string]: Module } = Object.create(null);
   static _pathCache = Object.create(null);
   static globalPaths: string[] = [];
-  // Proxy related code removed.
   static wrapper = [
-    "(function (exports, require, module, __filename, __dirname) { ",
+    // We provide non standard timer APIs in the CommonJS wrapper
+    // to avoid exposing them in global namespace.
+    "(function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) { ",
```
bnoordhuis, line 212: I hope this doesn't break any third-party code...

aside: in the future we should use `v8::script_compiler::compile_function_in_context()` but that's not currently exposed by deno.

bnoordhuis (approved):

bartlomieju (approved):
It's an improvement but that would still make packages using ESM modules that use global APIs break. However since I don't know of any packages that use ESM and depend on globals let's just merge it.

node/module.ts
bartlomieju: We'll handle that when rewriting all of this in Rust

PR #1841, opened 2022-01-24
refactor(node): move _errors to internal/errors

kt3k:
continued from #1722 (thanks @getspooky for the initial work), part of #1707

This PR moves `node/_errors.ts` to `node/internal/errors.ts`.

The motivation of this change is to follow the directory layout of Node.js repository more closely.


bartlomieju (approved):
LGTM

PR #1843, opened 2022-01-25
fix(testing): `assertObjectMatch` matches ArrayBuffer views correctly

### LINKED ISSUE #1842: `assertObjectMatch` does not correctly match `Uint8Array`
**Describe the bug**

When comparing `Uint8Array` values within an object, `assertObjectMatch` throws an `AssertionError` claiming that the `Uint8Array` fields are not equal.

**Steps to Reproduce**

Run the following test
```js
import { assertObjectMatch } from "https://deno.land/std@0.122.0/testing/asserts.ts";

Deno.test("should pass comparing Uint8Arrays", async () => {
  const body = new Uint8Array();

  function foo() {
    return { body, foo: "bar", fizz: 'buzz' };
  }

  assertObjectMatch(foo(), { body, foo: 'bar' });
});
```

which should log, failing the test:
```sh
    [Diff] Actual / Expected


    {
+     body: Uint8Array(0) [
+       ,
+     ],
-     body: {},
      foo: "bar",
    }

    at assertEquals (https://deno.land/std@0.122.0/testing/asserts.ts:269:9)
    at assertObjectMatch (https://deno.land/std@0.122.0/testing/asserts.ts:591:10)
    ...
```

**Expected behavior**

The assertion to pass successfully, since the expected value is a subset object of the actual object

**Environment**

- OS: macOS 11.2.3
- deno version: 1.18.0
- std version: 0.122.0

ah-yu:
Closes #1842

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -540,53 +540,51 @@ export function assertObjectMatch(
   expected: Record<PropertyKey, unknown>,
 ): void {
   type loose = Record<PropertyKey, unknown>;
-  const seen = new WeakMap();
-  function filter(a: loose, b: loose): loose {
-    // If the actual value is an array, let assertEquals do the assertion.
-    if (Array.isArray(a) || ArrayBuffer.isView(a)) {
-      return a;
-    }
-
-    // Prevent infinite loop with circular references with same filter
-    if ((seen.has(a)) && (seen.get(a) === b)) {
-      return a;
-    }
-    seen.set(a, b);
-    // Filter keys and symbols which are present in both actual and expected
-    const filtered = {} as loose;
-    const entries = [
-      ...Object.getOwnPropertyNames(a),
-      ...Object.getOwnPropertySymbols(a),
-    ]
-      .filter((key) => key in b)
-      .map((key) => [key, a[key as string]]) as Array<[string, unknown]>;
-    for (const [key, value] of entries) {
-      // On array references, build a filtered array and filter nested objects inside
-      if (Array.isArray(value)) {
-        const subset = (b as loose)[key];
-        if (Array.isArray(subset)) {
-          filtered[key] = value
-            .slice(0, subset.length)
-            .map((element, index) => {
-              const subsetElement = subset[index];
-              if ((typeof subsetElement === "object") && (subsetElement)) {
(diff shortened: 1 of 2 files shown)


testing/asserts_test.ts
```diff
@@ -601,6 +603,8 @@ Deno.test("testingAssertObjectMatching", function (): void {
     assertObjectMatch({ test: { a: 1 } }, { test: new A(1) });
     assertObjectMatch({ test: new A(1) }, { test: new A(1) });
   }
+  // subset with TypedArray instance inside
+  assertObjectMatch(m, { bar: new Uint8Array([1, 2]) });
```
kt3k, line 607: The fix looks good to me, but this test case seems passing even without the change.

If I changed the assertion to `assertObjectMatch(m, m)`, then the issue described in #1842 reproduces as a test case. So maybe let's add that assertion here?

kt3k (commented):
@ah-yu Thanks for the PR. Left one comment

ah-yu:
@kt3k Thanks for review! I think I made a bad fix before.
This bug caused by using same `weakmap` between `actual` computes and `expected` computes.
```js
const foo = new Map(); // Array/TypedArray/etc will cause the same problem
assertObjectMatch({ foo, bar:"bar"},{foo}); // fails 
```

testing/asserts.ts
```diff
@@ -540,53 +540,51 @@ export function assertObjectMatch(
   expected: Record<PropertyKey, unknown>,
 ): void {
   type loose = Record<PropertyKey, unknown>;
-  const seen = new WeakMap();
-  function filter(a: loose, b: loose): loose {
-    // If the actual value is an array, let assertEquals do the assertion.
-    if (Array.isArray(a)) {
```
ah-yu, line 546: Maybe we could remove this?

ah-yu:
Also fix https://github.com/denoland/deno_std/issues/1288 
Seems that https://github.com/denoland/deno_std/pull/1289 is not a right fix.

ah-yu:
Fixes https://github.com/denoland/deno_std/pull/1289#issuecomment-922692148

kt3k (approved):
@ah-yu LGTM. Thanks!

PR #1844, opened 2022-01-25
chore: reduce circular dependencies under std/node and check it in CI

### LINKED ISSUE #1609: node: Add a test case checking for circular imports in std/node
We started hitting following errors:
```
error: Uncaught ReferenceError: Cannot access 'EventEmitter' before initialization
class Stream extends EventEmitter {
                     ^
    at file:///Users/biwanczuk/dev/deno_std/node/_stream/stream.ts:14:22
```
#1608

```
error: Uncaught ReferenceError: Cannot access 'Writable' before initialization
  const stream = new Writable({
                 ^
    at createWritableStdioStream (file:///Users/biwanczuk/dev/deno_std/node/process.ts:151:18)
    at file:///Users/biwanczuk/dev/deno_std/node/process.ts:201:23
```
#1585

These are most likely caused by circular dependencies. Node's runtime code is written in CJS format and there are several "lazy loads" of modules, eg. 
```
if (something) {
  const fs = require("fs")
}
```
Right now we are reworking them to be static imports at the top of the file, but that might lead to strange errors like the ones shown above. I believe a test case that would catch circular imports might help solve this issue and help us organize the code better.

CC @bnoordhuis @Soremwar

kt3k:
closes #1609

This PR reduces the circular module dependencies under `std/node`, and also checks if there's any circular deps in there in CI. If there's any circular dependencies in `std/node`, now CI fails.


node/_process/stdio.js
```diff
@@ -0,0 +1,7 @@
+// Copyright 2018-2021 the Deno authors. All rights reserved. MIT license.
+// Copyright Joyent, Inc. and Node.js contributors. All rights reserved. MIT license.
+
+// Lazily initializes the actual stdio objects.
+// This trick is necessary for avoiding circular dependencies between
+// stream and process modules.
+export const stdio = {};
```
bartlomieju, line 7: 👍 nice, this is exactly the situation we wanted to catch

bartlomieju (approved):
Great work @kt3k, thank you for looking into it. LGTM.

One question I have is if "node/internal/streams/stdio.js" is required or just something that was left behind by accident?

kt3k:
> One question I have is if "node/internal/streams/stdio.js" is required or just something that was left behind by accident?

Ah, that was a mistake! Thanks for catching it

PR #1846, opened 2022-01-25
feat(node): add os.networkInterfaces

kt3k:
This PR adds `os.networkInterfaces()` API in std/node.

part of denoland/deno#13420


bartlomieju (approved):
LGTM, nice work

PR #1849, opened 2022-01-26
fix(node/process): can not pass exit code as a string

### LINKED ISSUE #1845: node: process.exit should accept string as parameter
Currently `process.exit` in std/node only accepts number or undefined as the argument, but it should also accept string type.

https://github.com/denoland/deno_std/blob/2b84a2be4a4d912ed65467962fd767a31e09e5f3/node/process.ts#L58

Nautigsam:
Fixes #1845


kt3k (approved):
@Nautigsam LGTM. Nice work. Thanks!

bartlomieju:
Nice work @Nautigsam!

PR #1850, opened 2022-01-26
chore(node): remove cli.ts

kt3k:
Now `deno run --compat --unstable` works better as a CLI entry point of node compatibility mode. So now we can remove this version of entry point.


bartlomieju (approved):
Sweet!

PR #1851, opened 2022-01-26
feat(node): add url.resolve

kt3k:
Continued from #1692 (Thanks @getspooky for the initial work)

This PR adds `url.resolve` to std/node.

part of https://github.com/denoland/deno/issues/13420


bartlomieju (approved):
LGTM, nice work!

Issue #1658, opened 2021-11-30
Modernize `std/archive/tar`


lucacasonato:
The current `std/archive/tar` builds on `Deno.Reader` and friends. We should rewrite it to use WHATWG streams. So a `TarDecoderStream` and `TarEncoderStream`.

PR #1855, opened 2022-01-27
feat: archive use WHATWG stream

gliheng:
This is an initial version that implements WHATWG archive stream referred in this issue https://github.com/denoland/deno_std/issues/1658  
A `TarEncoderStream` and `TarDecoderStream` is provided.
I'll add more tests this weekend.

I don't know if this stream version should replace the original Tar Untar class or not. So it's kept in separate file for now.


crowlKats:
It shouldnt use the Deno namespace whatsoever

gliheng:
File's content added to TarEncoderStream is still a Reader. Do you mean that need to change into a stream as well?

crowlKats:
@gliheng yes. There should be no reference to the Deno namespace whatsoever.

PR #1856, opened 2022-01-27
fix(node/http): fix http.request

kt3k:
This PR fixes `this.cb is not a function` error when you call `http.request` without callback function.

Also this change ignores `connection closed before message completed` error while making request. Node.js seems ignoring this error. With this change, we can pass the compatiblity test cases `test-http-url.parse-*.js`

Resolves a part of yarn blocker issues https://github.com/denoland/deno/issues/13420


bartlomieju (approved):
🚀

PR #1857, opened 2022-01-28
feat(node): add punycode module

kt3k:
This PR adds "punycode" module to std/node.

part of yarn locker issues https://github.com/denoland/deno/issues/13420


kt3k:
punycode module seems needing to print deprecation messages only when it's `require()`d to pass the test cases.

node/module_all.ts
```diff
@@ -118,6 +119,15 @@ export default {
   path,
   "perf_hooks": perfHooks,
   process,
+  get punycode() {
+    process.emitWarning(
```
kt3k, line 123: Lazily prints the deprecation warning because this message should be shown only when the punycode module is required.

node/module.ts
```diff
@@ -787,25 +784,30 @@ const m = {
...
+  if (nativeModulePolyfill.has(request)) {
+    return nativeModulePolyfill.get(request);
+  }
+  const mod = nodeMods[request];
+  if (mod) {
+    const nodeMod = createNativeModule(request, mod);
+    nativeModulePolyfill.set(request, nodeMod);
+    return nodeMod;
+  }
+  return undefined;
```
bartlomieju, line 804: Why are these changes necessary?

kt3k: The current code calls `createNativeModule(request, mod)` for all modules at the start up time. That causes the above deprecation message showing always even when `punycode` is not required by the user's script.

With this change, the above `get punycode() { process.emitWarning(...) ... }` part is called only when the punycode module is actually required by the user's script. This behavior is necessary for passing warning message expectations (`common.expectWarning(...)` calls) in some of compatibility tests e.g. `test-dns-lookup.js`, `test-fs-rmdir-recursive.js`, etc

bartlomieju: I see, this change seems quite invasive and might have surprising side-effects due to lazy evaluation. Could you please verify that all tests related to compat mode work in CLI with this PR applied?

kt3k: With the below diff applied, `cargo test compat` passed without an error. So I think this change doesn't break the basic feature of compat mode:

```diff
diff --git a/cli/compat/esm_resolver.rs b/cli/compat/esm_resolver.rs
index 300b76f31..fa883eddb 100644
--- a/cli/compat/esm_resolver.rs
+++ b/cli/compat/esm_resolver.rs
@@ -1218,7 +1218,7 @@ mod tests {
     let cwd = testdir("basic");
     let main = Url::from_file_path(cwd.join("main.js")).unwrap();
     let expected =
-      Url::parse("https://deno.land/std@0.123.0/node/http.ts").unwrap();
+      Url::parse("https://raw.githubusercontent.com/kt3k/deno_std/feat/node-punycode/node/http.ts").unwrap();
 
     let actual = node_resolve("http", main.as_str(), &cwd).unwrap();
     assert!(matches!(actual, ResolveResponse::Esm(_)));
diff --git a/cli/compat/mod.rs b/cli/compat/mod.rs
index abfffb2d2..5fc10ce38 100644
--- a/cli/compat/mod.rs
+++ b/cli/compat/mod.rs
@@ -16,7 +16,7 @@ pub(crate) use esm_resolver::NodeEsmResolver;
 // each release, a better mechanism is preferable, but it's a quick and dirty
 // solution to avoid printing `X-Deno-Warning` headers when the compat layer is
 // downloaded
-static STD_URL_STR: &str = "https://deno.land/std@0.123.0/";
+static STD_URL_STR: &str = "https://raw.githubusercontent.com/kt3k/deno_std/feat/node-punycode/";
 
 static SUPPORTED_MODULES: &[&str] = &[
   "assert",
```

> this change seems quite invasive and might have surprising side-effects due to lazy evaluation.

In my view, this change doesn't make that much effect because the commonjs 'module's are now created lazily, but the underlying ES modules are still eagerly evaluated (in `node/modules_all.ts`)

bartlomieju (approved):
LGTM

PR #1858, opened 2022-01-28
fix(node): mock response.socket object

kt3k:
This PR adds `socket` property to IncomingMessage (http response). This property is used for checking the http connection status in some libraries. e.g. https://github.com/request/request/blob/3c0cddc7c8eb60b470e9519da85896ed7ee0081e/request.js#L944-L945

part of yarn blocker issues https://github.com/denoland/deno/issues/13420


bartlomieju (approved):
🦕

PR #1864, opened 2022-01-29
feat(node/child_process): add support for AbortSignal in ChildProcess

### LINKED ISSUE #1848: node: ChildProcess should support AbortSignal
In class `ChildProcess`, we can pass `signal` among the options but we don't use it.
https://github.com/denoland/deno_std/blob/d0232ce805a159555dcbc6a7aca4a2beea54a5ee/node/internal/child_process.ts#L300-L303
We must support this option in `spawn` function and all its derivatives.

Nautigsam:
Fixes #1848


node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
 files, since that may cover up inconsistencies between the node library and
 actual node behavior.
 
+### Working with child processes ? Think about DENO_NODE_COMPAT_URL
```
bartlomieju, line 173:
```suggestion
### Working with child processes? Use `DENO_NODE_COMPAT_URL`
```

node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
...
+
+When working with `child_process` modules, you will have to run tests pulled
+from Node.js. These tests usually spawn deno child processes via the use of
+`process.execPath`. The deno executable will use its own embedded version of std
```
bartlomieju, line 177:
```suggestion
`process.execPath`. The `deno` executable will use its own embedded version of std
```

node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
...
+modules, then you may get the impression your code is not really working as it
+should.
+
+To prevent this, set DENO_NODE_COMPAT_URL with the absolute path to your
```
bartlomieju, line 181:
```suggestion
To prevent this, set `DENO_NODE_COMPAT_URL` with the absolute path to your
```

node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
...
+deno_std repo:
```
bartlomieju, line 182:
```suggestion
`deno_std` repo, ending with a trailing slash:
```

node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
...
+
+```
+export DENO_NODE_COMPAT_URL=$PWD/
```
bartlomieju, line 185:
```suggestion
export DENO_NODE_COMPAT_URL=$PWD/
# or
export DENO_NODE_COMPAT_URL=file:///path/to/deno_std/dir/
```

node/README.md
```diff
@@ -170,6 +170,23 @@ to pass certain tests. However, avoid doing such manual changes to the test
...
+```
+
+Then, deno will use your std code.
```
bartlomieju, line 188:
```suggestion
Then, `deno` will use local copy of `deno_std` instead of latest version.
```

Nautigsam: IMHO I would be better like this, what do you think?
```suggestion
Then, `deno` will use your local copy of `deno_std` instead of latest version.
```

bartlomieju: Sure, sounds good

node/_tools/suites/parallel/test-child-process-execfile.js
```diff
@@ -94,9 +93,10 @@ const denoOptions = [
     assert.strictEqual(err.name, 'AbortError');
     assert.strictEqual(err.signal, undefined);
   });
-  execFile(process.execPath, [echoFixture, 0], { signal }, check);
+  execFile(process.execPath, [...denoOptions, echoFixture, 0], { signal }, check);
 }
 
+/* TODO(Nautigsam) This test requires to check the type but isn't this already covered by tsc?
```
kt3k, line 99: > isn't this already covered by tsc?

That's true when user import it in typescript context, but node.js generally doesn't assume the user use typescript. So they check most parameter types at runtime

kt3k (approved):
@Nautigsam LGTM. Nice work!

PR #1866, opened 2022-01-31
chore: set compat env var automatically on testing

kt3k:
This PR sets `DENO_NODE_COMPAT_URL` env var to the local std root while node.js compat testing. So the contributors can simply run `deno test` without setting env vars.

### DIFF
--- a/node/_tools/test.ts
+++ b/node/_tools/test.ts
@@ -16,6 +16,7 @@ const filters = Deno.args;
  */
 
 const toolsPath = dirname(fromFileUrl(import.meta.url));
+const stdRootUrl = new URL("../../", import.meta.url).href;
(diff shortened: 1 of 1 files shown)


node/_tools/test.ts
```diff
@@ -70,7 +74,7 @@ for await (const path of testPaths) {
         console.log(`Error: "${path}" failed`);
         console.log(
           "You can repeat only this test with the command:",
-          magenta(cmd.join(" ")),
+          magenta(`deno test -A node/_tools/test.ts -- ${path}`),
```
kt3k, line 77: This shows the message like: `You can repeat only this test with the command: deno test -A node/_tools/test.ts -- parallel/test-child-process-execfile.js`

bartlomieju (approved):
LGTM

PR #1874, opened 2022-02-03
fix(node): fix fs.createWriteStream

### LINKED ISSUE #1873: node: fs.createWriteStream can't create new file
The following node.js script should create a new file `foo.txt` with the content `hi`, but our current `fs.createWriteStream` implementation can't create it.

```js
const fs = require("fs")
const write = fs.createWriteStream("foo.txt")
write.write("hi");
```

kt3k:
This PR fixes misc bugs of `fs.createWriteStream`

- Adds correct `flags` and `mode` defaults. Closes #1873 
- Passes stream options to parent `Writable` class. e.g. highWaterMark
- Adds some more compat tests.


bartlomieju (approved):
LGTM, nice work

PR #1877, opened 2022-02-03
feat(dotenv): add dotenv library

pietvanzoen:
Per discussion in https://github.com/denoland/deno/issues/13396, I'd like to add https://github.com/pietvanzoen/deno-dotenv to the Deno_std repo.

I'm not really sure how this should be setup. I've removed most of the cruft from my repo that didn't make sense here. 

I'm not sure how to integrate the tests here with the existing ones in this repo.

Advice appreciated. :)


dotenv/util.ts
```diff
@@ -0,0 +1,17 @@
+// deno-lint-ignore no-explicit-any
+export function removeEmptyValues(
+  obj: Record<string, any>,
+): Record<string, any> {
```
crowlKats, line 4: why `any` over `unknown`?

dotenv/README.md
```diff
@@ -0,0 +1,177 @@
...
+```sh
+# .env
+GREETING=hello world
+```
+
+Then import the configuration using the `config` function.
+
+```ts
+// app.ts
+import { config } from "https://deno.land/x/dotenv/mod.ts";
```
crowlKats, line 18: update the import

dotenv/README.md
```diff
@@ -0,0 +1,177 @@
...
+the process environment:
+
+```sh
+# .env
+GREETING=hello world
+```
+
+```ts
+// app.ts
+import "https://deno.land/x/dotenv/load.ts";
```
crowlKats, line 60: ditto

dotenv/README.md
```diff
@@ -0,0 +1,177 @@
...
+# .env.example
+GREETING=
+export LOG_PATH="provide a value here"
+```
+
+Then import the configuration with `safe` option set to `true`.
+
+```ts
+// app.ts
+import { config } from "https://deno.land/x/dotenv/mod.ts";
```
crowlKats, line 84: same here

dotenv/Makefile
```diff
@@ -0,0 +1,14 @@
...
+	curl -fsSL https://deno.land/x/install/install.sh | sh
+
+test:
+	deno test --reload --allow-env --allow-read
+	deno fmt --check
+
+fmt:
+	deno fmt
+
+.PHONY: test fmt install
```
crowlKats, line 14: no need for this makefile

dotenv/README.md
```diff
@@ -0,0 +1,177 @@
+# Dotenv [![Build Status](https://travis-ci.com/pietvanzoen/deno-dotenv.svg?branch=master)](https://travis-ci.com/pietvanzoen/deno-dotenv) [![GitHub tag (latest SemVer)](https://img.shields.io/github/v/tag/pietvanzoen/deno-dotenv)](https://github.com/pietvanzoen/deno-dotenv/releases)
```
crowlKats, line 1: remove these

dotenv/Makefile
crowlKats, line 14: no need for this makefile

crowlKats (changes requested):

GerbenRampaart:
Thanks @pietvanzoen for your effort and thanks @crowlKats for the review

dotenv/.env
```diff
@@ -0,0 +1,2 @@
+GREETING=hello world
+DO_NOT_OVERRIDE=overridden
```
bartlomieju, line 2: I suggest to move all of these files into `dotenv/testdata/` directory

dotenv/util.ts
```diff
@@ -0,0 +1,17 @@
+// deno-lint-ignore no-explicit-any
```
bartlomieju, line 1: Please add copyright headers

dotenv/.env
pietvanzoen: I've moved most of the test files to `./testdata`. I left `.env` and `.env.defaults` as they are used to test the default behaviour.

dotenv/test_deps.ts
```diff
@@ -0,0 +1,5 @@
+export {
+  assertEquals,
+  assertRejects,
+  assertThrows,
+} from "https://deno.land/std@0.112.0/testing/asserts.ts";
```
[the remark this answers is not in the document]
pietvanzoen: updated

dotenv/README.md
pietvanzoen: updated

dotenv/util.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+// deno-lint-ignore no-explicit-any
```
crowlKats, line 3:
```suggestion
```

crowlKats, line 3:
```suggestion

```

bartlomieju:
> Does anyone have tips for testing `./load.ts`? Currently it's failing in CI since the tests run from the root directory, then load is expecting a `.env` in the root. Putting `.env` in the root doesn't seem like a good solution.

In your `Deno.test()` you want to spawn a subprocess for that purpose - you'll be able to set a CWD et al, see https://github.com/denoland/deno_std/blob/3239e1b7999bc7dc1fafd27813d6e56acb33bcf3/node/process_test.ts#L118-L152 for example

dotenv/README.md
```diff
@@ -0,0 +1,171 @@
...
+```
+
+### Safe Mode
+
+To enable safe mode, create a `.env.example` file in the root of the project.
+
+```sh
+# .env.example
+GREETING=
+export LOG_PATH="provide a value here"
```
bartlomieju, line 75: Is this correct? I've never seen `export` in dotenv file before. What's the reason for it?

dotenv/README.md
```diff
@@ -0,0 +1,171 @@
...
+// app.ts
+import { config } from "https://deno.land/std@$STD_VERSION/dotenv/mod.ts";
+
+console.log(config());
+```
+
+Then run your app.
+
+```
+> deno app.ts
```
bartlomieju, line 24: This doesn't seem correct - missing `run` subcommand and it should require `--allow-read` flag

dotenv/README.md
```diff
@@ -0,0 +1,171 @@
...
+
+```ts
+// app.ts
+import "https://deno.land/std@$STD_VERSION/dotenv/load.ts";
+
+console.log(Deno.env.get("GREETING"));
+```
+
+```
+> deno --allow-env --allow-read app.ts
```
bartlomieju, line 64: Missing `run` subcommand

dotenv/mod.ts
```diff
@@ -0,0 +1,234 @@
...
+const defaultConfigOptions = {
+  path: `.env`,
+  export: false,
+  safe: false,
+  example: `.env.example`,
+  allowEmptyValues: false,
+  defaults: `.env.defaults`,
+};
+
+export function config(options: ConfigOptions = {}): DotenvConfig {
```
bartlomieju, line 59: I would suggest to flip the naming - `configSync`/`config` instead of `config` and `configAsync` - this would match most of built-in `Deno` APIs

bartlomieju (commented):
@pietvanzoen looks great, I got a few nitpicks. Let me know if you need help rewriting tests to use subprocess. 

Also I want to ask again if it's possible to move `.env` and `.env.example` files to `testdata/` directory (especially if tests are gonna spawn subprocesses - you can set `cwd` for each test)

timreichen:
There is a bit much magic going on for my taste. I think the functionality can be structured a bit better.
Imo the process should contain the following separate steps to be as flexible as possible and not to bloat a single function:
- reading files
- parsing and verifying
- set Deno.env

### reading files
Reading files should not be part of the mod imo. I know this is done for convenience but doing it outside will prevent problems with magic filepath resolutions for subdirs and makes an async version of the mod obsolete.

### paring and verifying
The parse function should verify the data with an `allowEmptyValues` option as well as `requiredKeys` and `requiredExports` (which can be from a .env.example file):
```ts
export function parse(dotEnvObject: DotEnvObject, { allowEmptyValues = false, requiredKeys, requiredExports }: {
    allowEmptyValues?: boolean;
    requiredKeys?: string[];
    requiredExports?: string[];
  } = {}) {
...
}
```

### set Deno.env
This would be as simple as
```ts
 Object.entries(object).forEach(([key, value]) => Deno.env.set(key, value));
 ```
 and could be abstracted into an `assign` function, similar to `Object.assign`:
 ```ts
export function assign(
  env: { set(key: string, value: string): void },
  object: Record<string, string>,
) {
  Object.entries(object).forEach(([key, value]) => env.set(key, value));
}
```

So the whole process would look like this:

Simple:
```ts
import { parse, assign } from "https://deno.land/std/dotenv/mod.ts"
const source = await Deno.readTextFile("path/to/.env")
const object = parse(source)
assign(Deno.env, object)
```
With example:
```ts
import { parse, assign } from "https://deno.land/std/dotenv/mod.ts"
const exampleSource = await Deno.readTextFile("path/to/.env.example")
const { env, exports } = parse(exampleSource)
const source = await Deno.readTextFile("path/to/.env")
const object = parse(source, { requiredKeys: Object.keys(env), requiredExports: exports })
assign(Deno.env, object)
```

So I think it would be worth to split up these steps to make it more clear what is happening and also 
though one function might be more convenient but less flexible.

Any thoughts on this?

kt3k:
@timreichen Thanks for taking a look, but does that suggestion solve any specific problem? If not, I think we should avoid the structural change at this point. Maybe let's do it after this first pass landed

timreichen:
> @timreichen Thanks for taking a look, but does that suggestion solve any specific problem? If not, I think we should avoid the structural change at this point. Maybe let's do it after this first pass landed

It does not solve a problem in a sense of functionality, but it would have a range of positive effects on that mod:
- less and cleaner code
- better code alignment to web apis and std in general
- less opinionated
- avoiding changing a global state (`Deno.env`) behind the scenes. I think it is better to actually pass the object (like expressed with `assign(Deno.env, object)`) so a change is expected.

bartlomieju:
@timreichen this is good feedback, and I'd be happy to expose lower level parts of the API for advanced use cases. 

However looking at this example:
```ts
import { parse, assign } from "https://deno.land/std/dotenv/mod.ts"
assign(Deno.env, parse(await Deno.readTextFile("path/to/.env")))
```

I think it's needlessly complicated if I just want to load a dotenv file. In most of the cases users just want to load the file and populate environmental variables with the contents of the file and don't care about nitty-gritty details of how it's implemented.

I agree with @kt3k opinion that we should iterate on the solution after landing the initial version. Would you mind opening a PR with your alternative proposal?

timreichen:
> @timreichen this is good feedback, and I'd be happy to expose lower level parts of the API for advanced use cases.
> 
> However looking at this example:
> 
> ```ts
> import { parse, assign } from "https://deno.land/std/dotenv/mod.ts"
> assign(Deno.env, parse(await Deno.readTextFile("path/to/.env")))
> ```
> 
> I think it's needlessly complicated if I just want to load a dotenv file. In most of the cases users just want to load the file and populate environmental variables with the contents of the file and don't care about nitty-gritty details of how it's implemented.

That might be true. I think the mod should expose the core functionality of parsing and verifying, maybe even an additional `stringify` to be able to generate .env files to complete the parse/stringify pair as in JSON.
I like the idea in this proposal of `load` and `loadSync` in a separate load.ts file like in this proposal for convenience, as we have file_server.ts in std/http for example.
Feel free to have a look at the PR https://github.com/denoland/deno_std/pull/1937 for more detail.

dotenv/README.md
pietvanzoen: It's a convenience feature that was recently added by a contributor of my original library (https://github.com/pietvanzoen/deno-dotenv/pull/61).

It allows using a `.env` file via this lib and via `source .env`.

pietvanzoen:
> @pietvanzoen looks great, I got a few nitpicks. Let me know if you need help rewriting tests to use subprocess.
> 
> Also I want to ask again if it's possible to move `.env` and `.env.example` files to `testdata/` directory (especially if tests are gonna spawn subprocesses - you can set `cwd` for each test)

Ok... got everything into `./testdata`. :)

dotenv/README.md
pietvanzoen: Hmmmm.... I didn't realise it excludes non-exported variables. :/ that's not ideal.

pietvanzoen: Ok... I've removed this functionality.

pietvanzoen:
@bartlomieju All tests updated and working. Removed the exports behaviour to match with `npm dotenv` behaviour more.

All tests/checks are passing. Going to rebase and push one more time and I think we're mostly good. :)

kt3k (approved):
@pietvanzoen LGTM. Nice work!

pietvanzoen:
> @pietvanzoen LGTM. Nice work!

Thanks @kt3k! I don't have permissions to merge unfortunately. 😛

pietvanzoen:
@kt3k thanks!

GerbenRampaart:
> > @timreichen Thanks for taking a look, but does that suggestion solve any specific problem? If not, I think we should avoid the structural change at this point. Maybe let's do it after this first pass landed
> 
> It does not solve a problem in a sense of functionality, but it would have a range of positive effects on that mod:
> 
> * less and cleaner code
> * better code alignment to web apis and std in general
> * less opinionated
> * avoiding changing a global state (`Deno.env`) behind the scenes. I think it is better to actually pass the object (like expressed with `assign(Deno.env, object)`) so a change is expected.
> 
> Imo it would be worth to spend some more time and polish it before the first pass.
> 
> Needs more tests and a better readme but here the branch as an idea: https://github.com/timreichen/deno_std/tree/dotenv/dotenv

I agree with all of your points, though the exact point of dotenv (also in the nodejs world) has always been to introduce a tiny bit of magic/opinionation about how to treat config files. 

A user can still do it any way they like. They can have their own config.json, read that, parse that, use object.assign and achieve the same goal.

I really like that this is added to std because whenever I see a .env in a node project I know exactly whats going on and I feel Deno is also better with this (slight) opinionated design.

Frankly, all those problems I've encountered in node with linting, formatting, transpiling ts, jest not working with esm, and so on, I love any (sane) opinionated choice a std library makes to make me more productive.

timreichen:
> I agree with all of your points, though the exact point of dotenv (also in the nodejs world) has always been to introduce a tiny bit of magic/opinionation about how to treat config files.

As I wrote in a comment in the alternative dotenv proposal:
> I think every std mod has the potential of three abstraction layers:

> - access to low level API (in this case parse)
> - abstraction for convenience but with the possibility to set options (like load(env, options) in this proposal)
> - abstraction of the most important use case without options (what you are describing with import load.ts)
> Imo access to low level API is a must for std, because it is the most flexible, though not the most convenient.
For the other two abstraction layers we can discuss what is the most sensible. I guess I could change load.ts so no fn call and is needed and we suggest to handle all non-standard use cases manually (by parsing etc)?

I think it is possible to achieve both low level API access as well as convenience for users. A bit of magic is fine for some use cases as long as there is a non-magic way to do the same thing:)
Maybe we should continue the discussion there (https://github.com/denoland/deno_std/pull/1937) as this PR is merged.

PR #1881, opened 2022-02-05
chore: Add `getUid` to `_deno_unstable.ts`

uki00a:
`Deno.getUid()` is still an unstable API, so this PR adds it to `_deno_unstable.ts`

Related to #1819


bartlomieju (approved):
LGTM, thanks!

PR #1883, opened 2022-02-06
docs(path/README): Fix missed syntax highlighting language in code block

dcdunkan:
Isn't it supposed to be a TypeScript code block?

- Added TypeScipt syntax highlighting for the missed code block.


kt3k (approved):
@dcdunkan LGTM. Thanks!

PR #1885, opened 2022-02-07
fix(node/buffer): fix base64 decode

### LINKED ISSUE #1875: node: Buffer.from(str, "base64") should ignore data after invalid characters
Current implementation of `Buffer.from(str, "base64")` in std/node throws with TypeError when the str includes invalid character, but node.js ignores data after invalid character.

[ssri](https://www.npmjs.com/package/ssri) (which is in dependencies of yarn) seems depending on the above feature of `Buffer.from`. So we need to make this behavior aligned to node.js to make ssri (and yarn) working

kt3k:
This PR fixes `Buffer.from(str, "base64")` call in std/node. Node.js ignores invalid chars in str, and also it considers padding characters `=` as the end of input.

part of yarn blocker issues https://github.com/denoland/deno/issues/13420

closes #1875


bartlomieju (approved):
LGTM

PR #1888, opened 2022-02-07
fix(node): support 6-arg fs.write

### LINKED ISSUE #1887: node: tar-fs extract doesn't work correctly
When extracting tar using `tar-fs` extract method, the items after the 2nd entry event fail to be read correctly.

This causes corrupted files while extracting tarballs from registries.

part of https://github.com/denoland/deno/issues/13420

kt3k:
closes #1887 

This PR adds support of 6-arg call of `fs.write` (and 5-arg call of `fs.writeSync`)

remaining tasks:
- [x] fix failing test case in `test-fs-write-buffer.js`


bartlomieju (approved):
LGTM

kt3k:
Sorry there was a mistake in the last fix (https://github.com/denoland/deno_std/pull/1888/commits/ae56a9b202714e8d42a874821b629f427b192325) in this PR. That fixed the test case, but it broke other cases which was needed by `yarn add <mod>` command..

I'll address it in another PR

PR #1902, opened 2022-02-11
fix(node/module): fix commonjs wrapper

### LINKED ISSUE #1901: node: providing timers in commonjs wrapper conflicts with the declaration of timers in modules
If an npm module has the following line:

```
const { setInterval } = require('node:timers');
```

This throws a SyntaxError: `error: Uncaught SyntaxError: Identifier 'setInterval' has already been declared`

kt3k:
Some npm modules have their own timer declarations (ref: https://github.com/discordjs/discord.js/blob/988a51b7641f8b33cc9387664605ddc02134859d/src/rest/RESTManager.js#L3 ), but those declarations conflicts with parameters of commonjs wrapper, and cause SyntaxError.

~~This PR adds the internal block inside the cjs wrapper and avoids the above error.~~ this doesn't work because now module can't control the strictness of the module. ("use strict" doesn't have effect anymore)

This PR adds another inner wrapper inside of the current one. The inner wrapper only provides the canonical commonjs items `exports`, `require`, `module`, `__filename`, `__dirname`, which prevents the conflicts of the declarations of timer functions.

closes #1901

### DIFF
--- a/node/module.ts
+++ b/node/module.ts
@@ -212,7 +212,7 @@ class Module {
     // to avoid exposing them in global namespace.
     `
     (function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) {
-      const inner = function(exports, require, module, __filename, __dirname) {
+      const inner = function (exports, require, module, __filename, __dirname) {
     `,
     `
       };
@@ -220,7 +220,6 @@ class Module {
     })
     `,
   ];
-
   // Loads a module at the given file path. Returns that module's
   // `exports` property.
   // deno-lint-ignore no-explicit-any
--- a/node/module.ts
+++ b/node/module.ts
@@ -210,14 +210,8 @@ class Module {
   static wrapper = [
     // We provide non standard timer APIs in the CommonJS wrapper
     // to avoid exposing them in global namespace.
-    `
-    (function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) {
-      (function (exports, require, module, __filename, __dirname) {
-    `,
-    `
-      }).call(this, exports, require, module, __filename, __dirname);
-    })
-    `,
+    "(function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) { (function (exports, require, module, __filename, __dirname) {",
+    "\n}).call(this, exports, require, module, __filename, __dirname); })",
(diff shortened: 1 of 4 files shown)


node/module.ts
```diff
@@ -210,10 +210,15 @@ class Module {
...
     // to avoid exposing them in global namespace.
-    "(function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) { ",
-    "\n});",
+    `
+    (function (exports, require, module, __filename, __dirname, setTimeout, clearTimeout, setInterval, clearInterval) {
+      (function (exports, require, module, __filename, __dirname) {
+    `,
+    `
+      }).call(this, exports, require, module, __filename, __dirname);
+    })
```
bartlomieju, line 219: I'm not a fan of this solution - this will completely break the source map support in CJS files; I think reverting injection of timers globals is a better solution.

Just FYI in Node's CJS loader, there's a mechanism that checks if wrapper was "patched" and then the loader behaves slightly differently.

kt3k: > this will completely break the source map support in CJS files

Can you elaborate how this breaks the cjs source maps? (Do we support that? If we make this wrapper a single line, can it be a workaround?)

> I think reverting injection of timers globals is a better solution.

I think reverting is not an option at the moment as it breaks the yarn compatibility. Yarn uses global `setInterval().unref()` [here](https://github.com/yarnpkg/yarn/blob/3119382885ea373d3c13d6a846de743eca8c914b/src/reporters/base-reporter.js#L169)

bartlomieju: > Can you elaborate how this breaks the cjs source maps? (Do we support that? If we make this wrapper a single line, can it be a workaround?)

Deno in not aware of this wrapper so when applying source maps it doesn't know that it should perform some additional calculations to take it into account. This more or less works fine if the wrapper is on a single line (you might get an information in the stack trace that position was out of bounds). But if we make it multiline, then it's pretty much guaranteed that stack traces will print wrong lines

kt3k: Thanks for the explanation! Then I made it a one-liner. What do you think about the current version?

bartlomieju: It's better than multiline :) let's land it

bartlomieju (approved):

node/module.ts
kt3k: Thanks!

Issue #1900, opened 2022-02-11
Type errors when importing createRequire


SheetJSDev:
**Describe the bug**

```
error: TS2339 [ERROR]: Property 'getUid' does not exist on type 'typeof Deno'.
    if (Deno.build.os !== "windows" && info.uid === Deno.getUid()) {
                                                         ~~~~~~
    at https://deno.land/std@0.125.0/node/_fs/_fs_access.ts:24:58

TS2339 [ERROR]: Property 'networkInterfaces' does not exist on type 'typeof Deno'.
      .networkInterfaces()
       ~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.125.0/node/os.ts:195:8
```

**Steps to Reproduce**

```bash
$ cd /tmp
$ pwd
/tmp
$ cat node.ts 
import { createRequire } from 'https://deno.land/std/node/module.ts';
$ deno run node.ts
Check file:///private/tmp/node.ts
error: TS2339 [ERROR]: Property 'getUid' does not exist on type 'typeof Deno'.
    if (Deno.build.os !== "windows" && info.uid === Deno.getUid()) {
                                                         ~~~~~~
    at https://deno.land/std@0.125.0/node/_fs/_fs_access.ts:24:58

TS2339 [ERROR]: Property 'networkInterfaces' does not exist on type 'typeof Deno'.
      .networkInterfaces()
       ~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.125.0/node/os.ts:195:8

Found 2 errors.
$ deno --version
deno 1.18.2 (release, x86_64-apple-darwin)
v8 9.8.177.6
typescript 4.5.2
```

**Expected behavior**

No errors with std types

**Environment**

- OS: macOS Monterey 12.1
- deno version: deno 1.18.2 (release, x86_64-apple-darwin)
- std version: 0.125.0 based on the URL

ah-yu:
@SheetJSDev You should add `unstable` flag.
`deno run --unstable node.ts`

PR #1904, opened 2022-02-11
chore: add `networkInterfaces` to `_deno_unstable.ts`

ah-yu:
`Deno.networkInterfaces` is still an unstable api.

Related: #1900


kt3k (approved):
@ah-yu LGTM. Thanks!

PR #1906, opened 2022-02-12
feat(testing): add `assertAlmostEquals`

### LINKED ISSUE #1894: testing/asserts: `assertAlmostEquals()`

I'm writing unit tests for a lot of math heavy functions, so floating point errors are all over the place. For instance:
```js
import { assertEquals } from "https://deno.land/std@0.125.0/testing/asserts.ts";

const result = sum(0.1, 0.2);

assertEquals(result, 0.3);
```
results in
```
Uncaught AssertionError: Values are not equal:

    [Diff] Actual / Expected

-   0.30000000000000004
+   0.3
```

A method for comparing numbers with a tolerance:

```js
assertAlmostEquals(result, 0.3, 0.0001 /* tolerance */);
```

alternative names:
- `assertCloseTo()` (from [chai](https://www.chaijs.com/api/bdd/#closetoexpected-delta-msg))
- `assertAlmostEqual()` or `assertAlmostEquals()` (from [Python](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertAlmostEqual))
- `assertRoughlyEquals()`

I'll use a helper function included with the tests, I'm working with vector math so I'm going to need a helper function for comparing vectors anyway, but I figured std could benefit from something like this as well.

lowlighter:
Closes #1894

I went with `assertAlmostEquals` but open to change the name but the first one proposed in #1894 sounded good

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -418,7 +418,7 @@ export function assertNotStrictEquals(
  * If the values are not almost equal then throw.
  *
  * ```ts
- * import { assertAlmostEquals } from "./asserts.ts";
+ * import { assertAlmostEquals, assertThrows } from "./asserts.ts";
  *
  * assertAlmostEquals(0.1, 0.2);
  *
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -420,7 +420,11 @@ export function assertNotStrictEquals(
  * ```ts
  * import { assertAlmostEquals } from "./asserts.ts";
  *
- * assertAlmostEquals(0.1, 0.2)
+ * assertAlmostEquals(0.1, 0.2);
+ * 
+ * // Using a custom epsilon value
+ * assertAlmostEquals(0.1 + 0.2, 0.3, 1e-16);
+ * assertThrows(() => assertAlmostEquals(0.1 + 0.2, 0.3, 1e-17));
  * ```
  */
 export function assertAlmostEquals(
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -432,16 +432,16 @@ export function assertAlmostEquals(
   if (actual === expected) {
     return;
   }
-  if (
-    (Number.isNaN(actual)) || (Number.isNaN(expected)) ||
-    (Math.abs(expected - actual) > epsilon)
-  ) {
-    const f = (n: number) =>
-      n.toFixed(-Math.log10(epsilon)).replace(/\.?0+$/, "");
-    throw new AssertionError(
-      msg ?? `actual: "${f(actual)}" expected to be close to "${f(expected)}"`,
-    );
+  const delta = Math.abs(expected - actual);
+  if (delta <= epsilon) {
+    return;
   }
+  const f = (n: number) => Number.isInteger(n) ? n : n.toExponential();
+  throw new AssertionError(
+    msg ??
+      `actual: "${f(actual)}" expected to be close to "${f(expected)}": \
+delta "${f(delta)}" is greater than "${f(epsilon)}"`,
+  );
 }
 
 /**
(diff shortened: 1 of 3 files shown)


bartlomieju:
This seems useful, I'm in favor. @lowlighter coukd you please open a PR to `manual` repo that updates "Testing" chapter with this assertion?

lowlighter:
@bartlomieju Opened https://github.com/denoland/manual/pull/222/files to cover these changes 👍

testing/asserts.ts
```diff
@@ -411,6 +411,39 @@ export function assertNotStrictEquals(
...
+    (Number.isNaN(actual)) || (Number.isNaN(expected)) ||
+    (Math.abs(expected - actual) > epsilon)
+  ) {
+    const f = (n: number) =>
+      n.toFixed(-Math.log10(epsilon)).replace(/\.?0+$/, "");
+    throw new AssertionError(
+      msg ?? `actual: "${f(actual)}" expected to be close to "${f(expected)}"`,
+    );
+  }
+}
```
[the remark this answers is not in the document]
djedr: > However there's one thing that I think look weird it to always force exponential notation.
> Maybe we could add a threshold, because printing `3e-1` instead of `0.3` doesn't feel natural.
> Like if less than 7 significant digits we should fallback to `toFixed()` notation ? I think it'd increase readability

True, it's just tricky to get this right. If you think the readability is worth it though, I have no objections. Just trying to err on the side of simplicity. :slightly_smiling_face:

testing/asserts.ts
```diff
@@ -411,6 +411,39 @@ export function assertNotStrictEquals(
...
+/**
+ * Make an assertion that `actual` and `expected` are almost equal numbers through
+ * a given epsilon. It can be used to take into account IEEE-754 double-precision
+ * floating-point representation limitations.
+ * If the values are not almost equal then throw.
+ *
+ * ```ts
+ * import { assertAlmostEquals } from "./asserts.ts";
+ *
+ * assertAlmostEquals(0.1, 0.2)
```
bartlomieju, line 423: Please add an example which uses the third argument too

bartlomieju (commented):
Looks great 👍 seems very useful

testing/asserts.ts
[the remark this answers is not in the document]
lowlighter: Yeah I think it's hard to get it right too. I'll let as is for now 😅

testing/asserts.ts
```diff
@@ -411,6 +411,43 @@ export function assertNotStrictEquals(
   );
 }
 
+/**
+ * Make an assertion that `actual` and `expected` are almost equal numbers through
+ * a given epsilon. It can be used to take into account IEEE-754 double-precision
```
kt3k, line 416: Not a change request, but I'm little unsure if this `epsilon` usage is enough descriptive to everyone. `difference` might be easier to understand, I guess

kt3k (approved):
@lowlighter LGTM. Looks like a nice addition!

testing/asserts.ts
bartlomieju: Good point

djedr: Another common name for this parameter is `delta`. But `difference` is the most likely of these three to be in the vocabulary of the average user.

ah-yu: `tolerance` may be a good option

lowlighter: I'll change it to `tolerance`, seems to be more close to the meaning of `epsilon` I guess

bartlomieju (approved):
LGTM too, nice addition @lowlighter!

testing/asserts.ts
kt3k: @lowlighter Thanks for updating!

Issue #1909, opened 2022-02-13
TS2322 [ERROR]: Type 'number | Partial<NetConnectOptions>' is not assignable to type 'number'.


idanran:
**Describe the bug**
TS2322 [ERROR]: Type 'number | Partial<NetConnectOptions>' is not assignable to type 'number'.
  Type 'Partial<TcpNetConnectOptions>' is not assignable to type 'number'.
    (options as TcpSocketConnectOptions).port = arg0;
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.122.0/node/net.ts:287:5

**Steps to Reproduce**
1. Create a file with the contents ...
2. Create a script with the contents ...
3. Execute the command ...
4. See error

**Expected behavior**
No error reporting.
<!-- A clear and concise description of what you expected to happen. -->

**Environment**

<!-- please complete the following information -->

- OS: Windows11
- deno version: 1.18.2
- std version: 0.125.0

djedr:
Reason for this is that type inference fails because the `arg === null` case is not handled.

This can be solved by replacing line 280 with:

```js
  if (arg0 !== null) if (typeof arg0 === "object") {
```

(+ wrap the enclosing `if`s body in curlies for proper style)

although whether that makes sense for the rest of the logic is another question (`options` will then remain empty).

idanran:
> Reason for this is that type inference fails because the `arg === null` case is not handled.
> 
> This can be solved by replacing line 280 with:
> 
> ```js
>   if (arg0 !== null) if (typeof arg0 === "object") {
> ```
> 
> (+ wrap the enclosing `if`s body in curlies for proper style)
> 
> although whether that makes sense for the rest of the logic is another question (`options` will then remain empty).

Maybe you can try to mention pr. I don't know much about typescript.

kt3k:
> Create a file with the contents ...
> Create a script with the contents ...

What is the 'contents' here?

I don't see a type error when I run the command `deno run --unstable https://deno.land/std@0.122.0/node/net.ts`. Do you have some customized compiler options?

idanran:
> > Create a file with the contents ...
> > Create a script with the contents ...
> 
> What is the 'contents' here?
> 
> I don't see a type error when I run the command `deno run --unstable https://deno.land/std@0.122.0/node/net.ts`. Do you have some customized compiler options?

I turned off strict mode in Typescript.

[...]

PR #1911, opened 2022-02-13
fix(node/net) TypeScript error TS2322

djedr:
Closes https://github.com/denoland/deno_std/issues/1909.

This is only a simple fix for the type error.
Not sure if it pans out with the logic.

First PR, not sure if I did everything right.

Tests (node/_tools/test.ts) pass.

Reasoning explained [here](https://github.com/denoland/deno_std/issues/1909#issuecomment-1038262002). Copy-paste:

> Reason for this is that type inference fails because the `arg === null` case is not handled.

> This can be solved by replacing line 280 with:

```js
  if (arg0 !== null) if (typeof arg0 === "object") {
```

> although whether that makes sense for the rest of the logic is another question (`options` will then remain empty).

### DIFF
--- a/node/net.ts
+++ b/node/net.ts
@@ -277,7 +277,7 @@ export function _normalizeArgs(args: unknown[]): NormalizedArgs {
   const arg0 = args[0] as Partial<NetConnectOptions> | number | string;
   let options: Partial<SocketConnectOptions> = {};
 
-  if (typeof arg0 === "object" && arg0 !== null) {
+  if (arg0 !== null) if (typeof arg0 === "object") {
     // (options[...][, cb])
     options = arg0;
   } else if (_isPipeName(arg0)) {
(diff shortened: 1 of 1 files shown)


djedr:
I see that https://github.com/denoland/deno_std/issues/1909 was closed, so I suppose this is not relevant anymore.

Not supporting loose mode is sensible. Although technically this now makes the `&& arg0 !== null` check superfluous from the typechecker's perspective (the code will remain correct when it is removed).

On the other hand the correctness hangs on the assumption from line 277:

```js
  const arg0 = args[0] as Partial<NetConnectOptions> | number | string;
```
which may not be true if `args` are truly `unknown`, i.e. come from unreliable source which may provide `null` or `undefined` -- which might be the case here (haven't dug deep enough to definitively say). In that case the assumption is wrong and the code actually should be guarded against both null and undefined.

I'll leave this for somebody more familiar with the logic to judge. Feel free to close this PR. Otherwise I am happy to update it to handle the null or undefined case -- if you tell me what would be the desired outcome then. The dumbest solution is to ignore them, as I did in this PR.

bartlomieju:
@kt3k do you think we should land this one?

kt3k:
Let's close this one because this patch is only meaningful for typescript 'loose' mode, which we don't want to support. The original issue was also closed by the same reason. #1909

PR #1917, opened 2022-02-14
fix(io/buffer): super and initialized prop

kitsonk:
This gets rid of the error messages like the following when using `io/buffer.ts` under tsc directly via dnt: 

```
TS2376: A 'super' call must be the first statement in the constructor when a class contains initialized properties, parameter properties, or private identifiers.
```

While this is being fixed in future version of TypeScript, it is a limitation that is easy to work around and allows people to use the `std` lib via dnt without type errors.

### DIFF
--- a/io/buffer.ts
+++ b/io/buffer.ts
@@ -723,11 +723,7 @@ export class BufWriter extends AbstractBufBase implements Writer {
   }
 
   constructor(writer: Writer, size: number = DEFAULT_BUF_SIZE) {
-    if (size <= 0) {
-      size = DEFAULT_BUF_SIZE;
-    }
-    const buf = new Uint8Array(size);
-    super(buf);
+    super(new Uint8Array(size <= 0 ? DEFAULT_BUF_SIZE : size));
     this.#writer = writer;
   }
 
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

I was actually seeing the same error while using dnt

kt3k:
@kitsonk I'm in favor of this change, but I'm not sure why this is ok in std and causes type error when using dnt transformation. Can this be somehow tested in std's CI?

kitsonk:
@kt3k no, because it only occurs when using `tsc` directly and not targeting esnext.  The fix is in TypeScript 4.6, and this is the PR that fixed it: https://github.com/microsoft/TypeScript/pull/29374.  You cannot force Deno to typecheck in non-esnext mode.  The only way to manifest this would be to run some sort of process the converts everything via dnt to "node" code and then runs tsc against it, and I see that seperate to this PR if such a thing is desired.

kitsonk:
I

> @kitsonk I'm in favor of this change, but I'm not sure why this is ok in std and causes type error when using dnt transformation. Can this be somehow tested in std's CI?

I opened #1918 to cover this.

kt3k:
Thanks for the explanation and creating the issue!

> https://github.com/microsoft/TypeScript/pull/29374.

Oh, that birthday cake PR!

PR #1921, opened 2022-02-14
test(node/path): Enable test-path-exists.js

wafuwafu13:


wafuwafu13:
https://github.com/denoland/deno_std/runs/5180211597?check_suite_focus=true
failures native modules are extensible

https://github.com/denoland/deno_std/runs/5244481502?check_suite_focus=true
same problem as https://github.com/denoland/deno_std/pull/1882#discussion_r800139492

node/module_all.ts
```diff
@@ -116,6 +116,8 @@ export default {
   "internal/util": internalUtil,
   net,
   os,
+  "path/posix": path,
```
kt3k, line 119: I think we need to pass something like the below here:

```ts
import pathPosix from "./path/posix.ts";
```
And that means we need to create default export of `path/posix.ts`

```ts
export default {
  sep,
  delimiter,
  ...
}
```
We probably also need to update `path/mod.ts` like the below

```diff
-import * as _posix from "./posix.ts";
+import _posix from "./posix.ts";
```

kt3k (commented):
Sorry for missing this PR. Left a comment

kt3k (approved):
@wafuwafu13 LGTM! Nice work!

PR #1934, opened 2022-02-17
Don't use globalThis.setTimeout types in node/timers

espindola:
Unlike globalThis.setTimeout in browsers, the setTimeout in node only
accepts functions, not strings.

The existing code works fine when typechecking with
lib.deno.shared_globals.d.ts, which declares setTimeout callback
argument as "(...args: any[]) => void". It is only a problem when
using lib.dom.d.ts, which declares the callback as having type
"string | Function". With the lib.dom.d.ts declaration, typechecking
fails because there is no .bind in string.


kt3k (approved):
@espindola This makes sense to me. Thanks!

PR #1937, opened 2022-02-18
feat(dotenv): Dotenv restructure

timreichen:
https://github.com/denoland/deno_std/pull/1877#issuecomment-1043461778


dotenv/README.md
```diff
@@ -2,113 +2,76 @@
...
 ```
 
+### load
+
 ```ts
 // app.ts
-import "https://deno.land/std@$STD_VERSION/dotenv/load.ts";
-
+import { load } from "https://deno.land/std@$STD_VERSION/dotenv/load.ts";
+await load(Deno.env);
```
kt3k, line 53: I'm not in favor of this change. In my view most of use cases of this module is this pattern. So this adds some boilerplates to the main use case. This seems making this library a little inconvenient

dotenv/README.md
```diff
@@ -2,113 +2,76 @@
...
 ### Options
 
-- `path?: string`: Optional path to `.env` file. Defaults to `./.env`.
-- `export?: boolean`: Set to `true` to export all `.env` variables to the
-  current processes environment. Variables are then accessable via
-  `Deno.env.get(<key>)`. Defaults to `false`.
-- `safe?: boolean`: Set to `true` to ensure that all necessary environment
-  variables are defined after reading from `.env`. It will read `.env.example`
-  to get the list of needed variables.
-- `example?: string`: Optional path to `.env.example` file. Defaults to
```
kt3k, line 37: Why delete `.env.example` as default?

dotenv/README.md
```diff
@@ -2,113 +2,76 @@
...
-  `./.env.example`.
+- `example?: DotEnvObject`: example object that the parsed data will be verified
+  against.
 - `allowEmptyValues?: boolean`: Set to `true` to allow required env variables to
   be empty. Otherwise it will throw an error if any variable is empty. Defaults
   to `false`.
-- `defaults?: string`: Optional path to `.env.defaults` file which defaults to
```
kt3k, line 42: Why delete `defaults` feature?

dotenv/README.md
timreichen: Do you mean adding `Deno.env` as an argument and/or the function call?

kt3k: I mean the function call looks redundant. People should be interested only in that use case. So it's better to do that automatically in my view.

jsejcksn:
(I haven't reviewed the code in this PR yet, however) I want to migrate some of the conceptual context from the introductory PR to this one so that it's more accessible during review and discussion.

@timreichen You [previously said](https://github.com/denoland/deno_std/pull/1877#issuecomment-1051765888):

> I think it is possible to achieve both low level API access as well as convenience for users. A bit of magic is fine for some use cases as long as there is a non-magic way to do the same thing :)

My interpretation of that statement is that the module should be written in a style that uses composable functions for achieving each of its goals/results and expose those functions as exports for consumers to use so that they can replicate the process in stages (without being required to participate in the entire chain of effects/mutations).

For example, AFAICT, if a consumer wants to load a `.env` file using the `load` module, there's no way to determine which environment variables came from the `.env` and which came from the pre-existing environment. **And** there's no simple way to do that with the provided tools. Or perhaps the consumer wants to load _part_ of the `.env`, filtering some entries.

It seems to me that the kinds of changes that you are suggesting are in the spirit of making things like that possible/simpler for the consumer.

If that interpretation is incorrect/incomplete, can you clarify the explicit goals here for any reviewers?

dotenv/load.ts
```diff
@@ -1,5 +1,47 @@
-// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { DenoEnv, parse } from "./mod.ts";
```
jsejcksn, line 1:
```suggestion
import { type DenoEnv, parse } from "./mod.ts";
```

Use [`type` modifier on type import names](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names)

dotenv/mod.ts
```diff
@@ -1,187 +1,107 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
```
jsejcksn, line 2:
```suggestion
// This module is browser compatible.

```

This should be marked if it's true. (I didn't test with the compiler, but it appears to be.)

dotenv/README.md
jsejcksn: In effort toward preserving existing behavior: How about moving everything in `load.ts` to a new module (e.g. `loader.ts`), then import the `load` function into `load.ts` and invoke it at the top level?

dotenv/mod_test.ts
```diff
@@ -1,434 +1,153 @@
...
 
-const testOptions = {
-  path: path.join(testdataDir, "./.env"),
-  defaults: path.join(testdataDir, "./.env.defaults"),
-};
-
-Deno.test("parser", () => {
-  const testDotenv = Deno.readTextFileSync(
-    path.join(testdataDir, "./.env.test"),
+Deno.test("parse", () => {
```
jsejcksn, line 10: I think it makes sense to use the [test steps API](https://deno.land/manual@v1.19.2/testing#test-steps) here, because the assertions test independent behaviors of the function.

e.g.

```ts
Deno.test("parse", async (ctx) => {
  // ...
  await ctx.step("parses a basic variable", () => assertEquals(env.BASIC, "basic"));
  await ctx.step("skips empty lines", () => assertEquals(env.AFTER_EMPTY, "empty"));
  // etc...
});
```

dotenv/mod.ts
```diff
@@ -1,187 +1,107 @@
...
-export class MissingEnvVarsError extends Error {
-  constructor(message?: string) {
-    super(message);
-    this.name = "MissingEnvVarsError";
-    Object.setPrototypeOf(this, new.target.prototype);
-  }
-}
+// would be a nice to be able to generate .env files and complete parse/stringify pair like in JSON
+// export function stringify(object: DotEnvObject) {
+// }
```
jsejcksn, line 107: I'd very much like to see a variation on this functionality. See https://github.com/pietvanzoen/deno-dotenv/issues/42. I envision the use of `.env` as an accessible and simplistic file-based kvstore of string data.

dotenv/mod.ts
```diff
@@ -1,187 +1,106 @@
...
 
-export interface ConfigOptions {
-  path?: string;
-  export?: boolean;
-  safe?: boolean;
-  example?: string;
-  allowEmptyValues?: boolean;
-  defaults?: string;
+// Deno should expose a type for Deno.env
+// Ideally Map-like as discussed in https://github.com/denoland/deno/issues/10349
```
jsejcksn, line 10: Using Deno NS types will prevent type-safe browser compatibility. It is useful for these types to be declared independently.

timreichen:
> My interpretation of that statement is that the module should be written in a style that uses composable functions for achieving each of its goals/results and expose those functions as exports for consumers to use so that they can replicate the process in stages (without being required to participate in the entire chain of effects/mutations).

That is exactly right.

bartlomieju (changes requested):
@timreichen while this PR has a lot of useful APIs added, I'm not in favor of breaking existing code in this case. 

Loading and applying env variables from a file in a single API is more convenient than having to call 2-3 functions yourself. It seems to me like most of the current API and functionality could be preserved by composing various APIs introduced in this PR.

Could you restructure your PR in such a way that it preserves the current API while exposing lower-level APIs should the user need it?

timreichen:
@bartlomieju I restored the `load.ts` file to do what it did before, namely automatically reading, verifying and loading the variables from `.env` to `Deno.env`.
The `mod.ts` now has `parse`, `stringify`, `verify` as low-level APIs and  `load` and `loadSync` to do the process that is happening in `load.ts` but with the option of passing file paths.
So the only breaking change I think would be the renaming of `config` to `load` resp. `configSync` to `loadSync` which imo makes more sense. What do you think?

timreichen:
@bartlomieju Why is this being tested? Imo this shouldn't be part of the API. If it fails, it fails no? 
https://github.com/denoland/deno_std/blob/f415da4c62cdacce880c3c7e492ceee184c3be4a/dotenv/mod_test.ts#L156-L167

ayame113:
> Imo this shouldn't be part of the API. If it fails, it fails no?

This behavior is required to support deno deploy. deno deploy does not have Deno.readFileSync. It is necessary to avoid throwing an error even if there is no Deno.readFileSync. https://github.com/denoland/deno_std/issues/1957#issuecomment-1050497911

timreichen:
I think we should not do special handling for deno deploy in such a way. Since the new api has separate `load` and `loadSync` functions and `std/dotenv/load.ts` is just calling `loadSync`, we should encourage users to call `load` instead of importing `load.ts` for deno deploy.

kt3k:
Or how about providing `load_async.ts` endpoint for deno deploy compatibility?

ayame113:
I think the following development methods are common.

- Use synchronous version of dotenv in local development
- Use built-in environment variable support in deno deploy production (and ignore errors from dotenv)

So I'd like to keep the current behavior, but if there's another way to support deno deploy, I'm fine.

kt3k:
@ayame113 You're right. No need of `load_async.ts` to support deno deploy. I retract `load_async.ts` suggestion. Let's keep  this as is.

timreichen:
Then I suggest we put the check in `load.ts` instead of inside the `loadSync` function and log a warning what nothing was read. Would that be ok?

kt3k:
> Then I suggest we put the check in load.ts instead of inside the loadSync function and log a warning that nothing was read.

Sounds good to me

dotenv/mod.ts
```diff
@@ -1,191 +1,336 @@
...
+ * ```ts
+ * import { verify } from "https://deno.land/std@$STD_VERSION/dotenv/mod.ts";
+ *
+ * const object = { env: { GREETING: "hello world" }, exports: [] };
+ * const exampleEnv = { GREETING: "someValue" };
+ * const options = { allowEmptyValues: true };
+ * verify(object, exampleEnv, options);
+ * ```
+ */
+export function verify(
```
kt3k, line 31: `verify` looks like an internal utility to me. Should this be exported? (What is verified by this is not documented in jsdoc or README. I guess the users don't understand what's done by this.)

kt3k:
`load`, `loadSync`, `parse`, and `stringify` looks good to me.

kt3k:
note #2316

kt3k:
Also restored the inline comment feature introduced in #2316 in 7516bef290e6aa79fc9891781b5de1841e23550d

kt3k (approved):
LGTM

kt3k:
Summary of the change in this PR:
- Added 2 primitive APIs `parse` and `stringify`
- Added 2 new APIs `load` and `loadSync`, which are better version of `config` and `configSync`
- Deprecated 2 APIs `config` and `configSync`

kt3k:
@bartlomieju PTAL

dotenv/README.md
```diff
@@ -9,127 +9,131 @@ Setup a `.env` file in the root of your project.
...
+  Defaults to ".env.example".
+- `defaultsPath?: string | URL`: the path or url to the .env.defaults file.
+  Defaults to ".env.defaults".
+- `allowEmptyValues?: boolean`: set to `true` to allow variables with an empty
+  value.
+
+## Auto loading
+
+`load.ts` automatically loads the local `.env` file, verifies it against
+`.env.exanple` if existent, adds defaults from `.env.defaults` if existent and
```
bartlomieju, line 58:
```suggestion
`.env.example` if existent, adds defaults from `.env.defaults` if existent and
```

dotenv/README.md
```diff
@@ -9,127 +9,131 @@ Setup a `.env` file in the root of your project.
...
-```
+## Parsing Rules
 
-### Safe Mode
+The parsing engine currently supports the following rules:
 
-To enable safe mode, create a `.env.example` file in the root of the project.
+- `BASIC=basic` becomes `{BASIC: 'basic'}`
+- empty lines are skipped
+- lines beginning with `#` are treated as comments
```
bartlomieju, line 78: Maybe rephrase that anything after `#` is treated as a comment (ie. `FOO=bar # this is a comment` works)

dotenv/load.ts
```diff
@@ -1,5 +1,12 @@
...
 
-configSync({ export: true });
+if (!(Deno.readTextFileSync instanceof Function)) {
+  // deno deploy doesn't have Deno.readFileSync
+  console.warn(
+    `Deno.readTextFileSync is not a function: No .env data was read.`,
+  );
+} else {
+  loadSync();
+}
```
bartlomieju, line 12: I don't understand this bit, so automatic loading is always synchronous?

kt3k: > automatic loading is always synchronous?

Yes, otherwise synchronous top-level `Deno.env.get()` call happens before loading variables from `.env`

We fixed that in #1964

kt3k:
@timreichen Please add a test case to check the parse result of `dotenv/testdata/_config/.env.test`.

It's now like the below:
```js
{ env: { BASIC: "basic", AFTER_EMPTY: "empty" }, exports: [] }
```

It used to be like the below at bee7b39 (everything after the first comment line looks ignored now)
```js
{
  env: {
    BASIC: "basic",
    AFTER_EMPTY: "empty",
    EMPTY_VALUE: "",
    QUOTED_SINGLE: "single quoted",
    QUOTED_DOUBLE: "double quoted",
    MULTILINE: "hello\nworld",
    JSON: '{"foo": "bar"}',
    WHITESPACE: "    whitespace   ",
    WHITESPACE_DOUBLE: "    whitespace   ",
    MULTILINE_SINGLE_QUOTE: "hello\\nworld",
    EQUALS: "equ==als",
    THE_ANSWER: "42",
    VAR_WITH_SPACE: "var with space",
    VAR_WITH_ENDING_WHITESPACE: "value",
    V4R_W1TH_NUM8ER5: "var with numbers",
    INVALID: "var starting with a number",
    INDENTED_VAR: "indented var",
    INDENTED_VALUE: "indented value",
    TAB_INDENTED_VAR: "indented var",
    TAB_INDENTED_VALUE: "indented value"
  },
  exports: []
}
```

timreichen:
I added the env.test case but also the missing test cases for the parser.

dotenv/mod_test.ts
```diff
@@ -328,6 +435,37 @@ Deno.test("configure", () => {
   restoreDenoEnv();
 });
 
+Deno.test("configure .env.test", () => {
```
kt3k, line 452: Nice!

Now this looks backward compatible

kt3k (approved):
LGTM

timreichen (commented):
Deprecated functions should work as before now.

kt3k:
@timreichen Sorry. We can't land this PR as is because it includes too many irrelevant changes (changes of design, parsing algorithm, new API addition, change of code structure & testing structure, etc) and it's hard to confirm this doesn't include any regression. We also don't find enough value in the features this PR adds compared to the risk of bugs this can cause.

We are still open to review/discuss each change **one by one** in **small PRs**. If you're still interested in working on this. Please split them into small PRs.

PR #1939, opened 2022-02-19
feat(async): Add `abortable` to async module.

lambdalisue:
Hi. 

The `abortable` is a wrapper function that makes `Promise` and `AsyncIterable` cancelable.

For example, in the case of `Promise`, it looks like this

```typescript
import { abortable } from "https://deno.land/std/async/mod.ts";
import { delay } from "https://deno.land/std/async/mod.ts";
const p = delay(1000);
const c = new AbortController();
setTimeout(() => c.abort(), 100);
// Below throws `AbortedError` after 100 ms
await abortable(p, c.signal);
```

and for `AsyncIterable` as follows

```typescript
import { abortable } from "https://deno.land/std/async/mod.ts";
import { delay } from "https://deno.land/std/async/mod.ts";
const p = async function* () {
  yield "Hello";
  await delay(1000);
  yield "World";
};
const c = new AbortController();
setTimeout(() => c.abort(), 100);
// Below throws `AbortedError` after 100 ms
// and items become `["Hello"]`
const items: string[] = [];
for await (const item of abortable(p(), c.signal)) {
  items.push(item);
}
```

I hope you all like it.

### DIFF
--- a/async/abortable.ts
+++ b/async/abortable.ts
@@ -1,0 +1,68 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { deferred } from "./deferred.ts";
+
+export class AbortedError extends Error {
+  reason?: any;
+
+  constructor(reason?: any) {
+    super(reason ? `Aborted: ${reason}` : "Aborted");
+    this.name = "AbortedError";
+    this.reason = reason;
+  }
+}
+
+/**
+ * Make Promise or AsyncIterable abortable with a given signal.
+ */
+export function abortable<T>(p: Promise<T>, signal: AbortSignal): Promise<T>;
+export function abortable<T>(
+  p: AsyncIterable<T>,
+  signal: AbortSignal,
+): AsyncGenerator<T>;
+export function abortable<T>(
+  p: Promise<T> | AsyncIterable<T>,
+  signal: AbortSignal,
+): Promise<T> | AsyncIterable<T> {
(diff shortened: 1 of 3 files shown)


kt3k (approved):
Thanks for your contribution! This seems a reasonable addition to `std/async` to me.

LGTM

async/abortable.ts
```diff
@@ -0,0 +1,72 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { deferred } from "./deferred.ts";
+
+export class AbortedError extends Error {
```
kt3k, line 4: I wonder if it makes sense to use DOMException with AbortError  instead of this custom error:

```ts
new DOMException(reason ? `Aborted: ${reason}` : "Aborted", "AbortError")
```

crowlKats: Yes i think this would be preferable imo. Ideally it would be using error causes, but DOMException doesnt support that yet (https://github.com/whatwg/webidl/issues/969).

lambdalisue: Shall I rewrite it as `DOMException` with that?

kt3k: > Shall I rewrite it as DOMException with that?

I think that's more preferable (fetch API rejects with that error, for example). Let's do that

kt3k (approved):
@lambdalisue LGTM!

Issue #1933, opened 2022-02-17
Make the standard modules "noImplicitOverride"-compliant


elisee:
**Describe the bug**

Currently some standard modules will throw an error when imported with TypeScript's `noImplicitOverride` option enabled.  Although [it is not (yet?) enabled by default for Deno](https://github.com/denoland/deno/issues/11836), I believe it should be possible to use all standard modules when one opts into `noImplicitOverride`?

**Steps to Reproduce**

For instance:

```ts
import { assert } from "https://deno.land/std@0.126.0/testing/asserts.ts";
```

with a `deno.json` file containing:

```ts
{
  "compilerOptions": {
    "noImplicitOverride": true
  }
}
```

Results in the following error:

```
error: TS4114 [ERROR]: This member must have an 'override' modifier because it overrides a member in the base class 'Error'.
  name = "AssertionError";
  ~~~~
    at https://deno.land/std@0.126.0/testing/asserts.ts:20:3
```

**Expected behavior**

No error.

**Environment**

- OS: Windows 10 x64
- deno version: 1.19.0
- std version: 0.126.0

PR #1940, opened 2022-02-19
chore: Enable "noImplicitOverride" compiler option and fix errors

elisee:
This commit makes deno_std `noImplicitOverride`-compliant and enables the compiler option in `deno.json` so that further changes don't regress it. Closes https://github.com/denoland/deno_std/issues/1933

### DIFF
--- a/encoding/_yaml/error.ts
+++ b/encoding/_yaml/error.ts
@@ -14,7 +14,7 @@ export class YAMLError extends Error {
     this.name = this.constructor.name;
   }
 
-  public toString(_compact: boolean): string {
+  public override toString(_compact: boolean): string {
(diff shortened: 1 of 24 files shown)


deno.json
```diff
@@ -26,5 +26,8 @@
         "wasi/testdata"
       ]
     }
+  },
+  "compilerOptions": {
+    "noImplicitOverride": true
```
bartlomieju, line 31: This change should be added to [strict-ts44.tsconfig.json](https://github.com/denoland/deno_std/blob/main/strict-ts44.tsconfig.json) instead

elisee: Oh alright, fixed.

kt3k (approved):
LGTM

bartlomieju (approved):

Issue #1475, opened 2021-10-26
std/node: support the "tls" module


piscisaureus:
- [x] connect
- [x] createServer https://github.com/denoland/deno_std/pull/1948

bnoordhuis:
This is at least partially blocked on https://github.com/briansmith/webpki/issues/219 - webpki rejects certain types of certificates that are exceedingly common when dealing with self-signed CAs.

A workaround is to add x509v3 extensions to the server certificate to force it be a v3 certificate (note: an empty extension list won't work - also rejected), but that means our tls module is hardly a drop-in replacement.

piscisaureus:
I would suggest to pursue this without support for self signed X509 version 1 certificates.

If you need to generate a self signed certificate that works (e.g. for testing) you can use the following snippet (works on mac):

```
openssl req -new -x509 -nodes -days 365 -text -subj "/CN=localhost" -extensions v3_req \
    -config <(cat /etc/ssl/openssl.cnf <(printf "\n[v3_req]\nbasicConstraints=critical,CA:TRUE\nkeyUsage=nonRepudiation,digitalSignature,keyEncipherment\nsubjectAltName=DNS:localhost")) \
    -keyout server.key -out server.crt 
```

PR #1948, opened 2022-02-22
feat(node/tls): basic support of tls.createServer

kt3k:
This PR adds basic support of `tls.createServer` in `std/node`.

part of #1475


bartlomieju (approved):
LGTM

Issue #1952, opened 2022-02-23
node: move `node/_tools/suites` to `node/_tools/test`


kt3k:
Some test cases in Node.js repository suppose that the directory structure is `test/parallel`, etc. So we need to rename the dir `suites` to `test` to pass those test cases without modification.

kt3k:
Affected test cases are `test-path-dirname.js` and `test-fs-realpath-native.js` for now.

PR #1953, opened 2022-02-23
chore(node): rename _tools/suites to _tools/test

kt3k:
closes https://github.com/denoland/deno_std/issues/1952


bartlomieju (approved):
LGTM, good catch

PR #1964, opened 2022-02-25
fix(dotenv): avoid top-level-await in load.ts

### LINKED ISSUE #1957: Avoid top-level-await in dotenv/load.ts
Since the top level await is in `dotenv/load.ts`, `load.ts` runs asynchronously. Therefore, another file that is imported cannot use the environment variables that dotenv loads. (See example.)

```
# .env
CONFIG="hello"
```

```ts
// main.ts
import "https://deno.land/std@0.127.0/dotenv/load.ts";
import "./b.ts";
```

```ts
// b.ts
console.log(Deno.env.get("CONFIG")); //=> undefined (!?)
```

This is inconvenient when we want to manage environment variables used in external modules with dotenv. I think we need to replace the function used in load.ts with a synchronous function.

I'm the one who introduced this to dotenv for deno deploy compatibility (https://github.com/pietvanzoen/deno-dotenv/pull/55), but now I think this choice was wrong. For deno deploy compatibility, I think you should use try-catch and ignore `readFileSync is not a function`.
https://github.com/denoland/deno_std/blob/3cb1f799801fe0470e4d87ea14e70800ba06bfd3/dotenv/mod.ts#L106-L113

ayame113:
close #1957

### DIFF
--- a/dotenv/load.ts
+++ b/dotenv/load.ts
@@ -1,5 +1,5 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-import { config } from "./mod.ts";
+import { configSync } from "./mod.ts";
 
-await config({ export: true });
+configSync({ export: true });
--- a/dotenv/load_test.ts
+++ b/dotenv/load_test.ts
@@ -29,3 +29,28 @@ Deno.test({
     p.close();
   },
 });
+
+Deno.test({
+  name: "load when multiple files",
+  async fn() {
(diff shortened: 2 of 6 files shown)


kt3k (approved):
LGTM

CI seems blocked by #1961

ayame113:
> CI seems blocked by https://github.com/denoland/deno_std/pull/1961

Okay, let's wait for it

PR #1967, opened 2022-02-26
fix(testing): assertObjectMatch for RegExp/Map/Set

lowlighter:
The following currently passes, though it shouldn't:
```ts
assertObjectMatch({ foo: /abc*/ }, { foo: /abc+/ });   // different pattern
assertObjectMatch({ foo: /abc+/ }, { foo: /abc+/i });  // different flags
```

It's because `RegExp` are of type `"object"` so they enter in the `filter()` function.

For some reason, the regex source pattern isn't listed by both `Object.getOwnPropertyNames` and `Object.getOwnPropertySymbols` methods, which only returns `"lastIndex"` (which isn't' enough to differentiate between 2 regexs):
```ts
const r = /abc/i
console.log(typeof r, [...Object.getOwnPropertyNames(r), ...Object.getOwnPropertySymbols(r)])
// "object", Array [ "lastIndex" ]
```

This is a small patch which avoids regexs entering this condition and instead being treated like a primitive

### DIFF
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -382,6 +382,11 @@ Deno.test("testingAssertObjectMatching", function (): void {
   const k = { foo: [[1, [2, [3]]]], bar: true };
   const l = { foo: [[1, [2, [a, e, j, k]]]], bar: true };
   const m = { foo: /abc+/i, bar: [/abc/g, /abc/m] };
+  const n = {
+    foo: new Set(["foo", "bar"]),
+    bar: new Map([["foo", 1], ["bar", 2]]),
+    baz: new Map([["a", a], ["b", b]]),
+  };
 
   // Simple subset
   assertObjectMatch(a, {
@@ -446,6 +451,11 @@ Deno.test("testingAssertObjectMatching", function (): void {
   // Regexp
   assertObjectMatch(m, { foo: /abc+/i });
   assertObjectMatch(m, { bar: [/abc/g, /abc/m] });
+  //Built-in data structures
+  assertObjectMatch(n, { foo: new Set(["foo"]) });
+  assertObjectMatch(n, { bar: new Map([["bar", 2]]) });
+  assertObjectMatch(n, { baz: new Map([["b", b]]) });
+  assertObjectMatch(n, { baz: new Map([["b", { foo: true }]]) });
 
   // Missing key
   {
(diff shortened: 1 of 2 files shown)


ah-yu:
Good point. Seems to have the same problem when using `assertObjectMatch` with non-plain object.
For example:
```ts
assertObjectMatch({ foo: new Map([["key1", "value1"]]) }, {
  foo: new Map([["key2", "value2"]]),
}); // should fail but passes
```

lowlighter:
Good point.
I think we need to clarify what we want to achieve from these non-plain object starting from this statement.

Since #1419 handled how we treat class instances and #1843 for typed arrays, the only notable objects that come to my mind which would benefit a special treatment would be `Set` and `Map` (I think `Date` already work as expected, and I'm not sure if `WeakSet`/`WeakMap` can be handled easily)

With #906, the behaviour adopted for `Array` is that we allow partial values matching inside them, like:
```ts
assertObjectMatch({ foo: [{ foo: 1, bar: 2 }] }, { foo: [{ foo: 1 }] });
```

It shouldn't be too hard to implement for `Map` since they're essentially like arrays but with more complex key typing.
The folloing should hence pass:
```ts
assertObjectMatch({
  foo: new Map([["foo", { foo: true, bar: true }], ["bar", { bar: true }]]),
}, { 
  foo: new Map([["foo", { foo: true }]]),
});
```
In terms of code, I think it'd be just filtering commons keys, and applying the `filter()` on each value.

For `Set` it may be trickier though. 
Technically they keep their insertion order which mean we could treat them as array-like, but the `assertEquals` ignores it (`  assertEquals(new Set(["a", "b"]), new Set(["b", "a"]));`) which is also what makes the most sense to me too.

So I would expect the following to pass too:
```ts
assertObjectMatch({
  foo: new Set([{ foo: true, bar: true }, { bar: true }]),
}, {
  foo: new Set([{ bar: true }, { foo: true }]),
});
```
But currently I can't think of an efficient way of achieving the above, because in this example the match are crossed.

ah-yu:
Your suggestion is good , but it makes things too complicated. I think it is enough to do a match for plain object and array object. It is not a common scenario that doing a match for `Map/Set/etc`, and users can convert `Map` and `Set` to array object if needed

testing/asserts_test.ts
```diff
@@ -613,6 +618,20 @@ Deno.test("testingAssertObjectMatching", function (): void {
...
+    let didThrow;
+    try {
+      assertObjectMatch(m, { foo: /abc+/ });
+      assertObjectMatch(m, { foo: /abc*/i });
+      assertObjectMatch(m, { bar: [/abc/m, /abc/g] });
+      didThrow = false;
+    } catch (e) {
+      assert(e instanceof AssertionError);
+      didThrow = true;
+    }
```
kt3k, line 632: nit: I think we need to have try-catch block for each assertObjectMatch call

kt3k:
@lowlighter Good catch. Thanks! Left one comment

ah-yu (commented):
It seems that the bug about non-plain object(see [lodash isPlainObject](https://lodash.com/docs/4.17.15#isPlainObject)) still exsits.
```ts
class Foo {}
class Bar { hello() {} }
assertObjectMatch( { x : new Foo() } , { x : new Bar() } ); // should fail but passes
```
I think we should replace `typeof value === "object"` with `isPlainObject(value)` to avoid filtering on non-plain objects(`RegExp` is a kind of non-plain object)

lowlighter:
@ah-yu But it seems to be the expected behaviour according to https://github.com/denoland/deno_std/pull/1419, seems that we strip out `constructor`

ah-yu:
@lowlighter It looks like the expected behaviour, thanks for the info!

kt3k (approved):
LGTM. Thank you for working on this! @lowlighter 

and thank you for your reviews! @ah-yu

PR #1970, opened 2022-02-28
feat: streams based `Buffer`

crowlKats:
Equivalent of `std/io/buffer.ts`'s `Buffer` but with streams

### DIFF
--- a/streams/buffer.ts
+++ b/streams/buffer.ts
@@ -1,0 +1,168 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { assert } from "../_util/assert.ts";
+import { copy } from "../bytes/mod.ts";
+
+const MAX_SIZE = 2 ** 32 - 2;
+const DEFAULT_CHUNK_SIZE = 16_640;
+
+/** A variable-sized buffer of bytes with `read()` and `write()` methods.
+ *
+ * Buffer is almost always used with some I/O like files and sockets. It allows
+ * one to buffer up a download from a socket. Buffer grows and shrinks as
+ * necessary.
+ *
+ * Buffer is NOT the same thing as Node's Buffer. Node's Buffer was created in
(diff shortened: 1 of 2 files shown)


streams/buffer.ts
```diff
@@ -0,0 +1,168 @@
...
+  /** The read only capacity of the buffer's underlying byte slice, that is,
+   * the total space allocated for the buffer's data. */
+  get capacity(): number {
+    return this.#buf.buffer.byteLength;
+  }
+
+  /** Discards all but the first `n` unread bytes from the buffer but
+   * continues to use the same allocated storage. It throws if `n` is
+   * negative or greater than the length of the buffer. */
+  truncate(n: number): void {
```
kt3k, line 92: truncate doesn't seem tested?

crowlKats: Done

bartlomieju:
What's the use case for this? Why create a new class with the same name instead of adding this functionality to existing `Buffer`?

crowlKats:
@bartlomieju well, we want various Deno.Reader/Writer based APIs to be written/ported to streams. These should be in their own file so there in no dependency on the aforementioned APIs

bartlomieju (approved):
LGTM

jsejcksn:
> What's the use case for this? Why create a new class with the same name instead of adding this functionality to existing `Buffer`?

@crowlKats Does this also allow for browser compatibility?

crowlKats:
@jsejcksn nope, once again, typescript is behind the times with outdated typings

jsejcksn:
> @jsejcksn nope, once again, typescript is behind the times with outdated typings

@crowlKats There's a tracking issue for compat issues caused by outdated native TS types [here](https://github.com/denoland/deno_std/issues/1973). Will you add any candidate modules there?

crowlKats:
@jsejcksn done

PR #1978, opened 2022-03-02
feat(streams): TextLineStream

crowlKats:
Equivalent of `LineStream`, but `string` based.
In response to whatwg/encoding#283


lucacasonato (approved):
LGTM. Good change IMO

PR #1984, opened 2022-03-03
feat(testing/asserts): use assertion signature for "assertStrictEquals"

jsejcksn:
In the same spirit as https://github.com/denoland/deno_std/pull/969, this PR retypes the signature for `assertStrictEquals` such that it can be used as a [type assertion function](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions). The nature of the runtime assertion is now reflected in the type system.

I also added a test with comments to demonstrate this change.

Please note that — because this has an effect on the type system — it is possible that new compiler errors will appear in code where they did not before (this is intended and desirable for type safety).

### DIFF
--- a/async/debounce_test.ts
+++ b/async/debounce_test.ts
@@ -45,12 +45,15 @@ Deno.test("[async] debounce: flushed", function () {
 
 Deno.test("[async] debounce: with params & context", async function () {
   const params: Array<string | number> = [];
-  const d = debounce(function (param1: string, param2: number) {
-    assertEquals(d.pending, false);
-    params.push(param1);
-    params.push(param2);
-    assertStrictEquals(d, this);
-  }, 100);
+  const d: DebouncedFunction<[string, number]> = debounce(
+    function (param1: string, param2: number) {
+      assertEquals(d.pending, false);
+      params.push(param1);
+      params.push(param2);
+      assertStrictEquals(d, this);
+    },
+    100,
+  );
   // @ts-expect-error Argument of type 'number' is not assignable to parameter of type 'string'.
   d(1, 1);
   d("foo", 1);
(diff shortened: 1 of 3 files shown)


jsejcksn:
> Please note that — because this has an effect on the type system — it is possible that new compiler errors will appear in code where they did not before (this is intended and desirable for type safety).

There was an example of this in an unrelated test file (which I updated in my most recent commit).

kt3k (approved):
LGTM. Nice!

kt3k:
I'm a bit worried about the timing to land this because this could break the users' code. Maybe let's land this for the next CLI minor release? @bartlomieju

jsejcksn:
> I'm a bit worried about the timing to land this because this could break the users' code. Maybe let's land this for the next CLI minor release?

@kt3k I don't mind this either way, but here are some observations:

- `deno.land/std` is currently versioned independently from the CLI:

    From [README#Releases](https://github.com/denoland/deno_std/blob/0.127.0/README.md#releases):

    > ## Releases
    >
    > Standard library is currently tagged independently of Deno version. This will
change once the library is stabilized.
    >
    > To check compatibility of different version of standard library with Deno CLI
see [this list](https://raw.githubusercontent.com/denoland/deno_website2/main/versions.json).

- According to the [SemVer spec](https://semver.org/spec/v2.0.0.html#spec-item-4), item 4 describes versions in the `0.y.z` range:
    > Major version zero (`0.y.z`) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

    Which means that every release before `1.0.0` should be considered a breaking change from the previous release.

bartlomieju:
> I'm a bit worried about the timing to land this because this could break the users' code. Maybe let's land this for the next CLI minor release? @bartlomieju

We already missed the merge window for `deno_std`. Let's just merge it after the release. If something goes wrong we can revert and release a hot fix

kt3k:
Ok. Let's land now

Issue #1966, opened 2022-02-25
Misleading assertion error in assertNotEquals


avihavai:
Hi, 

The function [assertNotEquals in /testing/asserts.ts has a misleading assertion error](https://github.com/denoland/deno_std/blob/97dc90373893b83713b05b1891a109ad48f8abc2/testing/asserts.ts#L311). Instead of using:

```
`actual: ${actualString} expected: ${expectedString}`;
```

The error should indicate that the actual string should not be the same as the expected string, e.g. 

```
msg = `actual: ${actualString} expected not to be: ${expectedString}`;
```

Really nice work on Deno by the way, have enjoyed using it a lot :1st_place_medal:.

bartlomieju:
PRs are welcome

iuioiua:
@bartlomieju, I believe this can be closed.

PR #1989, opened 2022-03-03
fix(testing): Misleading assertion error in assertNotEquals

getspooky:
Ref https://github.com/denoland/deno_std/issues/1966


lucacasonato (approved):
LGTM, thanks!

Issue #1986, opened 2022-03-03
tracking: Deno streams to web streams conversion


crowlKats:
There is also the separate effort to replace any code that uses various deno ns classes that implement `Deno.Reader` & `Deno.Writer` that still uses the `read` & `write` functions over the `readable` & `writable` properties.

### io
- [x] buffer.ts
  - [x] `Buffer` (#1970)
  - [x] `BufReader`: replaced by `ReadableStreamBYOBReader.read(, {atLeast})`
  - [x] `BufWriter`: unnecessary (doesnt make much sense for web streams)
  - [x] `readDelim`: replaced by `streams/delimiter.ts`'s `DelimiterStream`
  - [x] `readStringDelim` (#2006)
  - [x] `readLines`: replaced by `streams/delimiter.ts`'s `TextLineStream`

- [x] files.ts
  - [x] `readRange` replaced by `streams/byte_slice_stream.ts`'s `ByteSliceStream`

- [x] readers.ts
  - [x] `StringReader`: replaced by `TextDecoderStream`
  - [x] `MultiReader`: replaced by `streams/merge.ts`
  - [x] `LimitedReader` (#2007)

- [x] util.ts
  Do we even want these?
  - [x] `copyN`
  - [x] `readShort`
  - [x] `readInt`
  - [x] `readLong`

- [x] writers.ts
  unnecessary.

### archive
- [ ] `tar` (https://github.com/denoland/deno_std/pull/1985)

### encoding
- [x] binary.ts
  - [x] `readExact`: replaced by `ReadableStreamBYOBReader.read(, {atLeast})`
  - [x] `getNBytes`: unecessary
  - [x] `readVarnum`: doesnt this somewhat clash with `io/util.ts` functions?
  - [x] `readVarbig`
  - [x] `writeVarnum`
  - [x]  `writeVarbig`

- [x] csv.ts
  - [x] uses `BufReader` (#1993, #2491)

### log
- [x] handler.ts https://github.com/denoland/deno_std/pull/4021
  - [x] `WriterHandler`
  - [x] `RotatingFileHandler`

### mime
- [x] ~~`multipart`~~ mime was removed in #2336

### textproto
not sure what to do with this one. Update: textproto is removed in #2757

please let me know if I missed any.

iuioiua:
I'll give `readRange` from `files.ts` a crack, if that's still something we want. I'm thinking `RangedTransformStream`.

Also, update:
1. `std/mime` is deprecated
[...]

PR #1993, opened 2022-03-03
feat: web streams based encoding/csv

uki00a:
This PR tries to implement web streams based `encoding/csv` module.

I'm not very familiar with web streams, so any feedback would be appreciated!

Part of #1986

## Changes

- Added `encoding/csv/stream.ts`
- Added `encoding/csv/_io.ts` to remove duplicate codes between `encoding/csv.ts` and `encoding/csv/stream.ts`


encoding/csv/stream.ts
```diff
@@ -0,0 +1,51 @@
...
+    },
+  });
+}
+
+export interface CSVStreamOptions {
+  separator?: string;
+  comment?: string;
+}
+
+export class CSVStream {
```
crowlKats, line 34: You could rather make `CSVStream` extend `TransformStream`, then the only thing you have to do is `super()` with the transformstream options; and the variables you defined can just be private properties.

This isnt necessary, but imo just a lot cleaner & direct

encoding/csv/stream.ts
```diff
@@ -0,0 +1,51 @@
+import { Buffer, BufReader } from "../../io/buffer.ts";
```
crowlKats, line 1: Now, this is going to be the problem: any of the new streams based APIs are not allowed to depend on any Deno.Reader/Writer APIs. You'll have to wait for #1970 to land

crowlKats (changes requested):

encoding/csv/stream.ts
crowlKats: @uki00a #1970 landed, so this should be now doable. `BufReader` & `BufWriter` dont exist anymore now (as they dont make sense for web streams). However, i see that csv also depends on textproto, which hasnt been ported yet

uki00a: Rewrote `CSVStream` not to depend on `Deno.Reader` in [c573f87](https://github.com/denoland/deno_std/pull/1993/commits/c573f876d6bbf7e85e915b1b0c724594f0495ef9) :slightly_smiling_face:

lucacasonato (commented):
I think the `CSVStream` should be a `TransformStream<string, string[]>` rather than `TransformStream<Uint8Array, string[]>`. The text decoding can be done with `TextDecoderStream` by the user.

encoding/csv/_io.ts
```diff
@@ -0,0 +1,241 @@
...
+      this.message = `record on line ${line}: ${message}`;
+    } else if (start !== line) {
+      this.message =
+        `record on line ${start}; parse error on line ${line}, column ${column}: ${message}`;
+    } else {
+      this.message =
+        `parse error on line ${line}, column ${column}: ${message}`;
+    }
+  }
+}
```
crowlKats, line 236: wouldn't it make more sense to use syntaxerror? I know it doesnt have `startLine`, but creating new error types is a bit... eh. If deemed necessary to keep, then extend syntaxerror

crowlKats (commented):
except the nitpick, and luca's comment, this looks good

crowlKats (approved):
LGTM, thanks! the PR is still a draft; do you still have things to add? if not, i'd like to get this landed soon

uki00a:
@crowlKats Thanks for the review! I'd like to add more tests later in a separate PR, since the tests are a little low

PR #1995, opened 2022-03-05
chore: move streams/buffer.ts to io/buffer_stream.ts

crowlKats:
Since we are doing https://github.com/denoland/deno_std/issues/1986, it would be weird to have all the stream versions of things in the `streams` folder instead of the corresponding folder

### DIFF
--- a/io/buffer_stream.ts
+++ b/io/buffer_stream.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
 import { assert } from "../_util/assert.ts";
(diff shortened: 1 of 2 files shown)


nayeemrmn:
This would break `import { Buffer } from "...std/io/mod.ts";` assuming `io/buffer_stream.ts` was added there. You can rename it to `StreamBuffer` or something, but I think the better solution is to keep it in `streams` with the intention of `streams` slowly replacing `io`. We've mostly already gone in that direction.

crowlKats:
yea, thats true. will close this then.

PR #1997, opened 2022-03-05
chore: don't call deprecated _tools scripts in CI

kidonng:
Follow up to #1833


.github/workflows/ci.yml
```diff
@@ -113,12 +113,10 @@ jobs:
         uses: denoland/setup-deno@v1.0.0
 
       - name: Format
-        shell: bash
-        run: ./_tools/format.js --check
+        run: deno fmt --config deno.json --check
```
bartlomieju, line 116:
```suggestion
        run: deno fmt --check
```

.github/workflows/ci.yml
```diff
@@ -113,12 +113,10 @@ jobs:
...
 
       - name: Lint
-        shell: bash
-        run: ./_tools/lint.js
+        run: deno lint --config deno.json
```
bartlomieju, line 119:
```suggestion
        run: deno lint
```

bartlomieju (commented):
Config file is automatically looked up so we don't need to specify `--config` anymore. We should also remote the scripts from `_tools/` directory

kt3k (approved):

PR #2004, opened 2022-03-07
fix(node): fix http.request for minipass-fetch 2.x

### LINKED ISSUE #2001: compat: npm@8.5.3 doesn't work with compat mode
`npm install` works with `v8.5.2` with compat mode, but it doesn't with `v8.5.3`

diff between npm 8.5.2 and 8.5.3 https://github.com/npm/cli/compare/v8.5.2...v8.5.3

kt3k:
This PR fixes the call of `http.request` without `href` option.

`minipass-fetch@>2` calls `http.request` without `href` option ([this commit](https://github.com/npm/minipass-fetch/commit/f96f3b13e68f3851fd9fadb762c58f441a4c3f48) started it).

`npm` started depending on `minipass-fetch@2` indirectly at [this commit](https://github.com/npm/cli/commit/3c17b6965f0c5fffd5ac908388568a307466a73f), and that [broke](https://github.com/denoland/deno_std/runs/5417966994?check_suite_focus=true) the integration test of `npm install` in compat mode in CI

closes #2001

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -123,14 +123,15 @@ class ClientRequest extends NodeWritable {
 
     const client = await this._createCustomClient();
     const opts = { body: this.body, method: this.opts.method, client };
-    const mayResponse = fetch(this._createUrlStrFromOptions(this.opts), opts).catch((e) => {
-      if (e.message.includes("connection closed before message completed")) {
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, nice find

PR #2016, opened 2022-03-10
fix(io/readers): use an array as a MultiReader constructor parameter to avoid Maximum call stack size exceeded

### LINKED ISSUE #1954: `MultiReader` spread notation in constructor argument leads to "maximum call stack exceeded" errors for many readers
**Describe the bug**
The [`MultiReader` in `std/io`](https://deno.land/std@0.126.0/io/readers.ts#L17) has this constructor:
```ts
constructor(...readers: Deno.Reader[]) {
  this.readers = readers;
}
```
It's of course used like this (for example on [this line in `std/archive/tar.ts`](https://deno.land/std@0.126.0/archive/tar.ts#L450)):
```js
new MultiReader(...readers);
```
This means that if there are too many elements in `readers`, an error will be thrown like this:
```
error: Uncaught (in promise) RangeError: Maximum call stack size exceeded
    return new MultiReader(...readers);
           ^
    at Tar.getReader (https://deno.land/std@0.126.0/archive/tar.ts:450:12)
    at file:///home/joe/Desktop/tar_batch.js:22:20
```

**Steps to Reproduce**
Run the following command-line code to produce a directory that contains many files:
```
mkdir many_files && cd many_files && split -d -a10 -n100000 /dev/null file
```
Now run `nano script.js` and paste this:
```js
import { Tar } from "https://deno.land/std@0.126.0/archive/tar.ts";
import { copy } from "https://deno.land/std@0.126.0/streams/conversion.ts";

let tar = new Tar();
for(let i = 0; i < 100000; i++) {
  let filename = `file${i.toString().padStart(10, "0")}`;
  await tar.append(filename, {filePath:filename}); 
} 

const writer = await Deno.open(`./many_files.tar`, { write: true, create: true });
await copy(tar.getReader(), writer);
writer.close();
```
Save it, and then run this:
```
deno run --allow-read=. --allow-write=. script.js
```
It will produce the error message mentioned above.

**Expected behavior**
The `tar` module should be able to handle hundreds of thousands of files per `tar` archive. It seems like this would require changes/extensions to `MultiReader` so that an array can be passed directly, rather than using the spread operator.

**Environment**

- OS: Ubuntu 20.04
- deno version: 1.18.2
- std version: 0.126.0

**Additional details**
As a temporary/hacky fix, I've made an edited version of `std/archive/tar.ts` and `std/io/reader.ts` [here](https://github.com/josephrocca/scrap/blob/main/deno_tar_fixed/mod.ts) which fixes the issue by just removing the spread from the `MultiReader` constructor and usage:
```js
import { Tar } from "https://raw.githubusercontent.com/josephrocca/scrap/main/deno_tar_fixed/tar.ts";
```

TrickyPi:
Closes #1954

### DIFF
--- a/archive/tar.ts
+++ b/archive/tar.ts
@@ -447,7 +447,7 @@ export class Tar {
 
     // append 2 empty records
     readers.push(new Buffer(clean(recordSize * 2)));
-    return new MultiReader(...readers);
+    return new MultiReader(readers);
   }
 }
 
--- a/io/readers.ts
+++ b/io/readers.ts
@@ -18,8 +18,8 @@ export class MultiReader implements Deno.Reader {
   private readonly readers: Deno.Reader[];
   private currentIndex = 0;
 
-  constructor(...readers: Deno.Reader[]) {
-    this.readers = readers;
+  constructor(readers: Deno.Reader[]) {
+    this.readers = [...readers];
   }
 
   async read(p: Uint8Array): Promise<number | null> {
--- a/io/readers_test.ts
+++ b/io/readers_test.ts
@@ -29,7 +29,7 @@ Deno.test("ioStringReader", async function () {
 });
 
 Deno.test("ioMultiReader", async function () {
-  const r = new MultiReader(new StringReader("abc"), new StringReader("def"));
+  const r = new MultiReader([new StringReader("abc"), new StringReader("def")]);
   const w = new StringWriter();
   const n = await copyN(r, w, 4);
   assertEquals(n, 4);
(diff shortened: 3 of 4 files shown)


kt3k:
I'm in favor of this change as this fixes the actual use case of `Tar`.

Though this is technically a breaking change, this particular function doesn't seem used that much ([github search result](https://github.com/search?l=TypeScript&q=MultiReader&type=Code))

Any concerns on landing this?

kt3k (approved):
LGTM

PR #2020, opened 2022-03-11
feat(node/_fs): Add `fs.writevSync(fd, buffers[, position])`

getspooky:
#648


kt3k:
Nice!

We also need to export it in `node/fs.ts` to pass the test cases

getspooky:
@kt3k Done!

kt3k (approved):
@getspooky LGTM. Nice work!

PR #2021, opened 2022-03-11
chore: Use deno task

bartlomieju:


bartlomieju:
Blocked by https://github.com/denoland/deno/pull/13918

README.md
```diff
@@ -66,10 +66,11 @@ After cloning don't forget to `git submodule update --init`.
 
 Before opening a PR make sure to:
 
-- Add tests that cover your changes.
-- `deno test --unstable --allow-all` passes.
-- `deno run -A _tools/format.js` passes without changing files.
-- `deno run -A _tools/lint.js` passes.
+- have the latest Deno version installed locally
+- add tests that cover your changes.
+- `deno task test:canary` passes.
```
dsherret, line 71: Should this just be `deno task test`? Looks like people won't necessarily be running canary (though, maybe it would be neat if it fetched the latest canary).

node/README.md
```diff
@@ -109,6 +109,8 @@ file, this will pull the configured tests in and then add them to the test
 workflow.
 
 ```zsh
+$ deno task node:setup
+// or
 $ deno run --allow-read --allow-net --allow-write node/_tools/setup.ts
```
dsherret, line 114: I think we should get rid of this because it's something that could get out of date with `deno task node:setup`. People should be using the latest deno if they're contributing.

README.md
bartlomieju: It's somewhat required to run on canary

kt3k (approved):
LGTM. Great! 🎉

kt3k (approved):
I think we should try this now

deno.json
```diff
@@ -32,5 +32,13 @@
         "wasi/testdata"
       ]
     }
+  },
+  "tasks": {
+    "test": "deno test --doc --unstable --allow-all --coverage=./cov --ignore=node/",
+    "test:faketime": "faketime '2021-12-31' deno test --unstable --allow-all datetime/test_2021_12_31.ts",
+    "test:browser": "git grep --name-only \"// This module is browser compatible.\" | grep -v \".github/workflows\" | xargs deno cache --config browser-compat.tsconfig.json",
```
dsherret, line 39: `xargs` doesn't work on Windows. Maybe we should add it to `deno task`?

dsherret: Actually, it would probably be better if we moved this to a deno script.

bartlomieju: If you feel like it, then sure :)

bartlomieju: I don't think anyone runs this command locally, but I might be wrong

Issue #1991, opened 2022-03-03
Add test_suite module to deno_std to help bridge a gap in testing experience between Deno and Node


bartlomieju:
### Discussed in https://github.com/denoland/deno_std/discussions/1942

<div type='discussions-op-text'>

<sup>Originally posted by **KyleJune** February 20, 2022</sup>
In the [discord](https://discord.com/channels/684898665143206084/688040863313428503/944799764397056020), @bartlomieju suggested opening an issue here about possibly adding my third party module to deno_std to help bridge a gap in testing experience between Deno and Node.

The [test_suite](https://deno.land/x/test_suite) module is not a direct port of Jasmine, Jest, or Mocha but has similar functionality. The `describe` and `it` functions take a test definition that is an extension of Deno.TestDefinition. For the `describe` and `it` functions, I copied all the call signatures Deno has for `Deno.test` and `t.step`.

I implemented the `only` option for nested tests which you currently cannot do with the `t.step`.

It depends on 1 third party module I wrote called [collections](https://deno.land/x/collections). It just uses the `Vector` class from there instead of `Array` to make it easier to remove steps when a new test step is focused. It wouldn't be difficult to re-write the parts that use `Vector` to not use it.

The test coverage depends on another third party module I wrote called [mock](https://deno.land/x/mock). It is not a direct port of sinon.js but is similar. I wrote it from scratch in TypeScript for Deno because when I originally started using Deno, I wasn't able to get sinon.js to work.

I have [92.90% test coverage](https://app.codecov.io/gh/udibo/test_suite/branch/main) currently. For getting rid of the third party module dependency, they would need to be re-written. Currently the testing works by mocking `Deno.test` and creating a fake `TestContext`. @bartlomieju said he thinks this could be verified by integration tests - spawning a subprocesses and checking if output matches instead of using mocking.

More information and examples can be found in the readme [here](https://deno.land/x/test_suite).

The interface is documented using [Deno Doc](https://doc.deno.land/https://deno.land/x/test_suite/mod.ts).</div>

kt3k:
I'm in favor of this proposal. I recently ported many npm modules from crypto-browserify to std, and converting all `describe`/`it` to `Deno.test`/`await t.step` was very painful work. If this module is available in std, I think it should promote the migration of node.js code to deno-first project.

(links to past similar discussions: #1389, #931, #1831)

[...]

PR #2022, opened 2022-03-12
feat(collections): Add BinaryHeap

KyleJune:
Binary heaps are common data structures in other languages but JavaScript does not have a built in BinaryHeap class. In Rust, they have a binary heap in std/collections that can be found [here](https://doc.rust-lang.org/stable/std/collections/struct.BinaryHeap.html). In Java they have one too but call it [PriorityQueue](https://docs.oracle.com/javase/7/docs/api/java/util/PriorityQueue.html). I thought it would be handy if Deno had one too. I decided to go with the name BinaryHeap to match Rust's BinaryHeap.

Originally I implemented this in a third party module called [collections](https://deno.land/x/collections). In an issue for adding my [test_suite](https://deno.land/x/test_suite) module to std/testing, the possibility of adding my [mock](https://deno.land/x/mock) module came up too. The mock module makes use of a Red Black Tree that is part of my collections module and @bartlomieju said he would be open to adding some of my other structures to std/collections too.

https://github.com/denoland/deno_std/issues/1991#issuecomment-1061315245
> Correct, our implementation of RB tree is internal and can't be exposed. I'm also open to adding this module to `std/collections` as well as other structures you have there. They seem very handy +1

As a first step I thought it would be better to start small and just add BinaryHeap first before attempting to add my red black tree implementation.

Below are links to the deno docs for the 2 files added as part of this PR.

- [binary_heap.ts](https://doc.deno.land/https://raw.githubusercontent.com/KyleJune/deno_std/kj-add-binary-heap/collections/binary_heap.ts)
- [comparators.ts](https://doc.deno.land/https://raw.githubusercontent.com/KyleJune/deno_std/kj-add-binary-heap/collections/comparators.ts)


collections/binary_heap_test.ts
```diff
@@ -0,0 +1,269 @@
...
+import { BinaryHeap } from "./binary_heap.ts";
+import { ascend, descend } from "./comparators.ts";
+
+export class MyMath {
+  multiply(a: number, b: number): number {
+    return a * b;
+  }
+}
+
+export interface Container {
```
bartlomieju, line 11: These two don't need to be `export`ed

collections/comparators.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/** Compares its two arguments for ascending order using JavaScript's built in comparison operators. */
+export function ascend<T>(a: T, b: T) {
+  return a < b ? -1 : a > b ? 1 : 0;
+}
+
+/** Compares its two arguments for descending order using JavaScript's built in comparison operators. */
+export function descend<T>(a: T, b: T) {
+  return a < b ? 1 : a > b ? -1 : 0;
+}
```
bartlomieju, line 11: I think this module is not needed; if it is used in `binary_heap.ts` we could just define those inline in that module.

bartlomieju (commented):
I'm in favor of landing this PR. @kt3k any objections?

collections/binary_heap_test.ts
KyleJune: My bad. I originally had these in a different file called test_common.ts which I use in multiple test files of mine. I forgot to remove the export when I copied it into collections/binary_heap_test.ts. If I'm going to be using these objects again in my binary search tree and red black tree test coverage, should I inline them like I did here or create a test data folder for storing these classes used in multiple test cases?

collections/comparators.ts
KyleJune: When I add BSTree and RBTree in separate PRs, they will use these comparators too for determining how to sort the values added to the tree. That's why I put them in their own file. These functions work with `Array.prototype.sort` too. I'm fine with moving them too though. Having them at least somewhere is useful to avoid having to re-implement these functions whenever you want to change the default sort behavior of a data structure or sort an existing one.

KyleJune: Here is a usage example for making the BinaryHeap be a Min Heap instead of a Max Heap.

```ts
import { BinaryHeap, ascend } from "https://deno.land/std@0.129.0/collections/mod.ts";
const maxHeap = new BinaryHeap<number>(); // descend is default compare function.
const minHeap = new BinaryHeap<number>(ascend);
```

KyleJune: Another option would be to keep them in their own separate file but also update binary_heap.ts to re-export them for easy access, that way you could import them all from binary_heap.ts instead of importing the entire mod.ts or having 2 separate import lines.

kt3k:
@bartlomieju I'm in favor of including these kinds of data structures in general. Should be useful for algorithmic works. Will review in more details soon

collections/binary_heap_test.ts
kt3k: We loosely have the rule in deno_std that if the filename started with `_`, then that file is considered private. So maybe we can define `collections/_test_utils.ts` if we need some common items only used for testing.

collections/comparators.ts
kt3k: > Another option would be to keep them in their own separate file but also update binary_heap.ts to re-export them for easy access

This sounds better option to me. And also I think it's better to prefix this module as `_comparators.ts` to prevent the direct dependency to this file.

collections/binary_heap_test.ts
KyleJune: Done, I've moved those 2 classes to `collections/_test_utils.ts`.

collections/comparators.ts
KyleJune: I added an underscore prefix to the `comparators.ts` and `comparators_test.ts` as requested and updated `binary_heap.ts` to export them.

I've also updated the README.md file to have a section for BinaryHeap, showing the efficiency of the functions. I included an example of both min and max heaps along with showing how the comparator functions can be used to compare properties of the values in a BinaryHeap.

bartlomieju (approved):
LGTM, thank you for this @KyleJune. I'll wait for @kt3k review before landing.

kt3k (approved):
@KyleJune LGTM. Great!

PR #2024, opened 2022-03-13
fix(node): Fix dynamic require

YXL76:


kt3k (approved):
LGTM. Thanks! @YXL76

kt3k:
Note:
- isEventTarget is actually a little different from `instanceof EventTarget` in node.js ([ref](https://github.com/nodejs/node/blob/625c1e841adee3e687225597e60c70df6510a696/lib/internal/event_target.js#L893)), but the current code doesn't make sense already. So this shouldn't break anything.
- The change in `node/internal/readline/interface.mjs` should be no problem because the circular dependency in std/node is checked in `lint` job in CI. So we don't have circular deps here and no need of lazily initialize `Readable` here.

PR #2028, opened 2022-03-14
feat(testing): add `assertInstanceOf`

### LINKED ISSUE #2025: testing/asserts: `assertInstanceOf()`

I would like to assert if an object is an instance of a specific type, but using `asserts()` does not cause TypeScript to know about the new type of a variable:

```ts
import {assert} from "https://deno.land/std@0.129.0/testing/asserts.ts";

const maybeBlob = new Blob() as unknown;
assert(maybeBlob instanceof Blob, "variable is not of type 'Blob'");

console.log(maybeBlob.byteLength);
//                    ^^^^^^^^^^---- Property 'byteLength' does not
//                                   exist on type 'unknown'.
```

A workaround is to not use `assert()` and simply check for `instanceof` in an `if` statement and throw manually if it's not an instance, but this is cumbersome if you have to do it many times.

`assertInstanceOf(object, type)` that both checks if `object instancoef type` and uses [TypeScript assertion functions](https://www.typescriptlang.org/play#example/assertion-functions) to let TypeScript know the variable now has a certain type. 

Alternative names:
- `assertInstanceOf()`
- `assertInstance()`
- `assertIsInstance()` (from [python](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance))

I'm currently using my own helper function for this, which is what I'll be using instead for now.

This is what I'm using:
```js
/**
 * @param {unknown} obj
 * @param {Function} Type
 * @returns {asserts obj is typeof Type}
 */
export function assertInstanceOf(obj, Type, msg = "") {
	if (!msg) msg = `${obj} is not an instance of ${Type.name}`;
	assert(obj instanceof Type, msg);
}
```
Though this is JavaScript with JSDoc, I tried converting it to TypeScript but couldn't get it to work right away. For some reason with TypeScript all assertions seem to be be converting variables to `Function`. Supposedly because of https://github.com/microsoft/TypeScript/pull/14907

jespertheend:
Closes #2025

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -449,22 +449,22 @@ new (...args: any) => infer C ? C
  * If not then throw.
  */
 export function assertInstanceOf<T extends AnyConstructor>(
-  obj: unknown,
-  type: T,
+  actual: unknown,
+  expectedType: T,
   msg = "",
-): asserts obj is GetConstructorType<T> {
+): asserts actual is GetConstructorType<T> {
   if (!msg) {
-    const expectedTypeStr = type.name;
+    const expectedTypeStr = expectedType.name;
 
     let actualTypeStr = "";
-    if (obj === null) {
+    if (actual === null) {
       actualTypeStr = "null";
-    } else if (obj === undefined) {
+    } else if (actual === undefined) {
       actualTypeStr = "undefined";
-    } else if (typeof obj === "object") {
-      actualTypeStr = obj.constructor.name;
+    } else if (typeof actual === "object") {
+      actualTypeStr = actual.constructor.name;
     } else {
-      actualTypeStr = typeof obj;
+      actualTypeStr = typeof actual;
     }
 
     if (expectedTypeStr == actualTypeStr) {
@@ -477,7 +477,7 @@ export function assertInstanceOf<T extends AnyConstructor>(
         `Expected object to be an instance of "${expectedTypeStr}" but was "${actualTypeStr}".`;
     }
   }
-  assert(obj instanceof type, msg);
+  assert(actual instanceof expectedType, msg);
 }
 
 /**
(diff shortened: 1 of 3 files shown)


testing/asserts.ts
```diff
@@ -438,6 +438,48 @@ delta "${f(delta)}" is greater than "${f(tolerance)}"`,
   );
 }
 
+// deno-lint-ignore no-explicit-any
+type AnyConstructor = new (...args: any[]) => any;
+type GetConstructorType<T extends AnyConstructor> = T extends // deno-lint-ignore no-explicit-any
+new (...args: any) => infer C ? C
+  : never;
```
jespertheend, line 445: The formatting here is rather weird, but afaik this really needs `any`. I tried `unknown` but that causes issues when passing constructors with parameters.

testing/asserts.ts
```diff
@@ -438,6 +438,48 @@ delta "${f(delta)}" is greater than "${f(tolerance)}"`,
...
+  if (!msg) {
+    const expectedTypeStr = expectedType.name;
+
+    let actualTypeStr = "";
+    if (actual === null) {
+      actualTypeStr = "null";
+    } else if (actual === undefined) {
+      actualTypeStr = "undefined";
+    } else if (typeof actual === "object") {
+      actualTypeStr = actual.constructor.name;
```
kt3k, line 465: This can throw when `actual` is created like `Object.create(null)`. Maybe let's use optional chaining and default value?

kt3k (commented):
I'm in favor of adding this feature. The capability of narrowing types looks useful for writing tests

testing/asserts.ts
jespertheend: Ah good call, should be taken care of now. I've made it default to `"Object"`.

kt3k (approved):
LGTM

Does anyone have any concern about including this feature?

bartlomieju (approved):
LGTM, @jespertheend please open a PR to `manual` repo that updates `testing` chapter with this assertion.

jespertheend:
Ah sorry, I forgot to link to it: https://github.com/denoland/manual/pull/248

PR #2032, opened 2022-03-16
feat(node): add NodeEventTarget

### LINKED ISSUE #1865: node(events): add internal/event_target (or equivalent)
Currently `node/_events.js` uses `require("internal/event_target")` but this file does not exist.
https://github.com/denoland/deno_std/blob/1c8e6ce506267edb3ce64ce667d9611161b35a42/node/_events.js#L790-L815
This function is used in several places e.g. tests copied from Node.js source. In order to convert it to static imports, we need to get this file from Node or wrap Deno's `EventTarget` and `EventEmitter`.

kt3k:
closes #1865


bartlomieju (approved):
LGTM 👍

Issue #2013, opened 2022-03-08
compat: postinstall fails for some modules during `npm install`


kt3k:
When the npm module has `postinstall` script in package.json, then `npm install` with compat mode stops with errors.

Known npm modules which cause this error:
- `es5-ext`
- `core-js-pure`

bartlomieju:
Right... `postinstall` script. We need to find some sane solution for that.

kt3k:
`es5-ext` has `node -e \"try{require('./_postinstall')}catch(e){}\"` in `postinstall`
https://github.com/medikoo/es5-ext/blob/406507807f96aa092ee252ab5fe759fc0dd4b778/package.json#L115

and that command passed to `child_process.spawn`.

`core-js-pure` also has the same postinstall script https://github.com/zloirock/core-js/blob/6eb66ebbcd9c50a8f337c87992076b51a4beecec/packages/core-js-pure/package.json#L55

kt3k:
dirty solution could be parsing the input at the beginning of `child_process.spawn`, and change the behavior if the input is `node` command with `-e` option...

kt3k:
- `_postinstall.js` script in ex5-ext https://github.com/medikoo/es5-ext/blob/edf9fbac6c0d72a6192de7e0fbfb022eb34c3ac7/_postinstall.js
- `postinstall.js` script in core-js https://github.com/zloirock/core-js/blob/master/packages/core-js/postinstall.js

These scripts seem technically irrelevant to make them working (political messages and donation requests). Maybe ignoring these postinstall scripts by hard coding would be a possible option

kt3k:
postinstall issue is tracked in https://github.com/denoland/deno/issues/16164

PR #2033, opened 2022-03-16
test: re-enable npm testing

kt3k:
This PR enables `npm install` testing again.

The problem was cause by the specific module in the dependencies of `gulp` (the issue is tracked in https://github.com/denoland/deno_std/issues/2013)

This PR changes the target module to `yargs`, and avoids the above problem.


bartlomieju (approved):
LGTM

PR #2034, opened 2022-03-16
feat(async): expose asyncPromise and asyncAsyncIterable

### LINKED ISSUE #2031: Exposing `abortablePromise` and `abortableAsyncIterable` directly.
### Discussed in https://github.com/denoland/deno_std/discussions/2030

<div type='discussions-op-text'>

<sup>Originally posted by **harrysolovay** March 15, 2022</sup>
The [`abortablePromise`](https://github.com/denoland/deno_std/blob/main/async/abortable.ts#L23) and [`abortableAsyncIterable`](https://github.com/denoland/deno_std/blob/main/async/abortable.ts#L38) utilities are exposed only through [`abortable`](https://github.com/denoland/deno_std/blob/main/async/abortable.ts#L8), which wraps the two. Might it make sense to export them directly? In my situation, I need only `abortablePromise`, and would prefer to not include the code of `abortableAsyncIterable`. Thoughts would be greatly appreciated!</div>

kt3k:
closes #2031

### DIFF
--- a/async/README.md
+++ b/async/README.md
@@ -47,6 +47,50 @@ for await (const item of abortable(p(), c.signal)) {
 }
 ```
 
+## abortablePromise
+
+`abortablePromise` takes the promise and `AbortSignal` and returns the
+cancelable version of the promise.
+
+```typescript
+import { abortablePromise } from "https://deno.land/std@$STD_VERSION/async/mod.ts";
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

PR #2035, opened 2022-03-16
fix(node/process): Fix `EventEmitter` methods

YXL76:
1. Add `process.kill`.
2. Fix `addListener` and `removeListener`, they should be alias for `on` and `off`.
3. Override `emit` and `prependListener` to handle signal events.

The `once` and `prependOnceListener` also need to override. But I think we should export `_onceWrap` first.

https://github.com/denoland/deno_std/blob/8e01ec6cb7e00eb82c0a34fbff3a748ca48dd6aa/node/_events.mjs#L506-L512

### DIFF
--- a/node/_process/process.ts
+++ b/node/_process/process.ts
@@ -98,36 +98,3 @@ export const versions = {
   unicode: "13.0",
   ...Deno.version,
 };
-
-function hrtime(time?: [number, number]): [number, number] {
-  const milli = performance.now();
-  const sec = Math.floor(milli / 1000);
-  const nano = Math.floor(milli * 1_000_000 - sec * 1_000_000_000);
-  if (!time) {
-    return [sec, nano];
-  }
-  const [prevSec, prevNano] = time;
-  return [sec - prevSec, nano - prevNano];
-}
-
-hrtime.bigint = function (): BigInt {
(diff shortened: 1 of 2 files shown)


node/_process/process.ts
```diff
@@ -98,36 +98,3 @@ export const versions = {
   unicode: "13.0",
   ...Deno.version,
 };
-
-function hrtime(time?: [number, number]): [number, number] {
```
kt3k, line 102: Good catch! These are dead code

node/process.ts
```diff
@@ -231,6 +232,27 @@ memoryUsage.rss = function (): number {
...
+    try {
+      Deno.kill(pid, sig);
+    } catch (e) {
+      if (e instanceof TypeError) {
+        throw new ERR_UNKNOWN_SIGNAL(sig);
+      }
+      throw e;
+    }
+  } else {
+    throw new ERR_UNKNOWN_SIGNAL(sig.toString());
```
kt3k, line 250: nit: It might be better to make this `notImplemented()` error (especially when the signal is a number)

kt3k (approved):
LGTM. Thanks for your contribution! @YXL76

PR #2037, opened 2022-03-16
feat(node): add Console constructor

### LINKED ISSUE #2027: node: require("console").Console is missing
[Console](https://nodejs.org/api/console.html#new-consolestdout-stderr-ignoreerrors) constructor is missing in `console` module.

Console constructor is used in [fancy-log](https://www.npmjs.com/package/fancy-log) module, which is used in most gulp related projects.

kt3k:
This PR adds Node.js `Console` constructor in `console` module.

`gulp` uses `Console` constructor for logging ([ref](https://github.com/gulpjs/fancy-log/blob/6686d1784d11db4ccc4d91ad854e580a14ebe330/index.js)). This should improve the compatibility of gulp and its plugins.

closes #2027


bartlomieju (approved):
LGTM, nice to see so many tests enabled

PR #2039, opened 2022-03-17
feat(testing): Implement "assertSnapshot"

hyp3rflow:
Closes https://github.com/denoland/deno/issues/3635

based on PR https://github.com/denoland/deno/pull/14007


hyp3rflow:
Oh.. How can I convert this PR to draft? I created this just for showing the naive implementation of `assertSnapshot` on this PR https://github.com/denoland/deno/pull/14007. but I accidentally clicked `Create Pull Request` instead of draft one :q

kt3k:
converted to draft

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+
+  if (!snapshotFile) {
+    const snapshotPath = getSnapshotPath();
+    await ensureFile(snapshotPath);
+    const file = Deno.readTextFileSync(snapshotPath);
+    snapshotFile = file ? JSON.parse(file) : {};
+  }
+  const snapshot = snapshotFile?.[name];
+  if (!context.update) {
+    assertEquals(actual, snapshot);
```
bcheidemann, line 886: The current implementation will error on:

```ts
const actual = {
  fn: () => { /* ... */ }
}
```

As snapshots need to be represented as strings anyway, I would argue it makes more sense to use the `_format` helper here to convert the snapshot into a string. We could store the snapshot like this:

```json
{
  "Snapshot Test > babo > merong": [
    "{",
    "  b: 2,",
    "  c: 3,",
    "}"
  ]
}
```

This would enable us to catch a much wider range of errors in our snapshots. E.g. the following would error on `assertSnapshot`:

```ts
// First run
const actual = {
  testClass: class Test {}
}
// --snip--
assertSnapshot(test, actual)

// Second run
const actual = {
  testClass: class NotTest {}
}
// --snip--
assertSnapshot(test, actual)
```

But would not error (or worse, always error) if we only store the JSON representation.

Additionally, this representation enables us to use the `diff` and `buildMessage` to produce helpful output.

bcheidemann (changes requested):
Obviously this is just a starting point so hopefully I've not been overzealous with my comments! :D

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
 export function unreachable(): never {
   throw new AssertionError("unreachable");
 }
+
+let snapshotFile: Record<string, unknown> | undefined = undefined;
+const updatedSnapshotFile: Record<string, unknown> = {};
+// const snapshotMap: Record<string, number> = {};
+
+export async function assertSnapshot(context: Deno.TestContext, actual: unknown) {
```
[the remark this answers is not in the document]
hyp3rflow: Good! I think we can implement that after https://github.com/denoland/deno/pull/14007 merged.

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+  const name = getName(context);
+  // const count = getCount(name);
+
+  if (!snapshotFile) {
+    const snapshotPath = getSnapshotPath();
+    await ensureFile(snapshotPath);
+    const file = Deno.readTextFileSync(snapshotPath);
```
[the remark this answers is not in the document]
hyp3rflow: ✅ [31abdb8](https://github.com/denoland/deno_std/pull/2039/commits/31abdb8ec8960b2d3622416283fa868972d08a78)

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+
+  function getName(context: Deno.TestContext): string {
+    if (context.parent) return `${getName(context.parent)} > ${context.name}`;
+    return context.name;
+  }
+
+  function getSnapshotPath() {
+    const testFile = fromFileUrl(context.origin);
+    const parts = parse(testFile);
+    return `${parts.dir}/${parts.name}.snap`;
```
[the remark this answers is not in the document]
hyp3rflow: ✅ [31abdb8](https://github.com/denoland/deno_std/pull/2039/commits/31abdb8ec8960b2d3622416283fa868972d08a78)

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+  }
+
+  async function writeSnapshotFile() {
+    const snapshotPath = await getSnapshotPath();
```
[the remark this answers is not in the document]
hyp3rflow: ✅ [31abdb8](https://github.com/denoland/deno_std/pull/2039/commits/31abdb8ec8960b2d3622416283fa868972d08a78)

testing/asserts.ts
```diff
@@ -864,3 +866,59 @@ export function unimplemented(msg?: string): never {
...
+  } else {
+    let isEqual = true;
+    try {
+      assertEquals(actual, snapshot);
+    } catch {
+      isEqual = false;
+    }
+    if (!isEqual) console.info("Snapshot updated", name);
+  }
+  updatedSnapshotFile[name] = actual;
```
kt3k, line 896: Does this mean this implementation supports only 1 snapshot per test case?

hyp3rflow: Does the test case mean one `Deno.test`?
This implementation supports 1 snapshot per test module (one .test.ts file)

for example, if I test `a.test.ts` with `--update` arg,
```typescript
Deno.test("Snapshot Test A", async (t) => {
  await assertSnapshot(t, { a: 1, b: 2 });
  await t.step("yaho", async (t) => {
    await assertSnapshot(t, { b: 2, c: 3 });
    await t.step("haha", async (t) => {
      await assertSnapshot(t, { b: 2, c: 4 });
    });
  });
});

Deno.test("Snapshot Test A2", async (t) => {
  await assertSnapshot(t, { a: 1, b: 2 });
  await t.step("what", async (t) => {
    await assertSnapshot(t, { b: 2, c: 3 });
    await t.step("good", async (t) => {
      await assertSnapshot(t, { b: 2, c: 4 });
    });
  });
});
```

the output, `a.test.snap` will be like this
```json
{
  "Snapshot Test A": {
    "a": 1,
    "b": 2
  },
  "Snapshot Test A > yaho": {
    "b": 2,
    "c": 3
  },
  "Snapshot Test A > yaho > haha": {
    "b": 2,
    "c": 4
  },
  "Snapshot Test A2": {
    "a": 1,
    "b": 2
  },
  "Snapshot Test A2 > what": {
    "b": 2,
    "c": 3
  },
  "Snapshot Test A2 > what > good": {
    "b": 2,
    "c": 4
  }
}
```

kt3k: How about the below case? I think the second `assertSnapshot` call overwrites the snapshot from the first call

```js
Deno.test("Snapshot Test A", async (t) => {
  await assertSnapshot(t, { a: 1, b: 2 });
  await assertSnapshot(t, { c: 3, d: 4 });
});
```

hyp3rflow: Aha! Sorry, now I know what you mean. That case will be handled with commented count. (Currently not implemented perfectly so I commented) @kt3k

kt3k: Ah, ok. I see. `getCount` does that 👍

testing/asserts.ts
```diff
@@ -882,7 +882,7 @@ export async function assertSnapshot(context: Deno.TestContext, actual: unknown)
     snapshotFile = file ? JSON.parse(file) : {};
   }
   const snapshot = snapshotFile?.[name];
-  if (!context.update) {
+  if (!Deno.args.includes('--update')) {
```
bcheidemann, line 885:
```suggestion
  if (!Deno.args.some(arg => arg === '--update' || arg === '-u')) {
```

Could do something like this to support `-u` and `--update`.

hyp3rflow: I will handle this after https://github.com/denoland/deno/pull/14007 merged!

testing/asserts.ts
hyp3rflow: I think [Deno.inspect()](https://doc.deno.land/deno/stable/~/Deno.inspect) can be a good choice for serialization method. What do you think? @bcheidemann

bcheidemann: Yeah this would definitely do the job. I would suggest to use the [`_format` helper](https://github.com/denoland/deno_std/blob/0d731bc7aeb6730d4c18da140c89c7d24651ba91/testing/asserts.ts#L31) which wraps `Deno.inspect`. This ensures the output will be consistent with other asserts 🙂

testing/asserts.ts
kt3k: BTW to pass `-u` option to `Deno.args`, you need to invoke test command like `deno test -- -u` because the flags before `--` are consumed by Deno CLI itself. I'm ok with that for the first pass implementation, but it might be non ideal solution, I guess

hyp3rflow: Right. I also noticed that the flag for updating snapshot can affect other codes that wanted to test.. but I think that using `Deno.args` is better than making an option on Deno test cli.

bcheidemann: Imo this is the best solution for first pass. But if/when snapshot testing is a more established then we should look to implement this on the Deno CLI instead; as @hyp3rflow mentioned, this can affect test execution.

hyp3rflow:
I changed the snapshot style to jest's snapshot module.

bartlomieju:
@hyp3rflow I just landed PR in Deno repo that adds necessary fields to `Deno.TestContext`

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+    const parts = parse(testFile);
+    return `${join(parts.dir, parts.name)}.snap`;
+  }
+  function writeSnapshotFileSync() {
+    const snapshotPath = getSnapshotPath();
+    ensureFileSync(snapshotPath);
+    const buf = ['export const snapshot = {};\n'];
+    for (const [key, value] of Object.entries(updatedSnapshotFile)) {
+      buf.push(`\nsnapshot[\`${key}\`] = \`\n${_format(value)}\n\`;\n`);
+    }
```
bcheidemann, line 919: This is nice because you could easily automate conversion of Jest snapshots to Deno snapshots.

However, we will need to escape backticks (`) because otherwise the following code would break:

```ts
await assertSnapshot(context, "This snapshot will break because I added a ` ");
```

The same will be true of the snapshot name.

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
 }
+
+let snapshotFile: Record<string, string> | undefined = undefined;
+const updatedSnapshotFile: Record<string, unknown> = {};
+const snapshotMap: Record<string, number> = {};
+
+export async function assertSnapshot(context: Deno.TestContext, actual: unknown) {
+  const name = getName(context);
+  const count = getCount();
+  const testName = `${name} #${count}`;
```
bcheidemann, line 877:
```suggestion
  const testName = `${name} ${count}`;
```

This makes it consistent with Jest snapshots, which don't have a #.

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+  if (isUpdate) {
+    let isEqual = true;
+    try {
+      assertEquals(_format(actual), snapshot.slice(1, -1));
+    } catch {
+      isEqual = false;
+    }
+    if (!isEqual) console.info("Snapshot updated", testName);
+    updatedSnapshotFile[testName] = actual;
+    globalThis.onunload = writeSnapshotFileSync;
```
bcheidemann, line 895: Seems like this function will be run once per `assertSnapshot` call. Can we refactor this so it only runs once?

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+    const snapshotPath = getSnapshotPath();
+    await ensureFile(snapshotPath);
+    const {snapshot} = await import(snapshotPath);
+    snapshotFile = snapshot;
+  }
+  const snapshot = snapshotFile?.[testName] ?? '';
+  if (isUpdate) {
+    let isEqual = true;
+    try {
+      assertEquals(_format(actual), snapshot.slice(1, -1));
```
bcheidemann, line 889: Given these are reused in both the `if` and `else` blocks, can we extract these out above the `if` statement for better readability/maintainability e.g.

```ts
// --snip--
const _actual = _format(actual);
const _expected = snapshot.slice(1, -1);
if (isUpdate) {
 // --snip--
```

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+  }
+  function writeSnapshotFileSync() {
+    const snapshotPath = getSnapshotPath();
+    ensureFileSync(snapshotPath);
+    const buf = ['export const snapshot = {};\n'];
+    for (const [key, value] of Object.entries(updatedSnapshotFile)) {
+      buf.push(`\nsnapshot[\`${key}\`] = \`\n${_format(value)}\n\`;\n`);
+    }
+    Deno.writeTextFileSync(snapshotPath, buf.join(""));
+    console.log('Snapshot updated!');
```
bcheidemann, line 921:
```suggestion
    console.log(green(bold(` > ${n} snapshots updated.`)));
```

Maybe we could mimic Jest here and log how many snapshots were updated?

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+  }
+  const snapshot = snapshotFile?.[testName] ?? '';
+  if (isUpdate) {
+    let isEqual = true;
+    try {
+      assertEquals(_format(actual), snapshot.slice(1, -1));
+    } catch {
+      isEqual = false;
+    }
+    if (!isEqual) console.info("Snapshot updated", testName);
```
bcheidemann, line 893:
```suggestion
```

Do we need this message?

bcheidemann (changes requested):
This looks good! Just a couple comments but nothing major 🙂 Could you also format the file?

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+    const testFile = fromFileUrl(context.origin);
+    const parts = parse(testFile);
+    return `${join(parts.dir, parts.name)}.snap`;
+  }
+  function writeSnapshotFileSync() {
+    const snapshotPath = getSnapshotPath();
+    ensureFileSync(snapshotPath);
+    const buf = ['export const snapshot = {};\n'];
+    for (const [key, value] of Object.entries(updatedSnapshotFile)) {
+      buf.push(`\nsnapshot[\`${key}\`] = \`\n${_format(value)}\n\`;\n`);
```
bcheidemann, line 918: This doesn't need to be done now, but Jest only adds the extra `\n`s when the formatted result contains a new line.

e.g.

```ts
// test file (jest)
expect("single line").toMatchSnapshot();

// snapshot file
exports[...] = `"single line"`;

// test file (jest)
expect("two\nlines").toMatchSnapshot();

// snapshot file
exports[...] = `
"two
lines"
`;
```

It would be nice if we could mimic this behaviour to make migration of tests easier. Again though, this can be a future improvement if you don't want to add it here.

bartlomieju:
@hyp3rflow any chance we could finish this PR for next week's release?

disjukr:
@bartlomieju He is absence now due to military issue (until 4/28 kst). I can fix the code on his branch, but it might be hard to finish for the next week's release.

bcheidemann:
When is next weeks release? I have a 4 day weekend so I can pick this up and should have time to finish it. I would need to create a new PR though as I don't have write access to this branch.

bartlomieju:
> @bartlomieju He is absence now due to military issue (until 4/28 kst). I can fix the code on his branch, but it might be hard to finish for the next week's release.

@disjukr thanks for letting me know.



> @bartlomieju When is next weeks release? I have a 4 day weekend so I can pick this up and should have time to finish it. I would need to create a new PR though as I don't have write access to this branch.

The release is on Wednesday, April 20th. Feel free to open up a new branch. If we miss the date, that's not a big deal though, we'll just land it the following week.

bcheidemann:
> > @bartlomieju He is absence now due to military issue (until 4/28 kst). I can fix the code on his branch, but it might be hard to finish for the next week's release.
> 
> @disjukr thanks for letting me know.
> 
> 
> 
> > @bartlomieju When is next weeks release? I have a 4 day weekend so I can pick this up and should have time to finish it. I would need to create a new PR though as I don't have write access to this branch.
> 
> The release is on Wednesday, April 20th. Feel free to open up a new branch. If we miss the date, that's not a big deal though, we'll just land it the following week.

Thanks, I'll open a new branch tomorrow 👍

testing/asserts.ts
bcheidemann: Removed this for now

testing/asserts.ts
bcheidemann: Done

testing/asserts.ts
bcheidemann: Done

testing/asserts.ts
bcheidemann: Done

testing/asserts.ts
bcheidemann: Done

testing/asserts.ts
```diff
@@ -864,3 +866,58 @@ export function unimplemented(msg?: string): never {
...
+    globalThis.onunload = writeSnapshotFileSync;
+  } else {
+    assertEquals(_format(actual), snapshot.slice(1, -1));
+  }
+  function getName(context: Deno.TestContext): string {
+    if (context.parent) return `${getName(context.parent)} > ${context.name}`;
+    return context.name;
+  }
+  function getCount() {
+    const count = snapshotMap?.[name] ? snapshotMap[name] : 1;
```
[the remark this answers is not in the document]
bcheidemann: Done

testing/asserts.ts
bcheidemann: Done

testing/asserts.ts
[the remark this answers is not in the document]
bcheidemann: It's hard to do this with the way we now read the snapshot file. I think this can probably be implemented separately if it's needed

testing/asserts.ts
bcheidemann: Fixed

testing/asserts.ts
bcheidemann: Done

bcheidemann:
@bartlomieju @kt3k Could you review this and remove the draft status? Thanks 🙂

bcheidemann:
Thanks for giving me write access @hyp3rflow! 🙂 I would also value your review if you are able to

testing/README.md
```diff
@@ -87,6 +88,17 @@ Deno.test("isNotStrictlyEqual", function (): void {
...
 
+Using `assertSnapshot()`:
+
+```ts
+import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+
+Deno.test("isSnapshotMatch", function (test): void {
+  const a = {};
+  assertSnapshot(test, a);
+});
```
bartlomieju, line 99: We could use some more data in this example - please add some fields to the object. Also mention the required flag to update the snapshot

bcheidemann: Added some fields to the example as requested. I also added a new `## Snapshot Testing` section as I felt there was too much that should be documented to put it under the usage example. Hope this is okay 🙂

bcheidemann: Also added some test data to the map in `asserts_test.ts` 👍

Ready for re-review @bartlomieju @kt3k

testing/asserts.ts
```diff
@@ -12,9 +12,12 @@ import {
   stripColor,
   white,
 } from "../fmt/colors.ts";
+import { fromFileUrl, join, parse } from "../path/mod.ts";
+import { ensureFile, ensureFileSync } from "../fs/mod.ts";
```
bartlomieju, line 16: The tests are failing, because this module is marked as "browser compatible" - see comment at the top of the file. Maybe let's move it to `testing/snapshot.ts` for now, since snapshotting requires to write files to disk and can't be done in browsers.

bcheidemann: @bartlomieju this is done now 🙂

bcheidemann:
`ci / test (canary, windows-2019)` is failing because of path issues when importing the snapshot file. I will try changing this to use the `file://` scheme as this _should_ work across all platforms.

bcheidemann:
> `ci / test (canary, windows-2019)` is failing because of path issues when importing the snapshot file. I will try changing this to use the `file://` scheme as this _should_ work across all platforms.

@bartlomieju I have pushed a commit which will hopefully fix the issue 😄

kt3k:
When I create `test.ts` like the below in current directory (at deno_std root):

```ts
import { assertSnapshot } from "./testing/snapshot.ts";
Deno.test("isSnapshotMatch", function (test): void {
  const a = { 
    hello: "world!",
    example: 123,
  };  
  assertSnapshot(test, a); 
});
```

and ran `deno test -A test.ts` then, it created the snapshot file at the path `./file:/Users/kt3k/denoland/deno_std/__snapshots__/test.ts.snap`. I feel this is a bit strange. Is this an intended behavior?

testing/README.md
```diff
@@ -87,6 +88,22 @@ Deno.test("isNotStrictlyEqual", function (): void {
...
 ```
 
+Using `assertSnapshot()`:
+
+For more usage information, see [Snapshot Testing](#snapshot-testing).
+
+```ts
+import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
+
+Deno.test("isSnapshotMatch", function (test): void {
```
kt3k, line 98: nit: I think most docs/tests use `t` for TestContext variable. Let's make it consistent

```suggestion
Deno.test("isSnapshotMatch", function (t): void {
```

testing/README.md
```diff
@@ -154,6 +171,76 @@ Deno.test("fails", async function () {
...
+
+The `assertSnapshot` function will create a snapshot of a value and compare it
+to a reference snapshot, which is stored alongside the test file in the
+`__snapshots__` directory.
+
+```ts
+// example_test.ts
+import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
+
+Deno.test("isSnapshotMatch", function (test): void {
```
kt3k, line 186:
```suggestion
Deno.test("isSnapshotMatch", function (t): void {
```

testing/snapshot.ts
```diff
@@ -0,0 +1,205 @@
...
+  context.updatedSnapshot.forEach((value, key) => {
+    let formattedValue = escapeStringForJs(value);
+    formattedValue = formattedValue.includes("\n")
+      ? `\n${formattedValue}\n`
+      : formattedValue;
+    const formattedKey = escapeStringForJs(key);
+    buf.push(`\nsnapshot[\`${formattedKey}\`] = \`${formattedValue}\`;`);
+  });
+  ensureFileSync(context.snapshotPath as string);
+  Deno.writeTextFileSync(context.snapshotPath as string, buf.join("\n"));
```
kt3k, line 45: Because `snapshotPath` is now file url, I think we need to convert it back to file path

kt3k:
Updating behavior is not covered by the unit tests. Can we somehow test it? (For example by creating tmp directory and run `deno test` command with `-u` option with `Deno.run()` API.)

kt3k:
Other part looks good to me. Snapshot testing itself looks nicely implemented 👍

bcheidemann:
> Updating behavior is not covered by the unit tests. Can we somehow test it? (For example by creating tmp directory and run `deno test` command with `-u` option with `Deno.run()` API.)

Yes I agree it seems sensible to test this. I'll see what I can come up with after work 🙂

testing/README.md
```diff
@@ -154,6 +171,76 @@ Deno.test("fails", async function () {
...
+can be limited to only include existing snapshot files, if so desired.
+
+### Version Control:
+
+Snapshot testing works best when changes to snapshot files are comitted
+alongside other code changes. This allows for changes to reference snapshots to
+be reviewed along side the code changes that caused them, and ensures that when
+others pull your changes, their tests will pass without needing to update
+snapshots locally.
+
```
bartlomieju, line 243: Great docs, could you also open a PR to https://github.com/denoland/manual that adds "Snapshot testing" page to "Testing" chapter?

testing/__snapshots__/snapshot_test.ts.snap
```diff
@@ -0,0 +1,141 @@
...
+  "",
+  "    [Diff] Actual / Expected",
+  "",
+  "",
+  '-   "Hello!"',
+  '+   "Hello World!"',
+  "",
+  "",
+]
+`;
```
bartlomieju, line 141: Nitpick, could generated snapshot file end with a newline so we don't get these markers on GitHub?

testing/snapshot.ts
bcheidemann: Done 🙂

testing/snapshot.ts
```diff
@@ -0,0 +1,205 @@
...
+ * of snapshots updated after all tests have run.
+ *
+ * This function can safely be called more than once and will only register the teardown
+ * function once.
+ *
+ * @param context Assert snapshot context
+ */
+function registerSnapshotTeardown(context: AssertSnapshotContext) {
+  if (context.teardownRegistered) return;
+  globalThis.onunload = () => {
```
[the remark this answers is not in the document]
bcheidemann: Good call! Done 🙂

testing/__snapshots__/snapshot_test.ts.snap
bcheidemann: Done 🙂

bartlomieju:
The examples from `testing/README.md` don't work unfortunately:
```ts
// snap_test.ts
import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";

Deno.test("isSnapshotMatch", function (t): void {
  const a = {
    hello: "world!",
    example: 123,
  };
  assertSnapshot(test, a);
});
```

```
../deno/target/debug/deno test ./snap_test.ts --allow-read -- --update
running 1 test from ./snap_test.ts
isSnapshotMatch ... FAILED (7ms)

failures:

./snap_test.ts > isSnapshotMatch
Test case is leaking async ops.

-

failures:

	./snap_test.ts
	isSnapshotMatch

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (24ms)

error: Test failed
```

```
../deno/target/debug/deno test ./snap_test.ts --allow-read -- --update
running 1 test from ./snap_test.ts
isSnapshotMatch ...
test result: FAILED. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (26ms)

error: Uncaught (in promise) PermissionDenied: Requires write access to "file:///Users/ib/dev/deno_std/__snapshots__", run again with the --allow-write flag
      await Deno.mkdir(dir, { recursive: true });
      ^
    at async Object.mkdir (deno:runtime/js/30_fs.js:109:5)
    at async ensureDir (file:///Users/ib/dev/deno_std/fs/ensure_dir.ts:22:7)
    at async ensureFile (file:///Users/ib/dev/deno_std/fs/ensure_file.ts:27:7)
    at async readSnapshotFile (file:///Users/ib/dev/deno_std/testing/snapshot.ts:187:7)
    at async assertSnapshot (file:///Users/ib/dev/deno_std/testing/snapshot.ts:105:27)

```

I think this should be caught be some integration tests. @[bcheidemann](https://github.com/bcheidemann) let me know if you need any help with that.

This is very close to landing 🚀

bcheidemann:
> Updating behavior is not covered by the unit tests. Can we somehow test it? (For example by creating tmp directory and run `deno test` command with `-u` option with `Deno.run()` API.)

@kt3k I have added a test which does exactly this

bcheidemann:
> The examples from `testing/README.md` don't work unfortunately:
> 
> ```ts
> // snap_test.ts
> import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
> 
> Deno.test("isSnapshotMatch", function (t): void {
>   const a = {
>     hello: "world!",
>     example: 123,
>   };
>   assertSnapshot(test, a);
> });
> ```
> 
> ```
> ../deno/target/debug/deno test ./snap_test.ts --allow-read -- --update
> running 1 test from ./snap_test.ts
> isSnapshotMatch ... FAILED (7ms)
> 
> failures:
> 
> ./snap_test.ts > isSnapshotMatch
> Test case is leaking async ops.
> 
> -
> 
> To get more details where ops were leaked, run again with --trace-ops flag.
> 
> failures:
> 
> 	./snap_test.ts
> 	isSnapshotMatch
> 
> test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out (24ms)
> 
> error: Test failed
> ```
> 
> ```
> ../deno/target/debug/deno test ./snap_test.ts --allow-read -- --update
> running 1 test from ./snap_test.ts
> isSnapshotMatch ...
> test result: FAILED. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (26ms)
> 
> error: Uncaught (in promise) PermissionDenied: Requires write access to "file:///Users/ib/dev/deno_std/__snapshots__", run again with the --allow-write flag
>       await Deno.mkdir(dir, { recursive: true });
>       ^
>     at async Object.mkdir (deno:runtime/js/30_fs.js:109:5)
>     at async ensureDir (file:///Users/ib/dev/deno_std/fs/ensure_dir.ts:22:7)
>     at async ensureFile (file:///Users/ib/dev/deno_std/fs/ensure_file.ts:27:7)
>     at async readSnapshotFile (file:///Users/ib/dev/deno_std/testing/snapshot.ts:187:7)
>     at async assertSnapshot (file:///Users/ib/dev/deno_std/testing/snapshot.ts:105:27)
> ```
> 
> I think this should be caught be some integration tests. @[bcheidemann](https://github.com/bcheidemann) let me know if you need any help with that.
> 
> This is very close to landing rocket

@bartlomieju I have updated the examples in `testing/README.md` and these should now be correct 🙂

What do you suggest for integration tests?

testing/README.md
bcheidemann: Done 🙂 https://github.com/denoland/manual/pull/294

kt3k:
> I wasn't able to reproduce the second error case (Test case is leaking async ops.)

This was probably caused by the absense of `await` before `assertSnapshot` in the previous version of readme example (I experienced a similar error).

Now I was able to run the README example on my end (with and without `-u` option) and snapshot files were created in documented places. Nice work 👍

testing/snapshot_test.ts
```diff
@@ -0,0 +1,175 @@
...
+      await assertSnapshot(t, [
+        1,
+        2,
+      ]);
+    });
+    `,
+    false,
+  );
+
+  await assertSnapshot(t, formatOutput(result1.output).split("\n"));
```
kt3k, line 134: haha, nice usage of `assertSnapshot` 👍

kt3k (approved):
LGTM. Thank you for your contributions! @bcheidemann @hyp3rflow

testing/snapshot_test.ts
```diff
@@ -0,0 +1,175 @@
...
+});
+
+Deno.test("Snapshot Test - Update", async (t) => {
+  const TEMP_DIR = ".tmp";
+
+  async function runTestWithUpdateFlag(test: string, deleteTempDir = true) {
+    const testDir = dirname(fromFileUrl(import.meta.url));
+    const tempDir = join(testDir, TEMP_DIR);
+    const tempTestFileName = "test.ts";
+    const tempTestFilePath = join(tempDir, tempTestFileName);
```
kt3k, line 83: I think it's fine for first pass, but ideally we like to use `Deno.makeTempDir` for making temp directory. Filed an issue for this https://github.com/denoland/deno_std/issues/2129

bcheidemann: My initial implementation did use this but I refactored as it resulted in a random output path which I would have had to obfuscate in the snapshot.

What's the advantage of using this by the way?

bartlomieju (approved):
LGTM! Thank you so much to @hyp3rflow and @bcheidemann, this is very useful feature and hopefully many users will find it helpful.

bcheidemann:
> LGTM! Thank you so much to @hyp3rflow and @bcheidemann, this is very useful feature and hopefully many users will find it helpful.

Thanks @bartlomieju @kt3k for all your support with this feature!

testing/snapshot_test.ts
kt3k: `Deno.makeTempDir` creates a directory under the system's default temp directory and they are automatically removed by the OS some time later. So we don't need care much about those files even if the test code failed to remove it. Currently we need to remove `testing/.tmp` manually when the test case failed to remove it for some reason, and that's a little inconvenient

bcheidemann: Thanks for the explanation! 🙂

jaydenseric:
Just noticing this now; sad to see opinionated magic around where snapshots are saved (a `__snapshots__` dir) instead of the `assertSnapshot` function just accepting 2 arguments; the actual value being asserted, and a file URL / string path to the snapshot file. That way just by reading argument intellisense it's predictable (and customisable) where the snapshots end up, and the snapshot can be held in the correct type of file for the data (`.json`, `.html`, `.css`, etc.) for syntax highlighting and such when directly viewing the snapshot file in an editor.

The simpler design also means a simpler implementation to write and maintain.

Here is an example of the (IMO) ideal `assertSnapshot` API in action on npm:

- https://www.npmjs.com/package/snapshot-assertion
- https://github.com/jaydenseric/snapshot-assertion
- https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L40-L77

Note the optional third argument `assertion` that accepts an assertion function for comparing the actual text with the snapshot file text, in case you want to use your own custom one for a special kind of diff or something:

https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L16-L19

kt3k:
Explicit snapshot path looks nice to me. But implicit snapshot paths are also convenient for the users who don't care about snapshot paths / snapshot file syntax. Can we somehow support both patterns by overloading `assertSnapshot`?

bcheidemann:
> Just noticing this now; sad to see opinionated magic around where snapshots are saved (a `__snapshots__` dir) instead of the `assertSnapshot` function just accepting 2 arguments; the actual value being asserted, and a file URL / string path to the snapshot file. That way just by reading argument intellisense it's predictable (and customisable) where the snapshots end up, and the snapshot can be held in the correct type of file for the data (`.json`, `.html`, `.css`, etc.) for syntax highlighting and such when directly viewing the snapshot file in an editor.
> 
> The simpler design also means a simpler implementation to write and maintain.
> 
> Here is an example of the (IMO) ideal `assertSnapshot` API in action on npm:
> 
> - https://www.npmjs.com/package/snapshot-assertion
> - https://github.com/jaydenseric/snapshot-assertion
> - https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L40-L77
> 
> Note the optional third argument `assertion` that accepts an assertion function for comparing the actual text with the snapshot file text, in case you want to use your own custom one for a special kind of diff or something:
> 
> https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L16-L19

I agree with a lot of what you're proposing but not necessarily the API. I also don't think it's too late to have a discussion around this - it's only the first pass of this feature. Why don't you raise an issue and we can continue the conversation there? 🤔

bcheidemann:
> Explicit snapshot path looks nice to me. But implicit snapshot paths are also convenient for the users who don't care about snapshot paths / snapshot file syntax. Can we somehow support both patterns by overloading `assertSnapshot`?

I don't see why not, and I have a couple of ideas how it could be done without breaking changes to the API. An issue seems like the more appropriate place to discuss this though.

hyp3rflow:
I just came back earlier due to the pandemic. Thanks, @bcheidemann for your awesome works 😊
I'm also excited that I can use this feature in release!

PR #2044, opened 2022-03-20
feat(http): Add add http methods enum

antspk:
This is an old issue that first was discussed in [5790](https://github.com/denoland/deno/issues/5790) and even had a PR opened (5854)(https://github.com/denoland/deno/pull/5854) however the work on this issue was stopped, I think this is still relevant and it would be convenient to have this in std library.

### DIFF
--- a/http/http_method.ts
+++ b/http/http_method.ts
@@ -1,0 +1,34 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+/**
+ * Enum of HTTP RFC 7231 & RFC 5789 methods.
+ *
+ * ```ts
+ * import {
+ *   Method
+ * } from "https://deno.land/std@$STD_VERSION/http/http_method.ts";
+ *
+ * console.log(Method.GET); //=> "GET"
+ * ```
+ */
+export enum Method {
+  /* RFC 7231, 4.3.1 */
+  GET = "GET",
+  /* RFC 7231, 4.3.2 */
+  HEAD = "HEAD",
+  /* RFC 7231, 4.3.3  */
+  POST = "POST",
+  /* RFC 7231, 4.3.4 */
+  PUT = "PUT",
+  /* RFC 7231, 4.3.5 */
+  DELETE = "DELETE",
+  /* RFC 7231, 4.3.6 */
+  CONNECT = "CONNECT",
+  /* RFC 7231, 4.3.7 */
+  OPTIONS = "OPTIONS",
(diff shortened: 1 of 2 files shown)


kt3k (commented):
I'm in favor of this PR if there's no strong objection from community or core members

kitsonk:
Generally, enums are a problem, as they require runtime emit.  Usually string literal unions are better, like `type HTTPMethods = "GET" | "DELETE";`

I wonder specifically what problem this solves or improves.

kt3k:
> Generally, enums are a problem, as they require runtime emit. Usually string literal unions are better, like type HTTPMethods = "GET" | "DELETE";

I agree that union types are better in general, but in the case of http methods, it seems that any string is allowed as http method (for example, fetch spec says it. ref: https://fetch.spec.whatwg.org/#methods). So having `type HTTPMethods = "GET" | "HEAD" | ... | "PATCH"` doesn't make much sense in my view.

> I wonder specifically what problem this solves or improves.

Not a big improvement, but the users can make sure they're not making any typo by using enum, and also can get better dev experience by the completion of method names (such completions are not useful for advanced users, but maybe useful for people who are new to http protocol)

kitsonk:
> So having `type HTTPMethods = "GET" | "HEAD" | ... | "PATCH"` doesn't make much sense in my view.

So why would an enum restriction, with it required emitted code, be suitable if any value is allowed?

You can do something like:

```ts
function isGET(method: string): method is "GET" {
 return method === "GET";
}
```

If you want some type narrowing. Lots of other tricks too.

antspk:
I understand that there are drawbacks, but there are benefits too, for example having method union type you couldn't use it as a variable:

```ts
fetch('http://domain.tld', { method: Method.POST });
```
it's not like it's a problem writing it like this:
```ts
fetch('http://domain.tld', { method: "POST" });
```

Another thing is that having `Status` feels like `Method` enum should be an obvious thing, and it's missing right now.

So I think that the benefits are only from the developer experience side:
1. It's consistent with the `Status` enum.
2. It's usable as a variable in cases e.g. `fetch`.
3. It's intuitive.

The and the drawbacks drawbacks:
1. It requires runtime emit.

So would having additional `Method` enum alongside the already existing `Status` enum would be a problem, considering its drawbacks?

bartlomieju:
@antspk thank you for the PR, but I'm going to close it without a merge. This PR sparked conversation among the core team to include this helper and we've decided that we're gonna upstream `oak_commons` module with this functionality. Thanks again.

PR #2050, opened 2022-03-21
chore: warn on non supported event, but still register listener

kt3k:
This PR changes the behavior of binding to 'unimplemented' events on `process` object.

With this change, the events are still not handled, but the listeners are actually registered on `process` object and it increased/decreased the `listenerCount` of the each event.

This makes mocha's unit testing happy, and now we can pass `npx nps test.node.jsapi` test category in compat mode with this change.

part of #12886


bartlomieju (approved):
LGTM

PR #2051, opened 2022-03-21
refactor(node): clean up async_hooks.ts

kt3k:
Currently we have two duplicated implementation of `async_hooks` utility in `node/_async_hooks.ts` and `node/internal/async_hooks.ts` (both based on the same node.js utility)

This PR merges them into `node/internal/async_hooks.ts`.

part of https://github.com/denoland/deno_std/issues/1707


bartlomieju (approved):
LGTM

PR #2063, opened 2022-03-24
chore: try to disable language stats for node/ directory

bartlomieju:


.gitattributes
```diff
@@ -3,4 +3,4 @@
 
 # Tell github these are vendored files.
 # Doesn't include them in the language statistics.
-/node/_tools/suites/ linguist-vendored
+node/_tools/suites/ linguist-vendored
```
kt3k, line 6: ah, `suites` are now renamed to `test` (because some test cases uses the directory name in the asssertions)

kt3k (approved):

PR #2069, opened 2022-03-28
feat(testing): Add utility for faking time

KyleJune:
Depends on: https://github.com/denoland/deno_std/pull/2048

My previous PR that this one depends on adds `mock.ts`. This PR adds a utility for faking time. It mocks time by replacing the Date, setTimeout, clearTimeout, setInterval, clearInterval globals with fake versions that are controlled by the user via the `FakeTime` instance.

The `README.md` section for faking time can be found [here](https://github.com/KyleJune/deno_std/blob/kj-add-fake-time/testing/README.md#faking-time).

[Here](https://doc.deno.land/https://raw.githubusercontent.com/KyleJune/deno_std/26844186a8bd9053a1f340aa1f5e7cdc9bc5ac5e/testing/time.ts) are the deno docs for `time.ts`.


bartlomieju:
@KyleJune is this ready for review?

KyleJune:
> @KyleJune is this ready for review?

Yes, all it needs is main to be merged back into it and Deno header to be added. I could do that tonight. If there are no other issues and you decide you want to get it merged, feel free to make those changes yourself.

bartlomieju:
> > @KyleJune is this ready for review?
> 
> Yes, all it needs is main to be merged back into it and Deno copyright header to be added. I could do that tonight. If there are no other issues and you decide you want to get it merged, feel free to make those changes yourself.

Thanks! I'll give it as review. @kt3k please take a look as well.

testing/README.md
```diff
@@ -793,3 +793,56 @@ Deno.test("randomMultiple uses randomInt to generate random multiples between -1
   assertSpyCalls(randomIntStub, 2);
 });
 ```
+
+### Faking time
+
+Say we have a function that has time based behavior that we would like to test.
+With real time, that could cause test's to take much longer than they should. If
```
bartlomieju, line 800:
```suggestion
With real time, that could cause tests to take much longer than they should. If
```

testing/README.md
```diff
@@ -793,3 +793,56 @@ Deno.test("randomMultiple uses randomInt to generate random multiples between -1
...
+export function secondInterval(cb: () => void): number {
+  return setInterval(cb, 1000);
+}
+```
+
+With `FakeTime` we can do that. When the `FakeTime` instance is created, it
+splits from real time. The Date, setTimeout, clearTimeout, setInterval, and
+clearInterval globals are replaced with versions that use the fake time until
+real time is restored. You can control how time ticks forward with the `tick`
+method on the `FakeTime` instance.
```
bartlomieju, line 816:
```suggestion
With `FakeTime` we can do that. When the `FakeTime` instance is created, it
splits from real time. The `Date`, `setTimeout`, `clearTimeout`, `setInterval` and
`clearInterval` globals are replaced with versions that use the fake time until
real time is restored. You can control how time ticks forward with the `tick`
method on the `FakeTime` instance.
```

I don't understand what "splits" means in this sentence

testing/time.ts
```diff
@@ -0,0 +1,436 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+/** This module is browser compatible. */
+
+import { ascend, RBTree } from "../collections/rb_tree.ts";
+import { DelayOptions } from "../async/delay.ts";
+
+export const _internals = {
```
bartlomieju, line 8: Why do these need to be exported?

testing/time.ts
```diff
@@ -0,0 +1,436 @@
...
+}
+
+function fakeClearTimeout(id?: number): void {
+  if (!time) throw new TimeError("no fake time");
+  if (typeof id === "number" && dueNodes.has(id)) {
+    dueNodes.delete(id);
+  }
+}
+
+/** The fake time replacement for se */
```
bartlomieju, line 144: Seems like this comment was cut-off by mistake

testing/time.ts
```diff
@@ -0,0 +1,436 @@
...
+let advanceRate: number;
+let advanceFrequency: number;
+let advanceIntervalId: number | undefined;
+let timerId: Generator<number>;
+let dueNodes: Map<number, DueNode>;
+let dueTree: RBTree<DueNode>;
+
+/**
+ * Overrides the real Date object and timer functions with fake ones that can be
+ * controlled through the fake time instance.
```
bartlomieju, line 223: Could you add an example in a codeblock?

bartlomieju (commented):
Mostly looks good to me, thank you @KyleJune, this is very useful!

Could I ask you to open a PR to https://github.com/denoland/manual that updates testing chapters with this functionality?

testing/time.ts
KyleJune: I'll look into what it was supposed to say tonight. My bad.

testing/time.ts
KyleJune: Sure

testing/time.ts
KyleJune: They are used by the test coverage to verify the real methods are called or not called when they should be. I believe the tests might replace them with stubs so that the real timer functions are never called. I could move them to a separate file with _ prefix to the name so that this constant is never exposed to end users.

testing/time.ts
KyleJune: This comment can just be deleted. The function is never exported and the fakeSetTimeout function doesn't have a comment above it either.

testing/time.ts
bartlomieju: I see, that makes sense, though I can see this not working properly if user's code replaced them before this file is loaded. I think this is fine as is, can you add a comment describing the purpose?

KyleJune: Sure I can do that tonight and I think I'll just move them to _test_utils.ts. alternatively I could create a new file called _time.ts that could export just it instead.

bartlomieju: `_test_utils.ts` seems fine 👍

testing/time.ts
KyleJune: Would the example from the README.md file be fine?

bartlomieju: Yeah, definitely!

testing/README.md
KyleJune: The suggestion caused lint to fail. I'll run deno fmt tonight when addressing the other feedback

KyleJune:
> Could I ask you to open a PR to https://github.com/denoland/manual that updates testing chapters with this functionality?

Looks like it doesn't have info regarding bdd.ts and mock.ts yet either. I could try updating the manual for those and time.ts this weekend.

bartlomieju:
> > Could I ask you to open a PR to https://github.com/denoland/manual that updates testing chapters with this functionality?
> 
> Looks like it doesn't have info regarding bdd.ts and mock.ts yet either. I could try updating the manual for those and time.ts this weekend.

@KyleJune that would be awesome, much obliged 🙏

kt3k (approved):
LGTM. Awesome work! Thanks for your contribution!

kt3k:
We have a few test cases which depend on some specific date time. We currently use `faketime` command for setting it up in CI, but I hope we can replace those set ups with this utility 

https://github.com/denoland/deno_std/blob/37b0b02f9114482ae7cd9bc7ea239785f3e31b02/datetime/test_2021_12_31.ts

testing/time.ts
KyleJune: I thought about this and think _time.ts would be better so that when people import time.ts it won't have to load _test_utils.ts too which contains irrelevant utilities and could potentially grow in the future.

KyleJune: I moved the _internals const to _time.ts and added a comment describing its purpose.

testing/time.ts
KyleJune: I deleted this comment.

testing/time.ts
KyleJune: I've added the example from the README.md file.

KyleJune:
> We have a few test cases which depend on some specific date time. We currently use `faketime` command for setting it up in CI, but I hope we can replace those set ups with this utility
> 
> https://github.com/denoland/deno_std/blob/37b0b02f9114482ae7cd9bc7ea239785f3e31b02/datetime/test_2021_12_31.ts

I went ahead and did that. The following PR branches off of this one.

https://github.com/denoland/deno_std/pull/2109

KyleJune:
I've also made the remaining requested changes. I believe this is ready to go now.

kt3k:
@KyleJune Thanks for updating!

PR #2073, opened 2022-03-30
fix(node/module): catch errors from core.evalContext

milahu:
quick stab at https://github.com/denoland/deno/issues/12939

not tested

expected result: better error reporting on SyntaxError (parse error)

`return [null, { thrown }]` should imitate the API of [Deno.core.evalContext](https://github.com/denoland/deno/blob/061090de7e95e8e7a97f3277bd1a72899ebd1570/core/bindings.rs#L686)

### DIFF
--- a/node/module.ts
+++ b/node/module.ts
@@ -1312,7 +1312,14 @@ function wrapSafe(
 ): RequireWrapper {
   // TODO(bartlomieju): fix this
   const wrapper = Module.wrap(content);
-  const [f, err] = core.evalContext(wrapper, filename);
+  const [f, err] = (() => {
+    try {
+      return core.evalContext(wrapper, filename);
+    }
+    catch (thrown: any) {
+      return [null, { thrown }];
+    }
+  })();
   if (err) {
     if (process.mainModule === cjsModuleInstance) {
       enrichCJSError(err.thrown);
--- a/node/module.ts
+++ b/node/module.ts
@@ -1315,8 +1315,7 @@ function wrapSafe(
   const [f, err] = (() => {
     try {
       return core.evalContext(wrapper, filename);
-    }
-    catch (thrown: any) {
+    } catch (thrown: any) {
       return [null, { thrown }];
     }
   })();


bartlomieju (approved):
Thanks, but we need a test for this change

bartlomieju (changes requested):
Ooops, approved by mistake

kt3k:
This swallows any error from commonjs module evaluation, and returns `null`ed module. I think this doesn't solve any problem (I also commented the original issue about how to execute pnpm correctly in compat mode

milahu:
> This swallows any error from commonjs module evaluation

not really, every error goes through catch &rarr; enrichCJSError &rarr; throw

or should we catch only some errors? for example SyntaxError

> I think this doesn't solve any problem

the goal is to reach this branch on SyntaxError

```ts

  if (err) {
    if (process.mainModule === cjsModuleInstance) {
      enrichCJSError(err.thrown);
    }
    throw err.thrown;
  }
```

kt3k:
@milahu Ah, ok. You're right. This doesn't swallow error, but changes the way how SyntaxError handled. Sorry for the misunderstanding. I'll try this branch later

kt3k:
I confirmed this doesn't break existing behavior, but also I couldn't observe any improvement in the error output with this change

test.ts
```ts
import { createRequire } from "./node/module.ts";
const require = createRequire(import.meta.url);
require("./a.js");
```

a.js
```js
const a: number = 1; // <= this is a syntax error
```

BEFORE this PR

```
$ deno run -A --unstable test.ts
Check file:///Users/kt3k/denoland/deno_std/test.ts
error: Uncaught SyntaxError: Missing initializer in const declaration
Warning Couldn't format source line: Column 184 is out of bounds (source may have changed at runtime)
    at wrapSafe (file:///Users/kt3k/denoland/deno_std/node/module.ts:1315:25)
    at Module._compile (file:///Users/kt3k/denoland/deno_std/node/module.ts:251:29)
    at Object.Module._extensions..js (file:///Users/kt3k/denoland/deno_std/node/module.ts:1334:10)
```

AFTER this PR
```
$ deno run -A --unstable test.ts
Check file:///Users/kt3k/denoland/deno_std/test.ts
error: Uncaught SyntaxError: Missing initializer in const declaration
Warning Couldn't format source line: Column 184 is out of bounds (source may have changed at runtime)
    at file:///Users/kt3k/denoland/deno_std/node/module.ts:1317:19
    at wrapSafe (file:///Users/kt3k/denoland/deno_std/node/module.ts:1321:5)
    at Module._compile (file:///Users/kt3k/denoland/deno_std/node/module.ts:251:29)
```

stack trace changes a little bit, but the syntax errors are shown in a similar way

milahu:
> I couldn't observe any improvement in the error output

hmm ...

expected result:

```
$ cd /tmp
$ echo "const a: number = 1;" >a.js
$ node a.js 
/tmp/a.js:1
const a: number = 1;
      ^

SyntaxError: Missing initializer in const declaration
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1032:15)
    at ...
```

the 4 lines before SyntaxError are generated by
[node/lib/internal/source_map/prepare_stack_trace.js](https://github.com/nodejs/node/blob/92567283f4027940d4a4df0e7fb7444b67842ed5/lib/internal/source_map/prepare_stack_trace.js#L141) or
[node/src/node_errors.cc](https://github.com/nodejs/node/blob/92567283f4027940d4a4df0e7fb7444b67842ed5/src/node_errors.cc#L52)
using location from stack and code from sourcemap

problem: SyntaxError has no location (also not in error.stack)
workaround: https://github.com/browserify/syntax-error

```js
var check = require('syntax-error')

const src = "const a: number = 1;"
const file = "/tmp/a.js";

try {
  eval(src) // /tmp/eval.js:7
}
catch (e) {
  console.log(check(src, file).annotated.split("\n").slice(1, -1).join("\n") + "\n");
  // slice -1: remove ParseError from acorn JS parser
  console.log(e);
}
```

result:

```
/tmp/a.js:1
const a: number = 1;
       ^

SyntaxError: Missing initializer in const declaration
    at Object.<anonymous> (/tmp/eval.js:7:6)
    at Module._compile (node:internal/modules/cjs/loader:1103:14)
    at ...
```

bartlomieju:
@kt3k this message: `Warning Couldn't format source line: Column 184 is out of bounds (source may have changed at runtime)` is printed by Deno, because Deno is not aware that we're wrapping the source code in additional statements, when it applies source maps. We can fix this once Deno gets "native" support for `require` instead of using userland polyfill from `std/node`

bartlomieju:
@kt3k do you still want to pursue landing this PR?

kt3k:
@bartlomieju Let's close for now as this one is not active

@milahu Please open a new PR if you find a solution that doesn't depend on an external package

PR #2076, opened 2022-03-31
fix(node/fs): Fix writing redundant data

YXL76:
If `buffer.length` is larger than `length`, more date will be writed into file.


kt3k:
Good catch! Thanks! Can you add some simple test cases where `buffer.length > length`

YXL76:
test has been added

kt3k:
> test has been added

Very nice! Thanks!

kt3k (approved):
LGTM

PR #2084, opened 2022-04-06
refactor(node/tls): use in-memory options for cert files

kt3k:
We currently creates temporary cert files in `TlsServer.listen` to pass them to `Deno.listenTls` because `Deno.listenTls` didn't support string args for cert files at tha time. We added `cert` and `key` option in https://github.com/denoland/deno/pull/13740, and now we can simply use those options to specify certs with strings.


bartlomieju (approved):
Nice cleanup!

PR #2085, opened 2022-04-06
chore(node): make the message for notImplemented required

kt3k:
This PR makes the 1st arg (message) for `notImplemented` utility required.

If the message is omitted, the program only shows `notImplemented` and it doesn't give any information, and therefore that's inconvenient. This PR make the message required and improve the situation.


bartlomieju (approved):
LGTM

Issue #2061, opened 2022-03-24
node: implement 'upgrade' event in http.Server


kt3k:
part of #2055

Necessary for supporting websocket implementation of hmr server of vite https://github.com/vitejs/vite/blob/c7fc7c34029f109bb716c6a2888022b2f4c40e3d/packages/vite/src/node/server/ws.ts#L39

kt3k:
resolved. ref: https://github.com/denoland/deno/issues/15398#issuecomment-1236406923

PR #2086, opened 2022-04-06
feat(node): `Duplex.fromWeb`

crowlKats:
Needed for #2061
not sure where to put `isReadableStream` & `isWritableStream`. 
Also it seems node has no test for this


node/internal/streams/duplex.mjs
```diff
@@ -2,21 +2,27 @@
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 // deno-lint-ignore-file
 
+import { validateObject, validateBoolean } from "../validators.mjs";
 import { createDeferredPromise } from "../util.mjs";
 import { destroyer } from "./destroy.mjs";
 import { isBlob } from "../blob.mjs";
+import { Buffer } from "../../buffer.mjs";
```
kt3k, line 9: typo?
```suggestion
import { Buffer } from "../../buffer.ts";
```

kt3k:
It looks like they are testing the internal utils (such as `newStreamDuplexFromReadableWritablePair`) instead of testing public interfaces directly https://github.com/nodejs/node/pull/39134/files#diff-d8ce88f0cd20640d6cdffb550abcd3a8fb9b4e543cc04bc87c2d827802e7dc60

Is it possible to somehow borrow these test cases to fix into our situation?

crowlKats:
@kt3k no idea. looking through those tests, they use an internal symbol for web streams called `kState`. we dont have that, so dont think we can even run those tests...

kt3k:
Then maybe let's invent something very simple for now. It also will be tested in the implementation of #2061

bartlomieju (approved):
I agree, let's land it now.

PR #2087, opened 2022-04-06
feat(node/net): add server.ref/unref methods

kt3k:
This PR adds `net.createServer().ref/unref` methods in `std/node`. This makes [get-port](https://www.npmjs.com/package/get-port) module working with compat mode, which is depended by remix.


bartlomieju (approved):
LGTM

PR #2092, opened 2022-04-07
chore(ci): fix release script to have git credentials

dsherret:
For next time.


bartlomieju (approved):

.github/workflows/ci.yml
```diff
@@ -21,7 +21,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           submodules: true
-          persist-credentials: false
```
dsherret, line 24: It seems this was setup when we initially created the CI. I don't think we need this as we don't have it on the CLI repo?

_tools/release/03_release.ts
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env -S deno run --allow-read --allow-write --allow-env --allow-net --allow-run=git --no-check
+#!/usr/bin/env -S deno run --allow-read --allow-write --allow-env --allow-net --allow-run=git,deno --no-check
```
kt3k, line 1: I'm curious where `deno` is used. (`--allow-run=deno` is almost close to `--allow-all`, so ideally we want to avoid this..

kt3k (approved):
LGTM

_tools/release/03_release.ts
dsherret: It's actually not used, but I just added it to match the other script. I removed it now.

I think we should essentially treat this code as `--allow-all` anyway though. Right now we're only importing from our own repos and transitively octokit on skypack. I should vendor that actually. I will open a PR in deno_automation.

Issue #2045, opened 2022-03-20
testing/asserts : `assertFalse`


UltiRequiem:
**Is your feature request related to a problem? Please describe.**

Currently, the testing module lacks a method to check if a value is False.

**Describe the solution you'd like**

```ts
export function assertFalse(expr: unknown, msg = ""): asserts expr is false {
  if (expr) {
    throw new AssertionError(msg);
  }
}
```
**Describe alternatives you've considered**

```ts
Deno.test("Text with letter is not numeric.", () => {
  assertEquals(isNumber("hello"), false);
  assert(isNumber("hello") === false);
  assert(!isNumber("hello"));
});
```

With the solution shown before we could 

```ts
Deno.test("Text with letter is not numeric.", () => {
  assertFalse(isNumber("hello"));
});
```

Which I think is a lot more readable. Similar to 👇🏽 
https://jestjs.io/docs/expect#tobefalsy
https://github.com/avajs/ava/blob/main/docs/03-assertions.md#falseactual-message

getspooky:
@UltiRequiem you can use `assert` or `assertEquals` so no need to add `assertFalse`

UltiRequiem:
> @UltiRequiem you can use `assert` or `assertEquals` so no need to add `assertFalse`

I already listed both options on `Describe alternatives you've considered`.

Despite that, an `assertFalse` function would upgrade readability and other libraries include it.

UltiRequiem:
Another library that has something similar is Ava.

https://github.com/avajs/ava/blob/main/docs/03-assertions.md#falseactual-message

kt3k:
This looks too trivial to include in std to me.

Is there any supportive opinions to this other than the above?

UltiRequiem:
> This looks too trivial to include in std to me.
> 
> Is there any supportive opinions to this other than the above?

I think the two strongest points are readability, and that other frameworks have it.

But I understand if it is not wanted here 😆

bcheidemann:
I don't have a strong opinion if this should be implemented or not.

However, it should be `assertNot` in my opinion to match the existing convention set by other assertions e.g. `assertEquals()` vs `assertNotEquals()`.

bcheidemann:
@bartlomieju @kt3k can we make a call yet on if we want to include this or not? The reasons for including it seem valid to me and I know I (and many other developers) would use it even though it is trivial 😉

bartlomieju:
[...]

PR #2093, opened 2022-04-07
feat(testing): assertFalse

UltiRequiem:
Implements #2045

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -223,6 +223,13 @@ export function assert(expr: unknown, msg = ""): asserts expr {
   }
 }
 
+/** Make an assertion, error will be thrown if `expr` have truthy value. */
+export function assertFalse(expr: unknown, msg = ""): asserts expr is false {
+  if (expr) {
+    throw new AssertionError(msg);
+  }
+}
+
 /**
  * Make an assertion that `actual` and `expected` are equal, deeply. If not
  * deeply equal, then throw.
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -6,6 +6,7 @@ import {
   assertArrayIncludes,
   assertEquals,
   assertExists,
+  assertFalse,
   assertInstanceOf,
   AssertionError,
   assertIsError,
@@ -40,11 +41,14 @@ Deno.test("testingEqualDifferentZero", () => {
 Deno.test("testingEqual", function (): void {
   assert(equal("world", "world"));
   assert(!equal("hello", "world"));
+  assertFalse(equal("hello", "world"));
   assert(equal(5, 5));
   assert(!equal(5, 6));
+  assertFalse(equal(5, 6));
   assert(equal(NaN, NaN));
   assert(equal({ hello: "world" }, { hello: "world" }));
   assert(!equal({ world: "hello" }, { hello: "world" }));
+  assertFalse(equal({ world: "hello" }, { hello: "world" }));
   assert(
     equal(
       { hello: "world", hi: { there: "everyone" } },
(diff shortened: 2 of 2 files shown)


testing/asserts.ts
```diff
@@ -223,6 +223,13 @@ export function assert(expr: unknown, msg = ""): asserts expr {
   }
 }
 
+/** Make an assertion, error will be thrown if `expr` have truthy value. */
+export function assertFalse(expr: unknown, msg = ""): asserts expr is false {
```
bcheidemann, line 227:
```suggestion
export function assertNot(expr: unknown, msg = ""): asserts not expr {
```

This matches other assertions like `assertNotEquals`.

Additionally, `assertFalse` is misleading as this implies only the literal value `false` will pass this assertion when in actual fact this is asserting whether the expression is truthy.

testing/asserts_test.ts
```diff
@@ -40,11 +41,14 @@ Deno.test("testingEqualDifferentZero", () => {
 Deno.test("testingEqual", function (): void {
   assert(equal("world", "world"));
   assert(!equal("hello", "world"));
+  assertFalse(equal("hello", "world"));
```
bcheidemann, line 44: All the tests follow the format `assertFalse(equal(/* --snpi */))`.

Is this not the same as `assertNotEquals` but the output is less helpful for debugging?

IMHO we should lead by example and write tests in an idiomatic way wherever possible. I think these tests are not idiomatic because they are effectively testing the `equals` function which is already covered.

Could we instead write tests which assert different falsy values like `assertFalse(false)` and `assertFalse(null)` etc?

Additionally, it would be nice to assert that this assertion will fail when passed truthy values.

bcheidemann (changes requested):

kt3k:
@bartlomieju I think we reached the consensus on this. What do you think?

kt3k (approved):
LGTM

bartlomieju (approved):
Yes, we reached consensus. LGTM

testing/asserts.ts
UltiRequiem: That syntax is not valid TypeScript, but you are right, is misleading...

I'm not sure what we could do, any ideas @kt3k  @bartlomieju @bcheidemann ?

testing/asserts_test.ts
UltiRequiem: I just add these tests, I didn't remove the ones related to `equals` but if that's wanted I can do it too.

testing/asserts.ts
bcheidemann: I looked into this `assert <<cond>>` syntax a bit because I didn't know much about it. I think we could do something like this:

```ts
type Falsy = false | null | undefined | "" | 0;
export function assertNot(expr: unknown, msg = ""): asserts expr is Falsy {
  // --snip--
}
```

The type inference for this is actually pretty smart:

image

image

testing/asserts_test.ts
bcheidemann: Unless you think they add significant value over the new tests you added (which look great btw 😊) I would lean towards removing them.

Reason being that if we add lots of tests which are not needed then the tests will run slower. Of course, these tests wont make a significant difference on their own, but if every new feature adds 5 unnecessary tests then the test suite quickly becomes bloated.

What do you think @UltiRequiem ?

bcheidemann (commented):
New tests look good :)

kt3k:
Note: We also considered `assertNot` and `assertFalsy`, but prefer `assertFalse` the best because it sounds clearer to other choices

kt3k:
@bcheidemann Thank you for reviewing. I created an issue for the better typing suggestion #2170

testing/asserts_test.ts
UltiRequiem: Yes, it makes sense.

testing/asserts_test.ts
```diff
@@ -385,8 +456,14 @@ Deno.test("testingAssertObjectMatching", function (): void {
   const m = { foo: /abc+/i, bar: [/abc/g, /abc/m] };
   const n = {
     foo: new Set(["foo", "bar"]),
-    bar: new Map([["foo", 1], ["bar", 2]]),
-    baz: new Map([["a", a], ["b", b]]),
+    bar: new Map([
```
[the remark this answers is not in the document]
UltiRequiem: Sorry

PR #2098, opened 2022-04-10
fix(node): Make global.ts evaluate synchronously

### LINKED ISSUE #2097: `std/node` should avoid TLA
See https://github.com/denoland/deno/issues/14243. Imagine side-effect-importing some polyfill from esm.sh at your application entrypoint. That polyfill was built for Deno to import things from `std/node`. If it is asynchronously evaluated, you will get an undesired dependency evaluation order and the conventional `import "https://cdn.esm.sh/reflect-metadata?deno"` at your application entrypoint might not suffice.

In this case there is a single easily fixed culprit: https://github.com/denoland/deno_std/blob/8824ae1ef3e53a6531b464f9add1eb40a60b9d64/node/internal/util/debuglog.ts#L108-L113

cc @shivam-tripathi

nayeemrmn:
Fixes #2097.

### DIFF
--- a/node/global_test.ts
+++ b/node/global_test.ts
@@ -91,21 +91,26 @@ Deno.test("clearImmediate is correctly defined", () => {
 // https://github.com/denoland/deno_std/issues/2097
 Deno.test("global.ts evaluates synchronously", async () => {
   const tempPath = await Deno.makeTempFile({ suffix: ".ts" });
-  await Deno.writeTextFile(
-    tempPath,
-    `\
-    import "data:application/javascript,import '${
-      new URL("global.ts", import.meta.url).href
-    }'; console.log(globalThis.async ? 'async' : 'sync')";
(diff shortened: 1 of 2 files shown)


node/global_test.ts
```diff
@@ -86,3 +87,30 @@ Deno.test("clearImmediate is correctly defined", () => {
   assertStrictEquals(globalThis.clearImmediate, timers.clearImmediate);
   assertStrictEquals(window.clearImmediate, timers.clearImmediate);
 });
+
+// https://github.com/denoland/deno_std/issues/2097
+Deno.test("global.ts evaluates synchronously", async () => {
```
kt3k, line 92: Nice test to prevent the regression 👍

kt3k (approved):
LGTM. Thanks! @nayeemrmn

bartlomieju (approved):
LGTM, good fix and test

Issue #2099, opened 2022-04-11
crypto/randomInt marks maximum as inclusive instead of exclusive


albert-sun:
**Describe the bug**
Most crypto libraries, upon calling `randomInt(max)` or `randomInt(min, max)` generate a random number within the range [min, max), or min inclusive and max exclusive. The Node.js crypto library (https://nodejs.org/api/crypto.html#cryptorandomintmin-max-callback) follows this convention, with the arguments being labeled as:

```
min <integer> Start of random range (inclusive). Default: 0.
max <integer> End of random range (exclusive).
```

**Steps to Reproduce**
Generate many random numbers and check whether any randomized numbers equal the maximum. 
In theory, this should never occur with inclusive minimum and exclusive maximum.

<!--
1. Run a demonstration `random.ts` through Deno with the following content:
```import randomInt from "https://deno.land/std@0.134.0/node/_crypto/randomInt.ts";

const maximum = 100;
const offset = 50;
const numToGenerate = 10000;
for(let i = 0; i < numToGenerate ; i++) {
    // Test with randomInt(max: number) signature
    const randomizedMax = randomInt(maximum);
    if(randomizedMax === maximum) {
        throw new Error(`randomInt(max) should never generate maximum: generated ${randomizedMax} with max ${maximum}`);
    }

    // Test with randomInt(min: number, max: number) signature
    const randomizedMinMax = randomInt(offset, maximum + offset);
    if(randomizedMinMax === maximum + offset) {
        throw new Error(`randomInt(min, max) should never generate maximum: generated ${randomizedMinMax} with max ${maximum + offset}`);
    }
}
```

**Expected behavior**

No errors should be outputted and the code should exit silently.

**Environment**

- OS: Windows 10
- deno version: 1.20.5 (release, x86_64-pc-windows-msvc)
- std version: 0.134.0

albert-sun:
Within [crypto/randomInt.ts](https://deno.land/std@0.134.0/node/_crypto/randomInt.ts), the fix to generate with maximum exclusive would be (referencing this StackOverflow post [here](https://stackoverflow.com/questions/1527803/generating-random-whole-numbers-in-javascript-in-a-specific-range) for the appropriate syntax):
```diff
- const result = Math.floor(randomNumber * (max - min + 1)) + min;
+ const result = Math.floor(randomNumber * (max - min)) + min;
```

albert-sun:
[...]

PR #2101, opened 2022-04-11
Change crypto/randomInt (and associated test) to generate maximum exclusive instead of inclusive

albert-sun:
Referencing #2099 - committed the `randomInt` fix and modified the `between` assertion within `randomInt_test.ts` to disallow maximum inclusion when checking generated value.


kt3k (approved):
LGTM.

This is going to be a breaking change, but this suggestion looks reasonable to me. Thanks for your work!

PR #2109, opened 2022-04-14
chore: use FakeTime for testing parse

KyleJune:
I went ahead and updated the datetime module to use `FakeTime` from `testing/time.ts`. in its tests instead of using the `faketime` command for setting it up in CI.

This branches off of my other branch for adding `testing/time.ts`. The only changes that currently belong to this PR are in the latest commit (https://github.com/denoland/deno_std/commit/ecc28a70aec97cef3e2f3bfb897341dafa46c71f).

> We have a few test cases which depend on some specific date time. We currently use `faketime` command for setting it up in CI, but I hope we can replace those set ups with this utility
> 
> https://github.com/denoland/deno_std/blob/37b0b02f9114482ae7cd9bc7ea239785f3e31b02/datetime/test_2021_12_31.ts

_posted by @kt3k in https://github.com/denoland/deno_std/issues/2069#issuecomment-1098017284_


datetime/test.ts
```diff
@@ -64,6 +65,94 @@ Deno.test({
...
+    );
+
+    assertEquals(
+      datetime.parse("2020", "yyyy"),
+      new Date(2020, 11, 30),
+    );
+    assertEquals(
+      datetime.parse("2022", "yyyy"),
+      new Date(2022, 11, 30),
+    );
```
KyleJune, line 150: I added these because there were only tests for combination of month and day. I thought it would be good to have additional test coverage for the `dd`, `MM`, and `yyyy` formats.

When writing these tests, I was a bit surprised that the day was changing from 31 to 30. I think that is a bug. If it is and gets fixed, these MM and yyyy test case's expectated dates will need updated to have 31 for the day instead of 30.

I found the cause is a line in `DateTimeFormatter.prototype.partsToDate` from `formatter.ts`. Apparently `date.setHour(0, 0, 0, 0)` will also change the day because the `setHour` method is using the Timezone of the system rather than UTC but the output is always in UTC. I'm in central timezone, so the time gets set to "00:00:00.000 GMT-0600" which ends up changing the date to being "2021-12-30T06:00:00.000Z"

```ts
Deno.test("example", () => {
  const a = new Date("2021-12-31");
  a.setHours(0, 0, 0, 0);
  const b = new Date("2021-12-31");
  assertEquals(a, b);
});
```

```
AssertionError: Values are not equal:

    [Diff] Actual / Expected

-   2021-12-30T06:00:00.000Z
+   2021-12-31T00:00:00.000Z
```

Here is an example showing that behavior in Firefox developer console.

image

KyleJune: I had to remove the `MM` and `yyyy` test cases I previously added. The CI runs in UTC time, so the above issue doesn't happen in CI but does for any developers in a different timezone than UTC.

kt3k (approved):
@KyleJune LGTM. Thanks for working on this!

PR #2111, opened 2022-04-14
tech(node/EventEmitter) Improve typings

Duelsik:
This is an example how I would tackle the issue https://github.com/denoland/deno_std/issues/2113

EDIT: Note that it might require dependentants (is it a word? :D) to update overloads to proper signature.


kt3k:
This looks useful to me. What tasks are remaining to be ready?

Duelsik:
@kt3k Generally I have to review that and update static methods, will try to push it further today.

Duelsik:
(Pushed a broken commit)
Stuck a bit on weird behavior - look on test file and compare type on `ExtractedWithMapping` and `ExtractedExtendedWithMapping`.

In first example everything is inferred in the intended way, the 2nd... has correct keys, but values are "any"-ised.
I think the reason is that accessing "super generic type(?)" is done in different way.

node/events_test.ts
```diff
@@ -0,0 +1,95 @@
...
+import { EventEmitter, getEventListeners } from "./events.ts";
+import type { EventListenerType } from "./events.ts";
+import { expectType } from "https://raw.githubusercontent.com/TypeStrong/ts-expect/736658bd9f1c23ad9e2676c27e9bdc9297309ae9/src/index.ts";
+import type {
+  TypeEqual,
+} from "https://raw.githubusercontent.com/TypeStrong/ts-expect/736658bd9f1c23ad9e2676c27e9bdc9297309ae9/src/index.ts";
+
+// TODO: 1. Those tests doesn't do anything, just fail at compile-time when types are broken
+//          maybe exclude it to separate task taking <filename>_test_types.ts?
+// TODO: 2. Create a module in Deno.text for expecting types:
```
Duelsik, line 11: test*

Duelsik:
Got help on typescript discord, unblocked ^^

Duelsik:
Status update:
- The main change of typings is ready, test are done (and working since they catched my mistakes in f.e. return types)
- Most part was doublechecked with node docs (latest - v18, but it wasn't changed in this version, so should work with older versions too)
- I have to update EventEmitter dependencies, because most of them are overriding stuff like emit, but they don't have proper override syntax (since the signature is template func now) for example process.ts
- The https://github.com/denoland/deno_std/issues/2149 makes it kinda hard to judge how many places are failing because typechecks

Can I ask for something like `unstable/event-emitter` branch to merge to? The line count grows kinda quick and I am not sure if it wouldn't be easier if I will tackle stuff like updating bigger EventEmitter consumers one by one in separate PRs to this branch.
cc @kt3k

Duelsik:
// Got info to push everything here

Duelsik:
`node/process.ts` is ready to comment on

Duelsik:
Since `node/dgram.ts` Socket is heavily overriden - I will skip it in this PR.

Duelsik:
Imo actual state of `node/worker_threads.ts` is very straight-forwardly written... but it is done in the way that reduces readabilty, I am leaving this file for further discussion.
My suggestion is to move ParentPort to proper EventEmitter, the modification there looks like the only thing it wants to do in the other way that EventEmitter was written is to not accept `symbol` as event name.

cc @kt3k (pinging because you might have more understanding on worker/main threads in node)

Duelsik:
> Since `node/dgram.ts` Socket is heavily overriden - I will skip it in this PR.

same story with `node/_stream.d.ts`

Duelsik:
> Since `node/dgram.ts` Socket is heavily overriden - I will skip it in this PR.

same with `node/_readline.d.ts`

Duelsik:
Rebased

I wanted to add extending schema feature (growing template), but typescript says it is too sus and prevents taking "base" template as typings for "base" definition.

Duelsik:
> This looks useful to me. What tasks are remaining to be ready?

@kt3k Ready for review ^^

node/net.ts
```diff
@@ -1807,6 +1807,9 @@ function _emitErrorNT(server: Server, err: Error) {
 function _emitListeningNT(server: Server) {
   // Ensure handle hasn't closed
   if (server._handle) {
+    // TODO(duelsik): afaik it differs from node docs, but if it works - let's leave it with
```
cmorten, line 1810: This is an exact port of the Node code, see https://github.com/nodejs/node/blob/master/lib/net.js#L1414

Duelsik: @cmorten I think I did a mistake here, rechecked docs and net.Server doesn't have "socket" pushed alongside listening event. Will fix :)

node/events_test.ts
```diff
@@ -0,0 +1,1138 @@
+import { assert } from "../testing/asserts.ts";
```
kt3k, line 1: nit: please add the license header
```suggestion
// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
import { assert } from "../testing/asserts.ts";
```

node/events_test.ts
```diff
@@ -0,0 +1,1138 @@
...
+import { EventEmitter, getEventListeners, on, once } from "./events.ts";
+import { assertCallbackErrorUncaught } from "./_utils.ts";
+import type { EventListenerT } from "./events.ts";
+import { expectType } from "https://raw.githubusercontent.com/TypeStrong/ts-expect/736658bd9f1c23ad9e2676c27e9bdc9297309ae9/src/index.ts";
+import type { TypeEqual } from "https://raw.githubusercontent.com/TypeStrong/ts-expect/736658bd9f1c23ad9e2676c27e9bdc9297309ae9/src/index.ts";
```
kt3k, line 6: We prefer to avoid external dependencies in std. Can you copy and paste these utils to somewhere in std (inlining them in this file also works) as testing-only utilities?

node/events_test.ts
```diff
@@ -0,0 +1,1138 @@
...
+
+// TODO: 1. Those tests don't do much while running tests, they aren't testing the implementation but types added on top
+//          of mjs code, so their focus is to fail at compile-time when types are invalid
+//          Maybe exclude it to separate task taking <filename>_test_types.ts?
+// TODO: 2. Create a module in Deno.test for expecting types:
+//          Issue link https://github.com/denoland/deno_std/issues/2145
```
kt3k, line 12: Testing of types is totally fine! Let's keep discussing public type assertion APIs in #2145

node/events_test.ts
```diff
@@ -0,0 +1,1138 @@
...
+// TODO: 3. Not sure if NodeEventTarget and EventTarget should be tested if they aren't exported - not sure about their use-case
```
kt3k, line 13: Let's skip them at first pass

node/events_test.ts
```diff
@@ -0,0 +1,1138 @@
...
+    this.emit("error"); // Not enough args
+
+    this.emit("close", "Reason"); // OK
+    this.emit("close"); // OK
+  }
+}
+
+class EventEmitterWithoutMapping extends EventEmitter {
+  async test() {
+    // Everything is OK - fallback to unsafe listener map
```
kt3k, line 46: Ah, this is nice! Stricter typing is opt-in 👍

kt3k (commented):
Mostly looks good to me. Great improvement in type inference of event emitter handlers! The fallback to unsafe listener map looks also nice to me. (So I expect the breakage in user code is none or very minimal). Please remove the external dependencies in testing.

node/inspector.ts
```diff
@@ -9,7 +9,435 @@ const messageCallbacksSymbol = Symbol("messageCallbacks");
...
+// By the empty type I mean that it was covered
+// type SchemaListenerMap = { };
+
+export type InspectorListenerMap =
+  & Console.ConsoleListenerMap
+  & Debugger.DebuggerListenerMap
+  & HeapProfiler.HeapProfilerListenerMap
+  & Profiler.ProfilerListenerMap
+  & Runtime.RuntimeListenerMap;
+// & SchemaListenerMap;
```
cjihrig, line 438: This looks like it was unintentionally left in. Can you either remove it or add another comment explaining it.

node/process.ts
```diff
@@ -254,7 +259,25 @@ export function kill(pid: number, sig: Deno.Signal | number = "SIGTERM") {
...
 
-class Process extends EventEmitter {
+type NotImplementedListener = EventListenerT;
+type SignalListener = (signal?: number) => void;
+type ProcessListenerMap =
+  & {
+    SIGBREAK: SignalListener;
+    // internally used
+    warning: (warning: Warning) => void;
+    exit: (exitCode?: string | number) => void; // TODO: not sure about exitCode typings
```
cjihrig, line 269: According to the Node docs, this should be an integer if present I think. When would we want a string here?

node/worker_threads.ts
```diff
@@ -126,6 +126,7 @@ interface NodeEventTarget extends
   removeListener: NodeEventTarget["off"];
 }
 
+// TODO(duelsik): Post an issue about making this place less imperative and just export different class depending on `isMainThread`
```
cjihrig, line 129: I'm not sure I understand this comment. Is this a TODO to open a GitHub issue? If so, I think it can be removed from here.

cjihrig:
It's been a few months since the last review with no follow up, and now there are merge conflicts. I'm going to close this out, but feel free to pick this up again in the future.

PR #2115, opened 2022-04-16
fix(node): Use Web Crypto API via globalThis

### LINKED ISSUE #2114: node: executing crypto.randomBytes says crypto.getRandomValues ​​is not a function
**Describe the bug**
When I run crypto.randomBytes, it says crypto.getRandomValues ​​is not a function
**Steps to Reproduce**
1. enter to deno repl
2. ```> crypto.randomBytes(10)```
3. > Uncaught TypeError: crypto.getRandomValues is not a function
    at generateRandomBytes (https://deno.land/std@0.134.0/node/_crypto/randomBytes.ts:24:12)
    at Object.randomBytes (https://deno.land/std@0.134.0/node/_crypto/randomBytes.ts:68:12)
    at <anonymous>:2:8

**Expected behavior**

```
> crypto.randomBytes(10)
<Buffer (hex)>
```

**Environment**

- OS: MacOS 11
- deno version: 1.20.5
- std version: 0.134.0

> I looked at the release notes for std 0.135.0, there was no such issue fix.
>
> It is likely that the issue will be reproduced in 0.135.0 as well.

ayame113:
close #2114

I think the cause of #2114 is that the global variable `crypto` in the repl references the Node's Crypto API instead of the Web Crypto API.
This PR will replace `crypto.xxx` with `globalThis.crypto.xxx`.

I'm not really sure if this change fixes #2114 (I don't know how to test it). Therefore, there is still no certainty that this approach is suitable.

### DIFF
--- a/node/_crypto/crypto_browserify/randombytes.ts
+++ b/node/_crypto/crypto_browserify/randombytes.ts
@@ -28,10 +28,12 @@ export function randomBytes(
       for (let generated = 0; generated < size; generated += MAX_BYTES) {
         // buffer.slice automatically checks if the end is past the end of
         // the buffer so we don't have to here
-        crypto.getRandomValues(bytes.slice(generated, generated + MAX_BYTES));
+        globalThis.crypto.getRandomValues(
+          bytes.slice(generated, generated + MAX_BYTES),
+        );
       }
     } else {
-      crypto.getRandomValues(bytes);
+      globalThis.crypto.getRandomValues(bytes);
     }
   }
 
--- a/node/_crypto/randomBytes.ts
+++ b/node/_crypto/randomBytes.ts
@@ -16,12 +16,12 @@ function generateRandomBytes(size: number) {
   //Work around for getRandomValues max generation
   if (size > MAX_RANDOM_VALUES) {
     for (let generated = 0; generated < size; generated += MAX_RANDOM_VALUES) {
-      crypto.getRandomValues(
+      globalThis.crypto.getRandomValues(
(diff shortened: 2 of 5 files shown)


ayame113:
I think the cause of CI failure is https://github.com/denoland/deno/issues/14291. I tried `deno fmt --check` locally and it succeeds on Deno1.20.5 but fails on Deno1.20.6.

kt3k (approved):
This can be tested like the below (starts the `deno repl` in compat mode with local change):

```sh
DENO_NODE_COMPAT_URL=file:///path/to/local/deno_std/ deno repl --compat --unstable
```

And I confirmed this fixes the issue. Thanks for the fix!

kt3k:
but it feels a little strange why `crypto` in `node/_crypto/randomBytes.ts` was shadowed by repl's `crypto` 🤔

ayame113:
> And I confirmed this fixes the issue.

Thanks!!

> but it feels a little strange why `crypto` in `node/_crypto/randomBytes.ts` was shadowed by repl's `crypto` 🤔

The chrome repl behaves the same:

image

It seems that the handling of global variables is different when imported from script context (that is, repl) and when imported from module context. I'm not sure why this is happening.

kt3k:
Node.js module preloading was added here https://github.com/denoland/deno/pull/13127/files

Node.js modules are declared as scoped variables. So this still doesn't describe the shadowing of `crypto` etc in my view.

But anyways the fix is confirmed. So let's try this change.

Issue #2107, opened 2022-04-12
http/file_server.ts is extremely slow


bartlomieju:
https://twitter.com/jarredsumner/status/1513878964531519494

Deno:
```
> deno run -A https://deno.land/std/http/file_server.ts
```

```
❯ wrk -t8 -c20 -d10s http://localhost:4507/foo.txt
Running 10s test @ http://localhost:4507/foo.txt
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    40.30ms  153.78ms   1.01s    94.01%
    Req/Sec   667.28    104.16   787.00     93.13%
  48709 requests in 10.02s, 3.45GB read
Requests/sec:   4863.46
Transfer/sec:    353.12MB
```

Node:
```js
const http = require("http");
const fs = require("fs");

http.createServer(function (req, res) {
    const filename = __dirname + req.url;

    const readStream = fs.createReadStream(filename);

    readStream.on("open", function() {
        readStream.pipe(res);
    });

    readStream.on("error", function(err) {
        res.end(err);
    })
}).listen(8081);
```

```
node server.js
```

```
❯ wrk -t8 -c20 -d10s http://localhost:8081/foo.txt
Running 10s test @ http://localhost:8081/foo.txt
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     0.98ms  463.68us  20.02ms   99.08%
    Req/Sec     2.07k   161.45     2.31k    74.38%
  166351 requests in 10.10s, 11.76GB read
Requests/sec:  16471.51
Transfer/sec:      1.16GB

```

I expect there's an `await` somewhere in `file_server.ts` that effectively limits the concurrency.

littledivy:
Fixed in https://github.com/denoland/deno/pull/14284 and https://github.com/denoland/deno/pull/14366

note that the Node.js server above is not equivalent to the std file_server. It's missing etag generation, header ranges and other complex stuff. Here's the benchmark anyways:

`sample.txt` is 5mb

Deno `std/http/file_server.ts`:
```
wrk -t8 -c20 -d10s http://localhost:4507/sample.txt
Running 10s test @ http://localhost:4507/sample.txt
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency    17.58ms    2.53ms  44.05ms   88.56%
    Req/Sec   114.07     12.06   141.00     67.20%
[...]

PR #2116, opened 2022-04-16
perf(http): optimize file server

littledivy:
Towards #2107 

Major changes:

- use native readable streams optimization. `Deno.File#readable` instead of a wrapper stream. See https://github.com/denoland/deno/pull/14284
- use fnv-1a default etag hash function. Adds an option to allow using others.
- ~~faster hex function for out of other algorithms (like `sha-1`), using hex lookup table.~~
- disables logging by default when invoked independently. Replaces the `--quiet` flag with `--verbose`. Public API remains unchanged.
- ~~use simpler extname extraction to avoid scanning manually. (downsides?)~~

Benchmarks:

`main`:
```
Running 10s test @ http://localhost:4507/README.md
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.44ms  522.77us  26.73ms   97.55%
    Req/Sec     1.40k   119.07     2.63k    97.39%
  112288 requests in 10.10s, 432.42MB read
Requests/sec:  11114.28
Transfer/sec:     42.80MB
```

This patch:
```
wrk -t8 -c20 -d10s http://localhost:4507/README.md
Running 10s test @ http://localhost:4507/README.md
  8 threads and 20 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.01ms  163.42us   6.32ms   85.73%
    Req/Sec     1.99k    67.51     2.09k    86.51%
  160068 requests in 10.10s, 611.53MB read
Requests/sec:  15848.72
Transfer/sec:     60.55MB
```

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -42,7 +42,8 @@ interface FileServerArgs {
   quiet: boolean;
 }
 
-const encoder = new TextEncoder();
+// @ts-ignore `Deno.core` is not a public API
+const encode = Deno.core.encode || new TextEncoder().encode;
 
 const MEDIA_TYPES: Record<string, string> = {
   ".md": "text/markdown",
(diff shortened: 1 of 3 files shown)


http/file_server.ts
```diff
@@ -42,7 +42,8 @@ interface FileServerArgs {
   quiet: boolean;
 }
 
-const encoder = new TextEncoder();
+// @ts-ignore `Deno.core` is not a public API
+const encode = Deno.core.encode || new TextEncoder().encode;
```
lucacasonato, line 46: Not in favor of this one.

http/file_server.ts
```diff
@@ -174,20 +175,51 @@ const MEDIA_TYPES: Record<string, string> = {
 
 /** Returns the content-type based on the extension of a path. */
 function contentType(path: string): string | undefined {
-  return MEDIA_TYPES[extname(path)];
+  const ext = path.split(".").pop();
```
lucacasonato, line 178: Optimize extname instead.

http/file_server.ts
```diff
@@ -174,20 +175,51 @@ const MEDIA_TYPES: Record<string, string> = {
...
+  for (let i = 0; i < buf.length; i++) {
+    hash ^= buf.charCodeAt(i);
+    // Equivalent to `hash *= 16777619` without using BigInt
+    // 32-bit FNV prime
+    hash += (hash << 1) + (hash << 4) + (hash << 7) + (hash << 8) +
+      (hash << 24);
+  }
+  // 32-bit hex string
+  return (hash >>> 0).toString(16);
+}
```
lucacasonato, line 208: Is this not available is std/crypto?

http/file_server.ts
```diff
@@ -174,20 +175,51 @@ const MEDIA_TYPES: Record<string, string> = {
...
+  hexLookup[n] = hexOctet;
+}
+
+function hex(buf: ArrayBuffer): string {
+  const u8arr = new Uint8Array(buf);
+  const len = u8arr.length;
+  let hexOctets = "";
+  for (let i = 0; i < len; ++i) hexOctets += hexLookup[u8arr[i]];
+  return hexOctets;
+}
```
lucacasonato, line 194: Use std/encoding/hex and put this optimized encoder in there

testing/asserts.ts
```diff
@@ -97,7 +97,9 @@ function buildMessage(
         ? createColor(detail.type, { background: true })(detail.value)
         : detail.value
     ).join("") ?? result.value;
-    diffMessages.push(c(`${createSign(result.type)}${line}`));
+    diffMessages.push(c(`${
+      createSign(result.type)
+    }${line}`));
```
lucacasonato, line 102: Unrelated change?

lucacasonato (commented):
Great that you have managed to optimize this. This is code we want people to read and understand though, so we shouldn't duplicate functionality from other std/ modules.

testing/asserts.ts
bartlomieju: I've seen this change recently in some other PR, seems to be related to dprint. CC @dsherret

http/file_server.ts
bartlomieju: Agreed, we should just use `TextEncoder`

http/file_server.ts
littledivy: nope, can't find it

http/file_server.ts
littledivy: Didn't know about encoding/hex, it is equivalent in terms of performance. Though we should add a Uint8Array -> hex string function.

http/file_server.ts
littledivy: fine, comes with a performance penalty but whatever. Updated the benchmark results.

http/file_server.ts
littledivy: extname is fine actually, reverted.

bartlomieju (approved):
LGTM

http/file_server.ts
```diff
@@ -233,16 +251,20 @@ function fileLenToString(len: number): string {
...
 ): Promise<Response> {
-  const [file, fileInfo] = await Promise.all([
-    Deno.open(filePath),
-    Deno.stat(filePath),
-  ]);
-
+  const file = await Deno.open(filePath);
+  if (fileInfo === undefined) {
+    fileInfo = await Deno.stat(filePath);
+  }
```
lucacasonato, line 267: This serializes the calls. You can still have them be parallel while adding this optimization.

http/file_server.ts
```diff
@@ -233,16 +251,20 @@ function fileLenToString(len: number): string {
...
  * @param filePath Path of the file to serve.
+ * @param etagAlgorithm The algorithm to use for generating the ETag. Defaults to "fnv1a".
+ * @param fileInfo An optional FileInfo object returned by Deno.stat. It is used
+ * for optimization purposes.
  */
 export async function serveFile(
   req: Request,
   filePath: string,
+  etagAlgorithm: EtagAlgorithm = "fnv1a",
+  fileInfo?: Deno.FileInfo,
```
lucacasonato, line 262: These two should be an option bag according to our style guide I think (public api)

lucacasonato (approved):
LGTM!

PR #2117, opened 2022-04-16
feat(testing): add TestContext argument to it

KyleJune:
The assertSnapshot function from https://github.com/denoland/deno_std/pull/2039 that is being worked on will need access to the `Deno.TestContext` for the test case it is used in. This updates `it` to take a function that will have the `Deno.TestContext` passed to it as its first and only argument. This will also enable making use of the test step api within an `it` function.

The `deno fmt` command also made me update the `assert.ts` file.


bartlomieju (approved):
LGTM, thank you

PR #2120, opened 2022-04-18
chore(ci): add informational Codecov status checks

thomasrockhu-codecov:
Hi, Tom from Codecov here. I noticed that you were using Codecov but weren't actually getting any notifications on pull requests. I figured it would be useful to get some idea if code being changed is being tested, but also not blocking CI/merging. Let me know if this makes sense or if we can do something that would be helpful.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -3,7 +3,11 @@ codecov:
   require_ci_to_pass: true
 coverage:
   status:
-    project: off
-    patch: off
+    project:
+      default:
+        informational: true
+    patch:
+      default:
+        informational: true
 ignore:
   - examples


kt3k:
@thomasrockhu-codecov Thanks for the PR.

> but also not blocking CI/merging.

Does this mean the status from codecov is always green with this settings?

kt3k:
@thomasrockhu-codecov Thanks for the clarification. Can you also sign the CLA from the above link?

I think this is worth trying.

@bartlomieju What do you think?

lucacasonato:
Seems fine.

bartlomieju:
Seems good to me. @thomasrockhu-codecov can you sign the CLA so we can land the PR?

littledivy (approved):

kt3k (approved):

Issue #2123, opened 2022-04-19
Unable to use one `AbortSignal` instance with multiple HTTP servers


jaydenseric:
**Describe the bug**

Currently one user provided `AbortSignal` instance can't be used to control multiple Deno std library HTTP servers running at the same time, because under the hood instead of using the `addEventListener` method to safely listen for `abort` events, `onabort` properties are set which overrides previously set handlers.

- https://github.com/denoland/deno_std/blob/f482c486446b72d603f493ba8e4044c5f59e6d84/http/server.ts#L512
- https://github.com/denoland/deno_std/blob/f482c486446b72d603f493ba8e4044c5f59e6d84/http/server.ts#L554
- https://github.com/denoland/deno_std/blob/f482c486446b72d603f493ba8e4044c5f59e6d84/http/server.ts#L609
- https://github.com/denoland/deno_std/blob/f482c486446b72d603f493ba8e4044c5f59e6d84/http/server.ts#L655
- https://github.com/denoland/deno_std/blob/f482c486446b72d603f493ba8e4044c5f59e6d84/http/server.ts#L707

Also, if the user already has an `onabort` property handler set on the `AbortSignal` instance, Deno std destroys it by overriding it. This is an unexpected behavior that could have significant consequences in a user's codebase.

I have a PR incoming to fix this bug, so please no one else start working on it :)

**Steps to Reproduce**

1. Construct a new `AbortController`.
2. Use that abort controller’s `signal` as the `signal` option for multiple Deno std library HTTP servers.
3. Call the abort controller’s `abort` method. Notice that only one of the running servers handles the `abort` event and closes.

**Expected behavior**

A single `AbortSignal` instance should be able to be used with multiple Deno std library HTTP servers; when the `abort` event is dispatched all the servers should close. Also, if the user already has an `onabort` property handler set on the `AbortSignal` instance, Deno std should not destroy it by overriding it.

**Environment**

- OS: MacOS 12.3.1
- deno version: 1.20.6 (release, aarch64-apple-darwin)
- std version: 0.135.0

PR #2124, opened 2022-04-19
fix(http): Use the `addEventListener` method instead of `onabort` properties

jaydenseric:
Fixes https://github.com/denoland/deno_std/issues/2123 .


kt3k (commented):
Good catch. Thanks! Can you add test cases for this fix? (Maybe let's skip `listenAndServe` and `listenAndServeTls` as they are deprecated now)

kt3k (approved):

PR #2136, opened 2022-04-21
Update file server tests

chenfeng3695:
https://dash.deno.com/examples/log.js


chenfeng3695:
> https://dash.deno.com/examples/log.js

PR #2137, opened 2022-04-21
feat(testing)!: Infer stub and spy return types from properties

### LINKED ISSUE #2121: testing/mock.ts: Infer spy function arguments when using replacing a function on an object.

Say you want to spy a function on an object:

```ts
const myObject = {
    myFunction(x: string, y: number) {}
}
const mySpy = spy(myObject, "myFunction");
```

Currently the returned spy has the type `Spy<Self, unknown[], any>`, making types less useful down the line when asserting against arguments and return values.

Infer the argument and return type.
[Here's a working prototype](https://www.typescriptlang.org/play?ssl=39&ssc=12&pln=39&pc=7#code/JYOwLgpgTgZghgYwgAgMoAcCeBhOAbPAHlQjxmQF5k4RMAaZAQSgHMBnZCAD0hABMOAVxABrEAHsA7iADaAXUrVa8hgCUIYQVBCKamAHzIA3gChk56qzYAuJlYDcZi1A1aQEPgH5b6zdscWnFBQ4lDeyACiwaEBFmykMOEkZI4AviYmYJjoKADiGgAKcFBwALYa0GwAYiGlBSHoxAkM9eLonDwQ-BwiEJji5MkwhlRDMq3oCty8AsgAFAB0S8XstqAw0HbsAJSUhnrInk7mzOzI1sfIwmJSsnKOmdl5Gr5uNeJ1DU1kLQ0dMz0+gM0AkRiCyOMGlNOt15ksFisbEpMPJdhRDOtNq9tIdLtidBdAtcJNIHqBILBECgMJhvuQqHoGKcONMurNibd5LplHI1K4cQzaIZTIE5mAABbAJFDBjwxG2ZnbHz8kCxcyhYAsUD4WxiyXS5rIOVWBVWNGGfFq5AIfB4aVYXAEOlMqx8vwgfTyK0uNhgUIeWwAI3E4jwEBo3ogvv9cyVyAAbuJgHw0iZuOhQmBkDBhAgwMBxDo2Fg6dz6FsWTD2aISXcyypkPjFMI+BAYKAPPpY7Yac6K263PpHOnM9nc-nC8hi7SZcz-myhDXObz8V2cyAELqJVKe4bjasK+bGyq472Zf3j+6h2muBmoFn13mC0WSzKJvPYb1+oNQXN4mRdx+ZB0AaaAslsL9gSGU9X0NfIwCKEpygpapagmPsJn0Bh4Pxd5PjaDCGn0a8R3vMcNwnF8Z2aOdWVhDlpBUVc5icR8AHkoCGcI5j1HdwRgWVlhNQ89kvNxdgAH34ugnBAtowMwcJIJ-H4TBg6igOZAdtGFS4XHdK5+DbDs+GoRcbmkMy0FgzTXTEnTUxMBBC19ZBShwEM8DYwMACsIDzRQRQsdyqnHZ85i4WwQEEUpA2gBhMFsX0oFAFhdiCwJzH0txkDAKBBAgK10nSJyXKzdyaUUac5nc7BPO8vy8wYAAiEKwsLZrtkcCqsEcAB6Pq3MwSrxTgDgVhirowA4GRIuQaLYvi5BEqnPLUoUGhTOynEshyZBg1DcMQCAA). Hover over `mySpy` at the bottom to see its type.
Let me know if this is desired behaviour, I'd be glad to do a PR.
Though keep in mind this would be a breaking change for any existing calls that already provided their own generic arguments. 

I realise you can pass the types in the generic arguments, i.e.:

```ts
const mySpy = spy<Self, [string, number], void>(myObject, "myFunction");
```

but I mostly work with JavaScript, rather than TypeScript, where this gets very verbose:

```js
/** @type {import("https://deno.land/std@0.135.0/testing/mock.ts").Spy<typeof myObject, [number, string], void>} */
const mySpy = spy(myObject, "myFunction");
```

and often the arguments themselves need to be imported inline as well, making this even worse.

jespertheend:
This has a breaking change because the generic parameters have changed from `spy<Self, Args, Return>()` to `spy<Self, Prop>()`.
Only the types have been changed so there should be no runtime difference.

There's still a type error that I'm not sure how to resolve, which is why this is still a draft. Initially I thought it was [a bug in TypeScript](https://github.com/microsoft/TypeScript/issues/48805). But it seems like that's not the case.
Adding `// @ts-ignore` seems to work without breaking any other types, but I'd prefer to find a better way around this.

Fixes #2121

### DIFF
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -371,8 +371,10 @@ export function stub<
   property: Prop,
   func: (
     this: Self,
-    ...args: Partial<GetParametersFromProp<Self, Prop>>
-  ) => GetReturnFromProp<Self, Prop>,
+    // deno-lint-ignore no-explicit-any
+    ...args: any[]
+    // deno-lint-ignore no-explicit-any
+  ) => any,
 ): Stub<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
 export function stub<
   Self,
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -371,8 +371,7 @@ export function stub<
   property: Prop,
   func: (
     this: Self,
-    // deno-lint-ignore no-explicit-any
-    ...args: any[]
+    ...args: GetParametersFromProp<Self, Prop>
   ) => GetReturnFromProp<Self, Prop>,
 ): Stub<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
 export function stub<
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -289,6 +289,21 @@ function methodSpy<
   return spy;
 }
 
+/** Utility for extracting the arguments type from a property */
+type GetParametersFromProp<
+  Self,
+  Prop extends keyof Self,
+> = Self[Prop] extends (...args: infer Args) => unknown ? Args
+  : unknown[];
+
+/** Utility for extracting the return type from a property */
+type GetReturnFromProp<
+  Self,
+  Prop extends keyof Self,
+> // deno-lint-ignore no-explicit-any
+ = Self[Prop] extends (...args: any[]) => infer Return ? Return
+  : unknown;
+
 /** Wraps a function or instance method with a Spy. */
 export function spy<
   // deno-lint-ignore no-explicit-any
@@ -304,9 +319,11 @@ export function spy<
 >(func: (this: Self, ...args: Args) => Return): Spy<Self, Args, Return>;
 export function spy<
   Self,
-  Args extends unknown[],
-  Return,
->(self: Self, property: keyof Self): Spy<Self, Args, Return>;
+  Prop extends keyof Self,
+>(
+  self: Self,
+  property: Prop,
+): Spy<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
 export function spy<
   Self,
   Args extends unknown[],
@@ -341,19 +358,22 @@ export interface Stub<
 /** Replaces an instance method with a Stub. */
 export function stub<
   Self,
-  // deno-lint-ignore no-explicit-any
-  Args extends unknown[] = any[],
-  Return = undefined,
->(self: Self, property: keyof Self): Stub<Self, Args, Return>;
-export function stub<
-  Self,
-  Args extends unknown[],
-  Return,
+  Prop extends keyof Self,
 >(
   self: Self,
-  property: keyof Self,
-  func: (this: Self, ...args: Args) => Return,
-): Stub<Self, Args, Return>;
+  property: Prop,
+): Stub<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
+export function stub<
+  Self,
+  Prop extends keyof Self,
+>(
+  self: Self,
+  property: Prop,
+  func: (
+    this: Self,
+    ...args: Partial<GetParametersFromProp<Self, Prop>>
+  ) => GetReturnFromProp<Self, Prop>,
+): Stub<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
 export function stub<
   Self,
   Args extends unknown[],
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -373,8 +373,7 @@ export function stub<
     this: Self,
     // deno-lint-ignore no-explicit-any
     ...args: any[]
-    // deno-lint-ignore no-explicit-any
-  ) => any,
+  ) => GetReturnFromProp<Self, Prop>,
 ): Stub<Self, GetParametersFromProp<Self, Prop>, GetReturnFromProp<Self, Prop>>;
 export function stub<
   Self,
--- a/testing/mock_test.ts
+++ b/testing/mock_test.ts
@@ -412,8 +412,23 @@ Deno.test("stub function", () => {
   );
   assertEquals(func.restored, true);
 
+  // @ts-expect-error Stubbing with incorrect argument types should cause a type error
+  stub(new Point(2, 3), "explicitTypes", (_x: string, _y: number) => true);
+
+  // @ts-expect-error Stubbing with an incorrect return type should cause a type error
+  stub(new Point(2, 3), "explicitTypes", () => "string");
+
+  // Stubbing without argument types infers them from the real function
+  stub(new Point(2, 3), "explicitTypes", (_x, _y) => {
+    // `toExponential()` only exists on `number`, so this will error if _x is not a number
+    _x.toExponential();
+    // `toLowerCase()` only exists on `string`, so this will error if _y is not a string
+    _y.toLowerCase();
+    return true;
+  });
+
   // Stubbing without argument types should not cause any type errors:
-  const explicitTypesFunc = stub(point, "explicitTypes", () => true);
+  const explicitTypesFunc = stub(new Point(2, 3), "explicitTypes", () => true);
 
   // Check if the returned type is correct:
   assertThrows(() => {
--- a/testing/mock_test.ts
+++ b/testing/mock_test.ts
@@ -127,6 +127,7 @@ Deno.test("spy function", () => {
   );
   assertEquals(func.restored, false);
 
+  // Check if the returned type is correct:
   const explicitTypesSpy = spy(point, "explicitTypes");
   assertThrows(() => {
     assertSpyCall(explicitTypesSpy, 0, {
@@ -411,7 +412,10 @@ Deno.test("stub function", () => {
   );
   assertEquals(func.restored, true);
 
-  const explicitTypesFunc = stub(point, "explicitTypes");
+  // Stubbing without arguments and return type should not cause any type errors:
+  const explicitTypesFunc = stub(point, "explicitTypes", () => {});
+
+  // Check if the returned type is correct:
   assertThrows(() => {
     assertSpyCall(explicitTypesFunc, 0, {
       // @ts-expect-error Test if passing incorrect argument types causes an error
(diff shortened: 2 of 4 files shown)


testing/mock.ts
```diff
@@ -341,19 +358,22 @@ export interface Stub<
...
+  Prop extends keyof Self,
 >(
   self: Self,
-  property: keyof Self,
-  func: (this: Self, ...args: Args) => Return,
-): Stub<Self, Args, Return>;
+  property: Prop,
+  func: (
+    this: Self,
+    ...args: Partial<GetParametersFromProp<Self, Prop>>
```
KyleJune, line 374: I think this should just be unknown[] since partial won't work. That would allow it to be replaced with a function that doesn't need any specific arguments.

jespertheend: That sounds fine by me, as I don't have to fix the type error then xD
It seems like it should be `any[]` even, otherwise the `stub()` call in the "assertSpyCallAsync method" test gives a type error saying that the `x?: number` and `y?: number` parameters are not assignable to `unknown`.

The downside of this is that like that there is no type checking performed at all when using the `stub` function.

KyleJune: The returned stub has its Args derived from the prop on the instance. So calling the stub function would still have type checking and so would the assertions on the stub calls.

The only thing that wouldn't have type checking is the func arg, which didn't have type checking before either. Before your proposed change, the stub could only be called directly with the same args as the func had.

testing/mock.ts
```diff
@@ -341,19 +358,24 @@ export interface Stub<
...
-  property: keyof Self,
-  func: (this: Self, ...args: Args) => Return,
-): Stub<Self, Args, Return>;
+  property: Prop,
+  func: (
+    this: Self,
+    // deno-lint-ignore no-explicit-any
+    ...args: any[]
+    // deno-lint-ignore no-explicit-any
+  ) => any,
```
KyleJune, line 377: I think it would be good to keep the return type how you had it unless it caused issues. That would help prevent people from using stubs that return the wrong type of values.

testing/mock.ts
jespertheend: Ah yeah good point. Fwiw, type checking did seem to work with the Partial type in place. But that obviously meant no other function signatures were allowed. So `any[]` seems like the better option here.

testing/mock.ts
jespertheend: It's causing some issues in `time_test.ts` when trying to stub symbols on `Date.prototype`. But I see now it was getting type errors there anyway.
I'm not happy about the need to cast to `any` in order to fix it, but I'm not sure if there's a better way.

KyleJune: I think that time test case just isn't needed. I made it to basically test that all the methods call through to the original, passing all the same arguments and returning same return value. I just used symbols since they were all different. I think it would be fine to just change that test to verifying one method and one symbol gets replaced correctly and we could change it to passing in the correct args instead of using symbols. Spy on the original date functions instead of stubbing them.

jespertheend: Hmm not sure if I did this correctly, but I've added an extra commit that makes it only check if toISOString passes through to the original.

jespertheend: Hey I just started using `stub()` in production tests and I think it could be pretty beneficial to have type checking for the stubbed function. Right now all parameters are basically `any` without causing an "explicit any" type error, so you don't get any warnings about it either.

Did you specifically want to allow for changing the types of arguments entirely, or were you mostly concerned about passing in a function with zero arguments? Because it seems like using the `GetParametersFromProp<Self, Prop>` from earlier (without `Partial`) still allows for that. I.e.:

```ts
const obj = {
    someFn(x: number, y: string) {}
}
stub(obj,"someFn", () => {})
```

works fine without any type errors.

The reason why I'm asking is because in JavaScript with JSDoc it's pretty tedious to add types to arrow functions that are immediately passed in other functions, since you have to declare the function separately and then pass it in.

So say you have `obj.someFn` with the signature `(x: number, y: string) => void`, and you want to stub it while maintaining argument types, in TypeScript this would look like.

```ts
stub(obj,"someFn", (x: number, y: string) => {
    // do something with x and y
})
```

In JavaScript with JSDoc this would look like:
```js
/**
 * @param {number} x
 * @param {string} y
 */
function replaceFn(x, y) {
    // do something with x and y
}
stub(obj,"someFn", replaceFn)
```

jespertheend: I've removed `any[]` as arguments and I've added a few more tests to verify that the inferred types are correct. There's also a test to check if mocking a function without any arguments is possible.
But if you prefer to leave it as `any[]` please let me know.

KyleJune: As long as the stub helpers like returnsNext will still work when stubbing an instance method that has args on all it's call signatures, your change is fine. It sounds like it would be fine since you said `stub(obj,"someFn", () => {})` didn't give you a type issue.

I can test your changes this weekend on a project of mine that uses spies and stubs and let you know if I find any issues.

jespertheend: > As long as the stub helpers like returnsNext will still work

Ah I see, I'll add an extra test for that just in case.

> I can test your changes this weekend on a project of mine that uses spies and stubs

Sounds good 👍

KyleJune:
I checked latest commit against one of my modules ([oauth2_server](https://deno.land/x/oauth2_server)). I ended up getting a bunch of warnings but I guess they are reasonable and I should probably improve my stubs.

Some seem a little silly but are easy to fix. For example, the following error is happening because it's resolving to a boolean. If I cast false as false, the issue gets resolved.

image

Another example was where a function could resolve to undefined, but if I do `Promise.resolve()`, the type would be `Promise<void>` instead of `Promise<undefined>`, causing a warning. Updating it to explicitly resolve undefined like `Promise.resolve(undefined)` resolved the issue.

I've got no issues with your proposed change.

testing/time_test.ts
```diff
@@ -69,62 +69,21 @@ Deno.test("Fake Date instance methods passthrough to real Date instance methods"
...
-        }
-      },
+
+    const func = spy(
+      _internals.Date.prototype,
+      "toISOString",
     );
+    try {
+      now.toISOString();
+      assertSpyCall(func, 0, {
```
KyleJune, line 79: For this test, I think it would also be good to test that a symbol calls through to the original symbol too.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/@@toPrimitive

For that, you would spy on Symbol.toPrimitive as the property. Then if you do `${now}` it should call it with 'string' as the arg and return a string.

jespertheend:
Thanks for looking into this!

I've added the test you mentioned. For the return value it's only checking if the type is a string, I'm not sure how standardised calls to toPrimitive are on different platforms.

testing/time_test.ts
```diff
@@ -70,19 +70,33 @@ Deno.test("Fake Date instance methods passthrough to real Date instance methods"
...
+    const func2 = spy(
+      _internals.Date.prototype,
+      Symbol.toPrimitive,
+    );
+    try {
+      `${now}`;
+      assertSpyCall(func2, 0, {
+        args: ["string"],
+      });
+      assert(typeof func2.calls[0].returned === "string");
```
KyleJune, line 97: maybe it would be better to do 0+now, I think that would cause Symbol.toPrimitive to be called with "number" and return a number value which would be consistent across systems.

jespertheend: This seems to come with a few other problems. First I'd have to cast `now as unknown as number` in order to not get a type error. And it seems to call toPrimitive with `"default"` rather than `"number"`, so it ends up returning the same string as with the `${now}` approach.

But I looked at the spec for a bit and it seems Date.toPrimitive is pretty well defined, so I think it'll be pretty safe to test for the exact string.

jespertheend: Ah never mind, seems like `Number(now)` calls it with argument `"number"`

testing/mock_test.ts
```diff
@@ -126,6 +126,17 @@ Deno.test("spy function", () => {
...
+  // Check if the returned type is correct:
+  const explicitTypesSpy = spy(point, "explicitTypes");
+  assertThrows(() => {
+    assertSpyCall(explicitTypesSpy, 0, {
+      // @ts-expect-error Test if passing incorrect argument types causes an error
+      args: ["not a number", "string"],
+      // @ts-expect-error Test if passing incorrect return type causes an error
+      returned: "not a boolean",
+    });
+  });
```
KyleJune, line 139: Could you add a call that is valid with an assertSpyCall is valid to show it won't give typescript errors when you pass expected args and returned types? Currently there are only tests for when it catches typescript errors.

Maybe something like.

```ts
explicitTypesSpy(1, "hello");
assertSpyCall(explicitTypesSpy, 0, {
  args: [1, "hello"],
  returned: true,
});
```

Same for stub.

KyleJune (approved):

bartlomieju:
@jespertheend could you explain how this is a breaking change?

@kt3k please take a look too

jespertheend:
@bartlomieju The breaking change is that the `spy()` and `stub()` functions take different generic parameters now:
`spy<Self, Args, Return>()` =>  `spy<Self, Prop>()`.
And because the returned type is not `unknown` anymore, this means typing will be more strict in cases where you do something with the returned type.

I don't think this will cause any major issues since the new functionality will largely allow you to omit the generic arguments altogether. The generic arguments will be inferred in most cases so you won't have to provide them yourself.
Any new type errors resulted by the returned type will likely only expose already existing issues.

@KyleJune tried this change out against [oauth2_server](https://deno.land/x/oauth2_server) (see [comment](https://github.com/denoland/deno_std/pull/2137#issuecomment-1107476286)) and reported that it caused a bunch of new warnings:
> I ended up getting a bunch of warnings but I guess they are reasonable and I should probably improve my stubs.

bartlomieju:
Thanks for explanation @jespertheend! This seems reasonable, I'm all for stricter typings that help catch more bugs.

kt3k:
This API is still very young (< 1 month). So I think it's ok to break a little bit.

PR #2138, opened 2022-04-22
fix(wasi): ContextOptions is an optional argument

tani:
According to the node.js docs,
we can use WASI class without arguments. I mean that the `ContextOptions` should be optional.
This is a problematic to load https://github.com/ruby/ruby.wasm .

See also: https://nodejs.org/api/wasi.html

### DIFF
--- a/wasi/snapshot_preview1.ts
+++ b/wasi/snapshot_preview1.ts
@@ -346,7 +346,7 @@ export default class Context {
 
   exports: Record<string, WebAssembly.ImportValue>;
 
-  constructor(options: ContextOptions) {
+  constructor(options?: ContextOptions = {}) {
     this.#args = options.args ?? [];
     this.#env = options.env ?? {};
     this.#exitOnReturn = options.exitOnReturn ?? true;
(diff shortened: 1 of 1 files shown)


wasi/snapshot_preview1.ts
```diff
@@ -346,7 +346,7 @@ export default class Context {
 
   exports: Record<string, WebAssembly.ImportValue>;
 
-  constructor(options: ContextOptions) {
+  constructor(options?: ContextOptions = {}) {
```
kt3k, line 349: This seems throwing an type error

```
error: TS1015 [ERROR]: Parameter cannot have question mark and initializer.
  constructor(options?: ContextOptions = {}) {
              ~~~~~~~
    at file:///Users/runner/work/deno_std/deno_std/wasi/snapshot_preview1.ts:349:15
```

kt3k (commented):
Looks like a good idea. `Context` is directly used as `wasi.Wasi` object in `std/node`, so this is also a fix for `node/wasi`. Left one comment

tani:
@kt3k Thank you for giving the review. I fixed it. Cheers,

kt3k (approved):
LGTM

PR #2141, opened 2022-04-23
fix(crypto): Browser compatibility

YXL76:
Chrome disables `WebAssembly.Module` and `WebAssembly.Instance` for buffers larger than 4 KB.

### DIFF
--- a/_wasm_crypto/_build.ts
+++ b/_wasm_crypto/_build.ts
@@ -116,8 +116,8 @@ import wasmBytes from "./crypto.wasm.mjs";
 ${
   generatedScript.replace(
     /^const wasm_url =.*?;\nlet wasmCode =.*?;\n.*?const wasmInstance =.*?;\n/sm,
-    `const wasmModule = new WebAssembly.Module(wasmBytes);\n` +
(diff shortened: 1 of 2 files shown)


bartlomieju (changes requested):
@YXL76 thank you for the PR, please explain the changes you did. There are no changes in tests, so I'm not sure what's being fixed here.

YXL76:
As what I say, this is for Chrome, which disables synchronous compilation and instantiation of  large buffer on the main thread. I just convert `WebAssembly.Module` and `WebAssembly.Instance`   to their async version, which follows the error message.

> Uncaught RangeError: WebAssembly.Compile is disallowed on the main thread, if the buffer size is larger than 4KB. Use WebAssembly.compile, or compile on a worker thread.

kt3k:
Side note: A little concern about this change is that now this module can't be converted to CJS by [dnt](https://github.com/denoland/dnt) because of the use of top-level-await. But I think that's enough minor thing (dnt users can still target ESM, also they can provide workaround in some ways)

kt3k (approved):

kt3k:
Closing as it's stale. Also this work now needs to be done in https://github.com/denoland/wasmbuild

PR #2148, opened 2022-04-24
http: serve() should export the interface ServeTlsInit

timonson:
I would like to see `ServeTlsInit` being exported the way `ServeInit` already is.


kt3k (approved):
LGTM as it is a part of already exported `serveTls` API.

Issue #2140, opened 2022-04-22
assertSnapshot crops long texts


oscarotero:
**Describe the bug**

I want to use assertSnapshot to test the pages generated by [Lume](https://github.com/lumeland/lume), but just noticed that values with long content (more than 115 characters) are cropped so I cannot test the whole page content. 

**Steps to Reproduce**

1. Create a file with the contents:
```js
import {assertSnapshot} from "https://deno.land/std@0.136.0/testing/snapshot.ts";

Deno.test("Testing a page", async (t) => {
  await assertSnapshot(t, {
    title: "Testing a page",
    content: `
    <h1>Testing a page</h1>
    <p>This is a test</p>
    <ul>
      <li>1</li>
      <li>2</li>
      <li>3</li>
      <li>4</li>
    </ul>
    `
  })
});
```

2. Run the test with `deno test --allow-read --allow-write test.ts -- --update`
3. Open the file `__snapshots__/test.ts.snap` and you can see the `content` key with the value cropped.

```js
export const snapshot = {};

snapshot[`Testing a page 1`] = `
{
  content: "\\n    <h1>Testing a page</h1>\\n    <p>This is a test</p>\\n    <ul>\\n      <li>1</li>\\n      <li>2</li>\\n  ...",
  title: "Testing a page",
}
`;
```

**Expected behavior**

The `content` key should contain the entire original string

**Environment**

- OS: MacOS 12
- deno version: 1.21.0
- std version: 0.136.0

oscarotero:
Note that if the content passed to the snapshot is a string, it will be saved entirely. So a workaround for this is using `JSON.stringify` to pass the content:

```ts
import {assertSnapshot} from "https://deno.land/std@0.136.0/testing/snapshot.ts";

Deno.test("Testing a page", async (t) => {
  await assertSnapshot(t, JSON.stringify({
    title: "Testing a page",
    content: `
    <h1>Testing a page</h1>
    <p>This is a test</p>
    <ul>
      <li>1</li>
      <li>2</li>
      <li>3</li>
      <li>4</li>
    </ul>
    `
  }))
});
```

bartlomieju:
CC @bcheidemann

bcheidemann:
@bartlomieju on it 👍

Issue #2135, opened 2022-04-21
Improvements to the "assertSnapshot" API


bartlomieju:
Just noticing this now; sad to see opinionated magic around where snapshots are saved (a `__snapshots__` dir) instead of the `assertSnapshot` function just accepting 2 arguments; the actual value being asserted, and a file URL / string path to the snapshot file. That way just by reading argument intellisense it's predictable (and customisable) where the snapshots end up, and the snapshot can be held in the correct type of file for the data (`.json`, `.html`, `.css`, etc.) for syntax highlighting and such when directly viewing the snapshot file in an editor.

The simpler design also means a simpler implementation to write and maintain.

Here is an example of the (IMO) ideal `assertSnapshot` API in action on npm:

- https://www.npmjs.com/package/snapshot-assertion
- https://github.com/jaydenseric/snapshot-assertion
- https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L40-L77

Note the optional third argument `assertion` that accepts an assertion function for comparing the actual text with the snapshot file text, in case you want to use your own custom one for a special kind of diff or something:

https://github.com/jaydenseric/snapshot-assertion/blob/5b7f7425a231081e8f97b58765be6a789cf72696/assertSnapshot.mjs#L16-L19

_Originally posted by @jaydenseric in https://github.com/denoland/deno_std/issues/2039#issuecomment-1104689896_

bcheidemann:
As suggested by @kt3k in [this comment](https://github.com/denoland/deno_std/pull/2039#issuecomment-1104699940) we could overload the `assertSnapshot` function.

When I was working on the implementation, this is something I considered because we could have avoided #2128 but I didn't want to over complicate/rush the implementation for first pass because we wanted to release this feature with Deno v1.21.

Anyway, this would be my preferred API though there are of course other options:

```ts
type SnapshotMode = 'update' | 'assert''

type SnapshotOptions<T> = {
  mode: SnapshotMode;
  snapshotDir?: string;
  serializer?: (actual: T) => string;
  // --snip--
  msg?: string;
}

async function assertSnapshot<T>(
  t: Deno.TestContext,
  actual: T,
  options: SnapshotOptions<T>,
): Promise<void>;
async function assertSnapshot(
  context: Deno.TestContext,
  actual: unknown,
  msg?: string,
): Promise<void>;
async function assertSnapshot<T>(
[...]

PR #2150, opened 2022-04-24
test(testing/snapshot): Fix unexpected truncation of long strings in snapshots

bcheidemann:
Adds a regression test for #2140 

This depends on [this PR in denoland/deno](https://github.com/denoland/deno/pull/14384) being merged.

### DIFF
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -4,7 +4,6 @@ import { ensureFile, ensureFileSync } from "../fs/mod.ts";
 import { bold, green, red } from "../fmt/colors.ts";
 import { AssertionError, equal } from "./asserts.ts";
 import { buildMessage, diff, diffstr } from "./_diff.ts";
-import { format } from "./_format.ts";
 
 const CAN_NOT_DISPLAY = "[Cannot display]";
 const SNAPSHOT_DIR = "__snapshots__";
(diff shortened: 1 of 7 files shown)


kt3k (approved):
LGTM, thanks!

bcheidemann:
Now that the `denoland/deno` PR is merged, how can I re-run GH actions? I think it should be passing now as the changes should be in canary.

kt3k:
Canary builds are failing right now in deno main repository (by the reason not related to the fix at all). Please just leave this as is. The maintainers take care of the rest.

testing/snapshot.ts
```diff
@@ -20,6 +19,19 @@ type AssertSnapshotContext = {
 
 let _assertSnapshotContext: AssertSnapshotContext;
 
+export function serialize(actual: unknown): string;
+export function serialize<T>(actual: T): string;
+export function serialize(actual: unknown): string {
```
bartlomieju, line 24: Why are these exported?

bcheidemann: Given the proposed API changes in #2135 I thought it would be nice to export the default `serialize` function so that people can wrap it when providing custom serializers to `assertSnapshot`. However, this has nothing to do with this PR so I am happy to remove the export keyword for now.

bartlomieju: That'd be great, thank you

bcheidemann: Done 👍

Issue #2142, opened 2022-04-23
assertSnapshot handles multiline strings incorrectly


bcheidemann:
**Describe the bug**

When `assertSnapshot` is called with a multiline string, the newlines are replaced with the characters `\\n` making it hard to read multiline snapshots.

**Steps to Reproduce**

1. Create a test file

```ts
// test.ts
Deno.test("Snapshot Test - Multi-Line String", async (t) => {
  await assertSnapshot(t, `
Hello World 1!
Hello World 2!
Hello World 3!`);
});
```

2. Run `deno test test.ts -- -u`
3. See resulting snapshot file

```js
// __snapshots__/test.ts.snap
export const snapshot = {};

snapshot[`Snapshot Test - Multi-Line String 1`] = `"\\nHello World 1!\\nHello World 2!\\nHello World 3!"`;
```

4. Observer the issue

**Expected behavior**

The snapshot file should be formatted as follows:

```js
// __snapshots__/test.ts.snap
export const snapshot = {};

exports[`Snapshot Test - Multi-Line String 1`] = `
"
Hello World 1!
Hello World 2!
Hello World 3!
"
`;
```

**Environment**

- OS: [e.g. Ubuntu 20.04, MacOS 11]
- deno version: 1.21
- std version: 0.136.0

**Workaround**

The current workaround is to write your test file like this:

```ts
// test.ts
Deno.test("Snapshot Test - Multi-Line String", async (t) => {
  await assertSnapshot(t, `
Hello World 1!
Hello World 2!
Hello World 3!`.spilt("\n"));
});
```

Which will result in more readable output:

```js
// __snapshots__/test.ts.snap
export const snapshot = {};

snapshot[`Snapshot Test - Multi-Line String 1`] = `
[
  "",
  "Hello World 1!",
  "Hello World 2!",
  "Hello World 3!",
]
`;
```

oscarotero:
I have the same problem. I was expecting multiline strings were stored with template literals ([like Jest](https://github.com/marketgoo/Ola/blob/development/src/Button/__snapshots__/test.js.snap)), so it's more easy to detect changes and see git diff.

I've also detected differences between Windows and Unix/Macos due endline (`\n` vs `\r\n`) but not sure if this is a bug or is the desired behavior.

bcheidemann:
[...]

PR #2166, opened 2022-04-28
fix(testing/snapshot): handle multiline strings

bcheidemann:
Fix for #2142

### DIFF
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -29,7 +29,7 @@ function serialize(actual: unknown): string {
     compact: false,
     iterableLimit: Infinity,
     strAbbreviateSize: Infinity,
-  });
+  }).replace(/\\n/g, "\n");
 }
 
 /**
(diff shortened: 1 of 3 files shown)


testing/snapshot.ts
```diff
@@ -29,7 +29,7 @@ function serialize(actual: unknown): string {
     compact: false,
     iterableLimit: Infinity,
     strAbbreviateSize: Infinity,
-  });
+  }).replace(/\\n/g, "\n");
```
kt3k, line 32: With this change, the expression `"..."` isn't canonical string literal anymore if it contains `\n`.

I think it's ok and works in most cases.

But does anyone have any concern about this transformation?

bcheidemann: Yeah this is also how it is in Jest but that doesn't necessarily mean it's how it should be done here. The alternative I considered was:

```ts
await assertSnapshot("Hello\nWorld!");

export[`snapshot`] = `
\`Hello
World!\`
`;
```

But I felt that this has two downsides:

1. The snapshot is less readable due to the presence of escape characters
2. It would require significant changes to `Deno.inspect` and I was concerned about cluttering the API with options that have no use case (at least not that I can think of!) outside of snapshot testing

bcheidemann: > With this change, the expression `"..."` isn't canonical string literal anymore if it contains `\n`.
> 
> I think it's ok and works in most cases.
> 
> But does anyone have any concern about this transformation?

@kt3k Are we still waiting for feedback on this?

kt3k: Let's land and try to see the community feedback. Thanks for your contribution!

kt3k (approved):
LGTM

Issue #882, opened 2021-04-26
std/encoding/csv.ts - `stringify` could use improvement and better documentation


dsherret:
`stringify` in std/encoding/csv.ts takes a lot of reading to figure it out. After looking over the documentation and source code, I've discovered that I need to provide objects then specify what each column is along with how each column should access a property on the objects.

Some suggestions:

1. Unfortunately the property access is not type safe and so it's possible to change a property name and break this code without being notified. The API could be improved if it were type safe between the property names and the objects being provided as this would prevent bugs.
2. It would be nice if there were something analogous to `JSON.stringify(data)` here, where I can just have `data` in memory being `string[][]` and then do `stringify(data)` and get the result. I don't want to bother having to specify column names and property names on an object.
3. It might be easier to understand if there were some examples, but I think examples wouldn't be required if it wasn't so complex.

By the way, a simple API I've found for csv parsing and stringifying is https://www.npmjs.com/package/csv-string -- I almost feel like the csv parsing in std should be really simple then people can map the data in and out to objects if they feel like. There's a lot going on right now in there.

luk3skyw4lker:
I'm taking this issue! I'll take a look and see what can I do.

luk3skyw4lker:
About number 2, I think that the best solution would be to move the columns argument into the options object, this way the user could specify which columns he wants in the output and if no option is provided, then all columns will de present into the output.

luk3skyw4lker:
About 1, would be the case to think about adding a generic parameter to stringify to specify the type of the input object or am I wrong? I didn't fully understand the type safety that this funciton would have, can you clarify it for me?

luk3skyw4lker:
And about 3, I think that the examples would do a good job, but I'll take a look at the code you provided as a simple solution and see if it fits into the case. I gave a quick look and at least for me the deno implementation looks simpler, but would benefit from a better documentation.

luk3skyw4lker:
I took a deeper look at the csv-string implementation and it really looks simpler. I still stand with the idea of passing columns through the options object though which it would add some complexity to deno's implementation. What do you guys think? @bartlomieju  @crowlKats @dsherret

ralyodio:
Columns in json objects don’t guarantee order.  I would assume an array of arrays is the only way or possible a “Map” that does retain order in keys.

luk3skyw4lker:
[...]

PR #2168, opened 2022-04-29
refactor(encoding): making column argument optional

luk3skyw4lker:
Insert columns argument inside options object to make it optional in case stringify is called.

I'm open to implementation ideas and anything else. This PR addresses #882


luk3skyw4lker:
I could also work at the type safety of the property accessors, but I would need some help. I'm going to send this PR in the deno discord server.

kt3k:
If I run the code like the below, it prints only empty lines
```ts
import { stringify } from "./encoding/csv.ts";
const str = await stringify([[1, 2, 3], [4, 5, 6]]);
console.log(str);
```
This default behavior doesn't seem useful to me.

kt3k:
This is going to be a breaking change, but how much is it difficult to support old behavior as well by function overloading?

If we can overload the function like:

```ts
export async function stringify(
  data: DataItem[],
  options: StringifyOptions = {},
): Promise<string>;
export async function stringify(
  data: DataItem[],
  columns: Column[],
  options: StringifyOptions = {},
): Promise<string>;
// ... impl here ...
```

and keep the old behavior, then it'd be much easier to land this PR

luk3skyw4lker:
The backwards compatibility could be done, I'll work on that. About the default behavior, I'll try to make something more useful too.

Thanks for the review!

kt3k:
> The backwards compatibility could be done, I'll work on that. About the default behavior, I'll try to make something more useful too.

Great! Thanks!

luk3skyw4lker:
@kt3k I took a look at TypeScript's documentation and it seems that the function overloading cannot be done. Typescript doesn't support function overloading with different number of parameters.

luk3skyw4lker:
But I'm still working on some useful behavior to make things right.

luk3skyw4lker:
@kt3k also I think that this default behavior actually makes sense, what could be done is if the user sends a two dimensions array but no accessor at the column option we could throw an error saying that no property acessor was provided in the options column.

If I try to make another default behavior out of this, it could lead to a major confusion with the data, I think that this is one of the problems that could be solved with improving the docs of the function. What do you think?

kt3k:
@luk3skyw4lker 
> I took a look at TypeScript's documentation and it seems that the function overloading cannot be done. Typescript doesn't support function overloading with different number of parameters.

Overloading with different number of parameters should be possible. We already have such APIs. See `assertThrows` in `testing` module for example
https://github.com/denoland/deno_std/blob/564d2ed8abe47f97e9198aebefe20b4fea6c3c47/testing/asserts.ts#L599-L619

> If I try to make another default behavior out of this, it could lead to a major confusion with the data, I think that this is one of the problems that could be solved with improving the docs of the function. What do you think?

Doc improvement instead of code change is also very welcome. Maybe let's follow that path in this case?

luk3skyw4lker:
@kt3k Right, maybe I got something wrong, I thought I saw something like this at Deno too, even got confused when checked the docs and they said it wasn't supported.

About the docs, I think I may have some idea to make it work without the property accessors anyway, but it would only work with arrays, which makes more sense to me what do you think? The docs improvement would also be welcoming, the original issue mentions that too.

luk3skyw4lker:
@kt3k I just did some refactoring and I think that the default behavior for arrays is more useful now, if you test it out with the same case that you pointed out earlier in the PR, the output should be:

1,2,3
4,5,6

luk3skyw4lker:
I'll redo this PR next weekend if it still makes sense (moving my changes to csv.ts)

@kt3k @bartlomieju @crowlKats please let me know if this changes would still be useful

timreichen:
Why not just pack `columns` into `options` instead of overloading?

luk3skyw4lker:
> I would be in favour of packing `columns` into `options` instead of overloading. std is not stable and it is versioned. Breaking changes are possible.

@timreichen That was the first implementation, but I don't think that overloading is much of a problem, it's already done btw. But if it seems like something that wouldn't be useful we can remove it.

timreichen:
@luk3skyw4lker Then I would suggest to implement it but to add a deprecation tag  to the old implementation overload, something along the lines of 
```ts
/**
  * @deprecated this overload will be removed soon. Use `stringify(data, { ...options, columns })` instead.
  */
```
 I think the `columns` as a property in `options` is cleaner and both overloads provide the same data.

luk3skyw4lker:
@timreichen Alright! I'll update this PR next weekend with the suggestions, thanks for the tips!

luk3skyw4lker:
@timreichen @kt3k I have just noticed that for this to land, it would be a breaking change anyway, because for it to be available even with the overloading, I have to change the order of the function parameters, so if it`s gonna be a breaking change anyway, I think that we don't need the overloading. What do you guys think?

kt3k:
@luk3skyw4lker 
> it would be a breaking change anyway, because for it to be available even with the overloading, I have to change the order of the function parameters, so if it`s gonna be a breaking change anyway, I think that we don't need the overloading.

Ok. Let's try without overloading

luk3skyw4lker:
@kt3k @timreichen could you guys take a look? I'm open to any suggestions on how to make it better, thanks for the patience, previous suggestions and reviews!

encoding/csv.ts
```diff
@@ -145,18 +159,18 @@ async function getValuesFromItem(
...
  */
 export async function stringify(
   data: DataItem[],
-  columns: Column[],
   options: StringifyOptions = {},
 ): Promise<string> {
+  const columns = options.columns || [];
   const { headers, separator: sep } = {
     headers: true,
     separator: ",",
```
timreichen, line 176:
```suggestion
  { columns = [], headers = true, separator: sep = "," }: StringifyOptions = {},
): Promise<string> {
```

luk3skyw4lker:
@timreichen alterations done!

timreichen:
LGTM!

kt3k (approved):
LGTM

kt3k:
We also introduced a few breaking changes to `encoding/csv` recently, but we didn't receive much response from the community.
https://github.com/denoland/deno_std/pull/2536
https://github.com/denoland/deno_std/pull/2491

I think it's still relatively safe to make breaking changes to `encoding/csv`.

Issue #2091, opened 2022-04-07
`[child_process spawn] 'spawn' event` test uses "echo" which is not available on windows


dsherret:
This test uses "echo", but that's not available on Windows except when executing through `cmd`:

```
{
  "0": NotFound: program not found
    at Object.opSync (deno:core/01_core.js:177:12)
    at opRun (deno:runtime/js/40_process.js:29:17)
    at Object.run (deno:runtime/js/40_process.js:116:17)
    at new ChildProcess (file:///V:/deno_std/node/internal/child_process.ts:137:28)
    at spawn (file:///V:/deno_std/node/child_process.ts:122:10)
    at file:///V:/deno_std/node/child_process_test.ts:176:22
    at testStepSanitizer (deno:runtime/js/40_testing.js:444:13)
    at asyncOpSanitizer (deno:runtime/js/40_testing.js:145:15)
    at resourceSanitizer (deno:runtime/js/40_testing.js:370:13)
    at Object.exitSanitizer [as fn] (deno:runtime/js/40_testing.js:427:15)
}
```

luk3skyw4lker:
Looking at it right now.

luk3skyw4lker:
Another tests on the same file also use echo, would it be the case to refactor them all or just this one?

dsherret:
Thanks for looking into it. I think only that test. The other one is ignored on Windows and the other one uses a shell.

luk3skyw4lker:
You're welcome, I'll try to open a PR for solving this still today. I'm looking out for probable replacements for echo if it's a windows environment.

luk3skyw4lker:
One last thing, in which case that error is thrown? Because as far as I know, echo is really a command in the windows CMD, but you made it look like you weren't in a cmd environment when describing the issue, could you clarify it for me?

iuioiua:
@kt3k, I believe this can be closed as #2353 has implemented this fix.

kt3k:
@iuioiua Thanks!

PR #2169, opened 2022-04-29
fix(node): child_process spawn breaking

luk3skyw4lker:
spawn test using echo on Windows when it's not available #2091


kt3k:
Closing in favor of #2353.

Sorry for the delay in review, but please try minimizing the diff next time. Irrelevant diffs slows down and distracts the reviews

PR #2173, opened 2022-05-01
:recycle: Move from testing/bench to Deno.bench

UltiRequiem:
Because testing/bench is now deprecated

https://github.com/denoland/deno_std/tree/main/testing#benching
> These APIs are deprecated. Use Deno.bench() instead.

Docs:

https://deno.land/manual/tools/benchmarker#benchmarking-tool
https://doc.deno.land/deno/unstable/~/Deno.bench

I didn't find the `runs` property equivalent, maybe I could just iterate, but doesn't seem idiomatic.


kt3k (approved):
LGTM, nice work!

kt3k:
Side note: The output looks much nicer in the new UI

Before
スクリーンショット 2022-05-03 22 05 22

After
スクリーンショット 2022-05-03 22 04 27

PR #2178, opened 2022-05-02
test(http): add catch handler to server promise

bartlomieju:
Ref https://github.com/denoland/deno/issues/14449#issuecomment-1114322139


kt3k (approved):
LGTM

PR #2179, opened 2022-05-02
feat(testing/snapshot): Add "assertSnapshot" options overload

### LINKED ISSUE #2128: test(snapshot): do not commit wrong snapshot
ref #2039

Currently we intentionally commit some wrong snapshots in `testing/__snapshot__` to test `assertSnapshot`, and that means we can't use `--update` option for updating that snapshot. We need to somehow fix this to avoid the maintenance difficulties

### LINKED ISSUE #2129: test(snapshot): use `Deno.makeTempDir` instead of custom tmp directory
ref #2039

and also let's remove tmp directory at assertion failures.

### LINKED ISSUE #2135: Improvements to the "assertSnapshot" API

bcheidemann:
Closes #2135 
Closes #2128 
Closes #2129

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         shell: bash
         # excludes tests, testdata, and generated sources from coverage report
         run: |
-         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules" > cov.lcov
+         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules|\\.snap" > cov.lcov
 
       - name: Upload coverage
         uses: codecov/codecov-action@v1
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
         shell: bash
         # excludes tests, testdata, and generated sources from coverage report
         run: |
-         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules|\\.snap" > cov.lcov
+         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules" > cov.lcov
 
       - name: Upload coverage
         uses: codecov/codecov-action@v1
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -22,7 +22,7 @@ export type SnapshotOptions<T = unknown> = {
   /**
    * Snapshot output directory. Snapshot files will be written to this directory.
    * This can be relative to the test directory or an absolute path.
-   * 
+   *
    * If both `dir` and `path` are specified, the `dir` option will be ignored and
    * the `path` option will be handled as normal.
    */
@@ -46,7 +46,7 @@ export type SnapshotOptions<T = unknown> = {
   /**
    * Snapshot output path. The shapshot will be written to this file. This can be
    * a path relative to the test directory or an absolute path.
-   * 
+   *
    * If both `dir` and `path` are specified, the `dir` option will be ignored and
    * the `path` option will be handled as normal.
    */
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -327,15 +327,10 @@ export async function assertSnapshot<T>(
   actual: T,
   options: SnapshotOptions<T>,
 ): Promise<void>;
-export async function assertSnapshot(
-  context: Deno.TestContext,
-  actual: unknown,
-  msgOrOpts?: string,
-): Promise<void>;
 export async function assertSnapshot<T>(
   context: Deno.TestContext,
   actual: T,
-  msgOrOpts?: string,
+  message?: string,
 ): Promise<void>;
 export async function assertSnapshot(
   context: Deno.TestContext,
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -117,7 +117,7 @@ class AssertSnapshotContext {
   ): AssertSnapshotContext {
     let path: string;
     if (options.path) {
-      path = options.path;
+      path = resolve(options.path);
     } else if (options.dir) {
       const testFilePath = fromFileUrl(testContext.origin);
       const testFileName = basename(testFilePath);
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -54,9 +54,9 @@ export type SnapshotOptions<T = unknown> = {
 /**
  * Default serializer for `assertSnapshot`.
  */
-function serialize(actual: unknown): string;
-function serialize<T>(actual: T): string;
-function serialize(actual: unknown): string {
+export function serialize(actual: unknown): string;
+export function serialize<T>(actual: T): string;
+export function serialize(actual: unknown): string {
   return Deno.inspect(actual, {
     depth: Infinity,
     sorted: true,
@@ -325,7 +325,7 @@ export async function assertSnapshot(
     context,
     options,
   );
-  const testName = getTestName(context);
+  const testName = getTestName(context, options);
   const count = assertSnapshotContext.getCount(testName);
   const name = `${testName} ${count}`;
   const snapshot = await assertSnapshotContext.getSnapshot(
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -94,9 +94,10 @@ function getMode(options: SnapshotOptions) {
   } else if (_mode) {
     return _mode;
   } else {
-    return Deno.args.some((arg) => arg === "--update" || arg === "-u")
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -38,7 +38,7 @@ jobs:
         shell: bash
         # excludes tests, testdata, and generated sources from coverage report
         run: |
-         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules" > cov.lcov
+         deno coverage ./cov/ --lcov --exclude="test\\.(ts|js)|wasm\\.js|testdata|node/_tools|node/_module/cjs|node_modules|\\.snap" > cov.lcov
```
bartlomieju, line 41: While this change makes sense, I think we should also tackle https://github.com/denoland/deno/issues/14408 so it's not accounted for by default

testing/snapshot.ts
```diff
@@ -1,27 +1,66 @@
...
+   * cause the diff not to be logged.
+   */
+  msg?: string;
+  /**
+   * Name of the snapshot to use in the snapshot file.
+   */
+  name?: string;
+  /**
+   * Snapshot output path. The shapshot will be written to this file. This can be
+   * a path relative to the test directory or an absolute path.
```
bartlomieju, line 45: Please add a comment that this option will override `dir` if the path is absolute

testing/snapshot.ts
```diff
@@ -33,51 +72,240 @@ function serialize(actual: unknown): string {
...
-  function escapeStringForJs(str: string) {
-    return str
-      .replace(/\\/g, "\\\\")
-      .replace(/`/g, "\\`")
-      .replace(/\$/g, "\\$");
+function getMode(options: SnapshotOptions) {
+  if (options.mode) {
+    return options.mode;
+  } else if (_mode) {
+    return _mode;
```
bartlomieju, line 96: Can you add a test that will exercise this line?

testing/snapshot.ts
```diff
@@ -33,51 +72,240 @@ function serialize(actual: unknown): string {
...
+  private getCurrentSnapshotsInitialized() {
+    assert(
+      this.currentSnapshots,
+      "Snapshot was not initialized. This is a bug in `assertSnapshot`.",
+    );
+    return this.currentSnapshots;
+  }
+
+  /**
+   * Write updates to the snapshot file and logs statistics.
```
bartlomieju, line 170:
```suggestion
   * Write updates to the snapshot file and log statistics.
```

.github/workflows/ci.yml
bcheidemann: Yeah good call 😊 I will implement a fix for https://github.com/denoland/deno/issues/14408 and then remove this.

testing/snapshot.ts
bcheidemann: Done 👍

testing/snapshot_test.ts
```diff
@@ -214,8 +372,11 @@ Deno.test("Snapshot Test - Update", async (t) => {
     });
   `);
 
-  await assertSnapshot(t, formatOutput(result3.output));
-  assert(!formatError(result3.error), "unexpected output to stderr");
+  assertNoError(result3.error);
+  await assertSnapshot(t, formatTestOutput(result3.output));
+
+  // Tidyup
+  await Deno.remove(tempDir, { recursive: true });
```
bartlomieju, line 379: I believe all of these calls won't run if the test case throws

bcheidemann: Yes that's correct. Any remaining directories should be cleaned up by the OS eventually but I will implement a solution to ensure that we tidy up after ourselves regardless of tests passing or failing 😊

testing/snapshot.ts
```diff
@@ -33,51 +78,240 @@ function serialize(actual: unknown): string {
...
+   * Returns an instance of `AssertSnapshotContext`. This will be retrieved from
+   * a cache if an instance was already created for a given snapshot file path.
+   */
+  static fromOptions(
+    testContext: Deno.TestContext,
+    options: SnapshotOptions,
+  ): AssertSnapshotContext {
+    let path: string;
+    if (options.path) {
+      path = resolve(options.path);
```
kt3k, line 130: `path` and `dir` are resolved relative to current directory (which is the directory the user invoked `deno test` command) here. I think another option might be to resolve it relative to the test file's directory (`parse(testFilePath).dir`).

I think the latter might be the better choice because the user probably want to place snapshot near the test file. What do you think?

cc @bartlomieju Do you have any preference on this?

bartlomieju: I agree @kt3k, the latter seems like a better option

bcheidemann: Whoops :P fixed this

.github/workflows/ci.yml
bartlomieju: This should now be addressed by https://github.com/denoland/deno/commit/253fbf9d2a15018cbba9c9ea2829bd70e9723362

testing/snapshot_test.ts
bcheidemann: Done 👍

testing/snapshot.ts
bcheidemann: This wasn't covered because I made a mistake in the implementation. This is fixed now and this line is covered without the need for a specific test.

.github/workflows/ci.yml
bcheidemann: Thanks! :) I have reverted this change

testing/snapshot_test.ts
```diff
@@ -23,6 +23,21 @@ function formatTestError(string: string) {
   return stripColor(string).replace(/^Check file:\/\/(.+)\n/g, "");
 }
 
+function testFnWithTempDir(fn: (t: Deno.TestContext, tempDir: string) => Promise<void>) {
+  return async (t: Deno.TestContext) => {
+    const tempDir = await Deno.makeTempDir();
+    console.log(tempDir);
```
bartlomieju: Remove

testing/snapshot_test.ts
```diff
@@ -23,6 +23,21 @@ function formatTestError(string: string) {
...
+    try {
+      await fn(t, tempDir);
+      await Deno.remove(tempDir, { recursive: true });
+    }
+    catch (err) {
+      await Deno.remove(tempDir, { recursive: true });
+      throw err;
+    }
+  }
```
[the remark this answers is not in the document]
kt3k: I think we can even omit `catch (err) { throw err }` part

```ts
try {
  await fn(t, tempDir);
} finally {
  await Deno.remove(tempDir, { recursive: true });
}
```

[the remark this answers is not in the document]
bcheidemann: Done 👍

testing/snapshot_test.ts
bcheidemann: Removed this

bcheidemann:
@bartlomieju @kt3k 

Any idea what's causing this error?

image

I haven't touched this file and `deno fmt --check` passes locally 🤔

testing/snapshot_test.ts
```diff
@@ -96,127 +127,331 @@ Deno.test("Snapshot Test - Multi-Line Strings", async (t) => {
...
+      await assertSnapshot(t, VALUE, {
+        dir: join(Deno.cwd(), "testing/__snapshots__/options_tests/"),
+      });
+    });
+  });
+
+  await t.step("path", async (t) => {
+    await t.step("relative", async (t) => {
+      await assertSnapshot(t, VALUE, {
+        path: "__snapshots__/options_tests/custom_path.snap",
```
kt3k, line 194: 👍

kt3k:
@bcheidemann `deno fmt` output has been changed a little bit in denoland/deno#14446 (The CI runs with deno canary, so it's affected by it)

kt3k (approved):
LGTM, nice work!

bartlomieju (approved):
LGTM too, thank you @bcheidemann!

PR #2180, opened 2022-05-02
feat(flags): infer argument types, names and defaults

### LINKED ISSUE #1898: feature(flags): Add generic type to `parse()`

The `parse()` function return `Args` type. This interface's keys types except `-` are `any` type. So, I suggent using generic type. Like this:

```typescript
export function parse<T extends Args>(
  args: string[],
  {
    "--": doubleDash = false,
    alias = {},
    boolean = false,
    default: defaults = {},
    stopEarly = false,
    string = [],
    unknown = (i: string): unknown => i,
  }: ArgParsingOptions = {},
): T
```

c4spar:
close #1898


flags/test.ts
```diff
@@ -1,6 +1,10 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 import { assertEquals } from "../testing/asserts.ts";
 import { parse } from "./mod.ts";
+import {
+  assert,
+  IsExact,
+} from "https://deno.land/x/conditional_type_checks@1.0.5/mod.ts";
```
kt3k, line 7: We prefer to avoid depending on `deno.land/x/` from `std`. Can you copy the necessary part of the module in std as test utility?

flags/test.ts
```diff
@@ -736,3 +740,382 @@ Deno.test("valueFollowingDoubleHyphenIsNotUnknown", function (): void {
 Deno.test("whitespaceShouldBeWhitespace", function (): void {
   assertEquals(parse(["-x", "\t"]).x, "\t");
 });
+
+/** ---------------------- TYPE TESTS ---------------------- */
+
+Deno.test("typesOfAllBoolean", function (): void {
```
kt3k, line 746: Nice test cases! These illustrate what's done in this change very well

kt3k:
@c4spar Can you fix the type error in CI?

```
error: TS2322 [ERROR]: Type 'unknown' is not assignable to type 'boolean | undefined'.
      showDirListing: serverArgs["dir-listing"],
      ~~~~~~~~~~~~~~
    at file:///Users/runner/work/deno_std/deno_std/http/file_server.ts:810:7
```

flags/test.ts
c4spar: Yep i know, it was just temporarly 😉 will do 👍

c4spar:
Oops, I assumed dashed parameters are camelcased 😅 is there a reason why we don't do that?

kt3k (dismissed):
Thanks for updating!

LGTM!

flags/mod.ts
```diff
@@ -3,17 +3,96 @@
...
+type BooleanType = boolean | string | undefined;
+type StringType = string | undefined;
+type ArgType = BooleanType | StringType;
+
+type Values<
+  B extends BooleanType,
+  S extends StringType,
+  D extends Record<string, unknown> | undefined,
+> = undefined extends ((false extends B ? undefined : B) & S) ? // deno-lint-ignore no-explicit-any
+Record<string, any>
```
c4spar, line 24: @kt3k Currently I use `Record<string, any>` as the default type for the arguments when no types are defined in the options object. Should we change this to `Record<string, unknown>` or do we watn to keep `Record<string, any>`?

c4spar: yes, would be `string | number | boolean | undefined`.
But I also think that `any` should be sufficient in this case. If someone wants it type-safe, then he would specify option types anyway.

bartlomieju: I agree, `any` seems like a safe bet here

flags/mod.ts
```diff
@@ -3,17 +3,117 @@
...
+    [K in Name]?: MapTypes<Rest, V>;
+  }
+  : T extends string ? Partial<Record<T, V>>
+  : Record<never, never>;
+
+type MapDefaults<T extends ArgType> = T extends string
+  ? Partial<Record<T, unknown>>
+  : Record<string, unknown>;
+
+type DedotRecord<T> = T extends Record<string, unknown> ? UnionToIntersection<
```
bartlomieju, line 77: Lots of magic here :D

bartlomieju (commented):
@c4spar this looks mostly good to me, could you please update the `README` file for `flags` module to describe all of the type checking improvements you did here. Basically we need to educate users that they can make their parsing type safe.

c4spar:
It looks like the tests are currently failing because of a bug with dotted options. And I also discovered a view more bugs/issues while testing.

- Throws an error: Cannot create property 'bar' on boolean 'true'.
  ```ts
  parse(["--foo"], {
    default: {
      "foo.bar.baz": 1,
    },
  });
  ```

- Throws an error: Cannot use 'in' operator to search for 'baz' in true.
  ```ts
  parse(["--foo", "--foo.bar"]);
  ```

- Default value for dotted boolean arg is ignored. works with string arg.
  ```ts
  args = parse([], {
    boolean: ["foo.bar.baz"],
    default: {
      "foo": 1,
    },
  });
  console.log(args);
  ```

- `--foo` is turned into an array with the default value for `foo.bar.baz`.
  ```ts
  const args = parse(["--foo"], {
    boolean: ["foo.bar.baz"],
    default: {
      "foo.bar.baz": 1,
    },
  });
  console.log(args); // { _: [], foo: [ { bar: { baz: 1 } }, true ] }
  ```
  Not sure if the last one is a bug or working as intended but no idea how i can add proper typings for this. Because boolean options are always defined by default so the type for the example would be `{ foo: { bar: { baz: boolean | number } } }` but it returns `{ foo: [ { bar: { baz: 1 } }, true ] }` if i call it with `--foo` 🤔.

- And another issue I discovered is that all values can be also an array if they are used multiple times. So `--foo=1 --foo=2` would return `{ foo: [1, 2] }`. But that's not really nice because then I always have to check if the value is an array or not 🤔.

c4spar:
hm.., and i forgot negatable options. all options can be also of type `false` if they are prefixed with `--no-`.

kt3k:
> And another issue I discovered is that all values can be also an array if they are used multiple times. So --foo=1 --foo=2 would return { foo: [1, 2] }. But that's not really nice because then I always have to check if the value is an array or not 🤔.

I agree it's not nice. I guess this feature isn't used that much. I'd suggest we just should remove it. cc @bartlomieju What do you think?

> and i forgot negatable options. all options can be also of type false if they are prefixed with --no-.

hmm now `false` appearing everywhere even for string-typed options, and that doesn't look nice/useful.. I feel current `--no-` handling for non-boolean options is a bug, and we can ignore `--no-` in typing. What do you think?

kt3k:
I'm also skeptical about the usefulness and user awareness of dotted options. How about just handling everything in flat namespace?

cc @bartlomieju What do you think about dotted options?

bartlomieju:
> > And another issue I discovered is that all values can be also an array if they are used multiple times. So --foo=1 --foo=2 would return { foo: [1, 2] }. But that's not really nice because then I always have to check if the value is an array or not 🤔.
> 
> I agree it's not nice. I guess this feature isn't used that much. I'd suggest we just should remove it. cc @bartlomieju What do you think?
> 

Seems ok, but we should check if there are any usages in the wild before removing this feature. https://cs.github.com/?scopeName=All+repos&scope=&q=flags%2Fmod.ts should help with it

> I'm also skeptical about the usefulness and user awareness of dotted options. How about just handling everything in flat namespace?

> cc @bartlomieju What do you think about dotted options?

Agree, but again we should first check for potential usages.

c4spar:
I agree, the current implementation is not realy useable. It makes not a big difference to just use `unknown` everywhere because in most cases i still have to check the type now.

But if we want to remove support for collectable, negatable and dotted options, we should also keep in mind that the flags module is also used by some other cli libraries, so we would basically remove the feature also from there.

> hmm now false appearing everywhere even for string-typed options, and that doesn't look nice/useful.. I feel current --no- handling for non-boolean options is a bug, and we can ignore --no- in typing. What do you think?

I think `--no-` would be still usefull to disable default values from string args. But we could require to define negatable options manually, than it does not apply to all options automatically like this:

```ts
parse(Deno.args, {
  string: ["foo", "no-foo", "bar"],
  default: {
    foo: "bar"
  }
});

type Args = {
  foo: string | false;
  bar?: string;
};
```

```console
$ deno run example.ts
{ foo: "bar" }
$ deno run example.ts --no-foo
{ foo: false }
```

And if we remove support for collectable options or make it configurable as well, than i think there should be also no issue with dotted options 🤔

kt3k:
> But if we want to remove support for collectable, negatable and dotted options, we should also keep in mind that the flags module is also used by some other cli libraries, so we would basically remove the feature also from there.

Right, but do you know any library which depends on `std/flags`?

> I think --no- would be still usefull to disable default values from string args. But we could require to define negatable options manually, than it does not apply to all options automatically like this:

hmm, ok let's keep it for now.

> And if we remove support for collectable options or make it configurable as well, than i think there should be also no issue with dotted options 🤔

Ok. Let's try to remove colletable options and keep dotted options. I'll check the usages of collectable options in code search.

kt3k:
Briefly read through the first page of [this query](https://cs.github.com/?scopeName=All+repos&scope=&q=flags%2Fmod.ts+language%3ATypeScript+-path%3Ahttp%2Ffile_server.ts+-path%3Aflags%2Fmod.ts+-path%3A%E2%80%8Enode%2Finternal_binding%2Fnode_options.ts+-path%3A%E2%80%8Eexamples%2Fflags.ts+-path%3Aexamples%2Fgist.ts). 2 tools out of 15 seem using collectable types. Looks like more than I expected..

The tools not using colletable options
- [tani/vip](https://github.com/tani/vip/blob/19519e66180cd3fd081921b082a08024f8e70ac3/vip.ts) `opt: boolean`
- [codehz/deno-pug](https://github.com/codehz/deno-pug/blob/5393f52e16fe09edf8d9c2fb50d3f9e8da9fcae6/mod.ts) `output: string, release: boolean`
- [UltiRequiem/deno-clear](https://github.com/UltiRequiem/deno-clear/blob/99f9c6fcc3872215da7c29c132b948a9a659224b/cli.ts) `full: boolean, delay: number, help: boolean`
- [tyrelh/battlesnake-typescript-deno](https://github.com/tyrelh/battlesnake-typescript-deno/blob/4bbc3a2e52ea0892d69b1f46e2d4378d3f6bd0b6/app.ts) `port: number`
- [Thejus-Paul/wyvern](https://github.com/Thejus-Paul/wyvern/blob/c1015461a4047d4542049d20f6406c296cce0359/src/server.ts) `port: number`
- [Simply007/kontent-deno-vr](https://github.com/Simply007/kontent-deno-vr/blob/85482d7e0e9184ad888e2647fb2de9cc84ba18a0/alosaur.ts) `port: number`
- [SpenceSellers/theft.js](https://github.com/SpenceSellers/theft.js/blob/7761e4266acc504c1dd6e8d598e01fe6235fc932/theft.ts) `expand: boolean`
- [snendev/simple-deno-react-stack](https://github.com/snendev/simple-deno-react-stack/blob/a4a1820aac46b841958138ce462aa657292929cf/build/build.ts) `minify: boolean, sourcemap: boolean`
- [kbaseIncubator/kbase-mock-services](https://github.com/kbaseIncubator/kbase-mock-services/blob/9327e13c2ed42e4a9a04152a7e49906c49ee91a4/src/index.ts) `data-dir: string, port: number`
- [kbaseIncubator/kbase-mock-services fetchSampleSet.ts](https://github.com/kbaseIncubator/kbase-mock-services/blob/9327e13c2ed42e4a9a04152a7e49906c49ee91a4/src/cli/fetchSampleSet.ts) `ref: string, env: string, token: string, dest: string, timeout: number`
- [nats-io/nats.ws](https://github.com/nats-io/nats.ws/blob/d5b33a54b62e59ac689486af7fab05c9cf58f4fa/bin/clone-nd.ts) `dir: string, lib: string`
- [ije/aleph-canary](https://github.com/ije/aleph-canary/blob/b5c606ed73876bec3e662ef3fe3702f17b942e24/lib/log.ts) `log-level: string`
- [rimbu-org/rimbu](https://github.com/rimbu-org/rimbu/blob/a5fb58579d00aac2af1825c74def67c73f2e8ae6/doc-gen/src/parse-args.ts) `help: boolean, version: boolean, input: string, output: string`


The tools which use collectable options
- [skymethod/denoflare](https://github.com/skymethod/denoflare)
  - in `tail` subcommand `header`, `ip-address`, `method`, etc are collectable options. ([ref](https://github.com/skymethod/denoflare/blob/05ed9f6747c07ce3bbe75430b3340ecf03babac0/cli/cli_tail.ts#L182))
- [retog/scuttlesaurus](https://github.com/retog/scuttlesaurus/blob/52c2383375eab5b7852d1df1dad65d3e0dc41217/scuttlesaurus/main.ts) `logLevel: string, baseDir: string, incoming: string, outgoing: string, web: collectable object!`
  - See also [usage](https://github.com/retog/scuttlesaurus/tree/52c2383375eab5b7852d1df1dad65d3e0dc41217/scuttlesaurus#standalone-usage)

kt3k:
Note: In case of [scuttlesaurus](https://github.com/retog/scuttlesaurus/tree/52c2383375eab5b7852d1df1dad65d3e0dc41217/scuttlesaurus#standalone-usage), they use collectable dotted options. Looks like heavily dependent on both features...

c4spar:
@kt3k

> Right, but do you know any library which depends on std/flags?

[Denomander](https://github.com/siokas/denomander) is using the flags module too. Looks like it supports negateable and collectable options but not dotted options.

> Can you make collectable options configurable? How about enabling type inference only when the collectable options are disabled?

I can take a deeper look at weekend. We could try making it configureable like this:

```ts
parse(Deno.args, {
  string: ["foo"],
  collect/array: ["foo"],
});
```

But maybe we need to add some kind of error handling for duplicated and invalid options.

- What should happen if a non collectable option is used multiple times?
  1. The flags module throws an error for duplicated options.
  2. Add a callback method for duplicated options, similar to the unknown callback option.
  3. The last value will override the previous values.
- What should happen if i define a dotted option `foo.bar.baz` and call it with `--foo`?
  1. The flags module throws an error for invalid options.
  2. Add a callback method for invalid options, similar to the unknown callback option.
  3. Set `foo` to `{}` instead of `true` if its a dotted option and if no default value is set.
- Do we want to make negatable options also configurable? If yes, i think we need to add some error handling for it as well.

kt3k:
@c4spar 
> I can take a deeper look at weekend. We could try making it configureable like this:

Thanks!

> What should happen if a non collectable option is used multiple times?

I prefer the option `i.` the best for this. I'm fine with `iii.` if it's easier to implement. The option `ii.` looks like a strange way of handling the irregular case.

> What should happen if i define a dotted option foo.bar.baz and call it with --foo?

I prefer the option `i.`. The option `iii.` also looks ok to me. The option `ii.` looks strange and complex.

c4spar:
@kt3k i added the typings for the collect option and also for negatable options.
but i just noticed that boolean values are not collectable. i thought they would be collectable too, but it looks like only numeric and string values are currently collectable.
i can fix the typings for boolean args, but maybe we want to make boolean args collectable too? one use case would be `-vvv` (`-v -v -v`) for different verbosity levels.

c4spar:
hmm, aliases are also added to the results object. i tried adding types for it but rn it works only with a const assertion:
```ts
parse(Deno.args, {
  string: ["foo"],
  alias: {
    foo: "f",
  } as const
});
```
Not sure if it makes sense to invest more time on this and make the types more complex or if we should remove the feature. because i'm also not sure if it makes much sense to add all aliases to the results object. i think normally i would expect they are not added. What do you think @kt3k?

kt3k:
I think it's fine without alias support as first pass.

I'll review this soon. Thanks for your work!

c4spar:
the types for the alias option work now. i will open a separate PR for the new collect option.

flags/test.ts
```diff
@@ -736,3 +737,958 @@ Deno.test("valueFollowingDoubleHyphenIsNotUnknown", function (): void {
...
+  assertType<
+    IsExact<
+      typeof argv,
+      & { [x: string]: unknown }
+      & {
+        bar: number | string;
+        foo: TextDecoder | Array<boolean>;
+        dotted: {
+          beep: boolean | Date;
+          boop: RegExp | Array<string>;
```
kt3k, line 1463: This boop seems possible to be `string` when single `--dotted.boop foo` is passed

kt3k: Ah, this might depend on #2247?

c4spar: Yep, it depends on the other PR. Collectable options will then always be an array.

c4spar:
I'm still unsure about the error handling. currently the flags module allows everything and doesn't throw any error. This allows also overriding flags that have already been defined, which I think is still a useful feature.
we have also some tests for it: https://github.com/denoland/deno_std/blob/5370d2c440003aff8965c2b77f1724b29f851089/flags/test.ts#L200-L234

Maybe it makes more sense to keep this behaviour by default and add a new option like `strict/throwErrors` to throw errors for duplicated and invalid options?

We could go with solution `iii.` for now and add a `strict/throwErrors` or`noStrict/noThrow` option in a second pass.

flags/mod.ts
```diff
@@ -6,17 +6,204 @@
...
+      & MapTypes<Exclude<T, C>, V>
+      & (T extends undefined ? Record<never, never> : RecursiveRequired<
+        MapTypes<Extract<C, T>, Array<V>>
+      >)
+      : MapTypes<T, V>
+  >;
+
+/** Same as `Record` but also supports dotted and negatable options. */
+type MapTypes<T extends ArgType, V> = undefined extends T ? Record<never, never>
+  : T extends `no-${infer Name}` ? MapTypes<Name, V | false>
```
c4spar, line 132: I think the last thing that is missing is the handling of negatable options. How do we want to proceed with them? do we want to make them configurable by prefixing them with `no-` or with a separate option?

kt3k: @c4spar 
Sorry for the delay in response.

> do we want to make them configurable by prefixing them with no- or with a separate option?

I don't think we should change it optional. That feels like an unnecessary breaking change to me. In my view `--no-foo` shorthand is always useful for boolean args. So I suggest we just keep the existing behavior

For negation of non-boolean args, I can't think of reasonable usages of it. So I'd suggest we should leave it undefined behavior and let's ignore it in typings. What do you think?

c4spar: I still think we should keep support for negation of non boolean args.
A good example is `deno run --no-config` 😉 It's a string arg with a default behavior that can be disabled.

But if we don't make it configurable than all string and collectable args can be also `false`, which is a bit annoying.

kt3k: Ok. `--no-config` looks like a useful example.

> But if we don't make it configurable than all string and collectable args can be also false, which is a bit annoying.

I agree with this. What API do you suggest for configurable negatable args?

Is the following example what you suggest for configurable negatable args? (I think this is not implemented though)
```
    boolean: ["foo", "no-bar", "no-dotted.tick", "dotted.tock"],
    string: ["beep", "no-boop", "no-dotted.zig", "dotted.zag"],
```

c4spar: > Is the following example what you suggest for configurable negatable args? (I think this is not implemented though)

This would be the easiest solution and the typings for this are already implemented in this PR. I have also implemented the logic but only locally, i can open another PR for this.

Another solution would be to add a new `negate` option, same like the `collect` option. But the typings would be more complex for this solution.

kt3k: @c4spar 
> This would be the easiest solution and the typings for this are already implemented in this PR. I have also implemented the logic but only locally, i can open another PR for this.

Thank you for your efforts on this, but this syntax (`boolean: ["no-bar"]`) for indicating negatable arg looks confusing/too implicit to me.

> Another solution would be to add a new negate option, same like the collect option. But the typings would be more complex for this solution.

This solution looks much better to me as it's more explicit and probably easier to document. Also this feels consistent with `collect` option for configuring arg behaviors.

c4spar: ok i got the types for the `negatable` option working now 😉

kt3k (approved):
Great improvement of the usability of `flags`. Nice work!

LGTM

flags/mod.ts
```diff
@@ -31,36 +240,36 @@ export interface ParseOptions {
...
    * non-option. */
   stopEarly?: boolean;
 
   /** A string or array of strings argument names to always treat as strings. */
-  string?: string | string[];
+  string?: S | ReadonlyArray<Extract<S, string>>;
 
   /** A string or array of strings argument names to always treat as arrays.
    * Collectable options can be used multiple times. All values will be
    * colelcted into one array. If a non-collectable option is used multiple
```
dcdunkan, line 266: There's a typo: `Colelcted` to `Collected`?

dcdunkan, line 266: There's a typo here: `colelcted` to `collected`?

c4spar: fixed. Thx @dcdunkan 👍

kt3k:
Ok. Let's land now!

Thanks again for the series of works for landing this PR @c4spar!

PR #2182, opened 2022-05-03
fix(node/worker_threads): Don't wait for parent message in web worker

### LINKED ISSUE #2181: `node/module.ts` hangs in a web worker
Because of node `worker_thread` detection. Fix coming.

nayeemrmn:
Fixes #2181. Uses a different method for detecting if the environment is in a node worker thread so that web workers don't count. That way it doesn't wait for a parent startup message that is never coming.

### DIFF
--- a/node/worker_threads.ts
+++ b/node/worker_threads.ts
@@ -30,6 +30,7 @@ export interface WorkerOptions {
 }
 
 const kHandle = Symbol("kHandle");
+const PRIVATE_WORKER_THREAD_NAME = "$DENO_STD_NODE_WORKER_THREAD";
 class _Worker extends EventEmitter {
   readonly threadId: number;
   readonly resourceLimits: Required<
@@ -54,7 +55,7 @@ class _Worker extends EventEmitter {
     const handle = this[kHandle] = new Worker(
       specifier,
       {
-        ...(options || {}),
+        name: PRIVATE_WORKER_THREAD_NAME,
         type: "module",
         // unstable
         deno: { namespace: true },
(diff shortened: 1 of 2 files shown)


node/worker_threads.ts
```diff
@@ -176,10 +177,12 @@ if (!isMainThread) {
     notImplemented("parentPort.removeAllListeners");
 
   // Receive startup message
+  console.trace();
```
bartlomieju, line 180: Remove?

node/worker_threads.ts
```diff
@@ -176,10 +177,12 @@ if (!isMainThread) {
...
   [{ threadId, workerData, environmentData }] = await once(
     parentPort,
     "message",
   );
+  console.trace(threadId, workerData, environmentData);
```
bartlomieju, line 185: Remove?

node/worker_threads.ts
```diff
@@ -54,7 +55,7 @@ class _Worker extends EventEmitter {
     const handle = this[kHandle] = new Worker(
       specifier,
       {
-        ...(options || {}),
+        name: PRIVATE_WORKER_THREAD_NAME,
```
bartlomieju, line 58: This will cause all workers to have the same name; maybe we could add some kind of counter?

nayeemrmn: Cross worker counter seems like a pain, I thought about `Date.now()` but that's unreliable as well...

The name doesn't serve another purpose and is deleted by the worker on startup, it's probably fine.

bartlomieju (approved):
LGTM

PR #2187, opened 2022-05-03
:bug: Export equalsSimd and equalsNaive

UltiRequiem:
These two methods are used on `bytes/equals_bench.ts` but are not exported, so it throws TypeScript and Runtime Errors, most notably 

```
error: Uncaught SyntaxError: The requested module './equals.ts' does not provide an export named 'equalsNaive'
import { equalsNaive, equalsSimd } from "./equals.ts";
```

I discover these errors while trying to port the remaining benchmarks to the new `Deno.bench` API, similar to #2173

### DIFF
--- a/bytes/equals.ts
+++ b/bytes/equals.ts
@@ -6,7 +6,7 @@
  * @param a first array to check equality
  * @param b second array to check equality
  */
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
Thanks!

PR #2188, opened 2022-05-03
:memo: Update path on docs

UltiRequiem:
The path was updated.

Now is https://github.com/denoland/deno_std/tree/main/node/_tools/test


kt3k (approved):
LGTM, thanks!

PR #2191, opened 2022-05-05
Improved error message for csv's parse function

fadwamahmoud:
Hey, I was using the csv parse function with a poorly coded csv file and the error message wasn't too helpful with tracking what exactly was wrong. 

I added a couple more hints:
- The number of fields found for the headers row
- The number of fields of values row (expected to match the headers)

When this is useful:
- When the number of headers doesn't match one of the subsequent rows

### DIFF
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -242,7 +242,7 @@ export async function parse(
 
     return r.map((e) => {
       if (e.length !== headers.length) {
-        throw `Error number of fields line:${i}`;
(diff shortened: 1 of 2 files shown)


kt3k:
Nice improvement! Can you also add a test case to check this error message?

kt3k:
And also please sign the CLA from the above link

fadwamahmoud:
> Nice improvement! Can you also add a test case to check this error message?

@kt3k Thank you and sorry for the late reply! 
I added the test case and ran `deno fmt`. Please let me know if I'm missing anything.

kt3k (approved):
LGTM. Thanks!

PR #2195, opened 2022-05-06
export createEtagHash,EtagAlgorithm,contentType,MEDIA_TYPES

masx200:
export createEtagHash,EtagAlgorithm,contentType,MEDIA_TYPES

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -46,7 +46,7 @@ interface FileServerArgs {
 const encoder = new TextEncoder();
 const decoder = new TextDecoder();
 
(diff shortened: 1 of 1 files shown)


kt3k:
I don't think these are part of public APIs of `http/file_server`. I recommend copy & pasting these code to your project if you need these features.

We used to have `std/media_types` module, but removed it in https://github.com/denoland/deno/pull/4594 (It's now [/x/media_types](https://deno.land/x/media_types@v3.0.2))

I feel we should add it back to std with some modification addressing https://github.com/denoland/deno/issues/4292

cc @bartlomieju, @ry, @piscisaureus

PR #2198, opened 2022-05-07
fix(node/timers): set custom promisify function

marcushultman:
Fixes promisified timers. Without this fix. this code:
```
import { setTimeout } from 'https://deno.land/std@0.138.0/node/timers/promises.ts';
await setTimeout(2000);
```
Would result in:
```
error: Uncaught TypeError: Callback must be a function. Received 2000
    throw new codes.ERR_INVALID_CALLBACK(callback);
```

I looked around for code on how to easily test this, but could not really find any good way.

### DIFF
--- a/node/timers_test.ts
+++ b/node/timers_test.ts
@@ -1,6 +1,8 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
+import { assert } from "../testing/asserts.ts";
 import timers from "./timers.ts";
+import timersPromises from "./timers/promises.ts";
 
 Deno.test("[node/timers setTimeout]", () => {
   {
@@ -39,4 +41,12 @@ Deno.test("[node/timers setImmediate]", () => {
     const id = timers.setImmediate(() => {});
     timers.clearImmediate(id);
   }
+});
+
+Deno.test("[node/timers/promises setTimeout]", () => {
+  const { setTimeout } = timersPromises;
+  const p = setTimeout(0);
+
+  assert(p instanceof Promise);
+  return p;
 });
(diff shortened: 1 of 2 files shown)


cjihrig:
Can you please add a test. Also, in Node, the [`enumerable` flag is set to `true`](https://github.com/nodejs/node/blob/3bc23f555a8501f5bb37a78cbe866102d02984b5/lib/timers.js#L171).

node/timers_test.ts
```diff
@@ -40,3 +42,11 @@ Deno.test("[node/timers setImmediate]", () => {
     timers.clearImmediate(id);
   }
 });
+
+Deno.test("[node/timers/promises setTimeout]", async () => {
+  const { setTimeout } = timersPromises;
+  const p = setTimeout(100);
```
cjihrig, line 48: I think you could make this value smaller for the purposes of testing.

node/timers_test.ts
```diff
@@ -40,3 +42,11 @@ Deno.test("[node/timers setImmediate]", () => {
...
+
+  assert(p instanceof Promise);
+  await p;
```
cjihrig, line 51: Would it work to just return `p` here?

cjihrig (commented):
Left a few small comments, but mostly LGTM.

It's worth noting that this implementation is the opposite of what Node does. Node implements a promisified version and then returns that when using the symbol, while this promisifies the "normal" version. Functionally, I don't think it should matter though.

kt3k (approved):
LGTM!

marcushultman:
@kt3k Feel free to merge

Issue #2046, opened 2022-03-20
Buffer.prototype.utf8Write fails when called without length


marcushultman:
**Describe the bug**

I don't know exactly where the bug is, so I'll describe what I'm seeing. I use `createRequire` (from `https://deno.land/std@0.130.0/node/module.ts`) in order to import another module that `require('protobufjs/minimal')`. When I end up protobuf encoding a string >= 40 length, protobufjs calls `utf8Write` on the buffer like this:
```
buf.utf8Write(val, pos);
```
(see https://github.com/protobufjs/protobuf.js/blob/master/src/writer_buffer.js#L61)

I notice 2 arguments are passed to the function. The `utf8Write` [function in std@0.130.0/node](https://github.com/denoland/deno_std/blob/0.130.0/node/_buffer.mjs#L777) has 3 arguments though:
```
Buffer.prototype.utf8Write = function utf8Write(string, offset, length) {
```

The last argument (the `length`) is not there in the call from protobufjs. Since `utf8Write` is not a documented API, it's hard for me to know if this is a Deno issue or a protobufjs issue.

**Steps to Reproduce**

1. `npm init`
2. `npm install protobufjs`
3. Create file `repro.ts`:
```
import { createRequire } from "https://deno.land/std@0.130.0/node/module.ts"

const { Writer } = createRequire(import.meta.url)('protobufjs');

// expected: [40, 97, 97, ...]
// actual: [40, 0, 0, ...]
console.log(Writer.create().string('a'.repeat(40)).finish());

// expected: [39, 97, 97, ...] ok!
console.log(Writer.create().string('a'.repeat(39)).finish());
```
4. `deno run --allow-read=. repro.ts`

**Expected behavior**

Expect long strings to be written in protobufjs

**Environment**

- OS: macOS 12.0.1
- deno version: deno 1.20.1
- std version: 0.130.0

This may be a bug in protobufjs - I have not filed an issue there, please close this if I should.

Thanks ⭐️

marcushultman:
Looking at [node source code](https://github.com/nodejs/node/blob/master/src/node_buffer.cc#L714), it looks like the `length` argument does exist, but it's optional.

marcushultman:
Perhaps a fix like this is reasonable: https://github.com/denoland/deno_std/commit/fdaa7b9cc77b387a04177fd1749d732c61e9ac4d

PR #2199, opened 2022-05-07
fix(node/module): interpret length parameter as optional in blitBuffer

marcushultman:
As indicated by the [node source code](https://github.com/nodejs/node/blob/ce29d28/src/node_buffer.cc#L726), the `length`parameter is optional and defaults to the length of the `src`.

Closes https://github.com/denoland/deno_std/issues/2046


kt3k (approved):
Confirmed this works in the same way as Node. LGTM

Thanks for the fix!

Issue #2122, opened 2022-04-19
crypto: add fnv1a hash


littledivy:
Like the one in `std/hash` https://github.com/denoland/deno_std/blob/main/hash/_fnv

luk3skyw4lker:
I can take this one. I'll take a look at the algorithm today and create an implementation.

luk3skyw4lker:
It seems like they're already implemented. There are two classes being exported of the module and one is the fnv algorithm and the other is the fnv-a

luk3skyw4lker:
Why the need for them to be at the crypto module? Is it because it's internal? I didn't fully understand the need for it.

KyleJune:
It's because hash module is deprecated. They are going to delete the whole hash folder. Instead of deleting fnv that can't be replaced with WebCrypto and std/crypto, I think they are just going to just move it to std/crypto folder instead of keeping this folder with just fnv algorithms in it.

luk3skyw4lker:
Dang! That totally slipped through me, I'm sorry!

I'll do the work of implementing the algorithms then, thanks for the heads up!

luk3skyw4lker:
In that case, wouldn't be the case to migrate all of the hashing algorithms to the crypto module? Or I am wrong?

KyleJune:
I think all the other ones in that directory can be done by WebCrypto or std/crypto. fnv is a non cryptographic hash, which is why it currently doesn't have a replacement. I'm not sure about internal discussions at Deno on this task, but I think the fnv implementation in hash could just be moved to std/crypto instead of re-implementing it.

luk3skyw4lker:
I'm checking with the maintainers about the internals, thanks for your tips!

littledivy:
Yup `std/hash` is deprecated so we need this in `std/crypto`. You can simply copy over the algorithm from `std/hash/_fnv` to expose this API:

```typescript
import { crypto } from "https://deno.land/std/crypto/mod.ts";

// 32bit
await crypto.subtle.digest("fnv32", new TextEncoder().encode("deno")); // 0x6e, 0xd5, 0xa7, 0xa9
await crypto.subtle.digest("fnv32a", new TextEncoder().encode("deno")); // 0x8e, 0xf6, 0x47, 0x11

// 64bit
await crypto.subtle.digest("fnv64", new TextEncoder().encode("deno")); // 0x14, 0xed, 0xb2, 0x7e, 0xec, 0xda, 0xad, 0xc9
[...]

PR #2200, opened 2022-05-07
feat: add fnv alg to crypto

luk3skyw4lker:
Add implementation of fnv32, fnv32a, fnv64, fnv64a to crypto module. Addressing #2122


crypto/test.ts
```diff
@@ -1281,10 +1340,44 @@ const digestCases: [
...
 
+Deno.test("[crypto/digest/fnv] fnv algorithm implementation", () => {
+  const inputString = "deno";
+  const inputBytes = new TextEncoder().encode(inputString);
+
+  const expectedDigest32 = "a9a7d56e";
+  const expectedDigest32a = "1147f68e";
+
+  const expectedDigest64 = "7eb2ed14c9addaec";
+  const expectedDigest64a = "67fbd9a5b1486e42";
```
kt3k, line 1351: These expected values seem reversed in each 4-bytes when I compared them to golang's output. How did you create these expectations?

スクリーンショット 2022-05-09 14 20 48

https://go.dev/play/p/R7B_HpWXw5u

kt3k: also tested with `@sindresorhus/fnv1a` npm package (which only provides `-a` variant)

test.mjs
```js
import fnv1a from '@sindresorhus/fnv1a';
console.log(fnv1a("deno", { size: 32 }).toString(16));
console.log(fnv1a("deno", { size: 64 }).toString(16));
```
```sh
$ node test.mjs
8ef64711
a5d9fb67426e48b1
```

crypto/mod.ts
```diff
@@ -158,16 +166,20 @@ const webCryptoDigestAlgorithms = [
...
 type DigestAlgorithmObject = {
   name: DigestAlgorithmName;
   length?: number;
 };
 
-type DigestAlgorithm = DigestAlgorithmName | DigestAlgorithmObject;
+type DigestAlgorithm =
+  | DigestAlgorithmName
+  | DigestAlgorithmObject
+  | FNVAlgorithms;
```
kt3k, line 179: I think `FNVAlgotithms` should be appended to `DigestAlgorithmName` type instead of `DigestAlgorithm` like:

```ts
type DigestAlgorithmName = WasmDigestAlgorithm | FNVAlgorithms;
```

With that change, we can call `digest` like `crypto.subtle.digest({ name: "FNV32" }, data)` (Current implementation only allows `crypto.subtle.digest("FNV32", data)` signature)

kt3k (commented):
@luk3skyw4lker Thank you for looking into this! Left a few comments.

crypto/mod.ts
luk3skyw4lker: Right, when I tried that Typescript complained about a lot of types at the test file. But I'll try it again anyway.

crypto/test.ts
luk3skyw4lker: I caught the implementation directly from the hash folder and made only some modifications about it being a simpler function. But I'll check it out, I must have done something wrong in the process.

luk3skyw4lker:
@kt3k thanks for the review!

luk3skyw4lker:
@kt3k I think I have solved the issues. I wasn't able to keep the return of the fnv functions as ArrayBuffers because something was happening when I tried to convert the hashes into Uint32Arrays and the result was reversed (I think it may be because of MSB and LSB problems, I didn't manage to find information or fix for that) so I just convert the string to hex inside the function itself and the result comes as a hex string (and with the right ordenation).

crypto/mod.ts
kt3k: again, why can't `FNVAlgorithms` be added to `DigestAlgorithmName`?

kt3k (commented):
Thanks for updating! The resulted values look correct. Nice progress, but left a few comments on some points.

crypto/mod.ts
luk3skyw4lker: Oh, I misunderstood the previous comment, sorry. I thought I made the right change but I didn't :x

crypto/mod.ts
```diff
@@ -61,10 +59,14 @@ const stdCrypto = ((x) => x)({
     async digest(
       algorithm: DigestAlgorithm,
       data: BufferSource | AsyncIterable<BufferSource> | Iterable<BufferSource>,
-    ): Promise<ArrayBuffer> {
+    ): Promise<ArrayBuffer | string> {
```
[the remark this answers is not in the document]
luk3skyw4lker: I'll take a look into it, but as I said earlier, the problem with the inverted values was because of the return being a ArrayBuffer (I think it's because when I convert the hash into the ArrayBuffer type it stores it as a little-endian value, when in fact we want the big-endian value). But I'll see what I can do and let you know.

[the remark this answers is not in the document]
luk3skyw4lker: About this, I sent the PR into the discord server to see if anyone could help me. I have ran out of ideas to make this work as a ArrayBuffer.

luk3skyw4lker:
@kt3k I think that everything is right now! I had a lot of trouble to make the returns come as ArrayBuffers but I did it

kt3k (approved):
@luk3skyw4lker LGTM. Thanks!

PR #2203, opened 2022-05-08
BREAKING(testing/asserts): remove implicit any types from assert signatures.

### LINKED ISSUE #2202: testing/asserts.ts: Consider making `assertEquals()` and friends more strictly typed.

At the moment, [`assertEquals()`](https://deno.land/std@0.138.0/testing/asserts.ts#L155), [`assertNotEquals()`](https://deno.land/std@0.138.0/testing/asserts.ts#L201), [`assertNotStrictEquals()`](https://deno.land/std@0.138.0/testing/asserts.ts#L297) and [`assertArrayIncludes()`](https://deno.land/std@0.138.0/testing/asserts.ts#L446) have multiple signatures, one of them not containing a generic parameters. This causes two issues:

 - It's essentially an implicit `any`, it's as if every call to `assertEquals()` is invoked like `assertEquals<any>(a,b)`, without any warnings that this is the case. Users might not realise that this is the case, causing them to write less strict code. Or they might simply forget to add the generic parameter.
 - In JavaScript it's not possible to add type checking in this case, because `assertEquals<number>(2,3)` is not valid syntax and there's no JSDoc equivalent for this.

(Also note that [`assertStrictEquals`](https://deno.land/std@0.138.0/testing/asserts.ts#L243) does not have multiple signatures, I think that might be a mistake.)

Change the types to not include a signature without a generic argument. It will cause some breakage, for instances where unexpected types are used as arguments, but these are very easy to workaround by either casting or passing `any` as generic argument.
I tried this on a project with about 800 unit tests and it resulted in 7 new type errors, all of which were only exposing existing type issues with the parameters passed in.

The issue is pretty easy to work around by simply writing a wrapper for these functions that don't have a signature with the missing generic, like so:
```ts
function assertEquals<T>(actual: T, expected: T, msg?: string): void {
    assertEquals(actual, expected, msg);
}
```
I reckon the amount of users using the asserts library with pure JavaScript is pretty small, so a wrapper function is definitely a viable option for these users. Though that still leaves TypeScript users with less safe typings with no warnings about this.

jespertheend:
This is a breaking change because things like
```ts
assertEquals(5, "string");
```
will now result in a type error, which can be worked around like so:
```ts
assertEquals<unknown>(5, "string");
```
or like so:
```ts
assertEquals(5, "string" as unknown as number);
```

Fixes #2202


jespertheend:
Let me know if this change is desired. I fixed all new type errors by adding `<unknown>` right now, which I think could be improved a bit by only casting specific parts of the types.

kt3k:
@jespertheend Can you elaborate on the benefit of this change? Regarding the changes in assertions in this PR, this seems only adding boilerplate `<unknown>` without catching any error in the codebase to me.

bartlomieju (commented):
@jespertheend what's the rationale for this PR? Seems like potentially a huge breaking change for users

jespertheend:
Oops sorry, forgot to link to the original issue: #2202

jespertheend:
> Regarding the changes in assertions in this PR, this seems only adding boilerplate <unknown> without catching any error in the codebase to me.

What prompted me to write this was that I made a change that caused a bunch of my tests to break. Something that would have been catched if assertEquals was actually checking if `actual` and `expected` are forced to have the same type.

So for instance if you have:
```ts
const someVariable : number = 3;
assertEquals(someVariable, 3);
```
and at some point you end up changing the type of `someVariable`, you'll immediately see where your tests will break because it will give a type error.

kt3k:
> and at some point you end up changing the type of someVariable, you'll immediately see where your tests will break because it will give a type error.

But you'll see the assertion errors anyway when you run the test cases. So I think it only adds subtle values, and I'm skeptical if it's worth the huge breaking changes in the users' code

jespertheend:
Another benefit I found is that this makes autocompletions work for the second argument. So if you have complex structures it's easier to fill in the expected value.

cmorten:
Would a subtler change for now be to make use of the of a fairly permissive generic parameter default, e.g. `any`. This way users keen on type safety can opt in with stricter typing, but otherwise it’s non breaking for everyone else?

Usage of assert can often be when you’re not confident of the value and _type_ of the “actual” value. I suspect in some cases it would simply prompt folks who are “looser” with their typing to start scattering `as …` or `unknown` simply to make the `assert*` not complain. Having a loose default make save on effort and LOC in userland.

REF: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults

Sorikairox:
My question is "Should we test that 2 variable of different type are equals?"

I believe that the answer is no, thus I think that this PR makes sense and leads developers towards a cleaner code.

Yes it is a breaking change but I find it justified.

cmorten:
> My question is "Should we test that 2 variable of different type are equals?"
> 
> I believe that the answer is no, thus I think that this PR makes sense and leads developers towards a cleaner code.
> 
> Yes it is a breaking change but I find it justified.

For the most part agree, just suspect (as demonstrated by this very PR) the reality will probably not to force folks to write cleaner, better typed code, but rather scatter `<unknown>` or type casting as required to work around the type violations. But I may be being overly cynical 😅 Just wonder whether an opt-in approach would allow folks to introduce stronger typing while not breaking everyone else would be a progressive enhancement?

Certainly during `0.x.x` is the better time to experiment and break than once go major.

jespertheend:
> Would a subtler change for now be to make use of the of a fairly permissive generic parameter default, e.g. any or unknown.

That is the current behaviour, this PR actually proposes removing that default.

> just suspect (as demonstrated by this very PR) the reality will probably not to force folks to write cleaner, better typed code, but rather scatter <unknown> or type casting as required to work around the type violations.

The reason why I scattered `<unknown>` in this PR was because I didn't want to spend a lot of time before being certain if the change would get approved. Also I wanted to demonstrate that the breakage could easily be worked around by passing `any` or `unknown`. For instance, a bunch of type errors in deno_std came from the usage of strings rather than enums in `_diff_test.ts`. So this can be worked around by replacing the strings with the `DiffType` enum.

Even if those who prefer to have more loosely typed asserts would scatter the occasional `<unknown>` that would still be a win in my opinion. Because then the majority of asserts is still type checking by default. So if you ever change the return value of a function somewhere, you'll get instant feedback about the change and see which tests will break.

testing/asserts.ts
```diff
@@ -152,17 +152,7 @@ export function assertFalse(expr: unknown, msg = ""): asserts expr is false {
...
-  expected: unknown,
-  msg?: string,
-): void;
-export function assertEquals<T>(actual: T, expected: T, msg?: string): void;
-export function assertEquals(
-  actual: unknown,
-  expected: unknown,
-  msg?: string,
-): void {
+export function assertEquals<T>(actual: T, expected: T, msg?: string): void {
```
dsherret, line 155: The one really nice thing about typing things this way, is that you get auto-complete on the expected values, which can be especially nice when providing an object. For example:

```ts
assertEquals(actual, {
  someProp: 5,
  // ...auto-complete here...
});
```

kt3k (approved):
Looks like there's enough support on this. I'm now in favor of this change. Let's try it. LGTM

jespertheend:
Alright, let me try and fix up some of those `<unknown>`s.

jespertheend:
Alright most `<unknowns>` are removed and fixed or replaced by a narrower cast.
Not sure, maybe this is overkill, but I reckon this makes the code at least somewhat more robust.

kt3k:
Are we ready to land this? @bartlomieju

bartlomieju (approved):
LGTM

PR #2205, opened 2022-05-09
feat(node/dgram): support dgram (udp) node module compat

### LINKED ISSUE #2110: compat: add dgram.Socket
`dgram.Socket` is needed to run [multicast-dns](https://www.npmjs.com/package/multicast-dns) in compat mode. (feature request from community)

cmorten:
Works towards #2110 

- There are a number of APIs that this PR doesn't implement, for the most part because not sure we can right now?
- Some implemented APIs may differ subtly from Node (inevitable perhaps when trying to match JS / Deno interfaces with libuv! 😅).


node/internal_binding/udp_wrap.ts
```diff
@@ -1,3 +1,494 @@
...
+    if (code) {
+      const errno = codeMap.get(code)!;
+      ctx.errno = errno;
+      ctx.code = code;
+      ctx.message = errorMap.get(errno)![1];
+      ctx.syscall = buffer ? "uv_recv_buffer_size" : "uv_send_buffer_size";
+
+      return;
+    }
+
```
cmorten, line 153: TODO: if is Linux then size*2

bartlomieju:
@cjihrig can you take a look too?

node/internal_binding/handle_wrap.ts
```diff
@@ -38,11 +38,11 @@ export class HandleWrap extends AsyncWrap {
   }
 
   ref() {
-    notImplemented("HandleWrap.prototype.ref");
+    unreachable();
```
bartlomieju, line 41: Why are these unreachable?

bartlomieju (commented):
Awesome work, lots of tests enabled! This looks mostly good to me, but I'll wait for review from @cjihrig before landing

node/internal_binding/handle_wrap.ts
cmorten: As far as can see currently these methods exist purely for defining an interface on classes rather than something that will ever be implemented - felt more accurate to fail with unreachable than not implemented.

Wonder whether they should be defined at all… or if we want them, make use of the inheritance better and not override in all of the other wraps

node/buffer_test.ts
```diff
@@ -604,3 +604,14 @@ Deno.test({
     });
   },
 });
+
+Deno.test({
+  name:
+    "utf8Write handle missing optional length argument (https://github.com/denoland/deno_std/issues/2046)",
```
kt3k, line 610: Thanks for moving this test here!

kt3k (approved):
LGTM, nice work!

cjihrig (approved):
LGTM. Nice work!

kt3k:
@cjihrig Thanks for your review!

PR #2209, opened 2022-05-10
fix: add toJSON to node/perf_hooks.ts

GJZwiers:
This is needed together with https://github.com/denoland/deno/pull/14548 to land performance.toJSON, but there is a circular dependency because the changes to `.d.ts` in https://github.com/denoland/deno/pull/14548 are needed for the changes in this PR to pass the type-checks, while the changes in this PR are also needed for the CI to pass in denoland/deno


kt3k:
Code change looks good to me.

@bartlomieju Should we land this before https://github.com/denoland/deno/pull/14548?

bartlomieju:
> Code change looks good to me.
> 
> @bartlomieju Should we land this before [denoland/deno#14548](https://github.com/denoland/deno/pull/14548)?

Yes, I will land it and update git submodule in Deno so we unblock that PR

bartlomieju (approved):

PR #2212, opened 2022-05-12
0.139.0

denobot:
Bumped version for 0.139.0

Please ensure:
- [ ] Version in version.ts is updated correctly
- [ ] Releases.md is updated correctly
- [ ] All the tests in this branch have been run against the CLI release being done (`../deno/target/release/deno test -A --unstable`)

To make edits to this PR:
```shell
git fetch upstream release_0_139.0 && git checkout -b release_0_139.0 upstream/release_0_139.0
```

cc @piscisaureus


piscisaureus:
Note that #2210 (update snapshot expectations to match deno/main output) causes the snapshot tests to fail with deno v1.21.3. 
There's not much we can do about that other than revert #2210 which I don't want to do.

piscisaureus (approved):

PR #2216, opened 2022-05-12
fix(encoding): fix jump link target

valtlai:
This fixes the `jsonc` jump link on the following page: https://deno.land/std@0.139.0/encoding#encoding


valtlai:
Oh, just noticed this is a duplicate of #2215, co closing.

PR #2218, opened 2022-05-13
feat(http): add onListen option to serve

### LINKED ISSUE #2071: Provide a way to tell when the server is ready

When making a web server with `std/http` and listening with the `serve()` function or the `Server` class, It is difficult to write code for logs and tests (etc) that rely on the server being ready, because readiness is never signaled and the promise never resolves while the server is running.

```js
import { Server } from 'https://deno.land/std@0.132.0/http/server.ts';

const server = new Server({
    handler() {
        return new Response('Hello, World!');
    },
    port : 3000
});

await server.listenAndServe();
console.log('Server ready');     // <-- This line is never reached
```

```js
import { Server } from 'https://deno.land/std@0.132.0/http/server.ts';

const server = new Server({
    handler() {
        return new Response('Hello, World!');
    },
    port : 3000
});

server.listenAndServe();               // <-- Doesn't wait for server to be ready
await fetch('http://localhost:3000');  // <-- Prone to race conditions
```

I would like the promise returned by [Server#listenAndServe()](https://github.com/denoland/deno_std/blob/7213d527ec552907daec32f16143f53b73d50a3c/http/server.ts#L176-L188) (and functions that use it) to resolve when the server is listening (i.e. when it is ready for requests) as opposed to never resolving, which isn't useful.

Having a callback separate from the promise resolution, as proposed in https://github.com/denoland/deno_std/issues/2047 is an alternative approach, but IMO that is less desirable.

### LINKED ISSUE #2186: http/server: disable startup message behavior for `serve` fn

Recently, a controversial PR #1641 was merged (see feedback in the PR for details), which changed the exported function [`serve`](https://github.com/denoland/deno_std/blob/0.137.0/http/server.ts#L525-L568), causing it to write to `stdout` when invoked (via a call to `console.log`). The content that is written is an English-localized message including information about the origin on which the server is listening. There is currently no way to disable this behavior, and it is not desirable in many cases.

It [was suggested](https://github.com/denoland/deno_std/pull/1641#issuecomment-981103163) to use the `listenAndServe` function as an alternative to avoid the behavior, but since [that function is deprecated](https://github.com/denoland/deno_std/blob/0.137.0/http/server.ts#L630), the suggestion is a non-option.

I think this PR should be reverted. Alternatively, a new property should be added to the options argument which controls this newly introduced behavior and **defaults to not writing a startup message**, making the behavior opt-in.

**Related thoughts**

Here are a some things which I think should be considered and weren't discussed in the PR:

- What if the user of the function doesn't read or write English and would prefer a startup message in their own language?

- What if the function is being called by another process which is parsing `stdout`, and it introduces unnecessary new complexity by writing this message? The person would need to refactor their parser to accommodate for this behavior.

- `serve` could be invoked in an environment where calling `console.log` will throw an exception.

- What if the user simply does not want anything written to `stdout` as a side-effect of using the function?

kt3k:
This PR adds `onListen` option to `serve` and `serveTls` functions. You can override the default logging bahavior of these functions.

closes #2186 
closes #2071 

Note: I didn't add `secure: boolean` option to params for `onListen` callback (suggested in https://github.com/denoland/deno_std/issues/2186#issuecomment-1125347398) as it should be obvious from the function you are using (i.e. `secure` is false when you use `serve`, and `secure` is true when `serveTls`.)

- [x] Add tests

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -475,6 +475,9 @@ export interface ServeInit extends Partial<Deno.ListenOptions> {
 
   /** The handler to invoke when route handlers throw an error. */
   onError?: (error: unknown) => Response | Promise<Response>;
+
+  /** The callback which is called when the server started listening */
+  onListen?: (params: { hostname: string; port: number }) => void;
 }
 
 /**
@@ -524,8 +527,8 @@ function hostnameForDisplay(hostname: string) {
 
 /** Serves HTTP requests with the given handler.
  *
- * You can specify an object with a port and hostname option, which is the address to listen on.
- * The default is port 8000 on hostname "0.0.0.0".
+ * You can specify an object with a port and hostname option, which is the
+ * address to listen on. The default is port 8000 on hostname "0.0.0.0".
  *
  * The below example serves with the port 8000.
  *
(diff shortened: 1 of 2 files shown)


jsejcksn:
Hi @kt3k: I didn't see this PR until after it was merged, but I have some feedback:

> Note: I didn't add `secure: boolean` option to params for `onListen` callback (suggested in [#2186 (comment)](https://github.com/denoland/deno_std/issues/2186#issuecomment-1125347398)) as it should be obvious from the function you are using (i.e. `secure` is false when you use `serve`, and `secure` is true when `serveTls`.)

While this is true, it means that a user must write two different functions, one for each case, rather than being able to rely on the parameter object to discriminate. It would be helpful to include the `secure` boolean property. If it's helpful to think about an example: imagine a case where someone would like to write a single function to log a startup message which is localized in a non-English language.

PR #2226, opened 2022-05-13
feat(testing/asserts): return error from `assertRejects` and `assertThrows`

### LINKED ISSUE #1362: std assertThrows is acting like assertThrowsError
I just started building everything in Deno - even when node is the target (after a bundle and some creative imports). Absolutely love it. 

I found something that seems odd to me in std / assert. I would like to make a PR for it, but first I would like to hear if there is a (good) reason it is currently implemented like this. 

The `assertThrows` as described in `https://deno.land/manual/testing/assertions` is documented as if it asserts that "something" is being thrown. Yet 

```
Deno.test('Naming is hard',()=>{
	assertThrows(()=>{
		throw "msg"
	})
})
```

returns `AssertionError: A non-Error object was thrown.`. You will have to return a proper `Error` to get a pass

```
Deno.test('Naming is hard',()=>{
	assertThrows(()=>{
		throw Error("msg")
	})
})
````

On [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw) the throw statement is defined as

image

It seems odd to *demand* a thrown exception is of a specific type (`Error`) when the name `assertThrows` indicates that it is testing **if** something is thrown. 

**I suggest that the current behaviour gets renamed to `assertThrowsError` and `assertThrows` is changed to disregard what type is thrown (also it's async sibling).** 

Let me know if there is some hidden knowledge or good reasons for the current behaviour. And if that is the case I'm happy to make a PR to the docs repo to explain it better to the next dev coming along this path of the world of Deno.  

I do understand that it is a breaking change, but better to sort out consistency early than carrying confusion into the future... right? If not, maybe a way forward would be to add `assertThrowsException` to verify that an exception _was_ thrown but disregard the type.

### LINKED ISSUE #2220: `assertRejects` and `assertThrows` should return the caught error
It'd be nice if these functions returned the error they caught, so users could do more intricate assertions on the error itself.

### PR DESCRIPTION
**Pull request by mrkldshv**
Makes `assertRejects` and `assertThrows` return caught error.
* Deprecate error callback from `assertRejects` and `assertThrows` signatures
* Allow throwing non-error values

Closes #1362, #2220.

### CODE REVIEW

File: testing/asserts.ts [hunk 1]
```diff
@@ -647,12 +647,12 @@ export function assertThrows<E extends Error = Error>(
   ErrorClass?: new (...args: any[]) => E,
   msgIncludes?: string,
   msg?: string,
-): void;
+): Error | void;
```
KyleJune, line 650: I think it would make sense to change this to returning `unknown` instead of `Error | void` since the thrown or rejected value could be a non string value. If called with an `ErrorClass` or `errorCallback`, I believe the return type should be `E` since it will assert that the thrown exception is of type `E`.

KyleJune: I think it would make sense to update the catch block to also only do the following if an ErrorClass or errorCallback is set. That way these functions can be used for checking that non error values are thrown or rejected. If we did that, this would solve this issue too. https://github.com/denoland/deno_std/issues/1362#issuecomment-940101918

```ts
if (ErrorClass || errorCallback) {
  if (error instanceof Error === false) {
    throw new AssertionError("A non-Error object was thrown.");
  }
  assertIsError(
      error,
      ErrorClass,
      msgIncludes,
      msg,
  );
  if (typeof errorCallback == "function") {
      errorCallback(error);
  }
}
```

KyleJune (changes requested):
I'm not on the core team, just another contributor. Giving my feedback because I worked in this area before and think the change in call signature issue is a good opportunity to also fix another issue related to `assertThrows` and `assertRejects`.

testing/asserts.ts
mrkldshv: Thank you for feedback. If understand correctly, we'd like to allow function throwing values not only of `Error` type. So the return type of function would change to `unknown` and it will look something like that:
```ts
function assertThrows<E extends Error = Error>(...): unknown {
  // ...
  catch (error) {
    if (ErrorClass || errorCallback) {
      if (error instanceof Error === false) {
        throw new AssertionError("A non-Error object was thrown.");
      }
      assertIsError(
        error,
        ErrorClass,
        msgIncludes,
        msg,
      );
      if (typeof errorCallback == "function") {
        errorCallback(error);
      }
    }
    err = error;
    doesThrow = true;
  }
  // ...
  return err;
}
```

KyleJune: Yes, that's how I think it should work, then have return type be E for call signatures that include ErrorClass or errorCallback.

You can wait for a core team member to respond if you want. This is just my opinion as it would solve 2 issues at once while preserving backward compatibility. If someone wants to assert something was thrown other than an error, they would just do `const exception = assertThrows(() => { /* ... */ })` then make their assertions about exception.

mrkldshv: Right, makes sense for me. Let's wait for the input of the team core member.

KyleJune: Here's what I meant in my last comment regarding call signatures and the reasoning behind them.

I think maybe something like this. The differences from the current call signature for the function are.
1. Added an `assertThrows` call signature that takes fn and optionally take msg.
2. Made `ErrorClass` not optional for the old first call signature.
3. Add a call signature without `ErrorClass` but with `msgIncludes`.
```ts
export function assertThrows(
  fn: () => unknown,
  msg?: string,
): unknown;
export function assertThrows<E extends Error = Error>(
  fn: () => unknown,
  ErrorClass: new (...args: any[]) => E,
  msgIncludes?: string,
  msg?: string,
): E;
export function assertThrows(
  fn: () => unknown,
  msgIncludes: string,
  msg?: string,
): Error;
export function assertThrows(
  fn: () => unknown,
  errorCallback: (e: Error) => unknown,
  msg?: string,
): Error;
export function assertThrows<E extends Error = Error>(
  fn: () => unknown,
  errorClassOrCallback?:
    | (new (...args: any[]) => E)
    | ((e: Error) => unknown),
  msgIncludesOrMsg?: string,
  msg?: string,
): E | Error | unknown {
  //...
}
```
With these call signatures, `assertThrows` could be used in the following ways.
```ts
// assert an exception was thrown. Type would be unknown.
const exception = assertThrows(() => { throw "oops" })
assertEquals(exception, "hello");
// asserts that the exception was an AssertionError that says oops. Returns it as type AssertionError so you can make additional assertions on it.
let error = assertThrows(() => { throw new AssertionError("oops") }, AssertionError, "oops")
// asserts that the exception was an Error of any kind that says oops. Returns it as type Error so you can make additional assertions on it.
error = assertThrows(() => { throw new AssertionError("oops") }, "oops")
// asserts that the exception was an Error of any kind that says oops. Returns it as type Error so you can make additional assertions on it.
error = assertThrows(() => { throw new AssertionError("oops") }, (error: Error) => {
  // here for backwards compatibility
  // it asserts the exception was an error before passing it to the callback.
})
```

We could also remove the last call signature where you can pass in an errorCallback since the same could be accomplished by doing the following then making your assertions on the returned value. I believe the only reason errorCallback was added was to work around the issue that it was previously decided that assertion function's cannot return values.

```ts
const error = assertThrows(() => { throw new Error("oops") }, Error);
// make additional assertions about error here instead of in an errorCallback.
```

We can let core team decide whether or not they want to remove the errorCallback. I'm personally think it should be removed.

kt3k: @KyleJune @mrkldshv 
I agree with removing errorCallback signature in the future, but let's first deprecate it. I think we can deprecate only specific overloaded signature by putting `@deprecated` jsdoc tag right above the deprecated one.

Suggested signature looks great to me 👍

mrkldshv: Alright, thanks for the feedback. I'll implement necessary changes 👍

collections/sliding_windows_test.ts
```diff
@@ -22,7 +22,7 @@ function slidingWindowsThrowsTest<T>(
     size: number,
     config?: { step?: number; partial?: boolean },
   ],
-  ErrorClass?: ErrorConstructor | undefined,
+  ErrorClass: ErrorConstructor,
```
mrkldshv, line 25: I've changed it to required, because none of the signatures of `assertThrows` function matches the way this function is called. I'm open to refactoring it to keep it optional though. It can be something like:
```ts
if (ErrorClass) {
    assertThrows(
    () => {
      slidingWindows(...input);
    },
    ErrorClass,
    msgIncludes,
    msg,
  );
} else {
  assertThrows(
    () => {
      slidingWindows(...input);
    },
    msg,
  );
}
```

testing/asserts.ts
```diff
@@ -641,66 +641,86 @@ export function assertIsError<E extends Error = Error>(
...
+      errorClassOrCallbackOrMsg.prototype instanceof Error ||
+      errorClassOrCallbackOrMsg.prototype === Error.prototype
+    ) {
+      // deno-lint-ignore no-explicit-any
+      ErrorClass = errorClassOrCallbackOrMsg as new (...args: any[]) => E;
+      msgIncludes = msgIncludesOrMsg;
+    } else {
+      errorCallback = errorClassOrCallbackOrMsg as (e: Error) => unknown;
+      msg = msgIncludesOrMsg;
+    }
```
mrkldshv, line 694: I've made couple changes here: 
* swapped non-strict equality to strict one
* removed `errorCallback = null`
* added init value
They seem to be valid, but I'm not sure if I'm missing something.

testing/asserts.ts
```diff
@@ -641,66 +641,86 @@ export function assertIsError<E extends Error = Error>(
  * callback which will be passed the error, usually to apply some custom
  * assertions on it.
```
mrkldshv, line 642: Was thinking to remove this bit about error callback from the documentation not to encourage developers to use it, because we're deprecating this parameter. What do you think?

testing/asserts.ts
```diff
@@ -641,66 +641,87 @@ export function assertIsError<E extends Error = Error>(
...
+  ErrorClass: new (...args: any[]) => E,
   msgIncludes?: string,
   msg?: string,
-): void;
+): E;
+export function assertThrows(
+  fn: () => unknown,
+  msgIncludes: string,
+  msg?: string,
+): Error;
```
kt3k, line 659: When I run the code like the below, it doesn't throw the assertion error (`msgIncludes` is not compared with thrown error message). I think this signature isn't supported by this change. Why is this signature added? How about just removing this?

```ts
import { assertThrows } from "./testing/asserts.ts";

assertThrows(
  () => {
    throw new Error("foo");
  },
  "bar",
  "baz",
);
```

testing/asserts.ts
kt3k: Looks fine to me. We don't seem supporting `errorCallback === null` case according to the typing.

node/_crypto/crypto_browserify/evp_bytes_to_key_test.ts
```diff
@@ -10,7 +10,6 @@ Deno.test("salt buffer length is 7", function () {
     function () {
       EVP_BytesToKey(Buffer.alloc(5), Buffer.alloc(7), 1, 1);
     },
-    undefined,
```
kt3k, line 13: I think this should be `Error`. Now we use `assertThrows(fn: () => unknown, msg?: string)` signature here and the 2nd arg is not compared with the thrown error message (it's used as just informative `msg`)

node/module_test.ts
```diff
@@ -131,7 +131,6 @@ Deno.test("EventEmitter is exported correctly", () => {
 Deno.test("Require .mjs", () => {
   assertThrows(
     () => require("./testdata/inspect.mjs"),
-    undefined,
```
kt3k, line 134: This has the same issue as the above one (`crypto_browserify/evp_bytes_to_key_test.ts`)

testing/asserts.ts
```diff
@@ -710,43 +731,59 @@ export function assertThrows<E extends Error = Error>(
...
+  ErrorClass: new (...args: any[]) => E,
   msgIncludes?: string,
   msg?: string,
-): Promise<void>;
+): Promise<E>;
+export function assertRejects(
+  fn: () => Promise<unknown>,
+  msgIncludes: string,
+  msg?: string,
+): Promise<Error>;
```
kt3k, line 749: This signature seems having the same problem as the case of `assertThrows` (`msgIncludes` is not compared with the throws error message).

```ts
import { assertRejects } from "./testing/asserts.ts";
await assertRejects(
  async () => {
    throw new Error("foo");
  },
  "bar",
  "baz",
);
```

testing/asserts.ts
```diff
@@ -641,66 +641,87 @@ export function assertIsError<E extends Error = Error>(
...
   msgIncludes?: string,
   msg?: string,
-): void;
+): E;
+export function assertThrows(
+  fn: () => unknown,
+  msgIncludes: string,
+  msg?: string,
+): Error;
+/** @deprecated */
```
kt3k, line 660: 👍

testing/asserts_test.ts
```diff
@@ -860,6 +968,26 @@ Deno.test("AssertRejectsWithErrorCallback", async () => {
   );
 });
 
+Deno.test("assertThrows returns caught error", () => {
```
kt3k, line 971: Can you also add similar test cases where the thrown values are not errors, but `string`s, `number`s, etc?

testing/asserts.ts
mrkldshv: Right, I see. I'll remove it as well. What do you think?

testing/asserts_test.ts
mrkldshv: Sure, will add more test cases.

testing/asserts.ts
kt3k: Removing sounds good to me

kt3k (approved):
@mrkldshv LGTM. Nice work!

Issue #2208, opened 2022-05-10
Proposal: Web stream based jsonlines parser


ayame113:
Would you like to add stream based [jsonlines](https://jsonlines.org/) parser as suggested in discord?

https://discord.com/channels/684898665143206084/712010403302866974/972924800852951040

```
{"some":"thing"}
{"foo":17,"bar":false,"quux":true}
{"may":{"include":"nested","objects":["and","arrays"]}}
```

```ts
import { JSONLinesParseStream } from "...";

const { body } = await fetch("...");
const readable = body!
  .pipeThrough(new TextDecoderStream())
  .pipeThrough(new JSONLinesParseStream());
for await (const data of readable) {
  console.log(data);
}
```

Implementations can be copied from https://deno.land/x/jsonlines@v1.1.0.

ayame113:
My another suggestion is to add `transformStreamFromGeneratorFunction`, a utility function used [inside jsonlines modules](https://deno.land/x/jsonlines@v1.1.0/mod.ts#L86), to the strams module.

```ts
import {
  readableStreamFromIterable,
  transformStreamFromGeneratorFunction
} from "...";

const reader = readableStreamFromIterable([0, 1, 2])
  .pipeThrough(transformStreamFromGeneratorFunction(async function* (src) {
    for await (const chunk of src) {
      yield chunk * 100;
    }
  }));

for await (const chunk of reader) {
  console.log(chunk);
}
// output: 0, 100, 200
```

crowlKats:
I am in favour of this, though unsure about the latter.

bartlomieju:
The name is very long, maybe `generatorToStream`?

kt3k:
I'm in favor of having this in std.

It seems that there's also a spec called NDJSON, which seems having a little bit more formal document https://github.com/ndjson/ndjson-spec, http://dataprotocols.org/ndjson/ Maybe better to also consider this as module name?

(Elasticsearch, for example, seems using ndjson for it https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

ayame113:
> The name is very long, maybe `generatorToStream`?

I think it's good 👍

kt3k:
> The parser can also parse the [json text sequence](https://datatracker.ietf.org/doc/html/rfc7464), passing options as arguments. My module also includes a [Concatenated JSON](https://en.wikipedia.org/wiki/JSON_streaming#Concatenated_JSON) parser.
[...]

PR #2227, opened 2022-05-13
feat(streams): toTransformStream

ayame113:
part of #2208

Introduce a `generatorToStream` that transforms the generator function into a TransformStream, as described in https://github.com/denoland/deno_std/issues/2208#issuecomment-1122175133.

```ts
import { readableStreamFromIterable, generatorToStream } from "https://deno.land/std@$STD_VERSION/streams/mod.ts";

const readable = readableStreamFromIterable([0, 1, 2])
  .pipeThrough(generatorToStream(async function* (src) {
    for await (const chunk of src) {
      yield chunk * 100;
    }
  }));

for await (const chunk of readable) {
  console.log(chunk);
}
// output: 0, 100, 200
```


crowlKats:
Havent looked into this closely yet, however, how does this differ from https://github.com/denoland/deno_std/blob/main/streams/conversion.ts#L174 (with the exception that this returns a transformstream instead of a readablestream)?

ayame113:
It's basically similar, but this helps users create custom TransformStream classes.

```ts
class MyCustumTransformStream implements TransformStream {
  readonly writable: WritableStream<number>;
  readonly readable: ReadableStream<number>;
  constructor() {
    const transformer = generatorToStream(this.#transformFunc);
    this.writable = transformer.writable;
    this.readable = transformer.readable;
  }
  async *#transformFunc(src: ReadableStream<number>) {
    for await (const chunk of src) {
      yield chunk * 100;
    }
  }
}
```

nayeemrmn:
I don't think the name `generatorToStream()` makes sense, generator is another name for the iterable iterator returned by the generator function.

crowlKats:
also on top of that, `ToStream` is too generic. it should be `ToTransformStream`

ayame113:
It is certainly so.
`transformStreamFromGeneratorFunction` is a long but accurate name, so is it better?

streams/conversion.ts
```diff
@@ -196,6 +196,78 @@ export function readableStreamFromIterable<T>(
   });
 }
 
+/**
+ * Convert the generator function into a TransformStream.
+ *
+ * ```ts
+ * import { readableStreamFromIterable, transformStreamFromGeneratorFunction } from "https://deno.land/std@$STD_VERSION/streams/mod.ts";
```
bartlomieju, line 203: This should probably use relative import like other APIs above

bartlomieju (commented):
The name of this API is a mouthful. Maybe we could use `transformStreamFromGenerator` instead?

streams/conversion.ts
```diff
@@ -196,6 +196,78 @@ export function readableStreamFromIterable<T>(
...
+        if (typeof iterator.throw == "function") {
+          try {
+            await iterator.throw(reason);
+          } catch {
+            /* `iterator.throw()` always throws on site. We catch it. */
+          }
+        }
+        await readable.cancel(reason);
+      },
+    }, readableStrategy),
```
kt3k, line 267: This part looks similar to `readableStreamFromIterable` except some error handlings. How about updating `readableStreamFromIterable` to this implementation and using it here?

ayame113: I think it is possible to take the common parts of the two functions and integrate them into a another internal function.
Is this what you meant?

```ts
export function readableStreamFromIterable<T>(
  iterable: Iterable<T> | AsyncIterable<T>,
): ReadableStream<T> {
  return internalReadableStreamFromIterable(iterable);
}

export function transformStreamFromGenerator<I, O>(
  transformer: (src: ReadableStream<I>) => Iterable<O> | AsyncIterable<O>,
  writableStrategy?: QueuingStrategy<I>,
  readableStrategy?: QueuingStrategy<O>,
): TransformStream<I, O> {
  const {
    writable,
    readable,
  } = new TransformStream<I, I>(undefined, writableStrategy);

  const iterable = transformer(readable);
  return {
    writable,
    readable: internalReadableStreamFromIterable(iterable, readableStrategy, readable),
  };
}

function internalReadableStreamFromIterable<T>(
  iterable: Iterable<T> | AsyncIterable<T>,
  readableStrategy?: QueuingStrategy<T>,
  upstream?: ReadableStream,
): ReadableStream<T> {
  // implementation
}
```

streams/conversion.ts
ayame113: Thank you for your review. Has been updated.

nayeemrmn:
> The name of this API is a mouthful. Maybe we could use `transformStreamFromGenerator` instead?

@bartlomieju 

> I don't think the name `generatorToStream()` makes sense, generator is another name for the iterable iterator returned by the generator function.

Honestly I think we should consider naming this `transformStream()` and renaming `readableStreamFromIterable()` to `readableStream()`. The thing they convert from is the basic JS representation of each (iterable for readable stream, (iterable -> iterable) for transform stream).

ayame113:
Regarding a function name, I would like to follow the decisions of the core team.

streams/conversion.ts
kt3k: I imagined like that, but now I see it's a little confusing (I missed `readable` dependency in `internalReadableStreamFromIterable` 😓  )

The current implementation looks just fine to me.

kt3k:
I'm in favor of @nayeemrmn's proposal (`transformStream()`). This utility seems useful for many cases. We can encourage the use of it by giving the short name.

What do you think? @crowlKats @bartlomieju

crowlKats:
overall I am in favour, though I worry that having a function with the same name as a class (with change of casing) can get quite confusing for people

kt3k:
> I worry that having a function with the same name as a class (with change of casing) can get quite confusing for people

Then how about `toTransformStream`?

crowlKats:
yea, that sounds fine to me

ayame113:
`toTransformStream` sounds good.👍 Has been updated.

kt3k (approved):
LGTM

We can still iterate on the name. Any further suggestions/opinions/discussions are welcome

bartlomieju (approved):

Issue #2134, opened 2022-04-21
recursive functions need to be checked for circular references


TomokiMiyauci:
**Describe the bug**
Recursive functions need to be checked for circular references.

I do not know about all modules, but the ones covered are:

- deepMerge

For `deepMerge`, there is a test, but It is not enough.

https://github.com/denoland/deno_std/blob/f4474d681790cc79c2d1142a1efcaa2aaba86991/collections/deep_merge_test.ts#L341-L345

**Steps to Reproduce**

```ts
import { deepMerge } from "https://deno.land/std@0.136.0/collections/deep_merge.ts";

const cycle: Record<PropertyKey, unknown> = {};
cycle[""] = cycle;

deepMerge(cycle, cycle); // break
```

**Expected behavior**

Not throw `RangeError`

**Environment**

- std version: 0.136.0

PR #2228, opened 2022-05-13
fix(collections): prevent cycles in deepMerge()

cjihrig:
This commit adds logic to `deepMerge()` to prevent infinite looping over cyclic data structures. The algorithm will now bail out before calling `mergeObjects()` if either object has already been processed.

Fixes: https://github.com/denoland/deno_std/issues/2134


bartlomieju (approved):
LGTM

Issue #2207, opened 2022-05-10
error conditional response in http/file_server.ts


masx200:
**Describe the bug**

error conditional response in http/file_server.ts

`If-None-Match: W/3ff580a0`

`etag: W/3ff580a0`

expect response status to be 304

**Steps to Reproduce**
start server
```
deno run -A https://deno.land/std@0.138.0/http/file_server.ts
```

request
```
GET /desktop.ini HTTP/1.1
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9
Accept-Encoding: gzip, deflate, br
Accept-Language: zh-CN,zh;q=0.9,en;q=0.8,en-GB;q=0.7,en-US;q=0.6
Cache-Control: max-age=0
Connection: keep-alive
DNT: 1
Host: localhost:4507
If-Modified-Since: Wed, 09 Feb 2022 11:23:53 GMT
If-None-Match: W/3ff580a0
Referer: http://localhost:4507/
Sec-Fetch-Dest: document
Sec-Fetch-Mode: navigate
Sec-Fetch-Site: same-origin
Sec-Fetch-User: ?1
Upgrade-Insecure-Requests: 1
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.4951.54 Safari/537.36 Edg/101.0.1210.39
sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="101", "Microsoft Edge";v="101"
sec-ch-ua-mobile: ?0
sec-ch-ua-platform: "Windows"
```
response
```
HTTP/1.1 200 OK
accept-ranges: bytes
etag: W/3ff580a0
content-type: text/plain
date: Tue, 10 May 2022 05:28:16 GMT
last-modified: Wed, 09 Feb 2022 11:23:53 GMT
server: deno
access-control-allow-origin: *
access-control-allow-headers: Origin, X-Requested-With, Content-Type, Accept, Range
vary: Accept-Encoding
content-encoding: gzip
transfer-encoding: chunked
```
**Expected behavior**

expect response status to be 304

**Environment**

v8 10.0.139.17
typescript 4.6.2

OS: Microsoft Windows [版本 10.0.19044.1682]
deno version:deno 1.21.2 (release, x86_64-pc-windows-msvc)
std version:0.138.0

[...]

PR #2229, opened 2022-05-14
http/file_server.ts  fix error conditional response 

masx200:
fix error conditional response 

https://github.com/denoland/deno_std/issues/2207

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -306,7 +306,7 @@ export async function serveFile(
     const ifNoneMatch = req.headers.get("if-none-match");
     const ifModifiedSince = req.headers.get("if-modified-since");
     if (
-      (ifNoneMatch && ifNoneMatch === simpleEtag) ||
+      (ifNoneMatch && ( ifNoneMatch === simpleEtag || 'W/' + ifNoneMatch === simpleEtag || ifNoneMatch === 'W/' +  simpleEtag ) ) ||
       (ifNoneMatch === null &&
         ifModifiedSince &&
         fileInfo.mtime.getTime() < new Date(ifModifiedSince).getTime() + 1000)


kt3k (commented):
@masx200 
Thanks for the fix! Can you also add a test case similar to the case `file_server returns 304 for requests with if-none-match set with the etag` but with `W/` prefixed etag in request headers.

kt3k:
and also can you run `deno fmt` to the changed source code?

http/file_server_test.ts
```diff
@@ -991,3 +991,94 @@ Deno.test(
...
+      assert(etag.length > 0);
+      assert(etag.startsWith("W/"));
+      {
+        const res = await fetch(
+          testurl,
+          {
+            headers: {
+              "if-none-match": etag,
+
+              "If-Modified-Since": last_modified,
```
kt3k, line 1067: I think there's no need to specify `If-Modified-Since` when etag is already set

http/file_server_test.ts
```diff
@@ -991,3 +991,94 @@ Deno.test(
...
+    try {
+      const testurl = "http://localhost:4507/testdata/desktop.ini";
+      const fileurl = new URL("./testdata/desktop.ini", import.meta.url);
+      let etag: string | undefined | null;
+      let last_modified: string | null | undefined;
+      {
+        const res = await fetch(
+          testurl,
+          {
+            headers: [
```
kt3k, line 1008: Most headers look irrelevant to the purpose of the test case. I guess we need only `Accept-Encoding` to get `W/` prefixed etag.

kt3k (approved):
Thanks for updating the test case!

LGTM

PR #2230, opened 2022-05-14
[WIP] feat: Add sqlite database driver

bartlomieju:
Upstreaming https://github.com/dyedgreen/deno-sqlite


iuioiua:
@bartlomieju, is this PR still in progress? IMO, the standard library shouldn't have any database modules, at least for now. They'd add too much managerial overhead for maintainers given there other higher priorities. Also, the maintainers of the existing modules appear to be handling their respective repos fairly well right now. Perhaps, this can be revisited in the future?

bartlomieju:
No, there's no more consensus on the team if we should do it or not. I'm gonna close it for now.

Issue #1302, opened 2021-09-22
assertRejects does not fail when an error is thrown, but a promise was not rejected


dsherret:
**Describe the bug**

```ts
import { assertRejects } from "https://raw.githubusercontent.com/denoland/deno_std/main/testing/asserts.ts";
await assertRejects(function () { throw new Error("test") });
```

**Expected behavior** Should throw because no promise was rejected.

Node's `assertRejects` behaviour:

```
> assert.rejects(function() { throw new Error("test"); }).catch(err => console.log(err));
> Error: test
    at REPL16:1:35
    at waitForActual (assert.js:786:21)
    at Function.rejects (assert.js:921:31)
    at REPL16:1:8
    at Script.runInThisContext (vm.js:134:12)
    at REPLServer.defaultEval (repl.js:486:29)
    at bound (domain.js:416:15)
    at REPLServer.runBound [as eval] (domain.js:427:12)
    at REPLServer.onLine (repl.js:819:10)
    at REPLServer.emit (events.js:387:35)
 ```

This might just be an issue with the renaming from `assertThrowsAsync` -> `assertRejects` and perhaps we should keep `assertThrowsAsync` then change `assertRejects` to assert that a promise is rejected?

kt3k:
hmm I'm not sure there's enough needs of the suggested version of `assertRejects`. I personally usually don't distinguish these 2 behaviors when writing async functions..

It feels a little too pedantic to me if we have both `assertReject` and `assertThrowsAsync` with the above difference.

cc @caspervonb 
What do you think about this suggestion?

caspervonb:
It does what it says on the tin:
> Executes a function which returns a promise, expecting it to throw or reject.

- Definitively not a bug, does what it is documented to do.
- assertThrowsAsync should never make a return imho, terrible name and the differences are too small.

> `await assertRejects(function () { throw new Error("test") });`

Thinking about this, I'm actually a bit surprised that this isn't a type-error.

I think **maybe** we want this to assert (both at compile time and runtime) that a promise was actually returned, you can always do `Promise.resolve` turn a value into a promise. Seems that is the real source of this category of errors, if the thing is a promise then everything is fine. But in this case it isn't and you can't really assert rejection on a non promise.

dsherret:
Here's showing the critical difference between the two and something that will pass type checking:

```ts
function getFunc(throwsSync: boolean) {
  return () => {
    if (throwsSync) {
      throw new Error();
    }

[...]

PR #2234, opened 2022-05-15
feat(assertRejects): fails on synchronous throw #1302

Sorikairox:
Based on issue #1302 

`assertRejects` now throw when the given function does not return a promise.

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -750,6 +750,7 @@ export async function assertRejects<E extends Error = Error>(
   }
   let doesThrow = false;
   let isPromiseReturned = false;
+  const msgToAppendToError = msg ? `: ${msg}` : ".";
   try {
     const possiblePromise = fn();
     if (possiblePromise instanceof Promise) {
@@ -757,6 +758,9 @@ export async function assertRejects<E extends Error = Error>(
       await possiblePromise;
     }
   } catch (error) {
+    if (!isPromiseReturned) {
+      throw new AssertionError(`Function throws when expected to reject${msgToAppendToError}`);
+    }
     if (error instanceof Error === false) {
       throw new AssertionError("A non-Error object was rejected.");
     }
@@ -771,12 +775,8 @@ export async function assertRejects<E extends Error = Error>(
     }
     doesThrow = true;
   }
-  msg = msg ? `: ${msg}` : ".";
-  if (!isPromiseReturned) {
-    throw new AssertionError(`Function throws when expected to reject${msg}`);
-  } else if (!doesThrow) {
-    msg = `Expected function to reject${msg}`;
-    throw new AssertionError(msg);
+  if (!doesThrow) {
+    throw new AssertionError(`Expected function to reject${msgToAppendToError}`);
   }
 }
 
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -757,9 +757,6 @@ export async function assertRejects<E extends Error = Error>(
       await possiblePromise;
     }
   } catch (error) {
-    if (!isPromiseReturned) {
-      throw new AssertionError("Function throws when expected to reject.");
-    }
     if (error instanceof Error === false) {
       throw new AssertionError("A non-Error object was rejected.");
     }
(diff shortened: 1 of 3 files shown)


testing/asserts.ts
```diff
@@ -749,9 +749,17 @@ export async function assertRejects<E extends Error = Error>(
...
   try {
-    await fn();
+    const possiblePromise = fn();
+    if (possiblePromise instanceof Promise) {
+      isPromiseReturned = true;
+      await possiblePromise;
+    }
   } catch (error) {
+    if (!isPromiseReturned) {
+      throw new AssertionError("An error was thrown but a promise was not rejected.");
```
KyleJune, line 761: For `assertError`, we return an error like this when an error is not thrown.
```ts
if (!doesThrow) {
  msg = `Expected function to throw${msg ? `: ${msg}` : "."}`;
}
```
I think we should have a similar message for when it was expected to reject. Maybe "Function throws when expected to reject".

Also the messages later in this assertRejects function should be updated.

- "A non-Error object was thrown or rejected." should be "A non-Error object was rejected."
- "Expected function to throw" should be "Expected function to reject"

KyleJune (changes requested):

testing/asserts.ts
```diff
@@ -749,11 +749,19 @@ export async function assertRejects<E extends Error = Error>(
...
+      isPromiseReturned = true;
+      await possiblePromise;
+    }
   } catch (error) {
+    if (!isPromiseReturned) {
+      throw new AssertionError("Function throws when expected to reject.");
+    }
     if (error instanceof Error === false) {
-      throw new AssertionError("A non-Error object was thrown or rejected.");
+      throw new AssertionError("A non-Error object was rejected.");
```
KyleJune, line 764: Only question left is whether or not these 2 errors should have msg appended to them like the "Expected function to reject" message at the end of this function. I don't have any strong opinions on that. If we do make that change, I think `assertThrows` AssertionError messaging should be updated too.

KyleJune (approved):

testing/asserts.ts
```diff
@@ -766,8 +771,11 @@ export async function assertRejects<E extends Error = Error>(
     }
     doesThrow = true;
   }
-  if (!doesThrow) {
-    msg = `Expected function to throw${msg ? `: ${msg}` : "."}`;
+  msg = msg ? `: ${msg}` : ".";
+  if (!isPromiseReturned) {
+    throw new AssertionError(`Function throws when expected to reject${msg}`);
+  } else if (!doesThrow) {
```
KyleJune, line 777: I think it would be better to have this where it was. If an error is thrown instead of rejected, we would incorrectly say "A non-Error object was rejected." instead of "Function throws when expected to reject" like we would before your most recent change.

testing/asserts.ts
```diff
@@ -766,8 +771,11 @@ export async function assertRejects<E extends Error = Error>(
     }
     doesThrow = true;
   }
-  if (!doesThrow) {
-    msg = `Expected function to throw${msg ? `: ${msg}` : "."}`;
+  msg = msg ? `: ${msg}` : ".";
```
KyleJune, line 774: I don't think we should be modifying the msg argument.

KyleJune (changes requested):

testing/asserts.ts
KyleJune: I'm fine with the change to add `: ${msg}` to the end of the message though when msg is set.

Sorikairox: I'll add a test case that reproduces this unwanted behavior and change the code accordingly then !

testing/asserts.ts
Sorikairox: It was already being modified, I only did that not to duplicate `msg ? `: ${msg}` : "."` and both error messages. But I agree that there is a better way to do that and keep the function as pure as possible 😄

KyleJune: Good point on it already being modified. Didn't notice that. I guess the modification part is fine. You can just move that line up to the top I guess, that way you can make use of it in both the catch and the last assertion at the end of the function.

Sorikairox: I removed the modification on my last push, would it be alright? If so I'll do it on `assertThrows` too, if not I'll revert this part.

KyleJune (approved):

kt3k (approved):
This is a breaking change, but I agree that this makes more sense.

LGTM

kt3k:
(Note: Only one assertion in std's test cases was broken by this change https://github.com/denoland/deno_std/pull/2234/commits/e3337d4df0bd57c0af96cd245c12328827f7623b

Sorikairox:
@kt3k Sorry for the one assertion I didn't see. Some other tests are failing on my MBP for something totally unrelated, so I missed the real error.
Sorry for forgetting to run `deno fmt`, I only ran `deno lint`, will think about it in my next contribution !

kt3k:
@Sorikairox No problem! Thanks for your contribution.

BTW I'll keep this open a little while to see other maintainers opinions

PR #2235, opened 2022-05-15
fix(testing/bdd): support using bdd tests with dnt

dsherret:
This should make it work on Node when using dnt (edit: yes, just tested it)

### DIFF
--- a/testing/_test_suite.ts
+++ b/testing/_test_suite.ts
@@ -124,37 +124,41 @@ export class TestSuiteInternal<T> implements TestSuite<T> {
         sanitizeOps,
         sanitizeResources,
         fn: async (t) => {
-          if (!TestSuiteInternal.running) TestSuiteInternal.running = true;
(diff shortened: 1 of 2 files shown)


testing/_test_suite.ts
```diff
@@ -124,37 +124,41 @@ export class TestSuiteInternal<T> implements TestSuite<T> {
...
-          const context = {} as T;
-          const { beforeAll } = this.describe;
-          if (typeof beforeAll === "function") {
-            await beforeAll.call(context);
-          } else if (beforeAll) {
-            for (const hook of beforeAll) {
-              await hook.call(context);
-            }
-          }
+          TestSuiteInternal.runningCount++;
```
dsherret, line 127: Change is to count how many tests are running instead of using a boolean.

bartlomieju (approved):
LGTM, it's great this will work with `dnt`!

Issue #2236, opened 2022-05-16
async/pool: `pooledMap` results are not in the order of iterable element argument


sntran:
**Describe the bug**

When `async/pool` [was introduced](https://github.com/denoland/deno/pull/6898), it was expected to be like `Promise.all(array.map(async () => {...}))`. However, while `Promise.all` returns values are in order of the Promises passed, the returned values of `pooledMap` are not in the same order of the argument iterable.

Also, the name `map`, at least to me, is always an 1 to 1 association.

**Steps to Reproduce**

`deno run` the following snippet:

```ts
import { pooledMap } from "https://deno.land/std@0.139.0/async/mod.ts";

function getRandomInt(min: number, max: number): number {
  min = Math.ceil(min);
  max = Math.floor(max);
  return Math.floor(Math.random() * (max - min) + min); //The maximum is exclusive and the minimum is inclusive
}

const results = pooledMap(
  2,
  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
  (i) => new Promise((r) => setTimeout(() => r(i), getRandomInt(5, 20) * 100)),
);

for await (const value of results) {
  console.log(value);
}
```

**Expected behavior**

The returned value should be `[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]`

**Actual behaviour**

The resulting array is out of order.

**Environment**

- OS: Mac 12.3
- deno version: 1.21.0
- std version: 0.139

Sorikairox:
I'm taking this, but if someone is already on it no problem (at least I would have learned how this work internally)

PR #2239, opened 2022-05-16
fix(pooledMap): return ordered result #2236

Sorikairox:
Based on issue #2236.
Pooled map result are now ordered correctly

### DIFF
--- a/async/pool.ts
+++ b/async/pool.ts
@@ -25,7 +25,14 @@ export function pooledMap<T, R>(
       p: Promise<R>,
       controller: TransformStreamDefaultController<R>,
     ) {
-      controller.enqueue(await p);
+      try {
+        const s = await p;
+        controller.enqueue(s);
+      } catch (e) {
+        if (e instanceof AggregateError) {
+          controller.error(e as unknown);
+        }
+      }
     },
   });
   // Start processing items from the iterator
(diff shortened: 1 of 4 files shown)


async/pool_test.ts
```diff
@@ -44,3 +44,24 @@ Deno.test("[async] pooledMap errors", async function () {
   });
   assertEquals(mappedNumbers, [3]);
 });
+
+Deno.test("pooledMap returns ordered items", async () => {
```
kt3k, line 48: nice test!

async/pool.ts
```diff
@@ -40,7 +47,7 @@ export function pooledMap<T, R>(
         // fail the race, taking us to the catch block where all currently
         // executing jobs are allowed to finish and all rejections among them
         // can be reported together.
-        p.then((v) => writer.write(Promise.resolve(v))).catch(() => {});
+        writer.write(p);
```
kt3k, line 50: What happens if p rejects with AggregateError?

kt3k, line 50: What happens if p rejects with AggregateError?

Sorikairox: Good point.

I think we should handle this with another Exception made specifically for this case, that could not be thrown in p?

However that would be a breaking change.

Maybe check AggregateError message in the condition too?

Sorikairox: @kt3k I improved the handling of `AggregateError` so `controller.error(e as unknown);` is called only when it is relevant.
No interference is `p` rejects with `AggregateError` anymore

kt3k: Thanks for updating! This looks much better 👍

kt3k (approved):
LGTM

Issue #2151, opened 2022-04-24
assertStrictEquals could use a better comparation


orl0:
**Is your feature request related to a problem? Please describe.**

`assertStrictEquals()` inherits the problem of strict equality operator (`===`), exactly the `===` operator treats the number values `-0` and `+0` as equal and `Number.NaN`/`NaN` as not equal.

This behavior doesn't seems for me to be expected when testing.

Expression `assertStrictEquals(func(), NaN)` shouldn't throw if `func` returns `NaN`

**Describe the solution you'd like**

Replace strict equality operator (`===`) with `Object.is` method [(MDN)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is).

like
```js
  if (actual === expected) {
    return;
  }
```
to
```js
  if (Object.is(actual, expected)) {
    return;
  }
```

cjihrig:
+1 to this. In my experience, this is generally the behavior that people want. This is also how Node's `assert` module works.

Alternatively, we could add new `assertIs()` and `assertIsNot()` if people wanted to leave `assertStrictEquals()` as is.

What do you think @bartlomieju? I'd be interested in working on this.

bartlomieju:
> +1 to this. In my experience, this is generally the behavior that people want. This is also how Node's `assert` module works.
> 
> Alternatively, we could add new `assertIs()` and `assertIsNot()` if people wanted to leave `assertStrictEquals()` as is.
> 
> What do you think @bartlomieju? I'd be interested in working on this.

I agree, sounds like a reasonable change.

PR #2244, opened 2022-05-18
BREAKING(testing): use Object.is() for strict equality

cjihrig:
This commit updates `assertStrictEquals()`, `assertAlmostEquals()`, and `assertNotStrictEquals()` to use `Object.is()` instead of the `===` operator. The reason is that `Object.is()` better matches user expectations of an assertion library in a couple edge cases. It also better aligns with the behavior of at least one other runtime's assertion library (Node.js).

Fixes: https://github.com/denoland/deno_std/issues/2151

### DIFF
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -1022,6 +1022,7 @@ Deno.test({
     assertStrictEquals(true, true);
     assertStrictEquals(10, 10);
     assertStrictEquals("abc", "abc");
+    assertStrictEquals(NaN, NaN);
 
     const xs = [1, false, "foo"];
     const ys = xs;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

testing/asserts_test.ts
```diff
@@ -1022,6 +1022,7 @@ Deno.test({
     assertStrictEquals(true, true);
     assertStrictEquals(10, 10);
     assertStrictEquals("abc", "abc");
+    assertStrictEquals(NaN, NaN);
```
kt3k, line 1025: 👍

kt3k (approved):
LGTM

PR #2247, opened 2022-05-18
breaking(flags): add collect option

### LINKED ISSUE #2246: Flags: Add collect option
This is a requirement for #1898 and has been discussed in #2180.

Currently all non boolean values are collactable. If we make the flags module type safe, we would always have to check if it's an array or not. It would be more usefull to make collectable options configurable.

Collect option:
```ts
{
  collect?: boolean | string | Array<string>;
}
```

Usage:
```ts
parse(Deno.args, {
  boolean: ["foo"],
  string: ["bar", "baz"],
  collect: ["foo", "bar"],
});
```

Result:
```ts
{
  foo: Array<boolean>;
  bar: Array<string>;
  baz?: string;
}
```

c4spar:
close #2246

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -190,7 +190,7 @@ export function parse(
     Object.keys(flags.bools).length > 0 ||
     Object.keys(flags.strings).length > 0;
 
-  if (typeof collect !== "undefined" && hasTypes) {
+  if (collect !== undefined) {
     if (typeof collect === "boolean") {
       flags.collectAll = collect;
     } else {
@@ -207,6 +207,8 @@ export function parse(
       }
     }
   }
+
+  const hasCollect = flags.collectAll || Object.keys(flags.collect).length > 0;
 
   const argv: Args = { _: [] };
 
@@ -235,11 +237,11 @@ export function parse(
     });
 
     const key = keys[keys.length - 1];
-    const collectable = collect && hasTypes &&
+    const collectable = collect &&
       (flags.collectAll || !!flags.collect[name]);
 
     if (
-      !collectable && (
+      hasCollect || hasTypes ? !collectable : (
         get(o, key) === undefined ||
         get(flags.bools, key) ||
         typeof get(o, key) === "boolean"
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -52,7 +52,10 @@ export interface ParseOptions {
   /** A string or array of strings argument names to always treat as strings. */
   string?: string | string[];
 
-  /** A string or array of strings argument names to always treat as arrays. */
+  /** A string or array of strings argument names to always treat as arrays.
+   * Collectable options can be used multiple times. All values will be
+   * colelcted into one array. If a non-collectable option is used multiple
+   * times, the last value is used. */
   collect?: string | string[];
 
   /** A function which is invoked with a command line parameter not defined in
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -240,13 +240,6 @@ export function parse(
         typeof get(o, key) === "boolean"
       )
     ) {
-      // if (hasTypes && get(o, key) !== undefined) {
-      //   throw new Error(
-      //     `Option "${
-      //       name.length > 1 ? "--" : "-"
-      //     }${name}" can be used only once.`,
-      //   );
-      // }
       o[key] = value;
     } else if (get(o, key) === undefined) {
       o[key] = [value];
@@ -261,7 +254,7 @@ export function parse(
     key: string,
     val: unknown,
     arg: string | undefined = undefined,
-    collect = true,
+    collect?: boolean,
   ): void {
     if (arg && flags.unknownFn && !argDefined(key, arg)) {
       if (flags.unknownFn(arg, key, val) === false) return;
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -228,7 +228,7 @@ export function parse(
     });
 
     const key = keys[keys.length - 1];
-    const collectable = collect && !!flags.collect[name];
+    const collectable = collect && !!get(flags.collect, name);
 
     if (!collectable) {
       o[key] = value;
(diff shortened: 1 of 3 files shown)


flags/mod.ts
```diff
@@ -199,12 +237,26 @@ export function parse(
...
+        typeof get(o, key) === "boolean"
+      )
     ) {
+      // if (hasTypes && get(o, key) !== undefined) {
+      //   throw new Error(
+      //     `Option "${
+      //       name.length > 1 ? "--" : "-"
+      //     }${name}" can be used only once.`,
+      //   );
+      // }
```
kt3k, line 256: please remove these lines

kt3k: Or should this be uncommented in the future?

flags/mod.ts
```diff
@@ -352,18 +400,31 @@ export function parse(
...
       if (aliases[key]) {
         for (const x of aliases[key]) {
-          setKey(argv, x.split("."), defaults[key]);
+          setKey(argv, x, value);
         }
       }
     }
   }
 
+  for (const [key] of Object.entries(flags.bools)) {
```
kt3k, line 415: nit: I think `const key of Object.keys(flags.bools)` makes more sense here.

Is it ok only to loop through `flags.bools`? What about the case when `flags.strings` are set and collectable and argv doesn't have key for it?

kt3k:
@bartlomieju
This change makes each key have either single value or multiple values depending on the configuration. This is going to be breaking change, especially for advanced users such as [denoflare](https://github.com/skymethod/denoflare), [scuttlesaurus](https://github.com/retog/scuttlesaurus/blob/52c2383375eab5b7852d1df1dad65d3e0dc41217/scuttlesaurus/main.ts), etc (BTW most users don't use this feature).

This makes the parsing behavior more explicit than implicit, and upgrading path is obvious. So I think this is worth trying. What do you think?

flags/mod.ts
```diff
@@ -52,6 +52,9 @@ export interface ParseOptions {
   /** A string or array of strings argument names to always treat as strings. */
   string?: string | string[];
 
+  /** A string or array of strings argument names to always treat as arrays. */
+  collect?: boolean | string | string[];
```
kt3k, line 56: I guess nobody would use `collect: true` as most cli tools tend to have `--help` or `--version` and they shouldn't be collectable. Do you think of any use case of `collect: true`?

flags/mod.ts
c4spar: It depends on how we want to handle duplicate options. You said you would prefer to throw an error for duplicate options. But currently we have tests to make sure that options can be overwritten.
If we want to keep the current behavior, we can remove those lines.
If we want to throw an error message for duplicate options, we can uncomment them.
If we want to make it configurable, I can add a todo to do that in a separate PR.

flags/mod.ts
c4spar: ah right, i changed the default value for collectible boolean args from `false` to `[]`. I would also need to set default values for collectible string args. is this fine for you or would you prefer not to set default values for collectible options?

flags/mod.ts
c4spar: yeah maybe it's not needed. i can remove it.

flags/mod.ts
kt3k: It looks fine to me to set default values to collectible options.

flags/mod.ts
kt3k: Current behavior (overwriting) makes enough sense to me. Let's try this as first pass.

kt3k:
Thanks for updating!

kt3k (approved):
LGTM

bartlomieju (approved):
LGTM

flags/test.ts
```diff
@@ -736,3 +736,94 @@ Deno.test("valueFollowingDoubleHyphenIsNotUnknown", function (): void {
...
+    "--foo",
+    "baz",
+    "--beep",
+    "boop",
+    "--bool",
+    "--bool",
+  ]);
+
+  assertEquals(argv, {
+    foo: ["bar", "baz"],
```
kt3k, line 753: @c4spar 
Ah, I realized we still collect values for arg when there's no string, boolean, collect options are specified. What do you think about making this case `"baz"`?

kt3k: Because we check `hasCollect || hasTypes` before checking `collectable` in `setKey`, setting any option name changes the parse result (see the example below):
```
> parse(["--foo", "bar", "--foo", "baz"])
{ _: [], foo: [ "bar", "baz" ] }
> parse(["--foo", "bar", "--foo", "baz"], { string: "x" })
{ _: [], foo: "baz" }
```
I think this behavior is a little hard to explain to the users. What do you think?

kt3k: How about changing `hasCollect || hasTypes ? !collectable : (...)` part in `setKey` to just `!collectable`

c4spar: Also fine for me. I just thought that we want to keep the old behavior when no options are set, so that the flags module can be still used without defining any options. But i can change it if you want.
Current behaviour is roughly described here: https://github.com/denoland/deno_std/blame/0d2ba0f223d993fd31a0d10193de769ffb4a8886/flags/README.md#L69-L83 (But there is still an open question about negatable options.)

kt3k: > keep the old behavior when no options are set

Changing the parse behavior based on such condition feels too complex to me.

> so that the flags module can be still used without defining any options.

I think flags module is still usable without any options even if we removed the old behavior. Do you mean such change is too breaking to existing users?

c4spar: > I think flags module is still usable without any options even if we removed the old behavior. Do you mean such change is too breaking to existing users?

i mean also with collectible args. but i think i agree, it can be confusing. i will remove it.

kt3k (approved):
@c4spar Thanks for updating. Now the implementation is much simpler and what `collect` does looks more straightforward to me. The docs also look great. Nice work!

LGTM

PR #2259, opened 2022-05-21
feat(http/util): add compareEtag

### LINKED ISSUE #2250: Add utility function for etag comparisons
https://github.com/denoland/deno_std/commit/3403116ca9d43d7b95c318ac0f367d74402d0aec adds support for handling weak etags in `serveFile`. It'd be nice if there was a utility function to do just the weak etag comparison:

```ts
/** Returns true if the etags match. Weak etag comparisons are handled. */
function compareEtag(a: string, b: string): bool;
```

kt3k:
closes #2250

### DIFF
--- a/http/util.ts
+++ b/http/util.ts
@@ -1,0 +1,14 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+/** Returns true if the etags match. Weak etag comparisons are handled. */
+export function compareEtag(a: string, b: string): boolean {
+  if (a === b) {
+    return true;
+  }
+  if (a.startsWith("W/") && !b.startsWith("W/")) {
+    return a.slice(2) === b;
+  }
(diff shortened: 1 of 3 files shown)


http/util.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+/** Returns true if the etags match. Weak etag comparisons are handled. */
+export function compareEtag(a: string, b: string): boolean {
```
bartlomieju, line 3: Maybe name this file headers.ts?

bartlomieju (approved):
LGTM

http/util.ts
kt3k: I prefer to keep it `util.ts` at this point as we don't have plan to provide exhaustive list of utilities about headers. Maybe let's reconsider it again when this module developed more.

Issue #2254, opened 2022-05-19
(node/crypto) TypeError: crypto.createHmac is not a function


jerry4718:
**Describe the bug**

**Steps to Reproduce**

**Expected behavior**

**Environment**

- OS: Deepin 20.5 x86_64
- deno version: 1.22.0
- std version: 0.140.0

kt3k:
Implemented in https://github.com/denoland/deno_std/pull/2664

Issue #981, opened 2021-06-21
Suggestion: Support `X509Certificate` within hash module


getspooky:
The new X509Certificate(buffer) is  an inbuilt constructor of class X509Certificate within hash module which is used to provide a **PEM** encoded **X509 Certificate**.

**Syntax:**

```ts
const cert = Deno.readFileSync("public-cert.pem");
new X509Certificate(cert);
// getting subject included in this certificate.
// by using x509.subject function
const value = x509.subject
// display the result
console.log("subject :- " + value)
```

**Output:**

```sh
subject :- C=IN
ST=Westbengal
L=Kolkata
O=Panco, Inc.
CN=Rohit Prasad
emailAddress=rofofof@gmail.com
```

**Specification**

- **x509.ca**: Type: `<boolean>` Will be true if this is a Certificate Authority (ca) certificate.
- **x509.checkEmail(email[, options])**: Checks whether the certificate matches the given email address.
- **x509.checkHost(name[, options])**: Checks whether the certificate matches the given host name.
- **x509.checkIP(ip[, options])**: Checks whether the certificate matches the given IP address (IPv4 or IPv6). 
- **x509.checkIssued(otherCert)**: Checks whether this certificate was issued by the given otherCert.
-**x509.subject**: The complete subject of this certificate.

Details: https://nodejs.org/docs/latest-v15.x/api/crypto.html#crypto_class_x509certificate

lino-levan:
@iuioiua Any opinions on this one?

iuioiua:
`crypto.X509Certificate` has been added to the Node compatibility layer. I think that might suffice. WDYT, @kt3k?

kt3k:
Right. The only purpose of this proposal looks like the compatibility to `crypto.X509Certificate` of Node.js. Let's close

PR #2263, opened 2022-05-23
feat(node/crypto): Stub out missing exports

cmorten:
Stub out the remainder of the `crypto` module with the correct interfaces, invoking `notImplemented()` if called.

With the inclusion of the browserify directory into `deno_std` I suspect there are some methods which could already be implemented, but to narrow scope here I have opted for using `notImplemented()` more often than not.

I have restructured the `crypto` implementation somewhat to align with the Node directory structure ( see #1707 ) but have left the `crypto_browserify` directory where it was as moving results in a ~250 file change which is probably best done as a chore in a separate PR to make for easier reviewing.

This structural PR should hopefully pave the way to make implementing the following easier:

- #2254
- #1331
- #981


node/internal/util/types.ts
```diff
@@ -22,7 +22,6 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import * as bindingTypes from "../../internal_binding/types.ts";
-export { isCryptoKey, isKeyObject } from "../crypto/keys.ts";
```
cmorten, line 25: Causes circular references. We can reference the crypto module directly if needed.

node/internal/crypto/_keys.ts
```diff
@@ -0,0 +1,15 @@
...
+  return (
+    obj != null && (obj as Record<symbol, unknown>)[kKeyType] !== undefined
+  );
+}
+
+export function isCryptoKey(obj: unknown): boolean {
+  return (
+    obj != null && (obj as Record<symbol, unknown>)[kKeyObject] !== undefined
+  );
+}
```
cmorten, line 15: This file avoids circular deps issues.

node/internal/crypto/cipher.ts
```diff
@@ -0,0 +1,285 @@
...
+
+export interface CipherGCMOptions extends TransformOptions {
+  authTagLength?: number | undefined;
+}
+
+export interface CipherOCBOptions extends TransformOptions {
+  authTagLength: number;
+}
+
+export interface Cipher extends TransformStream {
```
kt3k, line 48: I guess this should be Node.js `Transform` rather than global `TransformStream`. What do you think?

kt3k (commented):
@cmorten Thanks for doing this! Left one comment

node/internal/crypto/cipher.ts
cmorten: You’re spot on! https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/crypto.d.ts#L850

Will touch up later once back at laptop

cmorten: [85e8a6a](https://github.com/denoland/deno_std/pull/2263/commits/85e8a6a2a2dfe22a777bb138092471ad5ec33b1a)

kt3k (approved):
LGTM. Thanks!

PR #2265, opened 2022-05-24
feat: unsigned LEB128 varint encoding

MierenManz:
As discussed with Kt3k on discord

[convo](https://discord.com/channels/684898665143206084/775393009981849600/978615168386887750)


encoding/varint.ts
```diff
@@ -0,0 +1,53 @@
...
+ * console.log(decodedValue === 25565);
+ * ```
+ */
+export function decodeU32(val: Uint8Array): number {
+  return wasm.decode_u32(val);
+}
+
+/**
+ * ```ts
+ * import { decodeU32 } from "./varint.ts";
```
kt3k, line 43: This example seems using the wrong API

```suggestion
 * import { decodeU64 } from "./varint.ts";
```

encoding/varint.ts
```diff
@@ -0,0 +1,53 @@
+import * as wasm from "../_wasm_varint/varint.mjs";
+
+/**
+ * ```ts
```
kt3k, line 4: Can you also add an API description?

Maybe something like:
```md
Encodes the given number into Uint8Array with LEB128. The number needs to be in the range of `0` and `0xffffffff`.
```

encoding/varint_test.ts
```diff
@@ -0,0 +1,97 @@
...
+    assertEquals(encodeU32(127), new Uint8Array([127]));
+    assertEquals(encodeU32(128), new Uint8Array([128, 1]));
+    assertEquals(encodeU32(255), new Uint8Array([255, 1]));
+    assertEquals(encodeU32(25565), new Uint8Array([221, 199, 1]));
+    assertEquals(encodeU32(2097151), new Uint8Array([255, 255, 127]));
+    assertEquals(
+      encodeU32(2147483647),
+      new Uint8Array([255, 255, 255, 255, 7]),
+    );
+    assertEquals(encodeU32(-1), new Uint8Array([255, 255, 255, 255, 15]));
```
kt3k, line 17: Are negative values considered valid input?

kt3k:
When I pass invalid array like `decodeU32(new Uint8Array([255, 255]))`, it throws with `RuntimeError: unreachable`. I think this should be a RangeError or TypeError. Also the message should mention about the encoding being wrong. What do you think?

encoding/varint_test.ts
MierenManz: No, gonna add a range check for that

MierenManz:
> I think this should be a RangeError or TypeError.
I think RangeError would be the best here. Gonna add these changes.
Thank you for the feedback

encoding/varint.ts
```diff
@@ -0,0 +1,75 @@
...
+ * import { encodeU32 } from "./varint.ts";
+ *
+ * const encodedValue = encodeU32(42);
+ * // Do something with the encoded value
+ * ```
+ */
+export function encodeU32(val: number): Uint8Array {
+  if (!Number.isInteger(val)) throw new TypeError("Floats are not supported");
+  if (val < 0) throw new RangeError("Signed integers are not supported");
+  if (val > U32MAX) throw new RangeError();
```
kt3k, line 18: I think we should give more information in the error message like:

```ts
throw new RangeError(`The given number exceeds the limit of unsigned integer: ${val}`);
```

encoding/varint.ts
```diff
@@ -0,0 +1,75 @@
...
+ * // Do something with the decoded value
+ * console.log(decodedValue === 25565n);
+ * ```
+ */
+export function decodeU64(val: Uint8Array): BigInt {
+  if (val.length > 10) throw RangeError("Too many bytes");
+  try {
+    return wasm.decode_u64(val);
+  } catch {
+    throw new RangeError("Bad varint");
```
kt3k, line 73:
```suggestion
    throw new RangeError(`Bad varint: ${val}`);
```

encoding/varint.ts
```diff
@@ -0,0 +1,75 @@
...
+ * // Do something with the decoded value
+ * console.log(decodedValue === 25565);
+ * ```
+ */
+export function decodeU32(val: Uint8Array): number {
+  if (val.length > 5) throw RangeError("Too many bytes");
+  try {
+    return wasm.decode_u32(val);
+  } catch {
+    throw new RangeError("Bad varint");
```
kt3k, line 53:
```suggestion
    throw new RangeError(`Bad varint: ${val}`);
```

encoding/varint.ts
```diff
@@ -0,0 +1,75 @@
...
+ * ```ts
+ * import { encodeU64 } from "./varint.ts";
+ *
+ * const encodedValue = encodeU64(42n);
+ * // Do something with the encoded value
+ * ```
+ */
+export function encodeU64(val: bigint): Uint8Array {
+  if (val < 0) throw new RangeError("Signed integers are not supported");
+  if (val > U64MAX) throw new RangeError();
```
kt3k, line 33:
```suggestion
  if (val > U64MAX) throw new RangeError(`The given number exceeds the limit of unsigned long integer: ${val}`);
```

kt3k (commented):
Looks much better on invalid input handling. Left a few suggestions for error messages.

kt3k (approved):
LGTM!

kt3k:
@MierenManz Thanks for your contribution!

PR #2267, opened 2022-05-24
 feat(streams/delimiter): support `\r` in `TextLineStream`

crowlKats:


streams/delimiter.ts
```diff
@@ -106,32 +109,52 @@ export class TextLineStream extends TransformStream<string, string> {
...
+    } else { // \n and \r
+      if ((lfIndex - 1) === crIndex) { // \r\n
+        this.#mergeHandle(chunk, crIndex, lfIndex, controller);
+      } else {
+        if (crIndex < lfIndex) { // \r first
+          this.#mergeHandle(chunk, crIndex, crIndex, controller);
+        } else { // \n first
+          this.#mergeHandle(chunk, lfIndex, lfIndex, controller);
+        }
       }
```
kt3k, line 143: nit: very small point, but these can be a little less nested like:
```ts
if ((lfIndex - 1) === crIndex) { // \r\n
  // ...
} else if (crIndex < lfIndex) { // \r first
  // ...
} else { // \n first
  // ...
}
```

kt3k (approved):
LGTM

jsejcksn:
@crowlKats There is no description for this PR and it's not linked to an issue. Can you share why the change was implemented? Are there any modern environments which treat solo Carriage Returns (without being followed by a Line Feed) as [newline](https://en.wikipedia.org/wiki/Newline) delimiters? Should there be a new option to control this behavior?

crowlKats:
Because I needed it for something else, so I just upstreamed it. Various specifications for some encodings still support \r. Adding an option to enable it would be possible, but is it necessary? Doubtful.

jsejcksn:
> but is it necessary? Doubtful.

It is a breaking change (which I discovered because it broke my code). Moving it behind an option seems rational.

PR #2276, opened 2022-05-27
feat(mime): add mime/content_type module

kt3k:
This PR adds `contentType` method in `std/mime`. The design is inspired by `TypeByExtension` function from Go's mime package

refs:
- Go's mime package https://pkg.go.dev/mime
- past discussion about previous `std/media_types` module https://github.com/denoland/deno/issues/4292

### DIFF
--- a/mime/content_type.ts
+++ b/mime/content_type.ts
@@ -1,0 +1,53 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// Copyright (c) 2014 Jonathan Ong <me@jongleberry.com>
+// Copyright (c) 2015 Douglas Christopher Wilson <doug@somethingdoug.com>
+// Copyright (c) 2020-2022 the oak authors
+// mime_db is copied from https://raw.githubusercontent.com/jshttp/mime-db/v1.52.0/db.json
+
+import db from "./mime_db.json" assert { type: "json" };
+
+const EXTRACT_TYPE_REGEXP = /^\s*([^;\s]*)(?:;|\s|$)/;
+const TEXT_TYPE_REGEXP = /^text\//i;
+
+const extensionToType = new Map<string, string>();
+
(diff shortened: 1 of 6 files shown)


mime/content_type.ts
```diff
@@ -0,0 +1,53 @@
...
+
+const EXTRACT_TYPE_REGEXP = /^\s*([^;\s]*)(?:;|\s|$)/;
+const TEXT_TYPE_REGEXP = /^text\//i;
+
+const extensionToType = new Map<string, string>();
+
+for (const [name, type] of Object.entries(db)) {
+  if ("extensions" in type) {
+    let mimeType = name;
+    const cs = charset(mimeType, "charset" in type ? type.charset : undefined);
```
cjihrig, line 17: Can `"charset" in type ? type.charset : undefined` be replaced with just `type.charset`, or does TypeScript complain?

cjihrig (approved):

mime/content_type.ts
kt3k: Yup, typescript complain about it. It's typed as union of various shaped objects. Some have `charset` and some don't.

スクリーンショット 2022-05-30 19 08 18

bartlomieju:
Looks like this PR conflicts with https://github.com/denoland/deno_std/pull/2286

lucacasonato:
> Looks like this PR conflicts with https://github.com/denoland/deno_std/pull/2286

I would rather say #2286 conflicts with this PR, as this one came first :)

kitsonk:
> > Looks like this PR conflicts with #2286
> 
> I would rather say #2286 conflicts with this PR, as this one came first :)

Though I suggested in multiple channels, as maintained of `media_types` that I would work to re-integrate it back into `std`. 😕

kt3k:
closing in favor of #2286. #2286 is (roughly) a superset of this. So we don't need this anymore.

Issue #2279, opened 2022-05-28
doc: async docs needlessly avoid top-level await


matthewleon:
**Describe the bug**

The documentation of `tee` from async (https://github.com/denoland/deno_std/blob/main/async/README.md) needlessly avoids top-level `await`. It wraps invocation in an immediately-invoked `async` function. This appears to be redundant.

**Expected behavior**

The docs should show use of `for await` at the top level.

PR #2280, opened 2022-05-28
docs(async): use top-level await in tee docs

matthewleon:
Eliminate redundant function wrapping in documentation of async's tee method. Resolves https://github.com/denoland/deno_std/issues/2279


kt3k (approved):
LGTM. Thanks!

Issue #2283, opened 2022-05-28
tests: async pool_test uses deprecated assertRejects


matthewleon:
**Describe the bug**

`pool_test.ts` uses the deprecated, non-generic, callback-accepting version of `assertRejects`: https://github.com/denoland/deno_std/blob/6dcb19b0037d9d4e1e9b08a95d20e0d3da6fa026/async/pool_test.ts#L37

**Steps to Reproduce**

Open the file in IntelliJ with the deno plugin, and you'll be issued a deprecation warning, with the method in the editor being shown in ~~strikethrough~~ style.

**Expected behavior**

No deprecation.

**Environment**

deno 1.22.1
- OS: MacOS 12.4
- deno version: 1.22.1
- std version: git https://github.com/denoland/deno_std/commit/87883364abe91d0763cf030feeb93b973965b2fe

matthewleon:
related: https://github.com/denoland/deno_std/pull/2282

PR #2284, opened 2022-05-28
tests(async): use generic assertRejects in pool_test.ts

matthewleon:
Fixes a deprecation warning. Resolves https://github.com/denoland/deno_std/issues/2283


matthewleon:
Sorry, this was not ready.

kt3k:
Thanks for trying this. I think we need to pass `AggregateError` as the 2nd arg of assertRejects and use the rejected value as further assertion

```ts
const error = await assertRejects(fn, AggregateError, ERROR_WHILE_MAPPING_MESSAGE);
// ... assert about `error.errors

PR #2286, opened 2022-05-30
feat: add media_types

kitsonk:
Once upon a time [deno.land/x/media_types](https://deno.land/x/media_types) was inside `std` but it was removed due to there being a debate about the API and I took over maintaining it.

The PR intends to reverse this and will result in a series of PRs brining some of the common APIs common to handling HTTP requests into the std library.

@lucacasonato this includes the GoLang `mime` package APIs as well as some parts of the `jshttp/mime-types` APIs that `deno.land/x/media_types` was based on.  It should provide an easy migration path for users of `deno.land/x/media_types` but also provide more logical APIs aligned to what is in GoLang.

Once this is landed, a subsequent PR should be created so the fileserver uses this instead of it partial mime db.  I will also refactor oak and my other projects to use this as well.

### DIFF
--- a/media_types/mod.ts
+++ b/media_types/mod.ts
@@ -9,9 +9,7 @@
  * @module
  */
 
-import db from "https://raw.githubusercontent.com/jshttp/mime-db/v1.52.0/db.json" assert {
-  type: "json",
-};
+import db from "./vendor/mime-db.v1.52.0.json" assert { type: "json" };
 import { extname } from "../path/mod.ts";
 import {
   consumeMediaParam,
--- a/media_types/mod.ts
+++ b/media_types/mod.ts
@@ -10,7 +10,6 @@
  */
 
 import db from "./vendor/mime-db.v1.52.0.json" assert { type: "json" };
-import { extname } from "../path/mod.ts";
 import {
   consumeMediaParam,
   decode2331Encoding,
@@ -72,6 +71,10 @@ export const types = new Map<string, KeyOfDb>();
 /** Given an extension or media type, return a full `Content-Type` or
  * `Content-Disposition` header value.
  *
+ * The function will treat the `extensionOrType` as a media type when it
+ * contains a `/`, otherwise it will process it as an extension, with or without
+ * the leading `.`.
+ *
  * Returns `undefined` if unable to resolve the media type.
  *
  * ### Examples
@@ -81,8 +84,10 @@ export const types = new Map<string, KeyOfDb>();
  *
  * contentType(".json"); // `application/json; charset=UTF-8`
  * contentType("text/html"); // `text/html; charset=UTF-8`
+ * contentType("text/html; charset=UTF-8"); // `text/html; charset=UTF-8`
  * contentType("txt"); // `text/plain; charset=UTF-8`
  * contentType("foo"); // undefined
+ * contentType("file.json"); // undefined
  * ```
  */
 export function contentType(extensionOrType: string): string | undefined {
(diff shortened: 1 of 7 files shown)


bartlomieju (approved):
LGTM!

kt3k:
FYI Luca and I worked together on a similar API last week, which resulted in https://github.com/denoland/deno_std/pull/2276 (which only suggests `contentType` API in `std/mime`

media_types/mod.ts
```diff
@@ -0,0 +1,407 @@
...
+
+/** Provides utility functions for media types.
+ *
+ * This API is inspired by the GoLang [`mime`](https://pkg.go.dev/mime) package
+ * and [jshttp/mime-types](https://github.com/jshttp/mime-types).
+ *
+ * @module
+ */
+
+import db from "https://raw.githubusercontent.com/jshttp/mime-db/v1.52.0/db.json" assert {
```
kt3k, line 12: I think we should copy this file into std, and should avoid the 3rd party dependency

media_types/mod.ts
```diff
@@ -0,0 +1,407 @@
...
+        }
+      }
+
+      types.set(ext, type);
+    }
+  }
+})();
+
+/** Given an extension or media type, return a full `Content-Type` or
+ * `Content-Disposition` header value.
```
kt3k, line 75: Accepting both extension and media type feels strange and confusing to me. What is the benefit of accepting both? How about splitting the API or dropping the media type as input? (Currently what we need in file_server implementation is only `extension -> content type` function)

kitsonk: A lot of code based on `media_type` is written to accept both.  For example with oak, you can set the response body type to `html`/`json`/`text/html`/`application/json` to have the `Content-Type` set correctly.  This gives a path for less breaking changes as well as a migration path for people already consuming an API.

Having it clearly documented I think avoids any confusion.

media_types/mod.ts
kitsonk: Done...

media_types/mod.ts
kt3k: Ok. Makes sense to me.

kt3k:
Question about the module name: Should we call this `std/media_types` or `std/mime`?

Golang, for example, name the module [`mime`](https://pkg.go.dev/mime) instead of `media_types`. Also MDN seems calling it MIME types more times than media types ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types (also the page title and URL has `MIME` instead of media type

kitsonk:
> Question about the module name: Should we call this `std/media_types` or `std/mime`?
> 
> Golang, for example, name the module [`mime`](https://pkg.go.dev/mime) instead of `media_types`. Also MDN seems calling it MIME types more times than media types ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types (also the page title and URL has `MIME` instead of media type

MIME is an antiquated name for media types. When it was originally part of `std`, it was called `media_types` and moved out to `deno.land/x` as `media_types`, so my feeling is it should come back as `media_types`, especially because we want people to move back from `deno.land/x/media_types` to this.

While it is less common it is the proper name for them (as there is no mail interchange going on here).  From the page you linked to:

> A media type (also known as a Multipurpose Internet Mail Extensions or MIME type) indicates the nature and format of a document, file, or assortment of bytes. MIME types are defined and standardized in IETF's [RFC 6838](https://datatracker.ietf.org/doc/html/rfc6838).

And the [standard they are based on says](https://datatracker.ietf.org/doc/html/rfc6838):

> # Media Type Specifications and Registration Procedures
>
> This document defines procedures for the specification and registration of media types for use in HTTP, MIME, and other Internet protocols.

So _MIME_ is an extension to the e-mail protocol, a _MIME type_ is when you use a media type with MIME and the `Content-Type` and `Content-Disposition` headers of HTTP use a media type.

kt3k:
> While it is less common it is the proper name for them (as there is no mail interchange going on here)

> So MIME is an extension to the e-mail protocol, a MIME type is when you use a media type with MIME and the Content-Type and Content-Disposition headers of HTTP use a media type.

Sounds fair enough to me

kt3k (approved):
LGTM

PR #2287, opened 2022-05-30
tests(async): use generic assertRejects in pool_test.ts

matthewleon:
Fixes a deprecation warning. Resolves https://github.com/denoland/deno_std/issues/2283

This is a follow-up of https://github.com/denoland/deno_std/pull/2284

@kt3k I've basically followed your suggestion there, with the slight modification that instead of sending `ERROR_WHILE_MAPPING_MESSAGE` as a parameter to `assertRejects`, I maintain the strict equality check. It looks like `assertRejects` checks for inclusion of the passed-in message, rather than strict equality.

### DIFF
--- a/async/pool_test.ts
+++ b/async/pool_test.ts
@@ -34,17 +34,16 @@ Deno.test("[async] pooledMap errors", async () => {
     return n;
   }
   const mappedNumbers: number[] = [];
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM. Thanks!

PR #2288, opened 2022-05-30
feat(node/assert): add deepEqual and notDeepEqual 

tjosepo:
Those were skipped because they don't have test cases in `test-assert.js`.

I used the implementation of `isDeepEqual()` from [`./node/internal/utils/comparisons.ts`](https://github.com/denoland/deno_std/blob/main/node/internal/util/comparisons.ts), which I believe to be a direct copy of the code in Node.

These functions missing is what prevents [Recast](https://github.com/benjamn/recast) from working on Deno using esm.sh.

It's my first time contributing to the Node compat modules, so let me know if I missed anything 😊

### DIFF
--- a/node/assert.ts
+++ b/node/assert.ts
@@ -15,6 +15,24 @@ import {
 } from "./internal/errors.ts";
 import { isDeepEqual } from "./internal/util/comparisons.ts";
 
+function innerFail(obj: {
+  actual?: unknown;
+  expected?: unknown;
+  message?: string | Error;
+  operator?: string;
+}) {
+  if (obj.message instanceof Error) {
+    throw obj.message;
+  }
+
+  throw new AssertionError({
+    actual: obj.actual,
+    expected: obj.expected,
+    message: obj.message,
+    operator: obj.operator,
+  });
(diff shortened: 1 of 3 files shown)


node/assert.ts
```diff
@@ -317,21 +318,63 @@ function notStrictEqual(
...
+    throw message;
+  }
+
+  throw createAssertionError({
+    actual,
+    expected,
+    message,
+    operator: "notDeepEqual",
+    generatedMessage: true,
+  });
```
cjihrig, line 377: Is it possible to avoid all of this duplication. Look at the `innerFail()` function in Node's `assert` library as an example.

cjihrig (commented):
The changes LGTM, but it would be nice to cut down on some code duplication.

cjihrig (commented):
The changes LGTM, but it would be nice to cut down on some code duplication.

tjosepo:
Thank you for the review!

I added the `innerFail()` function.

cjihrig (approved):
LGTM. Thanks for the contribution.

PR #2290, opened 2022-05-31
combine files

timreichen:
The csv API is split over multiple files. Combine `csv.ts` and `csv_stringify.ts` files into one file as well as their test files as it is with all other encoding/* files.


timreichen:
Once https://github.com/denoland/deno_std/pull/2420 lands, both `parse` and `stringify` will be browser compatible, therefore possible to merge the files.

PR #2292, opened 2022-05-31
CI: Add License header check

### LINKED ISSUE #1415: Check license headers in CI

MierenManz:
closes #1415
I am not sure if this pr should also fix all the missing headers.


MierenManz:
I'll get on that then and make CI green :D

Then this should be able to land

MierenManz:
Seems like the script is failing on some code that isn't in my local copy of the repo.
Which is `node/internal/crypto/_keys.ts`. Would it be fine to add that to the exclude list?

bartlomieju:
> Seems like the script is failing on some code that isn't in my local copy of the repo. Which is `node/internal/crypto/_keys.ts`. Would it be fine to add that to the exclude list?

That's strange, it's definitely part of `deno_std` that is not a submodule. I can confirm that that file doesn't have a copyright header

kt3k:
@MierenManz Nice! Thanks!

kt3k:
> Seems like the script is failing on some code that isn't in my local copy of the repo.


`actions/checkout` checks out merge head instead of branch itself, and this branch was a little out-dated from main branch

スクリーンショット 2022-06-01 16 13 18

bartlomieju (approved):
LGTM too

PR #2293, opened 2022-06-01
chore(http/file_server.ts): export interface for serveDir/serveFile

ayame113:
I accessed https://doc.deno.land/https://deno.land/std@0.141.0/http/file_server.ts and noticed that there was no description of how to specify the options.

- Export some interfaces according to the [style guide](https://deno.land/manual/contributing/style_guide#export-all-interfaces-that-are-used-as-parameters-to-an-exported-member). ("Export all interfaces that are used as parameters to an exported member")
- Add a description to JSDoc.


kt3k (approved):
LGTM!

PR #2298, opened 2022-06-02
refactor(http): file_server uses std/media_types

kitsonk:
This PR refactors http/file_server to use the media_types module from std.

It also removes some dead code that was not being referenced in the file.


bartlomieju (approved):
LGTM

PR #2301, opened 2022-06-02
BREAKING(flags): introduce negatable option

### LINKED ISSUE #2300: Flags: Make negatable options configurable

We discussed in #2180 that we want to make negatable options configurable, because it would improve the typings we want to introduce in #2180.

We agreed for now on following solution:

```ts
parse(Deno.args, {
  string: ["config"],
  negatable: ["config"],
});
```

An alternative solution could be to prefix the option name with `no-`.

```ts
parse(Deno.args, {
  string: ["config", "no-config"],
});
```

c4spar:
close #2300

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -58,6 +58,10 @@ export interface ParseOptions {
    * times, the last value is used. */
   collect?: string | string[];
 
+  /** A string or array of strings argument names which can be negated
+   * by prefixing them with `--no-`, like `--no-config`. */
+  negatable?: string | string[];
+
   /** A function which is invoked with a command line parameter not defined in
    * the `options` configuration object. If the function returns `false`, the
    * unknown option is not added to `parsedArgs`. */
@@ -68,6 +72,7 @@ interface Flags {
   bools: Record<string, boolean>;
   strings: Record<string, boolean>;
   collect: Record<string, boolean>;
+  negatable: Record<string, boolean>;
   unknownFn: (arg: string, key?: string, value?: unknown) => unknown;
   allBools: boolean;
 }
@@ -135,6 +140,7 @@ export function parse(
     stopEarly = false,
     string = [],
     collect = [],
+    negatable = [],
     unknown = (i: string): unknown => i,
   }: ParseOptions = {},
 ): Args {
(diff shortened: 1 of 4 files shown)


flags/test.ts
```diff
@@ -197,37 +197,57 @@ Deno.test("booleanParsingTrueLike", function (): void {
...
+  const parsed2 = parse(["--foo", "--no-foo", "123"], {
+    boolean: ["foo"],
+    negatable: ["foo"],
+  });
   assertEquals(parsed2.foo, false);
 });
 
 Deno.test("booleanAfterBooleanNegation", function (): void {
-  const parsed = parse(["--no--foo", "--foo"], { boolean: ["foo"] });
+  const parsed = parse(["--no--foo", "--foo"], {
```
kt3k, line 214: nit: Not an error of this change, but I think this `"--no--foo"` should be `"--no-foo"`.

`parsed` here looks like `{ _: [], "no--foo": true, foo: true }`, and that doesn't check what the test name suggests.

kt3k (commented):
Mostly looks good. Nice work!

Left one comment

flags/mod.ts
```diff
@@ -201,6 +208,22 @@ export function parse(
     }
   }
 
+  if (negatable !== undefined) {
+    const negatableArgs = typeof negatable === "string"
```
cjihrig, line 212: Does it make sense to perform some input validation here? What if `negatable` is not a string, but also not an array? Or, what if it is an array that includes things besides strings. Or, what if the provided values are strings, but nonsense values? It feels like we should either have validation, tests, or both for these edge cases.

flags/test.ts
c4spar: fixed

kt3k (approved):
@c4spar Thanks for updating the test cases. LGTM.

kt3k:
Note: This is a breaking change when the user's CLI tool accepting `--no-xyz`-form flags. In that case the user's CLI tool needs to specify that arg name in `negatable` option (So the workaround is obvious and doesn't require lot of changes to their code)

The main motivation of introducing this breaking change is to make the typing (in #2180) more useful and reasonable. Without this change, all args can be `false` even when the arg is specified as `string`. This PR changes it, and the API now accepts `--no-xyz` as negation of `--xyz` only when `xyz` is specified in `negatable` option.

PR #2302, opened 2022-06-02
feat(http): add content negotiation

kitsonk:
This ports the content negotiation APIs from `oak_commons` into `std/http`


http/_negotiation/mediaType.ts
```diff
@@ -0,0 +1,237 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
```
kt3k, line 1: I think the filename should be in snake_case

http/negotiation.ts
```diff
@@ -0,0 +1,91 @@
...
+/** Returns an array of media types accepted by the request, in order of
+ * preference. If there are no media types supplied in the request, then any
+ * media type selector will be returned. */
+export function accepts(request: Request): string[];
+/** For a given set of media types, return the best match accepted in the
+ * request. If no media type matches, then the function returns `undefined`. */
+export function accepts(
+  request: Request,
+  ...types: string[]
+): string | undefined;
```
kt3k, line 29: Another design possibility is that taking `string | null` as the 1st arg and makes the usage `accepts(req.headers.get("accept"))`, which is more primitive and more composable in my view.

What do you think?

cc @bartlomieju @lucacasonato 

(Note: In the case of `http/cookie`, we switched from `getCookies(req)` to `getCookies(req.headers)` in the past https://github.com/denoland/deno_std/pull/1041 )

kitsonk: The primitives are there, they are hidden away in `_negotiation`.

The reasoning for #1041 makes more sense, because otherwise you have to have a request already created, when in fact you only want to deal with the headers on cookies because you may not yet have build the response and making it `Request` and `Headers` would just be confusing.

In this case you will _always_ be dealing with a request.  Not the types are not requiring a DOM `Request` but specifically an object which has a `headers` property which has a `get()` method.  It is as narrow of a type to be applicable in as many use cases.

kt3k: > In this case you will always be dealing with a request.

I agree with this. This design looks reasonable to me.

kt3k (approved):
LGTM

PR #2303, opened 2022-06-03
feat(crypto): add KeyStack for rotatable key data signing

kitsonk:
This adds a utility class which is a pre-cursor to being able to provide a cookie interface that can transparently sign and validate client cookies to prevent cookie tampering, which is a best practice we should be promoting.  It is currently part of `oak_commons` that I am contributing back to `std`.

### DIFF
--- a/crypto/keystack.ts
+++ b/crypto/keystack.ts
@@ -32,22 +32,6 @@ export interface KeyRing {
 
 const encoder = new TextEncoder();
 
-function assert(value: unknown, message = "Assertion error"): asserts value {
-  if (!value) {
-    throw new Error(message);
-  }
-}
-
-const REPLACEMENTS: Record<string, string> = {
-  "/": "_",
-  "+": "-",
-  "=": "",
-};
-
-function encodeBase64Safe(data: string | ArrayBuffer): string {
-  return base64.encode(data).replace(/\/|\+|=/g, (c) => REPLACEMENTS[c]);
-}
-
 function importKey(key: Key): Promise<CryptoKey> {
   if (typeof key === "string") {
     key = encoder.encode(key);
--- a/crypto/keystack.ts
+++ b/crypto/keystack.ts
@@ -8,27 +8,14 @@
  * @module
  */
 
-import * as base64 from "../encoding/base64.ts";
+import { timingSafeEqual } from "./timing_safe_equal.ts";
+import * as base64url from "../encoding/base64url.ts";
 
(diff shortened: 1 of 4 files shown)


crypto/keystack.ts
```diff
@@ -0,0 +1,212 @@
...
+  /** Sign the data, returning a string based digest of the data. */
+  sign(data: Data): Promise<string> | string;
+  /** Verifies the digest matches the provided data, indicating the data was
+   * signed by the keyring and has not been tampered with. */
+  verify(data: Data, digest: string): Promise<boolean> | boolean;
+}
+
+const encoder = new TextEncoder();
+
+function assert(value: unknown, message = "Assertion error"): asserts value {
```
kt3k, line 35: How about using one in `std/_util/assert.ts`?

crypto/keystack.ts
```diff
@@ -0,0 +1,212 @@
...
+
+/** Types of data that can be signed cryptographically. */
+export type Data = string | number[] | ArrayBuffer | Uint8Array;
+
+/** Types of keys that can be used to sign data. */
+export type Key = string | number[] | ArrayBuffer | Uint8Array;
+
+/** An abstract interface for a keyring which handles signing of data based on
+ * a string based digest. */
+export interface KeyRing {
```
kt3k, line 21: Why does this have a different name from the class name? How about naming both KeyStack?

crypto/keystack.ts
```diff
@@ -0,0 +1,212 @@
...
+  }
+}
+
+const REPLACEMENTS: Record<string, string> = {
+  "/": "_",
+  "+": "-",
+  "=": "",
+};
+
+function encodeBase64Safe(data: string | ArrayBuffer): string {
```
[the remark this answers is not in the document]
kitsonk: 👍  will do

crypto/keystack.ts
kitsonk: sorry, yeah, should have refactored that.

crypto/keystack.ts
kitsonk: As stated, it is an abstract interface of which `KeyStack` is _an_ implementation of it. What we are establishing here though is a generic interface people would then be able to produce their own implementations if they wanted to.  There isn't _yet_ a good home for this interface outside of here at the moment.

kt3k: Ok. Sounds fair enough

kt3k (approved):
LGTM 👍

PR #2304, opened 2022-06-03
fix(media_types): work around type issues under Node.js

kitsonk:
This works around an issue when you try to use `media_types` under Node.js via dnt.  Node.js doesn't have as good type inference as Deno does when using JSON imports, and causes some assignment issues.


kt3k (approved):
LGTM

PR #2308, opened 2022-06-04
fix(testing/bdd): Fix flat test grouping context

KyleJune:
When a test case runs, I had it make a copy of the context so that changes made to context by an individual test case do not apply to all tests in a suite like changes in a beforeAll do. What I missed is that by copying the context every time this function is called, it limits what the afterEach functions in parent suites have access to. To fix it, I updated it to only copy the context the first iteration, that way all the afterEach functions will have access to the same context used by the test case instead of only variables added in the beforeEach in the same suite level.

In my migrate module, I have a test suite that has an afterEach that should call end on the migrate object if it exists in the context of the test case.

https://github.com/udibo/migrate/blob/6aaa591a6d800786b3131600426fcd1af17a9920/postgres_test.ts#L25
```ts
const migrateTests = describe<MigrateTest>({
  name: "PostgresMigrate",
  async afterEach() {
    const { migrate } = this;
    if (migrate) {
      await migrate.end();
    }
  },
});
```

However, because of this bug, if migrate is set in a child test suite's beforeEach or test case, the above afterEach won't have access to it. Before this fix, the migrateLoadTests migrate object wouldn't get ended by the migrateTests afterEach hook. I had to add an afterEach to it to do the same thing the parent suite's afterEach does. After this fix, the afterEach below is no longer needed since the parent suite's afterEach will be accessing the same context, allowing it to call end on the migrate object.

https://github.com/udibo/migrate/blob/6aaa591a6d800786b3131600426fcd1af17a9920/postgres_test.ts#L213
```ts
const migrateLoadTests = describe<InitializedMigrateTest>({
  name: "load",
  suite: migrateTests,
  async beforeEach() {
    this.migrate = new PostgresMigrate(options);
    const { migrate } = this;
    await cleanupInit(migrate);
    await migrate.init();
  },
  // Remove after https://github.com/denoland/deno_std/pull/2308 is fixed
  async afterEach() {
    await this.migrate.end();
  },
});
```


kt3k:
I agree with this change. Can you add a test case for checking this behavior?

KyleJune:
Sorry about the late response. I went ahead and added test coverage to show that the afterEach hooks have access to the same context as nested afterEach hooks and test cases do.

kt3k (approved):
Thanks for updating test cases. LGTM!

PR #2311, opened 2022-06-05
Pass resolved ephemeral port to onListen

turuslan:
Provides valid `port` to `ServeInit.onListen`.  
Fixes default `Listening on http://hostname:0/` message.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -573,7 +573,7 @@ export async function serve(
   handler: Handler,
   options: ServeInit = {},
 ): Promise<void> {
-  const port = options.port ?? 8000;
+  let port = options.port ?? 8000;
   const hostname = options.hostname ?? "0.0.0.0";
   const server = new Server({
     port,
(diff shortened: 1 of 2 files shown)


http/server.ts
```diff
@@ -668,7 +670,7 @@ export async function serveTls(
     throw new Error("TLS config is given, but 'certFile' is missing.");
   }
 
-  const port = options.port ?? 8443;
+  let port = options.port ?? 8443;
```
cjihrig, line 673: It looks like the `serveTls()` changes are not covered.

http/server_test.ts
```diff
@@ -1372,6 +1372,21 @@ Deno.test("serve - onListen callback is called when the server started listening
   });
 });
 
+Deno.test("serve - onListen callback is called with ephemeral port", () => {
+  const abortController = new AbortController();
+  return serve((_) => new Response("hello"), {
+    port: 0,
+    async onListen({ hostname, port }) {
+      const responseText = await (await fetch(`http://localhost:${port}/`))
```
cjihrig, line 1380: It might be worth having an assertion that `port` is not 0 before this `fetch()` call just to make debugging a bit easier if this ever regresses.

kt3k (approved):
LGTM

cjihrig (approved):

PR #2314, opened 2022-06-07
chore(stream/conversion.ts): handle errors in `toTransformStream` correctly

ayame113:
> Hi, just letting you know, that an empty `catch` handler doesn't catch anything. You have to pass an empty function at least.

_Originally posted by @thekondr in https://github.com/denoland/deno_std/pull/2227#discussion_r890269446_

Based on the comments above, use `promise.catch(()=>{})` instead of `promise.catch()` that does nothing, and ignore the error correctly.

### DIFF
--- a/streams/conversion.ts
+++ b/streams/conversion.ts
@@ -243,7 +243,7 @@ export function toTransformStream<I, O>(
         } catch (error) {
           // Propagate error to stream from iterator
           // If the stream status is "errored", it will be thrown, but ignore.
-          await readable.cancel(error).catch();
+          await readable.cancel(error).catch(() => {});
           controller.error(error);
           return;
         }
--- a/streams/conversion_test.ts
+++ b/streams/conversion_test.ts
@@ -1,6 +1,6 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-import { assert, assertEquals } from "../testing/asserts.ts";
+import { assert, assertEquals, assertRejects } from "../testing/asserts.ts";
 import {
   copy,
   iterateReader,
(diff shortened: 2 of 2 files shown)


kt3k (approved):
LGTM.

Confirmed this correctly swallows the error.

```
$ deno
Deno 1.22.1+1c44f40
exit using ctrl+d or close()
> Promise.reject(new Error("foo")).catch()
Promise { <rejected> Error: foo
    at <anonymous>:2:16 }
> Promise.reject(new Error("foo")).catch(() => {})
Promise { undefined }
```

kt3k:
@ayame113 While this fix looks good, is it possible to write test to prevent the regression?

ayame113:
@kt3k 
Previously, the stream was canceled as a result of `pull()` being rejected.

```ts
// before
new ReadableStream({
  pull(controller) {
    await readable.cancel(error).catch(); // throw, then stream will be cancelled.
    controller.error(error); // This code does not execute.
  }
})
```


After this PR, the stream will be canceled by `controller.error()`.

```ts
// after
new ReadableStream({
  pull(controller) {
    await readable.cancel(error).catch(() => {});
    controller.error(error); // controller.error cancels the stream.
  }
})
```

Therefore, the behavior seen from the outside has not changed at all. I find it difficult to test this...

kt3k:
@ayame113 Both cases throw, but I think thrown errors are different. If I create an example like the below, it prints `Error: foo` in this branch and prints `TypeError: Cannot cancel a locked ReadableStream.` in main

```ts
import {
  readableStreamFromIterable,
  toTransformStream,
} from "./streams/conversion.ts";

const src = readableStreamFromIterable([0, 1, 2]);
const stream = toTransformStream(async function* (src) {
  src.getReader(); // lock the source stream to cause error at cancel
  throw new Error("foo");
});

try {
  for await (const _ of src.pipeThrough(stream)) {}
} catch (e) {
  console.log(e);
}
```

ayame113:
This test looks good to me.👍 I added a test case.

kt3k (approved):
LGTM again. Thanks for adding the test case!

PR #2316, opened 2022-06-07
fix(dotenv): support inline comment

ije:
support inline comment like:

```
FOO=bar # this is a comment
```


dotenv/testdata/.env.comments
```diff
@@ -0,0 +1,3 @@
+# This is a comment
+FOO=bar # This is a inline comment
```
cjihrig, line 2: tiny nit:

```suggestion
FOO=bar # This is an inline comment
```

cjihrig (approved):
Nice. I got bit by this bug before.

PR #2318, opened 2022-06-07
fix(streams/delimiter): avoid recursion in TextLineStream

0f-0b:
Currently `TextLineStream` could throw for large chunks due to limited stack size. This PR rewrites it to eliminate recursion and fix the bug where CRs at the end of chunks are considered newlines even without `allowCR`.


crowlKats:
Not exactly in favour of using a regex for this. and if large chunks are an issue, then we could split up chunks into smaller ones.

sjx233:
Updated to not use regexes.

kt3k (approved):
This looks nicely implemented in fewer lines. Also thanks for fixing bugs. LGTM

PR #2319, opened 2022-06-07
refactor: use wasmbuild

dsherret:


crypto/mod.ts
```diff
@@ -133,6 +134,7 @@ const stdCrypto = ((x) => x)({
         return fnv(algorithm.name, bytes);
       }
 
+      const wasmCrypto = instantiateWasm();
```
dsherret, line 137: This caches the result of previous runs, so calling this multiple times is perfectly fine and will be very fast the second time.

dsherret, line 137: Important to note that this is now instantiated lazily and not up front, so the first call to any code path that instantiates a module will be slower than before.

.github/workflows/ci.yml
```diff
@@ -195,7 +188,7 @@ jobs:
           echo "$modifications"
 
           if [[ "$modified" = "true" ]]; then
-            echo "::error ::Rebuilt WASM doesn't match committed WASM. Please rebuild and commit."
+            echo "::error ::Rebuilt Wasm doesn't match committed Wasm. Please rebuild and commit."
```
dsherret, line 191: I think maybe this verification is failing because I built it on Windows?

dsherret: I rebuilt on Linux and maybe we'll see if that works. We'll need a better solution than this in the future.

deno.json
```diff
@@ -41,6 +43,9 @@
     "node:unit": "DENO_FUTURE_CHECK=1 deno test --unstable --allow-all node/ --ignore=node/_tools/test.ts,node/_tools/versions/",
     "node:test": "deno test --unstable --allow-all node/_tools/test.ts",
     "node:setup": "deno run --allow-read --allow-net --allow-write node/_tools/setup.ts",
-    "node:check-unstable-api": "deno check --unstable ./node/module_all.ts"
+    "node:check-unstable-api": "deno check --unstable ./node/module_all.ts",
+    "build:crypto": "cd _wasm_crypto && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts --sync && mv lib/deno_std_wasm_crypto.generated.js lib/deno_std_wasm_crypto.generated.mjs",
+    "build:hash": "cd hash/_wasm && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts --sync && mv lib/deno_hash.generated.js lib/deno_hash.generated.mjs",
+    "build:varint": "cd _wasm_varint && deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts --sync && mv lib/deno_std_wasm_varint.generated.js lib/deno_std_wasm_varint.generated.mjs"
```
dsherret, line 49: What would be better to create a single source of truth is something like a `--cwd <dir>` option to say what directory deno task should use when it starts up (I like how it uses the same directory as the deno.json, but then it means in cases like this we can't create an alias).

```json
"build:crypto": "cd _wasm_crypto && deno task --cwd $PWD wasmbuild --sync",
"build:hash": "cd hash/_wasm && deno task --cwd $PWD wasmbuild --sync",
"build:varint": "cd _wasm_varint && deno task --cwd $PWD wasmbuild --sync"
"wasmbuild": "deno run --unstable -A --no-check https://deno.land/x/wasmbuild@0.5.0/main.ts"
```

bartlomieju (approved):

PR #2325, opened 2022-06-08
fix(encoding/csv/stream): properly handle CR characters

uki00a:
`CSVStream` had a problem where CRs in fields and consecutive trailing CRs were not handled properly. This PR fixes that problem.

The PR also adds some test cases.

### DIFF
--- a/encoding/csv/stream.ts
+++ b/encoding/csv/stream.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 import { defaultReadOptions, parseRecord } from "./_io.ts";
 import type { LineReader } from "./_io.ts";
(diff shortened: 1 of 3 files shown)


encoding/csv/stream.ts
```diff
@@ -47,9 +52,8 @@ export class CSVStream implements TransformStream<string, Array<string>> {
       ...options,
     };
 
-    const textLine = new TextLineStream();
-    this.#textLine = textLine;
-    this.#lineReader = new StreamLineReader(textLine.readable.getReader());
+    this.#lines = new TextDelimiterStream("\n");
```
uki00a, line 55: Uses `TextDelimiterStream` because consecutive trailing CRs should be keeped

ry (approved):
LGTM - thanks!

PR #2331, opened 2022-06-09
node: process.on("beforeExit")

bartlomieju:
<s>Blocked by https://github.com/denoland/deno/pull/14830</s>


node/process.ts
```diff
@@ -81,6 +80,9 @@ export const exit = (code?: number | string) => {
 
   if (!process._exiting) {
     process._exiting = true;
+    // FIXME(bartlomieju): this is wrong, we won't be using syscall to exit
+    // and thus the `unload` event will be emitted to properly trigget "emit"
```
cjihrig, line 84:
```suggestion
    // and thus the `unload` event will be emitted to properly trigger "emit"
```

process.js
```diff
@@ -0,0 +1,9 @@
+process.on("beforeExit", (code) => {
+  console.log("Process beforeExit event with code: ", code);
+});
+
+process.on("exit", (code) => {
+  console.log("Process exit event with code: ", code);
+});
+
```
bartlomieju, line 8: Note to self: rework into a test

node/_tools/config.json
```diff
@@ -407,8 +407,10 @@
       "test-path-win32-exists.js",
       "test-path-zero-length-strings.js",
       "test-path.js",
+      "test-process-beforeexit.js",
       "test-process-binding-internalbinding-allowlist.js",
       "test-process-env-allowed-flags.js",
+      "test-process-exit-from-before-exit.js",
```
kt3k, line 413: great!

andreubotella:
I have looked a bit into this, and from what I can tell, in Node next ticks aren't enough to keep the event loop alive. If there are any ticks enqueued in the `beforeExit` event listener, they will run, but they won't cause another `beforeExit` event to be fired.

We could fix this by removing the presence of next ticks from the conditions that keep an event loop alive, but then the various places in the CLI that fire a `beforeunload` event must make sure to also run the next ticks callback before firing the `unload` event.

bartlomieju:
@andreubotella i think that's preferable behavior. We did remove next tick checks but then had trouble with second to last callback. I think your solution is gonna alleviate that. @cjihrig what do you think?

cjihrig:
That sounds like a worthwhile plan to me.

andreubotella:
It looks like my description of the issue above was essentially correct, but the fix I proposed wasn't. Next ticks by themselves can't keep the event loop alive, but async ops in their callbacks _can_. Therefore, it's not enough to drain the next ticks queue once after firing the `beforeunload` event – the queue has to be drained as part of the event loop, not after it ends.

I opened nodejs/node#43787 to test this behavior.

Since the event loop's pending status determines whether the `beforeunload` is canceled, for `beforeExit`, I think the next ticks queue should be drained from the `beforeunload` event listener before checking the pending status. This would mean there's no need to drain the queue for regular `beforeunload` events (as I suggested above), but I think we still should do that, for consistency between `beforeunload` and `beforeExit`.

node/process.ts
```diff
@@ -258,6 +262,13 @@ class Process extends EventEmitter {
   constructor() {
     super();
 
+    globalThis.addEventListener("beforeunload", (e) => {
+      super.emit("beforeExit", process.exitCode || 0);
+      if (core.eventLoopHasMoreWork()) {
```
andreubotella, line 267: This event listener should call into `processTicksAndRejections` from next_tick.ts before checking if there's more work. That would be enough to fix this issue, with no need to change Deno's event loop.

cjihrig: Do you mean like this:

```js
    globalThis.addEventListener("beforeunload", (e) => {
      super.emit("beforeExit", process.exitCode || 0);
      processTicksAndRejections();
      if (core.eventLoopHasMoreWork()) {
        e.preventDefault();
      }
    });
```

If so, that doesn't seem to get `deno test -A --unstable node/_tools/test.ts -- parallel/test-process-beforeexit.js` passing.

andreubotella: It doesn't pass with Deno 1.23.4, but for me it's working with canary.

cjihrig: Ah, right - because https://github.com/denoland/deno/pull/14830 hasn't been released yet 🤦 . It's working for me with latest main.

bartlomieju: Thanks for figuring it out @andreubotella!

cjihrig (approved):

PR #2332, opened 2022-06-09
feat(encoding/csv): Restructure tests

timreichen:
Replace test case array loops with `Deno.test` for better overview.


kt3k:
The current code is based on [table driven test](https://dave.cheney.net/2013/06/09/writing-table-driven-tests-in-go) style. It's common to use this style when you need to test large number of patterns of inputs. I don't think the current code has problem and don't agree this suggested change is "better" than the current state.

timreichen:
I am currently working on the csv restructure and if a test fails, it points to the `Deno.test` and not the actual test data. So the table driven test is a pain for debugging with the current output because one has to search for the test data.

kt3k:
> I am currently working on the csv restructure and if a test fails, it points to the Deno.test and not the actual test data. So the table driven test is a pain for debugging with the current output because one has to search for the test data.

Each table entry has identical test name. It also shows the diff visually well (like the below). I don't understand what this PR improves

スクリーンショット 2022-06-10 20 02 38

timreichen:
Yes, the test works the same, but you cannot jump to the test directly, which means you have to search for the test name. With individual `Deno.test` you can jump directly via the provided url, which is really useful.
Bildschirmfoto 2022-06-10 um 17 24 28
.

kt3k:
hmm, ok, you got a point.

kt3k:
`readMatrix` API is not tested in `csv_test.ts`. What is the reason of this change?

Also this change removes many error class/error message assertions. Are these change intentional?

timreichen:
because `parse` is calling `readMatrix`. So all former `readMatrix ` are now `parse` tests.

ry:
If it was a large increase in code, I would agree with @kt3k, but it's only 30% more. I can see how having explicit tests like this would be better for working on it.

timreichen:
> Also this change removes many error class/error message assertions. Are these changes intentional?

This was not intentional. `csv_test.ts` had some data missing. I added them back.

kt3k (approved):
Ok. Let's go with this. LGTM

PR #2339, opened 2022-06-11
docs(README): using deno fmt --check

sigmaSd:
Actually deno fmt --check does fail currently, some files seems to be deliberately manually formatted, should those be added as an exception to deno.json?


kt3k (approved):
LGTM

kt3k:
> Actually deno fmt --check does fail currently

What deno version do you use? The format result was slightly changed recently in https://github.com/denoland/deno/pull/14725 (std was updated in #2278)

PR #2346, opened 2022-06-14
refactor: add back collections/mod.ts with a warning

dsherret:
This adds back collections/mod.ts that was removed in https://github.com/denoland/deno_std/pull/2321 , but adds it back with a warning. This will allow people to use collecitons/mod.ts in non-production scenarios for quick hacking and then convert their code to use the specific imports before deployment to make the module graph lean (see https://github.com/denoland/deno_std/pull/2321#issuecomment-1152423433 for more rationale about why we should have a mod.ts).

Overall, my preference would be to not even have a warning here and instead just document to not use it when publishing modules, but it was discussed during the design meeting that this would be a good compromise to try to get people away from using mod.ts when distributing or deploying modules, but still be able to use it in quick scripts.

### DIFF
--- a/collections/mod.ts
+++ b/collections/mod.ts
@@ -1,0 +1,58 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+/** This module is browser compatible.
+ * @module
+ */
+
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - Discussed during CLI meeting

PR #2352, opened 2022-06-15
refactor(fs/walk): move private exported functions

timreichen:
moves `_createWalkEntrySync` and `_createWalkEntry` functions from `walk.ts` inside `_util.ts` so they do not get exported in `walk.ts` and not re-exported in `mod.ts`. Since they use the `WalkEntry` interface, it is also declared in `_util.ts` and re-exported in `walk.ts`.


kt3k (approved):
I agree that this is better organized. LGTM.

PR #2353, opened 2022-06-15
chore: fix child_process_test that uses echo on Windows

dsherret:
Use something else that works on Windows.


node/child_process_test.ts
```diff
@@ -173,7 +173,7 @@ Deno.test({
 // TODO(uki00a): Remove this case once Node's `parallel/test-child-process-spawn-event.js` works.
 Deno.test("[child_process spawn] 'spawn' event", async () => {
   const timeout = withTimeout(3000);
-  const subprocess = spawn("echo", ["ok"]);
+  const subprocess = spawn("deno", ["eval", "console.log('ok')"]);
```
cjihrig, line 176: Maybe use `Deno.execPath()` instead of `deno`.

cjihrig (approved):
LGTM with an optional nit.

kt3k (approved):

PR #2354, opened 2022-06-15
chore: updates for using deno 1.23

dsherret:
- Uses `deno task` because that uses the current executable when running deno commands now
- Uses `deno task --cwd .` in order to move wasmbuild to a new task

### DIFF
--- a/_tools/release/02_create_pr.ts
+++ b/_tools/release/02_create_pr.ts
@@ -35,7 +35,7 @@ function getPrBody() {
     `Please ensure:\n` +
     `- [ ] Version in version.ts is updated correctly\n` +
     `- [ ] Releases.md is updated correctly\n` +
-    `- [ ] All the tests in this branch have been run against the CLI release being done (\`../deno/target/release/deno test -A --unstable\`)\n\n` +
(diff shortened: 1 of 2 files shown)


_tools/release/02_create_pr.ts
```diff
@@ -35,7 +35,7 @@ function getPrBody() {
     `Please ensure:\n` +
     `- [ ] Version in version.ts is updated correctly\n` +
     `- [ ] Releases.md is updated correctly\n` +
-    `- [ ] All the tests in this branch have been run against the CLI release being done (\`../deno/target/release/deno test -A --unstable\`)\n\n` +
+    `- [ ] All the tests in this branch have been run against the CLI release being done (\`../deno/target/release/deno task test\`)\n\n` +
```
dsherret, line 38: Should we also get people running `deno task node:test`?

kt3k: I think so. `deno test -A --unstable` equivalent is `deno task test` + `deno task node:unit` + `deno task node:test`.

kt3k (approved):
LGTM!

PR #2356, opened 2022-06-16
feat(std/deno): loadConfigFile and loadConfigFileSync

timreichen:
functions to load the [configuration file](https://deno.land/manual/getting_started/configuration_file#configuration-file) from a given directory. It reads the `deno.json` or `deno.jsonc` file in the given dir and parses and returns the data.


nayeemrmn (changes requested):
It would be much better to have a `resolveConfig(entrypoint?: string | URL): string | null` which walks upwards in accordance with Deno's config resolution and just returns the path if it finds one.

timreichen:
@nayeemrmn good point. In that case I think it would make sense to have 
```ts
export function resolveConfigFilePath(dir: string | URL): string | null
export async function readConfigFile(filePath: string)
export function readConfigFileSync(filePath: string)
```
What do you think?

ry:
I'm hesitant to land this because I don't want this to get out of sync with the runtime behavior. Why do you need this?

timreichen:
> I'm hesitant to land this because I don't want this to get out of sync with the runtime behavior. Why do you need this?

I created this because I needed access to the set compilerOptions and there was no easy way to get the data.
Imo it would make sense to make this data officially and easily accessible since 3rd party building tools might use them.
I think it would make more sense to expose them via deno namespace instead of 3rd party mods to prevent desynchronisation:
```ts
Deno.config // { compilerOptions: …, lint: … etc. }
```
I did a similar proposal for deno cache (https://github.com/denoland/deno/issues/11829) for the same reasons but it was turned down, thus the proposal for std.

timreichen:
@ry closed this by accident but it is still relevant. Would it make more sense to expose the config data in the deno namespace?

```ts
const { importMap, compilerOptions } = Deno.config
```

PR #2359, opened 2022-06-16
rename(path): fromFileUrl and toFileUrl

timreichen:
- renames `fromfileUrl` to `fromfileURL` and `toFileUrl` to `toFileURL` in order to have more consistency with the naming scheme on the web platform (compare [URL.createObjectURL](https://developer.mozilla.org/de/docs/Web/API/URL/createObjectURL), [HTMLCanvasElement.toDataURL](https://developer.mozilla.org/de/docs/Web/API/HTMLCanvasElement/toDataURL))
 - deprecates the old functions


kt3k:
Suggestion makes sense to me, but I think we need more broader consensus on this change to be accepted

timreichen:
The problem then is that while it is consistent in Deno namespace, it is inconsistent with the web apis. I think uppercase `URL` is used pretty much consistently across web apis, as is `CSS`, `DOM` `JSON`, `SVG` etc. Maybe we should ask Ecma TC39 to rename all web APIs with acronyms to camelcase and deprecate the old ones. 🤷

nayeemrmn:
1) It's a bad convention since it doesn't delineate adjacent acronyms which aren't first
2) We already have precedent for rejecting this particular convention in our stable interfaces

So I see every reason to switch to proper camel/pascal case in std APIs where we haven't already done so, and not switch the other way.

PR #2361, opened 2022-06-16
fix(http/file_server): uniform status & statusText

sntran:
All other `Response` here use `Status` enum and `STATUS_TEXT`,
may as well do that for 206 and 200.


cjihrig (approved):
LGTM if the CI passes. Thanks!

kt3k (approved):
LGTM. Thanks!

PR #2362, opened 2022-06-16
chore: upgrade wasmbuild to 0.6

dsherret:
- Uses new `--js-ext mjs` flag to output an `mjs` extension
- Uses new `--check` flag for verifying the output is up to date (works across operating systems)


kt3k (approved):
Oh, this is hugely simplified. Nice!

LGTM

_wasm_crypto/README.md
```diff
@@ -11,10 +11,3 @@ deno task build:crypto
 ```
 
 This will regenerate the files in the `./lib/` folder from the Rust source.
-
-### CI
-
-If CI fails in `Verify Wasm hasn't changed` step, you need to download built
```
kt3k, line 17: Does this mean there is no diffs anymore in the built wasm based on the platforms?

dsherret: Yeah there’s no diffs. If the inputs are the same then the CI will pass. If someone builds on Windows then the output will be slightly different than if someone build on Linux, but the output should be functionally the same. So if the CI fails they should only have to run the deno task regardless of their OS, then recommit and it should pass the CI.

kt3k: Thanks for the clarification! Makes sense

PR #2363, opened 2022-06-16
Refactor(encoding/base64url):  remove export

timreichen:
removes export for `addPaddingToBase64url` utility function


kt3k (approved):
I agree this shouldn't be part of public API.

LGTM

PR #2365, opened 2022-06-17
docs: update `path.dirname` to be clear it returns the directory path

dsherret:
`dirname` is a little ambiguous (when not coming from a linux or Node.js background) because it's not clear if it returns the name (base name) of the directory (ex. "directory" for `/some/directory/file.txt`) or the directory path (ex. `/some/directory/` for `/some/directory/file.txt`). I think using the terminology "path" to mean an absolute or relative path to a file and "name" to mean the actual basename is a bit more clear. So I think it would be good to say "path" here instead of "name".

### DIFF
--- a/path/posix.ts
+++ b/path/posix.ts
@@ -216,8 +216,8 @@ export function toNamespacedPath(path: string): string {
 }
 
 /**
- * Return the directory name of a `path`.
- * @param path to determine name for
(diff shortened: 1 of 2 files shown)


cjihrig (approved):

kt3k (approved):
LGTM

Issue #2368, opened 2022-06-19
missing module node/util/types.ts node/diagnostics_channel.ts


masx200:
**Describe the bug**
missing module node/util/types.ts node/diagnostics_channel.ts
**Steps to Reproduce**
```
 deno info  https://esm.sh/undici
```

```

│ ├── https://deno.land/std@0.143.0/node/diagnostics_channel.ts (missing)

│ ├── https://deno.land/std@0.143.0/node/util/types.ts (missing)

``
**Expected behavior**

<!-- A clear and concise description of what you expected to happen. -->
no errors.
**Environment**

<!-- please complete the following information -->

v8 10.3.174.6
typescript 4.6.2
- OS: Microsoft Windows [版本 10.0.22621.105]
- deno version:1.22.3 (release, x86_64-pc-windows-msvc)
- std version:0.143.0

**Comment by masx200** [comment]
import_map.json
```json

{
    "imports": {
        "https://deno.land/std@0.143.0/node/util/types.ts": "https://deno.land/std@0.143.0/node/util/types.mjs",
        "https://deno.land/std@0.143.0/node/diagnostics_channel.ts": "./diagnostics_channel.ts"
    }
}
```
diagnostics_channel.ts

```ts

export function channel() {
    throw new Error("not implemented");
}
export function hasSubscribers() {
    throw new Error("not implemented");
}
export function Channel() {
    throw new Error("not implemented");
}

export default { channel, Channel, hasSubscribers };


```

PR #2369, opened 2022-06-19
missing module node/util/types.ts ode/diagnostics_channel.ts 

masx200:
https://github.com/denoland/deno_std/issues/2368

### DIFF
--- a/node/util/types.ts
+++ b/node/util/types.ts
@@ -1,0 +1,1 @@
+export { default } from "./types.mjs";
--- a/node/diagnostics_channel.ts
+++ b/node/diagnostics_channel.ts
@@ -1,0 +1,13 @@
+
+export function channel() {
+    throw new Error("not implemented");
+}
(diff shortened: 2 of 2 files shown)


node/util/types.ts
```diff
@@ -0,0 +1 @@
+export { default } from "./types.mjs";
```
kt3k, line 1: It feels strange to have both `types.ts` and `types.mjs`. Can you rename `types.mjs` to `types.ts` instead of creating this new file?

kt3k (commented):
Nice!

Can you also update `node/module_all.ts` so that `diagnostics_channel` can be required from `require` function?

masx200:
> 好的！
> 
> 您还可以更新`node/module_all.ts`以便功能`diagnostics_channel`需要吗？`require`



> Nice!
> 
> Can you also update `node/module_all.ts` so that `diagnostics_channel` can be required from `require` function?

ok

PR #2372, opened 2022-06-19
Fix WASI initialize()

surma:
I have no idea where the official WASI spec lives, but according to the [Node documentation](https://nodejs.org/api/wasi.html#wasiinitializeinstance), `initialize()` invokes the exported `_initialize()` function on the Wasm module, if it exists. Currently, the function throws.

### DIFF
--- a/wasi/snapshot_preview1.ts
+++ b/wasi/snapshot_preview1.ts
@@ -1733,8 +1733,11 @@ export default class Context {
       );
     }
 
-    if (typeof _initialize == "function") {
-      _initialize();
+    if (_initialize && typeof _initialize != "function") {
+      throw new TypeError(
+        "WebAssembly.Instance export _start must be a function or not be defined",
+      );
     }
+    _initialize?.();
   }
 }
--- a/wasi/snapshot_preview1.ts
+++ b/wasi/snapshot_preview1.ts
@@ -1733,12 +1733,8 @@ export default class Context {
       );
     }
 
-    if (typeof _initialize != "function") {
-      throw new TypeError(
-        "WebAsembly.instance export _initialize must be a function",
-      );
+    if (typeof _initialize == "function") {
+      _initialize();
     }
-
-    _initialize();
   }
 }
--- a/wasi/snapshot_preview1.ts
+++ b/wasi/snapshot_preview1.ts
@@ -1735,7 +1735,7 @@ export default class Context {
 
     if (_initialize && typeof _initialize != "function") {
       throw new TypeError(
-        "WebAssembly.Instance export _start must be a function or not be defined",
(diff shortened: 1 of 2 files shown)


wasi/snapshot_preview1.ts
```diff
@@ -1733,12 +1733,8 @@ export default class Context {
       );
     }
 
-    if (typeof _initialize != "function") {
-      throw new TypeError(
-        "WebAsembly.instance export _initialize must be a function",
-      );
+    if (typeof _initialize == "function") {
```
cjihrig, line 1736: I think we should still throw if `_initialize` exists but is not a function. I think what we need is a check to see if `_initialize` exists. That's what [Node.js currently does](https://github.com/nodejs/node/blob/f8f2772572a14ff458ef6f781e9d0cbf7fd678d9/lib/wasi.js#L141-L156).

cjihrig (commented):
> I have no idea where the official WASI spec lives

It has moved around a bit over time 😅 ... I *think* this is the most recent version: https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md.

I think this is where `_initialize()` is discussed, but it's now in a folder called legacy, so I'm not 100% sure: https://github.com/WebAssembly/WASI/blob/59cbe140561db52fc505555e859de884e0ee7f00/legacy/application-abi.md

surma:
> It has moved around a bit over time 😅 ... I think this is the most recent version: 
> https://github.com/WebAssembly/WASI/blob/main/phases/snapshot/docs.md.

I actually know about that one, but it also uses stuff like `Result<size, errno>` and all of that is super vaguely defined. I don’t know. I find the WASI docs greatly unsatisfying.

Anyway, addressed your comment, @cjihrig. Thanks for the review!

wasi/snapshot_preview1.ts
```diff
@@ -1733,12 +1733,11 @@ export default class Context {
       );
     }
 
-    if (typeof _initialize != "function") {
+    if (_initialize && typeof _initialize != "function") {
       throw new TypeError(
-        "WebAsembly.instance export _initialize must be a function",
+        "WebAssembly.Instance export _start must be a function or not be defined",
```
cjihrig, line 1738: Should this be:

```suggestion
        "WebAssembly.Instance export _initialize must be a function or not be defined",
```

cjihrig (approved):
LGTM with a tiny nit.

wasi/snapshot_preview1.ts
surma: 🤦

surma:
Thanks for catching my lazy copy-paste!

bartlomieju (approved):

PR #2373, opened 2022-06-20
test(encoding/json): fix parse stream test expectation

kt3k:
`ConcatenatedJSONParseStream` internally uses `JSON.parse` and the test cases depend on the error message of the native `JSON.parse`.

Recently the error message for invalid json like `{foo: 1}` has been changed from `Unexpected token f` to `Expected property name or '}'` in the latest v8 update denoland/deno#14874. We need to update these expectations in these test cases.


bartlomieju (approved):
LGTM

PR #2384, opened 2022-06-23
fix(encoding/json): add `null` to JSONValue union

jsejcksn:
[`null` is a valid JSON type](https://datatracker.ietf.org/doc/html/rfc8259#section-3), but was not included in the union in the initial PR.


kt3k (approved):
LGTM. Thanks!

PR #2390, opened 2022-06-27
feat(dotenv): include missing vars in MissingEnvVarsError

### LINKED ISSUE #2366: Include missing dotenv key(s) in MissingEnvVarsError
When loading config from a dotenv file using `config({safe:true})` it can be desirable to add an additional fallback if a key is missing, such as prompting for input from the cmdline.  
Being able to check which keys were not defined by catching `MissingEnvVarsError` would make this more reliable.

Missing dotenv keys included as a field in `MissingEnvVarsError`.

- Hard-coding the required keys
- Reading the `.env.example` file separately to retrieve the required keys
- Attempting to parse the keys from the error message

coloursofnoise:
- Closes #2366 

Note that this PR makes the `message` parameter for `MissingEnvVarsError` **required** (was optional) in order to maintain a reasonable param order while ensuring the `missing` param is required.  
If this is an issue it can be reverted, but I believe the `missing` param should remain required for consistency.

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -175,14 +175,19 @@ function assertSafe(
       `If you expect any of these variables to be empty, you can set the allowEmptyValues option to true.`,
     ];
 
-    throw new MissingEnvVarsError(errorMessages.filter(Boolean).join("\n\n"));
+    throw new MissingEnvVarsError(
+      errorMessages.filter(Boolean).join("\n\n"),
+      missing,
+    );
   }
 }
 
 export class MissingEnvVarsError extends Error {
-  constructor(message?: string) {
+  missing: string[];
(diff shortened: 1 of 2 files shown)


cjihrig:
I think this is a good change. Do you think there is any value in making `missing` optional?

coloursofnoise:
> Do you think there is any value in making missing optional?

In my opinion, no. If there is a missing key it should be included in the error, and if there's no missing key to include then you're using the wrong error.

cjihrig (approved):
LGTM but would like to know what @kt3k thinks.

kt3k (approved):
Looks like a good improvement to me. LGTM. Thanks for your contribution! @coloursofnoise

PR #2401, opened 2022-06-28
fix(encoding/csv/stream): Cancel lineReader if readable is canceled

nkronlage:
This fixes an issue with CSVStream where it would leak the file if you canceled the stream before it finished. 

The problem was the CSVStream's lineReader was not canceled if the readable stream was canceled.

I've added a test case for this. Note, it only shows up for large files so I've had to introduce a csv file containing many rows of `A,B,C`.

## Before

```
>  deno test --allow-read encoding/csv/stream_test.ts --filter "leak"
Check file:///Users/neil/github/deno_std/encoding/csv/stream_test.ts
running 1 test from ./encoding/csv/stream_test.ts
[encoding/csv/stream] cancel CSVStream during iteration does not leak file ... FAILED (39ms)

 ERRORS 

[encoding/csv/stream] cancel CSVStream during iteration does not leak file => ./encoding/csv/stream_test.ts:313:6
error: AssertionError: Test case is leaking async ops.

 - 1 async operation to op_read was started in this test, but never completed.

    at assert (deno:ext/web/00_infra.js:294:13)
    at asyncOpSanitizer (deno:runtime/js/40_testing.js:228:13)
    at async resourceSanitizer (deno:runtime/js/40_testing.js:375:7)
    at async exitSanitizer (deno:runtime/js/40_testing.js:432:9)
    at async Object.applyPermissions [as fn] (deno:runtime/js/40_testing.js:533:9)
    at async runTest (deno:runtime/js/40_testing.js:813:7)
    at async Object.runTests (deno:runtime/js/40_testing.js:1095:22)

 FAILURES 

[encoding/csv/stream] cancel CSVStream during iteration does not leak file => ./encoding/csv/stream_test.ts:313:6

FAILED | 0 passed | 1 failed | 3 filtered out (78ms)

error: Test failed
```

## After

```
> deno test --allow-read encoding/csv/stream_test.ts --filter "leak"
Check file:///Users/neil/github/deno_std/encoding/csv/stream_test.ts
running 1 test from ./encoding/csv/stream_test.ts
[encoding/csv/stream] cancel CSVStream during iteration does not leak file ... ok (34ms)

ok | 1 passed | 0 failed | 3 filtered out (67ms)
```


kt3k (approved):
@nkronlage Sorry for the delay in review. Good catch! Thanks!

PR #2403, opened 2022-06-29
feat: add createAssertSnapshot

### LINKED ISSUE #2214: testing/snapshot: "assertSnapshot" should have a way to set options in a persistent way

Currently, if I want to set some options for each snapshot in a file, I need to pass these options to each call of `assertSnapshot`. For example:

```ts
Deno.test("example", async (t) => {
  assertSnapshot(t, "first", { dir: "__snaps__" });
  assertSnapshot(t, "second", { dir: "__snaps__" });
  assertSnapshot(t, "third", { dir: "__snaps__" });
  assertSnapshot(t, "fourth", { dir: "__snaps__" });
});
```

This makes tests harder to read.

Ideally, it would be possible to do something like this:

```ts
assertSnapshot.setOptions({ dir: "__snaps__" });

Deno.test("example", async (t) => {
  assertSnapshot(t, "first");
  assertSnapshot(t, "second");
  assertSnapshot(t, "third");
  assertSnapshot(t, "fourth");
});
```

EDIT: On further consideration, I am actually leading slightly towards Option 2 below but would still be happy if this (or both) were implemented.

#### Option 1 - Wrapper Function

It is already possible to wrap assert snapshot like this:

```ts
async function assertSnapshotWithDefaultOpts<T>(
  t: Deno.TestContext,
  actual: T,
  options: SnapshotOptions<T>,
): Promise<void>;
export async function assertSnapshotWithDefaultOpts<T>(
  context: Deno.TestContext,
  actual: T,
  message?: string,
): Promise<void>;
export async function assertSnapshotWithDefaultOpts(
  context: Deno.TestContext,
  actual: unknown,
  msgOrOpts?: string | SnapshotOptions<unknown>,
): Promise<void> {
  await assertSnapshot(t, actual, {
    dir: "__snaps__",
    ...(typeof msgOrOpts === "object" ? msgOrOpts : { msg: msgOrOpts }),
  });
}

Deno.test("example", async (t) => {
  assertSnapshotWithDefaultOpts(t, "first");
  assertSnapshotWithDefaultOpts(t, "second");
  assertSnapshotWithDefaultOpts(t, "third");
  assertSnapshotWithDefaultOpts(t, "fourth");
});
```

However, this is unnecessarily verbose in my opinion.

#### Option 2 - Curried Function

A function (e.g. "createAssertSnapshot") could be exposed from `testing/snapshot.ts`.

```ts
const assertSnapshot = createAssertSnapshot({ /* options */ });

Deno.test("example", async (t) => {
  assertSnapshot(t, "first");
  assertSnapshot(t, "second");
  assertSnapshot(t, "third");
  assertSnapshot(t, "fourth");
});
```

This has the advantage that you could create different assertions and export them from a test utility file. For example,

```ts
export const assertSnapshotStr = createAssertSnapshot<string>({
  serializer: (actual: string) => /* --snip-- */,
});
export const assertSnapshotReactNode = createAssertSnapshot<React.ReactNode>({
  serializer: (actual: React.ReactNode) => /* --snip-- */,
  dir: "__components__",
});
```

bcheidemann:
Closes #2214

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -276,7 +276,7 @@ with default options.
 ```ts
 // example_test.ts
 import { createAssertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
-import { stripColor } from "https://deno.land/std@$STD_VERSION/testing/fmt/colors.ts";
+import { stripColor } from "https://deno.land/std@$STD_VERSION/fmt/colors.ts";
 
 const assertSnapshot = createAssertSnapshot({
   dir: ".snaps",
--- a/testing/README.md
+++ b/testing/README.md
@@ -235,6 +235,71 @@ The `assertSnapshot` function will only attempt to read from and write to
 snapshot files. As such, the allow list for `--allow-read` and `--allow-write`
 can be limited to only include existing snapshot files, if so desired.
 
+### Options:
+
+The `assertSnapshot` function optionally accepts an options object.
+
+```ts
+// example_test.ts
+import { assertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
+
+Deno.test("isSnapshotMatch", async function (t): Promise<void> {
+  const a = {
+    hello: "world!",
+    example: 123,
+  };
+  await assertSnapshot(t, a, {
+    // options
+  });
+});
+```
+
+You can also configure default options for `assertSnapshot`.
+
+```ts
+// example_test.ts
+import { createAssertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
+
+const assertSnapshot = createAssertSnapshot({
+  // options
+});
+```
+
+When configuring default options like this, the resulting `assertSnapshot`
+function will function the same as the default function exported from the
+snapshot module. If passed an optional options object, this will take precedence
+over the default options, where the value provded for an option differs.
+
+It is possible to "extend" an `assertSnapshot` function which has been configured
+with default options.
+
+```ts
+// example_test.ts
+import { createAssertSnapshot } from "https://deno.land/std@$STD_VERSION/testing/snapshot.ts";
+import { stripColor } from "https://deno.land/std@$STD_VERSION/testing/fmt/colors.ts";
+
+const assertSnapshot = createAssertSnapshot({
+  dir: ".snaps",
+});
+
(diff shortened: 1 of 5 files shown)


kt3k (commented):
@bcheidemann Thanks for the PR. The API design and testing look good to me. Nice work!

Can you also add docs in `testing/README.md`?

testing/asserts.ts
```diff
@@ -680,10 +680,10 @@ export function assertThrows<E extends Error = Error>(
 }
 
 /** Executes a function which returns a promise, expecting it to reject. */
-export function assertRejects(
+export function assertRejects<T = unknown>(
```
bartlomieju, line 683: Seems like an unrelated change, what's the reason for it?

bartlomieju (commented):
Looking good!

testing/asserts.ts
kt3k: Ah, probably it's better to keep this unknown as the thrown value is not type checked

testing/snapshot_test.ts
```diff
@@ -491,3 +497,55 @@ Deno.test("Snapshot Test - Regression #2144", async (t) => {
...
+    );
+  });
+
+  await t.step("Message", async (t) => {
+    const assertMissingSnapshot = createAssertSnapshot<string>({
+      mode: "assert",
+      name: "[MISSING SNAPSHOT]",
+    });
+
+    const err = await assertRejects<AssertionError>(async () => {
```
kt3k, line 530: Let's use `assertRejects(fn, AssertionError)` call, that will give correct type to `err`

testing/asserts.ts
bcheidemann: Thanks for flagging, I will remove this change and refactor the test to use the format outlined below

bcheidemann:
> @bcheidemann Thanks for the PR. The API design and testing look good to me. Nice work!
> 
> Can you also add docs in `testing/README.md`?

Yeah no problem :)

testing/README.md
```diff
@@ -235,6 +235,71 @@ The `assertSnapshot` function will only attempt to read from and write to
 snapshot files. As such, the allow list for `--allow-read` and `--allow-write`
 can be limited to only include existing snapshot files, if so desired.
 
+### Options:
```
kt3k, line 238: Nice docs!

kt3k (approved):
LGTM!

PR #2409, opened 2022-07-02
fix(node/fs/exists): fix promisified exists

kt3k:
fs.exists has non-standard callback signature. We need to provide special implementation for promisify via the special property `Symbol.for("nodejs.util.promisify.custom")`

cf. https://github.com/nodejs/node/pull/13316


bartlomieju (approved):
Thanks! This fixes the problem I've encountered

PR #2412, opened 2022-07-03
feat(dotenv): stringify

timreichen:
- add `stringify` function to dotenv mod


dotenv/mod_test.ts
```diff
@@ -450,3 +456,80 @@ Deno.test("config defaults", async () => {
...
+        stringify({ "BASIC": "basic" }),
+        `BASIC=basic`,
+      ),
+  );
+  await t.step(
+    "comment",
+    () =>
+      assertEquals(
+        stringify({ "#COMMENT": "comment" }),
+        `#COMMENT=comment`,
```
kt3k, line 474: I wonder if this output makes sense

kt3k (approved):
The API design looks good to me. Left a few comments

dotenv/mod.ts
```diff
@@ -191,3 +191,39 @@ export class MissingEnvVarsError extends Error {
...
+  for (const [key, value] of Object.entries(object)) {
+    let quote;
+
+    let escapedValue = value ?? "";
+
+    if (escapedValue.includes("\n")) {
+      // escape inner new lines
+      escapedValue = escapedValue.replaceAll("\n", "\\n");
+      quote = `"`;
+    } else if (escapedValue.match(/[ \W]/)) {
```
kt3k, line 216: Isn't this regexp the same as `/\W/`?

dotenv/mod_test.ts
timreichen: I am not sure how `stringify` should handle comments. Maybe not to support them at all? An alternative to the current behaviour would be to ignore the key and either throw an error or print a warning that the key is invalid because it is a comment. What do you think would make more sense?

dotenv/mod.ts
timreichen: Yes it is. Fixed.

dotenv/mod_test.ts
kt3k: I personally think ignoring the key with printing some warnings makes most sense in this case. Throwing an error also works, but I guess some users might be annoyed by the error

timreichen: I added the check and a warn message.

kt3k (approved):
LGTM. Nice addition!

PR #2413, opened 2022-07-03
Refactor(dotenv): deno deploy

timreichen:
- moves the handling of deno deploy specific `Deno.readTextFileSync is not a function` from `mod.ts` into `load.ts`.


kt3k (approved):
LGTM. Thanks

dotenv/load.ts
```diff
@@ -2,4 +2,11 @@
 
 import { configSync } from "./mod.ts";
 
-configSync({ export: true });
+if (!(Deno.readTextFileSync instanceof Function)) {
```
jsejcksn, line 5: @timreichen Why did you change this from a `typeof` check to an `instanceof` check during the refactor?

I ask because it seems to me like the goal is simply to determine if the method exists and is callable (`typeof x === "function"`), but `x instanceof Function` has some undesirable prototype-related implications (read more at [MDN > `instanceof` > instanceof and multiple realms](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#instanceof_and_multiple_realms)). Am I missing something else?

PR #2416, opened 2022-07-03
refactor(dotenv): config to load

timreichen:
- deprecates `config`
- depreactes `configSync`
- deprecates `ConfigOptions`
- adds `load` and `loadSync` which work the same as `config` and `configSync` but take `LoadOptions`
- adds `LoadOptions` which renames `path` to `envPath`, `defaults` to `defaultsPath` and `example` to `examplePath` for more clarity. It removes the `safe` option as it will always work in safe mode except when `examplePath` is absent. it removes `export` as the `load` and `loadSync` will always populate `Deno.env`. Else `parse` should be used instead.


dotenv/mod.ts
```diff
@@ -178,7 +223,7 @@ function assertSafe(
 
   const missing = difference(
     Object.keys(confExample),
-    // If allowEmptyValues is false, filter out empty values from configuration
+    // If allowEmptyValues is false, filter out empty values from loaduration
```
kt3k, line 226:
```suggestion
    // If allowEmptyValues is false, filter out empty values from configuration
```

dotenv/mod.ts
```diff
@@ -64,71 +77,103 @@ export function parse(rawDotenv: string): DotenvConfig {
...
+  const conf = parseEnvFilesSync(options);
+  setDenoEnv(Deno.env, conf);
   return conf;
 }
 
-export async function config(
-  options: ConfigOptions = {},
-): Promise<DotenvConfig> {
-  const o: Required<ConfigOptions> = { ...defaultConfigOptions, ...options };
+// @timreichen move function code into load function once configSync is removed
```
kt3k, line 132: Please use `TODO(timreichen):` format for todo comments

Issue #2410, opened 2022-07-02
node: process.argv doesn't work if executing remote file


bartlomieju:
**Describe the bug**

If you execute a file from `esm.sh` that polyfills Node APIs and accesses `process.argv` it will throw. That's caused by `fromFileUrl` function because `process.argv[1]` is defined as follows:
```
// Overwrites the 1st item with getter.
Object.defineProperty(argv, "0", { get: Deno.execPath });
// Overwrites the 2st item with getter.
Object.defineProperty(argv, "1", { get: () => fromFileUrl(Deno.mainModule) });
```

**Steps to Reproduce**

```
deno run -A https://esm.sh/prisma@4.0.0 generate --help
error: Uncaught TypeError: Must be a file URL.
    throw new TypeError("Must be a file URL.");
          ^
    at fromFileUrl (https://deno.land/std@0.146.0/path/posix.ts:492:11)
    at Array.get (https://deno.land/std@0.146.0/node/process.ts:69:47)
    at Array.indexOf (<anonymous>)
    at C.exports (https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:20:31345)
    at ../../node_modules/.pnpm/supports-color@5.5.0/node_modules/supports-color/index.js (https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:20:31603)
    at https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:2:3398
    at ../../node_modules/.pnpm/chalk@2.4.2/node_modules/chalk/index.js (https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:21:1687)
    at https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:2:3398
    at ../../node_modules/.pnpm/@babel+highlight@7.17.12/node_modules/@babel/highlight/lib/index.js (https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:21:4845)
    at https://esm.sh/v86/prisma@4.0.0/deno/prisma.js:2:3398

```

**Expected behavior**

Don't throw when executing remote file. I think we could somehow use "fake" local path like `CWD/$deno$node.js`

**Environment**

- OS: [e.g. Ubuntu 20.04, MacOS 11] macOS
- deno version: 1.23.3
- std version: 0.146.0

CC @kt3k @cjihrig

PR #2418, opened 2022-07-03
node: process.argv[2] uses phony local module if executing remote module

bartlomieju:
Closes https://github.com/denoland/deno_std/issues/2410

### DIFF
--- a/node/process.ts
+++ b/node/process.ts
@@ -66,7 +66,13 @@ const argv = ["", "", ...Deno.args];
 // Overwrites the 1st item with getter.
 Object.defineProperty(argv, "0", { get: Deno.execPath });
 // Overwrites the 2st item with getter.
-Object.defineProperty(argv, "1", { get: () => fromFileUrl(Deno.mainModule) });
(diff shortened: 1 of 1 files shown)


node/process.ts
```diff
@@ -66,7 +66,15 @@ const argv = ["", "", ...Deno.args];
 // Overwrites the 1st item with getter.
 Object.defineProperty(argv, "0", { get: Deno.execPath });
 // Overwrites the 2st item with getter.
-Object.defineProperty(argv, "1", { get: () => fromFileUrl(Deno.mainModule) });
+Object.defineProperty(argv, "1", {
+  get: () => {
+    if (Deno.mainModule.startsWith("file:")) {
+      return fromFileUrl(Deno.mainModule);
+    } else {
+      return Deno.cwd() + "$deno$node.js";
```
kt3k, line 74: minor point, but maybe better to `path.join(Deno.cwd(), "$deno$node.js")` as `Deno.cwd()` doesn't end with path separator

Issue #2427, opened 2022-07-08
Include "semver" module


bartlomieju:
Ref https://zenn.dev/sosukesuzuki/articles/177c782f8110ba

cuobiezi:
I want to help, I worked seven years with JS. And I created a simple plan for it, maybe it is useful. Thanks 😊

kitsonk:
@cuobiezi the PR #2430 is not the approach that should be used.  We shouldn't be adding node_modules.  While the issue isn't abundently clear, we should add something like `/semver/mod.ts`, porting the code into clean TypeScript code.

That being said, @justjavac ported Node.js semver to TypeScript and made it available here: https://deno.land/x/semver@v1.4.0.  It has tests, but the licensing and attribution to the original Node.js package should be made clearer as well as get agreement from @justjavac to bring it into `std`.

justjavac:
I would move it to `std`, and I created https://semver.deno.dev

bartlomieju:
@justjavac that would be a great addition!

@cuobiezi thanks for the work, as pointed by @kitsonk I should have added description to hint that we shouldn't bring whole module from npm.

cuobiezi:
IC, Thanks for your explain.

justjavac:
This issue can be closed

PR #2430, opened 2022-07-10
feature/2427 Create a simple plan for semver

cuobiezi:


kitsonk:
Sorry, but this is not the approach we should take, as stated in #2427.

PR #2439, opened 2022-07-13
chore: workaround swc emit issues

dsherret:
We should update swc to not change identifiers that end in `1`. I think it's the hygiene and then fixer pass that causes that trouble.

### DIFF
--- a/testing/asserts_test.ts
+++ b/testing/asserts_test.ts
@@ -1300,14 +1300,14 @@ Deno.test("assert almost equals number", () => {
 Deno.test({
   name: "assertInstanceOf",
   fn(): void {
-    class TestClass1 {}
-    class TestClass2 {}
+    class TestClassA {}
+    class TestClassB {}
 
     // Regular types
     assertInstanceOf(new Date(), Date);
     assertInstanceOf(new Number(), Number);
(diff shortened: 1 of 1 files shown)


encoding/csv_stringify_test.ts
```diff
@@ -481,7 +481,7 @@ Deno.test(
     async fn() {
       const columns = [0];
       const data = [[(n: number) => n]];
-      const output = `0${NEWLINE}(n) => n${NEWLINE}`;
+      const output = `0${NEWLINE}(n)=>n${NEWLINE}`;
```
dsherret, line 484: This is a reasonable change. It just emits differently than typescript does. If anything now our emits will be more consistent than before.

testing/asserts_test.ts
```diff
@@ -1342,13 +1343,14 @@ Deno.test({
 
     let TestClassWithSameName: new () => unknown;
     {
-      class TestClass1 {}
-      TestClassWithSameName = TestClass1;
+      class TestClass3 {}
```
dsherret, line 1346: So swc renames this class to `TestClass31` because it's in the same scope as `TestClass3`. I will open an issue tomorrow to look into this.

kitsonk (approved):
LGTM

Issue #955, opened 2021-04-13
Deno 1.9.0, Uncaught InvalidData: invalid url scheme


FSou1:
I have recently published a module: https://deno.land/x/fsquery/mod.ts

It works locally and at [github actions](https://github.com/FSou1/fsquery/actions/runs/746091878).

Now I'm trying to test it:

```
import { fsselect } from 'https://deno.land/x/fsquery@deno-module-0.1/mod.ts';

await fsselect("select * from .");
```

Unfortunately, I have been receiving the error:
```
error: Uncaught InvalidData: invalid url scheme
    throw new Deno.errors.InvalidData("invalid url scheme");
          ^
    at fileURLToPath (https://deno.land/std@0.92.0/node/url.ts:53:11)
    at createRequire (https://deno.land/std@0.92.0/node/module.ts:535:18)
    at https://deno.land/x/fsquery@deno-module-0.1/parser.ts:3:17
```

Here is a tricky point, which actually causes this error:
```
import { createRequire } from "./deps.ts";
const require = createRequire(import.meta.url);
const parser = require("./peg-parser.js");
export default parser;
```

Update: I've changed the module declaration from CommonJS to ES6 module. The issue is resolved now, but still reproducible at the `fsquery@deno-module-0.1` version.

TotallyNotChase:
@lucacasonato I'd like to take this on. What do you think the error message here should be? I think, ideally, it should highlight that the `filename` url should be of `file` protocol. But it could also include a note about how `createRequire` is expected to be used from `file` modules (though this note is only relevant when the passed argument is `import.meta.url`)

lucacasonato:
Probably something along the lines of `createRequire only supports 'file://' URLs` would be enough.

Haprog:
It seems like the old error message is still being used for this even in Deno 1.23.3. I ran into this when trying to use:

```ts
import { create } from 'https://deno.land/x/deno_exif@0.0.2/mod.ts';
```

> `createRequire` can only be used from `file` modules. We should improve the error message here.

First few times I was reading this I wasn't even sure what this means. First I thought what even is a "file module"? And what is the case that is not working/supported then?

Do I understand it correctly now that `createRequire` does not work for importing via HTTP/HTTPS URLs?

[...]

PR #2440, opened 2022-07-13
fix(node/module): improve error message of createRequire

kt3k:
This PR fixes the error message when the http(s) urls are given to `createRequire`.

This addresses the issue reported in https://github.com/denoland/deno_std/issues/955#issuecomment-1180179073


bartlomieju (approved):
Lgtm

PR #2395, opened 2022-06-28
node: ChildProcess.ref()/.unref()

bartlomieju:
Depends on https://github.com/denoland/deno/pull/14991

PR #2450, opened 2022-07-19
node: Use "Deno.spawnChild" API for "child_process" polyfill

bartlomieju:
Prerequisite for https://github.com/denoland/deno_std/pull/2395


cjihrig (approved):
LGTM but the CI is failing.

kt3k:
It looks like that `end` events on `child.stdout` and `child.stderr` are not firing after the child is killed according to the test failure. I'll look into more details.

kt3k:
Removing `this.stdout.destroy()` and `this.stderr.destroy()` calls in `kill()` function seemed solving the issue.

bartlomieju:
Thanks for looking into it Yoshiya!

PR #2456, opened 2022-07-21
fix(io) StringWriter retaining references after write

jmandel:
The PR fixes  an error I'm seeing in a real-world use case when I try to copy entries from https://deno.land/std@0.141.0/archive/README.md#untar into a `StringWriter`, like:

```ts
const sw = new StringWriter();
const res = await copy(entry, sw);
 ```

~~I'm not sure whose responsibility it is to keep the buffered chunks inside `StringWriter.#chunks` isolated from the inputs to `.write`.~~

Given:
https://github.com/denoland/deno_std/blob/255f3fa8b2c7bd666018178e6dd3bea5422f23a3/io/types.d.ts#L62

... it's the `StringWriter`'s responsibility to copy data out of `p` during `write(p)`, rather than maintaining references to `p`.

### DIFF
--- a/io/writers_test.ts
+++ b/io/writers_test.ts
@@ -22,3 +22,20 @@ Deno.test("ioStringWriterSync", function (): void {
   w.writeSync(encoder.encode("\nland"));
   assertEquals(w.toString(), "deno\nland");
 });
+
+Deno.test("ioStringWriterIsolationTest", async function () {
+  const encoder = new TextEncoder();
+  const src =  "ABC";
+  const srcChunks = src.split("").map(c => encoder.encode(c))
+
+  const w = new StringWriter();
+  for (const c of srcChunks) {
+    const written = await w.write(c);
+    assertEquals(written, 1)
+  }
+
+  // These chunks all been written... but manipulating
+  // their values still changes the value of `w.toString()`
+  srcChunks[0][0] = 88;
+  assertEquals(w.toString(), src);
+});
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: io/writers_test.ts [hunk 1]
```diff
@@ -22,3 +22,20 @@ Deno.test("ioStringWriterSync", function (): void {
...
+
+  const w = new StringWriter();
+  for (const c of srcChunks) {
+    const written = await w.write(c);
+    assertEquals(written, 1)
+  }
+
+  // These chunks all been written... but manipulating
+  // their values still changes the value of `w.toString()`
+  srcChunks[0][0] = 88;
```
jmandel, line 36: These chunks all been written... so manipulating their values shouldn't change the value of `w.toString()`

io/writers_test.ts
```diff
@@ -22,3 +22,17 @@ Deno.test("ioStringWriterSync", function (): void {
...
+Deno.test("ioStringWriterIsolationTest", async function () {
+  const encoder = new TextEncoder();
+  const src =  "ABC";
+  const srcChunks = src.split("").map(c => encoder.encode(c))
+
+  const w = new StringWriter();
+  for (const c of srcChunks) {
+    const written = await w.write(c);
+    assertEquals(written, 1)
+  }
```
jmandel, line 35: These chunks all been written... so manipulating their values shouldn't change the value of w.toString()

io/writers_test.ts
```diff
@@ -22,3 +22,17 @@ Deno.test("ioStringWriterSync", function (): void {
...
+  srcChunks[0][0] = 88;
```
jmandel, line 36: The chunks all been written... so manipulating their values shouldn't change the value of w.toString()

kt3k (approved):
@jmandel Thanks for the PR!

Confirmed the bug exists and this fixes it. LGTM

PR #2457, opened 2022-07-22
node: Handle "upgrade" event

bartlomieju:


node/http.ts
```diff
@@ -462,7 +475,18 @@ class ServerImpl extends EventEmitter {
...
           const res = new ServerResponse(reqEvent);
-          this.emit("request", req, res);
+          if (req.upgrade && this.listenerCount("upgrade") > 0) {
+            Deno.upgradeHttp(req._webRequest).then(([conn, head]) => {
+              const socket = new Socket({
+                handle: new TCP(constants.SERVER, conn),
+              });
+              this.emit("upgrade", req, socket, new Buffer(head));
+            }).catch((err) => {
+              console.error("upgrade errored", err);
```
bartlomieju, line 485: This should probably emit `close` or `error` event

kt3k:
@bartlomieju Is it possible to write test case for upgrade behavior?

bartlomieju:
Yes, I will write a test (or enable some Node tests) before landing.

bartlomieju:
@kt3k I would like to land this PR without tests - I verified that it works with both Vite and express. Adding tests requires larger refactor that I'd like to tackle next week or two weeks after.

Issue #2451, opened 2022-07-20
http/http_errors: Optional headers argument


timonson:
The class `HttpError` doesn't have a `headers` property. I would suggest that an optional `headers` property and argument for `createHttpError`  might be useful for cases like `Unauthorized, 401` where a `WWW-Authenticate` header would be required.

@kitsonk would you agree to this?

Thanks!

kitsonk:
Could you give a better example of how you expect it work?

timonson:
Maybe something like this https://github.com/denoland/deno_std/pull/2462 ?

PR #2462, opened 2022-07-25
feat(http/http_errors): Add headers property

timonson:
This PR closes https://github.com/denoland/deno_std/issues/2451 by adding an optional `headers` property. Through this option, error specific headers could be added to the error (response) data. For example, if JWT verification failed, the error creation could look like this:

```ts
const httpError = createHttpError(401, "Unauthorized", {
  headers: new Headers({ "WWW-Authenticate": "Bearer" }),
});
console.log(httpError.headers); // Headers { "www-authenticate": "Bearer" }
```

### DIFF
--- a/http/http_errors.ts
+++ b/http/http_errors.ts
@@ -98,7 +98,7 @@ export type ErrorStatusKeys = keyof typeof ERROR_STATUS_MAP;
 
 export interface HttpErrorOptions extends ErrorOptions {
   expose?: boolean;
-  headers?: Headers;
+  headers?: HeadersInit;
 }
 
 /** The base class that all derivative HTTP extend, providing a `status` and an
--- a/http/http_errors.ts
+++ b/http/http_errors.ts
@@ -104,16 +104,14 @@ export interface HttpErrorOptions extends ErrorOptions {
 /** The base class that all derivative HTTP extend, providing a `status` and an
  * `expose` property. */
 export class HttpError extends Error {
-  #status: number;
+  #status: ErrorStatus = Status.InternalServerError;
   #expose: boolean;
   #headers?: Headers;
   constructor(
-    status: number,
     message = "Http Error",
     options?: HttpErrorOptions,
   ) {
     super(message, options);
-    this.#status = status;
     this.#expose = options?.expose === undefined
       ? isClientErrorStatus(this.status)
       : options.expose;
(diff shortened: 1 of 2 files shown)


http/http_errors.ts
```diff
@@ -98,6 +98,7 @@ export type ErrorStatusKeys = keyof typeof ERROR_STATUS_MAP;
 
 export interface HttpErrorOptions extends ErrorOptions {
   expose?: boolean;
+  headers?: Headers;
```
kitsonk, line 101:
```suggestion
  headers?: HeadersInit;
```

This will make it more usable.

http/http_errors.ts
```diff
@@ -132,6 +138,9 @@ function createHttpErrorConstructor(status: ErrorStatus): typeof HttpError {
       if (options?.expose === false) {
         this.#expose = false;
       }
+      if (options?.headers) {
+        this.#headers = options.headers;
```
kitsonk, line 142:
```suggestion
        this.#headers = new Headers(options.headers);
```

1. with changing it to `HeadersInit` we need to change how we get to the headers.
2. we should avoid "leaking" objects into an error object, by creating a new `Headers` instance we avoid that.

http/http_errors.ts
```diff
@@ -113,6 +114,10 @@ export class HttpError extends Error {
   get expose(): boolean {
     return false;
   }
+  /** The optional headers object that is set on the error. */
+  get headers(): Headers | undefined {
+    return undefined;
+  }
```
kitsonk, line 120:
```suggestion
  abstract get headers(): Headers | undefined;
```

Would like to avoid code that never gets called.

kitsonk (commented):
Some thoughts.

http/http_errors.ts
timonson: Wouldn't we have to add the `abstract` keyword to the class, as well? And I am not sure if you really want to do that because this would require additional changes.

kt3k: > Would like to avoid code that never gets called.

Should we also make `get expose` and `get status` abstract?

kitsonk: Actually, upon reflection, there is no need to override this in every descendent.  This value could be stored on `HttpError` as a private field, and the getter can be on `HttpError`.  The value can be picked from the options.

kitsonk: The same could actually be done with expose, but that is unrelated to this PR.

timonson: I can remove the last two commits again if you want that

kt3k (approved):
LGTM 👍

kitsonk (approved):
LGTM, thanks for the contribution!

PR #2471, opened 2022-07-27
feat(encoding): added XML parser

JuanM04:
To-do:
- [x] Implement parser
  - [x] Add multiple outputs
  - [ ] Add tests
- [ ] Implement stringify
  - [ ] Add tests
  - [ ] Support multiple inputs (?)
- [x] Added `escape` utility

### DIFF
--- a/encoding/_xml/escape.ts
+++ b/encoding/_xml/escape.ts
@@ -1,0 +1,30 @@
+/**
+ * Automatically escapes XML characters in a string.
+ *
+ * @example
+ * const unsafe = "<script>alert('Hello World')</script>";
+ * const result = escape`
+ *   <p>${unsafe}</p>
+ * `
+ *
+ * console.log(result) // => <p>&lt;script&gt;alert(&apos;Hello World&apos;)&lt;/script&gt;</p>
+ */
+export function escape(
+  templ: TemplateStringsArray,
+  ...values: unknown[]
+): string {
+  let result = templ[0];
+
+  for (let i = 1; i < templ.length; i++) {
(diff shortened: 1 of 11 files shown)


kitsonk:
This is a lot of work. Has there been a conversation with a core team member about this being a good contribution to `deno_std`?  If so, it would be great to link to it or an issue where conversation has gone on.

JuanM04:
I did not have any conversation about this. I was experimenting with XML and, since Deno has support for YAML, TOML, CSV, and other common formats, I thought this would be a nice addition. If you don't think this should be integrated in `deno_std`, I could easily extract it into its own package

kt3k:
Is it possible to make this compatible with [DOMParser](https://developer.mozilla.org/en-US/docs/Web/API/DOMParser) API?

andreubotella:
Note that @b-fuze is independently thinking of adding integration with the XML parser `xml5_ever` in their [deno_dom](github.com/b-fuze/deno-dom) library.

kitsonk:
Having XML parser as part of [deno_dom](https://deno.land/x/deno_dom@v0.1.33-alpha) would be a lot better in my opinion. YAML, TOML, CSV are far more simplistic of parsers than XML.

cjihrig:
This has been idle for a few months. @kitsonk @kt3k do we want to move forward with this? If not, can we close it out?

kt3k:
I'm in favor of closing this one for now.

I agree with @kitsonk . I think we should have xml parser as part of DOM API instead of invented API.

PR #2474, opened 2022-07-27
fix(node): correct import specifiers

### LINKED ISSUE #2472: node/internal: incorrect import
The following import doesn't exist:

https://github.com/denoland/deno_std/blob/48058b109aea531df44790a95e5b59221b3e9ff8/node/internal/streams/duplexify.mjs#L5

It is concerning that we don't have tests that identify this.

### LINKED ISSUE #2473: node/internal: incorrect import in lazy_transform
The following import doesn't exist:

https://github.com/denoland/deno_std/blob/48058b109aea531df44790a95e5b59221b3e9ff8/node/internal/streams/lazy_transform.mjs#L5

It is `util.ts`.

It is concerning that we don't have tests that discover this.

kt3k:
fixes #2472 
fixes #2473

### DIFF
--- a/node/internal/streams/duplexify.mjs
+++ b/node/internal/streams/duplexify.mjs
@@ -2,5 +2,5 @@
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 // deno-lint-ignore-file
 
-import { duplexify } from "./duplex.js";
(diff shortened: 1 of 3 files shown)


kitsonk (approved):
LGTM

Issue #1977, opened 2022-03-02
support parsing "set-cookie" header in http/cookie.ts getCookies method


andykais:
**Is your feature request related to a problem? Please describe.**
I need to parse the "set-cookie" response header of a request for a web scraper.

**Describe the solution you'd like**
`getCookie` could support parsing "set-cookie" response headers, similar to how it parses "cookie" request headers

**Describe alternatives you've considered**
implementing my own cookie parser logic, since https://deno.land/std/http/cookie.ts doesnt look for the "set-cookie" header

getspooky:
Cookie implementation is based on https://datatracker.ietf.org/doc/html/rfc6265 and doesn't include parsing "set-cookie" response headers.

andykais:
alright. I ended up just reimplementing the `getCookie` logic myself, which is pretty straight forward. If supporting `set-cookie` is out of the scope of std, then thats fine since I have my workaround. Ill let a @getspooky or another contributor close this issue if that is the case.

(side note)
this seems kind of redundant
```ts
    for (const kv of c) {
      const [cookieKey, ...cookieVal] = kv.split("=");
      assert(cookieKey != null);
      const key = cookieKey.trim();
      out[key] = cookieVal.join("=");
    }
```
why not just write
```ts
    for (const kv of c) {
      const [cookieKey, cookieVal] = kv.split("=", 2);
      assert(cookieKey != null);
      const key = cookieKey.trim();
      out[key] = cookieVal;
    }
```

getspooky:
@andykais thanks for feedback, i will check that.

jsejcksn:
> why not just write
> 
> ```ts
>     for (const kv of c) {
>       const [cookieKey, cookieVal] = kv.split("=", 2);
>       assert(cookieKey != null);
>       const key = cookieKey.trim();
>       out[key] = cookieVal;
>     }
> ```

@andykais Setting the [`limit`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#parameters) argument to `2` would exclude all of the parts of the string after a potential second occurrence of `"="`.

Looking at [the syntax in the spec](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.1), it's not evident to me that `"="` is an invalid token in a cookie value. A safer approach which also avoids creating an intermediate array would be:

```ts
for (const kv of c) {
  const sliceIndex = kv.indexOf("=");
[...]

PR #2475, opened 2022-07-27
feat(http/cookie): add set-cookie headers parser

iuioiua:


http/cookie.ts
```diff
@@ -218,3 +218,97 @@ export function deleteCookie(
...
+      "Cookies with names starting with `__Host-` must have path be `/`.",
+    );
+  }
+  return cookie;
+}
+
+/**
+ * Parse set-cookies of a header
+ * @param {Headers} headers The headers instance to get set-cookies from
+ * @return {Object} Object with cookie names as keys
```
kt3k, line 303:
```suggestion
 * @return {Cookie[]} List of cookies
```

http/cookie.ts
```diff
@@ -218,3 +218,97 @@ export function deleteCookie(
...
+        }
+        cookie.unparsed.push([key, value].join("="));
+    }
+  }
+  if (cookie.name.startsWith("__Secure-")) {
+    /** This requirement is mentioned in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie but not the RFC. */
+    assert(
+      cookie.secure,
+      "Cookies with names starting with `__Secure-` must be set with the secure flag.",
+    );
```
kt3k, line 279: I feel it's more user-friendly to show this message as warning and skip this 'set-cookie' value. I believe browsers work like that. (The same applies to the below 3 asserts)

What do you thinkg?

http/cookie.ts
```diff
@@ -218,3 +218,97 @@ export function deleteCookie(
...
+  for (const [key, value] of attrs.slice(1)) {
+    switch (key.toLocaleLowerCase()) {
+      case "expires":
+        cookie.expires = new Date(value);
+        break;
+      case "max-age":
+        cookie.maxAge = Number(value);
+        assert(
+          cookie.maxAge >= 0,
+          "Max-Age must be an integer superior or equal to 0",
```
kt3k, line 249: similar to the below comment, how about just showing warning instead of throwing?

kt3k (commented):
@iuioiua Sorry for the delay in review. Thanks for your contribution! Left a few comments

kt3k:
@iuioiua Can you also sign the CLA from the above link?

http/cookie.ts
iuioiua: Yeah, good point. Perhaps `console.assert` is more fitting. What do you think?

kt3k: Agree. `console.assert` works as well.

http/README.md
```diff
@@ -278,3 +278,19 @@ deleteCookie(headers, "deno", { path: "/", domain: "deno.land" });
 ```
 
 > Note: At the moment multiple `Set-Cookie` in a `Response` is not handled.
+
+### getSetCookies
+
+### getCookies
```
kt3k, line 284: This heading looks like a mistake?

http/README.md
```diff
@@ -278,3 +278,19 @@ deleteCookie(headers, "deno", { path: "/", domain: "deno.land" });
 ```
 
 > Note: At the moment multiple `Set-Cookie` in a `Response` is not handled.
+
+### getSetCookies
```
kt3k, line 282: I think this API is for client side handling of cookies (This is something you use when you're interacting server from client side)

Can you describe a little more about it as other methods are for server side handling of cookies, and users might be confused about the purpose of this API.

http/README.md
iuioiua: No, I named it `getSetCookies` deliberately. I understand the possible confusion given both `get` and `set` prefixes in the same function name, but I believe it makes sense once its purpose is made clear.

[Go's cookie.go](https://github.com/golang/go/blob/master/src/net/http/cookie.go) uses the alike `readSetCookies`, which makes sense to me, personally.

Perhaps, `parseSetCookies` might be a better-suited name. What do you think?

http/README.md
iuioiua: Yes, this is for client-side cookie handling. I can't think of any other context in parsing the `Set-Cookie` header that would be needed besides on the client.

Again, [Go's cookie.go](https://github.com/golang/go/blob/master/src/net/http/cookie.go) has the alike `readSetCookies` function, and there's a [comment in the current code](https://github.com/denoland/deno_std/blob/cac7764e3878ac8c2a3c26a1201f3dfcc15e6910/http/cookie.ts#L194-L195) that points out that such a function is needed (if I read it right).

True, it is the only client-side-based function here. But I think it is useful, and related functionality appears to be wanted by some.

http/cookie.ts
iuioiua: It turns out using `console.assert` might not be workable. My latest change instead uses `console.warn`. I'm unsure whether it's the best way to do it, but at least invalid cookies aren't thrown now. And I'm not sure whether logging to the console is desirable in tests and such... What do you think?

http/cookie.ts
kt3k: @iuioiua How about making this `console.warn` instead of throwing?

http/README.md
kt3k: `getSetCookies` looks fine to me, but there's another subsection with the title of `getCookies` here. Is this intentional?

http/cookie.ts
```diff
@@ -218,3 +217,105 @@ export function deleteCookie(
...
+  if (!headers.has("set-cookie")) {
+    return [];
+  }
+  return [...headers.entries()]
+    .filter(([key]) => key === "set-cookie")
+    .map(([_, value]) => value)
+    /** Parse each `set-cookie` header separately */
+    .map(parseSetCookie)
+    /** Skip empty cookies */
+    .filter(Boolean) as Cookie[];
```
kt3k, line 320: 👍

kt3k (commented):
@iuioiua Sorry for the delay in review. The change mostly looks good to me. Added a few comments about small details

http/cookie.ts
iuioiua: Oh, missed this. Fixed now.

http/README.md
iuioiua: Wow, I completely misread that! Apologies. Fixed now 😅

kt3k (approved):
@iuioiua Thanks for updating! LGTM

kt3k:
CI failures look irrelevant to this PR. I'll look into that

iuioiua:
Closes #1977.

PR #2480, opened 2022-07-28
fix(node/http(s)): Set the url protocol by default

gamer0mega:
Node automatically sets the protocol in http and https defaults, which makes some node modules break if they do not set a protocol in the options.

Feel free to improve the code added as I'm not that familiar with how ts works.


kt3k (approved):
@gamer0mega Thanks for the PR.

Confirmed this behavior is compatible with Node.js. LGTM

(I added a test case)

PR #1382, opened 2021-10-11
test(node): check star & default imports match

AaronO:
WIP, lots of mismatches, some fixes ...

PR #2481, opened 2022-07-29
test(node): check default export and named export

kt3k:
alternative approach to #1382 

This PR checks that default export items match to named export items for each node compatible modules.


bartlomieju (approved):
LGTN, is there some script/test case that verifies that exports are as expected?

kt3k:
@bartlomieju
> is there some script/test case that verifies that exports are as expected?

`node/module_all_test.ts` checks that `import x` and `import * as x` matches for all modules (except `events` and `stream` which have some specific issues)

PR #2484, opened 2022-07-31
docs(collections): fix code example in `include_value` function

nokazn:
This is a tiny fix for a code example of `collections/include_value` function. Added closing triple backquotes for the code block.

### Actual

image

### Expected

image


kt3k (approved):
@nokazn Good catch! Thanks!

PR #2487, opened 2022-08-01
feat(testing): allow for stubbing non existent instance methods

### LINKED ISSUE #2486: testing/mock.ts: Allow stubbing non existent instance methods

When writing unit tests, I often run into issues when importing an entire dependency. That's why I usually create a minimal version of the dependency that only implements a few methods required for the test to work.
I then cast the mock object to the original object, so that it can still be passed in existing apis.
When using `stub()` on these (cast) objects, I'm not getting any type errors, but at run time an error is thrown:
```
property is not an instance method
```

This often leads me to using patterns like
```js
mockObject.mockFunction = () => {}
stub(mockObject, "mockFunction", () => {});
```

In cases where the function is expected to return a value, this is even more tedious because I have to either `@ts-ignore`, cast, or actually return an object with the correct type in the first function. At that point I might as well not use `stub()` at all and just assign the function directly.

Allow stubbing methods when they don't exist, and don't throw the following error when a property is `undefined`:
https://github.com/denoland/deno_std/blob/3ea3c9fad88575bf1ff373e2bb9ebcea05d391a2/testing/mock.ts#L386-L388

- The alternative described above works for now.
- Not using `stub()` and assigning the function directly also works, but this doesn't benefit from autocompletion, which `stub()` does have.

jespertheend:
Fixes #2486

### DIFF
--- a/testing/_test_utils.ts
+++ b/testing/_test_utils.ts
@@ -20,3 +20,7 @@ export class Point {
 export function stringifyPoint(point: Point) {
   return point.toString();
 }
+
+export type PointWithExtra = Point & {
+  nonExistent: () => number;
+};
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -383,7 +383,7 @@ export function stub<
   property: keyof Self,
   func?: (this: Self, ...args: Args) => Return,
 ): Stub<Self, Args, Return> {
-  if (typeof self[property] !== "function") {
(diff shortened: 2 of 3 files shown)


kt3k (approved):
LGTM

kt3k:
Note: this supports the example in https://github.com/denoland/deno_std/issues/2486#issuecomment-1201466115

PR #2491, opened 2022-08-02
feat(encoding/csv): sync parse

timreichen:
- removes BufReader as input for `parse`
- makes `parse` sync and browser compatible


kt3k (approved):
LGTM

Nice to have this module browser compatible.

Dropping of BufReader support makes sense to me as we're leaving from Reader/Writer interfaces for IO

Issue #2494, opened 2022-08-03
node: fs.existsSync throws when returning false in Node


bartlomieju:
**Describe the bug**

Running `std/node` in Deno:
```
fs.existsSync("/dev/my-vue-app/vite.config.js/package.json");
Error: Not a directory (os error 20), stat '/dev/my-vue-app/vite.config.js/package.json'
```

Running in Node:
```
fs.existsSync("/dev/my-vue-app/vite.config.js/package.json");
false
```

**Expected behavior**

Should return false, this is needed for Vite compatibility.

Related https://github.com/denoland/deno_std/issues/1216
Ref https://github.com/denoland/deno_std/issues/2055

PR #2495, opened 2022-08-03
fix(node): fs.existsSync never throws

bartlomieju:
Fixes https://github.com/denoland/deno_std/issues/2494

`fs.existsSync` should never throw as per Node implementation


kt3k (approved):
LGTM

PR #2496, opened 2022-08-04
docs(io): remove title attribute

dcdunkan:
Having the `title` attribute in the code blocks, removes the syntax highlighting of that code block in the webpage. Either we should support attributes. But since it seems like the `title` attribute is used nowhere else in this repository READMEs, I removed those two.

### DIFF
--- a/io/README.md
+++ b/io/README.md
@@ -4,7 +4,7 @@
 
 Read reader[like file], line by line:
 
-```ts title="readLines"
+```ts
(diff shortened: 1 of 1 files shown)


kt3k (approved):
Good catch. Thanks! @dcdunkan 

Confirmed this isn't syntax highlighted in https://deno.land/std@0.150.0/io

PR #2501, opened 2022-08-07
node: remove import to global.ts from module_esm.ts

bartlomieju:
This import seems suspicious, just trying to remove it


kt3k (approved):
LGTM

PR #2502, opened 2022-08-08
feat(http): add CookieMap and SecureCookieMap

kitsonk:
This PR adds `CookieMap` and `SecureCookieMap` to `std/http`. It is part of the ongoing effort to migrate parts of oak/oak_commons into `std`.

`CookieMap` provides a `Map` like interface to request and response cookies, while `SecureCookieMap` provides a similar interface, but accepts a key ring object for transparent verifying and signing cookies, helping ensure cookies cannot be tampered with client side.  These features have been present in oak for an extended period of time and are well liked and used.

Because of the async nature that built in encryption, `SecureCookieMap` methods operate asynchronously.  Because of this, when @kt3k and I paired on it, we agreed that having a sync `CookieMap` which didn't support key signing would be useful.

Also, @lucacasonato was concerned about the need to provide response headers at construction time of such an object. During pairing @kt3k and I discussed it and this PR provides `mergeHeaders()` which makes it trivial to set the headers on a response, including anything that needs to be set from the cookie map, when formulating the response, and the maps do not require the headers or response object to be passed on construction.

The most straight forward use case, with automatic cookie signing and verification would look like this:

```ts
import {
  SecureCookieMap,
  mergeHeaders,
  type KeyRing,
} from "https://deno.land/std@$STD_VERSION/http/cookie_map.ts";

const request = new Request("https://localhost/", {
  headers: {
    "cookie": "bar=foo; bar.sig=S7GhXzJF3n4j8JwTupr7H-h25qtt_vs0stdETXZb-Ro"
  },
});

declare const keys: KeyRing;

const cookies = new SecureCookieMap(request, { keys });
console.log(await cookies.get("bar")); // logs "foo", with the signature being verified
await cookies.set("session", "1234567"); // this will be automatically signed

const response = new Response("test", {
  headers: mergeHeaders({
    "content-type": "text/plain",
  }, cookies);
});
```

This PR does not provide an implementation of the `KeyRing` interface, I already had a PR open for that (#2303) and will refresh it.

### DIFF
--- a/http/cookie_map.ts
+++ b/http/cookie_map.ts
@@ -330,27 +330,26 @@ export function mergeHeaders(
   return headers;
 }
 
-/** Provides an way to manage cookies in a request and response on the server
- * as a single iterable collection.
- *
- * The methods and properties align to {@linkcode Map}. When constructing a
- * {@linkcode Request} or {@linkcode Headers} from the request need to be
- * provided, as well as optionally the {@linkcode Response} or `Headers` for the
- * response can be provided. Alternatively the {@linkcode mergeHeaders}
- * function can be used to generate a final set of headers for sending in the
- * response. */
-export class CookieMap implements Mergeable {
-  #keys?: string[];
-  #requestHeaders: Headers;
-  #responseHeaders: Headers;
-  #secure: boolean;
+const keys = Symbol("#keys");
+const requestHeaders = Symbol("#requestHeaders");
+const responseHeaders = Symbol("#responseHeaders");
+const isSecure = Symbol("#secure");
+const requestKeys = Symbol("#requestKeys");
 
-  #requestKeys(): string[] {
-    if (this.#keys) {
-      return this.#keys;
+/** An internal abstract class which provides common functionality for
+ * {@link CookieMap} and {@link SecureCookieMap}. */
+abstract class CookieMapBase implements Mergeable {
+  [keys]?: string[];
+  [requestHeaders]: Headers;
+  [responseHeaders]: Headers;
+  [isSecure]: boolean;
+
+  [requestKeys](): string[] {
+    if (this[keys]) {
+      return this[keys];
     }
-    const result = this.#keys = [] as string[];
-    const header = this.#requestHeaders.get("cookie");
+    const result = this[keys] = [] as string[];
+    const header = this[requestHeaders].get("cookie");
     if (!header) {
       return result;
     }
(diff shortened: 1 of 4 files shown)


http/cookie_map.ts
```diff
@@ -0,0 +1,841 @@
...
+}
+
+export interface Mergable {
+  [cookieMapHeadersInitSymbol](): [string, string][];
+}
+
+export interface SecureCookieMapOptions {
+  /** Keys which will be used to validate and sign cookies. The key ring should
+   * implement the {@linkcode KeyRing} interface. */
+  keys?: KeyRing;
```
kt3k, line 134: Does it make sense to make this optional when we have non-secure alternative `CookieMap`? @kitsonk

http/cookie_map.ts
```diff
@@ -0,0 +1,841 @@
...
+}
+
+/** An object which contains a `headers` property which has a value of an
+ * instance of {@linkcode Headers}, like {@linkcode Request} and
+ * {@linkcode Response}. */
+export interface Headered {
+  headers: Headers;
+}
+
+export interface Mergable {
```
kt3k, line 127: typo? `Mergable` -> `Mergeable`

kt3k (commented):
Several methods seem identical in `CookieMap` and `SecureCookieMap` (e.g. `#requestKeys()`, `[cookieMapHeadersInitSymbol]()`, `[Symbol.for("Deno.customInspect")]()`, `[Symbol.for("nodejs.util.inspect.custom")]()`). How about making them common methods in the base class?

Also left a few questions

kt3k:
The overall interfaces of `CookieMap` and `SecureCookieMap` look good to me.

http/cookie_map.ts
kitsonk: I personally think so, because the APIs are sync/async. So if you want to write your code to always use the async APIs but still only sometimes pass keys, you would just always use the secure version.

This is what I would do with oak, as swapping between the two implementations is a problem.

kt3k: Ok. Makes sense to me

kitsonk:
> Several methods seem identical in `CookieMap` and `SecureCookieMap` (e.g. `#requestKeys()`, `[cookieMapHeadersInitSymbol]()`, `[Symbol.for("Deno.customInspect")]()`, `[Symbol.for("nodejs.util.inspect.custom")]()`). How about making them common methods in the base class?

I thought about it, but uutside of the custom inspect symbols, you cannot access private methods in extended classes, which means that I would have to make it "visible" somehow as well as expose the private data as well to make `[cookieMapHeadersInitSymbol]` work.  So it would end up leaking more and likely end up in more code, and so it seems abstraction for abstraction sake.

kt3k:
> you cannot access private methods in extended classes, which means that I would have to make it "visible" somehow as well as expose the private data as well to make [cookieMapHeadersInitSymbol] work.

I think we can use unique symbols for such properties/methods. If we don't export that symbol, the users still cannot access to them, and there shouldn't be any leakage

timreichen:
I wonder if it would be better to add something like `toHeaders()` to `CookieMap` and `SecureCookieMap` that returns a `Headers` which can be merged with the others instead of adding a new `Mergeable` with special handling in `mergeHeaders`.

```ts
export class CookieMap {
  toHeaders() {
    const headers = new Headers()
    for (const [key, value] of this.#responseHeaders) {
      if (key === "set-cookie") {
        headers.append(key, value);
      }
    }
    return headers
  }
}
```
which then can be used like so:
```ts
const headers = mergeHeaders(secureCookies.toHeaders(), cookies.toHeaders());
```

kitsonk:
> I wonder if it would be better to add something like `toHeaders()` to `CookieMap` and `SecureCookieMap` that returns a `Headers` which can be merged with the others instead of adding a new `Mergeable` with special handling in `mergeHeaders`.
> 
> ```ts
> export class CookieMap {
>   toHeaders() {
>     const headers = new Headers()
>     for (const [key, value] of this.#responseHeaders) {
>       if (key === "set-cookie") {
>         headers.append(key, value);
>       }
>     }
>     return headers
>   }
> }
> ```
> 
> which then can be used like so:
> 
> ```ts
> const headers = mergeHeaders(secureCookies.toHeaders(), cookies.toHeaders());
> ```

Why make it more complicated to use?

In most cases when forming a response, you would do something like this:

```ts
new Response(body, {
  headers: mergeHeaders(responseHeaders, secureCookies),
});
```

`.toHeaders()` is totally arbitrary and redundant.  Symbols were specifically introduced for cases like this where you specifically only want a method that you don't really want users to call, which we don't.

kitsonk:
@kt3k refactored.

timreichen:
> Why make it more complicated to use?

Because I think it is not good if a utility function has exclusive access to class data. It is impossible for example to write your own utility function that handles `CookieMap` or a class that inherits `CookieMap` because of the lack of access to the data. 

> Symbols were specifically introduced for cases like this where you specifically only want a method that you don't really want users to call, which we don't.

Why not?

kitsonk:
> Because I think it is not good if a utility function has exclusive access to class data. It is impossible for example to write your own utility function that handles `CookieMap` or a class that inherits `CookieMap` because of the lack of access to the data.

Not true, the symbol and the interface are both exported from the module, free for anyone else to create an implementation.  Yes, it is more difficult to access, but that is intentional.

> > Symbols were specifically introduced for cases like this where you specifically only want a method that you don't really want users to call, which we don't.
> 
> Why not?

Because it is stated managed private data that the whole abstraction exists to abstract.  Specifically because of the behaviour of `Set-Cookie` being a one off where it needs to be appended in the response, you have to carefully handle them.  Where as `mergeHeaders()` ensures this is done properly, other methods for trying to merge them can end up in responses that don't work.  It is best left as opaque as possible to the user.

kt3k:
@kitsonk 
The implementation looks good to me now.

Now I have a few questions about the semantics of CookieMap and SecureCookieMap.

I found some of Map methods are backed by requestHeaders (such as `*[Symbol.iterator]()`, `values`, `keys`, `entries`, `has`, `get`)

And some are backed by responseHeaders (`set` `delete`). Because of these differences, the semantics of CookieMap/SecureCookieMap might look confusing to the users for some cases.

For example:
```ts
import { CookieMap } from "./http/cookie_map.ts";
const cookies = new CookieMap(new Headers());
cookies.set("c", "3");
console.log(cookies.has("c"));
```
This prints `false`, but I think the users might expect `true` here. Similarly:

```ts
const cookies = new CookieMap(new Headers({ "cookie": "a=1;b=2" }));
cookies.delete("b");
console.log(cookies.get("b"));
```
This prints `2` instead of `undefined`.

My questions are:
- Is this behavior difference from Map intentional?
- Does it make sense to make all methods backed by `requestHeaders`?

kitsonk:
> This prints `false`, but I think the users might expect `true` here.

This is intentional. Cookies are not actually set until they are sent to the client and a subsequent request is made. `get()` and `has()` only represent cookies that are actually set on the client, and not ones that are yet to be set. This is the behaviour similar to the abstractions in express and koa and oak.

kt3k:
> This is intentional. Cookies are not actually set until they are sent to the client and a subsequent request is made. get() and has() only represent cookies that are actually set on the client, and not ones that are yet to be set. This is the behaviour similar to the abstractions in express and koa and oak.

Thanks for the clarification. Makes sense to me now.

kt3k (approved):
LGTM

PR #2503, opened 2022-08-08
feat(crypto): add subtle.timingSafeEqual()

kitsonk:
This adds `subtle.timingSafeEqual()` to `std/crypto` which is used to compare array buffers in a way that cannot be exploited in timing based attacks.

This is like to be added to WebCrypto at some point in the near future (see: https://github.com/w3c/webcrypto/issues/270) but until that point, we should have it in std. It is required #2303 which I am refactoring as part of the migration of oak commons to std.

It also refactors `node/crypto/timingSafeEqual` to use this instead of its own implementation (which is very close to this implementation, but supports Node.js's `Buffer` as well).

### DIFF
--- a/crypto/README.md
+++ b/crypto/README.md
@@ -1,5 +1,10 @@
 # crypto
 
+Extensions to the WebCrypto interface, which provides additional encryption
+algorithms that are not part of the web standard, as well as a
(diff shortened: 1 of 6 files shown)


kt3k (approved):
This sounds a good idea to me.

LGTM

PR #2504, opened 2022-08-08
fix(http/file_server): throw error inside of serveFile if filePath is a directory

timonson:
This behavior was unintentionally changed after `0.114.0`. The function `serveFile` only serves files (no directories).


kt3k:
Note: #1479 changed the timing of throwing when the directory path is given as `filePath` parameter. It used to immediately throw, but now it throws when the response body is read

This PR tries to restore the previous behavior

BTW in `file_server.ts`, fileInfo.isDirectory is already checked before passing filepath to `serveFile`. So this change only affects users who use `serveFile` API directly.

timonson:
> So this change only affects users who use serveFile API directly.
Yes, exactly. I should have made this clear.

kt3k:
@timonson 
I've looked at test cases of serveFile, and in some cases it returns `Response` object with error status code. Does it make sense to your use cases to return 404 `Response` object instead of throwing?

timonson:
That's a great point and I agree there should be consistent behavior with regards to *error handling* inside the function. For my specific use case it would work.
We could put everything inside a `try` block and invoke `serveFallback` inside the `catch` block.

kt3k:
Note: #2529 partially did what we planned

cjihrig:
Closing as https://github.com/denoland/deno_std/pull/2659 supersedes this PR.

PR #2508, opened 2022-08-10
feat(http): support cert, key options in serveTls

disjukr:
Since [`certFile` and `keyFile` option in `Deno.listenTls` is deprecated and will be removed in Deno 2.0](https://github.com/denoland/deno/pull/13740), so I changed to use `cert` and `key` option directly in `serveTls` and changed `ServeTlsInit` interface a bit.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -696,9 +696,8 @@ export async function serveTls(
     once: true,
   });
 
-  const key = options.key || await Deno.readTextFile(options.keyFile!);
-  const cert = options.cert || await Deno.readTextFile(options.certFile!);
-  if (server.closed) return;
+  const key = options.key || Deno.readTextFileSync(options.keyFile!);
+  const cert = options.cert || Deno.readTextFileSync(options.certFile!);
 
   const listener = Deno.listenTls({
     port,
--- a/http/server.ts
+++ b/http/server.ts
@@ -683,9 +683,6 @@ export async function serveTls(
     throw new Error("TLS config is given, but 'cert' is missing.");
   }
 
-  const key = options.key || await Deno.readTextFile(options.keyFile!);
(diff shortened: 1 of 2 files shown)


kt3k (commented):
@disjukr Thanks for the PR. This looks a nice change to me! But can you also add a test case in `http/server_test.ts` to check these options working?

We'd also appreciate if you update the examples in jsdoc of `serveTls` to use these new options.

disjukr:
@kt3k Done. Thank you for the review :)

disjukr:
@kt3k Sorry, there was the timing issue if abort signal fired while before reading the cert file.
Could you please run the CI one more time?

disjukr:
It runs fine on my m1 mac, but it seems to be crashing on linux.. I'll look into it a bit more later.

disjukr:
Maybe.. this will pass the CI now... 😔

kt3k (approved):
@disjukr LGTM. Nice work!

kt3k:
Note: The CI failure is caused by #2517, which is irrelevant of this change. We'll land this after #2517 is resolved

http/server.ts
```diff
@@ -683,7 +696,20 @@ export async function serveTls(
     once: true,
   });
 
-  const s = server.listenAndServeTls(options.certFile, options.keyFile);
+  const key = options.key || Deno.readTextFileSync(options.keyFile!);
+  const cert = options.cert || Deno.readTextFileSync(options.certFile!);
```
kt3k, line 700: (Side Note: `Deno.readTextFileSync` makes sense to me here because the current version of `serveTls` starts the listener synchronously. This keeps that behavior.)

disjukr:
> Note: The CI failure is caused by #2517, which is irrelevant of this change. We'll land this after #2517 is resolved

@kt3k Great. Thank you!!

PR #2515, opened 2022-08-11
docs: removed types from jsdoc directives

### LINKED ISSUE #2493: docs: remove types in jsdoc @param directive
We sometimes write jsdoc `@param` directive like the below:
```
@param {string} name Name of cookie
```

But, this should be:
```
@param name Name of cookie
```
as we use typescript.

iuioiua:
Closes #2493. This was done by replacing `@param\s+\{.*\}` with `@param`. The same was done for `@return` and `@returns`.

Might be worth noting that keeping these directives type-free might have to be done via a new rule for `deno lint`. Otherwise, it has to be done manually like this.

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -166,8 +166,8 @@ function validateDomain(domain: string): void {
 
 /**
  * Parse cookies of a header
- * @param {Headers} headers The headers instance to get cookies from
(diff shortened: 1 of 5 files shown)


kt3k (approved):
LGTM. Thanks!

Issue #2507, opened 2022-08-10
Suggest: `Deno.env.get("NODE_DEBUG")` should not permission prompt


dsherret:
I think probably `Deno.env.get("NODE_DEBUG")` should not permission prompt and do a permission check for the env var before requesting it (using `Deno.permissions.query`).

https://github.com/denoland/deno_std/blob/cac7764e3878ac8c2a3c26a1201f3dfcc15e6910/node/internal/util/debuglog.ts#L108

devadeka:
@dsherret I'd like to try my hand at this one if that's alright...

devadeka:
So had a go at it...
I found out how to check whether the permissions for the `NODE_DEBUG` environment variables have been granted used  `Deno.permissions.query`.

Could I please get more clarification on expected behaviour? I am assuming line 108 is only run when permissions are granted. What happens for the other permission states? Should it permission prompt? If so could you point me to docs on how to trigger permission prompts.

Thanks

kt3k:
Introducing TLA in `std/node` has undesired effects on some `esm.sh` modules. See https://github.com/denoland/deno/issues/14243 and https://github.com/denoland/deno_std/issues/2097 for the past discussions.

Because `Deno.permissions.query` is async, querying env permission before getting it doesn't work in this case. (And actually we removed `Deno.permissions.query` call before to avoid the above issues. See https://github.com/denoland/deno_std/pull/2098 )

dsherret:
Oh yeah, I forgot about that. I’m going to close this one, but we should come up with a way to remove this env var check eventually.

kt3k:
> but we should come up with a way to remove this env var check eventually.

I agree. Maybe we can enhance/change CLI somehow to achieve this

iuioiua:
> Introducing TLA in `std/node` has undesired effects on some `esm.sh` modules. See [denoland/deno#14243](https://github.com/denoland/deno/issues/14243) and #2097 for the past discussions.
> 
> Because `Deno.permissions.query` is async, querying env permission before getting it doesn't work in this case. (And actually we removed `Deno.permissions.query` call before to avoid the above issues. See #2098 )

I see. I’ll keep that in mind. Thanks!

nayeemrmn:
> Maybe we can enhance/change CLI somehow to achieve this

[...]

PR #2519, opened 2022-08-13
feat(node): prevent env permission prompt

iuioiua:
This is for #2507.


kt3k:
Thanks for your contribution, but see this comment https://github.com/denoland/deno_std/issues/2507#issuecomment-1214170337 in the original issue. We can't use top-level-await in `std/node` to keep the loading of `node/global.ts` synchronous.

Closing without merge. Thanks for your efforts anyways.

Issue #2426, opened 2022-07-08
Include an `accept-language` parser


lucacasonato:
I think it would be useful to include an `accept-language` parser in std. It is very useful when building internationalized websites. Maybe something along the lines of:

```ts
function parse(header: string): Intl.Locale[];
```

kitsonk:
@lucacasonato we have https://github.com/denoland/deno_std/blob/main/http/negotiation.ts#L74 now, which meets this requirement. I don't believe just parsing the header is as useful a content negotiation all together.

iuioiua:
I mostly agree, @kitsonk. My question was does `Intl.Locale` provide some added functionality or benefit to negotiation that I’m not aware of?

kitsonk:
If there is, we should adjust the acceptsLanguage to take `Intl.Locale | string` arguments and return `Intl.Locale` values.

iuioiua:
I like the idea of modifying the current `acceptsLanguages` parser, but I can't see any reason to have `Intl.Locale` as an argument. Are you able to explain?

kitsonk:
Because `acceptsLanguage` takes an [optional set of languages](https://deno.land/std@0.152.0/http/mod.ts?s=acceptsLanguages#overload_1) that are supported by a server.  These should be expressible by either `string` (as supported now) or an instance of `Intl.Locale`.  It would then be able to match on `.language` and `.baseName` without its current string parsing.  It is actually arguable the the implementation should convert strings passed in to `Intl.Locale` instances.

iuioiua:
Yeah, I think this new function might not be needed on second thought. If the user wants to use `Intl.Locale`, they can easily implement it themselves. I've never dealt with internationalisation but still can't see a good reason to use `Intl.Locale` in this instance.

lino-levan:
I agree with iuioiua and kitsonk on this one. I don't think this is really necessary for the reasons that they described.

kt3k:
Ok. Let's close.

@lucacasonato Please reopen if you have any further suggestion/opinion

PR #2520, opened 2022-08-13
feat(http): `accept-language` header parser

iuioiua:
For #2426. My questions are:
1. Is this located in the correct place or should it be in [http/negotiation.ts](https://github.com/denoland/deno_std/blob/main/http/negotiation.ts)?
2. Should we aim to replace [`parseAcceptLanguage`](https://github.com/denoland/deno_std/blob/2e7f2f521c77d6c20edd3fd19107ff31b5ed097f/http/_negotiation/language.ts#L69-L80) and make this function externally accessible?

### DIFF
--- a/http/util.ts
+++ b/http/util.ts
@@ -12,3 +12,22 @@ export function compareEtag(a: string, b: string): boolean {
   }
   return false;
 }
+
+/**
+ * Parses `accept-language` header
+ * @param header Request header value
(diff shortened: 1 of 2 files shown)


kitsonk:
I see this redundant with https://github.com/denoland/deno_std/blob/main/http/negotiation.ts#L74. I don't think adding a discreet parser helps, people want/need negotiation.

iuioiua:
Closing due to reasons explained in #2426.

PR #2523, opened 2022-08-14
fix(node): events no export static function

### LINKED ISSUE #2402: node: events.setMaxListeners()

Node.js introduced [`events.setMaxListeners()` function](https://nodejs.org/dist/latest-v18.x/docs/api/events.html#eventssetmaxlistenersn-eventtargets) since v15.4.0, but deno_std's node compatibility layer lacks this API.  Some existing libraries depending on this API does not work on Deno.

It would be great if deno_std's node compatibility layer also provides the same name of function.

rottenpen:
the fix #2402 by export the static function

## repuduction

image

## expect

image

### DIFF
--- a/node/_events.mjs
+++ b/node/_events.mjs
@@ -723,7 +723,7 @@ export function listenerCount(emitter, type) {
  * @param {string | symbol} type
  * @returns {number}
  */
-function _listenerCount(type) {
+const _listenerCount = function listenerCount(type) {
   const events = this._events;
 
   if (events !== undefined) {
@@ -737,7 +737,7 @@ function _listenerCount(type) {
   }
 
   return 0;
-}
+};
 
 /**
  * Returns an array listing the events for which
--- a/node/_events.mjs
+++ b/node/_events.mjs
@@ -701,22 +701,6 @@ EventEmitter.prototype.rawListeners = function rawListeners(type) {
   return _listeners(this, type, false);
 };
 
-EventEmitter.prototype.listenerCount = _listenerCount;
-
-/**
- * Returns the number of listeners listening to the event name
- * specified as `type`.
- * @deprecated since v3.2.0
- * @param {EventEmitter} emitter
- * @param {string | symbol} type
- * @returns {number}
- */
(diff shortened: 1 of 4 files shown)


node/_tools/test/parallel/test-event-emitter-method-names.js
```diff
@@ -38,5 +38,9 @@ assert.strictEqual(E.off, E.removeListener);  // Same method.
 Object.getOwnPropertyNames(E).forEach(function(name) {
   if (name === 'constructor' || name === 'on' || name === 'off') return;
   if (typeof E[name] !== 'function') return;
+  if (name === 'listenerCount') {
+    assert.strictEqual(E[name].name, '_listenerCount');
+    return;
+  };
```
kt3k, line 44: We want to avoid changing this file because this file is automatically copied from node.js code base.

Can we make `E[name].name` 'listenerCount' here?

node/_events.mjs
```diff
@@ -707,22 +711,19 @@ EventEmitter.prototype.rawListeners = function rawListeners(type) {
...
+  return _listenerCount.call(emitter, type);
+}
 /**
  * Returns the number of listeners listening to event name
  * specified as `type`.
  * @param {string | symbol} type
  * @returns {number}
  */
-function listenerCount(type) {
+function _listenerCount(type) {
```
kt3k, line 726: how about making this like
```ts
const _listenerCount = function listenerCount(type) { .... }
```
I guess that would avoid the change in test case below

kt3k (commented):
@rottenpen Thanks for your contribution. Left a few comments

node/_events.mjs
rottenpen: good idea!

rottenpen:
Is there anything like hasky in dene that can run test/lint/fmt before commit

kt3k:
> Is there anything like hasky in dene that can run test/lint/fmt before commit

I don't know anything in deno ecosystem.

kt3k (approved):
@rottenpen LGTM. Thanks for your contribution!

Issue #1730, opened 2021-12-19
async/delay DelayOptions should take an unref boolean


Sembiance:
With the new `Deno.unrefTimer` support in Deno v1.17, the `DelayOptions` interface for `async/delay.ts` should support a new `boolean` field `unref`

When `unref` is set to `true`, the id of the internal `setTimeout` should be automatically unreffed with `Deno.unrefTimer(i)`

Thus a call like this:
```javascript
await delay(1000, {unref : true});
```
wouldn't prevent Deno from exiting.

bartlomieju:
I think this is reasonable, @kt3k are you in favor?

kt3k:
I'm in favor.

getspooky:
Good idea.

iuioiua:
@kt3k, doing this will break browser compatibility unless a Deno namespace check is done. If that’s ok, I’m happy to work on this feature.

kt3k:
@iuioiua 
> doing this will break browser compatibility unless a Deno namespace check is done.

Right, let's check Deno namespace and keep the browser compatibility.

iuioiua:
@kt3k, I believe this issue can be closed.

PR #2527, opened 2022-08-15
feat(async/delay): add persistent option

iuioiua:
For #1730. Am I right in saying a test can't currently be done for this?

### DIFF
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -3,11 +3,13 @@
 
 export interface DelayOptions {
   signal?: AbortSignal;
+  /** Defines if the timer should be non-blocking. This is `false` by default. */
+  unref?: boolean;
 }
 
 /* Resolves after the given number of milliseconds. */
 export function delay(ms: number, options: DelayOptions = {}): Promise<void> {
-  const { signal } = options;
+  const { signal, unref } = options;
   if (signal?.aborted) {
     return Promise.reject(new DOMException("Delay was aborted.", "AbortError"));
   }
@@ -22,5 +24,8 @@ export function delay(ms: number, options: DelayOptions = {}): Promise<void> {
     };
     const i = setTimeout(done, ms);
     signal?.addEventListener("abort", abort, { once: true });
+    if (unref) {
+      Deno.unrefTimer(i);
+    }
   });
 }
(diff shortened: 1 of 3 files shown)


0f-0b:
`unrefTimer` should probably be added to `_deno_unstable.ts`.

kt3k:
> unrefTimer should probably be added to _deno_unstable.ts.

Right. Let's wrap it with type guard condition `typeof Deno.unrefTimer == "function"`

Also let's mention that the option depends on `--unstable` flag.

kt3k:
Another naming candidate for this option would be `persistent` as used in `fs.watch` of node.js https://nodejs.org/api/fs.html#fswatchfilename-options-listener

Which do you think is better?

async/delay.ts
```diff
@@ -3,11 +3,13 @@
 
 export interface DelayOptions {
   signal?: AbortSignal;
+  /** Defines if the timer should be non-blocking. This is `false` by default. */
```
kt3k, line 6: `the timer should be non-blocking` sounds strange to me as timers are always async operations. I would document this like `Indicates whether the process can exit while the timer still existing`

iuioiua: > Another naming candidate for this option would be `persistent` (with opposite effect of boolean value) as used in `fs.watch` of node.js https://nodejs.org/api/fs.html#fswatchfilename-options-listener
> 
> Which do you think is better?

Yeah, I like persistent.

async/delay.ts
```diff
@@ -1,13 +1,16 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
+import { unrefTimer } from "../_deno_unstable.ts";
 
 export interface DelayOptions {
   signal?: AbortSignal;
+  /** Indicates whether the process can exit before the timer ends. This is `false` by default. */
+  persistent?: boolean;
```
kt3k, line 8: If we use `persistent`, then the meaning (effect) is the opposite of `unref`. The doc would be `Indicates whether the process should continue to run as long as timer existing` and it defaults to true.

kt3k (commented):
@iuioiua Thanks for updating! Left one comment

async/delay.ts
iuioiua: My misunderstanding. Fixed now!

kt3k (approved):
LGTM, nice!

kt3k:
Let's wait a little while to see if there's any concern on this addition

PR #2528, opened 2022-08-15
fix(node): Stricter checking of TypedArray

petamoriken:
Changed to use [get %TypedArray%.prototype [ @@toStringTag ]](https://tc39.es/ecma262/#sec-get-%typedarray%.prototype-@@tostringtag) for checking TypedArray.


kt3k (approved):
@petamoriken LGTM, thanks!

PR #2529, opened 2022-08-15
feat(http/file-server) return 404 response if file not found

### LINKED ISSUE #2005: Return 404 from serveFile when resource is not found
When using the [serveFile](https://deno.land/std/http/file_server.ts) function like so:

```typescript
import { serve } from "https://deno.land/std@0.128.0/http/server.ts";
import { serveFile } from "https://deno.land/std@0.128.0/http/file_server.ts";

serve(async (req: Request) => await serveFile(req, `${Deno.cwd()}` + new URL(req.url).pathname));
```
this successfully returns content requested.  However, if the requested resource doesn't exist, the `serveFile` code returns 500 internal server error.  Instead it would ideally return a 404.

As an added thought, accepting a custom response handler for 404 and 500 responses would be nice.

iuioiua:
closes #2005

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -110,7 +110,7 @@ export interface ServeFileOptions {
 }
 
 /**
- * Returns an HTTP Response with the requested file as the body. Throws `Deno.errors.NotFound` if the requested file doesn't exist.
+ * Returns an HTTP Response with the requested file as the body.
  * @param req The server request context used to cleanup the file handle.
  * @param filePath Path of the file to serve.
  * @param options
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -123,145 +123,159 @@ export async function serveFile(
...
+  try {
+    let file: Deno.FsFile;
+    if (fileInfo === undefined) {
+      [file, fileInfo] = await Promise.all([
+        Deno.open(filePath),
+        Deno.stat(filePath),
+      ]);
+    } else {
+      file = await Deno.open(filePath);
+    }
```
kt3k, line 135: I think `Deno.errors.NotFound` can be thrown only from this part. How about surround only this part with `try {...} catch {...}` clause?

http/file_server_test.ts
```diff
@@ -932,6 +932,18 @@ Deno.test(
     assertEquals(await res.text(), localFile);
   },
 );
+
+Deno.test(
+  "file_server `serveFile` returns 404 due to file not found",
```
kt3k, line 937: nice test case 👍

http/file_server.ts
kt3k: Not a beautiful solution, but I think something like the below would work,
```ts
let file_: Deno.FsFile;
// ... try catch here
const file = file_!;
```

kt3k: BTW, a big `try {...} catch {...}` block blurs what is being caught in it. So it also hurts the readability in my opinion

iuioiua: There, fixed! I think when I last tried, I didn't have the `throw` in the catch block.

kt3k (approved):
LGTM. Thanks!

Issue #2444, opened 2022-07-17
Feat request: `assertIsNotError`


harrysolovay:
Feature request: an `assertIsNotError` util to compliment [assertIsError](https://github.com/denoland/deno_std/blob/66671c44955f9fd0b48dd9859127bea2f79c4037/testing/asserts.ts#L571) assertion. I frequently return (not throw) errors.

```ts
declare function tryGetThing(): TheThing | TryGetThingError;
```

To test such function requires us to perform checks to narrow the fn's result union type to exclude its error-based members.

```ts
const theThingOrError = tryGetThing();
if (!(theThingOrError instanceof Error)) {
  // `theThingOrError` is not an error within this scope
}
```

Ideally I could utilize a built-in assertion instead.

```ts
const theThingOrError = tryGetThing();
assertIsNotError(theThingOrError);
// `theThingOrError` is no longer an error as of this line
```

cuobiezi:
@harrysolovay There is assertIsError. Is it your want ?
```ts
export function assertIsError<E extends Error = Error>(
  error: unknown,
  // deno-lint-ignore no-explicit-any
  ErrorClass?: new (...args: any[]) => E,
  msgIncludes?: string,
  msg?: string,
): asserts error is E {
  if (error instanceof Error === false) {
    throw new AssertionError(`Expected "error" to be an Error object.`);
  }
  if (ErrorClass && !(error instanceof ErrorClass)) {
    msg = `Expected error to be instance of "${ErrorClass.name}", but was "${
      typeof error === "object" ? error?.constructor?.name : "[not an object]"
    }"${msg ? `: ${msg}` : "."}`;
    throw new AssertionError(msg);
  }
  if (
    msgIncludes && (!(error instanceof Error) ||
      !stripColor(error.message).includes(stripColor(msgIncludes)))
  ) {
    msg = `Expected error message to include "${msgIncludes}", but got "${
      error instanceof Error ? error.message : "[not an Error]"
    }"${msg ? `: ${msg}` : "."}`;
    throw new AssertionError(msg);
  }
}
```

harrysolovay:
As I specified above, this util would be...

> [...] to compliment [assertIsError](https://github.com/denoland/deno_std/blob/66671c44955f9fd0b48dd9859127bea2f79c4037/testing/asserts.ts#L571) assertion.

`assertIsError` does not provide the desired behavior. I'm looking for the inverse (to assert something is not an error instance).

cuobiezi:
[...]

PR #2530, opened 2022-08-16
feat(testing/asserts): assertNotInstanceOf

### LINKED ISSUE #2309: testing/asserts: `assertNotInstanceOf()`

I would like to assert that an object is NOT an instance of a specific type. (i.e. the complement to  #2025).

An `assertNotInstanceOf` to complement `assertInstanceOf`.

I've currently taken the source for the existing `assertInstanceOf` and transformed it into an `assertNotInstanceOf` but IMO it would be more ideal to not need to bring my own but already have this in `deno_std`:

```ts
type AnyConstructor = new (...args: any[]) => any;

/**
 * Make an assertion that `actual` is not an instance of `expectedType`.
 * If it is an instance of `expectedType` then throw.
 */
export function assertNotInstanceOf<T extends AnyConstructor>(
  actual: unknown,
  expectedType: T,
  msg = `Expected object to not be an instance of "${expectedType.name}".`
): void {
  assert(!(actual instanceof expectedType), msg);
}
```

iuioiua:
For #2444. Deliberately kept simple as to not mess with the natural order of things (`typeof actual` and `unexpectedType`).

closes #2309

### DIFF
--- a/testing/README.md
+++ b/testing/README.md
@@ -23,6 +23,8 @@ pretty-printed diff of failing assertion.
   `expected`, according to a given `epsilon` _(defaults to `1e-7`)_
 - `assertInstanceOf()` - Make an assertion that `actual` is an instance of
   `expectedType`.
+- `assertNotInstanceOf()` - Make an assertion that `actual` is not an instance
+  of `expectedType`.
 - `assertStringIncludes()` - Make an assertion that `actual` includes
   `expected`.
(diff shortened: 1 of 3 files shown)


kt3k (approved):
@iuioiua LGTM 👍

iuioiua:
For #2309.

PR #2531, opened 2022-08-17
removed(http/negotiation): custom Request type

iuioiua:


iuioiua:
This was done as the global `Request` type exists already.

kitsonk:
The code is intentional, as the type represents only what is required to support using the API.  Keeping a narrow un-exported type like this means that the functions can not only be used with the built in `Request` but other objects which have a headers property with a `get()` method that returns a string.

iuioiua:
Oh, I see. Apologies.
________________________________

PR #2532, opened 2022-08-18
fix: make setup and destroy sync

### LINKED ISSUE #2415: Top level await in log/mod.ts
**Describe the bug**
https://deno.land/std@0.146.0/log/mod.ts?codeview=#L216
Contains a top level await which when compiled to npm module with dnt will throw a warning unless targeting ESM.

**Steps to Reproduce**
Import std::log and use dnt to compile for node.

**Expected behavior**

The log library to be able to target node (maybe, not sure if actually a goal)

**Environment**

- deno version: 1.23.1
- std version: 146

crowlKats:
Fixes #2415


kt3k (approved):
LGTM

PR #2536, opened 2022-08-19
feat(encoding/csv): remove ColumnOptions

timreichen:
- removes `ColumnOptions`


kt3k (approved):
LGTM

kt3k:
We removed `parse` option from ColumnOptions in #1724. Since then this interface isn't meaningful anymore. This change makes sense to me.

PR #2537, opened 2022-08-20
feat(node): use Deno.serve() API for Node http polyfill

bartlomieju:
If running with "--unstable" flag the Node polyfill will now use new
"Deno.serve()" API instead of "Deno.serveHttp()".


bartlomieju:
I think we should print a console.warning telling users that some features might not be working unless `--unstable` flag is provided (think WS support)

kt3k (approved):
LGTM 🚀

Issue #2447, opened 2022-07-18
feature request: base58 encoding


harrysolovay:
The [base58 encoding scheme](https://tools.ietf.org/id/draft-msporny-base58-01.html) has become quite popular for a number of use cases. Any thoughts on including a standard implementation in this lib?

crowlKats:
I think it makes sense

luk3skyw4lker:
Also taking this one!

mtmulch1:
Hello, is anyone currently working on this? New around these parts. Correct me if I am wrong and someone else is working on the issue still.

I did not see a standard implementation of Go to port over. 
https://github.com/denoland/deno_std/tree/main/encoding    is the correct location for adding the exported deno stand library functionality of the encoding and test files placed in the same directory?

luk3skyw4lker:
I am currently working on this issue still but this would be the right directory to add the implementation files.

harrysolovay:
Hi @luk3skyw4lker, wondering if there are any updates on this front? It seems others (@mtmulch1) may be interested in implementing this as well.

mtmulch1:
I found another issue in this repository this one is all yours.

luk3skyw4lker:
Hello @harrysolovay 

I'm still finishing the work, there are still some details but I'm planning to open a PR this weekend to finish it

luk3skyw4lker:
BTW, sorry to take so long!

harrysolovay:
@luk3skyw4lker I appreciate you taking on this work –– thank you!

luk3skyw4lker:
Hi everyone! It weren't able for me to open a PR last weekend, but I'm still working on the issue. During the week I keep tangled at work and usually am not able to work here or provide updates, I mostly do the work at weekends, I think that this weekend I'll be able to surely open the PR

kt3k:
implemented in #2539

harrysolovay:
@luk3skyw4lker thank you so much for seeing this through. Very excited to use this.

luk3skyw4lker:
@harrysolovay you're welcome! If there is any bug you can open an issue and tag me already

PR #2539, opened 2022-08-21
feat: add base58 encoding/decoding feature

luk3skyw4lker:
Adding base58 encoding/decoding feature. Addressing #2447


encoding/base58_test.ts
```diff
@@ -0,0 +1,46 @@
...
+  }
+});
+
+Deno.test("[encoding/base58] testBase58EncodeBinary", () => {
+  for (const [input, output] of testSetBinary) {
+    assertEquals(encode(input), output);
+  }
+});
+
+Deno.test("[encoding/base64] testBase64EncodeBinaryBuffer", () => {
```
kt3k, line 35:
```suggestion
Deno.test("[encoding/base58] testBase58EncodeBinaryBuffer", () => {
```

encoding/base58_test.ts
```diff
@@ -0,0 +1,46 @@
...
+  for (const [input, output] of testSetBinary) {
+    assertEquals(encode(input.buffer), output);
+  }
+});
+
+Deno.test("[encoding/base64] testBase64DecodeBinary", () => {
```
kt3k, line 41:
```suggestion
Deno.test("[encoding/base58] testBase58DecodeBinary", () => {
```

kt3k (commented):
Thanks for your contribution!

Can you also add a test case of `encode` of some binary data which contains several `0`s in it as there's a special handling for `0` bytes?

Can you also add a test of `decode` with the invalid input chars?

luk3skyw4lker:
@kt3k will do tonight! Thanks for the review

luk3skyw4lker:
@kt3k Alterations done! I've added the case test with a invalid char and with zero bytes

kt3k (approved):
LGTM. Thanks! @luk3skyw4lker 

We land this PR after #2549 is resolved

timreichen:
@luk3skyw4lker I was wondering if `const mapBase58 = […]` is really necessary or if 
```ts
let carry = mapBase58[notZeroString.charCodeAt(idx)];
```
could be replaced with
```ts
let carry = base58alphabet.indexOf(char);
```

timreichen:
@luk3skyw4lker I was wondering if `const const mapBase58 = […]` is really necessary or if 
```ts
let carry = mapBase58[notZeroString.charCodeAt(idx)];
```
could be replaced with
```ts
let carry = base58alphabet.indexOf(char)
```

encoding/base58.ts
```diff
@@ -0,0 +1,383 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+const mapBase58 = [
```
aapoalas, line 3: I would consider using a deno-no-lint directive to stop this array from being split up like this, and write it in a 16x16 grid. It would look cute.

luk3skyw4lker: Could also be done, I'll open the PR with the refactor and include this change

aapoalas:
> @timreichen I took the map values from the bitcoin implementation, the -1 are invalid chars for the base58 decoding. I think it could be done the way you mentioned, I didn't thought of it when I was working on. I can open another PR with this refactor during the weekend.

It's probable that the array indexing is faster. ... Maybe.

luk3skyw4lker:
> > @timreichen I took the map values from the bitcoin implementation, the -1 are invalid chars for the base58 decoding. I think it could be done the way you mentioned, I didn't thought of it when I was working on. I can open another PR with this refactor during the weekend.
> 
> It's probable that the array indexing is faster. ... Maybe.

Yeah, probably, the indexOf(char) wouldn't be a option too because it would bring the index of the char, and not the corresponding value from the table. Maybe the Record<string, number> could be a better option, written as a 16x16 matrix, it would improve code comprehension for anyone that could be doing some maintenance work.

luk3skyw4lker:
Actually they're the same, I didn't noticed it, my mistake.

luk3skyw4lker:
@timreichen I managed to make the PR, if you could check #2563 please

Issue #2534, opened 2022-08-19
dotenv forces apps to grant access to all env vars, possibly allowing attackers to read Envs not consumed by the app on purpose


mechanoid:
### Discussed in https://github.com/denoland/deno_std/discussions/2533

<div type='discussions-op-text'>

<sup>Originally posted by **mechanoid** August 19, 2022</sup>
At first let me say how great it is, that dotenv is now part of the STD. 

I was wondering though, why it uses `Deno.env.toObject()` to parse all env variables, without ability to restrict which Env vars are of interest for the app/library.

This leaves no choice but to run the script with a global `--allow-env` possibly allowing an attacker to access all env vars provided to the apps host. 

I think the `dotenv.config()` method (as well as `dotenv/load()`) should be able to take a list of ENV var names and restrict the lookup to those. This would allow to provide a more restricted access to only a subset of the Envs with `--allow-env=SOME_VAR,...`) </div>

kt3k:
closed in #2544.

@mechanoid Thank you for your suggestion!

mechanoid:
Very welcome!

PR #2544, opened 2022-08-22
feat(dotenv): allow to restrict env lookup to specific Env variables

mechanoid:
Add `restrictEnvAccessTo` option for `config` and `configSync` to provide an interface corresponding to the `--allow-env=ENV1,ENV2,...` CLI interface. This allows to reduce the exposed Environment of a service to the minimum when using dotenv.

Addressing: #2534


dotenv/mod_test.ts
```diff
@@ -643,3 +644,55 @@ Deno.test("stringify", async (t) => {
...
+  });
+
+  const decoder = new TextDecoder();
+  const conf = JSON.parse(decoder.decode(stdout).trim());
+
+  assertEquals(conf.GREETING, "hello world", "fetches .env by default");
+  assertEquals(conf.DEFAULT1, "Some Default", "default value loaded");
+});
+
+Deno.test("use of restrictEnvAccessTo for an Env var, without granting read permissions still fails", async () => {
```
kt3k, line 684:
```suggestion
Deno.test("use of restrictEnvAccessTo for an Env var, without granting env permissions still fails", async () => {
```

kt3k (approved):

LGTM with nits

dotenv/mod_test.ts
mechanoid: The delay is very ok, happy that anyone  has time for that at all!

The suggested change is of course welcome 👍

mechanoid:
Ah, you have already pushed the fix for the description! Thank you very much, much appreciated!

Issue #2541, opened 2022-08-22
suggestion(fs): removal of deprecated `exists` and `existsSync` within `std/fs`


iuioiua:
**Is your feature request related to a problem? Please describe.**

`exists` and `existsSync` are deprecated. However, these functions are still used within `std/fs`.

**Describe the solution you'd like**

Removal of `exists` and `existsSync` within `std/fs`.

I'm happy to work on this.

PR #2546, opened 2022-08-22
remove(fs): `exists` and `existsSync` usage in `std/fs`

iuioiua:
For #2541. I'm wondering whether the handling of `move` and `moveSync` is optimal for when `overwrite` is `false`, particularly, throwing an error in the `try` block.


kt3k (approved):
LGTM, nice!

fs/copy_test.ts
```diff
@@ -82,25 +82,16 @@ testCopy(
...
 
-    assertEquals(
-      await exists(srcFile),
-      true,
-      `source should exist before copy`,
-    );
-    assertEquals(
-      await exists(destFile),
-      false,
+    assert(await Deno.lstat(srcFile), "source should exist before copy");
```
dsherret, line 85: It seems there's no unit test for this? I think all of these will throw with an error that won't be as descriptive?

dsherret (changes requested):
Some `exists(...)` calls were changed to `Deno.lstat(...)` and the `Deno.lstat` will now throw.

I like how this gets rid of the TOCTOU issues, but replacing `exists` with a function that will throw and has a less descriptive name for the same functionality doesn't seem that ideal.

fs/copy_test.ts
iuioiua: That is true. As an alternative, perhaps we could exclusively use `exists` within tests. What do you think?

dsherret: Oh, I didn't pay enough attention to the code to notice this was a test (did a very fast drive by). Yeah, using it in the tests seems like a way to make them more descriptive, or use some local `exists` function to the module.

I'm going to undo my "request changes" because this is just test code.

dsherret (approved):
Didn't notice this issue was just in test code, which is ok because if it throws the test will fail. So LGTM. That said, it would be nice to use a descriptive name like `exists` in the tests.

fs/copy_test.ts
iuioiua: Perhaps, just having a clear `msg` defined within assertion tests will suffice...

PR #2547, opened 2022-08-22
Allow: err instanceof Deno.errors.AlreadyExists

RefinedSoftwareLLC:
// replaced `new Error` with `new Deno.errors.AlreadyExists` when appropriate. Enables:
```
try {
  copySync('src',  'dest', { overwrite: false })
  copySync('src',  'dest', { overwrite: false }) // already exists
} catch(err) {
  if (err instanceof Deno.errors.AlreadyExists == false) { throw err } // only ignore AlreadyExists error
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
           submodules: true
 
       - name: Set up Deno
-        uses: denoland/setup-deno@v1.0.0
+        uses: denoland/setup-deno@v1
         with:
           deno-version: ${{ matrix.deno }}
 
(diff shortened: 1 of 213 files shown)


RefinedSoftwareLLC:
This already works:
```
try {
  Deno.mkdirSync('./temp')
  Deno.mkdirSync('./temp') // already exists
} catch(e) {
  if (e instanceof Deno.errors.AlreadyExists == false) { Deno.exit(2) }
}
```

RefinedSoftwareLLC:
ci/lint should not have failed, it is recommending to make the code worse and not match the rest of each file.

kt3k (commented):
Thanks for your contribution.

The changes in `fs/` and `testing/` make sense to me. I think the changes in `node/` don't make much sense because these are node compatible layer. We shouldn't throw deno specific error from the code in `node/`

RefinedSoftwareLLC:
[node/_fs/_fs_open.ts](https://github.com/denoland/deno_std/blob/9c269464b111517e305df9a61f5b6e6b55cf6701/node/_fs/_fs_open.ts) already expects and throws Deno Errors like in ```throw new Deno.errors.AlreadyExists(`A file exists at the destination: ${destStr}`);```

kt3k (approved):
LGTM

PR #2550, opened 2022-08-23
chore: remove `void` and `Promise<void>` return types

iuioiua:


kt3k (approved):
@iuioiua LGTM, thanks!

kt3k:
Note: We discussed this long time ago and we partially did similar changes before.

ref:
- https://github.com/denoland/deno/issues/8832
- https://github.com/denoland/deno_std/pull/818

PR #2552, opened 2022-08-23
rename(tests): remove brackets from test names

timreichen:
- remove bracket names and replace bracket nesting names with nested tests


kt3k:
Is anyone in favor of this change?

I think this change doesn't solve any issue and slightly reduces the readability and context of the test cases. So I'm not in favor

cjihrig:
I agree with @kt3k here. This is also a pretty noisy style change, which could cause conflicts in other PRs for no real gain. Thanks for the PR, but I'm going to close this out.

PR #2558, opened 2022-08-24
feat(testing): better assertNotInstanceOf types

iuioiua:
For #2444. Disclaimer: this idea came from @harrysolovay, which I totally agree with. This should've been done in #2530.


kt3k (approved):
LGTM. Thanks! @iuioiua

Issue #2170, opened 2022-04-30
`assertFalse` can have better typing


kt3k:
ref https://github.com/denoland/deno_std/pull/2093#discussion_r862336747

iuioiua:
@kt3k, is this able to be closed?

kt3k:
I think this can be still explored a bit more.

What @bcheidemann suggested was:

```ts
type Falsy = false | null | undefined | "" | 0;
export function assertFalse(expr: unknown, msg = ""): asserts expr is Falsy {
  // --snip--
}
```

I'm not 100% sure this is correct way of typing it though.

iuioiua:
Ok, why not make it complete? Officially, falsy would be defined by the [following](https://developer.mozilla.org/en-US/docs/Glossary/Falsy):
```ts
type Falsy = false | 0 | -0 | 0n | "" | '' | `` | null | undefined | NaN;
```
If this sounds good. I'm happy to do this.

kt3k:
@iuioiua Sounds good to me, but TypeScript seems regarding `0` and `-0` as the same (No type error with `const a: -0 = 0`) and also "", '', and `` as the same.

iuioiua:
Ah, that makes sense. Thanks for the correction. So, it'd be:
```ts
type Falsy = false | 0 | 0n | "" | null | undefined | NaN;
```
I'll submit the PR.

kt3k:
fixed by #2562

PR #2562, opened 2022-08-25
feat(testing): better assertFalse types

iuioiua:
For #2170. FYI `typeof Nan` is `number`, so `Nan` was omitted from the `Falsy` type.


kt3k (approved):
@iuioiua LGTM. Thanks!

PR #2563, opened 2022-08-25
refactor: using indexOf for getting values at decode

luk3skyw4lker:
Refactoring decode algorithm according to @timreichen suggestions, using indexOf instead of mapBase58 array indexing


timreichen:
Ref https://github.com/denoland/deno_std/pull/2539#issuecomment-1226873509

timreichen:
LGTM

aapoalas:
I would recommend running a few basic benchmarks on the two methods. Use the `Deno.bench` that you've been so graciously given :)

kt3k:
@luk3skyw4lker Can you perform some benchmarks for current and new implementations?

luk3skyw4lker:
@kt3k can do! I'll try to post it this weekend.

But regardless of the performance implications, I think that this implementation will make it easier to maintain the code instead of the big array filled with -1 that may cause confusion and worst code readability. It was my mistake not to see this through when I was building the feature.

Anyway I will post the benchmark results at this PR at saturday or sunday, thanks for the recomendations everyone!

luk3skyw4lker:
@kt3k @aapoalas @timreichen 

So, here are the benchmarks, not so surprisingly the array indexation is faster than indexOf (since the array indexation goes directly into the right memory position and indexOf has to search it at the whole array). I don't think that they're pretty far from each other when dealing with big decoding strings.

With the string '2NEpo7TZRRrLZSi2U' as an argument:
image

With this comment contents as an argument (up until this line):
image

luk3skyw4lker:
`import { decode, decodeRefactored } from './base58.ts';

Deno.bench("Benching Base58 (main's implementation)", () => {
	decode('2NEpo7TZRRrLZSi2U');
});

Deno.bench("Benching Base58 (refactors's implementation)", () => {
	decodeRefactored('2NEpo7TZRRrLZSi2U');
});
`

Here is the code for the benchmark.

luk3skyw4lker:
```js
import { decode, decodeRefactored } from './base58.ts';

Deno.bench("Benching Base58 (main's implementation)", () => {
	decode('2NEpo7TZRRrLZSi2U');
});

Deno.bench("Benching Base58 (refactors's implementation)", () => {
	decodeRefactored('2NEpo7TZRRrLZSi2U');
});
```

Here's the code that I used for benchmarking (decodeRefactored is the function at this PR)

kt3k:
@luk3skyw4lker Thanks for the benchmarks.

The difference looks significant to me. I think we should keep the main's implementation. What do you think?

timreichen:
What about an object? Might be faster than `indexOf`:

```ts
const mapBase58: Record<string, number> = {
  "1": 0,
  "2": 1,
  "3": 2,
  …
 }

 let carry = mapBase58[char];
```

luk3skyw4lker:
Another benchmark using the object's idea from @timreichen:

image

This time the difference is very small, the array indexation still wins when we're treating small strings. But when we get the bigger string, it seems that the object's implementation is a little bit faster at the max time:

image

@kt3k I don't think that the difference is too signficant, in both benchmarks the difference was 3 microseconds, it didn't even increased with the size of the string, it's kind of imperceptible in any case.

The object's implementation solves part of the coding readability problem when written in a 10x6 matrix, I guess we should go with it (I'll commit it and you guys can take a look).

luk3skyw4lker:
I think that the indexOf approach is a lot cleaner, but if you guys think that the 3 microseconds are not worth the shot, we can go with the object implemenation.

kt3k (approved):
@luk3skyw4lker Thanks for the further benchmarks. Object map implementation looks good to me 👍

PR #2565, opened 2022-08-26
fix(encoding/json): Improve safety of `JSONValue` object type

jsejcksn:
Improves type safety by adding `undefined` to the union of values of the string-indexed object type union member of `JSONValue`: now its values are effectively optional, which better represents that no object will have a not-undefined value at every possible key.


kt3k (approved):
I think this reflects the actual type better. LGTM

@ayame113 Do you agree with this change?

ayame113:
@kt3k This looks good to me.

@jsejcksn Thanks!👍

PR #2567, opened 2022-08-26
feat: add envalid module

roj1512:
https://deno.land/x/envalid ([roj1512/envalid](https://github.com/roj1512/envalid)) is a port of the [npm.im/envalid](https://npm.im/envalid) ([af/envalid](https://github.com/af/envalid)) with slight modifications.

It makes using environment variables super easy and type-safe.

### DIFF
--- a/envalid/README.md
+++ b/envalid/README.md
@@ -1,0 +1,118 @@
+# envalid
+
+Validate environment variables.
+
+## Usage
+
+```ts
+// You can also use it with dotenv:
+// import "https://deno.land/std@$STD_VERSION/dotenv/load.ts";
+import {
+  bool,
+  cleanEnv,
+  num,
+  str,
+} from "https://deno.land/std@$STD_VERSION/envalid/mod.ts";
+
+// You can also do
+// export default cleanEnv(
+// to be importable from other locations.
+const env = cleanEnv(Deno.env.toObject(), {
+  TEXT: str(),
+  IS_X: bool(),
+  APP_ID: num(),
+  SOMETHING: str({
+    default: "default value",
+    example: "some string",
+    desc: "description",
+    docs: "https://example.com/configuration#SOMETHING",
+  }),
+  CHOICE: str({ choices: ["can be this", "or this"] }),
+});
+```
+
+## Validators
+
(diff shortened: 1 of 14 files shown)


roj1512:
Related: https://github.com/denoland/deno/issues/13396

kt3k:
I feel this tool doesn't have enough common value/popularity to be included in deno_std.

I found this download graph, for example, about config management tools in npm
https://npmtrends.com/config-vs-convict-vs-dotenv-vs-envalid

dotenv is really popular, but envalid doesn't have comparable download counts.

roj1512:
@kt3k I want other opinions before we vanish the PR.

roj1512:
> I found this download graph, for example, about config management tools in npm
> [npmtrends.com/config-vs-convict-vs-dotenv-vs-envalid](https://npmtrends.com/config-vs-convict-vs-dotenv-vs-envalid)

1. We [need](https://github.com/denoland/deno/issues/13396) computed types.
2. If ever worked on, porting something like convict or config won't help be consistent, as they are not based on dotenv.

imjamesb:
-1

I think that there are MANY ways to do data validation and there are a lot of preferences here. Data validation is such a big topic and in my own opinion I don't think Deno is ready to have a standard data validation module yet, not because the core team or the contributors wouldn't be able to make something out, but because Deno is opinionated, if Deno should have a data validation module in the standards it should be a lot of discussion around implementation and design on that module so it is efficient and feature rich, but still aligns well with Deno and it's design.

GerbenRampaart:
I was not familiar with envalid but I am now, and I do like it a lot.

timreichen:
I agree with @imjamesb. This is a very specific use case of object validation. I think std should have a data validation module that is flexible, extendable and applicable to many use cases as well as non-opinionated as possible.

roj1512:
Also, this is not a data validator. It's envalid — environment validator.

roj1512:
If in the future a data validation module was added to the std, we can make it use that.

cjihrig:
I'm also -1 to including this in std. We can revisit the decision in the future, but I'm going to close this for now. Thank you for the PR though.

PR #2568, opened 2022-08-26
fix(encoding/jsonc): Improve safety of `JSONValue` object type in `jsonc.ts`

ayame113:
As a continuation of #2565, I'd like to improve not only `encoding/json/_parse.ts` but also `encoding/jsonc.ts` types.

I should have suggested it when I commented on https://github.com/denoland/deno_std/pull/2565#issuecomment-1228053348, but I forgot.


kt3k (approved):
LGTM. Thanks!

PR #2574, opened 2022-08-28
remove(io): obsolete check in readLine

### LINKED ISSUE #2391: Please do a Deno existence check in io/buffer.ts BufReader.readline
I'd like to use std streams in the browser, and the only reference to the Deno ns after bundling is this reference:

https://github.com/denoland/deno_std/blob/09e24d81ce0bacebe81b3c9f683a6ecf9263fc12/io/buffer.ts#L499

Mind doing an existence check here to make sure it runs in the browser?

Thanks,
\- John

iuioiua:
Fixes #2391.


kt3k (approved):
LGTM.

Note: `Deno.errors.BadResource` should be still thrown in line 511. This change shouldn't make any behavior difference.

Issue #802, opened 2021-03-13
opendir and opendirSync are missing 


majidsajadi:
i have tried to implement the node.js compatibility layer for `fs.opendir` and `fs.opendirSync` but looks like these methods are missing from `fs` std library. any plan to implement these or is there any workaround for this issue?

iuioiua:
@bartlomieju, I believe this can be closed as `opendir` and `opendirSync` were implemented in #2576.

PR #2576, opened 2022-08-29
compat(fs): `opendir` and `opendirSync`

iuioiua:
For #802 and #648. Note: my tests were previously passing fine on my machine but now they're not. I'm not sure what I changed.

Also, I'm not sure how to handle `encoding` and `bufferSize` options in regards to the current implementation of `Dir`.


node/_tools/config.json
```diff
@@ -34,6 +34,7 @@
       "test-fs-append-file.js",
       "test-fs-chmod-mask.js",
       "test-fs-chmod.js",
+      "test-fs-opendir.js",
```
kt3k, line 37: Great to have this enabled! 👍

node/_fs/_fs_opendir.ts
```diff
@@ -0,0 +1,91 @@
...
+    /** Throws if path is invalid */
+    Deno.readDirSync(path);
+
+    callback(null, new Dir(path));
+  } catch (err) {
+    callback(denoErrorToNodeError(err as Error, { syscall: "opendir" }));
+  }
+}
+
+export const opendirPromise = promisify(opendir) as (
```
kt3k, line 59: Nice to have the promise version 👍

node/_fs/_fs_opendir.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  options = typeof options === "object" ? options : {};
+  options = Object.assign({
+    encoding: "utf8",
+    bufferSize: 32,
+  }, options);
+
+  try {
+    /** Throws if encoding is invalid */
+    new TextDecoder(options.encoding);
```
kt3k, line 46: I think this needs to be `validateEncoding()` call, which is defined in `node/internal/validators.mjs`. You can pass, for example, `encoding: "binary"` in opendir in node.js, but `new TextDecoder("binary")` throws.

iuioiua: Ah, ok. Am I supposed to use `path` for the `data` parameter?

kt3k: Oops, sorry `validateEncoding` doesn't look correct for this according to the thrown error message. Probably `assertEncoding` (in `node/internal/fs/utils.mjs`) is the right choice here.

kt3k (approved):
LGTM. Nice work!

iuioiua:
Please don’t merge yet. I need to check that callbacks can’t be called twice.

iuioiua:
> Please don’t merge yet. I need to check that callbacks can’t be called twice.

Ok, done, alongside some other improvements. My only remaining question is: is there any issue in using `Deno.readDirSync()`, a sync call, within the async version of `opendir()`?

kt3k:
> My only remaining question is: is there any issue in using Deno.readDirSync(), a sync call, within the async version of opendir()?

Is that necessary for throwing error synchronously (to make it more compatible to node)? If so, that should be ok.

node/_fs/_fs_opendir_test.ts
```diff
@@ -0,0 +1,144 @@
...
+          assert(dir);
+        },
+      ),
+  );
+
+  await t.step("passes if callback isn't called twice", async () => {
+    const importUrl = new URL("./_fs_opendir.ts", import.meta.url);
+    await assertCallbackErrorUncaught({
+      prelude: `import { opendir } from "${importUrl}"`,
+      invocation: `opendir("${path}", `,
```
kt3k, line 96: I think we need to do something like `path.replaceAll("\\", "\\\\")` here to avoid the error below in windows:

```
    error: AssertionError: actual: "error: Uncaught SyntaxError: Octal escape sequences are not allowed in strict mode.
          opendir("C:\Users\RUNNER~1\AppData\Local\Temp\766483fe", (err) => {
                                                        ^
        at <anonymous> (file:///C:/actions-runner/work/deno_std/deno_std/$deno$eval.js:3:53)
    " expected to contain: "Error: success"
```

iuioiua: There, that should do it. I used `JSON.stringify()`.

kt3k: > I used JSON.stringify().

That's much better 👍

kt3k (approved):
LGTM again.

iuioiua:
Last note (I think): `test-fs-opendir.js` is currently under `ignore` option for the Node test config file and it currently passes. When it's moved under the proper `tests` option, it fails due to limited current support of the `Dirent` class. The generated file has not been modified so this doesn't seem right.

kt3k:
> test-fs-opendir.js is currently under ignore option for the Node test config file and it currently passes. When it's moved under the proper tests option, it fails due to limited current support of the Dirent class.

Ah, ok. I missed that point. There seem several things still missing for actually passing `test-fs-opendir.js`. I'd suggest we should revert the compatibility testing part (changes under `node/_tools/`). The current test cases in `node/_fs/_fs_opendir_test.ts` look enough for landing this

iuioiua:
> > test-fs-opendir.js is currently under ignore option for the Node test config file and it currently passes. When it's moved under the proper tests option, it fails due to limited current support of the Dirent class.
> 
> Ah, ok. I missed that point. There seem several things still missing for actually passing `test-fs-opendir.js`. I'd suggest we should revert the compatibility testing part (changes under `node/_tools/`). The current test cases in `node/_fs/_fs_opendir_test.ts` look enough for landing this

Ok, done.

kt3k:
Thanks for updating!

PR #2582, opened 2022-08-30
BREAKING: replace ALL CAPS acronyms in public API names

kt3k:
We usually use camelCase and PascalCase for public API names, but there were some exceptions to this rule in std.

- `JSONParseStream`
- `ConcatenatedJSONParseStream`
- `JSONStringifyStream`
- `JSONValue` (type)
- `readMIMEHeader` method in `TextProtoReader` class

We discussed about these exceptions internally and decided to not allow these naming exceptions.

This PR replaces these ALL CAPS acronyms in API names with camelCased versions. (The old APIs are still aliased with `@deprecated` tags)

---
See also https://github.com/denoland/manual/pull/343

### DIFF
--- a/encoding/README.md
+++ b/encoding/README.md
@@ -522,13 +522,13 @@ supports the following formats:
 ### Basic usage
 
 If you want to parse JSON separated by a delimiter, use `TextLineStream` (or
-`TextDelimiterStream`) and `JSONParseStream`. `JSONParseStream` ignores chunks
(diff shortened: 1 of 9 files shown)


ayame113:
Out of curiosity, what about `CSVStream`?

kt3k:
@ayame113 Good point. We need to replace that too

piscisaureus (approved):
~LGTM~ Lgtm!

PR #2583, opened 2022-08-30
chore(node): remove the usage of --compat in node/integrationtest

kt3k:
This PR removes the usage of `--comapt` flag in `node/integrationtest`.

I updated `express` and `mysql2` module test to make those use `npm:` schema.

I removed `npm` and `yarn` cli tests as these tests can't be translated into `npm:` format easily, and also because we have our own dependency resolver in CLI, so the priority of running those tools should be very low now.


bartlomieju (approved):
LGTM, thanks!

Issue #2551, opened 2022-08-23
std(textproto, bytes): empty tests


timreichen:
**Describe the bug**
There are some empty tests in std. Is this intentional or were they removed by accident?

https://github.com/denoland/deno_std/blob/d7cd0080b8aa8a2df315edb220f919f4b88c4086/bytes/bytes_list_test.ts#L212
https://github.com/denoland/deno_std/blob/6d33b18a8c89f7caf0a8e7588f9ea31efe5a45fc/textproto/test.ts#L15

- std version: [0.152.0](https://github.com/denoland/deno_std/releases/tag/0.152.0)

iuioiua:
Not sure why that ignored textproto test was added (commit [here](https://github.com/denoland/deno_std/commit/1ee7c0b6724da7fe859d986c16d450e744eebecd)). I'd imagine it'd look like this, once implemented:
```ts
Deno.test({
  name: "[textproto] Reader : DotBytes",
  async fn() {
    const input =
      "dotlines\r\n.foo\r\n..bar\n...baz\nquux\r\n\r\n.\r\nanot.her\r\n";
    const tpReader = reader(input);
    assertEquals(await tpReader.readLine(), "dotlines");
    assertEquals(await tpReader.readLine(), ".foo");
    assertEquals(await tpReader.readLine(), "..bar");
    assertEquals(await tpReader.readLine(), "...baz");
    assertEquals(await tpReader.readLine(), "quux");
    assertEquals(await tpReader.readLine(), "");
    assertEquals(await tpReader.readLine(), ".");
    assertEquals(await tpReader.readLine(), "anot.her");
    assertEquals(await tpReader.readLine(), null);
  },
});
```
@kt3k, is this correct?

iuioiua:
@kt3k, this can be closed.

PR #2586, opened 2022-08-31
fix(buffer): cleanup empty test

iuioiua:
Fixes half of #2551. Note: according to [coverage](https://app.codecov.io/gh/denoland/deno_std/blob/main/bytes/bytes_list.ts), [these lines](https://github.com/denoland/deno_std/blob/d7cd0080b8aa8a2df315edb220f919f4b88c4086/bytes/bytes_list.ts#L44-L46) weren't covered. This fix simply ensures these lines are covered and removes the obsolete `[bytes] ByteList.iterator() range` test. With 100% coverage and all tests passing, we can be confident in saying that it's safe to do so.


kt3k (approved):
Nice! LGTM.

PR #2587, opened 2022-08-31
feat(node): pseudo-tty tests

iuioiua:
Adds tests from `pseudo-tty`, including `stdin-setrawmode.js` mentioned in #1579.


iuioiua:
I noticed that the number of node tests hasn’t increase from 546 compared to other PRs, which would seem wrong.

PR #2590, opened 2022-09-01
fix(node/process): fix BadResource issue of stdin.isTTY

kt3k:
This PR fixes `BadResource` error while executing `DEBUG=* PRISMA_CLI_QUERY_ENGINE_TYPE=binary deno run --unstable -A npm:prisma db pull` command. (The root cause is https://github.com/denoland/deno/issues/15708 )

This PR stores `Deno.isatty?.(Deno.stdin.rid)` value at the start of program, instead of evaluating it each time. This works around the issue above.


bartlomieju (approved):
LGTM

PR #2591, opened 2022-09-01
chore(node/child_process): disable `fork` for now

kt3k:
We are going to remove unstable `--compat` flag feature in https://github.com/denoland/deno/pull/15678

This PR temporarily disables `child_process.fork` in `std/node`, which depends on `--compat` mode.


bartlomieju (approved):
LGTM

PR #2592, opened 2022-09-01
fix: remove `_deno_unstable.ts`

iuioiua:
Playing around with this repo when `_deno_unstable.ts` isn't used.


iuioiua:
I don't fully understand the purpose of `_deno_unstable.ts`. I believe it's there as a means of ignoring checks that would otherwise throw non-useful errors. I found that getting rid of most of the use of this file doesn't break any tests, possibly due to stability improvements and fixes from `--unstable`. This PR assumes that these changes are desirable. Refs:
* #1819
* https://github.com/denoland/deno/issues/12726

Note, the following issues:
1. `testing/snapshot_test.ts` fails when use of `DenoUnstable` is removed from `fs/copy.ts`.
2. `async/delay.ts` fails when use of `DenoUnstable` is removed.

kt3k:
The purpose of `_deno_unstable.ts` was to avoid showing type errors at unstable APIs even when the user doesn't specify `--unstable` flag.

But we removed full type checking in v1.22 and v1.23 (https://deno.com/blog/v1.22#no-type-checking-by-default). `_deno_unstable.ts` might be useless now.

What do you think? @bartlomieju @nayeemrmn @dsherret

PR #2593, opened 2022-09-01
chore(node): disable test-child-process-execfile.js

kt3k:
The compat test case `test-child-process-execfile.js` depends on unstable `--compat` flag feature of CLI. The `--compat` feature is going to be removed soon in https://github.com/denoland/deno/pull/15678 . There seems no alternative way to pass these test cases without `--compat`. This PR temporarily removes these test cases.


bartlomieju (approved):
LGTM

PR #2599, opened 2022-09-02
fix(node): enable some `pseudo-tty` tests

iuioiua:
Sorry, re-create of #2587. Initially, this aimed to enable the `stdin-setrawmode.js` test, mentioned in #1579. However, it's still not passing, which I hoped/guessed it would've. Nevertheless, I've enabled other tests from `pseudo-tty` which are passing.


kt3k:
I think we probably need to modify `getPathsFromTestSuites` (in `node/_tools/common.ts`) to actually include these test cases.

iuioiua:
> I think we probably need to modify `getPathsFromTestSuites` (in `node/_tools/common.ts`) to actually include these test cases.

Done.

kt3k (approved):
LGTM. Thanks!

kt3k:
Will land when the outage of windows-xl runner availability is resolved

PR #2600, opened 2022-09-02
fix: remove `_deno_unstable.ts`

iuioiua:
Sorry, re-create of #2592.


kt3k (approved):
OK. Let's try this. LGTM

kt3k:
Note:
- `_deno_unstable.ts` was a workaround for avoiding showing type error at unstable APIs, even when the user invoked deno without `--unstable` flag. https://github.com/denoland/deno_std/pull/1819
- In deno v1.22 we removed default full type checking, and the above issue was gone. https://deno.com/blog/v1.22
- So it now should be ok to remove `_deno_unstable.ts`

bartlomieju (approved):

kt3k:
@iuioiua Thank you for suggesting this!

PR #2601, opened 2022-09-02
fix(testing/asserts): temporarily disable background colors in diff

### LINKED ISSUE #2575: assertion diff for "Expected" is barely readable
Bg green on white text makes it very hard to read anything. It should have some contrast. 

https://github.com/denoland/deno_std/blob/fa6c014902d834203914afa9a77d83c69c60c228/testing/_diff.ts#L383

image

kt3k:
closes #2575

### DIFF
--- a/testing/_diff.ts
+++ b/testing/_diff.ts
@@ -377,6 +377,10 @@ function createColor(
   diffType: DiffType,
   { background = false } = {},
 ): (s: string) => string {
+  // TODO(@littledivy): Remove this when we can detect
+  // true color terminals.
+  // https://github.com/denoland/deno_std/issues/2575
+  backround = false;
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM

lucacasonato:
Why didn't we just change the foreground text to be black?

kt3k:
black text might work, but in my terminal (mac's default terminal app) white text also works

スクリーンショット 2022-09-05 23 49 29

スクリーンショット 2022-09-05 23 48 45

I think the problem here is that certain terminal renders green background too bright, and some terminal renders it darker. I think the correct solution is to use the more specific colors (by specifying color code) when the true color is available in that terminal

PR #2603, opened 2022-09-03
fix(async): restore `delay` browser compatibility

### LINKED ISSUE #2598: bug(async): `delay` is not browser compatible
[async/delay.ts](https://github.com/denoland/deno_std/blob/5a3206d6eb96a693b90c3c64c5106dc41ccfbae4/async/delay.ts) is not browser compatible. However, it has the header claiming that it is. Reason being is the use of `Deno.unrefTimer()` for the `persistent` option (which I added 😅).
What is the best way to fix such an issue?
1. Somehow check for the existence of the `Deno` namespace and function?
2. Remove the `// This module is browser compatible.` header?
3. Revert the blocking functionality to restore browser compatibility?

iuioiua:
Fixes #2598. Only uses `Deno.unrefTimer()` if within the `Deno` namespace.
Note: there's a small amount of overlap with this and #2600.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -212,3 +212,22 @@ jobs:
       - name: Rebuild Wasm and verify it hasn't changed
         if: success() && steps.source.outputs.modified == 'true'
         run: deno task build:varint --check
+
+  deploy-compat:
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v3
+
+      - name: Deploy to Deno Deploy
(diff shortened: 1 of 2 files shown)


async/delay.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// @ts-nocheck Bypass static errors for missing --unstable.
```
kt3k, line 2: How about using `@ts-ignore` in the line above the `Deno.unrefTimer()`?

kt3k: That keeps the rest of the source code type safe, I think

PR #2611, opened 2022-09-04
feat(encoding/csv): sync stringify function 

timreichen:
- removes optional async transformation function from options
- make `stringify()` sync


luk3skyw4lker (commented):
Wouldn't it be good to also remove all occurrences of await in the test file? Seems like you're just removed the test cases where the transform function is used

luk3skyw4lker (approved):

kt3k (approved):
LGTM. Thanks!

PR #2613, opened 2022-09-05
feat(tools): add deprecation check

timreichen:
Currently there is no way for a user to know, when a `@deprecated` export will be removed.
Also some deprecated exports exists for over twenty releases. It is not clear, if that is intentional or if they were forgotten to be removed.

This PR introduces **deprecation version** and a **deprecation check** to standardise and manage deprecation in std.

# deprecation version
`@deprecated` tags now need a deprecation version to let the user know, when it will be removed.

This is the current format:

`@deprecated (will be removed in 0.157.0)`

# deprecation check
The deprecation check makes sure that every `@deprecated` tag has a deprecation version.
It compares the current mod version with the deprecation version. If they match, it will log that the export must be removed.

Note: This PR bumps the deprecation version for all `@deprecated` tags to `0.157.0`.


kt3k (commented):
- I think we should exclude the entire `/node/` directory as we can't control when to remove those.
- I prefer the expression `will be removed after X.Y.Z` as we can't guarantee we can land the removal PR at the specific version.

_tools/check_deprecation.ts
```diff
@@ -0,0 +1,105 @@
...
+  ".git",
+  "node/_module",
+  "dotenv/testdata",
+  "fs/testdata",
+  "http/testdata",
+  "node/_module/cjs",
+  "node/_module/node_modules",
+  "node/_tools",
+  "node/testdata",
+];
```
kt3k, line 17: I think we should add `fs/exists.ts` to `EXCLUDED_PATHS` as there's some controversies about its deprecation.

timreichen:
> * I think we should exclude the entire `/node/` directory as we can't control when to remove those.

Agreed.
> * I prefer the expression `will be removed after X.Y.Z` as we can't guarantee we can land the removal PR at the specific version.

Why not? The point is that the lint test will fail if expired deprecations are not removed and therefore std cannot be published until that point. And when in doubt the expired version can always be increased.

_tools/check_deprecation.ts
timreichen: Ok, I'll put in on the list and log a warning, so it is not forgotten about.

kt3k:
> Why not? The point is that the lint test will fail if expired deprecations are not removed and therefore std cannot be published until that point. And when in doubt the expired version can always be increased.

std release is a part of a new CLI version release and there are already many tasks in it https://gist.github.com/denobot/a449a1bb5fa21d90eec855760c31aef5
and the release manager might be unfamiliar with std. So we don't force them to remove those APIs during the release.

And also some of deprecated APIs are not trivial to remove (for example, removal of the 3rd overload of `assertThrows` might be tricky)

How about showing warning in console instead of failing the CI if the version is expired?

timreichen:
I rather would have the `after` keyword and log a warning if the deprecation is expired and the test failing if a deprecation version is missing. But I am worried that it won't make any difference from the behavior we have now, since it only gives information on until what point it **won't** be removed as opposed to until when it **will**.

Maybe we should have both keywords `in` and `after` to fail the test if it is an explicit version and log a warning if it is *after* for certain deprecations?

iuioiua:
I like this idea! I was also thinking that something like this would be cool. It might be worth mentioning this new standard of deprecation in the contributing documentation.

timreichen:
@kt3k I added the `after` keyword and changed all deprecations to `after`.

So it is possible atm to use either
- `@deprecated (will be removed in X.Y.Z)` which will throw an error if version <= the std version

or

- `@deprecated (will be removed after X.Y.Z)` whicht will log a warning

I would suggest though to enforce the usage of `in` where ever possible.

kt3k (approved):
LGTM. Thanks!

Issue #1033, opened 2021-07-13
Check Deploy compatibility of some modules in CI


kt3k:
We started checking browser compatibility of some of std modules in CI in #995. We can do similar checks for Deploy in CI.

kt3k:
Closed in #2615

PR #2614, opened 2022-09-05
fix(node/process): revert workaround for stdin.isTTY (#2590)

kt3k:
This reverts commit 336da79918798704e2b4dff4616de5d117cf4d63.

The root issue (https://github.com/denoland/deno/issues/15708) has been resolved by https://github.com/denoland/deno/pull/15747

PR #2615, opened 2022-09-05
chore: test compatibility of std/node with Deno Deploy

kt3k:


kt3k:
The `deploy-compat` status check fails now because of the current incompatibility of `std/node`.

The status should be fixed when #2614 is landed (which is currently blocked by unavailability of windows runners)

iuioiua:
@kt3k, does this close #1033?

kt3k:
Yes, thanks!

PR #2616, opened 2022-09-05
BREAKING(dotenv): rename config to load

timreichen:
Remake of https://github.com/denoland/deno_std/pull/2416 as it was closed by accident. 

- deprecates config
- depreactes configSync
- deprecates ConfigOptions
- adds load and loadSync which work the same as config and configSync but take LoadOptions
- adds LoadOptions which renames path to envPath, defaults to defaultsPath and example to examplePath for more clarity.
- removes the safe option as it will always use safe mode except when examplePath is absent.


timreichen:
@kt3k ready for review

kt3k (approved):
The community looks supportive to this change (regarding Discord reactions). I'm also in favor of this change. `load` sounds more appropriate word for this functionality than `config`.

Thanks for your suggestion!

LGTM

PR #2617, opened 2022-09-05
feat(encoding/jsdoc): parse and stringify

timreichen:
- adds jsdoc `parse` and `stringify` functions


kitsonk:
I'm not supportive of this. JSDoc isn't really an encoding thing, like JSON or CSV. It is decoration on an AST and deno_doc handles it as part of an AST. Code generation off of an AST is totally different than a "stringify" of encoding too.

luk3skyw4lker:
I agree with the upper comment, I don't think that this should be placed at the encoding folder too because like @kitsonk said, JSDoc isn't in fact related with encoding

timreichen:
@kitsonk are you not in favour of the module as a whole or just current location?

kitsonk:
As a whole. deno_doc is where such parsing functionality should exist and codegen is a whole other kettle of fish.

timreichen:
@kitsonk Ok, I think deno_doc could be used for that purpose as well as deno_std/tools.
But could [doc()](https://github.com/denoland/deno_doc/blob/8ec795ba28061390775fc57e691e1230f2a11978/mod.ts#L74) add support for file paths next to urls? It works with `file://`, so I guess local paths shouldn't be a problem?

PR #2618, opened 2022-09-05
move(wasm): varint and crypto

timreichen:
- moves wasm directories for `varint` and `crypto ` into corresponding folders instead of existing on the top level.
- relocates `varint` logic and tests into `encoding/varint` and re-exports functions in `encoding/varint.ts`.


kt3k:
We also need to update the below lines in `.github/workflows/ci.yml` (detection of changes in wasm source):

```
declare modifications="$(git diff --name-only HEAD~ -- ./_wasm_crypto)"
...
declare modifications="$(git diff --name-only HEAD~ -- ./_wasm_varint)"
```

kt3k (approved):
LGTM. Thanks!

PR #2623, opened 2022-09-07
ci: check deploy-compat only with main

kt3k:
This PR disables the `deploy-compat` check for PRs. `denoland/deployctl` only works for origin branches. The check currently doesn't work for PRs from forks.


bartlomieju (approved):
LGTM

PR #2627, opened 2022-09-07
remove: outdated std import

iuioiua:
[This line](https://github.com/denoland/deno_std/blob/d3b14b2aa06db78dd642ba816a6388e88d5008e2/test_import_map.json#L4) doesn't seem right. This PR removes it.


kt3k (approved):
LGTM

PR #1590, opened 2021-11-16
ci: disable doctests for node/ directory

bartlomieju:
Fix #1589

PR #2629, opened 2022-09-07
remove: `Doctest /node` step in CI

iuioiua:
If this test is still desired, we can work towards bringing it back. Otherwise, reasons for removing:
1. There aren't many markdown files in `std/node` anyway.
2. The step uses a `strict-ts44.tsconfig.json` config file, which doesn't exist.
3. The comment hasn't been actioned since late 2021 (#1590).

https://github.com/denoland/deno_std/blob/d3b14b2aa06db78dd642ba816a6388e88d5008e2/.github/workflows/ci.yml#L100-L102

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,10 +97,6 @@ jobs:
         shell: bash
         run: deno task node:check-unstable-api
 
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Let's remove as we removed `--compat` in CLI

LGTM

PR #2634, opened 2022-09-08
fix(node/fs): make ReadStream and WriteStream callable with or without new

### LINKED ISSUE #2548: node - fs.ReadStream - Should work as function in addition to class
I ran this test in the CLI, which does `deno run npm:cpy-cli@4.1.0 source.txt destination`:

```rs
#[test]
fn deno_run_esm_module() {
  let _server = http_server();

  let deno_dir = util::new_deno_dir();

  deno_dir.write("source.txt", "test");

  let deno = util::deno_cmd_with_deno_dir(&deno_dir)
    .current_dir(deno_dir.path())
    .arg("run")
    .arg("--unstable")
    .arg("--allow-read")
    .arg("--allow-env")
    .arg("--allow-write")
    .arg("npm:cpy-cli@4.1.0")
    .arg("source.txt")
    .arg("destination")
    .env("NO_COLOR", "1")
    .envs(env_vars())
    .spawn()
    .unwrap();
  let output = deno.wait_with_output().unwrap();
  assert!(output.status.success());

  assert!(deno_dir.path().join("source.txt").exists());
  assert_eq!(deno_dir.read_to_string("destination.txt"), "test");
}
```

And I got:

```
Not implemented: process.on("unhandledRejection")
Cannot copy from `source.txt` to `C:\Users\david\AppData\Local\Temp\deno-cli-test-1661208962749-0\destination\source.txt`: Class constructor ReadStream cannot be invoked without 'new'
```

It seems we use a class for `ReadStream`, but it looks like it works as a function too.

PolarETech:
Fixes #2548


PolarETech:
Another type error was found. I have made an additional commit to fix it.

node/_fs/_fs_streams.ts
```diff
@@ -1,55 +1,89 @@
...
+
+  return this as ReadStream;
 }
 
+Object.setPrototypeOf(ReadStream.prototype, NodeReadable.prototype);
+
 export function createReadStream(
   path: string | URL,
-  options?: ReadStreamOptions,
+  opts?: ReadStreamOptions,
```
cjihrig, line 84: Why is this renamed? It looks like it is called `options` in the Node codebase and docs.

node/_fs/_fs_streams_test.ts
```diff
@@ -0,0 +1,203 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { createReadStream, ReadStream } from "./_fs_streams.ts";
+import * as path from "../../path/mod.ts";
+import { assertEquals } from "../../testing/asserts.ts";
+
+const moduleDir = path.dirname(path.fromFileUrl(import.meta.url));
+const testData = path.resolve(moduleDir, "testdata", "hello.txt");
+
+Deno.test({
```
cjihrig, line 9: Instead of adding a bunch of tests in this PR, is it possible to import more tests from the Node test suite? Ideally, we should rely on the Node test suite for Node compat and add tests for the things that the Node test suite might not cover.

node/internal/fs/streams_test.ts
```diff
@@ -0,0 +1,173 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { createWriteStream, WriteStream } from "./streams.ts";
+import { assertEquals } from "../../../testing/asserts.ts";
+
+Deno.test({
```
cjihrig, line 5: Same comment in this file - is it possible to leverage the Node test suite more?

cjihrig (commented):
Left a few comments. Ideally, I'd like to see us trying to leverage the Node test suite more here.

node/_fs/_fs_streams.ts
PolarETech: The original `ReadStream`, `WriteStream`, and `createWriteStream` arguments were all `opts`.
It seemed odd that only the `createReadStream` argument was `options`, so I changed it.
I agree that `options` would be better, so I will change all of them.

node/internal/fs/streams.ts
```diff
@@ -26,89 +31,131 @@ interface FS {
...
+    highWaterMark: opts?.highWaterMark,
+    decodeStrings: opts?.decodeStrings,
+    defaultEncoding: opts?.defaultEncoding,
+    objectMode: opts?.objectMode,
+    emitClose: opts?.emitClose,
+    autoDestroy: opts?.autoClose ?? opts?.autoDestroy,
+    signal: opts?.signal,
+  });
+
+  (this as WriteStream).fd = null;
```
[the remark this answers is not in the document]
PolarETech: You are right.
I will modify `this as WriteStream` and `this as ReadStream` like the code you provided.

node/_fs/_fs_streams_test.ts
PolarETech: Sure. Please give me some time to see which Node test suite can be imported.

PolarETech:
Hi @cjihrig
Thank you for reviewing.

Regarding importing tests from the Node test suite more, 
I have confirmed the following tests are passed.

- parallel/test-file-write-stream4.js
- parallel/test-fs-read-stream-autoClose.js
- parallel/test-fs-read-stream-concurrent-reads.js (requires fixtures/loop.js)
- parallel/test-fs-write-stream-end.js

However, it seems these tests were passed before this PR.
Is it appropriate to include the addition of these tests in this PR?

Unit test cases that have a duplicate perspective with the Node test suite have been removed.

node/_fs/_fs_streams.ts
PolarETech: I have modified these in https://github.com/denoland/deno_std/pull/2634/commits/c12f3f1cd5a3d0039e991b86c952a5075ddecf8d

node/internal/fs/streams.ts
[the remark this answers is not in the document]
PolarETech: I have modified these in https://github.com/denoland/deno_std/pull/2634/commits/d786bee5a21e6ee42cc3b65cee4e2bfc8647a174

node/_fs/_fs_streams_test.ts
PolarETech: Duplicate Unit test cases have been reorganized in https://github.com/denoland/deno_std/pull/2634/commits/67266461986bf0938bb03c18a5a777e1ab010b49 and https://github.com/denoland/deno_std/pull/2634/commits/a7a852cdcaa8adc9a34312ea3c8df7e2e752f464.
Regarding the Node test suite, please see the separate comment added below.

node/internal/fs/streams_test.ts
PolarETech: As with the other comment, duplicate tests have been reorganized in https://github.com/denoland/deno_std/pull/2634/commits/67266461986bf0938bb03c18a5a777e1ab010b49 and https://github.com/denoland/deno_std/pull/2634/commits/a7a852cdcaa8adc9a34312ea3c8df7e2e752f464.
Regarding the Node test suite, please see the separate comment added below.

PolarETech:
ci / lint failed.
I have checked and it seems to be caused by upgrading Deno to 1.25.2.
There are no lint errors in the files changed by this PR.

node/_fs/_fs_streams_test.ts
PolarETech: Updated the comment regarding the Node test suite. https://github.com/denoland/deno_std/pull/2634#issuecomment-1241522250

cjihrig (commented):
The changes here look ok, and since the CI passes, I'd be ok with merging this. However, one thing I noticed while reviewing is that the implementation looks a good bit different from Node's: https://github.com/nodejs/node/blob/main/lib/internal/fs/streams.js. I understand that we don't have 100% feature parity with Node at this point, but even considering that, I think ongoing maintenance will be easier if we try to match the Node implementation more.

@kt3k @bartlomieju what do you think?

bartlomieju:
@cjihrig I agree with that opinion. We should try to match it as closely as possible

kt3k:
@cjihrig I agree we should match the implementation to node.js as far as possible. But also I think this PR is in good enough state/has enough value to land at this point as it enables some modules like `cpy-cli`, `copy-file`, etc. I'd suggest we should iterate on further compatibility improvement in later PRs

kt3k (approved):
LGTM

cjihrig (approved):

PR #2649, opened 2022-09-13
test(node/fs): add compat tests for ReadStream and WriteStream

PolarETech:
I would like to add the tests that have already been passed, as reported in https://github.com/denoland/deno_std/pull/2634#issuecomment-1241522250.

- parallel/test-file-write-stream4.js
- parallel/test-fs-read-stream-autoClose.js
- parallel/test-fs-read-stream-concurrent-reads.js (requires fixtures/loop.js)
- parallel/test-fs-write-stream-end.js

For each test, I have confirmed the related features have already been implemented in the modules accordingly.


cjihrig (approved):
LGTM. Thanks!

PR #2653, opened 2022-09-14
fix(node/fs): improve compatibility of fs.ReadStream

PolarETech:
This is in regards to the comment on https://github.com/denoland/deno_std/pull/2634#pullrequestreview-1104619333.

## Scope
This PR aims to address the followings:

1. Make the implementation closer to Node
2. Support `fd`, `start`, `end`, and `fs` options
3. Pass the following tests

- [x] parallel/test-fs-empty-readStream.js
- [x] parallel/test-fs-read-stream-double-close.js
- [x] parallel/test-fs-read-stream-encoding.js
- [x] parallel/test-fs-read-stream-fd.js
- [x] parallel/test-fs-read-stream-inherit.js (*1)
- [x] parallel/test-fs-read-stream-patch-open.js
- [x] parallel/test-fs-read-stream-resume.js
- [x] parallel/test-fs-read-stream-throw-type-error.js

## Out of Scope
After examination, I'd like to transfer the following test-related implementations to later Issues and PRs, as they are difficult to address or require a large implementation outside of ReadStream.

- parallel/test-fs-read-stream-err.js (*1)
   - Overridden `fs.read/close` must be executed in the ReadStream. However, the default import of `fs` in the ReadStream module leads to circular dependencies, so it cannot be implemented in the same way as Node ([9305a95](https://github.com/denoland/deno_std/pull/2653/commits/9305a95700d628dff9878caf36df9d39842d20c9)). _Any advice on how to resolve this would be appreciated._
- parallel/test-fs-read-stream-fd-leak.js
   - Same as above. `fs.open/close` are overridden within the test.
- parallel/test-fs-read-stream-file-handle.js
   - Requires `FileHandle` class for `fs.promises.open`. ([Ref.](https://github.com/nodejs/node/blob/main/lib/internal/fs/promises.js))
- parallel/test-fs-read-stream-pos.js (*2)
   - Performance issue with fs.writeFileSync() causing a false positive.
- parallel/test-fs-read-stream.js (*1)
   - ReadStream module implementation is complete.
   - `child_process.spawnSync` is required within the test. (Waiting for PR 2637) 
   - `child_process.exec` is required within the test. (PR is in preparation)

## References

## Remarks
_(*1)_
"bufferSize" option is present in the following tests.

- parallel/test-fs-read-stream-err.js
- parallel/test-fs-read-stream-inherit.js
- parallel/test-fs-read-stream.js

The option was removed in Node v0.9.12, and the current behavior of Node's ReadStream is to ignore it.
Therefore, "bufferSize" option has not been implemented, but consider it fine.

_(*2)_
The following test repeatedly reads data for 90 seconds.

- parallel/test-fs-read-stream-pos.js

If no errors occur during the read, the assert is never triggered, and it seems to be the intent of the test.


node/_tools/test/parallel/test-fs-read-stream-pos.js
```diff
@@ -37,6 +39,7 @@ let stream;
 
 const readInterval = setInterval(() => {
   if (stream) return;
+  if (!appended) return;
 
   stream = fs.createReadStream(file, {
     highWaterMark: hwm,
```
PolarETech, line 45: This test was failing about 10 / 30 of the time, but it seems to be a false positive.
If `createReadStream()` is executed before the append to the test file is completed, the "data" event will not be called and `common.mustCallAtLeast()` verification will fail. 
It is most likely an I/O performance issue, not ReadStream. The performance of `fs.writeFileSync()` in Deno is about 5 or 6 times slower than in Node. (Confirmed with Deno 1.25.3 and Node.js v18.9.0 on my laptop)

To avoid the issue, I have modified the test to make sure that the append to the test file is complete before executing `createReadStream()`.

node/_tools/test/parallel/test-fs-read-stream-pos.js
```diff
@@ -79,12 +83,17 @@ const exitTest = () => {
...
+    } else {
+      assert(cur);
       process.exit();
-    });
-    stream.destroy();
-  } else {
-    process.exit();
-  }
+    }
+  }, 20);
```
PolarETech, line 98: A small wait time has been added because common.mustCallAtLeast() validation may also fail if stream.destroy() is called and the stream is terminated in the middle.
And `assert(cur)` has been added to detect a case where the test terminates without ever reading any data as a failure.

PolarETech:
As commented in [#discussion_r972879858](https://github.com/denoland/deno_std/pull/2653/#discussion_r972879858) and [#discussion_r972880948](https://github.com/denoland/deno_std/pull/2653/#discussion_r972880948), I have modified the contents of test-fs-read-stream-pos.js.
If there are any problems, please let me know.

There is one error in ci / node, but it appears to be unrelated to the changes in this PR.

node/_tools/config.json
```diff
@@ -2,7 +2,12 @@
   "nodeVersion": "18.8.0",
   "ignore": {
     "common": ["index.js", "tmpdir.js"],
-    "fixtures": ["child-process-spawn-node.js", "empty.txt"],
+    "fixtures": [
+      "child-process-spawn-node.js",
+      "elipses.txt",
+      "empty.txt",
+      "x.txt"
```
cjihrig, line 9: Sorry if I'm missing something, but why do these fixtures need to be ignored?

node/_tools/config.json
```diff
@@ -33,6 +38,7 @@
       "test-fs-append-file.js",
       "test-fs-chmod-mask.js",
       "test-fs-chmod.js",
+      "test-fs-read-stream-pos.js",
```
cjihrig, line 41: I'm thinking we should just not include this test for the time being. We can revisit it when the underlying issues are addressed.

cjihrig (commented):
The changes mostly LGTM, and the CI is green. Would you mind backing out the test I commented on?

node/_tools/config.json
PolarETech: Some tests will fail due to the headers that are automatically added when extracting `elipses.txt` and `x.txt` by `deno task node:setup`.
That is why I have added them to the ignore section, to keep the files without the headers.

cjihrig: Got it. Thanks. I think we should not modify fixtures (not something for you to address here, just a general comment).

node/_tools/config.json
PolarETech: I understand your intention. The test has been removed. e459b0b

cjihrig (approved):

PR #2654, opened 2022-09-14
fix(node/process): reduce required env permission range

kt3k:
This PR reduces the required env permission range for `x in process.env` and `process.env.hasOwnProperty(x)`.


ry (approved):
LGTM

PR #2655, opened 2022-09-14
fix(node): use `Buffer.from()` and `Buffer.alloc()` instead of `new Buffer()`

iuioiua:
`new Buffer()` is deprecated in Node-land in-favour of `Buffer.from()`. This is done in good practice, although this may not be functionally different in Deno. Tests have been ignored.


node/querystring.ts
```diff
@@ -433,7 +433,7 @@ const unhexTable = new Int8Array([
  * A safe fast alternative to decodeURIComponent
  */
 export function unescapeBuffer(s: string, decodeSpaces = false): Buffer {
-  const out = new Buffer(s.length);
+  const out = Buffer.from(s);
```
kt3k, line 436: This type of Buffer.from call seems causing TypeError.

```
> Buffer.from(10)
Uncaught:
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received type number (10)
```

cjihrig: @kt3k that matches the behavior in Node.

node/integrationtest/yarn.js
```diff
@@ -19583,7 +19583,7 @@ $exports.store = store;
      * @returns {boolean} Returns `true` if `value` is a buffer, else `false`.
      * @example
      *
-     * _.isBuffer(new Buffer(2));
+     * _.isBuffer(Buffer.from(2));
```
cjihrig, line 19586: This is just a comment, but this would throw.

node/integrationtest/yarn.js
```diff
@@ -102198,7 +102198,7 @@ var hasFullSupport = (function () {
       return false
     }
 
-    var buf = Buffer.alloc ? Buffer.alloc(4) : new Buffer(4)
+    var buf = Buffer.alloc ? Buffer.alloc(4) : Buffer.from(4)
```
cjihrig, line 102201: This will throw.

node/integrationtest/yarn.js
```diff
@@ -102375,7 +102375,7 @@ function bufferFrom (value, encodingOrOffset, length) {
 
   return isModern
     ? Buffer.from(value)
-    : new Buffer(value)
+    : Buffer.from(value)
```
cjihrig, line 102378: It looks like both sides of the ternary are the same now. Based on the `isModern` variable, my guess is that this change should be backed out.

node/integrationtest/yarn.js
```diff
@@ -124028,7 +124028,7 @@ module.exports = function (iconv) {
     // Node authors rewrote Buffer internals to make it compatible with
     // Uint8Array and we cannot patch key functions since then.
     // Note: this does use older Buffer API on a purpose
-    iconv.supportsNodeEncodingsExtension = !(Buffer.from || new Buffer(0) instanceof Uint8Array);
+    iconv.supportsNodeEncodingsExtension = !(Buffer.from || Buffer.from(0) instanceof Uint8Array);
```
cjihrig, line 124031: This will throw.

node/integrationtest/yarn.js
```diff
@@ -131284,7 +131284,7 @@ function isArrayLikeObject(value) {
  * @returns {boolean} Returns `true` if `value` is a buffer, else `false`.
  * @example
  *
- * _.isBuffer(new Buffer(2));
+ * _.isBuffer(Buffer.from(2));
```
cjihrig, line 131287: Again, this is a comment, but the code would throw.

node/querystring.ts
iuioiua: I replaced `Buffer.from(<integer>)` with `Buffer.alloc(<integer>)`.

node/integrationtest/yarn.js
iuioiua: I now replaced `Buffer.from(<integer>)` with `Buffer.alloc(<integer>)`.

node/integrationtest/yarn.js
iuioiua: Thanks. I've reverted this.

node/integrationtest/yarn.js
iuioiua: Thanks, reverted.

node/integrationtest/yarn.js
iuioiua: I now replaced `Buffer.from(<integer>)` with `Buffer.alloc(<integer>)`.

node/integrationtest/yarn.js
iuioiua: I now replaced Buffer.from(<integer>) with Buffer.alloc(<integer>).

iuioiua:
A network interfaces check in `parallel/test-os` is now failing. I don't think any of my changes had anything to do with this.

node/integrationtest/yarn.js
```diff
@@ -102180,7 +102180,7 @@ function allocUnsafe (size) {
   if (Buffer.allocUnsafe) {
     return Buffer.allocUnsafe(size)
   } else {
-    return new Buffer(size)
+    return Buffer.from(size)
```
kt3k, line 102183: This also should be Buffer.alloc?

iuioiua: Yep! Fixed.

kt3k (approved):
LGTM

node/integrationtest/yarn.js
cjihrig: I wonder if this should be left as it was. It looks like this was a fallback for the case where `Buffer.allocUnsafe()` did not exist. `Buffer.allocUnsafe()` and `Buffer.alloc()` were added in the same version of Node, so this might not be a good fallback. That said, those methods were added in Node 5, which has been end-of-life for years now, so it probably doesn't matter. 🤷

cjihrig (approved):
LGTM with one small comment.

node/integrationtest/yarn.js
iuioiua: I agree. Let's just match Node for now.

kt3k:
@iuioiua Thanks for your contribution! Nice work!

Issue #2557, opened 2022-08-24
Not implemented: crypto.Hmac error with Jsonwebtoken


penberg:
**Describe the bug**

I am trying to use jsonwebtoken with Deno's compatibility mode, but I am seeing the following error:

```console
Uncaught Error: Not implemented: crypto.Hmac
```

**Steps to Reproduce**

With the following webtoken.mjs file:

```javascript
import jwt from 'jsonwebtoken'

const token = jwt.sign({ foo: 'bar' }, 'shhhhh');
console.log(token);
```

`deno run` (in compat mode) fails as follows:

```console
% deno run --unstable --allow-env --allow-read --compat webtoken.mjs
error: Uncaught Error: Not implemented: crypto.Hmac
  throw new Error(message);
        ^
    at notImplemented (https://deno.land/std@0.152.0/node/_utils.ts:23:9)
    at new Hmac (https://deno.land/std@0.152.0/node/internal/crypto/hash.ts:136:5)
    at Object.createHmac (https://deno.land/std@0.152.0/node/crypto.ts:264:10)
    at Object.sign (file:///Users/penberg/src/chiselstrike/webtoken/node_modules/jwa/index.js:132:23)
    at Object.jwsSign [as sign] (file:///Users/penberg/src/chiselstrike/webtoken/node_modules/jws/lib/sign-stream.js:32:24)
    at Object.module.exports [as sign] (file:///Users/penberg/src/chiselstrike/webtoken/node_modules/jsonwebtoken/sign.js:204:16)
    at file:///Users/penberg/src/chiselstrike/webtoken/webtoken.mjs:3:19
```

**Expected behavior**

```console
% node webtoken.mjs
eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJmb28iOiJiYXIiLCJpYXQiOjE2NjEzMzIwOTR9.dmuExDy4XoLwze8PDox3xRH7UCRklkFbrbhLLXIaNrQ
```

**Environment**

- OS: MacOS 12.4
- deno version: 1.24.3
- std version: 0.152.0

luk3skyw4lker:
I'm taking a stab at this one!

penberg:
Looks like there's some more unimplemented bits. As pointed out by @ricardodalarme, Octokit seems to use `jsonwebtoken` with `RS256` algorith, which hits a different unimplemented path in `crypto`:

```
% cat webtoken.mjs
import jwt from 'jsonwebtoken'

[...]

PR #2664, opened 2022-09-15
feat: add hmac implementation

luk3skyw4lker:
Adding HMAC implementation according to [crypto-browserify](https://github.com/crypto-browserify/createHmac/blob/master/browser.js)

Addressing #2557


cjihrig:
Thanks for the PR. It looks like the CI is not happy with the current changes. Also, I think we would want to vendor the appropriate tests from the Node test suite for this.

cjihrig:
Ping @luk3skyw4lker. Are you planning to revisit this?

luk3skyw4lker:
@cjihrig Yes I am! I'm just a little caght up in some work stuff, I'll try to make some progress this weekend

luk3skyw4lker:
@cjihrig about the tests, I'll take a look at the node tests suite and add the tests there

luk3skyw4lker:
@cjihrig @kt3k @bartlomieju I am a little confused on how we should test this implementation, should I just create a test that validates the creation of the HMAC and check the consistency of the output (since it's a hash, same input should produce the same output) or something else? I can't think of a lot of options

cjihrig:
@luk3skyw4lker were you able to identify tests in Node's suite for this functionality? If so, you can update [this file](https://github.com/denoland/deno_std/blob/main/node/_tools/config.json) with those tests, and then run `deno run --allow-read --allow-net --allow-write node/_tools/setup.ts` to refresh the test suite.

luk3skyw4lker:
@cjihrig I didn't found the test suite in the deno_std repository, but I did found a test for it in the Node repository, so I should just copy the test file and update this file that you mentioned and everything should be done (assuming that the tests are approved)?

cjihrig:
You don't need to copy the test files. If you update the linked config file and regenerate the test suite, it will pull in the tests that you add.

luk3skyw4lker:
@cjihrig Alright! Thanks for the help!

luk3skyw4lker:
@cjihrig FYI: I am currently working into fixing the broken node tests, I have found some inconsistencies in the code :x

luk3skyw4lker:
@cjihrig you guys can take it over if needed, I'm pretty short on time to solve this, I wish I could do more, but it isn't possible for me :/

cjihrig (approved):
Nice work

PR #2669, opened 2022-09-18
fix(node/fs): position option of fs.read and fs.readSync works the same as Node

PolarETech:
The behavior of `position` option in `fs.read` and `fs.readSync` is different between Deno and Node.
This PR fixes the behavior of Deno to be the same as Node's.

### Node v18.9.0

If `position` is an integer, 

- Specifies the number of bytes to start reading counting from the ***beginning*** of the file
- The file position will be ***unchanged***

Ref: https://nodejs.org/api/fs.html#fsreadfd-buffer-offset-length-position-callback

### Deno 1.25.3 with std@0.156.0

If `position` is an integer, 

- Specifies the number of bytes to start reading counting from the ***current position*** of the file
- The file position will be ***updated***

## Sample code

```mjs
// import for Node
import fs from 'fs';
import { Buffer } from 'node:buffer';
// or
// import for Deno
import fs from "https://deno.land/std@0.156.0/node/fs.ts";
import { Buffer } from "https://deno.land/std@0.156.0/node/buffer.ts";


const file = "./test.txt";
const text = "hello, world!";
const bufSize = 3;
const positions = [2, null, 2];
const data = [];

fs.writeFileSync(file, text);
const fd = fs.openSync(file);

for (const pos of positions) {
  const buf = Buffer.alloc(bufSize);
  fs.readSync(fd, buf, 0, buf.byteLength, pos);
  data.push(buf.toString("utf8"));
}

fs.closeSync(fd);
console.log(data);
```

### Execution result in Node

```mjs
[ 'llo', 'hel', 'llo' ] // [ 2-4 byte, 0-2 byte, 2-4 byte ]
```

### Execution result in Deno

```mjs
[ "llo", ", w", "ld!" ] // [ 2-4 byte, 5-7 byte, 10-12 byte ]
```

## Tests

No applicable test for this issue has been found in the Node test suite. Unit tests have been updated.

## Remarks

This issue has been found while implementing compatibility improvements to fs.ReadStream. #2653

### DIFF
--- a/node/_fs/_fs_read.ts
+++ b/node/_fs/_fs_read.ts
@@ -98,20 +98,20 @@ export function read(
   let err: Error | null = null,
     numberOfBytesRead: number | null = null;
 
-  let currentPosition = 0;
-  if (typeof position === "number" && position >= 0) {
-    currentPosition = Deno.seekSync(fd, 0, Deno.SeekMode.Current);
-    Deno.seekSync(fd, position, Deno.SeekMode.Start);
-  }
-
   try {
+    let currentPosition = 0;
+    if (typeof position === "number" && position >= 0) {
+      currentPosition = Deno.seekSync(fd, 0, Deno.SeekMode.Current);
+      Deno.seekSync(fd, position, Deno.SeekMode.Start);
+    }
+
     numberOfBytesRead = Deno.readSync(fd, buffer);
+
+    if (typeof position === "number" && position >= 0) {
+      Deno.seekSync(fd, currentPosition, Deno.SeekMode.Start);
+    }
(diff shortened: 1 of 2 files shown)


PolarETech:
There are two ci / node failures, but they appear to be issues outside the scope of this PR.

PolarETech:
There was a lack of consideration regarding error handling.
A fix commit has been added.

kt3k (approved):
LGTM. Good catch!

kt3k:
The CI failures should be fixed by https://github.com/denoland/deno/pull/15954

cjihrig (approved):

PR #2670, opened 2022-09-19
debug test-os.js test failures

kt3k:


kt3k:
ipv4 addresses are in backward order?

```
lo: [
    {
      address: "1.0.0.127",
      netmask: "0.0.0.255",
      family: "IPv4",
      mac: "00:00:00:00:00:00",
      internal: false,
      cidr: "1.0.0.127/8"
    },
```

https://github.com/denoland/deno_std/actions/runs/3080237838/jobs/4977350369#step:6:667

probably affected by https://github.com/denoland/deno/pull/15914

PR #2675, opened 2022-09-20
fix(http/file_server): handles path with reserved char

blmarket:
Current implementation don't escape some reserved characters, which causing the browser to consider certain part is not the file path.

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -194,6 +194,27 @@ Deno.test(
   },
 );
 
+Deno.test(
+  "file_server serveFile with filename including hash symbol",
+  async () => {
+    await startFileServer({ target: "./testdata" });
+    try {
+      const res = await fetch("http://localhost:4507/file%232.txt");
+      assertEquals(
+        res.headers.get("content-type"),
+        "text/plain; charset=UTF-8",
+      );
+      const downloadedFile = await res.text();
+      const localFile = new TextDecoder().decode(
+        await Deno.readFile(join(testdataDir, "file#2.txt")),
+      );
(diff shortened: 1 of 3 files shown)


blmarket:
This commit improves the path handling by escaping more characters using encodeURIComponent instead of encodeURI, with exception of '/'.

Honestly speaking, this behavior was referenced from Python [urllib.parse.quote](https://github.com/python/cpython/blob/21b5af9072a43275d52737a68d5cda2fab47f730/Lib/urllib/parse.py#L818) which was used by Python `http.server`

http/file_server_test.ts
```diff
@@ -284,8 +317,8 @@ Deno.test("checkURIEncodedPathTraversal", async function () {
       "http://localhost:4507/%2F..%2F..%2F..%2F..%2F..%2F..%2F..%2F..",
     );
 
-    assertEquals(res.status, 404);
-    const _ = await res.text();
+    assertEquals(res.status, 200);
+    assertStringIncludes(await res.text(), "README.md");
```
blmarket, line 321: Correct me if I'm wrong, but I think this test was not correct... or simply I don't understand what it tests.

kt3k: Looks like this was the test for a security fix for path traversal vulnerability in very old version of file_server https://github.com/denoland/deno/pull/8134/files

and that attack doesn't seems possible now (even with this fix) because we process the path with `new URL().pathname` and `posix.normalize(path)`

I think this change is ok

kt3k (approved):
LGTM

cjihrig (approved):

PR #2676, opened 2022-09-21
fix(node/net): fix socket events order

### LINKED ISSUE #2666: node/net: `close` and `error` events happen in wrong order in net.Socket
The below script prints events in `close` -> `error` order.
```ts
import { createRequire } from "https://deno.land/std@0.156.0/node/module.ts";

const require = createRequire(import.meta.url);

const net = require('net');

const socket = net.createConnection('27009', 'doesnotexist');
socket.once('error', err => console.log('Emitted error', err));
socket.once('close', err => console.log('Emitted close'));

console.log('Done');
```

But the equivalent node.js script:
```ts
const net = require('net');

const socket = net.createConnection('27009', 'doesnotexist');
socket.once('error', err => console.log('Emitted error', err));
socket.once('close', err => console.log('Emitted close'));

console.log('Done');
```
prints `error` -> `close` order.

This difference prevents mongoose working correctly in deno (ref: https://github.com/denoland/deno/issues/15824 )

kt3k:
When `net.connect` fails with an immediate error, `error` event happens before `close` event in node.js, but in `std/node` the events happen in the opposite order. This PR fixes it.

closes #2666
ref https://github.com/denoland/deno/issues/15824

### DIFF
--- a/node/internal_binding/handle_wrap.ts
+++ b/node/internal_binding/handle_wrap.ts
@@ -34,7 +34,7 @@ export class HandleWrap extends AsyncWrap {
 
   close(cb: () => void = () => {}) {
     this._onClose();
-    cb();
+    queueMicrotask(cb);
   }
 
   ref() {
--- a/node/net.ts
+++ b/node/net.ts
@@ -1420,11 +1420,10 @@ export class Socket extends Duplex {
         this._handle = null;
         this._sockname = undefined;
 
-        cb(exception);
-
         debug("emit close");
         this.emit("close", isException);
       });
+      cb(exception);
     } else {
       cb(exception);
       nextTick(_emitCloseNT, this);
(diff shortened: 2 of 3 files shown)


node/internal_binding/handle_wrap.ts
```diff
@@ -34,7 +34,7 @@ export class HandleWrap extends AsyncWrap {
 
   close(cb: () => void = () => {}) {
     this._onClose();
-    cb();
+    queueMicrotask(cb);
```
kt3k, line 37: I also tried some other patterns:
- `nextTick(cb)` didn't work (the events happen in `close` -> `error` order)
- `setTimeout(cb, 0)` worked

cjihrig (approved):
LGTM. This seems like the type of test that Node should have in its own suite 😄

Issue #2674, opened 2022-09-20
node: need a way to spawn process.execPath with no deno specific flags


cjihrig:
**Describe the bug**
Many Node.js tests spawn the node executable in a child process. This is separate from `child_process.fork()`. As an example, see [this test](https://github.com/nodejs/node/blob/96f072231d88ed3ed4a5422f2e37485b59f39e4d/test/parallel/test-child-process-spawnsync-env.js). The problem in that test is that in Deno's compat mode, the `spawnSync()` call needs to be updated to pass in necessary Deno CLI flags (`-A`, `--unstable`, etc.).

For users, this is a problem because they can't run existing Node code as is. For us, it's also a problem because by modifying the Node tests, we lose the ability to automatically upgrade them to newer versions of Node.

A few ideas on how to approach this:
- In the compat code, detect when we are spawning the current executable and set an environment variable that allows Deno to start up properly with no CLI flags. This would require changes in core.
- In the compat code, detect when we are spawning the current executable and add the CLI flags. This could be done in std alone, but might have some edge cases.

**Steps to Reproduce**

Spawn `process.execPath` in Node compat mode without changing any of the CLI flags.

**Expected behavior**

Node compat code runs without modifications.

**Environment**

- OS: all
- deno version: 1.25.3
- std version: 0.156.0

kt3k:
> In the compat code, detect when we are spawning the current executable and set an environment variable that allows Deno to start up properly with no CLI flags. This would require changes in core.
In the compat code, detect when we are spawning the current executable and add the CLI flags. This could be done in std alone, but might have some edge cases.

The first solution might be ideal, but sounds adding lot of complexity to CLI entrypoint.

I think we should try the 2nd solution first and see how many cases it can cover.

cc @bartlomieju what do you think?

cmoog:
I also encountered this here: https://github.com/denoland/deno/issues/15836

bartlomieju:
> > In the compat code, detect when we are spawning the current executable and set an environment variable that allows Deno to start up properly with no CLI flags. This would require changes in core.
> > In the compat code, detect when we are spawning the current executable and add the CLI flags. This could be done in std alone, but might have some edge cases.
> 
> The first solution might be ideal, but sounds adding lot of complexity to CLI entrypoint.
> 
> I think we should try the 2nd solution first and see how many cases it can cover.
> 
> cc @bartlomieju what do you think?

I agree with that, we should go with 2nd solution for now.

[...]

PR #2678, opened 2022-09-21
chore(node): update test to be more like node

cjihrig:
This commit updates test-child-process-spawnsync-env.js to match the test from Node.js with two exceptions noted in a TODO comment. These TODOs should be addressable once Deno <--> Node CLI argument mapping is implemented and there is a way to start Deno with require() available.

Refs: https://github.com/denoland/deno_std/issues/2674


bartlomieju (approved):
LGTM

PR #2680, opened 2022-09-22
feat(node/process): `process.getuid()` and `process.getgid()`

iuioiua:
1. Is this the best way to conditionally define a method of a class? These methods don't exist on Windows platforms in Node.
2. How can `// @ts-ignore` be avoided in the tests?
3. [`test-process-uid-gid.js`](https://github.com/nodejs/node/blob/main/test/parallel/test-process-uid-gid.js) wasn't added because `process.setuid()`, `process.seteuid()`, `process.setgid()` and `process.setegid()` aren't and cannot currently be implemented, meaning the majority of that test would fail.

### DIFF
--- a/node/process.ts
+++ b/node/process.ts
@@ -560,18 +560,6 @@ class Process extends EventEmitter {
     return 0o22;
   }
 
-  /** https://nodejs.org/api/process.html#processgetuid */
-  getuid(): number {
-    // TODO(kt3k): return user id in mac and linux
-    return NaN;
-  }
-
-  /** https://nodejs.org/api/process.html#processgetgid */
-  getgid(): number {
-    // TODO(kt3k): return group id in mac and linux
-    return NaN;
-  }
-
   // TODO(kt3k): Implement this when we added -e option to node compat mode
   _eval: string | undefined = undefined;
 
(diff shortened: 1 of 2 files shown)


node/process.ts
```diff
@@ -560,18 +560,6 @@ class Process extends EventEmitter {
     return 0o22;
   }
 
-  /** https://nodejs.org/api/process.html#processgetuid */
-  getuid(): number {
```
kt3k, line 564: how about keeping only the typing like `getuid?: () => number;` and call it like `getuid?.()` in testing? (I guess that reduces the type errors in test cases)

Note: DefinitelyTyped defines this as `getuid?: () => number;` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/35b81ce6b7865843f9c2e49861c470621134ce3b/types/node/process.d.ts#L689

kt3k:
@iuioiua Thanks for looking into this! Left one comment

cjihrig:
I have to ask, are there any Node.js tests that can be ported over with these APIs?

iuioiua:
> I have to ask, are there any Node.js tests that can be ported over with these APIs?

See no. 3 in my opening comment.

cjihrig (approved):

PR #2686, opened 2022-09-22
fix(node/process): ignore SIGTERM on windows

### LINKED ISSUE #2677: node: Listening for SIGTERM should be a noop
Currently adding a listener for SIGTERM on windows will throw an error like so:
```
error when starting dev server:
TypeError: Windows only supports ctrl-c (SIGINT) and ctrl-break (SIGBREAK).
    at bindSignal (deno:runtime/js/40_signals.js:14:16)
    at Object.addSignalListener (deno:runtime/js/40_signals.js:54:21)
    at process.on (https://deno.land/std@0.156.0/node/process.ts:389:14)
    at process.once (https://deno.land/std@0.156.0/node/_events.mjs:525:8)
    at createServer (file:///V:/small_mammal_test/my-vue-app/node_modules/.deno/vite@3.1.3/node_modules/vite/dist/node/chunks/dep-557f29e6.js:62175:17)
    at async CAC.<anonymous> (file:///V:/small_mammal_test/my-vue-app/node_modules/.deno/vite@3.1.3/node_modules/vite/dist/node/cli.js:700:24)
```
but according to https://docs.libuv.org/en/v1.x/signal.html and @cjihrig's knowledge this signal is a noop on Windows. We should align `std/node/` to this behavior.

kt3k:
closes #2677

### DIFF
--- a/node/process_test.ts
+++ b/node/process_test.ts
@@ -148,6 +148,7 @@ Deno.test({
 
 Deno.test({
   name: "process.on signal",
+  ignore: Deno.build.os == "windows",
   async fn() {
     const promise = deferred();
     let c = 0;
@@ -172,6 +173,7 @@ Deno.test({
 
 Deno.test({
   name: "process.off signal",
+  ignore: Deno.build.os == "windows",
   async fn() {
     const promise = deferred();
     let c = 0;


bartlomieju (approved):
LGTM, thanks for looking into this!

PR #2688, opened 2022-09-22
fix(node): support mapping Node flags to Deno

cjihrig:
This commit adds initial support for mapping Node CLI flags to Deno CLI flags. This is done to better support the case where Node compat mode is used to spawn another Node subprocess (that is actually Deno in compat mode - not at all confusing).

Refs: https://github.com/denoland/deno_std/issues/2674


kt3k (approved):
Great! LGTM 👍

PR #2690, opened 2022-09-22
fix(node): `test-child-process-spawnsync-validation-errors.js`

iuioiua:
- [x] Circular dependencies need to be fixed between `/internal/util.mjs` and `/internal/errors.ts`
- [x] The test should pass once #2680 is implemented.


node/internal/errors.ts
```diff
@@ -164,6 +163,17 @@ export const uvExceptionWithHostPort = hideStackFrames(
   },
 );
 
+/** Copied from `../util.ts` to avoid a circular dependency. */
```
cjihrig, line 166: I don't think we should have multiple copies like this since the copies can get out of sync.

node/internal/child_process.ts
```diff
@@ -411,6 +446,173 @@ function normalizeStdioOption(
   }
 }
 
+export function normalizeSpawnArguments(
+  file: string,
+  args: string[],
+  options: SpawnSyncOptions,
```
cjihrig, line 452: FWIW, Node calls `normalizeSpawnArguments()` for both `spawn()` and `spawnSync()`. No need to change this now, but something to think about for the future.

node/internal/child_process.ts
```diff
@@ -411,6 +446,173 @@ function normalizeStdioOption(
...
+  let cwd = options.cwd;
+
+  // Validate the cwd, if present.
+  if (cwd != null) {
+    cwd = getValidatedPath(cwd, "options.cwd") as string;
+  }
+
+  // Validate detached, if present.
+  if (options.detached != null) {
+    validateBoolean(options.detached, "options.detached");
```
cjihrig, line 486: For the options that are not currently supported, I wonder if it makes sense to perform the validation so that the input is correct, but then immediately throw an exception saying that the option is unimplemented so that users don't get unexpected behavior.

node/internal/child_process.ts
```diff
@@ -411,6 +446,173 @@ function normalizeStdioOption(
...
+      }
+      // '/d /s /c' is used only for cmd.exe.
+      if (/^(?:.*\\)?cmd(?:\.exe)?$/i.exec(file) !== null) {
+        args = ["/d", "/s", "/c", `"${command}"`];
+        windowsVerbatimArguments = true;
+      } else {
+        args = ["-c", command];
+      }
+    } else {
+      /** Condition for Android platform is ignore in Deno */
```
cjihrig, line 548: Can you make this a `TODO` comment.

node/internal/child_process.ts
```diff
@@ -411,6 +446,173 @@ function normalizeStdioOption(
...
+  }
+
+  const env = options.env || Deno.env.toObject();
+  const envPairs: string[][] = [];
+
+  // process.env.NODE_V8_COVERAGE always propagates, making it possible to
+  // collect coverage for programs that spawn with white-listed environment.
+  copyProcessEnvToEnv(env, "NODE_V8_COVERAGE", options.env);
+
+  /** `isZOS` condition is ignored in Deno. */
```
cjihrig, line 571: Please make this a TODO.

node/internal/child_process.ts
```diff
@@ -411,6 +446,173 @@ function normalizeStdioOption(
...
+      ArrayPrototypePush(envPairs, `${key}=${value}`);
+    }
+  }
+
+  return {
+    // Make a shallow copy so we don't clobber the user's options object.
+    ...options,
+    args,
+    cwd,
+    envPairs,
```
cjihrig, line 609: Please include `detached` here even though it is unused.

cjihrig (commented):
Left a few comments - mostly minor things, but I don't think we should duplicate functions across the codebase.

node/internal/errors.ts
iuioiua: If imported, a circular dependency is introduced. What are some alternatives?

node/internal/child_process.ts
iuioiua: Agreed. I’ll do this in another PR (likely the one that fixes typings around these functions, as previously mentioned).

node/internal/errors.ts
cjihrig: I think you should be able to import [this file](https://github.com/denoland/deno_std/blob/main/node/internal/error_codes.ts) to get the error code this PR adds to `internal/util.mjs`.

iuioiua: Ok, I avoided using `getSystemErrorName` altogether as it's deprecated and instead used `uvErrmapGet`, as recommended [here](https://github.com/nodejs/node/blob/7e0097d8a33fa7adbc1f298cbf647f6d2fd403e8/lib/internal/errors.js#L628-L633). Funnily enough, even Node hasn't implemented this improvement yet.

iuioiua: So the initial issue was that importing `getSystemErrorName` from `internal/util.mjs` introduced the circular dependency. I've made the above change and now `parallel\test-dgram-recv-error.js` is failing on Windows because it uses an error number of `-1`, which isn't caught by `uvErrmapGet`, but was caught by `getSystemErrorName`.

I'm not sure what can be done. What do you think?

iuioiua: Ok, I've reverted my previous change. Now, `getSystemErrorName` has been moved from `node/util.ts` to `node/_utils.ts`. This removes the circular dependency and retains proper error handling behaviour in Windows.

kt3k (approved):
LGTM 👍

cjihrig (approved):
LGTM. Thanks!

PR #2693, opened 2022-09-24
fix: testing example pages failing to load

chrisirhc:
These pages were erroring out on loading due to the use of "ignored" language.

Lowlight was failing due to the use of "ignored" as language name, which is unsupported.
image

### DIFF
--- a/examples/catj.ts
+++ b/examples/catj.ts
@@ -9,7 +9,7 @@
  *
  * Install using `deno install`:
  *
- * ```ignore
+ * ```shellsession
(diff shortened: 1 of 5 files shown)


kt3k (approved):
Good catch! Thanks!

Can you sign the CLA from the above link?

kt3k:
I also created an issue in doc_component repository https://github.com/denoland/doc_components/issues/37

chrisirhc:
Signed!

bartlomieju (approved):

PR #2694, opened 2022-09-24
test(node/fs): add test-fs-read-stream.js

PolarETech:
This PR adds parallel/test-fs-read-stream.js, which was waiting for `child_process.spawnSync()` and `child_process.exec()` implementation in #2653.


node/_tools/test/parallel/test-fs-read-stream.js
```diff
@@ -0,0 +1,284 @@
...
+if (!common.isWindows) {
+  // Verify that end works when start is not specified, and we do not try to
+  // use positioned reads. This makes sure that this keeps working for
+  // non-seekable file descriptors.
+  tmpdir.refresh();
+  const filename = `${tmpdir.path}/foo.pipe`;
+  const mkfifoResult = child_process.spawnSync('mkfifo', [filename]);
+  if (!mkfifoResult.error) {
+    child_process.exec(`echo "xyz foobar" > '${filename}'`);
+    const stream = new fs.createReadStream(filename, common.mustNotMutateObjectDeep({ end: 1 }));
```
PolarETech, line 209: The test item which uses child_process will not be executed on Windows. Therefore, there is no need to add this test to "windowsIgnore" in config.json, as the test will not fail due to escaping issue on Windows.

kt3k (approved):
LGTM. Nice!

PR #2696, opened 2022-09-26
fix(node/fs): improve compatibility of fs.WriteStream

PolarETech:
This is in regards to the comment on https://github.com/denoland/deno_std/pull/2634#pullrequestreview-1104619333.

Part of #911
Similar to #2653

## Scope
This PR aims to address the followings:

1. Make the implementation closer to Node
2. Support `fd` and `start` options
3. Pass the following tests

- [x] parallel/test-file-write-stream3.js
- [x] parallel/test-fs-write-stream-autoclose-option.js
- [x] parallel/test-fs-write-stream-close-without-callback.js
- [x] parallel/test-fs-write-stream-double-close.js
- [x] parallel/test-fs-write-stream-fs.js
- [x] parallel/test-fs-write-stream-throw-type-error.js

## Out of Scope
After examination, I'd like to transfer the following test-related implementations to later Issues and PRs, as they are difficult to address or require a large implementation outside of WriteStream.

- parallel/test-fs-write-stream-change-open.js
   - Overridden `fs.open/close` must be executed in the WriteStream. However, the default import of `fs` in the WriteStream module leads to circular dependencies, so it cannot be implemented in the same way as Node. _Any advice on how to resolve this would be appreciated._
- parallel/test-fs-write-stream-err.js
   - Same as above. `fs.write/close` are overridden within the test.
- parallel/test-fs-write-stream-encoding.js
   - Requires support for latin1 encoding. https://github.com/denoland/deno_std/blob/main/node/string_decoder.ts
- parallel/test-fs-write-stream-file-handle.js
   - Requires `FileHandle` class for `fs.promises.open`. ([Refs.](https://github.com/nodejs/node/blob/main/lib/internal/fs/promises.js))
- parallel/test-fs-write-stream-patch-open.js
   - Requires `child_process.fork()`.

## References
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts
https://nodejs.org/dist/latest-v18.x/docs/api/fs.html


node/internal/fs/streams.ts
```diff
@@ -66,121 +144,231 @@ export function WriteStream(
...
+    // deno-lint-ignore ban-ts-comment
+    // @ts-ignore
+    self._write = null;
+  }
+  if (!self[kFs].writev) {
+    // TODO(PolarETech): Reduce @ts-ignore as much as possible.
+    // deno-lint-ignore ban-ts-comment
+    // @ts-ignore
+    self._writev = null;
+  }
```
PolarETech, line 204: It is difficult to properly resolve type errors, so @ts-ignore is given.

In this part, assign `null` to not override `this._write()` or `this._writev()` in the parent class if `fs.write` or `fs.writev` is not passed as the `fs` option.
Changing the WriteStream interface to allow null for these functions causes a type error because it conflicts with the type declaration of the inherited Writable class.
It is difficult for me to solve this while keeping the implementation close to Node.

cjihrig (approved):
LGTM. Thanks!

PR #2704, opened 2022-09-27
fix(node): revert using `Buffer.from` in setup

iuioiua:
Running `deno task node:setup` fails because `Buffer.from` was not a function. `Buffer` import was coming from the non-Node part of this library.


kt3k (approved):
LGTM

PR #2718, opened 2022-09-29
chore(encoding): remove deprecated APIs

iuioiua:


encoding/csv.ts
```diff
@@ -27,10 +27,6 @@ export type { ReadOptions } from "./csv/_io.ts";
 const QUOTE = '"';
 export const NEWLINE = "\r\n";
 
-export class StringifyError extends Error {
```
kt3k, line 30: I think 'this' StringifyError was not deprecated. (StringifyError in `csv_stringify.ts` was deprecated because that's duplicate of this one)

kt3k:
Left one comment. Other part looks good to me.

encoding/csv.ts
iuioiua: Oh, I see. I'll revert.

kt3k (approved):
LGTM. Thank you for doing this!

PR #2719, opened 2022-09-29
chore(hash): remove `std/hash`

iuioiua:


kt3k:
It seems there's some controversy about this deprecation. (The deprecation itself might have been wrong).

When we deprecated `std/hash`, we overlooked the lack of streaming capability of `std/crypto` while `std/hash` support it. I think it's safe to keep `std/hash` as is until stream support is added to Web Crypto API and `std/crypto`. (The stream support of Web Crypto API is being discussed in [this repo](https://github.com/wintercg/proposal-webcrypto-streams).)

kt3k:
We probably need some clarification about the situation in `hash/README.md`

iuioiua:
Yes, agreed. Is it worth instead only deprecating parts of `std/hash` that overlap with `std/crypto`?

cjihrig:
> Is it worth instead only deprecating parts of std/hash that overlap with std/crypto? This would involve keeping streaming compatibilities.

Is DX likely to suffer if there was a partial deprecation and people needed to use both?

iuioiua:
> Is DX likely to suffer if there was a partial deprecation and people needed to use both?

Yes, I think so. I think it'd be better to deprecate any functionality that overlaps with `std/crypto` and move streams functionality to a new `std/crypto/stream` or similar.

The current need for streams functionality is enough to justify closing this PR. I'll open an issue and we can see what can be done about `std/crypto` and `std/hash`.

PR #2720, opened 2022-09-29
chore: remove deprecated APIs across `collections`, `streams`, `textproto` and `uuid`

iuioiua:
Removes deprecated APIs across:
1. `std/collections`
2. `std/streams`
3. `std/textproto`
4. `std/uuid`

@kt3k, I'd also like to remove the deprecated overloads for `assertThrows` and `assertRejects`. Could you please clarify exactly which overloads are deprecated? Is it just the one, respectively?


kt3k:
> @kt3k, I'd also like to remove the deprecated overloads for assertThrows and assertRejects. Could you please clarify exactly which overloads are deprecated? Is it just the one, respectively?

Yes, the only one overload (with `errorCallback` in the 2nd arg) for each was deprecated. `assertThrows` and `assertRejects` return thrown values since #2226, and we recommend using them for further assertion on thrown value, instead of asserting in `errorCallback`

testing/asserts.ts
```diff
@@ -714,13 +707,6 @@ export function assertRejects<E extends Error = Error>(
...
-/** @deprecated (will be removed after 0.157.0) Use assertRejects(fn, msg) instead, which now returns rejected value
- * and you can assert on it. */
-export function assertRejects(
-  fn: () => PromiseLike<unknown>,
-  errorCallback: (e: Error) => unknown,
-  msg?: string,
-): Promise<Error>;
 export async function assertRejects<E extends Error = Error>(
   fn: () => PromiseLike<unknown>,
   errorClassOrCallbackOrMsg?:
```
kt3k, line 712: This variable now can be much simpler like `errorClassOrMsg: (new (...args: any[]) => E) | string`.

Also `let errorCallback` declaration below (and other related stuff) can be removed.

kt3k (approved):
LGTM. Nice work!

PR #2721, opened 2022-09-29
chore: upgrade `wasmbuild` to 0.10.0 and rust to 1.64.0

iuioiua:
Perhaps, we should start thinking about how we can easily keep these dependencies up-to-date.


kt3k (approved):
LGTM

PR #2726, opened 2022-09-30
fix(http): close server synchronously

### LINKED ISSUE #2717: Leaking async operation after closing server
While working on a test case involving a simple std/http server I encountered the following error:
```
error: AssertionError: Test case is leaking async ops.

 - 1 async operation to sleep for a duration was started in this test, but never completed. This is often caused by not cancelling a `setTimeout` or `setInterval` call. The operation was started here:
    at Object.opAsync (deno:core/01_core.js:177:42)
    at runAfterTimeout (deno:ext/web/02_timers.js:220:31)
    at initializeTimer (deno:ext/web/02_timers.js:186:5)
    at setTimeout (deno:ext/web/02_timers.js:323:12)
    at https://deno.land/std@0.158.0/async/delay.ts:25:15
    at new Promise (<anonymous>)
    at delay (https://deno.land/std@0.158.0/async/delay.ts:16:10)
    at Server.#accept (https://deno.land/std@0.158.0/http/server.ts:387:17)
    at async Server.serve (https://deno.land/std@0.158.0/http/server.ts:134:14)
    at async Server.listenAndServe (https://deno.land/std@0.158.0/http/server.ts:187:12)
```

This can be reproduced using the following code snippet:
```ts
import { Server } from "https://deno.land/std@0.158.0/http/server.ts";

Deno.test("Reproduce leaking test case", async () => {
  const handler = () => {
    return new Response("Test", { status: 200 });
  };
  const port = 4000;
  const server = new Server({ port, handler });
  server.listenAndServe();
  await fetch("http://localhost:" + port, { method: "GET" });
  server.close();
});
```

It seems to be related to the backoff here: https://deno.land/std@0.158.0/http/server.ts?source=#L387. Should this timer be cleared when closing the server?

kt3k:
This PR updates the internal error handling after `server.close()` called. Now `server.close()` clears the underlying ops/resources synchronously.

This makes it simpler/easier to write server related test cases.

closes #2717

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -362,6 +362,9 @@ export class Server {
         // Wait for a new connection.
         conn = await listener.accept();
       } catch (error) {
+        if (this.#closed) {
+          break;
+        }
         if (
           // The listener is closed.
           error instanceof Deno.errors.BadResource ||
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -894,11 +894,7 @@ Deno.test("Server should not reject when the socket is closed", async () => {
 });
 
 Deno.test(
-  {
-    name:
-      "Server should implement a backoff delay when accepting a connection throws an expected error and reset the backoff when successfully accepting a connection again",
-    ignore: true,
-  },
+  "Server should implement a backoff delay when accepting a connection throws an expected error and reset the backoff when successfully accepting a connection again",
   async () => {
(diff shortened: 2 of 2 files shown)


kt3k:
In current `std/http`, `close()` can stop the server synchronously, but with `Deno.serve()`, it can't close the server synchronously (because of multithreading nature). This change is probably not desirable.

PR #2727, opened 2022-09-30
test(node): make .mjs test files executable

PolarETech:
This PR makes the .mjs test files executable.

I have tried to use import maps to import std/node modules in each .mjs test file. If there are other better solutions, I would appreciate suggestions.
Refs: https://deno.land/manual@v1.26.0/node/import_maps

The following tests are currently passed, so they have been added as well.

- parallel/test-stream-asIndexedPairs.mjs
- parallel/test-stream-some-find-every.mjs


cjihrig (approved):
Nice work. LGTM.

PR #2730, opened 2022-09-30
chore(fs): document the deprecation of `exists` and `existsSync`

### LINKED ISSUE #2125: docs: give more detailed description/reasoning/alternative suggestions for the deprecation of fs.exists
Got feedback from the community that the reason of deprecation of fs.exists are not enough clear, not enough understandable and that alternative solutions are not clear. We should document these points in `deprecated` docs of the API.

### LINKED ISSUE #2594: Deprecating "exists" because of some use cases ignores legitimate use cases
Latest docs: https://deno.land/std@0.153.0/fs/mod.ts?s=exists

"Checking the state of a file before using it causes a race condition. Perform the actual operation directly instead."

So it's saying instead of:

1. Check if file/dir exists
2. If it exists, do the thing

 instead do

1. Just do the thing on the file/dir (maybe handle errors)

Rather than that flow, I have this flow:

1. Check if a file/dir exists
2. If it **doesn't** exist, do expensive thing

This doesn't fit into "Perform the actual operation directly instead." scenario. I don't want to perform the operation if the directory or file exists. But I need to check that state beforehand. I have this pattern in many places.

What would the alternative be? 

I can understand the documented scenario is an anti-pattern, but this feels like deprecating crucial functionality without a clear pattern for legitimate use cases.

iuioiua:
`exists` was deprecated because it causes a race condition when used before a file operation. However, it still has some legitimate uses cases, [even within `std`](https://github.com/denoland/deno_std/blob/1c2a47fe0fa1968e5a55fa340c910d525c2b657c/log/handlers.ts#L199-L204). The community has expressed more arguments here:
* #2594
* #2102

This PR:
1. Reverts the deprecation of `exists` and `existsSync`.
2. Advises against using `exists` before a file operation and clearly explains the reasoning.
3. Closes #2125.
4. Cleans up an illegitimate use of `exists` within the codebase.
5. Adds `URL` support for the argument.

A few legitimate use cases were also likely removed in #2546, as `Deno.lstat` is still used to check for the existence of files. It'd be a good idea to revisit these removals and consider re-adding the `exists`, but only where appropriate.

### DIFF
--- a/_tools/check_deprecation.ts
+++ b/_tools/check_deprecation.ts
@@ -14,6 +14,14 @@ const EXCLUDED_PATHS = [
   "http/testdata",
 ];
 
+console.warn(
+  colors.yellow("Warning"),
+  `ignore ${
+    colors.green(`"fs/exists.ts"`)
+  } until issue is resolved: https://github.com/denoland/deno_std/issues/2594`,
+);
+EXCLUDED_PATHS.push("fs/exists.ts");
+
 const ROOT = new URL("../", import.meta.url).pathname.slice(0, -1);
 
 const FAIL_FAST = Deno.args.includes("--fail-fast");
--- a/_tools/check_deprecation.ts
+++ b/_tools/check_deprecation.ts
@@ -14,14 +14,6 @@ const EXCLUDED_PATHS = [
   "http/testdata",
 ];
 
-console.warn(
-  colors.yellow("Warning"),
-  `ignore ${
-    colors.green(`"fs/exists.ts"`)
-  } until issue is resolved: https://github.com/denoland/deno_std/issues/2594`,
-);
-EXCLUDED_PATHS.push("fs/exists.ts");
-
 const ROOT = new URL("../", import.meta.url).pathname.slice(0, -1);
 
 const FAIL_FAST = Deno.args.includes("--fail-fast");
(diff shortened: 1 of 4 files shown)


kt3k:
The update of the document looks good to me. Also the removal of the usage in `log` is very nice!

But let's keep them deprecated for now because the core team is still in favor of deprecation.

Also in my view, the community is also not totally on the same page on reverting deprecation. Some people still prefer the deprecation, and there are suggestion of modified version of API instead of `exists`/`existsSync`. See the discussion in https://github.com/denoland/deno_std/discussions/2102 for details.

iuioiua:
Ok, I can reduce the scope of this PR by ignoring the actual deprecation of it for now.

Please check my previous comment about the use of `exists` within this repo in `std/log`. How do we cater for that use case without using `exists`?

kt3k:
Thanks for updating!

> However, it still has some legitimate uses cases, [even within `std`](https://github.com/denoland/deno_std/blob/1c2a47fe0fa1968e5a55fa340c910d525c2b657c/log/handlers.ts#L199-L204).

Ok. This looks more legitimate usage than typical ones. I also observe several usages in our testing like `assertEquals(existsSync(srcDir), true);`. I'll bring this up in core team meetings.

kt3k (approved):
LGTM

PR #2732, opened 2022-10-01
fix(http): remove unnecessary delay when closing server

KhaledSakr:
This conditionally removes the backoff delay when an error happens while accepting connections, if the server is already closed. This fixes the leaking async op encountered in the issue below, that can also be reproduced using the test case added in this PR.

closes https://github.com/denoland/deno_std/issues/2717

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -371,20 +371,22 @@ export class Server {
           error instanceof Deno.errors.ConnectionReset ||
           error instanceof Deno.errors.NotConnected
         ) {
-          // Backoff after transient errors to allow time for the system to
-          // recover, and avoid blocking up the event loop with a continuously
-          // running loop.
-          if (!acceptBackoffDelay) {
-            acceptBackoffDelay = INITIAL_ACCEPT_BACKOFF_DELAY;
-          } else {
-            acceptBackoffDelay *= 2;
-          }
+          if (!this.#closed) {
+            // Backoff after transient errors to allow time for the system to
+            // recover, and avoid blocking up the event loop with a continuously
+            // running loop.
+            if (!acceptBackoffDelay) {
+              acceptBackoffDelay = INITIAL_ACCEPT_BACKOFF_DELAY;
+            } else {
+              acceptBackoffDelay *= 2;
+            }
 
-          if (acceptBackoffDelay >= MAX_ACCEPT_BACKOFF_DELAY) {
-            acceptBackoffDelay = MAX_ACCEPT_BACKOFF_DELAY;
-          }
+            if (acceptBackoffDelay >= MAX_ACCEPT_BACKOFF_DELAY) {
+              acceptBackoffDelay = MAX_ACCEPT_BACKOFF_DELAY;
+            }
 
-          await delay(acceptBackoffDelay);
+            await delay(acceptBackoffDelay);
+          }
 
           continue;
         }
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -991,6 +991,16 @@ Deno.test(
   },
 );
 
+Deno.test("Server should not leak async ops when closed", () => {
+  const hostname = "127.0.0.1";
+  const port = 4505;
+  const handler = () => new Response();
+  const server = new Server({ port, hostname, handler });
+  server.listenAndServe();
+  server.close();
+  // Otherwise, the test would fail with: AssertionError: Test case is leaking async ops.
+});
+
 Deno.test("Server should reject if the listener throws an unexpected error accepting a connection", async () => {
   const conn = createMockConn();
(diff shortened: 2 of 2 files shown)


http/server.ts
```diff
@@ -371,21 +371,23 @@ export class Server {
...
+              acceptBackoffDelay = INITIAL_ACCEPT_BACKOFF_DELAY;
+            } else {
+              acceptBackoffDelay *= 2;
+            }
+
+            if (acceptBackoffDelay >= MAX_ACCEPT_BACKOFF_DELAY) {
+              acceptBackoffDelay = MAX_ACCEPT_BACKOFF_DELAY;
+            }
+
+            await delay(acceptBackoffDelay);
```
cjihrig, line 388: Isn't the following sequence still possible?

1. Server is accepting traffic.
2. Server enters `delay()`.
3. Server is closed while `delay()` is still executing.
4. Test runner reports leaked op.

KhaledSakr: Agreed, this should be solvable as well by aborting the delay when closing the server. I gave it a go [here](https://github.com/denoland/deno_std/pull/2732/commits/5da431fc42bfb95966fa4afd7bb4a55f1d2bf5f5). However, I think there is room to improve the test case to not have it depend on the leaking detection AssertionError.

http/server.ts
```diff
@@ -253,6 +254,8 @@ export class Server {
 
     this.#closed = true;
 
+    this.#acceptBackoffDelayAbortController.abort();
```
cjihrig, line 257: Does it work to do this after the `for...of` loop a couple lines down?

http/server.ts
```diff
@@ -371,21 +374,32 @@ export class Server {
...
-          // running loop.
-          if (!acceptBackoffDelay) {
-            acceptBackoffDelay = INITIAL_ACCEPT_BACKOFF_DELAY;
-          } else {
-            acceptBackoffDelay *= 2;
-          }
-
-          if (acceptBackoffDelay >= MAX_ACCEPT_BACKOFF_DELAY) {
-            acceptBackoffDelay = MAX_ACCEPT_BACKOFF_DELAY;
+          if (!this.#closed) {
```
cjihrig, line 377: I think this change can be reverted with the addition of the abort controller and my previous comment about moving the `abort()` down.

http/server.ts
```diff
@@ -371,21 +374,32 @@ export class Server {
...
+              acceptBackoffDelay *= 2;
+            }
+
+            if (acceptBackoffDelay >= MAX_ACCEPT_BACKOFF_DELAY) {
+              acceptBackoffDelay = MAX_ACCEPT_BACKOFF_DELAY;
+            }
+
+            await delay(acceptBackoffDelay, {
+              signal: this.#acceptBackoffDelayAbortController.signal,
+            }).catch((err: unknown) => {
```
cjihrig, line 393: Can you avoid mixing `async/await` with `Promise` APIs. In other words, make this a `try...catch`.

http/server_test.ts
```diff
@@ -991,6 +991,44 @@ Deno.test(
   },
 );
 
+Deno.test("Server should not leak async ops when closed", () => {
+  const hostname = "127.0.0.1";
+  const port = 4505;
+  const handler = () => new Response();
+  const server = new Server({ port, hostname, handler });
+  server.listenAndServe();
+  server.close();
```
cjihrig, line 1000: No need to make the change in this PR if this fixes the original issue, but I still think it would be helpful to be able to `await` on `server.close()`.

http/server.ts
KhaledSakr: Aha, that makes sense.

http/server.ts
KhaledSakr: I made the change here: https://github.com/denoland/deno_std/pull/2732/commits/d7498295896b29ce0824d68f9c48010cbc6674cb

http/server.ts
KhaledSakr: Made the change here: https://github.com/denoland/deno_std/pull/2732/commits/e65d8bc4ad5627ebb8a4e504320784ff1293d390

http/server.ts
KhaledSakr: Yes, that's reasonable. I made the change here: https://github.com/denoland/deno_std/pull/2732/commits/4f7ef9c987c4b3feba1b82649ccce3b646a82f5c

http/server_test.ts
KhaledSakr: In general I think it makes sense, specially if we want to wait until all the pending handlers are resolved. However, I'm not yet comfortable enough with the codebase to make this change and I would prefer to keep it separate.

cjihrig (approved):
LGTM. Thanks for the contribution!

PR #2735, opened 2022-10-03
fix(streams): don't use shared buffer for iterateReader outputs

kt3k:
`iterateReader` and `iterateReaderSync` use shared underlying buffer for its outputs. This causes the issue when the consumer of iterable is slow at handling these bytes. (Especially when used in `fetch` body, the request body can be corrupted in real world use cases. See https://github.com/denoland/deno/issues/14437 for details.)

closes https://github.com/denoland/deno/issues/14437

### DIFF
--- a/streams/conversion.ts
+++ b/streams/conversion.ts
@@ -494,8 +494,8 @@ export async function* iterateReader(
   },
 ): AsyncIterableIterator<Uint8Array> {
   const bufSize = options?.bufSize ?? DEFAULT_BUFFER_SIZE;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, nice fix!

0f-0b:
https://github.com/denoland/deno_std/blob/c1e1240c0343f019a6727e741e024fb34c450bbf/streams/conversion.ts#L485-L488

Documentation needs to be updated as well.

kt3k:
@0f-0b Good catch. Thanks!

PR #2739, opened 2022-10-03
refactor(http/file_server): simplify normalizeURL util

kt3k:
The most of normalizeURL validations/path processing now seem useless because the basic validations are already done in CLI layer. This PR updates and simplifies it.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -620,37 +620,7 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
 }
 
 function normalizeURL(url: string): string {
-  let normalizedUrl = url;
-
-  try {
-    //allowed per https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html
-    const absoluteURI = new URL(normalizedUrl);
-    normalizedUrl = absoluteURI.pathname;
-  } catch (e) {
-    //wasn't an absoluteURI
-    if (!(e instanceof TypeError)) {
-      throw e;
-    }
-  }
-
-  try {
-    normalizedUrl = decodeURIComponent(normalizedUrl);
-  } catch (e) {
-    if (!(e instanceof URIError)) {
-      throw e;
-    }
-  }
-
-  if (normalizedUrl[0] !== "/") {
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -620,37 +620,7 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
...
-      throw e;
-    }
-  }
-
-  if (normalizedUrl[0] !== "/") {
-    throw new URIError("The request URI is malformed.");
-  }
-
-  normalizedUrl = posix.normalize(normalizedUrl);
-  const startOfParams = normalizedUrl.indexOf("?");
```
kt3k, line 649: This check was for removing query string part, but we already take `pathname` part from URL object. This check should be now useless.

http/file_server.ts
```diff
@@ -620,37 +620,7 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
...
-
-  try {
-    normalizedUrl = decodeURIComponent(normalizedUrl);
-  } catch (e) {
-    if (!(e instanceof URIError)) {
-      throw e;
-    }
-  }
-
-  if (normalizedUrl[0] !== "/") {
```
kt3k, line 644: because we use `pathname` part of URL object. I think this can't be true

http/file_server.ts
```diff
@@ -620,37 +620,7 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
 }
 
 function normalizeURL(url: string): string {
-  let normalizedUrl = url;
-
-  try {
-    //allowed per https://www.w3.org/Protocols/rfc2616/rfc2616-sec5.html
-    const absoluteURI = new URL(normalizedUrl);
```
kt3k, line 627: because `url` of Request object is already parsed once ( https://github.com/denoland/deno/blob/7a7767262a3b373cc9a3f6260a8ebd45da7edd1c/ext/fetch/26_fetch.js#L370 ), this can't throw an error.

bartlomieju (approved):
LGTM, nice cleanup

cjihrig (approved):

PR #2740, opened 2022-10-03
feat(node/crypto): add base64url encoding to hash.digest()

dz4k:
This is supported in Node:
> - `'base64url'`: [base64url](https://tools.ietf.org/html/rfc4648#section-5) encoding as specified in [RFC 4648, Section 5](https://tools.ietf.org/html/rfc4648#section-5).
https://nodejs.org/dist/latest-v16.x/docs/api/buffer.html#buffers-and-character-encodings

### DIFF
--- a/node/crypto_test.ts
+++ b/node/crypto_test.ts
@@ -44,7 +44,7 @@ Deno.test("[node/crypto.Hash] basic usage - base64 output", () => {
 });
 
 Deno.test("[node/crypto.Hash] basic usage - base64url output", () => {
-  const d = createHash("sha1").update("abc").update("def").digest("base64");
+  const d = createHash("sha1").update("abc").update("def").digest("base64url");
   assertEquals(d, "H4rBDyPFtbwRZ72oS4M-XAV6d9I");
 });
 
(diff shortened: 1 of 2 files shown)


cjihrig (commented):
Thanks for the contribution. Can you add a test?

node/crypto_test.ts
```diff
@@ -38,6 +38,16 @@ Deno.test("[node/crypto.Hash] basic usage - hex output", () => {
   assertEquals(d, "1f8ac10f23c5b5bc1167bda84b833e5c057a77d2");
 });
 
+Deno.test("[node/crypto.Hash] basic usage - base64 output", () => {
+  const d = createHash("sha1").update("abc").update("def").digest("base64");
+  assertEquals(d, "H4rBDyPFtbwRZ72oS4M+XAV6d9I=");
+});
+
+Deno.test("[node/crypto.Hash] basic usage - base64url output", () => {
+  const d = createHash("sha1").update("abc").update("def").digest("base64");
```
cjihrig, line 47: Should this be base64url?

dz4k: It should, thanks.

cjihrig (approved):

PR #2743, opened 2022-10-03
fix(node): handle inherited output in spawnSync()

cjihrig:
Deno.spawnSync() returns getters for stdout and stderr which throw when 'inherit' is used. This commit updates Node's spawnSync() to handle this case.

Fixes: https://github.com/denoland/deno_std/issues/2742


bartlomieju (approved):
LGTM

PR #2744, opened 2022-10-04
feat(collections): improve types of `partition` module

TomokiMiyauci:
This allows for more accurate type inference when type guards are provided, like `Array.filter`.

```ts
declare const target: (string | number)[].

partition(target, (value: unknown): value is string => typeof value === "string") // [ string[], number[]]
```

Backward compatibility should not be a problem.


kt3k (approved):
LGTM. Nice improvement!

Please sign the CLA from the above link

PR #2749, opened 2022-10-05
fix(node/tls): fix TLSSocket constructor

### LINKED ISSUE #2701: node TLSSocket throws TypeError if opts undefined
**Describe the bug**

TypeError thrown if opts is undefined...

https://deno.land/std@0.157.0/node/_tls_wrap.ts?source=#L70

```typescript
  constructor(socket: any, opts: any) {
    const tlsOptions = { ...opts };

    let hostname = tlsOptions?.secureContext?.servername;
    hostname = opts.host;
    tlsOptions.hostname = hostname;
```

The line `hostname = opts.host;` throws an error if opts was undefined.

```
error: Uncaught TypeError: Cannot read properties of undefined (reading 'host')
    hostname = opts.host;
                    ^
    at new TLSSocket (https://deno.land/std@0.157.0/node/_tls_wrap.ts:70:21)
```

but also, it doesn't make sense, as hostname may already be set and then this line completely overrides the line above (or errors).

**Steps to Reproduce**

I was attempting to run `web-ext` cli directly from npm to see if it worked from Deno...

```
deno run --unstable --allow-env --allow-read npm:web-ext --version
```

**Expected behavior**

For web-ext to run and output it's version number.

**Environment**

- OS: MacOS 12.4
- deno version: 1.25.4 (release, x86_64-apple-darwin)
- std version: 0.157.0

kt3k:
closes #2701

### DIFF
--- a/node/_tls_wrap.ts
+++ b/node/_tls_wrap.ts
@@ -1,6 +1,7 @@
 // Copyright 2018-2022 Deno authors. All rights reserved. MIT license.
 // Copyright Joyent and Node contributors. All rights reserved. MIT license.
 // deno-lint-ignore-file no-explicit-any
+
 import {
   ObjectAssign,
   StringPrototypeReplace,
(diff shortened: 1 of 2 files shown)


node/_tls_wrap.ts
```diff
@@ -63,7 +63,7 @@ export class TLSSocket extends net.Socket {
   [kConnectOptions]: any;
   ssl: any;
   _start: any;
-  constructor(socket: any, opts: any) {
+  constructor(socket: any, opts: any = {}) {
```
cjihrig, line 66: Consider using `kEmptyObject` instead of creating a new object here.

cjihrig (approved):

PR #2755, opened 2022-10-06
fix(node/http): disable chunked request if Content-Length header is specified

### LINKED ISSUE #2754: node: selenium-webdriver fails to launch Chrome with `Uncaught WebDriverError: Unexpected HTTP response`
**Describe the bug**

When running [selenium-webdriver](https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver), I get an `Uncaught WebDriverError: Unexpected HTTP response` error and cannot launch Chrome.

Upon investigation, the following behaviors seem to be the cause of the error.

- chromedriver does not support chunked requests (https://bugs.chromium.org/p/chromedriver/issues/detail?id=1850)
- node/http.request() sends chunked requests to chromedriver

From the Node documentation, I think chunked encoding should be disabled if `Content-Length` header is specified.

https://nodejs.org/api/http.html#httprequesturl-options-callback

> There are a few special headers that should be noted.
> - Sending a 'Content-Length' header will disable the default chunked encoding.

**Steps to Reproduce**

1. To avoid #2746, prepare the environment to use deno_std after 19f9df7
2. Install Chrome and chromedriver ([Ref.](https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver#installation))
3. Create a script with the contents below:
```mjs
// index.mjs
// Ref: https://github.com/SeleniumHQ/selenium/tree/trunk/javascript/node/selenium-webdriver#usage

import { Builder, Browser, By, Key, until } from 'npm:selenium-webdriver';
  
(async function example() {
  let driver = await new Builder().forBrowser(Browser.CHROME).build();
  try {
    await driver.get('https://www.google.com/ncr');
    await driver.findElement(By.name('q')).sendKeys('webdriver', Key.RETURN);
    await driver.wait(until.titleIs('webdriver - Google Search'), 1000);
  } finally {
    await driver.quit();
    Deno.exit();
  }
})();
```
4. Execute the command `deno run -A --unstable index.mjs`
5. See error
```mjs
error: Uncaught WebDriverError: Unexpected HTTP response:
HTTP/1.1 0



    at parseHttpResponse (file:///Users/ptech/selenium/cache/npm/registry.npmjs.org/selenium-webdriver/4.5.0/lib/http.js:569:11)
    at Executor.execute (file:///Users/ptech/selenium/cache/npm/registry.npmjs.org/selenium-webdriver/4.5.0/lib/http.js:515:28)
    at Object.runMicrotasks (deno:core/01_core.js:336:30)
    at processTicksAndRejections (file:///Users/ptech/deno_std/node/_next_tick.ts:62:10)
```

**Expected behavior**

Chrome should launch and perform the actions described in the script.

**Environment**

- OS: MacOS 10.15.7
- deno version: 1.26.0+b487027
- std version: d2cf6c8
- selenium-webdriver: 4.5.0
- Chrome: 106.0.5249.103
- chromedriver: 106.0.5249.61

PolarETech:
Fixes #2754

This PR disables applying chunked encoding to the request body if the following conditions are met.

- `Transfer-Encoding: chunked` is not specified in the header options
- `Content-Length` is specified in the header options

## References

https://nodejs.org/api/http.html#httprequesturl-options-callback
https://datatracker.ietf.org/doc/html/rfc7230#section-3.3

## Tests

No applicable test for this issue has been found in the Node test suite. Unit tests have been added.


node/http_test.ts
```diff
@@ -148,3 +148,150 @@ Deno.test("[node/http] request default protocol", async () => {
...
 });
+
+Deno.test("[node/http] send request with non-chunked body", async () => {
+  const promise = deferred<void>();
+  let requestHeaders: Record<string, string> = {};
+  const requestBody: string[] = [];
+
+  const server = http.createServer((req, res) => {
+    req.on("data", (chunk) => {
+      requestBody.push(chunk.toString());
```
cjihrig, line 159: Does it work to call `req.setEncoding('utf8')` and then concatenating the chunks in a string. Same comment for the other tests.

PolarETech: Yes, it works fine in that case. Should I modify the tests?

node/http_test.ts
```diff
@@ -148,3 +148,150 @@ Deno.test("[node/http] request default protocol", async () => {
...
+    req.write("world");
+    req.end();
+  });
+  server.on("close", () => {
+    promise.resolve();
+  });
+  await promise;
+});
+
+Deno.test("[node/http] send request with chunked body", async () => {
```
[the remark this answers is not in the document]
PolarETech: I am aware of this.
I think it is difficult to make the behavior completely the same as Node because of the spec of Deno's fetch().

__For Node's http.request():__
If both `Content-Length` and `Transfer-Encoding` headers are specified, both `Content-Length` and `Transfer-Encoding` headers are sent. The request body is chunked encoding.

__For Deno's fetch():__
Even if both `Content-Length` and `Transfer-Encoding` headers are specified(*), if the body is a ReadableStream, only the `Transfer-Encoding` header is sent.

This unit test is designed to make sure that the `Content-Length` header is not sent depending on the current behavior of Deno's fetch(), so it is not passed by Node.

For reference, the HTTP logs obtained by Wireshark during the test run are attached below.

_Node:_

```
POST / HTTP/1.1
Content-Type: text/plain; charset=utf-8
Content-Length: 11
Transfer-Encoding: chunked
Host: localhost:56750
Connection: close

6
hello 
5
world
0

HTTP/1.1 400 Bad Request
Connection: close

```

_Deno:_

```
POST / HTTP/1.1
content-type: text/plain; charset=utf-8
transfer-encoding: chunked
accept: */*
accept-language: *
user-agent: Deno/1.26.1
accept-encoding: gzip, br
host: localhost:56390

6
hello 
5
world
0

HTTP/1.1 200 OK
Date: Fri, 07 Oct 2022 19:28:48 GMT
Content-Type: text/plain;charset=UTF-8
Content-Length: 2

ok
```

(*) The current implementation does not pass the headers option to fetch() in the relevant code of node/http.ts, but the same result will be obtained if it is changed to do so.
https://github.com/denoland/deno_std/pull/2755/files#diff-07717ecd86915a087f8759f720b0217ab7c3c9f473aa8f4772aeae5dbb638a92R127-R131
I would like to create a separate PR for not passing the headers option to fetch().

node/http_test.ts
cjihrig: I think it's safer to do it that way (less assumptions about chunk sizes and more resilient to multibyte characters being split across chunks).

PolarETech: I see. I have modified the unit tests. (7bd2bdf)

node/http.ts
```diff
@@ -152,6 +154,35 @@ class ClientRequest extends NodeWritable {
     this.destroy();
   }
 
+  async _createBody(
+    body: ReadableStream | null,
+    opts: RequestOptions,
+  ): Promise<Buffer | ReadableStream | null> {
+    if (!body) return null;
+
+    const isChunked = (opts: RequestOptions) => {
```
cjihrig, line 163: Can you either inline this code or move the function to the top level. The way this is currently written, it will need to unnecessarily create a new closure each time `_createBody()` is called.

PolarETech: Thank you for pointing this out. I have inlined the code. (a29e9d9)

cjihrig (commented):
Left another comment.

To be honest, I'm not sure if this should be landed or not. Your reasoning seems legit, but the fact that there are no Node tests and the change behaves differently from Node based on my testing makes me think we might need some other changes first.

@kt3k thoughts?

node/http_test.ts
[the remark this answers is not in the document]
kt3k: @PolarETech Do we need this 2nd test case? The first and third ones look enough for checking the fix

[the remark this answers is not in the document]
kt3k: Ok, I confirmed that Node.js sends `transfer-encoding: chunked` when both `content-length: N` and `transfer-encoding: chunked` are specified. This test case seems useful to me. But how about rewriting the server part of the test case with `serve` function of `std/http`? Because this test case depends on non-compliant behavior of `http.Server` and server part is not the subject of the test case, I think it makes more sense to use `std/http` here.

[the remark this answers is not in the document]
PolarETech: Thanks for your advice. I have changed the unit test cases. (d38c018)

kt3k (approved):
@PolarETech Sorry for the delay in review. I missed your last comment.. 😓 

Now this looks good to me. Thanks for your contribution! LGTM

PR #2760, opened 2022-10-07
feat(node): add readline/promises

### LINKED ISSUE #2752: Missing readline/promises
https://nodejs.org/api/readline.html#promises-api

PolarETech:
Closes #2752
Part of #911

## Related Tests

- [x] parallel/test-readline-promises-csi.mjs
- [x] parallel/test-readline-promises-interface.js

## TODO

- [x] readline/promises.mjs -> readline/promises.ts

## Out of Scope

- parallel/test-readline-promises-tab-complete.js

A similar test, parallel/test-readline-tab-complete.js, also fails with the same error.
The cause of the test failure seems to be outside the readline module and seems difficult to fix, so let me exclude it from the scope of this PR.

## References

https://nodejs.org/api/readline.html#promises-api
https://github.com/nodejs/node/blob/main/lib/readline/promises.js
https://github.com/nodejs/node/blob/main/lib/internal/readline/promises.js
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/readline/promises.d.ts


node/readline/promises.ts
```diff
@@ -0,0 +1,194 @@
...
+
+import { kEmptyObject } from "../internal/util.mjs";
+import type { Abortable } from "../_events.d.ts";
+
+// NOTE: These type redefinitions are intended to avoid circular dependencies.
+// import type {
+//   AsyncCompleter,
+//   Completer,
+//   ReadLineOptions,
+// } from "../_readline.d.ts";
```
cjihrig, line 22: Instead of duplicating things, can they be placed in a file that both implementations can import?

node/readline/promises.ts
```diff
@@ -0,0 +1,194 @@
...
+  output?: WritableStream | undefined;
+  completer?: Completer | AsyncCompleter | undefined;
+  terminal?: boolean | undefined;
+  history?: string[] | undefined;
+  historySize?: number | undefined;
+  prompt?: string | undefined;
+  crlfDelay?: number | undefined;
+  removeHistoryDuplicates?: boolean | undefined;
+  escapeCodeTimeout?: number | undefined;
+  tabSize?: number | undefined;
```
cjihrig, line 40: Do these need to be both optional and explicitly allow `undefined`?

node/readline/promises.ts
PolarETech: Thank you for your suggestion.
I have created another .d.ts file and modified them to import from the file. (caa1377)

node/readline/promises.ts
PolarETech: I had just copy-pasted the contents of _readline.d.ts.
As you suggested, I have placed these in a separate file. I left the `undefined` in that file as well, wondering if it would be better to keep the contents of the original forked file as much as possible.

node/_readline.d.ts
```diff
@@ -1,43 +1,54 @@
...
 import { Abortable, EventEmitter } from "./_events.d.ts";
+import * as promises from "./readline/promises.ts";
 import { ReadableStream, WritableStream } from "./_global.d.ts";
 import { Buffer } from "./buffer.ts";
+import type {
+  AsyncCompleter,
+  Completer,
+  CompleterResult,
+  ReadLineOptions,
+} from "./_readline_2.d.ts";
```
cjihrig, line 15: Can you give this new file a more meaningful name. Something like `_readline_shared_types.d.ts` for example.

PolarETech: Sorry, I couldn't name it appropriately. The file name has been revised. (eeb6ce0)

cjihrig (approved):
LGTM, thanks!

PR #2763, opened 2022-10-10
chore: Avoid using rid-based filesystem functions

ayame113:
I think Deno CLI is trying to move away from using rid based filesystem functions. (https://github.com/denoland/deno/issues/12107, https://github.com/denoland/deno/issues/15796#issue-1364389627, https://github.com/denoland/deno/pull/15933#issuecomment-1250044922)
This PR removes the use of rid based functions from the standard library and replaces them with:

- `Deno.read(file.rid, buf)` -> `file.read(buf)`
- `Deno.write(file.rid, buf)` -> `file.write(buf)`
- `Deno.close(file.rid)` -> `file.close()`

(However, I didn't change `node` and `wasi` because they still seem to need rid-based functions.)

This is just a source code cleanup and does not change existing behavior.

### DIFF
--- a/_util/download_file.ts
+++ b/_util/download_file.ts
@@ -20,8 +20,5 @@ export async function downloadFile(url: string, fileUrl: URL) {
 
   await ensureFile(fromFileUrl(fileUrl));
   const file = await Deno.open(fileUrl, { truncate: true, write: true });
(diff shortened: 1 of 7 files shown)


cjihrig (approved):
LGTM, thanks!

PR #2765, opened 2022-10-10
chore(examples): Converting Deno Streams to Web Streams

ayame113:
towards https://github.com/denoland/deno_std/issues/1986

Switch the Deno stream used in the `example/` directory to a web stream.
And resolve the todo comment below.

> // TODO(ry) Re-enable streaming in this example.
// Originally we did: await Deno.copy(res.body, Deno.stdout);
// But maybe more JS-y would be: res.pipeTo(Deno.stdout);

### DIFF
--- a/examples/xeval.ts
+++ b/examples/xeval.ts
@@ -33,7 +33,7 @@ OPTIONS:
 ARGS:
   <code>`;
 
-export type XevalFunc = (v: string) => unknown | Promise<unknown>;
+export type XevalFunc = (v: string) => Promise<unknown>;
 
 export interface XevalOptions {
   delimiter?: string;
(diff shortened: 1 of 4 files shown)


examples/xeval.ts
```diff
@@ -33,7 +33,7 @@ OPTIONS:
 ARGS:
   <code>`;
 
-export type XevalFunc = (v: string) => void;
+export type XevalFunc = (v: string) => unknown | Promise<unknown>;
```
cjihrig, line 36: It looks like `xEvalFunc` is an async function here. Won't that make the return type only be `Promise<unknown>`?

ayame113: It makes sense. 👍

cjihrig (approved):

Issue #2733, opened 2022-10-02
review `std/crypto` and `std/hash`


iuioiua:
Continuing the discussion from #2719, `std/hash` shouldn't be fully deprecated due to its desired streaming functionality that isn't yet available as a web spec. [Although, it's planned to one day be.](https://github.com/wintercg/proposal-webcrypto-streams)

However, `std/hash` does overlap with `std/crypto` in a few ways. I have a few suggestions that I think would prune some of this overlapped functionality:
1. Move stream-based functionality within `std/hash` to `std/crypto/streams`.
2. Deprecate APIs within `std/hash` that exist within `std/crypto`.
3. Move some `Hasher` methods such as `toString` as pure functions in `std/crypto`.

Put simply; I suggest we merge `std/hash` into `std/crypto`. Doing so would get rid of duplicate code and maintain streaming support. @kt3k and @cjihrig, what do you guys think?

cjihrig:
If we can make `std/hash` completely redundant, I'd be supportive of moving things into `std/crypto`. If we can't make it redundant, maybe we should just wait on streaming webcrypto.

iuioiua:
@kt3k, would you be able to point out where streaming functionality exists within `std/hash`. Having a quick look through, I'm unable to spot it. Maybe I'm missing something...

iuioiua:
I think `std/crypto` would benefit from having a `toHashString` function, inspired by the `Hash.toString()` method from `std/hash`. Seems like a convenient thing to have.

Apart from that, I don't see where `std/crypto` is lacking, compared to `std/hash` (see above comment).

iuioiua:
Perhaps, there may be some confusion between `std/hash` and `node:crypto` (which does support streams with hashes).

kt3k:
@iuioiua 
With `std/hash`, you can call `hash.update("...")` multiple times and that appends the data to the internal state.

```js
const hash = createHash("md5");
for await (const chunk of readable) {
  hash.update(chunk);
}
hash.digest().toString();
```

I think we can consider this capability as streaming support. (The same thing is not possible with current Web Crypto's `digest` API, I guess)

iuioiua:
[...]

PR #2766, opened 2022-10-10
test(crypto): test `ReadableStream` in `digest`

iuioiua:
I'm neither an expert on the Stream nor Web Crypto APIs, so please let me know of any concerns. Testing is basic for now but can be expanded if needed.

Part of #2733.


kt3k:
(Sorry for the delay in review. We were busy last week)

Oh, we already support `AsyncIterable<BufferSource>` source.. (That makes readable stream input automatically working..) I wasn't fully aware of it..

littledivy:
Maybe we should wait for https://github.com/wintercg/proposal-webcrypto-streams?

kt3k:
`std/crypto` seems deviating from web crypto from the beginning. ref. https://github.com/denoland/deno_std/pull/1025

And because it supports `AsyncIterable<BufferSource>` as input, the readable stream seems already supported

kt3k:
I think this change is unnecessary and my previous comment (`std/crypto` doesn't support streaming) was wrong. I think we can just deprecate & remove `std/hash` now.

iuioiua:
> I think this change is unnecessary and my previous comment (`std/crypto` doesn't support streaming) was wrong. I think we can just deprecate & remove `std/hash` now.

I'll tweak this PR so it just provides tests for digesting streams instead.

kt3k (approved):
LGTM 👍

PR #2767, opened 2022-10-10
feat(crypto): `toHashString` utility

iuioiua:
This is to be the `std/crypto` version of `Hasher.toString` from `std/hash`. The aim of this function is to contribute to the deprecation of `std/hash`, as part of #2733.


iuioiua:
Are there any concerns with this PR, @kt3k?

kt3k:
@iuioiua Sorry for the delay in review. (We, most core team members, were on the offsite meeting last week, and were a bit busy on it..)

This utility makes sense to me as most users probably want to use string representation of the digest results. The default encoding ('hex') also looks good to me.

Does anyone have any concern on adding this util to `crypto/mod.ts`?

iuioiua:
The failing test is unrelated to this change.

iuioiua:
Gently pinging @bartlomieju and @cjihrig.

cjihrig (approved):
No objections. LGTM

kt3k (approved):
LGTM. Nice utility to have!

PR #2773, opened 2022-10-12
feat(http): `createCommonResponse` internal utility

### LINKED ISSUE #1587: std/http: Provide classes for common HTTP response status codes

When building a web application, it's very common to have to re-use the same HTTP responses when handling errors.
Applications that do a lot of input-validation might find it annoying to have to redeclare the proper HTTP responses.
Others might find it annoying that there is no "proper" way of handling errors.

```ts

serve(req => {
  if (!validateRequest(req)) return new Response("Did not pass validation", { status: 400, statusCode: "Bad Request" }); 
  if (!pathExists(req)) return new Response("404 Not Found", { status: 404, statusCode: "404 Not Found" });
  if (shouldRedirect(req)) return new Response(undefined, { status: 308, location: "https://www.example.com" });

  return new Response("Hello world"));
}
```

We could add classes for common HTTP status responses to the `std/http` module.
These classes would include a default message, such as `404 Not Found`, and would return the correct status code and text.
This would remove some boilerplate from any project that makes use of the `Response` class, such as projects using `Deno.serveHttp()` or the `std/http` server.

```ts
import { 
  BadRequestResponse,
  NotFoundResponse,
  PermanentRedirectResponse,
} from "https://deno.land/std/http/responses.ts";

serve(req => {
  if (!validateRequest(req)) return new BadRequestResponse("Did not pass validation"); 
  if (!pathExists(req)) return new NotFoundResponse();
  if (shouldRedirect(req)) return new PermanentRedirectResponse("https://www.example.com");

  return new Response("Hello world"));
}
```

Example implementation (API doesn't have to be this, it's just an example):
__std/http/response.ts:__
```ts
//  The class should extend the `Response` class, so that 
// `new NotFoundResponse() instanceof Response` returns `true`.
export class NotFoundResponse extends Response {
  constructor(body: BodyInit = "404 Not Found", init: ResponseInit = {}) {
    super(body, {
      ...init,
      status: 404,
      statusText: "Not Found",
    });
  }
}

// This class is a bit more opinionated, but is still very generic.
class PermanentRedirectResponse extends Response {
  constructor(location: string) {
    super(undefined, {
      status: 308,
      statusText: "Permanent Redirect",
      headers: { location },
    });
  }
}
```

Because this suggestion is not hard to implement, the alternative would be to let programmers build it themselves.

However, because this is a fairly straightforward suggestion and because it could benefit any project, I think it would do no harm to add it to the `std/http` module.

Any feedback is welcomed.

iuioiua:
This utility is meant to be small, simple and reduce boilerplate. I've implemented it within `file_server` to exemplify its use cases.

Closes #1587.

### DIFF
--- a/http/mod.ts
+++ b/http/mod.ts
@@ -12,4 +12,3 @@ export * from "./http_errors.ts";
 export * from "./http_status.ts";
 export * from "./negotiation.ts";
 export * from "./server.ts";
-export * from "./util.ts";
--- a/http/mod.ts
+++ b/http/mod.ts
@@ -12,3 +12,4 @@ export * from "./http_errors.ts";
 export * from "./http_status.ts";
 export * from "./negotiation.ts";
 export * from "./server.ts";
+export * from "./util.ts";
--- a/http/README.md
+++ b/http/README.md
@@ -38,24 +38,6 @@ console.log(Status.NotFound); //=> 404
 console.log(STATUS_TEXT[Status.NotFound]); //=> "Not Found"
 ```
 
-## Common Response
-
-Small utility for returning a standardized response, automatically defining the
-body, status code and status text, according to the response type.
-
-```ts
-import {
-  commonResponse,
-  Status,
-} from "https://deno.land/std@$STD_VERSION/http/mod.ts";
-
-const response = commonResponse(Status.NotFound);
-
-console.log(await response.text()); // "Not Found"
-console.log(response.status); // 404
-console.log(response.statusText); // "Not Found"
-```
-
 ## HTTP errors
 
 Provides error classes for each HTTP error status code as well as utility
--- a/http/README.md
+++ b/http/README.md
@@ -38,6 +38,24 @@ console.log(Status.NotFound); //=> 404
 console.log(STATUS_TEXT[Status.NotFound]); //=> "Not Found"
 ```
 
+## Common Response
+
+Small utility for returning a standardized response, automatically defining the
+body, status code and status text, according to the response type.
+
+```ts
+import {
+  commonResponse,
+  Status,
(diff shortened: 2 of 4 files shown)


cjihrig (commented):
This definitely looks like a useful utility function, and the code changes LGTM.

@kt3k what do you think of the function name and whether it should be part of the public API?

kt3k:
It looks good at least as an internal utility, but I'm not sure this has general enough value as a public API. I'll discuss this and collect opinions of core members tomorrow at the offsite.

kt3k:
Discussed this shortly with Luca and Kitson, and we all think what this utility does is too trivial and doesn't have general enough purpose. So I'd suggest we should keep this as an internal utility.

iuioiua:
Yeah, I dig it. I'll make the appropriate adjustments.

timreichen:
@iuioiua Event if it is used internally, I think it would be better to start the name of a factory function with a verb, e.g. `createStatusResponse` or `createCommonResponse`.

cjihrig (approved):
LGTM, thanks!

cjihrig:
Looks like the linter is requesting changes 😄

http/mod.ts
```diff
@@ -12,3 +12,4 @@ export * from "./http_errors.ts";
 export * from "./http_status.ts";
 export * from "./negotiation.ts";
 export * from "./server.ts";
+export * from "./util.ts";
```
kt3k, line 15: Maybe let's not do this?

kt3k (approved):
LGTM with nits

http/mod.ts
iuioiua: Sorry, missed that.

kt3k:
@iuioiua Thanks for the updates!

PR #2774, opened 2022-10-13
feat(math): Add distribution functions.

retraigo:
This PR adds two functions under the `math` module.

`uniformRange` generates an array of numbers uniformly distributed between a range in ascending order.

`normalDistribution` generates an array of numbers normally distributed around a mean in random order.

Tests have been written under the `distributions_test.ts` file.

### DIFF
--- a/math/README.md
+++ b/math/README.md
@@ -14,7 +14,7 @@ parameters:
 This implementation makes use of the popular Box-Muller transform.
 
 ```ts
-import { normalDistribution } from "https://deno.land/std/math/distributions.ts";
+import { normalDistribution } from "https://deno.land/std@$STD_VERSION/math/distributions.ts";
 
 const distribution = normalDistribution(100, 16, 0.003)
 ```
--- a/math/README.md
+++ b/math/README.md
@@ -24,7 +24,7 @@ const distribution = normalDistribution(100, 16, 0.003)
 Generate an array of numbers uniformly spread across a range.
 
 ```ts
-import { uniformRange } from "https://deno.land/std/math/distributions.ts";
+import { uniformRange } from "https://deno.land/std@$STD_VERSION/math/distributions.ts";
 
 const range = uniformRange(10, {
   min: 1,
--- a/math/README.md
+++ b/math/README.md
@@ -1,0 +1,34 @@
+# std/math
+
+Provides math functions for generating distributions.
+
+## Random Normal Distribution
+
+A **Normal** or **Gaussian** distribution is a type of 
+continuous probability distribution dependent on two
+parameters:
+
+**μ** - The **mean** 
+**σ** - The **standard deviation**
+
+This implementation makes use of the popular Box-Muller transform.
+
+```ts
+import { normalDistribution } from "https://deno.land/std/math/distributions.ts";
+
+const distribution = normalDistribution(100, 16, 0.003)
+```
+
+## Uniform Range
+
+Generate an array of numbers uniformly spread across a range.
+
+```ts
+import { uniformRange } from "https://deno.land/std/math/distributions.ts";
+
+const range = uniformRange(10, {
+  min: 1,
+  max: 10,
+})
+// [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
+```
--- a/math/distributions.ts
+++ b/math/distributions.ts
@@ -1,0 +1,100 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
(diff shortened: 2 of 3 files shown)


math/distributions.ts
```diff
@@ -0,0 +1,90 @@
...
+ *   min: 0,
+ *   max: 100,
+ * })
+ * ```
+ */
+export function uniformRange(
+  n: number,
+  options: RangeOptions = BASE_RANGE_OPTIONS,
+): number[] {
+  if (typeof n !== "number" || n <= 0 || !Number.isInteger(n)) return [];
```
kt3k, line 37: I think this should throw TypeError instead of returning empty array

math/README.md
```diff
@@ -0,0 +1,34 @@
...
+continuous probability distribution dependent on two
+parameters:
+
+**μ** - The **mean** 
+**σ** - The **standard deviation**
+
+This implementation makes use of the popular Box-Muller transform.
+
+```ts
+import { normalDistribution } from "https://deno.land/std/math/distributions.ts";
```
kt3k, line 17: This should be like the below to pass the CI checks

```suggestion
import { normalDistribution } from "https://deno.land/std@$STD_VERSION/math/distributions.ts";
```

kt3k:
@retraigo Do you have any image about the scope of `math` module? What kind of items do you suggest we should add in `math` module?

retraigo:
> @retraigo Do you have any image about the scope of `math` module? What kind of items do you suggest we should add in `math` module?

Finding math-related JS/TS code on the internet is a really tedious process. There are only a few math-related modules like mathjs out there. Having a standard, third-party-dependency-free math module for Deno would be great.

It could support any math-related stuff that makes sense to do in JavaScript like,
- Range generation (like the python `range()`)
- ndarrays (a stricter version that involves a private variable maybe?)
- Gamma
- Complex numbers
or even some basic calculus.

But IMO it shouldn't contain math that operates on a large amount of data (like regression analysis, matrices, etc). Not because they're a bad feature, but they'd be making devs choose them over actually performant third-party modules that use native code.

math/distributions.ts
retraigo: Made it throw a TypeError in the latest commit.

cjihrig:
The CI is looking pretty red after the last set of changes.

math/distributions_test.ts
```diff
@@ -0,0 +1,103 @@
...
+      9,
+      10,
+    ]);
+  },
+});
+
+Deno.test({
+  name: "Negative, zero, non-integer value of n returns an empty array.",
+  fn() {
+    assertEquals(uniformRange(-10, { min: 1, max: 10 }).length, 0);
```
kt3k, line 57: Looks like this now should use `assertThrows` to check it throws TypeError

kt3k:
@retraigo Thanks for many suggestions!

> ndarrays (a stricter version that involves a private variable maybe?)

Is it like ndarrays in numpy? What does stricter version mean?

> Gamma

Can you elaborate? What is the use case of this?

> Complex numbers or even some basic calculus.

This sounds like common needs for math calculation to me.

> Range generation (like the python range())

There's [`Number.range` proposal](https://github.com/tc39/proposal-Number.range) (stage 1). Maybe we should follow their API if we add it.

> But IMO it shouldn't contain math that operates on a large amount of data (like regression analysis, matrices, etc). Not because they're a bad feature, but they'd be making devs choose them over actually performant third-party modules that use native code.

I agree with this. We shouldn't compete with/re-invent something like tensorflow.js

retraigo:
> Is it like ndarrays in numpy? What does stricter version mean?
I meant how you can have ndarrays by just nesting arrays inside each other. A stricter version as in, forcing types and dimensions.

> > Gamma
> Can you elaborate? What is the use case of this?
> > Complex numbers or even some basic calculus.
> This sounds like common needs for math calculation to me.
Maybe they are a bit too much. I was just listing common things people require math libraries for.


> There's [`Number.range` proposal](https://github.com/tc39/proposal-Number.range) (stage 1). Maybe we should follow their API if we add it.
I didn't know a proposal existed for that. It does sound cool.

kt3k:
We discussed this PR internally, and we are worried about the scope of `std/math` being too vague at this point. We'd appreciate if you create your own math library in `deno.land/x` as a 3rd party module, and collect some use cases first, but we don't think accepting random items one by one in `std/math` is a good idea.

cjihrig:
I'm going to close this per https://github.com/denoland/deno_std/pull/2774#issuecomment-1290633495, but please do create a module on `deno.land/x` and iterate on it. We can revisit this after a while.

PR #2775, opened 2022-10-14
fix(node/fs): make fs.access resolve on windows

kt3k:
On windows `fileInfo.mode` is null and we treat it as 0o000 permission, and that makes any `fs.access` usage fail. This PR change it, and it treats it as 0o777 permission. It's still not perfect, but works better in many cases for npm modules. Especially this resolves denoland/deno#16255


bartlomieju (approved):
LGTM, nice fix!

Issue #2571, opened 2022-08-26
`globstar` option of `expandGlob{,Sync}` is ignored


kidonng:
**Steps to Reproduce**

```ts
import {expandGlobSync} from 'https://deno.land/std@0.153.0/fs/mod.ts'
import {assertEquals} from 'https://deno.land/std@0.153.0/testing/asserts.ts'

const prepare = () => {
	const dir = Deno.makeTempDirSync()
	for (const item of ['foo', 'foo/nested', 'bar', 'bar/nested']) {
		Deno.mkdirSync(`${dir}/${item}`, {recursive: true})
	}

	return dir
}

Deno.test('globstar true', () => {
	const dir = prepare()

	assertEquals(
		[
			...expandGlobSync(`${dir}/**`, {
				globstar: true,
			}),
		].length,
		5,
	)

	Deno.removeSync(dir, {recursive: true})
})

Deno.test('globstar false', () => {
	const dir = prepare()

	assertEquals(
		[
			...expandGlobSync(`${dir}/**`, {
				globstar: false,
			}),
		].length,
		2,
	)

	Deno.removeSync(dir, {recursive: true})
})
```

**Expected behavior**

Tests pass

**Describe the bug**

```
running 2 tests from ./test.ts
globstar true ... ok (8ms)
globstar false ... FAILED (6ms)

 ERRORS 

globstar false => ./test.ts:28:6
error: AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   5
+   2

  throw new AssertionError(message);
        ^
    at assertEquals (https://deno.land/std@0.153.0/testing/asserts.ts:183:9)
    at file:///workspaces/infra/test.ts:31:2

 FAILURES 

globstar false => ./test.ts:28:6

FAILED | 1 passed | 1 failed (46ms)

error: Test failed
```

**Environment**

- OS: Ubuntu 18.04
- deno version: deno 1.25.0
- std version: 0.153.0

kidonng:
Slightly related, why does `**` globs the parent itself? This is different from shell behavior and in my opinion harmful.

nayeemrmn:
> Slightly related, why does `**` globs the parent itself? This is different from shell behavior and in my opinion harmful.

Hmm, it seems the conventional behaviour is to match the parent when there is a prefix before the globstar but not if it's on its own:
```
$ echo examples/**
[...]

PR #2779, opened 2022-10-15
fix(fs/expandGlob): globstar false does not take effect (#2571)

nkabbara:
This is in relation to #2571. When passing globstar false, and passing the glob segment `**`, globstar false isn't being respected.


kt3k (approved):
LGTM

cjihrig (approved):

PR #2780, opened 2022-10-15
Fix(GitHub Actions)： use `GITHUB_OUTPUT` instead of deprecated `set-output` 

TakahashiIkki:
## What I did

Used `GITHUB_OUTPUT` instead of `set-output` to manage output for each step of GitHub Actions

## Why do it?

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Using `set-output` in GitHub Actions will result in a warning message and future deprecation.

## Documentation for reference

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
         shell: bash
         # excludes tests, testdata, and generated sources from coverage report
         run: |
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM. Thanks!

PR #2782, opened 2022-10-17
fix(node/fs): enable to check error thrown on invalid values of bufferSize

wafuwafu13:

### DIFF
--- a/node/_fs/_fs_opendir.ts
+++ b/node/_fs/_fs_opendir.ts
@@ -2,11 +2,7 @@
 
 import Dir from "./_fs_dir.ts";
 import { Buffer } from "../buffer.ts";
-import {
-  assertEncoding,
-  getOptions,
-  getValidatedPath,
-} from "../internal/fs/utils.mjs";
+import { getOptions, getValidatedPath } from "../internal/fs/utils.mjs";
 import { denoErrorToNodeError } from "../internal/errors.ts";
 import { validateFunction, validateInteger } from "../internal/validators.mjs";
 import { promisify } from "../internal/util.mjs";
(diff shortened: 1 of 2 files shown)


node/_fs/_fs_opendir.ts
```diff
@@ -79,8 +79,16 @@ export function opendirSync(
...
-  const encoding = options?.encoding ?? "utf8";
-  const bufferSize = options?.bufferSize ?? 32;
+  let encoding: string;
+  let bufferSize: number;
+  if (options) {
+    ({ encoding = "utf8", bufferSize = 32 } = options);
+  } else {
+    encoding = "utf8";
+    bufferSize = 32;
+  }
```
cjihrig, line 89: Node uses a utility function called `getOptions()` here. We have that [here](https://github.com/denoland/deno_std/blob/3b9fb39ddd9a99e64aff159302039dc3cfddd36e/node/internal/fs/utils.mjs#L292) (although the second argument should be updated to default to `kEmptyObject`). I think it would be preferable to use that here.

node/_fs/_fs_opendir.ts
```diff
@@ -43,15 +31,13 @@ export function opendir(
...
-  options = typeof options === "object" ? options : {};
-  const encoding = options?.encoding ?? "utf8";
-  const bufferSize = options?.bufferSize ?? 32;
-
   let err, dir;
   try {
-    assertEncoding(encoding);
-
-    checkBufferSize(bufferSize, "options.bufferSize");
+    const { _encoding, bufferSize } = getOptions(options, {
```
cjihrig, line 36: We don't need to destructure `_encoding` out if it's not being used. Same comment below.

cjihrig (commented):
LGTM with a small nit.

node/_fs/_fs_opendir.ts
wafuwafu13: fixed https://github.com/denoland/deno_std/pull/2782/commits/fe69a42ce0484c8aebe626c2578d9456147ecc96

cjihrig (approved):

Issue #1216, opened 2021-09-09
bug(fs): fs.exists/fs.existsSync throws when querying subpath of existing file.


LumaKernel:
**Describe the bug** A clear and concise description of what the bug is.

fs.exists/fs.existsSync throws when querying subpath of existing file.

**To Reproduce** Steps to reproduce the behavior:

1. In any directory,
2. `touch a.txt`
3. `deno --unstable`
4. \> `import * as fs from "https://deno.land/std@0.106.0/fs/mod.ts";`
4. \> `await fs.exists("a.txt/b");`
6. Error is thrown. See error

**Expected behavior** A clear and concise description of what you expected to
happen.

Return `false`.

**Screenshots** If applicable, add screenshots to help explain your problem.

![image](https://user-images.githubusercontent.com/29811106/132624717-37dd92eb-3680-4ca0-80fd-222593f71075.png)

**Desktop (please complete the following information):**

- OS: Ubuntu 20.04 with WSL on Windows11
- Version

```
deno 1.13.2 (release, x86_64-unknown-linux-gnu)
v8 9.3.345.11
typescript 4.3.5
```

**Additional context** Add any other context about the problem here.

Other implementations:

- Python3 `os.path.exists`: Return false.
- Node.js v14 `require("fs").exists`: Resolves false.
- Ruby 2.7 `File.exist?`: Return false.

LumaKernel:
Related: https://github.com/nodejs/node/issues/39960

Implementing `exists` generally seems problematic. Deprecation is another option. Maybe, user of `exists` should be aware of its implementation and maybe user should use `stat` itself depending on the situations.

kt3k:
Deprecating fs.exists might make sense. It always causes TOCTOU issue and fs.exists can be considered as an anti-pattern.

What do you think? @bnoordhuis @piscisaureus @ry

martin-braun:
@bnoordhuis @kt3k A valid use-case is to check if a specific file exists that a third-party binary needs before starting that binary eventually. node introduced fs.accessSync() for that case.

kt3k:
@martin-braun Sounds curious to me. Can you link to more specific context?

> A valid use-case is to check if a specific file exists that a third-party binary needs before starting that binary eventually.

This still causes TOCTOU bugs. So I think the ideal solution to that situation is that the 3rd party binary handles the situation correctly and propagate the error to the user.

martin-braun:
> Sounds curious to me. Can you link to more specific context?
[...]

PR #2785, opened 2022-10-18
feat(fs): undo deprecation of `exists` and add permission and type check to it

martin-braun:
<s>Introduces `readable`</s> Undos the deprecation and enhances `exists` of the `fs` module.

`exists` was deprecated due to TOCTOU issues, but there are valid use cases to have such function: 

- Checking for path accessibility prior using third-party tools that rely on such given path which can potentially avoid spawning expensive processes for no reason
- Checking access rights on the given path without locking it through opening or handling exceptions with `try` / `catch`
- Avoiding unnecessary per-project user abstractions of `Deno.lstat` / `Deno.stat` to compensate the deprecation of `exists`

Improvements compared to the former `exists`:

- The former `exists` was returning `false` on Windows systems if the file was not readable, but it was returning `true` on POSIX systems ([with exceptions](https://github.com/denoland/deno_std/pull/2785#issuecomment-1317926293), even if the file was not readable, this is not the case anymore: Options were added that provide an `isReadable` flag to effectively check if the file can be read as well, for Windows and POSIX systems
- Since in almost every case the user either expect a directory or a file (but not both) the flags `isDirectory` and `isFile` were added to the options to address those cases, otherwise the user would still rely on manually calling `Deno.lstat` / `Deno.stat` afterwards which would defeat the purpose of this implementation and introduce more TOCTOU / race conditions
- All implementations abstract from `Deno.stat` instead `Deno.lstat` to follow symlinks and ensure the option flags are working properly on links as well

References:

- The problem originated from #1364
- The solution was discussed in detail at #2102 which was inspired by post-reactions at #1216

cc @lambdalisue @twome @jespertheend @dionjwa @dev-nicolaos


iuioiua:
This PR would also close #2594

iuioiua:
You might want to double-check the code examples in the separate functions as they seem to incorrectly use `readable` instead of the corresponding function.

martin-braun:
I would also like to feature tests that involves an existing file / directory where the user has no read-permission on, but the current way is not very suitable. I can commit a file or directory that I cannot read, obviously.

To include those tests, I'd need to create a file and directory, remove their user permissions, execute the test, add user permissions back and remove the file and directory again.

Is this overkill? Is this recommend?

martin-braun:
> You might want to double-check the code examples in the separate functions as they seem to incorrectly use `readable` instead of the corresponding function.

Maybe it's too late for me, could you please comment on the affected lines, directly?

iuioiua:
I like this. My concerns, neither of which, I have strong opinions on:
1. Would `readableSymlink` also be something we want?
2. I think function names like `isReadable`, `isReadableDir`, etc. would be clearer to the developer.

fs/readable.ts
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Test whether or not the given path is readable by checking with the file system. To check simultaneously if the path is either a file or a directory please use `readableFile` or `readableDir` instead.
+ *
+ * Note: do not use this function if performing a check before another operation on that file. Doing so creates a race condition. Instead, perform the actual file operation directly.
+ *
+ * Bad:
+ * ```ts
+ * import { readable } from "https://deno.land/std@$STD_VERSION/fs/mod.ts";
+ *
+ * if (await readable("./foo")) {
```
iuioiua, line 52: Incorrectly uses `readable` instead of `readableSync`.

fs/readable_dir.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Test whether or not the given path is readable directory by checking with the file system.
+ *
+ * Note: do not use this function if performing a check before another operation on that file. Doing so creates a race condition. Instead, perform the actual file operation directly.
+ *
+ * Bad:
+ * ```ts
+ * import { readableDir } from "https://deno.land/std@$STD_VERSION/fs/mod.ts";
+ *
+ * if (await readableDir("./foo")) {
```
iuioiua, line 51: Should `readableDir` instead be `readableDirSync`?

fs/readable_file.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * Test whether or not the given path is a readable file by checking with the file system.
+ *
+ * Note: do not use this function if performing a check before another operation on that file. Doing so creates a race condition. Instead, perform the actual file operation directly.
+ *
+ * Bad:
+ * ```ts
+ * import { readableFile } from "https://deno.land/std@$STD_VERSION/fs/mod.ts";
+ *
+ * if (await readableFile("./foo.txt")) {
```
iuioiua, line 51: Should this instead be `readableFileSync`?

martin-braun:
@iuioiua Thank you, I was missing the comments, copy/paste caught me here.

> 1. Would `readableSymlink` also be something we want?

Please have a look at our final posts in the discussion thread. 

Since we abstract from `stat` instead of `lstat`, symlinks will be followed. The usage of `readable*` should default to a behavior that is expected to the user. 

`readable*` should primarily be used to ensure a file or directory is available for any third-party process. If the Deno script needs to work with the file or directory directly, it should access it directly and thus not use `readable*` at all to prevent race-conditions. 

In my experience, most tools follow symlinks, so should `readable*` in my honest opinion. 

A user might create a symlink to have a specific folder or file in another location (i.e. to sync it with the cloud). A Deno script should not break due to this. 

So, we abstract from `stat` not `lstat`, a `readableSymlink` would imply that `readableDir` or `readableFile` do not follow symlinks.

Checking a symlink is such a rare occurrence that we would prefer the user falls back to an own `lstat` implementation. `readable*` has been created to fulfill just a common case, thus I think `readableSymlink` is not necessary.

> 2. I think function names like `isReadable`, `isReadableDir`, etc. would be clearer to the developer.

I'm not sure. I tried to follow existing code, there is no such file starting with `is_` and `readable` is an adjective term. `exists` was also named `exists` and not `doesExist`, but I have no strong opinion about it.

kt3k:
> I think function names like isReadable, isReadableDir, etc. would be clearer to the developer.

I agree with this opinion. We already use the word `readable` for `ReadableStream` in several places (for example FsFile.readable is ReadableStream of its contents. ref. https://doc.deno.land/deno/stable/~/Deno.FsFile ). So this usage of `readable` sounds confusing to me.

iuioiua:
Yeah, when I first saw the PR title, I thought it was related to streams 😅

martin-braun:
I'm currently rewriting the tests to get rid of usage of `testdata` files and to be able to include permission tests and to my surprise there was [some information](https://github.com/denoland/deno_std/issues/1216#issuecomment-937722077) that caught me off-guard: 

> 2. "Exists" is a nebulous concept. `node -p 'fs.existsSync("/root/.ssh/id_rsa")'` prints false even though the file exists on my system. It's there, it's just not accessible.

Thus, I falsy assumed that `Deno.stat` is unable to access a file without permissions spreading around that I cannot confirm. I'm not sure about Linux, but on macOS this test will fail: 

```ts
async function isReadable(path: string | URL): Promise<boolean> {
  try {
    await Deno.stat(path);
    return true;
  } catch (error) {
    if (error instanceof Deno.errors.NotFound) {
      return false;
    }
    throw error;
  }
}

Deno.test("failing test", async function () {
  const tempDirPath = await Deno.makeTempDir();
  const tempFilePath = path.join(tempDirPath, "0.ts");
  try {
    await Deno.chmod(tempFilePath, 0o000);
    assertEquals(await isReadable(tempFilePath), false); // exists, but not readable
  } catch (e) {
    throw e;
  } finally {
    await Deno.remove(tempDirPath, { recursive: true });
  }
});
```

So it seems a file that a user cannot read its contents from can still be accessed with `Deno.stat`. Unfortunately, `FileInfo.mode` is unstable which makes it unacceptable to alter `isReadable` with a permission check. I don't really think refactoring to `exists`, `existsDir` and `existsFile` would be a satisfying solution, but at this point I see no other way.

Even when I was able to check the permissions, it would rather encourage a re-implementation of node's `fs.access` instead.

I'm seeking some advice at this point.

jespertheend:
I'll have to admit I also expected `Deno.stat("/root/.ssh/id_rsa")` to reject.
Fwiw, `FileInfo.mode` is unstable, but it does not require the `--unstable` flag. So I'm not sure how unacceptable a permission check would really be.

To be honest, I think returning true in the case of failing permissions would be fine for most use cases. But then `isReadable*` wouldn't be the correct name anymore. That would bring us back into `exists()` territory which is less ideal I think.

martin-braun:
@jespertheend It's good to know that `FileInfo.mode` doesn't require the `--unstable` flag. The description probably says it's unstable, because it's `null` on Windows, `Deno.chmod` doesn't work properly on Windows as well (see #4357).

I booted Windows, created two files (`test.txt` and `test2.txt`) in `C:\ProgramData`. I created a new user `testuser` and denied any permissions to him for `test.txt`: 

Screen Shot 2022-10-18 at 8 11 44 PM

Now I launched deno as `testuser` with `runas /user:"testuser" deno` and tried to access both files with this result: 

```deno
> Deno.stat("C:\\ProgramData\\test.txt")
Promise {
  <rejected> PermissionDenied: Access is denied. (os error 5), stat 'C:\ProgramData\test.txt'
    at async Object.stat (deno:runtime/js/30_fs.js:314:17)
}
> Deno.lstat("C:\\ProgramData\\test.txt")
Promise {
  <rejected> PermissionDenied: Access is denied. (os error 5), stat 'C:\ProgramData\test.txt'
    at async Object.lstat (deno:runtime/js/30_fs.js:297:17)
}
> Deno.lstat("C:\\ProgramData\\test2.txt")
Promise {
  {
    isFile: true,
    isDirectory: false,
    isSymlink: false,
    size: 0,
    mtime: 2022-10-18T18:16:46.311Z,
    atime: 2022-10-18T18:16:46.311Z,
    birthtime: 2022-10-18T18:16:46.311Z,
    dev: null,
    ino: null,
    mode: null,
    nlink: null,
    uid: null,
    gid: null,
    rdev: null,
    blksize: null,
    blocks: null
  }
}
```

So, on Windows `stat` and even `lstat` fail to access the restricted file with a `PermissionError`. So when catching this error and returning `false` the function name `isReadable*` is very correct here.

On \*nix based systems `stat` will not fail when the file cannot be read, but `FileInfo.mode` should not be `null`, thus allow us to additionally check if the file is readable.

On Windows I tested the same with directories and the behavior is the same to files.

So in favor of consistent behavior, we should include the permission check, since it will raise an error on Windows anyways. I will catch `PermissionError` and check the value of `FileInfo.mode` if it's not `null`.

jespertheend:
Would it make sense to release without the permissions check first and then add it in a follow up PR once `Deno.getUid()` and `Deno.getGid()` are stable? Otherwise reading the file sounds less than ideal, but could also work as a temporary solution. I can't think of any other way.

martin-braun:
@jespertheend It would be possible, although my biggest concern is the changing behavior of the implementation between the platforms and between concurrent releases. I really want to thrive for a better solution.

I committed an implementation (incl. tests) that relies on the `--unstable` flag until #2791 is fixed. The implementation is complete, the tests lack some more test cases: 

- Permission checks don't include directory / file checks with items that you are not owner of, it's probably overkill given the simplicity of the implementation, especially since the tests should not create users / groups on your system temporary
- I'm unable to test symbolic links whose index (the link itself) has missing permissions (only relevant for MacOS and a few other BSD variants, where this is possible through `chmod -h`), the issue boils down to libc (which is used by Rust, which is used by Deno) not allowing to modify symlink permissions. ([I poked the Rust team about it](https://internals.rust-lang.org/t/std-set-perm-and-symlinks/17592), it might works to set those with `Deno.run` directly, though)

Whatsoever, these issues are neglectable, but should be mentioned.

martin-braun:
So far for `isReadable`. In regards of `exists` to address cases in which it is necessary to know if a file/directory exists without caring for its permissions, how about refactoring `isReadable` to `exists`, but providing an `isReadable` flag on the options to further reduce redundancy?

martin-braun:
I need to recover the scenes and validate cases where `--allow-read` is missing.

martin-braun:
> I need to recover the scenes and validate cases where `--allow-read` is missing.

Done, but the PR is blocked by https://github.com/denoland/deno/issues/15576, it won't work without this issue being fixed. After that this PR should be ready.

Btw @jespertheend, I had some new recent findings, at least for MacOS: `Deno.error.PermissionDenied` can indeed be raised by `Deno.stat` on POSIX systems. On MacOS, it will raise only and only if you try to `Deno.stat` a protected **directory** behind a link, so the behavior is not exclusive to Windows.

lambdalisue:
@martin-braun 

> Done, but the PR is blocked by https://github.com/denoland/deno/issues/15576, it won't work without this issue being fixed. After that this PR should be ready.

It seems you are using `await` version in this PR so I wonder why the issue is blocking this PR?

jespertheend:
https://github.com/denoland/deno/issues/15576 has been fixed now. Would it be possible to merge this or does this still need some work?
Fwiw, importing from `https://raw.githubusercontent.com/denoland/deno_std/7032f6856193b84f2046b6762698939754469c02/fs/mod.ts` seems to work flawlessly for me.

martin-braun:
@jespertheend Didn't forget it, but had my focus on other stuff that pays the bill. At the moment I can't really spend more hours on this, so I hope we can pass the tests without much hassle. My last status was some tests fail on other systems, I can only test on macOS. Maybe someone on another platform can assist if the logs aren't clear about the issue.

fs/exists.ts
```diff
@@ -1,24 +1,53 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
+export interface ExistsOptions {
```
iuioiua, line 3: Are you able to state the default values of these properties?

fs/README.md
iuioiua, line 1: Can you delete this file from the PR? The Standard Library no longer has per-module READMEs.

fs/exists.ts
```diff
@@ -1,24 +1,53 @@
...
+  isFile?: boolean;
+}
+
 /**
- * Test whether or not the given path exists by checking with the file system.
+ * Test whether or not the given path exists by checking with the file system. Please consider to check if the path is readable and either a file or a directory by providing additional `options`:
  *
- * Note: do not use this function if performing a check before another operation on that file. Doing so creates a race condition. Instead, perform the actual file operation directly.
+ * ```ts
+ * import { isReadable } from "https://deno.land/std@$STD_VERSION/fs/mod.ts";
```
iuioiua, line 24:
```suggestion
 * import { exists } from "https://deno.land/std@$STD_VERSION/fs/exists.ts";
```

martin-braun:
Almost there. For record:

**ci / test (canary, windows-2022) (pull_request)**

```
[fs] existsNotExist ... ok (40ms)
[fs] existsNotExistSync ... ok (30ms)
[fs] existsFile ... ok (34ms)
[fs] existsFileLink ... FAILED (26ms)
[fs] existsFileSync ... ok (32ms)
[fs] existsFileLinkSync ... FAILED (30ms)
[fs] existsDir ... FAILED (31ms)
[fs] existsDirLink ... FAILED (31ms)
[fs] existsDirSync ... FAILED (15ms)
[fs] existsDirLinkSync ... FAILED (16ms)
[fs] existsPermission test exists on real file without --allow-read ... ok (316ms)
[fs] existsPermission test existsSync on real file without --allow-read ... ok (311ms)
[fs] existsPermission test exists on real file with --allow-read ... ok (265ms)
[fs] existsPermission test existsSync on real file with --allow-read ... ok (265ms)
[fs] existsPermission test exists on fictional file without --allow-read ... ok (266ms)
[fs] existsPermission test existsSync on fictional file without --allow-read ... ok (265ms)
[fs] existsPermission test exists on fictional file with --allow-read ... ok (266ms)
[fs] existsPermission test existsSync on fictional file with --allow-read ... ok (233ms)
```

iuioiua:
`_tools/check_deprecation.ts` also has some code that specifically relates to `fs.exists()`, which can be removed.

jespertheend:
@martin-braun I have a Linux and Windows system available to try it out on. I'm not very familiar with the Windows filesystem but I can take a look later today if you like.

martin-braun:
@jespertheend Please do and thanks in advance. My Parallels license has expired and I currently have limited time to setup a new Windows VM in a different system. I also have disk issues at the moment (limited space, errors when trying to expand its space) as well as I couldn't get UTM (QEMU) to work in a short time. Would be good if I could focus on all that later.

Basically, you want to `deno test --allow-all --unstable --filter "[fs] exists" fs/exists_test.ts` while slowly commenting out each sub test to figure out which test fails and why it does.

jespertheend:
@martin-braun  It seems like the there are two issues:
- `Deno.chmod` not being supported, causing `NotSupported: The operation is not supported` to be thrown. These can be fixed by wrapping al of the `Deno.chmod` calls in the `finally` blocks with the same `if (Deno.build.os !== "windows")` statement as is used above it.
- `Deno.symlink` throws `Error: A required privilege is not held by the client`. I'm not sure if ci is being run with the right privileges, if so I don't think this should be an issue. If I run `deno test` in cmd as administrator I don't run into this error.

If you want I can commit the if statements to fix the first issue. Though I think you would have to give me write access to your fork, so it might be easier to just add these yourself. There are 6 places in total that need this.

martin-braun:
@jespertheend First of all, thanks a lot. The first issue was an easy fix, thanks to your help. I guess I forgot about these `chmod` calls in `finally`. In regards of the 2nd issue: Could you please check the contents of `tempLinkDirPath` please? It seems the path that is returned from `Deno.makeTempDir()` is not writable by the current user on Windows machines. Since it works as Admin, I assume the `--allow-write` is not missing during these tests. Maybe, you could also check if `const tempDirPath = await Deno.makeTempDir({ prefix: 'deno_exists_test_' });` fixes the issue?

jespertheend:
@martin-braun The folder is completely empty. Or at least that's the case for the 'existsDirLink' test. The prefix only adds a prefix to the name of the directory, but it doesn't change the location so it still runs into the same issue.

jespertheend:
Hey I think this should be good to go once the next workflow runs.
I created a PR of this branch on my own fork in order to make the workflow run and [all tests seem to pass](https://github.com/jespertheend/deno_std/actions/runs/4492939327/jobs/7903498701).

fs/exists.ts
```diff
@@ -27,39 +59,89 @@
...
- * @deprecated (will be removed after 0.157.0) Checking the state of a file before using it causes a race condition. Perform the actual operation directly instead.
  */
-export async function exists(filePath: string | URL): Promise<boolean> {
+export async function exists(
+  path: string | URL,
+  options?: ExistsOptions,
+): Promise<boolean> {
   try {
-    await Deno.lstat(filePath);
+    const stat: Deno.FileInfo = await Deno.stat(path);
```
kt3k, line 68: type annotation seems unnecessary as it's inferred from the assigned value

```suggestion
    const stat = await Deno.stat(path);
```

Can you also remove this kind of type annotations for local variables?

fs/exists_test.ts
```diff
@@ -6,128 +6,445 @@ import { exists, existsSync } from "./exists.ts";
 const moduleDir = path.dirname(path.fromFileUrl(import.meta.url));
 const testdataDir = path.resolve(moduleDir, "testdata");
 
+Deno.test("[fs] existsNotExist", async function () {
+  const tempDirPath: string = await Deno.makeTempDir();
+  try {
+    assertEquals(await exists(path.join(tempDirPath, "not_exists")), false);
+  } catch (error) {
+    throw error;
```
kt3k, line 14: This catch block seems unnecessary (if you don't catch it, it's thrown as is)

(Can you also remove the similar blocks in other test cases?)

fs/exists_test.ts
```diff
@@ -6,128 +6,445 @@ import { exists, existsSync } from "./exists.ts";
...
+  }
+});
+
+/**
+ * Scenes control additional permission tests by spawning new Deno processes with and without --allow-read flag.
+ */
+interface Scene {
+  read: boolean; // true to test with --allow-read
+  sync: boolean; // true to test sync
+  fictional: boolean; // true to test on non existing file
```
kt3k, line 357: `fictional` sounds confusing to me. How about `existing` (with reversed boolean value)?

martin-braun:
I felt too confident for a moment. :) Should pass the tests now, hopefully also the Windows tests.

kt3k:
Core team discussed this last week and now we are in favor of landing this. Thank you for your effort and discussions for long time

kt3k (approved):
LGTM

iuioiua:
image

lambdalisue:
:tada:

lino-levan:
🎉🎉🎉 Nice work! Glad it landed.

martin-braun:
I'm glad we made it eventually. :)
Thanks for your patience to everyone. 🙏

twome:
Thank you so much for all your work!

PR #2787, opened 2022-10-18
feat(node): list remaining Node tests in documentation

### LINKED ISSUE #2771: suggestion: measure Node compatibility
Node compatibility could be measured by how many tests within `std/node` exist and are passing.

I suggest having a programmatic way of measuring which and how many Node tests are passing, with clear and readable results easily accessible. It'd allow a contributor to easily find which tests to pursue. Doing so would also make raising issues for failing/missing Node tests redundant.

I explored this idea [here](https://github.com/iuioiua/deno_std/tree/measure-compat). By running `deno task measure-compat`, an object containing the lists of passing and remaining tests within `std/node` is printed.

The question is: how do we mould this data into a readable format? A markdown file, say `std/node/compat.md`, could be generated with the two lists present. This is a simple solution. However, there are over 4,500 tests. The file would be very long. A badge on `std/node/README.md` could point to this file.

An alternative could be an interactive web app, showing off fresh and using this data to make a searchable and interactive list. This could be easier for the user but more complex to maintain. A badge on `std/node/README.md` could also point to this web app.

iuioiua:
This PR adds a tool that lists the remaining Node tests in the documentation. This eliminates the need for issues that track remaining Node tests, of which plenty exist. I thought having a `TODO.md` to track these tests is the simplest way to get started going about doing this. It's dynamic and is refreshed in CI every commit to stay fresh.

In the future, we can add a dynamic badge to `node/README.md` under the title that summarises how many out of the total tests are passing, with a link to the remaining tests doco.

Already passing tests were omitted because knowing what they are doesn't provide much benefit. However, knowing _how many_ are passing might be helpful (see the previous paragraph).

Closes #2771.


_tools/node_list_remaining_tests.ts
```diff
@@ -0,0 +1,67 @@
...
+    .sha;
+}
+
+async function getNodeTests(sha: string): Promise<string[]> {
+  const url = NODE_BASE_URL + "/git/trees/" + sha + "?recursive=1";
+  const response = await fetch(url);
+  const body = await response.json();
+
+  return body.tree
+    .filter(({ path }: Object) => path.endsWith(".js"))
```
kt3k, line 26: I think the items under `fixtures/` should be filtered out as they are not test cases. Also items under `common/` look like the testing utilities, not test cases.

iuioiua: Ok, this now has me thinking. Which [Node test directories](https://github.com/nodejs/node/tree/main/test) do we want to ignore for the foreseeable future completely? I think `testpy` should also be ignored.

@cjihrig, do you have any additional thoughts?

_tools/node_list_remaining_tests.ts
```diff
@@ -0,0 +1,67 @@
...
+async function getDenoTests(): Promise<string[]> {
+  const files: string[] = [];
+  const denoTestDir = new URL("../node/_tools/test", import.meta.url);
+
+  for await (const { path } of walk(denoTestDir, { exts: [".js"] })) {
+    files.push(path);
+  }
+
+  return files
+    .map((file: string) => file.replace(denoTestDir.pathname + "/", ""));
```
cjihrig, line 39: You could do the `replace()` at the same time you call `files.push(path)` a few lines up.

_tools/node_list_remaining_tests.ts
cjihrig: We can probably skip addons, async-hooks, cctest, doctool, embedding, fixtures, fuzzers, js-native-api, node-api, overlapped-checker, report, testpy, tick-processor, tools, v8-updates, wasi, and wpt. As new functionality such as Node-API is added here, we could enable those tests.

To be honest, I kind of feel like maintaining this might not be worth the work. Having a script that dumps out the missing tests could be useful, but I'm not sure that we need to maintain the list in the repo. Just my opinion - if others feel it's useful then by all means, add it.

iuioiua: I plan to have an array containing the list of Node test directory names that should be ignored. In the future, I believe the extent of maintenance of this tool might include removing names from this array to have them to be tested. Nothing else beyond that, as far as I can see.

IMO, this is a cheap win - having a clear and up-to-date list of Node tests for contributors to pursue.

iuioiua:
I plan to have an array containing the list of Node test directory names that should be ignored. In the future, I believe the extent of maintenance of this tool might include removing names from this array to have them to be tested. Nothing else beyond that, as far as I can see.

IMO, this is a cheap win - having a clear and up-to-date list of Node tests for contributors to pursue.

Either way, it's best to look at the new node/TODO.md file and only proceed if it feels like it achieves its intended purpose.

cjihrig:
> look at the new node/TODO.md file and only proceed if it feels like it achieves its intended purpose.

I do think the file serves the intended purpose, but if we are going to maintain it, perhaps it should live in `node/_tools` and be run as part of the script that updates the test suite instead of running it on every CI run. I would dump the file there too so that it lives next to `config.json`.

Also, since the file will be regenerated from time to time, the checkboxes will get overwritten. That's fine IMO, but wanted to point it out.

iuioiua:
Actually, yes, good ideas. I've done so now. Please check out how I've done the tasks in `deno.json` now. Do they look good?

node/_tools/list_remaining_tests.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+async function getNodeTests(sha: string): Promise<string[]> {
+  const url = NODE_API_BASE_URL + "/git/trees/" + sha + "?recursive=1";
+  const response = await fetch(url);
+  const body = await response.json();
+
+  return body.tree
+    .filter(({ path }: Object) =>
+      path.includes("test") && path.endsWith(".js") &&
```
cjihrig, line 47: I think you could tighten this check up a little bit:

```suggestion
      path.includes("/test-") && path.endsWith(".js") &&
```

cjihrig (approved):

cjihrig:
Would you mind rebasing to pick up a fix for the CI failures.

PR #2793, opened 2022-10-19
fix(node/net): modify close event timing

### LINKED ISSUE #2788: npm: getting available port using `npm:get-port@5` doesn't work
The following script throws `error: Uncaught AddrInUse: Address already in use (os error 48)`

```js
import getPort from "npm:get-port@5";
import { createServer } from "https://deno.land/std@0.160.0/node/http.ts";

const availablePort = await getPort({ port: getPort.makeRange(5555, 5600) });
createServer((_, res) => res.end("hi")).listen(availablePort);
```

(the 'available port' is not actually available)

If we delay the timing of listening a little bit, it works (the below script works):

```js
import getPort from "npm:get-port@5";
import { createServer } from "https://deno.land/std@0.160.0/node/http.ts";

const availablePort = await getPort({ port: getPort.makeRange(5555, 5600) });
setTimeout(() => createServer((_, res) => res.end("hi")).listen(availablePort), 0);
```

This issue seems causing `AddrInUse` error in `deno run npm:prisma studio`. ref https://github.com/denoland/deno/issues/16282

kt3k:
This PR delays the timing of 'close' event when `.close()` is called on net.Server. With the current implementation, 'close' event is emitted a bit too early and the port is not available yet. That causes the module `get-port` failing to return 'available' port. This PR uses `setTimeout` instead of `nextTick` for calling `_emitCloseNT` and the timing is delayed, and that makes `get-port` working.

closes #2788

related https://github.com/denoland/deno/issues/16282

### DIFF
--- a/node/net.ts
+++ b/node/net.ts
@@ -2414,10 +2414,14 @@ export class Server extends EventEmitter {
       return;
     }
 
+    // We use setTimeout instead of nextTick here to avoid EADDRINUSE error
+    // when the same port listened immediately after the 'close' event.
+    // ref: https://github.com/denoland/deno_std/issues/2788
     defaultTriggerAsyncIdScope(
       this[asyncIdSymbol],
-      nextTick,
+      setTimeout,
       _emitCloseNT,
+      0,
       this,
     );
   }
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, nice fix

PR #2795, opened 2022-10-19
feat(streams): `ByteSliceStream`

iuioiua:
This PR adds `ByteSliceStream`, which is the streams version of `readRange` from `std/io`.

Part of #1986.


streams/buffer.ts
```diff
@@ -232,3 +232,43 @@ export class LimitedTransformStream<T> extends TransformStream<T, T> {
...
+            chunk = chunk.slice(start - (this.#offset - chunk.byteLength));
+          }
+
+          if (end <= this.#offset) {
+            chunk = chunk.slice(0, chunk.byteLength - (this.#offset - end));
+            controller.enqueue(chunk);
+            controller.terminate();
+          }
+
+          controller.enqueue(chunk);
```
cjihrig, line 269: Maybe I'm missing something, but this whole block seems slightly off. Should the two separate `if` statements be an `if...else`? Why is `chunk` potentially assigned twice? Why is `controller.enqueue()` possibly called twice depending on the path through the code?

kt3k:
How about dropping `Transform` from the class name?

Many TransformStream APIs don't have `Transform` in their name.

- TextDecoderStream, TextEncoderStream, CompressionStream, DecompressionStream from Web API
- CsvStream, JSONParseStream, JSONStrigifyStream, TextLineStream, DelimiterStream from std

(`LimitedTransformStream` and `LimitedBytesTransformStream` look rather exceptions.)

I also suggest using `Slice` instead of `Ranged`. How about `ByteSliceStream` or `SliceByteStream`?

iuioiua:
Thanks for the suggestions! @cjihrig, does the code make a bit more sense now? @kt3k, I've gone with `SliceByteStream` to more align with the adjacent `LimitedBytesTransformStream`.

streams/buffer.ts
```diff
@@ -232,3 +232,44 @@ export class LimitedTransformStream<T> extends TransformStream<T, T> {
...
+        assert(start >= 0, "`start` must be greater than 0");
+        end += 1;
+      },
+      transform: (chunk, controller) => {
+        this.#offsetStart = this.#offsetEnd;
+        this.#offsetEnd += chunk.byteLength;
+        if (this.#offsetEnd > start) {
+          if (this.#offsetStart < start) {
+            chunk = chunk.slice(start - this.#offsetStart);
+          }
```
kt3k, line 263: I think we can reduce the number of `slice` calls (which is expensive) if we store `start - this.#offsetStart` value and use it below.

iuioiua:
Actually, I realise this implementation is inefficient. Let me tweak it.

timreichen:
@kt3k  How about just `SliceStream`?

iuioiua:
@timreichen, I think that's best reserved for the non-binary-data version of this class.

streams/buffer.ts
iuioiua: Both `slice`s are called only in a rare case - when both `start` and `end` lie just outside the chunk for a given iteration. I think the current code strikes a good balance between clarity and efficiency. What do you think?

kt3k: hmm ok. Now this makes sense to me.

kt3k (approved):
LGTM

kt3k:
We internally chatted about the name candidates. `ByteSliceStream` seemed having enough support.

PR #2799, opened 2022-10-21
refactor(http): use hashing functions from `std/crypto`

iuioiua:
This PR removes duplicated hashing functionality in the file server and its testing code and instead uses code from `std/crypto`.


kt3k (approved):
LGTM. Thanks!

kt3k:
Note: We used local `fnv1a` because of this change https://github.com/denoland/deno_std/pull/2116

Because we now have `fnv1a` in `std/crypto`, so it should be reasonable to replace it with `std/crypto`

PR #2800, opened 2022-10-21
chore(node): update vendored streams implementation

iuioiua:
This PR:
1. Updates the Node streams implementation to v4.2.0.
2. Removes the manual version of `test-stream-writable-change-default-encoding` since `std/node` is now on Node v18.


cjihrig (approved):
LGTM, thanks!

PR #2801, opened 2022-10-23
#2663 feat(encoding/front_matter): add support for different formats of front matter

### LINKED ISSUE #2663: Support JSON in encoding/front_matter
It would be nice if our `encoding/front_matter` could support JSON syntax in addition to YAML

schwarzkopfb:
an approach to close #2663 and simplify inclusion of other formats


kt3k:
It feels unfortunate that `front_matter` always depends on both yaml and toml even when the user only needs one of them. Is it possible for users to have only necessary parsers in dependencies?

schwarzkopfb:
@kt3k the only idea comes to my mind ATM is that we can change the signature of `function extract<T = unknown>(str: string): Extract<T>` to be async and then lazily import the necessary parser on demand.

I'm not sure that worth the effort (and the break of backwards compat). However it's an easy thing to implement, so I can do that quickly if you encourage that solution. 

WDYT?

kt3k:
> @kt3k the only idea comes to my mind ATM is that we can change the signature of function extract<T = unknown>(str: string): Extract<T> to be async and then lazily import the necessary parser on demand.

That sounds good to me 👍 

Does anyone have any concern with this solution?

kt3k:
> I don't think it's a good idea - having dynamic imports implies that proper --allow-net permission must be given to enable these imports.

Right. Let's avoid it.

> Maybe just create separate entry-points for TOML and YAML based parsers?

That works. `encoding/front_matter/yaml.ts` for yaml-front-matter, `encoding/front_matter/toml.ts` for toml-front-matter.

Another option might be to provide generator of `extract` function which accept `(key, parser)` pair and the user can create custom `extract` function like:

```js
const extract = generateExtract({ toml: tomlParser });
// now extract can parse toml front matter.
```

schwarzkopfb:
@kt3k @bartlomieju I used both of your suggestions and also added a `front_matter/any.ts` because I didn't want to loose the auto-recognition functionality of the format. I think customisable static site generators/blog engines can profit from that. Please check my changes.

schwarzkopfb:
Also, could anyone provide me some hint on how to fix the broken CI checks?

kt3k:
@schwarzkopfb 
Thanks for the updates! The structure looks good to me. `front_matter/any.ts` makes sense to me. A few comments on details:

- `recognize` APIs in `front_matter/*.ts` don't look useful as a public API. How about making these private?
- Let's export `extract` and `test` (with yaml parser) from `encoding/front_matter.ts` with `@deprecated` tag to not immediately break the dependent code.

schwarzkopfb:
@kt3k 
* making `recognize` private totally makes sense 👍 TBH I wasn't sure that it's a good idea to expose that at all
* could you advise me about the planned deprecation version? What should go into `@deprecated (will be removed after x.y.z)`?

schwarzkopfb:
> * could you advise me about the planned deprecation version? What should go into `@deprecated (will be removed after x.y.z)`?

for now I added +10 versions from the current, let me know if I'd change that

kt3k:
> for now I added +10 versions from the current, let me know if I'd change that

10 minor versions sound good enough to me.

kt3k (approved):
LGTM

schwarzkopfb:
FYI: I found [this pull request](./2604) which addressed an issue that is already solved by this PR.

kt3k:
> Honestly, I think that having an extra file with the only benefit of letting the user exclude the /yaml part from the import url doesn't seem the way to go

Ok. Let's make it explicit.

kt3k:
Thanks for your contribution! @schwarzkopfb

PR #2805, opened 2022-10-24
feat(crypto): `createHash` utility

iuioiua:
This PR introduces the `createHash` utility, which takes care of the usual boilerplate of creating hashes.

Its practical uses can be seen within `http/file_server.ts` and `http/file_server_test.ts`.

I can understand if this utility is too trivial to be public. However, it may still be helpful internally.


cjihrig:
I think we should start with this as an internal utility for a while and see if it warrants being added to the public API, but other than that the code here LGTM.

kt3k:
> I think we should start with this as an internal utility for a while and see if it warrants being added to the public API

I agree with this opinion. I'm not sure it's worth the public API as it looks a little trivial. Let's make it private at this point.

iuioiua:
No problems. I've now made `createHash` private.

cjihrig:
Would you mind rebasing to pick up a fix for the CI failures.

cjihrig (approved):

PR #2813, opened 2022-10-26
chore: update to stabilized Deno.consoleSize() API

bartlomieju:


node/assertion_error.ts
```diff
@@ -26,9 +26,7 @@ import { stripColor as removeColors } from "../fmt/colors.ts";
 
 function getConsoleWidth(): number {
   try {
-    // TODO(bartlomieju):
-    // @ts-ignore Deno.consoleSize is getting stabilized and its signature changes
-    return Deno.consoleSize(Deno.stderr.rid).columns;
+    return Deno.consoleSize().columns;
```
schwarzkopfb, line 29: My idea behind the `try-catch` block around it was that if the process were running without `--unstable` then `Deno.consoleSize` was `undefined`, so we falled back to a common value instead of crashing.
Now it's stable, so I'm not sure that's necessary. Furthermore this whole functionality is wrapped into `getConsoleWidth()` just because I needed this fallback. Probably that can be factored out as well. WDYT?

bartlomieju: This function can still throw if stdio is not a TTY so I think it's resonable to fall back to some default value.

PR #2814, opened 2022-10-26
chore(node): update node test suite to 18.12.0 (LTS)

cjihrig:
Node 18 entered LTS yesterday. Let's update the Node test suite and see what breaks (🤞 hopefully not too much since we were already on 18.8.0).


bartlomieju (approved):
LGTM

PR #2817, opened 2022-10-26
chore: allow to return the proper type without casting

yordis:
Hey peeps, I have to keep manually casting things and force the compiler to cast to a given type, but even so I keep getting the following error:

> Deno: Type 'unknown' is not assignable to type 'SemanticConventionFile'.

```ts
const data: SemanticConventionFile  = parse(content)
```

Hopefully this will allow me to do 

```ts
const data = parse<SemanticConventionFile>(content)
```

### DIFF
--- a/encoding/_yaml/parse.ts
+++ b/encoding/_yaml/parse.ts
@@ -14,7 +14,7 @@ export type ParseOptions = LoaderStateOptions;
  * Returns a JavaScript object or throws `YAMLException` on error.
  * By default, does not support regexps, functions and undefined. This method is safe for untrusted data.
  */
-export function parse(content: string, options?: ParseOptions): unknown {
+export function parse(content: string, options?: ParseOptions): any {
   return load(content, options);
 }
 


kt3k:
> Hey peeps, I have to keep manually casting things and force the compiler to cast to a given type, but even so I keep getting the following error:
>
> > Deno: Type 'unknown' is not assignable to type 'SemanticConventionFile'.
>
> ```ts
> const data: SemanticConventionFile  = parse(content)
> ```

I'm not sure this change is useful. You should be able to do the below instead of your example:

```ts
const data = parse(content) as SemanticConventionFile;
```

which works almost the same as your suggested example (`const data = parse<SemanticConventionFile>(content)`)

kt3k:
I'd like to hear more opinions from the community:

```ts
const data = yaml.parse(input) as MyData;
```

```ts
const data = yaml.parse<MyData>(input);
```

When the meanings of the above 2 are the same, do people prefer the 1st or 2nd one?

timreichen:
I prefer to assign types with the `as` keyword. Imo generics are usually used if the function depend on the type internally. So I would not recommend it for a `parse` function.

yordis:
@timreichen that is good point, and I `dont` disagree with you, but that means you need to use `any` over `unknown`; otherwise, it is hard to typecast, as far as I can tell when using `any` TS is less problematic and friendly to infer things or allow you to do things like `const data: SemanticConventionFile` that I couldn't do and I had to switch to `as SemanticConventionFile`.

```ts
interface JSON {
    parse(text: string, reviver?: (this: any, key: string, value: any) => any): any;
}
```

lino-levan:
I agree with @timreichen. If this PR should do anything, it should change parse's return type from `unknown` to `any`. This should really be the case for all parsing functions unless a more specific definition is possible.

yordis:
✅ done

iuioiua:
I agree that `parse` should return `any`. I don't think a generic type is needed. And I slightly prefer option 3 for the same reason as @lino-levan. In other words, LGTM!

lino-levan:
LGTM

kt3k:
I don't think changing the return type to `any` is a good idea. It makes the function less type-safe.

With this change, the code like the below doesn't cause type error anymore, while it will never succeed in runtime.

```js
const data = yaml.parse(input);
data.hello();
```

This change also feels conflicting with the fact that we discourage the usage of `any` in the lint rule `no-explicit-any`.

iuioiua:
Hm... Those are good points...

kt3k:
Closing as it's stale and there's no consensus on the direction

Issue #2798, opened 2022-10-21
Env file loading with empty value is broken since 0.147


acrodrig:
**Describe the bug**

It is not possible to load `.env` files with empty values since version 0.147. It used to work until version 0.146.

**Steps to Reproduce**

Assume the following `.env` file:

```env
A="foo"
B="bar"
C=""
```

And `test.ts` below:

```typescript
// Any version after 0.147 will break all the way up to 0.160
import "https://deno.land/std@0.160.0/dotenv/load.ts";
console.log(Deno.env.get("A"));
```

It breaks as shown here:

```
$ deno run -A test.ts 
error: Uncaught TypeError: Cannot read properties of undefined (reading 'trim')
      : unquoted.trim();
                 ^
    at parse (https://deno.land/std@0.147.0/dotenv/mod.ts:55:18)
    at parseFile (https://deno.land/std@0.147.0/dotenv/mod.ts:138:12)
    at configSync (https://deno.land/std@0.147.0/dotenv/mod.ts:79:16)
    at https://deno.land/std@0.147.0/dotenv/load.ts:11:3
```

**Expected behavior**

When changing version to 0.146:

```typescript
import "https://deno.land/std@0.146.0/dotenv/load.ts";
console.log(Deno.env.get("A"));
```

Output is as shown:

```
$ deno run -A test.ts 
foo
```

**Environment**

- OS: Mac OS 12.6
- deno version: 1.26.2
- std version: 0.147 to 0.160

lino-levan:
This was fixed in #2819. Should we close this issue now, or wait until a new std version is published?

cjihrig:
Yes, this can be closed.

acrodrig:
Thanks.

PR #2819, opened 2022-10-28
fix(dotenv): Empty string crashing parsing

lino-levan:
Empty strings in .env files crashed the parsing of them. Looks like it was just a falsey value that was not tested for.

Simple reproc (.env):
```
A=""
```

Related: #2798

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -55,9 +55,9 @@ export function parse(
       keysForExpandCheck.push(key);
     }
 
-    env[key] = notInterpolated
+    env[key] = typeof notInterpolated === "string"
       ? notInterpolated
-      : interpolated || interpolated === ""
+      : typeof interpolated === "string"
       ? expandCharacters(interpolated)
       : unquoted.trim();
   }
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -57,7 +57,7 @@ export function parse(
 
     env[key] = notInterpolated
       ? notInterpolated
-      : interpolated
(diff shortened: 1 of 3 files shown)


dotenv/mod.ts
```diff
@@ -57,7 +57,7 @@ export function parse(
 
     env[key] = notInterpolated
       ? notInterpolated
-      : interpolated
+      : interpolated || interpolated === ""
```
cjihrig, line 60: Could this become `typeof interpolated  === "string"`?

dotenv/testdata/.env.test
```diff
@@ -5,6 +5,7 @@ AFTER_EMPTY=empty
 EMPTY_VALUE=
 QUOTED_SINGLE='single quoted'
 QUOTED_DOUBLE="double quoted"
+EMPTY_QUOTED=""
```
cjihrig, line 8: Should we include a test for empty single quotes as well?

dotenv/mod.ts
lino-levan: That's a lot more explicit, let's do that.

dotenv/testdata/.env.test
lino-levan: I don't think it's strictly necessary but having one more test to prevent regressions will not hurt.

lino-levan: It turns out it was strictly necessary, there was a bug with single quote parsing. Nice catch!

cjihrig (approved):
LGTM, thanks!

Issue #2772, opened 2022-10-11
suggestion: detect relative imports within code examples


iuioiua:
#2762 replaced relative imports within code examples with absolute imports. As @ayame113 pointed out, this allows examples within the documentation to be copied and pasted.

Having a tool that can detect and warn the user when a relative import within a code example would prevent this undesired from happening again. I think something similar to #2613 by @timreichen would do the trick.

ayame113:
I think it is good. For reference, the regex I used to find the relative imports from within the docs are `\*.*from.*"\.` and `\*.*import.*"\.`.

iuioiua:
This has been implemented in #2820.

PR #2820, opened 2022-10-28
feat(tools): add import path check in docs

schwarzkopfb:
Add tool to check import paths in JSDoc comments and readme files across std to help standardize them as suggested in #2772. Enforcing that all imports in examples are from `https://deno.land/std@$STD_VERSION/*`

Screenshot 2022-10-28 at 3 55 10


iuioiua:
Nice! I'm not a maintainer, but a few concerns:
1. Is the CI step you were supposed to add be `deno task lint:doc-imports` instead of `deno task lint:deprecations`?
2. Could `walk()` from `std/fs` replace your `walk()`?
3. IMO, regex is fairly complex. Is there a simpler alternative?

schwarzkopfb:
@iuioiua Thanks for the great suggestions!

1) that was a dumb mistake, thanks for catching it
2) you're absolutely right, std's built-in `walk()` is perfectly fine and makes the code much cleaner
3) I'll spend some time to think/experiment on how could that be simplified, probably there is room to cut off some parts of that

_tools/check_doc_imports.ts
```diff
@@ -0,0 +1,90 @@
...
+  const filePath = x.path;
+  const isExcluded = EXCLUDED_PATHS
+    .map((x) => filePath.includes(x))
+    .some((x) => x);
+
+  if (
+    x.isDirectory || isExcluded
+  ) {
+    continue;
+  }
```
iuioiua, line 73: I think this whole section could be trimmed by using:
```ts
for await (
  const { path } of walk(ROOT, {
    exts: EXTENSIONS,
    includeDirs: false,
    skip: [
      /(node|.github).*/g,
    ],
  })
)
```

_tools/check_doc_imports.ts
```diff
@@ -0,0 +1,90 @@
...
+import * as colors from "../fmt/colors.ts";
+import { walk } from "../fs/walk.ts";
+
+const EXTENSIONS = [".mjs", ".js", ".ts", ".md"];
+const EXCLUDED_PATHS = [
+  ".git",
+  "node",
+];
+
+const ROOT = new URL("../", import.meta.url).pathname.slice(0, -1);
```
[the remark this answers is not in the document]
schwarzkopfb: in line #49 `ROOT` is also used for making the output a bit more readable by displaying file locations relative to the std root.

Without that it looks like something like this: 
Screenshot 2022-10-28 at 13 59 21

I prefer the current, shorter form but I'm convincible.

_tools/check_doc_imports.ts
schwarzkopfb: I was thinking about that, but I'd like to keep it easily configurable by `EXCLUDE_PATHS` for further extension in the future. One way is to dynamically generate the `skip` regexp based on the paths array, but I'm a bit concerned about that solution.

schwarzkopfb:
@iuioiua I came up with some changes based on your comments. WDYT?

cjihrig:
This looks like something that will be nice to have. My only concerns are:

- The complexity of the regular expressions if we need to do any maintenance on them. I'm not sure how much we can do about that though.
- Since there is nothing to report in the codebase right now, I guess we just have to take your word for it that it works based on your screenshots 😄

timreichen:
> * The complexity of the regular expressions if we need to do any maintenance on them. I'm not sure how much we can do about that though.

I think the solution is to use a proper ast tree walker. That would also be great to establish for future tools.

The [typescript](https://esm.sh/typescript) module works for creating and walking ast trees, but there is also [deno_ast](https://github.com/denoland/deno_ast), which is a rust mod but could (and imo should) be made runnable in deno scripts, like [deno_doc](https://github.com/denoland/deno_doc) is.

So the task could be done like so:
`deno_doc: get jsdoc comments` -> `deno_ast: walk code from comments, find and verify imports, log warnings`

iuioiua:
> @iuioiua I came up with some changes based on your comments. WDYT?

Looking better! However, I'm still wondering whether `countChecked` is needed...

schwarzkopfb:
@timreichen I implemented the AST-based approach based on the [typescript](https://esm.sh/typescript) module. I think it's good for a first iteration until denoland/deno_ast#120 gets resolved. Later we can upgrade to the Rust-based, more performant AST generation subsystem.

@cjihrig I added some tests so we have a better proof than just my words for that it's working. 😀 On windows it's failing for some weird, file path related issue. I'll investigate that soon, but until then I'd appreciate any feedback on the implementation.

@iuioiua The counter is only used to display a message on exit which informs the user about that something happened, but no error raised in mora handy way than just exiting with 0 or a non-zero exit code. Just like `deno fmt` or `deno lint` does. Also can be useful when someone adjusts the `EXCLUDE_PATHS` and can easily see a stat on how many files included in the scan.

iuioiua:
> Also can be useful when someone adjusts the EXCLUDE_PATHS and can easily see a stat on how many files included in the scan.

In this case, seeing a change in the number of files justifies the existence of the variable, IMO. Well explained.

iuioiua:
Would it perhaps be better to use `Deno.spawn` over `Deno.run`? That's where things are moving towards.

schwarzkopfb:
> Would it perhaps be better to use `Deno.spawn` over `Deno.run`? That's where things are moving towards.

Great suggestion, makes it a bit less messy, applied. Thanks!

schwarzkopfb:
@timreichen FYI: I also made [an experiment](https://schwarzkopfb.codes/deno_std/tree/tools_doc_import_check__with_deno_doc) which is using `deno_doc` for jsDoc parsing that produced surprising results. It is _orders of magnitude_ slower even with a smaller set of files to work with. And it breaks if I run the scan on the whole std repo. So for now I stick with the regexp for identifying comments and extracting code blocks from it and then using the AST generated by the TypeScript compiler to check import statements.

Screenshot 2022-10-30 at 11 10 12

Screenshot 2022-10-30 at 11 10 28

cjihrig (approved):

PR #2824, opened 2022-10-29
fix(flags): set boolean aliases to false by default

### LINKED ISSUE #2778: bug(flags): all boolean arguments included in alias are not set to false by default
**Describe the bug**
When setting an alias and if one of the alias are included in the boolean argument, all arguments inside the alias should have their value to false.

image

**Steps to Reproduce**

```ts
import { parse } from 'https://deno.land/std@0.159.0/flags/mod.ts';

const options = parse(Deno.args, {
  boolean: ['h'],
  alias: {
    h: "help"
  },
});
console.log(options);
```

**Expected behavior**

`options.h` and `options.help` should be false.

Also the type-checking is trying to guarantee the values are boolean:

image

**Environment**

- OS: Fedora 36
- deno version: 1.26.1
- std version: 0.159.0

c4spar:
close #2778

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -367,6 +367,7 @@ export function parse<
     unknown = (i: string): unknown => i,
   }: ParseOptions<B, S, C, N, D, A, DD> = {},
 ): Args<V, DD> {
+  const aliases: Record<string, string[]> = {};
   const flags: Flags = {
     bools: {},
     strings: {},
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks for the quick fix!

sant123:
Thank you so much!!

PR #2826, opened 2022-10-31
chore(semver): rename inc and diff

timreichen:
- rename abbreviations: `inc` to `increment` and `diff` to `difference`.
This is more in alignment to the js naming convention.


cjihrig:
This seems like more of a breaking change than a chore. If we are going to make this change, I think we should have the old and new names for a while and deprecate the old names.

kt3k (approved):
LGTM

cjihrig (approved):

PR #2831, opened 2022-11-02
docs(encoding): remove `await`

liruifengv:


littledivy (approved):

kt3k (approved):
LGTM.

Note: We made `parse` sync at v0.153.0 https://github.com/denoland/deno_std/commit/33afdfe0adf56fcb34ecbabc29a1223fe6408c3b

kt3k:
```
   89 const hostname = os.hostname();
   90 is.string(hostname);
   91 assert.ok(hostname.length > 0);
```
The assertion at line 91 (in `node/_tools/test/parallel/test-os.js`) seems failing on windows. Might have been affected by https://github.com/denoland/deno/pull/16441?

PR #398, opened 2019-05-13
walk() should not use deprecated FileInfo.path

ry:
walk() made extensive use of `FileInfo.path` which will be removed in future versions. 

cc @kt3k 

Note: this is necessary to bump CI to v0.5 https://github.com/denoland/deno_std/pull/393

Necessary for https://github.com/denoland/deno/pull/2313 too

PR #2836, opened 2022-11-03
Remove misleading docs from fs.walk

matklad:
It doesn't actually visit files in any particular order

### DIFF
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -53,9 +53,7 @@ export interface WalkOptions {
 export type { WalkEntry };
 
 /** Walks the file tree rooted at root, yielding each file or directory in the
- * tree filtered according to the given options. The files are walked in lexical
- * order, which makes the output deterministic but means that for very large
- * directories walk() can be inefficient.
+ * tree filtered according to the given options.
  *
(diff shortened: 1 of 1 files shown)


kt3k:
That part of the doc was added in https://github.com/denoland/deno_std/pull/398/files, and which seems to have been just copied from the doc of Walk in `path/filepath` in Go's std lib. https://pkg.go.dev/path/filepath#Walk

`Deno.readDir` uses `std::fs::read_dir` always (the first commit https://github.com/denoland/deno/pull/887/files), which doesn't guarantee the order of items. So `walk` has never guaranteed the order of the items.

It seems we never intended to guarantee the order of walk results. The removal looks reasonable to me.

kt3k (approved):
LGTM

matklad:
Yeah, this all makes sense. 

As a user, I would appreciate the guaranteed order, on theoretical grounds of not adding undue non-determinism to programs, and on practical grounds of this biting me in the back several times in Rust )

`Deno.readDir` is tough though: it wants to be both nice to use (sorted) and a thin abstraction over os (non sorting).

PR #2840, opened 2022-11-05
docs(encoding): remove `ColumnDetails["fn"]`

proudust:
`ColumnDetails["fn"]` was removed in #2611, but it remains in the docs. Therefore, it has been removed.


kt3k (approved):
LGTM

PR #2844, opened 2022-11-06
refactor(encoding): move `varint/_wasm_varint/` to `varint/_wasm/`

iuioiua:
This PR is more organisational and doesn't change any functionality. It was done simply because it doesn't make much sense to have `/_wasm_varint/` within `/encoding/varint/`.

Some old naming and documentation inconsistencies have also been fixed.


kt3k (approved):
This change makes sense to me. LGTM

PR #2845, opened 2022-11-06
refactor(crypto): move `crypto/_wasm_crypto/` to `crypto/_wasm/`

iuioiua:
This PR is more organisational and doesn't change any functionality. It was done simply because it doesn't make much sense to have `/_wasm_crypto/` within `/crypto/`.

Some old naming and documentation inconsistencies have also been fixed.


kt3k (approved):
LGTM. Thanks!

PR #2846, opened 2022-11-06
remove(_util): `assert`

iuioiua:
`_util/assert.ts` was added in denoland/deno#5838 with the following reasoning:
> This PR was ultimately focused on two intertwined outcomes. The first was reducing dependency on the Deno global (as seen in std/testing/asserts.ts) to improve browser compatibility. The second was on removing dependencies on std/testing/asserts.ts from non-test code for better separation of concerns. https://github.com/denoland/deno/pull/5847 resolves the first by making std/testing/asserts.ts browser compatible. What's left in this re-worked PR is the separation of concerns.

It's now been almost three years, and I think there are some valid reasons to get rid of it.
1. `/testing/asserts.ts` is browser compatible and doesn't use anything from the Deno namespace, thanks to denoland/deno#5847.
2. It has barely been used, being imported only 24 times across the codebase. On the other hand, `/testing/asserts.ts` has been imported just under 300 times across the codebase in non-test code. AFAIK, this has gone unnoticed because it's not really a problem anyway. Put simply, it was intended to be used for non-teest but hasn't.
3. `assert` is only one of many assertion functions. Why should an exception, for the above-quoted reasons, be made for just one assertion function and not the rest?
4. While known from the start, this is duplicated code. Furthermore, its existence is unclear unless one does some digging.

This PR removes `/_util/assert.ts` and replaces the relative few times it's used for `assert` in `/testing/asserts.ts`. My only concern is that `deno task test:browser` doesn't seem to check `/testing/asserts.ts`.


kt3k:
> On the other hand, /testing/asserts.ts has been imported just under 300 times across the codebase in non-test code.

I doubt `/testing/asserts.ts` is used such many places except testing/examples. In my view, non-test code should only use `_util/assert.ts` because it has less dependencies.

iuioiua:
> I doubt `/testing/asserts.ts` is used such many places except testing/examples.

This can be confirmed in VSCode:
* Search `/testing/asserts.ts`
* Include `*.ts, *.js`
* Exclude `testing`

At the time of writing this, I see 331 results.

> In my view, non-test code should only use _util/assert.ts because it has less dependencies.

In one way, I agree with this reasoning. However, this hasn't been done in the codebase. Also, why should this rule only apply to one of the many assertions?

iuioiua:
Perhaps, there should be a `testing/asserts` or `asserts` folder with one file per assertion... I'll close this PR and open an issue for further discussion.

kt3k:
> > I doubt `/testing/asserts.ts` is used such many places except testing/examples.
> 
> This can be confirmed in VSCode:
> 
> * Search `/testing/asserts.ts`
> * Include `*.ts, *.js`
> * Exclude `testing`
> 
> At the time of writing this, I see 331 results.



Use this search instead:

- Search: `/testing/asserts.ts`
- Include: `*.ts,*.js`
- Exclude: `testing,testdata,*test.ts,*test.js,*bench.ts,test/util.js,node/assert.ts,_test_utils.ts`

It shows 58 results, and most items are in code examples in JSDoc. I found below 10 actual usages, but I think these are bugs:
- 5 in `/node/internal_bindings`
- 1 in `/node/crypto`
- 1 in `/node/_utils.ts`
- 1 in `/media_types/mod.ts`
- 1 in `/io/files.ts`
- 1 in `/crypto/timing_safe_equal.ts`

iuioiua:
Oh, yes. Sloppy analysis on my part.

PR #2847, opened 2022-11-06
remove(_util): `deepAssign`

iuioiua:
This PR removes `_util/deepAssign.ts`, where it's used only in `encoding/_toml/parser.ts`, in favour of `collections/deep_merge.ts`.


kt3k (approved):
LGTM. Nice!

PR #2852, opened 2022-11-07
remove(bytes): `equalsNaive`

iuioiua:
This PR:
1. Removes `equalsNaive`, which was a private function.
2. Renames `equals32Bit` to `equals` which can handle binary arrays of any length.
3. Removes the now unneeded benchmark.


kt3k:
These removals make sense to me.

These were only necessary for landing https://github.com/denoland/deno_std/pull/1349

kt3k (approved):
LGTM

0f-0b:
If I understand correctly, the purpose of `equalsNaive` is to avoid slowing down `equals` for small arrays. For this reason I don't think `equalsNaive` should be removed.

To discourage direct import of these private functions, their names should perhaps be prefixed with an underscore, as [described in the readme](https://deno.land/std@0.162.0#how-to-use).

kt3k:
Ah, right. `equalsNative` was used for shorter input..

PR #2853, opened 2022-11-08
chore(node): temporarily disable flash

kt3k:
This PR temporarily disables flash in `node/http` server as it doesn't perform streaming correctly https://github.com/denoland/deno/issues/16567

Because of the issue, static file server (using `express.static` for example) doesn't work. That partially causes https://github.com/denoland/deno/issues/16282

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -453,8 +453,12 @@ class ServerImpl extends EventEmitter {
 
   constructor(handler?: ServerHandler) {
     super();
+    // TODO(kt3k): Temporarily disables flash. Enable when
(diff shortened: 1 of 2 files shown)


bartlomieju (changes requested):
I'm -1 on disabling it entirely. This will break projects like Vite or anything else using WS. Let's dofucs on fixing the issue in Deno

kt3k:
> This will break projects like Vite

Ah, right. I forgot 'upgrade' only works with flash. Ok. Let's fix https://github.com/denoland/deno/issues/16567

Issue #2849, opened 2022-11-07
suggestion: remove `_util/assert.ts`


iuioiua:
Please read #2846 before reading this. I suggest moving `std/testing/asserts.ts` to either an `std/testing/asserts` or `std/asserts` folder with an assertion-per-file. The folder would look more like `std/fs`.

Indeed, there is a large number of assertions, meaning there will be a large number of files. However, I think this is worth it to gain leaner assertion-related dependencies.

The files will have to be moved out, and then `std/testing/asserts.ts` functions deprecated. It'd be a bit of work, but I'd be happy to do it.

bartlomieju:
Strong -1 from me; this module is probably one of the most popular modules from std, it would be a huge disruptive change. I don't think it's worth it, since very often you need more than one assertion. Having to import them from separate files will be cumbersome for no obvious gain.

lino-levan:
Personally, I agree with @bartlomieju on this one. I think that if this was better for DX even if it was a disruptive change it could be seriously considered (deno itself was a disruptive change, less so now that we have pretty good compat). I struggle to see how splitting up assertions into different files would give *any* advantage to anything, especially DX?

>However, I think this is worth it to gain leaner assertion-related dependencies.

I'd be curious to hear you elaborate on your point about this. Maybe I'm misunderstanding something and this is more valuable than I currently think.

iuioiua:
Non-test code is supposed to use `assert` from `_util/assert.ts` to:
1. Avoid importing other unneeded code from `testing/asserts.ts`.
2. Avoid browser compatibility issues from `testing/asserts.ts`.
However, now, `testing/asserts.ts` is browser compatible. I suggest breaking the code up to cater to number 1.

Again, what's gained is that importing unnecessary assertions can be avoided. There can also exist a `mod.ts` so that multiple assertions can be imported if needed. This would remove the need for `_util/assert.ts`.

iuioiua:
> Non-test code is supposed to use `assert` from `_util/assert.ts`

My question is: why is `assert` the only assertion that this rule applies to? Other assertions are used in non-test code.

Alternatively, these other assertions could move into `_util/asserts.ts`. It would be less disruptive.

<img width="406" alt="image" src="https://user-images.githubusercontent.com/29347852/200405699-c9e81b75-c5f9-4ccd-999c-4a4144e7e183.png">

iuioiua:
Again, check out #2846. These are my points exactly. Perhaps, I closed the PR prematurely.

[...]

PR #2857, opened 2022-11-08
refactor: use `_util/asserts.ts` for non-test code

iuioiua:
This PR:
1. More appropriately renames `_util/assert.ts` to `_util/asserts.ts`.
2. Adds missing `unreachable` assertion to `_util/asserts.ts`, along with corresponding test in `_util/asserts_test.ts`.
3. Corrects non-test code within the codebase that previously used `testing/asserts.ts` to use `_util/asserts.ts`.
5. Adds a small explainer to `_util/assert.ts` and the "Contributing" section of the repo readme.

This is half of what was discussed in #2849. I'll create a tool to enforce the rule after this PR lands.


node/_crypto/crypto_browserify/asn1.js/base/node.js
```diff
@@ -3,7 +3,7 @@
 
 import { Reporter } from "./reporter.js";
 import { DecoderBuffer, EncoderBuffer } from "./buffer.js";
-import { assert, assertEquals } from "../../../../../testing/asserts.ts";
+import { assert, assertEquals } from "../../../../../_util/asserts.ts";
```
kt3k, line 6: Let's drop this `assertEquals` usage (it only compares numbers. It should be easily replaced by `assert`). `assertEquals` has fancy formatting feature of value diffs and that's not necessary for internal assertions.

_util/asserts.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+/**
+ * These assertions exist to create a separation of concerns between internal and testing assertions.
+ * All non-testing code within the codebase must use these assertions.
+ */
+import * as asserts from "../testing/asserts.ts";
```
kt3k, line 8: Let's drop this dependency. This introduces many unnecessary dependencies in several places. Internal assertions don't need fancy formatting or colors.

node/_utils.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 import { deferred } from "../async/mod.ts";
-import { assert, assertStringIncludes, fail } from "../testing/asserts.ts";
+import { assert, assertStringIncludes, fail } from "../_util/asserts.ts";
```
kt3k, line 3: `fail` looks only used by `mustCall` and it doesn't seem used by anywhere. I think we should remove `fail` and `mustCall`

kt3k: In case of `assertStringIncludes`, it's only used from `assertCallbackErrorUncaught` and that seems like an testing utility (only used from tests).

So we don't need to create `assertStringIncludes` in `_util/asserts.ts`, but rather need to move `assertCallbackErrorUncaught` to somewhere like `node/_test_util.ts`.

kt3k:
I think what we actually need in `_util/asserts.ts` are only `assert` and `unreachable` and those don't need to depend on `testing/asserts.ts`.

iuioiua:
Thanks for the good suggestions! I've made the changes, except for moving `assertCallbackErrorUncaught` into `node/_test_util.ts`, as I don't believe a whole file should exist for a single function. Perhaps, it can be revisited in another PR.

I've redone my initial comment on this PR to reflect the new changes.

README.md
```diff
@@ -109,3 +109,10 @@ accepted.
 _For maintainers_:
 
 To release a new version a tag in the form of `x.y.z` should be added.
+
+### Internal Assertions
+
+All internal non-test code, that is files that do not have `test` or `bench` in
+the name, must use the assertion functions within `_utils/asserts.ts` and not
+`testing/asserts.ts`. This is to create a separation of concerns between
+internal and testing assertions.
```
kt3k, line 118: Nice. Thanks!

kt3k (approved):
LGTM

PR #2858, opened 2022-11-09
perf(node/string_decoder): use native decoder for GenericDecoder

littledivy:
51x improvement in `bigUint8`
4x improvement in `smallUint8`

micro-benchmark taken from https://twitter.com/yagiznizipli/status/1590031518734651392

Before:
```
cpu: Apple M1
runtime: deno 1.27.1 (aarch64-apple-darwin)

benchmark       time (avg)             (min … max)       p75       p99      p995
-------------------------------------------------- -----------------------------
smallUint8    1.15 µs/iter      (1.15 µs … 1.2 µs)   1.16 µs    1.2 µs    1.2 µs
bigUint8    111.01 µs/iter (106.92 µs … 219.67 µs) 111.29 µs 154.25 µs 160.63 µs
```

After:
```
cpu: Apple M1
runtime: deno 1.27.1 (aarch64-apple-darwin)

benchmark       time (avg)             (min … max)       p75       p99      p995
-------------------------------------------------- -----------------------------
smallUint8  299.26 ns/iter (295.05 ns … 331.38 ns) 299.87 ns    320 ns 331.38 ns
bigUint8      2.17 µs/iter     (2.12 µs … 2.33 µs)   2.19 µs   2.33 µs   2.33 µs
```

Node 18.8
```
cpu: Apple M1
runtime: node v18.8.0 (arm64-darwin)

benchmark       time (avg)             (min … max)       p75       p99      p995
-------------------------------------------------- -----------------------------
smallUint8  354.58 ns/iter (348.33 ns … 417.61 ns) 355.44 ns 387.07 ns 417.61 ns
bigUint8       5.3 µs/iter     (5.27 µs … 5.46 µs)   5.29 µs   5.46 µs   5.46 µs
```


node/string_decoder.ts
```diff
@@ -227,6 +228,16 @@ function simpleWrite(
   return buf.toString(this.encoding);
 }
 
+function simpleUtf8Write(
```
kt3k, line 231: Does it make sense to replace `utf8Write` with this implementation?

littledivy: no, that would require a stateful streaming decoder op. We can do that if we ever move `std/node` to the runtime :)

kt3k: ok, this only supports simple case (doesn't handle incomplete char bytes)

kt3k (approved):
LGTM

node/string_decoder.ts
kt3k: > no, that would require a stateful streaming decoder op. We can do that if we ever move std/node to the runtime :)

Maybe we only need a little modified version of `core.decode`, which accepts 2 Uint8Arrays and returns a tuple of (string, remaining bytes)?

node/string_decoder.ts
```diff
@@ -227,6 +228,16 @@ function simpleWrite(
...
+  this: StringDecoderBase,
+  buf: Buffer | string,
+): string {
+  if (typeof buf === "string") {
+    return buf;
+  }
+  return core.decode(buf);
```
lucacasonato, line 238: This is not a public API….

node/_core.ts
```diff
@@ -22,6 +22,9 @@ if (Deno?.core) {
     encode(chunk: string): Uint8Array {
       return new TextEncoder().encode(chunk);
     },
+    decode(chunk: Uint8Array): string {
+      return new TextDecoder().decode(chunk);
```
anonrig, line 26: I wonder how will the performance be impacted if the initialized text decoder will only be initialized once for the subsequent calls.

littledivy: This is a fallback implementation, but to answer your question: yes it will be slightly faster to inititalize once but only noticable for very small inputs in micro benchmarks.

PR #2861, opened 2022-11-09
feat(fmt): add `prettyDuration`

retraigo:
A smaller version of [duration.js](https://github.com/retraigo/duration.js) that only performs formatting milliseconds into strings.


fmt/README.md
```diff
@@ -17,47 +17,47 @@ below.
 
 This is very much a work-in-progress. I'm actively soliciting feedback.
 
-- What useful features are available in other languages apart from Golang and C?
+-   What useful features are available in other languages apart from Golang and C?
```
cjihrig, line 20: Please undo all of the style only changes like this. It make the diff bigger and PRs more difficult to review.

fmt/numbers.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+// A module to format numbers.
+
+/** Add zeros to the beginning of a number till it reaches a certain digit count. */
```
cjihrig, line 4:
```suggestion
/** Add zeros to the beginning of a number until it reaches a certain digit count. */
```

fmt/numbers.ts
```diff
@@ -0,0 +1,8 @@
...
+export function addZero(num: number, digits = 3): string {
```
cjihrig, line 5: Would something simpler like this work:

```js
number.toString().padStart(digits, '0')
```

fmt/numbers_test.ts
```diff
@@ -0,0 +1,19 @@
...
+Deno.test({
+  name: "random number",
+  fn() {
+    const num = Math.round(Math.random() * 9787923);
+    assertEquals(num.toString(), addZero(num, num.toString().length));
+  },
+});
+
+Deno.test({
+  name: "bond",
```
cjihrig, line 15:
```suggestion
  name: "pads the start of a number",
```

fmt/duration.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { addZero } from "./numbers.ts";
+
+interface DurationObj {
+  d: number;
+  h: number;
+  m: number;
+  s: number;
+  ms: number;
+  µs: number;
```
cjihrig, line 12: Can we use `us` instead.

fmt/README.md
retraigo: done

fmt/numbers.ts
retraigo: removed the entire file because it was pointless

fmt/numbers.ts
retraigo: Yeah forgot padStart existed.

fmt/README.md
```diff
@@ -268,6 +268,26 @@ prettyBytes(1337, { locale: "de" });
 //=> '1,34 kB'
 ```
 
+# Pretty Duration
+
+Get formatted time duration from milliseconds.
+
+```ts
+import { prettyBytes } from "https://deno.land/std@$STD_VERSION/fmt/duration.ts";
```
kt3k, line 276:
```suggestion
import { prettyDuration } from "https://deno.land/std@$STD_VERSION/fmt/duration.ts";
```

retraigo: My bad

cjihrig (approved):

kt3k (approved):
LGTM

Note: There's a very similar ES feature proposal https://github.com/tc39/proposal-intl-duration-format but it's unclear when it's available in V8. So I think it makes sense to include this to std until that feature is implemented in V8

retraigo:
Yeah I'm looking forward to Temporals becoming part of the spec.

timreichen:
As a succession PR with some enhancements: https://github.com/denoland/deno_std/pull/2871

PR #2863, opened 2022-11-09
fix(node/http): fix non-string buffer response

### LINKED ISSUE #2862: node/http: non-string buffer response is broken when `--unstable` is specified
The following script can't response correctly when `--unstable` is specified (= when Deno.serve is used)

```js
import zlib from "http://deno.land/std@0.163.0/node/zlib.ts";
import { Buffer } from "http://deno.land/std@0.163.0/node/buffer.ts";
import http from "http://deno.land/std@0.163.0/node/http.ts";

const server = http.createServer((req, res) => {
  res.setHeader("content-encoding", "gzip");
  res.end(zlib.gzipSync(Buffer.from("a".repeat(100), "utf8")));
});
server.listen(3000);
```

kt3k:
This PR fixes binary (non-text) response of flash-backed node/http server.

fixes #2862

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -337,7 +337,7 @@ export class ServerResponse extends NodeWritable {
     const body = singleChunk ?? (final ? null : this.#readable);
     if (this.#isFlashRequest) {
       this.#resolve!(
-        new Response(Buffer.isBuffer(body) ? body.toString() : body, {
+        new Response(body, {
           headers: this.#headers,
           status: this.statusCode,
           statusText: this.statusMessage,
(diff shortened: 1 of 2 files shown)


node/http.ts
```diff
@@ -337,7 +337,7 @@ export class ServerResponse extends NodeWritable {
     const body = singleChunk ?? (final ? null : this.#readable);
     if (this.#isFlashRequest) {
       this.#resolve!(
-        new Response(Buffer.isBuffer(body) ? body.toString() : body, {
```
kt3k, line 340: `body.toString()` is unnecessary (because a Buffer is a Uint8Array) and also problematic when body contains non-text content.

cjihrig (approved):

PR #2867, opened 2022-11-09
chore(node): upgrade `denoflate` to 2.0.2 for setup script

iuioiua:
`node/_tools/TODO.md` was also updated as a side effect.


kt3k (approved):
LGTM

kt3k:
BTW should we number the remaining test list? Because all lines are numbered, the change in the middle of the list causes the large diff like this.

iuioiua:
> BTW should we number the remaining test list? Because all lines are numbered, the change in the middle of the list causes the large diff like this.

Yes, good point. Will do.

kt3k:
Thanks!

PR #2871, opened 2022-11-10
BREAKING(fmt/duration): rename prettyDuration to format

timreichen:
<del>- adds `parse` function to reverse formatting</del>

The following name changes are inspired by https://github.com/tc39/proposal-intl-duration-format 
- renames `prettyDuration()` to `format()`
- renames `formatType` to `style`
- renames `time`  formatType to `digital`
- renames `short`  formatType to `narrow`


fmt/README.md
```diff
@@ -270,22 +270,40 @@ prettyBytes(1337, { locale: "de" });
...
 // "1m 39s 674ms"
-prettyDuration(99674, { ignoreZero: true });
+format(99674, { ignoreZero: true });
 
 // "1 minutes, 39 seconds, 674 milliseconds"
-prettyDuration(99674, { formatType: "full", ignoreZero: true });
+format(99674, { style: "full", ignoreZero: true });
+```
+
+parse time duration to milliseconds.
```
cjihrig, line 291:
```suggestion
Parse time duration to milliseconds.
```

fmt/README.md
```diff
@@ -270,22 +270,40 @@ prettyBytes(1337, { locale: "de" });
 
 # Pretty Duration
 
-Get formatted time duration from milliseconds.
+format milliseconds to time duration.
```
cjihrig, line 273:
```suggestion
Format milliseconds to time duration.
```

fmt/duration.ts
```diff
@@ -25,7 +25,7 @@ const keyList: Record<keyof DurationObj, string> = {
 };
 
 /** Parse milleseconds into a duration. */
-function parseDuration(ms: number): DurationObj {
+function millisecondsToObject(ms: number): DurationObject {
```
cjihrig, line 28:
```suggestion
function millisecondsToDurationObject(ms: number): DurationObject {
```

fmt/duration.ts
```diff
@@ -53,33 +53,45 @@ function durationArray(
   ];
 }
 
+function objectToMilliseconds(object: DurationObject) {
```
cjihrig, line 56:
```suggestion
function durationObjectToMilliseconds(object: DurationObject) {
```

fmt/duration.ts
```diff
@@ -121,7 +133,75 @@ export function prettyDuration(
...
+    ms: 0,
+    us: 0,
+    ns: 0,
+  };
+  let match;
+  const valueRegex = /(?<value>\d+)\s*(?<name>[\µa-z]+)/g;
+  while ((match = valueRegex.exec(value)) !== null) {
+    const groups = match.groups!;
+    const name = groups.name;
+    const value = parseInt(groups.value);
```
cjihrig, line 167:
```suggestion
    const value = Number.parseInt(groups.value, 10);
```

fmt/duration_test.ts
```diff
@@ -1,64 +1,113 @@
...
 
 Deno.test({
-  name: "short duration ignore zero",
+  name: "format narrow duration ignore zero",
   fn() {
-    assertEquals(prettyDuration(99674, { ignoreZero: true }), "1m 39s 674ms");
+    assertEquals(format(99674, { ignoreZero: true }), "1m 39s 674ms");
+  },
+});
+
```
cjihrig, line 65: Can you add some tests to verify that `parse(format(value)) === value`.

timreichen: done

kt3k:
I'm not in favor of adding this feature to std. What this `parse` parses are random duration format, and doesn't look useful to me.

retraigo:
I didn't want to add `parse` in my earlier PR because I didn't feel like it needed to be part of a "standard library", considering people represent durations in a lot of ways.

There is a more complete version of `parse` in [duration.js][(https://github.com/retraigo/duration.js](https://github.com/retraigo/duration.js/blob/main/src/duration.ts#L560)) (though the module itself is quite big as a result) if you would like to refer.

Apart from renaming `prettyDuration` to `format` (had it as `prettyDuration` to maintain consistency with `prettyBytes`, the other renamings seem helpful.

timreichen:
> I'm not in favor of adding this feature to std. What this `parse` parses are random duration format, and doesn't look useful to me.

I am not quite following you, what do you mean exactly by random? It parses the possible outputted formats of `prettyDuration`, or did I miss something?
I think it is useful to provide a solution to reverse a formatting process, similar to what we do in std/datetime. There is no harm in providing it.

> I didn't want to add `parse` in my earlier PR because I didn't feel like it needed to be part of a "standard library", considering people represent durations in a lot of ways.
> 
> There is a more complete version of `parse` in [duration.js](https://github.com/retraigo/duration.js/blob/main/src/duration.ts#L560) (though the module itself is quite big as a result) if you would like to refer.

That is true, but `parse` supports the current formatting options.  Though it could expanded, I don't think it needs to provide more than what the counterpart function is capable of formatting atm.

> Apart from renaming `prettyDuration` to `format` (had it as `prettyDuration` to maintain consistency with `prettyBytes`, the other renamings seem helpful.

I think it would work better the other way around. Web APIs have some consistent naming schemes, `format` is one of them and I think we should adapt them, so `prettyBytes` would be better of called `format` as well imo.

kt3k:
> > I'm not in favor of adding this feature to std. What this `parse` parses are random duration format, and doesn't look useful to me.
> 
> I am not quite following you, what do you mean exactly by random? It parses the possible outputted formats of `prettyDuration`, or did I miss something? I think it is useful to provide a solution to reverse a formatting process, similar to what we do in std/datetime. There is no harm in providing it.

It's random because it doesn't have any formal specification. What is correct format or what is bug is not well-defined here. Users might report bugs about unspecified/untested format and we need to address those. That'll cost maintenance effort.

timreichen:
@kt3k Ok, I removed the parse function for now.

cjihrig (approved):

retraigo:
I think `prettyBytes` should be updated to match this naming convention too.

timreichen:
> I think `prettyBytes` should be updated to match this naming convention too. Maybe in another PR...

I'll open a separate PR 👍

cjihrig:
@kt3k are you still opposed to this after the latest changes?

kt3k (approved):
I'm now in favor. `format` sounds more 'standard' than `prettyDuration`. I think this change is worth a breaking change.

LGTM

PR #2873, opened 2022-11-10
refactor: move circular dependencies check to `node/_tools`

iuioiua:
This PR moves `_tools/check_circular_deps.ts` to the more appropriate `node/_tools/check_circular_deps.ts`, as it only checks Node code.

Would it also make more sense to rename the task from `lint:circular-deps` to `node:circular-deps`?


cjihrig:
> Would it also make more sense to rename the task from lint:circular-deps to node:circular-deps?

Maybe `node:lint-circular-deps`?

iuioiua:
> > Would it also make more sense to rename the task from lint:circular-deps to node:circular-deps?
> 
> 
> 
> Maybe `node:lint-circular-deps`?

I like that!

kt3k (approved):
LGTM. Nice!

PR #2876, opened 2022-11-11
feat(tools): check assertions

### LINKED ISSUE #2849: suggestion: remove `_util/assert.ts`

iuioiua:
This PR adds a tool that checks whether `testing/asserts.ts` is not incorrectly imported instead of `_util/asserts.ts`.

The check has already picked up some errors. Defining non-test code as code that a user can pull, these assertion imports need to be fixed to use `_util/asserts.ts`. We may need to add `assertRejects` to the internal assertions.

Closes #2849.


kt3k:
> The check has already picked up some errors. Defining non-test code as code that a user can pull, these assertion imports need to be fixed to use _util/asserts.ts.

```
- file:///home/runner/work/deno_std/deno_std/encoding/front_matter/_test_utils.ts
- file:///home/runner/work/deno_std/deno_std/testing/mock.ts
- file:///home/runner/work/deno_std/deno_std/testing/snapshot.ts
```

These look like false positives to me as they are testing utilities. They also should be excluded in some way

iuioiua:
I've reduced the `skip` parameter to include anything that includes `test`, `bench` or `node/`.

kt3k (approved):
Thanks for updating! LGTM

PR #2879, opened 2022-11-11
chore(node/_tools): update node setup tool

kt3k:
This PR updates the setup script of node compat test cases.

Summary of changes:
- call `list_remaining_tests.ts` from `setup.ts`
- don't call github API from `list_remaining_tests.ts`, instead use `node/_tools/versions/node-v$VERSION` directory structure for looking up remaining test cases.
- link to v18.12.0 tag instead of main.


iuioiua:
Nice! I was thinking my implementation was suboptimal when I initially wrote this. Good to see it can be done in a much cleaner manner.

cjihrig (approved):

Issue #2851, opened 2022-11-07
compat: static file serving using express.static hangs


kt3k:
Run the below script with some large (>100KB) file in `assets/`:

```js
import express from "npm:express";
const app = express();
app.use("/assets", express.static("assets"));
app.get('/', (req, res) => {
  res.send("hello");
});
app.listen(3000);
```

The request `http://localhost:3000/assets/<file>` often (not always) hangs.

If the file is larger (>1M), then the request also ends with `Malformed encoding found in chunked-encoding`.

kt3k:
This blocks denoland/deno#16282

kt3k:
Smaller reproduction code (`express.static` uses `serve-static` internally)

```ts
import http from "https://deno.land/std@0.162.0/node/http.ts";
import serveStatic from "npm:serve-static";
const serve = serveStatic("assets")
const server = http.createServer((req, res) => {
  serve(req, res);
});
server.listen(3000);
```

kt3k:
More reduced version:
```ts
import http from "https://deno.land/std@0.162.0/node/http.ts";
import send from "npm:send";
const server = http.createServer((req, res) => {
  send(req, req.url, { root: "assets" }).pipe(res);
});
server.listen(3000, () => console.log("listening 3000"));
```

kt3k:
The above is reduced to:
```js
import http from "https://deno.land/std@0.162.0/node/http.ts";
import fs from "https://deno.land/std@0.162.0/node/fs.ts";
const server = http.createServer((req, res) => {
  fs.createReadStream("assets" + req.url).pipe(res);
});
server.listen(3000, () => console.log("listening 3000"));
```

kt3k:
Looks like flash (Deno.serve) can't stream large file correctly.

This script hangs.
```js
Deno.serve((_) => new Response(Deno.openSync("data.txt").readable));
```

kt3k:
created an issue for flash https://github.com/denoland/deno/issues/16567

PR #2880, opened 2022-11-11
fix(node/fs): improve fs.read compatibility

kt3k:
Currently `fs.read` calls the callback synchronously. That partially caused https://github.com/denoland/deno_std/issues/2851. This PR makes `fs.read` call the callback asynchronously.

### DIFF
--- a/node/_fs/_fs_read.ts
+++ b/node/_fs/_fs_read.ts
@@ -115,8 +115,8 @@ export function read(
       let nread: number | null;
       if (typeof position === "number" && position >= 0) {
         const currentPosition = await Deno.seek(fd, 0, Deno.SeekMode.Current);
-        // The following seek/read/seek calls need to be sync
-        // because those need to happen atomically
+        // We use sync calls below to avoid being affected by others during
+        // these calls.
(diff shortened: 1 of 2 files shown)


node/_fs/_fs_read.ts
```diff
@@ -108,35 +108,26 @@ export function read(
...
-  }
-
-  return;
+  (async () => {
+    try {
+      let nread: number | null;
+      if (typeof position === "number" && position >= 0) {
+        const currentPosition = await Deno.seek(fd, 0, Deno.SeekMode.Current);
+        // The following seek/read/seek calls need to be sync
+        // because those need to happen atomically
```
cjihrig, line 119: Making these calls synchronous doesn't make any guarantees about atomicity... but the implementation was already using these APIs, so I guess no need to change it here.

cjihrig (approved):
LGTM since the CI failures appear to be network related.

node/_fs/_fs_read.ts
kt3k: Right, it was a wrong word. Updated

PR #2892, opened 2022-11-15
fix(node): issues with browser

roj1512:


kt3k:
How did you get affected by this?

roj1512:
I am `deno bundle`ing a project that is using `/node/fs.ts` and importing it from browser.

cjihrig (approved):

kt3k (approved):
LGTM

kt3k:
@roj1512 Do you know how `node/fs.ts` is depended in your dependency tree?

If the usage looks common/useful enough, maybe we can add some checks in CI to prevent this happening again.

PR #2893, opened 2022-11-15
fix: allow creating http.IncomingRequest without url being set

Skn0tt:
Congrats on launching the NPM compatibility, this is huge!
I'm trying to get an express app running on Netlify Edge Functions, and using https://github.com/dougmoscrop/serverless-http to do so. That library is [creating a `http.IncomingMessage`](https://github.com/dougmoscrop/serverless-http/blob/80bfb3e940057d694874a8b0bc12ad96d2abe7ab/lib/request.js#LL7) without `url` being specified, and sets it [in a second step](https://github.com/dougmoscrop/serverless-http/blob/80bfb3e940057d694874a8b0bc12ad96d2abe7ab/lib/request.js#LL29) (reproduced in the test).

That currently results in an uncaught exception with Deno's polyfill. This PR fixes that.

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -419,7 +419,7 @@ export class IncomingMessageForServer extends NodeReadable {
     });
     // TODO: consider more robust path extraction, e.g:
     // url: (new URL(request.url).pathname),
-    this.url = req.url.slice(req.url.indexOf("/", 8));
+    this.url = req.url?.slice(req.url.indexOf("/", 8));
     this.method = req.method;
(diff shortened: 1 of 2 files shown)


cjihrig:
It looks like there are some linting and type checking issues in the CI.

Skn0tt:
Made updates to fix the failures, should look better now.

cjihrig (approved):
LGTM, thanks!

PR #2896, opened 2022-11-16
(fmt/bytes): rename prettyBytes

timreichen:
- rename `prettyBytes` to `format`
- deprecate `prettyBytes`
- rename `PrettyBytesOptions` to `FormatOptions`
- deprecate `PrettyBytesOptions`

### DIFF
--- a/fmt/bytes.ts
+++ b/fmt/bytes.ts
@@ -17,7 +17,7 @@ type LocaleOptions = {
 };
 
 /**
- * @depreacted (will be reomved after 0.170.0) use `FormatOptions` instead
+ * @depreacted (will be removed after 0.170.0) use `FormatOptions` instead
  */
 export interface PrettyBytesOptions {
   /** Uses bits representation. Default is false. */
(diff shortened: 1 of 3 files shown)


fmt/bytes.ts
```diff
@@ -17,7 +17,7 @@ type LocaleOptions = {
 };
 
 /**
- * The options for pretty printing the byte numbers.
+ * @depreacted (will be removed after 0.170.0) use `FormatOptions` instead
```
cjihrig, line 20: This is spelled incorrectly in a few places:

```suggestion
 * @deprecated (will be removed after 0.170.0) use `FormatOptions` instead
```

cjihrig (commented):
The code changes LGTM minus my comment. @kt3k is this a name change that we want to make?

MierenManz:
The rename seems like a good idea to me. Consistency is best

retraigo:
The renaming would make the `fmt` module itself consistent if #2871 is going to be merged. It will be weird if `fmt/duration` is named `format` and `fmt/bytes` is named `prettyBytes` while they both do similar stuff.

kt3k:
I'm now slightly in favor of this change as 'format' sounds like a more appropriate name for _standard_ modules than 'prettyBytes'.

cjihrig (approved):

lino-levan:
+1 on the name change, feels a lot more "standard module" and it has the bonus of being closer to web naming specs.

kt3k:
Looks like we are all supportive to this.

@timreichen Thanks for your suggestion!

kt3k (approved):
LGTM

Issue #2859, opened 2022-11-09
node/string_decoder: GenericDecoder doesn't handle incomplete chars correctly


kt3k:
If the multi-byte chars are cut in the middle in the input, GenericDecoder (in node/string_decoder.ts) doesn't handle them correctly. The following script prints invalid chars.

```js
import { StringDecoder } from "./node/string_decoder.ts";
const text = "あいう";
const bytes = new TextEncoder().encode(text);
const dec = new StringDecoder();
let str = "";
str += dec.write(bytes.slice(0, 4));
str += dec.write(bytes.slice(4));
console.log(str);
```

For reference, the following node.js script prints correct original text.

```js
import { StringDecoder } from "string_decoder";
const text = "あいう";
const bytes = new TextEncoder().encode(text);
const dec = new StringDecoder();
let str = "";
str += dec.write(bytes.slice(0, 4));
str += dec.write(bytes.slice(4));
console.log(str);
```

rush1506:
Has anyone works on this yet? If not, can I take this up as my first issue? :)

cjihrig:
@rush1506 go for it.

kt3k:
fixed in #2897

PR #2897, opened 2022-11-16
fix(node/string_decoder) proper buffer type casting and fix default logic

rush1506:
Issues: https://github.com/denoland/deno_std/issues/2859
I found out there are 4 problems when checking on this:
1) decoder needs `Buffer` while what received was `TypedArray`. Even though those two are similar, there are some fundamental difference between the pair. Most notably the `copy` method and how the `Buffer View` works.
2) With the current code in the issue, if you add specific encoding `utf8` it would also not work
3) The current `string_decoder` has different encoding normalisation logic from `node` code. Can reference here - https://github.com/nodejs/string_decoder .By default, if not specified, all encoding should be `utf8`, `GenericDecoder` is for unrecognised `encoding` only. 
 4) Current `deno` `string_decoder` is using old `string_decoder` logic from nodes, the current `node` has converted the function to handle in `C`. 

So in this PR, I use the existing function `normalizeEncoding` to follow `node` defaulting logic. And add a `TypedArray` check before start writing `Buffer`. If it's `TypeArray`, we would cast it properly to `Buffer`

### DIFF
--- a/node/string_decoder.ts
+++ b/node/string_decoder.ts
@@ -202,7 +202,11 @@ function utf8Write(
   } else {
     i = 0;
   }
-  if (i < buf.length) return r ? r + this.text(normalizedBuffer, i) : this.text(normalizedBuffer, i);
+  if (i < buf.length) {
+    return r 
+      ? r + this.text(normalizedBuffer, i) 
+      : this.text(normalizedBuffer, i);
+  }
   return r || "";
 }
 
--- a/node/string_decoder.ts
+++ b/node/string_decoder.ts
@@ -21,7 +21,6 @@
 // USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 import { Buffer } from "./buffer.ts";
-import { core } from "./_core.ts";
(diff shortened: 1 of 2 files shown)


kt3k:
@rush1506 Can you find some test cases in node.js test cases which this PR enables? If not, can you add test cases which this PR fixes?

This undos the performance tuning introduced in #2858, but I think we should land this PR for now as #2858 only works for limited cases (no incomplete char cases). What do you think? @littledivy

rush1506:
@kt3k Let me add test cases which this PR fixes. As the new node version that use `C` for `string_decoder` would not have this problem. Only old codes does, and it was not fixed as well.

rush1506:
Any update on this guys?

littledivy (commented):
Seems fine by me.

kt3k (approved):
LGTM

kt3k:
@littledivy Thanks for checking. Let's do perf improvement again with stateful fast decoder

kt3k:
@rush1506 Thanks for a lot of corrections!

PR #2898, opened 2022-11-16
fix(node/http): request with headers

lideming:
This PR makes `http.request()` with the `headers` option actually send the headers.

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -96,6 +96,7 @@ export interface RequestOptions {
   href?: string;
 }
 
+// TODO: Implement ClientRequest methods (e.g. setHeader())
 /** ClientRequest represents the http(s) request from the client */
 class ClientRequest extends NodeWritable {
   defaultProtocol = "http:";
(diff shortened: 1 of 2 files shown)


lideming:
> Are there tests from the Node.js test suite that could be ported for this instead of writing our own?

I found tests for [automatic-headers](https://github.com/nodejs/node/blob/main/test/parallel/test-http-automatic-headers.js), [blank-header](https://github.com/nodejs/node/blob/main/test/parallel/test-http-blank-header.js) and [default-headers](https://github.com/nodejs/node/blob/main/test/parallel/test-http-client-default-headers-exist.js), but seems no test for exactly this PR (request did send headers).

kt3k:
> I found tests for [automatic-headers](https://github.com/nodejs/node/blob/main/test/parallel/test-http-automatic-headers.js), [blank-header](https://github.com/nodejs/node/blob/main/test/parallel/test-http-blank-header.js) and [default-headers](https://github.com/nodejs/node/blob/main/test/parallel/test-http-client-default-headers-exist.js), but seems no test for exactly this PR (request did send headers).

Fair enough. Thanks for checking

kt3k (approved):
LGTM

PR #2904, opened 2022-11-16
fix(fmt/bytes): default local decimal and group symbols

timreichen:
Fix tests with `decimal` and `group` symbols on unspecified `locale: true`.
This test failed before when run on default german system.


kt3k (approved):
LGTM. Thanks!

PR #2908, opened 2022-11-17
test(node): ignore Node API test on aarch64 darwin

### LINKED ISSUE #2856: "require Node-API module" test fails on Mac M1
**Describe the bug**

The test fails on Mac M1, because there's no Node-API file for this arch.
It should be ignored on this architecture.

kt3k:
closes #2856


bartlomieju (approved):
Thanks!

PR #2915, opened 2022-11-20
fix(node/fs): enable type narrowing with instanceof for ReadStream and WriteStream

### LINKED ISSUE #2891: TypeScript error with `WriteStream`
**Describe the bug**

Described in the title.

**Steps to Reproduce**

```ts
import { WriteStream } from "https://deno.land/std@0.164.0/node/fs.ts";

if (something instanceof WriteStream) {
  something.path // ts error
}
```

**Expected behavior**

Should behave like /std@0.155.0/node/fs.ts.

**Environment**

N/A

PolarETech:
Fixes #2891

References:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/fs.d.ts

### DIFF
--- a/node/fs.ts
+++ b/node/fs.ts
@@ -53,6 +53,7 @@ import {
   writeFileSync,
 } from "./_fs/_fs_writeFile.ts";
 import { Stats } from "./internal/fs/utils.mjs";
+// @deno-types="./internal/fs/streams.d.ts"
 import {
   createReadStream,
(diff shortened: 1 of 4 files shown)


kt3k (approved):
Looks better aligned to what node.js does. Thanks for the PR @PolarETech 

LGTM

cjihrig (approved):

PR #2919, opened 2022-11-21
fix(node/timers): implement timeout.refresh

kt3k:
part of undici issue https://github.com/denoland/deno/issues/16710

This PR implements `refresh` method of Timeout object.

(Note: after this change requests to `http` servers seem working, but undici still has problem with `https` servers.)


bartlomieju (approved):
LGTM, I tested this change with npm tests in Deno and they seem fine

PR #2922, opened 2022-11-21
feat: Add process._kill and shim process.kill(pid, 0) correctly

Soremwar:
`process.kill` has a special property in which it allows to check for a process existence with a signal of 0

This PR, enables that behaviour and adds `process._kill`, an API used for testing but that is likely being used in the wild as well


node/_tools/setup.ts
```diff
@@ -19,6 +19,8 @@ import { copy, readAll, writeAll } from "../../streams/conversion.ts";
 import { downloadFile } from "../../_util/download_file.ts";
 import { updateToDo } from "./list_remaining_tests.ts";
 
+Foras.initSyncBundledOnce();
```
Soremwar, line 22: Denoflate requires the WASM bundle to be initialized before it can be used

cjihrig:
The test added here is failing in the CI. Also, please use camelCase instead of snake_case in the code you've changed.

node/_fs/_fs_access.ts
```diff
@@ -1,7 +1,8 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
 import { type CallbackWithError, makeCallback } from "./_fs_common.ts";
-import { fs, os } from "../internal_binding/constants.ts";
```
Soremwar, line 4: All this instances of `os.errno` were incorrect. In Node the error code is based on the codes from "uv" rather than "constants"

node/dgram.ts
```diff
@@ -455,12 +455,12 @@ export class Socket extends EventEmitter {
...
       }
 
-      let flags = 0;
+      let flags: number | undefined = 0;
 
       if (state.reuseAddr) {
         flags |= UV_UDP_REUSEADDR;
       }
-      if (state.ipv6Only) {
+      if (state.ipv6Only && UV_UDP_IPV6ONLY) {
```
Soremwar, line 463: Bitwise assigned won't change the variable if the assignment is undefined, however TypeScript complains

node/internal_binding/constants.ts
```diff
@@ -1,145 +1,633 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
```
Soremwar, line 1: Manually obtained the constants for all 3 os, this one is the right one. However the error codes won't match those in UV (they are not supposed to)

node/os.ts
```diff
@@ -306,56 +309,6 @@ export function userInfo(
   notImplemented(SEE_GITHUB_ISSUE);
 }
 
-export const constants = {
```
Soremwar, line 309: I understand what the intent behind this was, however it's best to throw on unimplemented for `Deno.kill` rather than change Node's behaviour and face incompatibility issues in the future

node/constants.ts
```diff
@@ -55,7 +55,6 @@ export const {
   SIGBUS,
   SIGCHLD,
   SIGCONT,
-  SIGEMT,
```
Soremwar, line 58: Sadly since this has to be a named export, all the constants have to be explicitly named. There are probably quite a few that weren't included here because of that (or even this one that I couln't find anywhere)

node/os_test.ts
```diff
@@ -173,17 +173,6 @@ Deno.test({
   },
 });
 
-Deno.test({
```
Soremwar, line 176: This was never right, IDK why this was done this way

node/_fs/_fs_access.ts
```diff
@@ -57,7 +58,7 @@ export function access(
       );
       e.path = path;
       e.syscall = "access";
-      e.errno = os.errno.ENOENT;
+      e.errno = codeMap.get("EACCES");
```
kt3k, line 61: Should this be `codeMap.get("ENOENT")`?

node/_fs/_fs_access.ts
```diff
@@ -110,7 +111,7 @@ export function accessSync(path: string | Buffer | URL, mode?: number) {
...
```
kt3k, line 114: Should this be `codeMap.get("ENOENT")`?

node/dgram.ts
kt3k: I guess we should do `UV_UDP_IPV6ONLY!` below if this is only TypeScript issue

node/process.ts
```diff
@@ -243,22 +253,72 @@ memoryUsage.rss = function (): number {
   return memoryUsage().rss;
 };
 
-export function kill(pid: number, sig: Deno.Signal | number = "SIGTERM") {
+// Returns a negative error code than can be recognized by errnoException
+function _kill(pid: number, sig: number): number {
+  let err_code;
```
kt3k, line 258:
```suggestion
  let errCode;
```

node/process.ts
```diff
@@ -243,22 +253,72 @@ memoryUsage.rss = function (): number {
...
+      });
+    }
+
+    if (!status.success) {
+      err_code = uv.codeMap.get("ESRCH");
+    }
+  } else {
+    // Reverse search the shortname based on the numeric code
+    const maybe_signal = Object.entries(constants.os.signals).find((
+      [_, numeric_code],
```
kt3k, line 278:
```suggestion
      [_, numericCode],
```

node/process.ts
```diff
@@ -243,22 +253,72 @@ memoryUsage.rss = function (): number {
...
+        args: ["-0", pid],
+      });
+    }
+
+    if (!status.success) {
+      err_code = uv.codeMap.get("ESRCH");
+    }
+  } else {
+    // Reverse search the shortname based on the numeric code
+    const maybe_signal = Object.entries(constants.os.signals).find((
```
kt3k, line 277:
```suggestion
    const maybeSignal = Object.entries(constants.os.signals).find((
```

node/os_test.ts
kt3k: Right. These should be numbers, instead of strings

node/_fs/_fs_access.ts
Soremwar: My bad

Soremwar: Kinda worried that the current tests didn't catch this though

Soremwar:
Yeah, I'm sorting it out

node/internal_binding/constants.ts
```diff
@@ -1,11 +1,15 @@
...
+// All properties have to be initialized since they'll be reexported as named exports down the line in node/constants
+
 let os: {
   dlopen: {
-    RTLD_DEEPBIND?: number;
-    RTLD_GLOBAL?: number;
-    RTLD_LAZY?: number;
-    RTLD_LOCAL?: number;
-    RTLD_NOW?: number;
+    RTLD_DEEPBIND: number | undefined;
```
Soremwar, line 8: Making sure TS complains if the properties aren't initialized

Soremwar:
Ok, so this gets really complicated because CommonJS

Basically, in order for everything to work the way it should we need to conditionally use named exports based on the OS, however this is not possible with ESM

If we export everything regardless of OS, things like the spawnChild break, because it looks at all the signals available in the "constants.signals" export, even if they are undefined

If we don't, we can't ensure that the default export and the named exports match, so chances are there'll be a lot more bugs there

What should we do?

kt3k:
@Soremwar Let's just skip that test case in `module_all_test.ts`. Thanks for looking into it.

kt3k (approved):
LGTM

PR #2924, opened 2022-11-22
feat(node): export `process.argv`

### LINKED ISSUE #2917: node: "process" module should have `argv` export
```
import { argv } from 'process'
```

https://nodejs.org/api/process.html#processargv

Should be added in https://github.com/denoland/deno_std/blob/main/node/process.ts

iuioiua:
Fixes #2917.

### DIFF
--- a/node/process.ts
+++ b/node/process.ts
@@ -23,7 +23,7 @@ import { _exiting } from "./_process/exiting.ts";
 export {
   _nextTick as nextTick,
   arch,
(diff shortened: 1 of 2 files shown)


cjihrig:
It would be nice to have a regression test for this.

lino-levan (approved):
Other than no regression test, LGTM

cjihrig (approved):

PR #2926, opened 2022-11-22
fix(node/tls): implement secureConnect event

kt3k:
part of https://github.com/denoland/deno/issues/16710

This PR implements `secureConnect` event of `TLSSocket` class. This change is necessary for undici to work with `npm:` specifier.

This PR also changes `tls.Server` behavior to pass the compat test case `node/_tools/test/parallel/test-tls-connect-simple.js`.

This PR also updates `node/_tools/setup.ts` to enable copying of more files (in this case `fixtures/keys/agent1-cert.pem` and `fixtures/keys/agent1-key.pem`) under Node.js `test/` directory.


kt3k:
blocked by #2927

cjihrig (approved):

PR #2929, opened 2022-11-22
feat(async): retry

### LINKED ISSUE #2829: suggestion: retry algorithms
I like the idea of having a retry function with a BYO algorithm. A common use case is initiating a TCP connection for a database or service. My idea is to have both the retry function and a few popular algorithms, perhaps in `std/async/retry.ts`. Example:
```ts
const conn = await retry(
  async () => await Deno.connect({ port: 6379 }),
  exponentialBackoff({
    multiplier: 2,
    maxInterval: 5000,
    minInterval: 500,
  }),
);
```
This function would retry until the algorithm reaches its retry limit and then reject.

I'd like to work on this, if it sounds like a good idea.

lino-levan:
closes #2829


async/retry_test.ts
```diff
@@ -0,0 +1,24 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+import { retry } from "./retry.ts";
+import { assert, assertRejects } from "../testing/asserts.ts";
+
+function generateErroringFunction(errorsBeforeSucceeds: number) {
+  let errorCount = 0;
+
+  return () => {
+    if (errorCount >= errorsBeforeSucceeds) return true;
```
kt3k, line 9: maybe let's return `errorCount` and assert that number?

async/retry.ts
```diff
@@ -0,0 +1,50 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+const defaultRetryOptions = {
+  multiplier: 2,
+  maxTimeout: 60000,
+  maxAttempts: 5,
+  minTimeout: 100,
```
kt3k, line 7: 100 feels a little short to me. What about 1000, which is the default of popular [retry](https://www.npmjs.com/package/retry) module?

iuioiua:
I think a new `RetryError` or similar should be created, separately exported and thrown once retries are exhausted, like `async/deadline`.

async/retry.ts
```diff
@@ -0,0 +1,50 @@
...
+  fn: (() => Promise<T>) | (() => T),
+  opts?: RetryOptions,
+) {
+  opts = {
+    ...defaultRetryOptions,
+    ...opts,
+  };
+
+  if (opts.maxTimeout >= 0 && opts.minTimeout > opts.maxTimeout) {
+    throw "minTimeout is greater than maxTimeout";
```
iuioiua, line 31: Should this throw a `RangeError`?

async/README.md
```diff
@@ -236,3 +236,26 @@ const delayedPromise = delay(1000);
 // Below throws `DeadlineError` after 10 ms
 const result = await deadline(delayedPromise, 10);
 ```
+
+# retry
+
+Creates a retry promise which resolves to the value of the input. If the input
+promise throws, it will be retried `maxAttempts` amount of times. It will retry
```
cjihrig, line 243:
```suggestion
promise throws, it will be retried `maxAttempts` number of times. It will retry
```

async/retry.ts
```diff
@@ -0,0 +1,50 @@
...
+  let error = undefined;
+
+  for (let i = 0; i < opts.maxAttempts; i++) {
+    try {
+      return await fn();
+    } catch (err) {
+      await new Promise((r) => setTimeout(r, timeout));
+      timeout *= opts.multiplier;
+      timeout = Math.max(timeout, opts.minTimeout);
+      if (opts.maxTimeout >= 0) timeout = Math.min(timeout, opts.maxTimeout);
```
cjihrig, line 44: Please move the `timeout` assignment to its own line.

async/retry_test.ts
cjihrig, line 9: Please move the `return` to its own line.

lino-levan:
@iuioiua 
> I think a new RetryError or TimeoutError or similar should be created, separately exported and thrown once retries are exhausted, like async/deadline.

I disagree on this one. I think there's a lot of value in throwing the error that the input function gave, and it seems to be the standard among other modules that do `retry` iirc. We could definitely take a different route, but I would be interested in hearing the reasoning for this one.

lino-levan:
Sorry about that! Did not realize how the GitHub review system works with multiple reviewers.

iuioiua:
Throwing a `RetryError` or similar indicates whether the failure is flaky or not. This would be more valuable in third-party dependencies where the developer may not even know that a retry mechanism is in play until they see the error.

I agree the original error thrown from the function failure should also be preserved. Both errors would be valuable. I think the best situation would be to throw a `RetryError` with a [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) property defined.

lino-levan:
@iuioiua does this look good to you?

iuioiua:
LGTM! Good stuff!

async/retry.ts
```diff
@@ -0,0 +1,63 @@
...
+    try {
+      return await fn();
+    } catch (err) {
+      await new Promise((r) => setTimeout(r, timeout));
+      timeout *= options.multiplier;
+      timeout = Math.max(timeout, options.minTimeout);
+      if (options.maxTimeout >= 0) {
+        timeout = Math.min(timeout, options.maxTimeout);
+      }
+      error = err as string;
```
kt3k, line 58: This type assertion looks wrong to me. How about passing error to RetryError constructor as unknown type?

async/retry.ts
```diff
@@ -0,0 +1,63 @@
+// Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
+
+export class RetryError extends Error {
+  constructor(cause: string) {
+    super("Retry");
```
kt3k, line 5: I think this error message should be more descriptive like:

```js
`Exceeded max retry count (${count})`
```

async/retry.ts
lino-levan: I agree, that definitely felt a bit dubious. Will do.

kt3k (approved):
LGTM. Nice work!

lino-levan:
@kt3k any idea why tests aren't running? Is it because of the merge conflict? Should I do anything about it?

cjihrig:
@lino-levan there is a merge conflict. Can you resolve that and push - hopefully that will kick start the CI.

lino-levan:
@cjihrig does this look good?

lino-levan:
@cjihrig Given that we have moved away from having README files for each sub-module, yes. It was causing a merge conflict anyways.

Async currently doesn't have a readme: https://github.com/denoland/deno_std/tree/main/async

cjihrig (approved):
LGTM. Thanks!

PR #2931, opened 2022-11-23
fix: wording for cookie value validation error

Skn0tt:
Before, when setting a cookie with a value that contained a space (like `United Kingdom`), the error message implied that the value was `" "`. This can be confusing, because what it really means is that the value *contained* `" "`! This PR updates that error message, so it's more descriptive of the error.


kt3k (approved):
LGTM. Thanks!

PR #2933, opened 2022-11-23
feat(dotenv): support type inference based on `restrictEnvAccessTo` option

### LINKED ISSUE #2930: suggestion(dotenv): support type inference

`std/flags` supports type inference based on the options passed. It would be useful to support type inference in `std/dotenv` as well.

For example, `std/dotenv` provides `restrictEnvAccessTo` option, so return type could be inferred based on this.

```typescript
const conf = await config({
  restrictEnvAccessTo: ["MONGO_URL"],
});

conf.MONGO_URL; // This property will be inferred

type Config = { MONGO_URL: string };
const conf2: Config = conf; // This should not be error
```

uki00a:
Closes #2930

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -11,7 +11,17 @@ export interface DotenvConfig {
   [key: string]: string;
 }
 
-type StringList = Array<string> | undefined;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

Issue #2928, opened 2022-11-22
Allow Cookie['expires'] to be a number


AleksandrHovhannisyan:
**Is your feature request related to a problem? Please describe.**

I was trying to set a cookie on [Netlify Edge](https://docs.netlify.com/edge-functions/api/), which runs on Deno under the hood, using code that looks like this:

```js
context.cookies.set({
  name: 'name',
  value: `value`,
  // 10 seconds from now—this is the important part
  expires: Date.now() + 10000,
});
```

However, it was failing with this error:

```
TypeError: date.getUTCDate is not a function
    at toIMF (https://deno.land/std@0.114.0/datetime/mod.ts:99:24)
    at toString (https://deno.land/std@0.114.0/http/cookie.ts:80:24)
    at setCookie (https://deno.land/std@0.114.0/http/cookie.ts:195:13)
    at https://637b7052e167bb00082f54f1--edge.netlify.app/bootstrap/cookie_store.ts:51:11
    at Array.forEach ()
    at CookieStore.apply (https://637b7052e167bb00082f54f1--edge.netlify.app/bootstrap/cookie_store.ts:40:14)
    at FunctionChain.run (https://637b7052e167bb00082f54f1--edge.netlify.app/bootstrap/function_chain.ts:237:18)
    at async handleRequest (https://637b7052e167bb00082f54f1--edge.netlify.app/bootstrap/handler.ts:58:22)
    at async Server.#respond (https://deno.land/std@0.136.0/http/server.ts:298:18)
```

Looking at the definition for Deno's `setCookie` and [the `Cookie` interface](https://deno.land/std@0.114.0/http/cookie.ts?s=Cookie), I realized that the type for `Cookie['expires']` is strictly a `Date` object. Netlify claimed that this cookie store mirrors the web standard, so I was trying to follow [this example from the MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore/set#examples):

```js
const day = 24 * 60 * 60 * 1000;
cookieStore.set({
  name: "cookie1",
  value: "cookie1-value",
  expires: Date.now() + day,
  domain: "example.com",
});
```

Because the web standard (still experimental) types `expires` as a number:

> A timestamp, given as [Unix time](https://developer.mozilla.org/en-US/docs/Glossary/Unix_time) in milliseconds, containing the expiration date of the cookie.

**Describe the solution you'd like**

[...]

PR #2935, opened 2022-11-24
docs(cookie): clarify properties and their usage

AleksandrHovhannisyan:
There isn't an issue specifically for this, but I figured it would be helpful to clarify some of these properties and provide examples for `Expires` since it now supports two types. Prompted by #2928 and the changes in #2932.


kt3k (approved):
Looks nice improvements to me. Thanks!

cjihrig (approved):

PR #2939, opened 2022-11-24
add image/jxl mime type

Joshix-1:
[JPEG XL](https://jpeg.org/jpegxl/) is a new image format.
[Wikipedia](https://en.wikipedia.org/wiki/JPEG_XL), [jpegxl.info](https://jpegxl.info/), [Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1539075) and [Palemoon](https://repo.palemoon.org/MoonchildProductions/UXP/commit/fb0c204f2b59485e796b93ce89f73af552b05c2b) say the mime type is image/jxl
It is even supported by at least one [browser](https://www.palemoon.org/releasenotes.shtml).

That's why I think the mime type db should be updated.

### DIFF
--- a/media_types/vendor/mime-db.v1.52.0.ts
+++ b/media_types/vendor/mime-db.v1.52.0.ts
@@ -7152,6 +7152,9 @@ export default {
     "compressible": false,
     "extensions": ["jpx", "jpf"]
   },
+  "image/jxl": {
+    "extensions": ["jxl"]
+  },
   "image/jxr": {
     "source": "iana",
     "extensions": ["jxr"]


kt3k:
We don't maintain this file by ourselves, but we follow [mime-db](https://github.com/jshttp/mime-db) updates. Please follow their instruction for updating data.

https://github.com/jshttp/mime-db#registering-media-types

> Registering Media Types
The best way to get new media types included in this library is to register them with the IANA. The community registration procedure is outlined in [RFC 6838 section 5](https://tools.ietf.org/html/rfc6838#section-5). Types registered with the IANA are automatically pulled into this library.

Issue #2944, opened 2022-11-25
node compat test `parallel/test-http-outgoing-finish-writable.js` fails


kt3k:
The test seems to have started failing since https://github.com/denoland/deno/pull/16616 

(it's fine with https://github.com/denoland/deno/commit/b6f49cf4790926df125add2329611a8eff8db9da, started failing with https://github.com/denoland/deno/commit/fd023cf7937e67dfde5482d34ebc60839eb7397c

PR #2945, opened 2022-11-25
test(node/http): modify writable of ClientRequest

kt3k:
This PR tries to fix https://github.com/denoland/deno_std/issues/2944


cjihrig (approved):
This feels wrong, but it does get the CI passing again, so LGTM

kt3k:
> This feels wrong, but it does get the CI passing again, so LGTM

Thanks! 🙏  Yeah, let's revisit soon.

Issue #2918, opened 2022-11-21
`normalize` doesn't remove *all* `.` and `..` segments


jflatow:
**Describe the bug**

The documentation says `normalize` resolves `..` and `.` segments, but these can remain in the output if they are the *only* components of the resulting path. Probably the documentation should point this out, if it is intended.

**Steps to Reproduce**

e.g. `normalize('.')` results in `.` and `normalize('..')` results in `..`

cjihrig:
@jflatow would you like to update the docs?

jflatow:
> @jflatow would you like to update the docs?

Can you point me to the source?

iuioiua:
`normalize` lives in `path/posix.ts` and `path/win32.ts`.

jflatow:
> `normalize` lives in `path/posix.ts` and `path/win32.ts`.

The docs are split between those files?

iuioiua:
Yes, `normalize` works differently internally on POSIX systems than on Windows systems.

I'm not sure what platform you're issue is regarding, but I'm guessing you're looking to modify the JSDocs for one or both functions.

cjihrig:
Documented in https://github.com/denoland/deno_std/pull/2946

PR #2946, opened 2022-11-25
docs(path): clarify the meaning of resolve for normalize

jflatow:
Addresses #2918.


kt3k:
The doc looks good to me, but does the same note apply to `normalize` in `path/win32.ts`?

jflatow:
> The doc looks good to me, but does the same note apply to `normalize` in `path/win32.ts`?

I'm not sure and I can't test it, so I left it off

kt3k (approved):
LGTM

cjihrig (approved):
LGTM. I think we'll have to wait to merge until the CI is fixed.

PR #2950, opened 2022-11-27
Revert "test(node/http): modify writable of ClientRequest (#2945)"

cjihrig:
This reverts commit 5f5413df9ae7bf27819f2a112250a00822e811bb.

This change was made in response to https://github.com/denoland/deno/pull/16616. However, that change was reverted in https://github.com/denoland/deno/pull/16839 due to it breaking Vite. Let's revert it here as well.


lino-levan:
A bit of a bummer because that fix in theory was quite nice. This revert LGTM

kt3k (approved):
LGTM

Issue #2921, opened 2022-11-21
node: Get rid of Top-Level Await in "node/worker_threads.ts"


bartlomieju:
I'm trying to bundle "std/node/module_all.ts" into a single file, so it could be easily embedded into Deno, but I found one file that uses TLA that makes it harder to bundle:
https://github.com/denoland/deno_std/blob/5e5309ecb42a8ce3d1f82437e7f8995bc8aa290e/node/worker_threads.ts#L178

Ideally we could remove it this TLA and keep the module working the same.

cjihrig:
@bartlomieju would something like this work? `deno task node:test` passes locally for me with this change.

```diff
diff --git a/node/worker_threads.ts b/node/worker_threads.ts
index db525562..f5eedc82 100644
--- a/node/worker_threads.ts
+++ b/node/worker_threads.ts
@@ -175,10 +175,12 @@ if (!isMainThread) {
     notImplemented("parentPort.removeAllListeners");
 
   // Receive startup message
-  [{ threadId, workerData, environmentData }] = await once(
-    parentPort,
-    "message",
-  );
+  (async function() {
+    [{ threadId, workerData, environmentData }] = await once(
+      parentPort,
+      "message",
+    );
+  })()
 
   // alias
   parentPort.addEventListener("offline", () => {
```

bartlomieju:
@cjihrig yes, it appears this solves the problem!

kt3k:
Is this a still blocker of bundling work?

bartlomieju:
@kt3k I believe it still is

kt3k:
worker_thread is now disabled in `ext/node`

vixalien:
@kt3k this is a bummer. it prevents some node wasm modules from working.

vixalien:
(ffmpeg-wasm) specifically

PR #2951, opened 2022-11-27
chore: remove TLA in node worker_threads module

cjihrig:
This is reported to solve the problem in the linked issue.

Fixes: https://github.com/denoland/deno_std/issues/2921


cjihrig:
Doesn't work

PR #2955, opened 2022-11-29
refactor(bytes): move to single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the bytes module.

This PR also:
1. Removes the exporting of `equalsNaive` and `equals32Bit`, as they're private functions.
2. Removes the likely no longer needed `equals_bench.ts`.


kt3k (approved):
LGTM. Nice work!

PR #2957, opened 2022-11-29
refactor(streams): move to single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the streams module.


cjihrig (approved):

kt3k:
I feel that `streams/mod.ts` is a little unfinished. Do we deprecate this file, or keep it as an entrypoint of `streams`?

iuioiua:
> I feel that `streams/mod.ts` is a little unfinished. Do we deprecate this file, or keep it as an entrypoint of `streams`?

I agree. I'll submit a cleanup PR shortly. That was one of the first restructures I did. I think my approach now is much better.

We should keep `mod.ts` as an entry point for the module and not deprecate it.

PR #2958, opened 2022-11-29
refactor(archive): move to single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the archive module.


iuioiua:
Do we want `_common.ts` and `_test_common.ts` or `_util.ts` and `_test_util.ts`? I'm impartial.

archive/mod.ts
```diff
@@ -0,0 +1,40 @@
...
+
+/**
+ * Provides a `Tar` and `Untar` classes for compressing and decompressing
+ * arbitrary data.
+ *
+ * @module
+ */
+/** @deprecated (will be removed after 0.169.0) TODO: export all once overlapping objects/types are removed */
+export { Tar, type TarData, type TarDataWithSource } from "./tar.ts";
+export * from "./untar.ts";
```
kt3k, line 40: `mod.ts` doesn't exist now. Does it make sense to add these deprecated versions here?

iuioiua: Previously, `tar.ts` acted as the module entry point. Now that `Tar` and `Untar` classes have been separated, `mod.ts` acts as the entry point for both classes.

iuioiua: It'll also provide a module description where [one currently doesn't exist for `std/archive`](https://deno.land/std?doc).

kt3k:
> Do we want _common.ts and _test_common.ts or _util.ts and _test_util.ts? I'm impartial.

hmm both seem having examples:

```
crypto/_util.ts
fs/_util.ts
media_types/_util.ts
node/path/_util.ts
path/_util.ts

vs

node/_fs/_fs_common.ts
node/_http_common.ts
node/_tls_common.ts
streams/_common.ts
streams/_test_common.ts
uuid/_common.ts
```

I'm ok with either way.

iuioiua:
Ok, I'll leave it as is then.

kt3k (approved):
LGTM

PR #2963, opened 2022-11-30
chore(fs): enable timestamp tests with copy

lino-levan:
https://github.com/denoland/deno/issues/2411 was resolved two years ago. Creating this PR to see if we can enable this test. 

Edit: We can.


cjihrig (approved):
Thanks!

PR #2969, opened 2022-12-01
fix(bytes): minor tweaks and fixes for full cov.

iuioiua:
This PR applies a few adjustments to the bytes module to be more correct and fully covered by tests.


kt3k (approved):
LGTM. Thanks!

PR #2975, opened 2022-12-03
BREAKING refactor(io): single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the `io` module.


kt3k (approved):
LGTM 👍

PR #2977, opened 2022-12-04
feat(http): Add --header option to file_server

lino-levan:
A big thanks to @plohm12 for the initial work in #2589.

This PR adds `--cache` (or `-C`) as an option for file_server. It follows a similar spec to [http_server](https://github.com/http-party/http-server) over on the node side where `--cache=-1` means `no-store`.


timreichen:
Is there a use case of `maxAge=0`? Or why is it -1 for no-store?

lino-levan:
@timreichen good question. `maxAge=0` essentially behaves something like `must-revalidate`. `no-store` is subtly different than this:
- `must-revalidate` caches the response on the client, but always checks with the server to get the latest information
- `no-store` NEVER caches the response on the client. This is usually for security reasons.

kt3k:
Adding specific option for each header feels problematic to me as there are many HTTP headers. How about adding option for appending any kind of headers by the pairs of keys and values.

lino-levan:
@kt3k 100% agree in theory, but I think this should be an addition feature. I think that caching is a very basic header that should really have been an option.

What are you thinking for cli inputs for adding arbitrary http headers? 

I would imagine something like: `--headers="Set-Cookie: example=hi"` but I have no idea how we would parse multiple headers. Is there a standard separator that we could use (I'm pretty sure's it is newlines but we can't exactly use newlines in this situation)?

kt3k:
> I would imagine something like: --headers="Set-Cookie: example=hi" but I have no idea how we would parse multiple headers. Is there a standard separator that we could use (I'm pretty sure's it is newlines but we can't exactly use newlines in this situation)?

How about following curl's `-H, --header` option?

```
--header "set-cookie: example=hi" --header "cache-control: no-cache" -H "x-custom-header: xyz"
```

`parse` of `flags` module has `collect` option, which enables the usage of the same cli arg multiple times

```ts
parse(Deno.args, { collect: ["header"] });

**Comment by lino-levan** [comment]
@kt3k that makes a lot of sense. So now to move forward, should we :
- Change this PR to add `--header` and remove `--cache`
- Merge this PR, and then open a separate PR for `--header`

I'm game for either, though I personally do see value in `--cache` so I would lean towards option 2.

**Comment by lino-levan** [comment]
Sounds good to me, I will rework this PR when I get the chance (probably this weekend).

**Comment by lino-levan** [comment]
@kt3k PTAL

Thanks!

File: http/file_server.ts [hunk 1]
```diff
@@ -537,6 +542,14 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
 
   if (!opts.quiet) serverLog(req, response!.status);
 
+  if (opts.headers) {
+    for (const header of opts.headers) {
+      const name = header.split(":")[0];
+      const value = header.split(":").slice(1).join(":");
```
kt3k, line 548: I think this should split only once.
```suggestion
      const headerSplit = header.split(":");
      const name = headerSplit[0];
      const value = headerSplit.slice(1).join(":");
```

http/file_server.ts
```diff
@@ -630,17 +647,18 @@ USAGE:
...
-  --no-cors           Disable cross-origin resource sharing
-  -v, --verbose       Print request level logs
-  -V, --version       Print version information
+  -h, --help            Prints help information
+  -p, --port <PORT>     Set port
+  --cors                Enable CORS via the "Access-Control-Allow-Origin" header
+  --host     <HOST>     Hostname (default is 0.0.0.0)
+  -c, --cert <FILE>     TLS certificate file (enables TLS)
+  -k, --key  <FILE>     TLS key file (enables TLS)
+  -H, --header <HEADER> Sets a header on every request
```
kt3k, line 656: Let's give an example to show the syntax. Also let's mention the multi-headers capability.
```suggestion
  -H, --header <HEADER> Sets a header on every request.
                        (e.g. --header "Cache-Control: no-cache")
                        This option can be specified multiple times.
```

kt3k (approved):
LGTM. Nice work!

PR #2978, opened 2022-12-04
refactor(dotenv): remove duplicated functionality

iuioiua:
This PR removes utilities that duplicate functionality within the standard library. These utilities appear to be internal, as they're general and are used exclusively within `std/dotenv`, so I think they're safe to remove without needing deprecation first.


aminmarashi (approved):

lino-levan:
While I think this PR makes sense, I'm a little skeptical of our ability to just remove `util.ts` given that it doesn't follow the rules of what to link to:
> - Has a name or parent with an underscore prefix: _foo.ts, _util/bar.ts.
> - Is that of a test module or test data: test.ts, foo_test.ts, testdata/bar.txt.

I doubt anyone has specifically linked to it, but the fact that it isn't `_util.ts` makes me a little nervous.

kt3k:
> the fact that it isn't _util.ts makes me a little nervous.

Right. It should have been named `_util.ts`

kt3k (approved):
LGTM

PR #2981, opened 2022-12-05
feat(tools): auto-add copyright headers

### LINKED ISSUE #2959: suggestion: automatically add copyright header to files missing one
Currently, the licence check tool prints the list of files missing a copyright header and throws an error. It'd be cool if the tool could instead insert the copyright at the top of the file as defined by the `COPYRIGHT` constant. It should be easy to implement.

The tool is located at `_tools/check_licence.ts` and is run using `deno task lint:licence-headers`.

iuioiua:
This PR changes the licence check tool to automatically add licence headers to the top instead of failing if a file doesn't contain the copyright headers. It also changes the task from `lint:licence-headers` to `fmt:licence-headers`.

This'll be most useful in those PRs that involve many new files.

Closes #2959.


_tools/check_licence.ts
```diff
@@ -32,14 +30,8 @@ for await (
 ) {
   const content = await Deno.readTextFile(path);
   if (!content.includes(COPYRIGHT)) {
-    console.error(`Missing/incorrect copyright header: ${path}`);
-    if (!failed) {
-      failed = true;
-    }
+    const contentWithCopyright = COPYRIGHT + "\n" + content;
```
cjihrig, line 33: Should there also be a `\r` on Windows?

cjihrig (commented):
LGTM with one question.

lino-levan:
This doesn't LGTM in its current form. Left a comment.

lino-levan (changes requested):

_tools/check_licence.ts
```diff
@@ -32,14 +30,8 @@ for await (
 ) {
   const content = await Deno.readTextFile(path);
   if (!content.includes(COPYRIGHT)) {
```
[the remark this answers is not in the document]
iuioiua: That'd only happen in the new year when the current year's value changes. It's usually updated in one of the first PRs of the year. In other words, the case where a second copyright header is created is unlikely to happen.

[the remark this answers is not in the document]
lino-levan: If @kt3k approves, sure. I still think it should be automated as it's some pretty trivial regex, but I don't particularly care either way.

kt3k:
It looks like `fmt:licence-headers` never fails with this change. This now doesn't seem detecting a commits with wrong/missing copyright headers in CI.

iuioiua:
Well... True... I suppose @timreichen is right. I can add a `--check` option, which retains much of the original functionality of the tool that fails if no copyright header is included.

iuioiua:
CI failures are not occurring on the latest stable version of Deno (v1.28.3).

_tools/check_licence.ts
```diff
@@ -31,10 +32,17 @@ for await (
...
   const content = await Deno.readTextFile(path);
+
   if (!content.includes(COPYRIGHT)) {
-    console.error(`Missing/incorrect copyright header: ${path}`);
-    if (!failed) {
+    if (Deno.args.includes("--check")) {
+      console.error(`Missing/incorrect copyright header: ${path}`);
       failed = true;
+    } else {
+      const newline = isWindows ? "\r\n" : "\n";
```
kt3k, line 41: I'm not sure about windows situation of this, but when a windows user pushes the commits to github, these `\r\n`s are changed to `\n`?

kt3k:
> I'm now a little sceptical of the value of this new functionality compared to the complexity it adds. I'll let you guys be the judges.

I'm in favor of landing this as it doesn't look so complex

kt3k:
Let's land after https://github.com/denoland/deno_std/pull/2992 is landed

_tools/check_licence.ts
timreichen: Would it help to add a `.gitattributes` file with `eol` (https://git-scm.com/docs/gitattributes#_eol)?

lino-levan:
@kt3k https://github.com/denoland/deno_std/pull/2992 landed

_tools/check_licence.ts
kt3k: Deno CLI repo seems having that setting: `* text=auto eol=lf`

https://github.com/denoland/deno/blob/9daf6e197a642a88d79614fb53888b5bb954463a/.gitattributes#L2

kt3k: and we already have a similar setting in .editorconfig.

https://github.com/denoland/deno_std/blob/eca4dbfab0a8f5ab5e80daa255c4024e76c042ce/.editorconfig

iuioiua: > Deno CLI repo seems having that setting: `* text=auto eol=lf`

So, add it here too?

kt3k: Ah, deno_std also already have it. https://github.com/denoland/deno_std/blob/6bd933286079d9afbb588092e963e0565c230dfb/.gitattributes#L2

I think we don't need to use "\r\n" on windows. Let's always use "\n" here

kt3k (approved):
@iuioiua Thanks for updating!

LGTM

_tools/check_licence.ts
iuioiua: @kt3k was in favour of leaving it as `\n`.

PR #2982, opened 2022-12-05
fix(encoding): base58 encoding

kratico:
Fix base58 encoding logic

The following was filtering filter `0`s and `undefined`s because `0` is falsy in `.filter`
```
const validOutput = output.filter((item) => item ?? false);
```


lino-levan:
LGTM

kt3k (approved):
LGTM. Thanks!

PR #2965, opened 2022-11-30
refactor: add more functions to node/_core.ts

bartlomieju:
Working on https://github.com/denoland/deno/pull/16881 I noticed some APIs
are not listed here.

PR #2983, opened 2022-12-06
ci: check deploy compatibility in PR

kt3k:
This PR checks the compatibility of `std/node` with Deno Deploy in each PR.

Currently deploy compatibility is only check in `push` event of main branch. So we don't know if a PR breaks the deploy compat before landing it, and it caused an error like https://github.com/denoland/deno_std/pull/2965. This PR prevents it


bartlomieju (approved):
LGTM

PR #2985, opened 2022-12-06
docs(encoding/front_matter): fix doc error when render page

liruifengv:
The page https://deno.land/std@0.167.0/encoding/front_matter/mod.ts is wrong.

fix this https://github.com/denoland/dotland/issues/2616


kt3k (approved):
Thanks!

PR #2990, opened 2022-12-07
chore: update windows VM version in CI

iuioiua:
`ubuntu-22.04-xl` will soon be deprecated (see [here](https://github.blog/changelog/2022-12-01-github-actions-larger-runners-using-ubuntu-latest-label-will-now-use-ubuntu-22-04/)).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,8 +15,8 @@ jobs:
       matrix:
         deno: [canary]
         os:
-          - ${{ github.repository_owner == 'denoland' && 'ubuntu-20.04-xl' || 'ubuntu-20.04' }}
-          - ${{ github.repository_owner == 'denoland' && 'windows-2019-xl' || 'windows-2019' }}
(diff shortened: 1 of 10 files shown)


kt3k:
The change to `windows-2022-xl` looks fine, but `ubuntu-22.04-xl` doesn't seem running in the CI. It looks stuck with the message `Waiting for a runner to pick up this job...`

I guess the release/deprecation schedule for `-xl` runners are different from normal sized runners.

iuioiua:
Are larger runners supposed to be created at the organisation level? Documentation [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#adding-a-larger-runner-to-an-organization).

kt3k:
> Are larger runners supposed to be created at the organisation level? Documentation [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners#adding-a-larger-runner-to-an-organization).

Thanks for the info. This looks interesting! I'll look into it.

kt3k (approved):
LGTM

PR #2991, opened 2022-12-07
feat(CI): add latest stable Deno to CI

iuioiua:
This makes regressions in the canary version clearer. Does it need any further tweaking?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        deno: [canary]
+        deno: [canary, vx.x.x]
         os:
           - ${{ github.repository_owner == 'denoland' && 'ubuntu-20.04-xl' || 'ubuntu-20.04' }}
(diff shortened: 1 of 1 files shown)


iuioiua:
Viola, we can see that the latest stable Deno version, `vx.x.x`, is passing and canary is failing, making a regression in canary clearer.

kt3k:
We discussed this a few times before. ref: https://github.com/denoland/deno_std/issues/870

`std@main` CI needs to pass with `cli@main` (canary) and doesn't necessarily need to pass with the latest stable. (CLI canary can break std sometimes, and std needs to update to follow such changes.)

iuioiua:
I see. Are you suggesting we make CI using Deno stable not required, or are you -1 for the change? Thinking more about it, I think the more straightforward existing setup works fine.

kt3k:
I'm -1 at this point. It can be very complex sometimes to pass both canary and stable, and we want to avoid that situation.

This might be changed when std reached 1.0, but we don't have a plan to release 1.0 in near future.

PR #2993, opened 2022-12-08
refactor: replace `String.prototype.substr()` with `String.prototype.slice()`

iuioiua:
`String.prototype.substr()` is deprecated (see [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr#browser_compatibility)).

### DIFF
--- a/http/_negotiation/media_type.ts
+++ b/http/_negotiation/media_type.ts
@@ -109,7 +109,7 @@ function parseMediaType(
 
     for (const [key, val] of kvps) {
       const value = val && val[0] === `"` && val[val.length - 1] === `"`
-        ? val.slice(1, val.length - 2)
+        ? val.slice(1, val.length - 1)
         : val;
 
       if (key === "q" && value) {
--- a/http/_negotiation/media_type.ts
+++ b/http/_negotiation/media_type.ts
@@ -109,7 +109,7 @@ function parseMediaType(
 
     for (const [key, val] of kvps) {
       const value = val && val[0] === `"` && val[val.length - 1] === `"`
-        ? val.substr(1, val.length - 2)
+        ? val.slice(1, val.length - 2)
         : val;
 
       if (key === "q" && value) {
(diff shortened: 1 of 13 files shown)


http/_negotiation/media_type.ts
```diff
@@ -109,7 +109,7 @@ function parseMediaType(
 
     for (const [key, val] of kvps) {
       const value = val && val[0] === `"` && val[val.length - 1] === `"`
-        ? val.substr(1, val.length - 2)
+        ? val.slice(1, val.length - 2)
```
kt3k, line 112: I think this should be:
```suggestion
        ? val.slice(1, val.length - 1)
```

node/_http_agent.mjs
```diff
@@ -344,7 +344,7 @@ function calculateServerName(options, req) {
         // Leading '[', but no ']'. Need to do something...
         servername = hostHeader;
       } else {
-        servername = hostHeader.substr(1, index - 1);
+        servername = hostHeader.slice(1, index - 1);
```
kt3k, line 347:
```suggestion
        servername = hostHeader.slice(1, index);
```

kt3k:
Let's avoid updating `node/integrationtest/yarn.js` as it's vendored code, not ours.

testing/fast_check_example.ts
```diff
@@ -156,7 +156,7 @@ Deno.test("Can use fast-check to property test indexOf function", async (t) => {
             const pattern = b;
             const index = indexOf(text, pattern);
             return index === -1 ||
-              text.substr(index, pattern.length) === pattern;
+              text.slice(index, pattern.length) === pattern;
```
kt3k, line 159:
```suggestion
              text.slice(index, index + pattern.length) === pattern;
```

iuioiua:
> Let's avoid updating `node/integrationtest/yarn.js` as it's vendored code, not ours.

Done

kt3k (approved):
LGTM. Thanks!

PR #2997, opened 2022-12-13
chore: update Deno.Command usages

kt3k:
This PR updates Deno.Command usages. The default of `stdout` and `stderr` is changed from `piped` to `inherit` https://github.com/denoland/deno/pull/17025


crowlKats (approved):
LGTM

PR #2998, opened 2022-12-13
feat(tools): auto-update copyright headers

schwarzkopfb:
Current implementation of `deno task fmt:licence-headers` adds a secondary header if the year is not correct. This PR extends the tool to update the year instead if there is a header already. 

This will be handy to create a "happy new year 🥳" PR soon.

Related: #2981


kt3k (approved):
LGTM

PR #3001, opened 2022-12-14
refactor(media_types): single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the `media_types` module.


kt3k (approved):
LGTM. Nice work!

kt3k:
Marked the commit as `feat` as new single export files can be considered as new features of media_types module.

PR #3007, opened 2022-12-16
refactor(datetime): single-export files

iuioiua:
This is part of #2936, which aims for leaner dependencies and a more straightforward structure by only having one export per file within the `datetime` module.


kt3k (approved):
LGTM. Thanks!

PR #3008, opened 2022-12-16
fix(http): prevent downstream connections from getting closed when the response stream throws an error

arnauorriols:
Stream errors does not warrant closing the downstream connection. This PR prevents that.

I've added a non-regression test for the errors thrown within new `ReadableStream`s. I've tried to write one for fetch upstream streams but have failed. Any suggestion on how to write such test would be appreciated!

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -310,10 +310,11 @@ export class Server {
       // Send the response.
       await requestEvent.respondWith(response);
     } catch {
-      // `respondWith()` can throw for various reasons, including downstream but also upstream connection errors,
-      // as well as errors thrown during streaming of the response content.
-      // In order to avoid false negatives, we ignore the error here and let `serveHttp` close the connection on
-      // the following iteration if it is in fact a downstream connection error.
+      // `respondWith()` can throw for various reasons, including downstream and
+      // upstream connection errors, as well as errors thrown during streaming
+      // of the response content.  In order to avoid false negatives, we ignore
+      // the error here and let `serveHttp` close the connection on the
+      // following iteration if it is in fact a downstream connection error.
     }
   }
 
--- a/http/server.ts
+++ b/http/server.ts
@@ -287,12 +287,10 @@ export class Server {
    * Responds to an HTTP request.
    *
    * @param requestEvent The HTTP request to respond to.
-   * @param httpCon The HTTP connection to yield requests from.
    * @param connInfo Information about the underlying connection.
    */
   async #respond(
     requestEvent: Deno.RequestEvent,
-    httpConn: Deno.HttpConn,
     connInfo: ConnInfo,
   ) {
     let response: Response;
@@ -312,10 +310,10 @@ export class Server {
       // Send the response.
       await requestEvent.respondWith(response);
     } catch {
-      // respondWith() fails when the connection has already been closed, or there is some
-      // other error with responding on this connection that prompts us to
-      // close it and open a new connection.
-      return this.#closeHttpConn(httpConn);
+      // `respondWith()` can throw for various reasons, including downstream but also upstream connection errors,
+      // as well as errors thrown during streaming of the response content.
+      // In order to avoid false negatives, we ignore the error here and let `serveHttp` close the connection on
+      // the following iteration if it is in fact a downstream connection error.
     }
   }
 
@@ -344,7 +342,7 @@ export class Server {
 
       // Respond to the request. Note we do not await this async method to
       // allow the connection to handle multiple requests in the case of h2.
-      this.#respond(requestEvent, httpConn, connInfo);
+      this.#respond(requestEvent, connInfo);
     }
 
     this.#closeHttpConn(httpConn);
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -1142,16 +1142,21 @@ Deno.test("Server should not close the http2 downstream connection when the resp
   const url = `https://${listenOptions.hostname}:${listenOptions.port}/`;
 
   let n = 0;
+  const a = deferred();
   const connections = new Set();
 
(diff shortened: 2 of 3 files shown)


satyarohith (commented):
> I've tried to write one for fetch upstream streams but have failed. Any suggestion on how to write such test would be appreciated!

Do you mean fetch() shouldn't close a connection during a POST request if the body stream errors? Or am I misunderstanding something?

satyarohith:
>  This PR prevents the downstream connection to be closed in such cases. I have reproduced the bug manually, but I'm out of ideas on how to write a (reliable) automated test for this particular case.

1. ServerA - responds using fetch(ServerB)
2. ServerB - streams a response and errors

Assert that connection from test to ServerA isn't closed when `fetch(ServerB)` errors.

Does this work?

arnauorriols:
>     2. ServerB - streams a response and errors

The difficulty lies in making the `fetch` stream throw. The only way I've been able to reproduce it manually is by breaking the upstream connection abruptly (closing the upstream server is not enough). I guess I could achieve it with some advance use of the `mockListener`, but I'd like some input first about the feasability of this approach.

http/server.ts
```diff
@@ -312,10 +310,10 @@ export class Server {
...
       await requestEvent.respondWith(response);
     } catch {
-      // respondWith() fails when the connection has already been closed, or there is some
-      // other error with responding on this connection that prompts us to
-      // close it and open a new connection.
-      return this.#closeHttpConn(httpConn);
+      // `respondWith()` can throw for various reasons, including downstream but also upstream connection errors,
+      // as well as errors thrown during streaming of the response content.
+      // In order to avoid false negatives, we ignore the error here and let `serveHttp` close the connection on
+      // the following iteration if it is in fact a downstream connection error.
```
ry, line 316: Does it make sense to log something here? Errors that are completely swallowed scare me.

nit: wrap at 80 columns

http/server_test.ts
```diff
@@ -1130,6 +1130,50 @@ Deno.test("Server should not reject when the handler throws", async () => {
...
+  const handler = () => {
+    return new Response(
+      new ReadableStream({
+        async start() {
+          await delay(0);
+          try {
+            throw new Error("test-error");
+          } finally {
+            n++;
+            postRespondWith.resolve();
```
ry, line 1152: It's a little concerning that this deferred is resolved multiple times. It can only be resolved once tho.

arnauorriols: We could remove the deferred. After all, unless somebody has a better idea, this test needs to rely on `await delay(10)` for the following 2 requests (as explained in the code). We might as well use the same pattern for the first request, and at least be consistent.

ry: Maybe just create a new deferred on the same variable on each request? Then you don't need the `delay`

arnauorriols: this is what I mean in the code comment. If I do that and the test fails, the second `await postRespondWith;` hangs forever, cause the connection got broken and the `handler` is not executed again

lucacasonato (changes requested):
As discussed out of band, this test is relying on a bug in ext/http (and is currently incorrect).

http/server_test.ts
arnauorriols: nvm, this test is wrong, and passes because there's a bug in Deno (broken streams should actually close the underlying connection in http1). Going to rewrite the test with http2 with a pattern suggested by @lucacasonato (thanks!)

arnauorriols:
> As discussed out of band, this test is relying on a bug in ext/http (and is currently incorrect).

test rewritten, PTAL

ry (approved):
LGTM

Nice test.

I'm a little concerned nothing is logged in the blanket catch.

lucacasonato (changes requested):
This test is still wrong.

- it still tests requests sequentially
- it still tests on h1 (needs to be h2)
- the response body should only error once all three requests have returned their response headers

One of the requests should fail. The other two should succeed.

arnauorriols:
>  * it still tests on h1 (needs to be h2)

Fixed that. I've had to add the `--unsafely-ignore-certificate-errors` to the test command unfortunately. Any alternative suggestion to avoid it?

> - it still tests requests sequentially
> - the response body should only error once all three requests have returned their response headers

Why are these requirements necessary? I know we discussed them but this implementation is simpler and (I thought was) equally valid.

kt3k:
> Fixed. I've had to add the --unsafely-ignore-certificate-errors to the test command unfortunately. Any alternative suggestion to avoid it?

Have you tried custom HttpClient option? (like https://github.com/denoland/deno_std/blob/651bf8a27aa14354d64555923b6f91598dd8fbc6/http/server_test.ts#L1553-L1556

http/server_test.ts
```diff
@@ -1130,6 +1130,52 @@ Deno.test("Server should not reject when the handler throws", async () => {
   await servePromise;
 });
 
+Deno.test("Server should not close the http2 downstream connection when the response stream throws", async () => {
+  const listenOptions = {
+    hostname: "localhost",
+    port: 4505,
+    certFile: join(testdataDir, "tls/localhost.crt"),
+    keyFile: join(testdataDir, "tls/localhost.key"),
+    alpnProtocols: ["h2", "http/1.1"],
```
ry, line 1139: maybe leave out "http/1.1" ?

arnauorriols:
> Have you tried custom HttpClient option?

Thanks for the tip!

lucacasonato:
@arnauorriols It is invalid because we are still relying on the ext/http bug that requests finish cleanly on error. Really req.text() should be erroring when you throw in the response body.

lucacasonato (approved):
LGTM

PR #3013, opened 2022-12-16
Revert "fix(node/http): do not buffer first chunk (#2989)"

bartlomieju:
This reverts commit eca4dbfab0a8f5ab5e80daa255c4024e76c042ce.

This commit caused performance regression in `express` server from
25k rps to 20k rps.

CC @kt3k

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -285,6 +285,7 @@ export class ServerResponse extends NodeWritable {
   // used by `npm:on-finished`
   finished = false;
   headersSent = false;
+  #firstChunk: Chunk | null = null;
   // Used if --unstable flag IS NOT present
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

Can you show how to reproduce this bug so we can re-introduce the fix, but without the regression?

kt3k:
The below server responds immediately with chunk "A", and then "B" 1 sec later with the fix, but it responds "AB" after 1 sec without the fix.
```js
import * as http from "./node/http.ts";
http.createServer((_, res) => {
  res.write("A");
  setTimeout(() => {
    res.write("B");
    res.end();
  }, 1000)
}).listen(3000);
```

PR #3015, opened 2022-12-17
feat(streams): transformStreamFromProcess

davbrito:
Useful for when you want to pipe several processes in a pipeline. Or when there is a process that transforms the input such as `deno fmt`.

```ts

const dest = await Deno.open("code.ts", { write: true });

const fmtProcess = Deno.run({
  cmd: [Deno.execPath(), "fmt", "-"],
  stdin: "piped",
  stdout: "piped",
});

await generateFile()
    .pipeThrough(transformStreamFromProcess(fmtProcess))
    .pipeTo(dest.writable);
```

I first tried with a naive approach like this:
```ts
inputStream
    .pipeThrough({
       readable: process.stdout.readable,
       writable: process.stdin.writable,
     })
    .pipeTo(outputStream);
```

But it didn't work correctly because it doesn't handle the case when the process returns a failure status. 
This utility function handles that case propagating the error through the stream. If the user wants to ignore the failure status it can pass a flag as an option.

---

An idea that can also comes with this (out of the scope of this repo though) is to include this on the core `Deno.Process`, such as on a `.transform` property getter.
Something like:
```ts
inputStream
  .pipeThrough(processA.transform)
  .pipeThrough(processB.transform)
  .pipeTo(outputStream);
```


lucacasonato:
The new `Deno.Child` API returned from the unstable `Deno.Command` has `.readable: ReadableStream<Uint8Array>` and `.writable: WritableStream<Uint8Array>` methods. As such that new API (which we hope to stabilize soon), already implements TransformStreamLike and as such should be usable directly with `.pipeThrough` :)

kt3k:
I don't think it's good idea to implement this based on `Deno.run` and `Deno.Process` because they are likely to be deprecated soon in favor of `Deno.Command` and `Deno.ChildProcess`.

> already implements TransformStreamLike

Currently [`Deno.ChildProcess`](https://deno.land/api@v1.29.1?unstable=&s=Deno.ChildProcess) has `stdin: WritableStream<Uint8Array>` and `stdout: ReadableStream<Uint8Array>`, but they are not `writable` / `readable`. So `ChildProcess` doesn't look like a TransformStreamLike as is

I'm interested in the idea of making this CLI's feature (including error propagation), like:

```js
const child = new Deno.Command("some-command", { stdin: "piped", stdout: "piped" }).spawn();
const resultStream = source.pipeThrough(child)
```

PR #3016, opened 2022-12-18
fix(media_types): incorrect documentation

lino-levan:
This information is just wrong. I assume that at one point this used to be `application/javascript` but it is no longer the case and should be updated.


lino-levan:
Deploy failed because page wasn't up after 5 retries (it is up now). We should really space out those retries more or allow more of them.

lino-levan:
@kt3k should I push a dummy commit or is this fine?

kt3k (approved):
LGTM

PR #3017, opened 2022-12-19
refactor(io): deprecate `types.d.ts`

iuioiua:
`Reader`, `ReaderSync`, `Writer` and `WriterSync` interfaces are now supported in Deploy.

FYI `types.d.ts` was added in #912, in May 2021.

### DIFF
--- a/io/_test_common.ts
+++ b/io/_test_common.ts
@@ -1,7 +1,5 @@
 // Copyright 2018-2022 the Deno authors. All rights reserved. MIT license.
 
-import type { Reader } from "./types.d.ts";
-
 export const MIN_READ_BUFFER_SIZE = 16;
 export const bufsizes: number[] = [
   0,
@@ -16,7 +14,7 @@ export const bufsizes: number[] = [
   4096,
 ];
 
-export class BinaryReader implements Reader {
+export class BinaryReader implements Deno.Reader {
   index = 0;
 
   constructor(private bytes: Uint8Array = new Uint8Array(0)) {}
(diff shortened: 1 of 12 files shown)


ayame113:
I thought this was part of the work being done on https://github.com/denoland/deno/issues/9795, as https://github.com/denoland/deno_std/pull/912#discussion_r629825680 says. maybe we don't need to remove this? (sorry if i'm wrong)

iuioiua:
Hm... Perhaps, the opposite should be done. Instead of removing `types.d.ts`, it should be extended, so all `Deno.Reader`/`Deno.Writer` instances are replaced with these generic types. Thoughts?

kt3k:
I'm in favor of the opposite of this PR. As @ayame113 points, we're deprecating and removing `Deno.Reader` etc in the future. These types in `io/types.d.ts` should be used everywhere in std instead of `Deno.Reader` etc.

iuioiua:
Another benefit would be gaining web compatibility in some areas, as long as users implement their own `Reader`/`Writer` classes.

@bartlomieju, WDYT? If we all agree, I'd be happy to implement it.

bartlomieju:
> I'm in favor of the opposite of this PR. As @ayame113 points, we're deprecating and removing `Deno.Reader` etc in the future. These types in `io/types.d.ts` should be used everywhere in std instead of `Deno.Reader` etc.

I agree with that as well.

iuioiua:
Closing in favour of #3030.

PR #3021, opened 2022-12-19
fix: don't use windows-xl runners, too expensive

ry:


lino-levan:
LGTM

iuioiua:
Should the same be done with the `ubuntu-20.04-xl` runner?

lino-levan:
@iuioiua pricing for runners varies immensely depending on the OS

ref: https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#per-minute-rates

iuioiua:
True. However, seeing as nothing particularly intensive is going on in the repo, `ubuntu-20.04-xl` may not be needed.

Issue #2538, opened 2022-08-20
clear files structure for submodules in std


timreichen:
**Is your feature request related to a problem? Please describe.**

Submodules as in `std/encoding` have an unnecessary messy file structure and it will become even messier, if we don't introduce a strict file structure for submodules.

An example of that was the introduction of `encoding/csv/stream.ts`, which expands a submodule and necessarily added a `encoding/csv` directory for utility files. Now std has a `encoding/csv` directory only for the stream API while the normal `csv` API lives top level NOT in a directory called `csv`? This is very confusing and unexpected.

All of the submodules have a top level entry point (`encoding/csv.ts`, `encoding/hex.ts` for example), some have additional API (`encoding/csv/stream.ts`) or/and need util or/and wasm files, some need test files. As of now, multiple modules have their files distributed all over the `encoding` dir and some even outside std (see `std/_wasm_varint`)
Here some examples, how disconnected their data and files are:

### toml
- has `toml.ts` as top level entry point
- has `encoding/_toml` for utilities
- has test data in `encoding/testdata`

### varint
- has `varint.ts` as top level entry point
- has `../_wasm_varint` on `std` top level!

**Describe the solution you'd like**

- Let every submodule have it's own directory living in `encoding`. These contain a `mod.ts` for the main logic (basically, what most of the submodules have in their top entry point), test files, util and wasm files/subdirectories.
This ensures that all needed data and files for a submodule are grouped inside one directory and can be expanded by adding files inside the directory if necessary.
- Let every submodule keep a top level entry point which reexports mod.ts and other functions if necessary.
This will ensure that import will not break as it will forward the same data.

### toml
- encoding
  - toml.ts
  - toml
    - mod.ts
    - mod_test.ts
    - parser.ts
    - parser_test.ts
    - testdata
      - CRLF.toml
      - …
### varint
- encoding
  - varint.ts
  - varint
    - _wasm_varint
    - mod.ts
    - mod_test.ts
    
This is a big structural change, but worth it, because it introduces a functioning system for future additions and will get harder to fix over time.

**Describe alternatives you've considered**

### keeping the file structure as-is
The more features and modules are introduced, the more messy it will get. `testdata` will contain a lot of unrelated files from different submodules. If more submodules are introduced that need a wasm directory, it will bloat top level `std`.

lino-levan:
I feel like this is being done with https://github.com/denoland/deno_std/issues/2936. Thoughts @iuioiua?

[...]

PR #3024, opened 2022-12-20
refactor(encoding/csv): unified dir and move to single-export files

timreichen:
ref: https://github.com/denoland/deno_std/issues/2538

This PR moves all csv related stuff into the same dir.
- encoding/csv.ts → encoding/csv/mod.ts
- encoding/test_csv.ts → encoding/csv/mod_test.ts
- encoding/testdata/\*.csv → encoding/csv/testdata/\*.csv

splits logic into single export files.

⚠️ `https://deno.land/std@$STD_VERSION/encoding/csv.ts` is deprecated.


iuioiua:
I don't think the original files with public exports should be deleted. Instead, a deprecation note should be added and the exported functions should come from the new function locations.

This gives devs enough time to migrate.

I did this here for the single-export restructure: https://github.com/denoland/deno_std/blob/7afe49284d0b5357eaf0f5668080cd26f13a1d8b/streams/buffer.ts

iuioiua:
Also, it might be a good idea to also action #2936 at the same time, if possible.

timreichen:
@iuioiua good points. I added the re-exports file and split `parse` and `stringify` into separate files.

encoding/csv/parse.ts
```diff
@@ -0,0 +1,116 @@
...
+  columns?: string[];
+}
+
+/**
+ * Csv parse helper to manipulate data.
+ * Provides an auto/custom mapper for columns.
+ *
+ * @example
+ * ```ts
+ * import { parse } from "https://deno.land/std@$STD_VERSION/encoding/csv/mod.ts";
```
iuioiua, line 37: For imports within examples, it might be better to link to the most direct file. For example:
```ts
import { parse } from "https://deno.land/std@$STD_VERSION/encoding/csv/parse.ts";
```

iuioiua:
Related: #2936.

timreichen:
closing in favour of https://github.com/denoland/deno_std/issues/2936

PR #3034, opened 2022-12-21
test(http): fix test case of http2 stream error

kt3k:
This PR fixes the test case `Server should not close the http2 downstream connection when the response stream throws` in `http/server_test.ts` for canary

After the change https://github.com/denoland/deno/pull/17126, the 3rd fetch call throws immediately (Looks like it's more ideal behavior according to the comment in the test case)


bartlomieju (approved):
LGTM

PR #3037, opened 2022-12-21
fix(node/util): reference error of 'process'

kt3k:
Current implementation of `util.deprecate` depends on global `process` object and it doesn't work with `npm:` feature. (For example, `connect` from `npm:mongoose` doesn't work. See https://github.com/denoland/examples/pull/12 )

This PR fixes the above issue.

As a side effect of the fix, this PR moves the implementation of `deprecate` from `node/internal/utils.mjs` to `node/util.ts`. This was necessary for avoiding circular dependencies. (See also https://github.com/denoland/deno_std/pull/1697#discussion_r767247360


lino-levan:
LGTM

bartlomieju (approved):
LGTM

PR #3040, opened 2022-12-22
fix(flags): parse method looses types in certain cases with collect option

### LINKED ISSUE #3012: bug: std/flags parse loses typing when collect and string are defined
**Describe the bug**

Under some conditions, `parse` loses all of the magic typing which is kind of awful. Specifically, this occurs when `collect` AND `string` (or `boolean`) are defined.

**Steps to Reproduce**

1. Make a file like below
```typescript
const parsed = parse(Deno.args, {
  string: ["header"],
  collect: ["header"],
  alias: {
    h: "header"
  }
})
```
2. Run it with arguments `--header test --header 2` to prove that this is valid and it works
3. See that `parsed` is now of `unknown` type

**Expected behavior**

`parsed` should get the same magic typing it does in other cases in this case.

**Environment**

- OS: MacOS 11
- deno version: 1.29.1
- std version: 0.158.0

c4spar:
close #3012

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -162,7 +162,7 @@ type CollectValues<
       & (Extract<TArgs, TCollactable> extends never ? Record<never, never>
         : RecursiveRequired<
           MapTypes<Extract<TArgs, TCollactable>, Array<TType>, TNegatable>
-      >)
+        >)
     : MapTypes<TArgs, TType, TNegatable>
 >;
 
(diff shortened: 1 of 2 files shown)


lino-levan:
This is some pretty impressive typescript. Thank you for bringing this up

lino-levan:
LGTM

kt3k (approved):
LGTM. Thanks! @c4spar

PR #3041, opened 2022-12-22
chore: update node tests

lino-levan:
This change:
- updates the `deno_graph` dep
- adds proper instructions to `TODO.md` on how to update
- runs `deno task node:setup` again causing the rest of these changes

In https://github.com/denoland/deno_std/pull/2993, `.substr` was replaced by `.slice`, even in the vendored node test code which I think probably wasn't the best idea. Running `node:setup` reverts this change incidentally.

It looks like the changes to `TODO.md` were caused by people updating it by hand instead of changing the config and just running `node:setup` (I am one of those guilty of doing this). I have added a line in `TODO.md` to remedy this.

### DIFF
--- a/node/_tools/test/parallel/test-path-dirname.js
+++ b/node/_tools/test/parallel/test-path-dirname.js
@@ -10,7 +10,7 @@ const common = require('../common');
 const assert = require('assert');
 const path = require('path');
 
-assert.strictEqual(path.dirname(__filename).slice(-13),
+assert.strictEqual(path.dirname(__filename).substr(-13),
                    common.isWindows ? 'test\\parallel' : 'test/parallel');
 
 assert.strictEqual(path.posix.dirname('/a/b/'), '/a');
--- a/node/_tools/TODO.md
+++ b/node/_tools/TODO.md
@@ -1,5 +1,7 @@
 # Remaining Node Tests
 
+NOTE: This file should not be manually edited. Please edit `config.json` and run `deno task node:setup` instead.
+
 Total: 2822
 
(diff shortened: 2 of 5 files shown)


lino-levan:
Now that I think about it, the clarification should probably be done in `node/README.md` and not the main `README.md` of the repo.

iuioiua:
I didn't notice that I'd mistakenly changed Node's vendored tests. A note should be added to the `TODO.md` that the file should not be manually modified.

lino-levan:
Should I add the note in this PR or make another one clarifying it?

iuioiua:
I'm not one to say haha

I'd guess it's fine to include it in this PR. To be clear, I'm referring to modifying the script that generates the `TODO.md` so the notice is part of the generated text.

lino-levan:
> I'm not one to say haha

Fair, I feel like it is in the spirit of this PR so I will quickly update it.

>To be clear, I'm referring to modifying the script that generates the TODO.md so the notice is part of the generated text.

My wording was unclear but this was what I was referring to as well. Good to see we are on the same page.

kt3k (approved):
LGTM. Nice!

node/_tools/test/parallel/test-path-dirname.js
```diff
@@ -10,7 +10,7 @@ const common = require('../common');
 const assert = require('assert');
 const path = require('path');
 
-assert.strictEqual(path.dirname(__filename).slice(-13),
+assert.strictEqual(path.dirname(__filename).substr(-13),
```
kt3k, line 13: I missed this while reviewing #2993

PR #3043, opened 2022-12-22
fix(flags): types for aliases defined as array are ignored

### LINKED ISSUE #3042: flags: types for aliases defined as array are ignored
When aliases are defined as array the are not available in the types.

```ts
const parsed = parse(Deno.args, {
  string: ["foo"],
  alias: {
    foo: ["bar", "baz"],
  },
});
```

Types of `parsed`:
```ts
const parsed: {
    [x: string]: unknown;
    foo: string;
    _: (string | number)[];
}
```

Expected types:

```ts
const parsed: {
    [x: string]: unknown;
    foo: string;
    bar: string;
    baz: string;
    _: (string | number)[];
}
```

I will open a PR for this.

c4spar:
close #3042

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -92,16 +92,18 @@ type Aliases<T = string, V extends string = string> = Partial<
 >;
 
 type AddAliases<
-  T,
-  A extends Aliases | undefined,
-> = { [K in keyof T as AliasName<K, A>]: T[K] };
+  TArgs,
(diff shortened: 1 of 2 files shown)


lino-levan:
LGTM. Also noticed this while messing around with flags but thought it was intentional. Thanks!

kt3k (approved):
LGTM 👍

PR #3044, opened 2022-12-23
refactor(flags): use meaningful names instead of single chars for generic params

c4spar:
This makes the typings more readable and was also requested by @dsherret in this PR https://github.com/denoland/deno_std/pull/2180#discussion_r886863723 but I did not have the time to do this in the original PR.

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -102,10 +102,10 @@ type AddAliases<
   TArgs,
   TAliases extends Aliases | undefined,
 > = {
-  [TArgName in keyof TArgs as AliasName<TArgName, TAliases>]: TArgs[TArgName];
+  [TArgName in keyof TArgs as AliasNames<TArgName, TAliases>]: TArgs[TArgName];
 };
 
-type AliasName<
+type AliasNames<
   TArgName,
   TAliases extends Aliases | undefined,
 > = TArgName extends keyof TAliases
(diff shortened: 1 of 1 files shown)


timreichen:
While I think it is better to move away from single letters, I dont like the how the types looks with prefixes. It is clear that they are types, so instead of a prefix, a more descriptive name should be chosen. It breaks the std guide line of no abbreviations and something we try to avoid in naming conventions (https://deno.land/manual@v1.29.1/references/contributing/style_guide#naming-convention) (technically it isn't the same, but the outcome looks like it).

c4spar:
This is a widely used convention, not only in typescript, also in other languages.

The prefix exists not to make it clear that these are types, it's to avoid naming conflicts with other types and to make it clear it's a **generic** type and only valid in this context, e.g:

```ts
function createFilter<T extends Filter>() {}
```

In this example you cannot name the generic parameter Filter because the filter type already exists and if you prefix the generic with `T` than it's easier to see that this is a generic `Filter` type.

```ts
function createFilter<TFilter extends Filter>() {}
```

And i don't think this is in conflct with our naming convention because there is no clear convention mentioned for generic parameters in our style guide and it's still pascal case.

But if more people prefere to not prefix generics with `T` i can adapt it.

It would propably make sense to add a section for generic parameters to our style guide after we have agreed on a convention.

kt3k:
T prefix for generic types looks fine to me

lino-levan:
T prefix seems fine to me. I don't have a problem with it, but if there are people who strongly care either way then it's worth a discussion + addition to documentation.

kt3k (approved):
LGTM. Thanks!

PR #3046, opened 2022-12-23
refactor: remove unnecessary returns

lino-levan:
Noticed a few of these while looking at codebase, and I thought it could use a refactor.

LMK if there is a reason for these but to me these just look like dubious code.


encoding/testdata/jsonc/test262/assert.js
```diff
@@ -38,7 +38,6 @@ assert.sameValue = function (actual, expected, message) {
     }
   } catch (error) {
     throw new Test262Error(message + ' (_isSameValue operation threw) ' + error);
-    return;
```
kt3k, line 41: This file is vendored from test262 project. Let's avoid changing these

kt3k (approved):
LGTM. Thanks!

PR #3049, opened 2022-12-23
fix: Do not mutate tokens when creating details

kamilogorek:
Input:
```ts
import { assertEquals } from "https://deno.land/std@0.166.0/testing/asserts.ts";

assertEquals(`foo bar`, `fooo baz`);
```

Before:
```diff
-   foo bar
+   fooobaz
```

After:
```diff
-   foo bar
+   fooo baz
```

---

Input:
```ts
import { assertEquals } from "https://deno.land/std@0.166.0/testing/asserts.ts";

assertEquals(`foo bar`, `foo baz`);
```

Before/after:
```diff
-   foo bar
+   foo baz
```

### DIFF
--- a/testing/_diff.ts
+++ b/testing/_diff.ts
@@ -318,8 +318,8 @@ export function diffstr(A: string, B: string) {
       ) {
         return {
           ...result,
-          type: t[i - 1].type
-        }
+          type: t[i - 1].type,
+        };
       }
       return result;
     });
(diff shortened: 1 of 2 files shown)


kamilogorek:
Oh, thanks @kt3k

kt3k (approved):
LGTM. Thanks!

PR #3050, opened 2022-12-23
feat(node/diagnostics_channel): initial implementation

lino-levan:
This PR does the initial implementation of `node/diagnostics_channel`.

The main thing it is missing is the [built-in channels](https://nodejs.org/api/diagnostics_channel.html#built-in-channels). This means the following tests can't be enabled:
- [parallel/test-diagnostics-channel-http-server-start.js](https://github.com/nodejs/node/tree/v18.12.0/test/parallel/test-diagnostics-channel-http-server-start.js)
- [parallel/test-diagnostics-channel-http.js](https://github.com/nodejs/node/tree/v18.12.0/test/parallel/test-diagnostics-channel-http.js)

Additionally, this test also cannot be enabled due to `node/process` not being fully poly-filled:
- [parallel/test-diagnostics-channel-safe-subscriber-errors.js](https://github.com/nodejs/node/tree/v18.12.0/test/parallel/test-diagnostics-channel-safe-subscriber-errors.js)


lino-levan:
I think adding a few of these built-in channels would actually be pretty easy, let me try my hand at that.

lino-levan:
The http channels are kind of a headache so I may revisit it at some later point, I think this PR is ready for review.

kt3k (approved):
LGTM. Nice work!

Issue #1436, opened 2021-10-19
Support missing OS functionality


cknight:
I am looking into the Node polyfill for 'os' ( denoland/deno#3403 ).  The following Node functionality does not seem to be available in Deno:

- [ ] `cpus()` - to get information about the logical CPU cores.  See [node API](https://nodejs.org/api/os.html#os_os_cpus).
- [x] `endianness()` -  the endianness of the CPU for which the Deno binary was compiled.  See [node API](https://nodejs.org/api/os.html#os_os_endianness) https://github.com/denoland/deno/pull/3833
- [ ] `freemem()` - the amount of free system memory.  See [node API](https://nodejs.org/api/os.html#os_os_freemem)
- [ ] `getPriority([pid])` - the scheduling priority for the process specified by pid.  See [node API](https://nodejs.org/api/os.html#os_os_getpriority_pid) https://github.com/denoland/deno/pull/4202
- [x] `loadavg()` - an array containing the 1, 5, and 15 minute load averages.  See [node API](https://nodejs.org/api/os.html#os_os_loadavg)
- [ ] `networkInterfaces()` - Returns an object containing network interfaces that have been assigned a network address.  See [node API](https://nodejs.org/api/os.html#os_os_networkinterfaces)
- [x] `platform()` - a string identifying the operating system platform.  See [node API](https://nodejs.org/api/os.html#os_os_platform)
- [x] `release()` - the operating system as a string. See [node API](https://nodejs.org/api/os.html#os_os_release) https://github.com/denoland/deno/pull/4065
- [ ] `setPriority([pid, ]priority)` - Attempts to set the scheduling priority for the process specified by pid.  See [node API](https://nodejs.org/api/os.html#os_os_setpriority_pid_priority) https://github.com/denoland/deno/pull/4202
- [x] `tmpdir()` - Returns the operating system's default directory for temporary files as a string. See [node API](https://nodejs.org/api/os.html#os_os_tmpdir) https://github.com/denoland/deno/pull/4213
- [ ] `totalmem()` - total amount of system memory in bytes as an integer.  See [node API](https://nodejs.org/api/os.html#os_os_totalmem)
- [ ] `type()` - the operating system name as returned by uname(3). See [node API](https://nodejs.org/api/os.html#os_os_type)
- [ ] `uptime()` - the system uptime in number of seconds.  See [node API](https://nodejs.org/api/os.html#os_os_uptime)
[...]

PR #2839, opened 2022-11-05
refactor(node): use internal namespace for unstable APIs

bartlomieju:
Counterpart to https://github.com/denoland/deno/pull/16520

PR #3052, opened 2022-12-24
feat(node): Add support for os.uptime

kamilogorek:
Dependent on https://github.com/denoland/deno/pull/17179

A part of https://github.com/denoland/deno_std/issues/1436
This will also allow for out-of-the-box support for `npm:@sentry/node`, ref: https://github.com/getsentry/sentry-javascript/issues/3009#issuecomment-1364503874

### DIFF
--- a/node/os.ts
+++ b/node/os.ts
@@ -307,7 +307,7 @@ export function type(): string {
 
 /** Not yet implemented */
 export function uptime(): number {
-  notImplemented(SEE_GITHUB_ISSUE);
+  return Deno.uptime();
 }
 
 /** Not yet implemented */
--- a/node/os.ts
+++ b/node/os.ts
@@ -31,6 +31,9 @@ export const constants = os;
 
 const SEE_GITHUB_ISSUE = "See https://github.com/denoland/deno_std/issues/1436";
 
+// @ts-ignore Deno[Deno.internal] is used on purpose here
+const DenoOsUptime = Deno[Deno.internal]?.nodeUnstable?.osUptime || Deno.osUptime;
+
 interface CPUTimes {
   /** The number of milliseconds the CPU has spent in user mode */
   user: number;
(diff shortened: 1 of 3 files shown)


lino-levan:
LGTM. Nice work!

lino-levan:
@kamilogorek this should probably be un-drafted given that https://github.com/denoland/deno/pull/17179 got merged?

kt3k:
@bartlomieju Do you think it's ok to require `--unstable` for `os.uptime` usage in npm compat? Or should we use `Deno[Deno.internal].nodeUnstable` namespace for bypassing `--unstable` requirement? (like https://github.com/denoland/deno_std/pull/2839)

bartlomieju:
> @bartlomieju Do you think it's ok to require `--unstable` for `os.uptime` usage in npm compat? Or should we use `Deno[Deno.internal].nodeUnstable` namespace for bypassing `--unstable` requirement? (like #2839)

Ah, good point; I forgot about that. Yeah I think it would be preferable to avoid `--unstable` here. Should I open a PR to do that?

kamilogorek:
I believe there's no need to rerun the tests after internal changes, as they were tested in `denoland/deno` itself.

node/os.ts
```diff
@@ -305,9 +305,9 @@ export function type(): string {
   }
 }
 
-/** Not yet implemented */
+/** Returns the Operating System uptime in number of seconds. */
 export function uptime(): number {
-  notImplemented(SEE_GITHUB_ISSUE);
+  return Deno.osUptime();
```
bartlomieju, line 310: Please add 
```
// @ts-ignore Deno[Deno.internal] is used on purpose here
const DenoOsUptime = Deno[Deno.internal]?.nodeUnstable?.osUptime || Deno.osUptime;
```
at the top of the file and use `DenoOsUptime()` here

bartlomieju (changes requested):

bartlomieju (approved):
LGTM, thanks @kamilogorek

Issue #2895, opened 2022-11-16
suggestion: Add supports-color to the standard library


lino-levan:
## Problem
- Currently, it is not trivial to make CLI tools with color in deno (at least using the std), because it is unclear whether or not the terminal has access to color (and if it does, how many of them).
- This is something I've run into [at CookieDB](https://github.com/cookiedb/CookieDB/issues/5), and it's a real struggle. In a true-color terminal running `cookie start` looks like the top demo below. If it's not supported, it looks like the middle demo below.
- The current workaround to make it not completely unusable is either disable color entirely, disable it through a flag, or enable color through a flag (this is done in the bottom demo below).

## Solutions
I see three main possible solutions:

- A: Add a module / new method in `std/fmt` (either in `std/fmt/color` or maybe a new file `std/fmt/supports_color`) that allows for the client to do the check themselves, and handle it however they want
- B: Do the same as above BUT when the client tries to use a color method in a non-color terminal have it do nothing
- C: Do the same thing as the first one BUT the client can configure what the fallback behavior will look like if color is not supported

A very simple demo of the two options in my mind.
```typescript
// Option A
import { rgb24, supportsColor } from "https://deno.land/std/fmt/color.ts";

if(supportsColor.stderr.has16m) {
    console.log(rgb24("Hello, World!", {
        r: 41,
        g: 42,
        b: 43,
    }));
} else {
    console.log("Hello, World!");
}

// Option B
import { colors } from "https://deno.land/std/fmt/color.ts";

console.log(rgb24("Hello, World!", { // behind the scenes, this does nothing if true-color is not supported
    r: 41,
    g: 42,
    b: 43,
 }));
 
// Option C
// ??? - I don't know what this API would look like
```

## Related Images

true-color demo
<img width="454" alt="Screenshot 2022-11-15 at 3 56 43 PM" src="https://user-images.githubusercontent.com/11367844/202049890-9b58bd15-584e-4433-8b2d-d74d0366c9ba.png">

non true-color demo
<img width="469" alt="Screenshot 2022-11-15 at 3 57 48 PM" src="https://user-images.githubusercontent.com/11367844/202049900-cc778b83-193a-4d38-b464-79db1cd5dc0e.png">

workaround demo
<img width="460" alt="Screenshot 2022-11-15 at 3 59 05 PM" src="https://user-images.githubusercontent.com/11367844/202050115-25dbbad4-3905-4e47-b3d3-bb1b51f0066b.png">

[...]

PR #3053, opened 2022-12-25
feat(fmt): supports color module

lino-levan:
This PR is part one of completing https://github.com/denoland/deno_std/issues/2895.

This was heavily inspired `chalk/supports-color` but the api was tweaked a little bit to fit a little bit better into typescript. The tests were also taken from `chalk/supports-color` but also heavily modified.

Part two (which would be in another PR) would be:
> When the client tries to use a color method in a non-color terminal have it do nothing

This is a change that I'm not entirely sold on. I think it would be better for DX but it might be a little too "magic" for a standard library.


lino-levan:
Can't reproc test failures in local env. Will play around with GH actions.

lino-levan:
It turns out that this is just an issue with the tests that `chalk/supports-color` has. They do not do CI testing on windows and the tests thus fail on windows (even if the result of the modules is correct). I will modify the affected tests.

lino-levan:
The current test failures are unrelated to this change. Looks like something in canary broke std.

fmt/supports_color.ts
```diff
@@ -0,0 +1,247 @@
...
+  {
+    streamIsTTY,
+    sniffFlags = true,
+    args = Deno.args,
+    os = Deno.build.os,
+    osRelease = Deno.osRelease(),
+  }: SupportsColorOptions = {},
+) {
+  // Read flags to see if anything matches to force colors or to force no colors
+  const { _, color, colors } = parse(args, {
```
kt3k, line 72: I think we shouldn't automatically read from cli flags

kt3k:
This seems introducing too many env permission prompts to cli tools. Also the dependencies to CI vendor specific env vars look concerning. I feel this feature better belong to 3rd party tools, rather than std.

lino-levan:
SGTM, closing this PR

lino-levan:
For anyone curious about this in the future, I just moved the code to a /x/ module: https://deno.land/x/supports_color

Issue #3023, opened 2022-12-19
std/fs: Use original Error class for `walk` errors


binyamin:
Related to https://github.com/denoland/deno_std/issues/1310

### Summary

Whenever the "walk" function encounters an error, it transforms it using the code below. In the process, it uses a generic `Error` class as the constructor. This means that there's no easy way to identify the type of error.

https://github.com/denoland/deno_std/blob/7afe49284d0b5357eaf0f5668080cd26f13a1d8b/fs/walk.ts#L32-L42

### Proposed Solution
Use `new err.constructor()` instead of `new Error()`, when instantiating the new error class.

lino-levan:
Could you elaborate on your use case here, I'm curious. In what situation would the error message not be good enough?

Not against this change, I think it makes sense(?) but I would appreciate further elaboration.

binyamin:
@lino-levan Sure.

First of all, a CLI tool might want custom error messages. It also might want to add additional context to the error. For example, the filepath, or the name of a plugin.

Secondly, if the walk function receives a path which doesn't exist, it throws. But If I check first that the path exists, I seem to be [introducing a race condition](https://github.com/denoland/deno_std/blob/main/fs/exists.ts). But I can't handle the error easily, because it's transformed to look like a generic error. In other cases, I would use `instanceof` or `error.name`.

The only way I've found to check if it's a "Not found" error, is to parse the stack-trace. Which is not ideal, and could get complicated.

lino-levan:
@iuioiua thoughts on this proposal? I would borderline push for a `WalkError` class with an `err.cause` defined as the error?

lino-levan:
After some further thinking, I think I like the idea of a `WalkError` with a cause defined as the actual error. I will open a PR for this to further the discussion.

kt3k:
@binyamin Does the suggested solution by @lino-levan support your use case? You can use the proposed fix from `https://raw.githubusercontent.com/lino-levan/deno_std/fix-wrap-error/fs/walk.ts`

```ts
import { walk } from "https://raw.githubusercontent.com/lino-levan/deno_std/fix-wrap-error/fs/walk.ts";
```

binyamin:
@kt3k Yes. It would be nice if `WalkError` had a path property, reused from the constructor arguments. Just an idea.

PR #3054, opened 2022-12-25
feat(fs/walk): WalkError class

lino-levan:
closes https://github.com/denoland/deno_std/issues/3023.

The concept is that if you want to catch an error and see what caused it, you should be able to use `err.cause instanceof ErrorClassYouWantToCatch`.


fs/walk.ts
```diff
@@ -11,6 +11,14 @@ import {
   WalkEntry,
 } from "./_util.ts";
 
+export class WalkError extends Error {
+  constructor(err: unknown, path: string) {
+    super(`${err instanceof Error ? err.message : err} for path "${path}"`);
+    this.name = "RetryError";
```
kt3k, line 17: typo?
```suggestion
    this.name = "WalkError";
```

fs/walk.ts
```diff
@@ -31,14 +39,7 @@ function include(
 
 function wrapErrorWithRootPath(err: unknown, root: string) {
   if (err instanceof Error && "root" in err) return err;
-  const e = new Error() as Error & { root: string };
-  e.root = root;
```
kt3k, line 35: Can we remove `root` property from the thrown error?

fs/walk.ts
lino-levan: Yup

fs/walk.ts
lino-levan: We can? It's just not optimal in my opinion. Are we renaming it to `path` instead or completely removing it?

kt3k: `path` sounds good to me

kt3k:
@lino-levan Can you also add a simple test case which checks WalkError being thrown?

lino-levan:
SGTM, I'll get around to it when I get around to it. Maybe tomorrow if I get the chance.

lino-levan:
@kt3k I tried my hand at writing a test for this but can't think of a way for this to error out consistently (without relying on something like removing a folder after we already started walking it). Do you have any ideas?

fs/walk.ts
```diff
@@ -11,6 +11,16 @@ import {
...
 } from "./_util.ts";
 
+export class WalkError extends Error {
+  path: string;
+  constructor(err: unknown, path: string) {
+    super(`${err instanceof Error ? err.message : err} for path "${path}"`);
+    this.name = "WalkError";
+    this.path = path;
+    this.cause = err;
+  }
```
jsejcksn, line 21:
```suggestion
  override cause: unknown;
  override name = "WalkError";
  path: string;

  constructor(cause: unknown, path: string) {
    super(
      `${cause instanceof Error ? cause.message : cause} for path "${path}"`,
    );
    this.path = path;
    this.cause = cause;
  }
```

See usage info for `override` in the [TS handbook](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-3.html#override-and-the---noimplicitoverride-flag), and other usage examples in this repo — for example, in `AssertionError` at [`std/testing/asserts.ts#L20`](https://github.com/denoland/deno_std/blob/0.178.0/testing/asserts.ts#L20).

This also improves typing for this error subclass by removing the [optional modifier](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties) from the `cause` property — because it is guaranteed to be set to some (`unknown`) value.

lino-levan: Thanks for the review, I think this makes sense. I don't think this is mergable because I don't know if this is testable. If you could come up with a test case to test this code then I would love to move this PR forward.

jsejcksn: @lino-levan I don't have write permission for this PR to update the test file (I'm not an org member, nor am I a collaborator on your branch) — so I can't contribute a commit to revise it. FYI: There's a types assertion module in this repo at [`std/testing/types.ts`](https://github.com/denoland/deno_std/blob/main/testing/types.ts).

lino-levan: Ah sorry, I wasn't being super clear. This PR in general (outside of this change) is not landable because I can't write tests for it. As far as I can tell, this error class only gets used when some race conditions occur (and those can't be reproc'd). I would change this PR to a draft but I can't seem to figure out how to do that.

jsejcksn: Understood — but what's holding you back? Your [initial idea](https://github.com/denoland/deno_std/pull/3054#issuecomment-1405840646) seemed like a contrived, but perfectly fine approach for generating an exception in a test.

lino-levan: Fair enough. I'll look into it this weekend. This change looks good, I'll commit it. Thanks again for the review!

jsejcksn: > I tried my hand at writing a test for this but can't think of a way for this to error out consistently (without relying on something like removing a folder after we already started walking it). Do you have any ideas?

@lino-levan I just tested this locally and it reliably throws every run:

`test.ts`:
```ts
// deno test --allow-read=exampleRootTestDir --allow-write=exampleRootTestDir

import { assertRejects } from "https://deno.land/std@0.178.0/testing/asserts.ts";
import { walk } from "https://deno.land/std@0.178.0/fs/walk.ts";

Deno.test("throws", async () => {
  const dir = "exampleRootTestDir";

  await assertRejects(async () => {
    await Deno.mkdir(dir);
    for await (const _walkEntry of walk(dir)) await Deno.remove(dir);
  });
});

```

```
 % deno --version
deno 1.31.1 (release, aarch64-apple-darwin)
v8 11.0.226.13
typescript 4.9.4

% deno test --allow-read=exampleRootTestDir --allow-write=exampleRootTestDir
running 1 test from ./test.ts
throws ... ok (6ms)

ok | 1 passed | 0 failed (20ms)
```

Check out the [`testWalk`](https://github.com/denoland/deno_std/blob/0.178.0/fs/walk_test.ts#L11) function in the test file.

jsejcksn:
> I tried my hand at writing a test for this but can't think of a way for this to error out consistently (without relying on something like removing a folder after we already started walking it). Do you have any ideas?

@lino-levan I just tested this locally and it reliably throws every run:

`test.ts`:
```ts
// deno test --allow-read=exampleRootTestDir --allow-write=exampleRootTestDir

import { assertRejects } from "https://deno.land/std@0.178.0/testing/asserts.ts";
import { walk } from "https://deno.land/std@0.178.0/fs/walk.ts";

Deno.test("throws", async () => {
  const dir = "exampleRootTestDir";

  await assertRejects(async () => {
    await Deno.mkdir(dir);
    for await (const _walkEntry of walk(dir)) await Deno.remove(dir);
  });
});

```

```
 % deno --version
deno 1.31.1 (release, aarch64-apple-darwin)
v8 11.0.226.13
typescript 4.9.4

% deno test --allow-read=exampleRootTestDir --allow-write=exampleRootTestDir
running 1 test from ./test.ts
throws ... ok (6ms)

ok | 1 passed | 0 failed (20ms)
```

Check out the [`testWalk`](https://github.com/denoland/deno_std/blob/0.178.0/fs/walk_test.ts#L11) function in the test file.

lino-levan:
LGTM, thanks for your help on this PR!

kt3k (approved):
LGTM. Thanks! @lino-levan

PR #3056, opened 2022-12-26
refactor(io): rework single-export files

iuioiua:
This change simplifies how functions and types are re-exported while maintaining deprecation warnings introduced in #2975. This also fixes an issue blocking https://github.com/denoland/deno/pull/16881.

CC @bartlomieju

### DIFF
--- a/io/buffer.ts
+++ b/io/buffer.ts
@@ -2,18 +2,6 @@
 import { assert } from "../_util/asserts.ts";
 import { copy } from "../bytes/copy.ts";
 import type { Reader, ReaderSync } from "../types.d.ts";
-import {
-  BufferFullError as _BufferFullError,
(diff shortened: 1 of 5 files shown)


iuioiua:
Note to self: this needs to be done, as a proactive measure, for the `archive` and `streams` modules.

kt3k (approved):
LGTM.

Let's continue on denoland/deno#16881 with this change

lino-levan:
@iuioiua Congrats on your 100th commit to deno_std 🎉🎉🎉

iuioiua:
> @iuioiua Congrats on your 100th commit to deno_std 🎉🎉🎉

Wow! Didn't notice 🤣

🎉🍾🕺🏾

PR #3059, opened 2022-12-26
chore(node): update version to 18.12.1

### LINKED ISSUE #3057: node: process.version gives v16.17.0 instead of v18.x
We are targeting LTS release of Node for our compat layer. We did update to v18 in https://github.com/denoland/deno_std/pull/2814, but the version update was skipped. We should change the value of `process.version` to match what we are testing against.

(`process.versions` should be updated as well)

lino-levan:
These changes were obtained by running in node 18.12.0

```javascript
const { versions } = require('node:process');

console.log(versions)
```

Closes #3057


lino-levan:
Now that I think about it, do we want to be targeting 18.12.1? The OpenSSL security issue is patched in it.

node/_process/process.ts
```diff
@@ -100,7 +100,7 @@ export const platform = isWindows ? "win32" : Deno.build.os;
  * it pointed to Deno version, but that led to incompability
  * with some packages.
  */
-export const version = "v16.17.0";
+export const version = "v18.12.1";
```
kt3k, line 103: 👍

kt3k:
The update to v18.12.1 also makes sense to me.

kt3k (approved):
LGTM

PR #3060, opened 2022-12-26
refactor(archive): rework single-export files

iuioiua:
This change simplifies how functions and types are re-exported while maintaining deprecation warnings introduced in #2958. It also cleans up a couple of little remaining bits for single-export files in `io`.


kt3k (approved):
LGTM. Thanks!

PR #3061, opened 2022-12-26
refactor(streams): rework single-export files

iuioiua:
This change simplifies how functions and types are re-exported while maintaining deprecation warnings introduced in #2957.


kt3k (approved):
LGTM, Thanks!

PR #3062, opened 2022-12-27
fix(node/http): avoid empty chunk issue of flash

kt3k:
This solves a part of astro issues https://github.com/denoland/deno/issues/16659

Currently flash seems having an issue in handling empty chunks in ReadableStream response body (https://github.com/denoland/deno/issues/17194). This PR tries to avoid that issue by skipping enqueuing empty chunks.


bartlomieju (approved):
LGTM. Tried it locally and it actually fixes a problem with Vite as well.

lino-levan:
LGTM

PR #3067, opened 2022-12-28
fix(node/http): ignore body when status code is one of 101, 204, 205, 304

kt3k:
Currently the http server of `std/node` can't respond when the status code is one of 101, 204, 205, 304 and response body is specified. `Response` constructor throws in those cases.

(This problem seems causing the issue https://github.com/denoland/deno/issues/17142 )

This PR ignores specified body and makes body null when the status code is one of those.

closes https://github.com/denoland/deno/issues/17142

### DIFF
--- a/node/http.ts
+++ b/node/http.ts
@@ -299,6 +299,12 @@ export class ServerResponse extends NodeWritable {
     }
   }
 
+  /** Returns true if the response body should be null with the given
+   * http status code */
+  static #bodyShouldBeNull(status: number) {
+    return status === 101 || status === 204 || status === 205 || status === 304;
+  }
+
   constructor(
     reqEvent: undefined | Deno.RequestEvent,
     resolve: undefined | ((value: Response | PromiseLike<Response>) => void),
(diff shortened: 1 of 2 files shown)


node/http.ts
```diff
@@ -299,6 +299,12 @@ export class ServerResponse extends NodeWritable {
     }
   }
 
+  /** Returns true if the response body should be null with the given
+   * http status code */
+  static #bodyShouldBeNull(status: number) {
+    return status === 101 || status === 204 || status === 205 || status === 304;
```
kt3k, line 305: This condition matches with https://github.com/denoland/deno/blob/8bdf66c59c7424759f441e19047d7ffc0bf13ef3/ext/fetch/23_response.js#L82

bartlomieju (approved):
LGTM, before merging please ensure that this PR doesn't regress `express` performance.

EDIT: Verified myself that it doesn't regress performance.

kt3k:
@bartlomieju Thanks for verification!

Issue #3063, opened 2022-12-28
basename cant work correctly


familyboat:
**Describe the bug**
the result of posix.basename('aaaa/bbb////', 'a/bbb') is 'bbb////'. the trailing directory separator should be removed.

**Steps to Reproduce**
```js
import { posix } from "https://deno.land/std@0.170.0/path/mod.ts";

console.log(posix.basename('aaaa/bbb////', 'a/bbb')) // bbb////
```
**Expected behavior**

the trailing directory separator should be removed.

**Environment**

- OS: [MSYS_NT-10.0-19044]
- deno version:1.28.3
- std version: 0.170.0

kamilogorek:
Notes:
- it works fine without suffix from 2nd argument
- it's also broken in the exact same way in the latest node `path.basename`
- unix `basename` works correctly

kt3k:
https://github.com/denoland/deno_std/pull/3068#issuecomment-1367869543

familyboat:
I want to explain why i issued this bug. But my English is poor. Because i wanted to use   `basename` in my project, so i checked the relevant code, and found https://github.com/denoland/deno_std/blob/1aa78c207e2aa492916a4d97dba2f74a3a52c959/path/posix.ts#L302 would cause buggy behavior in some case. So i issued this bug.
I had a solution for this bug, but i was not confident about my solution. Maybe it's a ugly solution.
I found this pr https://github.com/denoland/deno_std/pull/3068 dont correct this bug completely. Then i searched the manual page of the unix `basename`, [unix basename manual page](https://man7.org/linux/man-pages/man1/basename.1.html), and executed this command in shell, `basename /// bbb`, the result is `/`. But i thought the result would be `''`. 
So this is my thought on this bug, maybe i could make a pr😄.

kamilogorek:
This issue (trimming trailing directories) will be impossible to fix without either:

a) changing how `parse` works - always trim trailing slashes, just as `basename` and `dirname` works in `coreutils`, this makes it impossible to reassure that `path.format(path.parse(some)) === some`
b) making `parse` and `basename` not symmetrical, where `parse(some).base !== path.basename(some)` and `parse(some).dir !== path.dir(some)`

Reasoning:
```js
path.parse("/home/user/somedir//file.zip")
// {
//   root: '',
//   dir: '"/home/user/somedir/',
//   base: 'file.zip"',
//   ext: '.zip"',
//   name: 'file'
// }
```

```shell
[...]

PR #3068, opened 2022-12-28
fix(path): Correctly handle trailing slashes for posix basename

kamilogorek:
The only "real" change here is marking `end = firstNonSlashEnd` when we first encounter the non-slash character. This way, if we don't match the whole suffix (which happens when using input like `a/bb`, as it contains a slash, and can never be matched), it will still trim all trailing slashes.

Fixes https://github.com/denoland/deno_std/issues/3063

### DIFF
--- a/path/basename_test.ts
+++ b/path/basename_test.ts
@@ -22,6 +22,7 @@ Deno.test("basename", function () {
   assertEquals(path.basename("/aaa/bbb", "a/bbb"), "bbb");
   assertEquals(path.basename("/aaa/bbb", "bbb"), "bbb");
   assertEquals(path.basename("/aaa/bbb//", "bbb"), "bbb");
+  assertEquals(path.basename("/aaa/bbb//", "a/bbb"), "bbb");
   assertEquals(path.basename("/aaa/bbb", "bb"), "b");
   assertEquals(path.basename("/aaa/bbb", "b"), "bb");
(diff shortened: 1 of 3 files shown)


kamilogorek:
Tbh I'd happily refactor this out into a common function. Other than `isPathSeparator` vs `isPosixPathSeparator` (which is already a part of `isPathSeparator` internally anyway) and  this piece:

```js
  // Check for a drive letter prefix so as not to mistake the following
  // path separator as an extra separator at the end of the path that can be
  // disregarded
  if (path.length >= 2) {
    const drive = path.charCodeAt(0);
    if (isWindowsDeviceRoot(drive)) {
      if (path.charCodeAt(1) === CHAR_COLON) start = 2;
    }
  }
```

those are exactly the same functions, 1:1, so we effectively end up with 78 redundant LOC.

kt3k (approved):
LGTM. Nice fix!

> those are exactly the same functions, 1:1, so we effectively end up with 78 redundant LOC.

That's also very welcome!

familyboat:
sorry to say you dont correct this bug completely. if no part of the path is legal character, eg: '/////', in this case, the result is '/////'. i think https://github.com/denoland/deno_std/blob/1aa78c207e2aa492916a4d97dba2f74a3a52c959/path/posix.ts#L302 is the source of wrong behavior.

PR #3070, opened 2022-12-30
chore(node): enable some tests

lino-levan:
Just trying to bring the number down a little bit.

I've been doing this by hand but I'll probably write an automated tool at some point to just try every test individually.


iuioiua:
> I've been doing this by hand but I'll probably write an automated tool at some point to just try every test individually.

Is there `deno test --json`?

lino-levan:
> Is there deno test --json?

There is not

```
$ deno test --json
error: Found argument '--json' which wasn't expected, or isn't valid in this context

        If you tried to supply `--json` as a value rather than a flag, use `-- --json`
```

kt3k (approved):
LGTM. Thanks!

PR #3072, opened 2022-12-30
feat(collections): refactored binary search tree and red black tree to remove by node

### LINKED ISSUE #3055: Add getRoot and removeTreeNode methods to RedBlackTree and BinarySearchTree.

A node can be removed from a binary tree with O(1) time complexity using the removeTreeNode method.

Add getRoot and removeTreeNode methods to RedBlackTree and BinarySearchTree.

https://deno.land/std@0.170.0/collections/binary_search_tree.ts?source

https://deno.land/std@0.170.0/collections/red_black_tree.ts?source

```ts
  getRoot() {
        return this.root;
    }
```

https://github.com/masx200/leetcode-test/blob/b52ad069dbeb43f8f155ccbc3687c87a631bfdbe/dinner-plate-stacks/RedBlackTree.ts#L8

https://www.npmjs.com/package/@datastructures-js/binary-search-tree#root

pting-me:
Fixes #3055

Changes `removeNode` to take in a node as a parameter, without exposing the functions as public. This should allow for extending these functions for further use.

@KyleJune These are more or less what you've suggested so would appreciate your thoughts.

### DIFF
--- a/collections/binary_search_tree.ts
+++ b/collections/binary_search_tree.ts
@@ -235,29 +235,34 @@ export class BinarySearchTree<T> implements Iterable<T> {
     return null;
   }
 
+  /** Removes the given node, and returns the node that was physically removed from the tree. */
   protected removeNode(
     node: BinarySearchNode<T>,
   ): BinarySearchNode<T> | null {
-    if (node) {
-      const successorNode: BinarySearchNode<T> | null =
-        !node.left || !node.right ? node : node.findSuccessorNode()!;
-      const replacementNode: BinarySearchNode<T> | null = successorNode.left ??
-        successorNode.right;
-      if (replacementNode) replacementNode.parent = successorNode.parent;
+    /**
+     * The node to physically remove from the tree.
+     * Guaranteed to not have both left and right children.
+     */
+    const flaggedNode: BinarySearchNode<T> | null = !node.left || !node.right
+      ? node
+      : node.findSuccessorNode()!;
+    /** Replaces the flagged node. */
+    const replacementNode: BinarySearchNode<T> | null = flaggedNode.left ??
+      flaggedNode.right;
 
-      if (!successorNode.parent) {
-        this.root = replacementNode;
-      } else {
-        successorNode.parent[successorNode.directionFromParent()!] =
-          replacementNode;
-      }
-
-      if (successorNode !== node) {
-        node.value = successorNode.value;
-        node = successorNode;
-      }
-      this._size--;
+    if (replacementNode) replacementNode.parent = flaggedNode.parent;
+    if (!flaggedNode.parent) {
+      this.root = replacementNode;
+    } else {
+      flaggedNode.parent[flaggedNode.directionFromParent()!] = replacementNode;
     }
+    if (flaggedNode !== node) {
+      /** Swaps values, in case value of the removed node is still needed by consumer. */
+      [node.value, flaggedNode.value] = [flaggedNode.value, node.value];
+      node = flaggedNode;
+    }
+
+    this._size--;
     return node;
   }
 
@@ -275,7 +280,8 @@ export class BinarySearchTree<T> implements Iterable<T> {
    */
   remove(value: T): boolean {
     const node: BinarySearchNode<T> | null = this.findNode(value);
-    return !!(node && this.removeNode(node));
+    if (node) this.removeNode(node);
+    return node !== null;
   }
 
   /** Returns node value if found in the binary search tree. */
--- a/collections/binary_search_tree.ts
+++ b/collections/binary_search_tree.ts
@@ -261,12 +261,10 @@ export class BinarySearchTree<T> implements Iterable<T> {
       const swapValue = node.value;
       node.value = flaggedNode.value;
       flaggedNode.value = swapValue;
-
-      node = flaggedNode;
     }
 
     this._size--;
-    return node;
+    return flaggedNode;
   }
 
   /**
--- a/collections/binary_search_tree.ts
+++ b/collections/binary_search_tree.ts
@@ -236,14 +236,12 @@ export class BinarySearchTree<T> implements Iterable<T> {
   }
 
   protected removeNode(
-    value: T,
+    node: BinarySearchNode<T>,
   ): BinarySearchNode<T> | null {
-    let removeNode: BinarySearchNode<T> | null = this.findNode(value);
-    if (removeNode) {
+    // let node: BinarySearchNode<T> | null = this.findNode(value);
+    if (node) {
       const successorNode: BinarySearchNode<T> | null =
-        !removeNode.left || !removeNode.right
-          ? removeNode
-          : removeNode.findSuccessorNode()!;
+        !node.left || !node.right ? node : node.findSuccessorNode()!;
       const replacementNode: BinarySearchNode<T> | null = successorNode.left ??
         successorNode.right;
       if (replacementNode) replacementNode.parent = successorNode.parent;
@@ -255,13 +253,13 @@ export class BinarySearchTree<T> implements Iterable<T> {
           replacementNode;
       }
 
-      if (successorNode !== removeNode) {
-        removeNode.value = successorNode.value;
-        removeNode = successorNode;
+      if (successorNode !== node) {
+        node.value = successorNode.value;
+        node = successorNode;
       }
       this._size--;
     }
-    return removeNode;
+    return node;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


collections/binary_search_tree.ts
```diff
@@ -277,7 +274,8 @@ export class BinarySearchTree<T> implements Iterable<T> {
    * Returns true if found and removed.
    */
   remove(value: T): boolean {
-    return !!this.removeNode(value);
+    const node: BinarySearchNode<T> | null = this.findNode(value);
+    return !!(node && this.removeNode(node));
```
KyleJune, line 278: I know I suggested this before, but lets assume that if a node is found, it will be removed.

```ts
const node: BinarySearchNode<T> | null = this.findNode(value);
if (node) this.removeNode(node);
return !!node;
```

collections/binary_search_tree.ts
```diff
@@ -236,14 +236,11 @@ export class BinarySearchTree<T> implements Iterable<T> {
   }
 
   protected removeNode(
```
KyleJune, line 238: It looks like currently this function previously either returned the removeNode or the successorNode if it found the node. The reason was so that the caller would know whether or not a node was found/removed. Now that we always pass in a node that was found, I think this function can be changed to instead only return the successorNode.

The remove function on the RedBlackTree doesn't need to get the removedNode, it only needs the successorNode so that it can rebalance the tree.

collections/binary_search_tree.ts
```diff
@@ -255,13 +252,13 @@ export class BinarySearchTree<T> implements Iterable<T> {
           replacementNode;
       }
 
-      if (successorNode !== removeNode) {
-        removeNode.value = successorNode.value;
-        removeNode = successorNode;
+      if (successorNode !== node) {
+        node.value = successorNode.value;
```
KyleJune, line 256: I don't recall why I had this one line. The value on the node being removed shouldn't matter. This if condition could be completely removed and we can change this function to only return the successorNode if it's different than the node.

collections/binary_search_tree.ts
```diff
@@ -255,13 +252,13 @@ export class BinarySearchTree<T> implements Iterable<T> {
...
+        node = successorNode;
       }
       this._size--;
     }
-    return removeNode;
+    return node;
```
KyleJune, line 261: From my previous comment, I would change this to `return successorNode !== node ? successorNode : null;`

I'd also suggest adding a documentation comment for this function indicating it will remove the node from the tree and return the successor node if there is one.

collections/red_black_tree.ts
```diff
@@ -254,7 +254,17 @@ export class RedBlackTree<T> extends BinarySearchTree<T> {
...
    */
   override remove(value: T): boolean {
-    const node = this.removeNode(value) as (RedBlackNode<T> | null);
+    const nodeToRemove = this.findNode(value) as (RedBlackNode<T> | null);
+
+    if (!nodeToRemove) {
+      return false;
+    }
+
+    const node = this.removeNode(nodeToRemove) as (
```
KyleJune, line 263: From another comment, I mentioned changing removeNode to only return a successorNode if there is one. The following change to this function should get it so that it will still return true/false based on if the node was found, then it will only do the fixup if there is a successorNode that requires it.

```js
override remove(value: T): boolean {
  const node = this.findNode(value) as (RedBlackNode<T> | null);
  const successorNode = node && this.removeNode(node) as (RedBlackNode<T> | null);
  if (successorNode && !successorNode.red) {
    this.removeFixup(successorNode.parent, successorNode.left ?? successorNode.right);
  }
  return !!node;
}
```

KyleJune (changes requested):

collections/binary_search_tree.ts
timreichen: Maybe `node != null` instead of `!!node`

collections/binary_search_tree.ts
pting-me: Half the tests fail when I try deleting this. Doing some digging to further understand why it's needed. Will include a comment.

collections/red_black_tree.ts
```diff
@@ -254,7 +254,17 @@ export class RedBlackTree<T> extends BinarySearchTree<T> {
    * Returns true if found and removed.
    */
   override remove(value: T): boolean {
-    const node = this.removeNode(value) as (RedBlackNode<T> | null);
+    const nodeToRemove = this.findNode(value) as (RedBlackNode<T> | null);
+
+    if (!nodeToRemove) {
+      return false;
+    }
```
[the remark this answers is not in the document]
pting-me: Any particular reason? I'm seeing a lot of early returns throughout `deno_std`, and it's not really mentioned in the style guide. Personally find it easier to narrow types and also quickly glance to ensure unwanted code will not be run.

collections/binary_search_tree.ts
pting-me: Think I figured it out.

Screen Shot 2022-12-31 at 10 49 21 AM

`successorNode` is actually the node we are physically removing from the tree, and not `node`. `replacementNode` is the one that will take the old place of `successorNode`. So this line is correct.

I tried recreating the algorithm on my own, and one particular piece I was trying to figure out was how we were guaranteed to delete without creating any orphan nodes. In this implementation, `successorNode` is guaranteed to only have one child, because that's the nature of `findSuccessorNode`. I think that's worth writing in the comments above.

Lastly, looks like `node = successorNode` meant that we weren't returning the successorNode. We were returning the node that was removed. I think we should probably keep it that way and preserve the logic of `removeFixup`, but should definitely be documented.

collections/binary_search_tree.ts
pting-me: Added a comment to reflect my findings. Returned node should be the node physically removed from the tree, and not the successor node. It seems like `RedBlackTree` does need the removed node.

collections/red_black_tree.ts
pting-me: It seems like `RedBlackTree` really needs the removed node, and we can't just give it a `successorNode`. I'll preserve the logic as it was previously written. It also conceptually makes more sense to me that the `remove` function would return the node that was removed.

collections/binary_search_tree.ts
```diff
@@ -235,33 +235,35 @@ export class BinarySearchTree<T> implements Iterable<T> {
...
-          ? removeNode
-          : removeNode.findSuccessorNode()!;
-      const replacementNode: BinarySearchNode<T> | null = successorNode.left ??
-        successorNode.right;
-      if (replacementNode) replacementNode.parent = successorNode.parent;
+    /**
+     * The node to physically remove from the tree.
+     * Guaranteed to not have both left and right children.
+     */
+    const flaggedNode: BinarySearchNode<T> | null = !node.left || !node.right
```
pting-me, line 246: Changed the naming here because the node that we are removing is not always the "successor" node, and I feel the naming here has caused confusion.

The node flagged for removal should only be the successor node when the node we're trying to remove has two children. Otherwise, we should be removing the provided node.

collections/binary_search_tree.ts
```diff
@@ -235,33 +235,36 @@ export class BinarySearchTree<T> implements Iterable<T> {
...
-      }
-      this._size--;
+    if (replacementNode) replacementNode.parent = flaggedNode.parent;
+    if (!flaggedNode.parent) {
+      this.root = replacementNode;
+    } else {
+      flaggedNode.parent[flaggedNode.directionFromParent()!] = replacementNode;
+    }
+    if (flaggedNode !== node) {
+      /** Swaps values, in case value of the removed node is still needed by consumer. */
```
pting-me, line 260: Technically `RedBlackTree` doesn't need the value, but included it for completeness sake.

Originally wrote this using a tuple swap, but this is the most performant.

lino-levan:
This PR is in the red, please run `deno fmt`. I'll take some time to review it later today.

pting-me:
@lino-levan Just realized it's failing because of the license headers. I'll rebase after those are in and request new build.

collections/red_black_tree.ts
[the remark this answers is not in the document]
lino-levan: I believe early returns are fine. This case LGTM

lino-levan:
LGTM. Nice work!

collections/red_black_tree.ts
lino-levan: +1 for remove returning the node that was removed and not the successorNode

lino-levan (approved):

kt3k (approved):
LGTM. Thanks for the enhancement and clean up. @pting-me 

Also thanks for your reviews! @KyleJune @lino-levan

PR #3073, opened 2022-12-30
chore(node/_http_outgoing): cleaned up OutgoingMessage to use class

lino-levan:
I'm on a quest to fix node compat's `http.ClientRequest`. This PR is step zero in this process.

`ClientRequest` currently extends `NodeWritable` when it should really be extending `OutgoingMessage`.

This PR makes `OutgoingMessage` a class (without changing any other semantics about it)

### DIFF
--- a/node/_http_outgoing.ts
+++ b/node/_http_outgoing.ts
@@ -675,11 +675,11 @@ export class OutgoingMessage extends Stream {
 
   end(
     // deno-lint-ignore no-explicit-any
-    chunk: any,
+    chunk?: any,
     // deno-lint-ignore no-explicit-any
-    encoding: any,
+    encoding?: any,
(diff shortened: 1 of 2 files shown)


lino-levan:
It looks like for some reason extending `OutgoingMessage` makes some tests hang permanently. I'll look into this later, it isn't necessary for this PR anyways.

kt3k:
I don't think this is a good idea as node.js implement OutgoingMessage as `function` https://github.com/nodejs/node/blob/15d673dfbe02134994f63c9f4bbd20ce697ed495/lib/_http_outgoing.js#L103

We often do the opposite (class -> function) change to make `std/node` more aligned to node. e.g. https://github.com/denoland/deno_std/pull/2634

lino-levan:
Closing this then. Thanks for the review!

PR #3077, opened 2023-01-02
fix(encoding/jsonc): avoid prototype pollution in Node.js and Browser

ayame113:
related: [CVE-2022-46175](https://github.com/advisories/GHSA-9c47-m6qq-7p4h)

Using `encoding/jsonc` with Node.js and browsers is currently vulnerable to prototype pollution (in the form described in the [CVE-2022-46175](https://github.com/advisories/GHSA-9c47-m6qq-7p4h)).

This PR changes the behavior of `jsonc.parse` when trying to set `__proto__` as a key in Node.js and browsers.

This PR does not change behavior when running in Deno. This is because `__proto__` has been removed within Deno and is already robust against prototype pollution attacks.


encoding/jsonc_test.ts
```diff
@@ -171,3 +173,51 @@ Deno.test({
...
+
+      // 2. create tmp file
+      await Deno.writeTextFile(tempTile, code);
+
+      // 3. run in Node.js
+      const command = new Deno.Command("node", {
+        stdout: "inherit",
+        stderr: "inherit",
+        args: [tempTile],
+      });
```
ayame113, line 214: I added test in Node.js environment. However, if you don't want to have a dependency on Node.js, I can remove this test.

lino-levan:
Not sure if this makes sense for the std. It seems to be making the code harder to read with very little gain for its intended place of consumption. I do see an argument for browsers since iirc this module is considered browser compatible but given that it is non-trivial to get this running directly in a browser due to typescript, I would not be concerned about `__proto__` vulnerabilities personally. I'm not super strongly against this change, just my 2 cents.

Also, I'm personally -1 for adding a dependency to node here.

ayame113:
We can use [esb.deno.dev](https://esb.deno.dev), [deno.land/x/emit](https://deno.land/x/emit) and [dnt](https://deno.land/x/dnt) to run this module in the browser or Node.js. When comparing code readability and security, security should take precedence.

In my opinion, when it comes to code readability and security, security should take precedence.

> Also, I'm personally -1 for adding a dependency to node here.

I thought about this again, by overriding the setter of `__proto__` and testing it, it may be possible to test without relying on Node.js. I'll try later.

encoding/jsonc_test.ts
ayame113: I have updated the code so there is no longer a dependency on Node.js.

encoding/jsonc_test.ts
```diff
@@ -171,3 +172,47 @@ Deno.test({
     assertValidParse('  ["foo\\\\\\\\", "bar"]', ["foo\\\\", "bar"]);
   },
 });
+
+Deno.test({
+  name: "[jsonc] avoid prototype pollution",
```
kt3k, line 177: Is it meaningful to run this test case in Deno?

kt3k:
Fixing for browser security makes sense to me

encoding/jsonc_test.ts
ayame113: This was just a test case reinforcementation. (This is the code I used to check the current Deno behavior.)

I've left this code in case someone wants to see how Deno works now, but it's not that important. Should it be deleted?

kt3k: I think the below one (`[jsonc] use Object.defineProperty when setting object property`) is the only test necessary for this fix as this test case already passes without this PR's change (It checks Deno cli behavior instead of `std/encoding/jsonc` behavior in my view)

ayame113: It makes sense. I removed this test.

ayame113: Thank you for your review!

lino-levan:
Sorry about not having time to review this earlier. In its new form, this LGTM!

My biggest concern before was regarding CI but now that this has been fixed, this seems like an obvious fix.

kt3k (approved):
LGTM

PR #3083, opened 2023-01-02
chore: make license checker work on windows

ayame113:
Note: I created #3082 using the code I fixed in this branch.


_tools/check_licence.ts
```diff
@@ -12,7 +13,7 @@ const EXCLUDED_DIRS = [
   "node/testdata",
   "crypto/_wasm/target",
   "encoding/varint/_wasm/target",
-];
+].map((path) => path.replaceAll("/", isWindows ? "\\\\" : "/"));
```
schwarzkopfb, line 16: what about

```ts
import { SEP } from "../path/separator.ts";

[ /* paths */ ].map((path) => path.replaceAll("/", SEP));
```

or even
```ts
import { join } from "../path/mod.ts";

const EXCLUDED_DIRS = [
  [ "node", "testdata" ],
  [ "crypto", "_wasm", "target" ],
  [ "e", "t", "c", "." ]
].map(p => join(...p));
```

?

ayame113: This didn't work because the RegExp constructor expects backslashes to be double-escaped. Thanks for the suggestion though!

timreichen: Maybe
```ts
import { join } from "../path/mod.ts";

[ /* paths */ ].map((path) => join(...path.split("/")));
```

Since join uses the right `SEP` by default?

timreichen: Rather than messing with the paths directly I would suggest to either use `posix.globToRegExp` with `{ os: "linux" }`, which also is kinda hacky or to replace all paths with `RegExp`s in the first place.

ayame113: @timreichen I updated the code based on your suggestions. Thank you!

kt3k (approved):
LGTM

PR #3084, opened 2023-01-02
fix(testing/bdd): fix cases where `only` keyword skips tests that should run

### LINKED ISSUE #2979: testing: bdd ".only" option might not work
I haven't tested it but here's a report that it might not work: https://www.reddit.com/r/Deno/comments/zchl3q/showcase_my_first_project_in_deno_and_an_early/

pting-me:
Fixes #2979 

There seems to be unexpected behavior when using test suites with `describe.only`.

The file `only_test.ts` will not be committed. I'm using it to demonstrate which tests I believe should be running, but are not. There are still some edge cases which don't seem to be working properly, and I have documented.

If someone could please verify that the cases I laid out in [`only_test.ts`](https://github.com/denoland/deno_std/pull/3084/files#diff-5e15a097d4c27adf0d56823581f9e783418093ab95c39dd519082a8f47a80909) are correctly marked as "Should run", I'll start writing tests for those cases.


testing/bdd_examples/only_test.ts
```diff
@@ -0,0 +1,221 @@
...
+it("it 4.4", () => {
+  assertEquals(0, 0);
+});
+
+describe("describe 1.2", () => {
+  /**
+   * Should run
+   * Normal
+   */
+  it(onlyTests, "it 1.2.1", () => {
```
kt3k, line 65: Do we allow the mixture of flat and nested test grouping?

kt3k:
The expectations in only_test.ts looks good to me, but let's not consider flat test grouping nested inside other suites like the below:

```ts
describe("outer 2.2", () => {
  describe(onlyTests, "inner 2.2.1", () => {
    /**
     * Should run
     * Fixed
     */
    it("it 2.2.1", () => {
      assertEquals(0, 0);
    });
  });
});
```

I don't think we support this situation

testing/bdd_examples/only_test.ts
pting-me: Yes this specific case is actually [in the documentation](https://deno.land/manual@v1.28.3/basics/testing/behavior_driven_development#mixed-test-grouping)

pting-me:
> The expectations in only_test.ts looks good to me, but let's not consider flat test grouping nested inside other suites like the below:

The specific one you mentioned is actually fixed, although I can remove it if you prefer. There are three cases in the file marked "Bugged." I will make notes on them.

testing/bdd_examples/only_test.ts
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * Should not run
+   * Normal
+   */
+  it("it 6.2", () => {
+    assertEquals(0, 0);
+  });
+});
+
+describe(onlyTests, "describe 7.1", () => {
```
pting-me, line 183: @kt3k I think this one (and the similar one below it on [R193](https://github.com/denoland/deno_std/pull/3084/files#diff-5e15a097d4c27adf0d56823581f9e783418093ab95c39dd519082a8f47a80909R193)) is the main question I have. Because the `describe` is part of one test suite but the `it` is part of another, I think it's registering differently than expected.

testing/bdd_examples/only_test.ts
```diff
@@ -0,0 +1,221 @@
...
+  /**
+   * Should not run
+   * Normal
+   */
+  it("it 1.2.4", () => {
+    assertEquals(0, 0);
+  });
+});
+
+describe(onlyTests, "outer 2.1", () => {
```
pting-me, line 91: @kt3k This one is similar to the one you mentioned, although the suite is linked in the outer describe, rather than the inner describe. I think we can leave this one out if it's not a supported case.

kt3k:
Added snapshot test case to check the output of the given example.

Now the below existing test cases are failing:

```
global ... describe ... only ... signature 1 ... all options => ./testing/bdd_test.ts:954:17
global ... describe ... only ... signature 3 ... all options => ./testing/bdd_test.ts:994:17
global ... describe ... only ... signature 4 => ./testing/bdd_test.ts:1010:15
global ... describe ... only ... signature 5 => ./testing/bdd_test.ts:1022:15
global ... describe ... only ... signature 6 ... minimum options => ./testing/bdd_test.ts:1035:17
global ... describe ... only ... signature 6 ... all options => ./testing/bdd_test.ts:1047:17
global ... describe ... only ... signature 7 ... minimum options => ./testing/bdd_test.ts:1063:17
global ... describe ... only ... signature 7 ... all options => ./testing/bdd_test.ts:1075:17
global ... describe ... only ... signature 8 ... minimum options => ./testing/bdd_test.ts:1092:17
global ... describe ... only ... signature 8 ... all options => ./testing/bdd_test.ts:1104:17
```

@pting-me Do you have any idea what is going wrong?

iuioiua:
Gentle ping, @pting-me.

kt3k:
Closing as it is stale now. Let's revisit when someone start complaining about this

PR #3088, opened 2023-01-03
fix(path): Correctly handle trailing slashes for posix basename

familyboat:


kt3k:
The original commit was authored by @castle-sky's email address (I assume this is also your account). I changed it to @familyboat's email address to make the CLA checker work

kt3k (approved):
LGTM. Thanks for the fix!

kamilogorek:
The thing here is that the test assertion is incorrect 😅 it should return `/` according to spec.

jsejcksn:
> The thing here is that the test assertion is incorrect 😅 it should return `/` according to spec.

@kamilogorek Can you share a link to the spec text so we can evaluate that?

kamilogorek:
@jsejcksn sure - https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html

The order of operation for `path.basename("//", "bbb")` should be:
1. extraction
2. slash trimming
3. suffix trimming

```
input: // (slashes only, apply step 3)
value: /
remove suffix: bbb (step 6, no matches found, leave string intact)
output: /
```

I do agree that spec says that for `//` case its "implementation-defined", but GNU coreutils is trimming `//` down to `/`, so I assume this as the base for discussion (as its same for osx as well).

jsejcksn:
@kamilogorek Thanks!

@kt3k In regard to https://github.com/denoland/deno_std/pull/3088#issuecomment-1382318403, the only relevant comment at the top of the associated modules ([posix](https://github.com/denoland/deno_std/blob/0.172.0/path/posix.ts), [win32](https://github.com/denoland/deno_std/blob/0.172.0/path/win32.ts), [test](https://github.com/denoland/deno_std/blob/0.172.0/path/basename_test.ts)) is:

```js
// Ported from https://github.com/browserify/path-browserify/
```

and I don't find other context. What is the external guiding criteria for the algorithm in this library and its tests?

familyboat:
@jsejcksn The functionality of the `basename` in https://github.com/browserify/path-browserify/ exists flaw. So is same in node.js. Here i just made a little change to make `basename` behave reasonable.
See more discussion in https://github.com/denoland/deno_std/issues/3063 and https://github.com/denoland/deno_std/pull/3089

PR #3089, opened 2023-01-03
ref(path): Rework basename and dirname to be coreutils compatible

kamilogorek:
Note: tests are failing due to `parse` assertions being affected by this change.

---

This whole PR will invalidate some of the behaviors of `parse` function, which means that it would need to be reworked as well. However it's slightly more work, so before doing this, I'd like to ask what you even think about the expected behavior.

Should we follow `coreutils`? Should we follow `Node.js`?

Would it also be reasonable to implement `parse` as a superset of `basename/dirname/extname` calls instead? This way it would never get out of sync.

---

`basename` now passes all `coreutils` tests // https://github.com/coreutils/coreutils/blob/master/tests/misc/basename.pl which it did not before.

After reworking, the whole core is shared between `posix` and `win32` modules.

Some behaviors, like `basename(".js", ".js") === ""` are not valid anymore, as per spec, the suffix cannot consume the whole path.

> If the suffix operand is present (...) and is identical to a suffix of the characters remaining in string, the suffix suffix shall be removed from string.

Tests had to be factored out into table-test, as they also should be shared between `win32`/`posix`.
Some test variants are definitely doubled, but I did not want to remove any of them just for the sake of removing lines of code.

Implemented behavior is based on https://pubs.opengroup.org/onlinepubs/9699919799/utilities/basename.html and https://pubs.opengroup.org/onlinepubs/9699919799/functions/basename.html.

Fixes https://github.com/denoland/deno_std/issues/3063


path/_util.ts
```diff
@@ -26,8 +26,12 @@ export function isPosixPathSeparator(code: number): boolean {
   return code === CHAR_FORWARD_SLASH;
 }
 
+export function isWindowsPathSeparator(code: number): boolean {
```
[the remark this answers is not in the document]
kamilogorek: That's why I only use it in `isPathSeprator`, but agree that the naming might be confusing. Will change that for sure.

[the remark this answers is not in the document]
familyboat: Then you should not export it. Right?

kt3k:
Aligning to coreutils makes sense to me.

> Some behaviors, like basename(".js", ".js") === "" are not valid anymore, as per spec, the suffix cannot consume the whole path.

This also seems ok to me.

kamilogorek:
`path.basename` and `path.dirname` are now `coreutils` compatible. However, we cannot use `dirname` and `basename` in `parse`/`format` tests anymore, as they are not aligned and I'm not sure if they will/should ever be.

I explained the issue in this thread: https://twitter.com/kamilogorek/status/1610743549510680590

kt3k:
Can you give more examples of breaking changes by this PR?

> path.basename and path.dirname are now coreutils compatible. However, we cannot use dirname and basename in parse/format tests anymore, as they are not aligned and I'm not sure if they will/should ever be.

If we introduce this change, I think we should also update `base` part of `parse` result.

> And because of that, `p === path.format(path.parse(p))` would effectively stop holding true for all the cases.

I think `p === path.format(path.parse(p))` already doesn't hold for `p === "//"`, or am I missing something?

kt3k:
It's true that `path.format(path.parse("/foo///bar"))` becomes `"/foo///bar"` (triple (or more) slashes in the middle are kept), but I'm not sure that's desired/intended behavior. (Is it useful for any meaningful case?)

kamilogorek:
This PR is now complete, with some breaking changes, that are all making it coreutils-compliant.

All coreutils tests are now also passing with both `dirname` and `basename`.

Changes:
- `basename` prefix cannot be longer than path itself, it it is, skip trimming
- `basename` prefix cannot consume whole path, if it does, skip trimming
- `basename` will fallback to `/` in case nothing is found
- `basename` will trim all trailing slashes
- `dirname` will trim all trailing slashes
- `dirname` will fallback to `.` in case no path is provided
- `dirname` will fallback to `/` in case there is no match, be it only leading slashes or a single-segment path
- `parse` and will processed path in the exact same way as `basename` and `dirname` (including changes listed above), which means invalid paths like `/foo//bar//` will be correctly parsed to `/foo` and `bar`, then formatted to `/foo/bar`
- `format` will return `dir` in case `base === "/"` to prevent default value from returning `///`

path/_util.ts
```diff
@@ -131,3 +132,63 @@ export function encodeWhitespace(string: string): string {
...
+
+export function lastPathSegment(
+  path: string,
+  isSep: (char: number) => boolean,
+  start = 0,
+): string {
+  let matchedNonSeparator = false;
+  let end = path.length;
+
+  for (let i = path.length - 1; i >= 0; --i) {
```
familyboat, line 144: `i >= 0` should be `i >= start`.

path/basename_test.ts
```diff
@@ -4,33 +4,94 @@
...
+  [["aaa\\bbb", "\\bbb"], "bbb"],
+  [["aaa\\bbb", "a\\bbb"], "bbb"],
+  [["aaa\\bbb", "bbb"], "bbb"],
+  [["aaa\\bbb\\\\\\\\", "bbb"], "bbb"],
+  [["aaa\\bbb", "bb"], "b"],
+  [["aaa\\bbb", "b"], "bb"],
+  [["/aaa/bbb", "bb"], "b"],
+  [["C:"], ""],
+  [["C:."], "."],
+  [["C:\\"], ""],
```
familyboat, line 77: Because the result of `posix.basename('/')` is `/`, so i think the result of `win32.basename('C:\\')` is `\\`.

kamilogorek: Fixed

path/_util.ts
kamilogorek: Good catch. It also exposed one bug, which I also fixed.

path/basename_test.ts
```diff
@@ -4,33 +4,94 @@
...
+  [["basename.ext\\\\"], "basename.ext"],
+  [["foo"], "foo"],
+  [["aaa\\bbb", "\\bbb"], "bbb"],
+  [["aaa\\bbb", "a\\bbb"], "bbb"],
+  [["aaa\\bbb", "bbb"], "bbb"],
+  [["aaa\\bbb\\\\\\\\", "bbb"], "bbb"],
+  [["aaa\\bbb", "bb"], "b"],
+  [["aaa\\bbb", "b"], "bb"],
+  [["/aaa/bbb", "bb"], "b"],
+  [["C:"], "\\"],
```
familyboat, line 75: Should the result be `""`? The result of `posix.basename('')` is `""`.

familyboat: `C;` represents a relative path from the current directory of the C: drive. `C:\\` represents a root directory of the C: dirve. I grab it in https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats

kamilogorek: Ah, great catch, thanks! Updated

kt3k:
Trying to summarize the major breaking changes introduced by this, by examples:

- `basename("abc", "abc")` => BEFORE `""`, AFTER `"abc"`
- `basename("/")` => BEFORE `""`, AFTER `"/"`
- `dirname("aaa///bbb")` => BEFORE `"aaa//"`, AFTER `"aaa"`

These changes are introuced by aligning the API to `basename`. I think these all make more sense than the current behavior.

> parse will process path in the exact same way as basename and dirname (including changes listed above), which means invalid paths like /foo//bar// will be correctly parsed to /foo and bar, then formatted to /foo/bar

I think this change just make sense. It feels rather strange if it keeps double or triple slashes in the middle.

The below change is also introduced inspired by the change of `basename("/") => "/"`.

- `win32.basename("c:\\")` => BEFORE `""`, AFTER `"\\"`

This also looks reasonable to me.

I'm in favor of landing these changes.

kt3k (approved):
LGTM

PR #3102, opened 2023-01-12
fix(node): support ref & unref of TCP handle

kt3k:
closes https://github.com/denoland/deno/issues/16710

remaining:
- [x] add a test case


bartlomieju (approved):
LGTM

PR #3096, opened 2023-01-09
Added option to TextLineStream to ignore newlines at the end of the last line

impactaky:
Rel https://github.com/denoland/deno_std/issues/3075
This PR allows you to delete blank lines that would appear if "\n" was at the end of a chunk.

example
```example.ts
const test_case = ["foo\n\nbar\n", "foo\nbar\n\n", "foo\nbar"];

for (const str of test_case) {
  const textLineStream = new File([str], "untitled")
    .stream()
    .pipeThrough(new TextDecoderStream())
    .pipeThrough(new TextLineStream());
  console.log("");
  let lineNumber = 0;
  for await (const line of textLineStream) {
    lineNumber += 1;
    console.log(lineNumber, line);
  }
}
```

example output
```
1 foo
2
3 bar
4

1 foo
2 bar
3
4

1 foo
2 bar
```

Output when using this PR and `.pipeThrough(new TextLineStream({ignoreLastLF: false}));`
```
1 foo
2
3 bar

1 foo
2 bar
3

1 foo
2 bar
```

PR #3103, opened 2023-01-12
fix(streams): prevent artificial terminal newline in `TextLineStream`

### LINKED ISSUE #3075: `TextLineStream` creates extra newline after the end of stream
@crowlKats You're the original author of the `TextLineStream` class, so I'm tagging you here. I know you didn't author the PR that introduced the code in question below (#2318), but it looks like you did review it:

Any idea why it flushes with an extra newline? (See https://deno.land/std@0.170.0/streams/text_line_stream.ts?source#L26)

```ts
flush: (controller) => this.#handle("\r\n", controller)
```

I recently answered [this question on Stack Overflow](https://stackoverflow.com/q/74905946/438273) and it made me wonder why this behavior exists.

My understanding of the conventional definition of _variable length lines_ is:

> each line is an EOL character sequence of 1+ bytes (e.g. `\n`), preceded by a character sequence of 0+ bytes containing no EOL character sequences

So — when a text file consists of lines: if the file doesn't terminate with a newline, then it's malformed. I'm not debating those semantics or claiming that every text stream must follow that rule, but I wonder if the class was written to expect the exception to the rule, and whether we can update it to fix the issue described on SO.

Any thoughts?

jsejcksn:
Resolves #3075. (Alternative to #3096)

For details, see: https://github.com/denoland/deno_std/issues/3075#issue-1515492150 — quick summary:

`TextLineStream` is for breaking streams of text content into chunks of _lines_. My understanding of the conventional definition of _variable length lines_ is:

> each line is an EOL character sequence of 1+ bytes (e.g. `\n`), preceded by a character sequence of 0+ bytes containing no EOL character sequences

It seems that the class was written to expect the exception to this definition (the stream does not terminate with an EOL character sequence), and this PR updates the class to handle both cases.

### DIFF
--- a/streams/text_line_stream.ts
+++ b/streams/text_line_stream.ts
@@ -23,7 +23,16 @@ export class TextLineStream extends TransformStream<string, string> {
   constructor(options?: TextLineStreamOptions) {
     super({
       transform: (chunk, controller) => this.#handle(chunk, controller),
-      flush: (controller) => this.#handle("\r\n", controller),
+      flush: (controller) => {
+        if (this.#buf.length > 0) {
+          if (
+            this.#allowCR &&
+            this.#buf.length > 1 &&
+            this.#buf[this.#buf.length - 1] === "\r"
+          ) controller.enqueue(this.#buf.slice(0, -1));
+          else controller.enqueue(this.#buf);
+        }
+      },
     });
     this.#allowCR = options?.allowCR ?? false;
   }
--- a/streams/text_line_stream.ts
+++ b/streams/text_line_stream.ts
@@ -27,7 +27,6 @@ export class TextLineStream extends TransformStream<string, string> {
         if (this.#buf.length > 0) {
           if (
             this.#allowCR &&
-            this.#buf.length > 1 &&
             this.#buf[this.#buf.length - 1] === "\r"
           ) controller.enqueue(this.#buf.slice(0, -1));
           else controller.enqueue(this.#buf);
--- a/streams/text_line_stream_test.ts
+++ b/streams/text_line_stream_test.ts
@@ -119,3 +119,78 @@ Deno.test("[streams] TextLineStream - large chunks", async () => {
   }
   assertEquals(lines, 20000);
 });
+
+Deno.test(
+  "[streams] TextLineStream - no final empty chunk",
+  async (t) => {
+    await t.step("with terminal newline", async () => {
+      const inputChunks = [
+        "abc\n",
+        "def\nghi\njk",
+        "l\nmn",
+        "o\np",
+        "qr",
+        "\nstu\nvwx\n",
+        "yz\n",
+      ];
+
+      const textLineStream = new ReadableStream<string>({
+        start(controller) {
+          for (const chunk of inputChunks) controller.enqueue(chunk);
+          controller.close();
+        },
+      }).pipeThrough(new TextLineStream());
+
+      const lines = [];
+
+      for await (const chunk of textLineStream) lines.push(chunk);
+
+      assertEquals(lines, [
+        "abc",
+        "def",
+        "ghi",
+        "jkl",
+        "mno",
+        "pqr",
+        "stu",
+        "vwx",
+        "yz",
+      ]);
+    });
+
+    await t.step("without terminal newline", async () => {
+      const inputChunks = [
+        "abc\n",
+        "def\nghi\njk",
+        "l\nmn",
+        "o\np",
+        "qr",
+        "\nstu\nvwx\n",
+        "yz",
+      ];
+
+      const textLineStream = new ReadableStream<string>({
+        start(controller) {
+          for (const chunk of inputChunks) controller.enqueue(chunk);
+          controller.close();
+        },
+      }).pipeThrough(new TextLineStream());
+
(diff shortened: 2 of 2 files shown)


streams/text_line_stream.ts
```diff
@@ -23,7 +23,16 @@ export class TextLineStream extends TransformStream<string, string> {
   constructor(options?: TextLineStreamOptions) {
     super({
       transform: (chunk, controller) => this.#handle(chunk, controller),
-      flush: (controller) => this.#handle("\r\n", controller),
+      flush: (controller) => {
+        if (this.#buf.length > 0) {
+          if (
+            this.#allowCR &&
+            this.#buf.length > 1 &&
+            this.#buf[this.#buf.length - 1] === "\r"
```
0f-0b, line 31:
```suggestion
            this.#buf.endsWith("\r")
```

jsejcksn: Thanks — you're right that line 30 is unnecessary. I originally considered the `endsWith` method, but is it faster than the equality comparison against the indexed string value? I didn't have test data to compare when writing it.

jsejcksn: Update: I wrote and ran a [benchmark](https://deno.land/manual@v1.29.3/tools/benchmarker). According to the results (included below), it looks like the indexed equality comparison approach is faster for all strings except empty strings (which aren't possible in this context):

<details>
<summary>Benchmark info:</summary>

The benchmark tests are run on static strings. The medium and long ones are generated [lorem ipsum](https://en.wikipedia.org/wiki/Lorem_ipsum) texts using the npm package [`lorem-ipsum@2.0.8`](https://github.com/knicklabs/lorem-ipsum.js/tree/2.0.8)

`string_ending.bench.ts`:
```ts
import { loremIpsum } from "https://esm.sh/lorem-ipsum@2.0.8?pin=v102";

function generateLoremIpsumWordString(count: number): string {
  return loremIpsum({
    count,
    format: "plain",
    // suffix: "\n", // Only needed for paragraphs
    units: "words",
  });
}

const cases: [name: string, str: string][] = [
  ["empty string", ""],
  ["CR only", "\r"],
  ["short string", "lorem ipsum\r"],
  ["medium string", `${generateLoremIpsumWordString(2e2)}\r`],
  ["long string", `${generateLoremIpsumWordString(1.5e3)}\r`],
];

for (const [name, str] of cases) {
  Deno.bench({
    name: `${name}: indexed equality comparison`,
    fn: () => {
      str[str.length - 1] === "\r";
    },
    group: name,
    baseline: true,
  });

  Deno.bench({
    name: `${name}: endsWith`,
    fn: () => {
      str.endsWith("\r");
    },
    group: name,
  });
}

```

Output:

```txt
% deno bench string_ending.bench.ts
cpu: Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz
runtime: deno 1.29.3 (x86_64-apple-darwin)

file:///Users/deno/string_ending.bench.ts
benchmark                                       time (avg)             (min … max)       p75       p99      p995
---------------------------------------------------------------------------------- -----------------------------
empty string: indexed equality comparison    76.94 ns/iter  (70.61 ns … 112.65 ns)  80.03 ns  95.73 ns  97.04 ns
empty string: endsWith                       12.74 ns/iter   (11.46 ns … 40.73 ns)   12.9 ns   19.9 ns  22.42 ns

summary
  empty string: indexed equality comparison
   6.04x slower than empty string: endsWith

CR only: indexed equality comparison         12.09 ns/iter   (11.23 ns … 25.77 ns)  12.61 ns  16.84 ns  18.14 ns
CR only: endsWith                             17.5 ns/iter      (16 ns … 30.95 ns)  18.24 ns  24.02 ns  24.76 ns

summary
  CR only: indexed equality comparison
   1.45x faster than CR only: endsWith

short string: indexed equality comparison    11.86 ns/iter      (11 ns … 23.91 ns)  12.34 ns  16.71 ns  17.44 ns
short string: endsWith                       17.42 ns/iter      (16 ns … 36.81 ns)  18.24 ns  23.94 ns  25.69 ns

summary
  short string: indexed equality comparison
   1.47x faster than short string: endsWith

medium string: indexed equality comparison   13.45 ns/iter   (12.37 ns … 28.26 ns)  13.91 ns  18.52 ns   20.2 ns
medium string: endsWith                      19.25 ns/iter   (17.62 ns … 34.19 ns)  19.86 ns  26.23 ns  27.48 ns

summary
  medium string: indexed equality comparison
   1.43x faster than medium string: endsWith

long string: indexed equality comparison     13.45 ns/iter  (12.25 ns … 124.09 ns)  13.91 ns  19.07 ns  20.44 ns
long string: endsWith                        19.11 ns/iter   (17.62 ns … 34.06 ns)  19.85 ns  26.38 ns  27.62 ns

summary
  long string: indexed equality comparison
   1.42x faster than long string: endsWith

```

</details>

crowlKats:
could a test that doesnt have any form of newline at the end of the textStream be added?

jsejcksn:
> could a test that doesnt have any form of newline at the end of the textStream be added?

@crowlKats Added!

crowlKats (approved):
LGTM

PR #3109, opened 2023-01-15
tools: bundle std/node using esbuild

bartlomieju:
Needed for https://github.com/denoland/deno/pull/16748


_tools/bundle_node.ts
```diff
@@ -0,0 +1,45 @@
...
+);
+
+await Deno.writeTextFile(
+  `${PROJECT_ROOT}${BUNDLE_FILENAME}`,
+  sourceLines.join("\n"),
+);
+console.log(
+  "Bundle generated successfully at",
+  `${PROJECT_ROOT}${BUNDLE_FILENAME}`,
+);
```
bartlomieju, line 44: This bit should probably be done in deno itself, since it's specific to `ext/node` in Deno repo

_tools/bundle_node.ts
```diff
@@ -0,0 +1,45 @@
+import * as esbuild from "https://deno.land/x/esbuild@v0.16.17/mod.js";
+
+const PROJECT_ROOT = new URL("..", import.meta.url).pathname;
+const BUNDLE_FILENAME = "node_bundle.js";
```
bartlomieju, line 4: This file should be uploaded as an artifact for the release

bartlomieju:
@kt3k any idea why the CI is failing even though I'm up to date with `main` branch?

kt3k:
Looks like that test case (`[encoding/csv/stream] cancel CsvStream during iteration does not leak file`) became very flaky recently (maybe in 1 or 2 days? I observed it 4-5 times in main and PRs). I'll look into more details

bartlomieju:
Not needed because of https://github.com/denoland/deno/pull/17724

PR #2271, opened 2022-05-26
feat(node/cluster): `cluster` module for Node compat

cmorten:
Implements majority `cluster` module for Node compat.

Tests and full functionality not implemented owing to missing:

- `process.send()` - https://nodejs.org/api/process.html#processsendmessage-sendhandle-options-callback
- `process.connected` - https://nodejs.org/api/process.html#processconnected
- `process.on("message")` - https://nodejs.org/api/process.html#event-message
- `process.on("disconnect")` - https://nodejs.org/api/process.html#event-disconnect
- `ChildProcess.prototype.send()` - https://nodejs.org/api/child_process.html#subprocesssendmessage-sendhandle-options-callback
- `ChildProcess.prototype.connected` - https://nodejs.org/api/child_process.html#subprocessconnected
- `domain.create` - https://nodejs.org/api/domain.html#domaincreate

Which are likely blocked by https://github.com/denoland/deno/pull/13566 so can be implemented in a follow-up.

PR #3111, opened 2023-01-15
fix: revert "feat(node/cluster): `cluster` module for Node compat (#2271)"

kt3k:
This temporarily reverts #2271. #2271 requires access to some env vars (`NODE_CLUSTER_SCHED_POLICY` and `NODE_UNIQUE_ID`) at top level and that affects the all Deno programs (because `std/node` is loaded at start up of CLI).

closes https://github.com/denoland/deno/issues/17414


kt3k:
We can avoid the requirements of env permission when `Deno.permission.querySync` (https://github.com/denoland/deno/pull/17019) is available. We'll try to land this again when https://github.com/denoland/deno/pull/17019 landed.

bartlomieju (approved):
LGTM

PR #3112, opened 2023-01-15
test(node): prevent leak of permission requirement

kt3k:
This test case detects the error like https://github.com/denoland/deno/issues/17414, and prevents similar errors

This PR will pass the CI when the current env permission error is fixed in CLI (needs another release) and STD ( https://github.com/denoland/deno_std/pull/3111 ) repos.


bartlomieju (approved):
LGTM, very nice!

Issue #2974, opened 2022-12-03
Streaming stdin using `process.stdin` does not work


lucacasonato:
```ts
import { process } from "https://deno.land/std@0.167.0/node/process.ts";

function readStdin(): Promise<string> {
  return new Promise((resolve, reject): void => {
    let code = "";

    process.stdin.setEncoding("utf8");

    process.stdin.on("readable", function () {
      const chunk = process.stdin.read();
      if (chunk !== null) code += chunk;
    });

    process.stdin.on("end", function () {
      resolve(code);
    });
    process.stdin.on("error", reject);
  });
}

console.log(await readStdin());
```

This blocks parts of `@babel/cli` from running in NPM compat.

lucacasonato:
It looks to be the `process.stdin.on("readable", ...)` call.

kt3k:
@PolarETech 

> highWaterMark: Deno.isatty?.(Deno.stdin.rid) ? 0 : undefined

Does Node.js set the same value to `highWaterMark` of stdin?

PolarETech:
@kt3k

I checked the actual value in Node.js with the following sample code.

```mjs
// test.mjs
import process from "node:process";

(() => {
  new Promise((resolve, reject) => {
    console.log("highWaterMark:", process.stdin.readableHighWaterMark);

    process.stdin.setEncoding("utf8");
    process.stdin.on("readable", () => {
      console.log("chunk:", process.stdin.read());
    });
    process.stdin.on("end", () => {
      console.log("end");
      resolve;
    });
    process.stdin.on("error", reject);
  });
})();

// dummy.txt
foo
```

The execution results are as follows.

```bash
% node test.mjs
highWaterMark: 0
foo # User Input
chunk: foo

% node test.mjs < dummy.txt
highWaterMark: 65536
chunk: foo
chunk: null
end
```

From the results, I suppose the `highWaterMark` is:

- for tty: `0`, the same as `tty.ReadStream`
  https://github.com/nodejs/node/blob/9ca57fae11836af6cc8044efea89ec392356ccbb/lib/tty.js#L60
- for file: `65536`, the same as `fs.ReadStream`
  https://github.com/nodejs/node/blob/9ca57fae11836af6cc8044efea89ec392356ccbb/lib/internal/fs/streams.js#L155

---

Sorry, I have just noticed that the default value of `highWaterMark` in `Readable` is `16384`, not `65536`. I should have passed `64 * 1024` instead of `undefined`.
[...]

PR #3113, opened 2023-01-15
fix(node/process): null is not returned when reaching end-of-file in stdin

PolarETech:
Fixes: #2974

Currently, the `highWaterMark` value of `Readable` for `process.stdin` is fixed at 0. Therefore, when piping a file, `null` is not returned after the file is read to the end, and the `end` event is not emitted.
This PR will fix this behavior by setting the appropriate highWaterMark value based on the type of stdin.

__TODO__
Actual `highWaterMark` values in Node.js.

| stdin | Linux/Mac                        | Windows                          | Related streams          |
| ----- | -------------------------------- | ---------------------------------| ------------------------ |
| TTY   | <ul><li>[x] 0</li></ul>          | <ul><li>[x] 0</li></ul>          | tty.ReadStream (Duplex)  |
| FILE  | <ul><li>[x] 65536</li></ul>      | <ul><li>[x] 65536</li></ul>      | fs.ReadStream (Readable) |
| PIPE  | <ul><li>[x] 16384</li></ul>      | <ul><li>[x] 16384 (*1)</li></ul> | net.Socket (Duplex)      |
| TCP   | <ul><li>[x] 16384 (*1)</li></ul> | - (will not happen)              | net.Socket (Duplex)      |
| null  | <ul><li>[x] 65536</li></ul>      | <ul><li>[x] 65536 (*1)</li></ul> | fs.ReadStream (Readable) |
| other | <ul><li>[x] 16384</li></ul>      | <ul><li>[ ] 16384</li></ul>unresolved (*2) | Readable       |

(*1) The accuracy of type identification is insufficient and needs further improvement.
(*2) Non-console Windows applications are assumed ([ref](https://github.com/nodejs/node/blob/main/lib/internal/bootstrap/switches/is_main_thread.js#L236-L242)), but I couldn't find a way to get this state in Deno. The value in Node.js was checked in an electron app.

__Will not be addressed in this PR__

- Implementation of `tty.ReadStream`
   Use `Duplex` instead
- Implementation of `net.Socket` with `fd` option
   Use `Duplex` instead
- Implementation of `guessHandleType()` in "node/internal_binding/util.ts"

__Refs__
The highWaterMark value to be set is based on actual Node.js behavior and some source code.
https://github.com/nodejs/node/blob/v18.12.1/lib/internal/bootstrap/switches/is_main_thread.js#L189
https://github.com/nodejs/node/blob/v18.12.1/src/node_util.cc#L257
https://github.com/nodejs/node/blob/v18.12.1/deps/uv/src/unix/tty.c#L333
https://github.com/nodejs/node/blob/v18.12.1/deps/uv/src/win/handle.c#L31
https://github.com/nodejs/node/blob/v18.12.1/lib/tty.js#L60
https://github.com/nodejs/node/blob/v18.12.1/lib/internal/fs/streams.js#L155
https://github.com/nodejs/node/blob/v18.12.1/lib/net.js#L329
https://github.com/nodejs/readable-stream/blob/v4.2.0/lib/internal/streams/state.js#L11-L13

__Node test suite__
["pseudo-tty/test-readable-tty-keepalive.js"](https://github.com/nodejs/node/blob/v18.12.1/test/pseudo-tty/test-readable-tty-keepalive.js) test seems to be related to this change but does not pass the test. The `data` is not received as expected in `process.stdin.on('readable', ...)`. 
The test has not been added because it appears to be failing due to factors other than this change, and it may take time to investigate. Perhaps some kind of PTY is needed in addition.

__Memo__
Some issues with the node compat layer have been found while working on this PR.

- stdin of `stdio` option is ignored in `child_process.spawnSync()` ([src](https://github.com/denoland/deno_std/blob/d6d3aac45e0822ed89ebc8d255d0e50373743d13/node/internal/child_process.ts#L749))
- `input` option is ignored in `child_process.spawnSync()` ([src](https://github.com/denoland/deno_std/blob/d6d3aac45e0822ed89ebc8d255d0e50373743d13/node/internal/child_process.ts#L749))


node/_process/streams.mjs
```diff
@@ -103,7 +103,7 @@ export const stdout = stdio.stdout = createWritableStdioStream(
 
 /** https://nodejs.org/api/process.html#process_process_stdin */
 export const stdin = stdio.stdin = new Readable({
-  highWaterMark: 0,
+  highWaterMark: Deno.isatty?.(Deno.stdin.rid) ? 0 : 64 * 1024,
```
kt3k, line 106: `Deno.stdin` also needs to be guarded by `?` to make this compatible with Deno Deploy
```suggestion
  highWaterMark: Deno.isatty?.(Deno.stdin?.rid) ? 0 : 64 * 1024,
```

kt3k:
Can you add some custom test case in `node/process_test.ts` which exercise the case this PR fixes?

node/process_test.ts
```diff
@@ -326,6 +326,66 @@ Deno.test({
...
+    await promise;
+    assertEquals(process.stdin.readableHighWaterMark, 0);
+    assertEquals(data, expected);
+  },
+});
+
+Deno.test({
+  name: "process.stdin readable with piping a file",
+  async fn() {
+    const expected = ["65536", "foo", "bar", "null", "end"];
```
kt3k, line 363: In node.js, the first number becomes `16384`. Is this difference intentional?

node/process_test.ts
```diff
@@ -326,6 +326,66 @@ Deno.test({
   },
 });
 
+Deno.test({
+  name: "process.stdin readable with a TTY",
+  ignore: !Deno.isatty(Deno.stdin.rid),
```
kt3k, line 331: Leave a TODO comment like: `TODO: Run this test even in non tty environment`

node/process_test.ts
PolarETech: My confirmation was lacking.
As I mentioned in the [issue](https://github.com/denoland/deno_std/issues/2974#issuecomment-1382845551), the highWaterMark is `65536` with piping a file, But when running the code in child_process with piping stdin, the highWaterMark is a different value.

- tty(tty.ReadStream): 0
- piping files(fs.ReadStream): 65536
- piping stdin(Stream?): 16384

I'll look into it further.
Thank you for pointing this out.

node/process_test.ts
PolarETech: A TODO comment has been added. (c10b492)

node/_process/streams.mjs
```diff
@@ -101,9 +101,27 @@ export const stdout = stdio.stdout = createWritableStdioStream(
   "stdout",
 );
 
+function _adjustHighWaterMark() {
+  if (Deno.isatty?.(Deno.stdin?.rid)) return 0;
+
+  try {
+    // TODO(PolarETech): On Windows, isFile() returns true even for a stream,
```
justinmchase, line 108: Can/should this be fixed lower down in the native code?

node/_process/streams.mjs
```diff
@@ -101,9 +101,27 @@ export const stdout = stdio.stdout = createWritableStdioStream(
...
+    // so it cannot distinguish a stream from a file.
+    // When stdin is a stream, it should return `undefined` (= 16 * 1024), not 64 * 1024.
+    if (Deno.fstatSync(Deno.stdin?.rid).isFile) return 64 * 1024;
+  } catch (_) {
+    // Avoid error that occurs when stdin is null on Windows.
+    return 64 * 1024;
+  }
+
+  // TODO(PolarETech): On Linux and Mac, 64 * 1024 should be returned when stdin is null.
```
justinmchase, line 117: This seems weird to me, is there any more info you can add to this comment for the following:

A) why only the two?
B) why this number?

node/_process/streams.mjs
PolarETech: Thanks for taking a look at this PR.
I am currently checking the native code as well, including the points you mentioned.

node/_process/streams.mjs
PolarETech: A) The behavior is different on Linux/Mac and Windows. On Windows, it is able to return 64 * 1024 when stdin is null. (line 114)
B) 64 * 1024 is based on actual value in Node.js.

justinmchase: Sure but is the value in nodejs meaningful or just some sentinal value they happened to pick? I would find a note here useful, to be honest. Probably anyone who needs to touch this area again will need to build up a lot of context, any hints will probably be useful.

node/_process/streams.mjs
```diff
@@ -101,9 +101,36 @@ export const stdout = stdio.stdout = createWritableStdioStream(
   "stdout",
 );
 
+function _adjustHighWaterMark() {
+  if (Deno.isatty?.(Deno.stdin?.rid)) return 0; // stdin is a TTY
+
+  if (Deno.build.os !== "windows") {
+    const fileInfo = Deno.fstatSync?.(Deno.stdin?.rid);
+    if (fileInfo.isFile) return 64 * 1024; // stdin is a redirected file
```
justinmchase, line 109: I see this constant appearing multiple times in this file, it seems like it could be beneficial to break it out into a constant or enum and reference that instead. A descriptive name could help a lot.

I'm not totally sure what each value signifies so I'm guessing below, but you get the idea.

```ts
enum HighWaterMark {
  Interactive = 0,
  Ignore = 64 * 1000,
}
```
```ts
if (fileInfo.isFile) return HighWaterMark.Ignore;
```

node/_process/streams.mjs
```diff
@@ -101,9 +101,36 @@ export const stdout = stdio.stdout = createWritableStdioStream(
...
+
+  // Avoid error that occurs when stdin is null on Windows.
+  try {
+    const fileInfo = Deno.fstatSync?.(Deno.stdin?.rid);
+    // TODO(PolarETech): Need a better way to identify a piped stdin.
+    // On Windows, `Deno.fstatSync(rid).isFile` returns true even for a piped stdin.
+    // Therefore, a piped stdin cannot be distinguished from a file by this property.
+    // The mtime, atime, and birthtime of the file are "2339-01-01T00:00:00.000Z",
+    // so this is used as a workaround.
+    if (fileInfo.birthtime.valueOf() === 11644473600000) return undefined; // stdin is "pipe"
```
justinmchase, line 123: should it include both?

```ts
    if (fileInfo.isFile && fileInfo.birthtime.valueOf() === 11644473600000) return undefined; // stdin is "pipe"
```

node/_process/streams.mjs
PolarETech: The value in Node.js is probably determined by performance and other considerations.
Regardless of the validity of the value, Deno members have indicated their policy that the implementation of the Node compat layer should be aligned with Node.js as much as possible, and I'm making an effort to follow the policy.

node/_process/streams.mjs
PolarETech: Thanks for the advice.
I will do some code refactoring to take into account the feedback I have received before changing this PR from Draft to Open.

node/_process/streams.mjs
PolarETech: I have added information to the Description that was found during the work on this PR. I will continue to update accordingly.

justinmchase: No, problem the code is basically fine I'm just probing for some hints about the magic numbers which I realize are coming from an attempt to be compatible with node. Lookin good.

node/process_test.ts
```diff
@@ -326,6 +326,66 @@ Deno.test({
...
+    assertEquals(data, expected);
+  },
+});
+
+Deno.test({
+  name: "process.stdin readable with piping a file",
+  async fn() {
+    const expected = ["65536", "foo", "bar", "null", "end"];
+    const scriptPath = "./node/testdata/process_stdin.ts";
+    const file = await Deno.readFile(`./node/testdata/process_stdin_dummy.txt`);
```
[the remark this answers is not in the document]
PolarETech: The test was divided into two tests, one for file and one for pipe. And I used inline Uint8Array in the test for pipe.
https://github.com/denoland/deno_std/pull/3113/files#diff-f27eacddc948ff5c5ce7c851fccf64c00f313b8edc28862b5066c211e2b14d72R402

PolarETech:
The stdin implementation has been changed to be more in line with Node.js.
Some TODOs remain, but they seem difficult to solve in a short time and I would like to be allowed to postpone them.

kt3k (approved):
LGTM. Nice work! Thanks! @PolarETech

PR #3115, opened 2023-01-15
fix(fs): Change globstar default to true for expandGlob and expandGlobSync

kamilogorek:
https://deno.land/std@0.172.0/fs/expand_glob.ts?s=ExpandGlobOptions extends https://deno.land/std@0.172.0/path/mod.ts?s=GlobOptions which states that `globstar: boolean = true`.

Fixes https://github.com/denoland/deno_std/issues/3099


kt3k (approved):
Makes sense. Thanks for looking into this.

LGTM

Issue #3110, opened 2023-01-15
semver.format function is dropping build data


justinmchase:
**Describe the bug**
When formatting a semver object, the build data is not printed even if its on the original semver object.

**Steps to Reproduce**

```ts
import { parse } from "https://deno.land/std@0.172.0/semver/mod.ts";
const v = parse('1.2.3+4')
console.log(v.format()) // 1.2.3, expected 1.2.3+4
```

**Expected behavior**

I expect the build metadata to continue to be printed as part of the version

**Environment**

- deno 1.28.3 (release, x86_64-apple-darwin)
- std version: 0.172.0

### Recommended Function
```ts
function format(semver: SemVer): string {
  const main = `${semver.major}.${semver.minor}.${semver.patch}`;
  const pre = semver.prerelease.join(".");
  const build = semver.build.join(".");
  return [[main, pre].filter(_ => _).join('-'), build].filter(_ => _).join("+")
}
```

justinmchase:
Additionally there is no function to replace the build values. You can alter the pre-release by calling the `.increment` function and providing a new pre-release name but there is no way to increment or even replace the build number with out doing string manipulation and re-parsing.

kamilogorek:
This is exact same implementation as https://www.npmjs.com/package/semver uses, so there's a question whether Deno should be compatible with it, or be "more correct" with some of its methods.
`format` is used for version comparisons in some places, so just adding `build` to the `format` breaks few tests right away.

justinmchase:
I'm sure you're aware of it but semver.org [says](https://semver.org/#spec-item-10):

> Build metadata MUST be ignored when determining version precedence.

Perhaps `.format()` is fine as is but then the `.toString()` implementation is what should retain the build metadata? I could see how changing this would affect some tests perhaps but I think the intent of the tests themselves is likely still the same, meaning their comparisons should still yield the same results regardless of the build metadata present.

Would you prefer this issue be raised there first and then merged here if/after accepted there?

justinmchase:
Here is a related thread:
https://github.com/npm/node-semver/issues/174

It seems like the developers doing the maintaining perhaps don't even understand the spec itself very well.

Here is what I would recommend:

A) Do it the right way here, call it a feature advantage for Deno over node.
B) Keep the parsing as is, keep the format function as is, it has the side effect of setting the internal `this.version` field, which has only the parts of the version which are significant for determining precedence.
[...]

PR #3117, opened 2023-01-15
feat(semver): Allow for passing includeBuild into format method

kamilogorek:
ref: https://github.com/denoland/deno_std/issues/3110

### DIFF
--- a/semver/mod.ts
+++ b/semver/mod.ts
@@ -542,10 +542,13 @@ export class SemVer {
     this.format();
   }
 
-  format(): string {
+  format({ includeBuild } = { includeBuild: false }): string {
     this.version = this.major + "." + this.minor + "." + this.patch;
     if (this.prerelease.length) {
       this.version += "-" + this.prerelease.join(".");
+    }
+    if (includeBuild && this.build.length) {
+      this.version += "+" + this.build.join(".");
     }
     return this.version;
   }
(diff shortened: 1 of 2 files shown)


semver/mod.ts
```diff
@@ -542,11 +542,14 @@ export class SemVer {
     this.format();
   }
 
-  format(): string {
+  format({ includeBuild } = { includeBuild: false }): string {
```
kt3k, line 545: Should this be optional? How about showing the build metadata always?

kamilogorek: We cannot do that, as `format()` call is used for semver comparison, ref: https://github.com/denoland/deno_std/issues/3110#issuecomment-1383221174

Also:
> Although format itself is not pure function, so maybe an alternative just for printing would be preferable.

justinmchase:
It may be wise to just adjust the `.toString()` function to return the version with metadata and leave the format function alone, for backwards compatibility reasons. 

Also, it would be nice if there was a function which returned a new `SemVer` with the build set. since there is now way to apply build metadata to a version without doing string manipulation at the moment.

```ts
var v0 = semver.parse('1.2.3')
var v1 = v0.metadata('4')
console.log(v1.toString()) // 1.2.3+4
```

justinmchase:
I made a more complete version of this PR with some feedback from the discussion:

https://github.com/denoland/deno_std/pull/3126

kamilogorek:
Closed in favor of https://github.com/denoland/deno_std/pull/3126

PR #3122, opened 2023-01-17
fix(node/util): Stricter runtime type checking

petamoriken:
In accordance with the ECMAScript specification, internal slots are used for runtime type checking whenever possible.


kt3k (approved):
LGTM

PR #3124, opened 2023-01-17
feat(strings): add strings library

naltun:
Add strings library for preliminary review. deno_std does not have a string manipulation library, so I ported (part of) Golang's [strings](https://pkg.go.dev/strings@go1.19.5) module.

Having reviewed the contributing notes, I understand that deno_std can be seen as a loose port of the Go standard library, so feel free to consider this for incorporation. I am new to TypeScript/JavaScript and have followed Deno since its announcement, and I am using this project as a way to learn TypeScript while also learning how Deno works. If there are any improvements that are needed, please let me know and I will address them.

I will continue work on porting the remaining functions I believe are useful while a reviewer can take a moment review whether Deno needs a strings library.


iuioiua:
Frankly, I think these functions are a little trivial and likely unneeded thanks to the in-built versatility [string methods](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) in JS. In other words, this functionality already exists without the need to import any external code.

Either way, the intent and enthusiasm behind this PR is great! If you're looking to get started, try searching issues with the "good first issue" tag 😁

timreichen:
I am bot against a string mod per se, but imo it does not make sense to double functionality in a mod that is covered natively or a one-liner. Which functions would this mod include?

naltun:
> I am bot against a string mod per se, but imo it does not make sense to double functionality in a mod that is covered natively or a one-liner. Which functions would this mod include?

Go's [strings module](https://pkg.go.dev/strings@go1.19.5#pkg-overview) has some functions I think are unneeded, but has some that are useful. These include `Cut()`, `LastIndex()`, `Title()`, and `HasPrefix()`. If the `strings` module is unneeded, that's OK. @iuioiua had a good idea with the `good first issue` tags.

timreichen:
> Go's [strings module](https://pkg.go.dev/strings@go1.19.5#pkg-overview) has some functions I think are unneeded, but has some that are useful. These include `Cut()`, `LastIndex()`, `Title()`, and `HasPrefix()`. If the `strings` module is unneeded, that's OK. @iuioiua had a good idea with the `good first issue` tags.

- `Cut` seems like a very specific case of [split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split?retiredLocale=de) to me:
```ts
function cut(string: string, separator: string) {
  const fragments = string.split(separator)
  return [fragments.shift(), fragments.join(separator), fragments.length > 0]
}
```
Is there a common and frequent use case for this?
- `LastIndex` is covered by [lastIndexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/lastIndexOf)
- `Title` is part of string casings, which there is a [case](https://deno.land/x/case@2.1.1) mod in /x.
So the question is if we should include a general string casing mod somewhere in std which would make sense to me.
- `HasPrefix` is covered by [startsWith](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith).

naltun:
@timreichen 

> Cut seems like a very specific case of split
> Is there a common and frequent use case for this?

So Go's `strings.Cut` is similar to Python's `str.partition`. JS's `String.from.split()` is useful, but something like `cut()` saves the user from having to grab an index of the returned split, e.g. `"my string".split(" ")[-1]`. For this, I use `str.partition` whenever I need to split strings (in Python).

Go's `strings` module is _a bit_ unneeded in JS, since Go strings behave differently. That said, there might be some useful helpers.

I'm still OK if we deem a `strings` library is unneeded.  If we want to include a small `strings` library`, we can determine which functions might be useful and I can work on those.

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * ```ts
+ * import { contains } from "https://deno.land/std@$STD_VERSION/strings/strings.ts";
+ *
+ * contains("Hello, Deno!", "Deno"); // true
+ * ```
+ */
+
+export function contains(s: string, substr: string): boolean {
+  return s.includes(substr);
+}
```
justinmchase, line 37: I don't like this, people should be using the native function in this case.

At _worst_ this function show throw a NotSupportedError instead and have an `@obsolete` attribute which says please use `s.includes(substr)` which would help people who gravitate to the word "contains" to be redirected to the right place which is the built-in function called `includes`.

But I would not just wrap standard functions and have people write a bunch of code that differs from the rest of the community.

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * count("cheese", "e"); // 3
+ * count("five", ""); // 5
+ * ```
+ */
+
+export function count(str: string, substr: string): number {
+  if (substr === "") return str.length + 1;
+  return str.split(substr).length - 1;
+}
```
justinmchase, line 55: This one feels like a homework assignment more than a function someone would actually use. 

Also, I'd probably use a regex

```ts
"cheese".match(/e/g)?.length ?? 0 // 3
```

Not a big fan of having functions that won't really be used.

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * ```
+ */
+
+export function hasPrefix(str: string, prefix: string): boolean {
+  if (str.length >= prefix.length && str.slice(0, prefix.length) === prefix) {
+    return true;
+  } else {
+    return false;
+  }
+}
```
justinmchase, line 96: This is already on a string:

`str.startsWith(prefix)`

Not a big fan of alternate implementations of standard functions.

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * let [before, after, found] = cut("Deno", "n"); // "De", "o", true
+ * let [before, after, found] = cut("Hello, World!", " "); // "Hello," "World!", true
+ * let [before, after, found] = cut("Fresh", "i"); // "Fresh", "", false
+ */
+
+export function cut(str: string, sep: string): [string, string, boolean] {
+  const s = str.split(sep);
+  if (s.length == 2) return [s[0], s[1], true];
+  return [str, "", false];
+}
```
justinmchase, line 74: Same with this, it doesn't seem like really something people could use reliably. It just seems like a wrapper on split.

```ts
const [l, r, ...rest] = s.split(sep)
if (rest) // err
else // ok
```

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * hasSuffix("Calvin", "n"); // true
+ * hasSuffix("Calvin", ""); // true
+ * hasSuffix("Calvin", "Cal"); //false
+ * ```
+ */
+
+export function hasSuffix(str: string, suffix: string): boolean {
+  if (str.endsWith(suffix)) return true;
+  return false;
+}
```
justinmchase, line 115: Same here.

Can be reduced to

```suggestion
export function hasSuffix(str: string, suffix: string): boolean {
  return str.endsWith(suffix)
}
```

strings/strings.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @example
+ * ```ts
+ * import { toLower } from "https://deno.land/std@$STD_VERSION/strings/strings.ts";
+ *
+ * toLower("Hello, Deno!"); // hello, deno!
+ * ```
+ */
+export function toLower(s: string): string {
+  return s.toLowerCase();
+}
```
justinmchase, line 129: Again, I just can't get behind wrapping and changing the names of standard functions, it will just cause code confusion for people coming to a code base using these functions. They will all have to wonder if this toLower is somehow different from the built-in toLowerCase and will have to go and dig into this library to figure out what its doing differently and ultimately wonder why its even here at all.

justinmchase (commented):
I would not recommend making this change, the functions in this PR don't make sense to me as actual functions that should be added to this library.

I suppose it's possible that there are other string functions that could be added, I wouldn't want to make the case that a strings module shoudn't exist at all but these particular functions seem too low of value.

Some others that may be useful if they're not already in another module:

* toCamelCase()
* toSnakeCase()
* toPascalCase()
* toTitleCase()
* toBase64()
* sha256()
* hash()

etc.

There is another possible motivation for this PR I'd like to come out against and that would be the desire to change the form of standard javascript functions from the form of `str.fn()` to `fn(str)`.

I do agree that additional string functions from a library should be in the form of `fn(str)` vs modifying `String.prototype` but they should not be gratuitous wrappers over existing standard javascript functions. Even at the risk of mixing styles, I would not support that. I think that will add confusion and an unnecessary layer indirection and probably a lot of dependence on this library where it was not needed.

Perhaps it would be wise to make a case for existing 3rd party libraries which have valuable string functions, that are popular in the community and if such a module could be identified maybe a case could be made at that point to pull those functions into this module.

pting-me:
I agree with @justinmchase that we should not rename functions simply to match Golang std.

However, I do think there's a case to be made for _feature_ parity. There are a couple of these functions which have multiple ways to write, which I think we can make a case for: `count` and `cut`. Even if it may seem easy to find _one_ solution to this, I don't think it's very obvious on how to find the _fastest_ solution to this.

Another approach I'd consider would be to implement some of the TC39 proposals for String, but it doesn't look like there were any submissions for `count` and `cut`. Not sure how the community feels about this one.

justinmchase:
@pting-me I heard the op say something similar about this being roughly like the go lang std library... this seems super weird to me. I could understand Rust maybe but Go? Weird decision.

strings/strings.ts
pting-me: I think it's okay. Seems useful enough to be included in [Voca](https://vocajs.com/#countSubstrings), [Ruby's String class](https://ruby-doc.org/3.2.0/String.html#method-i-count), and most importantly [Golang std](https://pkg.go.dev/strings@go1.19.5#Count).

We should build for performance rather than whatever is most idiomatic. In this case, [`match` is definitely faster than `split`](https://measurethat.net/Benchmarks/Show/7822/0/strmatch-vs-strsplit), but probably hard to beat a naive loop. Could probably look at [Voca](https://github.com/panzerdp/voca/blob/master/src/count/count_substrings.js) or maybe there's some clues in [v8](https://chromium.googlesource.com/v8/v8/+/4.3.49/src/string.js?autodive=0%2F%2F#1018) that we could build on.

strings/strings.ts
pting-me: The main draw of this function over `split` is that this only splits the _first_ instance of the separator, rather than every instance.

- [Golang `Cut`](https://pkg.go.dev/strings#Cut) does seem like a very recent addition
- [Ruby `partition`](https://ruby-doc.org/3.2.0/String.html#method-i-partition) is very similar
- Doesn't seem like Voca has anything like it
- [Rust `split_once`](https://doc.rust-lang.org/std/string/struct.String.html#method.split_once)

Which also means this implementation is actually incorrect.

Rather than using `split`, (which would require a `join` after) I think we can make get more performance with something like this:

```typescript
// might be able to get a faster implementation if we use a loop instead of iterating 3 times (with indexOf, 2x slice)
const flag = s.indexOf(sep);
if (flag === -1) return [str, '', false];
return [
  str.slice(0, flag),
  str.slice(flag + sep.length),
  true,
];
```

pting-me:
My understanding is that there's a lot of good things about the Golang std library (Ryan in particular seems to be a huge fan), and the primary goal of `deno_std` was to port over that functionality.

In this particular case I think the right approach would be to go through the strings library to see where we are missing functionality that isn't covered by the native `String` and work from there. I'd imagine the functions to be used similarly to tools like Voca and Lodash.

ayame113:
(As others have also pointed out in the comments) some functions are substitutable with built-in methods like this:

- `compare` -> `String.prototype.localeCompare` or `Intl.Collat​​or.prototype.compare`
- `contains` -> `String.prototype.includes`
- `count` -> `String.prototype.match`
- `hasPrefix` -> `String.prototype.startsWith`
- `hasSuffix` -> `String.prototype.endsWith`
- `toLower` -> `String.prototype.toLowerCase`

On the other hand, `cut` looks similar to the [Reversible string split proposal](https://github.com/tc39/proposal-reversible-string-split). How about providing a function with the same behavior as the Reversible string split proposal?

Also suggestions like `toCamelCase`/`toSnakeCase`/`toPascalCase`/`toTitleCase` sound good to me.

naltun:
I am closing this as right now I am preoccupied writing useful Deno code elsewhere. This PR initially started out as an exercise to learn Deno while having a meaningful contribution to `deno_std`. It seems like there is interest in some form of a `strings` library, but I think it best to start with fresh code that solves existing issues. Thanks to everyone who participated in the discussion.

PR #3126, opened 2023-01-19
feat(semver): Add support for build metadata

justinmchase:
# Overview

Currently the build metadata is parsed but is not printed out in any form. A semver which is parsed will not be identical to 
the printed output:

```ts
console.log(semver.parse("1.2.3+4").format()) // 1.2.3
```

Additionally, once set, there is no way to alter the metadata on the SemVer instance without formatting into a new string and re-parsing:

```ts
const build = "5";
const v0 = semver.parse("1.2.3+4").increment("pre");
const v1 = semver.parse(`${v0}+${build}`);
```

Therefore this PR attempts to address these two issues with the following APIs:

1. Add a new, optional, options object parameter onto the `.format()` function which contains a single field `style`. The style will dictate the format of the _output_ of the function, it will not affect the format internal mutated fields `this.version` or `this.raw`, to preserve backwards compatibility and not require an extensive refactor.
2. The `.increment()` function will now have a new, optional, `metadata: string` parameter which will set the `this.build` field with the new metadata or reset it to empty. Incrementing with no build data will not retain the previous build metadata.

```ts
console.log(semver.parse("1.2.3+4").format({ style: "full" )) // 1.2.3+4

const build = "5";
const v0 = semver.parse("1.2.3+4").increment("pre", undefined, build);
console.log(v0.format({ style: "full" })) // 1.2.3-0+5
```

## Related To

* fixes https://github.com/denoland/deno_std/issues/3110
* replaces #3117
* related https://github.com/npm/node-semver/issues/174

### Notes
* According to the semver spec the build metadata must not be used during the comparison calculations. For this reason the internal `this.version` and `this.raw` fields are preserved as they were since those fields can be part of the comparison calculations. The only modifications are to the _output_ of the format function.
* Currently the parsing functions do actually _parse_ the build metadata correctly, so there was no need to modify this. It handles it safely it simply does not output it or support modification correctly.

### DIFF
--- a/semver/format_test.ts
+++ b/semver/format_test.ts
@@ -4,7 +4,11 @@ import { assertEquals } from "../testing/asserts.ts";
 import * as semver from "./mod.ts";
 
 Deno.test("format", async (t) => {
-  const versions: [string, "release" | "prerelease" | "build" | "full" | undefined, string][] = [
+  const versions: [
+    string,
+    "release" | "prerelease" | "build" | "full" | undefined,
+    string,
+  ][] = [
     ["1.2.3", undefined, "1.2.3"],
     ["1.2.3", "release", "1.2.3"],
     ["1.2.3", "prerelease", "1.2.3"],
@@ -47,9 +51,9 @@ Deno.test("format", async (t) => {
       name: `format(${version} ${style} ${expected})`,
       fn: () => {
         const v = semver.parse(version)!;
-        const actual = v.format({ style })
+        const actual = v.format({ style });
         assertEquals(actual, expected);
-      }
-    })
+      },
+    });
   }
 });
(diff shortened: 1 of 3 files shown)


semver/format_test.ts
```diff
@@ -0,0 +1,59 @@
+// Copyright Isaac Z. Schlueter and Contributors. All rights reserved. ISC license.
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { assertEquals } from "../testing/asserts.ts";
+import * as semver from "./mod.ts";
+
+Deno.test("format", async (t) => {
+  const versions: [
```
kt3k, line 7: nice exhausive tests!

kt3k (approved):
LGTM. Looks nice enhancement. Thanks for your contribution!

PR #3128, opened 2023-01-20
fix(encoding/csv): Escape cells containing newlines

lionel-rowe:
Escapes CSV cells containing newlines, so that they can be displayed correctly when opened in MS Excel.

**Old behavior:**

```js
stringify([['(1)\n(2)', '(3)']], { headers: false })
// => (1)\n(2),(3)\r\n
```

**New behavior:**

```js
stringify([['(1)\n(2)', '(3)']], { headers: false })
// => "(1)\n(2)",(3)\r\n
```

Note that this PR does _not_ normalize newlines within cells, so this results in a file with plain LFs within a cell but CRLFs delimiting rows. I'd argue that this doesn't constitute inconsistent line endings, as only the end-of-row ones are really "line endings" within the context of the file. In any case, the user can normalize the text within cells however they wish, if necessary.

Note also that this doesn't resolve all MS Excel compatibility issues — for files containing Unicode text, a BOM (`U+FEFF`) still needs to be prepended to force encoding to UTF-8. However, that's out of the scope of this PR.

I've changed the name of ~~the exported~~ `const NEWLINE`, which is used in testing, to `CRLF` ("newline" typically refers to LF, i.e. `\n`). Almost all changed lines within `encoding/csv_test.ts` are due to this change — the only substantive change to a test is "Value with LF":

```diff
  await t.step(
    {
      name: "Value with LF",
      fn() {
        const columns = [0];
        const data = [["foo\n"]];
-       const output = `0${CRLF}foo\n${CRLF}`;
+       const output = `0${CRLF}"foo\n"${CRLF}`;
        assertEquals(stringify(data, { columns }), output);
      },
    },
  );
```

### DIFF
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -25,7 +25,8 @@ export {
 export type { ReadOptions } from "./csv/_io.ts";
 
 const QUOTE = '"';
-export const NEWLINE = "\r\n";
+const LF = "\n";
+const CRLF = "\r\n";
 
 export class StringifyError extends Error {
   override readonly name = "StringifyError";
@@ -40,7 +41,7 @@ function getEscapedString(value: unknown, sep: string): string {
 
   // Is regex.test more performant here? If so, how to dynamically create?
   // https://stackoverflow.com/questions/3561493/
-  if (str.includes(sep) || str.includes(NEWLINE) || str.includes(QUOTE)) {
+  if (str.includes(sep) || str.includes(LF) || str.includes(QUOTE)) {
     return `${QUOTE}${str.replaceAll(QUOTE, `${QUOTE}${QUOTE}`)}${QUOTE}`;
   }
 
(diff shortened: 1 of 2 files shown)


encoding/csv.ts
```diff
@@ -25,7 +25,8 @@ export {
 export type { ReadOptions } from "./csv/_io.ts";
 
 const QUOTE = '"';
-export const NEWLINE = "\r\n";
+const LF = "\n";
+export const CRLF = "\r\n";
```
kt3k, line 29: Maybe let's not export this, but define another `CRLF` in testing to avoid this being considered as a public API.

encoding/csv_test.ts
```diff
@@ -1231,7 +1226,7 @@ Deno.test({
         fn() {
           const columns = [0];
           const data = [["foo\n"]];
-          const output = `0${NEWLINE}foo\n${NEWLINE}`;
+          const output = `0${CRLF}"foo\n"${CRLF}`;
```
kt3k, line 1229: Note: This is the actual change of the test case by this fix

kt3k (approved):
LGTM. Thanks!

Issue #2446, opened 2022-07-18
Don't unnecessarily request to read environment variable


timfish:
My library uses some npm dependencies from esm.sh but this causes the user to be prompted for access to the environment due to this code:

https://github.com/denoland/deno_std/blob/8b34f7363501d17b059ab65575b985902ef190ba/node/internal/util/debuglog.ts#L106-L115

It looks like this file is included via `https://deno.land/std@0.148.0/node/process.ts` which esm.sh is injecting because the npm module checks for the existence of `process` to detect node.js.

Since the code falls back to an empty string, would it not be better if it the environment variable was only read if permissions were already granted?

iuioiua:
@timfish, I also attempted this in #2519. However, top-level `await` needs to be avoided in `std/node`.
> We can't use top-level-await in std/node to keep the loading of node/global.ts synchronous. - @kt3k

kt3k:
Right, we can't use top-level-await in `std/node` for compatibility issues.

One option to solve this might be, in my view, to introduce `Deno.permissions.querySync` in Deno CLI, and use that before `Deno.env.get("NODE_DEBUG")` call.

iuioiua:
I like the sound of that 👀. It help with other similar situations.

kt3k:
Deno doesn't ask env permission for accessing `NODE_DEBUG` (ref: https://github.com/denoland/deno/pull/15893)

PR #3132, opened 2023-01-25
fix(node): don't unnecessarily ask for env access

lino-levan:
Props to @iuioiua for getting `querySync` landed (https://github.com/denoland/deno/pull/17019).

Closes https://github.com/denoland/deno_std/issues/2446.

### DIFF
--- a/node/internal/util/debuglog.ts
+++ b/node/internal/util/debuglog.ts
@@ -104,7 +104,7 @@ export function debuglog(
 }
 
 let debugEnv;
-if (Deno.permissions.querySync({ name: "env", variable: "NODE_DEBUG" })) {
+if (Deno.permissions.querySync({ name: "env", variable: "NODE_DEBUG" }).state === "granted") {
   debugEnv = Deno.env.get("NODE_DEBUG") ?? "";
 } else {
   debugEnv = "";
--- a/node/internal/util/debuglog.ts
+++ b/node/internal/util/debuglog.ts
@@ -104,7 +104,10 @@ export function debuglog(
 }
 
 let debugEnv;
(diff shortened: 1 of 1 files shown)


lino-levan:
Canary is quite literally not built yet, this *should* pass when it is finished.

iuioiua:
Quick turnaround! 🚀😳

kt3k:
Looks good to me. Thanks for your contribution!

But now this is blocked by Deno Deploy compatibility as it doesn't have `querySync` yet. Let's wait until it updates permissions API

bartlomieju:
> BTW currently we exceptionally skip checking `env` permission for this particular env variable. See [denoland/deno#15893](https://github.com/denoland/deno/pull/15893)
> 
> Maybe we should remove that exceptional handling in CLI after this landed. (@bartlomieju What do you think about this?)

Sure, sounds good to me!

node/internal/util/debuglog.ts
```diff
@@ -104,15 +104,12 @@ export function debuglog(
 }
 
 let debugEnv;
-try {
+if (Deno.permissions.querySync({ name: "env", variable: "NODE_DEBUG" })) {
```
kt3k, line 107: This needs to be:

```suggestion
if (Deno.permissions.querySync({ name: "env", variable: "NODE_DEBUG" }).state === "granted") {
```

lino-levan: Nice catch!

iuioiua:
Can we re-create this PR in the main repo if it's still needed?

lino-levan:
Closing this PR as it is stale.

Issue #1396, opened 2021-10-13
node/dns.ts crashes with UNKNOWN error if `--allow-net` permission is missing


talentlessguy:
**Describe the bug**

node/dns.ts `Server` crashes with `UNKNOWN` error if `--allow-net` is not provided. This case should be handled like in a regular http server.

![image](https://user-images.githubusercontent.com/35937217/137198236-e2b27550-803c-4dd5-a802-7d4869c0a30e.png)

**Steps to Reproduce**

```ts
import { Server } from 'https://deno.land/std/node/net.ts'

const s = new Server()

s.listen(3000)
```

```sh
$ deno run example.js
```

**Expected behavior**

Throw permission error

**Environment**

- OS: Linux
- deno version: 1.15.1
- std version: 0.111.0

PR #3133, opened 2023-01-25
fix(node): throw permisison error instead of unknown error

lino-levan:
Closes https://github.com/denoland/deno_std/issues/1396


kt3k:
Looks reasonable changes. Can you add test cases to exercise those cases?

kt3k (approved):
LGTM

PR #3150, opened 2023-01-30
refactor(node): allow partial availability of Deno.core

bartlomieju:
This change allows to only a part of `Deno.core` namespace to be available,
while still missing some APIs. This is needed for `AsyncLocalStorage` polyfill
in Deno Deploy.

### DIFF
--- a/node/_core.ts
+++ b/node/_core.ts
@@ -7,7 +7,7 @@
 // usages.
 
 // deno-lint-ignore no-explicit-any
-export let core: any;
+let DenoCore: any;
 
 // deno-lint-ignore no-explicit-any
 const { Deno } = globalThis as any;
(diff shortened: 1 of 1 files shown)


bartlomieju:
## Please merge on approval

kt3k (approved):
LGTM

kt3k:
Landing as it looks fine, but which part is actually necessary for implementing AsyncLocalStorage? @bartlomieju

bartlomieju:
> Landing as it looks fine, but which part is actually necessary for implementing AsyncLocalStorage? @bartlomieju

`setPromiseHooks` and `getPromiseDetails`. That said Luca doesn't want to expose every API listed here on Deploy (eg. `op_napi_open`).

kt3k:
Thanks for clarification. Makes sense!

PR #3151, opened 2023-01-31
fix(node): disable worker_threads

crowlKats:
This is a blocker for embedding `std/node` in the Deno binary: https://github.com/denoland/deno/pull/16748

When trying to get rid of Top-Level Await we found out that this module is buggy, because it behaves differently
depending if user import `worker_threads` module inside the worker or not. Disabling it temporarily to enable 
embedding std/node in the binary. We'll revisit it after that PR lands.

### DIFF
--- a/node/worker_threads.ts
+++ b/node/worker_threads.ts
@@ -51,6 +51,7 @@ class _Worker extends EventEmitter {
     if (options?.eval === true) {
       specifier = `data:text/javascript,${specifier}`;
     } else if (typeof specifier === "string") {
+      // @ts-ignore This API is temporarily disabled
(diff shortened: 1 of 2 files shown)


bartlomieju:
@kt3k please take a look

kt3k (approved):
LGTM

kt3k:
I'm in favor of landing this to make progress in https://github.com/denoland/deno/pull/16748

Note for future rework of worker_thread: One solution might be add `workerData` option to [`WorkerOptions.deno`](https://github.com/denoland/deno/pull/16748) and use it for passing `threadId`, `environmentData`, and `workerData` synchronously to `worker_threads.Worker`

PR #3152, opened 2023-01-31
fix: Accept cookies with value containing = character

tlgimenes:
When parsing cookies containing an `=` character in the cookie's value, deno's current library implementation splits this value, thus breaking some systems.

## Example: 
Try parsing the following set-cookie header:
```a=__ofid=b; domain=localhost; expires=Sun, 30-Jul-2023 14:58:46 GMT; path=/```

Using the current implementation of `getSetCookie` function, deno's library would give us:
```
[
  {
    name: "a",
    value: "__ofid",
    domain: "localhost",
    expires: 2023-07-30T14:58:46.000Z,
    path: "/"
  }
]
```

After the proposed changes, we have:
```
[
  {
    name: "a",
    value: "__ofi=b",
    domain: "localhost",
    expires: 2023-07-30T14:58:46.000Z,
    path: "/"
  }
]
```

### CODE REVIEW

File: http/cookie.ts [hunk 1]
```diff
@@ -291,12 +291,10 @@ export function deleteCookie(
...
   const attrs = value
     .split(";")
-    .map((attr) =>
-      attr
-        .trim()
-        .split("=")
-        .map((keyOrValue) => keyOrValue.trim())
-    );
+    .map((attr) => {
+      const [key, ...values] = attr.trim().split("=").map((keyOrValue) => keyOrValue.trim());
```
kt3k, line 295: `.map((keyOrValue) => keyOrValue.trim())`
Is ^--this part correct?

What happens if the key-value pair is like `a=foo= bar`?

kt3k (commented):
Thanks for the fix! Commented on one point.

Also can you add a test case which test the case where the value has `=` characters?

http/cookie.ts
tlgimenes: I think you are right @kt3k!

kt3k (approved):
LGTM. Thanks for the fix!

PR #3153, opened 2023-02-01
fix: make encoding/front_matter work in a browser

kitsonk:
This PR ensures that encoding/front_matter can work in a browser context by accessing the Deno namespace safely.


kitsonk:
This is like pulling on a thread, if something uses the `/node` part of the standard library, it simply will not work in a browser.  There appears to be a lot of work to make `/node` work in a browser without throwing errors everywhere, which means that importing libraries into things like fresh islands is a major pitfall as well as other parts of the std library.  I can't easily fix it and will simply have to look at a non-Deno library for somethings.  Frustrating.

kitsonk:
Actually, raised a PR to address the other issue I was encountering: https://github.com/denoland/deno-gfm/pull/35 which means that it is only the safe access in the encoding/front_matter needed to make that part of std browser compatible, so I got rid of extraneous changes and am re-opening.

PR #3157, opened 2023-02-01
fix(semver): Allow unsetting build metadata

justinmchase:
The main issue is if you pass `""` as the build metadata to the `semver.increment` function it doesn't actually unset the build metadata as expected because of a falsy comparison of the parameter rather than a strict `metadata === undefined`. 

This will now alter the behavior of the increment function to have no changes when you specify `undefined` but to actually set the build metadata to `[]` if you were to pass `""` into the function.

Additionally, there was what I think to be a "bug" discovered [in the constructor](https://github.com/denoland/deno_std/compare/main...justinmchase:deno_std:feature/semver-unsettable-build-metadata?expand=1#diff-8ab0d644da7168c4e5e4629a14418028025bfc0d4ab710f307eb7b5144595231L507-L508) which was making it very hard to test, which is a bug where the `this.raw` field was set before the call to `this.format()`, which ends up retaining a wrong / different version than if you call increment.

Namely if you create a semver with build metadata the build metadata will be in the raw field. If you then _increment_ it, then it will fall out of the raw field even if the build metadata is unchanged.

This happens because in the increment function, it is calling `this.format()` and _then_ setting the raw field to the version but in the constructor it does the reverse. This means the raw will change based on if you've called increment or not.

Also, the whole design of this class probably needs a major revamp to be immutable. Having mutations happen on functions such as `.format()` is really problematic when the object is this complex. If the std lib ever hits `1.0.0` I would strongly recommend putting this chore onto the backlog.

## Before

```ts
const v = semver.parse("1.2.3+0");
v.increment("patch", undefined, "");
console.log(v.format({ style: "full" })); // 1.2.4+0
```

## After

```ts
const v = semver.parse("1.2.3+0");
v.increment("patch", undefined, "");
console.log(v.format({ style: "full" })); // 1.2.4
```
```ts
const v = semver.parse("1.2.3+4");
v.increment("patch");
console.log(v.format({ style: "full" })); // 1.2.4+0
```


kt3k (approved):
LGTM. Thanks for the fix!

> Also, the whole design of this class probably needs a major revamp to be immutable.

This makes sense. Created an issue https://github.com/denoland/deno_std/issues/3164

Issue #3149, opened 2023-01-30
npm compat tracking issue


kt3k:
Tracking issue for npm/node compat issues in `std/node`

## High priority items
- [ ] socket option in tls.connect https://github.com/denoland/deno_std/issues/3047 (affects `npm:mssql`)
- [ ] crypto.Sign https://github.com/denoland/deno_std/issues/3006
- [ ] crypto.createSecretKey https://github.com/denoland/deno_std/issues/3097 (affects `npm:next-auth`, `npm:jsonwebtoken`)
- [ ] `npm:aws-cdk init` throws TypeError https://github.com/denoland/deno_std/issues/2916 (affects `npm:aws-cdk` (windows)

## Medium priority items
- [ ] primisify(execFile) https://github.com/denoland/deno_std/issues/3145 
- [ ] crypto.Cipheriv https://github.com/denoland/deno_std/issues/3118 (affects [tuyapi](https://github.com/codetheweb/tuyapi))
- [ ] crypto.generateKeyPairSync https://github.com/denoland/deno_std/issues/2899 (affects `npm:mineflayer`)
- [ ] missing APIs in `node:os` https://github.com/denoland/deno_std/issues/1436
- [ ] X509Certificate https://github.com/denoland/deno_std/issues/1331
- [ ] http.ClientRequest.setTimeout https://github.com/denoland/deno_std/issues/2476
- ...

#### Note about priorities

The priorities are higher when the issue affects popular npm modules (when used via `npm:` specifier).

---
#### Note to contributors

You can replace the internal node shim implementation by the command:

```shell
DENO_NODE_COMPAT_URL=file:///path/to/your/deno_std/ deno run script.ts
```

By setting `DENO_NODE_COMPAT_URL` env var, Deno uses your local deno_std/node as Node APIs, and you can test, check, or inspect the internals by modifying your local deno_std.

iuioiua:
I'll give `crypto.createSecretKey()` a crack.

PolarETech:
I would like to take up `promisify(execFile)`. I have already implemented `promisify(exec)`, so I think this one will be ready to go soon.

PolarETech:
I have started to try `npm:aws-sdk init`.

Still investigating, but it seems to be an issue with `npm:fs-extra.copy()`.
[...]

PR #3161, opened 2023-02-02
fix(node/child_process): enable promisify(execFile)

### LINKED ISSUE #3145: Promisified `execFile` returns `undefined` in Deno
The following code works in Node 18.1 but not in Deno 1.30:

```js
// exec.mjs
import * as cp from 'node:child_process';
import { promisify } from 'node:util';

const execFile = promisify(cp.execFile);

const child = await execFile('deno', ['--version']);
console.log(child.stdout);
```

With Deno both `child.stdout` and `child.stderr` are undefined
```
deno run --allow-run=deno --allow-read ./exec.mjs

undefined
```

With Node:
```
node exec.mjs
deno 1.30.0 (release, x86_64-pc-windows-msvc)
v8 10.9.194.5
typescript 4.9.4
```

PolarETech:
Fixes #3145
Part of #3149

Refs:
https://github.com/nodejs/node/blob/v18.12.1/lib/child_process.js#L523-L527
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/child_process.d.ts#L1103


kt3k (approved):
LGTM. Nice work!

PR #3162, opened 2023-02-03
test(encoding): re-enable csv stream leak test on non-windows

kt3k:
This PR re-enables the test case `[encoding/csv/stream] cancel CsvStream during iteration does not leak file` on non-windows platforms.


littledivy (approved):
LGTM!

PR #3165, opened 2023-02-03
csv: rename bom to byteOrderMark

timreichen:
- rename `bom` option to `byteOrderMark` for more clarity

### DIFF
--- a/encoding/csv.ts
+++ b/encoding/csv.ts
@@ -234,7 +234,7 @@ export type StringifyOptions = {
    *
    * @default {false}
    */
-  bom?: boolean;
+  byteOrderMark?: boolean;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


kt3k:
I'm not sure renaming this makes it 'clearer'.

I googled `byte order mark csv` and `bom csv` and the latter gives more results. (`byte order mark csv` 600K hits, `bom csv` 3M hits on my side)

Does anyone have any opinion on this?

cc @lionel-rowe @bartlomieju

kt3k:
(Note: #3143 isn't released yet. So this isn't breaking change for now)

bartlomieju:
Looking at some Web APIs, it seems the "bom" abbreviation is usually used: https://developer.mozilla.org/en-US/docs/Web/API/TextDecoderStream/ignoreBOM. From my perspective, it's perfectly fine to keep the current option name.

Issue #2916, opened 2022-11-19
`npm:aws-cdk init` throws TypeError (windows only)


GJZwiers:
```
deno run -A npm:aws-cdk init --language typescript

Applying project template app for typescript
TypeError [ERR_INVALID_ARG_TYPE]" argument must be of type number. Received undefined
    at __node_internal_ (https://deno.land/std@0.165.0/node/internal/validators.mjs:112:15)
    at parseFileMode (https://deno.land/std@0.165.0/node/internal/validators.mjs:49:3)
    at Object.chmod (https://deno.land/std@0.165.0/node/_fs/_fs_chmod.ts:8:12)
    at Object.chmod (file:///C:/Users/GJZwiers/AppData/Local/deno/npm/registry.npmjs.org/node_modules/graceful-fs/polyfills.js:252:19)l-fs/polyfills.js:252:19)                                                                                     /lib/copy/copy.js:131:13)
    at setDestMode (file:///C:/Users/GJZwiers/AppData/Local/deno/npm/registry.npmjs.org/node_modules/fs-extray.js:98:12/lib/copy/copy.js:131:13)                                                                                     fs.js:186:16
    at file:///C:/Users/GJZwiers/AppData/Local/deno/npm/registry.npmjs.org/node_modules/fs-extra/lib/copy/copy.js:98:12
    at file:///C:/Users/GJZwiers/AppData/Local/deno/npm/registry.npmjs.org/node_modules/graceful-fs/graceful-fs.js:186:16
    at https://deno.land/std@0.165.0/node/_fs/_fs_common.ts:230:31
    at https://deno.land/std@0.165.0/node/_fs/_fs_copy.ts:30:49
```

```
deno --version

deno 1.28.1 (release, x86_64-pc-windows-msvc)
v8 10.9.194.1
typescript 4.8.3
```

Edit: Only happens on Windows

cjihrig:
Since this is Windows only, would you care to help debug or fix, @GJZwiers ?

GJZwiers:
I can do some debugging! Is there a way to make my local deno binary/cargo build work with my local copy of `std`, so I can test some changes?

cjihrig:
> Is there a way to make my local deno binary/cargo build work with my local copy of std, so I can test some changes?

Yep! Here is what I use:

`DENO_NODE_COMPAT_URL='file:///path/to/your/std/' /path/to/cli/deno run -A --unstable your_script.js`

GJZwiers:
[...]

PR #3168, opened 2023-02-06
fix(node/fs): chmod function throws unnecessary TypeError on Windows

PolarETech:
Fixes: #2916
Part of: #3149

This PR fixes an error that occurs when an invalid value is unintentionally passed to the `mode` parameter of the `chmod` and `chmodSync` functions on Windows.
This change will make `npm:fs-extra.copy()` work on Windows.

___NOTE: This change must be reverted if `Deno.chmod` and `Deno.chmodSync` were supported on Windows.___

### DIFF
--- a/node/_fs/_fs_chmod.ts
+++ b/node/_fs/_fs_chmod.ts
@@ -12,7 +12,18 @@ export function chmod(
   callback: CallbackWithError,
 ) {
   path = getValidatedPath(path).toString();
-  mode = parseFileMode(mode, "mode");
+
+  try {
+    mode = parseFileMode(mode, "mode");
+  } catch (error) {
+    // TODO(PolarETech): Errors should not be ignored when Deno.chmod is supported on Windows.
+    // https://github.com/denoland/deno_std/issues/2916
+    if (Deno.build.os === "windows") {
+      mode = 0; // set dummy value to avoid type checking error at Deno.chmod
+    } else {
+      throw error;
+    }
+  }
 
   Deno.chmod(pathModule.toNamespacedPath(path), mode).catch((error) => {
     // Ignore NotSupportedError that occurs on windows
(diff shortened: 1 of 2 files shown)


node/_fs/_fs_chmod.ts
```diff
@@ -12,7 +12,18 @@ export function chmod(
...
 ) {
   path = getValidatedPath(path).toString();
-  mode = parseFileMode(mode, "mode");
+
+  try {
+    mode = parseFileMode(mode, "mode");
+  } catch (error) {
+    // TODO(PolarETech): Errors should not be ignored when Deno.chmod is supported on Windows.
+    // https://github.com/denoland/deno_std/issues/2916
+    if (Deno.build.os === "windows") {
```
kt3k, line 21: nit: How about branching by this condition before doing try block?

kt3k (commented):
Mostly looks good to me. Left one comment

node/_fs/_fs_chmod.ts
PolarETech: Then how about the following?

```ts
  path = getValidatedPath(path).toString();

  if (Deno.build.os === "windows" && mode === null) {
    mode = 0;
  }
  mode = parseFileMode(mode, "mode");

  Deno.chmod(pathModule.toNamespacedPath(path), mode).catch((error) => {
```

I realized that we might need to consider the case where `Deno.chmod` supports Windows in the future, and I assume that if `Deno.chmod` works, then `parseFileMode` must be run.
It could be the case that `Deno.chmod` is supported on Windows but the code here is left as is.
To prepare for such a case, I thought `parseFileMode` would run, and if an error occurred, setting `mode` to a dummy value (0 would be safest) would solve the current issue.

However, the concern remains that if a situation arises in the future where `Deno.chmod` supports Windows but `Deno.FileInfo.mode` does not, the permissions will be set to 0 unexpectedly.

kt3k: Thanks for the explanation. Now the current code just makes sense to me.

kt3k (approved):
LGTM

Issue #3164, opened 2023-02-03
refactor(semver): make SemVer class immutable


kt3k:
SemVer class mutates its properties based on the method calls. Especially the mutation of `this.raw` property is problematic.

We should make this class immutable.

See the comment in https://github.com/denoland/deno_std/pull/3157 for more context

justinmchase:
I would be willing to take a shot at this when / if the time is right, it would be a breaking change that would probably justify a major version increment.

bartlomieju:
> I would be willing to take a shot at this when / if the time is right, it would be a breaking change that would probably justify a major version increment.

@justinmchase since `deno_std` is still `0.x.y` it's considered unstable. A breaking change is fine if this is the preferable API.

justinmchase:
Ok well I'll take a crack at it and we can see how we like it.

timreichen:
I wonder if we even need a semver class or should just remove it. Semver parse, format, compare, increment etc. is pretty much stateless, so functions would work well enough.

justinmchase:
No, I don't think you do. I would just export everything as functions. Turn the `class` into a simple `type` or `interface` have the behavior be in the exported functions rather than as members of a class, always return a _new_ object (or string as appropriate) for those functions.

Some of the functions have an `if (semver instanceof SemVer)` check right now, which is problematic. If two different libraries import slightly different versions of semver and pass instsances too each other then they'll have problems possibly. It would just be a lot better in general to get rid of the class.

justinmchase:
Here's a WIP

https://github.com/denoland/deno_std/pull/3169

justinmchase:
Yeah, precisely. There are a lot of weirdnesses like that. The raw can change based on whether or not you've called certain other functions, not good.

Can you try out #3169 and see if it works more like you'd expect? Try out the API and see if there is anything you'd like to be different? I think some community feedback may help the PR progress.

PR #188, opened 2019-02-11
Redesign of http server module

keroxp:
- added `HttpServer` type for general purpose http server
- ported `path-to-regexp` module
- refactored server module 
   - changed `ServerRequest` from class to interface
- added e2e test for server
- let `serve()` cancelable

PR #3169, opened 2023-02-06
BREAKING(semver): rewrite semver

justinmchase:
# Major SemVer Refactor

This library was forked from the nodejs package of the same name [semver](https://www.npmjs.com/package/semver). This is a major breaking change refactor of that library with the intent of rectifying some of the less than desirable aspects of the api that likely accrued due to backwards compatibility concerns.

That being said the _logic_ of the code should be compatible with the previous logic, there is no proposed changes to the expected behaviors other than possibly in a few edge cases which were poorly defined. There are extensive unit tests from before which were instrumental in ensuring correct behavior and even more tests now to further tighten the expected behavior in many scenarios.

### Change Summary

* All classes are now interfaces, such as: `SemVer`, `SemVerComparator`, `SemVerRange`.
* All functions are now non-member functions. 
* Function names are patterened consistently based on the classes they apply to (e.g. `parse`, `parseRange`, `parseComparator`).
* Prefers `undefined` not `null`
* All input objects are treated as immutable and all functions return new instances if a mutation occurs
* Most functions now require a `SemVer` rather than `SemVer | string`
* Everything was in a single file now its split into many smaller files with more targetted functions
* Unit tests are added for every single sub module
* Adds some consistency to function names:
    * `parse`, `parseComparator`, `parseRange`
    * `tryParse`, `tryParseComparator`, `tryParseRange`
    * `isSemVer`, `isSemVerComparator`, `isSemVerRange`
    * etc.
* Prefers to return `SemVer` or throw an error rather than returning `undefined`
* Always compares full version (including pre and build)
* Does a full format by default (including pre and build)
* Every single exported function or type has fully valid jsdoc comments now.

## String vs Number in pre-release bug

There appeared to be a bug in the old code where the following two versions would not compare deterministically:

```
1.0.0-foo <=> 1.0.0-0
```

The two versions would be parsed with the following pre-release values:
```
{ prerelease: [ "foo" ] }
{ prerelease: [ 0 ] }
```
Then during the comparison the following would be run `0 < "foo"` or `"foo" < 0`, which always returns false and therefore the two versions are equal rather than one being greater than the other. The new logic will treat numbers as greater than strings.

## Build comparisons

Versions with build metadata will be sorted alphabetically. Versions with build metadata are considered "greater than" the same semantic version without build metadata.

# Related To

* https://github.com/denoland/deno_std/issues/3164
* https://github.com/denoland/deno_std/pull/3157


semver/mod.ts
```diff
@@ -274,6 +274,14 @@ export interface Options {
   includePrerelease?: boolean;
 }
 
+export interface SemVer {
```
timreichen, line 277: How about calling the interface `Version`?

semver/mod.ts
```diff
@@ -431,351 +439,94 @@ for (let i = 0; i < R; i++) {
...
-    options = {
-      includePrerelease: false,
-    };
-  }
-
-  if (version instanceof SemVer) {
-    return version;
-  }
-
+export function parse(version: string | SemVer): SemVer {
```
timreichen, line 442: I think `parse` should only accept a strings,

justinmchase: Fair enough, that would be cleaner.

semver/mod.ts
justinmchase: Considering its coming from the "semver" module named after the "semver" spec, I felt like this was a pretty nice name while Version is kind of vague.

Also, theoretically there are some other version specs that may be worth supporting eventually such as Python, Nuget, Windows versions etc. It may be worth keeping it the name more specific to give it room.

justinmchase: And by support, I mean there may be some situations where parsing other version types and having a method for normalizing them into SemVer may be a good idea, and possibly formatting back out to those other formats.

iuioiua: I agree with you for the same reasons, @justinmchase.

semver/mod.ts
```diff
@@ -274,6 +274,14 @@ export interface Options {
...
+  major: number;
+  minor: number;
+  patch: number;
+  prerelease: ReadonlyArray<string | number>;
+  build: ReadonlyArray<string>;
+}
+
 // Note: this is the semver.org version of the spec that it implements
 // Not necessarily the package version of this code.
 export const SEMVER_SPEC_VERSION = "2.0.0";
```
timreichen, line 287: Why do we export the spec version? Shouldn‘t a comment be enough?

justinmchase: I have no idea, I'll look to see if there are references and then clean this up. I'm going to have a little room to take this back off the shelf today.

semver/mod.ts
justinmchase: I refactored the whole thing such that the `parse()`, `parseComparator()` and `parseRange()` functions all only take strings now, and all of the other functions take either a `SemVer`, `Comparator` or `Range` argument specifically.

Its hard to understate how much complexity it adds to these apis to have every function attempting to check all possible types.

justinmchase:
@timreichen I did a _big_ wave of updates on the main module here. I'm going to go through all the tests now and see what I screwed up but I'd appreciate another peek at this.

So the biggest change in here is to replace a lot of the comparison code, which seemed to have a lot of duplicate code per behavior. It was also pretty confusing how it was laid out.

So I altered a few things to make this simpler, which I wouldn't mind a sanity check on the approach:

1. Added a MIN = 0.0.0
2. Added a MAX = Infinity.Infinity.Infinity
3. Comparators now have a min() and max() function
    * The min is the lowest possible version that will match the comparator
    * The max is the largest possible version that will match the comparator
4. Intersects functions are now calculated by doing simple comparisons to the min/max of the comparator
5. Outside functions are now calculated by doing simple comparisons to the min/max of the range
6. All comparisons are done using the same version comparison functions

jsejcksn (commented):
Hi @justinmchase. I saw your [comment](https://github.com/denoland/deno_std/issues/3164#issuecomment-1484123639) and took a quick look at this PR — I didn't examine most of the code, but the immutability component is desirable IMO.

On that topic, I left a thought for you to consider regarding types.

semver/format.ts
```diff
@@ -0,0 +1,91 @@
...
+  //   ┌───┴───┐   │
+  //   ├───────────── primary
+  // ┌─┴─┐     │   │
+  // 1.2.3-pre.1+b.1
+  // │ │ │ └─┬─┘ └┬┘
+  // │ │ │   │    └── build
+  // │ │ │   └─────── pre
+  // │ │ └─────────── patch
+  // │ └───────────── minor
+  // └─────────────── major
```
jsejcksn, line 39: This is a nice visual aid 🔍 👀

semver/validity.ts
```diff
@@ -0,0 +1,163 @@
...
+    Array.isArray(build) &&
+    prerelease.every((v) => typeof v === "string" || typeof v === "number") &&
+    prerelease.filter((v) => typeof v === "string").every((v) =>
+      isValidString(v)
+    ) &&
+    prerelease.filter((v) => typeof v === "number").every((v) =>
+      isValidNumber(v)
+    ) &&
+    build.every((v) => typeof v === "string" && isValidString(v));
+  if (!result) console.log({ isSemVer: false, value });
```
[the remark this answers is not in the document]
justinmchase: Sorry that was for debugging and should definitely not be there, you are right.

semver/semver.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * A SemVer object parsed into its consituent parts.
+ */
+export interface SemVer {
+  major: number;
+  minor: number;
+  patch: number;
+  prerelease: ReadonlyArray<string | number>;
+  build: ReadonlyArray<string>;
+}
```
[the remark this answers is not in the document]
justinmchase: I think they should be `readonly` there is not a good reason they were not marked as readonly except that the type evolved from the original class. I'll add the `readonly` keyword.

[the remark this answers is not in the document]
justinmchase: For someone who wants to mutate versions for some reason I think there would be a couple of reasonable options, beyond just using the increment api in this library already:


```ts
const v0 = { major: 0, minor: 1, patch: 0, prerelease: [], build: [] };
v0.major = 5; // its not readonly here
console.log(semver.format(v0));
```

```ts
const v0 = semver.parse("0.1.0");
console.log(semver.format({ ...v0, major: 5 ));
```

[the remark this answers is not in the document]
jsejcksn: > For someone who wants to mutate versions for some reason I think there would be a couple of reasonable options, beyond just using the increment api in this library already:

@justinmchase That's true when the value is not an immutable object returned by this library, but when working with a type that conforms to the fully readonly `SemVer`, it's more complicated — and that's why I suggested providing a helper function for getting a mutable copy. Here's an example to illustrate:

```ts
type SemVer = {
  readonly major: number;
  readonly minor: number;
  readonly patch: number;
  readonly prerelease: ReadonlyArray<string | number>;
  readonly build: ReadonlyArray<string>;
};

declare function parse(version: string): SemVer;

// Get an immutable SemVer from a function in this module:
const semver = parse("v1.1.1");

// Can't mutate:
semver.minor = 2; /* Error
       ~~~~~
Cannot assign to 'minor' because it is a read-only property.(2540) */

// Types used by function below:
type Mutable<T> = { -readonly [K in keyof T]: T[K] };
type MutableSemVer = Mutable<{ [K in keyof SemVer]: Mutable<SemVer[K]> }>;

// Help a user get a mutable copy of a SemVer
// The fn name is just a Rust naming convention (not necessarily a suggestion)
function toMutable(semver: SemVer): MutableSemVer {
  return structuredClone(semver);
}

{ // Easy for user with a helper function and type: 🤩
  const semverMut = toMutable(semver);
  semverMut.minor = 2; // Ok
}

// Complicated for user to do correctly without those: 🥵

{ // Either tedious nested spread syntax (and prone to
  // breakage if the SemVer structure should ever change):
  const semverMut = {
    ...semver,
    prerelease: [...semver.prerelease],
    build: [...semver.build],
  };
  semverMut.minor = 2; // Ok
}

{ // Or complex type manipulation to annotate a correct, narrow return type
  // because the return type of the structuredClone function is `any`:
  const semverMut = structuredClone(semver) as {
    -readonly [K in keyof SemVer]: SemVer[K] extends readonly any[]
      ? SemVer[K][number][]
      : SemVer[K];
  };
  semverMut.minor = 2; // Ok
  //^? const semverMut: { major: number; minor: number; patch: number; prerelease: (string | number)[]; build: string[]; }

  const defaultValue = structuredClone(semver);
      //^? const defaultValue: any
}

```
[playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&target=99&jsx=4&exactOptionalPropertyTypes=true&isolatedModules=true&noImplicitOverride=true&ssl=52&ssc=5&pln=1&pc=1#code/C4TwDgpgBAyhC2A1CAnKBeKBvAUFKKEAhgCYD2AdgDYhTxEBWZKAXFBQK7wBGqA3HgLFy1WvACWFZm049+gwqUo0oYIsADGACxldeKAfkUiVYQoSrEAzhDYAlYcpABBFCiIgAPFeApJAcygAH3Y9VAA+QyElUShuDnEqEntHUVd3Lx8-Cn9InABfARwSCA0qIkIoADMOCg1gcUpVCpsACgA3VCtGijYsgIBKNjgkeRwAenGoAHEIYCgiCihxeHgOYCJuS1gEZDQqlDJ4Bera+p7lpeAtcSs6MhIOSxYcDUofKBt4TrRMNRQ2gAidoARgAdOCQYCBkVJlAAMKLADk8zWG2AthwXx+YIkUl+UAATHwoOMAFRQACibmYgnw9IAfkymThERQpPMiFZuv4rmQoEi8cwkXFSkQODZlvNbidjABaJyqQ6QFCgMGtQkAVgALAAGAZQMnjHATKYAFXAEDuEogJDitBqdQaTV4VDIAHcXqBIFAALLrTaWTxm8IYbBQOXGRUAbQA0pcoABrCAgMhVKBmgC6bDNcczUEKOG90H9Gy2EBGezDpcDEE8WCgcYTydT6crqGzfoD5c87ZQedD+TypqgAAkIFQwCcbWh-HMTmja1A3mBaGmTn2R2atNAqksKER4NAZQwJZyoHYz+xDwFl5ROhRnUtWhz2KUrVYKolaERPhx-HOPg9AMOCOucTTAGQNblq02KoMMuyoEMXZlpYfbYAocwcCgSxZBw9TYba8JuhQECwQgPwwgUJoNnClJcg6zBQDOUDuuI1wnDuk6oKcToXIsdrFmwgB8G4AlLuCG8FAfHBKClmGkHQZY5HfEhUQyaWuKSExmDEqSUwAPKJtRI7wkcYBUOIGjqLa1RMSxkFQOQd5uKUwAqGx1xkOsUCeTYImAK67NF6VS7E7mgGIkI0Epvj4NlWGYwifCAj5EAAHlArQCUqlDQJBghwtwiiJkQc7LOm1zQOheEEZUVhaF5SRQBAPzLloixzkMEnvPM6neZguD0lAYLDTJAA0dJKqgE7WLYjbDWCMlgglFgzZm42DfEiTJHNI0UagYKbUka2CIW+C9cAmn4mGulwoZxm0QZaBvPA5kQOlxZ0Is4hgE86gXA5iwctZJxvC59SjdeNLukIwDYVclr5VMvBWTaPk7jDcM+ZaUDrhVny+PhsOECQxE5bx4FLDKAAGiwgFTLz4JJ0l7bJfX4yghOESTJFkTJBpchhg2RqkKhNpISYpuufadn2eZNalGIUCQdxRrEtPRpmE34AA-Dsoz9rGmbRrI+hG5rg34Ah+t5lEp2fCzGlCgSN0GUZ+CTAAetrJqM91jkQFU4pUMAiBEFQHDQJg1VE0RPPKZRUSDZ73v5DgQA)

semver/mod.ts
jsejcksn: @justinmchase Did you decide against the string-only argument — or just haven't modified that part yet?

https://github.com/denoland/deno_std/pull/3169/files#diff-581da22f7b2e7456281ca11fb4976d2ab48b8f5e89063da36861efe16e1e2a66R182

https://github.com/denoland/deno_std/blob/d5feeab5cae4f49b709d05d2e60ecfe276d416ad/semver/parse.ts#L182

timreichen:
I am not sure about this, but would it make sense to have utility functions like lt gt, satisfies etc. Accept only semver objects? That way no type check is needed and parse is not called internally which i think is better.

justinmchase:
@timreichen I definitely agree, most of the basic operators are all like that now.

#### lt
https://github.com/denoland/deno_std/pull/3169/files#diff-5a3730dfb1e103261a88e803ead03990911c01d5136d83df8c6eda08d6996ca6R11-R16

#### gt
https://github.com/denoland/deno_std/pull/3169/files#diff-f1e39df0ee27b17153625a47d7072ac0ec77b7ef05bf555fa21aba77682c3419R11-R16

Satisfies on the other hand is still `string | SemVer` because:

1. Its basically just a wrapper over `rangeTest`
2. Its a big commonly used api that I didn't want to break completely or make people add a bunch of lines just to get the one function call.
3. If you already have a `SemVer` just use `rangeTest`

#### satisfies
https://github.com/denoland/deno_std/pull/3169/files#diff-8ab0d644da7168c4e5e4629a14418028025bfc0d4ab710f307eb7b5144595231R284

semver/mod.ts
```diff
@@ -231,1720 +230,91 @@
...
-}
-
-/** Less than or equal comparison */
-export function lte(
-  v1: string | SemVer,
-  v2: string | SemVer,
-  options?: Options,
-): boolean {
-  return compare(v1, v2, options) <= 0;
+export function valid(value: string | undefined): string | undefined {
```
timreichen, line 272: Not sure if we should provide this function in the long run. I think the user should use a try-catch block on parse if he wants to check. Maybe add a deprecation note?

semver/mod.ts
```diff
@@ -231,1720 +230,91 @@
...
+ * @returns The maximum version for the given range
  */
-export function prerelease(
-  version: string | SemVer,
-  options?: Options,
-): ReadonlyArray<string | number> | null {
-  const parsed = parse(version, options);
-  return parsed && parsed.prerelease.length ? parsed.prerelease : null;
+export function maxVersion(
+  value: string | SemVerComparator,
```
timreichen, line 314: maybe remove string parameter type for `value`.

justinmchase:
Alright I changed all of the compatibility functions in the module to be more strict about their inputs and marked them as deprecated, gave docs about what to use instead. I also updated the example code to reflect those changes and I moved the formatting diagram to its own section in the main part of the documentation.

semver/format.ts
justinmchase: I moved it to a formatting section in the main module documentation.

semver/semver.ts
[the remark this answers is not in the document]
justinmchase: Yeah I think you're right that its kind of annoying that its immutable, and frankly none of these functions actually care about mutability anymore since they're all stateless. So I think I'll just remove the `readonly` keywords, none of these functions will mutate your objects but if you want to go right ahead.

semver/range_test.ts
```diff
@@ -92,298 +97,382 @@ Deno.test("range", function () {
...
+        const range = parseRange(r);
+        const s = parse(v);
+        const found = rangeTest(s, range);
+        assert(!found);
+      });
+    }
+  },
 });
+
+// Deno.test("unlockedPrereleaseRange", function () {
```
kt3k, line 274: Should these commented out test cases be removed?

semver/validity.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+/**
+ * Checks to see if the value is a valid Operator string.
+ *
+ * Adds a type assertion if true.
+ * @param value The value to check
+ * @returns True if the value is a valid Operator string otherwise false.
+ */
+export function isValidOperator(value: unknown): value is Operator {
```
kt3k, line 106: I think this and the below 2 functions have little value to the users. How about making them private/unexported?

semver/semver.ts
```diff
@@ -0,0 +1,61 @@
...
+export const INVALID: SemVer = {
+  major: Number.NaN,
+  minor: Number.NaN,
+  patch: Number.NaN,
+  prerelease: [],
+  build: [],
+};
+
+/**
+ * ANY is a sentinal value used by some range calculations. It is not a valid
```
kt3k, line 51:
```suggestion
 * ANY is a sentinel value used by some range calculations. It is not a valid
```

semver/range.ts
```diff
@@ -0,0 +1,111 @@
...
+    return r1.ranges.some((r11) => {
+      return r00.every((c0) => {
+        return r11.every((c1) => comparatorIntersects(c0, c1));
+      });
+    });
+  });
+}
+
+/**
+ * Returns the highest version in the list that satisfies the range, or `null`
```
kt3k, line 82:
```suggestion
 * Returns the highest version in the list that satisfies the range, or `undefined`
```

semver/range.ts
```diff
@@ -0,0 +1,111 @@
...
+  versions: SemVer[],
+  range: SemVerRange,
+): SemVer | undefined {
+  const satisfying = versions.filter((v) => rangeTest(v, range));
+  const sorted = sort(satisfying);
+  return sorted.pop();
+}
+
+/**
+ * Returns the lowest version in the list that satisfies the range, or `null` if
```
kt3k, line 98:
```suggestion
 * Returns the lowest version in the list that satisfies the range, or `undefined` if
```

kt3k (commented):
Sorry for the delay in review. Left several comments.

kt3k:
I wonder if it's good idea to remove `string |` overload from the common APIs like `rangeTest`, `gt`, `lt`. Requiring `SemVer` input always increases the cognitive cost and typing amount even for very casual usages like one in `_tools/check_deprecation.ts`. What do people think about it?

kt3k:
Looks like some APIs are immediately breaking without deprecation notice.

kt3k:
(closed once by mistake, and reopened)

semver/validity.ts
```diff
@@ -0,0 +1,162 @@
...
+  return (
+    typeof value === "number" &&
+    !isNaN(value) && (
+      value === Number.POSITIVE_INFINITY ||
+      value === Number.NEGATIVE_INFINITY ||
+      (
+        value <= Number.MAX_SAFE_INTEGER &&
+        value >= 0
+      )
+    )
```
jsejcksn, line 143: Nit:

After eliminating `NaN` as a possibility (see [`Number.isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN) vs [global `isNaN()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/isNaN)), testing for `±∞` can be collapsed into a single functional check by negating [`Number.isFinite()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isFinite)

```suggestion
    !Number.isNaN(value) && (
      !Number.isFinite(value) ||
      (0 <= value && value <= Number.MAX_SAFE_INTEGER)
    )
```

jsejcksn:
> I wonder if it's good idea to remove `string |` overload from the common APIs like `rangeTest`, `gt`, `lt`. Requiring `SemVer` input always increases the cognitive cost and typing amount even for very casual usages like one in `_tools/check_deprecation.ts`. What do people think about it?

@kt3k I agree that the cognitive cost is higher, but I also think it's useful (important?) for the user to acknowledge the fact that parsing is needed for these operations: that more than string comparison is happening.

Also — because parsing is fallible — allowing strings would make all of the functions parse internally and therefore fallible as well... when they don't need to be otherwise. [Fallibility is currently difficult/impossible to signal in the type system](https://github.com/microsoft/TypeScript/issues/13219) (e.g. [`never` is absorbed in unions](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&removeComments=true&target=99&jsx=4&exactOptionalPropertyTypes=true&inlineSourceMap=true&inlineSources=true&isolatedModules=true&noImplicitOverride=true#code/GYVwdgxgLglg9mABAWwIYE8BGBTA4tqAEVSmwAoBKALkWNMQB9ExsA3bAJ0QG8AoRRDGCIyAWRIALAHQdUYACZxklRAB5EABikBWCog4EQHJCwDutEuQoBufoigSOcc2cQBRDk45kARADk4KHsJGABnexhkbB8bXgBfXl4AeiTEeUtECDgAG2zUAAdQ7HCoOEQAKxBQoLpsETBAi3omFnYOakRAXg3AND3eLLBqtIyAXhQMHHwiS0pbAQEUgD0Afl4gA)), and marking each of these functions that way in the JSDoc might not be simple to explain or understand — there'd be a need to explain that each one "is fallible if any string arguments are provided, but not if `SemVer` arguments are exclusively provided."

justinmchase:
> I wonder if it's good idea to remove string | overload from the common APIs like rangeTest, gt, lt. Requiring SemVer input always increases the cognitive cost and typing amount even for very casual usages like one in _tools/check_deprecation.ts. What do people think about it?

I actually kind of agree, which is why I tried to preserve some of the more friendly named functions as `string | SemVer` while keeping all more heavy functions as `SemVer` specifically...

But I will say there are two problems with the `string | SemVer` inputs:

1. If they are all that way then it gets confusing about who is supposed to do the parsing and you end up with a lot of duplicated code.
2. Parsing can fail and the mechanism for identifying those errors with parsing is exception throwing currently and that actually causes a lot of confusion with the apis because some were just eating the errors on a failed parse and returning `undefined`, leaving it difficult to tell what the errors were.

So separating the parsing and error handling of the parsing from the operations makes all of that much more explicit and controllable for the callers.

Yet... It seems entirely possible that a lot of users are totally casual and they really don't care. So from a "UX" perspective of a library interface, I think it would be pretty reasonable to strike a balance. Make it _possible_ to be very explicit and handle errors with high quality messages and deterministic behavior... but also have some low barrier to entry vanity functions.

semver/operators/lt_test.ts
```diff
@@ -0,0 +1,42 @@
...
+      ["2.0.0", "1.2.3"],
+      ["1.2.3", "1.2.3-asdf"],
+      ["1.2.3", "1.2.3-4"],
+      ["1.2.3", "1.2.3-4-foo"],
+      ["1.2.3-5", "1.2.3-5-foo"], // numbers > strings, `5-foo` is a string not a number
+      ["1.2.3-5", "1.2.3-4"],
+      ["1.2.3-5-foo", "1.2.3-5-Foo"],
+      ["3.0.0", "2.7.2+asdf"],
+      ["1.2.3-a.10", "1.2.3-a.5"],
+      ["1.2.3-a.5", "1.2.3-a.b"],
```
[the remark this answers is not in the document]
justinmchase: Interesting, good catch.

Here is the place in code actually doing this comparison:
https://github.com/justinmchase/deno_std/blob/0421330f380d21486b239d6b813b2e6344599bb4/semver/operators/compare.ts#L73

```ts
    } else if (typeof a === "string" && typeof b === "number") {
      // string < number
      return -1;
    } else if (typeof a === "number" && typeof b === "string") {
      // number > string
      return 1;
    } else if (a < b) {
```

Which is indeed the _opposite_ of what the spec is saying. If I recall correctly and look in the original code, this was actually an untested situation and it was simply doing `a < b` and it was actually a bug because...

```ts
"a" < 1 // false
1 < "a" // false
```

Which then I either misinterpretted or inverted the spec somewhere along the way.

semver/range_test.ts
justinmchase: No, it looks like I missed them and I need to fix em up and get them running again.

justinmchase:
Ok so I re-enabled those overlooked range tests and that was pretty nice because it caught a few small things and really helped cleanup a few details.

One big issue caught by those tests is that range intersection can only be true on satisfiable ranges, which was a miss before these tests.

## Possible Bug Found

One test was inverted, because I think there was a bug in the original code and now its more correct but please double check my logic.

`^` should match any version up to but excluding the next major version. A pre-release version such as `2.0.0-rc1` is _less than_ `2.0.0` but it should also be considered out of range of the `^1.0.0` version range.

Previously this test was returning true, I believe incorrectly.

```
rangeTest("^1.0.0", "2.0.0-rc1") == false
```

kt3k:
@jsejcksn @justinmchase OK. Now I'm convinced we should go without `string | SemVer` overload. Thanks for your inputs.

justinmchase:
I just wanted to gentle ping on this PR and say that I have also created this related tool and published it as OSS through my company:

https://github.com/optum/semver-cli

I plan to update to this new version of semver once merged and I also internally have a GitHub action as well as an App, which I hope to make available publicly as well over time. All related to managing semantic versions across projects.

I will also add, that it can bump versions in README's and code files.

justinmchase:
All conversations should be resolved and I don't know of any further work that needs to be done here. How are we feeling about this one?

jsejcksn (approved):

kt3k:
I think the structure of having `semver/operator` subdirectory isn't aligned to the rest of the standard library. How about moving all files directly under `semver/`?

semver/range.ts
```diff
@@ -0,0 +1,131 @@
...
+  //[ [1 and 2] or [2 and 3] ] = [ 2 or 3 ] = 2
+  return sort(
+    range.ranges.map((r) =>
+      sort(r.filter((c) => rangeTest(c.min, range)).map((c) => c.min)).pop()!
+    ).filter((v) => v),
+  ).shift() ?? INVALID;
+}
+
+/**
+ * The maximum valid SemVer for a given range or INVALID
```
kt3k, line 35: Why INVALID instead of `undefined`?

semver/semver.ts
```diff
@@ -0,0 +1,61 @@
...
+  minor: 0,
+  patch: 0,
+  prerelease: [],
+  build: [],
+};
+
+/**
+ * A sentinel value used to denote an invalid SemVer object
+ * which may be the result of impossible ranges or comparator operations.
+ * @example `eq(parseRange("<0.0.0"), INVALID)`
```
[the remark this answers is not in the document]
justinmchase: ok thanks I'll check it out, I think I fixed it once but then there were some adjustments.

semver/range.ts
justinmchase: good question, I'll think about it and either alter it or give you a good answer.

justinmchase:
@kt3k 

* I can move all the files into the root, it just seemed like a lot of files in one dir, and that felt like a relevant way to structure it. But technically speaking there is no reason why it can't all be in one big folder.
* Having some/many/all of the apis being `string | SemVer` is also achievable, we just have gone back and forth on it a few times and I'd really like to make sure we have a consensus before changing it again.

jsejcksn:
> While I agree to remove `string | SemVer` overload in the final shape, we usually don't make immediate breaking changes in single verion bump. I wonder if it's possible to make them deprecated first and then removed after several version.

@kt3k This is a concern you've [shared before](https://github.com/denoland/deno_std/pull/1984#issuecomment-1057769086). I want to mention this again here (as it seems especially on-topic contextually):

- `deno.land/std` is currently versioned independently from the CLI:

    From [README#Releases](https://github.com/denoland/deno_std/blob/0.185.0/README.md#releases):

    > ## Releases
    > 
    > Standard library is currently tagged independently of Deno version. This will change once the library is stabilized.
    > 
    > To check compatibility of different version of standard library with Deno CLI see [this list](https://raw.githubusercontent.com/denoland/dotland/main/versions.json).

- According to the [SemVer spec](https://semver.org/spec/v2.0.0.html#spec-item-4), item 4 describes versions in the `0.y.z` range:
    > Major version zero (`0.y.z`) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

    Which means that every release before `1.0.0` should be considered a breaking change from the previous release.

So, regardless of which version of Deno is being used, consumers should be prepared to refactor every time they upgrade a `std` dependency (until the library reaches `1.0.0`).

If this is not the intention of `std`, then perhaps a discussion is needed to determine how to update the versioning strategy.

kt3k:
That's technically true ragarding semver spec, but in practice we had backlash from the community about certain breaking change (https://github.com/denoland/deno_std/pull/188 ) in the past and we ended up reverting it (https://github.com/denoland/deno_std/pull/206 ) (Deno was still 0.x at that time)

I think the users still expect some level of stability even at 0.x versions.

kt3k:
The following expressions throw with confusing error message: `Error: Comparison against non-numbers`. Are these expected behaviors?

- `eq(parse("1.1.1"), ANY)`
- `rangeMax(parseRange(">=0.0.0"))`

semver/compare.ts
```diff
@@ -0,0 +1,92 @@
...
+): 1 | 0 | -1 {
+  if (s0 === s1) return 0;
+  return (
+    compareNumber(s0.major, s1.major) ||
+    compareNumber(s0.minor, s1.minor) ||
+    compareNumber(s0.patch, s1.patch) ||
+    checkIdentifier(s0.prerelease, s1.prerelease) ||
+    compareIdentifier(s0.prerelease, s1.prerelease) ||
+    checkIdentifier(s1.build, s0.build) ||
+    compareIdentifier(s0.build, s1.build)
```
kt3k, line 27: Comparing `build` by default feels wrong to me as the spec says the below

> Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence.

https://semver.org/#spec-item-10

I think we should keep `compareBuild` API to support the case where the users want to compare `build` as well

kt3k:
Restored deprecated overloads (with `@deprecated` tag) to the below APIs:
- `compare`
- `difference`
- `eq`
- `gt`
- `gte`
- `gtr`
- `lt`
- `lte`
- `ltr`
- `neq`
- `outside`
- `rcompare`

Restored the below APIs because they seem still necessary:
- `compareBuild`

Restored the below APIs with `@deprecated` tag only for compatibility purpose.
- `compareIdentifiers`

Remaining:
- [ ] `cmp`
- [ ] `increment`
- [ ] `intersects`
- [ ] `major`
- [ ] `maxSatisfying`
- [ ] `minor`
- [ ] `minSatisfying`
- [ ] `minVersion`
- [ ] `parse`
- [ ] `patch`
- [ ] `prerelease`
- [ ] `rcompareIdentifiers`
- [ ] `rsort`
- [ ] `satisfies`
- [ ] `sort`
- [ ] `toComparators`
- [ ] `valid`
- [ ] `validRange`

semver/format.ts
```diff
@@ -0,0 +1,83 @@
...
+ * @param comparator
+ * @returns A string representation of the comparator
+ */
+export function comparatorFormat(comparator: SemVerComparator) {
+  const { semver, operator } = comparator;
+  return `${operator}${format(semver)}`;
+}
+
+/**
+ * Formats teh range into a string
```
mxdvl, line 63:
```suggestion
 * Formats the range into a string
```

PR #3173, opened 2023-02-08
fix(encoding/yaml): avoid prototype pollution in Node.js and Browser

ayame113:
Similar to #3077, try to avoid prototype pollution of the Yaml parser when running in browsers and Node.js.

Related link: Changes when preventing prototype pollution in js-yaml -> https://github.com/nodeca/js-yaml/commit/a00312147f287a1703bf6b7e104d27c8a0caacee


kt3k (approved):
LGTM. Thanks!

Issue #3171, opened 2023-02-06
Ensure "node/" directory is "standalone"


bartlomieju:
With the plans to move "std/node" into the the `deno` repository and embed it directly into the binary (https://github.com/denoland/deno/pull/16748), we need to ensure that "node/" directory is standalone - ie. that it doesn't depend on any other modules from the standard library. This is gonna be tricky as I know that eg. "crypto" relies on the "crypto/" directory.

If it's not possible to remove these dependencies easily, then we should at least compile a list of dependencies that will need to be brought over to the CLI (or replace them with other APIs available in the `deno` repository).

lino-levan:
This sounds very tricky. I know that this was originally the way it was, but maybe this would be easier if we moved everything over to the main repository? Or maybe take advantage of git submodules?

bartlomieju:
> This sounds very tricky. I know that this was originally the way it was, but maybe this would be easier if we moved everything over to the main repository? Or maybe take advantage of git submodules?

We have no way of including WASM files in the snapshot.

lino-levan:
Ah I understand the problem now. There really are no good answers here. I'm still a little bit hesitant to support moving `/node` into the main Deno repository but I understand the reasoning for it.

bartlomieju:
Following script produces a list of files in `std/node/` that reach to other modules in the standard library:

```ts
import * as colors from "https://deno.land/std@0.177.0/fmt/colors.ts";

const FROM_RE = /from\s+["']([^"']+)["']/g;

const STD_ROOT = new URL("../deno/test_util/std/", import.meta.url);
const STD_NODE_ROOT = new URL("./node/", STD_ROOT);
await readDir(STD_NODE_ROOT, STD_NODE_ROOT);

function relativeToRoot(root: URL, url: URL): string {
  return url.pathname.substr(root.pathname.length);
}

async function readDir(stdNodeUrl: URL, currentUrl: URL) {
  for await (const entry of Deno.readDir(currentUrl)) {
    if (entry.name == "integrationtest") {
      continue;
    } else if (entry.name.endsWith("_test.ts")) {
      continue;
    }

    if (entry.isDirectory) {
      await readDir(stdNodeUrl, new URL(`./${entry.name}/`, currentUrl));
      continue;
    }

    const fileUrl = new URL(entry.name, currentUrl);
    const filePath = fileUrl.pathname;
    // console.log(fileUrl.pathname);
[...]

PR #3174, opened 2023-02-09
refactor(node): move std/_util usages to node/_util

kt3k:
part of #3171 

This PR moves the references to `std/_util` to `node/_util` (by creating copies of `_util/asserts.ts` and `_util/os.ts` in `node/_util`)

(Note: this reduces the number of outside imports from `std/node` to `std` from 70 to 40.)


bartlomieju (approved):
LGTM, considering collapsing added files into a single file.

kt3k:
> considering collapsing added files into a single file.

I can't think of a good name for the combined version of these 2. Let's iterate after moved to CLI repo.

PR #3176, opened 2023-02-09
refactor(node): replace std/async usages with its own version

kt3k:
This PR replaces `std/async` usages from `std/node` with its own version of them.

part of #3171


bartlomieju (approved):
LGTM, though I would suggest to put both `deferred` and `delay` in the same file. The fewer files we need to move the better

iuioiua:
I think we should add a comment that this code was vendored from `std`. In the future, a question one might have is where this code came from. Also, this'd help copy any bug fixes applied from `std` to the new node dir.

kt3k:
Moved the APIs to `node/_util/async.ts` and added comment about the source.

PR #3177, opened 2023-02-09
refactor(node): replace std/fmt usages in std/node

kt3k:
This PR replaces `std/fmt` usages in `std/node`.

part of #3171


bartlomieju (approved):
LGTM

PR #3179, opened 2023-02-09
chore(crypto): remove deprecated `util.ts`

iuioiua:
This change removes `crypto/util.ts`, which is deprecated.


kt3k (approved):
LGTM. Thanks!

PR #3180, opened 2023-02-09
chore(wasmbuild): upgrade to 0.10.3

iuioiua:
This change upgrades `wasmbuild` from 0.10.2 to 0.10.3, due to the fix in denoland/wasmbuild#69 and rebuilds `crypto` and `encoding/varint` source files.


kt3k (approved):
LGTM. Thanks!

PR #3183, opened 2023-02-10
test(http): make file_server_test less flaky

### LINKED ISSUE #3195: test: `http/file_server_test.ts` is flaky on windows
It often fails with `error: PermissionDenied: Access is denied. (os error 5)`

kt3k:
'Access is denied' error is sometimes thrown at `child.kill("SIGKILL")` call only on windows. This PR tries to reduce the probability of flaky test failure caused by it.

closes #3195


kt3k:
Look like it's almost non-flaky now.

bartlomieju (approved):

PR #3185, opened 2023-02-11
test(node): enable more child_process testing 

PolarETech:
Part of: #911

During recent PR work, I found some tests that have already passed and would like to add them.
I also modified some existing tests (e.g., remove the `--unstable` flag, which is no longer required).


PolarETech:
CI fails, but it does not seem to be due to this change.

kt3k (approved):
LGTM. Thanks!

PR #3186, opened 2023-02-11
feat(http/file_server): etag value falls back to `DENO_DEPLOYMENT_ID` if `fileInfo.mtime` is not available

### LINKED ISSUE #3182: Feature request: use `DENO_DEPLOYMENT_ID` as Etag in fileserver.ts
Currently etags are not supported when using `http/file_server.ts` within deno deploy. (ref: https://github.com/denoland/deploy_feedback/issues/334)
This is because deno deploy does not support file modification timestamps.

Until deno deploy supports file timestamps, how about having the etag value fall back to `DENO_DEPLOYMENT_ID` if timestamps aren't available?

(Since the `DENO_DEPLOYMENT_ID` changes with each new deployment, I thought I could just use it as the etag value.)

ayame113:
close #3182 

some questions:

- Now always fallback to `DENO_DEPLOYMENT_ID`. Should this be optionally configurable?
    ```ts
    interface ServeDirOptions {
      etagFallbackToDenoDeploymentId?: boolean; // default is true
    }
    // or
    interface ServeDirOptions {
      defaultEtag?: string; // default value is Deno.env.get("DENO_DEPLOYMENT_ID")
    }
    ````

Note: This PR is not necessary if https://github.com/denoland/deploy_feedback/issues/334 is expected to be resolved soon. If so, feel free to close this.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -24,13 +24,13 @@ interface EntryInfo {
   name: string;
 }
 
+const encoder = new TextEncoder();
 const envPermissionStatus =
   Deno.permissions.querySync?.({ name: "env", variable: "DENO_DEPLOYMENT_ID" })
     .state ?? "granted"; // for deno deploy
-const DENO_DEPLOYMENT_ID = envPermissionStatus === "granted"
-  ? Deno.env.get("DENO_DEPLOYMENT_ID")
+const hashedDenoDeploymentId = envPermissionStatus === "granted"
+  ? toHashString(encoder.encode(Deno.env.get("DENO_DEPLOYMENT_ID")))
   : undefined;
-const encoder = new TextEncoder();
 
 function modeToString(isDir: boolean, maybeMode: number | null): string {
   const modeMap = ["---", "--x", "-w-", "-wx", "r--", "r-x", "rw-", "rwx"];
@@ -132,7 +132,7 @@ export async function serveFile(
         `${fileInfo.mtime.toJSON()}${fileInfo.size}`,
       ),
     )
-    : DENO_DEPLOYMENT_ID;
+    : hashedDenoDeploymentId;
 
   // Set last modified header if last modification timestamp is available
   if (fileInfo.mtime) {
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -150,12 +150,11 @@ export async function serveFile(
     const ifNoneMatch = req.headers.get("if-none-match");
     const ifModifiedSince = req.headers.get("if-modified-since");
     if (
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -24,6 +24,13 @@ interface EntryInfo {
   name: string;
 }
 
+const envPermissionStatus = await Deno.permissions.query({
```
kt3k, line 27: Top-level-await often causes unexpected behavior to dependents. I'd suggest we should use querySync instead. (And querySync is now unavailable in Deploy. I think we should do something like the below for now: https://github.com/denoland/deno_std/pull/3178/files#diff-a65e7827e30b4089194fd08ed758f7344da9052d00a6d45c41ebae713c959522R37-R39

ayame113: Makes sense... updated!

kt3k (approved):
LGTM

ayame113:
Thank you for your review!

http/file_server.ts
```diff
@@ -118,28 +124,34 @@ export async function serveFile(
...
+  const etag = fileInfo.mtime
+    ? toHashString(
       await createHash(
         etagAlgorithm ?? "FNV32A",
-        `${lastModified.toJSON()}${fileInfo.size}`,
+        `${fileInfo.mtime.toJSON()}${fileInfo.size}`,
       ),
-    );
+    )
+    : DENO_DEPLOYMENT_ID;
```
lucacasonato, line 135: Please also hash the deployment id.

ayame113: I changed it like that. For reference, can you tell me why I should do that?

kt3k: I guess the purpose is not to reveal `DENO_DEPLOYMENT_ID` to the visitors. Exposing `DENO_DEPLOYMENT_ID` is not directly exploitable/abusable, but if the attacker keeps memorizing them and find vulnerability in some old version of the deployment, then they can use old `DENO_DEPLOYMENT_ID` to invoke 'old' version of deployment by accessing `https://project-name--<that DENO_DEPLOYMENT_ID>.deno.dev/`

The current way of transforming `DENO_DEPLOYMENT_ID` is reversible. I think we should apply `createHash` before pass it to `toHashString`

ayame113: It makes a lot of sense. thank you!

Another question. The hashing algorithm for the deployment id is currently fixed to FNV32A. Should the `etagAlgorithm` optionally passed by the user be honored here?

kt3k: I think it's ok with hard coded `FNV32A`. Thanks for updating!

kt3k (approved):
LGTM again!

PR #3184, opened 2023-02-10
feat(encoding/csv/streaming): Added `skipFirstRow` and `columns` options

ayame113:
close #2559

This PR adds the ability to parse csv with headers to the WebStream based csv parser.

The additional options (`skipFirstRow` and `columns`) work exactly the same as when specify them in the existing synchronous csv parser.

<details>
<summary>how it works</summary>

```ts
import { parse } from "./csv.ts";
import { CsvStream } from "./csv/stream.ts";
import { readableStreamFromIterable } from "../streams/readable_stream_from_iterable.ts";

const csv = `a,b,c
d,e,f
`;

// It shows how each option works.
const readable1 = readableStreamFromIterable(csv).pipeThrough(new CsvStream({ skipFirstRow: true }));
for await (const data of readable1) {
  console.log(data); // { a: "d", b: "e", c: "f" }
}

const readable2 = readableStreamFromIterable(csv).pipeThrough(new CsvStream({ columns: ["aaa", "bbb", "ccc"] }));
for await (const data of readable2) {
  console.log(data); // { aaa: "a", bbb: "b", ccc: "c" }, { aaa: "d", bbb: "e", ccc: "f" }
}

const readable3 = readableStreamFromIterable(csv).pipeThrough(new CsvStream({ skipFirstRow: true, columns: ["aaa", "bbb", "ccc"] }));
for await (const data of readable3) {
  console.log(data); // { aaa: "d", bbb: "e", ccc: "f" }
}

// The output is the same as the existing synchronous parse function.
console.log(parse(csv, { skipFirstRow: true })); // [ { a: "d", b: "e", c: "f" } ]
console.log(parse(csv, { columns: ["aaa", "bbb", "ccc"] })); // [ { aaa: "a", bbb: "b", ccc: "c" }, { aaa: "d", bbb: "e", ccc: "f" } ]
console.log(parse(csv, { skipFirstRow: true, columns: ["aaa", "bbb", "ccc"] })); // [ { aaa: "d", bbb: "e", ccc: "f" } ]
```

</details>

PR #3187, opened 2023-02-11
fix: change `BigInt` type to `bigint` type

ayame113:
Fix the currently failing CI. (ref: https://github.com/denoland/deno_std/pull/3184#issuecomment-1426354288)
The cause of CI failures is that BigInt types are now reported as lint errors.(https://github.com/denoland/deno_lint/pull/1124)
This PR fixes CI by replacing BigInt with bigint.


kt3k (approved):
LGTM. Thanks!

PR #3189, opened 2023-02-11
feat(streams/delimiter_stream): add disposition option

denizdogan:
Adds the option to include the delimiter in the output chunks, either as a suffix in the found chunk or as a prefix to the subsequent chunk. The default behavior is preserved, but can be explicitly enabled by passing `{ disposition: "discard" }`.

Let me know what you think!

### DIFF
--- a/streams/delimiter_stream.ts
+++ b/streams/delimiter_stream.ts
@@ -3,6 +3,12 @@
 import { BytesList } from "../bytes/bytes_list.ts";
 import { createLPS } from "./_common.ts";
 
+type DelimiterDisposition =
+  | "suffix" // delimiter included in chunk
+  | "prefix" // delimiter included in subsequent chunk
+  | "discard" // delimiter discarded
+;
+
 /** Transform a stream into a stream where each chunk is divided by a given delimiter.
  *
  * ```ts
@@ -20,8 +26,12 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
   #matchIndex = 0;
   #delimLen: number;
   #delimLPS: Uint8Array;
+  #disp?: DelimiterDisposition;
 
-  constructor(delimiter: Uint8Array) {
+  constructor(
+    delimiter: Uint8Array,
+    options?: { disposition?: DelimiterDisposition },
+  ) {
     super({
       transform: (chunk, controller) => {
         this.#handle(chunk, controller);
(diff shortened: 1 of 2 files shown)


streams/delimiter_stream.ts
```diff
@@ -3,15 +3,48 @@
 import { BytesList } from "../bytes/bytes_list.ts";
 import { createLPS } from "./_common.ts";
 
-/** Transform a stream into a stream where each chunk is divided by a given delimiter.
+/** Disposition of the delimiter. */
+type DelimiterDisposition =
```
crowlKats, line 7:
```suggestion
export type DelimiterDisposition =
```

streams/delimiter_stream.ts
```diff
@@ -20,8 +53,12 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
   #matchIndex = 0;
   #delimLen: number;
   #delimLPS: Uint8Array;
+  #disp?: DelimiterDisposition;
 
-  constructor(delimiter: Uint8Array) {
+  constructor(
+    delimiter: Uint8Array,
+    options?: { disposition?: DelimiterDisposition },
```
crowlKats, line 60: I'd rather prefer if the type declaration was an interface instead of inlined

streams/delimiter_stream.ts
denizdogan: Fixed!

streams/delimiter_stream.ts
denizdogan: Fixed!

denizdogan:
I don't think the remaining errors are due to my code, but if it is, please let me know!

streams/delimiter_stream.ts
```diff
@@ -3,15 +3,52 @@
...
+ *       new TextEncoder().encode(";"),
+ *       { disposition: "suffix" },
+ *     )
+ *   )
+ *   .pipeThrough(new TextDecoderStream());
+ * ```
+ *
+ * @param delimiter Delimiter byte sequence
+ * @param options Options for the transform stream
+ * @param options.disposition Disposition of the delimiter
```
crowlKats, line 50: move to comment in `DelimiterStreamOptions`'s `disposition`

crowlKats (approved):
LGTM

PR #3190, opened 2023-02-11
[WIP] feat(node/crypto): `Sign` and `Verify`

### LINKED ISSUE #3006: node: add crypto.Sign
`crypto.Sign` is not implemented now, but is necessary to use `npm:@google-cloud/translate`

ref https://github.com/denoland/deno_std/discussions/3002

iuioiua:
All that should remain to do is to add any functionality that [`browserify-sign`](https://github.com/crypto-browserify/browserify-sign) is missing that the Node API has. This should be good to go once the `parallel/test-crypto-sign-verify.js` test passes.

Closes #3006 and is part of #3149.


bartlomieju:
Thanks for working on it @iuioiua, I suggest to hold up a bit as we're in the process of moving `std/node` to the CLI in https://github.com/denoland/deno/pull/17724. To not duplicate work it'd be good to add this functionality there once the initial PR lands.

iuioiua:
Ok, I'll create a new PR once `std/node` has moved to the main repo.

PR #3193, opened 2023-02-11
chore(collections): full test coverage

iuioiua:
This change adds some tests and removes some unneeded functionality to bring the test coverage up to 100% for this module.


collections/binary_search_tree.ts
```diff
@@ -188,11 +188,6 @@ export class BinarySearchTree<T> implements Iterable<T> {
     const replacementDirection: Direction = direction === "left"
       ? "right"
       : "left";
-    if (!node[replacementDirection]) {
```
kt3k, line 191: I think this check is still useful for the user who extends this class. You can test this validation by the code like below:

```ts
import { BinarySearchTree  } from "./collections/binary_search_tree.ts";
class MyTree<T> extends BinarySearchTree<T> {
  foo() {
    super.rotateNode(this.root!, "right");
  }
}
const tree = new MyTree();
tree.insert(1);
tree.foo();
```

collections/zip.ts
```diff
@@ -31,7 +31,7 @@ import { minOf } from "./min_of.ts";
 export function zip<T extends unknown[]>(
   ...arrays: { [K in keyof T]: T[K][] }
 ): T[] {
-  const minLength = minOf(arrays, (it) => it.length) ?? 0;
+  const minLength = minOf(arrays, (it) => it.length)!;
```
kt3k, line 34: This value can be undefined when the input is an empty array. So this non-null assertion doesn't look good to me. If we want to increase the branch coverage here, we can call `zip()` and that should cover `?? 0` part.

kt3k (commented):
Test cases look good to me. Left some comments on the changes in implementation

collections/zip.ts
iuioiua: Ah, my previous assumption was that this wasn't the case. Thanks!

collections/binary_search_tree.ts
iuioiua: I see. Yes, it throws a useful error. I reverted and actually used this snippet as a test instead.

kt3k (approved):
LGTM. Nice work!

PR #2238, opened 2022-05-16
[WIP] feat: Add redis client

uki00a:
See #2224

## TODO

- [ ] Remove third-party dependencies
  - `db/redis/experimental/cluster/mod.ts` depends on `lodash-es` and [cluster-key-slot](https://github.com/invertase/cluster-key-slot)
  - benchmark code depends on [benny](https://github.com/caderek/benny)
- [ ] Run tests (`db/redis/tests`) and benchmarks (`db/redis/benchmark`) in CI

### Things to consider

- [deno-redis](https://github.com/denodrivers/redis/) contains experimental features in the `experimental/` directory (How should these be treated in `deno_std`?)
- How should tests be run in CI? (**currently ignored**)
  - [The test code requires that `redis-server` and `redis-cli` be installed locally](https://github.com/denodrivers/redis/blob/master/.github/workflows/build.yml)
- I assume that breaking changes may still happen to some APIs in `deno-redis`

## Note

- `deno-redis` contains benchmark code in `benchmark/` directory
  - It runs the same benchmark cases in both `deno-redis` and `ioredis`

Issue #2224, opened 2022-05-13
Include /x/redis as /std/redis


kt3k:
There seem strong demands/needs of redis client in the community. We should have one in std

ref: https://deno.land/x/redis

ref: https://news.ycombinator.com/item?id=31333025

External dependencies of /x/redis:
- lodash-es/shuffle
- lodash-es/sample
- lodash-es/uniqBy
- cluster-key-slot

erfanium:
what makes `redis` different from mysql, postgres, mongodb, elasticsearch,... ? They are all popular. we can't just include `redis` driver in std and ignore the rest.

kt3k:
@erfanium 
We chose to try migrating `redis` (and `sqlite`) first because they have comparatively less external dependencies. (Also the maintainer welcomes the transition in case of redis)

We haven't decided on other drivers. We might take different path for other drivers depending on their situations

piscisaureus:
I'm -1 on this. This is out of scope for std.

> There seem strong demands/needs of redis client in the community.

If there is no community provided driver for an important database like redis, we can consider starting the effort.
But I still think they should be separate projects. The discoverability issue can be fixed with improved search.

iuioiua:
> I'm -1 on this. This is out of scope for std.

I agree. I think we should close this issue.

PR #3200, opened 2023-02-14
feat: redis module

iuioiua:
This PR introduces the `redis` module to the Standard Library. It's a port of the [`r2d2` module](https://github.com/iuioiua/r2d2).

I understand there has yet to be a consensus on whether database connectors should exist within the Deno Standard Library. I am not 100% for it myself. Doing so may be outside the scope of this repo and may add too much managerial overhead. Nevertheless, having the implementation done would make the debate easier to settle, whether the consensus is for or against it.

Previously, #2238 attempted to port over [`deno-redis`](https://deno.land/x/redis) to the Standard Library. This implementation has some advantages that, I believe, may fit into this repo better:
1. It has no external dependencies outside of the Standard Library.
1. It has 100% test coverage but doesn't add time to CI workflows. It does this using a [Redis service container](https://docs.github.com/en/actions/using-containerized-services/about-service-containers). Note: A separate `redis` job had to be created for CI as the Redis service as it's only supported in Ubuntu runners.
1. Performant - The fastest Redis library in JavaScript (up to 35% faster than the next implementation, AFAIK).
1. It was written to be low-maintenance and easily understood. This is mostly thanks to it being functional instead of object-oriented. Also, it deliberately does not implement individual commands but offers a general-purpose but full-featured interface to Redis without unneeded intermediate abstractions.
1. Supports [RESPv2](https://github.com/iuioiua/r2d2#respv2), [RESP3](https://github.com/iuioiua/r2d2#resp3), [raw data](https://github.com/iuioiua/r2d2#raw-data), [pipelining](https://github.com/iuioiua/r2d2#pipelining), [pub/sub](https://github.com/iuioiua/r2d2#pubsub), [transactions](https://github.com/iuioiua/r2d2#transactions), [eval scripts](https://github.com/iuioiua/r2d2#eval-script) and [Lua scripts](https://github.com/iuioiua/r2d2#lua-script).
1. Compatible with [timeouts](https://github.com/iuioiua/r2d2#timeouts) and [retries](https://github.com/iuioiua/r2d2#retries).
1. It has been tested in production on my own site, [anhii.com](https://anhii.com/). I haven't had any real issues with it since adding it in. Actual implementation code can be seen [here](https://github.com/anhii/anhii/blob/main/routes/index.tsx#L8), [here](https://github.com/anhii/anhii/blob/main/routes/search.tsx#L25) and [here](https://github.com/anhii/anhii/blob/main/utils/redis.ts).

Todo:
- [ ] Expose the `redis` folder to Codecov.

Related:
- #2224

I'd love to hear what others think and any concerns.


lino-levan:
Oh wow. This will surely spark some vivid discussion. I'll start off with my take: not quite sure how to feel about this.

In general, standard libraries (that I know of) don't really implement any database drivers (outside of *maybe* sqlite). I don't know if there is good reasoning for this (outside of managerial overhead). I'm sure this has been proposed elsewhere and we should verifying that the reasoning that it didn't work for other std libs doesn't apply to us.

iuioiua:
Few things:
1. I realise I should delete `redis/README.md`.
2. The failing tests appear to be unrelated to this PR.
3. How could I ensure coverage is measured for Redis tests and the rest?

Before proceeding with further work on this PR, @kt3k, how do you and the team feel about this PR?

kt3k:
Sorry for the delayed response

> I understand there has yet to be a consensus on whether database connectors should exist within the Deno Standard Library.

The current policy is that we don't include library/tools for software from specific vendors, and redis client falls into that category. We also didn't land sqlite client (#2230) for the same reason. So I don't think we can land this as is.

iuioiua:
I understand, except for the "as is" part. Are you entertaining the implementation if it were tweaked? If not, all good. I'll close this PR.

kt3k:
I was wondering if RESP can be considered as a vendor independent specification (in that case, RESP parse/serializer doesn't violate std's policy), but also RESP protocol doesn't seem adopted except redis. So I'd suggest we should close this for now.

PR #3205, opened 2023-02-17
chore: remove powershell note in snapshot docs

GJZwiers:
Updating snapshots works on Powershell nowadays using `-- --update` rather than `"--" --update`, so maybe this note can be cleaned up now.

I am on `deno` 1.30.3, `std` 0.177.0 and `pwsh` 7.3.2.


dsherret (approved):
LGTM. Thanks!

Issue #3207, opened 2023-02-17
timingSafeEqual: RangeError: Offset is outside the bounds of the DataView


leahneukirchen:
**Describe the bug**

[This may be an error in the internals of argontwo, but even in this case the error message should be improved.]

timingSafeEqual raises `RangeError: Offset is outside the bounds of the DataView` when comparing two Uint8Array(32).

**Steps to Reproduce**

Run this:
```ts
import { encode, decode } from "https://deno.land/std/encoding/base64url.ts";
import { timingSafeEqual } from "https://deno.land/std/node/crypto.ts";

import { hash } from "https://deno.land/x/argontwo/mod.ts";

let x = hash('bong', decode('4ereginClY4QuqtT2j-8_g'));
// x = decode(encode(x));
console.log(timingSafeEqual(x,
			    decode('QAEyt51JgnxV8sc0wVKv63OcLO5BAzqFooCxXAnuBUo')))
```

This throws an exception.

When uncommenting the line that copies x, it works fine.

**Expected behavior**

The script prints true.

Both arrays are of same length and same contents as can be inspected easily.

**Environment**

- OS: Void Linux x86_64
- deno version: deno 1.30.3
- std version: std@0.177.0

PR #3208, opened 2023-02-17
fix: Create DataView with correct byteLength in timingSafeEqual

kamilogorek:
The example presented in https://github.com/denoland/deno_std/issues/3207 uses `argontwo`, which under the hood returns `Uint8Array` which memory buffer is taken directly from a WASM instance, making original `b.getUint8(i)` access out of bounds for 2nd argument to `timingSafeEqual`.

ref: https://deno.land/x/argontwo@0.1.1/wasm/mod.ts?source#L14
ref: https://deno.land/x/argontwo@0.1.1/mod.ts?source#L126

Fixes https://github.com/denoland/deno_std/issues/3207

Note#1: I use `SharedArrayBuffer` for cleaner assertions.
Note#2: changed names of 2 tests, because VS Code integration deduplicates tests with the same name.

### DIFF
--- a/crypto/timing_safe_equal.ts
+++ b/crypto/timing_safe_equal.ts
@@ -12,10 +12,14 @@ export function timingSafeEqual(
     return false;
   }
   if (!(a instanceof DataView)) {
-    a = ArrayBuffer.isView(a) ? new DataView(a.buffer, a.byteOffset, a.byteLength) : new DataView(a);
+    a = ArrayBuffer.isView(a)
+      ? new DataView(a.buffer, a.byteOffset, a.byteLength)
+      : new DataView(a);
   }
   if (!(b instanceof DataView)) {
-    b = ArrayBuffer.isView(b) ? new DataView(b.buffer, b.byteOffset, b.byteLength) : new DataView(b);
+    b = ArrayBuffer.isView(b)
+      ? new DataView(b.buffer, b.byteOffset, b.byteLength)
+      : new DataView(b);
   }
   assert(a instanceof DataView);
   assert(b instanceof DataView);
--- a/crypto/timing_safe_equal.ts
+++ b/crypto/timing_safe_equal.ts
@@ -12,10 +12,10 @@ export function timingSafeEqual(
     return false;
   }
   if (!(a instanceof DataView)) {
-    a = new DataView(ArrayBuffer.isView(a) ? a.buffer : a, 0, a.byteLength);
+    a = ArrayBuffer.isView(a) ? new DataView(a.buffer, a.byteOffset, a.byteLength) : new DataView(a);
   }
   if (!(b instanceof DataView)) {
-    b = new DataView(ArrayBuffer.isView(b) ? b.buffer : b, 0, b.byteLength);
+    b = ArrayBuffer.isView(b) ? new DataView(b.buffer, b.byteOffset, b.byteLength) : new DataView(b);
   }
   assert(a instanceof DataView);
   assert(b instanceof DataView);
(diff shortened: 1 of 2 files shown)


crypto/timing_safe_equal.ts
```diff
@@ -12,10 +12,10 @@ export function timingSafeEqual(
     return false;
   }
   if (!(a instanceof DataView)) {
-    a = new DataView(ArrayBuffer.isView(a) ? a.buffer : a);
+    a = new DataView(ArrayBuffer.isView(a) ? a.buffer : a, 0, a.byteLength);
```
kt3k, line 15: This looks fine when `a` is an ArrayBuffer, but the second argument seems to need to be `a.byteOffset` when `a` is a TypedArray.

Especially the example given in #3207 doesn't return true with this change

crypto/timing_safe_equal_test.ts
```diff
@@ -62,11 +62,59 @@ Deno.test({
...
     const a = encoder.encode("hello deno");
     const b = encoder.encode("hello Deno");
     assert(!timingSafeEqual(a, b));
   },
 });
+
+Deno.test({
+  name: "[timingSafeEqual] - Uint8Array w. different buffer sizes (a > b)",
+  fn() {
+    const a = new SharedArrayBuffer(4);
```
kt3k, line 77: minor point, but do we need to use SharedArrayBuffer here? How about ArrayBuffer?

kt3k:
@kamilogorek Thanks for looking into this! I left a few comments

crypto/timing_safe_equal.ts
kamilogorek: Great catch. Added the test to cover it as well.

crypto/timing_safe_equal_test.ts
kamilogorek: `ArrayBuffer` is a transferable object, so we won't be able to create `Uint8Array` after manipulating it with the `DataView`. And we do want to do that, to make `Uint8Array` a specific slice of the buffer for the purpose of these specific tests.

crypto/timing_safe_equal.ts
kamilogorek: Tested original case as well to double-check:

```ts
import { hash } from "https://deno.land/x/argontwo/mod.ts";
import { encode, decode } from "https://deno.land/std/encoding/base64url.ts";
import { timingSafeEqual } from "../../deno/deno_std/node/crypto.ts";

let x = hash("bong", decode('4ereginClY4QuqtT2j-8_g'));
let dec = decode('QAEyt51JgnxV8sc0wVKv63OcLO5BAzqFooCxXAnuBUo');

console.log(timingSafeEqual(dec, x)); // true
```

crypto/timing_safe_equal_test.ts
```diff
@@ -62,11 +62,83 @@ Deno.test({
...
+});
+
+Deno.test({
+  name: "[timingSafeEqual] - Uint8Array w. non-0 byteOffset",
+  fn() {
+    const a = new SharedArrayBuffer(4);
+    const va = new DataView(a);
+    va.setUint8(1, 212);
+    va.setUint8(2, 213);
+    const ua = new Uint8Array(a, 1, 2);
```
kt3k, line 129: 👍

kt3k (approved):
LGTM. Thanks!

bartlomieju:
@kt3k I believe this API might have been copied into `ext/node` in the Deno repo and if so should be updated as well

Issue #3123, opened 2023-01-17
Proposal: Move csv/yaml/toml/json/jsonc/front_matter to top level


kt3k:
These 6 modules in `std/encoding` are relatively complex and look worth top level modules.

Also there are claims from contributors that the current structures of these modules are unreasonable/inconsistent (#2538). For example, csv related APIs are exported from `encoding/csv.ts` and `encoding/csv/stream.ts`, and because of this structure `encoding/csv` isn't in the correct form of module. (There's another proposal to move `encoding/csv.ts` to `encoding/csv/mod.ts`, but by doing this these APIs will become less discoverable)

I propose we should move these like the below:

- `encoding/csv.ts` -> `csv/mod.ts`
- `encoding/yaml.ts` -> `yaml/mod.ts`
- `encoding/toml.ts` -> `toml/mod.ts`
- `encoding/json/*.ts` -> `json/mod.ts`
- `encoding/jsonc.ts` -> `jsonc/mod.ts`
- `encoding/front_matter.ts` -> `front_matter/mod.ts`

related #2538

lino-levan:
I wasn't really following the discussion in many of the related issues. Three quick notes:
- This is obviously a breaking change, are there any other changes we can make at the same time to avoid a double breaking change on these? (I think @timreichen has at least one proposal on csv parsing)
- Is the discoverability an issue with the directory structure or an issue with dotland? Imo, it feels like a dotland issue.
- This is just me spitballing, should we consider merging the json and jsonc folders along with this change?

iuioiua:
> * This is obviously a breaking change, are there any other changes we can make at the same time to avoid a double breaking change on these? (I think @timreichen has at least one proposal on csv parsing)

The move towards single-export files, which @timreichen aimed to do in his PRs as part of #2936.

> * Is the discoverability an issue with the directory structure or an issue with dotland? Imo, it feels like a dotland issue.

Directory structure - perhaps having the various formats under the one directory was once okay, but now the implementations seemed to have outgrown `std/encoding`, and each deserves to be a module on its own.

> * This is just me spitballing, should we consider merging the json and jsonc folders along with this change?

I don't think we should. Even though they're both JSON-related, they don't have much overlap.

lino-levan:
> (I think @timreichen has at least one proposal on csv parsing)

I was referring to https://github.com/denoland/deno_std/issues/2660 in regards to this. We should probably make a decision on this as we are moving the files.

pting-me:
[...]

PR #3213, opened 2023-02-21
refactor(encoding/varint): native implementation

iuioiua:
This change replaces the Wasm implementation of `encoding/varint` with a native-TypeScript implementation ported from the [x/varint module](https://deno.land/x/varint@v2.0.0) by @keithamus.

My benchmarks show that this implementation is 2x slower than its Wasm predecessor. However, the benefit is the implementation is simpler, and a step in CI is removed.

See https://github.com/denoland/deno_std/issues/3123#issuecomment-1437857122


encoding/varint.ts
```diff
@@ -1,9 +1,166 @@
...
+ * is of the decoded varint 32-bit number.
  *
- * @module
+ * VarInts are _not 32-bit by default_ so this should only be used in cases
+ * where the varint is _assured_ to be 32-bits.
+ *
+ * To know how many bytes the VarInt took to encode, simply negate `offset`
+ * from the returned new `offset`.
  */
+export function decodeU32(buf: Uint8Array, offset = 0): number {
```
keithamus, line 119: You’ll definitely want to somehow return how many bytes the varint was encoded in, so that consumers know which byte offset to continue reading data from.

Also, thanks ☺️

iuioiua: Good point. In that case, we'll have to deprecate the current API as the form of the return values will change. @kt3k, are you in favour?

kt3k: I agree that we should return the used bytes somehow.

> we'll have to deprecate the current API as the form of the return values will change

I wonder if there's other design possibilities other than returning 2-tuple (Creating an array for each decode sounds a little expensive to me). For example, `npm:varint` sets the used bytes at `varint.decode.bytes` ref. https://github.com/chrisdickinson/varint#varintdecodebytes

How about exporting used bytes like `export let decodedBytes` and setting that value at each decode?

keithamus: My vote would be against magic variables or properties on functions as they can release Zalgo in async contexts (though I don't have a compelling usecase for unpacking varints async :joy:)

One option could be to simply add additional methods like `decodeU32AndGetOffset(): [number, number]`. I'm sure you can think of a better name though :smiley:

kt3k: I doubt that could cause zalgo as `decode` doesn't involve any async operations. I think the code like the below works reliably:

```js
const n = decode(...);
const bytes = SOME_MAGIC_VAR;
```

> One option could be to simply add additional methods like decodeU32AndGetOffset(): [number, number]. I'm sure you can think of a better name though 😃

This seems obvious and a good option too 👍

keithamus: I was thinking more about if you had something like:

```js
const n = decode(...);
await someOp
const bytes = SOME_MAGIC_VAR
```

I can't think of a reason why you'd need to await between extraction... perhaps if you're streaming responses... but then you can shuffle the lines around without much difficulty. It's just a bit of an awkward api.

iuioiua: > I agree that we should return the used bytes somehow.
> 
> > we'll have to deprecate the current API as the form of the return values will change
> 
> I wonder if there's other design possibilities other than returning 2-tuple (Creating an array for each decode sounds a little expensive to me). For example, `npm:varint` sets the used bytes at `varint.decode.bytes` ref. https://github.com/chrisdickinson/varint#varintdecodebytes
> 
> How about exporting used bytes like `export let decodedBytes` and setting that value at each decode?

How is it expensive to return a 2-tuple? I don't understand how it is.

If it turns out to be inexpensive to return a 2-tuple, would it be better to return `{ result: number, offset: number }` for `decodeX()` and `{ buffer: UInt8Array, offset: number }` for `encodeX()` instead?

We should introduce these as `encode32Bit()`, `decode64Bit()`, etc., in conjunction with the deprecation.

kt3k: @keithamus 

> I can't think of a reason why you'd need to await between extraction... perhaps if you're streaming responses... but then you can shuffle the lines around without much difficulty. It's just a bit of an awkward api.

Ok. Now I agree that magic var design is confusing

kt3k: @iuioiua 

> We should introduce these as encode32Bit(), decode64Bit(), etc., in conjunction with the deprecation.

If we change the API names, how about following `x/varint` naming style? (`decode`, `decode32`, `encode`). Fortunately they don't have overlaps with the existing APIs, and also they look simpler than the current APIs.

iuioiua: Ok, we'll do a straight port of x/varint. However, I think it's worth splitting `encode()` into `encode32()` and `encode64()` if there are performance gains to be had. I'll create a new issue for the exploration of this idea.

iuioiua:
Closing in favour of #3215.

PR #3214, opened 2023-02-22
feat(ini): Module for handling INI file format

aaronhuggins:
INI files are a common cross-platform format found on Windows and Linux. Common use cases are manipulating desktop entry files on Linux and Windows profiles. This PR adds a module for handling the format with flexible parsing and formatting options. Related discussion #3209.


iuioiua:
Hi @aaronhuggins, the corresponding test file for a `mod.ts` file is usually `test.ts` rather than `mod_test.ts`.

aaronhuggins:
Hey @iuioiua, fixed that test file name convention. Thanks!

kt3k:
@aaronhuggins Can you also add the examples of `parse` and `stringify` usages at `@module` document?

kt3k:
cc @timreichen Do you see any issue in the design or implementation?

timreichen:
A few thoughts:
- Imo ini should be located in the same dir as yaml or toml, so either in std/encoding or they should be top level.
- The implementation logic seems to be the wrong way around. The `parse` and `stringify` functions should contain the logic rather than creating an instance of the ini class.
- I wonder why/if we should even have a class implementation in the first place.

aaronhuggins:
@timreichen 
> * Imo ini should be located in the same dir as yaml or toml, so either in std/encoding or they should be top level.

I started out with the files in `encoding`, but based on @kt3k feedback I moved it up to its own dir. Happy to move it back if consensus is `encoding`.

> * The implementation logic seems to be the wrong way around. The `parse` and `stringify` functions should contain the logic rather than creating an instance of the ini class.
> * I wonder why/if we should even have a class implementation in the first place.

Based on experience with other [languages](https://docs.python.org/3/library/configparser.html) and the requirements of the best documented [spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s03.html), a class implementation is useful for fine-grained control over the data, especially preserving and manipulating comments in addition to plain key/value pairs.

There's a use case for very simple `parse`/`stringify` for users who don't care about more advanced features of INI specs and just need to access values or convert to JSON. Since a class implementation also feels to me like the best place to locate the logic for super-users, I chose to provide convenience methods which wrapped the class instead of duplicating logic.

> * The test suite is not complete

The test suite appeared complete to me before submitting, I'll review again and try to improve the suite.

> * The `[ini] use Object.defineProperty when setting object property` test seems to be unassociated with ini itself. Why is this tested?

It seemed necessary to cover `__proto__` as a key in an INI for cases where users call `toObject` or the top-level `parse`. Someone could theoretically craft a malicious INI key or section to modify the prototype of the resulting object without properly setting via `Object.defineProperty`.

aaronhuggins:
@timreichen Refactored tests. Great call. Found some misses which could have led to unexpected behavior.

@kt3k you said that file formats were going to be moved from `encoding` to their own top-level dirs. Do you and @timreichen agree to move this as `encoding/ini.ts`?

iuioiua:
I think putting this implementation in a top-level folder is the way to go, given the complexity of the implementation. See #3123 for more context.

kt3k:
I agree with @iuioiua. #3123 is almost at the consensus in my view, and the current directory position is aligned to it.

timreichen:
> I agree with @iuioiua. #3123 is almost at the consensus in my view, and the current directory position is aligned to it.

Then I agree with having the ini mod top level.

> Based on experience with other [languages](https://docs.python.org/3/library/configparser.html) and the requirements of the best documented [spec](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s03.html), a class implementation is useful for fine-grained control over the data, especially preserving and manipulating comments in addition to plain key/value pairs.

instead of a map class, I think one possible solution would be to provide a `IniFormatter` with a `formatToParts` and maybe `partsToString` function (similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatToParts and how datetime is implemented in std/datetime https://github.com/denoland/deno_std/blob/b23a76a47adaf63b5493761a87e6dd0dc0de0bb8/datetime/_common.ts#L403 https://github.com/denoland/deno_std/blob/b23a76a47adaf63b5493761a87e6dd0dc0de0bb8/datetime/_common.ts#L604) to have fine control. This design is more like a web API.

> It seemed necessary to cover __proto__ as a key in an INI for cases where users call toObject or the top-level parse. Someone could theoretically craft a malicious INI key or section to modify the prototype of the resulting object without properly setting via Object.defineProperty.

I am not sure if we should have special pleading in ini for that. Else we'd need to do some checks in every mod where objects are used, no?

There are two additional points that need to be discussed:

-Should comments be parsed?
Lots of implementations in std (for example toml and jsonc) skip comments during parsing. I think with ini we should do the same for the simple `parse`. This would make the function about 40 lines long.
- Should ParseOptions even have a reviver function?
Seems to me that the data can be modified after parsing if necessary.
I know that `JSON.parse` takes it as an argument but I think that is a leftover from early, not too well thought out API design.
We also removed the reviver function in std/csv because the function is required to be sync.

aaronhuggins:
@timreichen I'll give some thought to those critiques. You're points about bringing it in line with web standards may be right, but I'm not sure and might want to play a little with that.

One thing that strikes me is this:

> I am not sure if we should have special pleading in ini for that. Else we'd need to do some checks in every mod where objects are used, no?

I think this check should be present in parsers which result in an object. Otherwise, they're vulnerable to prototype pollution attacks.

iuioiua:
> * Should comments be parsed?
>   Lots of implementations in std (for example toml and jsonc) skip comments during parsing. I think with ini we should do the same for the simple `parse`. This would make the function about 40 lines long.

I think not.

> * Should ParseOptions even have a reviver function?
>   Seems to me that the data can be modified after parsing if necessary.
>   I know that `JSON.parse` takes it as an argument but I think that is a leftover from early, not too well thought out API design.
>   We also removed the reviver function in std/csv because the function is required to be sync.

I think maybe not.

lino-levan:
This is a lot of great work! I'm in agreement with @iuioiua and @timreichen.

 - In regards to comments being parsed. I've never heard of a config file parser handling comments and from a quick look at the code it seems to add a lot of complexity.
 - In regards to a reviver method, I'm pretty sure this existing for `JSON.parse()` is just leftover from a badly designed API.

iuioiua:
Is this ready to merge, @aaronhuggins?

not-my-profile:
> INI files are a common cross-platform format found on Windows and Linux.

That's not true. There is no specification that establishes "INI" cross-platform. The Desktop Entry Specification doesn't even mention "INI" once. The Python [configparser](https://docs.python.org/3/library/configparser.html) module clearly states (emphasis mine):

> This module provides the ConfigParser class which implements a basic configuration language which provides a structure **similar** to what’s found in Microsoft Windows INI files.

E.g. the [Desktop Entry Specification](https://specifications.freedesktop.org/desktop-entry-spec/latest/ar01s03.html) only specifies `#` as a comment prefix, `configparser` only recognizes `#` and `;` by default while your implementation apparently recognizes `#`, `;`, or `//` by default and Windows supports who knows what. This is just an example I am sure that there are many more subtle differences.

With the lack of an INI specification in mind I don't think that it makes sense for a standard library to include an `ini` module. I think it would make more sense to name the module `configparser` and model the parser directly after the Python implementation.

Sidenote: The Python [configparser documentation](https://docs.python.org/3/library/configparser.html) basically recommends TOML over INI:

> TOML is a well-specified format for application configuration files. It is specifically designed to be an improved version of INI.

I think it would make sense for a Deno implementation to do so as well.

iuioiua:
I just fixed the typo and rebased. @aaronhuggins, are you still happy to push this PR forward? If so, please sign the CLA and let us know once it's ready for review 🙂

kt3k:
We'll continue to work on this feature in #3871

PR #3215, opened 2023-02-23
refactor(encoding/varint): deprecate Wasm implementation in favour of native TypeScript

iuioiua:
This change replaces the Wasm implementation of `encoding/varint` with a native-TypeScript implementation ported from the [x/varint module](https://deno.land/x/varint@v2.0.0) by @keithamus.

My benchmarks show that this implementation is 2x slower than its Wasm predecessor. However, the benefit is the implementation is simpler, and a step in CI is removed.

This PR supersedes #3213.

Ref:
* https://github.com/denoland/deno_std/issues/3123#issuecomment-1437857122
* https://github.com/denoland/deno_std/pull/3213#discussion_r1115252163

### DIFF
--- a/encoding/varint.ts
+++ b/encoding/varint.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// Copyright 2020 Keith Cirkel. All rights reserved. MIT license.
 
 /**
  * Functions for encoding typed integers in array buffers.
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

iuioiua:
@lino-levan, WDYT?

lino-levan:
The wasm varint was always a nightmare. The decision seems pretty clear to me (thanks for the benchmark!). I don't see a lot of downsides to this change. LGTM!

PR #3216, opened 2023-02-23
fix(fs/ensure_symlink): lstat relative symlink properly

### LINKED ISSUE #2312: fs.ensureSymlink() can't create relative symlinks
**Describe the bug**

It appears that `fs.ensureSymlink()` won't create links to files that are relative to the link.

It also produces broken links when give a destination path relative to the current working directory but is not relative to the link.

**Steps to Reproduce**

```ts
// relLinkTest.ts

import * as fs from "https://deno.land/std@0.142.0/fs/mod.ts";

fs.emptyDirSync("dir");

fs.ensureFileSync("dir/files/a.txt");
Deno.writeTextFileSync("dir/files/a.txt", "A");

// This link is broken because there is no file "dir/link/dir/files/a.txt"
fs.ensureSymlinkSync("dir/files/a.txt", "dir/links/a.txt");
/*
  dir/
  ├── files
  │   └── a.txt
  └── links
      └── a.txt -> dir/files/a.txt <--- broken

  2 directories, 2 files
*/
try {
  console.log(Deno.readTextFileSync("dir/links/a.txt"));
} catch (e) {
  console.error(e); // NotFound: No such file or directory (os error 2)
}

try {
  // can't create a relative symlink:
  fs.ensureSymlinkSync("../files/a.txt", "dir/links/rel.txt");
} catch (e) {
  console.error(e); // NotFound: No such file or directory (os error 2), stat '../files/a.txt'
}

// create the link with `ln` command
const p = Deno.run({ cmd: ["ln", "-s", "../files/a.txt", "dir/links/ln.txt"] });
await p.status();
/*
  dir
  ├── files
  │   └── a.txt
  └── links
      ├── a.txt -> dir/files/a.txt <--- broken
      └── ln.txt -> ../files/a.txt <--- valid
*/

console.log(Deno.readTextFileSync("dir/links/ln.txt")); // A
```

**Expected behavior**

I have trouble understanding how `ensureSymlink()` is useful if it can't work with relative symlinks. I would expect the same behavior as `ln`.

**Environment**

- OS: BTW I use Arch
- deno version: 1.22.2
- std version: 0.142.0

locriacyber:
Closes #2312

### DIFF
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -4,6 +4,16 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, toPathString } from "./_util.ts";
 import { isWindows } from "../_util/os.ts";
 
+export function resolveSymlinkSrc(src: string, dest: string): string;
+export function resolveSymlinkSrc(src: URL, dest: URL): URL;
+export function resolveSymlinkSrc(src: string | URL, dest: string | URL) {
+  if (typeof src == "string" && typeof dest == "string") {
+    return path.resolve(path.dirname(dest), src);
+  } else {
+    return new URL(src, dest)
+  }
+}
+
 /**
  * Ensures that the link exists.
  * If the directory structure does not exist, it is created.
@@ -11,8 +21,11 @@ import { isWindows } from "../_util/os.ts";
  * @param src the source file path
  * @param dest the destination link path
  */
-export async function ensureSymlink(src: string | URL, dest: string | URL) {
-  const srcStatInfo = await Deno.lstat(src);
+export function ensureSymlink(src: string, dest: string);
+export function ensureSymlink(src: URL, dest: URL);
+export function ensureSymlink(src: any, dest: any) {
+  const srcReal = resolveSymlinkSrc(src, dest);
+  const srcStatInfo = await Deno.lstat(srcReal);
   const srcFilePathType = getFileInfoType(srcStatInfo);
 
   await ensureDir(path.dirname(toPathString(dest)));
@@ -39,8 +52,11 @@ export async function ensureSymlink(src: string | URL, dest: string | URL) {
  * @param src the source file path
  * @param dest the destination link path
  */
-export function ensureSymlinkSync(src: string | URL, dest: string | URL) {
-  const srcStatInfo = Deno.lstatSync(src);
+export function ensureSymlinkSync(src: string, dest: string);
+export function ensureSymlinkSync(src: URL, dest: URL);
+export function ensureSymlinkSync(src: any, dest: any) {
+  const srcReal = resolveSymlinkSrc(src, dest);
+  const srcStatInfo = Deno.lstatSync(srcReal);
   const srcFilePathType = getFileInfoType(srcStatInfo);
 
   ensureDirSync(path.dirname(toPathString(dest)));
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -4,10 +4,9 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, toPathString } from "./_util.ts";
 import { isWindows } from "../_util/os.ts";
 
-export function resolveSymlinkSrc(src: string, dest: string): string;
-export function resolveSymlinkSrc(src: URL, dest: URL): URL;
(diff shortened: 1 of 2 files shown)


locriacyber:
I agree to the CLA, but I'm not signing up to this shady service.
image

fs/ensure_symlink.ts
```diff
@@ -4,15 +4,28 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, toPathString } from "./_util.ts";
 import { isWindows } from "../_util/os.ts";
 
+export function resolveSymlinkSrc(src: string, dest: string): string;
+export function resolveSymlinkSrc(src: URL, dest: URL): URL;
+export function resolveSymlinkSrc(src: string | URL, dest: string | URL) {
```
kt3k, line 9: This looks like an internal utility. So Let's not export this?

fs/ensure_symlink.ts
```diff
@@ -4,15 +4,28 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
...
  * If the directory structure does not exist, it is created.
  *
  * @param src the source file path
  * @param dest the destination link path
  */
-export async function ensureSymlink(src: string | URL, dest: string | URL) {
-  const srcStatInfo = await Deno.lstat(src);
+export function ensureSymlink(src: string, dest: string);
+export function ensureSymlink(src: URL, dest: URL);
+export function ensureSymlink(src: any, dest: any) {
```
kt3k, line 26: I think we shouldn't change the typing here. This now doesn't support the usage like `ensureSymlink(src: string, dest: URL)` which we currently support.

I think we should align `src` and `dest` to either URL or string by checking each type

kt3k:
Thanks for your contribution!

Can you add test cases which exercises the this fix?

kt3k:
> I agree to the CLA, but I'm not signing up to this suspicious service.

I guess you're trying to sign up to CLA assistant as CLA owner (which is not required here). You should be able to sign our CLA with less permissions by navigating from the above like (one in @CLAassistant's comment).

fs/ensure_symlink.ts
locriacyber: ok

locriacyber:
> > I agree to the CLA, but I'm not signing up to this suspicious service.
> 
> I guess you're trying to sign up to CLA assistant as CLA owner (which is not required here). You should be able to sign our CLA with less permissions by navigating from the above like (one in @CLAassistant's comment).

I used that link

fs/ensure_symlink.ts
locriacyber: I'm not sure why both URL and string are supported.

Does creating symlink to URL make any sense?

locriacyber:
> Thanks for your contribution!
> 
> Can you add test cases which exercise this fix?

How to add test cases

locriacyber:
Two problems:
- `ensureSymlink` Also checkes if the link target exist. This is not documented.
- If the link target is different, nothing is done.

kt3k:
Do you click the bottom button in the below screenshot?

Screen Shot 2023-02-24 at 20 58 57

That should navigate to the below authorization screen (which only asks Email (read-only) permission)

Screen Shot 2023-02-24 at 13 14 05

kt3k:
> How to add test cases

Add test cases in `fs/ensure_symlink_test.ts` with the fixture setup in `fs/testdata` dir, which simulates the situation described in #2312

locriacyber:
> Two problems:
> 
>     * `ensureSymlink` Also checkes if the link target exist. This is not documented.
> 
>     * If the link target is different, nothing is done.

@kt3k How do I deal with this?

kt3k:
Thanks for updating! The implementation now looks good to me.

> * `ensureSymlink` Also checkes if the link target exist. This is not documented.

Please add this to the jsdoc.

> * If the link target is different, nothing is done.

What does this mean?

kt3k:
Can you also run `deno fmt` to pass lint CI check?

locriacyber:
> > * If the link target is different, nothing is done.
> 
> What does this mean?

If `link` points to `different-file`, and you call `ensureSymlink(link, target)`, nothing will happen

kt3k (approved):
Thanks for updating! Confirmed this fixes the issue. LGTM

kt3k:
> If link points to different-file, and you call ensureSymlink(link, target), nothing will happen

Sounds like a bug as it doesn't ensure the link targets to the given file, but let's address this later

PR #3220, opened 2023-02-24
fix(http/file_server): Redirect directory URLs that don't end with a slash

### LINKED ISSUE #3219: `http/file_server` should redirect directory URLs that don't end with a slash to URLs that do

If `showIndex` is true, and `./some/directory/index.html` exists, all of the following URLs will load that `index.html` file:

- Case 1: `http://example.com/some/directory/index.html`
- Case 2: `http://example.com/some/directory/` (with a final slash)
- Case 3: `http://example.com/some/directory` (without a final slash)

However, the version without a final slash is not equivalent in terms of how relative URLs resolve against it.

Say that index.html loads an image using a relative URL: ``. Both case 1 and 2 will resolve that URL as being `http://example.com/some/directory/test.png`. Case 3 will resolve it as `http://example.com/some/test.png`.

[Here's an example that shows this in practice.](https://slash-directory-test.deno.dev/) ([Github repo](https://github.com/andreubotella/slash-directory-test))

If `showIndex` is true, and a directory contains an `index.html` file, loading a URL to that directory that doesn't end with a slash should HTTP-redirect to the URL with the slash, rather than serving the `index.html` file.

This is the behavior of other static servers, including Apache HTTPD, nginx and lighttpd. They all use a 301 redirect.

An alternative could be to add an option to `ServeDirOptions`. However, it would have to default to redirecting, since that seems to be the expected behavior.

andreubotella:
When `showIndex` is true and `directory/index.html` exists, loading `http://example.com/directory` (without a final slash) can cause relative URLs in the index file to resolve against the wrong directory. This change redirects such URLs to the version ending with a slash to avoid this.

Fixes #3219.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -520,8 +520,8 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
             // relative URLs in the index file will resolve against the parent
             // directory, rather than the current directory. To prevent that, we
             // return a 301 redirect to the URL with a slash.
-            const url = new URL(req.url);
-            if (!url.pathname.endsWith("/")) {
+            if (!fsPath.endsWith("/")) {
+              const url = new URL(req.url);
               url.pathname += "/";
               return Response.redirect(url, 301);
             }
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -516,6 +516,15 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
           const path = posix.join(fsPath, "index.html");
           const indexFileInfo = await Deno.lstat(path);
           if (indexFileInfo.isFile) {
+            // If the current URL's pathname doesn't end with a slash, any
+            // relative URLs in the index file will resolve against the parent
+            // directory, rather than the current directory. To prevent that, we
+            // return a 301 redirect to the URL with a slash.
+            const url = new URL(req.url);
+            if (!url.pathname.endsWith("/")) {
+              url.pathname += "/";
+              return Response.redirect(url, 301);
+            }
             response = await serveFile(req, path, {
               etagAlgorithm: opts.etagAlgorithm,
               fileInfo: indexFileInfo,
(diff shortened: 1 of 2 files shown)


kt3k:
I'm in favor of the idea. Can you add a test case?

http/file_server.ts
```diff
@@ -516,6 +516,15 @@ export async function serveDir(req: Request, opts: ServeDirOptions = {}) {
           const path = posix.join(fsPath, "index.html");
           const indexFileInfo = await Deno.lstat(path);
           if (indexFileInfo.isFile) {
+            // If the current URL's pathname doesn't end with a slash, any
+            // relative URLs in the index file will resolve against the parent
+            // directory, rather than the current directory. To prevent that, we
+            // return a 301 redirect to the URL with a slash.
+            const url = new URL(req.url);
+            if (!url.pathname.endsWith("/")) {
```
kt3k, line 524: Can this check be performed with some existing variable? For example, does `fsPath.endsWith("/")` work?

kt3k: If so, we can `new URL` in the if block and keep the perf a bit better

andreubotella: I thought `fsPath.endsWith("/")` would not work because of the path normalization, but I've checked it, and it doesn't seem to remove trailing slashes. However, creating a `URL` object is pretty much needed inside the `if` block, since doing `req.url + "/"` will not work if there's a query string.

http/file_server_test.ts
```diff
@@ -1201,6 +1201,32 @@ Deno.test(
...
+
+Deno.test(
+  "serveDir redirects a directory URL not ending with a slash correctly even with a query string",
+  async () => {
+    const url = "http://localhost:4507/http/testdata/subdir-with-index?test";
+    const res = await serveDir(new Request(url), { showIndex: true });
+    assertEquals(res.status, 301);
+    assertEquals(
+      res.headers.get("Location"),
+      "http://localhost:4507/http/testdata/subdir-with-index/?test",
```
kt3k, line 1225: 👍

kt3k (approved):
LGTM

andreubotella:
Can this be merged?

PR #3154, opened 2023-02-01
fix: make encoding/front_matter work in a browser

kitsonk:
This safely access the Deno namespace so encoding/front_matter can be used in a browser (like an interactive island in Fresh).

Issue #1973, opened 2022-03-01
Some modules cannot be marked as "browser compatible" because of outdated TS DOM types


jsejcksn:
## Context

The current browser compatibility CI test [uses TypeScript's native DOM types](https://github.com/denoland/deno_std/blob/main/browser-compat.tsconfig.json) when type-checking modules. This is appropriate.

However, the current DOM types appear to be outdated (technically, _misaligned_ from what's being commonly shipped by browsers), so it's not possible to get tests to pass for some modules which actually _are_ compatible in that sense.

## Problem

The above described scenario prevents marking some modules as being "browser compatible" and prevents their discovery as such. I've included links to examples at the end of this message.

Refs:
- https://github.com/denoland/deno_std/pull/1972#issuecomment-1054901936
- https://github.com/denoland/deno_std/pull/1972#issuecomment-1054923976

jsejcksn:
@crowlKats I don't see [any activity at all in the TS repo which mentions `ReadableByteStreamController`](https://github.com/microsoft/TypeScript/issues?q=ReadableByteStreamController). Am I missing something or should we file an issue?

crowlKats:
yep, we probably should. Major missing structs are `ReadableByteStreamController` & `ReadableStreamBYOBReader`

PR #3229, opened 2023-03-06
refactor: improved browser-compatibility checking

iuioiua:
When submitting this PR, ~76% of public code is browser compatible, and ~24% is not browser compatible. Currently, each browser-compatible file requires a `// This module is browser compatible.` comment for type-checking via the `test:browser` task.

This change inverts the current approach to browser-compatibility declaration by instead having files that are browser-_incompatible_ have a declaration. In other words, the code in the Standard Library can safely be assumed to be browser-compatible unless stated otherwise by a `// This module is not browser compatible.` comment.

Secondly, the task has been improved to detect and notify the developer when a declaration must be added or removed.

Lastly, browser-related compatibility declarations have been adjusted to the new approach across the codebase.


lino-levan:
Super interesting concept. I think this makes sense but this is a subtle """breaking change""" in the (public facing) philosophy of the deno std. I think this makes sense. Thanks for updating the README for clarity.

LGTM!

README.md
```diff
@@ -3,8 +3,8 @@
 [![codecov](https://codecov.io/gh/denoland/deno_std/branch/main/graph/badge.svg?token=w6s3ODtULz)](https://codecov.io/gh/denoland/deno_std)
 
 These modules do not have external dependencies and they are reviewed by the
-Deno core team. The intention is to have a standard set of high quality code
-that all Deno projects can use fearlessly.
+Deno core team. The intention is to have a standard set of high-quality,
+browser-compatible code that all Deno projects can use fearlessly.
```
kt3k, line 7: I'm not sure we can say this here as we haven't run test cases in browsers, also we don't have docs about how to use deno_std in browsers.

iuioiua: Reverted

kt3k (approved):
LGTM

jsejcksn:
@iuioiua Thanks for creating this PR. See also: https://github.com/denoland/deno_std/issues/1973 — I haven't checked those in while.

Additionally, I'm curious about how you created the changes in this PR. Did you:

- update every file manually?
- use an IDE's GUI find/replace feature?
- use some CLI tool(s)?
- use some combination of the above?

...or something else?

iuioiua:
> 

I removed all instances of the `// This module is browser compatible.` declaration via VSCode's find and replace. I then, by hand, added the new declaration to those browser-incompatible files thanks to [this script](https://gist.github.com/iuioiua/bcea1e771c659ab8ee3c6937eb18a881).

kt3k:
I changed my mind. I'm not in favor of this change because of 2 reasons:

(1) `isBrowserCompatible` check is not always true because some files include the reference to `Deno` namespace, but still browser compatible (ex. front_matter https://github.com/denoland/deno_std/pull/3154). So the proposed `This module is not browser compatible.` header is not accurate. On the other hand, the current `This module is browser compatible.` header is probably more accurate because if the source passes browser type checking, it probably works with browsers (at least no exception is known for now).

(2) I think the default policy of Deno Standard Library is still `it works in Deno` and not `it works in Deno and browsers by default`. This can be changed in the future, but at this point, we don't run test cases in actual browsers in CI and also we don't document how to use deno_std in browsers. So deno_std is not ready enough to declare it's browser compatible by default

iuioiua:
There are two components to this PR:
1. The re-defining of this repo as being browser-compatible-first.
2. The automation of browser compatibility checks.

@kt3k, the points you raise against component 1 are totally valid, and I agree. I'll close this PR in favour of a new one focusing solely on component 2, as long as the task takes little time to complete. Thanks for your input, @lino-levan.

PR #3206, opened 2023-02-17
Remove std/node, it was merged into Deno itself

ry:
Ref https://github.com/denoland/deno/pull/17785 https://github.com/denoland/deno/pull/17777

PR #3234, opened 2023-03-06
chore: remove `check_deploy` tool

iuioiua:
This tool appears to be no longer in use.


kt3k:
related: https://github.com/denoland/deno_std/pull/3206

PR #3028, opened 2022-12-20
refactor(encoding/json): unified dir

timreichen:
ref: #2538

This PR moves all json related stuff into the same dir.

encoding/testdata/json → encoding/json/testdata

PR #3236, opened 2023-03-07
BREAKING(json): move to top-level folder and towards single-export files

iuioiua:
Part of:
* #3123
* #2538
* #2936

Supercedes:
* #3028.


kt3k (approved):
LGTM

PR #3238, opened 2023-03-07
refactor(CI): enable ubuntu to upload coverage

iuioiua:
Uploading coverage from Ubuntu was disabled in denoland/deno#10420 due to a panic caused by `readAll()` within `std/streams`. This change is an experiment that re-enables coverage upload.


kt3k (approved):
LGTM. Thanks!

The issue (https://github.com/denoland/deno/issues/10420) seems now resolved.

PR #3239, opened 2023-03-07
feat: add http/server_sent_event

kitsonk:
This ports the server sent event abstraction from oak into the standard library, which allows users to create an event target for server sent events, which then provides a response to the client that can be used to dispatch events which then get raised in the client.


http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Provides {@linkcode ServerSentEvent} and {@linkcode SSEStreamTarget}
+ * which provide an interface to sending sending server sent events to a
```
kt3k, line 5: typo?
```suggestion
 * which provide an interface to sending server sent events to a
```

http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
...
+   *             of `"message"` will be handled specifically as a message
+   *             server-side event.
+   * @param data  arbitrary data to send to the client, data this is a string
+   *              will be sent unmodified, otherwise `JSON.parse()` will be used
+   *              to serialize the value
+   * @param eventInit initialization options for the event
+   */
+  constructor(
+    type: string,
+    data: unknown,
```
kt3k, line 129: How about including `data` in `ServerSentEventInit`?

`CustomEvent` has similar structure as this class and it has `details` property in the init option bag. https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent

http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
...
+  #closed = false;
+  #controller?: ReadableStreamDefaultController<Uint8Array>;
+  // we are ignoring any here, because when exporting to npm/Node.js, the timer
+  // handle isn't a number.
+  // deno-lint-ignore no-explicit-any
+  #keepAliveId?: any;
+
+  // deno-lint-ignore no-explicit-any
+  #error(error: any) {
+    console.log("error", error);
```
kt3k, line 273: Is this debug logging?

http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
...
+   * }, { port: 8000 });
+   * ```
+   */
+  dispatchEvent(event: CloseEvent | ErrorEvent): boolean;
+}
+
+/** An implementation of {@linkcode ServerSentEventTarget} that provides a
+ * readable stream as a body of a response to establish a connection to a
+ * client. */
+export class SSEStreamTarget extends EventTarget
```
kt3k, line 261: Maybe we should rename this to `ServerSentEventStreamTarget` for consistency? (SSE acronym doesn't look used in other places)

(Also if we use `SSE` acronym, it should be `Sse`. We decided not to use ALL CAPS acronyms in public API https://github.com/denoland/deno_std/pull/2582)

http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
...
+
+  /** The optional ID associated with the event that will be sent to the client
+   * and be made available in the `EventSource`. */
+  get id(): number | undefined {
+    return this.#id;
+  }
+
+  override toString(): string {
+    const data = `data: ${this.#data.split("\n").join("\ndata: ")}\n`;
+    return `${this.#type === "__message" ? "" : `event: ${this.#type}\n`}${
```
kt3k, line 163: `__message` type is used for representing unnamed event type, but not documented. It also feels a little artificial. How about using empty string for representing unnamed event type?

http/server_sent_event.ts
kitsonk: Good point, also on the `MessageEvent`, which is the client side event type, it's constructor takes `data` in the options bag. I will rework it.

http/server_sent_event.ts
```diff
@@ -0,0 +1,411 @@
...
+ * client. */
+export class SSEStreamTarget extends EventTarget
+  implements ServerSentEventTarget {
+  #bodyInit: ReadableStream<Uint8Array>;
+  #closed = false;
+  #controller?: ReadableStreamDefaultController<Uint8Array>;
+  // we are ignoring any here, because when exporting to npm/Node.js, the timer
+  // handle isn't a number.
+  // deno-lint-ignore no-explicit-any
+  #keepAliveId?: any;
```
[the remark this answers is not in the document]
kitsonk: Is there a disadvantage to keeping the std library type checked against Node.js as well as Deno?  We have preserved the inspection logic in other contributions.

http/server_sent_event.ts
kitsonk: Agreed, I have renamed it.

http/server_sent_event.ts
kitsonk: No, it is a sigil to identify a "message" versus an event, which gets called by a different handler on the client side.  It isn't documented because it is "private" and only used by `.dispatchMessage()`.  I will convert it to a symbol though to avoid any confusion.

kitsonk: Actually tried that and the problem is that you can't call `super()` on the `Event` with that, and we need some sort of string sigil that identifies the non-event message and a `""` feels like a risky one.  Again it isn't "undocumented" it is just not intended to be publically accessible, therefore something that is unlikely to collide with a legitimate event type.

kt3k (approved):
LGTM. Thanks!

Issue #3233, opened 2023-03-06
`ensureDir` and `ensureDirSync` are racy


lucacasonato:
They do `lstat` then `mkdir` rather than doing only `mkdir` and throwing an error.

If you call `ensureDir` quickly enough in parallel, one of the two `ensureDir`'s will throw "AlreadyExists".

"exists" checks using something like `lstat` are wrong 99% of times, and this is no exception :)

kt3k:
We switch the order of `mkdir` and `lstat` in #3242, but that caused another DX problem. Now `ensureDir` always asks for `write` permission to the given directory even when the directory already exists and it actually doesn't need to write to it.

Now a PR was created to mitigate that situation, but the solution looks very complex. #4008

Should we follow the direction of #4008 or should we revert back to pre-#3242 state, accepting some racy nature of it?

(In my view, `ensureDir` is usually called for creating special purpose paths such as cache dir, config dir, etc. I can't imagine `ensureDir` is called in high concurrency. We can also document that it can be racy in highly concurrent situation)

iuioiua:
Accepting the racy nature of the function seems more reasonable and better than a complex solution as long as there's clear notice within the documentation, like [`exists()`](https://deno.land/std@0.210.0/fs/mod.ts?s=exists).

lucacasonato:
These functions should not be racy. If that means you sometimes need write permissions instead of read permissions, that seems like a reasonable trade-off to me.

lucacasonato:
@kt3k i opened this bug because I was using it in high concurrency and my code would randomly crash. I think this function being racy is not acceptable.

PR #3240, opened 2023-03-07
chore(fs): Refactor the test of ensureFile and ensureDir

ayame113:
When I was investigating #3233, these test cases seemed incorrect, so I changed some.

The following three points have been updated.

- Changed `assertThrows()` that looks like nothing ([comment](https://github.com/denoland/deno_std/pull/3240#discussion_r1128484467))
- Use try-finally to make the directory clean up correctly
- A small test case that is useful for debugging when implementing [#3233](https://github.com/denoland/deno_std/issues/3233) has been added. (`ensureDirShouldNotSwallowErrors` and `ensureDirSyncShouldNotSwallowErrors`)

### DIFF
--- a/fs/ensure_dir_test.ts
+++ b/fs/ensure_dir_test.ts
@@ -110,7 +110,7 @@ Deno.test("ensureDirSyncIfItAsFile", function () {
 Deno.test("ensureDirIfInvalidPath", async function () {
   await assertRejects(
     async () => {
-      await ensureDir("<invalid>");
+      await ensureDir("<invalid\0>");
     },
     Error,
   );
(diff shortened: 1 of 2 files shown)


fs/ensure_dir_test.ts
```diff
@@ -11,97 +11,116 @@ Deno.test("ensureDirIfItNotExist", async function () {
...
+    Deno.mkdirSync(testDir, { recursive: true });
 
-  ensureDirSync(testDir);
+    ensureDirSync(testDir);
 
-  assertThrows(() => {
+    // test dir should still exists.
     Deno.statSync(testDir);
-    throw new Error("test dir should still exists.");
-  });
```
ayame113, line 70: The following code seemed to be nothing tested because the error was always thrown.

```ts
assertThrows(() => {
  Deno.statSync(testDir);
  throw new Error("test dir should still exists.");
});
```

Therefore, it was rewritten as follows.

```ts
// test dir should still exists.
Deno.statSync(testDir);
```

kt3k (approved):
LGTM. Thank you!

PR #3026, opened 2022-12-20
refactor(encoding/toml): unified dir and move to single-export files

timreichen:
ref: #2538

This PR moves all toml related stuff into the same dir.

encoding/toml.ts → encoding/toml/mod.ts
encoding/test_toml.ts → encoding/toml/mod_test.ts
encoding/testdata/*.toml → encoding/toml/testdata/*.toml

splits logic into single export files.

⚠️ `https://deno.land/std@$STD_VERSION/encoding/toml.ts` is deprecated.

PR #3241, opened 2023-03-07
BREAKING(toml): move to top-level folder and towards single-export files

iuioiua:
Part of:
1. #3123
2. #2538
3. #2936

Supercedes #3026


kt3k (approved):
LGTM

PR #3242, opened 2023-03-07
fix(fs/ensureDir): Fix racy ensureDir

### LINKED ISSUE #3233: `ensureDir` and `ensureDirSync` are racy

ayame113:
close #3233

This PR builds on top of #3240.

Note: I tried to add a test for this, but I couldn't reproduce it locally so I didn't.

### DIFF
--- a/fs/ensure_dir.ts
+++ b/fs/ensure_dir.ts
@@ -15,6 +15,12 @@ import { getFileInfoType } from "./_util.ts";
  */
 export async function ensureDir(dir: string | URL) {
   try {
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #3027, opened 2022-12-20
refactor(encoding/jsonc): unified dir and move to single-export files

timreichen:
ref: #2538

This PR moves all jsonc related stuff into the same dir.

encoding/jsonc.ts → encoding/jsonc/mod.ts
encoding/test_jsonc.ts → encoding/jsonc/mod_test.ts
encoding/testdata/*jsonc testsuite* → encoding/jsonc/testdata

splits logic into single export files.

⚠️ https://deno.land/std@$STD_VERSION/encoding/jsonc.ts is deprecated.

PR #3243, opened 2023-03-08
BREAKING(jsonc): move to top-level folder and single-export files

iuioiua:
Part of:
1. #3123
2. #2538
3. #2936

Supercedes #3027


kt3k (approved):
LGTM

PR #3244, opened 2023-03-08
Update _yaml/parse.ts: YAMLException doesn't exist

Atlas48:
YAMLException doesn't exist, changed documentation to actual name.

This is a single-line change that does not substantially effect how the script runs.


kt3k (approved):
LGTM. Thanks!

PR #3245, opened 2023-03-08
feat: add http/etag

kitsonk:
This is based on the etag handling in oak as well as the file_server as part of std and extracts that out into a re-usable library.


http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+  weak?: boolean;
+}
+
+function isFileInfo(value: unknown): value is FileInfo {
+  return Boolean(
+    value && typeof value === "object" && "mtime" in value && "size" in value,
+  );
+}
+
+async function calcEntity(
```
timreichen, line 55: Imo the should be written out, `calculateEntity` instead of `calcEntity`

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+) {
+  // a short circuit for zero length entities
+  if (!entity.length && !algorithm) {
+    return "811c9dc5";
+  }
+
+  return toHashString(await createHash(algorithm ?? DEFAULT_ALGORITHM, entity));
+}
+
+async function calcFileInfo(fileInfo: FileInfo, { algorithm }: ETagOptions) {
```
timreichen, line 67: Imo the should be written out, `calculateFileInfo` instead of `calcFileInfo`

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+ *   assert(etag);
+ *   if (!ifNoneMatch(ifNoneMatchValue, etag)) {
+ *     return new Response(null, { status: 304, headers: { etag } });
+ *   } else {
+ *     return new Response(body, { status: 200, headers: { etag } });
+ *   }
+ * });
+ * ```
+ */
+export function ifNoneMatch(
```
timreichen, line 181: Is this really needed? Couldn't this be done with `!ifMatch()`?

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+ *   assert(etag);
+ *   if (!ifMatchValue || ifMatch(ifMatchValue, etag)) {
+ *     return new Response(body, { status: 200, headers: { etag } });
+ *   } else {
+ *     return new Response(null, { status: 412, statusText: "Precondition Failed"});
+ *   }
+ * });
+ * ```
+ */
+export function ifMatch(
```
timreichen, line 137: Imo the name is a bit misleading. In the js landscape, `match` (in String, RegExp etc.) normally returns a matched content. I think `test` would make more sense as a name as it returns a boolean.

http/etag.ts
kitsonk: They are internal and not exported. The term is not unclear.

http/etag.ts
kitsonk: No the logic is quite different. If-None-Match uses weak matching, If-Match doesn't.

http/etag.ts
kitsonk: They align to the header names, ergo align to the logic the header represents.

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+  return HASHED_DENO_DEPLOYMENT_ID;
+}
+
+/** Calculate an ETag for an entity. When the entity is a specific set of data
+ * it will be fingerprinted as a "strong" tag, otherwise if it is just file
+ * information, it will be calculated as a weak tag.
+ *
+ * ```ts
+ * import { calculate } from "https://deno.land/std@$STD_VERSION/http/etag.ts";
+ * import { assert } from "https://deno.land/std@$STD_VERSION/_util/asserts.ts"
```
lino-levan, line 85: I don't think we should recommend that users use std's internal assert.

```suggestion
 * import { assert } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts"
```

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+ * See MDN's [`If-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match)
+ * article for more information on how to use this function.
+ *
+ * ```ts
+ * import {
+ *   calculate,
+ *   ifMatch,
+ * } from "https://deno.land/std@$STD_VERSION/http/etag.ts";
+ * import { serve } from "https://deno.land/std@$STD_VERSION/http/server.ts";
+ * import { assert } from "https://deno.land/std@$STD_VERSION/_util/asserts.ts"
```
lino-levan, line 121: Same as above

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+ * See MDN's [`If-None-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-None-Match)
+ * article for more information on how to use this function.
+ *
+ * ```ts
+ * import {
+ *   calculate,
+ *   ifNoneMatch,
+ * } from "https://deno.land/std@$STD_VERSION/http/etag.ts";
+ * import { serve } from "https://deno.land/std@$STD_VERSION/http/server.ts";
+ * import { assert } from "https://deno.land/std@$STD_VERSION/_util/asserts.ts"
```
lino-levan, line 165: Same as above

lino-levan (changes requested):
Outside of a few nitpicks, this looks good to me.

http/etag.ts
iuioiua: "calc" is widely-known to mean calculate. It's fine 🙂

http/etag.ts
iuioiua: Ditto ^

http/etag.ts
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * See further information on the `ETag` header on
+ * [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag).
+ *
+ * @module
+ */
+
+import { type DigestAlgorithm } from "../crypto/crypto.ts";
+import { toHashString } from "../crypto/to_hash_string.ts";
+import { createHash } from "../crypto/_util.ts";
```
kt3k, line 17: This introduces the dependency to the entire `std/crypto`. I feel that is too excessive for this module as hash algorithm doesn't matter very much for this module in my view. How about just using web crypto API for creating hash?

kitsonk: Fair point. I ported the stuff from file server, so there was consistency, but I agree it doesn't make a huge amount of sense.

kt3k (approved):
LGTM. Thanks!

http/util.ts
```diff
@@ -4,7 +4,10 @@
 import { Status, STATUS_TEXT } from "./http_status.ts";
 import { deepMerge } from "../collections/deep_merge.ts";
 
-/** Returns true if the etags match. Weak etag comparisons are handled. */
+/** Returns true if the etags match. Weak etag comparisons are handled.
+ *
+ * @deprecated (will be removed in 0.184.0) use `etag/ifMatch` and `etag/ifNoneMatch` instead.
```
kt3k, line 9: Let's deprecate at v0.187.0
```suggestion
 * @deprecated (will be removed after 0.187.0) use `etag/ifMatch` and `etag/ifNoneMatch` instead.
```

Issue #2461, opened 2022-07-22
assertEquals can't show a helpful diff and a provided message


stuartlangridge:
**Describe the bug**

`assertEquals` (from `std/testing`) shows a nice diff when comparing two multiline strings. However, if an optional `msg` is passed to `assertEquals` then it shows the message but _not_ the diff; there is no way to get both.

**Steps to Reproduce**

Obviously assertEquals could compare two arrays all in one go, but this is a simplified example. The key point is that a test may run more than one assert, but it's hard to know which one it failed on without outputting a message along with the assert, and then Deno.test suppresses the diff. 

```
$ cat deno-assert-example.js
import { assertEquals } from "https://deno.land/std@0.149.0/testing/asserts.ts";

Deno.test("Nice diff, but I don't know which entry the compare failed on", () => {
    const list1 = ["abc\ndef", "ghi\jkl", "mno\npqr", "stu\nvwx"];
    const list2 = ["abc\ndef", "ghi\jkl", "mno\npqr", "AHA\nvwx"];
    for (let i=0; i<list1.length; i++) {
        assertEquals(list1[i], list2[i]);
    }
});

Deno.test("See the entry index number, but no nice diff", () => {
    const list1 = ["abc\ndef", "ghi\jkl", "mno\npqr", "stu\nvwx"];
    const list2 = ["abc\ndef", "ghi\jkl", "mno\npqr", "AHA\nvwx"];
    for (let i=0; i<list1.length; i++) {
        assertEquals(list1[i], list2[i], `failed on entry ${i}`);
    }
});

$ deno test deno-assert-example.js
running 2 tests from .../deno-assert-example.js
test Nice diff, but I don't know which entry the compare failed on ... FAILED (9ms)
test See the entry index number, but no nice diff ... FAILED (3ms)

failures:

Nice diff, but I don't know which entry the compare failed on
AssertionError: Values are not equal:

    [Diff] Actual / Expected

-   stu\n
+   AHA\n
    vwx

    at assertEquals (https://deno.land/std@0.149.0/testing/asserts.ts:177:9)

See the entry index number, but no nice diff
AssertionError: failed on entry 3
    at assertEquals (https://deno.land/std@0.149.0/testing/asserts.ts:177:9)
```

**Expected behavior**

[...]

PR #3253, opened 2023-03-14
feat!: include `msg` in assertion errors

bajtos:
Rework assertion functions to always include custom `msg` text in the failure message when `msg` is provided.

Since this change touches almost all assertion functions, I used the opportunity to improve the assertion error messages to be more consistent.

This commit also fixes `assertObjectMatch` to propagate the `msg` argument to the underlying `assertEquals()` call. (It was always ignored before this change.)

**BREAKING CHANGE:**
Assertion error messages have slightly different content now. This may break downstream consumers that depend on the exact error message strings.

Resolve https://github.com/denoland/deno_std/issues/2461

_Note: the changes are best viewed with whitespace-only changes ignored._

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -308,7 +308,9 @@ export function assertNotStrictEquals<T>(
 
   const msgSuffix = msg ? `: ${msg}` : ".";
   throw new AssertionError(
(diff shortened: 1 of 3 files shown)


kt3k (approved):
Nice! LGTM

lino-levan:
LGTM, nice clean up!

Issue #3256, opened 2023-03-16
My deno deploy site is broken because of this bug. PLEASE HELP


JacobWeisenburger:
`std@0.180.0/encoding/_yaml/loader/loader.ts` is missing, this is breaking my deno deploy site.

https://deno.land/std@0.180.0/encoding/_yaml/loader/loader.ts

![image](https://user-images.githubusercontent.com/31667350/225654319-0f4190f9-6fa8-431f-b2ca-c6541b24131d.png)

![image](https://user-images.githubusercontent.com/31667350/225654966-f6c8529f-0a44-4b42-8e1c-a697f342f91c.png)

**Describe the bug**
```
Error The deployment failed: HTTP status client error (404 Not Found) for url (https://deno.land/std@0.180.0/encoding/_yaml/loader/loader.ts)
```

**Steps to Reproduce**
1. push to my github repo
2. deno deploy tries to download the code to run it
3. the above error stops the deploy

**Expected behavior**
I'm not using `std@0.180.0`, so this shouldn't be breaking my deployment.

**Environment**
```ts
// import_map.json
{
    "imports": {
        "std/": "https://deno.land/std@0.179.0/"
    }
}
```

deno 1.31.3 (release, x86_64-pc-windows-msvc)
v8 11.0.226.19
typescript 4.9.4

JacobWeisenburger:
Someone was able to help me on the deno discord.

One of my dependencies was importing a file that no longer exists in the `std@0.180.0`. So he told me to remap the import url in my `import_map.json` and that worked.

PR #3257, opened 2023-03-16
feat(yaml): make APIs public

iuioiua:
Looking at [js-yaml](https://github.com/nodeca/js-yaml), it appears as though `load()`, `dump()`, and related functions/interfaces should be public. This change does that.

It's possible that interfaces within `loader/loader_state.ts` and `dumper/dumper_state.ts` should not be included in the public API. A second opinion would be appreciated.

Related:
* #3256


kt3k:
`load` and `dump` are very similar (almost identical) to `parse` and `stringify`. It doesn't make much sense to export them in addition to the existing APIs in my view.

lino-levan:
We should investigate why https://github.com/ubersl0th/markdown/pull/17 needed to use `load` over `parse`. Maybe there is a use case?

iuioiua:
> `load` and `dump` are very similar (almost identical) to `parse` and `stringify`. It doesn't make much sense to export them in addition to the existing APIs in my view.

Oh, yeah, they're identical. That was something I should've obviously checked.

> We should investigate why [ubersl0th/markdown#17](https://github.com/ubersl0th/markdown/pull/17) needed to use `load` over `parse`. Maybe there is a use case?

Check out https://github.com/denoland/deno_std/blob/main/yaml/parse.ts

lino-levan:
They are quite literally identical, nevermind.

PR #3261, opened 2023-03-17
refactor(CI): pin runner to macOS-12

iuioiua:
For the same reason that Ubuntu and Windows runner versions are pinned.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         os:
           - ubuntu-22.04
           - windows-2022
-          - macOS-latest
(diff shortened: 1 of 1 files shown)


lino-levan:
I'm curious on what the reasoning was for pinning the versions of the other runners?

kt3k:
`denoland/deno` uses explicit version number since https://github.com/denoland/deno/pull/6884 and [this](https://github.com/denoland/deno/pull/6880#issuecomment-664000672) looks like the reasoning for it. I agree with it. Let's use the explicit version

kt3k (approved):
LGTM

PR #3264, opened 2023-03-17
chore(encoding/yaml): restore deprecated exports

kt3k:
`x/markdown@v2.0.0` depends on this file with non-versioned std url.

https://github.com/ubersl0th/markdown/blob/v2.0.0/src/block-lexer.ts#L25

### DIFF
--- a/encoding/_yaml/loader/loader.ts
+++ b/encoding/_yaml/loader/loader.ts
@@ -1,0 +1,8 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+export {
+  /** @deprecated (will be removed after 0.182.0) use `parse` from std/yaml/mod.ts instead */
+  load,
+  /** @deprecated (will be removed after 0.182.0) use `parseAll` from std/yaml/mod.ts instead */
+  loadAll,
+} from "../../../yaml/_loader/loader.ts";


iuioiua:
From the README, which `x/markdown` appears to have missed:

> Don't link to / import any module whose path:
> * Has a name or parent with an underscore prefix: _foo.ts, _util/bar.ts.
> ...
>
> These elements are not considered part of the public API, thus no stability is guaranteed for them.

This seems like a simple correction that needs to be made in the module, as I suggest [here](https://github.com/ubersl0th/markdown/pull/17#issuecomment-1473046791).

kt3k:
Even when it's fixed in `x/markdown`, the tools which depend on `x/markdown@v2.0.0` are still affected (for example https://github.com/kawarimidoll/deno-diplodocus/blob/0.0.3/deps.ts#L7 ), and I guess some of them are inactive..

timreichen:
I don't think this is really a std problem, since the readme is clear on underscore files.

But a simple fix is to make an importmap entry to fix the problem:
```json
"https://deno.land/std/encoding/_yaml/loader/loader.ts": "https://deno.land/std@0.180.0/yaml/_loader/loader.ts"
```

lino-levan:
I'm -1 on this as well. Not only is `/x/markdown@2.0.0` breaking the stability rules (for no reason), they are also using unversioned imports. I don't think any action is really necessary on the side of std. @timreichen's point should be good enough for people still relying on the broken version of `/x/markdown` in the future.

ayame113:
I am +1.
x/markdown has been using versionless imports since July 2020. I think removing `_yaml/loader/loader.ts` without a transition period would have too much impact.
Deleting this file is easy, but it's not a good developer experience when people suddenly get a Module not found error. Even if we leave this file, I don't think any particular problem will occur.
I think we should not remove `_yaml/loader/loader.ts` for the time being to give people time to migrate.

iuioiua:
Again, such changes should be expected for internal code and importing such internal code is clearly advised against in the README.

This is a mistake on the part of the module maintainer and shouldn't be compensated for here in this repo.

timreichen:
There are two solutions for this problem which are an immediate fix:
1. let the user fix the problem with an importmap entry as stated above
2. let the maintainer fix the problem in the mod

If the module is not maintained, we can discuss including a markdown mod in std directly to circumvent the problem altogether.

Std should not compensate for disregarding guidelines and stay as clean as possible. Else it will become a legacy support nightmare over time.

ayame113:
People using deno deploy with `x/markdown@2.0.0` are faced with the situation where one day they suddenly can't deploy to production. From the user's point of view, the import of `x/markdown@2.0.0` is correctly versioned, but the deployment to production fails. The users importing `x/markdown@2.0.0` are totally innocent and should not be affected.

Another suggestion from me, how about adding `console.warn("This file will be removed in v0.182.0. If any of your dependencies import this file, please switch to the newer version.")` to the beginning of the file to encourage people to switch to newer versions of `x/markdown`? That way I think I can give a warning to the user who is importing this file.
Then I think we can remove this file in subsequent releases.

iuioiua:
> People using deno deploy with `x/markdown@2.0.0` are faced with the situation where one day they suddenly can't deploy to production. From the user's point of view, the import of `x/markdown@2.0.0` is correctly versioned, but the deployment to production fails. In my opinion, the users importing `x/markdown@2.0.0` are totally innocent and should not be affected.

Well, it's a bug. It's up to the maintainers to fix the bug and then for consumers to upgrade to the patched version. Unfortunately, it looks as though the maintainers may no longer be maintaining the module. It's been 3 years since the last update, and [the PR to fix this critical issue](ubersl0th/markdown#17) has not had any updates. Perhaps, it is worth creating `std/markdown`, as @timreichen says...

> Another suggestion from me, how about adding `console.warn("This file will be removed in v0.182.0. If any of your dependencies import this file, please switch to the newer version.")` to the beginning of the file to encourage people to switch to newer versions of `x/markdown`? That way I think I can give a warning to the user who is importing this file.

I like the idea of `console.warn()`ing deprecations at the top of deprecated files. It makes things super-obvious without being annoying by only being printed upon startup. What do the rest of us think?

kt3k:
Discussed this among core members, and we decided to not land this. Sorry for the confusions!

We recommend using import maps (like the one given in https://github.com/denoland/deno_std/pull/3264#issuecomment-1473718031) to fix this if the users can't replace `x/markdown@v2.0.0` dependency for some reason

PR #3270, opened 2023-03-21
feat(csv): add `CsvStringifyStream`

### LINKED ISSUE #3258: CSVStream: support outputting to a writable stream

I like to use `CSVStream` for reading CSV files. I want to also stream my output but today `CSVStream` only supports readables and not writables.

A way to use `CSVStream` (or a similar class) with `TextEncoderStream` to pipe to a writable (e.g. to `Deno.stdout.writable`).

forking the code in "https://deno.land/std@0.180.0/csv/mod.ts" for `stringify` to accept a "write" callback so that a different destination other than an in-memory string can be supplied (`data: DataItem[]` would also have to be changed to `data: Iterable<DataItem>`)

uki00a:
Closes #3258

This PR adds `CsvStringifyStream` (`std/csv/csv_stringify_stream`) to support streaming CSV records

### DIFF
--- a/csv/csv_stringify_stream.ts
+++ b/csv/csv_stringify_stream.ts
@@ -1,0 +1,67 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { stringify } from "./stringify.ts";
+
+export interface CsvStringifyStreamOptions {
+  /**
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM. Nice work!

PR #3277, opened 2023-03-23
test(crypto): update 'deno run' usage

kt3k:
It seems that `deno run -` now handles the input as JavaScript by default. This PR explicitly specify it as `ts` in `--ext` option to fix the test cases.

maybe related: https://github.com/denoland/deno/pull/17172


kt3k:
The change was unintentional and it's now fixed in https://github.com/denoland/deno/pull/18391

This change is not necessary now.

PR #3289, opened 2023-03-28
fix(fs): fix NotFound error when moving src to itself with overwrite: true

### LINKED ISSUE #3288: `std/fs/move` doesn't check if `src` and `dest` differ
**Describe the bug**
The actual issue is that `move` with `overwrite` option set to `true` removes `dest` as the first action. The result is that `src` ends up deleted and deno complains that it can't find the path to rename.

**Steps to Reproduce**
1. Execute in a directory containing directory `a`:
```ts
move("a", "a", { overwrite: true });
```
2. Directory `a` is gone.

**Expected behavior**

Nothing happens.

**Environment**

- OS: Windows 10
- deno version: 1.31.1
- std version: 0.181.0

tetraphobia:
Resolves #3288

Moving a file to itself with the overwrite option set to true results in the original file being deleted and a NotFound error being thrown. This implements a simple guard against these scenarios.


fs/move.ts
```diff
@@ -30,6 +31,14 @@ export async function move(
     );
   }
 
+  const srcAbsolute = src instanceof URL ? src.pathname : resolve(src)
+  const destAbsolute = dest instanceof URL ? dest.pathname : resolve(dest)
+
+  if (srcAbsolute === destAbsolute) {
+    await Deno.rename(src, dest)
```
kt3k, line 38: Is this rename meaningful?

fs/move.ts
```diff
@@ -30,6 +31,14 @@ export async function move(
     );
   }
 
+  const srcAbsolute = src instanceof URL ? src.pathname : resolve(src)
+  const destAbsolute = dest instanceof URL ? dest.pathname : resolve(dest)
+
+  if (srcAbsolute === destAbsolute) {
```
kt3k, line 37: Can this check made into a common util in `fs/_util.ts`? (like `isSubdir`)

fs/move_test.ts
```diff
@@ -374,3 +374,93 @@ Deno.test("moveSyncIntoSubDir", function () {
...
+  // ensure test data exists
+  await Deno.mkdir(srcDir, { recursive: true });
+  assert(await Deno.lstat(srcDir));
+  await Deno.writeFile(srcFile, srcContent);
+
+  // move file to itself with overwrite
+  await move(srcFile, srcFile, { overwrite: true });
+  await move(srcFile, srcUrl, { overwrite: true });
+  await move(srcUrl, srcFile, { overwrite: true });
+  await move(srcUrl, srcUrl, { overwrite: true });
```
kt3k, line 415: I think this should move directory, according to the test title
```suggestion
  await move(srcDir, srcDir, { overwrite: true });
```

fs/move_test.ts
```diff
@@ -374,3 +374,93 @@ Deno.test("moveSyncIntoSubDir", function () {
...
+
+  // test file should be untouched
+  assertEquals(new TextDecoder().decode(await Deno.readFile(srcFile)), "src");
```
kt3k, line 395: You can use `Deno.readTextFile()` and skip using `TextDecoder`

fs/move.ts
tetraphobia: I don't feel like it is, since the result will be the same. I put the rename there just in case it would be overriding some behavior during rename.

fs/move.ts
tetraphobia: Will do.

fs/_util.ts
```diff
@@ -1,6 +1,22 @@
...
+
+/**
+ * Test whether `src` and `dest` resolve to the same location
+ * @param src src file path
+ * @param dest dest file path
+ */
+export function isSamePath(
+  src: string | URL,
+  dest: string | URL,
+  sep: string = path.sep,
```
kt3k, line 13: The 3rd parameter looks only used in testing. Do we really need this?

kt3k: (We usually don't need to specify platform specific params explicitly because they are abstracted by path module)

tetraphobia: > The 3rd parameter looks only used in testing. Do we really need this?

Idealy not. I had some trouble producing consistent results in the tests involving mixed URL and string params using win32 paths. It's probably just a testing issue, and I will look for a better solution in the meantime.

tetraphobia: Today I discovered the `ignore` option and fixed the issue :)

kt3k (approved):
LGTM. Nice work!

Issue #3291, opened 2023-03-29
slice() method of BytesList class not working


mjljm:
**Describe the bug**
The method does not work.

**Steps to Reproduce**

```ts
import { BytesList } from "https://deno.land/std@0.181.0/bytes/bytes_list.ts";

const b = new BytesList();
const te = new TextEncoder();
const td = new TextDecoder();

const u1 = te.encode("Some dummy example to  show it does not work");
b.add(u1);

const res1 = b.slice(5, 9);
console.log(td.decode(res1)); // got "some" instead of "dumm"
```

**Expected behavior**

Result should be "dumm", not "some"

**Environment**

- OS: Windows 10
- deno version: 1.32.1
- std version: 0.181.0

I looked into bytes_list_tests.ts. It works because tests are made against Uint8Arrays filled with 0s. So only the length of the result is tested, not its contents.

I submitted pull request #3292 to correct this bug.

kamilogorek:
Fixed in https://github.com/denoland/deno_std/pull/3292

lino-levan:
As @kamilogorek correctly points out, it was fixed in #3292. Could this be closed @kt3k?

PR #3292, opened 2023-03-29
Corrected slice() method of BytesList class and added 2 tests

mjljm:
Should correct issue #3291 :
- corrected the slice() method of the BytesList class in bytes/bytes_list.ts
- added 2 tests in bytes/bytes_lists_test.ts


kt3k (approved):
LGTM. Good catch! Thanks

Issue #3294, opened 2023-03-30
Add isUTCLeap function


RyanClementsHax:
**Is your feature request related to a problem? Please describe.**

The difference between [Date.prototype.getFullYear](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getFullYear) and [Date.prototype.getUTCFullYear](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getUTCFullYear) can be surprising for those working with the [isLeap](https://deno.land/std@0.102.0/datetime/mod.ts?s=isLeap) which depends upon the former to determine the year when passed a `Date` object.

This means that `isLeap('2000-01-01')` returns true if you live east of GMT, but false if you live west of GMT.

**Describe the solution you'd like**

A function named `isUTCLeap` that works the same as `isLeap` but returns whether the date given to it is a leap year converted to UTC time.

This means that `isUTCLeap('2000-01-01')` always returns true.

I'm not picky on the name, but I chose `isUTCLeap` instead of `isLeapUTC` to keep with the naming conventions of `getFullYear` and `getUTCFullYear`.

**Describe alternatives you've considered**

I thought of adding a parameter to the `isLeap` function, but it seemed more idiomatic to make a new function for this.

**Side note**

This fixes the assertions in `is_leap_test.ts` where some would fail on machines set to GMT-<anything> time. I'm ok with just fixing the tests to not assert against edge cases like this, but I also figured that this would be a handy, and easy feature to add so I just went ahead and did the work.

kt3k:
closed by #3295

PR #3295, opened 2023-03-30
feat(isUtcLeap): Add function to determine a leap year in UTC to avoi…

RyanClementsHax:
…d timezone issues

- Add isUtcLeap that is the same as isLeap but uses getUTCFullYear to avoid timezone issues
- Add tests for isUtcLeap
- Fix tests for isLeap that only failed on machines running in GMT-<anything>

Implements https://github.com/denoland/deno_std/issues/3294


lionel-rowe:
TBH the whole idea of determining whether something's a leap year from a `Date` object seems massively error-prone and difficult to reason about. For example, a developer might reasonably believe `isUTCLeap` will interpret a `Date` object as if it's UTC even when it was constructed in a non-UTC timezone, which isn't true.

```js
// in my current timezone, UTC+8
isLeap(new Date(2000, 0 /* zero-indexed month */, 1)) // true (consistently true everywhere)
isUTCLeap(new Date(2000, 0 /* zero-indexed month */, 1))) // false (but would be true if I lived somewhere with a zero or negative UTC offset)
isLeap(new Date("2000-01-01")) // true (but would be false if I lived somewhere with a negative UTC offset)
isUTCLeap(new Date("2000-01-01")) // true (consistently true everywhere)
```

IMO it'd be better to simply deprecate or remove the `(year: Date) => boolean` signature.

RyanClementsHax:
I'm ok with either. I can update the PR if needed. How will I know if this is the way we want to go? Does a maintainer need to give the final say?

lionel-rowe:
> I'm ok with either. I can update the PR if needed. How will I know if this is the way we want to go? Does a maintainer need to give the final say?

I'd assume so — I'm just a random contributor so I can't speak authoritatively, just adding my 2 cents as I've come across far too many `Date`-related timezone bugs. Incidentally, this type of confusion is a big part of the motivation for [`Temporal`](https://tc39.es/proposal-temporal/docs/index.html), though that's still at stage 3 and AFAIK not implemented in any browsers yet, even behind a flag.

iuioiua:
How come this functionality isn't being added to `isLeap()` instead of having a separate function? Having 2 separate functions that have so much overlap may confuse users.

If a separate function is justified, it should be named `isUtcLeap()` as acronyms should written in camelCase (see #2582).

Disclaimer: I have not looked at this implementation or discussion, so please excuse any ignorance.

RyanClementsHax:
I could add an option object to is leap to toggle this feature but it seemed more idiomatic to make a second function to keep in line with the api of JavaScript's date object. This is also why I named the function the way I did. For more details see the issue linked.

I'm not opposed to changing these details btw. I don’t feel very strongly about them.

iuioiua:
I see, I see. In that case, a separate function is reasonable.

Again though, the function name would have to be `isUtcLeap()` with the acronym being in camelCase 👍🏾

RyanClementsHax:
I've changed the function to be named `isUtcLeap`

kt3k (approved):
LGTM

Issue #3296, opened 2023-03-31
Add string width calculation for TTY-based text layout


lionel-rowe:
Split from https://github.com/denoland/deno/issues/18477.

**Is your feature request related to a problem? Please describe.**

The Deno runtime exposes `Deno.consoleSize`, which gives the physical width of the console in TTY-like environements in columns and rows. For printable ASCII text and many other characters, 1 column == 1 character; however, across all Unicode characters, there also exist many that are zero columns wide, and many others that are two columns wide.

In order for `Deno.consoleSize` to be generally reliable for TTY-based text layout, such as calculating where line wrap will occur, some method of getting the width of a string also needs to be exposed.

**Describe the solution you'd like**

Internally, the Deno repl uses the `rustyline` crate, a readline implementation for Rust, which in turn relies on the `unicode_width` crate to measure text width. This is also the crate [used by the Rust compiler when determining how to "underline" errors](https://www.reddit.com/r/rust/comments/gpw2ra/how_is_the_rust_compiler_able_to_tell_the_visible/).

`unicode_width` is based on the character width rules laid out in [Unicode Standard Annex #11](http://www.unicode.org/reports/tr11/).

It does _not_ deal with ANSI escape codes (such as `"\x1b[36mCYAN\x1b[0m"`, which is displayed in cyan color in TTYs but remains 4 columns wide).

Adding a port of `unicode_width` to `deno_std` would allow userland code to implement TTY-based text layout in a way that's both standards-based and gives results consistent with tooling used in the runtime.
**Describe alternatives you've considered**

My initial preference was for this to be exposed by the runtime itself, but @dsherret convinced me that's not necessary.

The code for rendering `console.table` results in the runtime uses its own logic, which is significantly less rigorous than `unicode_width`. It simply strips ANSI colors and runs a few regexes of known double-width chars, but doesn't handle zero-width chars.

ANSI-stripping functionality is already provided in deno_std by `fmt/stripColor.ts`.

jsejcksn:
Link to review discussion thread related to visual columns per line: https://github.com/denoland/deno_std/pull/3297#discussion_r1166289430

PR #3297, opened 2023-03-31
feat(string_width): add stringWidth for TTY text layout

lionel-rowe:
Closes https://github.com/denoland/deno_std/issues/3296.

Get the expected physical column width of a string in TTY-like environments.

Combines a TypeScript port of the [unicode-width Rust crate](https://github.com/unicode-rs/unicode-width/), which looks up the nominal width of Unicode characters, with a port of [chalk/strip-ansi](https://github.com/chalk/strip-ansi), which strips out ANSI escape sequences.


iuioiua:
I'm -1 on this PR as I don't believe it's worth the added complexity of a Rust implementation. We recently deprecated the Rust implementation of `encoding/varint` for this same reason.

I'm also unsure whether this functionality is sought after enough to be in the Standard Library. I don't feel strongly about this, though.

That's not to say such functionality would be valuable in a 3rd party module! 🙂

lionel-rowe:
> I don't believe it's worth the added complexity of a Rust implementation

To clarify, this is a TypeScript port of a Rust crate, not a WASM implementation. The Rust code is just a thin wrapper around the crate, used for checking equivalency of results during testing, and isn't imported by the module code itself.

string_width/_strip_ansi.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// Ported from chalk/strip-ansi, Copyright (c) Sindre Sorhus <sindresorhus@gmail.com> (https://sindresorhus.com). MIT License.
+
+export function stripAnsi(str: string) {
```
kt3k, line 4: This already exists in `fmt/colors.ts` as `stripColor`

string_width/README.md
```diff
@@ -0,0 +1,75 @@
+# string_width
```
kt3k, line 1: We don't create README in deno_std anymore. Instead we write module-level documentation in `mod.ts` with `@module` jsdoc tag.

string_width/string_width.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+import { stripAnsi } from "https://deno.land/std@$STD_VERSION/string_width/_strip_ansi.ts";
+import { unicodeWidth } from "https://deno.land/std@$STD_VERSION/string_width/_unicode_width.ts";
```
kt3k, line 4: Use relative module specifiers (like `./_unicode_width.ts`) in the runtime code of the module. (`https://deno.land/std@$STD_VERSION` should be used only in the examples in docs.)

kt3k:
`std/string_width` top level module sounds too specific to me. How about `std/console` (or something similar)?

string_width/string_width_test.ts
```diff
@@ -0,0 +1,161 @@
...
+    await t.step("NFD", () => {
+      const nfd = str.normalize("NFD");
+
+      assertEquals(nfd.length, 2);
+      assertEquals(stringWidth(nfd), 1);
+    });
+  });
+});
+
+Deno.test("docs", async (t) => {
```
kt3k, line 55: Docs (examples) are at least type checked by `deno test --doc` command (and we have a plan to run them in the future). I think we don't need to run these examples as a unit test case here.

kt3k:
> `string_width/_scripts/generateData.ts`

We prefer `_tools` over `_scripts`. Also we usually use snake case for file names (`generateData.ts` -> `generate_data.ts`

string_width/string_width_test.ts
```diff
@@ -0,0 +1,161 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+import { stringWidth } from "https://deno.land/std@$STD_VERSION/string_width/string_width.ts";
```
kt3k, line 3: These module specifiers also should be relative paths (`./string_width.ts`)

kt3k:
It's bikeshedding topic, but I'm wondering between 2 api name candidates: `stringWidth` and `unicodeWidth`. Does anyone have preference between these? cc @iuioiua @lino-levan @timreichen @jsejcksn

lionel-rowe:
> It's bikeshedding topic, but I'm wondering between 2 api name candidates: `stringWidth` and `unicodeWidth`

FWIW, while I don't think the name `unicodeWidth` is ideal as it could perpetuate the myth that Unicode strings are some special type of strings distinct from "ASCII strings", it at least gives a hint at the type of functionality it provides (grabbing width data for the constituent characters as specified by Unicode), while not implying it does something it doesn't (e.g. stripping ANSI codes to give a more accurate physical width in TTYs). A previous version of this PR used the name `stringWidth` as it combined both functions, until @kt3k pointed out that the ANSI-stripping functionality is already provided in `fmt/colors.js`.

lino-levan:
> It's bikeshedding topic, but I'm wondering between 2 api name candidates: stringWidth and unicodeWidth

I've done a bit of pondering (my original response before I cut it down to this was ~half a page long), but I think that `stringWidth` makes the most sense to me. At the end of the day, this method needs to be discoverable and I think that even if `stringWidth` isn't the best description of what it doesn't do, it is a very clear description of what it does. `unicodeWidth` is harder to search for and the functionality (imo) isn't immediately obvious. That being said, I don't feel super strongly either way. Just my 2 cents.

console/unicode_width_test.ts
```diff
@@ -0,0 +1,38 @@
...
+      assertEquals(unicodeWidth(nfc), 1);
+    });
+
+    await t.step("NFD", () => {
+      const nfd = str.normalize("NFD");
+
+      assertEquals(nfd.length, 2);
+      assertEquals(unicodeWidth(nfd), 1);
+    });
+  });
```
jsejcksn, line 37:
```suggestion
  });

  await t.step("Emoji", () => {
    assertEquals(unicodeWidth("👩"), 2); // Woman
    assertEquals(unicodeWidth("🔬"), 2); // Microscope
    assertEquals(unicodeWidth("👩‍🔬"), 4); // Woman Scientist
  });
```

I think additional tests could be useful — they serve as documentation of behavior. That's one from the upstream [readme](https://github.com/unicode-rs/unicode-width/blob/v0.1.17/README.md?plain=1#L27-L40).

jsejcksn:
> It's bikeshedding topic, but I'm wondering between 2 api name candidates: `stringWidth` and `unicodeWidth`. Does anyone have preference between these?

I'm in favor of keeping `unicodeWidth`.

- Socratic: What's the difference between _string width_ and _string length_?

  The term _width_ is ambiguous/overloaded. The usage here is very [technical](https://www.unicode.org/reports/tr11/), and, for that reason alone, I think it makes sense to use _unicode_ because it helps disambiguate.

- This is a port and I think porting the name is sensible as well — there's nothing special about the context of Deno which invalidates it (compared to other ports where idioms don't translate), and keeping the name will give confidence to Deno users who are already familiar with the upstream library.

console/unicode_width_test.ts
kt3k: Good point. I don't think this is an intentional behavior. What we want here is to calculate the visual width of characters in terminals.

Note: `npm:string-width` returns 2 for 👩‍🔬

```js
import stringWidth from 'npm:string-width';
console.log(stringWidth("👩‍🔬"));
```

jsejcksn: @kt3k

> I don't think this is an intentional behavior.

It is intentional. The term _width_ **doesn't** mean "visual column width" in either the upstream crate or the unicode technical report from which it is derived.

> What we want here is to calculate the visual width of characters in terminals.

Are you sure? [I don't think that's going to be possible without much deeper knowledge about the terminal environment](https://stackoverflow.com/q/71228251/438273) — and I'm not sure whether the required info can be reliably obtained (especially cross-platform in many terminals).... at the least, it will be a difficult task.

> Note: `npm:string-width` returns 2 for 👩‍🔬

`👩‍🔬` occupies `1` column in my current terminal environment.

kt3k: > It is intentional. The term width doesn't mean "visual column width" in either the upstream crate or the unicode technical report from which it is derived.

That's true for the upstream crate, but I don't think it's true for this module.

This PR starts from this issue https://github.com/denoland/deno/issues/18477 (and https://github.com/denoland/deno_std/issues/3296 )

We want to know how much characters can fit in `Deno.consoleSize()`

kt3k: > 👩‍🔬 occupies 1 column in my current terminal environment.

What OS and terminal do you use? How does the script `console.log("👩‍🔬".repeat(Deno.consoleSize().columns))` render in your terminal?

jsejcksn: @kt3k

> This PR starts from this issue [denoland/deno#18477](https://github.com/denoland/deno/issues/18477) (and #3296 )
> 
> We want to know how much characters can fit in `Deno.consoleSize()`

Understood — thanks for the linked information.

> What OS and terminal do you use?

```
% sw_vers
ProductName:		macOS
ProductVersion:		13.3.1
BuildVersion:		22E261
```

```
% system_profiler -json SPApplicationsDataType | gojq '.SPApplicationsDataType[] | select(._name=="Terminal")'
{
  "_name": "Terminal",
  "arch_kind": "arch_arm_i64",
  "lastModified": "2023-04-01T16:46:50Z",
  "obtained_from": "apple",
  "path": "/System/Applications/Utilities/Terminal.app",
  "signed_by": [
    "Software Signing",
    "Apple Code Signing Certification Authority",
    "Apple Root CA"
  ],
  "version": "2.13"
}
```

> How does the script `console.log("👩‍🔬".repeat(Deno.consoleSize().columns))` render in your terminal?

```
% deno eval 'console.log("👩‍🔬".repeat(Deno.consoleSize().columns))'
👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬👩‍🔬
```

I'm also using a customized profile in Terminal.

kt3k: Thanks for sharing!

The way Terminal.app handles it looks strange (I was also able to reproduce it.)

On the other hand, iTerm2 seems handling the combined emoji in a better way.

In iTerm2
Screen Shot 2023-04-15 at 18 30 33

In the above (iterm2) case returning the width (or columns?) 2 feels sensible to me.

Anyway the combined emoji cases look complicated and even having incompatibilities among terminal software. I'd suggest we'd leave this as is and iterate on them later.

jsejcksn: > I'd suggest we'd leave this as is and iterate on them later.

Does that mean including the [suggested test](https://github.com/denoland/deno_std/pull/3297#discussion_r1166289430)?

timreichen:
`stringWidth` seems more familiar to me, but I don't have a strong opinion.

console/unicode_width_test.ts
kt3k: Let's add them with comments referencing to this PR

jsejcksn: > Let's add them with comments referencing to this PR

[Comment added](https://github.com/denoland/deno_std/pull/3297#discussion_r1166289430)

console/unicode_width_test.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  await t.step("Emoji", () => {
+    assertEquals(unicodeWidth("👩"), 2); // Woman
+    assertEquals(unicodeWidth("🔬"), 2); // Microscope
+    // Note: Returns 4 for the below case, following the upstream crate
+    // `unicode_width`. Another possible might be returning 2, which is what
```
kt3k, line 43:
```suggestion
    // `unicode_width`. Another possibility might be returning 2, which is what
```

kt3k (approved):
LGTM. Thanks for your contribution @lionel-rowe 

Also thanks all for discussions and reviews

PR #3300, opened 2023-04-02
feat(testing): add .skip alias to bdd test API

marvinhagemeister:
This PR adds `.skip()` as an alias of `.ignore()`. That alias is much more common than `.ignore()` in popular bdd-style testing APIs in other test runners.

| Test runner | Skipping APIs | 
|---|---|
| [mocha](https://github.com/mochajs/mocha) | `describe.skip()`, `it.skip()` |
| [jasmine](https://github.com/jasmine/jasmine) | `xdescribe()`, `xit()` |
| [jest](https://github.com/facebook/jest) | `describe.skip()`, `it.skip()` or `xdescribe()`, `xit()` |
| [vitest](https://vitest.dev/guide/filtering.html#skipping-suites-and-tests) | `describe.skip()`, `it.skip()` |

### DIFF
--- a/testing/bdd.ts
+++ b/testing/bdd.ts
@@ -439,6 +439,12 @@ export interface it {
 
   /** Registers an individual test case with ignore set to true. */
   ignore<T>(...args: ItArgs<T>): void;
+
+  /**
+   * Registers an individual test case with ignore set to true. Alias of
+   * `.ignore()`.
+   */
+  skip<T>(...args: ItArgs<T>): void;
 }
 
 /** Registers an individual test case. */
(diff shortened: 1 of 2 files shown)


kt3k:
Adding `.skip` makes sense to me (`.ignore` looks exceptional on that design)

I don't think we should add `xit` or `xdescribe`. They are useful only when these keywords are all global variables (you can quickly switch between `it` <-> `xit` in that case). If those are exported via ES modules, you need to add `import { xit } from ...` before switching to `xit` (and also delete that import when you stop using it). So they are cumbersome to use and have little value in my view.

marvinhagemeister:
Fair point! I've updated the PR to only add `.skip()`.

PR #3302, opened 2023-04-03
test(datetime): make isLeap test non-sensitive to timezone

### LINKED ISSUE #3299: is_leap tests are sensitive to time zone
**Describe the bug**

The `is_leap` function works, but the test cases construct a date that is sensitive to time zones.

https://github.com/denoland/deno_std/commit/8cf6b45e2cdb202235945928dcec641d390ae579

The test should probably use a date other than Jan 1 to ensure that the year is consistent in all time zones, for example:

```
> new Date("1972-01-01").getFullYear()
1971
> new Date("1972-02-01").getFullYear()
1972
```

kt3k:
This PR makes isLeap test cases non-sensitive to timezone.

If the timezone offset is < 0, `new Date("1970-01-01").getFullYear()` becomes `1969`, for example, and the current assertions don't pass.

Screen Shot 2023-04-03 at 1 42 36

closes #3299

### DIFF
--- a/datetime/is_leap_test.ts
+++ b/datetime/is_leap_test.ts
@@ -9,9 +9,9 @@ Deno.test({
     assert(isLeap(2000));
     assert(!isLeap(2003));
     assert(!isLeap(2007));
-    assert(!isLeap(new Date("1970-01-01")));
-    assert(isLeap(new Date("1972-01-01")));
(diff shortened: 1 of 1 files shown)


kt3k:
Closing in favor of #3295

PR #3305, opened 2023-04-04
fix(path): Remove non-node symbols

RyanClementsHax:
Addresses part of https://github.com/denoland/deno/issues/18177

I didn't add the removal of `fromFileUrl` or `toFileUrl` because they are heavily depended upon in deno core


kt3k:
https://github.com/denoland/deno/issues/18177 is about `node:path`, but this PR changes `std/path`, which is not necessarily compatible with `node:path`.

I think what we want to do in denoland/deno#18177 is removing those APIs from Node.js polyfills in Deno CLI repo (https://github.com/denoland/deno/blob/686fe477496cf6eee398eb8ac195692528d3f427/ext/node/polyfills/path/mod.ts)

RyanClementsHax:
Oh my bad. What is the difference between the two modules?

I'll make the change in Deno core

RyanClementsHax:
Oh that makes sense ty

PR #3306, opened 2023-04-04
feat(dotenv): allow reading from `.env` files without granting env access

jsejcksn:
This PR improves security/privacy by allowing the user to read env data from `.env` files without granting actual environment permissions.

Continuation of [#3221 (comment)](https://github.com/denoland/deno_std/pull/3221#issuecomment-1444784865):

> Note to self and for a potential follow up PR: After refactoring the new test to be stricter, I realized that there doesn't seem to be a way to load environment variables purely from a file: without granting access to at least some actual environment variables. In my view, this seems like an oversight and could be fixed rather easily by:
> 
> * not using an empty array `[]` as the default parameter for `restrictEnvAccessTo` in most of the functions in this module, and
> * refactoring `readEnv` to interpret the case of an empty array as "allow access to no variables" (a no-op), and just return `{}` without attempting to use the `Deno.env` API at all

Technically, I think this is a breaking change — but if I'm wrong about that, please feel free to change the title. There are no tests to support my hypothesis that it's a breaking change, but I think that it could break existing code if a consumer previously provided an explicit empty array and expected for variables to be read from the environment.

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -178,7 +178,7 @@ const RE_ExpandValue =
 
 export function parse(
   rawDotenv: string,
-  restrictEnvAccessTo: StringList = [],
+  restrictEnvAccessTo?: StringList,
 ): Record<string, string> {
   const env: Record<string, string> = {};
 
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks

> Technically, I think this is a breaking change

I agree, but the default behavior and non-empty `restrictEnvAccessTo` usages are the same. I think this can be considered as a new feature.

PR #3309, opened 2023-04-08
feat(http): add HTTP_METHODS, HttpMethods, and isHttpMethod

kitsonk:
This provides a couple of utilities to be able to safely handle common HTTP methods.

This comes from oak/oak_commons.

### DIFF
--- a/http/method.ts
+++ b/http/method.ts
@@ -3,27 +3,57 @@
 
 /**
  * Contains the constant {@linkcode HTTP_METHODS} and the type
- * {@linkcode HttpMethods} and the type guard {@linkcode isHttpMethod} for
+ * {@linkcode HttpMethod} and the type guard {@linkcode isHttpMethod} for
  * working with HTTP methods with type safety.
  *
  * @module
  */
 
-/** A constant array of common HTTP methods. */
+/** A constant array of common HTTP methods.
+ *
+ * This list is compatible with Node.js `http` module.
+ */
 export const HTTP_METHODS = [
-  "HEAD",
-  "OPTIONS",
+  "ACL",
+  "BIND",
+  "CHECKOUT",
+  "CONNECT",
+  "COPY",
+  "DELETE",
   "GET",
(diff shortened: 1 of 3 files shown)


iuioiua:
I very much like these HTTP helpers you've been adding. LGTM!

http/method.ts
```diff
@@ -0,0 +1,29 @@
...
+  "OPTIONS",
+  "GET",
+  "PUT",
+  "PATCH",
+  "POST",
+  "DELETE",
+] as const;
+
+/** A type representing string literals of each of the common HTTP method. */
+export type HttpMethods = typeof HTTP_METHODS[number];
```
kt3k, line 24: How about `HttpMethod` (singular)?

http/method.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/** A constant array of common HTTP methods. */
+export const HTTP_METHODS = [
+  "HEAD",
+  "OPTIONS",
+  "GET",
+  "PUT",
+  "PATCH",
+  "POST",
+  "DELETE",
```
kt3k, line 20: I wonder if we should also list `CONNECT` and `TRACE` which are mentioned in RFC 9110 https://www.rfc-editor.org/rfc/rfc9110.html#section-9.1

(RFC 9110 also mentions [HTTP Method Registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml), which list 39 methods)

kt3k: (Note: Node.js http.METHODS array has 34 items. 28 items overlap with the IANA list, 6 items don't.)

http/method.ts
kitsonk: agreed.

http/method.ts
kitsonk: It makes sense to align to Node.js `http` methods.

kt3k (approved):
LGTM

Issue #3223, opened 2023-02-28
encoding/ascii85.ts: `encode()`  returns a wrong result with a 4-byte subarray


dahlia:
#### Describe the bug

The `encode()` function from *encoding/ascii.ts* returns a wrong result with a subarray.  More specifically, if a subarray is a 4-byte (or multiples of 4 bytes) slice of a buffer longer than it, and the slice refers to a non-zero offset in the buffer, the function encodes the first 4 bytes of the buffer instead of the subarray bytes.

#### Steps to Reproduce

```typescript
import { encode } from "https://deno.land/std@0.178.0/encoding/ascii85.ts";
import { assertEquals } from "https://deno.land/std@0.178.0/testing/asserts.ts";

Deno.test("base85 encoding", () => {
  const data = new Uint8Array([0x73, 0x70, 0x61, 0x6d]);
  const encoded = encode(data);
  assertEquals(encoded, "F)YQ)");
});

Deno.test("base85 encoding with a 4-byte subarray", () => {
  const data = new Uint8Array(
    [0x01, 0x02, 0x03, 0x04, 0x73, 0x70, 0x61, 0x6d],
  );
  const encoded = encode(data.subarray(4));

  // The following assertion fails; it returns "!<N?+" instead, which encodes
  // the first 4 bytes of the array: [0x01, 0x02, 0x03, 0x04].
  assertEquals(encoded, "F)YQ)");
});
```

#### Expected behavior

The below test should pass, but it does not.

#### Environment

- OS: Ubuntu 22.04 & Windows 11 Pro
- Deno version: 1.30.3
- std version: 0.178.0

kamilogorek:
Probably the same issue that I've fixed here - https://github.com/denoland/deno_std/pull/3208
ref: https://github.com/denoland/deno_std/blob/69fa1e956c2469f168756d757b1653323327ed53/encoding/ascii85.ts#L78-L79

PR #3310, opened 2023-04-09
fix(encoding/ascii85): fix `encode()` returns a wrong result with a 4-byte subarray

babiabeo:
Fixes: #3223

### DIFF
--- a/encoding/ascii85.ts
+++ b/encoding/ascii85.ts
@@ -65,6 +65,9 @@ const Z85 =
  * @param [options.delimiter] whether to use a delimiter, if supported by encoding standard
  */
 export function encode(uint8: Uint8Array, options?: Ascii85Options): string {
+  // Create a new Uint8Array from the given uint8 instead of using it
+  uint8 = new Uint8Array(uint8);
+
   const standard = options?.standard ?? "Adobe";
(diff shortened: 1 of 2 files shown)


encoding/ascii85.ts
```diff
@@ -65,6 +65,8 @@ const Z85 =
  * @param [options.delimiter] whether to use a delimiter, if supported by encoding standard
  */
 export function encode(uint8: Uint8Array, options?: Ascii85Options): string {
+  // Create a new Uint8Array from the given uint8 instead of using it
+  uint8 = new Uint8Array(uint8);
```
kt3k, line 69: Can you try to fix the line `const view = new DataView(uint8.buffer);` instead of creating new uint8 array here. This causes unnecessary copy of underlying ArrayBuffer, and makes the function slower.

I think that line should be something like `new DataView(uint8.buffer, uint8.byteOffset, uint8.byteLength)`.

kt3k:
The test case looks good to me.

encoding/ascii85.ts
babiabeo: Thanks for your suggestion!

kt3k (approved):
LGTM. Thanks!

PR #3314, opened 2023-04-13
docs(http): fix default etag algorithm

kt3k:
Since #3245, the default etag hash algorithm is `SHA-256` instead of `fnv1a` (because `http/etag` only supports digest algorithms available for `crypto.subtle.digest`).

This PR fixes the doc for it. Also removes the reference to `DigestAlgorithm` type from `std/crypto`.

Because of the removal of the dependency to `std/crypto` this PR reduces the file_server size from 646.44KB to 389.77KB.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -428,7 +428,7 @@ export interface ServeDirOptions {
    *
    * @default {"SHA-256"}
    */
-  etagAlgorithm?: DigestAlgorithm;
+  etagAlgorithm?: AlgorithmIdentifier;
   /** Headers to add to each response
    *
    * @default {[]}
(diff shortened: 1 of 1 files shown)


http/file_server.ts
```diff
@@ -427,7 +426,7 @@ export interface ServeDirOptions {
   quiet?: boolean;
   /** The algorithm to use for generating the ETag.
    *
-   * @default {"fnv1a"}
+   * @default {"SHA-256"}
    */
   etagAlgorithm?: DigestAlgorithm;
```
iuioiua, line 431:
```suggestion
  etagAlgorithm?: AlgorithmIdentifier;
```

kt3k: Thanks

bartlomieju (approved):
LGTM, I like the size decrease!

PR #3316, opened 2023-04-16
fix(http): Move deno deploy specific logic from `etag.ts` to `file_server.ts`

ayame113:
Move logic related to `DENO_DEPLOYMENT_ID` in `etag.ts` to `file_server.ts`.

This logic was originally in `file_server.ts` but was moved to etag.ts at #3245. But since this logic is mainly for file servers, wouldn't it be better to put it in `file_server.ts` rather than `etag.ts`?

### DIFF
--- a/http/etag.ts
+++ b/http/etag.ts
@@ -26,20 +26,6 @@ type Entity = string | Uint8Array | FileInfo;
 const encoder = new TextEncoder();
 
 const DEFAULT_ALGORITHM: AlgorithmIdentifier = "SHA-256";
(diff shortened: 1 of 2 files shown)


ayame113:
For reference, a related Discord discussion: https://discord.com/channels/684898665143206084/775393009981849600/1095848426102591558

kt3k (approved):
I agree with this. LGTM

Falling back to the hash value based on `DENO_DEPLOYMENT_ID` feels specific to the application (in this case `file_server`) and not universal in etag calculation.

PR #3319, opened 2023-04-17
fix(http/file_server.ts): Respond to Range requests with correct byte length

### LINKED ISSUE #3198: Erroneous HTTP/2 responses to range requests with std/http's `serveFile`
As title suggests; here is a playground: https://dash.deno.com/playground/weak-duck-73. Ordinary requests correctly receive a 200, but a range request results in the following:
```plain
$ curl -v -H "Range: bytes=0-10" https://weak-duck-73.deno.dev
(—— snip ——)
> GET / HTTP/2
> Host: weak-duck-73.deno.dev
> user-agent: curl/7.77.0
> accept: */*
> range: bytes=0-10
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* Connection state changed (MAX_CONCURRENT_STREAMS == 4294967295)!
* HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
* stopped the pause stream!
* Connection #0 to host weak-duck-73.deno.dev left intact
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

I am not familiar with the details of HTTP/2 or TLS, but I suspect this is an issue with the interaction between both the Deploy platform and the library implementation because:
- I do not reproduce the issue on localhost (HTTP/1.1 connections) and my own remote setup (Nginx proxying incoming HTTP/2 TLSv1.2 requests to its localhost by HTTP/1.1);
- A modified `serveFile` implementation clears the problem, see below.

Details: the `serveFile` function pipes a file using its `FsFile.readable` stream (https://deno.land/std@0.168.0/http/file_server.ts?source#L192); this stream is not truncated and relies on the lower level to stop piping at the right position by inspecting `Content-Length`/`Content-Range` headers. I modified the function to pipe the file's stream through a custom `TransformStream` that truncates it, after which the 206 response is correctly constructed.

Thanks in advance for looking into this, and wish you a merry Christmas! 🎄

ayame113:
fix #3198

Note: As commented by https://github.com/denoland/deno_std/issues/3198#issuecomment-1510884991, the problem is already fixed for most users using file_server.ts. However, users who wanted to use serveFile() by itself were still getting the wrong response. This PR fixes it.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -10,6 +10,7 @@ import { contentType } from "../media_types/content_type.ts";
 import { serve, serveTls } from "./server.ts";
 import { calculate, ifNoneMatch } from "./etag.ts";
 import { Status } from "./http_status.ts";
+import { ByteSliceStream } from "../streams/byte_slice_stream.ts";
 import { parse } from "../flags/mod.ts";
 import { assert } from "../_util/asserts.ts";
 import { red } from "../fmt/colors.ts";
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks!

PR #3323, opened 2023-04-19
perf(http/file_server): Avoid calculating Content-Type when 304 Not Modified response

### LINKED ISSUE #3318: perf(http/file_server): `Content-Type` header is not required when returning a 304 Not Modified response
304 Not Modified responses currently include a `Content-Type` header.

image

But according to the spec this doesn't seem to be necessary.
(ref: https://stackoverflow.com/questions/31099113/why-response-304-doesnt-have-content-type-header)

It should be slightly faster by not calculating the Content-Type header for 304 responses.
I would like to submit a PR later.

ayame113:
close #3318 

If the file server returns a 304 Not Modified, calls `Deno.open()` and `contentType()` are unnecessary. Change those functions to be called on a later line.

This change seems to make the 304 response generation slightly (100µs) faster. (446µs -> 340µs)

before:
```
benchmark                      time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------- -----------------------------
fileServer_200OK           551.74 µs/iter(388.4 µs … 1.58 ms) 621.5 µs 1.01 ms 1.24 ms
fileServer_304NotModified  446.54 µs/iter(350.7 µs … 1.23 ms) 486.3 µs 883.2 µs 996.6 µs
```

after:
```
benchmark                      time (avg)             (min … max)       p75       p99      p995
----------------------------------------------------------------- -----------------------------
fileServer_200OK           536.2 µs/iter(364.9 µs … 51.06 ms) 542.5 µs 1.02 ms 1.2 ms
fileServer_304NotModified  340.07 µs/iter(257.8 µs … 2.58 ms) 365 µs 653.9 µs 695.1 µs
```

<details>
<summary>
benchmark code:
</summary>

```ts
import { serveFile } from "./file_server.ts";

{
  const req = new Request("http://localhost/");
  Deno.bench(async function fileServer_200OK() {
    await serveFile(req, "./README.md");
  });
}
{
  const etag =
    (await serveFile(new Request("http://localhost/"), "./README.md"))
      .headers.get("etag")!;

  const req = new Request("http://localhost/", {
    headers: { "if-none-match": etag },
  });
  Deno.bench(async function fileServer_304NotModified() {
    await serveFile(req, "./README.md");
  });
}
```

</details>

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -111,20 +111,11 @@ export async function serveFile(
     return createCommonResponse(Status.NotFound);
   }
 
-  const file = await Deno.open(filePath);
-
   const headers = setBaseHeaders();
 
-  // Set mime-type using the file extension in filePath
-  const contentTypeValue = contentType(extname(filePath));
-  if (contentTypeValue) {
-    headers.set("content-type", contentTypeValue);
-  }
-
   // Set date header if access timestamp is available
-  if (fileInfo.atime instanceof Date) {
-    const date = new Date(fileInfo.atime);
-    headers.set("date", date.toUTCString());
+  if (fileInfo.atime) {
+    headers.set("date", fileInfo.atime.toUTCString());
   }
 
   const etag = fileInfo.mtime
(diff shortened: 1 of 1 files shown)


aapoalas (approved):
Nice improvement! Looks good to me.

kt3k (approved):
LGTM!

PR #3327, opened 2023-04-22
feat(jsonc): annotate return types

jsejcksn:
Previously, return type annotations were missing and the type `JsonValue` was duplicated.


kt3k (approved):
LGTM. Thanks!

Issue #3194, opened 2023-02-11
Path module has both `sep` and `SEP` exports, which are the same


dsherret:
```
> import * as path from "https://deno.land/std@0.177.0/path/mod.ts";
undefined
> path.sep
"\\"
> path.SEP
"\\"
```

Is this intentional? Maybe they should be consolidated?

kalisjoshua:
Looking at this, the "fix" seems to be pretty straightforward but the concern might be that removing functionality after so long might be problematic. Using `git blame` it appears that this has been in existence for 2 years at this point.

I've never contributed before so I am completely new to the protocol for deprecating "features".

kt3k:
We usually first put `@deprecate` jsdoc tag to the APIs to deprecate, and then remove them like 5 minor versions later. (Examples: https://github.com/denoland/deno_std/blob/934cb11a7e41e262dcd12be88e5a757e985359e7/http/util.ts#L9 https://github.com/denoland/deno_std/blob/934cb11a7e41e262dcd12be88e5a757e985359e7/csv/stream.ts#L5

In `std/path` there's also `SEP_PATTERN` exported. So I think we should keep `SEP` and deprecate `sep`. What do you think?

iuioiua:
> We usually first put `@deprecate` jsdoc tag to the APIs to deprecate, and then remove them like 5 minor versions later.

Nit: @kt3k, most deprecations have been removed 3 minor versions later.

> So I think we should keep `SEP` and deprecate `sep`. What do you think?

I agree.

timreichen:
Is uppercase instead of camelcase for constants a common design pattern in std? If so, why?
It says in the [style guide naming convention](https://deno.com/manual@v1.32.5/references/contributing/style_guide#naming-convention)  to always use camel or pascal case and specifically to avoid uppercase (for acronyms). Isn't that a breach of that convention?

kt3k:
@timreichen Sounds a good point. I created an issue for it #3331

@kalisjoshua Node compatibility can be worth considering, but I don't think it's absolutely necessary as Deno now provides `node:path` as builtin module.

kalisjoshua:
Good point @kt3k.

I created a PR for review and continued discussion on specifics. I'm happy to help in any way that I can. I'm just trying to get my fingers into the code and make improvements where I can.

PR #3332, opened 2023-04-25
fix(std/path) duplicate export sep/SEP and naming conventions

kalisjoshua:
issue: #3194

**Summary**

  1. Created new variable names to align with the - current, published - naming convention
  2. Kept the non-standard - to be deprecated - variables as alias to the standard versions
  3. Added `@deprecated` tags to each variable that will be removed in a future version
  4. Moved the `separators.ts` file to `_separators.ts` to align with the naming convention for [files that should not be linked to directly](https://deno.com/manual@v1.32.5/references/contributing/style_guide#if-a-filename-starts-with-an-underscore-_foots-do-not-link-to-it)
  5. Added comment to `_separators.ts` to detail why the file is necessary

I am completely open to suggestion/debate on anything proposed herein.

### DIFF
--- a/path/_separator.ts
+++ b/path/_separator.ts
@@ -1,0 +1,12 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+// NOTE: This file exists to break circular referencecs; it should NOT be
+// imported directly, outside of the "std/path" module.
+
+import { isWindows } from "../_util/os.ts";
+import { sep as win32Sep } from "./win32.ts";
+import { sep as posixSep } from "./posix.ts";
+
+export const SEP = isWindows ? win32Sep : posixSep;
+export const SEP_PATTERN = isWindows ? /[\\/]+/ : /\/+/;
--- a/path/common.ts
+++ b/path/common.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { SEP } from "./separator.ts";
+import { SEP } from "./_separator.ts";
 
 /** Determines the common path from a set of paths, using an optional separator,
  * which defaults to the OS default separator.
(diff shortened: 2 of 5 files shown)


kalisjoshua:
I just noticed there are a lot of formatting changes that aren't intentional and I am working on removing them from this PR.

path/_separator.ts
```diff
@@ -0,0 +1,16 @@
...
+import { isWindows } from "../_util/os.ts";
+import { sep as win32Sep } from "./win32.ts";
+import { sep as posixSep } from "./posix.ts";
+
+export const sep = isWindows ? win32Sep : posixSep;
+// @deprecated (will be removed after 0.187.0) use `sep`
+export const SEP = sep;
+export const sepPattern = isWindows ? /[\\/]+/ : /\/+/;
+// @deprecated (will be removed after 0.187.0) use `sepPattern`
+export const SEP_PATTERN = sepPattern;
```
nayeemrmn, line 16: The outcome of the discussions was to keep using upper snake case for constants: https://github.com/denoland/manual/pull/625

kalisjoshua: @nayeemrmn  good to know thank you. I will take a look and see if anything is necessary for this PR to continue or not. I will close it if I don't think anything is needed to change.

kalisjoshua:
Reviewing the changes I made the only thing that remains as a possible change and likely very low priority is bullet 4 (above) changing the filename of `separator.ts` to `_separator.ts`. I fully understand if that is not significant enough of a change to incorporate.

kt3k:
Keeping `SEP` looks good to me, but `sep` seems still exported from `mod.ts`.

> Moved the separators.ts file to _separators.ts

I think this change is not desired. We usually prefer to allow importing items from its own file (to allow optimizing the dependency graph). (See https://github.com/denoland/deno_std/issues/2936 for reference). We use `_foo.ts` pattern when the items in it are only private APIs.

kalisjoshua:
Closing because there is nothing more to change for this issue or branch.

PR #3333, opened 2023-04-26
fix(http/server): flaky 'address in use' error

### LINKED ISSUE #3320: http/file_server: Occasional wrong error when port is already in use
**Describe the bug**

When the port is already in use, `file_server` occasionally fails with the wrong error.

**Steps to Reproduce**

1. Run `file_server .` in a folder
2. Run `file_server .` in a separate folder, multiple times until the error reproduces (usually <5 is enough, if it doesn't happen wait a minute and try again)

**Expected behavior**

Since both the servers are trying to listen on port 4507, the second one correctly fails.

Usually it fails with this error:
```
➜ file_server .
error: Uncaught (in promise) AddrInUse: Address already in use (os error 48)
    const listener = Deno.listen({
                          ^
    at Object.listen (ext:deno_net/01_net.js:414:39)
    at Server.listenAndServe (https://deno.land/std@0.182.0/http/server.ts:182:27)
    at serve (https://deno.land/std@0.182.0/http/server.ts:604:20)
    at main (https://deno.land/std@0.182.0/http/file_server.ts:663:5)
    at https://deno.land/std@0.182.0/http/file_server.ts:697:3
```

However, sometimes it fails with this other unhelpful error:
```
➜ file_server .
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'port')
  port = (server.addrs[0] as Deno.NetAddr).port;
                                           ^
    at serve (https://deno.land/std@0.182.0/http/server.ts:606:44)
    at main (https://deno.land/std@0.182.0/http/file_server.ts:663:5)
    at https://deno.land/std@0.182.0/http/file_server.ts:697:3
```

**Environment**

- OS: macOS 13.3
- deno version: 1.32.4
- std version: 0.182.0

tetraphobia:
Resolves #3320 

When server.addrs[0].port is assigned to a variable, it sometimes fails because the call to `server.listenAndServe` has not finished yet, resulting in a rather unhelpful error being thrown.

Awaiting the call to `server.listenAndServe` fixes the bug.

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -601,7 +601,13 @@ export async function serve(
     once: true,
   });
 
-  const s = await server.listenAndServe();
+  const listener = Deno.listen({
+    port,
+    hostname,
+    transport: "tcp",
+  });
+
+  const s = server.serve(listener);
 
   port = (server.addrs[0] as Deno.NetAddr).port;
 
@@ -611,7 +617,7 @@ export async function serve(
     console.log(`Listening on http://${hostnameForDisplay(hostname)}:${port}/`);
   }
 
-  return s;
+  return await s;
 }
 
 export interface ServeTlsInit extends ServeInit {
--- a/http/server.ts
+++ b/http/server.ts
@@ -601,7 +601,7 @@ export async function serve(
     once: true,
   });
 
-  const s = server.listenAndServe();
+  const s = await server.listenAndServe();
 
   port = (server.addrs[0] as Deno.NetAddr).port;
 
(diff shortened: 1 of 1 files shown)


tetraphobia:
Turns out, awaiting the call to `server.listenAndServe` causes file_server to hang, so I went with the solution proposed by @lucacasonato.

kt3k (approved):
Confirmed this fixes the issue. LGTM

Note: When I tested the original issue, I was able to reproduce it only when I invoked file server with file url or http(s) url. It didn't reproduce if I invoked it as relative path (like `deno run -A ./http/file_server.ts`).

Issue #3330, opened 2023-04-25
Regex escaping


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

https://github.com/tc39/proposal-regex-escaping/ has been in TC39 purgatory for almost 8 years now and it's still at stage 1, despite what seems to be fairly wide consensus that it's a useful feature that should be standardized. In light of that, it seems like a good candidate for inclusion in a standard library, for later deprecation if and when the proposal ever gets rolled out.

**Describe the solution you'd like**

A function called something like `regexEscape` with an API something like this:

```js
export declare function regexEscape(str: string): string
```

Ideally, it'd support contexts both within and outside of character classes, as well as all flavors of regex, including the upcoming [`v` flag](https://github.com/tc39/proposal-regexp-v-flag). It'd escape the minimal set of characters needed to be agnostic to all those contexts (i.e. no escaping of non-ASCII characters or always-literal characters like `_`), and for other characters, it would use the most readable escape where possible (e.g. `\.` for a literal period), falling back to hex escapes where that's impossible (e.g. [`\x3c` for `<`](https://github.com/tc39/proposal-regex-escaping/issues/55)).

**Describe alternatives you've considered**

Wait another 8 years I guess? 😅

Alternative APIs would involve versions that take options (e.g. specifying flags or whether the context is within a character class or not) or exporting different versions of the function for these use cases. Both of these options would make the implementation significantly more complex.

lino-levan:
I'm in favor of this proposal. @lionel-rowe are you willing to open a PR?

lionel-rowe:
> @lionel-rowe are you willing to open a PR?

Happy to at some point, not sure when I'll get around to it though.

PR #3334, opened 2023-04-27
feat(regexp): add escape function

lionel-rowe:
Fixes https://github.com/denoland/deno_std/issues/3330

Escaping is fairly conservative to account for context sensitivity and forward compatibility with the `v` flag — the only non-alphanumeric printable ASCIIS that aren't escaped are ` "'_`, and of the characters that are escaped, almost half use the most conservative `\x` hex notation. That could be reduced a little (but not much) by ignoring characters in `ClassSetReservedDoublePunctuator`, which only need escaping if repeated adjacent to each other in a within a character class in a `v`-flagged regex.


kt3k:
Thanks for the suggestion!

I think the directory name should be `regexp/` instead of `regex/` as JS has `RegExp` class, and also the API name should be `regexpEscape` (or `regExpEscape`?) instead of `regexEscape`. What do you think?

lionel-rowe:
> I think the directory name should be `regexp/` instead of `regex/` as JS has `RegExp` class, and also the API name should be `regexpEscape` (or `regExpEscape`?) instead of `regexEscape`. What do you think?

This is probably one of those "gif/jif" things that everyone has their own strong opinions about, but it seems like `regex` is the more common usage within the community at large by a significant margin (a Google search for ["javascript regex"](https://www.google.com/search?q=%22javascript+regex%22) gives 597k results, whereas ["javascript regexp"](https://www.google.com/search?q=%22javascript+regexp%22) gives 175k; meanwhile on GitHub, [regex](https://github.com/search?l=JavaScript&q=regex&type=Repositories) gives 6,970 JS repository results, whereas [regexp](https://github.com/search?l=JavaScript&q=regexp&type=Repositories) gives 2,063).

Still, this function would typically be used when interpolating/concatenating content into strings passed to the `RegExp` constructor, so that's something in favor of `regexp`. Also, grepping for `regex` gives results for both variants, whereas grepping for `regexp` only gives results for `regexp`.

I started writing this answer arguing in favor of `regex`, but I think I've convinced myself that `regexp` is the better option 😅 happy to change unless anyone has a more convincing argument for `regex`.

iuioiua:
> Still, this function would typically be used when interpolating/concatenating content into strings passed to the `RegExp` constructor, so that's something in favor of `regexp`. Also, grepping for `regex` gives results for both variants, whereas grepping for `regexp` only gives results for `regexp`.

+1 for `regexp`

kt3k:
@lionel-rowe Thanks for updating to `std/regexp`!

I now wonder if we could make API name just `escape` instead `regExpEscape`. I think that's more aligned to the rest of deno_std. (For example, `parse` in `std/yaml` not `yamlParse`, `format` in `fmt/bytes` instead of `formatBytes` (or `bytesFormat`), etc

lionel-rowe:
> @lionel-rowe Thanks for updating to `std/regexp`!
> 
> I now wonder if we could make API name just `escape` instead `regExpEscape`. I think that's more aligned to the rest of deno_std. (For example, `parse` in `std/yaml` not `yamlParse`, `format` in `fmt/bytes` instead of `formatBytes` (or `bytesFormat`), etc

Done and done!

kt3k (approved):
LGTM 👍

Issue #3329, opened 2023-04-25
Encode/decode HTML entities


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

Encoding/decoding HTML entities is one of those tasks that's so ubiquitous in JS that it's kind of unbelievable it's never been added to the web platform or as a browser API. There must be tens of thousands of userland implementations of functions with names like `encodeHtmlEntities`, `escapeHtml`, `sanitizeHtml`, etc., with various subtly different semantics and APIs ([searching GitHub](https://github.com/search?l=JavaScript&q=escapeHtml&type=Code) gives around 1.7 million JS results for `escapeHtml`, though there's no easy way to tell how many are independent implementations). Many implementations use the DOM itself to do the encoding/decoding, which is a great idea but only works in the browser or with a weighty dependency such as jsdom.

**Describe the solution you'd like**

Suggested API something like this:

```ts
export type EntityList = Record<string, { codepoints: number[], characters: string }>
type EncodeHtmlEntitiesOptions = Partial<{ /* any relevant options*/ }>
type DecodeHtmlEntitiesOptions = Partial<{ entityList: EntityList, /* any other relevant options */ }>

export declare function encodeHtmlEntities(str: string, options?: EncodeHtmlEntitiesOptions): string
export declare function decodeHtmlEntities(str: string, options?: DecodeHtmlEntitiesOptions): string
```

With no options passed, `encodeHtmlEntities` maps `&<>'"` to `&amp; &lt; &gt; &#39; &quot;`, and `decodeHtmlEntities` maps in the opposite direction, plus handling `&apos;` and all decimal/hex entities.

By passing [the list of named HTML entities](https://html.spec.whatwg.org/entities.json) as `options.entityList` of `decodeHtmlEntities`, all named entities are also decoded:

```js
import { default as entityList } from 'https://html.spec.whatwg.org/entities.json' assert { type: 'json' }

decodeHtmlEntities('&thorn;&eth;', { entityList }) // 'þð'
```

This isn't done by default, as that JSON file is around 145KB (100KB minified), and it's probably not necessary to handle those in most cases, given that modern implementations typically won't encode them.

**Describe alternatives you've considered**

Might also be worth having equivalent functions for XML entities, but dealing with XML is more complicated as it can have custom entities, plus writers/consumers of XML are more likely to already be using existing XML libraries that handle that functionality.

PR #3335, opened 2023-04-27
feat(html): Add escape and unescape functions for HTML entities

lionel-rowe:
Fixes https://github.com/denoland/deno_std/issues/3329


lino-levan:
Maybe we should rename `html_entities` to just `entities` since it is in the html top level folder?

iuioiua:
> Maybe we should rename `html_entities` to just `entities` since it is in the html top level folder?

+1

_tools/check_doc_imports.ts
```diff
@@ -26,6 +26,11 @@ const RX_CODE_BLOCK = /`{3}([\w]*)\n([\S\s]+?)\n`{3}/gm;
 let shouldFail = false;
 let countChecked = 0;
 
+const allowedUrlPrefixes = [
+  "https://deno.land/std@$STD_VERSION/",
+  "https://html.spec.whatwg.org/",
```
lionel-rowe, line 31: Is it OK to add a special case for this? Rationale — it's a JSON version of [a spec page provided by WHATWG](https://html.spec.whatwg.org/multipage/named-characters.html), which is the authoritative reference for named HTML entities and which states:

> _This list is static and [will not be expanded or changed in the future](https://github.com/whatwg/html/blob/main/FAQ.md#html-should-add-more-named-character-references)._

The alternative to importing directly from the WHATWG-provided JSON would be duplicating that JSON file within `std`.

lionel-rowe:
Bikeshedding: encode/decode vs. escape/unescape vs [any other options]? I don't have a strong preference.

_tools/check_doc_imports.ts
kt3k: I'd prefer to duplicate the file in std with some note about the original. The spec doesn't seem guaranteeing the url and the shape of that JSON file.

timreichen:
+1 for escape and unescape

_tools/check_doc_imports.ts
lionel-rowe: Fixed — one other advantage of importing internally is it doesn't have to use the same structure, which is more verbose than it needs to be for our purposes.

kt3k (approved):
LGTM

kt3k:
Memo:

- php - [htmlspecialchars](https://www.php.net/manual/en/function.htmlspecialchars.php), [htmlspecialchars_decode](https://www.php.net/manual/en/function.htmlspecialchars-decode.php) (global builtins)
- ruby - [CGI.escapeHTML](https://docs.ruby-lang.org/en/3.2/CGI/Util.html#method-i-escapeHTML), [CGI.unescapeHTML](https://docs.ruby-lang.org/en/3.2/CGI/Util.html#method-i-unescape_html)
- golang - [EscapeString](https://pkg.go.dev/html#EscapeString), [UnescapeString](https://pkg.go.dev/html#UnescapeString) (html standard library)
- python - [html.escape](https://docs.python.org/3/library/html.html#html.escape), [html.unescape](https://docs.python.org/3/library/html.html#html.unescape) (html standard library)

kt3k:
@lionel-rowe Thanks for your contribution!

PR #3341, opened 2023-04-29
perf(collections/includes_value): support objects with non-enumerable properties

babiabeo:
Since [`for...in` loop can only visit enumerable properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Enumerability_and_ownership_of_properties#traversing_object_properties), this PR updates `includesValue` function to work with non-enumerable properties.

### DIFF
--- a/collections/includes_value.ts
+++ b/collections/includes_value.ts
@@ -23,7 +23,9 @@ export function includesValue<T>(
   record: Readonly<Record<string, T>>,
   value: T,
 ): boolean {
-  for (const i in record) {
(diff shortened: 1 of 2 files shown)


levex:
Thanks for the PR! Are these results inverted? 1 microseconds is 1000 nanoseconds.

babiabeo:
> Thanks for the PR! Are these results inverted? 1 microseconds is 1000 nanoseconds.

Oops! Sorry, my bad

babiabeo:
Closed due to mistake

PR #3342, opened 2023-04-30
BREAKING(path): deprecate path.sep

### LINKED ISSUE #3194: Path module has both `sep` and `SEP` exports, which are the same

kt3k:
closes #3194

### DIFF
--- a/path/mod.ts
+++ b/path/mod.ts
@@ -43,10 +43,12 @@ export const {
   parse,
   relative,
   resolve,
-  sep,
   toFileUrl,
   toNamespacedPath,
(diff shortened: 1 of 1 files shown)


iuioiua (approved):

bartlomieju (approved):

dsherret (approved):
LGTM

PR #3344, opened 2023-04-30
debug xeval_test

kt3k:


kt3k:
The below `new AsyncFunction` call seems swallowing things...

```
const AsyncFunction = Object.getPrototypeOf(async function () {}) 
  .constructor;
new AsyncFunction("$", "(")
```

Issue #3343, opened 2023-04-30
Enable `xevalCliSyntaxError` case in `examples/xeval_test.ts`


kt3k:
This test started failing with Deno canary 683dbd7f3bdd91c33005b816ad26b82a4343931a.
The last passing version was Deno canary 6cd62ea5e969de258b1d308daf5bec91e73e79d3.

The failure doesn't reproduce with my local mac (M1)

kt3k:
Did some debugging in https://github.com/denoland/deno_std/pull/3344

Something wrong seems happening around:
https://github.com/denoland/deno_std/blob/aa57bf337bbf5e061752b92164b003a60e728951/examples/xeval.ts#L93

PR #3345, opened 2023-04-30
chore: update assertions following Deno.inspect change

kt3k:
This PR updates the assertion of various outputs which was affected by `Deno.inspect` change in https://github.com/denoland/deno/pull/18691

### DIFF
--- a/examples/xeval_test.ts
+++ b/examples/xeval_test.ts
@@ -62,7 +62,8 @@ Deno.test({
   },
 });
 
-Deno.test("xevalCliSyntaxError", async function () {
+// TODO(kt3k): Enable this test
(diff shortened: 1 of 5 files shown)


examples/xeval_test.ts
```diff
@@ -62,7 +62,8 @@ Deno.test({
   },
 });
 
-Deno.test("xevalCliSyntaxError", async function () {
+// TODO(kt3k): Enable this test
+Deno.test("xevalCliSyntaxError", { ignore: true }, async function () {
```
kt3k, line 66: This is also disabled temporarily (because it's failing with canary). I created issue for it https://github.com/denoland/deno_std/issues/3343

crowlKats (approved):

PR #3347, opened 2023-04-30
feat(async): support `signal` on `deadline()` like `delay()`

lambdalisue:
As the subject suggests, I have added support for the signal parameter in the `deadline()` function. This change is useful to clean up dangling promises for tests like:

```ts
Deno.test("...", async () => {
  const { signal } = controller;
  assertEquals(await Promise.race([
    deadline(new Promise(() => {}), 100, { signal }),
    Promise.resolve("Hello"),
  ], "Hello");
  controller.abort();
});
```

Without the signal parameter, the above test fails with the following error:

```
error: Leaking async ops:
  - 1 async operation to sleep for a duration was started in this test, but never completed. This is often caus
ed by not cancelling a `setTimeout` or `setInterval` call.
```

Note that the comment on `deadline` suggests using `AbortSignal.timeout(...)` instead. However, the signal itself cannot be manually aborted ([yet](https://github.com/w3ctag/design-reviews/issues/737)), so `deadline()` is still useful in such cases.

### DIFF
--- a/async/deadline.ts
+++ b/async/deadline.ts
@@ -1,12 +1,17 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { deferred } from "./deferred.ts";
+import { delay } from "./delay.ts";
+
+export interface DeadlineOptions {
+  /** Signal used to abort the deadline. */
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks!

PR #3349, opened 2023-05-01
docs(http): add missing default values for `ServeDir` options of `file_server`

vwkd:
Documents the default values for the `fsRoot`, `urlRoot` and `showIndex` options of `file_server`'s `serveDir` function which are currently missing in the [documentation](https://deno.land/std@0.185.0/http/file_server.ts?s=ServeDirOptions).


iuioiua (approved):

kt3k (approved):
LGTM. Thanks!

PR #3350, opened 2023-05-02
feat(async/pool): use browser compatible APIs

### LINKED ISSUE #3348: [async/pooledMap] res.readable[Symbol.asyncIterator] is not a function in most browsers
**Describe the bug**

When trying to use `async/pooledMap` in Chrome, I encountered that error. Testing with other browsers, only Firefox does not raise the error.

When the error occurs, the iteratorFn still runs, but we can't iterate the results.

Understandingly, this is on the browser vendors to implement `asyncIterator` for `ReadableStream`, but if there is another way to handle that in `async/pooledMap`, that would be great!

The relevant bugs are here:

Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=929585
Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1525852
WebKit: https://bugs.webkit.org/show_bug.cgi?id=194379

**Steps to Reproduce**

The easiest way is to use `async/pooledMap` in a Fresh island. It's hard to provide a standalone example. The example from `async/pooledMap` documentation is sufficient:

```ts
import { pooledMap } from "https://deno.land/std@0.185.0/async/pool.ts";

const results = pooledMap(
  2,
  [1, 2, 3],
  (i) => new Promise((r) => setTimeout(() => r(i), 1000)),
);

for await (const value of results) {
  // ...
}
```

**Expected behavior**

Should not see error, and the result can be iterated

**Environment**

- OS: Mac OS 13.2.1
- deno version: 1.32.4
- std version: 0.185
- Browsers: Chrome 112

jsejcksn:
Resolves #3348

### DIFF
--- a/async/pool.ts
+++ b/async/pool.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
 
 export const ERROR_WHILE_MAPPING_MESSAGE = "Threw while mapping.";
 
@@ -91,5 +92,17 @@ export function pooledMap<T, R>(
       )).catch(() => {});
     }
   })();
-  return res.readable[Symbol.asyncIterator]();
+  // Feature test until browser coverage is adequate
+  return Symbol.asyncIterator in res.readable &&
+      typeof res.readable[Symbol.asyncIterator] === "function"
+    ? (res.readable[Symbol.asyncIterator] as () => AsyncIterableIterator<R>)()
+    : (async function* () {
+      const reader = res.readable.getReader();
+      while (true) {
+        const { done, value } = await reader.read();
+        if (done) break;
+        yield value;
+      }
+      reader.releaseLock();
+    })();
 }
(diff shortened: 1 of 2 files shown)


jsejcksn:
Do we already have a utility for the `ReadableStream` to `AsyncIterableIterator` conversion? I didn't find one in a quick search.

async/pool.ts
```diff
@@ -91,5 +92,17 @@ export function pooledMap<T, R>(
       )).catch(() => {});
     }
   })();
-  return res.readable[Symbol.asyncIterator]();
+  // Feature test until browser coverage is adequate
+  return Symbol.asyncIterator in res.readable &&
+      typeof res.readable[Symbol.asyncIterator] === "function"
+    ? (res.readable[Symbol.asyncIterator] as () => AsyncIterableIterator<R>)()
```
kt3k, line 98: nit: Is this type assertion (`as () => AsyncIterableIterator<R>`) necessary?

kt3k: Looked unnecessary

kt3k: ah, it's necessary for browser type checking..

jsejcksn:
@kt3k Thanks for adding the test.

kt3k (approved):

PR #3352, opened 2023-05-02
feat(uuid): add pre-defined namespace UUIDs

babiabeo:
These namespaces are useful for generating uuid v3 or v5.

Ref: https://www.ietf.org/rfc/rfc4122.html#appendix-C

### DIFF
--- a/uuid/constants.ts
+++ b/uuid/constants.ts
@@ -1,0 +1,43 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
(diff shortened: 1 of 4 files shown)


iuioiua:
Great addition! However, it is best to move these constants into a separate file. It's not a big deal, but in its current state, if someone only wanted to import the constants, they'd import all of `std/uuid` too.

babiabeo:
@iuioiua Fair point! I will move them when I have time. Thanks for your suggestion!

babiabeo:
Added missing constants export.

Issue #3361, opened 2023-05-05
http/file_server: Improving file server implementations


ayame113:
The file server seems to have some points where the implementation could be refactored.

- HEAD request support
    [Golang's std server](https://github.com/golang/go/blob/fe10464358057778732e9c958683039beb64e61a/src/net/http/fs.go#L352) and [hyper-staticfile](https://github.com/stephank/hyper-staticfile/blob/8e917de3aa25f442e2c8e738093f5f8ab7bace75/src/util/file_response_builder.rs#L184) seem to support this.
    I think it's easy to implement just by conditionally branching with `req.method === "HEAD"`.

- In recent versions of Deno, the Date header is automatically set if it's not present. So I think adding the Date header inside `setBaseHeader()` is unnecessary.
    https://github.com/denoland/deno_std/blob/699ba98572f28e6dfb423abe959502ba6a96c551/http/file_server.ts#L328-L334

- Inside `serveDirIndex` there are serial await calls. It seems that it can be speeded up by calling this in parallel.
    ```ts
    for await (const entry of Deno.readDir(dirPath)) {
      ...
      const fileInfo = await Deno.stat(filePath);
      ...
    } // https://github.com/denoland/deno_std/blob/699ba98572f28e6dfb423abe959502ba6a96c551/http/file_server.ts#L282
    ````

- Currently it returns content with 200 OK for paths like `/path//////to////file.txt`. I think such requests should be redirected to a normalized path.
- It would be nice to use JSDoc's `@module` tag to enhance the documentation.

ayame113:
It looks like HEAD requests are already supported. My comment above was a mistake. This issue can be closed.

PR #3364, opened 2023-05-05
chore(http/file_server): remove unnecessary Date header generation

ayame113:
part of #3361

As far as I know, when this code was first introduced we had to manually set the Date header. The CLI now automatically sets the Date header, so we no longer need to manually set the Date header.

There is one more place to set the Date header ([link](https://github.com/denoland/deno_std/blob/699ba98572f28e6dfb423abe959502ba6a96c551/http/file_server.ts#L157)), but I left it that way because it seemed meaningful.

Also renamed `setBaseHeader` to `createBaseHeader` as I thought it would be more appropriate.

---

I apologize for submitting the PR all at once. Feel free to close these PRs if you don't think you need them.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -150,7 +150,7 @@ export async function serveFile(
     return createCommonResponse(Status.NotFound);
   }
 
-  const headers = setBaseHeaders();
+  const headers = createBaseHeaders();
 
   // Set date header if access timestamp is available
   if (fileInfo.atime) {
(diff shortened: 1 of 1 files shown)


iuioiua:
> The CLI now automatically sets the Date header, so we no longer need to manually set the Date header.

Is there anywhere that references this fact?

kt3k:
> Is there anywhere that references this fact?

The Date header should be set by [hyper](https://github.com/hyperium/hyper) as it's mandatory by HTTP spec.

https://www.rfc-editor.org/rfc/rfc9110.html#section-6.6.1-6

> An origin server with a clock (as defined in [Section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110.html#http.date)) MUST generate a Date header field in all [2xx (Successful)](https://www.rfc-editor.org/rfc/rfc9110.html#status.2xx), [3xx (Redirection)](https://www.rfc-editor.org/rfc/rfc9110.html#status.3xx), and [4xx (Client Error)](https://www.rfc-editor.org/rfc/rfc9110.html#status.4xx) responses

(I couldn't find exact code, but I found this issue https://github.com/hyperium/hyper/issues/2998

kt3k (approved):
LGTM

PR #3365, opened 2023-05-06
feat(collections): add `partitionEntries`

scarf005:
## Rationale

The collection functions for arrays and objects are mostly in parity, such as `filter{keys,values,entries}`. However, [partition][partition] is available only for arrays.

[partition]: https://github.com/denoland/deno_std/blob/9dd2253e591d4bf2f66a4d38d75a7e60b1cb7c74/collections/partition.ts

## Implementaion and testing
bulk of the [code][code] and [test][test] are modified from `filterEntries`, as the only difference was not discarding entries that does not match given predicate.

[code]: https://github.com/denoland/deno_std/blob/9dd2253e591d4bf2f66a4d38d75a7e60b1cb7c74/collections/filter_entries.ts
[test]: https://github.com/denoland/deno_std/blob/9dd2253e591d4bf2f66a4d38d75a7e60b1cb7c74/collections/filter_entries_test.ts

## Example usage

<details><summary>Example usages</summary>

```ts
const args = {
  "armor_bullet": 30,
  "id": "item_id1",
  "name": "item1",
  "armor_bash": 46,
  "material": "nanite",
  "armor_cut": 30,
  "weight": "460 g",
  "volume": "430 ml",
  "armor_acid": 80,
  "armor_foo": 123,
  "armor_bar": 456,
  "armor_baz": 789,
  "flags": ["foo", "bar", "baz"],
}
```
before:
```ts
import { partition } from "https://deno.land/std/collections/partition.ts"

const [armors, notArmors] = partition(Object.entries(args), ([key]) => key.startsWith("armor_"))
const armorskeys = armors.map(([key, value]) => [key.replace("armor_", ""), value])

const result = {
  armor: Object.fromEntries(armorskeys),
  ...Object.fromEntries(notArmors),
}
```
after:
```ts
import { partitionEntries } from "https://deno.land/std/collections/partition_entries.ts"

const [armors, notArmors] = partitionEntries(args, ([key]) => key.startsWith("armor_"))
const armorskeys = mapKeys(armors, (key) => key.replace("armor_", ""))
  
const result = {
  armor: armorskeys,
  ...notArmors,
}
```

</details>

### DIFF
--- a/collections/mod.ts
+++ b/collections/mod.ts
@@ -50,6 +50,7 @@ export * from "./map_keys.ts";
 export * from "./map_not_nullish.ts";
 export * from "./map_values.ts";
 export * from "./partition.ts";
+export * from "./partition_entries.ts";
 export * from "./permutations.ts";
 export * from "./find_single.ts";
 export * from "./sliding_windows.ts";
--- a/collections/partition_entries.ts
+++ b/collections/partition_entries.ts
@@ -1,0 +1,50 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+/**
+ * Returns a tuple of two records with the first one containing all entries of
+ * the given record that match the given predicate and the second one containing
+ * all that do not.
+ *
+ * @example
+ * ```ts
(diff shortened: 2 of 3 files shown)


kt3k (approved):
Looks like a reasonable addition to me. LGTM

PR #3366, opened 2023-05-06
refactor(http/file_server): use utils from std

kt3k:
This PR refactors `http/file_server` by using `format` util from `std/fmt/bytes` and `isRedirectStatus` from `http/http_status`. Also removes a few redundant `await` keywords.


bartlomieju (approved):
LGTM!

Issue #3086, opened 2023-01-02
suggestion(datetime): `difference()` changes


iuioiua:
1. Deprecate `DifferenceOptions`, as it provides needless customisation. Deprecating it would simplify the function by just returning the values for all units, which is computationally cheap.
2. Return `0` instead of `undefined` for units for relevant cases. It makes more sense to do so intuitively, IMO.

kt3k:
I don't recommend we should work on `datetime` difference further because Temporal is now actively developed in V8 and it seems having a good diff calculation support (zonedDateTime.until and zonedDateTime.since ref: https://tc39.es/proposal-temporal/docs/zoneddatetime.html#until )

PR #3367, opened 2023-05-06
fix(datetime): negative months, quarters, and years

denizdogan:
`difference` will currently return negative months, quarters, and years, when `to` < `from`, while all other units are always non-negative.

I know [we're not doing more work on `difference`](#3086), but this is just a simple bugfix, so please consider it.

```
    {
      days: 730,
      hours: 17520,
      milliseconds: 63072000000,
      minutes: 1051200,
-     months: -24,
-     quarters: -8,
+     months: 23,
+     quarters: 7,
      seconds: 63072000,
      weeks: 104,
-     years: -2,
+     years: 1,
    }
```

### DIFF
--- a/datetime/difference.ts
+++ b/datetime/difference.ts
@@ -21,9 +21,9 @@ export type DifferenceOptions = {
 };
 
 function calculateMonthsDifference(from: Date, to: Date): number {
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #3371, opened 2023-05-07
fix(http/file_server): dealing with dir listing view that contain system files

### LINKED ISSUE #3141: file_server different path open results unexpectedly so large
deno install -A --root "D:\scoop_user_apps\apps\deno" https://deno.land/std/http/file_server.ts

file_server .

If the current path is in E:\test\ws, the result of executing the above command is

```
Index of home /
Mode | Size | Name

(unknown mode)		chat-go/

```

If the current path is in E:\test, the result of executing the above command is
```
选择器

div-sub
```
If the current path is in `E:\`, the result of executing the above command is
```
Internal Server Error
```

ayame113:
close #3141

When I launch a file server in the Windows root directory and access http://localhost:4507/, I get a 500 Internal Server Error.
It seems that the following error occurs internally. These are Windows system files that reject `Deno.stat` calls.

```
The process cannot access the file because it is being used by another process. (os error 32): stat 'C:\/DumpStack.log.tmp'
The process cannot access the file because it is being used by another process. (os error 32): stat 'C:\/hiberfil.sys'
The process cannot access the file because it is being used by another process. (os error 32): stat 'C:\/pagefile.sys'
Access is Denied. (os error 5): stat 'C:\/PerfLogs'
Access is Denied. (os error 5): stat 'C:\/Recovery'
The process cannot access the file because it is being used by another process. (os error 32): stat 'C:\/swapfile.sys'
Access is Denied. (os error 5): stat 'C:\/System Volume Information'
```

This PR adds a try-catch around the `Deno.stat` call so that directory listing views are generated even for directories containing system files.
After this PR, the display of the root directory page looks like this:
image

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -257,6 +257,7 @@ async function serveDirIndex(
   options: {
     showDotfiles: boolean;
     target: string;
+    quiet: boolean | undefined;
   },
 ): Promise<Response> {
   const { showDotfiles } = options;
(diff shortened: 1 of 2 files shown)


http/file_server_test.ts
```diff
@@ -57,7 +58,7 @@ async function startFileServer({
     ],
     cwd: moduleDir,
     stdout: "piped",
-    stderr: "null",
+    stderr: "inherit",
   });
```
ayame113, line 62: Adding this line will make debugging easier as error messages will be displayed instead of discarded. It was a minor change so I've included it in this PR.

http/file_server.ts
```diff
@@ -645,12 +656,18 @@ async function createServeDirResponse(
...
-    return serveDirIndex(fsPath, { showDotfiles, target });
+    return serveDirIndex(fsPath, { showDotfiles, target, quiet });
   }
 
   return createCommonResponse(Status.NotFound);
 }
 
+function logError(error: unknown) {
+  console.error(
+    red(error instanceof Error ? error.message : "[non-error thrown]"),
```
kt3k, line 667: Not directly related to this change, but `"[non-error thrown]"` message doesn't look useful. How about just printing `error` itself?

ayame113: Makes sense. It has changed! (https://github.com/denoland/deno_std/pull/3371/commits/d8137cb34cc33182d830e1c8cfc94700832c700c)

kt3k (approved):
LGTM

PR #3374, opened 2023-05-11
chore(csv): remove internal dead code

ayame113:
I removed the `readRecord()` in `csv/_io.ts` as it is not used anywhere.

### DIFF
--- a/csv/csv_parse_stream.ts
+++ b/csv/csv_parse_stream.ts
@@ -41,8 +41,8 @@ class StreamLineReader implements LineReader {
     }
   }
 
-  isEOF(): Promise<boolean> {
-    return Promise.resolve(this.#done);
+  isEOF(): boolean {
+    return this.#done;
   }
 
   cancel() {
@@ -87,7 +87,7 @@ export class CsvParseStream<
 
   #headers: readonly string[] = [];
 
-  constructor(options: T = defaultReadOptions as T) {
+  constructor(options?: T) {
     this.#options = {
       ...defaultReadOptions,
       ...options,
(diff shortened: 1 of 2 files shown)


csv/csv_parse_stream.ts
```diff
@@ -41,8 +41,8 @@ class StreamLineReader implements LineReader {
     }
   }
 
-  isEOF(): Promise<boolean> {
-    return Promise.resolve(this.#done);
+  isEOF(): boolean {
+    return this.#done;
   }
```
ayame113, line 46: Since `isEOF` no longer needs to be async, I made it sync.

csv/csv_parse_stream.ts
```diff
@@ -87,7 +87,7 @@ export class CsvParseStream<
 
   #headers: readonly string[] = [];
 
-  constructor(options: T = defaultReadOptions as T) {
+  constructor(options?: T) {
     this.#options = {
       ...defaultReadOptions,
       ...options,
```
ayame113, line 93: defaultReadOptions is set on line 92, so there is no need to use it as a default value for the constructor here.

iuioiua (approved):

csv/csv_parse_stream.ts
```diff
@@ -41,8 +41,8 @@ class StreamLineReader implements LineReader {
     }
   }
 
-  isEOF(): Promise<boolean> {
-    return Promise.resolve(this.#done);
+  isEOF(): boolean {
+    return this.#done;
```
kt3k, line 45: Note: `StreamLineReader` (`LineReader` interface) is only used internally in `CsvParseStream`. So we can change this without breaking user facing APIs.

kt3k (approved):
LGTM

PR #3377, opened 2023-05-12
feat: handle spaces around comparators

mxdvl:
## What does this change?

Handle notations using spaces between comparators and versions.

## Why is this change required?

Accepts a wider ranges of version strings, which are often found to describe ranges in NPM packages. For example, [`@swc-node/register`’s peerDependencies](https://www.npmjs.com/package/@swc-node/register?activeTab=code).


kt3k:
Thanks for the suggestion, but we are now completely rewriting semver module in https://github.com/denoland/deno_std/pull/3169. Could you send a PR targeting that branch?

kt3k:
@mxdvl Thanks! I'll include https://github.com/denoland/deno_std/commit/7ab6dc3cc8371130cc4b7a899ed34700a0f252e7 to #3169

Issue #3378, opened 2023-05-13
async/retry.ts doesn't allow applying jitter to exponential backoff


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

It seems like one of the major benefits of exponential backoff as a retry strategy is that it can space out the retries with random jitter to prevent bottlenecks, but [async/retry.ts](https://github.com/denoland/deno_std/blob/main/async/retry.ts) currently doesn't implement this or provide it as an option.

[Stack Overflow — Why is random jitter applied to back-off strategies?](https://stackoverflow.com/questions/46939332/why-is-random-jitter-applied-to-back-off-strategies)

> Suppose you have multiple clients that send messages that collide. They all decide to back off. If they use the same deterministic algorithm to decide how long to wait, they will all retry at the same time -- resulting in another collision. Adding a random factor separates the retries.

**Describe the solution you'd like**

Add an option to apply jitter (probably just a boolean, or maybe something more fine-grained).

Possibly have this option enabled by default — would there be any disadvantages to doing that?

**Describe alternatives you've considered**

N/A

lionel-rowe:
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/ gives 5 examples of backoff functions, of which it concludes Decorrelated Jitter and Full Jitter are the optimal ones.

Those 5 functions could be implemented something like this:

```ts
export type BackoffFn = (
  cap: number,
  base: number,
  attempt: number,
  prev: number,
) => number;

export const backoffFns = {
  none: () => {
    return 0;
  },
  exponential: (cap, base, attempt) => {
    return Math.min(cap, base * 2 ** attempt);
  },
  fullJitter: (cap, base, attempt) => {
    return randomBetween(0, Math.min(cap, base * 2 ** attempt));
  },
  equalJitter: (cap, base, attempt) => {
    const temp = Math.min(cap, base * 2 ** attempt);
    return temp / 2 * randomBetween(0, temp / 2);
  },
  decorrelatedJitter: (cap, base, _attempt, prev) => {
    return Math.min(cap, randomBetween(base, prev * 3));
  },
} satisfies Record<string, BackoffFn>;
```

Of these, the current implementation used by `retry` is functionally equivalent to `exponential`.

They're all fairly simple, so providing all of them out-of-the-box and allowing consumers to supply a custom `BackoffFn` would allow a nice degree of flexibility.

PR #3379, opened 2023-05-13
feat(async): add jitter to retry exponential backoff

lionel-rowe:
Fixes https://github.com/denoland/deno_std/issues/3378


lino-levan:
I'm pretty strongly -1 against complicating the API to allow for arbitrary backoff functions. When I made the PR to introduce `async/retry`, I made a conscious decision to not do that (because outside of exponential backoff, what algorithm would you need?). I chose not to add jitter at the time because I didn't understand what the use case was but I think you did a good job in #3378 explaining why it might be useful.

TLDR; I think adding an option for jitter would be fine, but please don't overcomplicate the API with backoff functions.

lino-levan:
I never thought about making jitter enabled by default but honestly that makes sense (plus it removes an option which is always nice). Thoughts @iuioiua?

lionel-rowe:
> I never thought about making jitter enabled by default but honestly that makes sense (plus it removes an option which is always nice). Thoughts @iuioiua?

Ah I thought that was what you meant, but re-reading your post I see I misunderstood.

**Pros of always using jitter:**

1. Likely to significantly decrease number of failed API calls (especially those due to 429 Too Many Requests) in real-world scenarios
2. Likely to significantly decrease time-to-completion in real-world scenarios, mainly as a result of 1, but also because the average time between calls is spaced closer together
3. API remains completely unchanged

**Pros of having an option for jitter, disabled by default:**

1. Default behavior remains completely unchanged
2. More deterministic... sort of? I'm not sure whether an unknown power of `m` between `x` and `y` is meaningfully more deterministic than a completely random number between `x` and `y` here. Is there any reason a consumer would specifically want the number of seconds between each retry to be a power of 2?

**Pros of having an option for jitter, enabled by default:**

1. Gives the most-likely preferred behavior as default but allows opt-in to the old behavior

async/retry.ts
```diff
@@ -12,15 +12,15 @@ export class RetryError extends Error {
 export interface RetryOptions {
   /** How much to backoff after each retry. This is `2` by default. */
   multiplier?: number;
-  /** The maximum milliseconds between retries. This is `60000` by default. */
+  /** The maximum milliseconds between retries. This is `60000` by default. `-1` indicates no maximum. */
```
kt3k, line 15: > `-1` indicates no maximum.

Why is this handling necessary? The users can pass `maxTimeout: Infinity` if they don't want maximum timeout

async/retry_test.ts
```diff
@@ -31,11 +31,90 @@ Deno.test("[async] retry fails after max errors is passed", async function () {
...
+      for (let i = 0; i < 10; ++i) {
+        const result = _exponentialBackoffWithJitter(cap, base, i, 2);
+        total += Math.round(result);
+        results.push(Math.round(result));
+      }
+
+      assertEquals(results as unknown, row);
+    }
+
+    assertEquals(total, expectedTimingData.total);
```
kt3k, line 116: Asserting total value feels redundant to me

async/retry.ts
lionel-rowe: I just added that to document the existing behavior as of the current version. I guess removing the behavior wouldn't be a breaking change as it's currently undocumented.

async/retry_test.ts
lionel-rowe: Removed

kt3k (approved):
LGTM. Nice!

PR #3383, opened 2023-05-17
test(http): update cert error assertions

kt3k:
This PR fixes the CI on main.

The cert error message of http client seems a bit changed by this change ( https://github.com/denoland/deno/pull/18499 ) in CLI repo.


bartlomieju (approved):
LGTM, yeah, I was surprised that these messages changed

PR #3385, opened 2023-05-17
BREAKING(semver): rewrite semver

### LINKED ISSUE #3164: refactor(semver): make SemVer class immutable

kt3k:
continuation of #3169.

This PR rewrites `semver` module, and make `SemVer` object immutable.

Remaining tasks. We needs to restore deprecated signatures to the below APIs:
- [x] `intersects`
- [x] `maxSatisfying`
- [x] `minSatisfying`
- [x] `minVersion`
- [x] `toComparators`
- [x] `validRange`

closes #3164


justinmchase (approved):

semver/parse.ts
```diff
@@ -0,0 +1,635 @@
...
+    re[i] = new RegExp(src[i]);
+  }
+}
+
+/**
+ * Returns the parsed version, or undefined if it's not valid.
+ * @param version The version string to parse
+ * @returns A valid SemVer or `undefined`
+ */
+export function tryParse(version?: string): SemVer | undefined {
```
justinmchase, line 166: I think if you alter the version to be `version?: string | SemVer` that may reduce some dulicate code elsewhere which is attempting to differentiate on inputs. Also it would be consistent with the parse function below.

semver/compare.ts
```diff
@@ -0,0 +1,165 @@
...
+  s1: string | SemVer,
+  options?: { includePrerelease: boolean },
+): 1 | 0 | -1;
+export function compare(
+  s0: string | SemVer,
+  s1: string | SemVer,
+  options?: { includePrerelease: boolean },
+): 1 | 0 | -1 {
+  const v0 = typeof s0 === "string" ? parse(s0, options) : s0;
+  const v1 = typeof s1 === "string" ? parse(s1, options) : s1;
```
justinmchase, line 29: It seems like parse itself supports strings so why not just call it and have the logic to differentiate between string and SemVer in that one spot?

```suggestion
  const v0 = parse(s0, options);
  const v1 = parse(s1, options);
```

semver/compare.ts
```diff
@@ -0,0 +1,165 @@
...
+  s1: string | SemVer,
+  options?: { includePrerelease: boolean },
+): 1 | 0 | -1;
+export function compareBuild(
+  s0: string | SemVer,
+  s1: string | SemVer,
+  _options?: { includePrerelease: boolean },
+): 1 | 0 | -1 {
+  const v0 = typeof s0 === "string" ? parse(s0) : s0;
+  const v1 = typeof s1 === "string" ? parse(s1) : s1;
```
justinmchase, line 76:
```suggestion
  const v0 = parse(s0);
  const v1 = parse(s1);
```

semver/difference.ts
```diff
@@ -0,0 +1,51 @@
...
+  s1: string | SemVer,
+  options?: { includePrerelease: boolean },
+): ReleaseType | undefined;
+export function difference(
+  sv0: string | SemVer,
+  sv1: string | SemVer,
+  options?: { includePrerelease: boolean },
+): ReleaseType | undefined {
+  const s0 = typeof sv0 === "string" ? parse(sv0) : sv0;
+  const s1 = typeof sv1 === "string" ? parse(sv1) : sv1;
```
justinmchase, line 28:
```suggestion
  const s0 = parse(sv0);
  const s1 = parse(sv1);
```

semver/compare.ts
kt3k: Thanks. This makes sense.

kt3k:
This should be ready now.

Renamed `rangeTest`, `comparatorTest` -> `testRange`, `testComparator` because the file names `range_test.ts`, `comparator_test.ts` causes multiple issues (because they are auto-detected as test files).

semver/mod.ts
```diff
@@ -232,1720 +249,204 @@
...
+  if (value == null) return null;
+  const v = typeof value === "string" ? tryParse(value) : value;
+  if (v == null) return null;
+  return format(v);
 }
 
 /**
- * Returns the version incremented by the release type
- * (major, minor, patch, or prerelease), or null if it's not valid.
+ * @deprecated (will be removed after 0.191.0) Use `rangeTest()` instead
```
timreichen, line 351:
```suggestion
 * @deprecated (will be removed after 0.191.0) Use `testRange()` instead
```

semver/mod.ts
```diff
@@ -232,1720 +249,204 @@
...
- * down to a prerelease of that major version. `preminor`, and `prepatch` work
- * the same way.
- *
- * If called from a non-prerelease version, the `prerelease` will work the same
- * as `prepatch`. It increments the patch version, then makes a prerelease. If
- * the input version is already a prerelease it simply increments it.
- *
- * If the original version has build metadata and the `metadata` parameter is
- * `undefined`, then it will be unchanged.
+ * A compatibility function that calls rangeTest
```
timreichen, line 353:
```suggestion
 * A compatibility function that calls testRange
```

kt3k:
Summary of changes:
- Remove SemVer class, instead add SemVer interface. SemVers are now represented as immutable plain objects. This closes #3164
- Deprecate `string` overload for semver input from APIs. (Breaking change)
- Add single-file export entrypoint for every API.

semver/gt.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { SemVer } from "./types.ts";
+import { compare } from "./compare.ts";
+
+/** Greater than comparison */
+export function gt(
+  s0: SemVer,
+  s1: SemVer,
+): boolean;
+/**
+ * @deprecated (will be removed after 0.195.0) Use `gt(s0: SemVer, s1: SemVer)` instead.
```
bartlomieju, line 11: It's great that the backwards compatibility is still preserver for know - I suggest we give it anywhere between 1-3 months before removing the deprecated API.

So maybe let's be more generous here and deprecate in `0.200.0`?

bartlomieju (approved):
While I haven't reviewed the actual implementation, I read through the API changes. The is gonna be quite a breakage for users, but I think it's for the better - it makes it way more obvious which operations are more costly and requires more care from the user.


LGTM

semver/gt.ts
kt3k: 0.200.0 sounds good to me 👍 Updated

PR #3388, opened 2023-05-17
[WIP] refactor(semver): move to single-export files

iuioiua:
Please don't merge yet. Tests have deliberately been left till last. If existing tests pass, this means deprecated exports are working correctly.

### DIFF
--- a/semver/validity.ts
+++ b/semver/validity.ts
@@ -1,12 +1,9 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-import { ALL, NONE, SemVerComparator } from "./comparator.ts";
-import { SemVerRange } from "./range.ts";
-import { ANY, INVALID, SemVer } from "./semver.ts";
-import { Operator } from "./types.ts";
-
 export const MAX_LENGTH = 256;
 
 /**
+ * @deprecated (will be removed after 0.189.0) Import from `std/semver/is_semver.ts` instead.
+ *
  * Checks to see if value is a valid SemVer object. It does a check
  * into each field including prerelease and build.
  *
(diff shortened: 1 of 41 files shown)


semver/validity.ts
```diff
@@ -83,76 +46,39 @@ export function isSemVerComparator(value: unknown): value is SemVerComparator {
...
  */
-export function isValidString(value: unknown): value is string {
-  return (
-    typeof value === "string" &&
-    value.length > 0 &&
-    value.length <= MAX_LENGTH &&
-    !!value.match(/[0-9A-Za-z-]+/)
-  );
-}
+export { isValidString } from "./is_valid_string.ts";
```
kt3k, line 84: I think we don't need this kind of re-export from subpath because currently the all APIs are only exported from `mod.ts` under `semver/`.

iuioiua: Devs might be importing from files more directly rather than `mod.ts`. Removing these re-exports would break their setups without notice.

iuioiua:
Rename suggestions:
1. `rangeTest()` to `isWithinRange()`
2. `comparatorTest()` to `isWithinComparator()`
WDYT?

semver/validity.ts
kt3k: I mean the file `validity.ts` doesn't exist in the current main. This is introduced in the rewrite

iuioiua: If a function is exported and not prefixed with `_`, it probably should be added to `mod.ts`, as it's not private.

iuioiua:
All tests pass, meaning the existing API is maintained. I'll refactor tests now.

kt3k (approved):
LGTM. Nice work!

PR #3391, opened 2023-05-19
feat: very strongly typed collections (feedback request)

### LINKED ISSUE #3384: very strongly typed `collections/` functions

return type of some of the functions of `collections/` functions can be inferred in compile time.
inspiration: 
- https://github.com/dilan-dio4/strz
- https://github.com/type-challenges/type-challenges

for example, return type of distinct can be improved with:

```ts
type Arr = readonly unknown[]

type Distinct<T extends Arr, Rest extends Arr = []> 
    = T extends readonly [infer Head, ...infer Tail]
        ? Distinct<Tail, Head extends Rest[number] ? Rest : [...Rest, Head]>
        : Rest

function distinct<const T extends Arr>(array: T): Distinct<T> {
    const set = new Set(array);
    return Array.from(set) as Distinct<T>;
}

const d = distinct([1, 1, 1, 2, 'abc', 'abc', 2, 3]) // [1, 2, "abc", 3]
console.log(d) // [1, 2, "abc", 3]
```

[Playground Link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggTnKBeKcIEMAmB7AdgGxCgFdcBrXbAd1wG0BdAKEdEigBEBLAZ2E9wDGwADwAVKBAAewCLkzdYCADRQAShF4Tps+YsQoGAPiiMoZ5FHFSZchWix5CUWvwBmERAAkMmFQDoAtw9LdE58JnNIqAB+Dh4+QRFRUPwVbywtG111XlpcYgBbACMPehi1DWAoAC5nAL8c4DSfekNTKLNaxuZXUiFOPChMeP4hYQE8TSttWz1DAAp0BHQQWtEASlquXlGk4wBvdrMJ3E1uCCqUXAgqKABlC8XlkHWAbiPUC+I4XD0Vv1ccGwBXm52A6yg6AU2wSY1EhneAF9mCdNJgLMMdol5rQAIwqfFQQkAJhUAHJ0EUBGTyZTqSpSVAAMz0CEAejZzhJKgARHSeSoWYxUdh8BA-PhsABzeaYdmcvEM3n8wX0IA)

scarf005:
## Summary

- resolves #3384

feedbacks are greatly appreciated as i've hit some roadblocks in correctly typing functions.

## Problems

I've encountered many problems while implementing the feature, for example 
- i wasn't able to distinguish between readonly array and regular array
- resulting array type may be incorrectly marked as readonly
- non-readonly array input resulting in `[]`
- type uses structural compare which is different than referential comparison used by Set

More at: https://github.com/denoland/deno_std/issues/3384#issuecomment-1554301973

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -10,6 +10,7 @@
  * @module
  */
 
+import { Equal, NotEqual } from "../collections/_type_utils.ts";
 import { red, stripColor } from "../fmt/colors.ts";
 import { buildMessage, diff, diffstr } from "./_diff.ts";
 import { format } from "./_format.ts";
@@ -135,7 +136,14 @@ function constructorsEqual(a: object, b: object) {
     !a.constructor && b.constructor === Object;
 }
 
+type IsTruthy<T> = T extends Falsy ? false : true;
+type IsFalsy<T> = T extends Falsy ? true : false;
+
 /** Make an assertion, error will be thrown if `expr` does not have truthy value. */
+export function assert<const T>(
+  expr: T,
+  msg?: string,
+): IsTruthy<T> extends true ? void : never;
 export function assert(expr: unknown, msg = ""): asserts expr {
   if (!expr) {
     throw new AssertionError(msg);
@@ -144,6 +152,10 @@ export function assert(expr: unknown, msg = ""): asserts expr {
 
 /** Make an assertion, error will be thrown if `expr` have truthy value. */
 type Falsy = false | 0 | 0n | "" | null | undefined;
+export function assertFalse<const T>(
+  expr: T,
+  msg?: string,
+): IsFalsy<T> extends true ? void : never;
 export function assertFalse(expr: unknown, msg = ""): asserts expr is Falsy {
   if (expr) {
     throw new AssertionError(msg);
@@ -166,6 +178,11 @@ export function assertFalse(expr: unknown, msg = ""): asserts expr is Falsy {
  * });
  * ```
  */
+export function assertEquals<const X, const Y>(
+  actual: X,
+  expected: Y,
+  msg?: string,
+): Equal<X, Y> extends true ? void : never;
 export function assertEquals<T>(actual: T, expected: T, msg?: string) {
   if (equal(actual, expected)) {
     return;
@@ -202,6 +219,11 @@ export function assertEquals<T>(actual: T, expected: T, msg?: string) {
  * assertNotEquals<number>(1, 2)
  * ```
  */
+export function assertNotEquals<const X, const Y>(
+  actual: X,
+  expected: Y,
+  msg?: string,
+): NotEqual<X, Y> extends true ? void : never;
 export function assertNotEquals<T>(actual: T, expected: T, msg?: string) {
   if (!equal(actual, expected)) {
     return;
@@ -604,6 +626,7 @@ export function assertObjectMatch(
 /**
  * Forcefully throws a failed assertion
  */
+// @ts-expect-error
 export function fail(msg?: string): never {
   const msgSuffix = msg ? `: ${msg}` : ".";
   assert(false, `Failed assertion${msgSuffix}`);
--- a/collections/_type_utils.ts
+++ b/collections/_type_utils.ts
@@ -1,0 +1,43 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+export type Arr = readonly any[];
+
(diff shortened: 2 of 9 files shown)


testing/asserts.ts
```diff
@@ -135,7 +136,14 @@ function constructorsEqual(a: object, b: object) {
...
 
+type IsTruthy<T> = T extends Falsy ? false : true;
+type IsFalsy<T> = T extends Falsy ? true : false;
+
 /** Make an assertion, error will be thrown if `expr` does not have truthy value. */
+export function assert<const T>(
+  expr: T,
+  msg?: string,
+): IsTruthy<T> extends true ? void : never;
 export function assert(expr: unknown, msg = ""): asserts expr {
```
jsejcksn, line 147: This comment applies to this function and the others on which overload signatures have been added.

Adding non-assertion [overload signatures](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads) (without `assert` predicates) to [assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions) is an anti-pattern because the purpose of assertion functions in the type system is for narrowing of values' types at runtime.

Here's an example to demonstrate what I mean:

[TS Playground](https://www.typescriptlang.org/play?noUncheckedIndexedAccess=true&removeComments=true&target=99&jsx=4&exactOptionalPropertyTypes=true&inlineSourceMap=true&inlineSources=true&isolatedModules=true&noImplicitOverride=true#code/CYUwxgNghgTiAEEQBd4CMBc8AM8A+8AjANwBQpA3vAPTXwCCAdsgJYC0ADlMsiDI1gAiAexABnRgHJUjWDGEB3eMgCeHBMIBm8AG5QIAVwRRNvGPANiQpeMrUIAYvrEr4AXnibnCArl+N8eAAiIMDGAwgIQINGUE0WRhBgMltVdXgASTEAFRgDZAALFQAebIA+d3hs+BAAD15YsXgnCBd4AH5Pb3gsZDyQMhtPGLBWYQCoMSsYZGKwcbFUcoAKOo4YLGyASiws3Pyi0oq6huAmvqMO3WEWYB74RJ0+FOHGUZZx+Enp5FXa9awMQA1oxFIwdl8pnxkE01uYKABfci2WjBNCheYRO5oBC4YTmQjwdYfczIYTKArGKEzD4CIbfaHLNBbF62Nm0AB67WRNDoQXR8DEBWEWPg4wgrhxRC+qEKLCaHBuzAwQzQpE53KRlF58AAcnJFE1JjV-uBeMAhpoRmMJtTfnDAYwQWCIQyZrD-vCtbY3b9may2byuTy1RrSEigA)

```ts
declare let b: 0 | 1;

{ // Anti-pattern: Doesn't narrow type of value after use
  type Falsy = false | 0 | 0n | "" | null | undefined;
  type IsTruthy<T> = T extends Falsy ? false : true;

  function assert<const T>(expr: T): IsTruthy<T> extends true ? void : never;
  function assert(expr: unknown): asserts expr {}

  // "b" could be 0 or 1 prior to the assertion:
  assert(b);
       //^? let b: 0 | 1

  // "b" should only be 1 at this point:
  b
//^? let b: 0 | 1
}

{ // Narrows as expected
  function assert(expr: unknown): asserts expr {}

  assert(b);
       //^? let b: 0 | 1

  b
//^? let b: 1
}

```

jsejcksn: @scarf005 I also noticed that you linked to my comment in your PR message. Please don't take that or this critique personally, but do take them seriously: correct typings can be complicated — I just want to help ensure that if this kind of feature is implemented, it's done with thorough consideration, testing, and review.

kt3k:
@scarf005 
Could you work on each API in a separate PR? We can't see the impact of the PR as it is.

Question: Do you think this breaks any existing code?

I'd appreciate if you could give more examples which show how this change is practically useful.

kt3k:
@scarf005 
Could you work on each API in a separate PR? We can't see the impact of the PR as it is.

Question: Do you think this breaks any existing code?

I'd appreciate if you could give more examples which show how this change is practically useful.

iuioiua:
Hi @scarf005, I'll close this without merging, as it's gone stale. If you'd still like to pursue this, can you please open a new PR for each function, as previously mentioned, so we can better review each change? If we do it that way, we'll have a higher chance of getting them through the door if the design looks good. Either way, thank you.

scarf005:
sure, as I was hitting roadblocks. Will open a new PR if i find a better solution.

PR #3392, opened 2023-05-19
Add BLAKE2B-224 hashing algorithm in crypto

quentinadam:
I added the BLAKE2B-224 algorithm which was relatively trivial given that BLAKE2B-256 and BLAKE2B-384 had been added previously.
I recompiled the crypto wasm file with `deno task build:crypto` with the latest version of rust (1.69.0).


kt3k (approved):
LGTM. Thanks!

PR #3395, opened 2023-05-24
fix(streams/mergeReadableStreams): better error handling

jupegarnica:
'try catch' must be wrapping the inner body of the async function in order to be effective. In this way not unhandle errors will be thrown


crowlKats:
could we get a test?

jupegarnica:
I guess so, i will this weekend

jupegarnica:
@crowlKats tested 
:dancers:

kt3k (approved):
LGTM. Thanks!

PR #3396, opened 2023-05-24
feat(collections/zip): refine type definition

ayame113:
Improve the type definition of the zip function.

Inspired by #3391, [type-challenges/transpose](https://github.com/type-challenges/type-challenges/tree/d11ce46a603e916e29f7c9dde0a6be8de1471914/questions/25270-medium-transpose), [type-challenges/maximum](https://github.com/type-challenges/type-challenges/tree/d11ce46a603e916e29f7c9dde0a6be8de1471914/questions/09384-hard-maximum), [type-challenges/greater-than](https://github.com/type-challenges/type-challenges/issues/14098) and https://github.com/microsoft/TypeScript/issues/35660

image


ayame113:
The JSDoc for this function was misplaced, so no documentation was shown. I fixed that too.
https://deno.land/std@0.189.0/collections/zip.ts?s=zip
image

kt3k:
Sorry for the delay in review. I'm closing this without merge.

This seems adding better handling of types when the input array is tuple type. I think that is rare usage pattern of zip. So the added benefit in typing doesn't look worth it, compared to the added complexity in typing.

Thanks for your contribution anyway.

PR #3401, opened 2023-05-25
feat(collections): switch functions to take iterables when possible

Aplet123:
Many functions in the standard library already work with all iterables, so the argument types should be updated to reflect that. I also made small changes with negligible performance implications to `joinToString` and `permutations` to make them support iterables.

Unrelated to the rest of the PR, I also changed `unzip` to use the `Array` length constructor instead of `Array.from` since `Array.from` is for some reason much, much slower.

### DIFF
--- a/collections/associate_by.ts
+++ b/collections/associate_by.ts
@@ -26,7 +26,7 @@
  * ```
  */
 export function associateBy<T>(
-  array: readonly T[],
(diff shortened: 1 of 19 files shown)


Aplet123:
@kt3k I've moved the comment back down to where it was originally.

kt3k (approved):

Issue #3402, opened 2023-05-25
TOML misses a lot of edge cases


Aplet123:
There are a lot of edge cases that the current TOML parsing/stringifying code misses:

`toml.stringify({"": "a"})`:
Actual:
```toml
 = "a"
```
Expected:
```toml
"" = "a"
```
`toml.parse("a = [0.1]")`:
Actual:
```text
Uncaught Error: Parse error on line 1, column 6: Array is not closed
```
Expected:
```json
{"a":[0.1]}
```

`toml.stringify({"a":{"b":{}}})`:
Actual:
```text
(empty string)
```
Expected:
```toml
[a.b]

# or

a.b = {}
```

`toml.parse("a = {}")`:
Actual:
```text
Uncaught Error: Parse error on line 1, column 5: Invalid token after "{"
```
Expected:
```json
{"a": {}}
```

`toml.stringify({"a": [{"/": "b"}, "c"]})`:
Actual:
```toml
a = [{/ = "b"},"c"]
```
Expected:
```toml
a = [{"/" = "b"},"c"]
```

ry:
Patches welcome!

Aplet123:
@ry Unfortunately I'm much better at breaking parsers than making them so I'd appreciate if someone who's better than me at parsing could help out :)

brianknight10:
Did #3403 resolve this issue, or is there more work to be done? I'd be willing to work on it if there is.

Aplet123:
@brianknight10 #3403 fixes the 3 edge cases listed here relating to `toml.stringify`; the 2 cases relating to `toml.parse` are still unfixed as far as I know.

PR #3403, opened 2023-05-25
Various edge case fixes for `toml.stringify`

Aplet123:
Fixes the `stringify` errors in #3402

### DIFF
--- a/toml/test.ts
+++ b/toml/test.ts
@@ -711,3 +711,15 @@ aaaaa = 1
     assertEquals(actual, expected);
   },
 });
+
+Deno.test({
+  name: "[TOML] stringify empty key",
+  fn() {
+    const src = {
+      "": "a",
+      "b": { "": "c" },
+    };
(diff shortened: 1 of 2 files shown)


toml/test.ts
```diff
@@ -711,3 +711,38 @@ aaaaa = 1
...
+
+Deno.test({
+  name: "[TOML] stringify empty key",
+  fn() {
+    const src = {
+      "": "a",
+      "b": { "": "c" },
+    };
+    const roundTrip = parse(stringify(src));
+    assertEquals(src, roundTrip);
```
kt3k, line 723: I think we should directly assert the stringified result, instead of parsing it back to js object

kt3k (commented):
Thanks for the fix! Left one comment

Aplet123:
@kt3k I adjusted the tests.

kt3k (approved):
LGTM. Thanks!

PR #3419, opened 2023-06-01
feat(collections/sort_by): descending order can be specified in options

### LINKED ISSUE #3397: Should sort_by support descending order?
### Discussed in https://github.com/denoland/deno_std/discussions/3393

<div type='discussions-op-text'>

<sup>Originally posted by **sant123** May 22, 2023</sup>
Hi, I see that `collections/sortBy` function orders by ascending. Could it be possible to support descending? 

Also not sure if another overloading would be good or creating a new function for this.

Thanks.</div>

I can create a PR with a new function called `collections/sortByDesc`. Please let me know 🙏🏼

sant123:
Closes #3397

### DIFF
--- a/collections/sort_by.ts
+++ b/collections/sort_by.ts
@@ -1,10 +1,18 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
+/** Order */
+export type Order = "asc" | "desc";
+
+/** Options for sortBy */
+export type SortByOptions = {
+  order: Order;
+};
+
 /**
  * Returns all elements in the given collection, sorted by their result using
  * the given selector. The selector function is called only once for each
- * element.
+ * element. Ascending or descending order can be specified.
  *
  * @example
  * ```ts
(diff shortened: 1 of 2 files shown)


kt3k:
I think it's too verbose to duplicate all test cases for desc ordering (Also it's waste of CI time). I suggest we should add a single test case with title `[collections/sortBy] desc ordering` which exercises one or two typical cases.

collections/sort_by.ts
```diff
@@ -64,7 +97,7 @@ export function sortBy<T>(
     const b = selectors[bi];
     if (a === null) return 1;
     if (b === null) return -1;
```
kt3k, line 99: Do these lines also need to be changed?

How about swapping `a` and `b` above if the order is "desc"? (In that case I think we don't need to switch `comparator`)

sant123: Now they are swapped, good catch! Also the idea of `comparator` is to extract the function ordering once and call it when `a` and `b` are not null. However I do not understand why coverage says there's no coverage in `comparator` 🤔

collections/sort_by.ts
```diff
@@ -48,23 +68,39 @@ export function sortBy<T>(
...
+
   indexes.sort((ai, bi) => {
     const a = selectors[ai];
     const b = selectors[bi];
-    if (a === null) return 1;
-    if (b === null) return -1;
-    return a > b ? 1 : a < b ? -1 : 0;
+    if (a === null) return aIsNull;
+    if (b === null) return bIsNull;
+    return comparator(a, b);
```
kt3k, line 103: How about swapping a, b like the below? The current implementation looks hard to follow to me..

```js
  indexes.sort((ai, bi) => {
    let a = selectors[ai];
    let b = selectors[bi];
    if (order === "desc") {
      [a, b] = [b, a];
    }
    if (a === null) return 1;
    if (b === null) return -1;
    return a > b ? 1 : a < b ? -1 : 0;
  });
```

sant123: Definitely better, now got what you meant. Thank you so much @kt3k!

kt3k (approved):
LGTM

PR #3422, opened 2023-06-03
refactor(testing): remove unnecessary context processing in snapshot teardown

wafuwafu13:
https://github.com/denoland/deno_std/blob/2d9dd290f03e6edf8913fb0de7b9b0becb934e16/testing/snapshot.ts#L337

This `contexts` length is always 1 because `teardown` is only called once in a context.

https://github.com/denoland/deno_std/blob/2d9dd290f03e6edf8913fb0de7b9b0becb934e16/testing/snapshot.ts#L530
https://github.com/denoland/deno_std/blob/2d9dd290f03e6edf8913fb0de7b9b0becb934e16/testing/snapshot.ts#L410-L415


testing/snapshot_test.ts
```diff
@@ -459,6 +478,148 @@ Deno.test(
...
+          3,
+          4,
+        ]);
+      });`,
+    );
+
+    assertNoError(result1.error);
+    await assertSnapshot(t, formatTestOutput(result1.output), {
+      name: "Snapshot Test - Different Dir - New snapshot",
+    });
```
kt3k, line 567: I think this assertion is enough for checking this change isn't breaking things. Can you remove the rest of assertions and reduce the new snapshots?

kt3k:
> This contexts length is always 1 because teardown is only called once in a context.

I think `contexts.length` is at most 1 because each test file is run in different worker (since v1.10.0 of CLI). ref: (https://github.com/denoland/deno/pull/9815 , https://deno.com/blog/v1.10#improvements-to-deno-test )

testing/snapshot_test.ts
wafuwafu13: removed https://github.com/denoland/deno_std/pull/3422/commits/e7853d18390f880d036dafe77d07f2ee27c07b15

kt3k (approved):
LGTM

PR #3424, opened 2023-06-04
fix(http/file_server): use platform specific `resolve`

Chocobo1:
On Windows, running `posix.resolve("C:\\Users\\abc")` within `C:\Users\abc\Desktop\folderabc` would resolve to
`C:\Users\abc\Desktop\folderabc/C:\Users\abc` which is not expected.


Chocobo1:
>Can you write a test case for this? 

Sorry, I'm not sure what kind of test would do. Any ideas?

kt3k:
Start the server like this line: https://github.com/denoland/deno_std/blob/2d9dd290f03e6edf8913fb0de7b9b0becb934e16/http/file_server_test.ts#L488

With the first argument as a path starting with `C:/`. And then check a correct file is served, and stop the server.

Chocobo1:
I've added a test, please take a look.

kt3k (approved):
Thanks for adding a test case! LGTM

PR #3427, opened 2023-06-06
fix(async/retry): fix retry bugs and introduce jitter option

### LINKED ISSUE #3425: Retry is unreliable
We are using a retry to handle the situation where a peer service is still starting up.
For this we want to have a reliable time that the retry will back-off at least.

So with the initial retry implementation, the default implementation should have yielded a reliable retry of `31s = 1s + 2s + 4s + 8s + 16s`.
Now with the jitter introduced this would effectively be halved.

There are several problems with this however:
1. The jitter is not properly documented for the retry function.
2. If unlucky there could be five times `0s` back-off.
3. The implementation is faulty and will actually back-off (Math.rand times) `1s + 1s + 2s + 4s + 8s` by default and thus unexpectedly yield an even lower back-off timespan.
4. Looking more closely at the implementation there is one useless wait delay after the last attempt. So actually the real retry time is only `8s = 1s + 1s + 2s + 4s`
    - fn()
    - wait 1s * Math.rand
    - ...
    - fn()
    - wait 8s * Math.rand
    - raise RetryError

**Steps to Reproduce**

```typescript
import { retry } from "https://deno.land/std@0.190.0/async/retry.ts";
async function retryTime() {
  let start = new Date().getTime();
  try {
    await retry(() => { throw new Error("Failed") });
  } catch(exc) {}
  let stop = new Date().getTime();
  return stop - start;
};
await Promise.all([retryTime(), retryTime(), retryTime(), retryTime(), retryTime(), retryTime(), retryTime()])
```

**Expected behavior**

The back-off should be at-least `minTimeout` for each step and thus result in a total retry duration of at least `minTimeout * maxAttempts`.
This should be apparent from the documentation, as well as what the expected mean and maximum retry durations will be.

**Environment**

- OS: Arch Linux
- deno version: 1.33.2
- std version: 0.190.0

bodograumann:
Fixes two bugs in `async/retry` and introduces a new option `jitter`.
The option controls the relative amount of jitter to subtract from the retry delay.
It defaults to `1` to preserve the previous intended behaviour.
Using `0` completely disables jitter.

Resolves #3425

CC @kt3k


async/retry.ts
```diff
@@ -64,35 +93,36 @@ export async function retry<T>(
     options.minTimeout <= options.maxTimeout,
     "minTimeout is greater than maxTimeout",
   );
+  assert(options.jitter <= 1, "jitter is greater than 1");
```
kt3k, line 96: Maybe we should also assert `options.jitter >= 0`?

async/retry_test.ts
```diff
@@ -44,7 +71,7 @@ Deno.test(
 );
 
 Deno.test(
-  "[async] retry throws if minTimeout is less than 0",
+  "[async] retry throws if maxTimeout is less than 0",
```
kt3k, line 74: Thanks!

kt3k:
Mostly looks good to me. Left one question.

Also can you run `deno fmt` to format the source code?

async/retry.ts
bodograumann: I thought about that, but didn't seem necessary to me. It would simply increase the delay beyond the minimum then.

kt3k (approved):
LGTM. Nice work!

PR #3430, opened 2023-06-07
BREAKING(collections): change mapValues and reduceGroups return types

bekh6ex:


bekh6ex:
An interesting question to follow: Why do we have `K extends string` constraint, but not `K extends keyof any` as `Record` has? Why shouldn't we allow numbers and symbols as well?

kt3k:
This reverts #1880. I'm not sure this is an improvement..

> Why shouldn't we allow numbers and symbols as well?

Allowing symbols makes sense to me

bekh6ex:
Thanks for the reference. I was failing to come up with use-case where `Partial` makes sense. It looks like I took a wrong route.

Maybe `mapValues` signature has to be changed instead. Let me try something else

bekh6ex:
About allowing symbols: apparently symbol keys are ignored by `Object.keys()`, `Object.entries()` and `for ... in` loops. So I don't think it makes sense to add them. 

Adding numbers also doesn't make much sense, because they are getting transformed into strings as soon as become keys.

bekh6ex:
I've marked it as breaking change, but not sure how do you treat it.
The implementation hasn't changed, but existing scripts might start failing due to compilation errors.

bekh6ex:
There is a `license/cla` check hanging because of my incorrect local GIT setup (GitHub does not understand that it is me as well).

@kt3k can you or other maintainers just ignore it, or I need to redo the PR?

kt3k (approved):
LGTM

This handles union types of string literal types better. There's no change in runtime code.

I tried to add deprecated signature as overload, but it didn't seem working (https://github.com/denoland/deno_std/pull/3430/commits/fa1ac73212ff8ac01b8feedb3c521e124ece9a21 ). I think there's no way to land this with deprecation notice.

kt3k:
when K = string, the return type is changed from `Record<string, O>` to `Partial<Record<string, O>>` and the downstream code could break, but the difference of these 2 types looks subtle. I think it's ok to break in this case.

kt3k:
If we overload mapValue like the below, I think it won't be a breaking change:

```ts
export function mapValues<T, O, K extends string>(
  record: Readonly<Record<K, T>>,
  transformer: (value: T) => O,
): Record<K, O>;
export function mapValues<T, O, K extends string>(
  record: Readonly<Partial<Record<K, T>>>,
  transformer: (value: T) => O,
): Partial<Record<K, O>>;
export function mapValues<T, O, K extends string>(
  record: Readonly<Partial<Record<K, T>>>,
  transformer: (value: T) => O,
): any { ... }
```

Also by this typing, `Record<K, ...>` input returns `Record<K, ...>` which seems more desirable result.

What do you think?

iuioiua:
Hi @bekh6ex, are you able to apply Yoshiya's suggestion above? I'd like to push this PR forward if Yoshiya's in favour of it.

kt3k:
Closing in favor of #3978

PR #3434, opened 2023-06-08
docs(async/retry): consistently talk about attempts

### LINKED ISSUE #3433: (async/retry): what the argument maxAttempts of retry function does mean indeed? All attempts or all attempts except first attempt?
According to this line https://github.com/denoland/deno_std/blob/1ec581bf7e1c0ac6fc9a6c61d6e9603d628a0c1f/async/retry.ts#LL19C7-L19C35 (this line is introduced in https://github.com/denoland/deno_std/pull/3427, which is introduced in version 191 of std), I am confused about the meaning of the argument `maxAttempts`. I think number of retry attempts means attempts except first attempt, which is not a retry attempt. 
```ts
import { retry } from "https://deno.land/std@0.191.0/async/mod.ts";

let count = 0

function panic() {
  if (!count) {
    console.log('It is just an attempt.');
  } else {
    console.log(`It is ${count}th retry attempt.`);
  }
  count++;
  throw new Error()
}

// the argument maxAttempts is 5 by default.
retry(
  panic
)
```
The result is
```ts
It is just an attempt.
It is 1th retry attempt.
It is 2th retry attempt.
It is 3th retry attempt.
It is 4th retry attempt.
error: Uncaught (in promise) RetryError: Exceeded max retry count (5)
        throw new RetryError(error, options.maxAttempts);
              ^
    at retry (https://deno.land/std@0.191.0/async/retry.ts:104:15)   
Caused by: Error
    at panic (file:///D:/familyblue/codebase/deno_std/test.ts:12:9)  
    at retry (https://deno.land/std@0.191.0/async/retry.ts:101:20) 
```

bodograumann:
Attempts is always one more than retries, so it is confusing talking about retries, when we have `maxAttempts` as an option.
Updated docs and error message.

Resolves #3433

### DIFF
--- a/async/retry.ts
+++ b/async/retry.ts
@@ -4,8 +4,8 @@
 import { assert } from "../_util/asserts.ts";
 
 export class RetryError extends Error {
-  constructor(cause: unknown, count: number) {
-    super(`Exceeded max retry count (${count})`);
+  constructor(cause: unknown, attempts: number) {
+    super(`Retrying exceeded the maxAttempts (${attempts}).`);
     this.name = "RetryError";
     this.cause = cause;
   }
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM. Thanks!

kt3k:
Landing after the ongoing release process completed

PR #3439, opened 2023-06-09
fix(http/file_server): resolve empty subdir correctly on Windows

Chocobo1:
One goal of this commit is to avoid tripping over the 'current directory read permission' prompt from deno.
The prompt should not happen in the case when we already specified a path to the `--allow-read` argument.
This prompt can be seen when navigating to a subdir in a browser while running a deno server from Windows. This is because file_server mixed different path separators when building paths and then pass it to `Deno.stat()`. However `Deno.stat()` doesn't seem to handle it well and therefore generated the prompt.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -31,7 +31,14 @@
  * @module
  */
 
-import { extname, posix, resolve } from "../path/mod.ts";
+import {
+  extname,
+  join,
+  posix,
+  relative,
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

kt3k:
ref https://github.com/denoland/deno_std/pull/3444

Issue #3125, opened 2023-01-18
add string case utilities 


andrewbrey:
Today, there are no string case utilities in the `std lib`, but one need only look around to `npm` briefly to see that there is a healthy appetite for shared modules to handle the complexities of converting strings between different cases:

- [case](https://www.npmjs.com/package/case) (~500k weekly downloads)
- [case-anything](https://www.npmjs.com/package/case-anything) (~90k weekly downloads)
- [pascalccase](https://www.npmjs.com/package/pascalcase) (~16M weekly downloads)
- [camel-case](https://www.npmjs.com/package/camel-case) (~16M weekly downloads)
- [lodash.camelcase](https://www.npmjs.com/package/lodash.camelcase) (~8M weekly downloads)
- [lodash.deburr](https://www.npmjs.com/package/lodash.deburr) (~1M weekly downloads)

> Note that `lodash` has _lots_ of string utilities that would be excellent to have "official" std lib support for

Some of these examples are "general" string case conversion modules, and others are for a specific case conversion (and in the case of `lodash.deburr` it is a module for removing diacritical marks from Latin character sets to make the case conversions more reliable).

I myself incorporate `case` via `npm:` specifiers in my largest Deno project because having access to a well tested and well implemented module to convert a string to:

- upper (UPPER CASE)
- lower (lower case)
- title (Title Case)
- snake (snake_case)
- pascal (PascalCase)
- camel (camelCase)
- kebab (kebab-case)
- constant (CONSTANT_CASE)

which handles the edge cases around diacritical marks, weird spacing, etc is very helpful.

I saw https://github.com/denoland/deno_std/pull/3124 which attempts to establish a `strings` module category for the `std lib`, so these are related requests, though neither depends on the other.

Thank you for considering, and thank you for Deno (and the excellent `std lib`)!

naltun:
@andrewbrey do you see this being part of an existing module or something new?

andrewbrey:
> do you see this being part of an existing module or something new?

That's not something I have an opinion on - I mentioned the related PR because it's attempting to establish a `strings` top-level folder in the stdlib, where I could see the outcome of _this feature request_ living, but I could also see _this feature request_ tucking neatly within the existing `fmt` top-level folder as well.

iuioiua:
This has been mostly completed in #4082. Functions for more casing will be handled in new issues. Thank you for the suggestions and contributions.

guy-borderless:
[...]

PR #3440, opened 2023-06-09
feat(fmt/case): Add casing helpers

Gustrb:
This PR implements a few casing helpers, as were asked in the issue #3125 
It adds a few helpers methods such as:
- toPascalCase
- toCamelCase
- toSnakeCase
- toScreamingSnakeCase
- toKebabCase
- toTitleCase

Any suggestions on more methods would be appreciated, I've just implemented the ones I could find any use for


aapoalas:
Title case maybe?

timreichen:
I think it would make more sense to have an implementation that doesn't just split by whitespaces but also other separators and by words so a case can be applied to another like in [lodash](https://github.com/lodash/lodash/blob/ddfd9b11a0126db2302cb70ec9973b66baec0975/lodash.js#L14139).
Example:
```ts
toKebabCase('Foo Bar') // => 'foo-bar'
toKebabCase('fooBar') // => 'foo-bar'
toKebabCase('__FOO_BAR__') // => 'foo-bar'
```

Gustrb:
@timreichen true, it would make a lot more sense, I'll take a look at their implementation then :)

guy-borderless:
toSentenceCase

iuioiua:
Closing this in favour of #4082, which continues the work started here. Thank you for your contribution, @Gustrb!

Issue #3426, opened 2023-06-06
YAML returns `null` or `undefined`


oscarotero:
**Describe the bug**

This issue comes up here: https://github.com/lumeland/lume/pull/431#issuecomment-1578393330

Parsing an empty string returns `undefined`, but a string with only a comment, returns `null`.

There's a [relevant issue in the js-yaml library](https://github.com/nodeca/js-yaml/issues/567).

**Steps to Reproduce**

Run the following in REPL mode:

```
> import {parse} from "https://deno.land/std@0.189.0/yaml/parse.ts";
> parse('')
undefined
> parse('# test')
null
```

**Expected behavior**

I'd expect to return always `undefined` in all cases (or `null` if you prefer, but not both).

**Environment**

- OS: macOS
- deno version: 1.34.0
- std version: all

not-my-profile:
It's not just about comments the same odd behavior can also be observed with just newlines:

```
> parse('\n')
undefined
> parse('\n\n')
null
```

not-my-profile:
I guess this can be closed now that #3442 has been merged? (thanks btw!)

PR #3442, opened 2023-06-11
fix(yaml): parse always return null when file is empty, whitespace or only comments

JakeAve:
In issue #3426 there's differences with how YAML parse is treating empty YAML files. Sometimes they're undefined, but if you have extra lines, you can get back null. I don't really care what value we want to put in (null, undefined, ""), but the null was already being triggered when adding a yaml file with `\n`, so that's the value I rolled with. It also makes sense to me if loadDocuments returns something, either a parsed document, empty string or null - `undefined` gives the vibe that sometimes this function returns something and sometimes nothing. At least null or "" looks intentional.

### DIFF
--- a/yaml/_loader/loader.ts
+++ b/yaml/_loader/loader.ts
@@ -1788,7 +1788,7 @@ export function load(input: string, options?: LoaderStateOptions): unknown {
   const documents = loadDocuments(input, options);
 
   if (documents.length === 0) {
-    return;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks!

kt3k:
Does anyone have a preference of `undefined` over `null` in this situation?

Issue #1114, opened 2021-08-08
datetime: `dayOfYear()` wrong output when date is instantiated with individual date components


mamereu:
I stumbled on this bug while using dayOfYear inside a loop. Out of [the 4 ways to instantiate a Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date), dayOfYear fails on the fourth one, when individual date components are passed to the constructor, as in the following:

```
dayOfYear(new Date())                        // output 220, ok, today is 08/08/2021
dayOfYear(new Date("2021-08-08"))   // output 220, ok

const unixtime=new Date().getTime()
dayOfYear(new Date(unixtime))         // output 220, ok

dayOfYear(new Date(2021,7,8))         // output 219, wrong !!!  Where has been lost 1 day ? 

//let's go back...
dayOfYear(new Date(2021,0,1))       // output 366, of course should be 1 
dayOfYear(new Date(2021,0,2))      // output 1,  of course should be 2  

```

- OS: fedora 34
```
deno 1.12.2 (release, x86_64-unknown-linux-gnu)
v8 9.2.230.14
typescript 4.3.5
```

satyarohith:
Related https://github.com/denoland/deno_std/issues/925

iuioiua:
Is this issue fixed by #1384?

kt3k:
`dayOfYear` seems handling the given date as UTC. If the system's timezone has the negative offset, `new Date(2021, 0, 1)` becomes 2020-12-31 in UTC. So I think the above examples work as intended

lino-levan:
As far as I can tell, this issue was roughly related to the two PRs mentioned, but was never directly a bug and instead just weirdness with time zones. Should this be closed?

PR #3443, opened 2023-06-12
fix(datetime/day_of_year): respect time zone of date

familyboat:
Maybe fix https://github.com/denoland/deno_std/issues/1114


kt3k:
I'm not sure #1114 is a 'bug'.

Does it make sense to have both UTC and local timezone versions of APIs (like https://github.com/denoland/deno_std/pull/3295 ) ?

familyboat:
UTC version of `dayOfYear` does not make sense to me. When I call `dayOfYear`, I want to know what day of the year it is in my local time zone, which I think is the functionality of `dayOfYear`.
Even though, `dayOfUtcYear` (maybe another name) also makes sense to me.

familyboat:
When Intantiating `Date` with different arguments, the time zone of date varies. Some is utc time zone, some is your local time zone, some is your specified time zone, which will be converted to your local time zone. See more information in https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#date_time_string_format.
In this line https://github.com/denoland/deno_std/blob/aea85151e872338f9d6c52041e58ed14fbf21650/datetime/day_of_year.ts#L13 , the time zone of date is your local time zone, so the reault may be 70, or 69.
Here what I want to say is that developer should know the time zone of date instance, and time zone used in `dayOfYear` should be pointed out in docs. Leave the responsibility to the developer.

datetime/day_of_year.ts
```diff
@@ -13,14 +13,39 @@ import { DAY } from "./constants.ts";
...
+ * @example
+ * ```ts
+ * import { dayOfUtcYear } from "https://deno.land/std@$STD_VERSION/datetime/mod.ts";
+ *
+ * dayOfUtcYear(new Date("2019-03-11T03:24:00.000Z")) // output 70
+ * ```
+ *
+ * @return Number of the day in the year in UTC time
+ */
+export function dayOfUtcYear(date: Date): number {
```
kt3k, line 43: `Utc` in the middle of `dayOfYear` sounds strange to me. How about `utcDayOfYear` or `dayOfYearUtc`?

familyboat: I prefer `dayOfYearUtc`. Updated.

kt3k (approved):
LGTM

PR #3445, opened 2023-06-13
BREAKING(asserts): move `std/testing/asserts` to `std/assert`

iuioiua:
This change moves the functions in [std/testing/asserts.ts](https://github.com/denoland/deno_std/blob/aea85151e872338f9d6c52041e58ed14fbf21650/testing/asserts.ts) to their separate files in a new top-level `std/asserts` folder. Use of `_utils/asserts.ts` has also been replaced by `asserts/asserts.ts`. The reasoning behind this is:
1. Assertions can be called in both production and testing code.
2. It's at a sufficient size that it deserves its own submodule.
3. Currently, importing any number of assertion functions also imports all other assertions, bloating dependency imports sometimes. E.g. [Deno KV OAuth](https://deno.land/x/deno_kv_oauth) uses a custom [`assert()`](https://github.com/denoland/deno_kv_oauth/blob/main/src/_core.ts#L111) for this reason - to keep dependencies lean.

Related previous migrations:
* #3252
* #3251 
* #3243
* #3241 
* #3236
* #3228


testing/asserts.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
-/** A library of assertion functions.
+/**
+ * @deprecated (will be removed after 0.200.0) Import from `std/asserts` instead.
```
kt3k, line 4: Maybe let's wait for a little more than usual deprecation period? `testing/asserts` are heavily depended by the ecosystem.

iuioiua: The usual deprecation period is 3 minor versions. This is 9 minor versions, like the recent `std/semver` migration (#3385). Happy to increase this if you still think it should be pushed out.

ry (commented):
I'm ok with the move, but I think we should maintain backwards compatibility.

iuioiua:
> I'm ok with the move, but I think we should maintain backwards compatibility.

Assuming the usual release cadence of a minor release every two weeks, the deprecation would be in place for over 4 months before removal.

bartlomieju:
I would suggest we give it at least 3 months time before removing `std/testing/asserts`

iuioiua:
I've moved the assertion test and extended the removal version to 0.204.0. That's ~3 months from now, assuming a weekly release cadence.

iuioiua:
Do we want to replace the use of `_util/asserts.ts` with `std/asserts`? I'm for it.

iuioiua:
Ready to merge.

iuioiua:
Whoops. I just found out about https://github.com/denoland/deno_lint/issues/1159

kt3k:
Researched assertion features in other languages and their std libs.

- Node.js has [`assert`](https://nodejs.org/api/assert.html) module, which includes various assertions.
- Rust has [`assert`](https://doc.rust-lang.org/std/macro.assert.html), [assert_eq](https://doc.rust-lang.org/std/macro.assert_eq.html), etc macros in std. 
- PHP has:
  - [`assert`](https://www.php.net/manual/en/function.assert.php) builtin syntax.
  - doesn't have assertions as testing helper. (They are available as 3rd party libs such as PHPUnit)
- Python has:
  - [`assert` statement](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) in language.
  - [`assertEqaul`](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertEqual) in `unittest` std module.
- Kotlin has:
  - [assert](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/assert.html) in `kotlin`
  - [assertEquals](https://kotlinlang.org/api/latest/kotlin.test/kotlin.test/assert-equals.html) in `kotlin.test`
- Swift has:
  - [assert](https://developer.apple.com/documentation/swift/assert(_:_:file:line:)) builtin func.
  - [XCTAssertEqual](https://developer.apple.com/documentation/xctest/2142776-xctassertequal) in XCTest builtin framework.
- Ruby has various assertions under [`Minitest::Assertions`](https://ruby-doc.org/stdlib-3.0.2/libdoc/minitest/rdoc/Minitest/Assertions.html)
- Java has:
  - [`assert` statment](https://docs.oracle.com/javase/8/docs/technotes/guides/language/assert.html)
  - doesn't have assertions as testing helper. (They are available as 3rd party libs such as JUnit)
- Go doesn't have assertions in the std lib. Recommends using if statement and `t.Error` combination.

---

Node.js and Rust seems having unified concept of 'assert' which works for both runtime assertion and assertion as testing helper.

On the other hand, Python, Kotlin, and Swift have assertion for runtime checks and assertion for testing in different places.

PHP and Java only has assert for runtime checks in language. Assertions for testing are available from 3rd party libs.

Go doesn't seem providing assertions from language or std.

kt3k:
Other note:
non-test usage of asserts already look common to some extent by code search.
- https://github.com/keroxp/servest/blob/9a2892c05ca5f862fd1033dd3a2d7b853df1f13d/middleware.ts#L40
- https://github.com/hayd/deno-lambda/blob/e464ccf4ad0f391b160a3babde5680842acbfe17/_generate_types.ts#L7
- https://github.com/chiselstrike/chiselstrike/blob/3274f966dd67ee2a7963fff5fc48ee1eb9a33a41/tsc_reflection/src/main.ts#L90

If this is common expectation from the users of std, then we should provide single-export endpoints, and if we provide single-export endpoint for each assertion, `std/testing` doesn't look right place to do that as the module is already very crowded (`bdd`, `snapshot`, `mock`, etc)

piscisaureus:
Probably a good idea, although I'd call it std/assert or std/assertions. "asserts" is not a a proper noun.

testing/asserts.ts
```diff
@@ -1,6 +1,9 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
-/** A library of assertion functions.
+/**
+ * @deprecated (will be removed after 0.204.0) Import from `std/asserts` instead.
```
kt3k, line 4: We discussed about the removal version in the core team, and we consider these APIs are too common to remove soon. We'd prefer to set the removal version to 1.0.0 (We expect 1.0.0 will happen this year)

kt3k (approved):
LGTM

fbaltor:
Hey, I don't know if this is the right place to ask, but from my understanding we must change the asserts being used in the `deno` as well, right? 

image

bartlomieju:
@fbaltor it's very strange that you get these warnings - in `deno` repository we have a git submodule for `deno_std` (https://github.com/denoland/deno_std/tree/b23a76a47adaf63b5493761a87e6dd0dc0de0bb8) which hasn't been updated to latest version yet.

fbaltor:
You are absolutely right. I made some confusion by inadvertently updating the submodules. Thank you, @bartlomieju!

PR #3446, opened 2023-06-13
chore: parallel tests

iuioiua:
This change will make bugs present themselves, which are worth fixing.
Before: 1m37s
After: 20s (4.85x faster)

Inspired by https://github.com/denoland/fresh/pull/1281


iuioiua:
I've created https://github.com/denoland/deno/issues/19852 to fix this test flakiness. I think parallel tests are worth pursuing.

iuioiua:
So far, I've removed `fs/exists` tests and refactored `fs/walk` tests so they don't create and delete files during tests. Instead, test files and directories are static.

Now, there are flaky tests in `std/http`.

kt3k:
The changes in `fs/exists` test cases look good as there's no coverage diffs.

I don't see the problems in the existing `fs/walk` test cases with `--parallel` flag. What was the issues?

> fs/walk tests so they don't create and delete files during tests. Instead, test files and directories are static.

I think creating files are fine as they are created in temp dirs. They don't interfere other test cases.

iuioiua:
> I don't see the problems in the existing `fs/walk` test cases with `--parallel` flag. What was the issues?
> 
> > fs/walk tests so they don't create and delete files during tests. Instead, test files and directories are static.
> 
> I think creating files are fine as they are created in temp dirs. They don't interfere other test cases.

I'm not entirely sure, but possibly race conditions. These tests were the cause of many flaky tests. Once I removed the dynamic creation/deletion of the related folders, the errors stopped. Also, the previous had quite a lot going on which made it difficult to understand, IMO.

kt3k:
I think the flakiness comes from the dependency to `cwd` in test cases. We probably need to get rid of it.

In my view the previous tests were slightly better in the sense that they set up the fixture necessary to each case locally (You can understand why the test case passes only looking at each test block). With the current structure, which fixtures are used which test cases are unclear

iuioiua:
The first argument of `assertWalkPaths()` is `rootPath`, which should clarify which fixture is used in each test. I would've thought no setup is the best setup.

If you're certain the new tests are not more straightforward, I can fix the old tests instead. But I do think the new ones are clearer.

kt3k:
> The first argument of assertWalkPaths() is rootPath, which should clarify which fixture is used in each test. I would've thought no setup is the best setup.

Looking at the relation between test cases and fixtures again, they don't look that complex. Thanks for the explanation!

kt3k (approved):
LGTM. Thanks!

PR #3447, opened 2023-06-13
BREAKING(testing/snapshot): change tab char serialization

wafuwafu13:
Example test
```ts
Deno.test("Snapshot Test - Default serializer", async (t) => {
  await assertSnapshot(t, "a\nb\tc");
});
```

Before
```
snapshot[`Snapshot Test - Default serializer 1`] = `
"a
b\\tc"
`;
```

After
```
snapshot[`Snapshot Test - Default serializer 1`] = `
"a
b	c"
`;
```

### DIFF
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -204,7 +204,7 @@ export function serialize(actual: unknown): string {
     compact: false,
     iterableLimit: Infinity,
     strAbbreviateSize: Infinity,
-  }).replace(/\\n/g, "\n");
+  }).replace(/\\n/g, "\n").replace(/\\t/g, "\t");
 }
 
 /**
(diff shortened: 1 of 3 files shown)


testing/snapshot.ts
```diff
@@ -204,7 +204,7 @@ export function serialize(actual: unknown): string {
     compact: false,
     iterableLimit: Infinity,
     strAbbreviateSize: Infinity,
-  }).replace(/\\n/g, "\n");
+  }).replace(/\\n/g, "\n").replace(/\\t/g, "\t");
```
kt3k, line 207: We added option (`escapeSequences`) for disabling the escape of this kind of chars in https://github.com/denoland/deno/pull/19337. Let's use that option instead.

kt3k: Updated to use `escapeSequences: false`

kt3k (approved):
LGTM. Thanks for the initial idea!

lambdalisue:
Note that this change made it impossible for the same snapshot file to coexist on the old Deno and the new Deno...
The specific version is not known, but Deno without https://github.com/denoland/deno/pull/19337 still produces escaped output, so the old and new snapshots will be different.

Issue #1359, opened 2021-10-06
`walk` and `walkSync` skip over symlinks when `followSymlinks: false`


lucacasonato:
Instead I would expect symlinks to be yielded just like files: if `includeSymlinks` is set & the symlink matches the `exts`, `match`, and `skip` options.

iwinux:
The skip logic was introduced by https://github.com/denoland/deno_std/pull/1792.

Would @bnoordhuis kindly explain why are symlinks skipped in this way?

iwinux:
Forget the deleted comment, my mistake.

The "if not follow symlink then continue (skip)" logic actually dates back to the initial `fs.walk` implementation (see https://github.com/denoland/deno_std/pull/192), and gets carried over since then.

Would someone who knows the background explain the decision behind this? It seems like an easy fix (just remove the `continue` branch) but I'm afraid that something might break if we change that.

joehillen:
Here is some more context from the duplicate issue I posed:

This is really 2 issues:

1) no symbolic link files are returned when using `walk()`. See below to reproduce.

2) when using `followSymlinks: true`, the path to the destination is returned and labeled `{ isFile: false, isSymlink: true }`, even though the destination is a regular file. 


**Steps to Reproduce**

```ts
// walkTest.ts

import * as fs from "https://deno.land/std@0.142.0/fs/mod.ts";

fs.emptyDirSync("dir");

fs.ensureFileSync("dir/files/a.txt");
Deno.writeTextFileSync("dir/files/a.txt", "A");

// FIXME: See https://github.com/denoland/deno_std/issues/2312
// fs.ensureSymlinkSync("../files/a.txt", "dir/links/a.txt");

fs.ensureDirSync("dir/links");
await Deno.run({
  cmd: ["ln", "-s", "../files/a.txt", "dir/links/a.txt"],
}).status();

/*
  dir
  ├── files
  │   └── a.txt
  └── links
      └── a.txt -> ../files/a.txt
*/

console.log(">>> followSymlinks: false >>>>");
for (const f of fs.walkSync("dir", { followSymlinks: false })) {
  console.log(f); // dir/links/a.txt is not returned
}

console.log(">>> followSymlinks: true >>>>");
for (const f of fs.walkSync("dir", { followSymlinks: true })) {
  console.log(f);
  // returns:
  //  {
  //    path: "/home/joe/tmp/walkTest/dir/files/a.txt", <-- this is the path to the destination, not the link file.
[...]

PR #3453, opened 2023-06-16
fix(fs/walk): yield broken symbolic links

### LINKED ISSUE #3452: `fs.walk()` throws on broken symlinks
**Describe the bug**

When using [walk()](https://github.com/joehillen/deno_std/blob/main/fs/walk.ts#L81) with `followSymlinks` set to true
and there is a symlink that points to a destination that does not exist; the function will throw an error.

It's not possible to handle this error as a user because the error is thrown in
the `for await` loop deep in the recursive call stack.

**Steps to Reproduce**

On a Unix system, create a symlink that points to a non-existing destination:

```bash
mkdir /tmp/link-test
cd /tmp/link-test
ln -s /tmp/non-existing-dir/ link
```

Then run the following code:

```ts
import { walk } from "https://deno.land/std@0.192.0/fs/walk.ts";

walk("/tmp/link-test", { followSymlinks: true })
  .then((files) => {
    console.log(files);
  });
```

The following exception will be thrown:

```txt
error: Uncaught (in promise) WalkError: No such file or directory (os error 2): realpath '/tmp/link-test/link' for path "/tmp/link-test"
  return new WalkError(err, root);
         ^
    at wrapErrorWithPath (https://deno.land/std@0.192.0/fs/walk.ts:48:10)
    at walk (https://deno.land/std@0.192.0/fs/walk.ts:134:11)
    at eventLoopTick (ext:core/01_core.js:182:11)
    at async file:///tmp/link-test.ts:3:18
Caused by: NotFound: No such file or directory (os error 2): realpath '/tmp/link-test/link'
    at async walk (https://deno.land/std@0.192.0/fs/walk.ts:112:16)
    at async file:///tmp/link-test.ts:3:18
```

**Expected behavior**

I don't think it's useful to throw when trying to `walk()` a broken link, as there is nothing the caller can do to handle it when it happens.

The simplest solution would be to catch and ignore the `NotFound` exception
when calling `realpath()` on a symlink.

**Environment**

- OS: Arch Linux
- deno version: 1.34.1
- std version: 0.192.0

joehillen:
Fixes #3452


luk3skyw4lker (changes requested):

fs/walk.ts
```diff
@@ -109,11 +109,19 @@ export async function* walk(
...
+          // Caveat emptor: don't assume |path| is not a symlink. realpath()
+          // resolves symlinks but another process can replace the file system
+          // entity with a different type of entity before we call lstat().
+          ({ isSymlink, isDirectory } = await Deno.lstat(path));
+        } catch (err) {
+          // Ignore broken symlinks.
+          if (err instanceof Deno.errors.NotFound) {
+            continue;
+          }
+          throw err;
```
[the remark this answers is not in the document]
joehillen: I thought of that. Why would someone want to try to follow a dead link if they pick `followSymlinks = true`? You should only get a file or directory back in that scenario.

Unix programs don't error when following symlinks. For example, `find`:

```sh
+ cd /tmp/link-test
+ touch a
+ mkdir b
+ touch b/c
+ ln -s b l2b
+ ln -s /this-path-does-not-exist badlink
+ find -L /tmp/link-test
/tmp/link-test
/tmp/link-test/badlink
/tmp/link-test/l2b
/tmp/link-test/l2b/c
/tmp/link-test/b
/tmp/link-test/b/c
/tmp/link-test/a
```

Having to think about handling links defeats the purpose of the option. This is compounded by https://github.com/denoland/deno_std/issues/1359, which will never return `isFile = true`; this makes  `followSymlinks` useless in its current form.

There is the added problem that `followSymlinks = false` won't even return a symlink:

```ts
for await (const entry of walk("/tmp/link-test")) {
  console.log({ path: entry.path, isSymlink: entry.isSymlink });
}
```

Notice that neither `l2b` or `badlink` are returned:

```
└> deno run --allow-read test.ts
{ path: "/tmp/link-test", isSymlink: false }
{ path: "/tmp/link-test/b", isSymlink: false }
{ path: "/tmp/link-test/b/c", isSymlink: false }
{ path: "/tmp/link-test/a", isSymlink: false }
```

This is also bad because it means there is currently no good way to handle dead symlinks with `walk()`. (I only just now noticed this second problem, so I'll open a new issue for that.)

fs/walk.ts
```diff
@@ -109,11 +109,19 @@ export async function* walk(
...
+        try {
+          path = await Deno.realPath(path);
+          // Caveat emptor: don't assume |path| is not a symlink. realpath()
+          // resolves symlinks but another process can replace the file system
+          // entity with a different type of entity before we call lstat().
+          ({ isSymlink, isDirectory } = await Deno.lstat(path));
+        } catch (err) {
+          // Ignore broken symlinks.
+          if (err instanceof Deno.errors.NotFound) {
+            continue;
```
kt3k, line 121: How about yielding the broken symlink as it is instead of ignoring?

iuioiua:
Hi @joehillen, are you still happy to pursue this PR? If so, can you apply Yoshiya's previous suggestion and fix the merge conflicts?

joehillen:
@iuioiua Thank you for the reminder. I had forgotten about this. Unfortunately I'm busy most of this week, but I'll be able to pick this up later this week or next.

iuioiua:
All good! Thanks for letting me know 👍🏾

joehillen:
I've updated the PR and made a couple of tweaks and added `destination` to `WalkEntry`.

@iuioiua @kt3k  Ready for review.

joehillen:
@iuioiua @kt3k Ready for review!

I've made several additional tweaks which are broken into small commits for easy review.

deno.json
```diff
@@ -13,6 +13,7 @@
   },
   "tasks": {
     "test": "deno test --doc --unstable --allow-all --parallel --coverage --trace-ops",
+    "debug": "deno test --doc --unstable --allow-all --parallel --trace-ops --inspect-wait",
```
iuioiua, line 16: This doesn't appear to be related. Can you please revert?

iuioiua (changes requested):

deno.json
iuioiua: Please open a new issue discussing the suggestion and we can go from there 🙂

iuioiua: I've gone ahead and reverted this change.

iuioiua:
PTAL @kt3k

kt3k:
Closing as we have better way to handle broken symlinks. See https://github.com/denoland/deno_std/issues/3452#issuecomment-2058863401 for more details.

Also the API changes to expandGlob don't seem discussed very well. I don't find they are desirable changes.

PR #3455, opened 2023-06-17
fix(fs/walk): return symbolic links (#1359)

### LINKED ISSUE #1359: `walk` and `walkSync` skip over symlinks when `followSymlinks: false`

joehillen:
Symbolic links were not returned. This means that `walk()` is not listing all the contents of a directory.

Some use cases for this is are:

- Copying a directory recursively.
- Finding invalid symbolic links.
- Modify link targets without resolving them.

Other changes:

- walk_test.ts: Rename `walkArray()` to `collectPaths()` because it is more descriptive.
- walk_test.ts: Add `collectEntries()`

closes #1359

### DIFF
--- a/fs/expand_glob_test.ts
+++ b/fs/expand_glob_test.ts
@@ -51,6 +51,7 @@ Deno.test("expandGlobWildcard", async function () {
     "abc",
     "abcdef",
     "abcdefghi",
+    "link",
     "subdir",
   ]);
 });
@@ -70,6 +71,7 @@ Deno.test("expandGlobParent", async function () {
     "abc",
     "abcdef",
     "abcdefghi",
+    "link",
     "subdir",
   ]);
 });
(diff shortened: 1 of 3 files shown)


kt3k:
Looks like reasonable changes to me. Thanks for your suggestions!

Can you take a look into the test failures in `fs/expand_glob_test.ts`? Can you also run `deno fmt` to pass the lint check?

joehillen:
@kt3k Should be fixed now. Sorry for the delay. It's been a crazy month.

joehillen:
I think I fixed the test failure. I don't have access to windows, so I can't be sure until it runs again

kt3k:

After reading the original issue again, I think we should add `includeSymlinks` option instead of unconditionally returning the symlinks. We also need to check the matching of `match`, `exts`, `skip` before yielding.

Closing this one in favor of #3464, which implements `includeSymlinks` option.

Thanks for creating this PR anyway.

PR #3464, opened 2023-06-24
feat(std/fs.walk): include symlink option

MrKleeblatt:
Instead of completely ignoring symlinks if `followSymlinks == false` there should be an option to at least return the symlink (and don't resolve the real path)


kt3k (approved):

PR #3466, opened 2023-06-24
feat(json): Primitives at top level without separator

grimly:
Hello,

I am consuming an API that uses concatenated JSON when streaming.
I came across the issue this API sent `nullnull` in the stream.

I can't really blame the API because they are clearly saying they have no separator between JSON values and null is valid for their API.

I updated the transformer to allow for `null`, `true` and `false` at top level without any separation at all.

Best regards,
Michel


kt3k (approved):
LGTM

PR #3467, opened 2023-06-25
feat: add Pull Request auto labeler action

niklasmtj:
Add a GitHub action that automatically labels the PR with the names of the modules changed. Currently, it seems that PRs don't always follow the conventional commit naming convention, resulting in PR names that are not directly associated with the changed modules/code.

The action used is maintained in GitHub's actions repository and is used by ~47k repositories. The action takes ~14s and automatically sets the configured labels. You can see that it needs [`pull-requests: write`](https://github.com/actions/labeler/#permissions) permissions. There is also an option to include dot directories like `.github` if desired.

Information about other options and configurations can be found [here](https://github.com/actions/labeler/#match-object).

Recent:
- Added the labels of the `std` library modules only.

Idea:
- Add the `tests` label when tests have been added in a PR.

## How does this help?
Get a quicker and easier overview of the changed code in the PRs without having to check each PR individually.

If there are more labels that could be helpful, let me know or push the PR branch :)

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,66 +1,66 @@
 archive:
-- archive/**
+  - archive/**
 async:
-- async/**
+  - async/**
 bytes:
-- bytes/**
+  - bytes/**
 collections:
-- collections/**
+  - collections/**
 console:
-- console/**
+  - console/**
 crypto:
-- crypto/**
+  - crypto/**
 csv:
-- csv/**
+  - csv/**
 datetime:
-- datetime/**
+  - datetime/**
 dotenv:
-- dotenv/**
+  - dotenv/**
 encoding:
-- encoding/**
+  - encoding/**
 examples:
-- examples/**
+  - examples/**
 flags:
-- flags/**
+  - flags/**
 fmt:
-- fmt/**
+  - fmt/**
 front_matter:
-- front_matter/**
+  - front_matter/**
 fs:
-- fs/**
+  - fs/**
 html:
-- html/**
+  - html/**
 http:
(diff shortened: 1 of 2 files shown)


kt3k:
I'm not sure we're interested in introducing this.

Some notes:

query to find the user repos: https://github.com/search?q=path%3A.github%2Flabeler.yml&type=code

some examples:
- Kong https://github.com/Kong/kong/blob/3a491c81b460aa6d5376a28cd8b953c166598d06/.github/labeler.yml
- Rome https://github.com/rome/tools/blob/e18ebe62e6718ff228324ed2b58a15436e0d1c93/.github/labeler.yml
- https://github.com/CleverRaven/Cataclysm-DDA/blob/595f4f73a6787d9f863ba91d5079fbe20cb4d54c/.github/labeler.yml
  - This repo seems using it in a different way (language detection, doc PR detection etc):

niklasmtj:
You can also just close the PR if it's not interesting for the project. It was just an idea that might help getting an overview over the recent PRs. But I'm not mad if this get's closed without a merge :-D

kt3k:
I'll let this open for a while to see the feedback from the maintainers and community.

iuioiua:
I'd have to see it in action to be convinced, so I think it's worth a try.

kt3k (approved):
Fair enough. Let's try

niklasmtj:
Nice to see! If you have any feedback let me know. The tagging will work starting with the next created PR. If there are tags that should be added / removed let me know. Happy to work on that.

Issue #3451, opened 2023-06-16
`assertObjectMatch` throws error for non-object


vwkd:
**Describe the bug**

`assertObjectMatch` should check partial equality of two objects.

Currently, when calling `assertObjectMatch` with a non-object in the first argument, it throws a cryptic `Uncaught TypeError: Invalid value used as weak map key` error and if it's the second argument `Uncaught TypeError: Cannot use 'in' operator to search for 'yyy' in xxx`.

Confusing cryptic errors depending on argument order aside, for top-level non-object inputs we should use `assertEquals`. But the above errors are also thrown for any nested property due to the use of recursion.

For example, currently `assertObjectMatch` breaks if some deep property that expects a nested object turns out to be `null` or `undefined`. This should pass, for example to check a partially successful GraphQL response.

**Steps to Reproduce**

```js
import { assertObjectMatch } from "https://deno.land/std@0.191.0/testing/asserts.ts";

assertObjectMatch({ foo: null }, { foo: { bar: 42 }});
// error: Uncaught TypeError: Invalid value used as weak map key

assertObjectMatch({ foo: { bar: 42 }}, { foo: null });
// switch arguments and it works

assertObjectMatch(null, { foo: 42 });
// error: Uncaught TypeError: Invalid value used as weak map key

assertObjectMatch({ foo: 42 }, null);
// error: Uncaught TypeError: Cannot use 'in' operator to search for 'foo' in null

assertObjectMatch(21, 42);
// error: Uncaught TypeError: Invalid value used as weak map key
```

**Expected behavior**

`assertObjectMatch` should handle non-object inputs. It should pass a superset of inputs that pass `assertEquals`, meaning, it should throw less often, not more often.

**Environment**

- OS: macOS 12.6.6
- deno version: deno 1.34.1
- std version: 0.191.0

luk3skyw4lker:
I'm taking this issue!

luk3skyw4lker:
Just an update: I'm still working on it, but I'm having a little trouble in understanding the code behind the function, I'll be sending a PR anytime soon

JakeAve:
[...]

PR #3468, opened 2023-06-26
fix: Added catches for primitive/null values passed as "actual" in assertObjectMatch

JakeAve:
Having issues when null and undefined are passed in the 'actual' argument for `assertObjectMatch`. Typeof null is 'object', so in recursion null kept being sent as a weakmap, but should have been assigned as `filtered[key] = value`.

Primitive types also could be passed in as the 'actual' value in JS, and they gave a cryptic weakmap error, but it's just a typeerror.

I slightly disagree with the AssertionError instead of the TypeError in issue #3451, but I'm open to suggestions.

### DIFF
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -62,10 +62,7 @@ export function equal(c: unknown, d: unknown): boolean {
       return true;
     }
     if (a && typeof a === "object" && b && typeof b === "object") {
-      if (
-        a && b &&
-        !constructorsEqual(a, b)
-      ) {
+      if (a && b && !constructorsEqual(a, b)) {
         return false;
       }
       if (a instanceof WeakMap || b instanceof WeakMap) {
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -545,11 +545,7 @@ export function assertObjectMatch(
         seen.set(a, b);
       } catch (err) {
         if (err instanceof TypeError) {
-          throw new TypeError(
-            `Cannot assertObjectMatch ${
-              a === null ? null : `type ${typeof a}`
-            }`,
-          );
+          return a;
         } else throw err;
       }
       // Filter keys and symbols which are present in both actual and expected
--- a/testing/asserts.ts
+++ b/testing/asserts.ts
@@ -64,10 +64,7 @@ export function equal(c: unknown, d: unknown): boolean {
     if (a && typeof a === "object" && b && typeof b === "object") {
       if (
         a && b &&
-        !constructorsEqual(
-          a as unknown as Record<string, unknown>,
-          b as unknown as Record<string, unknown>,
-        )
+        !constructorsEqual(a, b)
       ) {
         return false;
       }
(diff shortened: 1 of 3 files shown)


JakeAve:
I know that @luk3skyw4lker looked at this too. Can we get his blessing in case he has something better?

luk3skyw4lker:
Hey @JakeAve maybe you should run the `deno fmt` command because there are many diffs that are mainly code formats and deno already has a pattern for code format.

testing/asserts.ts
```diff
@@ -564,20 +573,23 @@ export function assertObjectMatch(
           filtered[key] = value;
           continue;
         } // On nested objects references, build a filtered object recursively
-        else if (typeof value === "object") {
+        else if (typeof value === "object" && value !== null) {
```
JakeAve, line 576: Real change is checking that the value is not null so it doesn't get sent for another recursion

testing/asserts.ts
```diff
@@ -539,10 +536,22 @@ export function assertObjectMatch(
 
     function fn(a: loose, b: loose): loose {
       // Prevent infinite loop with circular references with same filter
-      if ((seen.has(a)) && (seen.get(a) === b)) {
+      if (seen.has(a) && seen.get(a) === b) {
         return a;
       }
-      seen.set(a, b);
+
+      try {
```
JakeAve, line 543: Real change is catching `WeakMap.set()`, which will throw a TypeError for primitives and non-iterables. The other changes are `deno fmt`

JakeAve:
@luk3skyw4lker I agree. @kt3k I'm on 
```
deno 1.34.3 (release, aarch64-apple-darwin)
v8 11.5.150.2
typescript 5.0.4
```
but my `deno fmt` is changing a lot more lines than I feel like it should.

kt3k:
@JakeAve I think these formatting changes were caused by your editor settings (do you enable any auto formatter?). Deno doesn't seem introducing these changes recently (and `deno fmt` allows/respects some formatting variations). Could you revert the change, disable auto formatter, and redo the actual changes?

testing/asserts.ts
```diff
@@ -541,7 +541,17 @@ export function assertObjectMatch(
...
-      seen.set(a, b);
+      try {
+        seen.set(a, b);
+      } catch (err) {
+        if (err instanceof TypeError) {
+          throw new TypeError(
+            `Cannot assertObjectMatch ${
+              a === null ? null : `type ${typeof a}`
+            }`,
+          );
```
kt3k, line 552: How about returning `a` here instead of throwing `TypeError`? That makes the case `assertObjectMatch(null, { foo: 42 })` throwing more useful AssertionError instead of TypeError.

JakeAve: I'm open to that. I feel like assertObjectMatch is supposed to be for higher ordered objects, so I've been wanting to do a typeerror, but I understand you could have a function that returns either null/undefined or an object. The problem I have is there are ts warnings whenever you try to pass something that couldn't be made into a weakmap, so if we go that route, I wonder if we update the argument types to accept more values than 
```
actual: Record<PropertyKey, any>,
expected: Record<PropertyKey, unknown>,
```
I'm pushing the updates so we can see how it looks.

kt3k: > I feel like assertObjectMatch is supposed to be for higher ordered objects, so I've been wanting to do a typeerror

Ah, right. The previous version might make more sense. Thanks for the clarification of the reasoning.

kt3k (approved):
LGTM

PR #3470, opened 2023-06-28
chore: fix lint errors

kt3k:
`object` type is removed from `ban-types` list in canary. 

ref:
- https://github.com/denoland/deno/pull/19619
- https://github.com/denoland/deno_lint/issues/1159


bartlomieju (approved):
LGTM

PR #3119, opened 2023-01-16
feat: new `std/webgpu` sub-module

crowlKats:
Closes #3036

PR #3476, opened 2023-07-04
chore(ci): test on stable

iuioiua:
This change ensures CI passes on the latest stable release of Deno.


kt3k:
Compatibility with stable deno is sometimes very difficult to keep (if deno canary has some changes with breaking nature), and in that case we have to prioritize canary (The latest CLI release needs to pass the latest std test cases). That is the reason we only run canary in std CI.

refs:
- https://github.com/denoland/deno_std/issues/870
- https://github.com/denoland/deno_std/pull/2991

(I think there was also a case where it was impossible to pass both canary and stable CI in the past, but I don't find it now..)

iuioiua:
> (I think there was also a case where it was impossible to pass both canary and stable CI in the past, but I don't find it now..)

Yeah, I agree. To elaborate, this would be because the Deno and Standard Library APIs are becoming more stable over time.

iuioiua:
This PR is ready for review.

kt3k:
I think we should start CI with stable CLI when we released stable version of std. ref https://github.com/denoland/deno_std/issues/3489

iuioiua:
Actually, I think we should merge this before v1, so it's confirmed working fine for v1.

kt3k (approved):
Let's try and see.

kt3k:
ref https://github.com/denoland/deno_std/pull/3119

PR #3477, opened 2023-07-04
chore: fix type errors caused by canary

kt3k:
Some type errors started appearing because of TS upgrade in canary (See https://github.com/denoland/deno_std/actions/runs/5450255896/jobs/9915333243 for details). This PR fixes those type errors.

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -130,7 +130,7 @@ async function fetchExactPath(
   const encoder = new TextEncoder();
   const decoder = new TextDecoder();
   const request = encoder.encode("GET " + path + " HTTP/1.1\r\n\r\n");
-  let conn: void | Deno.Conn = undefined;
+  let conn: undefined | Deno.Conn;
   try {
     conn = await Deno.connect(
       { hostname: hostname, port: port, transport: "tcp" },
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -130,7 +130,7 @@ async function fetchExactPath(
   const encoder = new TextEncoder();
   const decoder = new TextDecoder();
   const request = encoder.encode("GET " + path + " HTTP/1.1\r\n\r\n");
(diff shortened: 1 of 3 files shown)


dsherret (approved):
Interesting… seems worse at figuring stuff out.

http/file_server_test.ts
```diff
@@ -130,7 +130,7 @@ async function fetchExactPath(
   const encoder = new TextEncoder();
   const decoder = new TextDecoder();
   const request = encoder.encode("GET " + path + " HTTP/1.1\r\n\r\n");
-  let conn: void | Deno.Conn;
+  let conn: void | Deno.Conn = undefined;
```
aapoalas, line 133: The `void` here is strange. I'd definitely use `undefined` here instead. Would it work without assignment with that?

http/server_sent_event.ts
```diff
@@ -345,10 +345,11 @@ export class ServerSentEventStreamTarget extends EventTarget
    * {@linkcode ResponseInit} needed to create a response that will establish
    * a SSE connection with the client. */
   asResponseInit(responseInit: ResponseInit = {}): [BodyInit, ResponseInit] {
-    responseInit.headers = new Headers(responseInit.headers);
+    const headers = new Headers(responseInit.headers);
```
aapoalas, line 348: This ends up being a fairly okay perf boost as well, nice! :)

aapoalas (approved):
Looks good to me, though I'd personally change one thing.

http/file_server_test.ts
kt3k: Oh, that seems working. Thanks!

aapoalas:
> Interesting… seems worse at figuring stuff out.

At work my friend was saying that basic if-else type clauses have stopped working in some cases in newer TS. There may be some sort of regression happening.

PR #3479, opened 2023-07-04
fix(async/delay): reject with existing AbortSignal reason

jsejcksn:
Currently the `delay` function always rejects with a custom `DOMException` instance when aborted by an `AbortSignal`, ignoring the abort [`reason`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/reason) and nullifying its purpose. This PR updates the function to reject with the existing reason value. This will allow for better exception-handling in user code and enhance exception diagnostic debugging.


async/delay.ts
```diff
@@ -39,12 +39,20 @@ export interface DelayOptions {
 export function delay(ms: number, options: DelayOptions = {}): Promise<void> {
   const { signal, persistent } = options;
   if (signal?.aborted) {
-    return Promise.reject(new DOMException("Delay was aborted.", "AbortError"));
+    return Promise.reject(
+      typeof signal.reason === "undefined"
```
justinmchase, line 43: I think this can be condensed to `signal.reason ?? 'Delay was aborted.'` but in principal this change looks like a no brainier to me.

justinmchase (commented):
Maybe use the `??` operator to simplify but 👍

async/delay.ts
jsejcksn: Thames for the feedback! I considered that, but then realized it would prevent `null` from being forwarded, and it's just as meaningful a value as any other.

kt3k: It looks like that `signal.reason` can't be `undefined` when it's aborted.

```
> let ac = new AbortController()
undefined
> ac.abort(undefined)
undefined
> ac.signal.reason
DOMException: The signal has been aborted
```

, which means `new DOMException("Delay was aborted.", "AbortError")` part is never reached.

How about checking if `reason` is DOMException and AbortError, and returning the default error in that case (assuming that the user didn't give `reason` in that case)

jsejcksn: @kt3k

> It looks like that `signal.reason` can't be `undefined` when it's aborted.
> 
> ```
> > let ac = new AbortController()
> undefined
> > ac.abort(undefined)
> undefined
> > ac.signal.reason
> DOMException: The signal has been aborted
> ```
> 
> , which means `new DOMException("Delay was aborted.", "AbortError")` part is never reached.

According to the [spec](https://dom.spec.whatwg.org/#interface-AbortSignal), that's right — and Deno's implementation appears to conform in that way. The only reason I special-cased `undefined` was as a grace measure for a potential case where a nonstandard `AbortSignal` is provided (one that doesn't create a value when no argument is provided to the `abort` method) — this module is marked as browser-compatible after all, and Deno isn't the only runtime where it's used.

In hindsight, that might have been overengineering — I reversed those changes, removing the default reason entirely, and updated the tests to clarify the intent of the change — **to reject the exact `reason` value provided by the signal** instead of ignoring it and using another value.

> How about checking if `reason` is DOMException and AbortError, and returning the default error in that case (assuming that the user didn't give `reason` in that case)

In the case that a user provides a custom `DOMException` with the name `"AbortError"`, it seems like quite an arbitrary decision to ignore that value. Is there a way to positively identify Deno's default parameter instance? Something like this wouldn't even be enough AFAICT:

```ts
const r = signal.reason;

// Could still be a user-created value:
const looksLikeDenosDefaultParameter =
  r instanceof DOMException &&
  r.name === "AbortError" &&
  r.message === "The signal has been aborted";
```

Unless Deno brands all of its internally created values with a unique symbol or something like that, I'm not sure how we could discriminate it.

async/delay_test.ts
```diff
@@ -1,6 +1,17 @@
...
+  assertInstanceOf,
+  assertRejects,
+  assertStrictEquals,
+} from "../testing/asserts.ts";
+
+// https://dom.spec.whatwg.org/#interface-AbortSignal
+function assertIsDefaultAbortReason(reason: unknown) {
+  assertInstanceOf(reason, DOMException);
+  assertStrictEquals(reason.name, "AbortError");
+}
```
jsejcksn, line 14: I've updated the existing tests to stop checking for the custom `"message"` property string in [Deno's implementation](https://github.com/denoland/deno/blob/v1.34.3/ext/web/03_abort_signal.js#L83) of the default `DOMException`, and instead check that the value complies with the attributes described in the [spec](https://dom.spec.whatwg.org/#interface-AbortSignal).

kt3k: This makes sense to me

kt3k (dismissed):
LGTM

async/delay.ts
justinmchase: Yeah, in fact you should just take it if the reason is any kind of an Error object. And if its not, frankly you should not use the signal reason as the error message either but add it to a field. Such as this logic:

```ts
class AbortError extends DOMException {
  constructor(public readonly reason: unknown) {
    super("The signal has been aborted")
  }
}

const err = signal.reason instanceof Error
  ? signal.reason
  : new AbortError(signal.reason)
```

This will provide an Error object consistently, with a consistent message but will also provide the reason on a field should that be provided by the aborter.


This covers these cases:
```ts
controller.abort() // DOMException: The signal has been aborted
controller.abort('some other reason') // DOMException: The signal has been aborted { reason: 'some other reason' }
controller.abort(null) // DOMException: The signal has been aborted { reason: null }
```

The abort reason can be any value, so I would put it onto a field on the exception rather than try to make it the message.

async/delay_test.ts
justinmchase: I would leave that in there to be honest, don't change the exception message, add the reason onto a field on the exception instead.

async/delay.ts
jsejcksn: @justinmchase I definitely agree with you that it's best practice to only throw/reject instances of the [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) class (which includes [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) instances) — even if the language allows for other types of values.

However, the whole point of this PR is for this function to proxy the user's value verbatim (whether or not they choose to follow that practice). See the newly-added test, specifically [the step in lines 50–57](https://github.com/denoland/deno_std/pull/3479/files#diff-30bc1a4d7912bd1b9ba4c411ccb0589508f5099b3e9dea07035ce48902b38ee1R50-R57) — [`assertStrictEquals`](https://github.com/denoland/deno_std/blob/0.193.0/testing/asserts.ts#L253) uses [`Object.is`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is) for identity comparison, so moving the user value to a [property](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause) on a different value would cause that test to fail — and it would work against the goal of this PR, which is primarily to simplify exception-handling for users. By rejecting the reason argument provided to `abortSignal.abort(reason)`, the user can handle this value directly in their `catch` code, without having to learn about yet another abstraction in a function in `std`.

Here's a contrived example, but it illustrates what I described:

```ts
const reason = new Error("That took too long!");

delay(500).then(() => controller.abort(reason));

try {
  await delay(1000, { signal: controller.signal });
} catch (cause) {
  if (controller.signal.aborted && Object.is(cause, reason)) {
    console.log("I know that my code caused the exception because:", cause.message);
  }
}
```

If we create a new value (even if we attach the user's reason to it somehow), the user has to learn about that abstraction in order to handle the exception cause: which creates an unnecessary challenge.

async/delay_test.ts
kt3k: @justinmchase Do you mean adding `.reason` field to the existing `new DOMException("Delay was aborted.", "AbortError")` error object?

jsejcksn:
> cancel

@kt3k I noticed that you posted this after I merged in the latest changes from the `main` branch — can you clarify?

kt3k:
@jsejcksn I'm still wondering whether this version or @justinmchase's [suggestion](https://github.com/denoland/deno_std/pull/3479#discussion_r1256116881) is better.

This directly throws with `reason` given in `controller.abort(reason)`. The consumer of the error will be unaware of `delay` involved in-between in that case. I feel that could make debugging slightly more difficult when `delay` is used deep in dependencies. (I'm not completely sure though)

jsejcksn:
@kt3k I see — allow me to share more insight about why I created this PR:

I modelled the changes after the [`fetch` specification](https://fetch.spec.whatwg.org/), which describe the abort behavior to work in that way. Please see the spec links for full details, and I'll inline the most relevant parts here for ease of continuity:

https://fetch.spec.whatwg.org/#fetch-method

> ## 5.6. Fetch method
> 
> The `fetch(input, init)` method steps are:
> 
> 1. Let _p_ be a new promise.
> 
> …
> 
> 4. If _requestObject_’s signal is aborted, then:
>     1. Abort the `fetch()` call with _p_, _request_, null, and _requestObject_’s signal’s abort reason.
>     2. Return _p_.

https://fetch.spec.whatwg.org/#abort-fetch

> To **abort a `fetch()` call** with a _promise_, _request_, _responseObject_, and an _error_:
> 
> 1. Reject _promise_ with _error_.

Here's an example to demonstrate that behavior (which the changes in this PR mimic):

`fetch_abort_reason.mjs`:
```js
(async () => {
  const reason = new Error("Test finished");

  const ac = new AbortController();
  ac.abort(reason);

  try {
    await fetch("https://deno.land/", { signal: ac.signal });
    throw new Error("Unreachable");
  } catch (cause) {
    console.log({
      isUserReason: Object.is(cause, reason),
      cause,
    });
  }
})();

```

Deno implements the spec:

```
% deno --version
deno 1.35.1 (release, aarch64-apple-darwin)
v8 11.6.189.7
typescript 5.1.6

% deno run --allow-net fetch_abort_reason.mjs
{
  isUserReason: true,
  cause: Error: Test finished
    at file:///Users/deno/fetch_abort_reason.mjs:2:18
    at file:///Users/deno/fetch_abort_reason.mjs:16:3
}
```

Node.js implements the spec:

```
% node --version
v18.16.1

% node fetch_abort_reason.mjs
{
  isUserReason: true,
  cause: Error: Test finished
      at Object.fetch (node:internal/deps/undici/undici:11457:11)
      at async file:///Users/deno/fetch_abort_reason.mjs:8:5
}
```

Google Chrome appears to deviate from the spec:

> To reproduce: Copy the code from the module above and paste into the JS developer console and evaluate. I tested on Chrome version `114.0.5735.198 (Official Build) (arm64)`.

```
{
  isUserReason: false,
  cause: DOMException {
    code: 20,
    name: "AbortError",
    message: "Failed to execute 'fetch' on 'Window': The user aborted a request.",
    stack: "Error: Failed to execute 'fetch' on 'Window': The user aborted a request.\n    at <anonymous>:8:11\n    at <anonymous>:16:3"
  }
}
```

I hope you find this information to be supportive of the change. I think that it's less surprising to follow/mimic spec behavior rather than invent new behavior.

kt3k (approved):
@jsejcksn Thanks for elaborating. Now I'm convinced this design is reasonable. LGTM

kt3k:
> Google Chrome appears to deviate from the spec:

This sounds interesting. I checked other runtimes, and found that Firefox, Node.js, and Bun throw the abort reason directly (following the spec), and Safari does the similar to Chrome.

kt3k:
off-topic note:

wpt for testing fetch being aborted with abort reason
https://github.com/web-platform-tests/wpt/blob/fe4dd46252b60b7bcebcfab09e15475b91dcfb54/fetch/api/abort/general.any.js#L37-L45

chrome and safari seem actually failing with them https://wpt.fyi/results/fetch/api/abort/general.any.html?label=experimental&label=master&aligned

async/delay_test.ts
justinmchase: Yeah I suppose I did.

PR #3485, opened 2023-07-10
feat(fmt): enable `setColorEnabled` in browsers

halvardssm:
Allowed for programmatic override of enable color.

When Deno is not available in global this, and we want to use the colors library in e.g. browsers without shiming (dnt), it is not possible to view colors in the console, or set it to enabled. This pr changes this behavior so that it is possible to programmatically override the enable variable.

### DIFF
--- a/fmt/colors.ts
+++ b/fmt/colors.ts
@@ -76,10 +76,6 @@ let enabled = !noColor;
  * @param value
  */
 export function setColorEnabled(value: boolean) {
-  if (noColor) {
-    return;
-  }
-
   enabled = value;
 }
 
--- a/fmt/colors.ts
+++ b/fmt/colors.ts
@@ -76,6 +76,10 @@ let enabled = !noColor;
  * @param value
  */
 export function setColorEnabled(value: boolean) {
+  if (Deno?.noColor) {
+    return;
+  }
+
   enabled = value;
 }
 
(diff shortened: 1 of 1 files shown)


kt3k:
This makes sense to me. `Deno.noColor` is still respected (as the default value), but it can be overridable by `setColorEnabled(true)` call.

Does anyone have any concerns?

kt3k:
Changes for browsers make sense, but if the runtime is Deno, maybe we still want to keep that if block (to prevent accidentally enable the color in that case)

halvardssm:
That sounds like a good alternative. I will change the code accordingly.

halvardssm:
@kt3k Also, for this condition

```
const noColor = typeof Deno?.noColor === "boolean"
  ? Deno.noColor as boolean
  : true;
```

Wouldn't it be better to change it to 

```
const noColor = typeof Deno?.noColor === "boolean"
  ? Deno.noColor as boolean
  : false;
```

To have this as opt out rather than opt in?

kt3k:
I think that would make sense, but I'm not sure about the exact reasoning.

@nayeemrmn Do you remember why `noColor = true` by default in browsers? https://github.com/denoland/deno/pull/5847

nayeemrmn:
> Do you remember why `noColor = true` by default in browsers? [denoland/deno#5847](https://github.com/denoland/deno/pull/5847)

It was because browser consoles don't support ANSI sequences. Though I just checked and apparently Chromium does (now?)! But Firefox doesn't. I'm in favour of enabling it by default

halvardssm:
Updated according to our discussion, but let me know if there is anything else

kt3k (approved):
LGTM 👍

PR #3486, opened 2023-07-10
BREAKING(streams): deprecate readableStreamFromIterable

### LINKED ISSUE #3484: Deprecate `readableStreamFromIterable` in favor of `ReadableStream.from`
Since `ReadableStream.from` was introduced in https://github.com/denoland/deno/pull/19446, I think we can remove `readableStreamFromIterable` instead.

Note: It seems that there is no type definition for `ReadableStream.from` as of v1.35.0, but it seems to have been added later at https://github.com/denoland/deno/pull/19438.
https://github.com/denoland/deno/blame/1edc8693bfc4b75e3bb61fc8dc531bdf569b332d/ext/web/lib.deno_web.d.ts#L808

lino-levan:
Closes #3484


kt3k (approved):
LGTM. Thanks!

PR #3488, opened 2023-07-12
feat: add `std/text` with word-similarity helpers

jeff-hykin:
Simple argument helper, but could also be useful in frontend search components as well


lino-levan:
I'm in favor of adding this, but should we made split this out into a different file? Also, I feel like there is value is exporting `levenshteinDistanceBetween` (though I would probably rename it to `levenshteinDistance`).

jeff-hykin:
Putting it in a different file and renaming the function sounds good to me.

I also found a, probably more efficient, implementation of levenshteinDistance on deno.land. I could just copy it, but I wasn't sure about the license so I left my function in there.

flags/mod.ts
```diff
@@ -785,3 +785,120 @@ export function parse<
...
+ *
+ * @param {Object} options - The options for spell checking.
+ * @param {string} options.givenWord - The word to be checked for possible corrections.
+ * @param {string[]} options.possibleWords - An array of possible words to compare against.
+ * @param {boolean} [options.caseSensitive=false] - Flag indicating whether the spell check should be case sensitive. Default is false.
+ * @param {boolean} [options.autoThrow=false] - Flag for throwing automatically if the word is not a direct match
+ * @param {number} [options.suggestionLimit=Infinity] - Number of results to return
+ * @returns {string[]} An array of possible correct spellings for the given word.
+ */
+export function didYouMean(
```
kt3k, line 860: I think the name should be something like `guessWord`, `guessCandidate`, etc (No prior example of this kind of naming in std)

flags/mod.ts
```diff
@@ -785,3 +785,120 @@ export function parse<
...
+  {
+    givenWord,
+    possibleWords,
```
kt3k, line 863: We usually make mandatory arguments positional args, instead of putting them in option bag. ref. https://deno.land/manual@v1.35.1/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object

kt3k:
I agree with @lino-levan . I think this API should be in somewhere else. I also agree that `levenshteinDistance` is also worth a public API. (PHP seems having [levenshtein](https://www.php.net/manual/en/function.levenshtein.php) builtin function)

@lino-levan Do you have any suggestion about the place for these APIs? What if we create `std/text` or `std/string` module?

flags/mod.ts
```diff
@@ -785,3 +785,120 @@ export function parse<
 
   return argv as Args<TArgs, TDoubleDash>;
 }
+
+/**
+ * Calculates the Levenshtein distance between two strings.
+ *
+ * @param {string} str1 - The first string.
```
kt3k, line 792: Types in `@param` directive seem redundant as they are given in TypeScript. We usually omit them in our docs (For example, see https://github.com/denoland/deno_std/blob/6b1da3240ca5f9460df2571096d64566a7cb3c38/async/pool.ts#L31

lino-levan:
I'm of the opinion that this and #3440 ought to be in std/text. I think std/string would limit our future options.

jeff-hykin:
I'm actually glad you guys bring up `std/text` cause I was planning on asking for such a thing in a future PR.

And just as a catch-all I don't really care about any small renaming, argument-change, file location kind of stuff. I'll probably implement all the suggestions tomorrow (including having it in a different file)

jeff-hykin:
alright I believe I implemented all the feedback

flags/mod.ts
```diff
@@ -784,4 +784,4 @@ export function parse<
   }
 
   return argv as Args<TArgs, TDoubleDash>;
-}
+}
```
lino-levan, line 787: Looks like unrelated change, could you revert?

flags/test.ts
```diff
@@ -1906,3 +1906,41 @@ Deno.test("typesOfParseOptionsGenericDefaults", function () {
     >
   >(true);
 });
+
+Deno.test("basicDidYouMean", function () {
```
lino-levan, line 1910: could we move this to a seperate testing file in `std/text`?

text/mod.ts
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * @param {Object} options - The options for spell checking.
+ * @param {string} options.guessWord - The word to be checked for possible corrections.
+ * @param {string[]} options.possibleWords - An array of possible words to compare against.
+ * @param {boolean} [options.caseSensitive=false] - Flag indicating whether the spell check should be case sensitive. Default is false.
+ * @param {boolean} [options.autoThrow=false] - Flag for throwing automatically if the word is not a direct match
+ * @param {number} [options.suggestionLimit=Infinity] - Number of results to return
+ * @returns {string[]} An array of possible correct spellings for the given word.
+ */
+export function didYouMean(
```
lino-levan, line 72: discussion: Should we follow the single-function file pattern that we do elsewhere in std here? Does it make any sense to?

kt3k: I think we should follow single-file export pattern. Some who only wants `levenshteinDistance` would want to import from `std/text/levenshtein_distance.ts`

kt3k:
Could you also address `didYouMean` -> `guessWord` rename, and jsdoc typing clean up?

flags/mod.ts
lino-levan: I'm a fan of `guessWord`.

lino-levan:
I have to agree with @timreichen. `autoThrow` is a weird option, both in naming and in functionality. +1 for removing it.

jeff-hykin:
@lino-levan

> Could you also address `didYouMean` -> `guessWord` rename, and jsdoc typing clean up?

Sorry I misinterpreted your original request (I renamed `givenWord` to `guessWord`).

I do want to discuss this name change though. Based on evidence, including [cliffy for deno](https://github.com/c4spar/deno-cliffy/blob/6f3ba2959f4d78973ec2a51bc8654df82556629d/flags/_utils.ts#L46), [npm library names](https://www.npmjs.com/package/didyoumean), [pip module names](https://pypi.org/project/didyoumean3/), [stack overflow questions](https://stackoverflow.com/questions/41424/how-do-you-implement-a-did-you-mean), etc people generally call this kind of behavior/feature "did you mean".

I'd say there's already a bit of visibilty problem with deno-std, like how 3rd party libraries show up first for features like `deep merge`. If that function, for example, was called `recursiveMerge` in deno std, I think it just makes it all that much harder to find even if the name is still technically accurate.

So for the sake of convention/searchability, I think `didYouMean` is the only name I have strong feelings about.

lino-levan:
> I'd say there's already a bit of visibilty problem with deno-std, like how 3rd party libraries show up first when searching for deep merge or debounce

I strongly believe that `/x` should show standard library submodules at the top of each search for things like "deep merge" and "debounce". This would solve a lot of issues in my opinon.

jeff-hykin:
@lino-levan  for the JS doc code, I added the import lines to the examples and fixed up the examples. Was there anything else to do for the jsdoc comments?

jeff-hykin:
> I strongly believe that `/x` should show standard library submodules at the top of each search for things like "deep merge" and "debounce". This would solve a lot of issues in my opinon.

agreed, my point being if they seach "did you mean" but in deno std its called "guessWord" its not going to show up even with that^ feature implemented

lino-levan:
JSDoc comments look good. Is there a reason you still have tests in `flags/test.ts`?

jeff-hykin:
> JSDoc comments look good. Is there a reason you still have tests in `flags/test.ts`?

Yes, the reason is because I forgot twice in a row to remove them haha

text/did_you_mean.ts
```diff
@@ -0,0 +1,78 @@
+import { levenshteinDistance } from "./levenshtein_distance.ts";
+import { assert } from "../testing/asserts.ts";
```
kt3k, line 2: `testing/asserts.ts` is deprecated. Let's import from `../assert/assert.ts`

kt3k:
@jeff-hykin Thanks for updating. Looks much better to me now.

> `text/word_similarity.ts`

I'd suggest we should name this as `_utils.ts` and not expose items in it from `mod.ts`. Let's revisit when users ask for them to be exposed.

> So for the sake of convention/searchability, I think didYouMean is the only name I have strong feelings about.

Ok. Now I'm convinced `didYouMean` is fine if there are many precedent examples, and I'd suggest we should avoid `assertDidYouMean`, which sounds strange to me..

kt3k:
Also can you fix lint errors? (You can check it with `deno lint` command in the repo root)

lino-levan:
Could you fix formatting?

lino-levan:
Also, after doing some soul searching, I think `didYouMean` should potentially be in flags? I feel like flags should really be renamed to `cli` or something.

text/did_you_mean.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    const suggestions: string[] = wordSimilaritySort(givenWord, possibleWords, {
+      caseSensitive: caseSensitiveDistance,
+    }).slice(0, suggestionLimit);
+    if (suggestionLimit == 1) {
+      throw new DidYouMeanError(
+        `For ${JSON.stringify(givenWord)}, did you mean ${
+          JSON.stringify(suggestions[0])
+        }?`,
```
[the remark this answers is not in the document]
jeff-hykin: is there a deno-test way to show code coverage? (or lack thereof)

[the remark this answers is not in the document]
jeff-hykin: haha actually this was tested the test was just wrong 😅

jeff-hykin:
> flags should really be renamed to `cli` or something.

I agree with this. It also would open up a slew of simple CLI tools I would like to add such as
- terminalColorSupport
- askFor.yesNo
- askFor.int
- askFor.float
- askFor.oneOf
- askFor.filePath

jeff-hykin:
want me to go ahead an do that change to create std/cli? @lino-levan

jeff-hykin:
actually I kinda just want to get this merged in and then debate creating the cli/ folder afterwords. This was supposed to be a really small addtion to std (supposed to be just `didYouMean` but then I abstracted out the distance function). I've got a bunch of future PR's planned, but I kind of wanted to get this one out of the way to get familiar with the process.

lino-levan:
I didn't realize this was your first contribution! Wow!

> I've got a bunch of future PR's planned, but I kind of wanted to get this one out of the way to get familiar with the process.

Feel free top open PRs! I have a feeling that smaller ones would be easier to get merged faster.

jeff-hykin:
Thanks for the changes @kt3k. I got halfway through them yesterday after updating from denoland:main but ran out of time. 

Could you explain the `_util.ts` vs `mod.ts`, like why are some things in mod but others are in _util?

text/word_similarity.ts
```diff
@@ -0,0 +1,131 @@
...
+export function closest(
+  givenWord: string,
+  possibleWords: string[],
+  options?: {
+    caseSensitive?: boolean;
+  },
+): string {
+  assert(
+    possibleWords.length > 0,
+    `When using closest(), the possibleWords array must contain at least one word`,
```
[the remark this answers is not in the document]
jeff-hykin: done

text/word_similarity.ts
```diff
@@ -0,0 +1,131 @@
...
+  );
+  const { caseSensitive } = { ...options };
+
+  if (!caseSensitive) {
+    givenWord = givenWord.toLowerCase();
+  }
```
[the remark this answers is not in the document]
jeff-hykin: done

text/_util.ts
```diff
@@ -0,0 +1,129 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { levenshteinDistance } from "./levenshtein_distance.ts";
+import { assert } from "../assert/mod.ts";
```
lino-levan, line 3: @kt3k confused by this one, why is this `mod.ts` and not `assert.ts`?

text/_util.ts
```diff
@@ -0,0 +1,129 @@
...
+ * @param possibleWords - The string-array that will be sorted
+ * @param options.caseSensitive - Flag indicating whether the distance should include case. Default is false.
+ * @returns A sorted copy of possibleWords
+ * @note
+ * the ordering of words may change with version-updates
+ * e.g. word-distance metric may change (improve)
+ * use a named-distance (e.g. levenshteinDistance) to
+ * guarantee a particular ordering
+ */
+export function closest(
```
lino-levan, line 31: I feel like this is missing a word here. `closestWord`? Not sure.

text/_util.ts
```diff
@@ -0,0 +1,129 @@
...
+ * // words most-similar to "hep" will be at the front
+ * words.sort(similarityCompare("hep"));
+ * ```
+ * @note
+ * the ordering of words may change with version-updates
+ * e.g. word-distance metric may change (improve)
+ * use a named-distance (e.g. levenshteinDistance) to
+ * guarantee a particular ordering
+ */
+export function similarityCompare(
```
lino-levan, line 80: `compareSimilarity`?

text/_util.ts
jeff-hykin: Not a strong preference, but it was more generic; I could do a find-and-replace-all from the old location to the new location, but only if I used "mod.ts" (so that imports like `assertEquals, assertThrows` kept working)

text/_util.ts
jeff-hykin: could be closestString since spaces are allowed

text/_util.ts
lino-levan: My hope would be that in the std we try to minimize the module graph, so not sure I support this move.

text/_util.ts
lino-levan: `closestString` makes sense to me. I don't feel strongly either way, just brought this up for bikeshedding.

text/_util.ts
```diff
@@ -0,0 +1,129 @@
...
+ * // force case sensitive
+ * wordSimilaritySort("hep", possibleWords, { caseSensitive: true });
+ * ```
+ *
+ * @param givenWord - The string to measure distance against
+ * @param possibleWords - The string-array that will be sorted
+ * @param options.caseSensitive - Flag indicating whether the distance should include case. Default is false.
+ * @returns {string[]} A sorted copy of possibleWords
+ */
+export function wordSimilaritySort(
```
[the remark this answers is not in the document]
jeff-hykin: I'll leave that up to you guys, its the difference between:

```
let suggestions = [...possibleWords].sort(similarityCompare("hep", { caseSensitive })).join(", ")
```
and 
```
let suggestions = wordSimilaritySort("hep", possibleWords).join(", ")
```

text/_util.ts
jeff-hykin: okay I'll change it

text/_util.ts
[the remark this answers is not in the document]
lino-levan: I'm not understanding why one would need to do `[...possibleWords]`?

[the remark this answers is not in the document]
jeff-hykin: cause otherwise it mutates `possibleWords`, which might be sorted in a particular order already (for example an order that is used when the help command is invoked)

text/_util.ts
jeff-hykin: done

text/_util.ts
jeff-hykin: done

text/_util.ts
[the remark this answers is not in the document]
lino-levan: Ah sorry, I should have clarified. Why not use `.toSorted()`? It would just be
```ts
let suggestions = possibleWords.toSorted(similarityCompare("hep", { caseSensitive })).join(", ")
```
which is a little more readable imo.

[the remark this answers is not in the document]
jeff-hykin: I'd prefer older compatibility (e.g. avoiding bleeding-edge ES2023 features), and plus toSorted is literally only one character shorter than the more explicit version.

kt3k:
@jeff-hykin 
> Could you explain the `_util.ts` vs `mod.ts`, like why are some things in mod but others are in `_util`?

APIs exported from `mod.ts` are public APIs of the module, and ones exported from `_util.ts` are not (i.e. They are internal utilities). Once we expose the APIs as public APIs, we can't change them easily in the future. I think we are ready to expose `levenshteinDistance` and `didYouMean`, but others (`closestString`, `compareSimilarity`, `wordSimilaritySort`) don't look ready to be exposed

kt3k:
> I do want to discuss this name change though (`didYouMean` -> `guessWord`). Based on evidence, including [cliffy for deno](https://github.com/c4spar/deno-cliffy/blob/6f3ba2959f4d78973ec2a51bc8654df82556629d/flags/_utils.ts#L46), [npm library names](https://www.npmjs.com/package/didyoumean), [pip module names](https://pypi.org/project/didyoumean3/), [stack overflow questions](https://stackoverflow.com/questions/41424/how-do-you-implement-a-did-you-mean), etc I believe people generally call this kind of behavior/feature "did you mean".

I revisited these links, but what the above pip module does is different from what this PR does (They seem correcting typo based on the dictionary). Stackoverflow question refers to `Did you mean` feature of google search, which is also different from what this PR does. cliffy has `didYouMean` util, but that is not a public API. `npm:didyoumean` is probably the only one example similar to this, but the behavior is very different (they don't throw automatically).

I now feel unsure this design of `didYouMean` is appropriate for deno_std

(BTW I think `levenshteinDistance` definitely deserve std API as there's [a prior example](https://www.php.net/manual/en/function.levenshtein.php) in php)

jeff-hykin:
This is starting to feel a bit like review purgatory 😅

I'm just simply trying to contribute:
   - 2 tools people can find by searching "did you mean"
   - 1. an assert that validates things like enum arguments (not everyone uses TS) with a helpful runtime "That's not one of the options, did you mean X" error message when it's an invalid option
   - 2. a pure function that returns the most similar word to a give dictionary (closestString/Spell check/NPM didYouMean/Ruby didYouMean)

> cliffy has didYouMean util, but that is not a public API

Yes that's the point; I can't use his implementation and the cliffy author, myself, and others have all had to reimplement the same tool because there isn't one available in the std. Even if it was public in cliffy, it would be weird to import all of cliffy just to validate an enum argument in a function that's going to be used in a web interface. "I need to guarantee that this string belongs to a predefined list of strings, and throw a helpful error if that's not the case" is a common need, and for every duplicate implementation, there's probably 10 more devs that decided "I'm just going to throw an unhelpful error message, after all, I'm writing a tiny JS function and it doesn't make sense to create 2x the work implementing name similarity just to improve the error"

> They seem correcting typo based on the dictionary

This is why it feels like purgatory; the original PR did correct typos based on a list of strings but based on feedback we renamed it to "closestString" and then the _utils change made it private. It feels like we're going around in circles while also blocking generic and commonly needed functionality.


I agree naming is important, and I feel like things that throw should be an `assertX` or `requireX`. So, would it be sufficient to:
- rename+change the current didYouMean to assertOneOf, make it handle more generic types gracefully, and have it detect the case of input=String, and options=Array[String] in which case it provides a "did you mean" error.
- Then rename or alias closestString to didYouMean

kt3k:
Sorry for pressuring!

> rename+change the current didYouMean to assertOneOf, make it handle more generic types gracefully, and have it detect the case of input=String, and options=Array[String] in which case it provides a "did you mean" error.

I feel `assertOneOf` throwing "did you mean" error is too arbitrary design, and feels confusing.

I looked through the APIs again. Now I feel 3 APIs in `_util.ts` are more appropriate for deno_std in API naming and functionality (sorry for changing my mind again!), and if we have these 3 APIs, the feature like `didYouMean` can be easily implemented using `closestString` or `wordSimilaritySort`.

I'd like to suggest the below:
- Make `levenshteinDistance`, `closestString`, `compareSimilarity`, and `wordSimilaritySort` public APIs.
- In jsdoc example of `wordSimilaritySort`, we describe how `didYouMean` can be implemented using it.

If we have `didYouMean` function in the jsdoc example, that'll be indexed in our doc search in Deno homepage, and the user should have access to it easily. What do you think?

iuioiua:
Hi @jeff-hykin, I spoke with @kt3k. Are you still happy to pursue this PR with the above comment's suggestions?

jeff-hykin:
Thanks for the ping, and sorry for leaving this PR hanging. I am still interested in it.

Yeah, I can make the changes and get this branch back up to date with main.

> If we have didYouMean function in the jsdoc example, that'll be indexed in our doc search in Deno homepage, and the user should have access to it easily

This is good, thanks for addressing one of my core concerns.

> assertOneOf throwing "did you mean" error is too arbitrary design, and feels confusing.

I'm onboard with this sentiment; purity/cleanliness is a top priority for std. Even small things, like `--_` causing problems for the std flag parser bother me.


That said, I still feel there is a gap here. Maybe I we can merge those four to get this PR closed and open up a new PR to continue this discussion. For now I'll continue it here.

The "let the user implement it" is exactly what I'm trying to avoid. I mean, if the expression is so easy that the end user would just inline it; ok thats fine. But when they need to define a helper function, a function that isn't customized to their specific project in any way, then I think theres a problem.

Me, my coworker, and my student shouldn't each need our own stdlib. But if deno std doesn't implement stuff like assertOneOf, then we will (either that or we need to copy and paste our implementation from project to project which also seems terrible).

iuioiua:
I've created #3752 to continue the discussion of `didYouMean()` in `std/text`. @jeff-hykin, can you please run `deno fmt` and make any final adjustments, then let us know once it's ready?

jeff-hykin:
> If we have didYouMean function in the jsdoc example

Should I put all 80 lines of the didYouMean function in a jsdoc example? Or should we just merge as-is and discuss this on #3752 ? @kt3k

jeff-hykin:
Thanks @kt3k !

PR #3491, opened 2023-07-12
chore: simplify getSetCookies

### LINKED ISSUE #3487: suggestion: deprecate `getSetCookies()`
Deno now supports the [`getSetCookie()` method](https://developer.mozilla.org/en-US/docs/Web/API/Headers/getSetCookie) for the `Headers` object. `getSetCookies()` should either be deprecated or use this new method.

lino-levan:
Closes #3487.

I think simplifying it makes more sense rather than deprecating it given that this is a nice utility anyways.


lino-levan:
It seems to claim that this isn't browser compatible. It definitely is? Are the TS type definitions out of date?

kt3k:
`getSetCookie` was added recently in TS https://github.com/microsoft/TypeScript/pull/54725 ( https://github.com/microsoft/TypeScript-DOM-lib-generator/pull/1553 ). It looks like it will be available in TS 5.2 (Deno uses TS 5.1.6)

kt3k:
Note: TS 5.2 iteration plan https://github.com/microsoft/TypeScript/issues/54298

We probably need to wait another month.

I think landing this with `@ts-ignore` and `TODO` comment might be an option.

lino-levan:
I'm game to land this with a ts-ignore. Let me update the PR.

kt3k (approved):
LGTM

Issue #1779, opened 2022-01-02
Consider adding "testing/expect.ts" module


bartlomieju:
Jest is arguably the most popular testing framework for Node ecosystem at the moment. It uses [`expect`](https://jestjs.io/docs/expect) style assertions, which seems to [be very popular](https://www.npmjs.com/package/expect) way of exercising tests too. 

I was prompted to consider it by a [recent issue in `vitest`](https://github.com/vitest-dev/vitest/issues/43) (which seems to be somewhat compatible with Jest.

To be clear: I prefer to write tests using `testing/asserts.ts`, but we shouldn't limit ourselves to only a single style, if the other is also very popular (> 8mm weekly downloads) then we should consider adding it too.

getspooky:
Good suggestion, but I believe it would be better to create a separate library rather than include it in deno std.

bartlomieju:
@getspooky is there a reason why you would prefer it as a third party lib? The whole idea is to have it in `deno_std` so it's easy to discover and could be used reliably.

kt3k:
I like the idea. This makes the migration of existing libraries from node to deno very much easier.

getspooky:
@bartlomieju can you please list of functions will be supported ?

standvpmnt:
I might not be understanding the full context here but, could someone instead just add the expect package from jest into their project instead, by adding something like [expect package from skypack](https://www.skypack.dev/view/expect) ?

bartlomieju:
> I might not be understanding the full context here but, could someone instead just add the expect package from jest into their project instead, by adding something like [expect package from skypack](https://www.skypack.dev/view/expect) ?

Yes, that would be equivalent (if the library indeed works with Deno), however we would like to make sure that this library is guaranteed to work with Deno and tested as part of our CI pipeline. I haven't tested this URL so can't say if it would just work.

> @bartlomieju could you please specify list of functions will be supported ?

Ideally we would support all APIs listed in the link from the first comment, but this list is a good start.

standvpmnt:
Yes, simply importing it doesn't work in deno I suppose because of some module and dependency resolution.

bartlomieju:
> Yes, simply importing it doesn't work in deno I suppose because of some module and dependency resolution.

Right, so ideally we'd have first-class support by having this module in `deno_std`, so users can safely rely on it for crucial part of project which is test code.

standvpmnt:
[...]

PR #3494, opened 2023-07-15
[DRAFT] `expect` function for testing

Sorikairox:
Draft to get feedback on implementing them as wrappers for `assert` functions.

Tests for `BeEqual` matcher are exactly the same as `assertEquals`, so we might be able to refactor them into one test suit calling both instead of having duplication.

Base code for expect from https://github.com/allain/expect/tree/master which is ISC License.

Modified it to have Matchers get a context and throw as they use `assert` function. Also fixed some typing.

Related to issue #1779


expect/expect.ts
```diff
@@ -0,0 +1,129 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+//ISC License
```
kt3k, line 3: I think we can use some short form of license header like:

```
Copyright 2019 Allain Lalonde. All rights reserved. ISC License.
```

Sorikairox: Done !

kt3k:
I think `matcher/` dir and `types.ts` should be named `_matcher/` and `_types.ts` respectively to minimize the publicly exposed APIs (I suppose the only meaningful API for users to import is `expect`).

bartlomieju:
Pinging @marvinhagemeister since he brought up a need for this API recently :)

marvinhagemeister:
Nothing to say other than I really like this and can't wait to have this natively in deno 👍

bartlomieju:
> @bartlomieju @marvinhagemeister Do you have any opinion/preference about the export path of this module? This PR suggests `std/expect/expect.ts`. I think that would make sense. But The original issue suggested `std/testing/expect.ts` that also seems reasonable to me.

I would err on `std/exect/mod.ts` (similar to how `assert` was made top-level). `std/testing/expect.ts` works too 👍

aapoalas:
One argument for why assert was made top level was that it has valid runtime uses as well. Expect kind of doesn't. I'd put it under testing.

Sorikairox:
FWIW, I was going to put it in `std/testing/expect.ts` at first, and then I saw `assert` at the top level and moved everything. So I think `testing/expect.ts` fit better.

marvinhagemeister:
@Sorikairox I'd really like us to re-open this PR. I've been using `npm:expect` in one project recently and it's super slow. It does a lot of allocations and collects error stack traces upfront and other things. It's super slow even if you do just `expect(1).toEqual(1)` and none of that is needed.

| assertion library | time |
|---|--:|
| `npm:expect` | 786ms |
| assert | 144ms |

Made a super tiny wrapper around `assert` to mirror the `expect` API like in this PR and the speedup is too good to ignore. I want us to ship `expect` ourselves. I really want to see this land.

bartlomieju:
I would also be in favor of landing as is and adding additional APIs in follow up PRs.

Sorikairox:
Hey guys. My apologies, life happened and I did not manage to make time to continue on this.

Due to rebasing and force pushing, looks like I cannot re-open this one. Opening https://github.com/denoland/deno_std/pull/3627

PR #3496, opened 2023-07-21
feat(assert): add inequality asserts

ratson:
Add inequality asserts

- `assertGreaterOrEqual`
- `assertGreater`
- `assertLessOrEqual`
- `assertLess`


kt3k:
Note: Some testing libs seems having these kinds of assertions (python's unittest, PHPUnit, XCTest, Jest, etc), while some others (ruby's minitest, kotlin.test, Node's assert, JUnit, etc) don't.

- python unittest https://docs.python.org/3/library/unittest.html
  - assertLess
  - assertLessEqual
  - assertGreater
  - assertGreaterEqual
- PHPUnit https://docs.phpunit.de/en/10.0/assertions.html
  - assertLessThan
  - assertLessThanOrEqual
  - assertGreaterThan
  - assertGreaterThanOrEqual
- XCTest https://developer.apple.com/documentation/xctest/comparable_value_assertions
  - XCAssertLessThan
  - XCAssertLessThanOrEqual
  - XCAssertGreaterThan
  - XCAssertGreaterThanOrEqual
- Jest https://jestjs.io/docs/expect
  - toBeLessThan
  - toBeLessThanOrEqual
  - toBeGreaterThan
  - toBeGreaterThanOrEqual

assert/assert_greater.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Make an assertion that `actual` is greater than `expected`.
+ * If not then throw.
+ */
+export function assertGreater<T>(actual: T, expected: T, msg?: string) {
+  if (actual > expected) return;
+
+  const actualString = format(actual);
+  const expectedString = format(expected);
+  throw new AssertionError(msg ?? `Expect ${actualString} > ${expectedString}`);
```
kt3k, line 14: I think we should show both default and given message if `msg` is passed. Maybe let's align to what `assertEquals` does for the given `msg` input. ref. https://github.com/denoland/deno_std/blob/b5c05eea16b542ea6951ef59bfb70a9de0aaca10/assert/assert_equals.ts#L29-L30

kt3k:
It looks reasonable to include these to `std/assert` as there seem several prior examples.

assert/assert_greater.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { format } from "./_format.ts";
+import { AssertionError } from "./assertion_error.ts";
+
+/**
+ * Make an assertion that `actual` is greater than `expected`.
+ * If not then throw.
+ */
+export function assertGreater<T>(actual: T, expected: T, msg?: string) {
```
lino-levan, line 9: Does using a generic here (with no `extends` or similar) make sense? There are a limited amount of JS primitives that can be compared using `>`/`<` operators. I think we should consider enumerating them.

kt3k: I think `T` without `extends` is fine because comparing behavior can be customized by `valueOf` method of user defined object, like:

```js
console.log({ valueOf() { return 1 } } < { valueOf() { return 2 } }); // => true
console.log({ valueOf() { return 1 } } < { valueOf() { return 0 } }); // => false
```

kt3k:
The idea and implementation look good to me. Thanks for the suggestion! @ratson 

I think there's still a room to bikeshed the naming. I think there are 3 ways to name these (pick greater-equal as example):
- `assertGreaterThanOrEqual` (similar to PHPUnit, XCTest, Jest)
- `assertGreaterOrEqual` (similar to NUnit (C#), this PR)
- `assertGreaterEqual` (similar to Python)

Any opinions?

kt3k:
The core team looked in favor of `assertGreaterOrEqual`

kt3k (approved):

PR #3497, opened 2023-07-22
Fix typo in std/testing/asserts.ts deprecation message

rotu:
`std/testing/asserts.ts` was moved to `std/assert` not `std/asserts`.


kt3k (approved):
LGTM. Thanks!

PR #3499, opened 2023-07-23
fix(http): fix handling of string port number

kt3k:
Addressed https://github.com/denoland/deno/issues/10064#issuecomment-1637427260


bartlomieju (approved):
LGTM, thank you!

PR #3505, opened 2023-07-27
test(fs/walk): clean up socket file after testing

kt3k:
A unix socket file is not cleaned up after testing, and it causes test failure after the 2nd run of the same test case (The issue seems introduced in #3446)

This PR adds the clean up of it, and fixes the issue.


bartlomieju (approved):
LGTM!

PR #3507, opened 2023-07-28
BREAKING(testing/snapshot): fix regression of serialization of long strings

### LINKED ISSUE #3404: Snapshots break after Deno 1.33.2
**Describe the bug**
Deno 1.33.2 changed the way to serialize multiline strings and, instead of a single string, it returns a concatenation of several strings, one per line. This makes all tests to fail: 

imaxe

**Steps to Reproduce**

1. Create a snapshot in Deno < 1.33.2
2. Upgrade to Deno >= 1.33.2
3. Run the tests

**Expected behavior**

The solution is update the snapshots with `-- --update` (obviously), but I think the previous format was much clearer to see the difference between two snapshots.

**Environment**

- OS: All
- deno version: >= 1.33.2
- std version: 0.186.0

kt3k:
Now the snapshots of long strings are split into additions of smaller string, and those are very hard to read as pointed in #3404. This behavior is unintentionally introduced by the merge of `Deno.inspect` and `util.inspect` (Node compat). ref: https://github.com/denoland/deno/pull/18691

 This PR uses `breakLength: Infinity` option of `Deno.inspect` (which was added in https://github.com/denoland/deno/pull/19337), and restores the original behavior.

fixes #3404 

cc @oscarotero

### DIFF
--- a/testing/__snapshots__/snapshot_test.ts.snap
+++ b/testing/__snapshots__/snapshot_test.ts.snap
@@ -100,32 +100,19 @@ snapshot[`Snapshot Test - Adverse String \\ \` \${} 1`] = `"\\\\ \` \${}"`;
 
 snapshot[`Snapshot Test - Multi-Line Strings > string 1`] = `
 "
-" +
-  "<html>
-" +
-  "  <head>
-" +
-  "    <title>Snapshot Test - Multi-Line Strings</title>
-" +
-  "  </head>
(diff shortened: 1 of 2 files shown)


crowlKats (approved):
LGTM

oscarotero:
Oh, thanks a lot for this!!

PR #3516, opened 2023-08-04
Draft: dotenv: don't crash when permission is not granted to an env file

alexgleason:
I'm having a problem with using `--allow-read=.env`

It wants to read `.env.defaults` and `.env.example`, which I don't care about.

I think if you don't grant permission to those files, the application should not crash. It should continue on without them. For now, the only way around it is with `--allow-read=.env,.env.defaults,.env.example` (too verbose), or to call `loadSync({ export: true, examplePath: null, defaultsPath: null })` (also too verbose)

So I have made it not crash when the files are denied permission.

But there's another problem. `Deno.errors.PermissionDenied` can occur both for Deno permissions AND for OS file permissions.

That seems like a flaw in Deno itself. You should be able to tell whether `PermissionDenied` is due to OS permissions or Deno permissions.

The only way around that is a verbose check of the file against the permissions API.

It seems wrong. I wouldn't merge this code as-is. But I'm wondering if someone has ideas about what to do.

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -315,6 +315,7 @@ function parseFileSync(
     return parse(Deno.readTextFileSync(filepath), restrictEnvAccessTo);
   } catch (e) {
     if (e instanceof Deno.errors.NotFound) return {};
+    if (e instanceof Deno.errors.PermissionDenied && Deno.permissions.querySync({ name: "read", path: filepath }).state !== "granted") return {};
     throw e;
   }
 }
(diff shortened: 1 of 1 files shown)


kt3k:
> I'm having a problem with using --allow-read=.env
> It wants to read .env.defaults and .env.example, which I don't care about.

I understand this motivation, but I don't think it's a good idea to change the behavior based on given permission.

Maybe we can create an alternative endpoint (like `dotenv/load_simple.ts` (or whatever name)) that only loads `.env` to explicitly opt out from `.env.defaults` and `.env.example` behaviors. What do you think?

iuioiua:
Closing this as stale. In `parse()`'s current state, you can set the `examplePath` and `defaultPath` options to `null` to prevent this behavior.

PR #3524, opened 2023-08-08
fix(http/cookie_map): Add maxAge to set/delete options

### LINKED ISSUE #3523: `CookieMapSetDeleteOptions` is missing `maxAge`
**Describe the bug**

The `CookieMap` class supports setting cookies with a `maxAge`. The `Cookie` class [correctly handles](https://deno.land/std@0.197.0/http/cookie_map.ts?source=#L266) this option (setting the right `expires` attribute), but the [options interface](https://deno.land/std@0.197.0/http/cookie_map.ts?source=#L104) doesn't allow passing `maxAge`. I assume this was accidentally excluded since there's no discussion of leaving it out in #2502.

**Steps to Reproduce**

```ts
let response = new Headers();
let cookies = new CookieMap(new Headers(), { response });

cookies.set("foo", "bar", {
  maxAge: 60 * 60 * 24,
});

console.log(response.get("set-cookie"));
// foo=bar; path=/; expires=Tue, 08 Aug 2023 11:17:06 GMT; httponly
```

This works at runtime but doesn't typecheck:

```
Argument of type '{ maxAge: number; }' is not assignable to parameter of type 'CookieMapSetDeleteOptions'.
  Object literal may only specify known properties, and 'maxAge' does not exist in type 'CookieMapSetDeleteOptions'
```

**Expected behavior**

The code should typecheck when working options are passed.

**Environment**

- OS: macOS 13.4
- deno version: 1.36.0
- std version: 0.197.0

oliverjam:
Adds `maxAge` to the `CookieMapSetDeleteOptions` and `SecureCookieMapSetDeleteOptions`, so the interface matches the implementation. Closes #3523.

I added a test to make sure `maxAge` keeps working, since there wasn't one. The option gets converted to `expires`, so I needed to mock the time with `FakeTime`.

### DIFF
--- a/http/cookie_map.ts
+++ b/http/cookie_map.ts
@@ -106,6 +106,8 @@ export interface CookieMapSetDeleteOptions {
   domain?: string;
   /** When the cookie expires. */
   expires?: Date;
+  /** Number of seconds until the cookie expires */
+  maxAge?: number;
   /** A flag that indicates if the cookie is valid over HTTP only. */
   httpOnly?: boolean;
   /** Do not error when signing and validating cookies over an insecure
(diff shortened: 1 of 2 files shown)


lino-levan (approved):
LGTM. Nice catch! Thanks for the test.

kt3k (approved):
LGTM. Thanks!

PR #3527, opened 2023-08-08
feat(std): Add `std/url` module.

### LINKED ISSUE #3515: Add std/url with URL utilities
Should have similar functionality to path: parent, basename, ext

but operate on JS native URL objects

tr1ckydev:
Utilities for working with URL paths (#3515).

closes #3515


url/dirname_test.ts
```diff
@@ -0,0 +1,26 @@
...
+  ],
+  [
+    new URL("https://deno.land/std/assert/mod.ts"),
+    new URL("https://deno.land/std/assert"),
+  ],
+  [
+    new URL("https://deno.land/std/assert/mod.ts?foo=bar"),
+    new URL("https://deno.land/std/assert"),
+  ],
+  [new URL("https://deno.land///"), new URL("https://deno.land")],
```
lino-levan, line 19: You should probably include a test with a hash in it

url/extname_test.ts
```diff
@@ -0,0 +1,18 @@
...
+
+import { assertEquals } from "../assert/mod.ts";
+import * as url from "./mod.ts";
+
+const TESTSUITE = [
+  ["https://deno.land/std/assert/mod.ts", ".ts"],
+  [new URL("https://deno.land/std/assert/mod.ts"), ".ts"],
+  [new URL("https://deno.land/std/assert/mod.ts?foo=bar"), ".ts"],
+  [new URL("https://deno.land/std/assert/mod."), "."],
+  [new URL("https://deno.land/std/assert/mod"), ""],
```
lino-levan, line 11: Again, hash test would be nice.

url/join_test.ts
```diff
@@ -0,0 +1,26 @@
...
+    new URL("https://deno.land/std/assert/mod.ts"),
+  ],
+  [["https://deno.land/"], new URL("https://deno.land/")],
+];
+
+Deno.test("join", function () {
+  for (const [[test_url, ...paths], expected] of TESTSUITE) {
+    assertEquals(url.join(test_url, ...paths), expected);
+  }
+});
```
lino-levan, line 26: Now that I'm looking at the tests, I'm not sure if join even makes sense.

url/normalize.ts
```diff
@@ -0,0 +1,14 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+import { posixNormalize } from "../path/_normalize.ts";
+
+/**
+ * Normalize the `url`, resolving mutiple `'/'`s into `'//'` after protocol and remaining into `'/'`.
```
lino-levan, line 7: I'd expect normalize to normalize away `..` -- Is this the case?

url/dirname_test.ts
tr1ckydev: Added now.

url/extname_test.ts
tr1ckydev: Added.

url/join_test.ts
tr1ckydev: Why is it so?

lino-levan: I guess it handles edgecases around double `/` so maybe it's still useful. Not entirely sure what the utility is. Thoughts @marvinhagemeister?

url/normalize.ts
tr1ckydev: Can you provide me with a test case?

lino-levan: Something like
```ts
  [
    "https://deno.land/std/assert/../async/retry.ts",
    new URL("https://deno.land/std/async/retry.ts"),
  ],
  ```
  I'd also probably test that `https://deno.land/..` normalizes to `https://deno.land/`

Comment (tr1ckydev) [reply url/normalize.ts:7]: Both the cases pass the test.

Comment (tr1ckydev) [reply url/normalize.ts:7]: Added the cases.

Comment (lino-levan) [reply url/normalize.ts:7]: Could you add that it handles this to the JSDoc?

Comment (tr1ckydev) [reply url/normalize.ts]: Added.

**Comment by tr1ckydev** [comment]
Should `common` from std/path be implemented for url? Need suggestion.

**Comment by lino-levan** [comment]
I'm trying to think of a usecase and can't come up with one off of the top of my head. I think a lot of these things can be done as a follow-up PR. So no pressure to get everything in right now!

**Comment by tr1ckydev** [comment]
I have fixed couple of bugs and added new edge cases tests. Will be adding them later by today.

**Comment by timreichen** [comment]
What would you think of adding a `isUrl(input: string)` function, either in this PR or the future?
Seems like both [isurl](https://www.npmjs.com/package/isurl) and [is-url](https://www.npmjs.com/package/is-url) have millions of weekly downloads.

**Comment by 0f-0b** [comment]
> What would you think of adding a `isUrl(input: string)` function, either in this PR or the future?

There is [`URL.canParse`](https://developer.mozilla.org/docs/Web/API/URL/canParse_static), although browser support is currently poor.

**Comment by tr1ckydev** [comment]
https://github.com/tr1ckydev/pathurl
This is my package with all the proposed path functions for URLs along with various tests. If it's acceptable then I will port all the functions from here into `std/url`.

File: url/normalize.ts [hunk 5]
```diff
@@ -0,0 +1,14 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+import { posixNormalize } from "../path/_normalize.ts";
+
+/**
+ * Normalize the `url`, resolving `'..'` and `'.'` segments and mutiple `'/'`s into `'//'` after protocol and remaining into `'/'`.
```
kt3k, line 7:
```suggestion
 * Normalize the `url`, resolving `'..'` and `'.'` segments and multiple `'/'`s into `'//'` after protocol and remaining into `'/'`.
```

kt3k:
`dirname` might sound strange because it returns a URL instead of a string (it doesn't feel like 'name' to me), but this looks good enough for first pass.

kt3k (approved):
@tr1ckydev LGTM. Nice work!

tr1ckydev:
> `dirname` might sound strange because it returns a URL instead of a string (it doesn't feel like 'name' to me), but this looks good enough for first pass.

Agree, but as it's from url module so it can be expected. Even though dirname from path module returns parent path so works in case of url too imo.

PR #3530, opened 2023-08-08
BREAKING: move `std/flags` to new `std/cli` sub-module

### LINKED ISSUE #3526: proposal: Move `std/flags` to `std/cli`
Currently, we have this top level module which is both confusingly named (what is flags referring to? -- not immediately clear imo) and incredibly limiting in scope. I think this would be a lot better if the TLM was named `std/cli`. This would also open a lot of opportunities for new utilities that I would find quite useful imo.

See https://github.com/denoland/deno_std/pull/3488 for context on initial discussion.

cc @jeff-hykin @kt3k @iuioiua

lino-levan:
Closes #3526. Marking as draft to prevent accidental merge before discussion is complete.


cli/flags.ts
iuioiua, line 1: I think we should name this file `parse_flags.ts` and the function `parseFile()`. WDYT?

cli/mod.ts
iuioiua, line 1: Please don't forget to include the JSDocs from `flags/mod.ts`.

flags/mod.ts
```diff
@@ -2,6 +2,8 @@
 // This module is browser compatible.
 
 /**
+ * @deprecated (will be removed after 1.0.0) Import from `"https://deno.land/std@$STD_VERSION/cli/flags.ts` instead.
```
iuioiua, line 5: Perhaps, we're okay to remove after 0.208.0. @kt3k, WDYT?

iuioiua (changes requested):
Let's go ahead with this move.

cli/flags.ts
lino-levan: I'd made rename it to `parse_args.ts` and `parseArgs()`.

nhrones: > I would rename it to `parse_args.ts` and `parseArgs()` if it was up to me.

In my very ancient experience.  Flags were almost always Boolean indicators.  Often masking bits to pack flags.
Of course that just one old mans take on naming.

nhrones:
https://en.wikipedia.org/wiki/Flag_(programming)

cli/flags.ts
iuioiua: Actually, it's parsing `Deno.args()`. My bad. Yep, let's go with `parseArgs()`.

cli/mod.ts
```diff
@@ -0,0 +1,4 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+export * from "./parse_args.ts";
```
iuioiua, line 4: How about a short description for this sub-module? Maybe something like "Tools for interacting with the Deno CLI."

flags/mod.ts
```diff
@@ -2,6 +2,8 @@
 // This module is browser compatible.
 
 /**
+ * @deprecated (will be removed in 0.208.0) Import from `std/cli/parse_args.ts` instead
```
iuioiua, line 5: I think we should punt the removal version back to 0.209.0 now.

iuioiua (changes requested):
Looking good! Couple nits.

flags/mod.ts
kt3k: `std/flags` exists for long time and is heavily used. I think we shouldn't remove this that soon

lino-levan: Is `1.0.0` too late or what do you think about the timing?

kt3k: I think removal at 1.0.0 is safer and more appropriate regarding the usage of this module.

cli/mod.ts
iuioiua: Happy to approve once this is addressed, @lino-levan.

iuioiua (approved):
LGTM. Thank you, Lino.

lino-levan:
wait, don't merge yet

lino-levan:
Alright, should be mergeable now.

iuioiua:
> wait, don't merge yet

All good. I don't merge these more important PRs without Bartek's or Yoshiya's approval anyway 😛

kt3k (approved):
LGTM

The core team is also in favor of `std/cli` idea.

PR #3531, opened 2023-08-09
fix(testing/time): fix FakeTime.restoreFor accuracy for sync callbacks

Milly:
Fixed `FakeTime.restoreFor` to atomically restore and re-override `globalThis.setTimeout` etc.
This also fixes `FakeTime.prototype.delay` and `FakeTime.prototype.runMicrotasks` and others.

Of course, we don't get atomicity when an async function is specified to `restoreFor`.
However, `FakeTime.prototype.delay` that uses an non-async function will be fixed.

Example of `runMicrotasks` (that calls `delay` that calls `restoreFor`):
```ts
import { FakeTime } from "./testing/time.ts";

const testTarget = async() => {
   console.log(1, setTimeout);
   await Promise.resolve();
   console.log(2, setTimeout);
   await Promise.resolve();
   console.log(3, setTimeout);
};

const time = new FakeTime();
const promise = testTarget(); // no await, microtask is queued
time.runMicrotasks();
await promise; // already resolved
time.restore();

// Outputs before PR:
// 1 [Function: fakeSetTimeout]
// 2 [Function: setTimeout]
// 3 [Function: fakeSetTimeout]

// Outputs this PR:
// 1 [Function: fakeSetTimeout]
// 2 [Function: fakeSetTimeout]
// 3 [Function: fakeSetTimeout]
```

### DIFF
--- a/testing/time.ts
+++ b/testing/time.ts
@@ -258,7 +258,7 @@ export class FakeTime {
     // deno-lint-ignore no-explicit-any
     ...args: any[]
   ): Promise<T> {
-    if (!time) throw new TimeError("no fake time");
+    if (!time) return Promise.reject(new TimeError("no fake time"));
     restoreGlobals();
     try {
       const result = callback.apply(null, args);
--- a/testing/time.ts
+++ b/testing/time.ts
@@ -252,21 +252,26 @@ export class FakeTime {
   /**
    * Restores real time temporarily until callback returns and resolves.
    */
-  static async restoreFor<T>(
+  static restoreFor<T>(
     // deno-lint-ignore no-explicit-any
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM.

Wording of PR title might sound confusing, but this PR only changes the timing of `overrideGlobals` being called inside `FakeTime.restoreFor` when the given callback is sync. By this change, `FakeTime.restoreFor` works more accurately when the callback is sync, and that seems fixing the edge cases described in the above.

Milly:
Additional fix pushed.
Reject instead of throw on TimeError.
Match the original behavior.

kt3k:
Thanks. That makes more sense

Issue #3420, opened 2023-06-01
file_server fallback to index.html


kyeotic:
**Is your feature request related to a problem? Please describe.**

Using statically-generated single page applications, such as those produced by Vite, return 404 for dynamic deep links, since those paths are not backed by real files.

**Describe the solution you'd like**

An option to fallback to `S{fsRoot}/index.html}` for missing paths.

**Describe alternatives you've considered**

* Use hash routes. Sub-optimal because it impacts user-facing routes
* Use oak. Sub-optimal for Deno Deploy when compared to
```
cd dist/ && deployctl deploy --prod --project=$PROJECT_NAME https://deno.land/std@0.171.0/http/file_server.ts
```

PR #3536, opened 2023-08-12
feature(http/file_server): add 404 page support

### LINKED ISSUE #3533: Add Custom 404 Page Support
## Feature Description
Currently, the file server responds with a standard "Not Found" status (404) when a requested resource isn't available. To improve the user experience and provide more informative error handling, I propose supporting custom 404 pages.

## Proposed Solution
Add error handling logic to the `createServeDirResponse` function in the server. The new logic should:

1. Check for the existence of a `404.html` file in the root directory when a request for a non-existent resource is made.
2. If present, serve this `404.html` file to the client in place of the standard 404 status.
3. If the `404.html` file doesn't exist, continue returning the standard "Not Found" status.

mooxl:
Adds fallback logic for a custom 404 page as discussed and fixes #3533. 
The PR modifies the serveDirResponse logic. Instead of immediately throwing an error when a specified file in the `URL` is not found, it now performs the following checks before throwing an error:

- If `fallback` is set to false, an error is thrown.
- It checks if a custom file is specified. If yes, it uses that file; otherwise, it defaults to `404.html`.
- It retrieves the file, serves it and if it is not found, an error is thrown.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -528,6 +528,11 @@ export interface ServeDirOptions {
    * @default {false}
    */
   quiet?: boolean;
+  /** Serves fallback file when the requested resource is not found.
+   *
+   * @default {true}
+   */
+  fallback?: boolean | string;
   /** The algorithm to use for generating the ETag.
    *
    * @default {"SHA-256"}
@@ -615,6 +620,7 @@ async function createServeDirResponse(
   const target = opts.fsRoot || ".";
   const urlRoot = opts.urlRoot;
   const showIndex = opts.showIndex ?? true;
+  const fallback = opts.fallback ?? true;
   const showDotfiles = opts.showDotfiles || false;
   const { etagAlgorithm, showDirListing, quiet } = opts;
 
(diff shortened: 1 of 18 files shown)


lino-levan (commented):
There seems to be a lot of formatting changes which makes this pretty hard to review (did you format with prettier?). Are you able to revert this? 

From what I've been able to see, it looks great! Nice work.

lino-levan:
Noting here that it also solves https://github.com/denoland/deno_std/issues/3420.

mooxl:
> There seems to be a lot of formatting changes which makes this pretty hard to review (did you format with prettier?). Are you able to revert this?
> 
> From what I've been able to see, it looks great! Nice work.

Apologies for that! It's my fisr contribution to the codebase, so it's a one-time occurrence. (i hope)

You were right, the formatting problem was indeed caused by using Prettier. I've switched to ﻿`deno fmt` instead, and now ﻿`deno fmt --check` doesn't raise any complaints. However, I have noticed some occasional formatting changes where parentheses are removed or line breaks are altered. This might be due to changes in ﻿`deno fmt` itself? 

Thank you for bringing this to my attention!

kt3k:
`deno fmt` allows more formatting variations than prettier. So it doesn't format back to the original state unfortunately.. Probably it's better to revert back to main, and copy paste the relevant parts..

kt3k:
@mooxl 
I think hard coded `404.html` is enough for a first pass. I'm skeptical if we need to support custom 404 file path as there's no such request from the users

mooxl:
> @mooxl I think hard coded `404.html` is enough for a first pass. I'm skeptical if we need to support custom 404 file path as there's no such request from the users

Hey @kt3k,
Thank you for the prompt response! I believe we should add support for a custom 404 file to address issue #3420. I have made the necessary formatting changes in the updated PR.

kt3k:
I think the case #3420 can be solved by copying `index.html` to `404.html`.

Also `file_server` is historically not intended to support all use cases, but it's more for showcasing how a simple file server can be written in Deno. So it's also important file_server.ts can be read without much difficulties by curious users. With that policy in mind, we should be very careful about adding new features to `file_server.ts`

mooxl:
Hi @kt3k,
I get your point about simplicity in ﻿file_server.ts. Let's stick to the default ﻿404.html. I'll adjust the PR accordingly. Thanks for your guidance.

http/file_server_test.ts
```diff
@@ -1622,3 +1622,36 @@ Deno.test(
...
   },
 );
+
+Deno.test("serveDir serves 404.html when resource is not found", async () => {
+  const url = "http://localhost:4507/not-found.html";
+  const expectedText = "This-is-the-content-of-the-fallback-file/404.html";
+  const res = await serveDir(new Request(url), {
+    fsRoot: "./http/testdata",
+  });
+  assertEquals(res.status, 200);
```
kt3k, line 1632: I think the http status should be still 404 when serving 404.html.

I think we don't need to use `serveFile` for returning the contents of 404.html. I think we can do something like `new Response((await Deno.open(path)).readable, { status: 404 })`

lino-levan:
@kt3k could you take another look?

http/file_server.ts
```diff
@@ -643,30 +649,43 @@ async function createServeDirResponse(
...
-  }
-
-  // if target is file, serve file.
-  if (!fileInfo.isDirectory) {
-    return serveFile(req, fsPath, {
-      etagAlgorithm,
-      fileInfo,
-    });
+  let fileInfo: Deno.FileInfo | undefined;
+  try {
```
kt3k, line 653: I think we shouldn't do `try-catch` here. Because we catch the error here, the error logging in [this line](https://github.com/denoland/deno_std/blob/main/http/file_server.ts#L580) is affected (The not found error is not logged when `--verbose` option is given).

There are 3 patterns of returning NotFound HTTP response in `createServeDirResponse` (Search for `return createCommonResponse(Status.NotFound)`). We should replace those 3 with some helper function calls.

I think we can use the helper like the below:

```ts
async function createNotFound(target: string, fallback: boolean) {
  if (fallback) {
    try {
      const path = join(target, "404.html");
      const fileInfo = await Deno.lstat(path);
      if (fileInfo?.isFile) {
        return new Response((await Deno.open(path)).readable, {
          status: Status.NotFound,
        });
      }
    } catch {
      // pass
    }
  }
  return createCommonResponse(Status.NotFound);
}
```

There are also 2 places in `serveFile` where `createCommonResponse(Status.NotFound)` is returned. But those 2 place can be ignored because those code paths are only used when the user use `serveFile` API directly.

lino-levan:
@mooxl Are you able to push this forward? I'm willing to pick it up if not.

mooxl:
Hey @lino-levan,
I'm currently preparing for my exams, so I don't have time to work on this. You can go ahead with it. :)

iuioiua:
Hey guys, are we still happy to push this forward?

lino-levan:
Personally a bit busy atm, hoping either you our Max can push it forward.

iuioiua:
@kt3k, how would you like to proceed with this one?

iuioiua:
There doesn't seem to be much demand for this feature and this PR has gone stale. I suggest we close without merging. WDYT, @kt3k?

lino-levan:
I'm personally in favor of this.

iuioiua:
I'm closing this without merging, as this PR has gone stale. We'd be happy to look at a new PR if anyone is willing to finish this. Thank you, either way, @mooxl!

PR #3543, opened 2023-08-15
fix(log): rotating file handler sync setup and destroy

### LINKED ISSUE #3538: Log handler setup function is inconsistent

When creating logger handlers some of the handler setup functions are `sync` while others are `async`.

For example, I have a function that creates a logger for my modules like so:

```ts
export function createLogger(
  config: ErgomaticConfig,
  name: string,
): Logger {
  const logHandlers = [];
  const level = config.logLevel;
  const fileHandler = new handlers.FileHandler(level, {
    filename: join(logsDir(), "file.log"),
  });

  fileHandler.setup();

  logHandlers.push(
    new handlers.ConsoleHandler(level, {
      formatter,
    }),
    fileHandler,
  );

  return new Logger(name, level, {
    handlers: logHandlers,
  });
}
```

That I was using in the constructor of my modules:

```ts
class Module {
  constructor() {
    this.logger = createLogger(...)
  }
}
```

I decided I wanted to move over to [`RotatingFileHandler`](https://deno.land/std@0.198.0/log/handlers.ts?s=RotatingFileHandler) for log file rotation. My previous setup no longer works because `RotatingFileHandler.setup` is async whereas `FileHandler.setup` is sync, so this requires me to refactor all my classes if I want to use it.

Make the API consistent for log handlers, either all async or all sync.

IMO there could have been `setup` & `setupSync` like other Deno APIs but this probably can't be changed while preserving backwards compatibility.

Alternatively add a `SyncRotatingFileHandler` that can be used in `sync` contexts

lino-levan:
Closes #3538


lino-levan:
Failing on the same flaky test, rather unfortunate. Why is this only coming up recently?

kt3k:
> Failing on the same flaky test, rather unfortunate. Why is this only coming up recently?

Probably because of the parallel execution of tests #3446

kt3k (approved):
LGTM

PR #3546, opened 2023-08-17
chore(examples): remove examples folder

### LINKED ISSUE #3463: Moving the `examples` directory
The `examples` directory currently resides directly next to the directories containing the public API and uses the same naming convention as the other directories containing the public API. Likewise it is also [listed in the documentation](https://deno.land/std@0.192.0?doc) next to the public APIs.

I think it would make sense to move `examples` to e.g. `_examples` to clarify that it's not actually part of the standard library (and I think the `_` prefix would also result in the directory being excluded from the documentation page).

lino-levan:
Closes #3463. Do we want to move these somewhere? Maybe a new category can be added to denobyexample for "complete app examples" or something? Maybe these make more sense in the manual?


kt3k:
This change looks fine to me at this point. We have https://examples.deno.land/ and https://github.com/denoland/examples . Those work better as the collections of examples.

kt3k (approved):
LGTM

kt3k:
I'll label this as `chore` as these have never been intended as public APIs of std.

bartlomieju (approved):
LGTM

Issue #3547, opened 2023-08-18
std/log doesn't work well with objects with bigint


eseiker:
**Describe the bug**

Currently std/log uses `JSON.stringify` to represent `object`. (https://deno.land/std@0.198.0/log/logger.ts?source#L141)
This occurs object with properties which type are `bigint` fail to be serialized, while `JSON.stringify` doesn't take care of `bigint` by default.

**Steps to Reproduce**

```ts
> import * as log from "https://deno.land/std@0.198.0/log/mod.ts";
undefined
> const obj = { a: 123n }
undefined
> log.info(obj)
Uncaught TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at Logger.asString (https://deno.land/std@0.198.0/log/logger.ts:141:19)
    at Logger.#_log (https://deno.land/std@0.198.0/log/logger.ts:110:25)
    at Logger.info (https://deno.land/std@0.198.0/log/logger.ts:161:22)
    at Module.info (https://deno.land/std@0.198.0/log/mod.ts:385:31)
    at <anonymous>:2:5
>
```

**Expected behavior**

Objects consisting bigint properties should be logged.
While I'd better like to see `console.log` like output rather than just JSON.stringify result,
JSON.stringify with reviver could be most simple solution, such as
```ts
> JSON.stringify(obj, (_, v) => typeof v === 'bigint' ? v.toString() : v)
'{"a":"123"}'
>
```
(However, this implementation has a caveat, the result of reviver is shown as "123", not 123n or 123)

At least, there could be options to inject `asString` manually while setting up log.

**Environment**

- OS: macOS 13.5
- deno version: deno 1.36.1 (release, aarch64-apple-darwin) / v8 11.6.189.12 / typescript 5.1.6
- std version: 0.198.0

PR #3548, opened 2023-08-18
fix(log): fix logger does not serialize object with bigint values

babiabeo:
Fixes: #3547
Related: https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-521460510


babiabeo:
Close in favor of #3550

PR #3550, opened 2023-08-19
fix(log): fix serializing BigInt value in object

### LINKED ISSUE #3547: std/log doesn't work well with objects with bigint

eseiker:
Another approach to fix #3547

While my brief implementation in original issue and #3548 represent bigint value as quoted string, this PR serializes value without quotes.

Since this is my first PR submitting to deno, I hope I'm doing right to follow community guidelines😅

### DIFF
--- a/log/logger.ts
+++ b/log/logger.ts
@@ -123,8 +123,9 @@ export class Logger {
     return msg instanceof Function ? fnResult : msg;
   }
 
-  asString(data: unknown): string {
+  asString(data: unknown, isProperty = false): string {
     if (typeof data === "string") {
+      if (isProperty) return `"${data}"`;
       return data;
     } else if (
       data === null ||
@@ -138,7 +139,11 @@ export class Logger {
     } else if (data instanceof Error) {
       return data.stack!;
     } else if (typeof data === "object") {
-      return JSON.stringify(data);
+      return `{${
+        Object.entries(data)
+          .map(([k, v]) => `"${k}":${this.asString(v, true)}`)
+          .join(",")
+      }}`;
     }
     return "undefined";
   }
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks!

PR #3563, opened 2023-08-22
refactor: move `_util/diff` to `assert/_diff`

kt3k:


bartlomieju (approved):
LGTM, not super pretty but it solves the problem at hand nicely.

Can you remove `_util/diff.ts` in this PR as well?

kt3k:
Added `formatter` option instead of `serialized` option, and now `assertEquals` is less artificial in my view.

> Can you remove _util/diff.ts in this PR as well?

`_util/diff.ts` is renamed to `assert/_diff.ts`

iuioiua (approved):

bartlomieju (approved):

PR #3564, opened 2023-08-22
refactor(_util): remove `_util/os`

kt3k:
We're trying to remove top-level `_util` module because it's inconvenient for publishing each submodule independently.

This PR removes `_util/os.ts` with below changes:
- Copies `_util/os.ts` to `path/_os.ts` as `path` uses all exported items of `_util/os.ts`
- Replaces other usages with `const isWindows = Deno.build.os === "windows"` because they only use `isWindows` and also those usages don't require browser compatibility

related #3563


path/glob.ts
```diff
@@ -1,11 +1,11 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { isWindows, osType } from "../_util/os.ts";
+import { isWindows, osType } from "./_os.ts";
```
iuioiua, line 4:
```suggestion
import { isWindows, osType, type OSType } from "./_os.ts";
```

iuioiua (changes requested):

bartlomieju (approved):
LGTM

iuioiua (approved):

PR #3566, opened 2023-08-23
chore: cleanup usage of double equals

lino-levan:
Originally reported on discord. A lot of our vendored implementations use double equals which is unfortunate.

### DIFF
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -2,7 +2,6 @@
 // Documentation and interface for walk were adapted from Go
 // https://golang.org/pkg/path/filepath/#Walk
 // Copyright 2009 The Go Authors. All rights reserved. BSD license.
-import { assert } from "../assert/assert.ts";
 import { join } from "../path/join.ts";
 import { normalize } from "../path/normalize.ts";
 import {
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -137,7 +137,6 @@ export async function* walk(
   }
   try {
     for await (const entry of Deno.readDir(root)) {
-      assert(entry.name != null);
       let path = join(root, entry.name);
 
       let { isSymlink, isDirectory } = entry;
@@ -207,7 +206,6 @@ export function* walkSync(
     throw wrapErrorWithPath(err, normalize(root));
   }
   for (const entry of entries) {
-    assert(entry.name != null);
     let path = join(root, entry.name);
 
     let { isSymlink, isDirectory } = entry;
--- a/log/handlers.ts
+++ b/log/handlers.ts
@@ -42,7 +42,7 @@ export class BaseHandler {
       const value = logRecord[p1 as keyof LogRecord];
 
       // do not interpolate missing values
-      if (value == null) {
+      if (value === null) {
         return match;
       }
 
--- a/log/handlers.ts
+++ b/log/handlers.ts
@@ -42,7 +42,7 @@ export class BaseHandler {
       const value = logRecord[p1 as keyof LogRecord];
 
       // do not interpolate missing values
-      if (value === null) {
+      if (value === undefined) {
         return match;
       }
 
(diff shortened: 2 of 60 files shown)


iuioiua:
It might be also worth getting rid of instances of `!=`.

iuioiua (commented):
Does this cover all instances of `==` and `!=`? I see some cases other within JSDocs, which could also be corrected for clarity.

lino-levan:
I wasn't sure if it was valuable to update the JSDocs since many of them just assume that `==` and `===` are the same (or `!=` and `!==`). We could update them if you think it's valuable to do so.

iuioiua:
Updating those would provide extra context for those who understand that level of "type-strictness ".

log/handlers.ts
```diff
@@ -42,7 +42,7 @@ export class BaseHandler {
       const value = logRecord[p1 as keyof LogRecord];
 
       // do not interpolate missing values
-      if (value == null) {
+      if (value === null) {
```
kt3k, line 45: This looks like a check for missing property of the object. I think this should compare with `undefined`
```suggestion
      if (value === undefined) {
```

media_types/_util.ts
```diff
@@ -109,7 +109,7 @@ function indexOf<T>(s: Iterable<T>, fn: (s: T) => boolean): number {
 }
 
 export function isIterator<T>(obj: unknown): obj is Iterable<T> {
-  if (obj == null) {
+  if (obj === null) {
```
kt3k, line 112: Maybe this should be `obj === null || obj === undefined`?

path/glob_test.ts
```diff
@@ -14,7 +14,7 @@ function match(
   path: string,
   opts: GlobToRegExpOptions = {},
 ): boolean {
-  if (opts.os == null) {
+  if (opts.os === null) {
```
kt3k, line 17:
```suggestion
  if (opts.os === undefined) {
```

semver/is_semver.ts
```diff
@@ -19,7 +19,7 @@ import { isValidNumber, isValidString } from "./_shared.ts";
  * @returns True if value is a valid SemVer otherwise false
  */
 export function isSemVer(value: unknown): value is SemVer {
-  if (value == null) return false;
+  if (value === null) return false;
```
kt3k, line 22:
```suggestion
  if (value === null || value === undefined) return false;
```

semver/is_semver_comparator.ts
```diff
@@ -15,7 +15,7 @@ import { ALL, NONE } from "./constants.ts";
  * @returns True if the object is a SemVerComparator otherwise false
  */
 export function isSemVerComparator(value: unknown): value is SemVerComparator {
-  if (value == null) return false;
+  if (value === null) return false;
```
kt3k, line 18:
```suggestion
  if (value === null || value === undefined) return false;
```

semver/is_semver_range.ts
```diff
@@ -13,7 +13,7 @@ import { isSemVerComparator } from "./is_semver_comparator.ts";
  * @returns True if its a valid SemVerRange otherwise false.
  */
 export function isSemVerRange(value: unknown): value is SemVerRange {
-  if (value == null) return false;
+  if (value === null) return false;
```
kt3k, line 16:
```suggestion
  if (value === null || value === undefined) return false;
```

fs/walk.ts
iuioiua, line 1: I gather these lines were removed simply because they were redundant. Is that right?

iuioiua (approved):
Nice!

fs/walk.ts
lino-levan: That's correct, yes.

media_types/_util.ts
kt3k:
```suggestion
  if (obj === null || obj === undefined) {
```

kt3k (approved):
LGTM

PR #3567, opened 2023-08-23
BREAKING(permissions): deprecate permissions module

### LINKED ISSUE #3557: Deprecate `permissions` module
`permissions` module is a very thin wrapper of `Deno.permissions` APIs. This module doesn't look used in the ecosystem. I think we should deprecate this module before the stabilization.

part of #3489

lino-levan:
Closes #3557

### DIFF
--- a/permissions/mod.ts
+++ b/permissions/mod.ts
@@ -2,7 +2,7 @@
 /**
  * Helpers for interacting with Deno's permissions system.
  * @module
- * @deprecated (will be removed after 1.0.0) Use the Deno permission API instead
+ * @deprecated (will be removed in 1.0.0) Use the Deno permission API instead
  */
 
 const { PermissionDenied } = Deno.errors;
@@ -47,7 +47,7 @@ function getPermissionString(descriptors: Deno.PermissionDescriptor[]): string {
  * If one of the permissions requires a prompt, the function will attempt to
  * prompt for it.  The function resolves with all of the granted permissions.
  *
- * @deprecated (will be removed after 1.0.0) Use the Deno permission API instead
+ * @deprecated (will be removed in 1.0.0) Use the Deno permission API instead
  */
 export async function grant(
   ...descriptors: Deno.PermissionDescriptor[]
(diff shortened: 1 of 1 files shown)


permissions/mod.ts
```diff
@@ -2,6 +2,7 @@
 /**
  * Helpers for interacting with Deno's permissions system.
  * @module
+ * @deprecated (will be removed after 1.0.0) Use the Deno permission API instead
```
iuioiua, line 5:
```suggestion
 * @deprecated (will be in 1.0.0) Use the Deno permission API instead
```
I think this is clearer. I'd argue this is how all deprecations should be done.

lino-levan: agreed

iuioiua: Sidenote: perhaps, the "will be removed after..." check in [_tools/check_deprecation.ts](https://github.com/denoland/deno_std/blob/b34c646c45bcc3d4f7e797b78439a3cfd4d0b1b3/_tools/check_deprecation.ts) should be removed too (in a separate PR).

permissions/mod.ts
```diff
@@ -44,7 +45,10 @@ function getPermissionString(descriptors: Deno.PermissionDescriptor[]): string {
  * ```
  *
  * If one of the permissions requires a prompt, the function will attempt to
- * prompt for it.  The function resolves with all of the granted permissions. */
+ * prompt for it.  The function resolves with all of the granted permissions.
+ *
+ * @deprecated (will be removed in 1.0.0) Use the Deno permission API instead
```
iuioiua, line 50: Is the deprecation notice present when using every overload or just this one? If just the one, I'd have a sweep through the codebase and other current PRs to make the appropriate adjustments.

lino-levan: In this specific case, it's fine since we've deprecated ever individual overload (outside of the base one, which seems inaccessible? Could I get you to check that?), but I'd definitely make sure we've properly deprecated every overload.

Speaking of checking the inaccessibility of the base thing. These two work:
```ts
grant({name: "net"}, {name: "net"})
grant([{name: "net"}, {name: "net"}])
```

This doesn't:
```ts
grant([{name: "net"}], {name: "net"})
```

Why? Am I reading this type wrong?

iuioiua: Ah, whoops didn't see that each overload had the declaration. Either way, I'd check elsewhere if an overload is missing the declaration.

kt3k (approved):
LGTM. Thanks!

Issue #3562, opened 2023-08-22
Add back `check_circular_deps.ts` to `_tools/`


kt3k:
Context: We plan to publish each sub module of std independently to the registry at the stabilization of it. In that case, the circular dependencies between sub modules prevent them to be published. So now we'd want to avoid circular dependencies inside std.

Add back `check_circular_deps.ts` from https://github.com/denoland/deno_std/blob/0.177.0/node/_tools/check_circular_deps.ts, and check the circular deps of the entire std modules.

part of #3489

kt3k:
```
archive depends on:
  - assert
  - io
  - bytes
  - streams
assert depends on:
  - fmt
async depends on:
  - assert
bytes depends on nothing
collections depends on nothing
console depends on:
  - assert
crypto depends on:
  - assert
  - encoding
datetime depends on nothing
dotenv depends on:
  - collections
encoding depends on nothing
flags depends on:
  - assert
fmt depends on nothing
front_matter depends on nothing
fs depends on:
  - _util
  - assert
  - path
html depends on nothing
http depends on:
  - assert
  - datetime
  - encoding
  - async
io depends on:
  - assert
  - bytes
json depends on:
  - streams
jsonc depends on:
  - assert
  - json
log depends on:
  - assert
  - fmt
  - fs
  - io
  - bytes
media_types depends on nothing
msgpack depends on:
  - bytes
path depends on:
  - _util
  - assert
permissions depends on nothing
regexp depends on nothing
semver depends on nothing
signal depends on:
  - async
streams depends on:
  - assert
  - bytes
  - async
  - io
testing depends on:
  - assert
  - fmt
  - fs
  - path
  - _util
  - async
  - collections
toml depends on:
  - collections
uuid depends on:
  - assert
  - bytes
  - crypto
wasi depends on:
  - path
  - _util
yaml depends on:
  - io
  - assert
  - bytes
```

kt3k:
![graphviz](https://github.com/denoland/deno_std/assets/613956/838b4256-6381-4f11-9283-2efc498ce620)

lino-levan:
Ideally we'd take out the modules we're deprecating/removing from this graph.

iuioiua:
Would it be worth automatically writing the image file to the root of the repo in CI? We could even include it in the README. WDYT?

kt3k:
closed by #3569

kt3k:
[...]

PR #3569, opened 2023-08-23
chore(_tools): check dependencies between submodules

kt3k:
This PR adds a tool for checking circular dependencies between std sub-modules.

```
$ deno task lint:circular
Task lint:circular deno run --allow-env --allow-read ./_tools/check_circular_submodule_dependencies.ts
34 submodules checked.
No circular dependencies found.
```

Optionally (with `--graph` option) the tool can output graphviz graph source code. Which can be rendered like the below with graphviz cli

graphviz


iuioiua:
Closes #3562

PR #3574, opened 2023-08-24
chore: revert test fix temporarily

mmastrac:
Fails due to rusty_v8 downgrade -- will revert at a later date.


kt3k (approved):
Makes sense. LGTM

PR #3577, opened 2023-08-25
Properly quote/escape strings in assertIsError

jeff-hykin:
Currently messages will look like
- `expected X but got "["howdy", 2]"` instead of
- `expected X but got "[\"howdy\", 2]"`

Its misleading, but also inconvenient since often its nice to copy out the `instead got` and use it to fix the test case.

### DIFF
--- a/assert/assert_is_error.ts
+++ b/assert/assert_is_error.ts
@@ -22,18 +22,18 @@ export function assertIsError<E extends Error = Error>(
     );
   }
   if (ErrorClass && !(error instanceof ErrorClass)) {
-    msg = `Expected error to be instance of "${ErrorClass.name}", but was "${
-      typeof error === "object" ? error?.constructor?.name : "[not an object]"
-    }"${msgSuffix}`;
+    msg = `Expected error to be instance of ${JSON.stringify(ErrorClass.name)}, but was ${
+      typeof error === "object" ? JSON.stringify(error?.constructor?.name) : "[not an object]"
+    }${msgSuffix}`;
     throw new AssertionError(msg);
   }
   if (
     msgIncludes && (!(error instanceof Error) ||
       !stripColor(error.message).includes(stripColor(msgIncludes)))
   ) {
-    msg = `Expected error message to include "${msgIncludes}", but got "${
-      error instanceof Error ? error.message : "[not an Error]"
-    }"${msgSuffix}`;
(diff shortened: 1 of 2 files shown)


assert/assert_is_error.ts
```diff
@@ -22,18 +22,18 @@ export function assertIsError<E extends Error = Error>(
     );
   }
   if (ErrorClass && !(error instanceof ErrorClass)) {
-    msg = `Expected error to be instance of "${ErrorClass.name}", but was "${
-      typeof error === "object" ? error?.constructor?.name : "[not an object]"
-    }"${msgSuffix}`;
+    msg = `Expected error to be instance of ${JSON.stringify(ErrorClass.name)}, but was ${
+      typeof error === "object" ? JSON.stringify(error?.constructor?.name) : "[not an object]"
```
jeff-hykin, line 26: I know Debatebly the error.name shouldn't need JSON.stringify, but runtime performance is not an issue and this will handle really janky cases (e.g. I can make ErrorClass.name contain quotes cause this is JavaScript we're talking about)

assert/assert_is_error.ts
```diff
@@ -22,18 +22,18 @@ export function assertIsError<E extends Error = Error>(
...
   }
   if (
     msgIncludes && (!(error instanceof Error) ||
       !stripColor(error.message).includes(stripColor(msgIncludes)))
   ) {
-    msg = `Expected error message to include "${msgIncludes}", but got "${
-      error instanceof Error ? error.message : "[not an Error]"
-    }"${msgSuffix}`;
+    msg = `Expected error message to include ${JSON.stringify(msgIncludes)}, but got ${
+      error instanceof Error ? JSON.stringify(error.message) : "[not an Error]"
```
jeff-hykin, line 35: Also, intentionally, only the `error.message` is given quotes. I think `"[not an Error]"` is misleading since its not a string. Arguably I think `[not an Error]` should be replaced with something more descriptive about what has happened.

assert/assert_is_error.ts
kt3k: The below script throws with `TypeError: Cannot assign to read only property 'name' of function 'class ErrorClass {}'`
```js
class ErrorClass {}
ErrorClass.name = 'foo"bar'
```
I don't think it's possible to make Class.name contain quotes.

kt3k (approved):
LGTM. Thanks!

Issue #3572, opened 2023-08-24
dotenv: do not read all env vars while expanding variables


kt3k:
Currently `parse` reads the entire env vars by default while expanding variables in `.env` file, but that should be unnecessary because we only need referenced env vars.

cknight:
Now that I finally understand the situation, I'd be happy to take a crack at this.

cknight:
@kt3k I think this can be closed now.

Issue #3561, opened 2023-08-22
[std/dotenv] - change .env.example to .env.required


cknight:
The dotenv module has the concept of a `.env.example` file which dictates a set of environment variables which must be present (either through `.env`/`.env.default` files or via the process environment).  If any are missing after the dotenv is loaded then an exception is thrown.  A secondary use case is that it can be used to document the purpose of these environment variables.  I propose renaming `.env.example` to `.env.required` to be clearer as to its purpose.  'Example' does not convey that the entries are required to exist.

iuioiua:
We've now removed the `examplePath` option from `@std/dotenv/load`. See #5449 for details. Either way, thank you for your suggestion.

PR #3578, opened 2023-08-25
BREAKING(dotenv): Fix dot env permissions

cknight:
Addresses #3572 

## Headline changes
* BREAKING: remove `restrictEnvAccessTo` configuration option
* Additional documentation added, including commentary on permissions
* Removed all module dependencies
* Fixed bug (see below)
* Refactored tests, added additional tests

## Overview
This is a breaking change, primarily to remove the `restrictEnvAccessTo` configuration option.  To allow expansion of `.env` variables to use values from the process environment, the original implementation read the entire process environment into memory.  This mean all users of dotenv required the `--allow-env` permission.  Accessing all process variables was undesirable if you only needed to expand a single env variable, so `restrictEnvAccessTo` was added to restrict which env variables dotenv had access to and allow the granular `--allow-env=SOME_VAR` to work properly.  This setup deviates away from how Deno handles permissions.

This change now allows for per-env-var access with no code configuration required and if you don't export any variables or expand any process variables then no `--allow-env` permissions are needed at all.  The below examples show before and after permission handling:

## Permission example 1
```sh
#.env
HELLO=world
```
```ts
// mod.ts
import { load } from "https://deno.land/std@0.199.0/dotenv/mod.ts";
await load();
```
**Before change**: Run with  `deno run --allow-read --allow-env mod.ts`

**After change**:  Run with `deno run --allow-read mod.ts`

## Permission example 2
```sh
#.env
HELLO=${A}
```
```ts
// app.ts
import { load } from "https://deno.land/std@0.199.0/dotenv/mod.ts";
await load();
```
**Before change**: Run with  `A=world deno run --allow-read --allow-env mod.ts` (full env access still required)

**After change**:  Run with `A=world deno run --allow-read --allow-env=A mod.ts` (granular access allowed)

## Permission example 3
```sh
#.env
HELLO=${A}
```
**Before**:
```ts
import { load } from "https://deno.land/std@0.199.0/dotenv/mod.ts";
await load({restrictEnvAccessTo: ["A"]});
```
Run with  `A=world deno run --allow-read --allow-env=A mod.ts` (granular access only works with code grant to `A` via `restrictEnvAccessTo`)

**After change**: 
```ts
import { load } from "./mod.ts";
await load();
```
Run with  `A=world deno run --allow-read --allow-env=A mod.ts` (no code change required for granular access)

## Bug fix
When expanding a process environment var who value was blank, the existing code was setting the configuration value to undefined rather than empty string.

```sh
#.env
HELLO=${A}
```
**Before**:
```ts
// mod.ts
import { load } from "https://deno.land/std@0.199.0/dotenv/mod.ts";
console.log(await load());
```
Run with  `A= deno run --allow-read --allow-env mod.ts` yields:
```
{ HELLO: "undefined" }
```

**After change**:
```ts
// mod.ts
import { load } from "./mod.ts";
console.log(await load());
```
Run with  `A= deno run --allow-read --allow-env=A mod.ts` yields:
```
{ HELLO: "" }
```
## Additional consideration
As this is a breaking change, it may be desirable to consider #3561 as well, which would also be a breaking change to dotenv.  However, as there is no feedback on that proposal to date no PR for that change has been progressed.


cknight:
Any ideas why the linter is failing?  Also failing locally for me, but I'm not sure how to fix it.

iuioiua:
`std/streams/readable_stream_from_iterable.ts` needs to be deleted as it's deprecated. Best doing that in another PR.

dotenv/mod.ts
```diff
@@ -221,13 +221,6 @@ export interface LoadOptions {
    * @default {"./.env.defaults"}
    */
   defaultsPath?: string | null;
-
-  /**
-   * List of Env variables to read from process. By default, the complete Env is
-   * looked up. This allows to permit access to only specific Env variables with
-   * `--allow-env=ENV_VAR_NAME`.
-   */
-  restrictEnvAccessTo?: StringList;
```
kt3k, line 230: Let's keep this option only in typing in a few versions to prevent immediately breaking users code (in type checks)

dotenv/mod.ts
```diff
@@ -1,6 +1,8 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 /**
- * Load environment variables from a `.env` file.
+ * Load environment variables from a `.env` file.  Loaded variables are accessible
+ * in a configuration object returned by the `load()` function, as well as optionally
+ * exporting them to the process environment using the `export` option.
```
kt3k, line 5: Thanks for adding these lines.

kt3k (approved):
LGTM. Nice clean up and fix!

Issue #3522, opened 2023-08-08
proposal: `std/ulid`


iuioiua:
Not exclusively, [ULIDs](https://github.com/ulid/spec) can be valuable in Deno KV scenarios. See [here](https://deno.land/manual@e017a74bd9f1fd50d980799c78499fc04e105fee/runtime/kv/key_space#universally-unique-lexicographically-sortable-identifiers-ulids) for a soon-to-be-added section in the KV docs that explains why and [here](https://github.com/denoland/deno_kv_oauth/pull/151) for an example.

@kt3k's [x/ulid](https://deno.land/x/ulid) module implements ULID functionality. However, that code may provide greater value if it lived in the Standard Library regarding visibility and opportunities for optimisation.

The migration would be an easy 1:1 copy/paste of that code to here.

iuioiua:
We already have other single-file modules, and it seems fine.

iuioiua:
I realise that ULIDs in Deno KV can:
1. Eliminate the need for `*_by_time` indexes.
2. Eliminate the need for `createdAt: Date` entry properties.

The result of these two benefits can lead to far simpler codebases. E.g. try picturing [Deno SaaSKit's KV logic](https://github.com/denoland/saaskit/blob/e27502ed610489e0e94bb7bee9eafee7a610a04e/utils/db.ts) using ULIDs and seeing how much simpler it could make things. I'd even suggest that ULIDs for Deno KV entry IDs be made an official recommendation. Either way, I'm more strongly about this being part of the Standard Library.

PR #3582, opened 2023-08-26
feat(ulid): port /x/ulid module

lino-levan:
Closes https://github.com/denoland/deno_std/issues/3522.

I think this is valuable, particularly for Deno KV.


iuioiua (changes requested):
We probably only want to export the following from `mod.ts`:
* `decodeTime()`
* `factory()`
* `monotonicFactory()`
* `PRNG`
* `ULID`
I think the others should be moved to a `_util.ts` or similar file.

iuioiua:
We also probably want to include the examples from the README in the JSDocs.

ulid/test.ts
```diff
@@ -0,0 +1,233 @@
...
+  assert,
+  assertEquals,
+  assertStrictEquals,
+  assertThrows,
+} from "../assert/mod.ts";
+
+import * as ULID from "./mod.ts";
+const ulid = ULID.factory();
+
+Deno.test("ulid", async (t) => {
```
iuioiua, line 16: Nit: I don't see the benefit of having this extra nesting level with tests. It could be more flat if done as separate `Deno.test()`s. Not a strong opinion, though.

iuioiua (approved):
Nice!

ulid/mod.ts
```diff
@@ -0,0 +1,129 @@
...
+  return time;
+}
+
+/**
+ * Generates a ULID function given a PRNG
+ *
+ * @example To use your own pseudo-random number generator, import the factory, and pass it your generator function.
+ * ```ts
+ * import { factory } from "https://deno.land/std@$STD_VERSION/ulid/mod.ts";
+ * import prng from "somewhere";
```
iuioiua, line 65: This line is causing the CI errors. Perhaps, just use `detectPrng()`?

ulid/mod.ts
```diff
@@ -0,0 +1,129 @@
...
+ * ulid(150000); // 000XAL6S41ACTAV9WEVGEMMVRC
+ *
+ * // Even if a lower timestamp is passed (or generated), it will preserve sort order
+ * ulid(100000); // 000XAL6S41ACTAV9WEVGEMMVRD
+ * ```
+ *
+ * @example You can also pass in a prng to the monotonicFactory function.
+ * ```ts
+ * import { monotonicFactory } from "https://deno.land/std@$STD_VERSION/ulid/mod.ts";
+ * import prng from "somewhere";
```
iuioiua, line 99: Ditto

iuioiua (approved):
LGTM again

kt3k:
@iuioiua Do you know the usage of `monotonicFactory`? I'm not sure what it's used for.

I'm also unsure we need to export `factory`. I guess nobody would replace pseudo random number generator.

I'd suggest we should only expose `ulid` and `decodeTime` functions (and related types) at least for first pass.

iuioiua:
I think `monotonicFactory()` is worth keeping. It's used for generating monotonically increasing ULIDs, which would be valuable in guaranteeing no conflicts when high numbers of ULIDs are generated within the same millisecond.

I agree with the rest of your points.

kt3k:
Ok. Monotonicity looks useful. ([This doc](https://github.com/ulid/spec#monotonicity) also mentions about guaranteeing of generation order).

While monotonicity is useful, `monotonicFactory` looks still redundant as it is for specifying alternative random number generator. I'd suggest we should replace it with non-factory version (such as `ulidMonotonic` or maybe `generateMonotonic`)

iuioiua:
Yeah, it seems a little funny. It only increments monotonically if the last timestamp is less than or equal to the current timestamp (now). I think it's fine how it is with the name `monotonicFactory`. However, that's not a strong opinion.

Half-related: I think deviating from the ULID's JavaScript implementation is fine. Otherwise, we'd be putting ourselves into a box regarding capability. E.g. I found a possible tweak that could improve the performance of ULID generation by 10x while maintaining cryptographically random strings. Though, I'd have to validate it's correctness first 🤣

iuioiua:
WDYT, @kt3k?

lino-levan:
I would almost flip the naming to have `ulid` be monotonically increasing and some sort of non-monotonic `ulid` exist on the side? It feels like outside of migration, everyone should really use monotonic `ulid`s.

kt3k:
With monotonic version, the random bits are not random anymore, and it seems having far higher possibility of collision in distributed usage. I think monotonic version is not something we should recommend by default.

lino-levan:
> With monotonic version, the random bits are not random anymore

I missed that. This makes sense then. Completely agree that no one will replace the PRNG function. I'll rework the PR tonight if I can.

iuioiua:
Oh, I completely missed that you suggested removing the random number generator parameter. Yes, that parameter looks mostly useless to me.

ulid/_util.ts
```diff
@@ -0,0 +1,145 @@
...
+  return str;
+}
+
+export function encodeRandom(len: number, prng: PRNG): string {
+  let str = "";
+  for (; len > 0; len--) {
+    str = randomChar(prng) + str;
+  }
+  return str;
+}
```
iuioiua, line 56:
```suggestion
export function encodeRandom(len: number): string {
  let str = "";
  const randomBytes = crypto.getRandomValues(new Uint8Array(len));
  for (let i = 0; i < len; i++) {
    str += ENCODING[randomBytes[i] % ENCODING_LEN];
  }
  return str;
}
```
This function is 10x faster, retains cryptographic randomness and removes the need for the `prng` parameter. Doing this might need tweaking elsewhere in the implementation.

lino-levan: I'll run some statistical tests tomorrow, but this makes sense to me.

lino-levan: This sort of breaks a lot of tests unfortunately. I'll try to pick this up on the bus tomorrow.

lino-levan:
Return of the flaky tests I guess. This is ready for review @iuioiua.

iuioiua (changes requested):
Cool to see that change implemented! Few things:
* I don't think `monotonicFactory()` should have an `encodeRand` parameter or any parameter.
* This implementation deviates from the original JS implementation (e.g. faster `encodeRandom()` and no `prng` parameter. I think these differences should be pointed out in the JSDocs somewhere.

ulid/_util.ts
```diff
@@ -0,0 +1,111 @@
...
+export function encodeTime(now: number, len: number = TIME_LEN): string {
+  if (now > TIME_MAX) {
+    throw new Error("cannot encode time greater than " + TIME_MAX);
+  }
+  if (now < 0) {
+    throw new Error("time must be positive");
+  }
+  if (Number.isInteger(now) === false) {
+    throw new Error("time must be an integer");
+  }
```
iuioiua, line 32:
```suggestion
  if (!Number.isInteger(now) || now < 0 || now > TIME_MAX) {
    throw new Error("Time must be a positive integer less than " + TIME_MAX);
  }
```
Nit

ulid/_util.ts
```diff
@@ -0,0 +1,111 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+interface ULID {
+  (seedTime?: number): string;
+}
+
+interface PRNG {
+  (): number;
+}
```
iuioiua, line 9:
```suggestion
```
I believe this is no longer needed. Right?

ulid/_util.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Generates a ULID function
+ */
+export function factory(): ULID {
+  return function ulid(seedTime: number = Date.now()): string {
+    return encodeTime(seedTime, TIME_LEN) + encodeRandom(RANDOM_LEN);
+  };
+}
```
iuioiua, line 78: How come this is here instead of `mod.ts`?

ulid/_util.ts
```diff
@@ -0,0 +1,111 @@
...
+ * ulid(150000); // 000XAL6S41ACTAV9WEVGEMMVR9
+ * ulid(150000); // 000XAL6S41ACTAV9WEVGEMMVRA
+ * ulid(150000); // 000XAL6S41ACTAV9WEVGEMMVRB
+ * ulid(150000); // 000XAL6S41ACTAV9WEVGEMMVRC
+ *
+ * // Even if a lower timestamp is passed (or generated), it will preserve sort order
+ * ulid(100000); // 000XAL6S41ACTAV9WEVGEMMVRD
+ * ```
+ */
+export function monotonicFactory(encodeRand = encodeRandom): ULID {
```
iuioiua, line 99: How come this is here instead of `mod.ts`?

ulid/_util.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Generates a ULID function
+ */
+export function factory(): ULID {
+  return function ulid(seedTime: number = Date.now()): string {
+    return encodeTime(seedTime, TIME_LEN) + encodeRandom(RANDOM_LEN);
+  };
+}
+
+/**
+ * Generates a monotonically increasing ULID, optionally given a PRNG.
```
iuioiua, line 81:
```suggestion
 * Generates a monotonically increasing ULID.
```

iuioiua (changes requested):

lino-levan:
> I don't think monotonicFactory() should have an encodeRand parameter or any parameter.

Strongly disagree. It is impossible to test without this parameter. It is not exposed to end users so I don't think this matters.

> This implementation deviates from the original JS implementation (e.g. faster encodeRandom() and no prng parameter. I think these differences should be pointed out in the JSDocs somewhere.

Again, not exposed to the user so I don't know if that's relevant.

ulid/_util.ts
lino-levan: See above.

ulid/_util.ts
lino-levan: If I'm reading Yoshiya's comments correctly, he doesn't believe these should be exposed to end users. I have to agree. If they are unable to change the PRNG there is no use for this. If it was up to me I'd remove the factory functions entirely but it makes testing for monotonic harder.

lino-levan: I've gone ahead and removed this factory function entirely. I've kept the monotonic factory for now.

ulid/_util.ts
```diff
@@ -0,0 +1,92 @@
...
+    if (charIndex === -1) {
+      throw new Error("incorrectly encoded string");
+    }
+    if (charIndex === maxCharIndex) {
+      str = replaceCharAt(str, index, ENCODING[0]);
+      continue;
+    }
+    return replaceCharAt(str, index, ENCODING[charIndex + 1]);
+  }
+  throw new Error("cannot increment this string");
```
iuioiua, line 58: Is this line reachable? If not, it should be removed.

ulid/_util.ts
iuioiua: Nice

ulid/mod.ts
```diff
@@ -0,0 +1,80 @@
...
+ * monotonicUlid(150000); // 000XAL6S41ACTAV9WEVGEMMVR9
+ * monotonicUlid(150000); // 000XAL6S41ACTAV9WEVGEMMVRA
+ * monotonicUlid(150000); // 000XAL6S41ACTAV9WEVGEMMVRB
+ * monotonicUlid(150000); // 000XAL6S41ACTAV9WEVGEMMVRC
+ *
+ * // Even if a lower timestamp is passed (or generated), it will preserve sort order
+ * monotonicUlid(100000); // 000XAL6S41ACTAV9WEVGEMMVRD
+ * ```
+ */
+export const monotonicUlid = monotonicFactory();
```
iuioiua, line 66: I totally misread your initial suggestion, @kt3k. Now, I understand, and I like it 😎

iuioiua (approved):
Regardless of my previous unreachable line comment, LGTM! This implementation is actually better than the original JavaScript implementation because:
1. Random strings are generated 10x faster.
2. The redundant `prng` parameter is removed. Fewer options are better.
3. Monontic ULIDs are possible without having first to create the factory.

Good stuff, @lino-levan and @kt3k!

ulid/_util.ts
lino-levan: This was supposed to be reachable, but due to a bug it wasn't. Did not affect behavior in any way, but nice catch!

iuioiua:
Anything else pending for this PR?

lino-levan:
Nothing pending afaik. Waiting for any comments from @kt3k.

iuioiua:
I've noticed a bug with `monotonicUlid()` in https://github.com/denoland/saaskit/pull/514. For some reason, ULIDs aren't respecting their `seedTime` parameters. However, `ulid()` works fine.

lino-levan:
Could you write a reproduction? I suspect it may be a feature but I'm not too sure. @kt3k may know more about it.

iuioiua:
Yeah, I'll get a repro snippet when I can.

lino-levan:
Alright I've been staring at this code for too long. I don't see this bug. I'd appreciate a fresh set of eyes.

iuioiua:
Sorry, I haven't had the chance to look into this further. @kt3k, would `std/ulid` be considered an unstable module once merged? If so, I think it'd be worth merging as-is.

kt3k (approved):
LGTM. Thank you for your efforts! @lino-levan 

Note: Core team also looked in favor of including this.

kt3k:
> would std/ulid be considered an unstable module once merged?

Yes, we are planning to keep ulid unstable (versioned as 0.x) for a while.

PR #3584, opened 2023-08-30
BREAKING(encoding): deprecate encoding/binary

### LINKED ISSUE #3559: Clean up `encoding` module
Most APIs in `encoding` are simple encoders & decoders, but `encoding/binary` includes APIs based on legacy Reader/Writer interface (and others in it look only thin wrappers around `DataView` class). I think we should deprecate `encoding/binary`.

part of #3489

kt3k:
closes #3559

### DIFF
--- a/encoding/binary.ts
+++ b/encoding/binary.ts
@@ -12,7 +12,7 @@ import type { Reader, Writer } from "../types.d.ts";
 type RawBaseType = "int8" | "int16" | "int32" | "uint8" | "uint16" | "uint32";
 type RawNumberType = RawBaseType | "float32" | "float64";
 type RawBigType = RawBaseType | "int64" | "uint64";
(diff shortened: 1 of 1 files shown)


encoding/binary.ts
```diff
@@ -10,20 +12,30 @@ import type { Reader, Writer } from "../types.d.ts";
 type RawBaseType = "int8" | "int16" | "int32" | "uint8" | "uint16" | "uint32";
 type RawNumberType = RawBaseType | "float32" | "float64";
 type RawBigType = RawBaseType | "int64" | "uint64";
+/**  * @deprecated (will be removed after 0.205.0) */
```
iuioiua, line 15:
```suggestion
/** @deprecated (will be removed after 0.205.0) */
```
Nit

iuioiua (commented):
Do these deprecations need an explanation?

PR #3586, opened 2023-08-30
BREAKING(yaml): change binary handling

### LINKED ISSUE #3585: Remove dependency to `io` in `yaml`
We are stabilizing `yaml` and deprecating `io`, but `yaml` depends on `io`. We need to remove this dependency.

part of #3489

kt3k:
Currently `yaml` module parses `!!binary "..BASE64.."` notation to `Buffer` class, and serializes `Buffer` (or any object with `readSync` method) to `!!binary "..BASE64.."` notation.

This PR replaces `Buffer` in the above usage with `Uint8Array`. I think this makes more sense in JavaScript.

Also this change is necessary to remove the dependency to `std/io` from `std/yaml`. We're going to deprecate `std/io` soon. This change is necessary for stabilizing `yaml` in the future.

closes #3585

### DIFF
--- a/yaml/parse_test.ts
+++ b/yaml/parse_test.ts
@@ -230,3 +230,13 @@ Deno.test({
     assertEquals(parse(yaml3), expected);
   },
 });
+
+Deno.test({
+  name: "parse bianry type",
+  fn() {
+    const yaml = `message: !!binary "SGVsbG8="`;
+    assertEquals(parse(yaml), {
+      message: new Uint8Array([72, 101, 108, 108, 111]),
+    });
+  },
+});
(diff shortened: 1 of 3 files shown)


yaml/parse_test.ts
```diff
@@ -230,3 +230,13 @@ Deno.test({
     assertEquals(parse(yaml3), expected);
   },
 });
+
+Deno.test({
+  name: "parse bianry type",
```
bartlomieju, line 235:
```suggestion
  name: "parse binary type",
```

bartlomieju (approved):

Issue #3489, opened 2023-07-12
Release a stable version 1.0


piscisaureus:
- Clean ups
  - [x] denoland/deno_std#3600
  - [x] denoland/deno_std#3655
  - [x] denoland/deno_std#3601
  - [x] denoland/deno_std#3551
  - [x] denoland/deno_std#3617
  - [x] denoland/deno_std#3618
  - [x] denoland/deno_std#3549
  - [x] denoland/deno_std#3559
  - [x] denoland/deno_std#3599
  - [x] denoland/deno_std#3602
  - [x] denoland/deno_std#3616
  - [x] denoland/deno_std#3623
- Deprecations
  - [x] denoland/deno_std#3552
  - [x] denoland/deno_std#3557
  - [x] denoland/deno_std#3558
- Others
  - [x] denoland/deno_std#3562

### dependencies between std sub modules

![graphviz](https://github.com/denoland/deno_std/assets/613956/6dd5d466-a2dc-49d8-b38e-f9ab07fee1be)

- White (double circle): ready
- Yellow (circle): needs clean up
- Pink (box): not ready (needs feedback, needs re-design, too young)
- Gray (polygon): deprecated / remove

lino-levan:
We should go over every api we offer to make sure we're not accidentally stabilizing an API that really shouldn't exist.

jsejcksn:
Re: version `1.0`:

The readme document currently includes [this](https://github.com/denoland/deno_std/blob/0.194.0/README.md?plain=1#L11-L18):

> ## Releases
> 
> Standard library is currently tagged independently of Deno version. This will change once the library is stabilized.
> 
> To check compatibility of different version of standard library with Deno CLI see [this list](https://raw.githubusercontent.com/denoland/dotland/main/versions.json).

I interpret this to mean that the intention is to version `std` with the same version numbers as the CLI after stabilization, being released in-step. Is that interpretation accurate? Should the original statement be modified or removed?

boywithkeyboard:
I agree with @jsejcksn that the std library should follow the versioning of Deno as shared by various Deno staff members before.

lino-levan:
We really need to re-write `std/path` before `1.0`. That would be a blunder to leave the way it is.

kt3k:
@lino-levan 
> We really need to re-write std/path before 1.0. That would be a blunder to leave the way it is.

Can you create an independent issue for it with more details?

lino-levan:
[...]

PR #3589, opened 2023-08-30
feat(bytes): Deprecate BytesList class

### LINKED ISSUE #3549: Clean up `bytes` module
Most APIs of bytes module are simple functions that work on Uint8Arrays, but only `BytesList` has exceptional design and purpose.

`BytesList` was introduced for optimizing `readDelim` method of `io` (then `bufio`) module. https://github.com/denoland/deno_std/pull/867 It was more like internal utility for optimization rather than a public API.

I think we should deprecate this API before stabilizing `std/bytes` (but we might better to check the actual usages in the ecosystem)

part of #3489

aapoalas:
As part of std cleanup, this PR deprecates the `BytesList` class and suggests plain `Uint8Array[]` usage instead. The two non-deprecated users of this class are also transformed to use a plain `Uint8Array[]`.

These are:
1. msgpack: Here the transformation is trivial and the `bytes/concat` function can be utilized directly.
2. DelimiterStream: Here the transformation is very much nontrivial.

### DelimiterStream
Switching to a plain Array based approach made quite a few optimisations more directly "visible" in the stream implementation. These are in no particular order:
1. The generally expected case is that a chunk will include a delimiter. It's generally thus the case that the list of previous chunks is relatively short, and that at the end of a chunk's iteration all previous chunks will usually have been fully "forgotten" by the stream. This makes it a bit easier to reason about what is going on in the stream.
2. All chunks are always fully iterated, meaning that the iteration index is always a local property: At the start of a chunk's iteration it is at the beginning of that chunk and at the end it is at its end. There is no need to keep it in-between iterations.
3. Match index only needs to be "persisted" at the end of the chunk iteration. Otherwise it can be a purely local variable.
4. **Controversial / uncertain / API contract change:** If a particular slice of data is coming from a single chunk, then there is no strict need to reallocate memory if we trust that the buffers we've been given are "ours" to give. This means that the buffers that our DelimiterStream outputs may be sub-views of (or direct references to) the buffers that it has been given, and any consumers should consider themselves free to touch or mutate only those sub-views they've been given. This is likely a fair performance boost, but it does mean that the API contract is much more strict suddenly.

closes #3549
part of #3489

### DIFF
--- a/streams/delimiter_stream.ts
+++ b/streams/delimiter_stream.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { BytesList } from "../bytes/bytes_list.ts";
+import { concat } from "../bytes/concat.ts";
 import { createLPS } from "./_common.ts";
 
 /** Disposition of the delimiter. */
@@ -52,13 +52,11 @@ export interface DelimiterStreamOptions {
  * @returns Transform stream
  */
 export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
-  #bufs = new BytesList();
+  #bufs: Uint8Array[] = [];
   #delimiter: Uint8Array;
-  #inspectIndex = 0;
   #matchIndex = 0;
-  #delimLen: number;
   #delimLPS: Uint8Array;
-  #disp?: DelimiterDisposition;
+  #disp: DelimiterDisposition;
 
   constructor(
     delimiter: Uint8Array,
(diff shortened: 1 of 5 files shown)


streams/delimiter_stream.ts
```diff
@@ -83,32 +80,109 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
...
-          const shift = this.#disp === "prefix" ? start : this.#inspectIndex;
-          this.#bufs.shift(shift);
-          this.#inspectIndex = this.#disp === "prefix" ? this.#delimLen : 0;
-          this.#matchIndex = 0;
+          matchIndex = 0;
+          const delimiterStartIndex = inspectIndex - delimLen;
+          const delimitedChunkEnd = disposition === "suffix"
+            ? inspectIndex
+            : delimiterStartIndex;
+          if (delimitedChunkEnd <= 0 && bufs.length === 0) {
```
aapoalas, line 104: **comment:** Kind of an obvious downside here: I've written out all the possible cases in excruciating detail to optimise each of them individually, and added a fair bit of comments. The number of lines has gone way up.

kt3k (approved):
I think lower level utilities like `DelimiterStream` should be optimized for performance rather than ergonomics. The changes in it look fair and reasonable to me. LGTM

johnspurlock:
Heads up - this seems to have broken `DelimiterStream`: see #3609

PR #1985, opened 2022-03-03
feat: web streams based archive/tar

crowlKats:
still WIP

closes #1658

PR #3593, opened 2023-08-31
[std/archive] Improve docs and general code cleanup

cknight:
std/archive needs some TLC.  This is the first in a series of improvements to this module.  Specifically, this PR adds or improves documentation and makes cosmetic code changes primarily for readability.

### DIFF
--- a/archive/_common.ts
+++ b/archive/_common.ts
@@ -15,17 +15,18 @@ export interface TarInfo {
 
 export interface TarOptions extends TarInfo {
   /**
-   * append file
+   * Filepath of the file to append to the archive
    */
   filePath?: string;
 
   /**
-   * append any arbitrary content
+   * A Reader of any arbitrary content to append to the archive
    */
   reader?: Reader;
 
   /**
-   * size of the content to be appended
+   * Size of the content to be appended.  This is only required
+   * when passing a reader to the archive.
    */
   contentSize?: number;
 }
@@ -46,7 +47,7 @@ export enum FileTypes {
   "contiguous-file" = 7,
 }
 
-export const recordSize = 512;
+export const HEADER_LENGTH = 512;
 
 /*
 struct posix_header {           // byte offset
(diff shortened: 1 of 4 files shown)


kt3k (approved):
LGTM. Thanks!

kt3k:
BTW we've been trying to redesign this module. This module is based on legacy Reader/Writer interfaces, but we've been deprecating them in favor of ReadableStream/WritableStream interfaces.

We tried to make `std/archive/tar` to use ReadableStream/WritableStream in https://github.com/denoland/deno_std/pull/1985, but it needed to depend on [unfinished web stream proposal](https://github.com/whatwg/streams/pull/1145), and we couldn't land it.

We'd appreciate if you could make progress along with this direction.

cknight:
Thanks @kt3k, that's good to hear.  My further proposals for this module are around quality of life usage (adding compression and convenience functions to make the API much more user friendly).  They build upon the existing Reader usage however.  Given this, and the desire to redesign the module, I can hold off if the redesign is now able to be progressed.  I'd be happy to help progress, though streams are new to me.  However, it looks like the unfinished web stream proposal is still unfinished and therefore continues to block the completion of #1985?

kt3k:
Yeah, it still blocks #1985, but I was wondering if there's some alternative way which avoids depending on the proposal, but still replaces Reader/Writer with web streams..

iuioiua:
> Yeah, it still blocks #1985, but I was wondering if there's some alternative way which avoids depending on the proposal, but still replaces Reader/Writer with web streams..

[Apparently not](https://github.com/denoland/deno_std/pull/1985#issuecomment-1362053375)

cknight:
@crowlKats Are you able to articulate in layman's terms what is incomplete with your efforts in #1985?  I've read your PR and the corresponding incomplete spec proposal but am missing the gap it plugs.

crowlKats:
well, the spec proposal itself needs to land to be able to have #1985

cknight:
Ah, are you saying then that the code you've written matches the proposed spec, but we can't land the PR until the spec lands first?  Makes sense, thanks.

PR #3595, opened 2023-08-31
fix(fmt): make printf respect NO_COLOR

### LINKED ISSUE #3575: [std/fmt/printf.ts] (s)printf doesn't honor NO_COLOR
**Describe the bug**
The `printf` and `sprint` functions do not honor NO_COLOR when using `%i` or `%I`.

**Steps to Reproduce**
This one line demonstrates the problem. Setting `NO_COLOR=true` and/or `colors.setColorEnabled(false)` will set `Deno.noColor=true` and yet `sprintf` and `printf` produce color with no way to disable the color.

```bash
$ export NO_COLOR=true; deno eval --ext=ts 'import * as fmt from "https://deno.land/std@0.200.0/fmt/printf.ts"; import * as colors from "https://deno.land/std@0.200.0/fmt/colors.ts"; colors.setColorEnabled(false); const obj = {a: 1, b: 2, c: "d",}; fmt.printf("Deno.noColor: %t\n%T\n%#v\n%j\n%i\n%I", Deno.noColor, obj, obj, obj, obj, obj);'
```

The cli output has color.
image

When viewed from a system that doesn't support ANSI, the escape codes make it hard to read.
image

**Expected behavior**

I expect the [NO_COLOR](http://no-color.org/) environmental variable be honored and the coloring would be disabled. My use case is calling Deno from another program and all output is logged using a `ConsoleHandler` function which uses `sprintf` for objects. I prefer the readability of `%i` but the ANSI escape codes make it hard to read.

Note: I included `colors.setColorEnabled(false)` in an attempt to disable coloring even though `std/fmt/colors.ts` is a separate module, and has no effect on `std/fmt/printf.ts`.

Note 2: (related) PR #182 added NO_COLORS to the colors module.

**Environment**

- OS: openSUSE Leap 15.4
- deno version: 1.36.3
- std version: 0.200.0

lino-levan:
Closes #3575

### DIFF
--- a/fmt/printf.ts
+++ b/fmt/printf.ts
@@ -907,7 +907,7 @@ class Printf {
    */
   fmtI(val: unknown, compact: boolean): string {
     return Deno.inspect(val, {
-      colors: true,
+      colors: !Deno?.noColor,
       compact,
       depth: Infinity,
       iterableLimit: Infinity,


iuioiua (approved):

kt3k (approved):
LGTM

PR #3607, opened 2023-09-01
fix(std): Bug fixes for all std/url functions and new tests.

tr1ckydev:
Fixes few bugs when handling urls with search params and #s.
Added new tests.


url/extname.ts
```diff
@@ -10,5 +11,6 @@ import { posixExtname } from "../path/_extname.ts";
  */
 export function extname(url: string | URL) {
   url = new URL(url);
-  return posixExtname(url.pathname);
```
kt3k, line 13: I think this was more appropriate

tr1ckydev: The only difference in new vs old is, in the new one if the url is like `deno.land`, it can extract `.land` from it which i believe can be a useful feature.

url/strip.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+/**
+ * Strips any hash (eg. `#header`) or search parameters (eg. `?foo=bar`) from the provided URL.
+ *
+ * (Mutates the original url provided)
+ * @param url url to be stripped.
+ */
+export function strip(url: URL) {
```
kt3k, line 10: I think this is too trivial to expose as a public API (`url.hash = ""; url.search = "";` is enough easy to perform)

url/extname.ts
kt3k: > in the new one if the url is like deno.land, it can extract .land from it

I don't think that's what users expect as `.land` is top level domain, not an extension of the url.

kt3k:
Other part looks good to me

url/extname.ts
tr1ckydev: Done.

url/strip.ts
tr1ckydev: Done.

kt3k (approved):
Thanks for updating! LGTM

PR #3608, opened 2023-09-01
feat(stripAnsiCode): escape erase character

sigmaSd:
I run into this while debugging a windows terminal app

A quick search on github shows that it seems to be the erase character sequence https://github.com/search?q=ESC%5B%23X&type=code


kt3k (approved):
I found the sequence in this docs https://learn.microsoft.com/en-us/windows/console/console-virtual-terminal-sequences#text-modification

LGTM

PR #3611, opened 2023-09-03
Fix(streams): DelimiterStream regression

### LINKED ISSUE #3609: `DelimiterStream` regression in 0.201.0, drops the first character of certain lines
DelimiterStream drops the first character of certain lines starting in 0.201.0 - found due to data corruption in a production system, so I hope the Deno folks can fix this quickly before other Deno users' data is silently corrupted.

To repro, run the following minimal repro deno script, which compares the lines from v200 vs v201 of DelimiterStream.

Nothing should be printed, every line should be exactly the same in both versions.  But note that a few of the lines are missing the first char in v201, and it's not always deterministic between runs! 

Looks like this class was refactored in 201, maybe this can be included as a test going forward?

```ts
import { DelimiterStream as DelimiterStream200 } from 'https://deno.land/std@0.200.0/streams/delimiter_stream.ts';
import { DelimiterStream as DelimiterStream201 } from 'https://deno.land/std@0.201.0/streams/delimiter_stream.ts';

// collect lines from a large-ish text file using two different versions of DelimiterStream
const results: Record<string, string[]> = {};
const newline = new Uint8Array([ 10 ]);
for (const [ name, DelimiterStream ] of Object.entries({ DelimiterStream200, DelimiterStream201 })) {
    results[name] = [];
    const stream = (await fetch('https://raw.githubusercontent.com/publicsuffix/list/master/public_suffix_list.dat')).body!;
    const allChunks = stream.pipeThrough(new DelimiterStream(newline));
    for await (const chunk of allChunks) {
        const line = decoder.decode(chunk);
        results[name].push(line);
    }
}

// notice that 201 drops the first char of some lines, also not deterministic!
results.DelimiterStream200.forEach((line200, i) => {
    const line201 = results.DelimiterStream201[i];
    if (line200 !== line201) console.log(`line ${i + 1}: ${line200} -> ${line201}`);
});
```

- OS: MacOS Ventura 13.5.1 (22G90)
- deno version: deno 1.36.4 (release, aarch64-apple-darwin)
- std version: 0.201.0

aapoalas:
Two regressions appeared in the recent DelimiterStream refactoring:
1. #3609 The pushing of chunk remnants was off by one, which would mean that if an incoming chunk's last byte was the first byte to be pushed out into an outgoing chunk then that byte would accidentally be left out.
2. Unreported issue but related to the first that appeared once it was fixed was that if an incoming chunk started with a separator then the chunk starting pointer was not properly moved forward.

A regression test was added testing for these two errors.

closes #3609

### DIFF
--- a/streams/delimiter_stream.ts
+++ b/streams/delimiter_stream.ts
@@ -105,6 +105,7 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             // Our chunk started with a delimiter and no previous chunks exist:
             // Enqueue an empty chunk.
             controller.enqueue(new Uint8Array());
+            chunkStart = disposition === "prefix" ? 0 : inspectIndex;
           } else if (delimitedChunkEnd > 0 && bufs.length === 0) {
             // No previous chunks, slice from current chunk.
             controller.enqueue(chunk.subarray(chunkStart, delimitedChunkEnd));
@@ -130,6 +131,8 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
               // for a suffix disposition this branch would mean that the previous
               // chunk ended with a full match but was not enqueued.
               chunkStart = inspectIndex;
+            } else {
+              chunkStart = 0;
             }
           } else if (delimitedChunkEnd < 0 && bufs.length > 0) {
             // Our chunk started by finishing a partial delimiter match.
@@ -147,6 +150,7 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             if (disposition === "prefix") {
               // Must keep last bytes of last chunk.
               bufs.push(last.subarray(lastSliceIndex));
+              chunkStart = 0;
             } else {
               chunkStart = inspectIndex;
             }
@@ -155,10 +159,10 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             const chunkSliced = chunk.subarray(chunkStart, delimitedChunkEnd);
             const result = concat(...bufs, chunkSliced);
             bufs.length = 0;
+            controller.enqueue(result);
             chunkStart = disposition === "prefix"
               ? delimitedChunkEnd
               : inspectIndex;
-            controller.enqueue(result);
           } else {
             throw new Error("unreachable");
           }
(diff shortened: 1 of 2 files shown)


streams/delimiter_stream.ts
```diff
@@ -105,6 +105,7 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             // Our chunk started with a delimiter and no previous chunks exist:
             // Enqueue an empty chunk.
             controller.enqueue(new Uint8Array());
+            chunkStart = disposition === "prefix" ? 0 : inspectIndex;
```
aapoalas, line 108: **comment:** This is the fix for the second issue: We need to push our chunk start index forward unless we're going include the delimiter in the next chunk.

streams/delimiter_stream.ts
```diff
@@ -130,6 +131,8 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
               // for a suffix disposition this branch would mean that the previous
               // chunk ended with a full match but was not enqueued.
               chunkStart = inspectIndex;
+            } else {
+              chunkStart = 0;
```
aapoalas, line 135: **comment:** This is actually unnecessary: The chunk start is necessarily already zero here but I decided to add this here for readability reasons.

streams/delimiter_stream.ts
```diff
@@ -147,6 +150,7 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             if (disposition === "prefix") {
               // Must keep last bytes of last chunk.
               bufs.push(last.subarray(lastSliceIndex));
+              chunkStart = 0;
```
aapoalas, line 153: **comment:** Same thing here, this is unnecessary: The chunk start is necessarily already zero here but I decided to add this here for readability reasons.

streams/delimiter_stream.ts
```diff
@@ -155,10 +159,10 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             const chunkSliced = chunk.subarray(chunkStart, delimitedChunkEnd);
             const result = concat(...bufs, chunkSliced);
             bufs.length = 0;
+            controller.enqueue(result);
```
aapoalas, line 162: **comment:** This reordering is done so that all branches would agree on the order of actions: First enqueue, then move chunk start pointer.

streams/delimiter_stream.ts
```diff
@@ -179,7 +183,7 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
     this.#matchIndex = matchIndex;
     if (chunkStart === 0) {
       bufs.push(chunk);
-    } else if (chunkStart !== length - 1) {
+    } else if (chunkStart < length) {
```
aapoalas, line 186: **comment:** This is the fix for the first issue: This is an off-by-one-error here.

kt3k (approved):
Confirmed this passes the check given in #3609 

LGTM. Thank you @aapoalas for the quick fix!

PR #3612, opened 2023-09-04
test(http): reduce http server test flakiness

kt3k:
This PR updates `http/server_test.ts`. Now we use different port number for each test case.

This should reduce the possibility of rare test failure in linux with the error message `AddrInUse: Address already in use (os error 98)` (ex https://github.com/denoland/deno_std/actions/runs/6067720722/job/16459754435 )


iuioiua (approved):
LGTM! Nice monotonically increasing ports 🤣

bartlomieju (approved):

Issue #3600, opened 2023-09-01
Clean up platform specific APIs of `path` module


kt3k:
We currently expose platform specific APIs like the below:

```js
import { posix, win32 } from "https://deno.land/std@0.200.0/path/mod.ts";
```

or
```js
import * as win32 from "https://deno.land/std@0.200.0/path/win32.ts";
import * as posix from "https://deno.land/std@0.200.0/path/posix.ts";
```

I think this has 2 problems:
- There's no way to import single implementation of platform specific API
- 'posix' and 'win32' keywords don't match the way we usually specify the platform name (i.e. it doesn't match to `Deno.build.os`)

One way to solve this is to implement #687

part of #3489

lino-levan:
> 'posix' and 'win32' keywords don't match the way we usually specify the platform name

I'm in favor of renaming `win32.ts`->`windows.ts`. It's unclear to me what `posix.ts` could be renamed to while keeping it's semantic meaning.

> One way to solve this is to implement https://github.com/denoland/deno_std/issues/687

As I've mentioned before, I'm not quite sure how to go about implementing this. There are some APIs (`resolve` and friends) which don't fit well into adding an optional argument. Breaking changes would almost certainly be required.

kt3k:
> There are some APIs (resolve and friends) which don't fit well into adding an optional argument.

Something like the below seems working. Does this help designing it?:

```ts
type OS = "windows" | "linux" | "darwin" | "freebsd";

function resolve(...args: string[]): string;
function resolve(...args: [...string[], { os: OS }]): string;
function resolve(...args: any[]): string {
    //
    return "...";
}

resolve("a", "b");
resolve("a", "b", { os: "windows" });
```

lino-levan:
Ah, fair enough, I never really got the hang of overloading typescript functions :sweat_smile:. Seems like a reasonable and non-breaking change. I'll pick this up later today.

jollytoad:
I made a suggestion for this here... https://github.com/denoland/deno_std/discussions/3553

Adding an os option is overkill and unnecessary.

kt3k:
#3553 might be fine too. I'm curious what people think about these options.

lino-levan:
Either way makes sense to me. The complexity is about the same either way.

kt3k:
[...]

PR #3613, opened 2023-09-04
feat(path): add option os argument

### LINKED ISSUE #687: path: Make os an option argument

lino-levan:
Closes #687 and #3600

I took this opportunity to unify test names within the module.


path/basename.ts
```diff
@@ -11,8 +12,12 @@ import { posixBasename, windowsBasename } from "./_basename.ts";
  * @param path - path to extract the name from.
  * @param [suffix] - suffix to remove from extracted name.
  */
-export function basename(path: string, suffix = ""): string {
-  return isWindows
+export function basename(
+  path: string,
+  suffix = "",
+  options?: PathOptions,
```
lino-levan, line 18: Ideally this would be a single argument like
```ts
options?: {
  suffix?: string
  os?: OSType
}
```
but we don't live in an ideal world. Not sure what to do about this.

kt3k:
@lino-levan Thank you for your work anyway!

PR #3622, opened 2023-09-04
BREAKING(media_types): deprecate typeByExtension

### LINKED ISSUE #3616: Clean up `media_types` module
`typeByExtention` API looks redundant as it's covered by `contentType`. I think we should deprecate it before stabilizing it.

part of #3489

lino-levan:
Closes #3616


kt3k (approved):
LGTM. Thanks!

PR #3626, opened 2023-09-06
fix(assert/assert_false): Fix incorrect JSDoc placement

hugoalh:
That JSDoc should belong to `function assertFalse` but not `type Falsy`.


kt3k (approved):
LGTM. Thanks

PR #3628, opened 2023-09-08
BREAKING(collections): move RedBlackTree, BinarySearchTree, and BinaryHeap to 'unstable' subdir

### LINKED ISSUE #3599: Clean up `collections` module
Most APIs in `collections` are simple functions inspired by lodash or kotlin.collectioins. However the following 5 classes are very different in design and purpose.

- BinaryHeap
- BinarySearchNode
- BinarySearchTree
- RedBlackNode
- RedBlackTree

I think we should move them under `std/collections/unstable` directory for now before stabilizing it, and keep considering the more appropriate way to expose them as public APIs.

part of #3489

kt3k:
This PR moves RedBlackTree, BinarySearchTree, and BinaryHeap to `collections/unstable` sub directory to prepare for the stabilization of `std/collections`.

Existing APIs are all aliased from the APIs from `collections/unstable`. There should not be immediate breakage in the users' code.

closes #3599


bartlomieju (approved):
LGTM, provided lint stage is fixed.

PR #359, opened 2019-04-24
http: Cookie extend

zekth:
Add `delCookie` and `setCookie`.

I have a concern about the multiple `Set-Cookie` header. It's common to use BUT `Headers` does not give us the possibility to do this. Do you think any workaround is possible? I think we can keep it like this for the moment and work on multiple `Set-Cookie` headers handling in a future PR.

Also Days / Months format are hardcoded in the module because ATM there is some discussions about embedding or not the ICU in Deno which may break this module. Ref: https://github.com/denoland/deno/issues/1968

PR #3633, opened 2023-09-09
BREAKING(datetime): deprecate to_imf.ts

lino-levan:
I know we're planning on deprecating the whole module but this one is especially not useful.

### DIFF
--- a/datetime/to_imf.ts
+++ b/datetime/to_imf.ts
@@ -2,6 +2,8 @@
 // This module is browser compatible.
 
 /**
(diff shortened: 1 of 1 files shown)


lino-levan:
I was curious on how this function even came to be (https://github.com/denoland/deno_std/pull/359) and it looks like `.toUTCString()` was just overlooked. Interesting rabbit-hole.

kt3k (approved):
LGTM. Thanks

> I was curious on how this function even came to be (https://github.com/denoland/deno_std/pull/359) and it looks like .toUTCString() was just overlooked. Interesting rabbit-hole.

Thanks for the research!

Issue #3601, opened 2023-09-01
Clean up `front_matter` module


kt3k:
`front_matter` module doesn't strictly follow the single-export pattern. We should clean it up and make it follow single-pattern strictly before stabilizing.

part of #3489

JakeAve:
When you say single-export pattern, do you mean add the any, json, toml and yaml exports to the mod.ts?

```
export * as any from './any.ts'
export * as json from './json.ts'
export * as toml from './toml.ts'
export * as yaml from './yaml.ts'
```

kt3k:
Single-export pattern means every API is implemented in its own file. In the case of `front_matter`, there are APIs like `createExtractor` and `test`, which are directly implemented in `mod.ts`, that breaks the pattern. We'd like to move them into its own file.

kt3k:
I think we should also do the below to make the module comformant to single export pattern:
- deprecate default export from `any.ts`, `json.ts`, `yaml.ts`, and `toml.ts`
- deprecate `test` export from `any.ts`, `json.ts`, `yaml.ts`, and `toml.ts`
- deprecate `Format` type and use union type instead for format params

PR #3634, opened 2023-09-12
fix(front_matter): Move exports to their own module

JakeAve:
Cleaning up front_matter per #3601. Moved `createExtractor` and `test` to their own modules. Copied tests into respective test files.

I'm looking for guidance because in trying to recycle the regex patterns by moving them to a `formats` module, we technically have two exports there.


front_matter/createExtractor.ts
kt3k, line 1: let's use snake_case for file names

front_matter/formats.ts
kt3k, line 1: I'd suggest we should keep this private (i.e. prefix the file name with `_`) as this looks exposing too much internal details.)

kt3k:
Thanks for the PR. Mostly looks good to me. Left a few comments

kt3k (approved):
LGTM. Thanks!

PR #3351, opened 2023-05-02
refactor(_tools): update check_doc_imports.ts

kt3k:
This PR uses `npm:typescript@5.0.2` instead of `https://esm.sh/typescript@5.0.2`. This change removes the dependency to `Deno.core` (`Deno.core.runMicrotasks` are used in `std@0.177.0/node` and it's in dependnecy of `https://esm.sh/typescript@5.0.2`)

closes #3337

PR #3635, opened 2023-09-13
chore(_tools): fix type errors in `_tools`

kt3k:
This PR fixes the below type errors in `_tools`. Currently scripts in `_tools` are not type-checked because they are not imported from any test cases. This PR also adds the type check task for them and run it in CI.

- `semver.increment(ver: string)` is now type error since #3591 (`_tools/check_deprecation.ts`)
- There are type errors in `_tools/check_doc_imports.ts` introduced in https://github.com/denoland/deno_std/pull/3351

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,19 +83,7 @@ jobs:
         run: deno fmt --check
 
       - name: Lint
-        run: deno lint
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

PR #3636, opened 2023-09-13
chore(_tools): update status towards stabilization

kt3k:
This PR updates the status of sub modules towards the stabilization.


lino-levan:
To be clear, this PR doesn't stabilize these modules right?

bartlomieju (approved):

kt3k:
@lino-levan 
> To be clear, this PR doesn't stabilize these modules right?

No. This tool is just informational thing. This doesn't define the stabilization.

PR #3638, opened 2023-09-13
fix(testing/time): fix FakeTime.next to return false if all timers are cleared

Milly:
Fix `FakeTime.next()` and `nextAsync()` to return `false` if all timers are cleared.

## Problem

The document for `FakeTime.next()` says:
```
Advances time to when the next scheduled timer is due.
If there are no pending timers, time will not be changed.
Returns true when there is a scheduled timer and false when there is not.
```

However, if the timer is cleared before calling `next()`, it will return true even if there is no scheduled timer.

```typescript
import { FakeTime } from "https://deno.land/std@0.201.0/testing/time.ts";

const time = new FakeTime();
const start = time.now;

const id = setTimeout(() => {}, 1000);
clearTimeout(id);
const hasNext = time.next();

console.log(hasNext); // true, but expected false
console.log(time.now - start); // 1000, but expected 0
```

### DIFF
--- a/testing/time.ts
+++ b/testing/time.ts
@@ -157,6 +157,16 @@ function* timerIdGen() {
   while (true) yield i++;
 }
 
+function nextDueNode(): DueNode | null {
+  for (;;) {
+    const dueNode = dueTree.min();
+    if (!dueNode) return null;
+    const hasTimer = dueNode.timers.some((timer) => dueNodes.has(timer.id));
+    if (hasTimer) return dueNode;
+    dueTree.remove(dueNode);
+  }
+}
+
 let startedAt: number;
 let now: number;
 let initializedAt: number;
@@ -375,7 +385,7 @@ export class FakeTime {
    * Returns true when there is a scheduled timer and false when there is not.
    */
   next(): boolean {
-    const next = dueTree.min();
+    const next = nextDueNode();
     if (next) this.now = next.due;
     return !!next;
   }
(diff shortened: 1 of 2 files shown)


testing/time.ts
```diff
@@ -157,6 +157,16 @@ function* timerIdGen() {
   while (true) yield i++;
 }
 
+function nextDueNode(): DueNode | null {
+  for (;;) {
+    const dueNode = dueTree.min();
+    if (!dueNode) return null;
+    const hasTimer = dueNode.timers.some((timer) => dueNodes.has(timer.id));
+    if (hasTimer) return dueNode;
+    dueTree.remove(dueNode);
```
kt3k, line 166: Removing dueTree item here feels confusing to me. How about fixing `fakeClearTimeout` and `fakeClearInterval` instead?

```diff
 function fakeClearTimeout(id?: number) {
   if (!time) throw new TimeError("no fake time");
-  if (typeof id === "number" && dueNodes.has(id)) {
-    dueNodes.delete(id);
+  if (typeof id === "number") {
+    const dueNode = dueNodes.get(id);
+    if (dueNode) {
+      dueNodes.delete(id);
+      dueTree.remove(dueNode);
+    }
   }
 }
```

kt3k (commented):
Good catch. Thanks! Left one comment

testing/time.ts
Milly: I did this because I wanted to define the deletion process in one place.

And the code example you provided is incomplete.
`DueNode.timers` contains timers with other ids, so it should not simply be removed from the dueTree.
I added a test about it.

kt3k: hmm, we already do similar deletion in `now` setter. Now I think it's fine to delete tree item here.

kt3k (approved):
LGTM

PR #3641, opened 2023-09-15
BREAKING(front_matter): deprecate Format enum, use union type instead

kt3k:
part of #3601 #3489 

This PR deprecates `Format` enum, and use the union type `"yaml" | "toml" | "json" | "unknown"` instead for `format` parameters.


lino-levan (approved):
LGTM. Makes a lot of sense.

front_matter/create_extractor.ts
```diff
@@ -1,11 +1,12 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 
 import {
-  Format,
   MAP_FORMAT_TO_EXTRACTOR_RX,
   MAP_FORMAT_TO_RECOGNIZER_RX,
 } from "./_formats.ts";
 
+type Format = "yaml" | "toml" | "json" | "unknown";
```
iuioiua, line 8: Should this be exported?

lino-levan: Nice catch, I would think so.

kt3k: The name 'Format' is currently exported as enum value, which this PR deprecates. I think exporting this 'Format' in this PR could break typing of the user's code.

How about using literal type (`"yaml" | "toml" | "json" | "unknown"`) directly in public interfaces?

PR #3649, opened 2023-09-18
refactor: split path into per-os modules

jollytoad:
Splits path functions into a module per platform and common fns.
Also applied the pattern to glob.

As discussed in #3553
Also see #3600


path/posix/glob.ts
```diff
@@ -0,0 +1,125 @@
...
+export { isGlob } from "../_common/is_glob.ts";
+
+const constants: GlobConstants = {
+  sep: "/+",
+  sepMaybe: "/*",
+  seps: ["/"],
+  globstar: "(?:[^/]*(?:/|$)+)*",
+  wildcard: "[^/]*",
+  escapePrefix: "\\",
+};
```
kt3k, line 23: This looks nice

kt3k:
Maybe web can move `normalizeGlob` and `joinGlobs` to `_common`?

path/mod.ts
```diff
@@ -23,12 +35,17 @@
  */
 
 import { isWindows } from "./_os.ts";
-import * as _win32 from "./win32.ts";
-import * as _posix from "./posix.ts";
+import * as _windows from "./windows/mod.ts";
+import * as _posix from "./posix/mod.ts";
 
-export const win32 = _win32;
+/** @deprecated */
```
kt3k, line 41: 👍

path/win32.ts
kt3k, line 1: I wonder if we should also deprecate this path (and also `path/posix.ts`).

kt3k:
This should be backward compatible from my understanding. Mostly looks good to me.

I wonder if it makes sense to have `path/posix/to_namespaced_path.ts` as it doesn't have effect in non-window platforms.

path/glob.ts
```diff
@@ -1,53 +1,26 @@
...
+  joinGlobs as windowsJoinGlobs,
+  normalizeGlob as windowsNormalizeGlob,
+} from "./windows/glob.ts";
+
+export { isGlob } from "./_common/is_glob.ts";
+
+export type { GlobOptions };
+
+export type GlobToRegExpOptions = GlobOptions & {
   os?: OSType;
```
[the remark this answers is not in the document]
jollytoad: This was mainly so that the tests still work, personally I think the `os` option should be ditched entirely otherwise it's inconsistent with the other path fns, but the tests will need reworking to account for that.

jollytoad:
> I wonder if it makes sense to have `path/posix/to_namespaced_path.ts` as it doesn't have effect in non-window platforms.

I think it makes sense to keep it purely for the sake of API consistency between `/path`, `/path/posix` & `/path/windows`.

path/win32.ts
jollytoad: It would be useful to be able to deprecate an entire module, so that it's import statement and/or all imported members display with usual deprecated strike-through, but I'm not sure there is a way to do that is there?

kt3k:
> I think it makes sense to keep it purely for the sake of API consistency between /path, /path/posix & /path/windows.

Ok. Makes sense.

kt3k:
Is there anything remaining with this PR?

path/win32.ts
kt3k: Probably need to alias items and deprecate the aliases.

```js
import * as path from "./windows/mod.ts";
/** @deprecated (will be removed after 0.210.0) Import from `path/windows/mod.ts` instead. */
export const sep = path.sep;
/** @deprecated (will be removed after 0.210.0) Import from `path/windows/mod.ts` instead. */
export const delimiter = path.delimiter;
/** @deprecated (will be removed after 0.210.0) Import from `path/windows/mod.ts` instead. */
export const resolve = path.resolve;
...
```

jollytoad:
> Maybe web can move `normalizeGlob` and `joinGlobs` to `_common` like `isGlob`?

I was just looking into this and noticed that there is an inconsistency in the options. The GlobOptions jsdoc says `globstar` defaults to `true`, which is the case for `globToRegExp`, but in `normalizeGlob` & `joinGlob` it actually defaults to `false`!

Also, these fns are pretty simple and rely on platform specific imports (eg. SEP, SEP_PATTERN, normalize, join), which would have to be passed into a common fn, I think it may just be simplier to retain the platform specific implementations of each.

kt3k (approved):
LGTM. Nice work!

PR #3653, opened 2023-09-19
BREAKING(front_matter): deprecate default exports

kt3k:
This PR deprecates default export items from `front_matter`. In deno std, we usually don't use default exports except these 4 (and also `wasi/snapshot_preview1.ts`).

This aligns `front_matter` design to other modules.

part of #3601 #3489

PR #3654, opened 2023-09-19
BREAKING(front_matter): deprecate language-specific `test` functions

kt3k:
This PR deprecates language specific `test` functions. This PR makes front_matter module aligned to single-export pattern together with #3653 #3641 

part of #3601 #3489

### DIFF
--- a/front_matter/any.ts
+++ b/front_matter/any.ts
@@ -6,6 +6,7 @@ import { parse as parseYAML } from "../yaml/parse.ts";
 import { parse as parseTOML } from "../toml/parse.ts";
 
 export { Format } from "./_formats.ts";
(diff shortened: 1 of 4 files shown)


kt3k:
The motivation of this is:
1. to make `front_matter` conformant to single-export pattern
2. `test` seems like an advanced usage of front matter (`extract` is the basic way to use it). So we probably don't need to expose `test` for each language, but it probably is sufficient to provide only generic `test` function, which supports all combination of languages.

iuioiua (approved):
LGTM

lino-levan (approved):
LGTM. Anything keeping this open? Just waiting for community feedback?

lucacasonato (approved):
LGTM

Issue #3617, opened 2023-09-04
Clean up `encoding` module, part 2


kt3k:
`encoding` module has the following encodings and corresponding decode and encode methods:

- ascii85
  - encode: Uint8Array -> string
  - decode: string -> Uint8Array
- base32
  - encode: Uint8Array -> string
  - decode: string -> Uint8Array
- base58
  - encode: ArrayBuffer | string -> string
  - decode: string -> Uint8Array
- base64
  - encode: ArrayBuffer | string -> string
  - decode: string -> Uint8Array
- base64url
  - encode: ArrayBuffer | string -> string
  - decode: string -> Uint8Array
- hex
  - encode: Uint8Array -> Uint8Array
  - decode: Uint8Array -> Uint8Array
- varint
  - encode: bigint | number -> [Uint8Array, number]
  - decode: Uint8Array -> [bigint, number]

Encodings except `varint` are similar in typing, but there are discrepancies (The case of `hex` is pointed in #3605). I think these 6 encodings should have the same typings, for example `encode: Uint8Array | ArrayBuffer | string -> string; decode: string -> Uint8Array`

part of #3489

lino-levan:
@kt3k is this completed?

kt3k:
@lino-levan Yeah, let's close this. Thanks for your contributions!

PR #3660, opened 2023-09-20
BREAKING(encoding): rename encode/decode to encodeFoo/decodeFoo

### LINKED ISSUE #3643: Proposal: deprecate `encode`/`decode`, instead introduce `encodeFoo/decodeFoo`
Currently `encode` and `decode` methods in `std/encoding` have incompatible signatures:

For `encode`, there're 3 types: `Uint8Array -> Uint8Array` `Uint8Array -> string` `ArrayBuffer | string -> string`

For `decode`, there're 2 types: `string -> Uint8Array` `Uint8Array -> Uint8Array`

I think these should be aligned to the same type (The most possible candidate for that type would be `ArrayBuffer | Uint8Array | string -> string` for `encode` and `string -> Uint8Array` for `decode` in my view).

This transition can't be done without breaking changes. To avoid the confusion, I'd suggest we should deprecate all existing `encode`/`decode`, and add `encodeFoo`/`decodeFoo` (ex. `encodeBase64`/`decodeBase64`) instead.

cc @bartlomieju @ry @piscisaureus @lucacasonato

related #3605 #3617 #3489

lino-levan:
closes #3643 and #3617


lino-levan:
We export `byteLength` from `base32.ts`. I don't know if that makes sense.

lino-levan:
Should be ready to review now.

kt3k (approved):
LGTM. Thank you!

PR #3661, opened 2023-09-20
BREAKING(http): clean up module

### LINKED ISSUE #3646: Rename `http_status.ts` and `http_errors.ts`
They're already in the `http` folder. No need to restate the name.

### LINKED ISSUE #3655: Clean up plan for `http` module
There are 12 public paths in `std/http`:
- `cookie.ts`
- `cookie_map.ts`
- `etag.ts`
- `file_server.ts`
- `http_errors.ts`
- `http_status.ts`
- `method.ts`
- `negotiation.ts`
- `server.ts`
- `server_sent_event.ts`
- `user_agent.ts`
- `util.ts`

I'd suggest we should stabilize the following in first pass:
- `cookie.ts`
- `etag.ts`
- `file_server.ts`
- `http_status.ts` (The path might be renamed #3646)
- `negotiation.ts`
- `user_agent.ts`

and I'd suggest we move the below to `unstable/` sub directory of `std/http`
- `cookie_map.ts` (Looks overly complex. Not sure signed cookie is best practice today. KeyRing abstraction might be confusing)
- `http_errors.ts` (Not sure error status code as Error object is the general enough abstraction)
- `method.ts` (Our mothod list deviates from [IANA's method registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml), instead aligned to http.METHODS of node.js. Not sure this is correct)
- `server.ts` (`Server` is confusing as now CLI has [`Deno.Server`](https://deno.land/api?s=Deno.Server). Also the usage of this is niche now as CLI has Deno.serve).
- `server_sent_event.ts` (The issue #3645 looks concerning)
- `util.ts` (Not sure it's worth exposing this (createCommonResponse) as std API)

lino-levan:
Closes #3655 and closes #3646


http/unstable/errors.ts
iuioiua, line 1: How come this is part of the unstable API? I'm sure I'm missing context, but the move surprises me. We use errors in SaaSKit to handle serving error responses in the REST API by throwing these HTTP errors (https://github.com/denoland/saaskit/blob/main/plugins/error_handling.ts).

iuioiua (commented):
I think prepending `unstable_` instead of having an `unstable` folder is a bit tidier. Either way, seeing the `http_` prefix being dropped from these files is cool!

kt3k:
> I think prepending unstable_ instead of having an unstable folder is a bit tidier.

Might be a good idea. @lino-levan What do you think?

BTW we already have `unstable/` sub directory in `crypto` and `collections`

http/unstable/errors.ts
kt3k: In my view, representing http errors as JS Error objects is an opinionated approach.

In most cases when you throw http error in your route handler or middleware, you can also directly return `Response` object with error status code.

For example, the below code:
```js
throw createHttpError(status, message)
```
can usually be replaced by
```js
return new Response(message, { status });
```
or similar. If this replacement can be applied universally, then HttpError abstraction might be just redundant.

iuioiua: If a request handler has a call stack size of `n`, using returns and finally returning an error response may require up to `n` conditionals. An error needs to be handled once to return an error response. I previously used the former method in SaaSKit before realising it could be done much easier once I discovered the HTTP error code in the Standard Library. It certainly made things easier to deal with.

To be clear, if we decide that HTTP errors should be unstable, fair enough. I just wanted to provide a valid counterargument.

kt3k: That might be a valid use case. Let's keep the discussion in an independent issue

lino-levan:
I think we're at least a little committed to the `unstable` folder convention, but I think `unstable_x` reads better. The point of unstable APIs anyways is that they're little used so it'd probably be fine to move them around one more time.

iuioiua:
Gentle ping, @lino-levan.

lino-levan:
I have some concerns. Following up on discord.

lino-levan:
This should be ready to be merged now. Please review again, when you both get the chance.

http/http_errors.ts
```diff
@@ -46,160 +49,7 @@
...
-  options?: HttpErrorOptions,
-): HttpError {
-  return new errors[Status[status] as ErrorStatusKeys](message, options);
-}
-
-/** A type guard that determines if the value is an HttpError or not. */
-export function isHttpError(value: unknown): value is HttpError {
-  return value instanceof HttpError;
-}
+export * from "./unstable_errors.ts";
```
iuioiua, line 55: Are the original JSDocs retained or does the deprecation notice overwrite them all? Same question elsewhere.

lino-levan: I'm not sure... I suspect it's up to the editor (or maybe it's spec'd out and I'm completely wrong)? My editor LSP is kind of borked so I might not be the best person to ask.

iuioiua: I’m sure it completely overwrites the rest of the JSDocs. I recommend exporting each function/class/whatever with its respective JSDoc with the prepended deprecation notice.

lino-levan: Will do.

kt3k:
Looks good to me. Thanks for your contribution! @lino-levan 

Let's land when Asher's point is addressed.

lino-levan:
I believe this is ready to be merged. Please review. I would like a second pair of eyes to see if I re-exported everything properly.

kt3k:
It looks like the below structure doesn't work for deprecating re-export.
```js
/** @deprecated */
export { foo } from "..";
```

Instead, we need to create aliases and deprecate them.
```js
import { foo as foo_ } from "..";
/** @deprecated */
export const foo = foo_;
```
I'll make these changes in this branch

kt3k (approved):
LGTM. Thank you for your contribution!

lino-levan:
Thanks for the final push @kt3k! Good to know what the proper reexport structure is.

iuioiua:
We should check whether other deprecated sub-modules have included the correct deprecation JSDocs.

Issue #3594, opened 2023-08-31
Should we slow down deprecation/removal to support the usage in browsers?


kt3k:
We deprecated and removed `readableStreamFromIterable` in `std/streams` because Deno [implemented](https://github.com/denoland/deno/pull/19446) `ReadableStream.from` in v1.35, which overlaps with `readableStreamFromIterable`.

However the above removal affected the user who uses `readableStreamFromIterable` in Web browsers (Note: `ReadableStream.from` is not yet available in browsers except firefox). See https://github.com/denoland/deno_std/pull/3579#issuecomment-1700637852 )

Should we slow down our deprecation/removal pace to better support such use cases? (Especially 'removal')

kt3k:
I think it makes sense to keep some APIs only for supporting browsers because we support and promote [Fresh](https://github.com/denoland/fresh) for web app development, and it encourages the mixed style of server-side and client-side programming.

lino-levan:
I feel like there's almost an opportunity for a `/std/shim` which contains these "moved" APIs (and has room for other possible shims). Just a thought.

kt3k:
Do you mean we should provide `ReadableStream.from` shim in `std/shim`?

lino-levan:
That was my feeling, yes. We could also not do shims and tell people to look for more well-known solutions (corejs), but I'm not really too sure about that. It feels like shims for semi-modern (post IE) runtimes could be useful.

lino-levan:
We could just provide some ponyfills if needed. I'm kind of swaying back and forth on this one. The docs should be really clear what method to use instead in modern runtimes.

jollytoad:
My specific use case was for shunting some server-side Deno code into a ServiceWorker, it would certainly be handy if std provider something like the suggested `std/shim`, `ReadableStream.from` and `URLPattern` are two good polyfill candidates for this. I'd be happy to help contribute to this effort if it was agreed upon as a path forward.

iuioiua:
I like the idea of a polyfill until sufficient browser support for a particular API exists. I think it should have the following properties:
- Be identical to the upcoming API, as much as possible, including type signature, behaviour and name.
   ```ts
   // streams/readable_stream_from.ts
   function readableStreamFrom<R>(asyncIterable: AsyncIterable<R> | Iterable<R | PromiseLike<R>>): ReadableStream<R>
   ```
- Declare that the function is for client-side code while browser support is pending and that the native equivalent is recommended outside of that use case.

[...]

PR #3579, opened 2023-08-26
BREAKING(streams): remove readable_stream_from_iterable

lino-levan:
Failing CI in https://github.com/denoland/deno_std/pull/3578

PR #3663, opened 2023-09-21
BREAKING(collections): deprecate `groupBy()`

NotWoods:
`groupBy` can be replaced by `Object.groupBy` in Deno 1.37. Creating this as a draft for now because the TypeScript lib isn't updated, but running this without typechecking passes fine.


lino-levan:
Related to #3594.

collections/group_by.ts
```diff
@@ -2,6 +2,8 @@
 // This module is browser compatible.
 
 /**
+ * @deprecated Use `Object.groupBy` instead.
```
iuioiua, line 5:
```suggestion
 * @deprecated (will be removed in 0.207.0) Use {@linkcode Object.groupBy} instead.
```

iuioiua (changes requested):

iuioiua (approved):
This change looks reasonable and good to me. Thanks!

kt3k:
Deprecation makes sense to me, but safari still doesn't support this according to MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy

I'm afraid we might get similar reaction to the case of `readableStreamFromIterable()` https://github.com/denoland/deno_std/pull/3579#issuecomment-1700637852

lino-levan:
Not sure about this. I'm still considering `std/shim` or `std/ponyfill` or something.

kt3k:
We updated the deprecation policy in https://github.com/denoland/deno_std/pull/3807

Now we recommend the users to pin the URL to the old std version if they still need them.

kt3k (approved):
LGTM

PR #3665, opened 2023-09-21
Avoid internally using strings for numbers

NotWoods:
Small change to avoid using numbers that are converted to strings then converted back to numbers.

`mergeReadableStreams` and `zipReadableStreams` both used `Object.entries(array)`, which turned the array indexes into strings. Using `array.entries()` instead lets us stick with numbers the whole time. `zipReadableStreams` can additionally just use a set instead of working with indexes at all.

file_server_test used `port?: string` as its config type instead of `port?: number`, which also required coersion.

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -23,7 +23,7 @@ let child: Deno.ChildProcess;
 let stdout: ReadableStream<string>;
 
 interface FileServerCfg {
-  port?: string;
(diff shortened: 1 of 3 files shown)


lino-levan (approved):
LGTM. Thanks!

kt3k (approved):
LGTM. Thanks!

PR #3666, opened 2023-09-21
fix(encoding): add validation of the input types

kt3k:
This PR adds the validation of the input type for encoding APIs.

### DIFF
--- a/encoding/_util.ts
+++ b/encoding/_util.ts
@@ -1,0 +1,29 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+const encoder = new TextEncoder();
+
+function getTypeName(value: unknown): string {
+  const type = typeof value;
+  if (type !== "object") {
+    return type;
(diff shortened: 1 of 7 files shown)


lino-levan (commented):
This PR in it's current state LGTM. It solves the problem in the title. I just question whether it makes sense to add runtime validation for these things. Typescript should prevent people from putting in an invalid input? I don't know if this is worth the additional runtime cost.

kt3k:
A user can opt-out from typescript in Deno. In that case, they don't see type error for wrong input types.

Also a user can specify wrong types by mistake:
```js
const result: { data: string }  = await response.json();
encodeBase64(result.data); // `result.data` can be a number if the response is not intended one
```

In those cases, throwing validation error will be useful, I think.

lino-levan:
Makes sense to me. Just making sure this was the reasoning.

bartlomieju (approved):
LGTM 👍

PR #3667, opened 2023-09-21
fix(datetime): Correctly format midnight in 12-hour time

alexanderreiff:
I found that a Date object in the midnight hour will be formatted as `0:xx` in a `hmm` date time format.

Deno's [datetime.format documentation](https://deno.land/std@0.202.0/datetime/mod.ts) for `h`/`hh` says this should be from 1-12, as does the underlying [Unicode specification](https://www.unicode.org/reports/tr35/tr35-dates.html#dfst-hour).


kt3k (approved):
LGTM. Thanks!

PR #3672, opened 2023-09-24
feat(fmt): add significantParts option to duration format

ArmaanAS:
Added the `significantParts` option to the std/fmt/duration `format()` function. This can reduce the number of date parts returned, showing only the first N signficant parts. Added 4 test cases for this new option. 

## Example
```js
format(10_000_000)

"0d 2h 46m 40s 0ms 0µs 0ns"
```

With `significantParts: 2`
```js
format(10_000_000, {
  significantParts: 2
})

"0d 2h 46m"
```

With `ignoreZero: true`
```js
format(10_000_000, {
  ignoreZero: true,
  significantParts: 2
})

"2h 46m"
```

### Notes
No rounding is performed, only truncation. Perhaps rounding may be desirable as mathematically "significant figures" does round.

Removed seemingly redundant template literal strings surrounding `.join(" ")` functions. If anyone can explain why it was used then please leave a reply.

In my opinion, the `ignoreZero` option should be `true` by default. It makes the duration easier to read. This may be considered a breaking change though.

### DIFF
--- a/fmt/duration.ts
+++ b/fmt/duration.ts
@@ -102,29 +102,21 @@ export function format(
   switch (opt.style) {
     case "narrow": {
       if (opt.ignoreZero) {
-        return `${
-          durationArr.filter((x) => x.value).map((x) =>
-            `${x.value}${x.type === "us" ? "µs" : x.type}`
-          )
-            .join(" ")
-        }`;
+        return durationArr.filter((x) => x.value).map((x) =>
+          `${x.value}${x.type === "us" ? "µs" : x.type}`
+        ).join(" ");
       }
-      return `${
-        durationArr.map((x) => `${x.value}${x.type === "us" ? "µs" : x.type}`)
-          .join(" ")
-      }`;
+      return durationArr.map((x) =>
+        `${x.value}${x.type === "us" ? "µs" : x.type}`
+      ).join(" ");
     }
     case "full": {
       if (opt.ignoreZero) {
-        return `${
-          durationArr.filter((x) => x.value).map((x) =>
-            `${x.value} ${keyList[x.type]}`
-          ).join(", ")
-        }`;
+        return durationArr.filter((x) => x.value).map((x) =>
+          `${x.value} ${keyList[x.type]}`
+        ).join(", ");
       }
-      return `${
-        durationArr.map((x) => `${x.value} ${keyList[x.type]}`).join(", ")
-      }`;
+      return durationArr.map((x) => `${x.value} ${keyList[x.type]}`).join(", ");
     }
     case "digital": {
       const arr = durationArr.map((x) =>
--- a/fmt/duration.ts
+++ b/fmt/duration.ts
@@ -89,9 +89,11 @@ export interface PrettyDurationOptions {
   ignoreZero: boolean;
   /**
    * Number of significant parts to display.
-   * e.g. 1h 10m 30s 500ms with significantParts=2 will be displayed as 1h 10m.
+   * e.g. `"1h 10m 30s 500ms"` with `significantParts: 2` will be displayed as `"1h 10m"`.
+   * Removes leading and trailing zeros.
+   * Overrides `ignoreZero` option.
    */
-  significantParts?: number;
+  significantParts: number;
 }
 
 export function format(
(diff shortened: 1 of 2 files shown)


kt3k:
Thanks for the suggestion.

> In my opinion, the ignoreZero option should be true by default. It makes the duration easier to read. This may be considered a breaking change though.

I think the combination of `ignoreZero: true` and `significantPart > 1` has an issue when the 2nd (or higher) significantPart is 0.

How about making `significantParts > 0` cause stripping of leading 0s?

```js
format(10_000_000, {
  significantParts: 2
})

// returns "2h 46m" instead of "0d 2h 46m"
```

When we do the above stripping of leading 0, we can return `2h 0m` for `7200000`

```js
format(2 * 3600 * 1000, { significantParts: 2 }) // => may return `2h 0m`
format(2 * 3600 * 1000, { significantParts: 2, ignoreZero: true }) // => returns `2h`
```

ArmaanAS:
Yes, I think you're correct. `ignoreZero: true` removes all 0 values when it should really just remove leading zeroes. I think `ignoreZero` is a problematic option as a more suitable option would be `leadingZeros: true | false`. 

Additionally the behaviour of `ignoreZero` differs between the three styles. `"full"` and `"narrow"` behave the same in that all zeros are filtered out but `"digital"` only removes the trailing zeros:

### `style: "narrow"`

```js
format(10_000_000, {
  style: "narrow",
})

"0d 2h 46m 40s 0ms 0µs 0ns"

// Removes all 0 values
format(10_000_000, {
  style: "narrow",
  ignoreZero: true,
})

"2h 46m 40s"
```

### `style: "digital"`

```js
format(10_000_000, {
  style: "digital",
})

"00:02:46:40:000:000:000"

// Removes zeros at the end only
format(10_000_000, {
  style: "digital",
  ignoreZero: true,
})

"00:02:46:40"
```

## Proposal

Instead of alter this function further, we should deprecate it and create a new function called `formatDuration()` which takes the easier to understand `leadingZeros` and `trailingZeros` booleans instead of `ignoreZero`:

1. `style: "narrow" | "digital" | "full" = "narrow"`
2. `leadingZeros: boolean = false`
3. `trailingZeros: boolean = false`
4. `significantParts: number = -1`

The zeros options should be `false` by default to make the output shorter and easier to read. I believe that the shorter format will be the most desired, but that's up to debate.

For `style: "digital", the `leadingZeros` option will have no effect. It will always show leading zeros.

_Additionally I can fix the plural forms of the units, e.g. "1 minutes" should be "1 minute"_

kt3k:
I don't understand why `ignoreZeros` are problematic.

> Additionally the behaviour of ignoreZero differs between the three styles. "full" and "narrow" behave the same in that all zeros are filtered out but "digital" only removes the trailing zeros:

I think this is fine because `digital` style doesn't have units in its output. If it removes the leading zeros, there are no distinction between, for example, 2 hours and 2 minutes. Also this is explicitly documented in the API doc.

I think we should apply `significantParts` option only to `narrow` and `full` styles (It doesn't make sense to do this with `digital` in my view). We can document that in the API doc.

ArmaanAS:
I just meant it's problematic because of the issue you pointed out. If we add `trailingZeros` and `leadingZeros` options, they work better with the `significantParts` option to make an overall better `format()` function. What is your opinion on these 2 zeros options compared to `ignoreZero`? Personally it makes better sense to me because, removing zeros between the "significant" parts makes the formatted duration harder to read.

e.g.
### leadingZeros / trailingZeros

```javascript
{ 
  significantParts: 3, 
  leadingZeros: false, 
  trailingZeros: false,
}

// Easier to read / understand
"2 days, 0 hours, 5 minutes"
```

### ignoreZeros
```javascript
{ 
  significantParts: 3, 
  ignoreZero: true, 
}

// Harder to understand at a glance
"2 days, 5 minutes"
```

The decision is up to you (@kt3k). If you think ignoring all zeros is easier to use / read / understand then I'll go with that. Let me know what changes I should make.

kt3k:
> ### leadingZeros / trailingZeros
> ```js
> { 
>   significantParts: 3, 
>   leadingZeros: false, 
>   trailingZeros: false,
> }
> 
> // Easier to read / understand
> "2 hours, 0 minutes, 5 seconds"
> ```

Did the above mean the below?

```js
{ 
  significantParts: 3, 
  leadingZeros: true, 
  trailingZeros: true,
}

// Easier to read / understand
"2 hours, 0 minutes, 5 seconds"
```

My suggestion is that specifying `significantParts` automatically causes stripping of leading zeros and trailing zeros (without requiring `ignoreZero: true`).

```js
{ 
  significantParts: 3, 
  // no need of ignoreZeros
}
"2 hours, 0 minutes, 5 seconds"
```

That covers better the case when the significant parts have trailing zeros.

```js
{ 
  significantParts: 3,
}
"2 hours, 0 minutes, 0 seconds"
```
The last result (`2 hours, 0 minutes, 0 seconds`) can't be obtained when `leadingZeros` are true.

ArmaanAS:
As I've been trying to implement what you are suggesting, @kt3k, and it occured to me that the "digital" format is also a bit odd. It assumes that durations are large rather than small in the way that it removes trailing zeros rather than leading zeros.

e.g.

```javascript
format(5.5005, {
  style: "digital",
  ignoreZero: true,
})

"00:00:00:00:005:500:500"
```

If instead `ignoreZero` removed the leading zeros (as well as formatted microseconds and nanoseconds as a continued fraction rather than : colon separated values), it should return the following:

```javascript
"05.5005"
```

The digital format isn't very well defined and it's real-world use is alluding me. In which scenario would this current implementation be a useful display format?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/format

Today I discovered this proposed built-in duration format function for `Intl`. It also includes a `fractionalDigits` option similar to `significantParts`. Unfortunately this function is only supported in Safari which makes it hard for me to use it as a reference. 

It seems that it formats "digital" styles by keeping hours, minutes and seconds, milliseconds, e.t.c. but it includes more granular options for each datetime part. I tried to read the [tc39 specification](https://tc39.es/proposal-intl-duration-format/#sec-getdurationunitoptions) but I couldn't fully understand how they format digital durations.

Honestly, I'm starting to feel hesitant to work on this further because I feel increasingly pessimistic about it's usefulness. I've pushed the changes you suggested with an added test case, but I don't fully agree with this implementation of `format`.

Let me know if this works for you. Thanks.

lucacasonato:
Can we just use `Intl.DurationFormat` now?

iuioiua:
`Intl.DurationFormat` has not yet been implemented in the runtime, but we should pursue that instead. I've opened https://github.com/denoland/deno/issues/21450 to track that addition. For that reason, I'll close this PR without merging. Thank you very much for your contribution, either way, @ArmaanAS. It's much appreciated.

PR #3673, opened 2023-09-25
BREAKING(encoding): deprecate old encode/decode in hex.ts

lino-levan:
I realized that these were not deprecated in the last PR. Maybe an oversight? Might have missed something.


kt3k (approved):
LGTM. Thanks!

PR #3674, opened 2023-09-25
fix(semver): Add a necessary grouping

futsuuu:
This PR fixes an issue where if the prerelease starts and ends with a number, it is judged as a number.

before

```typescript
const version = "1.0.0-1b3" // prerelease: "1b3", but matches with /^0|[1-9]\\d*$/
parse(version).prerelease   // -> [ 1 ]
```
after

```typescript
const version = "1.0.0-1b3" // prerelease: "1b3", doesn't match with /^(0|[1-9]\\d*)$/
parse(version).prerelease   // -> [ "1b3" ]
```


kt3k (approved):
LGTM. Thanks for the fix.

kt3k:
Note: Ids like `1b3` should be considered alphanumeric identifier (because it includes `b` char), but currently it's handled like numeric identifier because of this error. This PR fixes it. ref. https://semver.org/#spec-item-9

PR #3682, opened 2023-09-30
perf base64url

mimiMonads:
Reasons for potential speed improvement:

 - Fewer String Parses: In the original function, the string is parsed three times due to three replace calls. In the new function, the string is parsed at most twice using replace, and even this is optimized using quick endsWith checks.

 - Optimized Checks for '=': The new function uses the endsWith method, which is optimized for checking the end of a string and is faster than scanning the entire string as in the replace method.

 - Slice Over Replace: The new function uses slice to remove trailing '=' characters. slice is generally faster for such operations since it simply involves adjusting a pointer/reference to the string start/end, whereas replace would have to parse through the string.

also, this operation is safe because is a base64 where `=` only can be at the end

It's proximally 15% faster than the current implementation

### DIFF
--- a/encoding/base64url.ts
+++ b/encoding/base64url.ts
@@ -50,9 +50,14 @@ function convertBase64urlToBase64(b64url: string): string {
   return addPaddingToBase64url(b64url).replace(/\-/g, "+").replace(/_/g, "/");
 }
 
-function convertBase64ToBase64url(b64: string): string {
-  return b64.replace(/=/g, "").replace(/\+/g, "-").replace(/\//g, "_");
-}
+
+const convertBase64ToBase64url = (b64:string) =>
+       b64.endsWith('=')
+        ? b64.endsWith('==')
+          ? b64.replace(/\+/g, "-").replace(/\//g, "_").slice(0, -2)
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

kt3k:
It might be also worth considering using a pattern like `b64.replace(/[\+\/\=]/g, (c) => {...})`

mimiMonads:
I already did, and it was slower, I had another good candidate

kt3k:
Oh, interesting. Thanks for the researches

ycmjason:
I have just sumble upon this piece of code and found it interesting and ended up at this PR.

FWIW... I have done some benchmark and have found that this actually stacks up pretty well with `slice`:

```
b64.replace(/=?=$/, "").replace(/\+/g, "-").replace(/\//g, "_")
```

https://jsbm.dev/Jz884zkOfva86

It is about 1% slower but I think it's a good tradeoff for a cleaner implementation.

mimiMonads:
Long story short, I usually work with deno, node and bun and after a lot of testing for an impl I just got to the conclusion that that was the fastest way , I mean, sure I don't mind it but I just wanted to share it with the std

ycmjason:
It's so cool you have found this approach! 🎉

PR #3685, opened 2023-10-03
fix(assert): fix swapping of multiline str diff

### LINKED ISSUE #3039: `assertEquals` shows an incorrect diff
**Describe the bug**

The diff shown by `assertEquals` is not correct if the number of newlines differs in the left and right hand side of the assertion, and if there are repeated lines in the left hand side.

**Steps to Reproduce**

```
import { assertEquals } from "https://deno.land/std@0.170.0/testing/asserts.ts";

assertEquals("bar\nHello\nbar\nWorld1", "Hello2\nWorld");
```

results in the following output:

```
error: Uncaught AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   bar\n
-   Hello\n
-   bar\n
-   World1
+   Hello2\n
+   World


        ^
    at assertEquals (https://deno.land/std@0.170.0/testing/asserts.ts:190:9)
    at file:///mnt/artemis/Projects/github.com/denoland/deno_std/assert.ts:3:1
```

**Expected behavior**

The diff should look like this:

```
error: Uncaught AssertionError: Values are not equal:


    [Diff] Actual / Expected


-   bar\n
-   Hello\n
-   bar\n
-   World1
+   Hello2\n
+   World


        ^
    at assertEquals (https://deno.land/std@0.170.0/testing/asserts.ts:190:9)
    at file:///mnt/artemis/Projects/github.com/denoland/deno_std/assert.ts:3:1
```
**Environment**

- OS: Ubuntu 20.04
- deno version: 1.29.1
- std version: 0.170.0

### LINKED ISSUE #3681: Assert diff swaps actual and expected when comparing multiline strings
**Describe the bug**

When `assertEquals` is used to compare a multiline string and the check fails, then the diff output confusingly swaps the order of actual and expected.

Screenshot 2023-09-29 at 14 05 02

**Steps to Reproduce**

```tsx
import { assertEquals } from "https://deno.land/std@0.203.0/assert/mod.ts";

Deno.test("multiline", () => {
  const actual = `<div>
  <div>
    <p>
    foo</p>
  </div>
  [object Object],barbaz</div>
</div>  
`;

  const expected = `<div>
  <div>
    <p>foo</p>
    barbaz
  </div>
</div>
`;
});
```

**Expected behavior**

It should not swap the order. It should mark actual as "actual" and expected as "expected".

**Environment**

- OS: [e.g. Ubuntu 20.04, MacOS 11]
- deno version: 1.37.1
- std version: 0.203.0

kt3k:
This PR fixes the bug of multiline str diff when there are more diff lines in `actual` than `expected`.

The multiline diff algorithm ([`diffstr`](https://github.com/denoland/deno_std/blob/4fe9fd6d3910589a6fbae2f5063981d3510c965a/assert/_diff.ts#L250)) performs diffing in 2 different ways: First it takes diff of the entire multiple string, comparing each line. Then it takes diff of words for each line over the combination of added (`expected`) lines and removed (`actual`) lines.

When computing word diffs for each line combination, the code swaps [the main array](https://github.com/denoland/deno_std/blob/4fe9fd6d3910589a6fbae2f5063981d3510c965a/assert/_diff.ts#L345) to use for the loop (probably for optimization), but it forgot to swap [diffing target](https://github.com/denoland/deno_std/blob/4fe9fd6d3910589a6fbae2f5063981d3510c965a/assert/_diff.ts#L353-L356). It resulted in swapped diff when there are more removed (`actual`) lines.

closes #3039 
closes #3681

ref #948

### DIFF
--- a/assert/_diff.ts
+++ b/assert/_diff.ts
@@ -342,18 +342,21 @@ export function diffstr(A: string, B: string) {
   }
 
   // Compute word-diff
-  const aLines = added.length < removed.length ? added : removed;
-  const bLines = aLines === removed ? added : removed;
+  const hasMoreRemovedLines = added.length < removed.length;
+  const aLines = hasMoreRemovedLines ? added : removed;
+  const bLines = hasMoreRemovedLines ? removed : added;
   for (const a of aLines) {
     let tokens = [] as Array<DiffResult<string>>,
       b: undefined | DiffResult<string>;
     // Search another diff line with at least one common token
     while (bLines.length) {
       b = bLines.shift();
-      tokens = diff(
+      const tokenized = [
         tokenize(a.value, { wordDiff: true }),
         tokenize(b?.value ?? "", { wordDiff: true }),
-      );
+      ] as string[][];
+      if (hasMoreRemovedLines) tokenized.reverse();
+      tokens = diff(tokenized[0], tokenized[1]);
       if (
         tokens.some(({ type, value }) =>
           type === DiffType.common && value.trim().length
--- a/assert/_diff_test.ts
+++ b/assert/_diff_test.ts
@@ -314,3 +314,32 @@ Deno.test({
     ]);
   },
 });
+
+Deno.test({
+  name: "multiline diff with more removed lines",
+  fn() {
+    const diffResult = diffstr("a\na", "e");
+    assertEquals(diffResult, [
+      {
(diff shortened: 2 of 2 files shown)


bartlomieju (approved):
LGTM, thanks for fixing this!

PR #3686, opened 2023-10-04
fix(archive/untar.ts): Cannot access symlinks in archives

### LINKED ISSUE #3232: Cannot access symlinks in archives
When traversing the contents of a Tar Archive, there doesn't appear to be a way to access the reference to which a sym link points. The TarEntry structure indicates that the type is `symlink`, but there is no other metadata with information about where the sym link points.

```
import { join } from "https://deno.land/std@0.171.0/path/mod.ts";
import { ensureDir, ensureFile } from "https://deno.land/std@0.171.0/fs/mod.ts";
import { copy } from "https://deno.land/std@0.171.0/streams/mod.ts";
import { Untar } from "https://deno.land/std@0.171.0/archive/mod.ts";

export async function extractTarArchive(archive: string, destination: string) {
  const reader = await Deno.open(archive, { read: true });
  const untar = new Untar(reader);
  for await (const entry of untar) {
    if (entry.type === "symlink") {
      console.log(entry);
      continue;
    }
    if (entry.type === "directory") {
      await ensureDir(join(destination, entry.fileName));
      continue;
    }
    await ensureFile(join(destination, entry.fileName));
    const file = await Deno.open(join(destination, entry.fileName), {
      write: true,
    });
    await copy(entry, file);
  }
}

extractTarArchive("tarball.tar", "output_folder");
```

Here is a snippet that will attempt to read the contents of a Tar Archive. If it encounters a `symlink` it prints the metadata, but there doesn't appear to be a way to see what the sym link points too. In the case of a symlink I see this:

```
TarEntry {
  fileName: "folder/linked_file.txt",
  fileMode: 493,
  mtime: 1677876636,
  uid: 501,
  gid: 20,
  owner: "irbull",
  group: "staff",
  type: "symlink",
  fileSize: 0
}
```
This `link_file.txt` should point to `file.txt`. I've attached a tar archive that includes a sym link. (sorry, I had to zip the tar archive, as GitHub doesn't accept `.tar` files.)

[tarball.tar.zip](https://github.com/denoland/deno/files/10886124/tarball.tar.zip)

cwirving:
Fixes #3232

Added a `linkName` optional property to class `TarEntry` so that the destination of symbolic link entries can be read. The entry is optional so that existing code that processes files, directories, etc. don't see an unexpected new property, but the property is set for entries of type "symlink".

The change is minimal so that existing tests are unchanged but a new test that extracts an archive containing a file and a symbolic link to this file demonstrates the new behavior.

A couple of existing lines were reformatted by `deno fmt`, but I assume that we want all files formatted...


kt3k (approved):
LGTM. Thanks!

PR #3687, opened 2023-10-04
refactor: remove top-level arrow sytnax use

iuioiua:
My main concern is whether anything is broken in `std/crypto`.

See https://docs.deno.com/runtime/manual/references/contributing/style_guide#top-level-functions-should-not-use-arrow-syntax


lino-levan:
Does this have an impact on user code at all?

iuioiua:
I don't think so. If anything, it might be `std/crypto`.

lino-levan:
I hunted down the original source for the code in crypto and [it's actually been there the entire time](https://github.com/denoland/deno_std/pull/1025/files). My main concern had been this, since it looks like code that might come out of a bug fix. It turns out all of the binding was done ahead of time? I'm not sure how much of that was necessary. If it turned out to indeed have been necessary which is very possible, we should add tests.

iuioiua:
@jeremyBanks, are you able to comment on whether the changes to `std/crypto` here are reasonable and safe?

jeremyBanks:
@iuioiua :wave: Hello! It's been a while so my apologies if I'm getting mixed-up, but I believe there are a couple problems with this change as currently written, though they unfortunately aren't caught by the test suite.

1. The change breaks the re-export of the built-in crypto methods that we don't modify/wrap with further logic.

Current behaviour on `main`:

```
$ deno eval 'console.log((await import("./crypto/crypto.ts")).crypto)'
{
  getRandomValues: [Function: bound getRandomValues],
  randomUUID: [Function: bound randomUUID],
  subtle: {
    decrypt: [AsyncFunction: bound decrypt],
    deriveBits: [AsyncFunction: bound deriveBits],
    deriveKey: [AsyncFunction: bound deriveKey],
    digest: [AsyncFunction: digest],
    encrypt: [AsyncFunction: bound encrypt],
    exportKey: [AsyncFunction: bound exportKey],
    generateKey: [AsyncFunction: bound generateKey],
    importKey: [AsyncFunction: bound importKey],
    sign: [AsyncFunction: bound sign],
    unwrapKey: [AsyncFunction: bound unwrapKey],
    verify: [AsyncFunction: bound verify],
    wrapKey: [AsyncFunction: bound wrapKey],
    digestSync: [Function: digestSync],
    timingSafeEqual: [Function: timingSafeEqual]
  }
}
```

Behaviour on this PR branch:

```
$ deno eval 'console.log((await import("./crypto/crypto.ts")).crypto)'
{
  subtle: {
    digest: [AsyncFunction: digest],
    digestSync: [Function: digestSync],
    timingSafeEqual: [Function: timingSafeEqual],
    [Symbol("[[webidl.brand]]")]: Symbol("[[webidl.brand]]")
  },
  [Symbol("[[webidl.brand]]")]: Symbol("[[webidl.brand]]")
}
```

Most properties/methods of the global `crypto` object and its child `crypto.subtle` are non-enumerable, so attempting to `...` spread them directly into the `stdCrypto` object causes them to be ignored, while the existing code named them all explicitly in the `webCrypto` object (from which they were enumerable).

If I recall correctly, this isn't covered by existing tests because I only thought to add tests where we had our own logic, not where we were just re-exporting an existing method unmodified (except for the binding). At the time I initially wrote this code, I _think_ that Deno's native WebCrypto implementation wasn't complete, so asserting that many of them existed would have failed.

This issue could be fixed by listing them all out explicitly without the binding, however...

2. The native `crypto` and `crypto.subtle` methods throw an error if they're not called with the appropriate receiver.
 
See https://github.com/denoland/deno/blob/2167a52d693e79c661f91d29164f59111f3bc3c9/ext/crypto/00_crypto.js#L4683C5-L4683C25 for an example of this check in Deno's WebCrypto implementation. (This module is also marked as browser-compatible, and browsers also enforce this, although from a brief search I'm not sure where this behaviour is specified.) This is why they were all explicitly bound. I think the tests miss this for the same reason: we're only testing methods where we have our own additional logic, so and if those pass-through to the built-in implementations their logic is already making sure the appropriate receiver is used.

---

Both of these failures could be caught by adding a test case like the following, which passes on `main`:

```typescript
Deno.test("[crypto/getRandomValues] pass-through to native crypto.getRandomValues()", () => {
  stdCrypto.getRandomValues(new Uint8Array(1));
});
```

On your branch as-written, it fails because the native method isn't re-exported:

```
[crypto/getRandomValues] pass-through to native crypto.getRandomValues() => ./crypto/crypto_test.ts:1422:6
error: TypeError: stdCrypto.getRandomValues is not a function
  stdCrypto.getRandomValues(new Uint8Array(1));
            ^
    at file:///mnt/c/Users/_/denoland/deno_std/crypto/crypto_test.ts:1423:13
```

If I tweak your branch to explicitly re-export that method, it then fails because of the incorrect receiver due to the lack of binding:

```
[crypto/getRandomValues] pass-through to native crypto.getRandomValues() => ./crypto/crypto_test.ts:1422:6
error: TypeError: Illegal invocation
  stdCrypto.getRandomValues(new Uint8Array(1));
            ^
    at Module.assertBranded (ext:deno_webidl/00_webidl.js:1011:11)
    at Object.getRandomValues (ext:deno_crypto/00_crypto.js:4683:12)
    at file:///mnt/c/Users/_/denoland/deno_std/crypto/crypto_test.ts:1423:13
```

iuioiua:
Thank you, @jeremyBanks. Very well explained. I've reverted the breaking changes to `std/crypto`.

PR #3692, opened 2023-10-11
fix(fs): `expandGlob`/`expandGlobSync` don't require full `--allow-read` perms on granted read permissions

### LINKED ISSUE #2652: `expandGlob` always requires read access to CWD
Regardless of the root specified, running `expandGlob` will prompt for `--allow-read` access to the CWD.

main.ts:
```
import {expandGlob} from 'https://deno.land/std@0.95.0/fs/expand_glob.ts';

for await (const entry of expandGlob('**/*.js', {root: './subfolder'})) {
	console.log(entry);
}
```

Running `deno run main.ts --allow-read=./subfolder`

will produce:

```
⚠️  ️Deno requests read access to <CWD>. Run again with --allow-read to bypass this prompt.
   Allow? [y/n (y = yes allow, n = no deny)]  n
error: Uncaught PermissionDenied: Requires read access to <CWD>, run again with the --allow-read flag
    : joinGlobs([Deno.cwd(), root], globOptions);
                      ^
    at Object.opSync (deno:core/01_core.js:172:12)
    at Object.cwd (deno:runtime/js/30_fs.js:65:17)
    at expandGlob (https://deno.land/std@0.95.0/fs/expand_glob.ts:87:23)
    at expandGlob.next (<anonymous>)
    at file:///C:/path/to/main.ts:5:18
```

Swapping both or either of the root or --allow-read paths for absolute paths produces a similar permission denial:

```
⚠️  ️Deno requests read access to "C:\path\to\subfolder". Run again with --allow-read to bypass this prompt.
   Allow? [y/n (y = yes allow, n = no deny)]  n
error: Uncaught (in promise) PermissionDenied: Requires read access to "C:\path\to\subfolder", run again with the --allow-read flag
  const info = await Deno.stat(path);
               ^
    at async Object.stat (deno:runtime/js/30_fs.js:230:17)
    at async _createWalkEntry (https://deno.land/std@0.95.0/fs/walk.ts:25:16)
    at async expandGlob (https://deno.land/std@0.95.0/fs/expand_glob.ts:108:21)        
    at async file:///C:/path/to/main.ts:5:18        
```

**Expected behavior**

Using a root path that is within the boundary granted by `--allow-read` should just work, it should not prompt for read permissions of the entire CWD.

**Environment**

- OS: Windows 11
- deno version: 1.22.2
- std version: 0.95.0

lowlighter:
This makes the following changes to `expandGlob`/`expandGlobSync`:
- `options.root` now defaults to `Deno.cwd()` ***only after*** checking that it was not provided, meaning that full `--allow-read` permissions is not required anymore when passing both a pattern and a root (unless the root is not absolute, meaning that currently it still needs to use path resolution)
- The `isAbsoluteGlob` initialization has been moved earlier to make it possible to defaults `options.root` to `/` (or windows root equivalent if defined) when an absolute glob pattern has been detected, meaning once again full `--allow-read` permissions are not needed when just passing only an absolute pattern
- On Windows, for absolute globs there is a resolution needed even for root path because the `resolve` tries to check if the cwd has a different drive name or not. To mitigate this and avoid breaking changes, since we still have access to `resolvedDevice` (and there was already an handling for `path === undefined`) it lets users using denying read permissions while still be able to run the function 
```ts
//test.ts
import { expandGlobSync } from "./fs/expand_glob.ts";

console.log("1. Specified root")
console.log(...[...expandGlobSync("ab*", {root:"/workspaces/deno_std/fs/testdata/glob"})].map(({name}) => name))

console.log("2. Absolute pattern")
console.log(...[...expandGlobSync("/workspaces/deno_std/fs/testdata/glob/ab*")].map(({name}) => name))
```

```markdown
➜ /workspaces/deno_std $ deno run --no-prompt --allow-read=/workspaces/deno_std/fs/testdata/glob test.ts
1. Specified root
abc abcdef abcdefghi
2. Absolute pattern
abc abcdef abcdefghi
```

Closes #2652


path/windows/resolve.ts
```diff
@@ -32,7 +32,14 @@ export function resolve(...pathSegments: string[]): string {
...
       }
-      path = Deno.cwd();
+      try {
+        path = Deno.cwd();
+      } catch (error) {
+        if (!(error instanceof Deno.errors.PermissionDenied)) {
+          throw error;
+        }
+        path = undefined as unknown as string;
+      }
```
kt3k, line 42: `resolve("c:")` now resolves to `c:\` instead of throwing permission error. I think this is a breaking change to `windows/resolve`

Can we avoid changing `resolve` here, but only making changes to expandGlob?

lowlighter: Seems like using directly the root if `isGlobAbsolute` is true instead of `resolve(root)` do the trick

kt3k (approved):
LGTM. Thanks!

PR #3698, opened 2023-10-13
fix(msgpack): encode huge objects #3696

### LINKED ISSUE #3696: msgpack.encode fails with `RangeError: Maximum call stack size exceeded` on big objects
**Describe the bug**
My app may receive huge jsons (with thousands of items and very complex structure) and shall re-encode the json into msgpack. 
JSON.parse works perfectly fine but msgpack.encode fails with `Uncaught RangeError: Maximum call stack size exceeded`.

**Steps to Reproduce**
Here is a simple code to reproduce it:

```ts
const bigObject = {};

for (let i = 0; i < 50000; i++) {
  Object.defineProperty(bigObject, `prop_${i}`, {
    value: i,
    enumerable: true
  });
}

import * as msgpack from "https://deno.land/std@0.204.0/msgpack/mod.ts";
const m = msgpack.encode(bigObject);
```

**Expected behavior**
Should not fail and return the encoded Uint8Array.

**Environment**

- OS: [WSL-Ubuntu 22.04]
- deno version: `{ deno: "1.37.0", v8: "11.8.172.3", typescript: "5.2.2" }`
- std version: 0.204.0

pavel-lechenko:
This PR resolves #3696

Also required to adjust `concat` function so that it also supports `Uint8Array[][]`.


bytes/concat.ts
```diff
@@ -8,18 +8,32 @@
  * const a = new Uint8Array([0, 1, 2]);
  * const b = new Uint8Array([3, 4, 5]);
  * console.log(concat(a, b)); // [0, 1, 2, 3, 4, 5]
+ * ```
  */
-export function concat(...buf: Uint8Array[]): Uint8Array {
+export function concat(...buf: Uint8Array[] | Uint8Array[][]): Uint8Array {
```
lino-levan, line 13: This isn't how I would accomplish this overload. I'd probably have two possible signatures: One `buffer: Uint8Array[]` and one `...buf: Uint8Array[]`.

lino-levan (changes requested):

bytes/concat.ts
aapoalas: Agreed: Check if length is one and then check type:
* Array? Presume array of Uint8Arrays and concat them directly.
* Uint8Array? Return directly as concating a single item requires no copy work.

pavel-lechenko: The approach of having such overloads causes signature incompatibility. Do you have an idea how to avoid the conflict?

Also Microsoft recommends using the union as a good practice in case of multiple types: https://www.typescriptlang.org/docs/handbook/2/functions.html#writing-good-overloads

pavel-lechenko: Another approach is to copy-paste `concat` into msgpack/encode.ts so it accepts a single argument of type `Uint8Array[]`. 
What do you think?

pavel-lechenko: Like this:
```ts
export function encode(object: ValueType) {
  const byteParts: Uint8Array[] = [];
  encodeSlice(object, byteParts);
  
  // Concat byteParts into output 
  let length = 0;
  for (const b of byteParts) {
    length += b.length;
  }
  const output = new Uint8Array(length);
  let index = 0;
  for (const b of byteParts) {
    output.set(b, index);
    index += b.length;
  }
  return output;  
}
```

aapoalas: A possible union typing is `Uint8Array[] | [Uint8Array[]]`.

pavel-lechenko: @aapoalas implemented per your suggestion together with tests. 
Used `(Uint8Array | Uint8Array[])[]` for better readability and LSP support.

bytes/concat.ts
iuioiua, line 1: What about factoring the new logic into a `concatDeep()`? It might be a bit easier to follow.

iuioiua: On top of this, can you please run `deno fmt`?

bytes/concat.ts
```diff
@@ -8,18 +8,41 @@
  * const a = new Uint8Array([0, 1, 2]);
  * const b = new Uint8Array([3, 4, 5]);
  * console.log(concat(a, b)); // [0, 1, 2, 3, 4, 5]
+ * ```
  */
-export function concat(...buf: Uint8Array[]): Uint8Array {
+export function concat(...buf: (Uint8Array | Uint8Array[])[]): Uint8Array {
```
lino-levan, line 13: Could you overload these operators like so?
```suggestion
export function concat(buf: Uint8Array[]): Uint8Array;
export function concat(...buf: Uint8Array[]): Uint8Array;
export function concat(...buf: (Uint8Array | Uint8Array[])[]): Uint8Array {
```

lino-levan (changes requested):

iuioiua:
@pavel-lechenko, PTAL at https://github.com/denoland/deno_std/issues/3696#issuecomment-1776424324

bytes/concat.ts
pavel-lechenko: Sure, but in this case it won't work for multiple arrays of Uint8Array. So it will support only one array.

I will make a change as suggested.

iuioiua:
I've played around with this, and it seems that all that's needed is to remove the spread operator from `concat()` and add the new tests for `concat()` and `encode()`. IIUC, `concat()` doesn't need to accept anything more than `Uint8Array[]`, as all tests pass fine.

lino-levan (approved):
I think this LGTM. Obviously not the only way to solve the problem.

aapoalas (approved):
LGTM as well.

iuioiua:
@pavel-lechenko, please see my previous comment.

kt3k (approved):
LGTM

kt3k:
@iuioiua Do you suggest removing spread operator? I think we should always avoid breaking changes if possible.

iuioiua:
Yes, I don't think `concat()` needs to cater for arrays deeper than 1. Having it accept an array makes much more sense to me. I think it's better to make the breaking change than complicating it.

kt3k:
@pavel-lechenko Thanks for your contribution!

We'll continue working on @iuioiua's point in later PRs.

PR #3699, opened 2023-10-15
refactor: move unstable `std/crypto` files

iuioiua:
To match the changes in #3661


crypto/keystack.ts
```diff
@@ -1,7 +0,0 @@
-// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-// This module is browser compatible.
-
-/**
- * @deprecated (will be removed after 0.209.0) Import from `std/crypto/unstable/keystack.ts` instead.
- */
-export * from "./unstable/keystack.ts";
```
kt3k, line 7: Looks like we still need to keep this file

iuioiua: Whoops. Fixed!

kt3k (approved):
LGTM. Thanks!

PR #3701, opened 2023-10-15
docs: remove `@$STD_VERSION` references

iuioiua:
Including this in the import specifier seems redundant.

### DIFF
--- a/_tools/check_doc_imports.ts
+++ b/_tools/check_doc_imports.ts
@@ -45,7 +45,7 @@ function checkImportStatements(
     const importPath = (moduleSpecifier as ts.StringLiteral).text;
     const isRelative = importPath.startsWith(".");
     const isInternal = importPath.startsWith(
-      "https://deno.land/std@$STD_VERSION/",
+      "https://deno.land/std/",
     );
     const { line } = sourceFile.getLineAndCharacterOfPosition(
       moduleSpecifier.pos,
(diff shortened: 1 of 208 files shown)


lino-levan:
What's the upside to doing this? This will unpin all versions (and I think it will mess up `deno check`?

iuioiua:
Perhaps a better question is, "What is the upside of including `@$STD_VERSION` in the import specifiers?" If there's no justifiable answer, there's no reason to include it. I assume it was initially included to remind devs to consider the version of the Standard Library they require. Everyone knows about module versions when importing dependencies. This doesn't need to be re-iterated in every import specifier in every JSDoc example. I doubt removing this string will cause any issues.

A newcomer might also see the `@$STD_VERSION` and wonder if it means something beyond a placeholder for the Standard Library version. I know I did.

It won't mess up `deno check`, as the `imports` field in `deno.json` was also adjusted to match.

cknight:
"Everyone knows about module versions when importing dependencies" is a bold statement which I don't think will hold true.  Many beginners to ESM, Deno and javascript in general won't understand the concept of module pinning.  This change will move the documentation from:

```ts
import { Tar } from "https://deno.land/std@0.204.0/archive/tar.ts";
//code example using Tar
```
to
```ts
import { Tar } from "https://deno.land/std/archive/tar.ts";
//code example using Tar
```
This will encourage many people to use unpinned versions after they copy and paste the code, when module pinning is best practice.

iuioiua:
> This will encourage many people to use unpinned versions after they copy and paste the code, when module pinning is best practice.

Actually, that might be a justifiable answer.

PR #3703, opened 2023-10-15
fix: leaky `cancel CsvParseStream during iteration does not leak file` test

### LINKED ISSUE #3694: `[csv/csv_parse_stream] cancel CsvParseStream during iteration does not leak file` is flaky
https://github.com/denoland/deno_std/actions/runs/6479639086/job/17593573910

iuioiua:
Frankly, I just applied the fix suggested by the CLI without much thought. Let's see if it works.

Fixes #3694


iuioiua:
So I guess it doesn't work... I'll play around a little.

iuioiua:
Ok, this didn’t work. Closing.

PR #3705, opened 2023-10-15
chore: remove deprecated `restrictEnvAccessTo` option from `std/dotenv/load`

iuioiua:


lino-levan (approved):
LGTM

kt3k (approved):
LGTM. Thanks!

kt3k:
Note: Now dotenv only accesses to the env vars which are necessary to read `.env` file, and this option is not necessary. See https://github.com/denoland/deno_std/pull/3578 for more details

PR #3706, opened 2023-10-16
chore: add `check:types` task

iuioiua:


kt3k:
Most of files are type-checked by `deno test` command (via `deno task test`). The reason why we have `lint:tools-types` is that the types of tools are not checked by `deno test` (because they are not imported from any test cases)

iuioiua:
I don't see any indication in the `test` task or `deno test` sub-command that types are checked. Maybe you're referring to the `--doc` flag, which type-checks JSDoc code blocks. This task checks all types in all implementations.

kt3k:
See this blog post https://deno.com/blog/v1.23#no-type-checking-by-default. `deno test` checks types for local typescript files.

kt3k:
The type error in CI seems happening because unrelated tool dependency is included and that affects the global types.

The below command shows the same error
```
deno check --unstable testing/time.ts console/_tools/compare_with_rust.ts
```

iuioiua:
The check for `console/_tools/compare_with_rust.ts` might be failing because `--unstable` isn't being used in the check. I don't understand the reasoning for `testing/time.ts`. I think it's just that `fakeClearTimeout()` doesn't match all of the overloads of `globalThis.clearTimeout()`. We can discuss these later.

Either way, `deno test` automatically type-checking is good enough to warrant closing this PR. Thanks!

PR #3709, opened 2023-10-19
feat(semver): canParse

### LINKED ISSUE #3707: SemVer: Method returning boolean for validation
I used to use `semver.valid` to check if a string is a valid semver. Since the deprecation of the old library, is there a similar method which returns a boolean? I prefer Booleans compared to TypeErrors, what `semver.parse` throws.

If not, can we add one?

Thank you!

lino-levan:
closes #3707


iuioiua (approved):
LGTM

lino-levan:
This flaky test makes me sad :(

timreichen:
How about just testing against a regexp?

lino-levan:
> How about just testing against a regexp?

Unfortunately it's not that simple (there are a lot of other things to test). This is, in my opinion, the cleanest way of doing it. Anything else would require a lot of duplicated logic.

kt3k (approved):
LGTM

PR #3710, opened 2023-10-21
chore: remove optional `std/console` tests

iuioiua:
My thinking is that these tests being optional means it likely fine for them not to be included at all in the codebase. Removing these tests has no effect on the user or regular testing.


kt3k:
I think these tools can be useful for future development or investigation of potential issues

iuioiua:
If they might be helpful in the future, I suggest we make them part of the test suite. If I understand correctly, this would incur the requirement of Rust for tests, which might be cumbersome.

I can see optional tests like these being forgotten in the future and raising questions by code readers. We could reference these tests somewhere in the documentation if ever required. But given that the tests are optional, it's tough to say they'll ever be needed.

kt3k:
I don't see the issue of keeping them as a reference.

> We could reference these tests somewhere in the documentation if ever required.

Maybe let's consider these as documentation about how to compare our implementation with the Rust crates.

BTW `console/_tools/generate_data.ts` is the generator of `console/_data.json`, which will be necessary when we optimize or refactor the structure of that file in the future.

PR #3711, opened 2023-10-22
BREAKING(path): split up glob into multiple files

lino-levan:
Closes https://github.com/denoland/deno_std/issues/3600

Please take a careful look, I might've missed something.


lino-levan:
Looks like several flaky tests :/

Any idea why these are happening now @iuioiua?

iuioiua:
I don’t know the fix. I’ve asked Leo if he has any ideas.

path/glob.ts
```diff
@@ -1,28 +1,26 @@
...
 
-export type GlobToRegExpOptions = GlobOptions & {
-  os?: OSType;
-};
+/**
+ * @deprecated (will be removed in 0.215.0) Import from "std/path/windows/is_glob.ts"
+ *
+ * Test whether the given string is a glob
+ */
+export const isGlob = _isGlob;
```
iuioiua, line 18:
```suggestion
export {
  /**
   * @deprecated (will be removed in 0.215.0) Import from "std/path/windows/is_glob.ts"
   *
   * Test whether the given string is a glob
   */
  isGlob,
} from "./is_glob.ts";
```
I think this approach is cleaner. Ditto for the other instances.

iuioiua (changes requested):

path/glob.ts
iuioiua: Yep 👍🏾

iuioiua (commented):
None of these glob functions are platform-specific. Their implementations are identical and duplicated across both platforms. They should be split up at the top level of `std/path` but not in platform-specific folders. Am I missing something here?

iuioiua:
Oh, I see they use different `normalize()` functions... Nevermind. Please allow a little time for me to comb through this.

kt3k (approved):
LGTM

iuioiua (approved):
LGTM. However, there's some duplicated code between the two functions. E.g. `normalizeGlob()` are essentially the same on both platforms, except that they rely on different underlying `normalize()` and `SEP_PATTERN` symbols. Let's discuss options for making this better in an issue.

kt3k:
@lino-levan Thank you for your contribution!

PR #3717, opened 2023-10-24
fix: flaky `http/file_server.ts` tests

### LINKED ISSUE #3716: bug: `serveDirIndex TLS` and `partial TLS arguments fail` tests are flaky
See https://github.com/denoland/deno_std/actions/runs/6625309968/job/17996042005?pr=3715

iuioiua:
Closes #3716


http/file_server_test.ts
```diff
@@ -540,31 +545,22 @@ Deno.test(
...
+      client.close();
+
+      assertStringIncludes(await res.text(), "<title>Deno File Server</title>");
     } finally {
       await killFileServer();
     }
   },
 );
 
 Deno.test("partial TLS arguments fail", async function () {
```
kt3k, line 562: I don't think we need to change this test case because according to this run log https://github.com/denoland/deno_std/actions/runs/6625309968/job/17996042005?pr=3715 this test case was affected by the one above (i.e. the flakiness wasn't caused by this test case itself)

iuioiua: I understand. However, passing by, I realised the test could rely on more reliable APIs. Hence, the change.

kt3k:
flakiness happened in other places, but this one looks fine to me.

kt3k (approved):
LGTM

PR #3722, opened 2023-10-26
refactor: remove unneeded `assertPath()`

iuioiua:
This function is redundant since we're using TypeScript.


kt3k:
I don't agree with it. We still support JavaScript users. These assertions are useful for them.

Also even when the user uses TypeScript, they could give wrong types to their code (for example, one can easily do `JSON.parse(json) as WrongType`). In that case runtime assertions are useful.

iuioiua:
I talked with Yoshiya. A change like this will require more discussion. Closing for now.

PR #3723, opened 2023-10-26
refactor: cleanup `assertArg/s()`

iuioiua:
These functions aren't actually doing anything useful.


iuioiua:
This change will require more discussion.

Issue #3718, opened 2023-10-25
tracking: cleanup deprecated APIs


iuioiua:
Upon deprecating an API, removing its use throughout the Standard Library is best. This exposes the implications of the deprecation and is an exercise in what the migration will involve for devs using the API. I think it's best to have this done before removal to weed out any potential issues early. These are some of the deprecated APIs I've noticed that are still in use throughout the Standard Library:

- [x] [`deferred()`](https://deno.land/std@0.205.0/async/mod.ts?s=deferred) (which was deprecated in #3758)
- [x] [`BytesList()`](https://deno.land/std@0.204.0/bytes/mod.ts?s=BytesList) (to be removed in 0.205.0, but is being used in `readDelim()`, which is being removed in v1.0.0)
- [x] [`toHashString()`](https://deno.land/std@0.204.0/crypto/mod.ts?s=toHashString) (in JSDocs)
- [x] [`StringReader()`](https://deno.land/std@0.204.0/io/mod.ts?s=StringReader) (used in std/csv test, which is stable)
- [x] [`typeByExtension()`](https://deno.land/std@0.204.0/media_types/mod.ts?s=typeByExtension) (used in std/media_type, which is stable)
- [x] [`iterateReader()`](https://deno.land/std@0.204.0/streams/mod.ts?s=iterateReader) (used in std/http, which is stable)
- [x] [`readableStreamFromReader()`](https://deno.land/std@0.204.0/streams/mod.ts?s=readableStreamFromReader) (used in std/csv test, which is stable)

Note: I've done my best only to include deprecated APIs used in other stable APIs or deprecated APIs with some time till removal. Please let me know if I've missed any or included some needlessly.

0scvr:
I have gone through the code quickly and it seems that most of these functions are only used in deprecated functions/types (therefore not needing to be replaced) or in test files.

iuioiua:
Yes, I've included test cases deliberately. Either way, to be sure, I double-checked my list, added reasoning behind each inclusion, and removed `readerFromIterable()` and `writeAllSync()`. Thanks for checking!

git001:
I'm getting confused when I tak a look into `Deno.Buffer` [Deno.Buffer](https://deno.land/api@v1.37.2?s=Deno.Buffer) which links to [Buffer](https://deno.land/std@0.204.0/io/mod.ts?s=Buffer) and this mention this.

> Use Web Streams instead

[...]

PR #3725, opened 2023-10-26
chore: remove `readableStreamFromReader()` & `StringReader()` in CSV stream test

0scvr:
Removed the use of `readableStreamFromReader` and `StringReader` in csv stream parse tests since they are deprecated.
See #3718

### DIFF
--- a/csv/csv_parse_stream_test.ts
+++ b/csv/csv_parse_stream_test.ts
@@ -2,7 +2,6 @@
 import { CsvParseStream } from "./csv_parse_stream.ts";
 import type { CsvParseStreamOptions } from "./csv_parse_stream.ts";
 import { ERR_QUOTE, ParseError } from "./_io.ts";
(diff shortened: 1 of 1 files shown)


csv/csv_parse_stream_test.ts
```diff
@@ -351,9 +349,7 @@ x,,,
 });
 
 function createReadableStreamFromString(s: string): ReadableStream<string> {
-  return readableStreamFromReader(new StringReader(s)).pipeThrough(
-    new TextDecoderStream(),
-  );
+  return ReadableStream.from(s);
 }
```
iuioiua, line 353: We can remove the use of `createReadableStreamFromString()` and just use `ReadableStream.from()` instead.

iuioiua (changes requested):

iuioiua (approved):
LGTM. Thank you, Oscar 🙏🏾

PR #3726, opened 2023-10-26
chore: remove deprecated `BytesList()` in `readDelim()`

0scvr:
Removed use of deprecated `BytesList` in `readDelim` in favour of  `Uint8Array`.

See #3718

### DIFF
--- a/io/read_delim.ts
+++ b/io/read_delim.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { concat } from "https://deno.land/std@$STD_VERSION/bytes/concat.ts";
+import { concat } from "../bytes/concat.ts";
 import type { Reader } from "../types.d.ts";
 
 /** Generate longest proper prefix which is also suffix array. */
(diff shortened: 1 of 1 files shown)


iuioiua:
Hm, I would’ve thought `concat()` would be needed here. WDYT, @kt3k?

io/read_delim.ts
```diff
@@ -48,15 +47,15 @@ export async function* readDelim(
...
       // Yield last chunk.
-      yield chunks.concat();
+      yield chunks;
       return;
     } else if (result < 0) {
       // Discard all remaining and silently fail.
       return;
     }
-    chunks.add(inspectArr, 0, result);
+    chunks = new Uint8Array([...chunks, ...inspectArr.slice(0, result)]);
```
iuioiua, line 56: Let's use `concat()` here.

iuioiua (changes requested):

io/read_delim.ts
0scvr: Ah yes, concat would be a lot simpler

iuioiua (approved):
I made a tweak, and now, LGTM! Thank you.

PR #3727, opened 2023-10-26
chore: replace deprecated `typeByExtension()` in `content_type.ts`

0scvr:
Replaced call to deprecated `typeByExtension` in `content_type.ts` in favour of private `typeByExtension` function.

See #3718 

Not sure if this is what you were looking for. Seems odd to deprecate typeByExtension when it is needed by content type and would need to be reimplemented for content type when removed.

### DIFF
--- a/media_types/content_type.ts
+++ b/media_types/content_type.ts
@@ -2,10 +2,10 @@
 // This module is browser compatible.
 
 import { parseMediaType } from "./parse_media_type.ts";
-import { typeByExtension } from "./type_by_extension.ts";
 import { getCharset } from "./get_charset.ts";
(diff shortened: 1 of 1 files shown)


iuioiua:
Ah, these were the types of issues I was hoping to catch when I first created the issue. I’m not certain what to do here.

Either way, can you please fix the failing tests?

0scvr:
> Ah, these were the types of issues I was hoping to catch when I first created the issue. I’m not certain what to do here.
> 
> Either way, can you please fix the failing tests?

Yes, my mistake. I forgot to import types.

iuioiua (approved):
On second thought, LGTM! Thank you, Oscar.

PR #3731, opened 2023-10-27
docs: re-write README

iuioiua:
These changes aim to make the README more concise, thorough, and friendlier to newcomers. I'm not strongly opinionated on most changes, so other opinions are welcome.

Further improvements on the README will come in the following PRs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,13 +13,13 @@ import { copy } from "https://deno.land/std@$STD_VERSION/fs/copy.ts";
 await copy("./foo", "./bar");
 ```
 
-See [here](#recommended-usage-patterns) for recommended usage patterns.
+See [here](#recommended-usage) for recommended usage patterns.
 
 ## Documentation
 
 Check out the documentation [here](https://deno.land/std?doc).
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,50 +1,80 @@
-# Deno Standard Modules
+# Deno Standard Library
```
lino-levan, line 1: Are we giving up on "module" as the official naming for Deno modules? If so, we should probably update /manual.

lino-levan (commented):
One nit. Otherwise looks like a more clear README to me.

README.md
iuioiua: Ah, you might be misinterpreting. The title change is because this project is the Deno Standard Library. It's as simple as that. My thinking is that otherwise, a newcomer might see the current title and wonder whether to refer to this as "Deno Standard Modules" or "Deno Standard Library". This is a consistency improvement.

lino-levan: Ah alright, seems reasonable.

kt3k (approved):
LGTM

PR #3732, opened 2023-10-28
BREAKING: deprecate `std/wasi`

iuioiua:
After a discussion with @kt3k, there appears to be nearly no interest or development in this module. Best it be removed.


kt3k (approved):
Let's deprecate as there seem no known usages. Also the download count of `std/wasi` is very low. Also we don't particularly encourage using wasi in Deno.

LGTM

PR #3734, opened 2023-10-28
feat: `std/data_structures`

iuioiua:
This change moves the unstable `BinaryHeap()`, `BinarySearchTree()`, `ascend()`, `descend()` and `RedBlackTree()` APIs to the new `std/data_structures` sub-module. After a discussion with @kt3k, we decided this location for these APIs makes the most sense. This will also open up opportunities for more data-structure-related APIs in the future.

The unstable files within `std/collections` were deleted, as they were... unstable. I've updated and cleaned up the previous deprecation documentation and added the sub-module to CI processes, etc.


kt3k:
Any suggestion, objection, bikeshedding, etc from the community is welcome.

iuioiua:
There seems to be no concern.

.github/CODEOWNERS
bartlomieju, line 1: I think we should remove this file altogether (in a separate PR)

bartlomieju (approved):
LGTM

iuioiua:
I've corrected the removal version.

kt3k:
Let me try some bike shedding. How about singular form: `std/data_structure`?

@KyleJune Do you have any other/better ideas for this new submodule name?

iuioiua:
I think `data_structures` makes more sense as the folder contains implementations for multiple data structures.

KyleJune:
Of the two, my preference would be std/data_structures to make it clear there are many. std/collections is also plural.

kt3k (approved):
LGTM 👍

PR #3735, opened 2023-10-28
BREAKING: remove deprecated `std/collections` APIs

iuioiua:
Please don't merge until the published version is 0.206.0.

Prerequisite #3734


iuioiua:
I think adding a deprecation period for the unstable APIs is the way to go. I'll do that in another PR. Closing.

PR #3736, opened 2023-10-28
chore(http): remove `http/unstable_errors`

iuioiua:
After a discussion with @kt3k, we agreed that `std/http/http_errors` is likely not needed in the Standard Library. It overlaps with `std/http/status` and is more complex than using custom error classes.

I've deleted the unstable files so things are clearer, updated the previous deprecation documentation and changed the removal version to 0.207.0.


kt3k (approved):
LGTM.

BTW I don't think we need to tag this as `BREAKING` as they are already deprecated.

kitsonk:
This has made it so oak cannot be upgraded to the latest version of std.

A custom error class, which is common in other frameworks like express and koa encapsulates a lot of functionality that allows end users to throw HTTP exceptions and have an appropriate handling throughout the middle ware.  There are specific properties and the like which middleware frameworks can key off of to determine the display.

It was a lot more than "an overlap with status".

I am disappointed and now will have to take this back into oak after it had been contributed to std. 😢

iuioiua:
Hi Kitson. Yes, we were aware that Oak used this API in our discussions. Initially, I was also against the move because SaaSKit used the API (which I manage). I [refactored SaaSKit to instead use custom error classes directly](https://github.com/denoland/saaskit/pull/638), and it made error handling logic more straightforward. 

Yes, our reasoning was more than just overlapping with `std/http/status`. As stated, it seemed a little more complex than just using custom errors. It also seemed a little too opinionated for a Standard Library and had nearly no usage, according to analytics. Perhaps I could've been better in my initial communication. My bad.

While we think the API might not belong in a Standard Library, it doesn't mean that it might not be useful in other contexts, such as a HTTP framework such as Oak, which is more reasonable to be more opinionated. Perhaps I haven't convinced you, but I hope, if anything, this makes our reasoning behind the decision clearer 🙂

kitsonk:
It should not have been accepted in the first place then. There was a lot of pressure to make std more opionated in regards to HTTP which I pushed back on but eventually said, that I would rwfactor out parts of oak to std to reduce that pressure, only to have all that unwound a few months later. 😞

PR #3738, opened 2023-10-28
BREAKING: stabilise `std/http/methods`

iuioiua:
Another list of HTTP methods worth considering is [this](https://www.iana.org/assignments/http-methods/http-methods.xhtml). However, I think it's best to support the commonly used HTTP methods initially and only expand to include the uncommonly used ones if there's demand for them.

This is a continuation of #3655.


kt3k:
I have a few questions about this.

In what case is this module useful? I've never needed this kind of check (`isHttpMethod`) or list (`HTTP_METHODS`) in web development. What is the valid use case?

> I think it's best to support the commonly used HTTP methods initially

I don't see enough motivation to limit them to ones mentioned in rfc 9110. rfc 9110 says:

> Additional methods, outside the scope of this specification, have been specified for use in HTTP. All such methods ought to be registered within the "Hypertext Transfer Protocol (HTTP) Method Registry", as described in [Section 16.1](https://www.rfc-editor.org/rfc/rfc9110.html#method.extensibility).

So the ones in [HTTP method registry](https://www.iana.org/assignments/http-methods/http-methods.xhtml) are also HTTP methods. Why do we exclude them?

> expand to include the uncommonly used ones if there's demand for them.

Any such change is going to be a breaking change (as isHttpMethod returns different values) and requires major version bump of `std/http` after v1. Is that acceptable situation?

iuioiua:
Hm... Now that you say it, I don't think this file has a valid use case. What do usage data say? We can also ask the community on Discord. Let's confirm it's even needed in the first place before proceeding to discuss these changes.

iuioiua:
Hm... Perhaps, we should just deprecate then remove this file: https://discord.com/channels/684898665143206084/775393009981849600/1173370592003625090

iuioiua:
Closing in favour of #3834 for now.

PR #3739, opened 2023-10-28
perf(streams): add single-character fast path for `DelimiterStream()`

aapoalas:
DelimiterStream is often used to delimit formats that use a single char as the separator such as most commonly used flavours of CSV, new-line delimited streamable JSON and others. Multi-character delimiters require extra handling that is entirely unnecessary for char delimiters. As such, it makes sense to special-case char delimiter streams for extra performance.

The performance benefit is at most a few percentage points.


streams/delimiter_stream.ts
```diff
@@ -55,24 +55,47 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
...
-      },
+      transform,
+      flush,
     });
 
     this.#delimiter = delimiter;
-    this.#delimLPS = createLPS(delimiter);
+    this.#delimLPS = lps;
     this.#disp = options?.disposition ?? "discard";
   }
```
iuioiua, line 100: We can simplify the constructor by doing something like:
```ts
constructor(
  delimiter: Uint8Array,
  options?: DelimiterStreamOptions,
) {
  super({
    transform: (chunk, controller) =>
      delimiter.length === 1
        ? this.#handleChar(chunk, controller)
        : this.#handle(chunk, controller),
    flush: (controller) => this.#flush(controller),
  });

  this.#delimiter = delimiter;
  this.#delimLPS = delimiter.length > 1 ? createLPS(delimiter) : null;
  this.#disp = options?.disposition ?? "discard";
}
```
I think we should also factor out the `flush()` logic into its own separate `#flush()` method to match the style of the rest of the class. WDYT?

streams/delimiter_stream_test.ts
```diff
@@ -79,6 +79,76 @@ Deno.test("[streams] DelimiterStream, prefix", async () => {
...
+  "_", // chunk with only delimiter
+  "qwertzu", // no delimiter
+  "iopasd_mnbvc", // one delimiter in the middle
+  "xylkjh_gfdsap_oiuzt", // two separate delimiters
+  "euoi__aueiou", // two consecutive delimiters
+  "rewq098765432", // more than one intermediate chunks with no delimiters
+  "349012i491290", // more than one intermediate chunks with no delimiters
+  "asdfghjkliop", // more than one intermediate chunks with no delimiters
+  "ytrewq_mnbvcxz", // one delimiter in the middle after multiple chunks with no delimiters
+  "_asd", // chunk starts with delimiter
```
iuioiua, line 95: Great comments! 👍🏾

iuioiua (changes requested):
Thanks for this! I think we can make this a little simpler.

streams/delimiter_stream.ts
iuioiua: Gentle ping @aapoalas

iuioiua (approved):
I've gone ahead and applied my suggestion by factoring the `flush()` logic into it's own method and simplifying the constructor. Now, LGTM! Thanks for this, @aapoalas!

PR #3741, opened 2023-10-28
chore: cleanup `collections/CONTRIBUTING.md`

iuioiua:
We already have a [CONTRIBUTING.md](https://github.com/denoland/deno_std/blob/main/CONTRIBUTING.md). This file is not needed.


kt3k:
The philosophy part is not covered by `/CONTRIBUTING.md` at the top level. I think we should keep them somehow somewhere (especially the notes about the side effects and the triviality of the some of APIs seem worth keeping)

iuioiua:
Sorry, I should've stated, I'm going to use parts of this in the new contributing file and perhaps even the `std/collection` JSDocs.

kt3k (approved):
Fair enough. LGTM

PR #3742, opened 2023-10-28
chore: cleanup old `std/fmt` files

iuioiua:
It appears these files aren't needed.


kt3k (approved):
These are from this PR https://github.com/denoland/deno_std/pull/566 and they look very stale. Let's remove.

LGTM

PR #3745, opened 2023-10-29
docs: add "Stability" README section

iuioiua:
This list will change often enough that I'm uncertain whether we should have this here or in a pinned tracking issue. Either way, here it is.


kt3k (approved):
LGTM

PR #3749, opened 2023-10-30
chore: add `ok` task

iuioiua:
This change adds the `ok` task, which is a holistic task checking that formatting, linting and tests all look good.


kt3k:
I prefer to keep `lint` CI job. By this change `lint` job runs on all platforms and with 2 deno versions. That looks unnecessary and waste of CI time

iuioiua:
Yes, the `lint` job would run on all platforms and with 2 versions of the Deno runtime. However, I think this is a worthwhile tradeoff. Do consider:
- We gain simpler CI.
- We use normal-sized runners, which are free.
- The difference in CI times is arguably imperceptible compared to the current CI setup (see [here](https://github.com/denoland/deno_std/actions/workflows/ci.yml)).

kt3k:
`deno fmt` sometimes has breaking changes. In that case, this CI setup doesn't pass.

Another downside is that `lint` failure is now not visible in PR page.

> We gain simpler CI.

I would choose faster CI setup rather than simpler CI setup

iuioiua:
Fair enough. I'll revert the CI changes. We'd still have the benefit of contributors being able to use `deno task ok` for local development.

kt3k (approved):
LGTM.

BTW `deno task lint` doesn't include `deno fmt --check`. Maybe you'd like to include it to `deno task ok` as well?

kt3k (approved):
LGTM again!

PR #3751, opened 2023-10-30
feat: `ServerSentEventStream()`

### LINKED ISSUE #3693: proposal: replace `ServerSentEvent()` with `ServerSentEventStream()`

I am suggesting we replace the `http/server_sent_events.ts` implementation to be a simple `TransformStream` that performs the encoding of SSEs to strings.

Here is an example I just threw together with @kt3k about how this should work:

```ts
const sse_stream = new SSEStream();

sse_stream.writable.getWriter().write({
  data: "Hello World",
  id: "",
  delay: 213,
});

sse_stream.readable.getReader().read();
// data: "Hello World"\r\n
// id: ""
// delay: 213\r\n\r\n

const stream = new ReadableStream({
  start(controller) {
    setInterval(() => {
      controller.enqueue({ data: "Hello World", id: "", delay: 213 });
    }, 1000);
  },
});

const body = stream.pipeThrough(new SSEStream()).pipeThrough(new TextDecoderStream())
new Response(body, { headers: { "Content-Type": "text/event-stream" } })
```

iuioiua:
You can test this via:
```ts
import {
  type ServerSentEventMessage,
  ServerSentEventStream,
} from "./http/server_sent_event_stream.ts";

Deno.serve({ port: 8000 }, (req) => {
  if (req.url.endsWith("/")) {
    const body = `
    <!DOCTYPE html>
    <html>
      <head>
        <meta charset=&quot;utf-8&quot; />
      </head>
      <body>
        <script>
          const source = new EventSource('/sse');
          source.onmessage = function(e) {
            document.body.innerHTML += e.data + '<br>';
          };
        </script>
      </body>
    </html>
    `;
    return new Response(body, {
      headers: { "content-type": "text/html; charset=utf-8" },
    });
  }

  if (req.url.endsWith("/sse")) {
    const stream = ReadableStream.from<ServerSentEventMessage>([{
      data: "hello there",
      id: "123",
    }]).pipeThrough(new ServerSentEventStream());
    return new Response(stream, {
      headers: {
        "content-type": "text/event-stream",
        "cache-control": "no-cache",
      },
    });
  }

  return new Response("not found", { status: 404 });
});
```

My main concern is that the name is too wordy. However, the reason I chose it is because it's clearest and doesn't conflict with the current `ServerSentEvent()` in `std/http`.

Closes #3693


http/server_sent_event_message_stream.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Converts a server-sent message object into a string for the client.
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format}
+ */
+function stringify(message: ServerSentEventMessage) {
+  const lines = [];
+  if (message.comment) lines.push(`:${message.comment}`);
+  if (message.event) lines.push(`event:${message.event}`);
+  if (message.data) {
+    message.data.split(/\r?\n/).forEach((line) => lines.push(`data:${line}`));
```
lucacasonato, line 29: I think just \r is also a valid new line delimiter. Please check the spec tho

http/server_sent_event_message_stream.ts
```diff
@@ -0,0 +1,69 @@
...
+ *   .pipeThrough(new TextEncoderStream());
+ * new Response(stream, {
+ *   headers: {
+ *     "content-type": "text/event-stream",
+ *     "cache-control": "no-cache",
+ *   },
+ * });
+ * ```
+ */
+export class ServerSentEventMessageStream
```
lucacasonato, line 60:
```suggestion
export class ServerSentEventStream
```

lucacasonato (commented):
Yay

http/server_sent_event_message_stream.ts
iuioiua: Yep, this line splits `messsage.data` by `\r\n`, `\n` and `\r`. In case there's any confusion, this is different to how newlines are represented in the SSE final message string, which is safe to just be a `\n`, according to the spec (see [here](https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream)). Either way, I've added to the test to check that `message.data` is split correctly for various delimiters.

http/server_sent_event_message_stream.ts
iuioiua: Done, but I've kept the `ServerSentEventMessage` interface name as-is.

http/server_sent_event_stream.ts
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Converts a server-sent message object into a string for the client.
+ *
+ * @see {@link https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events#event_stream_format}
+ */
+function stringify(message: ServerSentEventMessage) {
+  const lines = [];
+  if (message.comment) lines.push(`:${message.comment}`);
+  if (message.event) lines.push(`event:${message.event}`);
```
lino-levan, line 28: Are newlines allowed for these? Should we validate that these fields don't contain any newlines?

http/server_sent_event_stream_test.ts
```diff
@@ -0,0 +1,23 @@
...
+    retry: 456,
+  }]).pipeThrough(new ServerSentEventStream());
+  let chunks = "";
+  for await (const chunk of stream) chunks += chunk;
+
+  assertEquals(
+    chunks,
+    `:a\nevent:b\ndata:c\ndata:d\ndata:e\ndata:f\nid:123\nretry:456\n\n`,
+  );
+});
```
lino-levan, line 23: Would be nice to have a test that validated only some fields being specified.

lino-levan (approved):
Overall, looks great! Nice work Asher. Two nits.

http/server_sent_event_stream.ts
iuioiua: I've added a check that asserts that the single-line fields don't contain any newlines.

lino-levan (approved):
LGTM

kt3k:
How about directly encoding into Uint8Array? Do we ever need string output of SSE?

iuioiua:
Good idea. Done.

kt3k (approved):
LGTM!

Issue #3754, opened 2023-10-31
tracking: format  test names


timreichen:
Ref: https://github.com/denoland/deno_std/issues/2553#issuecomment-1784006070

This issue tracks the implementation progress of adapting the following format for test names in each sub-module:
```
<symbol>() <description> // E.g. ensureDir() creates the directory if it doesn't exist
```

- [x] archive
- [x] assert
- [x] async
- [x] bytes
- [x] cli
- [x] collections
- [x] console
- [x] crypto
- [x] csv
- [x] datetime
- [x] dotenv
- [x] encoding
- [x] expect
- [x] flags
- [x] fmt
- [x] front_matter
- [x] fs
- [x] html
- [x] http
- [x] io
- [x] json
- [x] jsonc
- [x] log
- [x] media_types
- [x] msgpack
- [x] path
- [x] regexp
- [x] semver
- [x] streams
- [ ] testing
- [x] text
- [x] toml
- [x] ulid
- [x] url
- [x] uuid
- [x] webgpu
- [x] yaml

JakeAve:
I'm working on fs - should be done tomorrow

scarf005:
https://github.com/denoland/deno-docs/blob/2c90cc2eca70c484e5fbfc61d256bf7342a2e95e/runtime/manual/references/contributing/style_guide.md?plain=1#L310-L327

maybe the test name rules could be documented in deno manual. should I open PR/issue?

iuioiua:
Ah, yes. We do need to add something in the contributing guidelines. A PR would be welcome!

syhol:
Is there a way to make the linter enforce this?

syhol:
Would it be worth using the bdd.ts describe blocks to define the subject under test?

```ts
Deno.test({
  name: "[std/datetime] isLeap",
  fn() {
    assert(isLeap(1992));
    assert(isLeap(2000));
    assert(!isLeap(2003));
    assert(!isLeap(2007));
    assert(!isLeap(new Date("1970-01-02")));
    assert(isLeap(new Date("1972-01-02")));
    assert(isLeap(new Date("2000-01-02")));
    assert(!isLeap(new Date("2100-01-02")));
  },
});
```
to
```ts
import { describe, it } from "../testing/bdd.ts";

describe("isLeap()", () => {
  it("identifies leap years for numbers", () => {
    assert(isLeap(1992));
    assert(isLeap(2000));
    assert(!isLeap(2003));
[...]

PR #3755, opened 2023-10-31
refactor(path): format  test names

timreichen:

I added `posix` and `win32` as prefixes to the function name (for example ` posix.toFileUrl()`).

Question: lots of tests do not have a description. Most of these would repeat the information in the function name. Do all tests need a description?

### DIFF
--- a/path/glob_test.ts
+++ b/path/glob_test.ts
@@ -380,7 +380,7 @@ Deno.test({
 Deno.test({
   name: "globToRegExp() matches special RegExp characters in range",
   fn() {
-    // Excluding characters checksed in the previous test.
+    // Excluding characters checked in the previous test.
     assertEquals(globToRegExp("[\\\\$^.=]", { os: "linux" }), /^[\\$^.=]\/*$/);
     assertEquals(
       globToRegExp("[!\\\\$^.=]", { os: "linux" }),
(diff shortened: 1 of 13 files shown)


path/glob_test.ts
```diff
@@ -370,17 +370,17 @@ Deno.test({
...
     assertEquals(globToRegExp("[!?*+@!|]", { os: "linux" }), /^[^?*+@!|]\/*$/);
   },
 });
 
 Deno.test({
-  name: "[path] globToRegExp() Special RegExp characters in range",
+  name: "globToRegExp() matches special RegExp characters in range",
   fn() {
-    // Excluding characters checked in the previous test.
+    // Excluding characters checksed in the previous test.
```
kt3k, line 383:
```suggestion
    // Excluding characters checked in the previous test.
```

kt3k (approved):
LGTM. Thanks!

kt3k:
> Question: lots of tests do not have a description. Most of these would repeat the information in the function name. Do all tests need a description?

Can you provide an example?

timreichen:
> > Question: lots of tests do not have a description. Most of these would repeat the information in the function name. Do all tests need a description?
> 
> Can you provide an example?

For example in `basename_test.ts`: There are only two tests (`posix.basename()` and `posix.basename()`) I guess the description would be something like `returns the basename of a posix path` and `returns the basename of a win32 path`  respectively.

kt3k:
I think it's fine to omit the description in such cases

PR #3762, opened 2023-11-01
fix(fmt): format duration rounding error.

hastebrot:
this works-around rounding errors in `millisecondsToDurationObject()` of `std/fmt/duration.ts` by converting values in milliseconds or less into as string and use `String.slice()`. 

use-case: I get durations like `16.342` millis when I round my values to microseconds.

`main`:
```ts
> import { format } from "https://deno.land/std@0.204.0/fmt/duration.ts";
> format(16.342, { ignoreZero: true })
"16ms 341µs 999ns"
```

`hastebrot/fix-duration-format-rounding-error`:
```ts
Deno.test({
  name: "format duration rounding error",
  fn() {
    assertEquals(format(16.342, { ignoreZero: true }), "16ms 342µs");
  },
});
```

TODO:
- [x] provide example and tests
- [ ] check if rounding problems are fixed in #3672

related to #3672

### DIFF
--- a/fmt/duration.ts
+++ b/fmt/duration.ts
@@ -50,14 +50,15 @@ const keyList: Record<keyof DurationObject, string> = {
 function millisecondsToDurationObject(ms: number): DurationObject {
   // Duration cannot be negative
   const millis = Math.abs(ms);
+  const millisFraction = millis.toFixed(7).slice(-7, -1)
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks for the fix!

PR #3766, opened 2023-11-02
feat(testing/mock): enable `spy` to accept a class constructor

hasundue:
I encounter a situation where I have to create a wrapper class for `Deno.Command` to spy/mock its constructor, but it is not consistent with the mocking API provided by `std/testing/mock.ts`, which is awkward to me.

So this PR proposes enabling `spy` function to accept a class constructor as an argument, so that we can create a spy instance for a class constructor in the same manner as for a regular function.

This shouldn't be a breaking change, but does include some internal API changes. I didn't find a clean way to incorporate two types of spies in a single interface `Spy`, so I created a new `ConstructorSpy` interface, which is returned when you call `spy` on a constructor.

It is the first time for me to create a PR to this repository, so please kindly notice me if I miss something.

### DIFF
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -555,6 +555,58 @@ function methodSpy<
   return spy;
 }
 
+/** A constructor wrapper that records all calls made to it. */
+export interface ConstructorSpy<
+  // deno-lint-ignore no-explicit-any
+  Self = any,
+  // deno-lint-ignore no-explicit-any
+  Args extends unknown[] = any[],
+> {
+  new (...args: Args): Self;
+  /** The function that is being spied on. */
+  original: new (...args: Args) => Self;
+  /** Information about calls made to the function or instance method. */
+  calls: SpyCall<Self, Args, Self>[];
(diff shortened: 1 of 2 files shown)


hasundue:
There's some instability in file_server_test.ts apparently?
Found a similar failure in another PR: https://github.com/denoland/deno_std/actions/runs/6732690687/job/18299866301

iuioiua:
> There's some instability in file_server_test.ts apparently?
> 
> Found a similar failure in another PR: https://github.com/denoland/deno_std/actions/runs/6732690687/job/18299866301

Looking into it...

kt3k:
Looks like a reasonable extension of `spy` to me.

@KyleJune I'd appreciate your opinion on this

KyleJune:
> Looks like a reasonable extension of `spy` to me.
> 
> @KyleJune I'd appreciate your opinion on this

I think it's a reasonable extension. I didn't review the implementation, but the tests looked good.

kt3k (approved):
LGTM. Thanks for your contribution! @hasundue

PR #3768, opened 2023-11-03
chore: use `Deno.readTextFile()` and `Deno.writeTextFile()` in tests

iuioiua:
While looking at the failing HTTP tests, I realised these should be cleaned up.


lino-levan (approved):
Nice clean up. Love to see it.

kt3k (approved):
LGTM 👍

Issue #3770, opened 2023-11-06
proposal: rework and fix `std/http/file_server` and tests


iuioiua:
`std/http/file_server` tests are currently flaky. One reason this may be happening is that each of these tests opens a new process and makes `fetch()` calls. However, such network calls and new processes are not needed. Instead, we can use something like `const res = await serveDir(req)`. This would be a far more straightforward method of testing.

PR #3772, opened 2023-11-07
chore(http): fix file server test names

iuioiua:
This will make the diff easier to follow when transitioning to using `serveDir()` directly.

Towards #3770 and #3754


kt3k (approved):
LGTM

PR #3777, opened 2023-11-08
feat(cli): `promptSecret()`

lambdalisue:
The function is similar to `prompt`, but it prints the user's input as `*` to prevent the password from being shown.

Sometimes, I find the need for this type of function when creating CLI applications, so I decided to implement it. I'm not certain whether it should be included in the standard library, as it's somewhat challenging to test since it hooks into stdin. Therefore, I couldn't provide tests (or please let me know if there is a way).

Please inform me if this function shouldn't be included in the standard library, and I'll develop it as a third-party module. :+1:

### DIFF
--- a/cli/prompt_secret.ts
+++ b/cli/prompt_secret.ts
@@ -1,0 +1,57 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+const input = Deno.stdin;
+const output = Deno.stdout;
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+const LF = "\n".charCodeAt(0);
+const CR = "\r".charCodeAt(0);
+
+/**
+ * Shows the given message and waits for the user's input. Returns the user's input as string.
+ * This is similar to `prompt()` but it print user's input as `*` to prevent password from being shown.
+ * Use an empty `mask` if you don't want to show any character.
+ */
+export function promptSecret(
+  message = "Password ",
+  mask = "*",
+): string | null {
+  if (!Deno.isatty(input.rid)) {
+    return null;
+  }
+
+  const maskChar = encoder.encode(mask);
+  const callback = mask === "" ? () => {} : () => {
+    output.writeSync(maskChar);
+  };
+  output.writeSync(encoder.encode(message));
+
+  Deno.stdin.setRaw(true, { cbreak: true });
+  try {
+    return readLineFromStdinSync(callback);
+  } finally {
+    Deno.stdin.setRaw(false);
+  }
+}
+
+// Slightly modified from Deno's runtime/js/41_prompt.js
+// This implementation immediately break on CR or LF and accept callback.
+// The original version waits LF when CR is received.
+// https://github.com/denoland/deno/blob/e4593873a9c791238685dfbb45e64b4485884174/runtime/js/41_prompt.js#L52-L77
+function readLineFromStdinSync(callback: () => void): string {
+  const c = new Uint8Array(1);
+  const buf = [];
+
+  while (true) {
+    const n = input.readSync(c);
+    if (n === null || n === 0) {
+      break;
+    }
+    if (c[0] === CR || c[0] === LF) {
+      break;
+    }
+    buf.push(c[0]);
+    callback();
+  }
+  return decoder.decode(new Uint8Array(buf));
+}
--- a/cli/prompt_secret.ts
+++ b/cli/prompt_secret.ts
@@ -13,7 +13,7 @@ const CR = "\r".charCodeAt(0);
  * Use an empty `mask` if you don't want to show any character.
  */
 export function promptSecret(
-  message = "Password ",
+  message = "Secret ",
   mask = "*",
 ): string | null {
   if (!Deno.isatty(input.rid)) {
--- a/cli/prompt_secret.ts
+++ b/cli/prompt_secret.ts
@@ -10,6 +10,13 @@ const BS = "\b".charCodeAt(0); // ^H - Backspace on Linux and Windows
 const DEL = 0x7f; // ^? - Backspace on macOS
 const CLR = encoder.encode("\r\u001b[K"); // Clear the current line
 
+export type PromptSecretOptions = {
+  /** A character to print instead of the user's input. */
+  mask?: string;
+  /** Clear the current line after the user's input. */
+  clear?: boolean;
+};
+
 /**
  * Shows the given message and waits for the user's input. Returns the user's input as string.
  * This is similar to `prompt()` but it print user's input as `*` to prevent password from being shown.
@@ -17,14 +24,13 @@ const CLR = encoder.encode("\r\u001b[K"); // Clear the current line
  */
 export function promptSecret(
   message = "Secret ",
-  mask = "*",
+  { mask = "*", clear }: PromptSecretOptions = {},
 ): string | null {
   if (!Deno.isatty(input.rid)) {
     return null;
   }
 
-  const callback = mask === "" ? undefined : (n: number) => {
-    // Clear the current line
+  const callback = !mask ? undefined : (n: number) => {
     output.writeSync(CLR);
     output.writeSync(encoder.encode(`${message}${mask.repeat(n)}`));
   };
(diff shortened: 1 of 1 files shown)


iuioiua (changes requested):
Hi @lambdalisue, thanks for the contribution! This is related to #3526. Can you please move the implementation to a new `std/cli` top-level folder? We haven't confirmed the new sub-module, but I feel like it's likely to happen.

iuioiua:
Just a little update - we're going ahead with the `std/cli` sub-module.

lambdalisue:
I'll apply that change on this weekend 👍

lambdalisue:
@iuioiua Done. Please re-review.

cli/prompt_secret.ts
```diff
@@ -0,0 +1,57 @@
...
+  } finally {
+    Deno.stdin.setRaw(false);
+  }
+}
+
+// Slightly modified from Deno's runtime/js/41_prompt.js
+// This implementation immediately break on CR or LF and accept callback.
+// The original version waits LF when CR is received.
+// https://github.com/denoland/deno/blob/e4593873a9c791238685dfbb45e64b4485884174/runtime/js/41_prompt.js#L52-L77
+function readLineFromStdinSync(callback: () => void): string {
```
iuioiua, line 41:
```suggestion
function readLineFromStdinSync(callback?: () => void): string {
```
I think we should make the callback optional and only call it when defined.

cli/prompt_secret.ts
```diff
@@ -0,0 +1,57 @@
...
+const LF = "\n".charCodeAt(0);
+const CR = "\r".charCodeAt(0);
+
+/**
+ * Shows the given message and waits for the user's input. Returns the user's input as string.
+ * This is similar to `prompt()` but it print user's input as `*` to prevent password from being shown.
+ * Use an empty `mask` if you don't want to show any character.
+ */
+export function promptSecret(
+  message = "Password ",
```
iuioiua, line 16:
```suggestion
  message = "Secret ",
```
A secret may not always be a password.

iuioiua (changes requested):
Nice work so far. Few things:
1. I'm getting the following artefact on my macOS machine:
   ```bash
   Password **********% 
   ```
1. Backspace isn't being correctly processed.
1. Perhaps, we can open a new process using and simulate typing with [`Deno.Command()`](https://deno.land/api@v1.38.1?s=Deno.Command&unstable=).

lambdalisue:
> Perhaps, we can open a new process using and simulate typing with [Deno.Command()](https://deno.land/api@v1.38.1?s=Deno.Command&unstable=).

Ah, sounds nice. I'll try.

lambdalisue:
@iuioiua 

I thought that the following implementation would handle the Ctrl sequence on the side of the process invoked by `Deno.Command` to handle backspaces properly, and replied that it looked good, but it did not work. I think I misunderstood something, so could you please tell me what kind of implementation you were expecting?

**test.ts**

```ts
const cmd = new Deno.Command(Deno.execPath(), {
  args: ["run", "./cli/prompt_secret_script.ts"],
  stdin: "inherit",
  stdout: "piped",
});
try {
  Deno.stdin.setRaw(true, { cbreak: true });
  const proc = cmd.spawn();
  const { stdout } = await proc.output();
  const result = (new TextDecoder()).decode(stdout);
  console.log(`OUTPUT: ${result}`);
} finally {
  Deno.stdin.setRaw(false);
}
```

**prompt_secret_script.ts**

```ts
const input = Deno.stdin;
const output = Deno.stdout;
const LF = "\n".charCodeAt(0);
const CR = "\r".charCodeAt(0);

const c = new Uint8Array(1);
const buf = [];

while (true) {
  const n = input.readSync(c);
  if (n === null || n === 0) {
    break;
  }
  if (c[0] === CR || c[0] === LF) {
    break;
  }
  buf.push(c[0]);
}
output.writeSync(new Uint8Array(buf));
```

iuioiua:
I tried playing around with this and I couldn't get a test work. I would've thought that this could be accomplished by piping both input and output. Using `child.output()` is not ideal because you want to capture output bit-by-bit. Not all in one hit (most likely, IIUC).

lambdalisue:
I see. We want a stdin that doesn't display the input string, not a stdin that doesn't process it... I think it would make more sense to add a `conceal` mode or whatever to deno's `stdin` instead of using `setRaw` to force deno_std to do it. What do you think?

lambdalisue:
@iuioiua, I'm not certain whether my implementation works on Windows, but I've added control sequence handling to properly manage the backspace. I'm unsure if this approach is a good idea compared to my previous comment, but in case you find it preferable.

iuioiua:
Nice. Backspaces are working now. Frankly, I'm not sure which direction to go with this. @kt3k, do you have any advice on the implementation of this?

Another issue I found. Once I hit enter upon entering my secret, the `Secret *********` line disappears from the console. The line remains in the console when I do the same for `prompt()`. Also, check [this](https://github.com/caspervonb/deno-prompts/blob/master/mod.ts) out. It may be of value to you.

lambdalisue:
> Once I hit enter upon entering my secret, the Secret ********* line disappears from the console.

Actually, that's on purpose. I can make the behavior similar to `prompt()` but `Secret *****` shows the number of characters so I thought it should be removed.

iuioiua:
I think it's ok to have the `***` characters shown and be as close to `prompt()` as possible.

lambdalisue:
I made some adjustments to the interface to enable end users to customize the behavior.

https://github.com/denoland/deno_std/assets/546312/d8ee11ca-c874-4ee6-8d93-71a2be89fc8e

cli/prompt_secret.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+    Deno.stdin.setRaw(false);
+  }
+}
+
+// Slightly modified from Deno's runtime/js/41_prompt.js
+// This implementation immediately break on CR or LF and accept callback.
+// The original version waits LF when CR is received.
+// https://github.com/denoland/deno/blob/e4593873a9c791238685dfbb45e64b4485884174/runtime/js/41_prompt.js#L52-L77
+function readLineFromStdinSync(callback?: (n: number) => void): string {
```
iuioiua, line 56: Would this function be simpler if we instead passed `mask: string` as the parameter?

iuioiua (changes requested):
This is looking good! Just one final nit. Also, @kt3k, any idea how we can test this?

cli/prompt_secret.ts
lambdalisue: If the function reads from stdin and writes feedback mask characters to stdout, I find it a bit challenging to choose an appropriate name for the function. It seems that the responsibility of the function is somewhat complex, I feel.

iuioiua (approved):
This LGTM for first pass! I think we can improve in following PRs by adding tests (however that's done) and perhaps simplifying logic. Thank you, @lambdalisue!

iuioiua:
@ayame113, are you able to confirm whether this works correctly on Windows? IIRC, you've got a Windows machine.

ayame113:
Hi @iuioiua !

I tried running the snippet below on windows.

```
import { promptSecret } from "./cli/prompt_secret.ts"

const password = promptSecret("please input password")
console.log(password)
```

Then the following error message was output.

```
please input passworderror: Uncaught NotSupported: The operation is not supported
  Deno.stdin.setRaw(true, { cbreak: true });
             ^
    at Stdin.setRaw (ext:deno_io/12_io.js:256:9)
    at promptSecret (file:///C:/Users/ayame/work/pr-check/deno_std/cli/prompt_secret.ts:39:14)
    at file:///C:/Users/ayame/work/pr-check/deno_std/aaaaaaaa.ts:3:18
```

When I looked at the `Deno.stdin.setRaw` documentation, it seems that the `cbreak` option is not supported on Windows.

> The `cbreak` option can be used to indicate that characters that correspond to a signal should still be generated. When disabling raw mode, this option is ignored. This functionality currently only works on Linux and Mac OS.
> https://deno.land/api@v1.38.3?s=Deno.SetRawOptions

lambdalisue:
@ayame113 I see. I'll remove that option on Windows later but could you remove it from code and try if that works properly?

ayame113:
> @ayame113 I see. I'll remove that option on Windows later but could you remove it from code and try if that works properly?

@lambdalisue 
I tried removing the cbreak option and it worked fine on windows. 🎉🎉🎉

lambdalisue:
@ayame113 Thanks a lot. I removed the option on Windows 👍

kt3k (approved):
LGTM

PR #3779, opened 2023-11-08
fix: rework file server tests

### LINKED ISSUE #3770: proposal: rework and fix `std/http/file_server` and tests

iuioiua:
This change fixes HTTP file server flakiness by using `serveDir()` directly instead of starting a new process and making `fetch()` requests. This increases the ergonomics, simplicity and stability of these tests. Other cleanups were also included. These tests also complete faster, as a bonus.

Closes #3770
Towards #3718

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -913,7 +913,6 @@ Deno.test(
   "serveDir() resolves empty sub-directory without asking for current directory read permissions on Windows",
   {
     ignore: Deno.build.os !== "windows",
-    permissions: {},
   },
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
LGTM, nice cleanup 👍

PR #3781, opened 2023-11-08
BREAKING(http): deprecate `enum Status` in favour of `STATUS_CODES` object

iuioiua:
This change deprecated the `Status` enum in favour of a new `STATUS_CODES` object. The main reason is that objects align more with native JavaScript than enums. There are also other technical reasons for objects being more desirable over enums. This PR also contains other improvements to `std/http`:
- Corrects statuses to align with [documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status):
  - `RequestEntityTooLarge` to `ContentTooLarge`
  - `RequestURITooLong` to `URITooLong`
  - `RequestedRangeNotSatisfiable` to `RangeNotSatisfiable`
- Replaces internal `createCommonResponse()` with a simpler `createStandardResponse()`. Other non-standard responses are created explicitly. This makes things easier to understand.

My main concern is, do the new symbol names make the most sense?

References:
- https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums
- https://youtu.be/jjMbPt_H3RQ?feature=shared
- https://youtu.be/3aUHskjF7pc?feature=shared


kt3k:
How about `STATUS` or `STATUS_CODE` instead of `STATUS_CODES`? Singular form sounds better to me in this case.

iuioiua:
I went with `STATUS_CODE` to line up with `STATUS_TEXT`.

kt3k (approved):
LGTM

PR #3784, opened 2023-11-09
BREAKING(bytes): deprecate `concat()` signatures that don't use `Uint8Array[]` argument

iuioiua:
This change deprecates the `concat()` signatures that use spread and depth > 1 arguments in favour of a single argument of an array of `Uint8Array`s. This will enable the function to remain simple and performant while still taking care of scenarios where large numbers of `Uint8Array`s need to be concatenated.

This comes from work done in #3698.


lino-levan (approved):
Would be a good time to aim for 100% coverage in concat as a bonus for this PR, but this LGTM.

bytes/concat.ts
```diff
@@ -1,18 +1,57 @@
...
+ *
+ * @example
+ * ```ts
+ * import { concat } from "https://deno.land/std@$STD_VERSION/bytes/concat.ts";
+ *
+ * const a = new Uint8Array([0, 1, 2]);
+ * const b = new Uint8Array([3, 4, 5]);
+ * concat(a, b); // Uint8Array(6) [ 0, 1, 2, 3, 4, 5 ]
+ * ```
+ */
```
kt3k, line 47: I don't think we need to have this jsdoc. This is for implementatioin signature ([ref](https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature)) and that is not visible to the users.

iuioiua: Ah, yes, you did say that. Fixed.

kt3k (approved):
LGTM

PR #3785, opened 2023-11-09
feat(http): middleware

### LINKED ISSUE #1295: Concept Proposal: `std/http/middleware`

iuioiua:
`createHandler()` creates a handler from a given middleware chain. This implementation aims to be plain, simple, unopinionated and align with the Deno runtime and Web APIs as much as possible.

Hopefully, this can be used as a fundamental building block for web frameworks and spur the creation of middleware with a similar design philosophy.

Closes #1295


http/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Middleware function based on {@linkcode Deno.ServeHandlerInfo} but an added
+ * `next()` function which calls the next middleware function in the middleware
+ * chain.
+ */
+export type MiddlewareFunction = (
+  request: Request,
+  info: Deno.ServeHandlerInfo,
+  next: () => Response | Promise<Response>,
+) => Response | Promise<Response>;
```
iuioiua, line 12: Other possible function signatures like `(request, next)` and `(request, info)` have deliberately been omitted. I think having a plain function signature and just prefixing unused arguments in the middleware function with `_` is best.

http/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+ * const handler = createHandler([])
+ * ```
+ */
+export function createHandler(
+  middlewares: MiddlewareFunction[],
+): Deno.ServeHandler {
+  return (request, info) => {
+    function compose(index: number): Response | Promise<Response> {
+      if (index >= middlewares.length) {
+        throw new RangeError("Middleware chain exhausted");
```
iuioiua, line 48: Does this error make sense?

http/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * const middleware2: MiddlewareFunction = (request, info) => {
+ *   return Response.json({ request, info });
+ * };
+ *
+ * const handler = createHandler([])
+ * ```
+ */
+export function createHandler(
+  middlewares: MiddlewareFunction[],
```
iuioiua, line 43: I think it's fine to not have spread (`...args`) arguments.

lino-levan: +1 for this decision. I think spread args would make for a worse api.

http/middleware.ts
lino-levan: I'm not sure if a `RangeError` is the right Error class for this. Should this be a custom error?

http/middleware.ts
lino-levan: +1 on this decision. This is the right signature if we want middleware to be std-defined.

lino-levan:
I'm slightly -1 on this from a philosophical standpoint, but the actual code looks fine to me.

http/middleware.ts
iuioiua: Yeah. Sidenote: I'd prefer to never use them...

http/middleware.ts
iuioiua: I'm not certain either. I'd be keen to hear other's thoughts.

http/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+ *   const duration = performance.now() - start;
+ *   response.headers.set("x-request-time", duration.toString());
+ *   return response;
+ * };
+ *
+ * const middleware2: MiddlewareFunction = (request, info) => {
+ *   return Response.json({ request, info });
+ * };
+ *
+ * const handler = createHandler([])
```
kt3k, line 39: Do you mean the below?
```suggestion
 * const handler = createHandler([middleware1, middleware2])
```

kt3k:
Maybe let's explore how this `MiddlewareFunction` can be used in fresh and oak?

Also this middleware doesn't seem having ability to pass information to the later middleware such as auth info, etc (or is that possible?). I wonder if this 'MiddlewareFunction' interface is general enough to be 'standard' middleware

iuioiua:
@kitsonk and @marvinhagemeister, WDYT? Can you see this being useful for your respective frameworks? 🙂

timreichen:
I would suggest in general to rename the module:
- `middleware` -> `handlers` to be more inclusive for future features.

And also
- `MiddlewareFunction` -> something like `MiddlewareHandler` or `RequestHandler` to be more specific
- `createHandler` ->  `composeHandlers` for more clarity


> Also this middleware doesn't seem having ability to pass information to the later middleware such as auth info, etc (or is that possible?). I wonder if this 'MiddlewareFunction' interface is general enough to be 'standard' middleware

I think this feature would be a must for useful middleware handlers in std.

Passing data to the next handler could be achieved by extending `Deno.ServeHandlerInfo` and optionally passing data when calling `next()`. Maybe something like this:
```ts
export type MiddlewareHandler<I> = (
  request: Request,
  info: I,
  next: (info?: I) => Response | Promise<Response>,
) => Response | Promise<Response>;

export function composeHandlers<I extends Deno.ServeHandlerInfo>(
  middlewares: MiddlewareHandler<I>[],
): (request: Request, info: I) => Promise<Response> | Response {
  return (request, info) => {
    function compose(index: number, info: I): Response | Promise<Response> {
      const handler = middlewares[index];
      if (!handler) throw new RangeError("Middleware chain exhausted");
      return handler(
        request,
        info,
        (newInfo = info) => compose(index + 1, newInfo),
      );
    }
    return compose(0, info);
  };
}

```
So it can be used like this:
```ts
import { composeHandlers } from "https://deno.land/std/http/handlers.ts";

interface StateHandlerInfo extends Deno.ServeHandlerInfo {
  state: Record<PropertyKey, unknown>;
}

const composedHandler = composeHandlers<StateHandlerInfo>([
  (_request, info, next) => next({ ...info, state: { foo: "bar" } }),
  (_request, info) => Response.json(info.state),
]);

Deno.serve((request, info) => composedHandler(request, { ...info, state: {} }));
```

iuioiua:
> I would suggest in general to rename the module:
> 
> * `middleware` -> `handlers` to be more inclusive for future features.

This file is deliberately only for middleware-related APIs.

> And also
> 
> * `MiddlewareFunction` -> something like `MiddlewareHandler` or `RequestHandler` to be more specific

I agree. Done.

> * `createHandler` ->  `composeHandlers` or `createMiddlewareHandler` for more clarity

I don't want to go down a bikeshedding rabbit hole, but `createHandler()` that accepts multiple middlewares is clear to me. Adding the word "compose" might raise the question: "What exactly does it mean to compose?"

> I think this feature would be a must for useful middleware handlers in std.
> 
> Passing data to the next handler could be achieved by extending `Deno.ServeHandlerInfo` and optionally passing data when calling `next()`.

I think the important question to answer is: "Is a context object a requirement _here_ or can context be managed without adding a parameter to `MiddlewareHandler`?" Likely. But I'm not sure. Either way, I'm on the fence about extending `Deno.ServeHandlerInfo`. Another idea is:
```ts
interface MiddlewareHandlerOptions<T> {
  info?: Deno.ServeHandlerInfo;
  context?: T;
}

export type MiddlewareHandler<T> = (
  request: Request,
  next: () => Response | Promise<Response>,
  options?: MiddlewareHandlerOptions<T>,
) => Response | Promise<Response>;
```

kt3k:
> "Is a context object a requirement here or can context be managed without adding a parameter to MiddlewareHandler?"

I think the context info can be stored somewhere else by using WeakMap keyed by request object.

```js
const contextMap = new WeakMap();

export function getContext(request: Request) {
  return contextMap.get(request);
}
```
```js
const middleware1 = async (req, info, next) => {
  const ctx = getContext(req);
  ...
};
```

iuioiua:
Oh, I like that approach.

http/middleware.ts
```diff
@@ -0,0 +1,54 @@
...
+ * };
+ *
+ * const handler = createHandler([middleware1, middleware2])
+ * ```
+ */
+export function createHandler(
+  middlewares: MiddlewareHandler[],
+): Deno.ServeHandler {
+  return (request, info) => {
+    function compose(index: number): Response | Promise<Response> {
```
marvinhagemeister, line 46: Tiny nit: The name `compose` is a bit misleading, because this function doesn't compose anything. What we call `createHandler` here resembles much more the signature of a `compose` function.

http/middleware.ts
marvinhagemeister: The way the `createHandler` function is designed here assumes that the last middleware never calls `next()`. So in that sense it's fine, although it prevents composing `createHandler` itself. Not sure if that is a goal here though.

```js
const a = createHandler([m1, m2, m3]);
const b = createHandler([m1, m2, m3]);

// Not possible
const c = createHandler([a, b]);
```

marvinhagemeister (commented):
I like the idea of exposing a composable middleware primitive in the std. One thing I'm wondering if the current signature proposed in this PR is useful enough. It allows you to call a bunch of functions in sequence, but the way it's designed here it's not possible to share state among them.

iuioiua:
> I like the idea of exposing a composable middleware primitive in the std. One thing I'm wondering if the current signature proposed in this PR is useful enough. It allows you to call a bunch of functions in sequence, but the way it's designed here it's not possible to share state among them.

If it had a state parameter, could you imagine it being something Fresh would use?

http/middleware.ts
iuioiua: I've renamed `createHandler` to `composeHandler`.

http/middleware.ts
iuioiua: Yeah, it wasn't one of my goals to do that... I don't know the advantage of doing it and whether there'd be sufficient use cases for that.

iuioiua:
I've added a state parameter. PTAL, @marvinhagemeister.

I am still not 100% sure about having this in the Standard Library. This PR aims to explore the idea either way. The implementation does aim to be as plain and unopinionated as possible.

@kitsonk, WDYT?

iuioiua:
PTAL @keithamus and @LionC

http/middleware.ts
```diff
@@ -0,0 +1,71 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+/** Options for {@linkcode MiddlewareHandler}. */
+export interface MiddlewareOptions<T = undefined> {
```
keithamus, line 4: Once nice thing in https://github.com/denoland/deno_std/pull/1555/files#diff-d7730497c05f4870ef5336fe16c6eddb662c84d364b8ec55a1810a36a4901591R51 is that the Middleware was typed to allow for dependencies in the state which allowed for a very nice typing of middleware.

kitsonk:
My view (https://github.com/denoland/deno_std/issues/1295#issuecomment-973709156) has not changed, and considering other things that are being deemed too frameworky around HTTP servers are being removed it seems that is the direction of travel...

http/middleware.ts
```diff
@@ -0,0 +1,71 @@
...
+  state: T;
+}
+
+/**
+ * Middleware handler which extends {@linkcode Deno.ServeHandlerInfo}. Used
+ * in {@linkcode composeMiddleware}.
+ */
+export type MiddlewareHandler<T = undefined> = (
+  request: Request,
+  info: Deno.ServeHandlerInfo,
```
syhol, line 17: I know this is from the Deno.serve API, but might it be cleaner to move it into either the options object or the options state object?
1. It follows the deno_std guidelines "max 2 args, put the rest into an options object"
2. Deno.ServeHandlerInfo isn't a very interesting object, no offence 😅, but it adds noise to the API when I can imagine most people will never use it.

iuioiua: Yeah, I'm thinking that too.

Leokuma:
Is it mandatory that every middleware in a chain has the same State type? Doesn't it make it difficult to reuse middlewares accross different chains/endpoints?

LionC:
> Is it mandatory that every middleware in a chain has the same State type? Doesn't it make it difficult to reuse middlewares accross different chains/endpoints?

They do not have the same type - they state what they require to already be present to work (so inputs) and what they add to the context (outputs) and the chaining function then composes them, combining the types correctly and in order. This makes them completely composable, as a combined chain is just a new middleware with those inputs and outputs again.

Leokuma:
Thank you, Lion.

Another question: the way this PR implements it, isn't there a chance of "state collision" between middlewares? Like, a middleware declares a `state.cryptoKey` and another middleware also uses the same prop, but it's just a naming coincidence as they serve different purposes. When you chain them together, you have to clean up `state` to avoid data from the previous middleware leaking into the next?

LionC:
> Thank you, Lion.
> 
> Another question: the way this PR implements it, isn't there a chance of "state collision" between middlewares? Like, a middleware declares a `state.cryptoKey` and another middleware also uses the same prop, but it's just a naming coincidence as they serve different purposes. When you chain them together, you have to clean up `state` to avoid data from the previous middleware leaking into the next?

Sorry, I replied in regards to how my original proposal and PoC PR implemented it (linked in the first post here as well), not related to this MR.

iuioiua:
> > Is it mandatory that every middleware in a chain has the same State type? Doesn't it make it difficult to reuse middlewares accross different chains/endpoints?
> 
> In my draft (linked in the original post), they do not have the same type - they state what they require to already be in the context to work (so basically context inputs) and what they add to the context (context outputs) and the chaining function then composes them, combining the types correctly and in order. This makes them completely composable, as a combined chain is just a new middleware with those inputs and outputs again, and a listener is just a middleware with an empty context input.
> 
> I think if we go Typescript first, we should honor order to catch errors, as it is just a function chain in the end.

Do you see a way to add this functionality to this PR's implementation?

iuioiua:
I will close this for now because this has gone stale and is not a priority right now. I'd be happy to look at a future PR that addresses some of the concerns mentioned here.

PR #3786, opened 2023-11-10
refactor: remove use of `deferred()`

iuioiua:
`deferred()` from `std/async` is now deprecated.

Towards #3718

### DIFF
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -1081,8 +1081,8 @@ Deno.test(
     };
     const listener = Deno.listen(listenOptions);
 
-    const onRequest = Promise.withResolvers();
-    const postRespondWith = Promise.withResolvers();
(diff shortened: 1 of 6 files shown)


http/server_test.ts
```diff
@@ -1081,8 +1081,8 @@ Deno.test(
     };
     const listener = Deno.listen(listenOptions);
 
-    const onRequest = deferred();
-    const postRespondWith = deferred();
+    const onRequest = Promise.withResolvers<void>();
```
iuioiua, line 1084: Is having to define a `void` generic argument a bug?

bartlomieju: It might be - maybe it should be a default? Can you open an issue in `deno` repo?

bartlomieju (approved):
LGTM, nice cleanup!

PR #3788, opened 2023-11-10
chore: use `Array.fromAsync()` where possible

### LINKED ISSUE #3767: suggestion: use `Array.fromAsync()` and `ReadableSteam.from()` where possible
Deno now supports [`Array.fromAsync()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync) and [`ReadableStream.from()`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/from_static). Using these methods could clean up common patterns. It'd also be a great exercise for someone wanting to become familiar with these methods.

I suggest using these methods, where appropriate, throughout the codebase to make things a little tidier. An example of this being done can be seen in `text_line_stream_test.ts` in #3744. A few things should be considered:
1. These methods should **not** be used in web-compatible files (those with `// This module is browser compatible.` at the top). The reason is that these methods are not yet widely supported in browsers, and adding these methods would reduce compatibility.
2. It's fine to use these methods in server-side code, including tests, as they use the Deno runtime.

iuioiua:
Closes #3767

### DIFF
--- a/archive/tar_test.ts
+++ b/archive/tar_test.ts
@@ -116,9 +116,10 @@ Deno.test("directoryEntryType", async function () {
 
   const reader = await Deno.open(outputFile, { read: true });
   const untar = new Untar(reader);
-  for await (const entry of untar) {
-    assertEquals(entry.type, "directory");
-  }
+  await Array.fromAsync(
+    untar,
+    (entry) => assertEquals(entry.type, "directory"),
+  );
 
   reader.close();
   await Deno.remove(outputFile);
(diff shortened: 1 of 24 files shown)


bartlomieju (approved):
Very nice cleanup 👍

kt3k:
FYI `Array.fromAsync` is currently not available in Chrome https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync

This might change the browser compatibility

iuioiua:
Ah, good point. I'll review the changes when I can.

kt3k:
The usages in test cases should be just fine. The usages in 'browser compatible' modules might be affecting some users (for example when they use them in island component of fresh)

iuioiua:
I just checked, and this PR only touches test files. So we're all good.

PR #3790, opened 2023-11-12
chore: trim CODEOWNERS

iuioiua:
See https://github.com/denoland/deno_std/pull/3734#discussion_r1390172621


kt3k:
In my view this file is useful. If the concern here is the maintenance cost of this file, maybe let's leave the only below line:

```
/*            @kt3k
```

iuioiua:
What do we lose by removing this file? And WDYT, @bartlomieju?

kt3k:
If the user name is listed in this file, that account will get 'participating' notification, which is useful in my case.

crowlKats:
Due to the streams folder, this is useful for me, and especially once #3119 lands, having this will be even more useful for me. We can just simplify the file to:
```
* @kt3k
/streams @kt3k @crowlKats 
```
Then there should be no concerns in terms of maintainability.

iuioiua:
Fair points. I've adjusted it to be the Leo's snippet.

kt3k (approved):
LGTM

PR #3791, opened 2023-11-12
docs(path): fix `@deprecated` note for `isGlob()`

uki00a:
This PR fixes a `@deprecated` comment so that `isGlob` is imported from [`std/path/is_glob.ts`](https://deno.land/std@0.206.0/path/is_glob.ts) instead of [`std/path/windows/is_glob.ts`](https://deno.land/std@0.206.0/path/windows/is_glob.ts).

### DIFF
--- a/path/glob.ts
+++ b/path/glob.ts
@@ -6,7 +6,7 @@ export type { GlobToRegExpOptions } from "./glob_to_regexp.ts";
 
 export {
   /**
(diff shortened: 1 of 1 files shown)


lino-levan (approved):
Ah that's my bad 😅. Nice catch, LGTM!

iuioiua (approved):
Nice! Thank you.

PR #3793, opened 2023-11-12
feat(crypto): support `BLAKE2B-160` algorithm

quentinadam:
Added BLAKE2B-160 algorithm and recompiled the crypto WASM with latest version of rust (1.73)


crypto/crypto_test.ts
```diff
@@ -506,6 +509,7 @@ const digestCases: [
   }, {
     BLAKE2B: Error,
     "BLAKE2B-128": Error,
+    "BLAKE2B-160": "70e8ece5e293e1bda064deef6b080edde357010f",
```
iuioiua, line 512: Should this be `Error`?

quentinadam: To be honest, I don't know. I guess it should be Error because the BLAKE2B algorith is not supposed to be called with a { length } option, but what is a bit odd is that the test pass in either case (whether "70e8ece5e293e1bda064deef6b080edde357010f" or Error).

iuioiua: Interesting... I've played around too. It seems this is the case for the other algorithms. Would you have any idea why this is happening, @jeremyBanks?

jeremyBanks: Uh-oh. Thanks for the ping. It looks like there is a bug in the test logic I wrote, and test cases which expect an `Error` are not being checked correctly. Yikes. My mistake. 😨 

The assertEquals at `crypto/crypto_test.ts:1383` can fail, but then the `AssertionError` it raises will be caught and treated the same as if it were thrown by the code being tested. If the test is expecting an `Error`, the `AssertionError` will satisfy this expectation, and the test will pass even if it should have failed due to the assertion. 

Fixing the test logic causes the existing test vectors for `BLAKE3`, `SHAKE128`, and `SHAKE256` to fail. I'll prepare a PR to correct this.

With the fix, the current code in this PR fails if this line is changed to `Error`. I think the current logic is that we accept the `length` parameter for non-variable-length algorithms as long as the specified length is the same as their output's fixed length. By that logic, this line does not need to be changed, and should remain `"70e8ece5e293e1bda064deef6b080edde357010f"`.

If you feel that this behavior is not desirable, and we should strictly forbid the `length` property for non-variable-length algorithms, that should probably to be changed in a separate PR because it will require a `BREAKING` change to the Rust logic and to require several other test cases to be updated.

iuioiua: Alright, let's keep it as is.

iuioiua (approved):
LGTM! Thank you, @quentinadam.

kt3k (approved):
LGTM

PR #3794, opened 2023-11-12
chore: update tool imports

lino-levan:


iuioiua (approved):
Nice! Thanks, Lino. I think we should centralise all the dev dependencies somewhere. It'd make management a little easier and open up the possibility of automatic dependency updates. I've created #3795 to explore this further.

PR #3796, opened 2023-11-12
fix: adjust `lint:deprecations` task permissions

iuioiua:
I believe #3794 introduced this need because of the upgrade to `deno_doc@0.73.0`.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -11,7 +11,7 @@
     "test": "deno test --doc --unstable --allow-all --parallel --coverage=./cov --trace-ops",
     "test:browser": "git grep --name-only \"This module is browser compatible.\" | grep -v deno.json | grep -v .github/workflows | grep -v _tools | xargs deno check --config browser-compat.tsconfig.json",
     "fmt:licence-headers": "deno run --allow-read --allow-write ./_tools/check_licence.ts",
-    "lint:deprecations": "deno run --allow-read --allow-net ./_tools/check_deprecation.ts",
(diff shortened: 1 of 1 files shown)


_tools/check_doc_imports.ts
```diff
@@ -14,6 +14,7 @@ const EXCLUDED_PATHS = [
   ".git",
   ".github",
   "_tools",
+  "README.md",
```
kt3k, line 17: `deno task lint:doc-imports` passes in my env without this change. Is this really necessary?

deno.json
```diff
@@ -11,7 +11,7 @@
     "test": "deno test --doc --unstable --allow-all --parallel --coverage=./cov --trace-ops",
     "test:browser": "git grep --name-only \"This module is browser compatible.\" | grep -v deno.json | grep -v .github/workflows | grep -v _tools | xargs deno check --config browser-compat.tsconfig.json",
     "fmt:licence-headers": "deno run --allow-read --allow-write ./_tools/check_licence.ts",
-    "lint:deprecations": "deno run --allow-read --allow-net ./_tools/check_deprecation.ts",
+    "lint:deprecations": "deno run --allow-read --allow-net --allow-env=HOME ./_tools/check_deprecation.ts",
```
kt3k, line 14: 👍

_tools/check_doc_imports.ts
iuioiua: Strange, it failed before without this line. I just tested it again, and it's fine too. I've reverted it now.

kt3k (approved):
LGTM

PR #3799, opened 2023-11-13
fix(crypto): validate that digest length must be an integer between 0 and isize::MAX

### LINKED ISSUE #3798: crypto: variable-length hash algorithms do not throw errors for non-integer/non-safe-integer lengths, tests cases assume they do, but test logic is broken and suppresses the assertion failures
@quentinadam and @iuioiua noticed in https://github.com/denoland/deno_std/pull/3793#discussion_r1390512613 that the `crypto_test.ts` logic was failing to check that errors were being thrown when expected.

I've made a commit that fixes the test logic: https://github.com/jeremyBanks/deno_std/commit/e571f76600b1a07652edab40ab21791d77b49b7d. Upon fixing it, three tests start to fail for two cases each: the three variable-length hash algorithms, `BLAKE3`, `SHAKE128`, and `SHAKE256` do not throw errors for the cases `Non-integer length with options {"length":1.5})` (which they treat as `length: 1`) and `Too-large length with options {"length":8.112963841460666e+31})` (which they treat as `length: 0`), but our current test cases expect them to throw errrors. 

The test logic definitely needs to be fixed, so that commit should be applied, but obviously that can't be merged until we also fix the `crypto.ts` logic to perform the expected checks (which _might_ be considered `BREAKING` if we're being super-pedantic, but is probably safe in practice), or to change the expected result.

Some might argue that that truncating `1.5` to `1` is okay (I would prefer it to be an error), but I think all would agree that any length that's too large to fit into 32 bits (`usize` for WASM targets) should throw an error instead of being truncated to `0`.

jeremyBanks:
Currently, if we call `crypto.subtle.digest`/`.digestSync` with a variable-length hash algorithm and length that is...

- too large to fit in 32 bits/greater than Rust's `usize::MAX` for WASM: it will be truncated, usually to `0`, producing an incorrect result.
- too large to fit in in Rust's `isize::MAX` for WASM: it will trigger an internal `capacity_overflow` error from Rust's `Vec`.
- negative: it will overflow and wrap around, usually resulting in a very-large value that will trigger an internal `capacity_overflow` error from Rust's `Vec`.
- not an integer: it will be truncated to the nearest integer value.

Some of this behaviour is a result of the JS-WASM translation layer, so this PR adds a check on the JavaScript side to throw a `RangeError` in these cases, before calling the Rust/WASM (which already checks that the length is valid for the specific algorithm being used).

This also resolves #3798, which describes how our tests cases were already expecting errors for these cases, but the test logic had a bug that was causing the assertion errors to be suppressed. This fixes that test logic.

One might argue that this is a `BREAKING` change since it produces an observable change in behaviour (at least for the non-integer case). I'm not sure it's a significant enough to count, but feel free to rename this PR to indicate that it's `BREAKING` if you disagree.


jeremyBanks:
These canary test failures don't _seem_ to be related to my changes..?

crypto/crypto.ts
```diff
@@ -328,6 +349,13 @@ const webCryptoDigestAlgorithms = [
 export type FNVAlgorithms = "FNV32" | "FNV32A" | "FNV64" | "FNV64A";
 export type DigestAlgorithmName = WasmDigestAlgorithm | FNVAlgorithms;
 
+/*
+ * The largest digest length the current WASM implementation can support. This
+ * is the value of `isize::MAX` on 32-bit platforms like WASM, which is the
+ * maximum allowed capacity of a Rust `Vec`.
+ */
+const maximumDigestLength = 0x7FFF_FFFF;
```
iuioiua, line 357:
```suggestion
const MAX_32BIT_INTEGER = 0x7FFF_FFFF;
```
Nit: To match `Number.MAX_SAFE_INTEGER`

iuioiua (changes requested):
Great work! Couple of nits.

crypto/crypto.ts
```diff
@@ -220,6 +220,17 @@ const stdCrypto: StdCrypto = ((x) => x)({
...
+      if (
+        length !== undefined &&
+        (length < 0 || length > maximumDigestLength ||
+          !Number.isInteger(length))
+      ) {
+        throw new RangeError(
+          `length must be an integer between 0 and ${maximumDigestLength}, inclusive`,
+        );
+      }
+
```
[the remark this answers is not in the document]
jeremyBanks: I'm a bit inclined to go with the names `assertValidDigestLength` and `MAX_DIGEST_LENGTH`, because I find the alternatives somewhat imprecise.

`assertIs32BitInteger` would make me think that it's either asserting the range of a 32-bit unsigned integer, or a 32-bit signed integer, but in fact it's enforcing the upper bound of a 32-bit signed integer but the lower bound of zero/that of a 32-bit unsigned integer, and also allowing it to be undefined (if we entirely extract this logic). So a more accurate name would be like `assertIsNonNegative32BitSignedIntOrUndefined`.

I'm happy to switch to the names you suggested if you still prefer them, but just wanted to share my reasoning.

crypto/crypto.ts
jeremyBanks: Changed to uppercase, but didn't rename yet per above comment.

crypto/crypto.ts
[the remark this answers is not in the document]
iuioiua: Ok, I see. Yep, your suggested name makes sense 👍🏾

crypto/crypto_test.ts
```diff
@@ -1379,15 +1383,32 @@ for (const algorithm of digestAlgorithms) {
...
+            fail(
+              `${algorithm} of ${caption}${
+                i > 0 ? ` (but not until variation [${i}]!)` : ""
+              } with options ${
+                JSON.stringify(options)
+              }) expected an exception of type ${expected.name}, but none was thrown.`,
+            );
           }
+        } else {
+          throw new TypeError("expected value has an unexpected type");
```
iuioiua, line 1411: Note to self: these tests are a little difficult to follow. Perhaps, we can make things a little more _digestible_ in a follow-up PR 🤣

crypto/crypto.ts
```diff
@@ -328,6 +333,29 @@ const webCryptoDigestAlgorithms = [
...
+  if (
+    value !== undefined &&
+    (value < 0 || value > MAX_DIGEST_LENGTH ||
+      !Number.isInteger(value))
+  ) {
+    throw new RangeError(
+      `length must be an integer between 0 and ${MAX_DIGEST_LENGTH}, inclusive`,
+    );
+  }
+};
```
iuioiua, line 357:
```suggestion
function assertValidDigestLength(value?: number) {
  if (
    value !== undefined &&
    (value < 0 || value > MAX_DIGEST_LENGTH ||
      !Number.isInteger(value))
  ) {
    throw new RangeError(
      `length must be an integer between 0 and ${MAX_DIGEST_LENGTH}, inclusive`,
    );
  }
}
```
See https://docs.deno.com/runtime/manual/references/contributing/style_guide#top-level-functions-should-not-use-arrow-syntax

iuioiua (approved):
LGTM! Thank you, @jeremyBanks. Thorough work, as per usual.

crypto/crypto.ts
jeremyBanks: D’oh, I keep forgetting this. Thanks for the reminder.

kt3k (approved):
LGTM

Issue #3782, opened 2023-11-08
tracking: remove use of TypeScript enums


iuioiua:
The main reason is that objects align more with native JavaScript than enums. There are also other technical reasons for objects being more desirable over enums. Union types may also work, depending on the scenario.

Enums show up in the following spots:
- [x] assert/_diff.ts
- [x] console/_tools/generate_data.ts
- [x] datetime/week_of_year.ts
- [x] fmt/printf.ts
- [x] fs/eol.ts
- [x] http/http_status.ts and http/status.ts (see #3781)
- [x] jsonc/parse.ts
- [x] log/levels.ts
- [x] toml/stringify.ts

Note: the `enum`s in front_matter/_format.ts  and archive/_common.ts are deliberately omitted as the corresponding implementations will soon be deprecated.

References:
- https://www.typescriptlang.org/docs/handbook/enums.html#objects-vs-enums
- https://youtu.be/jjMbPt_H3RQ?feature=shared
- https://youtu.be/3aUHskjF7pc?feature=shared

Nkiriobasi:
Has this been fixed? if not, can I fix it.

iuioiua:
Great! Yep, it's still available. Would you like me to assign you?

Nkiriobasi:
Yes

realpha:
hi, is this is still open? (Seems to be without assignee). Could I take a stab at this?

Would it be possible to convert the tasks into individual issues or would you prefer to just reference the tracking issue in the individual prs?

iuioiua:
Best that we tackle them in separate PRs and reference this issue.

PR #3800, opened 2023-11-13
BREAKING(fs): replace `EOL` enum with OS-dependant constant

### LINKED ISSUE #3583: OS-normalized EOL
In Node we have [os.EOL](https://nodejs.org/api/os.html#os_os_eol) which abstracts the OS.

In Deno we have [std/fs/eol](https://deno.land/std/fs/eol.ts) which is an os-invariant enumeration. The [docs for writeFile](https://deno.land/typedoc/index.html#writefile) and similar just use denormalized `\n`.

Is this a case where we can prefer the Node pattern over the Go pattern, or is there some other recommended way to do this sort of thing? I'm simply trying to write a file with line breaks.

iuioiua:
This change replaces the `EOL` enum with an OS-dependent constant. The breaking change is made without deprecation because both implementations share the same `EOL` name. Frankly, I'm uncertain this is the best approach, so I'm open to similar ideas.

Closes #3583
Towards #3782

### DIFF
--- a/fs/eol.ts
+++ b/fs/eol.ts
@@ -1,5 +1,4 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-// This module is browser compatible.
 
 /** End-of-line character for POSIX platforms like Linux and macOS. */
 export const LF = "\n" as const;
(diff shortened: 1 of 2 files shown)


iuioiua:
Closing in favour of #3809.

Issue #3541, opened 2023-08-15
timingSafeEqual is provided from 2 places


kt3k:
ref https://github.com/denoland/deno_std/pull/2503

`timingSafeEqual` is provided from 2 places:
- as a separate function from `crypto/timing_safe_equal.ts`
- as a method of subtle crypto extension from `crypto/crypto.ts`

Do we need to keep both?

iuioiua:
I'd say that it should only be exported from `crypto/timing_safe_equal.ts` as [it doesn't seem apparent that `timingSafeEqual()` will be added to the Web Crypto API](https://github.com/w3c/webcrypto/issues/270).

PR #3803, opened 2023-11-15
BREAKING(crypto): remove `crypto.subtle.timingSafeEqual()` method

iuioiua:
Note: this doesn't remove the `timingSafeEqual()` function, which is still available. This only removes the `crypto.subtle.timingSafeEqual()` method from the `crypto.subtle` interface.

See #3541


kt3k (approved):
LGTM. Thanks!

PR #3806, opened 2023-11-15
test(http): fix op leak in file_server_test

kt3k:
The main CI looks failing because of the op leak in `serveDir() traverses path` test case with canary Deno CLI. This PR fixes it.


iuioiua (approved):
LGTM. Nice 👍🏾

PR #3807, opened 2023-11-15
docs(README): add note about deprecation

kt3k:


iuioiua:
Ah! This is that error I was talking about in #3796.

bartlomieju (approved):

PR #3808, opened 2023-11-16
chore: cleanup after removal of `std/wasi`

iuioiua:
This cleans up some remnants of the `std/wasi` deprecation and removal, some of which had to happen after the actual removal.


kt3k (approved):
LGTM. Thanks!

PR #3809, opened 2023-11-16
BREAKING(fs): deprecate `EOL` enum

### LINKED ISSUE #3583: OS-normalized EOL

iuioiua:
The `EOL` enum will be replaced by an OS-dependent constant in 0.209.0.

Closes #3583
Towards #3782

### DIFF
--- a/fs/eol.ts
+++ b/fs/eol.ts
@@ -1,7 +1,33 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-/** Platform-specific conventions for the line ending format (i.e., the "end-of-line"). */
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM 👍

PR #3813, opened 2023-11-16
chore(crypto): cleanup tests

iuioiua:
This aims to rely on higher-level methods and reduce code duplication.


kt3k (approved):
Nice clean ups!

PR #3627, opened 2023-09-07
Feat: Add expect syntax by wrapping assert syntax

Sorikairox:
Contains the following. Sadly couldn't make time to write tests (life happened). 

- [x] [expect(value)](https://jestjs.io/docs/expect#expectvalue)
- [x]  [.not](https://jestjs.io/docs/expect#not)
- [x]  [.resolves](https://jestjs.io/docs/expect#resolves)
- [x]  [.rejects](https://jestjs.io/docs/expect#rejects)
- [x]  [.toBe(value)](https://jestjs.io/docs/expect#tobevalue)
- [ ]  [.toHaveLength(number)](https://jestjs.io/docs/expect#tohavelengthnumber)
- [ ]  [.toHaveProperty(keyPath, value?)](https://jestjs.io/docs/expect#tohavepropertykeypath-value?)
- [x]  [.toBeCloseTo(number, numDigits?)](https://jestjs.io/docs/expect#tobeclosetonumber-numdigits?)
- [x]  [.toBeDefined()](https://jestjs.io/docs/expect#tobedefined)
- [x]  [.toBeFalsy()](https://jestjs.io/docs/expect#tobefalsy)
- [x]  [.toBeGreaterThan(number | bigint)](https://jestjs.io/docs/expect#tobegreaterthannumber--bigint)
- [x]  [.toBeGreaterThanOrEqual(number | bigint)](https://jestjs.io/docs/expect#tobegreaterthanorequalnumber--bigint)
- [x]  [.toBeLessThan(number | bigint)](https://jestjs.io/docs/expect#tobelessthannumber--bigint)
- [x]  [.toBeLessThanOrEqual(number | bigint)](https://jestjs.io/docs/expect#tobelessthanorequalnumber--bigint)
- [x]  [.toBeInstanceOf(Class)](https://jestjs.io/docs/expect#tobeinstanceofclass)
- [x]  [.toBeNull()](https://jestjs.io/docs/expect#tobenull)
- [x]  [.toBeTruthy()](https://jestjs.io/docs/expect#tobetruthy)
- [x]  [.toBeUndefined()](https://jestjs.io/docs/expect#tobeundefined)
- [x]  [.toBeNaN()](https://jestjs.io/docs/expect#tobenan)
- [ ]  [.toContain(item)](https://jestjs.io/docs/expect#tocontainitem)
- [ ]  [.toContainEqual(item)](https://jestjs.io/docs/expect#tocontainequalitem)
[...]

PR #3814, opened 2023-11-16
feat: add std/expect

kt3k:
Continued from #3627

This PR adds `std/expect` mostly compatible with `expect` of jest.

Remaining:
- [x] toHaveLength
- [x] toHaveProperty
- [x] toContain
- [x] toContainEqual
- [x] Fix toBeCloseTo tolerance handling

Mock related:
- [x] fn (mock)
- [x] toHaveBeenCalled
- [x] toHaveBeenCalledTimes
- [x] toHaveBeenCalledWith
- [x] toHaveBeenLastCalledWith
- [x] toHaveBeenNthCalledWith
- [x] toHaveReturned
- [x] toHaveReturnedTimes
- [x] toHaveReturnedWith
- [x] toHaveLastReturnedWith
- [x] toHaveNthReturnedWith

Testing:
- [x] add test cases

We'd avoid supporting snapshot related APIs for first pass (`toMatchSnapshot` `toMatchInlineSnapshot` `toThrowErrorMatchingSnapshot` `toThrowErrorMatchingInlineSnapshot`)

Done:
- [x] documentation

closes #3813 1779


andrewthauer:
Very much looking forward to this. Curious though, would it make sense to put this under `std/testing` or may `std/testing/expect` instead?

iuioiua:
We thought the `expect()` implementation, once complete with testing, would be of a sufficient size that it might justify having it as its own sub-module. Plus, it'd be more visible.

kt3k:
> Curious though, would it make sense to put this under std/testing or may std/testing/expect instead?

I'm in favor of keeping the implementation in `std/expect`, but I'm open to the idea of re-exporting this from somewhere in `std/testing`, such as from `std/testing/bdd.ts`.

andrewthauer:
Out of curiosity, is the plan to keep all modules at the top level or will there be any nesting?

I personally think I'd find it more intuitive to scan top level areas first, then drill into them further. For example I'd find `testing` and then look inside to see `bdd` and `expect`, etc.. A possible structure might look like:

```
└── testing
    ├── bdd
    │   └── mod.ts
    ├── expect
    │   └── mod.ts
    └── mod.ts

```

kt3k:
We generally prefer top level modules over nested structure. For example, we moved some complex modules from `encoding/` to top-level to avoid the nested structure in the past. cf https://github.com/denoland/deno_std/issues/3123

andrewthauer:
Right. For encoding that makes sense to me. I personally wouldn't think to look under that initially for json,yaml, etc.

I would say that `expect` is much less clear to me own its own at the top level then the encoding ones. Putting it under testing would give it context and make it more clear. My 2 cents anyways. I'll will say no more now 😀.

expect/_to_be.ts
```diff
@@ -0,0 +1,13 @@
...
+import { assertNotStrictEquals } from "../assert/assert_not_strict_equals.ts";
+import { assertStrictEquals } from "../assert/assert_strict_equals.ts";
+
+export function toBe(context: MatcherContext, expect: unknown): MatchResult {
+  if (context.isNot) {
+    assertNotStrictEquals(context.value, expect, context.customMessage);
+  } else {
+    assertStrictEquals(context.value, expect, context.customMessage);
+  }
+}
```
bartlomieju, line 13: What's the reason for splitting into so many small modules?

expect/_to_be_close_to.ts
```diff
@@ -0,0 +1,35 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+import { MatcherContext, MatchResult } from "./_types.ts";
+import { AssertionError } from "../assert/assertion_error.ts";
+
+// TODO(kt3k): tolerance handling is wrong
```
bartlomieju, line 6: Is there a test that covers this tolerance?

bartlomieju (commented):
This looks mostly good to me, I'm a bit worried about a number of modules that were added - if you're unit testing a file that has no dependencies besides pulling `expect/mod.ts` then the overhead of loading all these small modules might be non-trivial. Could you reconsider adding a single module that has all the matchers inside it?

expect/_to_be_close_to.ts
kt3k: Oops, this has been already resolved in https://github.com/denoland/deno_std/pull/3814/commits/357ecb3f2cd0f2ab9a2748cc7793c6f9f8f3c1ef

kt3k:
> if you're unit testing a file that has no dependencies besides pulling expect/mod.ts then the overhead of loading all these small modules might be non-trivial.

Fair enough. I'll move all matchers into `_matchers.ts`

bartlomieju (approved):

kt3k:
Landing this as `std/expect` for now. Let's revisit `std/expect` vs `std/testing/expect` before stabilizing this.

PR #3817, opened 2023-11-18
docs(path): correct `globToRegExp()` deprecation notice

GregTCLTK:


iuioiua (approved):
Nice catch! Thank you

Issue #3764, opened 2023-11-02
tracking: complete documentation


iuioiua:
Documentation is essential for obvious reasons and key for the stabilization of packages. Requirements for each package:
1. Pass without error for public entrypoints using `deno doc --lint <file>`
2. Adhere to the [documentation guidelines](https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md#documentation).

All packages need to have their documentation complete and validated:
- [x] archive #5636
- [x] assert #4876 
- [x] async
- [x] bytes
- [x] cli #4815
- [x] collections
- [x] crypto #4884
- [x] csv #4920
- [x] data-structures #4847 
- [x] datetime
- [x] dotenv #5095
- [x] encoding #4838 
- [x] expect #4836 
- [x] fmt #4829 
- [x] front-matter #4789 
- [x] fs #4788
- [x] html #4878
- [x] http
- [x] ini #5020 
- [x] io #5656
- [x] json #4798
- [x] jsonc #4814 
- [ ] log
- [x] media-types
- [x] msgpack #4832
- [x] net #4982
- [x] path #4900
- [x] regexp
- [x] semver #4846
- [x] streams #4852
- [x] testing #5033
- [x] text #4837 
- [x] toml #4799
- [x] ulid #4825 
- [x] url
- [x] uuid #4790 
- [x] webgpu
- [x] yaml

lucacasonato:
xpost: #3878

iuioiua:
It should be up-to-date as of now.

babiabeo:
Do you think we should use `{@link}` tag instead of markdown hyperlink?

kt3k:
related #4299

iuioiua:
Update: I've reset the checklist. The remaining packages should be validated to meet the updated documentation requirements.

kt3k:
Can we close this now?

iuioiua:
Not yet. The remaining packages should be documented, even if they're not getting stabilized. Even if they end up getting removed, documenting them will give us a better understanding of them too.

iuioiua:
`@std/log` is no longer around. Closing :)

PR #3818, opened 2023-11-19
refactor(collections): unify function overloads

iuioiua:
This refactor combines multiple function overloads within `std/collections`. The functions are now easier to understand, and the documentation required for this module has been reduced.

More functions can be cleaned up in follow-up PRs.

Towards #3764


bartlomieju (approved):
LGTM, really good clean up 👍

PR #3821, opened 2023-11-19
chore: centralise dev dependencies in `deno.json`

### LINKED ISSUE #3795: suggestion: centalise external dev dependencies
Having a `dev_deps.ts` or similar somewhere containing all the external development dependencies is a good idea. It'd allow us to manage/update these dependencies, even do it automatically. External dependencies will have to be audited to determine which approach is best.

iuioiua:
Did I miss any?

Closes #3795


bartlomieju (approved):
Cool, LGTM

PR #3823, opened 2023-11-21
docs(collection): document overloads (revert #3818)

kt3k:
#3818 removed overloads from several functions, but they seemed useful. Now these functions accept selectors of mixed return types such as `maxBy(arr, (item) => item === null ? 10 : "foo")` and what happens with these calls are not obvious (also not documented). I think we shouldn't allow these calls.

This PR instead documents all overloads of these functions towards #3764


iuioiua (approved):
LGTM

PR #3825, opened 2023-11-21
chore: configure dependabot for `std/crypto`

iuioiua:
This aims to automate keeping Rust dependencies for `std/crypto/_wasm` up-to-date.


kt3k (approved):
LGTM. Let's try

PR #3826, opened 2023-11-21
`assertVersions` for requiring a min/max version of Deno

jeff-hykin:
Note: parseSemver doesn't work for parsing the v8 version, which is the only reason I created `parseVersion`

I don't care where these functions are stored or how they're named.

### DIFF
--- a/assert/assert_versions.ts
+++ b/assert/assert_versions.ts
@@ -1,0 +1,107 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+import { parseVersion } from "../text/parse_version.ts";
+import { zip } from "../collections/zip.ts";
+import { AssertionError } from "./assertion_error.ts";
+
+/**
+ * Asserts that the specified versions of the given systems meet the required criteria.
+ *
+ * @example
+ * ```ts
+ * import { assertVersions } from "https://deno.land/std@$STD_VERSION/assert/assert_versions.ts";
+ * assertVersions({
+ *   deno: {
+ *     min: [1, 25],
+ *     max: [4],
+ *   },
+ *   v8: {
+ *     max: [20],
+ *     min: [11, 6],
+ *   },
+ *   typescript: {
+ *     max: [99],
+ *     min: [2, 3, 4],
+ *   },
+ * });
+ * ```
+ * @param {Object} options - An object containing version information for different systems.
+ * @param {Object} options.typescript - Version information for TypeScript.
+ * @param {number[]} [options.typescript.min] - Minimum required version for TypeScript.
+ * @param {number[]} [options.typescript.max] - Maximum allowed version for TypeScript.
+ * @param {Object} options.v8 - Version information for V8.
+ * @param {number[]} [options.v8.min] - Minimum required version for V8.
+ * @param {number[]} [options.v8.max] - Maximum allowed version for V8.
+ * @param {Object} options.deno - Version information for Deno.
+ * @param {number[]} [options.deno.min] - Minimum required version for Deno.
+ * @param {number[]} [options.deno.max] - Maximum allowed version for Deno.
+ *
+ * @throws {AssertionError} Throws an AssertionError if any system's version does not meet the specified criteria.
+ */
+export function assertVersions(options: {
+  typescript?: {
+    min?: number[];
+    max?: number[];
+  };
+  v8?: {
+    min?: number[];
+    max?: number[];
+  };
+  deno?: {
+    min?: number[];
+    max?: number[];
+  };
+}): void {
+  for (const [whichSystem, minMax] of Object.entries(options)) {
+    const { min, max } = { ...minMax };
+    const versionInfo = { ...Deno?.version }[whichSystem];
+    if (typeof versionInfo == "string") {
+      const existingVersion = parseVersion(versionInfo);
+      if (min instanceof Array) {
+        const throwMin = () => {
+          throw new AssertionError(
+            `Current version of ${whichSystem} is ${versionInfo}, however the code requires a minimum version of ${
+              min.join(".")
+            }`,
+          );
+        };
+        for (const [eachExisting, eachMin] of zip(existingVersion, min)) {
+          if (typeof eachExisting !== typeof eachMin) {
+            throwMin();
+          } else if (typeof eachExisting == "string") {
+            // @ts-ignore it complains "dont do string!=number" but beacuse of the if statement above this will never have that case
+            if (eachExisting !== eachMin) {
+              throwMin();
+            }
+          } else if (typeof eachExisting == "number") {
+            if (eachExisting < eachMin) {
+              throwMin();
+            }
+          }
+        }
+      }
+      if (max instanceof Array) {
+        const throwMax = () => {
+          throw new AssertionError(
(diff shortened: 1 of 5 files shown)


kt3k:
This looks too specific to certain purpose to me. Most assertions in `std/assert` deal with very general concept such as equality, inclusion, matching, etc..

jeff-hykin:
> std/assert deal with very general concept such as equality, inclusion, matching, etc..

Okay, where can I move it / what can I call it so that it won't be criticized?

jeff-hykin:
I should also mention: For security reasons, this functionality can't be handled by a 3rd party repo.

I know that might be surprising so let me explain. The PR is based on [this module](https://github.com/hyperupcall/ensure). e.g.

```
import { ensure } from 'https://deno.land/x/ensure/mod.ts'
ensure({
  denoVersion: "1.0.0",
  typeScriptVersion: "3.8"
})
```

The problem is that^ import NEEDS to be un-pinned.

For example, lets say we pin the import and have `assertVersions({deno: {max: [2]})`. Well there's no guarentee that Deno 3.0 will still use `Deno.versions.deno`,  meaning the pinned module might be completely broken on Deno 3.0 instead of doing its job of reporting that the version is too high.

That problem can be addressed with an unpinned import, and updating assertVersions to handle the (theoretical) Deno 3.0 API.

But that^ module is just some random 3-start repo. Having that assertion as an unpinned import at the top of all of my deno scripts would be a major security risk.

kt3k:
In my view this util is too specific to certain purpose, compared to other std modules. (std usually deals with js features (`async`, `bytes`, etc), other languages/protocols (`yaml`, `semver`, etc), common practices (`cli`, `dotenv`, etc).

I'm open to including this in std if this has enough support from the community, but I haven't seen this feature requested or discussed in the past.

jeff-hykin:
Can I request a separate deno-offical repo for this functionality? There's gotta be some offical way to require a min/max deno version.

I don't understand how Deno cares so much about pinning versions but forgets to pin itself.

Could I add a cross-runtime OS check (e.g. `isMac` that utilizes either `Deno.build` or
`navigator.platform`) and then put this in that section?

timreichen:
It seems to me that this is a trivial problem to solve with a simple string comparison:
```ts
const REQUIRED_DENO_VERSION = "1.2";
const denoVersion  = Deno.version.deno
assert(denoVersion >= REQUIRED_DENO_VERSION, `deno version must be at least ${REQUIRED_VESION} but is ${denoVersion}`);

const REQUIRED_V8_VERSION = "12.0.2";
// ...
const REQUIRED_TYPESCRIPT_VERSION = "5";
// ...
```
If so I don't think this needs to implemented in std.

Or did I miss something?

realpha:
Also, this is easier more robust than to just mention the `minimum required version` in a README

babiabeo:
In my opinion, the problem is that people might have to add the `assertVersions` function to every single file since Deno allows importing any file via http URLs. This will cause some inconvenience during development, especially with large projects like `deno_std`.

realpha:
I think this is something to be discussed in the context of deno and not `deno_std` as it codifies a convention, which could have ramificatios for the ecosystem as whole.

jeff-hykin:
> It seems to me that this is a trivial problem to solve with a simple string comparison: 
>
```> const REQUIRED_DENO_VERSION = "1.2";
> assert(
>   Deno.version.deno >= REQUIRED_DENO_VERSION,
>   `deno version must be at least ${REQUIRED_DENO_VERSION} but is ${Deno.version.deno}`,
> );
> ```
> did I miss something?

Yes:

```
> "1.2" > "1.10"
true
```

> Also, this is easier and more robust than to just mention the minimum required version in a README

Not if it's an upstream dependency. E.g. `blah` tool uses fsExtra, fsExtra requires X version.

> library authors to ensure their consumers don't run into trouble

As a consumer I have run into this problem and it was not obvious how to fix (wasn't obvious it was a deno version issue)

Also if `blah` updates its dependencies, and newer versions of fsExtra have newer requirements, they're going to need to go read the readme's to figure that out instead of getting the check for free.

jeff-hykin:
> people might have to add the `assertVersions` function to every single file since Deno allows importing any file via http URLs.

I think this is a great point. Both from a dev experience and runtime/code-complexity perspective I would much prefer some kind of builtin check, or official way to add version constraints in the `deno.jsonc`.

realpha:
> > It seems to me that this is a trivial problem to solve with a simple string comparison:
> > ```
> > const REQUIRED_DENO_VERSION = "1.2";
> > assert(
> >   Deno.version.deno >= REQUIRED_DENO_VERSION,
> >   `deno version must be at least ${REQUIRED_DENO_VERSION} but is ${Deno.version.deno}`,
> > );
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > did I miss something?
> 
> Um, yes @timreichen
> 
> ```
> > "1.2" > "1.10" // e.g.  (2 > 10)
> true
> ```
> 
> _
> 
> > Also, this is easier and more robust than to just mention the minimum required version in a README
> 
> @realpha Not if it's an upstream dependency. E.g. `blah` tool uses fsExtra, fsExtra requires X version.
> 
> Also if `blah` updates its dependencies, and newer versions of fsExtra have newer Deno requirements, they're going to need to go read the readme's to figure that out instead of getting the check for free.
> 
> _
> 
> > library authors to ensure their consumers don't run into trouble
> 
> As a consumer I have run into this problem. It was also not obvious how to fix, I had to open an issue on the module and then author and I debugged it together until we realized our deno versions were different.

That's what I meant. A way to enforce or warn about a version constraint akin to how it's handled in cargo or npm would be optimal. Sry for the misunderstanding

jeff-hykin:
> > As a consumer I have run into this problem. It was also not obvious how to fix, I had to open an issue on the module and then author and I debugged it together until we realized our deno versions were different.
> 
> That's what I meant. A way to enforce or warn about a version constraint akin to how it's handled in cargo or npm would be optimal. Sry for the misunderstanding

Oh I see, that was actually my mistake misreading it! I missed the `than`

bartlomieju:
Leaving a comment since I was pinged offline.

> I would much prefer some kind of builtin method, like the JSX comments, or official way to add version constraints in the deno.jsonc. This checker function is very much a workaround caused by having no other way to assert the requirement.

That is reasonable and I would encourage you to open a feature request in main Deno repository - we are working on new way to distribute code (and the standard library will use that mechanism), but I can definitely relate to the fact that you might be pinned to an older version for whatever reason (distro lagging, admins taking time to allow list newer versions of Deno).

I don't have a strong opinion if this function should be in `asserts/` module or any other, but I can see it being useful for people. That said, I'm not sure I agree with adding these these to every single file in the standard library. It's gonna be a nightmare to update and keep up-to-date. With `deno_std` inching closer to stability (1.0 release for most of the modules) we will guarantee that they work with all future versions of Deno without breaking changes. Probably a field in `deno.json` would be a good solution here - still I wouldn't hard-error, but maybe print a warning.

kt3k:
> I would much prefer some kind of builtin method, like the JSX comments, or official way to add version constraints in the deno.jsonc. This checker function is very much a workaround caused by having no other way to assert the requirement.

I agree with this. This should be discussed in Deno CLI context. Let's not start with a workaround.

Created https://github.com/denoland/deno/issues/21321

PR #3834, opened 2023-11-21
BREAKING: deprecate `std/http/method`

### LINKED ISSUE #3792: suggestion: deprecate `std/http/methods`
Does anyone use/need [`std/http/methods`](https://deno.land/std@0.206.0/http/method.ts)? If not, we should remove it.

iuioiua:
Closes #3792


kt3k (approved):
LGTM

PR #3835, opened 2023-11-22
refactor(fs): correct test names

JakeAve:
Updating the `fs` test descriptions per #3754. Also creating some more parity between async and sync functions.


fs/copy_test.ts
```diff
@@ -47,7 +47,7 @@ function testCopySync(name: string, cb: (tempDir: string) => void) {
 }
 
 testCopy(
-  "[fs] copy file if it does no exist",
+  "copy() throws if src does not exist",
```
iuioiua, line 50:
```suggestion
  "copy() rejects if src does not exist",
```
Ditto for all other async functions which reject.

iuioiua (changes requested):
Thanks for this, @JakeAve. Nit: can you please tweak the verb for async functions from `throws` to `rejects`?

fs/eol_test.ts
```diff
@@ -9,36 +9,37 @@ const LFinput = "deno\nis not\nnode";
 const NoNLinput = "deno is not node";
 
 Deno.test({
-  name: "[EOL] Detect CR LF",
+  name: "detect() detects CR LF as the end-of-line character",
```
iuioiua, line 12:
```suggestion
  name: "detect() detects CRLF as the end-of-line character",
```

fs/walk_test.ts
iuioiua, line 1: Excellent! Thanks for adding the missing tests!

iuioiua (approved):
Nice work! Thanks for this

PR #3836, opened 2023-11-22
refactor(datetime): replace enum with object

realpha:
This PR replaces the used typescript enum in the `datetime` module with a 
as `const object` as requested [here](https://github.com/denoland/deno_std/issues/3782)

As this enum is only used internally, merging this change wouldn't constitute a
breaking change.

History:
- chore(root): ignore .tool-versions
- feat(datetime): replace enum with const object

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ deno.lock
 /console/testdata/unicode_width_crate/target
 html_cov/
 cov.lcov
+.tool-versions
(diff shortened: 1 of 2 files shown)


.gitignore
iuioiua, line 1: Can you please revert this change? Best we stick to the objective of the PR 🙂

iuioiua (changes requested):

.gitignore
realpha: sure thing! Out of curiosity: Is the convention in deno_std to always use the latest `deno` cli version?

iuioiua: Yep, the latest version is always best 👍🏾

iuioiua (approved):
LGTM! Thank you 🙏🏾

PR #3837, opened 2023-11-22
refactor(jsonc): replace `enum` with string union type

realpha:
This PR replaces the used typescript enum in the `jsonc` module with a
`const object` as requested  [here](https://github.com/denoland/deno_std/issues/3782).

As this enum is only used internally, merging this change wouldn't constitute a
breaking change.

This change seems a little bit clunky but does the trick.

### DIFF
--- a/jsonc/parse.ts
+++ b/jsonc/parse.ts
@@ -51,30 +51,32 @@ export function parse(
   return new JSONCParser(text, { allowTrailingComma }).parse();
 }
 
-enum TokenType {
-  BeginObject,
-  EndObject,
-  BeginArray,
-  EndArray,
-  NameSeparator,
-  ValueSeparator,
-  NullOrTrueOrFalseOrNumber,
-  String,
-}
+const TokenType = {
+  BeginObject: "BeginObject",
+  EndObject: "EndObject",
+  BeginArray: "BeginArray",
+  EndArray: "EndArray",
+  NameSeparator: "NameSeparator",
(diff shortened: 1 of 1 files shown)


iuioiua (changes requested):
Would changing `TokenType` to a union type make more sense in this case?

realpha:
> Would changing `TokenType` to a union type make more sense in this case?

IMO yes. Wasn't sure if the convetion in the codebase in general is to prefer maps over simple string unions.

Will push the changes

jsonc/parse.ts
```diff
@@ -51,30 +51,32 @@ export function parse(
...
+  EndObject: "EndObject",
+  BeginArray: "BeginArray",
+  EndArray: "EndArray",
+  NameSeparator: "NameSeparator",
+  ValueSeparator: "ValueSeparator",
+  NullOrTrueOrFalseOrNumber: "NullOrTrueOrFalseOrNumber",
+  String: "String",
+} as const;
+
+type TokenType = keyof typeof TokenTypeMap;
```
iuioiua, line 65: I think it'd be simpler if we just had `TokenType` be a string union type without the intermediary `TokenTypeMap`. A quick skim suggested that might be fine.

iuioiua (changes requested):

jsonc/parse.ts
realpha: Done

iuioiua (approved):
LGTM! Thank you.

PR #3838, opened 2023-11-22
refactor(log): replace enums and update types

realpha:
This PR encompasses 2 changes:
1. The replacement of the `LogLevels` enum with a `const object` as requested [here](https://github.com/denoland/deno_std/issues/3782)
2. The subsequent introduction of a dedicated `LogLevel` union type 

This new union type isn't currently exposed by the module API, but transitively by `LogRecord` and `LogRecordOptions`.
Therefore this change can be considered breaking (although it doesn't alter the runtime behavior whatsoever).

If you are generally ok with these changes: Should the `LogLevel` union get exposed through the module API? Would make sense IMO

History:
- feat(log): replaced LogLevels enum with object and enforced stricter types
- feat(log): updated types
- chore(log): updated types in tests


iuioiua (approved):
LGTM. Nicely done. Thank you. @kt3k, how do you think we should approach the breaking change?

kt3k:
The existing `LogLevels` are translated like the below (according to [ts playground](https://www.typescriptlang.org/play?#code/KYDwDg9gTgLgBMAdgVwLZwDIQOYeAN2ABsBnOAbwCg44A5AeQBUBlAUUbgF44AGAGmpwAIqwBCAVQDiXOAEZ+ggJK0AYvRkAmBTQDqAQQBKtZdO4BmbXFYGD9AzIAslgMIHFjRc70YZAVgUAvkA):

```js
export var LogLevels;
(function (LogLevels) {
    LogLevels[LogLevels["NOTSET"] = 0] = "NOTSET";
    LogLevels[LogLevels["DEBUG"] = 10] = "DEBUG";
    LogLevels[LogLevels["INFO"] = 20] = "INFO";
    LogLevels[LogLevels["WARNING"] = 30] = "WARNING";
    LogLevels[LogLevels["ERROR"] = 40] = "ERROR";
    LogLevels[LogLevels["CRITICAL"] = 50] = "CRITICAL";
})(LogLevels || (LogLevels = {}));
```

I think we need to include reverse mapping from number to enum name like the below to avoid the breaking change:
```ts
export const LogLevels = {
  NOTSET: 0,
  DEBUG: 10,
  INFO: 20,
  WARNING: 30,
  ERROR: 40,
  CRITICAL: 50,
  0: "NOTSET",
  10: "DEBUG",
  20: "INFO",
  30: "WARNING",
  40: "ERROR",
  50: "CRITICAL",
} as const;
```

And probably we can deprecate `0`, `10`, .. part.

realpha:
Thx for the feedback! Just to clarify, I have 3 questions regarding this (and also the remaining PRs):
1. Was "not introducing breaking changes" a design goal here? To my knowledge, the reverse mapping feature of typescript `enum`s is also one of the reasons one shouldn't use them. Speaking generally now, replacing publicly exposed `enum`s (also for the remaining PR) by `const object`s will always be a breaking change - unless the reverse mapping is preserved. Which also entails that additional types (at least 2) have to be exposed to facilitate working with this mapping relationships as unions. IMO this is not really warranted, and especifically in this case because:
  - the exposed functions `getLevelByName` and `getLevelName` facilitate this capability
  - the resulting const object (incl. reverse mapping) would be rather clumsy and usage quite unidiomatic because of the asymetric property access
 ``` typescript
  const level = LogLevels.DEBUG; //or LogLevels["DEBUG"];
  const name = LogLevels[0];
```
2. Should I revert the change regarding the introduction of the explicit `LogLevel` type? Because this is a breaking change as well.
3. As far as I've understood it, the deprecation notice should be added to every numeric key, not just `0` and `10`, right?

iuioiua:
> 1. Was "not introducing breaking changes" a design goal here? To my knowledge, the reverse mapping feature of typescript `enum`s is also one of the reasons one shouldn't use them. Speaking generally now, replacing publicly exposed `enum`s (also for the remaining PR) by `const object`s will always be a breaking change - unless the reverse mapping is preserved. Which also entails that additional types (at least 2) have to be exposed to facilitate working with this mapping relationships as unions. IMO this is not really warranted, and especifically in this case because:
> 
> * the exposed functions `getLevelByName` and `getLevelName` facilitate this capability
> * the resulting const object (incl. reverse mapping) would be rather clumsy and usage quite unidiomatic because of the asymetric property access
> 
> ```ts
>  const level = LogLevels.DEBUG; //or LogLevels["DEBUG"];
>  const name = LogLevels[0];
> ```

Yes, the aim is not to introduce breaking changes. We can introduce non-breaking changes and add a deprecation notice that lasts ~3 minor versions until the breaking change is made.

> 2\. Should I revert the change regarding the introduction of the explicit `LogLevel` type? Because this is a breaking change as well.

Yeah, I think so. But we can add a deprecation notice that it'll change in the future (0.211.0). Example [here](https://github.com/denoland/deno_std/blob/199cff0db562340ad3bea90f8bdd7e9f712dca5e/io/buf_writer.ts#L43).

> 3\. As far as I've understood it, the deprecation notice should be added to every numeric key, not just `0` and `10`, right?

Yes, correct.

realpha:
Oops, seems like my latest rebase gone wrong - will revert. Sry!

realpha:
I'll just cherry pick my commits and open a new PR, sry for the inconvenience

PR #3840, opened 2023-11-23
refactor(toml): replace ArrayType enum with union

realpha:

Converted the `enum` simply to a string literal union. IMO, a `const obejct`
and a dedicated type for the discriminants would be preferable though (avoiding
string allocations in potential hot paths)

Only used internally.


realpha:
@iuioiua What's your opinion on this? Should we either stick with the union or replace it with a const object `<Typestring, number>`?

iuioiua (approved):
This implementation LGTM! Please mark as ready for review once ready 🙂

kt3k (approved):

PR #3841, opened 2023-11-23
refactor(front_matter): replace Format enum with const object

realpha:

While the actual change is trivial, it requires a spur of followup changes in the rest of the module (and beyond).
Given that this is a breaking change, and that this module is widely used (e.g. `fresh`), I wanted to check-in with you guys before commencing.

IMO the change is warranted, and the `enum` has already been marked as deprecated. So in principle I would be in favor of completing this.
On the other hand, it's really a low priority issue and low value change, which will require a not so trivial amount of effort across the ecosystem.

### DIFF
--- a/front_matter/_formats.ts
+++ b/front_matter/_formats.ts
@@ -2,13 +2,14 @@
 
 type Delimiter = string | [begin: string, end: string];
 
-/** @deprecated (will be removed after 1.0.0) Use literal types `"yaml" | "toml" | "json" | "unknown"`. */
-export enum Format {
-  YAML = "yaml",
-  TOML = "toml",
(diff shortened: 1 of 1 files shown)


realpha:
@iuioiua @kt3k

realpha:
Impact could be minimized by changing `Formats` back to `Format` as in the other PRs. Although this isn't used in this module, the absence of the reverse mapping inhernt to enums would still constitute a breaking change.

iuioiua:
Actually, let's not change this code at all. The enum will be deprecated at std v1. It's fine to remain how it is until then. Thanks for looking into this and providing the feedback either way. I'll remove this from the list.

PR #3842, opened 2023-11-23
refactor(fmt): replaced enums with const objects

realpha:

This change is only internal.


kt3k (approved):
LGTM

PR #3843, opened 2023-11-23
refactor(console): replace enums with const objects

realpha:
Refers to this [issue](https://github.com/denoland/deno_std/issues/3782).

This change is internal only.


iuioiua (approved):
LGTM! Thank you.

PR #3844, opened 2023-11-23
refactor(assert): replace DiffTypeEnum with const object

realpha:

This change is internal only. The changed `DiffType` was solely exported to facilitate tests in `_diff_test.ts`.

### DIFF
--- a/assert/_diff.ts
+++ b/assert/_diff.ts
@@ -16,11 +16,13 @@ interface FarthestPoint {
   id: number;
 }
 
-export enum DiffType {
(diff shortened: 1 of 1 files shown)


iuioiua (commented):
Would using a string union type make more sense here?

kt3k:
> Would using a string union type make more sense here?

Let's not do that in scope of #3782. `DiffType` object is used large number of times in the code.

kt3k (approved):
LGTM

PR #3845, opened 2023-11-23
refactor(archive): replace FileTypes enum with const object

realpha:

This change was a little bit tricky because the exposed interfaces do not enforce `FileTypes` in the input they receive. The old `enum` also wasn't exposed via the module API.
Consequently, the implementations also don't really trust their input but it seems that failing to provide a correct filetype doesn't necessarily lead to a panic.

I've tried to make this change as minimally invasive as possible, but IMO it doesn't improve anything. Maybe this refactor should be completed in tandem with the stabilization of the `archive` module API to actually contribute something to the overall maintainability and quality of this module.

### DIFF
--- a/archive/_common.ts
+++ b/archive/_common.ts
@@ -36,16 +36,26 @@ export interface TarMeta extends TarInfo {
   fileSize?: number;
 }
 
-export enum FileTypes {
-  "file" = 0,
-  "link" = 1,
-  "symlink" = 2,
-  "character-device" = 3,
-  "block-device" = 4,
-  "directory" = 5,
-  "fifo" = 6,
-  "contiguous-file" = 7,
-}
+export const FileTypes = {
+  "file": 0,
+  "link": 1,
+  "symlink": 2,
+  "character-device": 3,
+  "block-device": 4,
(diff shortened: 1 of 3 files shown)


realpha:
@iuioiua

iuioiua:
> This change was a little bit tricky because the exposed interfaces do not enforce `FileTypes` in the input they receive.

Are you talking about `TarInfo.type` being of type `string` instead of `keyof typeof FileTypes`? If so, this implementation will be deprecated once #1985 lands, and the new implementation addresses that.

> Consequently, the implementations also don't really trust their input but it seems that failing to provide a correct filetype doesn't necessarily lead to a panic.
> 
> I've tried to make this change as minimally invasive as possible, but IMO it doesn't improve anything. Maybe this refactor should be completed in tandem with the stabilization of the `archive` module API to actually contribute something to the overall maintainability and quality of this module.

Let's skip this enum, as the current implementation will be deprecated once the new implementation lands. I'll close this PR and remove it from the list. Are you able to create a new issue regarding panicking when providing the incorrect file type? Either way, thank you for your help!

PR #3849, opened 2023-11-24
BREAKING(datetime): remove `toIMF()`

iuioiua:
This was deprecated in favour of [`Date.toUTCString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toUTCString).


kt3k (approved):
LGTM

PR #3850, opened 2023-11-24
BREAKIING(fs): remove `EOL` enum, add OS-dependent `EOL`

iuioiua:
This was deprecated in favour of a plain JS object.


kt3k (approved):
This is a bit different type of breaking change from what we usually do, but let's try and see

LGTM

PR #3851, opened 2023-11-24
docs(http): fix `Status` enum deprecation notice

iuioiua:
This moves the removal version of the `Status` enum back to match the deletion of the `http/http_status.ts` file. Otherwise, the `Status` enum will get removed before the `http_status.ts` file is deleted.

### DIFF
--- a/http/status.ts
+++ b/http/status.ts
@@ -32,7 +32,7 @@
  */
 
 /**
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #3852, opened 2023-11-24
BREAKING(async): remove deprecated `deferred()`

iuioiua:
This was deprecated in favour of [`Promise.withResolvers()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers).


kt3k (approved):
LGTM

PR #3854, opened 2023-11-24
BREAKING(bytes): remove spread `concat()` overload

iuioiua:


kt3k (approved):
LGTM

Thanks for a lot of clean up works!

PR #3855, opened 2023-11-24
deprecate(log): remove enums and add deprecation notices

realpha:
References this [issue](https://github.com/denoland/deno_std/issues/3782) and is the successor of this [PR](https://github.com/denoland/deno_std/pull/3838)

- feat(log): replaced LogLevels enum with object and enforced stricter types
- feat(log): updated types
- chore(log): updated types in tests
- chore(log): retain bimap for const object and add deprecation notices
- refactor(log): added deprecation notices and hardened level setter
- feat(log): added LogLevel to module API
- chore(log): removed unnecessary type casts in tests

### DIFF
--- a/log/logger.ts
+++ b/log/logger.ts
@@ -65,7 +65,7 @@ export class Logger {
   /**
    * Use this to retrieve the current numeric log level.
    *
-   * @returns {number} - Deprecated (will return {@linkcode LogLevel} after 0.211.0)
+   * @returns - Deprecated (will return {@linkcode LogLevel} after 0.211.0)
    */
   get level(): number {
     return this.#level;
@@ -74,7 +74,7 @@ export class Logger {
   /**
    * Use this to set the numeric log level.
    *
-   * @param {number} level - Deprecated (will accept {@linkcode LogLevel} after 0.211.0)
+   * @param level - Deprecated (will accept {@linkcode LogLevel} after 0.211.0)
    */
   set level(level: number) {
     try {
--- a/log/logger.ts
+++ b/log/logger.ts
@@ -56,14 +56,17 @@ export class Logger {
     options: LoggerOptions = {},
   ) {
     this.#loggerName = loggerName;
-    this.#level = getLevelByName(levelName);
+    /* TODO: Remove this unnecessary typecast after 0.211.0 */
+    this.#level = getLevelByName(levelName) as LogLevel;
     this.#handlers = options.handlers || [];
   }
 
   /**
    * Use this to retrieve the current numeric log level.
+   *
+   * @returns {number} - Deprecated (will return {@linkcode LogLevel} after 0.211.0)
    */
-  get level(): LogLevel {
+  get level(): number {
     return this.#level;
   }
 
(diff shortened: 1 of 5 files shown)


realpha:
@iuioiua Again, sry for the inconvenience. I am at my best at fridays 😅

realpha:
Also wasn't sure regarding the module API export and the `LogLevel` deprecation notices. Happy to revert/change if you have any better ideas

log/logger.ts
```diff
@@ -56,22 +57,40 @@ export class Logger {
...
+    /* TODO: Remove this unnecessary typecast after 0.211.0 */
+    this.#level = getLevelByName(levelName) as LogLevel;
     this.#handlers = options.handlers || [];
   }
 
-  get level(): LogLevels {
+  /**
+   * Use this to retrieve the current numeric log level.
+   *
+   * @returns {number} - Deprecated (will return {@linkcode LogLevel} after 0.211.0)
```
iuioiua, line 68: There's no need to define types in JSDocs as the function itself already does that. Ditto to other instances in the PR.

iuioiua (changes requested):

log/logger.ts
realpha: Was my thinking as well, but how do you put out deprecation notices on certain parts of the signature otherwise?

Using a `@deprecated` tag marks the whole function as deprectated. To me that seemed controdictionary. But if that's how you want to roll here, I am down to change it.

kt3k (approved):
LGTM

kt3k:
Now `LogLevels` gets strikethrough in the IDE. I think that gives wrong impression about it.

Screenshot 2023-11-27 at 12 21 54

Also, removing deprecation note check because of this doesn't look desirable.

realpha:
> Sorry for the delay. I've changed the `number` types to `LogLevel`, where appropriate. I think it's OK to do this in this case as it's not breaking any behaviour. @kt3k, WDYT?
> 
> I'll sort out the deprecation check failure separately.

It is a breaking change. `LogLevel` is a super type of of `number`, so there are mulitple scenarios where this change might break user code. E.g.:
```typescript
type MyConfig = {
  appName: string,
  logLevel: number,
}

declare function getConfig(): MyConfig;

const config = getMyConfig();

const recordOpts: LogRecordOptions = {
   msg: "Log this",
  args: [];
  level: config.logLevel;
           ^^^^^^^^^^^^
  `-> Type number not assignable type LogLevel`
  loggerName: "MY_DEFAULT";
}
```

iuioiua:
Ok. Fair criticisms. I'll make the adjustments shortly.

realpha:
why not just revert to 62a2ab5, call it a day, and tidy the rest up with `0.211.0`? It's a low value change. Doesn't seem justified to me, to spend more effort on it.

iuioiua:
That's probably what I was going to do 🙂

iuioiua (approved):
I've reverted back to https://github.com/denoland/deno_std/commit/62a2ab5c2db72ceb66eac9c39040fcb75222170e and just removed the type declarations from JSDocs. @kt3k, please check once more and merge if happy.

Thanks again, @realpha. Sorry again for this taking a little longer than expected.

PR #3858, opened 2023-11-26
docs: add "Deprecations" section to contributing guide

iuioiua:
This change removes the deprecation check in favour of a section on making deprecations in the contributing guide. The tool seems needlessly overconstrained, requires adjustment any time the deprecation approach changes and can hinder changes at times ([example](https://github.com/denoland/deno_std/pull/3855)).

Instead, we can just have a section for both contributors and maintainers to follow when making changes to deprecations.


lino-levan:
I agree with Yoshiya. This has been quite useful in the past.

iuioiua:
I beg to differ. I think the tool is overkill for its intended objective. The same objectives can be easily achieved and managed with a bit of documentation and a small addition to the list of things that the maintainer checks for when reviewing a PR. The one argument I'd agree with for keeping the tool is that it checks for when a deprecated symbol requires removal. Specificially, the checking of the format of the deprecation notice seems unneeded. Frankly, I have a similar opinion of most of these automated tools in a similar context - the engineering cost outweighs the value provided.

Nevertheless, it's all good. Disagreements are inevitable, and consensus wins. I've reverted the changes to the tool itself and reoriented the PR to document how to add deprecation notices.

kt3k (approved):
LGTM

.github/CONTRIBUTING.md
```diff
@@ -37,3 +37,25 @@ before contributing.
...
+   /**
+    * @deprecated (will be removed in 0.215.0) Use {@linkcode bar} instead.
+    */
+   export function foo() {}
+   ```
+
+1. Submit a pull request starting with the following format:
+
+   ```
+   deprecation(sub): `foo()`
```
kt3k, line 60: 👍

PR #3865, opened 2023-11-27
chore: group Dependabot PRs

iuioiua:
Currently, Dependabot opens a new PR for each dependency update available. This change makes Dependabot open only one PR for any available updates, reducing maintenance.

Reference https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/customizing-dependency-updates#example-2


kt3k (approved):
LGTM.

PR #3867, opened 2023-11-28
fix(assert): fix diff of long string in objects

### LINKED ISSUE #939: Test output hides diff on long strings in objects
When comparing two objects that contain long strings with `assertEquals`, the strings get truncated for readability, but that ends up hiding the Exected / Actual diff in some cases.

[Here](https://gist.github.com/wperron/f4a7bf79172f5c354d5c75850bb7d511) is a quick gist that demonstrates the issue

kt3k:
closes #939 

Now [the original example](https://gist.github.com/wperron/f4a7bf79172f5c354d5c75850bb7d511) shows the diff like the below:

```
$ deno test a.js 
running 1 test from ./a.js
test ... FAILED (2ms)

 ERRORS 

test => ./a.js:3:6


    [Diff] Actual / Expected


    {
-     foo: "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of ERROR ERROR.",
+     foo: "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum.",
    }
```

### DIFF
--- a/assert/_format.ts
+++ b/assert/_format.ts
@@ -18,6 +18,7 @@ export function format(v: unknown): string {
       iterableLimit: Infinity,
       // getters should be true in assertEquals.
       getters: true,
+      strAbbreviateSize: Infinity,
     })
     : `"${String(v).replace(/(?=["\\])/g, "\\")}"`;
 }
--- a/assert/_format_test.ts
+++ b/assert/_format_test.ts
@@ -78,3 +78,16 @@ Deno.test("assert diff formatting", () => {
 ]`,
   );
 });
+
+Deno.test("format() doesn't truncate long strings in object", () => {
+  const str = format({
+    foo:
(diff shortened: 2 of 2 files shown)


iuioiua (approved):
LGTM

PR #3868, opened 2023-11-28
chore(streams): cleanup use of `readAll()`

iuioiua:
Towards #3718


iuioiua:
Actually, this isn't needed.

PR #3869, opened 2023-11-28
chore(streams): cleanup use of `writeAll()`

iuioiua:
Towards #3718


iuioiua:
Actually, I realise this isn't needed.

Issue #3604, opened 2023-09-01
Feedback on proposal: Deprecate std/dotenv


cknight:
https://github.com/denoland/deno_std/issues/3489#issuecomment-1702521412 proposes deprecating std/dotenv as it overlaps with the new [`--env` CLI flag proposal](https://github.com/denoland/deno/pull/20300).  This issue has been raised to capture discussion on the topic.

cknight:
My own view is that as it currently stands, --env only provides a subset of the functionality which dotenv provides, and therefore dotenv provides sufficient benefit to the user to remain a std module.  Specifically, it allows:

* a `.env` file to be parsed as a configuration object.  This means there is no dependency on allowing env permissions. N.b. the --env proposal is considering the ability to automatically allow env permissions for env vars specified in the .env file without the need to specifically declare permissions for them, however this is still under discussion.  Another use case for the configuration object is the ability to log or output all env vars in use without knowing what the individual env var keys are.  This is not possible with the --env proposal, though you could output the entire process env environment with full env permission grant.
* dotenv offers ([the poorly named](https://github.com/denoland/deno_std/issues/3561)) `.env.example` file, allowing the project/module to dictate which environment variables it expects to be present in the .env file or process environment.  These can also offer documentation on each expected env var.  This is useful for setups where the .env file is dynamically sourced for each environment and offers auto-validation.
* dotenv offers the `.env.defaults` file to supply default env values when these are missing from the `.env` file.  Again, this is useful in environments where you are externally supplying a `.env` file.

Much of this functionality (except the configuration object) could be moved to the `--env` proposal.  However as it stands, I believe we shouldn't deprecate this module.

Leokuma:
> a .env file to be parsed as a configuration object

Wouldn't JSON or YAML be more appropriate for that? Because the purpose of an env file is to set the environment.

> ability to log or output all env vars in use without knowing what the individual env var keys are

Is there a use case for that other than debugging?

> This means there is no dependency on allowing env permissions

Could be achieved with `--env` too. And actually dotenv requires `--allow-read`, so we're basically swapping one permission flag for another.

> useful for setups where the .env file is dynamically sourced for each environment

Fair point. How worse would it be to use a `config.json` or `env.json` instead? 

[...]

PR #3874, opened 2023-11-29
feat: `std/env` with `assertEnvSet()` and `setEnvDefault()`

### LINKED ISSUE #2490: dotenv: different behaviour between `load.ts` and `mod.ts` when used together with `run --watch=`
### Describe the bug

When using Deno's watcher (`deno run --watch=file,folder/ …`) in conjunction with dotenv, I get different behaviour with auto loading compared to using `config()`.  **I'm not sure whether the fault lies with dotenv or `Deno.env`'s handling of changes, though.**

On a watcher-triggered reload, when using auto-loading, eventual changes in `.env` will not be available via `Deno.env`.  When using `config()`, they are available.

Background: I was implementing an `.env`-based Deno Fresh server configuration (setting the HTTP port etc.) when I noticed that the file changes weren't picked up.

### Steps to Reproduce

1. Create initial `.env` file:
```
MY_ENV_VAR=foo
```

2. Create `app-config.ts`:
```typescript
import { config } from "https://deno.land/std@0.150.0/dotenv/mod.ts";
console.log((await config())["MY_ENV_VAR"]);
````

3. Run `app-config.ts` and note the current output, "foo":
```
deno run -A --watch=.env app-config.ts
```

4. Create `app-load.ts`:
```typescript
import "https://deno.land/std@0.150.0/dotenv/load.ts";
console.log(Deno.env.get("MY_ENV_VAR"));
````

5. Run `app-load.ts` and note the current output, "foo":
```
deno run -A --watch=.env app-load.ts
```

6. Change contents of `.env` file:
```
MY_ENV_VAR=bar
```

7. The watcher will reload `app-config.ts`, the output will be "bar".
8. The watcher will reload `app-load.ts`, the output will still be "foo". 

### Expected behavior

I expect both scripts to behave the same, they should both output "foo".

### Environment

- OS: macOS 12.5 (21G72)
- deno version: 1.24.0
- std version: 0.150.0

### LINKED ISSUE #3555: [std/dotenv] - .env.example file is not validated
**Describe the bug**
If I have a .env.example file as follows:
```sh
# .env.example
SOME_KEY
```
dotenv will not enforce the SOME_KEY env variable is present in the environment (the correct syntax is `SOME_KEY=`).  Indeed, this key is ignored. This seems like an easy mistake to make, especially for beginners.  However, I wonder if allowing multi-line entries makes this to difficult to catch?

**Steps to Reproduce**

1. Create a `.env.example` file with the following contents:
```sh
SOME_KEY
```
2. Assuming there is a mod.ts, run:
```sh
deno run -A mod.ts
```
mod.ts runs without issue.

**Expected behavior**

dotenv validates the content of .env.example and throws an exception on invalid format

**Environment**
- deno version: 1.36.1
- std version: 0.198.0

Perhaps .env.example could be more strictly parsed to catch this?

### LINKED ISSUE #3561: [std/dotenv] - change .env.example to .env.required

iuioiua:
This PR adds a new `std/env` sub-module with `assertEnvSet()` and `setEnvDefault()` that, if accepted, replaces `std/dotenv`'s `load()`, when used with the new `--env` CLI argument.

`load()` does quite a bit. In short, it reads and parses a `.env` file, ensures that it contains required environment variables as defined by a `.env.example`, and sets any default environment variable values as defined by a `.env.defaults`. Once done, it sets these environment variables in the current process. There's more to it than that, but that's the gist. Documentation [here](https://deno.land/std@0.208.0/dotenv/mod.ts?s=LoadOptions).

`assertEnvSet()` and `setEnvDefault()` perform a subset of this functionality, but in a fundamentally different and simpler way - by directly interacting with [`Deno.env`](https://docs.deno.com/runtime/manual/basics/env_variables) and defining required and default environment variable information in-code instead of in `.env.*` files.

It allows the Deno runtime to take care of actually loading the `.env` file, using the `--env` CLI argument, and instead just checks the current processes environment variables set. In other words, these functions skip reading and parsing of `.env`.

Required environment variables and default environment variables are defined in-code instead of obscure `.env.*`. Previously, `.env.*` files were saved to repos anyway. This is safe to do, as these specific values (required and defined env. vars.) are not confidential.

One of the nice things about `load()`'s approach is that by having environment variables in `.env.*` files, one can copy and paste the keys to the actual `.env` file in the correct format. When `assertEnvSet()` throws because of a missing required environment variable, it also prints the error in the `.env` format, reducing any possible guesswork regarding the environment variable the error refers to.

I understand this differs from the `dotenv` convention, but I think it's a far simpler and cleaner way of ensuring environment variables are set up correctly.

If accepted in favour of `load()`:
* Closes #3561
* Closes #3555
* Closes #2490
* Closes #3516


timreichen:
I think the name `prepare` does not describe well what the function is actually doing.
If we include it in std, I think it should be more straight forward, maybe even split the functionality to something like:
```ts
import { setDefault, assert } from "https://deno.land/std@$STD_VERSION/dotenv/mod.ts";

// sets a default value in Deno.env if the key does not already exist.
setDefault("foo", "bar")

// throws an error if value does not exist in Deno.env
assert("foo")
```

iuioiua:
Yeah, fair points. I wasn't certain about the name either. I very much like the idea of splitting this up into 2 separate functions too. How about the names `setEnvDefault()` and `assertEnvSet()`?

timreichen:
> Yeah, fair points. I wasn't certain about the name either. I very much like the idea of splitting this up into 2 separate functions too. How about the names `setEnvDefault()` and `assertEnvSet()`?

Is the `env` in the names necessary in a mod called dotenv? Includeing `Env` could be confusing because one might think it handles/modifies/checks the Env object instead of a single key/value pair.

Maybe instead of `assert()`, `assertExists()` would be more consistent with std/assert/assert_exists.
I couldn't find a common name for a `setIfAbsent` function. Seems like every language has a different name for that. So maybe there is a better name out there instead of `setDefault()`...

But assuming the import is done like this:
```ts
import * as env from "https://deno.land/std@$STD_VERSION/dotenv/mod.ts";

// sets a default value in Deno.env if the key does not already exist.
env.setDefault("foo", "bar")

// throws an error if value does not exist in Deno.env
env.assertExists("foo")
```
seems pretty clear to me.

iuioiua:
> Is the `env` in the names necessary in a mod called dotenv? Includeing `Env` could be confusing because one might think it handles a whole Env object instead of a single key/value pair.

I'm thinking the same now too. These functions would probably best done in a new `std/env`.



> Maybe instead of `assert()`, `assertExists()` would be more consistent with std/assert/assert_exists. I couldn't find a common name for a `setIfAbsent` function. Seems like every language has a different name for that. So maybe there is a better name out there instead of `setDefault()`...
> 
> But assuming the import is done like this:
> 
> ```ts
> import * as env from "https://deno.land/std@$STD_VERSION/dotenv/mod.ts";
> 
> // sets a default value in Deno.env if the key does not already exist.
> env.setDefault("foo", "bar")
> 
> // throws an error if value does not exist in Deno.env
> env.assertExists("foo")
> ```
> 
> seems pretty clear to me.

But that's not always how the functions will be imported. They may get imported from an intermediary `deps.ts` file.

I've gone ahead with `setEnvDefault()` and `assertEnvSet()`.

timreichen:
ref [emplace](https://github.com/tc39/proposal-upsert) at stage 2.

kt3k:
I'm -1 to this.

> a new std/env sub-module with assertEnvSet() and setEnvDefault() that, if accepted, replaces std/dotenv's load().

I don't think these can replace `load` of `dotenv`. The main point of `load` is, as its name suggests, loading of the env variables, not validating or supplying defaults.

Also `assertEnvSet` and `setEnvDefault` both look too trivial to be included in std.

```js
if (!Deno.env.has("FOO")) Deno.env.set("FOO", "BAR");
```
is clearer in my view than:
```js
import { setEnvDefault } from "std/env/mod.ts";
setEnvDefault("FOO", "BAR")
```

kt3k:
> Seeing that people do in fact use load() for its ability to load .env files and define default and required environment variables, there is demand for this functionality.

From my observation, the features of checking required and supplying defaults are not essential part of `dotenv`. Rather people often seem annoyed or confused by that feature enabled by default (For example, [this](https://github.com/denoland/deno_std/pull/3516) suggests making them optional)

I don't think providing these 2 APIs help people migrating from `std/dotenv` to `--env`.

> I hope for --env, assertEnvSet() and setEnvDefault() to replace load()

To my understanding of the discussion in #3604, some people are in favor of keeping to use `std/dotenv` because of 'automatic' discovery capability of `.env` file (It's specified by source code, not command line flag). Because `--env` doesn't have it, I don't think migration is not an option to such users.

iuioiua:
PTAL @timreichen, @jsejcksn, @cknight, @andrewthauer and @Leokuma. I'd like to hear what you guys think about this.

timreichen:
I am unbiased about this. We also could just add a warning message in `load()` and `loadSync()` encouraging the usage of `--env`.

andrewthauer:
At a minimum, I would like to see both `parse` and `stringify` as part of std. I personally don't have a lot of need for the more opinionated `load` method tbh.

As for `assertEnvSet`, I would normally use `direnv` with `env_vars_required` for local development. Both `assertEnvSet` and `setEnvDefault` also cross into some other territory such as validation of the value. You might also need to parse the value into a different datatype, etc. which quickly becomes domain specific. So tbh, I'm not entirely sure if I would reach for either of these very often.

jsejcksn:
My understanding is that there is a continuum of use cases when working with plaintext files which consist of key-values pairs.

At one end: a user wants as little configuration and as much "magic" as possible — they expect the offered API to figure out where their ".env" file is (maybe some cascade of expected locations) and load the pairs into the runtime environment variable dictionary/map/etc.

At the opposite end: a user will somehow obtain a string of such data at some point during the program's lifecycle, and they need a clearly defined specification for the serialization/deserialization of the data, along with composable functional primitives to help with manually completing tasks which deal with key-value pairs and this format.

It appears that Deno's `--env` CLI argument wants to take on one end of the continuum (closer to the configuration-less magic), but there are many other users at each point along the rest of the continuum, and I hope that `dotenv` can provide usefulness to them.

For us: a clear challenge is to define exactly which combinations of primitives provide utility to users on the continuum, but I expect that the standard library will continue to offer primitives — and perhaps in the future: a de-/serialization specification so that (especially new) users can have confidence in the format of their env files without needing to fully review the source code of the module.

The spec that I'm describing would exhaustively answer questions that many users have about the format, such as:

- Are single-quoted values allowed? How about double-quoted?
- What about whitespace before/after a key/value?
- What if a value includes `=`?
- Are multiline values supported? What's the syntax for that?
- What types of newlines are supported? Just `LF` or `CRLF`?
- …etc.

iuioiua:
I'm going to close this PR for now. It seems as though we need a bit more discussion.

> For us: a clear challenge is to define exactly which combinations of primitives provide utility to users on the continuum, but I expect that the standard library will continue to offer primitives — and perhaps in the future: a de-/serialization specification so that (especially new) users can have confidence in the format of their env files without needing to fully review the source code of the module.
> 
> The spec that I'm describing would exhaustively answer questions that many users have about the format, such as:
> 
> * Are single-quoted values allowed? How about double-quoted?
> * What about whitespace before/after a key/value?
> * What if a value includes `=`?
> * Are multiline values supported? What's the syntax for that?
> * What types of newlines are supported? Just `LF` or `CRLF`?
> * …etc.

For future reference, let's try to stay on track with the issue at hand. While this is an interesting idea, it's best discussed in a new issue 🙂

PR #3875, opened 2023-11-29
fix(fs): don't use `path.resolve()` in `fs.copy()`

### LINKED ISSUE #643: bug: `fs.copy()` throws permission error because it requires CWD access
Here's a repository containing an example to replicate the bug:
https://github.com/ChiKler/-Deno-error-Uncaught-PermissionDenied-read-access-to-CWD-

I believe that's because you use `copySync()` from `std`, which (without a good reason) calls `path.resolve()` which requires CWD access. It shouldn't call `path.resolve()`.

_Originally posted by @nayeemrmn in https://github.com/denoland/deno/pull/2875#issuecomment-723579312_

iuioiua:
As it turns out, all tests pass if you remove the use of `path.resolve()` in `fs.copy()` 🤷🏾‍♂️

I'll open a separate issue exploring removing the use of `Deno.lstatSync()` in `fs.copy()` too.

Closes #643

### DIFF
--- a/fs/copy.ts
+++ b/fs/copy.ts
@@ -2,7 +2,6 @@
 
 import { basename } from "../path/basename.ts";
 import { join } from "../path/join.ts";
-import { resolve } from "../path/resolve.ts";
 import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, isSubdir, toPathString } from "./_util.ts";
 import { assert } from "../assert/assert.ts";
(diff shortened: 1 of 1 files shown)


kt3k:
I think this `resolve` is used for checking the equality of `src` and `dest` more robustly. (For example, if `src` is relative path and `dest` is absolute path, both representing the same path, then the below check (`src === dest`) can be wrong (I think we need a test case to check this). Is there any alternative way to check path equality without using `resolve`?

iuioiua:
I think we can account for path equality if `Deno.copyFile()` threw an error in that case, which it currently does not. Let's pursue that.

kt3k:
> we can account for path equality if Deno.copyFile() threw an error in that case, which it currently does not. Let's pursue that.

Sounds good!

PR #3876, opened 2023-11-29
refactor(fs): reduce use of `toPathString()`

iuioiua:


iuioiua:
Is it a problem if it's run on the same machine? Perhaps I'm missing something obvious. I don't see the problem.

kt3k:
For example, in `src.toString().split(sep)`, sep is `\` and `src.toString()` becomes something like `file:///C:/path/to/file` (slash-splitted) on windows. So the split call doesn't split anything in that case.

iuioiua:
Ok. I think it's best we address any testing holes before pushing for this then, to be sure.

PR #3879, opened 2023-11-29
chore(http): remove test files with symbols in the filename

iuioiua:


kt3k:
What is the motivation?

kt3k (approved):
Note: These files are blocker for 'workspace' introduction in deno_std, and we need to remove these files for now.

LGTM

PR #3882, opened 2023-11-30
chore(http): rename test file

iuioiua:
Note: none of the tests had anything to do with space-character handling, so replacing the space character in the filename with an underscore was safe.

This is a requirement for workspaces support.


lucacasonato (approved):
LGTM, but adding a test that checks that files with spaces are served correctly would be good.

PR #3884, opened 2023-11-30
chore(assert): move `_format.ts`

### LINKED ISSUE #3880: @std/expect depends on private function from @std/assert
`//expect/_matchers.ts` depends on `//assert/_format.ts`, which is not a public module in `//assert`. This code should be rewritten to not depend on private `//assert` APIs.

iuioiua:
I'm unsure whether this is compatible with workspaces. If it isn't, I'll just copy `_format.ts` to `std/expect` instead.

Closes #3880


lucacasonato:
Not compatible

PR #3885, opened 2023-11-30
deprecation: move `types.d.ts` to `std/io`

iuioiua:
This change is required for workspaces support.

### DIFF
--- a/_tools/check_circular_submodule_dependencies.ts
+++ b/_tools/check_circular_submodule_dependencies.ts
@@ -38,7 +38,7 @@ async function check(
     }
   }
   deps.delete(submod);
-  deps.delete("types.d.ts");
+  deps.delete("io/types.d.ts");
   return { name: submod, set: deps, state };
 }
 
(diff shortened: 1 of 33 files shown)


_tools/check_circular_submodule_dependencies.ts
```diff
@@ -38,7 +38,7 @@ async function check(
     }
   }
   deps.delete(submod);
-  deps.delete("types.d.ts");
+  deps.delete("io/types.d.ts");
```
lucacasonato, line 41: I think you should remove this line instead.

lucacasonato (commented):
Looks good

_tools/check_circular_submodule_dependencies.ts
kt3k: That's correct

iuioiua: Fixed

kt3k (approved):
LGTM

PR #3886, opened 2023-11-30
chore(assert): copy `_format.ts` to `std/expect`

### LINKED ISSUE #3880: @std/expect depends on private function from @std/assert

iuioiua:
Closes #3880


kt3k (approved):
LGTM

kt3k:
Maybe let's create an issue for merging these again in the future?

PR #3890, opened 2023-12-01
feat: net port utils

### LINKED ISSUE #3279: suggestion: `getAvailablePort()`
Are any of the following utilities good candidates for inclusion?

**`withSignal`**

Run an async function by passing in a fresh abort controller's signal. Upon completion, abort.

Use case: ensuring that the resolution of a promise triggers the abort, which may trigger other *things*.

```ts
async function withSignal<T>(cb: (signal: AbortSignal) => Promise<T>) {
  const controller = new AbortController()
  try {
    return await cb(controller.signal)
  } finally {
    controller.abort()
  }
}
```

**`port.getAvailable`**

Determine an open port.

Use case: get an unused port when launching processes that allow for specificity of port.

```ts
function getAvailable(): number {
  const tmp = Deno.listen({ port: 0 })
  const { port } = tmp.addr as Deno.NetAddr
  tmp.close()
  return port
}
```

**`port.ready`**

Poll whether a given port can be connected to

Use case: determine whether a server is prepared to accept connections.

```ts
async function ready(port: number, ms = 500): Promise<void> {
  while (true) {
    try {
      const connection = await Deno.connect({ port })
      connection.close()
      break
    } catch (e) {
      if (e instanceof Deno.errors.ConnectionRefused) await delay(ms)
      else throw e
    }
  }
}
```

harrysolovay:
Closes #3279

### DIFF
--- a/net/port.ts
+++ b/net/port.ts
@@ -1,0 +1,26 @@
+/** @returns an open port */
+export function getAvailable(): number {
+  const tmp = Deno.listen({ port: 0 })
+  const { port } = tmp.addr as Deno.NetAddr
+  tmp.close()
+  return port
+}
+
+/**
+ * @param port the port at which to check readiness
+ * @param ms the duration of the intervals between re-checking readiness
+ */
+export async function ready(port: number, ms = 500): Promise<void> {
+  while (true) {
+    try {
+      const connection = await Deno.connect({ port })
+      connection.close()
+      break
+    } catch (e) {
+      if (e instanceof Deno.errors.ConnectionRefused) {
+        await new Promise((resolve) => setTimeout(resolve, ms))
+      }
+      else throw e
+    }
+  }
+}


kt3k:
Please create a file for each API with corresponding file name, and re-export them from `mod.ts` (We follow that pattern in newer modules).

net/port.ts
```diff
@@ -0,0 +1,29 @@
+/** @returns an open port */
+export function getAvailable(): number {
```
kt3k, line 2: How about `getAvailablePort`? This belongs to `net` module. I think it's better to clarify what is available in API name.

net/port.ts
```diff
@@ -0,0 +1,29 @@
...
+}
+
+/**
+ * Get a promise that resolves upon successful connection to the specified port.
+ * 
+ * @param port the port at which to check readiness
+ * @param ms the duration of the intervals between re-checking readiness
+ * @throws if the inner `Deno.connect` call throws a non `Deno.errors.ConnectionRefused` error.
+ */
+export async function ready(port: number, ms = 500): Promise<void> {
```
kt3k, line 16: Let's make the second argument option object. We probably would like to add more options in the future (e.g. `signal: AbortSignal`, `timeout: number`. etc)

ref: https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object

harrysolovay: I got rid of this function altogether. I think [iuioiua's comment about using `retry`](https://github.com/denoland/deno_std/issues/3279#issuecomment-1836932882) is correct (even moreso in light of mention of abort behavior / ie. would the function throw?).

net/get_available_port.ts
```diff
@@ -0,0 +1,18 @@
...
+ * import { getAvailablePort } from "https://deno.land/std@$STD_VERSION/net/get_available_port.ts";
+ *
+ * const port = getAvailablePort();
+ * Deno.serve({ port }, () => new Response("Hello, world!"));
+ * ```
+ */
+export function getAvailablePort(): number {
+  const listener = Deno.listen({ port: 0 });
+  listener.close();
+  return (listener.addr as Deno.NetAddr).port;
```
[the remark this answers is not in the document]
kt3k: I agree, but we tried that before and failed to make it non-breaking, and reverted in the past.

- https://github.com/denoland/deno/pull/10012
- https://github.com/denoland/deno/pull/10255

Another attempt is probably welcome

iuioiua (approved):

PR #3891, opened 2023-12-04
chore(async): `exponentialBackoffWithJitter()` tweaks

iuioiua:
The function comes from an internal file, so the underscore prefix in the function name isn't needed. I've also moved the test to `_util_test.ts` to be more direct.


kt3k (approved):
LGTM. Nice!

PR #3892, opened 2023-12-04
chore(http): add test for file containing space in name

iuioiua:
See https://github.com/denoland/deno_std/pull/3882#pullrequestreview-1758729944

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -154,6 +154,25 @@ Deno.test("serveDir() with hash symbol in filename", async () => {
   await Deno.remove(filePath);
 });
 
+Deno.test("serveDir() with space in filename", async () => {
+  const filePath = join(testdataDir, "test file.txt");
+  const text = "Plain text";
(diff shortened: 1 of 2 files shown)


http/file_server_test.ts
```diff
@@ -154,6 +154,25 @@ Deno.test("serveDir() with hash symbol in filename", async () => {
...
+  const req = new Request("http://localhost/test%20file.txt");
+  const res = await serveDir(req, serveDirOptions);
+  const downloadedFile = await res.text();
+
+  assertEquals(res.status, 200);
+  assertEquals(
+    res.headers.get("content-type"),
+    "text/plain; charset=UTF-8",
+  );
+  assertEquals(downloadedFile, text);
```
kt3k, line 171: Can you surround these part with `try {}` block and call `Deno.remove(filePath)` in `finally {}` block? That structure avoids the leak of created file even when the test failed in the middle of the body.

kt3k (approved):
LGTM

Issue #3878, opened 2023-11-29
Run `deno doc --lint` on all exported files


lucacasonato:
To ensure all files have sufficient documentation, we should run `deno doc --lint`. This will expose errors such as:

- missing JSDoc comments on exported symbols
  - ```
    Missing JSDoc comment.
        at file:///Users/lucacasonato/projects/github.com/denoland/deno_std/data_structures/binary_search_tree.ts:177:3
    ```
- symbols that are part of the public API, but not exposed
  - ```
    Type 'BinarySearchTree.prototype.rotateNode' references type 'Direction' which is not exported from a root module.`
    ```
- missing explicit typings on arguments / return types
  - ```
    Missing return type.
        at file:///Users/lucacasonato/projects/github.com/denoland/deno_std/data_structures/comparators.ts:5:1
    ```

iuioiua:
Yep, for sure. It might also save a bit of time targeting any `mod.ts` files.

I think these points are best part of #3764 instead of a separate issue, as they refer to the same task.

iuioiua:
I'm closing this in favor of #3878, which requires running `deno doc—-lint` on exported files.

PR #3897, opened 2023-12-05
chore(fs): complete documentation

iuioiua:
Note: some `void` and `Promise<void>` return types have been declared. This is a bug in the `deno_doc` version in the stable version of the runtime. It was fixed in https://github.com/denoland/deno_doc/pull/405 which will be included in the next release of the runtime.

Towards #3764 and #3878.


kt3k (approved):
LGTM. Nice!

Issue #3713, opened 2023-10-23
tracking: improve test coverage


iuioiua:
The test coverage of each of the following sub-modules is below 100% and can be improved:
- [ ] assert
- [ ] async
- [ ] bytes
- [ ] console
- [ ] crypto
- [ ] csv
- [ ] datetime
- [ ] encoding
- [ ] flags
- [ ] fmt
- [ ] front_matter
- [ ] fs
- [ ] http
- [ ] json
- [ ] jsonc
- [ ] media_types
- [ ] msgpack
- [ ] path
- [ ] semver
- [ ] streams
- [ ] testing
- [ ] toml
- [ ] uuid
- [ ] yaml

Note: Deprecated APIs do not require additional tests. Removing unneeded functionality may also be reasonable instead of increasing testing for some APIs.

Paul-Stern:
Hello!
My name is Paul and I am just starting my GitHub activity. Given that this issue is a "good first" one, I would like to take part in fixing it. Can I be asssigned for it?

iuioiua:
Hi Paul! This task is best tackled in parts (one PR per sub-module). Which sub-modules would you like to tackle?

Paul-Stern:
Ok! I started digging in the csv sub-module so let it be csv.

miso-devel:
I would like to work on some test coverage improvements. 
I am working on `console` now.

scarf005:
hi, I'm working on `assert`, before #3771.

babiabeo:
> * [ ]  [console](https://app.codecov.io/gh/denoland/deno_std/tree/main/console)
> * [ ]  [toml](https://app.codecov.io/gh/denoland/deno_std/tree/main/toml)

`console` and `toml` have been fully covered in #4177 and #4211 respectively.

mbhrznr:
`media_types` has been fully covered in #4554.

timreichen:
I think this tracker is not up-to-date anymore.

iuioiua:
> I think this tracker is not up-to-date anymore.

Thanks. I've updated the initial comment so that it doesn't require as much attention.

PR #3898, opened 2023-12-05
fix(fs): improve `exists()` test coverage

vasucp1207:
Issue https://github.com/denoland/deno_std/issues/3713

New test cases cover,
- When both isDirectory and isFile set to true.
- - When user is the owner and can read.
- When user group is the owner and can read.


iuioiua (approved):
LGTM! Thank you for your contribution.

PR #3903, opened 2023-12-06
deprecation(io): IO types in favour of Web Streams API

iuioiua:
To be clear, `/types.d.ts` is being deprecated in favour of `/io/types.d.ts` with a removal version of 0.211.0. `/io/types.d.ts` is being deprecated in favour of the Web Streams API. It's being done this way to minimise breaking changes and allow code to still use `/io/types.d.ts` until its removal in 1.0.0.

This change assumes that `Deno.*` IO interfaces and methods will be removed in Deno v2 and Std. Lib. v1, at the same time.

See https://github.com/denoland/deno/pull/21465

### DIFF
--- a/io/types.d.ts
+++ b/io/types.d.ts
@@ -1,9 +1,19 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #3905, opened 2023-12-07
feat(http): `signCookie()`, `verifyCookie()` and `parseSignedCookie()`

iuioiua:
This implementation aims to explore the idea of a simpler functional alternative to the current `CookieMap()` implementation.

The implementation is simpler and complements the current cookie functions. However, I'm not fully certain this is the best approach yet. My main concern is that it may involve too much setup from the dev, which is a plus for `CookieMap()`, which just uses `.get()` and `.set()`.


kt3k:
@kitsonk PTAL

iuioiua:
As per discussions with Yoshiya, we'll be looking to introduce this functionality as unstable to begin with.

kt3k (approved):
LGTM

PR #3911, opened 2023-12-08
refactor(assert): correct test names

scarf005:
- refactor: extract test data
- refactor: use `assertThrows`
- refactor: use `Deno.test`


kt3k:
We generally prefer nicely sized (like < 300 lines) multiple PRs if it's not too difficult to split into pieces

Issue #3915, opened 2023-12-08
Trim single function import dependencies


jtoppine:
**Describe the bug**

We might wan't to take a look at the single function import dependency graphs to make sure no unnecessary dependencies are loaded, via util modules or such. 

For example `ensure_dir.ts` brings in 80 unique dependencies (102Kb). 

That's just to tell (in case of error) whether the target file is a file, directory or symlink. (`ensure_dir.ts` imports the very simple `getFileInfoType` function from `_util.ts` , and `_util.ts` imports a whole bunch of unneeded stuff ~80 dependencies total). This functionality could be achieved without any dependencies methinks.

ensureDir is just what I happened to notice, other functions around the std lib may have similar potential for "weight loss".

**Steps to Reproduce**

For example:
```
deno info https://deno.land/std@0.208.0/fs/ensure_dir.ts
```
You'll see a bunch of unnecessary stuff is imported.

I didn't go through all the files which might have such unnecessarily deep dependency graphs.

**Expected behavior**

Unneeded dependencies are not imported when using single function imports.

iuioiua:
This can be improved by:
1. Importing directly from implementation files and not `mod.ts` files.
2. Breaking up internal utilities into separate files - one function per file.

For `std/fs`, I took care of the first one in #3916. The second one is still available as a contribution opportunity. PRs are welcome!

PR #3916, opened 2023-12-08
refactor(fs): trim internal dependencies

iuioiua:
Towards fixing #3915. We could also split internal utilities into one file per function. But if we do that here, we'd have to do that in other sub-modules too. WDYT, @kt3k?

### DIFF
--- a/fs/_util.ts
+++ b/fs/_util.ts
@@ -1,6 +1,9 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
-import * as path from "../path/mod.ts";
-import { basename, normalize } from "../path/mod.ts";
+import { resolve } from "../path/resolve.ts";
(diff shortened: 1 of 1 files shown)


lino-levan:
Importing from `mod` inside a util file was definitely being silly I think this is fine. How big is the resulting graph now?

iuioiua:
From 80 modules down to 24 (I think). It'd go down to ~1 if we applied my above suggestion.

kt3k (approved):
Nice improvements! LGTM

kt3k:
> We could also split internal utilities into one file per function.

Sounds good to me. Let's do that when it actually reduces number of downloads count for any public API.

nhrones:
A Good read for this issue:  https://marvinh.dev/blog/speeding-up-javascript-ecosystem-part-7/

PR #3917, opened 2023-12-08
fix(assert): export parameter type alias for `assertArrayIncludes()`

jsejcksn:
Resolves https://github.com/denoland/deno_std/pull/3910#discussion_r1420087019

/cc @kt3k


iuioiua (approved):
LGTM. Thank you.

PR #3921, opened 2023-12-11
docs(crypto): complete documentation

iuioiua:
Towards #3764 and #3878.


crypto/unstable_keystack.ts
```diff
@@ -145,7 +146,9 @@ export class KeyStack {
     return -1;
   }
 
-  [Symbol.for("Deno.customInspect")](inspect: (value: unknown) => string) {
+  [Symbol.for("Deno.customInspect")](
```
iuioiua, line 149: How would we describe this?

kt3k: Maybe something like `/** Custom output for Deno.inspect */`

kt3k (approved):
LGTM

PR #3923, opened 2023-12-11
BREAKING(collections): remove `groupBy()`

iuioiua:
To be merged for v0.211.0.


kt3k (approved):
LGTM

av:
This change made me greatly confused while exploring `@std/collections`.

`groupBy` is the main reason why I'm reaching out to lodash-like libraries, so not being able to find it at first, and then having to discover it was removed (without a trace of explanation) was a bit frustrating.

iuioiua:
Apologies, @av. Yes, we should've included relevant info. `groupBy()` was replaced by the JS-native [`Object.groupBy()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy).

av:
Thanks for getting back to me. I figured it out, just wanted to leave some feedback for the team. Thank you!

PR #3925, opened 2023-12-11
refactor(encoding): cleanup deprecated API use

iuioiua:
None of these are breaking changes.


kt3k:
> None of these are breaking changes.

I think this is the breaking change by definition because we actually 'remove' the APIs.. Let's tag this as `BREAKING`

iuioiua:
Ah, whoops. I misread the function names. Will fix.

kt3k (approved):
LGTM 👍

PR #3929, opened 2023-12-11
chore(ci): temporarily disable Deno stable in CI

iuioiua:
See https://github.com/denoland/deno_std/pull/3119#issuecomment-1850139738


kt3k (approved):
LGTM

PR #3930, opened 2023-12-11
BREAKING(front_matter): remove deprecated APIs

iuioiua:
To be merged for v0.211.0.

PR #3931, opened 2023-12-11
deprecation(front_matter): move `Format` enum deprecation forward

iuioiua:
Removing in v1 seems like overkill. The deprecation shouldn't be too disruptive, and removing it earlier allows the `front_matter` to be cleaned up sooner alongside the cleanups in #3930.


kt3k (approved):
Ok. Makes sense. LGTM

PR #3932, opened 2023-12-12
chore(front_matter): cleanup deprecation-related code

iuioiua:
These changes should not be breaking.

### DIFF
--- a/front_matter/any.ts
+++ b/front_matter/any.ts
@@ -5,8 +5,10 @@ import { parse as parseYAML } from "../yaml/parse.ts";
 import { parse as parseTOML } from "../toml/parse.ts";
 
 export { Format } from "./_formats.ts";
-/** @deprecated (will be removed after 0.210.0) Import from {@link https://deno.land/std/front_matter/test.ts} instead. */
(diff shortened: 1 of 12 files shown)


front_matter/any.ts
```diff
@@ -5,8 +5,10 @@ import { parse as parseYAML } from "../yaml/parse.ts";
 import { parse as parseTOML } from "../toml/parse.ts";
 
 export { Format } from "./_formats.ts";
-/** @deprecated (will be removed after 0.210.0) Import from {@link https://deno.land/std/front_matter/test.ts} instead. */
-export { test } from "./test.ts";
+export {
+  /** @deprecated (will be removed after 0.210.0) Import from {@link https://deno.land/std/front_matter/test.ts} instead. */
+  test,
+} from "./test.ts";
```
iuioiua, line 11: Previously, the deprecation notice wasn't showing up in the code editor. This change fixes that.

kt3k: Good catch!

kt3k (approved):
Thanks for taking care of this. LGTM

Issue #3894, opened 2023-12-04
suggestion(streams): deprecate `toArrayBuffer()`, `toBlob()`, `toJson()` and `toText()`


iuioiua:
These functions can essentially be replaced as follows:
1. `toArrayBuffer(stream)` to `new Response(stream).arrayBuffer()`
2. `toBlob(stream)` to `new Response(stream).toBlob()`
3. `toJson(stream)` to `new Response(stream).json()`
4. `toText(stream)` to `new Response(stream).text()`

The main argument against using `Response` methods was it seems hacky. This argument doesn't justify duplicating functionality that comes for free, already built into the runtime, IMO. If our aim is to complement the Web API, we should remove these APIs.

lino-levan:
One example right away that I found is:
```ts
const stringStream = ReadableStream.from(["hello", " deno ", "world"]);
await new Response(stringStream).text(); // throws!
```
which I feel is a reasonably common case (and is supported by the lib we ported these functions from).

iuioiua:
That's a misuse of `Response` rather than a shortcoming. The correct use would just be:
```ts
const stringStream = ReadableStream.from(["hello", " deno ", "world"])
  .pipeThrough(new TextEncoderStream());
await new Response(stringStream).text();
```

lino-levan:
`Response` just isn't meant to do this type of work. Here's another example that is impossible afaik to write using `new Response(thing).method();`:

```ts
const mixedStream = ReadableStream.from(["hello", Uint8Array.from([1, 2, 3, 4, 5]), "world", [1, 2, 3, 4, 5]]);
await new Response(stringStream).blob(); // throws!
```

nhrones:
> `Response` just isn't meant to do this type of work. Here's another example that is impossible afaik to write using `new Response(thing).method();`:
> 
> ```ts
> const mixedStream = ReadableStream.from(["hello", Uint8Array.from([1, 2, 3, 4, 5]), "world", [1, 2, 3, 4, 5]]);
> await new Response(stringStream).blob(); // throws!
> ```

Did you mean  ` await new Response(mixedStream).blob(); // throws!`
Yes that also throws!

lino-levan:
Ah sorry, you're right, let me edit that.

iuioiua:
A stream of mixed types seems like an edge case, and I have yet to see one in testing code, let alone implementation code anywhere. Are you able to find any examples yourself? Catering to an edge case is not a good reason to have functionality within the Standard Library.

[...]

PR #3934, opened 2023-12-12
perf(streams): make `toBlob()` 20-30% faster

iuioiua:
It's faster just to use `Response.blob()`. This also makes the argument more stricter by only allowing `Uint8Array`s, which the previous implementation implied. Inspired by talks in #3894.

Benchmark source code:
```ts
import { toBlob } from "https://deno.land/std@0.208.0/streams/mod.ts";

Deno.bench("toBlob()", { group: "blob" }, async () => {
  const stream = ReadableStream.from([
    new Uint8Array([1, 2, 3, 4, 5]),
    new Uint8Array([6, 7]),
    new Uint8Array([8, 9]),
  ]);

  await toBlob(stream);
});

Deno.bench("Response.blob()", { group: "blob", baseline: true }, async () => {
  const stream = ReadableStream.from([
    new Uint8Array([1, 2, 3, 4, 5]),
    new Uint8Array([6, 7]),
    new Uint8Array([8, 9]),
  ]);

  await new Response(stream).blob();
});
```

Benchmark results:
```
cpu: Apple M2
runtime: deno 1.38.5 (aarch64-apple-darwin)

benchmark            time (avg)        iter/s             (min … max)       p75       p99      p995
--------------------------------------------------------------------- -----------------------------


toBlob()              5.89 µs/iter     169,894.7      (4.25 µs … 4.6 ms)   5.33 µs  10.08 µs  12.75 µs
Response.blob()       4.68 µs/iter     213,504.9     (4.37 µs … 6.61 µs)   4.55 µs   6.61 µs   6.61 µs

summary
  Response.blob()
   1.26x faster than toBlob()
```

### DIFF
--- a/streams/to_blob.ts
+++ b/streams/to_blob.ts
@@ -2,8 +2,8 @@
 // This module is browser compatible.
 
 /**
- * Converts a {@linkcode ReadableStream} of strings or {@linkcode Uint8Array}s
- * to a {@linkcode Blob}. Works the same as {@linkcode Response.blob}.
+ * Converts a {@linkcode ReadableStream} of {@linkcode Uint8Array}s to a
+ * {@linkcode Blob}. Works the same as {@linkcode Response.blob}.
  *
  * @example
  * ```ts
@@ -14,20 +14,7 @@
  * ```
  */
 export async function toBlob(
-  readableStream: ReadableStream,
+  stream: ReadableStream<Uint8Array>,
 ): Promise<Blob> {
-  const reader = readableStream.getReader();
-  const chunks: Uint8Array[] = [];
-
-  while (true) {
-    const { done, value } = await reader.read();
-
-    if (done) {
-      break;
-    }
-
-    chunks.push(value);
-  }
-
-  return new Blob(chunks);
+  return await new Response(stream).blob();
 }


kt3k:
> This also makes the argument more stricter by only allowing Uint8Arrays, which the previous implementation implied.

Does the previous implementation imply narrower typing? It looks supporting string inputs

iuioiua:
The function does support string streams. However, it's unclear whether this is deliberate. Reasons being:
1. The argument accepts `ReadableStream` without a defined type variable.
2. The JSDoc says, `Works the same as {@linkcode Response.blob}`, which strictly only accepts `ReadableStream<Uint8Array>`.
3. The JSDocs also says it supports string streams.
4. There are no tests for string stream arguments.

We need to make a decision on whether it should support `ReadableStream<string | Uint8Array>` or `ReadableStream<Uint8Array>`.

I think supporting only `ReadableStream<Uint8Array>` and having the simple implementation of using `Response.blob()` under the hood is the way to go. Doing so will make the design and expected behaviour of the function clear and eliminate future confusion. Tweaking code that uses this function is easily done too.

Again, I think that these `toJson()`, `toText()`, etc. functions are redundant. But having them perfectly mirror `Response` methods gives them a clear definition and simple implementation.

kt3k:
Now I think this is a very good idea as this doesn't load everything on-memory.

Streaming file upload with `FormData` can be done like:

```js
import { toBlob } from "https://deno.land/std@0.209.0/streams/mod.ts";

const file = await Deno.open("huge-file");
const form = new FormData();
form.set("file", toBlob(file.readable));
await fetch("url", { method: "POST", body: form });
```

With the current implementation, the above loads and stores the entire file contents in `chunks` in `toBlob` call, which is very inefficient when the file is large.

kt3k (approved):
LGTM

PR #3935, opened 2023-12-12
refactor(streams): simplify `toArrayBuffer()` implementation

iuioiua:
The current implementation is only ~2% faster according to benchmarks. IMO, having a simpler implementation wins over such a small difference in performance. I don't feel strongly about this change, so feel free to close if you disagree, Yoshiya.

Benchmark source code:
```ts
import { toArrayBuffer } from "https://deno.land/std@0.208.0/streams/to_array_buffer.ts";

Deno.bench("toArrayBuffer()", { baseline: true }, async () => {
  const stream = ReadableStream.from([
    new Uint8Array([0, 1, 2, 3]),
    new Uint8Array([4, 5, 6]),
    new Uint8Array([7, 8, 9]),
  ]);

  await toArrayBuffer(stream);
});

Deno.bench("Response.arrayBuffer()", async () => {
  const stream = ReadableStream.from([
    new Uint8Array([0, 1, 2, 3]),
    new Uint8Array([4, 5, 6]),
    new Uint8Array([7, 8, 9]),
  ]);

  await new Response(stream).arrayBuffer();
});
```

Benchmark results:
```
cpu: Apple M2
runtime: deno 1.38.5 (aarch64-apple-darwin)

benchmark                   time (avg)        iter/s             (min … max)       p75       p99      p995
---------------------------------------------------------------------------- -----------------------------
toArrayBuffer()               3.3 µs/iter     302,846.8   (2.42 µs … 267.88 µs)   2.71 µs   6.67 µs   7.75 µs
Response.arrayBuffer()       3.38 µs/iter     295,884.6     (3.31 µs … 3.71 µs)    3.4 µs   3.71 µs   3.71 µs

summary
  toArrayBuffer()
   1.02x faster than Response.arrayBuffer()
```

### DIFF
--- a/streams/to_array_buffer.ts
+++ b/streams/to_array_buffer.ts
@@ -1,11 +1,9 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { concat } from "../bytes/concat.ts";
-
 /**
  * Converts a {@linkcode ReadableStream} of {@linkcode Uint8Array}s to an
- * {@linkcode ArrayBuffer}. Works the same as{@linkcode Response.arrayBuffer}.
+ * {@linkcode ArrayBuffer}. Works the same as {@linkcode Response.arrayBuffer}.
  *
  * @example
  * ```ts
@@ -19,20 +17,7 @@ import { concat } from "../bytes/concat.ts";
  * ```
  */
 export async function toArrayBuffer(
-  readableStream: ReadableStream<Uint8Array>,
+  stream: ReadableStream<Uint8Array>,
 ): Promise<ArrayBuffer> {
-  const reader = readableStream.getReader();
-  const chunks: Uint8Array[] = [];
-
-  while (true) {
-    const { done, value } = await reader.read();
-
-    if (done) {
-      break;
-    }
-
-    chunks.push(value);
-  }
-
-  return concat(chunks).buffer;
+  return await new Response(stream).arrayBuffer();
 }


kt3k:
I personally don't think the current implementation is too complex, but I'd also like to hear other members' opinions as well.

aapoalas:
Hmm. I personally find `Response` usage here to be a bit off-putting: If the code has nothing to do with responses necessary, why is it suddenly using the Response constructor? "Because it's a neat trick."

Near tricks don't really feel proper.

iuioiua:
I get that it seems improper. But that is the only argument for having these functions in the Standard Library. Why does it matter if the function outputs the expected result for a given input without any other undesirable technical implications? Genuine question.

aapoalas:
Yeah, in some ways it definitely doesn't. Though from a purity perspective the function now has a side-effect that is hard to detect which is it's reliance on the Response constructor.

Theoretically, the previous implementation can be used in an environment where Response is not implemented or perhaps doesn't implement the `arrayBuffer()` method. The new one cannot, and you cannot know this until you run the code. Though, I do not understand how you'd end up in such an environment with Typescript as your source code :)

The previous code can similarly be blamed for requiring imports to work and would not work in an environment that doesn't support them. But that at least is easy to statically analyse.

But yeah, this is very much a matter of taste. I definitely wouldn't consider either code to be necessarily superior to the other, or to be "the correct way".

iuioiua:
I like that explanation. Yeah, I don't claim that this method is far superior to the previous. I just think it's a little simpler for near-no performance loss. Either way, I'm mostly impartial to whether this lands or not.

lino-levan:
I don't feel strongly either way. It's fine either way in my opinion. Deferring to whatever Yoshiya decides on this one.

iuioiua (commented):
Similar to https://github.com/denoland/deno_std/pull/3934#issuecomment-1855025901, the previous implementation stored in an array of chunks.

kt3k:
> Similar to https://github.com/denoland/deno_std/pull/3934#issuecomment-1855025901, the previous implementation stored in an array of chunks.

I think the same reasoning doesn't apply to this one because this returns the entire data in single ArrayBuffer. So the entire data is loaded in the runtime anyway.

iuioiua:
Actually, yes, I was wrong. My tests show that the current implementation is more memory efficient.

nhrones:
One less object creation.

PR #3936, opened 2023-12-12
fix(text): fix closestString

kt3k:
Currently `closestString` always returns the first element of the given `possibleWords`. This PR fixes it to choose closest string using levenshtein distance.

```js
import { closestString } from "https://deno.land/std@0.208.0/text/mod.ts";
console.log(closestString("a", ["b", "a"])); // this currently prints b
```

### DIFF
--- a/text/closest_string.ts
+++ b/text/closest_string.ts
@@ -46,7 +46,7 @@ export function closestString(
   }
 
   let nearestWord = possibleWords[0];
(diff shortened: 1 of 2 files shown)


aapoalas:
Nice catch

aapoalas (approved):

iuioiua (approved):

PR #3938, opened 2023-12-12
feat(expect): support message checks in expect().toThrow

syhol:
I'm attempting to support deno testing in valibot: https://github.com/fabian-hiller/valibot/pull/179 and currently trying to swap out the vitest expect for the new deno_std expect, but I came across an unsupported feature.

Currently the parameter for toThrow only supports an Error constructor, this PR also lets you pass a string, RegExp, or Error instance.

Excerpt from the Jest Expect documentation https://jestjs.io/docs/expect#tothrowerror

> You can provide an optional argument to test that a specific error is thrown:
> 
> - regular expression: error message matches the pattern
> - string: error message includes the substring
> - error object: error message is equal to the message property of the object
> - error class: error object is instance of class
> 
> For example, let's say that drinkFlavor is coded like this:
> 
> ```typescript
> function drinkFlavor(flavor) {
>   if (flavor == 'octopus') {
>     throw new DisgustingFlavorError('yuck, octopus flavor');
>   }
>   // Do some other stuff
> }
> ```
> 
> We could test this error gets thrown in several ways:
> 
> ```typescript
> test('throws on octopus', () => {
>   function drinkOctopus() {
>     drinkFlavor('octopus');
>   }
> 
>   // Test that the error message says "yuck" somewhere: these are equivalent
>   expect(drinkOctopus).toThrow(/yuck/);
>   expect(drinkOctopus).toThrow('yuck');
> 
>   // Test the exact error message
>   expect(drinkOctopus).toThrow(/^yuck, octopus flavor$/);
>   expect(drinkOctopus).toThrow(new Error('yuck, octopus flavor'));
> 
>   // Test that we get a DisgustingFlavorError
>   expect(drinkOctopus).toThrow(DisgustingFlavorError);
> });
> ```

This PR also updates the `assert/assert_is_error.ts` function to support the RegExp as a message expectation type, since thats what `expect().toThrow` uses under the hood.

I will shortly add some more comments inline.

### DIFF
--- a/assert/assert_is_error.ts
+++ b/assert/assert_is_error.ts
@@ -23,11 +23,11 @@ export function assertIsError<E extends Error = Error>(
   error: unknown,
   // deno-lint-ignore no-explicit-any
   ErrorClass?: new (...args: any[]) => E,
-  msgIncludes?: string,
+  msgIncludes?: string | RegExp,
   msg?: string,
 ): asserts error is E {
   const msgSuffix = msg ? `: ${msg}` : ".";
-  if (error instanceof Error === false) {
+  if (!(error instanceof Error)) {
     throw new AssertionError(
       `Expected "error" to be an Error object${msgSuffix}}`,
     );
@@ -38,12 +38,21 @@ export function assertIsError<E extends Error = Error>(
     }"${msgSuffix}`;
     throw new AssertionError(msg);
   }
-  if (
-    msgIncludes && (!(error instanceof Error) ||
-      !stripAnsiCode(error.message).includes(stripAnsiCode(msgIncludes)))
-  ) {
+  let msgCheck;
+  if (typeof msgIncludes === "string") {
+    msgCheck = stripAnsiCode(error.message).includes(
+      stripAnsiCode(msgIncludes),
+    );
+  }
+  if (msgIncludes instanceof RegExp) {
+    msgCheck = msgIncludes.test(stripAnsiCode(error.message));
+  }
+
+  if (msgIncludes && !msgCheck) {
     msg = `Expected error message to include ${
-      JSON.stringify(msgIncludes)
+      msgIncludes instanceof RegExp
+        ? msgIncludes.toString()
+        : JSON.stringify(msgIncludes)
     }, but got ${
       error instanceof Error
         ? JSON.stringify(error.message)
--- a/assert/assert_is_error.ts
+++ b/assert/assert_is_error.ts
@@ -23,7 +23,7 @@ export function assertIsError<E extends Error = Error>(
   error: unknown,
   // deno-lint-ignore no-explicit-any
   ErrorClass?: new (...args: any[]) => E,
-  msgIncludes?: string | RegExp,
+  msgMatches?: string | RegExp,
   msg?: string,
 ): asserts error is E {
   const msgSuffix = msg ? `: ${msg}` : ".";
@@ -39,20 +39,20 @@ export function assertIsError<E extends Error = Error>(
     throw new AssertionError(msg);
   }
   let msgCheck;
-  if (typeof msgIncludes === "string") {
+  if (typeof msgMatches === "string") {
     msgCheck = stripAnsiCode(error.message).includes(
-      stripAnsiCode(msgIncludes),
+      stripAnsiCode(msgMatches),
     );
   }
-  if (msgIncludes instanceof RegExp) {
-    msgCheck = msgIncludes.test(stripAnsiCode(error.message));
+  if (msgMatches instanceof RegExp) {
+    msgCheck = msgMatches.test(stripAnsiCode(error.message));
   }
 
-  if (msgIncludes && !msgCheck) {
+  if (msgMatches && !msgCheck) {
     msg = `Expected error message to include ${
-      msgIncludes instanceof RegExp
-        ? msgIncludes.toString()
-        : JSON.stringify(msgIncludes)
+      msgMatches instanceof RegExp
+        ? msgMatches.toString()
+        : JSON.stringify(msgMatches)
     }, but got ${
       error instanceof Error
         ? JSON.stringify(error.message)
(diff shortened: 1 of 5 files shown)


assert/assert_is_error.ts
```diff
@@ -23,11 +23,11 @@ export function assertIsError<E extends Error = Error>(
   error: unknown,
   // deno-lint-ignore no-explicit-any
   ErrorClass?: new (...args: any[]) => E,
-  msgIncludes?: string,
+  msgIncludes?: string | RegExp,
```
syhol, line 26: This is a change to the public API, however it only adds a new type and should not break any backwards compatibility.

assert/assert_is_error.ts
```diff
@@ -23,11 +23,11 @@ export function assertIsError<E extends Error = Error>(
...
   msg?: string,
 ): asserts error is E {
   const msgSuffix = msg ? `: ${msg}` : ".";
-  if (error instanceof Error === false) {
+  if (!(error instanceof Error)) {
```
syhol, line 30: This allows type narrowing so we don't have to do more checks later in the function (see line 42 in the old diff)

expect/expect.ts
```diff
@@ -79,8 +79,7 @@ export interface Expected {
   toReturnTimes(expected: number): void;
   toReturnWith(expected: unknown): void;
   toStrictEqual(candidate: unknown): void;
-  // deno-lint-ignore no-explicit-any
-  toThrow<E extends Error = Error>(expected?: new (...args: any[]) => E): void;
+  toThrow<E extends Error = Error>(expected?: unknown): void;
```
syhol, line 82: This is a change to the public API, however it only makes the parameter type wider so shouldn't break any backwards compatibility. I could provide a more correct type here, but the convention seems to be that any complex types in the expect API are just replaced with unknown. The proper type would be something like:
```typescript
interface Expect {
  toThrow<E extends Error = Error>(expected?: string | RegExp | E | (new (...args: any[]) => E)): void
}
```
Happy to do this if you think its better

expect/_to_throw_test.ts
syhol, line 1: I'm following the convention of sticking to 1 `Deno.test` per function and I've added the test description in comments. Happy to split this out into multiple tests if you would like.

assert/assert_is_error.ts
kt3k: This change makes sense to me. If we do this type change, I think the parameter name `msgMatches` would be more appropriate. What do you think?

expect/expect.ts
kt3k: The above type in your comment makes more sense to me. Can you update the typing to it?

syhol:
> I'm curious if this change is the only thing necessary for https://github.com/fabian-hiller/valibot/pull/179

It's worth emphasising that https://github.com/fabian-hiller/valibot/pull/179 passes as is, and I'm just looking to improve it further by swapping out npm:expect for deno_std expect.

Looks like it also depends on:
- [npm:expect-type](https://www.npmjs.com/package/expect-type) - Would you be interested in including this in the deno_std expect? For now I'll continue to use the npm package.
- [expect.any(Function)](https://jestjs.io/docs/expect#expectanyconstructor) - Not sure what I'm going to do about this as it seems like a lot more work to support. I might be able to refactor it out of valibot, as it looks like a messy solution in valibot anyway.
- The `toThrowError` alias of `toThrow` - Would you be interested in supporting jest expect aliases? I could also ask the owner of valibot if they would be happy with renaming the function calls to use `toThrow`. I can't imagine its a big issue.

Also I just realised that [testing/asserts.ts:assertThrows](https://deno.land/std@0.208.0/testing/asserts.ts?s=assertThrows) (soon to be [assert/assert_throws.ts](https://deno.land/std@0.208.0/testing/asserts.ts?s=assertThrows) ) exists. I think it makes much more sense to make expect.toThrow use that instead.

So this is my plan with this PR, please let me know what you think:
- Implement your change recommendations
  - Proper typing for toThrow
  - Renaming the param to msgMatches
- Applying Regex support to [assert/assert_throws.ts](https://deno.land/std@0.208.0/testing/asserts.ts?s=assertThrows) (I won't update testing/asserts.ts unless you recommend it, as it seems to be deprecated)
- Refactor expect.toThrow to use assertThrows under the hood
- Keep the assertIsError RegExp support because its called by assertThrows

How does that sound @kt3k?

syhol:
Or if you like I can just make the suggested changes and do the refactor to assertThrows in a followup PR

kt3k (approved):
LGTM

kt3k:
> [npm:expect-type](https://www.npmjs.com/package/expect-type)

Currently we don't have a plan to support this in deno_std. Please keep using `npm:expect-type` in your PR

> [expect.any(Function)](https://jestjs.io/docs/expect#expectanyconstructor) - Not sure what I'm going to do about this as it seems like a lot more work to support. I might be able to refactor it out of valibot, as it looks like a messy solution in valibot anyway.

Contribution for supporting these asymmetric matchers are welcome, but as you expect, this probably requires a lot of works, I guess..

syhol:
> We're interested in supporting aliases (We already support some of them). These additions are especially welcome

I just noticed that some aliases (including toThrowError) have been removed in the next release of jest: https://github.com/jestjs/jest/pull/14632/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR46 with that in mind, is it worth holding off on that for now?

Also I've discovered one more (last) issue that valibot has with this expect: Missing undefined properties throw an error for .toEqual where they wouldn't in jest expect. I'll create an issue for it.

kt3k:
> I just noticed that some aliases (including toThrowError) have been removed in the next release of jest: https://github.com/jestjs/jest/pull/14632/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR46 with that in mind, is it worth holding off on that for now?

Ah, right. Let's hold off

> Also I've discovered one more (last) issue that valibot has with this expect: Missing undefined properties throw an error for .toEqual where they wouldn't in jest expect. I'll create an issue for it.

Thanks!

PR #3940, opened 2023-12-12
perf(encoding): increase `varint.decode()` performance by 45x

MierenManz:


lino-levan:
Could you put some of the screenshots you shared on discord in this thread for context?

lino-levan (approved):
LGTM

MierenManz:
image
After changing the signature of the function slightly to match that of the current std implementation it's 45x faster :woman_shrugging:

encoding/varint.ts
```diff
@@ -34,22 +38,51 @@ const SHIFTN = 7n;
...
-    if (!(byte & MSB) && decoded > MaxUInt64) {
-      throw new RangeError("overflow varint");
+  U64_VIEW[0] = 0n;
+  let intermediate = 0;
+  let position = 0;
+  let i = offset;
+
+  if (buf.length === 0) throw new RangeError("Cannot read empty buffer");
+
+  let byte = buf[i];
```
iuioiua, line 48:
```suggestion
  let byte;
```

encoding/varint.ts
```diff
@@ -34,22 +38,51 @@ const SHIFTN = 7n;
...
-    decoded += BigInt((byte & REST) * Math.pow(2, shift));
-    if (!(byte & MSB) && decoded > MaxUInt64) {
-      throw new RangeError("overflow varint");
+  U64_VIEW[0] = 0n;
+  let intermediate = 0;
+  let position = 0;
+  let i = offset;
+
+  if (buf.length === 0) throw new RangeError("Cannot read empty buffer");
+
```
iuioiua, line 47: Can you please add a simple test case for this scenario?

encoding/varint.ts
```diff
@@ -34,22 +38,51 @@ const SHIFTN = 7n;
...
+    (adjustedOffset === 10 && intermediate > -1) ||
+    adjustedOffset === 11 || i > buf.length
+  ) {
+    throw new RangeError("malformed or overflow varint");
   }
-  throw new RangeError("malformed or overflow varint");
+
+  // Write the intermediate value to the "empty" slot
+  // if the first slot is taken. Take the second slot
+  U32_VIEW[Number(adjustedOffset > 4)] = intermediate;
```
iuioiua, line 83:
```suggestion
  U32_VIEW[adjustedOffset > 4] = intermediate;
```

iuioiua (changes requested):
Nice work! Just a couple of nits. I think the added performance is worth the added complexity.

encoding/varint.ts
lino-levan: What's funny is depending on some variables, removing the assignment here could reduce the performance due to how v8 does inlining. Incredibly weird. I think this should be noted in a comment here.

encoding/varint.ts
lino-levan: This should probably be mentioned in a JSDoc imo.

encoding/varint.ts
iuioiua: Which variables? Benchmarks on my machine actually showed a small bump in performance from this small change.

encoding/varint.ts
lino-levan: I'll let Skye advise on this one, but I suspect this is another optimization. Should be noted in a comment if so.

encoding/varint.ts
lino-levan: Skye and I were discussing this months (?) ago when she originally wrote this. I don't remember, deferring to Skye here. Perhaps a new version of v8 made this optimization worthless 🤷.

encoding/varint.ts
kt3k: This `Number` looks just necessary to me (You can't use boolean for indicating first and second slot):

```
> let buf = new Uint32Array([0, 1]);
undefined
> buf[true]
undefined
> buf[Number(true)]
1
```

iuioiua: Ah, yes, my misunderstanding. Nevermind.

lino-levan: Ah completely misread this (thought it was a bitshift). Nevermind!

encoding/varint.ts
MierenManz: it seems like it makes no difference anymore in the benchmarks. I guess this was just a style thing last time lol

MierenManz:
> I think the added performance is worth the added complexity.
For maintenance would it be a good idea if I documented / explained the function somewhere in detail? Like why it works and what makes it faster compared to the standard algorithm in typescript

iuioiua:
I think the JSDocs for that function is a good spot for those extra technical details. Does that sound good, @kt3k?

Also, please run `deno fmt`.

kt3k:
> For what it's worth would it be a good idea if I documented / explained the function somewhere in detail? Like why it works and what makes it faster compared to the standard algorithm in typescript

That sounds very helpful to maintainers & contributors.

MierenManz:
I've added more comments to the code explaining what it is and why it's there.
Most should be pretty obvious now (I hope)

iuioiua (approved):
Excellent work, Skye. Thank you.

PR #3943, opened 2023-12-13
chore(http): deprecate `CookieMap()`

### LINKED ISSUE #3942: proposal: deprecate `std/http/cookie_map`
The main reason for doing this is that the API overlaps with `std/http/cookie`, which is widely used and simpler. The gap between these APIs is the ability to work with signed cookies, which does seem like a valid use case.

There are two paths we can take:
1. Implement our own signed cookie implementation, which instead compliments the `std/http/cookie` API. #3905 explores this idea, which uses a simpler implementation approach.
2. Point to [x/jose](https://deno.land/x/jose), which is a well-maintained module that takes care of this functionality.

Either way, removing this API and, therefore, the competing implementations is probably best. Opinions are welcome.

Related #3655

iuioiua:
Closes #3942
Prerequisite #3905

### DIFF
--- a/http/cookie_map.ts
+++ b/http/cookie_map.ts
@@ -2,7 +2,7 @@
 // This module is browser compatible.
 
 /**
- * @deprecated (will be removed after 0.210.0) Import from {@link https://deno.land/std/crypto/unstable_cookie_map.ts} instead.
+
  *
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #3944, opened 2023-12-13
fix(cli): re-export promptSecret from mod.ts

kt3k:
This PR adds re-export of `promptSecret` from mod.ts

ref https://github.com/denoland/deno_std/pull/3777


iuioiua (approved):
Good catch

PR #3950, opened 2023-12-13
BREAKING(crypto): move to unstable `KeyStack()`

iuioiua:


iuioiua:
Hm... The more I look at `KeyStack`, the more I like it... Might be worth gathering some more info before doing this.

PR #3953, opened 2023-12-13
chore(semver): regex template strings

timreichen:
replaces regex strings with template strings.

### DIFF
--- a/semver/_shared.ts
+++ b/semver/_shared.ts
@@ -150,7 +150,7 @@ export const HYPHENRANGE_REGEXP = new RegExp(
 );
 
 // Star ranges basically just allow anything at all.
-export const STAR_REGEXP = new RegExp(`(<|>)?=?\\s*\\*`);
+export const STAR_REGEXP = /(<|>)?=?\s*\*/;
 
 /**
  * Returns true if the value is a valid SemVer number.
--- a/semver/_shared.ts
+++ b/semver/_shared.ts
@@ -71,13 +71,13 @@ export function compareIdentifier(
 // ## Numeric Identifier
 // A single `0`, or a non-zero digit followed by zero or more digits.
 
-const NUMERICIDENTIFIER = `0|[1-9]\\d*`;
+const NUMERICIDENTIFIER = "0|[1-9]\\d*";
 
 // ## Non-numeric Identifier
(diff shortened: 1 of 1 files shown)


semver/_shared.ts
```diff
@@ -71,13 +71,13 @@ export function compareIdentifier(
 // ## Numeric Identifier
 // A single `0`, or a non-zero digit followed by zero or more digits.
 
-const NUMERICIDENTIFIER = "0|[1-9]\\d*";
+const NUMERICIDENTIFIER = `0|[1-9]\\d*`;
```
iuioiua, line 74:
```suggestion
const NUMERIC_IDENTIFIER = `0|[1-9]\\d*`;
```
I think we should do the same with the other constants (if they're private) to make them more readable.

semver/_shared.ts
```diff
@@ -87,30 +87,23 @@ const MAINVERSION =
...
-  PRERELEASEIDENTIFIER +
-  ")*))";
+const PRERELEASE =
+  `(?:-(${PRERELEASEIDENTIFIER}(?:\\.${PRERELEASEIDENTIFIER})*))`;
 
 // ## Build Metadata Identifier
 // Any combination of digits, letters, or hyphens.
-
-const BUILDIDENTIFIER = "[0-9A-Za-z-]+";
+const BUILDIDENTIFIER = `[0-9A-Za-z-]+`;
```
iuioiua, line 101: This change is unnecessary, as no string interpolation is happening here. Ditto to other instances.

iuioiua (changes requested):
Mostly LGTM. Couple of nits.

semver/_shared.ts
timreichen: I agree, I was going to update the names in a separate PR. Do you want to do it in this PR?

iuioiua: Ah, actually, doing it in another PR is a good idea. It might be worth checking the rest of the sub-module for similar fix opportunities.

semver/_shared.ts
timreichen: done

iuioiua (approved):
LGTM. Nice, this makes this more readable and reduces LOC. Thank you.

PR #3955, opened 2023-12-13
chore(semver): rename internal constants for better readability

timreichen:
rename some internal constants for more readability.


iuioiua (approved):
LGTM. Nice.

PR #3957, opened 2023-12-13
deprecation(semver): rename `isSemVerComparator()`

timreichen:
renames
`SemVerComparator` => `Comparator`
`isSemVerComparator()` => `isComparator()`
~~`isSemVerRange()` => `isRange()`~~
~~`SemVerRange` => `Range`~~
All old functions and interfaces are deprecated.


semver/is_semver_comparator_test.ts
iuioiua, line 1: I think this file should be `is_comparator_test.ts`.

semver/is_semver_range.ts
```diff
@@ -11,16 +11,8 @@ import { isSemVerComparator } from "./is_semver_comparator.ts";
  * Adds a type assertion if true.
  * @param value The value to check if its a valid SemVerRange
  * @returns True if its a valid SemVerRange otherwise false.
+ * @deprecated (will be removed in 0.211.0) use `Range` instead.
```
iuioiua, line 14:
```suggestion
 * @deprecated (will be removed in 0.212.0) Use {@linkcode isRange} instead.
```

semver/is_semver_comparator.ts
```diff
@@ -13,18 +11,8 @@ import { ALL, NONE } from "./constants.ts";
  * Adds a type assertion if true.
  * @param value The value to check if its a SemVerComparator
  * @returns True if the object is a SemVerComparator otherwise false
+ * @deprecated (will be removed in 0.211.0) use `isComparator` instead.
```
iuioiua, line 14:
```suggestion
 * @deprecated (will be removed in 0.212.0) Use {@linkcode isComparator} instead.
```

semver/types.ts
```diff
@@ -46,6 +46,15 @@ export type FormatStyle =
...
  * @example >=0.0.0
  */
+export interface Comparator {
+  operator: Operator;
+  semver: SemVer;
+  min: SemVer;
+  max: SemVer;
+}
+/**
+ * @deprecated (will be removed in 0.211.0) use `Comparator` instead.
```
iuioiua, line 56:
```suggestion
 * @deprecated (will be removed in 0.212.0) Use {@linkcode Comparator} instead.
```

iuioiua (changes requested):

timreichen:
@iuioiua Before merging: There is also an opportunity to simplify `Range` and lose the single `ranges` property:

before
```ts
export interface Range {
  // The outer array is OR while each inner array is AND
  ranges: SemVerRangeOr;
}
```
after
```ts
// The outer array is OR while each inner array is AND
export type Range = Comparator[][];
```
WDYT?

semver/is_semver_comparator_test.ts
timreichen: Done

semver/is_semver_comparator.ts
```diff
@@ -13,18 +11,8 @@ import { ALL, NONE } from "./constants.ts";
...
-  if (typeof value !== "object") return false;
-  const { operator, semver, min, max } = value as SemVerComparator;
-  return (
-    isValidOperator(operator) &&
-    isSemVer(semver) &&
-    isSemVer(min) &&
-    isSemVer(max)
-  );
+  return isComparator(value);
 }
```
iuioiua, line 18:
```suggestion
export const isSemVerComparator = isComparator;
```
Sorry, should've spotted this earlier.

semver/is_semver_range.ts
```diff
@@ -11,16 +11,8 @@ import { isSemVerComparator } from "./is_semver_comparator.ts";
...
-  if (typeof value !== "object") return false;
-  const { ranges } = value as SemVerRange;
-  return (
-    Array.isArray(ranges),
-      ranges.every((r) =>
-        Array.isArray(r) && r.every((c) => isSemVerComparator(c))
-      )
-  );
+  return isRange(value);
 }
```
iuioiua, line 18:
```suggestion
export const isSemVerRange = isRange;
```

iuioiua (changes requested):
Sorry, I missed a couple of nits.

iuioiua (changes requested):
Sorry, I missed a couple of nits.

timreichen:
No problem.

> @iuioiua Before merging: There is also an opportunity to simplify `Range` and lose the single `ranges` property:
> 
> before
> 
> ```ts
> export interface Range {
>   // The outer array is OR while each inner array is AND
>   ranges: SemVerRangeOr;
> }
> ```
> 
> after
> 
> ```ts
> // The outer array is OR while each inner array is AND
> export type Range = Comparator[][];
> ```
> 
What is your opinion on this?

timreichen:
Actually, looking into Range a bit more, it seems that the current implementation is kinda strange.
It is defined like multiple segments on a line. Imo Range implies to be a single segment.

iuioiua:
Yeah, it is strange. I see repeated use of `range.ranges`. It should probably be just `ranges` (without the `ranges` property).

timreichen:
How about we redefine Range to `export type Range = Comparator[]` and remove one OR layer. The OR layer is based on `parseRange` implementation which can parse `range || range`, but  it should actually return Range[] instead of Range and be called `parseRanges` instead of `parseRange`.

iuioiua:
The first solution sounds better.

timreichen:
> Yeah, it is strange. I see repeated use of `range.ranges`. It should probably be just `ranges` (without the `ranges` property).

The more I investigate about the Range implementation, the more am I convinced that the two renames
- isSemVerRange() => isRange()
- SemVerRange => Range

are misleading.

Their actual definition of SemVerRange is not a single Range but an array of Ranges.
Instead of renaming Ranges, we should implement `Range` and also adjust the API functions to accept Range instead of Ranges. Basically all range* functions (`rangeFormat`, `rangeMin`, `rangeMax`, `rangeIntersects`, `minSatisfying`, `maxSatisfying`,  `testRange`, etc.) just loop over the Range array anyway. So instead I would suggest changing these to accept a single Range and get rid of all these loops. WDYT?

**Please do not merge** until this fundamental problem is solved.

iuioiua:
Are you able to open a new issue focusing on this weird nested range design? There may be a reason (or there may not be) for this design decision.

Either way, let's proceed with this PR as it focuses on removing the `SemVer` portion of these symbol names.

timreichen:
@iuioiua I will open a new issue for this. This misnomer is also present in the npm semver.  Even the bnf file https://github.com/npm/node-semver/blob/main/range.bnf calls it `range-set`.

Would you be ok if I reduced this PR to
- `isSemVerComparator()` => `isComparator()`
- `SemVerComparator` => `Comparator`

so it can be merged?
Else we potentially would need to re-deprecate the `Range` later again and wait a cycle until we can fix the problem.

iuioiua:
Yep, let's reduce the PR 👍🏾

semver/is_semver_comparator.ts
timreichen: done

timreichen:
> Yep, let's reduce the PR 👍🏾

done

iuioiua (approved):
LGTM! Thanks! Let's see what Yoshiya thinks.

kt3k (approved):
LGTM

PR #3958, opened 2023-12-13
deprecation(semver): deprecate `rcompare()`

timreichen:
add deprecation notice to `rcompare`

### DIFF
--- a/semver/rcompare.ts
+++ b/semver/rcompare.ts
@@ -7,7 +7,7 @@ import { compare } from "./compare.ts";
  * `1` and `-1` are inverted.
  *
  * Sorts in descending order if passed to `Array.sort()`,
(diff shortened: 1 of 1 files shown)


semver/rcompare.ts
```diff
@@ -7,6 +7,7 @@ import { compare } from "./compare.ts";
  * `1` and `-1` are inverted.
  *
  * Sorts in descending order if passed to `Array.sort()`,
+ * @deprecated (will be removed in 0.211.0) use `compare` instead.
```
iuioiua, line 10:
```suggestion
 * @deprecated (will be removed in 0.212.0) Use {@linkcode compare} instead.
```

iuioiua (changes requested):

iuioiua (approved):
LGTM

kt3k (approved):
LGTM

Issue #3394, opened 2023-05-23
JSON support in loggers


tmcw:
**Is your feature request related to a problem? Please describe.**

Logging services - in my case [logtail](https://logtail.com/) - are increasingly involved in parsing logs and giving access to structured logs. The most straightforward option is to log JSON, though some also support logfmt.

The Deno logging module is great, but it doesn't work for logging JSON: at an early stage, it stringifies anything that is logged:

https://deno.land/std@0.188.0/log/logger.ts?source#L95

**Describe the solution you'd like**

I'd prefer the stringify step in the logging module to be at the handler stage, not at the logger stage, so that a `JSONHandler`, similar to `ConsoleHandler`, could be written in userspace.

**Describe alternatives you've considered**

There are other logging modules, but it'd be nice to use the Deno-std one.

osdevisnot:
💯 This is especially relevant now that #golang added https://pkg.go.dev/log/slog for structured logging. Also, it seems the std logger is slightly unconventional to use `warning` instead of `warn` method; the latter of which is more common in javascript.

syhol:
The [JSON logging PR](https://github.com/denoland/deno_std/pull/3960) has been merged now, could we consider this issue closed?

iuioiua:
Thanks, @syhol!

PR #3960, opened 2023-12-14
feat(log): Added support for structured JSON logging

syhol:
Implementing a solution for this issue: https://github.com/denoland/deno_std/issues/3394

**What is it?**
Lets you configure the logger to emit structured json logs that will play nicely with log aggregators

**Why do it?**
People like structured json logs:
- https://stackify.com/what-is-structured-logging-and-why-developers-need-it/
- https://sematext.com/glossary/structured-logging/
- https://www.loggly.com/use-cases/what-is-structured-logging-and-how-to-use-it/

TLDR; Easier parsing makes it more reliable, cheap, and fast to gain insight into your logs, whether thats for monitoring, root cause analysis, or business intelligence.

**What does it look like?**

```ts
import * as log from "https://deno.land/std@$STD_VERSION/log/mod.ts";

log.setup({
  handlers: {
    default: new log.handlers.ConsoleHandler("DEBUG", {
      formatter: log.formatters.jsonFormatter,
      useColors: false,
    }),
  },
});

log.info("Hey");
// > {"level":"INFO","datetime":1702481922294,"message":"Hey"}

log.info("Hey", { product: "nail" });
// > {"level":"INFO","datetime":1702484111115,"message":"Hey","args":{"product":"nail"}}

log.info("Hey", 1, "two", [3, 4, 5]);
// > {"level":"INFO","datetime":1702481922294,"message":"Hey","args":[1,"two",[3,4,5]]}
```

The changes include:
- A new `log/formatters.ts` module with the new `jsonFormatter` function
  - Other formatters in future could include logfmt and maybe protobufs
  - DEFAULT_FORMATTER and FormatterFunction could be moved here
- Most log aggregators I've worked with didn't like colour codes, so `ConsoleHandler` now has a new optional property in its constructor parameter `useColors?: boolean` that defaults to true, keeping backwards compatibility. Only when explicitly set to false will the handler skip applying colours. 
  - I could have created a new handler but I couldn't think of a good name name.
- Exporting a new `const formatters` from the `log/mod.ts` so we can use the formatter in a similar way to how we use handlers: `log.formatters.jsonFormatter`
- A ton of docs, please note that I've documented the following topics & quirks:
  - How to configure the logger to emit non-coloured json
  - How the first argument is always treated as the message and will be double encoded if not a string
  - How the emitted json looks based on different types of input
  - How to create a custom json formatter

### DIFF
--- a/log/formatters_test.ts
+++ b/log/formatters_test.ts
@@ -1,9 +1,10 @@
 // Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
 import { assertEquals } from "../assert/mod.ts";
+import { FakeTime } from "../testing/time.ts";
 import { jsonFormatter } from "./formatters.ts";
 import { LogRecord } from "./logger.ts";
 
-const makeRecord = (msg: string, args: unknown[] = []) =>
+const log = (msg: string, args: unknown[] = []) =>
   new LogRecord({
     msg,
     args,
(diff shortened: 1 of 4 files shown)


log/formatters_test.ts
```diff
@@ -0,0 +1,45 @@
...
+    loggerName: "user-logger",
+  });
+
+Deno.test("jsonFormatter with just a message", function () {
+  const result = JSON.parse(jsonFormatter(makeRecord("User service exploded")));
+
+  assertEquals(result.level, "INFO");
+  assertEquals(typeof result.datetime, "number");
+  assertEquals(result.message, "User service exploded");
+  assertEquals(Object.keys(result), ["level", "datetime", "message"]);
```
syhol, line 20: If I mock the time then I can write much nicer assertions, I'm just waiting for https://github.com/denoland/deno_std/pull/3961 to land so the test look pretty.

kt3k (approved):
LGTM. Thanks!

PR #3961, opened 2023-12-14
feat(testing): enable the `using` keyword on `FakeTime()`

syhol:
Enable the "using" keyword on FakeTime by adding a `Symbol.dispose` method that just calls `restore()`.

Meaning we can change this code:

```ts
Deno.test("Testing something here?", () => {
  const time = new FakeTime(1234);
  try {
    const result = callMyThing();
    assertEquals(result, 1234)
  } catch {
    time.restore()
  }
});
```

to this:

```ts
Deno.test("Testing something here?", () => {
  using time = new FakeTime(1234);
  const result = callMyThing();
  assertEquals(result, 1234)
});
```


iuioiua (changes requested):
Nice! I like this a lot. We should have this anywhere we can in the Standard Library.

Are you able to modify any tests that use `FakeTime()` to take advantage of this new functionality?

syhol:
Updated all but one use of FakeTime, the one instance I didn't update needed to be reset in a strange place so I left it alone

iuioiua (approved):
Great work! Thank you.

PR #3962, opened 2023-12-14
deprecation(semver): rename `format*()` to `stringify*()`

timreichen:
renames
- `comparatorFormat()` => `stringifyComparator()`
- `format()` => `stringify()`
-~~`rangeFormat()` => `stringifyRange()`~~

deprecates all old functions.

**Note** `stringify()` has a different api than `format()`: It does not take a `FormatStyle` but stringifies everything that is passed as an `Partial<SemVer>`


iuioiua (changes requested):

iuioiua (changes requested):
Mostly LGTM. Few nits.

semver/stringify_comparator.ts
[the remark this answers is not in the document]
timreichen: On it.

semver/stringify_range.ts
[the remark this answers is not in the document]
timreichen: On it.

iuioiua (approved):
I've added the corresponding test files. Now, LGTM! Thank you. Let's see what Yoshiya thinks.

iuioiua (approved):
I've added the corresponding test files. Now, LGTM! Thank you. Let's see what Yoshiya thinks.

timreichen:
This also concerns `stringifyRange`: https://github.com/denoland/deno_std/pull/3957#issuecomment-1857116472

timreichen:
@iuioiua similar to https://github.com/denoland/deno_std/pull/3957, I also revert the `Range` changes here.

iuioiua (approved):
LGTM again

semver/stringify.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Format a SemVer object into a string.
+ *
+ * If any number is NaN then NaN will be printed.
+ *
+ * If any number is positive or negative infinity then '∞' or '⧞' will be printed instead.
+ *
+ * @param semver The semantic version to format
+ * @returns The string representation of a semantic version.
+ */
+export function stringify(semver: Partial<SemVer>) {
```
kt3k, line 25: Why support `Partial<SemVer>`? I think the API should only deal with full `Semver`

kt3k:
I don't see the motivation of this change well. `stringify` sounds confusing than `format` to me.

semver/stringify.ts
timreichen: The current `format` implementation supports partial stringification by passing a `FormatStyle`. `Partial<SemVer>` basically replaces the `FormatStyle`:

Before:
```ts
format({ major: 1, minor: 2, patch: 3, prerelease:[], build: [] }, "patch") // returns 3
format({ major: 1, minor: 2, patch: 3, prerelease:[], build: [] }, "primary") // returns 1.2.3
format({ major: 1, minor: 2, patch: 3, prerelease:["pre"], build: [] }, "release") // returns 1.2.3-pre
```
After: 
```ts
stringify({ patch: 3 }) // returns 3
stringify({ major: 1, minor: 2, patch: 3 }) // returns 1.2.3
stringify({ major: 1, minor: 2, patch: 3, prerelease: ["pre"] }) // returns 1.2.3-pre
```

timreichen:
> I don't see the motivation of this change well. `stringify` sounds confusing than `format` to me.

Reason is that the name fits the function better. The SemVer is not formatted but predicatively turned into a string. It is also more intuitive name because there are lots of  parse/stringify function pairs in std and web APIs (JSON for example) that do the reverse process from each other.

kt3k:
From my observation, `stringify` name is usually used for serialization of complex object/data to string representation (such as the case of toml, yaml, csv, json, etc), on the other hand, `format` is used for formatting something into human-readable form (such as the case of `fmt/bytes`, `fmt/duration`). I think `format` function of semver module belongs to the 2nd category because it can't contain complex data in it.

timreichen:
> From my observation, `stringify` name is usually used for serialization of complex object/data to string representation (such as the case of toml, yaml, csv, json, etc), on the other hand, `format` is used for formatting something into human-readable form (such as the case of `fmt/bytes`, `fmt/duration`). I think `format` function of semver module belongs to the 2nd category because it can't contain complex data in it.

I don't think complexity is a a parameter.  `format()` normally has options how it should be formatted, because there are multiple ways of a value representation (for example `style: "narrow" | "digital" | "full"` in fmt/duration or ` style: 'currency'` in `Intl.NumberFormat`. That is not the case with `stringify()`.
And a SemVer only has one string represention.

iuioiua:
I agree with Tim's points. These functions appear to be more fitting for `stringify*` names.

semver/stringify.ts
kt3k: `{ patch: 3 }` is not a semver. I don't see the benefit of dealing with it. What is the use case of such cases?

On the other hand, `format` only deals with actual semvers. That looks more reasonable design to me.

timreichen: I cannot think of any use case because one might just call `String(version.patch)` or `${version.patch}` or `version.patch.toString()` for a single stringified value.

I think the main point of annoyance with `SemVer` as a type is the need to pass `prerelease` and `build` even if they are empty arrays, because they can be inferred.

I think there are two ways to go about this:

1. Changing the type to
```ts
export function stringify(semver: {
    major: number;
    minor: number;
    patch: number;
    prerelease?: (string | number)[];
    build?: string[];
  }) { ... }
```

or use `SemVer` but openin a PR that makes `prerelease` and `build` optional in general. That would also simplify other parts of the mod.

WDYT?

kt3k: > 2. Use SemVer but opening a PR that makes prerelease and build optional in general. That would also simplify other parts of the mod.

This makes sense to me.

kt3k:
> That is not the case with stringify().

JSON.stringify has formatting options: `replacer` and `space`

I still don't see the motivation of the rename. The first line of jsdoc of `stringify` says `Format a SemVer object into a string`. Doesn't this imply `format` is the natural name for it?

I kind of agree that `style` option of `format` is not useful. Why not just deprecate that option?

timreichen:
> > That is not the case with stringify().
> 
> JSON.stringify has formatting options: `replacer` and `space`
> 

That is true. The replacer function is basically a property preprocessor function. `space` on the other hand actually changes the format of the output.

> I still don't see the motivation of the rename. The first line of jsdoc of `stringify` says `Format a SemVer object into a string`. Doesn't this imply `format` is the natural name for it?
> 

I think we just need to be consistent in std, because we have both names `format` and `stringify` which do the same task.
How would you differentiate between the two?

> I kind of agree that `style` option of `format` is not useful. Why not just deprecate that option?

I agree, that option is not useful.

kt3k:
> I think we just need to be consistent in std, because we have both names format and stringify which do the same task.
How would you differentiate between the two?

In my view `stringify` is synonymous to `serialize`. The output can be extremely long and unreadable to humans, but they are machine readable. The output of `format` is mainly for human readability on the other hand.

timreichen:
> In my view `stringify` is synonymous to `serialize`. The output can be extremely long and unreadable to humans, but they are machine readable. The output of `format` is mainly for human readability on the other hand.

Ok, I see what you mean. I will close this PR for now, but I think we should find an objective definition. `Human readable` is a really loose term, wasn't `json` designed to be human readable? Some of the std modules are using the terms interchangeably (toml `stringify` calls `dump` on an `Dumper` instance which internally calls `#format` for example).

PR #3968, opened 2023-12-15
feat(cli): command line spinner

darkdarcool:
These are just a a few functions similar to [`ora`](https://github.com/sindresorhus/ora)'s. 

Whenever I create a CLI app, I find myself not wanting to add a third-party spinner package and adding this code myself. I think this addition to the CLI module is a helpful one for some people. I'm not exactly sure how to add tests for this, since it depends a lot on timing and `stdout`, but I'm open to adding tests if someone finds a way. 

Like #3777, please tell me if this should not be included in the standard lib, and I'll just make a small third party package for it.


cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
+// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+
+const encoder = new TextEncoder();
+
+const LINE_CLEAR = encoder.encode("\r\u001b[K"); // From cli/prompt_secret.ts
+const COLOR_RESET = "\u001b[0m";
+const DEFAULT_SPEED = 75;
+
+/** The options for the spinner */
```
iuioiua, line 9:
```suggestion
/** Options for {@linkcode Spinner}. */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+
+  /**
+   * Creates a new spinner.
+   * @param message - The message to display next to the spinner.
+   * @param speed - The speed of the spinner.
+   * @param spinner - The spinner to use.
+   * @param color - The color of the spinner.
+   * @param cleanup - Whether to cleanup the output if the process is terminated.
+   */
+  constructor(message: string | undefined, speed: number, spinner: string[], color: string | undefined, cleanup: boolean) {
```
iuioiua:
```suggestion
  constructor(options?: SpinnerOptions) {
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+/**
+ * A spinner that can be used to indicate that something is loading.
+ */
+export class Spinner {
+  #message: string | undefined;
+  #speed: number;
+  #spinner: string[];
+  #color: string | undefined;
+  #cleanup: boolean;
+
```
iuioiua:
```suggestion
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+const COLOR_RESET = "\u001b[0m";
+const DEFAULT_SPEED = 75;
+
+/** The options for the spinner */
+export interface SpinnerOptions {
+  /** The spinner to be animated */
+  spinner: string[];
+  /** The message to display next to the spinner */
+  message?: string;
+  /** The speed of the spinner. Defaults to 75 */
```
iuioiua, line 15:
```suggestion
  /**
   * The speed of the spinner.
   *
   * @default {75}
   */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+
+const encoder = new TextEncoder();
+
+const LINE_CLEAR = encoder.encode("\r\u001b[K"); // From cli/prompt_secret.ts
+const COLOR_RESET = "\u001b[0m";
+const DEFAULT_SPEED = 75;
+
+/** The options for the spinner */
+export interface SpinnerOptions {
+  /** The spinner to be animated */
```
iuioiua, line 11:
```suggestion
  /**
   * The sequence of characters to be iterated through for animation.
   *
   * @default {["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"]}
   */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+  spinner: string[];
+  /** The message to display next to the spinner */
```
iuioiua, line 13:
```suggestion
  /** The message to display next to the spinner. */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+  message?: string;
+  /** The speed of the spinner. Defaults to 75 */
+  speed?: number;
+  /** The color of the spinner. Defaults to the default terminal color */
```
iuioiua, line 17:
```suggestion
  /** The color of the spinner. Defaults to the default terminal color. */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+  color?: string;
+  /** Whether to cleanup the output if the process is terminated. Defaults to true */
```
iuioiua, line 19:
```suggestion
  /**
   * Whether to cleanup the output if the process is terminated.
   *
   * @default {true}
   */
```

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+
+/** The default spinners */
+export const spinners = {
+  lines: ['-', '\\', '|', '/'],
+  binary: ['0000', '1000', '1100', '1110', '1111', '0111', '0011', '0001'],
+  dots: ['⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'],
+  dots2: ['⣾', '⣽', '⣻', '⢿', '⡿', '⣟', '⣯', '⣷'],
+  blink: ['◜', '◠', '◝', '◞', '◡', '◟'],
+}
+
```
iuioiua: I think we should provide a single reasonable default spinner. The `dots` spinner LGTM because it's the one Homebrew uses (my opinion).

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Creates a new spinner.
+   * @param message - The message to display next to the spinner.
+   * @param speed - The speed of the spinner.
+   * @param spinner - The spinner to use.
+   * @param color - The color of the spinner.
+   * @param cleanup - Whether to cleanup the output if the process is terminated.
+   */
+  constructor(message: string | undefined, speed: number, spinner: string[], color: string | undefined, cleanup: boolean) {
+    this.#message = message;
```
iuioiua, line 59: This should be an `options` object with a `SpinnerOptions` type.

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+  #intervalId: number | undefined;
+
+  /**
+   * Creates a new spinner.
+   * @param message - The message to display next to the spinner.
+   * @param speed - The speed of the spinner.
+   * @param spinner - The spinner to use.
+   * @param color - The color of the spinner.
+   * @param cleanup - Whether to cleanup the output if the process is terminated.
+   */
```
iuioiua, line 51: These JSDoc directives aren't needed if we go with the below suggestion. Are you able to provide an `@example` block here?

cli/spinner.ts
```diff
@@ -0,0 +1,163 @@
...
+ *  speed: 20
+ * });
+ * 
+ * mySpinner.start();
+ * ```
+ */
+export function createSpinner({ message, speed = DEFAULT_SPEED, spinner, color, cleanup = true }: SpinnerOptions): Spinner {
+  return new Spinner(message, speed, spinner, color, cleanup);
+}
+
```
iuioiua, line 152: I don't think this function is needed as it's just a wrapper around the `Spinner` constructor.

iuioiua (changes requested):

iuioiua (changes requested):
I'm unsure about adding this also. It seems like something useful that other modules could rely on. However, it could become too opinionated if we're not careful. Either way, I've added a few suggestions. I'm happy to defer to whatever the concensus is for this addition.

I think the only exported symbols here should be `Spinner` and `SpinnerOptions`.

darkdarcool:
@iuioiua Thanks for the feedback, I incorporated everything. Also, do you have any suggestions for how I might add tests for this?

iuioiua:
Sorry, I don't see the changes. Did you push your changes?

I'm not sure how to test this yet. We may need to do with manual testing for first pass in this case.

darkdarcool:
Think I fixed it now.

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+  #intervalId: number | undefined;
+  #active = false;
+
+  /**
+   * Creates a new spinner.
+   * @param options - The options for the spinner.
+   *
+   * @example
+   * ```ts
+   * const spinner = new Spinner({ message: "Loading..." });
```
iuioiua, line 50: For example blocks to work, you must also include imports. Example [here](https://github.com/denoland/deno_std/blob/main/collections/aggregate_groups.ts).

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+    this.#active = true;
+
+    let i = 0;
+
+    const message = this.#options.message ?? "";
+    const spinner = this.#options.spinner ?? DEFAULT_SPINNER;
+    const speed = this.#options.speed ?? DEFAULT_SPEED;
+    const cleanup = this.#options.cleanup ?? true;
+
+    const color = this.#options.color ?? "";
```
iuioiua, line 79: Some form of this block should be defined in the constructor.

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  speed?: number;
+  /** The color of the spinner. Defaults to the default terminal color. */
+  color?: string;
+  /**
+   * Whether to cleanup the output if the process is terminated.
+   *
+   * @default {true}
+   */
+  cleanup?: boolean;
```
iuioiua, line 32: Hm... I wonder whether this option and the associated functionality are needed. I feel like it might not be. WDYT?

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    if (Deno.stdout.writable.locked) return;
+
+    // Updates the spinner after the given interval.
+    const updateFrame = () => {
+      Deno.stdout.writeSync(LINE_CLEAR);
+      let frame;
+      // If color was specified, add it to the frame. Otherwise, just use the spinner and message.
+      if (this.#options.color) {
+        frame = encoder.encode(color + spinner + COLOR_RESET + " " + message);
```
iuioiua, line 89:
```suggestion
        frame = encoder.encode(color + spinner[i] + COLOR_RESET + " " + message);
```

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+      spinner: DEFAULT_SPINNER,
+      color: undefined,
+      cleanup: true,
+    };
+  }
+
+  /**
+   * Starts the spinner.
+   */
+  start() {
```
iuioiua, line 67: Are you able to add an example here?

cli/spinner.ts
```diff
@@ -0,0 +1,115 @@
...
+    // Cleanup if the process is terminated.
+    if (cleanup) {
+      Deno.addSignalListener("SIGINT", this.stop.bind(this));
+    }
+  }
+
+  /**
+   * Stops the spinner.
+   */
+  stop() {
```
iuioiua, line 107: Are you able to an example here?

iuioiua (changes requested):
Looking better! Few more suggestions.

cli/spinner.ts
darkdarcool: After looking at it from the lens of someone using it, I don't see it getting used very much, and it being a hindrance to debugging when you stop a program for it to just clean up the output by default.

iuioiua (approved):
I made some tweaks and fixed the formatting. Now, LGTM.

I'm impartial to adding this functionality, but I'm happy to go along with the consensus.

cli/spinner.ts
```diff
@@ -0,0 +1,109 @@
...
+   * @default {["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"]}
+   */
+  spinner?: string[];
+  /** The message to display next to the spinner. */
+  message?: string;
+  /** The speed of the spinner.
+   *
+   * @default {75}
+   */
+  speed?: number;
```
kt3k, line 24: `speed` sounds strange to me as lower number gives faster speed of spinner. How about `internal`? ref: https://github.com/sindresorhus/ora?tab=readme-ov-file#interval

cli/spinner.ts
```diff
@@ -0,0 +1,109 @@
...
+  /** The color of the spinner. Defaults to the default terminal color. */
+  color?: string;
```
kt3k, line 26: It seems this needs to be ANSI color code. On the other hand, `ora` supports human-readable color names (`'black' | 'red' | 'green' | 'yellow' | 'blue' | 'magenta' | 'cyan' | 'white' | 'gray'`). I wonder which is the better design..

darkdarcool: I think it definitely makes it easier to read and write quickly. I'll make it support both 👍

cli/spinner.ts
```diff
@@ -4,9 +4,26 @@ const encoder = new TextEncoder();
...
 const LINE_CLEAR = encoder.encode("\r\u001b[K"); // From cli/prompt_secret.ts
 const COLOR_RESET = "\u001b[0m";
-const DEFAULT_SPEED = 75;
+const DEFAULT_INTERVAL = 75;
 const DEFAULT_SPINNER = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"];
 
+// This is a hack to allow us to use the same type for both the color name and an ANSI escape code.
+// deno-lint-ignore ban-types
+type Ansi = string & { };
+type Color = 'black' | 'red' | 'green' | 'yellow' | 'blue' | 'magenta' | 'cyan' | 'white' | 'gray' | Ansi;
```
darkdarcool, line 13: @iuioiua I used this typescript hack to allow for type suggestions on colors and allowing for a ANSI string, is this ok?

See https://github.com/microsoft/TypeScript/issues/29729#issuecomment-460346421

kt3k (approved):
LGTM. Nice addition!

cli/spinner.ts
```diff
@@ -0,0 +1,139 @@
...
+export interface SpinnerOptions {
+  /**
+   * The sequence of characters to be iterated through for animation.
+   *
+   * @default {["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏"]}
+   */
+  spinner?: string[];
+  /** The message to display next to the spinner. */
+  message?: string;
+  /** The time between each frame of the spinner.
```
jsejcksn, line 51: It would be helpful to include the unit of time here (milliseconds) — otherwise, the user must review the implementation to understand. For example:

```diff
- The time between each frame of the spinner.
+ The time (milliseconds) between each frame of the spinner.
```

PR #3970, opened 2023-12-16
fix(path/extname): Fix incorrect import

hugoalh:
`windowsExtname` should import from `./windows/extname.ts` but not `./posix/extname.ts`.


kt3k (approved):
LGTM. Thanks!

PR #3974, opened 2023-12-17
deprecation(semver): comparator min and max properties

timreichen:
deprecate `SemVerComparator` `min` and `max` properties.

These properties do not represent anything in a comparator but are pre-calculated values.
They get obstructive, if one creates a comparator as an object and not parsing a string:

before:
```ts
import { MAX } from "deno.land/std/semver/constants.ts"
const comparator = {
  operator: ">=",
  semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] },
  min: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] },
  max: MAX,
}
```
after:
```ts
const comparator = {
  operator: ">=",
  semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] } 
}
```


semver/types.ts
```diff
@@ -49,8 +49,14 @@ export type FormatStyle =
...
   operator: Operator;
   semver: SemVer;
-  min: SemVer;
-  max: SemVer;
+  /**
+   * @deprecated (will be removed in 0.215.0) use {@linkcode comparatorMin} instead.
+   */
+  min?: SemVer;
+  /**
+   * @deprecated (will be removed in 0.215.0) use {@linkcode comparatorMax} instead.
```
iuioiua, line 57:
```suggestion
   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMin} instead.
   */
  min?: SemVer;
  /**
   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMax} instead.
```

semver/types.ts
```diff
@@ -76,3 +82,6 @@ export interface SemVerRange {
   // The outer array is OR while each inner array is AND
   ranges: SemVerRangeOr;
 }
+
+export type Range = [SemVerComparator, SemVerComparator];
+export type RangeSet = Range[];
```
iuioiua, line 87: Is this change related to the main objective of this PR?

semver/parse_comparator.ts
```diff
@@ -20,12 +18,5 @@ export function parseComparator(comparator: string): SemVerComparator {
...
   const semver = m[2] ? parse(m[2]) : ANY;
-  const min = comparatorMin(semver, operator);
-  const max = comparatorMax(semver, operator);
-  return {
-    operator,
-    semver,
-    min,
-    max,
-  };
+  return { operator, semver };
```
iuioiua, line 21: This is a breaking change.

iuioiua (changes requested):
Can you please revert any changes that are either breaking or unrelated?

semver/types.ts
timreichen: sorry, this slipped in. reverted.

semver/parse_comparator.ts
timreichen: I reverted the breaking change.

semver/stringify_range.ts
iuioiua, line 1: Is this related to the PR?

timreichen: Sorry, this change also slipped in. Reverted

semver/types.ts
```diff
@@ -49,8 +49,14 @@ export type FormatStyle =
...
-  min: SemVer;
-  max: SemVer;
+  /**
+   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMin} instead.
+   */
+  min?: SemVer;
+  /**
+   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMax} instead.
+   */
+  max?: SemVer;
```
iuioiua, line 59:
```suggestion
  min: SemVer;
  /**
   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMax} instead.
   */
  max: SemVer;
```
Sorry, I just noticed these were made optional. I think that should be reverted.

iuioiua (changes requested):
Sorry, one last suggestion.

iuioiua (changes requested):
These changes seem to include the deprecation of `SemVerComparator`. Can you please ensure the PR only makes a single change?

iuioiua:
Also, I'm not certain of this change. Either way, let's see what others think.

timreichen:
> These changes seem to include the deprecation of `SemVerComparator`. Can you please ensure the PR only makes a single change?

But `SemVerComparator` was deprecated in https://github.com/denoland/deno_std/pull/3957/files#diff-8dc159a3b198ba82f12b082d1fcb076072d5ff8f10648744ae4746446d749022

kt3k (approved):
LGTM

semver/types.ts
```diff
@@ -49,7 +49,13 @@ export type FormatStyle =
 export interface Comparator {
   operator: Operator;
   semver: SemVer;
+  /**
+   * @deprecated (will be removed in 0.212.0) use {@linkcode comparatorMin} instead.
+   */
   min: SemVer;
+  /**
```
iuioiua, line 56: Actually, I don't think we need to include the deprecation in the new type.

iuioiua (changes requested):

kt3k (approved):

iuioiua (approved):

Issue #3967, opened 2023-12-15
SemVer range represents range-set


timreichen:
**Is your feature request related to a problem? Please describe.**

Ref: https://github.com/denoland/deno_std/pull/3957#issuecomment-1857116472

The semver mod was taken over from node-semver. range actually represents a [range-set](https://github.com/npm/node-semver/blob/main/range.bnf).
I opened an [issue](https://github.com/npm/node-semver/issues/669) on node-semver to find out the reason why.

We should differentiate between `range` and `range-set` to avoid confusion.

**Describe the solution you'd like**

rename range files and functions to `range_set` aka. `RangeSet` and deprecate all old functions.
In a next step, reintroduce old range functions with the implementation of a real range.

**Describe alternatives you've considered**

iuioiua:
Another suggested name would be `Range`. If we proceed with the suggested change, we should also promote the `ranges` property of `SemVerRange` as the value itself. It being nested looks unnecessary. I.e. functions should use `ranges` instead of the current `range.ranges`.

Another requirement we should question is whether the `ranges` property needs to be a nested array. In other words, is the outer array being used for OR comparisons and inner arrays being used for AND comparisons actually necessary? I expect that it is necessary, but I think it should be questioned. If it were found to not be necessary, we could just have a ranges type that is a flat array with a depth of 1 that could make things much simpler.

I hope my thoughts make sense. Happy to try clarify if they don't.

timreichen:
> Another suggested name would be `Range`. If we proceed with the suggested change, we should also promote the `ranges` property of `SemVerRange` as the value itself. It being nested looks unnecessary. I.e. functions should use `ranges` instead of the current `range.ranges`.

What do you mean exactly with "promoting the `ranges` property"?

> Another requirement we should question is whether the `ranges` property needs to be a nested array. In other words, is the outer array being used for OR comparisons and inner arrays being used for AND comparisons actually necessary? I expect that it is necessary, but I think it should be questioned. If it were found to not be necessary, we could just have a ranges type that is a flat array with a depth of 1 that could make things much simpler.

imo a `Range` should be defined as the following:
```ts
type Range = [Comparator, Comparator]
```
which actually represents a range between two versions, like a segment on a line:
Example: `>=1.2.3 - <2.0.0`.

A `RangeSet` represents multiple segments on the line and therefore needs to be an array of ranges:
```ts
type RangeSet = Range[] // [Comparator, Comparator][]
```
[...]

PR #3976, opened 2023-12-17
deprecation(semver): introduce RangeSet

timreichen:
Ref: https://github.com/denoland/deno_std/issues/3967

This is a draft which introduces `RangeSet` in order to deprecate the current `SemVerRange` implementation so a proper `Range` implementation can be done after the removal of the deprecated functions.


iuioiua:
Are we able to confirm that `RangeSet` truly needs to be a nested array? Please see https://github.com/denoland/deno_std/issues/3967#issuecomment-1859293834.

iuioiua:
@timreichen, a lot is going on in `std/semver`, and this PR changes a lot of the API. Can we close this for now and revisit it once things have settled down?

PR #3977, opened 2023-12-18
fix(http): account for period in signed cookie name

### LINKED ISSUE #3972: Signed cookies incorrect if value contains dots
Signed cookie verifying and parsing in `http/unstable_signed_cookie.ts` is implemented in such a way that it gets confused if the original unsigned value contains dots.

Not sure how common dots are in cookie values are. I certainly don't use dots and wasnt "bitten" by the bug, just randomly happened to notice this looking at the source.

So `signCookie("some.value", key)` results in `"some.value.signature"`. Parsing that with `parseSignedCookie`results in `"some"` instead of `"some.value"`. And ´verifyCookie' similarly reads `"some"` as the value and `"value"` as the signature. The actual signature that comes after, is discarded. It should be able to extract the signature based on the _last_ dot, to read correctly value as `"some.value"` and signature as `"signature`".

iuioiua:
PTAL @jtoppine

Fixes #3972

### DIFF
--- a/http/unstable_signed_cookie.ts
+++ b/http/unstable_signed_cookie.ts
@@ -3,6 +3,12 @@ import { decodeHex, encodeHex } from "../encoding/hex.ts";
 
 const encoder = new TextEncoder();
 
+function splitByLast(value: string, separator: string): [string, string] {
+  const index = value.lastIndexOf(separator);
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Nice!

PR #3978, opened 2023-12-18
feat(collections): improve mapValues typing

kt3k:
continued from #3430. Key types are inferred better from mapValues inputs.

Also allow to accept `Partial<Record<..>>`

### DIFF
--- a/collections/map_values.ts
+++ b/collections/map_values.ts
@@ -26,15 +26,51 @@
  * );
  * ```
  */
-export function mapValues<T, O>(
-  record: Readonly<Record<string, T>>,
+export function mapValues<T, O, K extends string>(
+  record: Readonly<Record<K, T>>,
   transformer: (value: T) => O,
-): Record<string, O> {
-  const ret: Record<string, O> = {};
+): Record<K, O>;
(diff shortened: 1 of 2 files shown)


collections/map_values.ts
```diff
@@ -26,15 +26,51 @@
...
  * ```
  */
-export function mapValues<T, O>(
-  record: Readonly<Record<string, T>>,
+export function mapValues<T, O, K extends string>(
+  record: Readonly<Record<K, T>>,
   transformer: (value: T) => O,
-): Record<string, O> {
-  const ret: Record<string, O> = {};
+): Record<K, O>;
```
iuioiua, line 32: Should this overload be deprecated?

kt3k: That makes this change a breaking change (`Partial<Record>` is not assignable to `Record`). Also that makes this util less convenient if the input is actually `Record` in my view (`record[key]` is `O` for `Record<K, O>` but, it's `O | undefined` for `Partial<Record<K, O>>`)

iuioiua (approved):
LGTM

PR #3979, opened 2023-12-18
refactor(semver): use string literals

timreichen:
Replaces  `" + variable + "` strings with string literals.


iuioiua (approved):
LGTM. This makes things more readable. Thank you.

PR #3980, opened 2023-12-18
chore(semver): use capturing groups

timreichen:
Replaces `replace` functions with match and uses groups for more readability.


timreichen:
closed in favour of https://github.com/denoland/deno_std/pull/3983

PR #3981, opened 2023-12-18
refactor(semver): rename isX to isWildcard

timreichen:
Renames `isX()` to `isWildcard()` for more clarity, because a wildcard can also be "" or "*".


kt3k (approved):
Makes sense! LGTM

iuioiua (approved):

PR #3982, opened 2023-12-18
buffer pull only once data has been written

crowlKats:
Currently, when a Buffer is read from, and no data was written, it will instantly close. This PR adds logic so that the buffer needs to be written to first before starting to read.


crowlKats:
i think this might not work

iuioiua:
Closing because #1985 is now closed in favor of #4548.

PR #3984, opened 2023-12-18
refactor(archive): format test names

timreichen:

### DIFF
--- a/archive/tar_test.ts
+++ b/archive/tar_test.ts
@@ -43,7 +43,7 @@ Deno.test("createTarArchive", async function () {
   assertEquals(wrote, 3072);
 });
 
-Deno.test("deflateTarArchive", async function () {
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
LGTM. Thank you.

iuioiua:
Note to self: use these names in #1985.

kt3k:
I wonder if the expression like `Tar()` makes sense as `Tar` is a class here

iuioiua:
What do you suggest?

kt3k:
If a test case is simple enough, I think we can use something like:

```js
Deno.test("Tar", (t) => {
  t.step("append() blah blah", () => {});
  t.step("getReader() blah blah", () => {});
});
```

kt3k:
But if a test case is complex and needs to involve multiple APIs (in other words, integration test type ones), maybe we need to allow more arbitrary format

PR #3985, opened 2023-12-18
refactor(async): format test names

timreichen:

### DIFF
--- a/async/pool_test.ts
+++ b/async/pool_test.ts
@@ -8,7 +8,7 @@ import {
   assertStringIncludes,
 } from "../assert/mod.ts";
 
-Deno.test("[async] pooledMap", async function () {
+Deno.test("pooledMap()", async function () {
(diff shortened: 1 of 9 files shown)


async/pool_test.ts
```diff
@@ -22,7 +22,7 @@ Deno.test("[async] pooledMap", async function () {
   assert(diff < 3000);
 });
 
-Deno.test("[async] pooledMap errors", async () => {
+Deno.test("pooledMap() checks errors", async () => {
```
iuioiua, line 25:
```suggestion
Deno.test("pooledMap() handles errors", async () => {
```

iuioiua (changes requested):
The word "checks" is being a few times here, and it doesn't make much sense to me. Are you able to redo these test names?

iuioiua (approved):
LGTM. Thanks!

PR #3990, opened 2023-12-18
refactor(csv): format test names

timreichen:


csv/csv_parse_stream_test.ts
```diff
@@ -57,97 +57,93 @@ Deno.test({
...
-        separator: ";",
-      },
-      {
-        name: "MultiLine",
-        input: `"two
+// These test cases were originally ported from Go:
+// https://github.com/golang/go/blob/go1.12.5/src/encoding/csv/
+// Copyright 2011 The Go Authors. All rights reserved. BSD license.
+// https://github.com/golang/go/blob/master/LICENSE
+const testCases = [
```
iuioiua, line 64: I don't mind this change but I think it should be in a separate PR. It's best to keep PRs as lean and focused as possible.

iuioiua (changes requested):
Mostly LGTM once my comment and use of "checks" is addressed.

csv/csv_parse_stream_test.ts
timreichen: Ok, I'll revert the change.

timreichen:
> Mostly LGTM once my comment and use of "checks" is addressed.

I updated the test names

iuioiua (approved):
LGTM! Thank you.

Only thing is that @kt3k stated previously that he'd prefer class instances in test names to omit their brackets. This PR is fine to me as it is now. Having the brackets to me, while camel cased, tells me that it's a class instance.

kt3k (approved):
LGTM

PR #3992, opened 2023-12-18
refactor(datetime): format test names

timreichen:

@iuioiua do we really need the `constants_test.ts` file to check constants?


iuioiua (approved):
LGTM. Thank you.

iuioiua:
> @iuioiua do we really need the `constants_test.ts` file to check constants?

In my eyes, no. But let's handle that separately.

PR #3997, opened 2023-12-18
refactor: add some missing explicit types

dsherret:
This will improve the deno doc output.

### DIFF
--- a/cli/parse_args.ts
+++ b/cli/parse_args.ts
@@ -18,7 +18,9 @@
  */
 import { assert } from "../assert/assert.ts";
 
-/** Combines recursively all intersection types and returns a new single type. */
+/** Combines recursively all intersection types and returns a new single type.
+ * @internal
+ */
 type Id<TRecord> = TRecord extends Record<string, unknown>
   ? TRecord extends infer InferredRecord
     ? { [Key in keyof InferredRecord]: Id<InferredRecord[Key]> }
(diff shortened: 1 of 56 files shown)


cli/parse_args.ts
```diff
@@ -18,7 +18,9 @@
  */
 import { assert } from "../assert/assert.ts";
 
-/** Combines recursively all intersection types and returns a new single type. */
+/** Combines recursively all intersection types and returns a new single type.
+ * @internal
```
dsherret, line 22: This is to suppress the "type is used in public type, but not exported" error in `deno doc --lint`

expect/fn.ts
```diff
@@ -4,7 +4,7 @@
 
 import { MOCK_SYMBOL, MockCall } from "./_mock_util.ts";
 
-export function fn(...stubs: Function[]) {
+export function fn(...stubs: Function[]): (...args: any[]) => any {
```
iuioiua, line 7:
```suggestion
export function fn(...stubs: Function[]): Function {
```

cli/parse_args.ts
iuioiua: Hm... I thought exporting such types was actually a good thing. In my previous PRs for #3764, I exported them.

dsherret: Depends if someone might need to depend on the type or not. If it's just some internal alias then it might not be useful to export. Perhaps I should revert the `@internal` changes and someone with the proper knowledge can go through to figure out what to export?

iuioiua: Yeah, that's a good idea. I can take care of those.

dsherret: Done.

iuioiua (approved):

iuioiua (approved):
LGTM. Thanks!

iuioiua:
Towards #3764 and #3878

PR #3998, opened 2023-12-19
chore(datetime): remove constants test file

timreichen:
removes datetime constants test file. It seems unnecessary to test constant values.


iuioiua (approved):
This change sounds reasonable to me, as we don't test for constants in other parts of the Standard Library.

PR #4000, opened 2023-12-20
deprecation(encoding): deprecate `base32.byteLength()`

### LINKED ISSUE #3975: proposal: deprecate `base32.byteLength()` in the public API from `std/encoding`
The other APIs within `std/encoding` do not expose such a function. There may be no need for this as part of the public API.

iuioiua:
Closes #3975


kt3k (approved):
LGTM

I think it's very rare that the user only want to know the length of encoded data without actually encoding it.

PR #4001, opened 2023-12-20
refactor(semver): cleanup `parseRange()`

timreichen:
restructures and simplifies `parseRange` by combining multiple operator parsings and reducing string manipulations.


semver/parse_range.ts
```diff
@@ -234,34 +63,130 @@ function isWildcard(id: string): boolean {
...
+
+          const majorIsWildcard = isWildcard(groups.major);
+          const minorIsWildcard = isWildcard(groups.minor);
+          const patchIsWildcard = isWildcard(groups.patch);
+
+          const major = +groups.major;
+          const minor = +groups.minor;
+          const patch = +groups.patch;
+
+          switch (groups.operator) {
```
iuioiua, line 89: This `switch` statement is quite meaty and difficult to follow. I liked that the previous approach factored out this logic into separate functions. Maybe we can factor out the logic of each case as before but still keep the other cleanups in this PR. WDYT?

timreichen: Yeah, I agree. I thought about that too and figured separate functions would span lots of duplicate code, especially the `majorIsWildcard` checks and number type castings, nut I think it looks good enough and updated the PR.
**Note** I replaced all string return values with a `parseComparator()` wrapper, so the return type of the functions, so that these functions don't return an array of strings which have to be parsed later. We can replace these wrappers with the actual Operator `{ operator: ..., semver: ...}` value in a later PR, so no parsing step is needed.

semver/parse_range.ts
```diff
@@ -220,49 +59,265 @@ function hyphenReplace(range: string) {
...
+
+/**
+ * Parses a range string into a SemVerRange object or throws a TypeError.
+ * @param rangeSet The range set string
+ * @returns A valid semantic groups range
+ */
+export function parseRange(rangeSet: string): SemVerRange {
+  const ranges = rangeSet
+    .split(/\s*\|\|\s*/)
+    .map((range) => parseHyphenRange(range).flatMap(parseRangeString));
```
kt3k, line 321:
```suggestion
export function parseRange(range: string): SemVerRange {
  const ranges = range
    .split(/\s*\|\|\s*/)
    .map((range) => parseHyphenRange(range).flatMap(parseRangeString));
```

iuioiua (approved):
Nit: In hindsight, maybe this PR could've been broken up to make it a little easier to digest. Maybe not... Either way, LGTM!

iuioiua (approved):
LGTM

semver/_shared.ts
```diff
@@ -127,26 +127,16 @@ const XRANGE_IDENTIFIER = `${NUMERIC_IDENTIFIER}|x|X|\\*`;
 export const XRANGE_PLAIN =
   `[v=\\s]*(?<major>${XRANGE_IDENTIFIER})(?:\\.(?<minor>${XRANGE_IDENTIFIER})(?:\\.(?<patch>${XRANGE_IDENTIFIER})(?:${PRERELEASE})?${BUILD}?)?)?`;
 
-export const XRANGE_REGEXP = new RegExp(
-  `^(?<operator>${COMPARATOR})\\s*${XRANGE_PLAIN}$`,
+// Meaning is "reasonably at or greater than" or "at least and backwards compatible with"
+export const OPERATOR_REGEXP = new RegExp(
```
kt3k, line 131: @timreichen Is this change of variable name intentional? (`XRANGE_REGEXP` -> `OPERATOR_REGEXP`)

timreichen: Yes, it is specifically an xrange with operators and needs to be from `XRANGE` without the operator. Would you prefer `OPERATOR_XRANGE_REGEXP`?

kt3k: Fair enough.

> Would you prefer OPERATOR_XRANGE_REGEXP?

That sounds clearer to me!

timreichen: https://github.com/denoland/deno_std/pull/4056

PR #4002, opened 2023-12-20
refactor(uuid): format test names and improve test coverage 

babiabeo:
Ref: #3713, #3754

This PR updates the test names as well as improving the test coverage in `uuid`.

See: https://app.codecov.io/gh/denoland/deno_std/pull/4002/tree/uuid


iuioiua (approved):
I removed the version prefixes from the test names, as we can tell which UUID version we're testing from the file name.

Either way, LGTM! Thank you for this.

PR #4004, opened 2023-12-21
deprecation(signal): move deprecation forward

iuioiua:
This change moves the removal version of `std/signal` from v1 to v0.212.0.

This is being done as there's little usage of the sub-module and most likely not to be too impacting.

This will help clean up technical debt sooner.


kt3k (approved):
LGTM

PR #4005, opened 2023-12-21
chore(fs): remove flaky `emptyDir()` tests

iuioiua:
This PR removes flaky `emptyDir()` and `emptyDirSync()` [tests](https://github.com/denoland/deno_std/actions/runs/7283505986/job/19847485371).

These tests test Deno's permissions system, which is a runtime concern and outside the scope of the Standard Library. These tests were written a while ago in https://github.com/denoland/deno/pull/3501 and the permission system is a lot more robust today. I think it's fine for these tests to be removed.

Note: test coverage is not affected.


kt3k:
What is the source of flakiness?

iuioiua:
Most likely, the heavy use of `Deno.Command()` in parallel.

kt3k:
> These tests test Deno's permissions system, which is a runtime concern and outside the scope of the Standard Library.

I disagree with this. This tests `emptyDir` requiring expected permissions. I don't think it's a good idea to remove these.

iuioiua:
I'll close this for now until we confirm the cause of the flakes.

iuioiua:
Related #4022

PR #4008, opened 2023-12-21
fix: `ensureFile()` and `ensureFileSync()` for tight permissions

### LINKED ISSUE #3339: Fixing racy ensureDirSync broke ensureFileSync for tight permissions
I just updated our used version of deno std and the changes made to `ensureDir` and `ensureDirSync` in https://github.com/denoland/deno_std/pull/3242 break our unit tests. While I am in favor of fixing the race condition, I feel like one side effect has not been considered that could be easily solved (so I think, see proposal below).

The changes introduced in https://github.com/denoland/deno_std/pull/3242 make it so that `ensureFile` and `ensureFileSync` no longer work with write permissions limited to specific files, they would require write permissions to their parent directories now due to their internal usage of `ensureDir` and `ensureDirSync`

This can be easily tested with this example code (first two tests, rest is my proposed fix):
```ts
import { ensureFileSync as ensureFileSyncWithOldEnsureDirSync } from "https://deno.land/std@0.178.0/fs/mod.ts";
import {
  ensureDirSync,
  ensureFileSync as ensureFileSyncWithNewEnsureDirSync,
} from "https://deno.land/std@0.179.0/fs/mod.ts";
import {
  getFileInfoType,
  toPathString,
} from "https://deno.land/std@0.179.0/fs/_util.ts";
import * as path from "https://deno.land/std@0.179.0/path/mod.ts";

const testDirectory = "./tests";
const testFilepath = testDirectory + "/allowed-file";

ensureDirSync(testDirectory);

Deno.test(
  {
    name: "Works with old ensureDirSync",
    permissions: {
      read: [testDirectory],
      write: [testFilepath],
    },
  },
  () => {
    try {
      Deno.removeSync(testFilepath);
    } catch (err) {
      if (!(err instanceof Deno.errors.NotFound)) throw err;
    }
    ensureFileSyncWithOldEnsureDirSync(testFilepath);
  },
);

Deno.test(
  {
    name: "Does not work with new ensureDirSync",
    permissions: {
      read: [testDirectory],
      write: [testFilepath],
    },
  },
  () => {
    try {
      Deno.removeSync(testFilepath);
    } catch (err) {
      if (!(err instanceof Deno.errors.NotFound)) throw err;
    }
    ensureFileSyncWithNewEnsureDirSync(testFilepath);
  },
);

Deno.test(
  {
    name: "Would work with proposed ensureDirSync (but is it racy?)",
    permissions: {
      read: [testDirectory],
      write: [testFilepath],
    },
  },
  () => {
    try {
      Deno.removeSync(testFilepath);
    } catch (err) {
      if (!(err instanceof Deno.errors.NotFound)) throw err;
    }
    ensureFileSyncWithProposedEnsureDirSync(testFilepath);
  },
);

Deno.test(
  {
    name: "Proposed ensureDirSync also still works for directories",
    permissions: {
      read: [testDirectory],
      write: [testDirectory],
    },
  },
  () => {
    try {
      Deno.removeSync(testDirectory, { recursive: true });
    } catch (err) {
      if (!(err instanceof Deno.errors.NotFound)) throw err;
    }
    ProposedEnsureDirSync(testDirectory);
  },
);

Deno.test(
  {
    name:
      "Proposed ensureDirSync also still works for directories - throws on missing permission",
    permissions: {
      read: [testDirectory],
      write: [testDirectory],
    },
  },
  () => {
    try {
      Deno.removeSync(testDirectory, { recursive: true });
    } catch (err) {
      if (!(err instanceof Deno.errors.NotFound)) throw err;
    }
    Deno.permissions.revokeSync({ name: "write", path: testDirectory });
    ProposedEnsureDirSync(testDirectory);
  },
);

function ProposedEnsureDirSync(dir: string | URL) {
  try {
    Deno.mkdirSync(dir, { recursive: true });
  } catch (mkdirErr) {
    if (
      !(mkdirErr instanceof Deno.errors.AlreadyExists ||
        mkdirErr instanceof Deno.errors.PermissionDenied)
    ) {
      throw mkdirErr;
    }

    try {
      const fileInfo = Deno.lstatSync(dir);
      if (!fileInfo.isDirectory) {
        throw new Error(
          `Ensure path exists, expected 'dir', got '${
            getFileInfoType(fileInfo)
          }'`,
        );
      }
    } catch (lstatErr) {
      if (
        mkdirErr instanceof Deno.errors.PermissionDenied &&
        lstatErr instanceof Deno.errors.NotFound
      ) {
        throw mkdirErr;
      } else {
        throw lstatErr;
      }
    }
  }
}

// ensureFileSync from 0.185.0, just using the ensureDirSync proposed here
function ensureFileSyncWithProposedEnsureDirSync(filePath: string | URL) {
  try {
    // if file exists
    const stat = Deno.lstatSync(filePath);
    if (!stat.isFile) {
      throw new Error(
        `Ensure path exists, expected 'file', got '${getFileInfoType(stat)}'`,
      );
    }
  } catch (err) {
    // if file not exists
    if (err instanceof Deno.errors.NotFound) {
      // ensure dir exists
      ProposedEnsureDirSync(path.dirname(toPathString(filePath)));
      // create file
      Deno.writeFileSync(filePath, new Uint8Array());
      return;
    }
    throw err;
  }
}
```

Output
```
$ deno test --allow-read --allow-write minimal.test.ts
running 5 tests from ./minimal.test.ts
Works with old ensureDirSync ... ok (13ms)
Does not work with new ensureDirSync ... FAILED (5ms)
Would work with proposed ensureDirSync (but is it racy?) ... ok (5ms)
Proposed ensureDirSync also still works for directories ... ok (5ms)
Proposed ensureDirSync also still works for directories - throws on missing permission ... FAILED (5ms)

 ERRORS 

Does not work with new ensureDirSync => ./minimal.test.ts:35:6
error: PermissionDenied: Requires write access to "./tests", run again with the --allow-write flag
    Deno.mkdirSync(dir, { recursive: true });
         ^
    at Object.mkdirSync (ext:deno_fs/30_fs.js:118:7)
    at ensureDirSync (https://deno.land/std@0.179.0/fs/ensure_dir.ts:49:10)
    at ensureFileSync (https://deno.land/std@0.179.0/fs/ensure_file.ts:72:7)
    at file:///home/egf0s02d/projects/minion/minimal.test.ts:49:5

Proposed ensureDirSync also still works for directories - throws on missing permission => ./minimal.test.ts:89:6
error: PermissionDenied: Requires write access to "./tests", run again with the --allow-write flag
    Deno.mkdirSync(dir, { recursive: true });
         ^
    at Object.mkdirSync (ext:deno_fs/30_fs.js:118:7)
    at ProposedEnsureDirSync (file:///home/egf0s02d/projects/minion/minimal.test.ts:111:10)
    at file:///home/egf0s02d/projects/minion/minimal.test.ts:105:5

 FAILURES 

Does not work with new ensureDirSync => ./minimal.test.ts:35:6
Proposed ensureDirSync also still works for directories - throws on missing permission => ./minimal.test.ts:89:6

FAILED | 3 passed | 2 failed (82ms)

error: Test failed
```

As you can see I already included a proposed fix, that is changing the implementation from this
https://github.com/denoland/deno_std/blob/aa57bf337bbf5e061752b92164b003a60e728951/fs/ensure_dir.ts#L47-L64

to this
```ts
export function ensureDirSync(dir: string | URL) {
  try {
    Deno.mkdirSync(dir, { recursive: true });
  } catch (mkdirErr) {
    if (
      !(mkdirErr instanceof Deno.errors.AlreadyExists ||
        mkdirErr instanceof Deno.errors.PermissionDenied)
    ) {
      throw mkdirErr;
    }

    try {
      const fileInfo = Deno.lstatSync(dir);
      if (!fileInfo.isDirectory) {
        throw new Error(
          `Ensure path exists, expected 'dir', got '${
            getFileInfoType(fileInfo)
          }'`,
        );
      }
    } catch (lstatErr) {
      if (
        mkdirErr instanceof Deno.errors.PermissionDenied &&
        lstatErr instanceof Deno.errors.NotFound
      ) {
        throw mkdirErr;
      } else {
        throw lstatErr;
      }
    }
  }
}
```

so if the directory creation fails due to missing write privileges, but it already exists, then we are fine, if it does not exist we just rethrow the original error.

This being a friday evening fix, there may be circumstances I have not considered yet, so I'll just leave that here to get some input.

- OS: Rocky Linux 8 + 9
- deno version: 1.32.5
- std version: 0.179.0+

egfx-notifications:
Implements the solution I proposed in #3339

Fixes #3339

### DIFF
--- a/fs/ensure_dir.ts
+++ b/fs/ensure_dir.ts
@@ -16,18 +16,32 @@ import { getFileInfoType } from "./_get_file_info_type.ts";
 export async function ensureDir(dir: string | URL) {
   try {
     await Deno.mkdir(dir, { recursive: true });
-  } catch (err) {
-    if (!(err instanceof Deno.errors.AlreadyExists)) {
-      throw err;
+  } catch (mkdirErr) {
+    if (
+      !(mkdirErr instanceof Deno.errors.AlreadyExists ||
+        mkdirErr instanceof Deno.errors.PermissionDenied)
+    ) {
+      throw mkdirErr;
     }
 
-    const fileInfo = await Deno.lstat(dir);
-    if (!fileInfo.isDirectory) {
-      throw new Error(
-        `Ensure path exists, expected 'dir', got '${
-          getFileInfoType(fileInfo)
-        }'`,
-      );
+    try {
+      const fileInfo = await Deno.lstat(dir);
+      if (!fileInfo.isDirectory) {
+        throw new Error(
+          `Ensure path exists, expected 'dir', got '${
+            getFileInfoType(fileInfo)
+          }'`,
+        );
+      }
+    } catch (lstatErr) {
+      if (
+        mkdirErr instanceof Deno.errors.PermissionDenied &&
+        lstatErr instanceof Deno.errors.NotFound
+      ) {
+        throw mkdirErr;
+      } else {
+        throw lstatErr;
+      }
     }
   }
 }
@@ -47,18 +61,32 @@ export async function ensureDir(dir: string | URL) {
 export function ensureDirSync(dir: string | URL) {
   try {
     Deno.mkdirSync(dir, { recursive: true });
-  } catch (err) {
-    if (!(err instanceof Deno.errors.AlreadyExists)) {
-      throw err;
+  } catch (mkdirErr) {
+    if (
+      !(mkdirErr instanceof Deno.errors.AlreadyExists ||
+        mkdirErr instanceof Deno.errors.PermissionDenied)
+    ) {
+      throw mkdirErr;
     }
 
-    const fileInfo = Deno.lstatSync(dir);
-    if (!fileInfo.isDirectory) {
-      throw new Error(
-        `Ensure path exists, expected 'dir', got '${
-          getFileInfoType(fileInfo)
-        }'`,
-      );
+    try {
+      const fileInfo = Deno.lstatSync(dir);
+      if (!fileInfo.isDirectory) {
+        throw new Error(
+          `Ensure path exists, expected 'dir', got '${
+            getFileInfoType(fileInfo)
+          }'`,
+        );
+      }
+    } catch (lstatErr) {
+      if (
+        mkdirErr instanceof Deno.errors.PermissionDenied &&
+        lstatErr instanceof Deno.errors.NotFound
+      ) {
+        throw mkdirErr;
+      } else {
+        throw lstatErr;
+      }
     }
   }
 }
--- a/fs/ensure_dir.ts
+++ b/fs/ensure_dir.ts
@@ -13,8 +13,18 @@ import { getFileInfoType } from "./_get_file_info_type.ts";
  * ensureDir("./bar"); // returns a promise
  * ```
  */
-export async function ensureDir(dir: string | URL) {
+export async function ensureDir(
+  dir: string | URL,
+  onlyPromptIfNotExists = true,
+) {
   try {
+    if (
+      onlyPromptIfNotExists &&
+      (await Deno.permissions.query({ name: "write", path: dir })).state ===
+        "prompt"
+    ) {
+      throw new Deno.errors.PermissionDenied();
+    }
     await Deno.mkdir(dir, { recursive: true });
   } catch (mkdirErr) {
     if (
@@ -38,6 +48,14 @@ export async function ensureDir(dir: string | URL) {
         mkdirErr instanceof Deno.errors.PermissionDenied &&
         lstatErr instanceof Deno.errors.NotFound
       ) {
+        if (
+          onlyPromptIfNotExists &&
+          (await Deno.permissions.query({ name: "write", path: dir })).state ===
+            "prompt"
+        ) {
+          await ensureDir(dir, false);
+          return;
+        }
         throw mkdirErr;
       } else {
         throw lstatErr;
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
LGTM. Thank you.

kt3k:
Thanks for the fix. One thing I wonder about this solution is that this still prompt the user write permission for the parent dir. I wonder if reverting #3242 can be a better solution. If we revert #3242, `ensureFile()` even doesn't ask write permission for the parent dir when it already exists.

egfx-notifications:
> I wonder if reverting #3242 can be a better solution. If we revert #3242, `ensureFile()` even doesn't ask write permission for the parent dir when it already exists.

Yes, but then we would also reintroduce the race condition fixed by #3242

> One thing I wonder about this solution is that this still prompt the user write permission for the parent dir. 

You are right, I did not see this as the deno test runner uses `--no-prompt` implicitly and in my production code I also use `--no-prompt` explicitly. I just pushed another commit to only prompt after verifying that we do not have the permissions yet and the target directory does not exist. The only thing that bothers me is that I had to augment the function signature to avoid infinite recursion. If you can think of a better way to do this, I'd be happy to see it.  My first approach with `Deno.permissions.request()` did not work, unfortunately, as it would prompt even if `--no-prompt` is specified, so I resorted to recursion and letting `mkdir()` do the prompting.

The current approach behaves like this:
- succeed if directory exists regardless of write permissions
- prompt for write permission if directory does not exist and prompting is allowed
- fail with `PermissionDenied` error from `mkdir()` if prompting is not allowed or permission was explicitly denied

@lucacasonato maybe you can have a look at this, too? Since the original fix #3242 concerns a bug that you reported

This being said, the `ensureFile()` and `ensureFileSync()` functions also do `lstat()` then `writeFile()`, so should probably be refactored as well.

kt3k:
@egfx-notifications 
Thanks for updating, but the introduction of a new option looks too much to me. I'd suggest we would land with the previous version. Could you revert the last commit?

kt3k:
I came up with another way to solve this issue. If we call `lstat` first, then `mkdir`, then catch only `AlreadyExists`, this solves both race condition and unnecessary write perm requirement without introducing new option. I'll try that in another branch.

iuioiua:
I'm closing this in favour of #4041. Thank you for this, either way, @egfx-notifications.

Issue #4010, opened 2023-12-21
std/expect: invalid return typing


lowlighter:
**Describe the bug**

When using `resolves` in the `std/expect`, some matchers return a wrong signature.

For instance, the following snippet says that `resolves.toHaveLength` returns `void` but it should be `Promise<void>`
```ts
import { expect } from "https://deno.land/std@0.209.0/expect/expect.ts";
Deno.test("foo", async () => {
  await expect(new Promise(solve => setTimeout(() => solve([]), 100))).resolves.toHaveLength(0)
})
```

<img src="https://github.com/denoland/deno_std/assets/22963968/e1d16f38-c6a8-4c39-a010-b0579f39c7d8" width="700">

Because of this, the linter wrongly assumes that `await` is not required (because it's not a promise)
<img src="https://github.com/denoland/deno_std/assets/22963968/097b7930-10a3-4e6e-ab2e-4c7629d72d04" width="700">

But without the await statement, there will be leaking ops (since the promise is not awaited)
```
foo ... FAILED (2ms)

 ERRORS

foo => ./test.ts:12:6
error: Leaking async ops:
  - 1 async operation to sleep for a duration was started in this test, but never completed. This is often caused by not cancelling a `setTimeout` or `setInterval` call.
To get more details where ops were leaked, run again with --trace-ops flag.
```

**Steps to Reproduce**

See above

**Expected behavior**

Matchers should always return `Promise<void>` when using `resolves` previously in the chain

**Environment**

- OS: -
- deno version: 1.39.0
- std version: 0.209.0

PR #4011, opened 2023-12-21
fix(expect): invalid return typing

babiabeo:
I noticed that using the `unknown` type in the `Fn` parameter would result in incompatibility with other types, which would also affect the `Async` type.

Example:
```ts
function callFunc(func: Fn) {
  func();
}

callFunc(
  expect("").resolves.toHaveLength
//                      ^
//  Argument of type '(expected: number) => void' is not assignable to parameter of type 'Fn'.
//    Types of parameters 'expected' and 'args' are incompatible.
//      Type 'unknown' is not assignable to type 'number'.
);
```

This PR changes `unknown` to `any` for compatibility.

Fix: #4010


iuioiua (approved):
LGTM. Nicely done!

PR #4012, opened 2023-12-21
fix(cli/spinner): export private type aliases used in public API

jsejcksn:
Ref: https://github.com/denoland/deno_std/pull/3968

I saw this new feature in the [release notes for 0.210.0](https://github.com/denoland/deno_std/releases/tag/0.210.0) and noticed that there's no way to access the `Color` type alias which is used by the public options property at `SpinnerOptions.color`. This exports that union and one of its members which was also still private.

https://github.com/denoland/deno_std/blob/5ddf32a1eaeb04ce55d2680e32cac19e4dd8a8ca/cli/spinner.ts#L41-L58

https://github.com/denoland/deno_std/blob/5ddf32a1eaeb04ce55d2680e32cac19e4dd8a8ca/cli/spinner.ts#L13-L27


iuioiua:
Towards #3764 and #3878.

iuioiua (approved):
Nice catch! Thank you.

PR #4018, opened 2023-12-21
chore: add more explicit types

dsherret:
This will further help with deno doc output.

### DIFF
--- a/bytes/last_index_of_needle.ts
+++ b/bytes/last_index_of_needle.ts
@@ -20,7 +20,7 @@
 export function lastIndexOfNeedle(
   source: Uint8Array,
   needle: Uint8Array,
-  start = source.length - 1,
+  start: number = source.length - 1,
 ): number {
   if (start < 0) {
     return -1;
--- a/io/slice_long_to_bytes.ts
+++ b/io/slice_long_to_bytes.ts
@@ -10,7 +10,7 @@
  */
 export function sliceLongToBytes(
   d: number,
-  dest = Array.from<number>({ length: 8 }),
+  dest: number[] = Array.from<number>({ length: 8 }),
 ): number[] {
   let big = BigInt(d);
   for (let i = 0; i < 8; i++) {
--- a/log/mod.ts
+++ b/log/mod.ts
@@ -431,13 +431,7 @@ const state = {
  * For examples check source code of {@linkcode FileHandler}`
  * and {@linkcode TestHandler}.
  */
-export const handlers: {
-  BaseHandler: typeof BaseHandler;
-  ConsoleHandler: typeof ConsoleHandler;
-  WriterHandler: typeof WriterHandler;
-  FileHandler: typeof FileHandler;
-  RotatingFileHandler: typeof RotatingFileHandler;
-} = {
+export const handlers = {
   BaseHandler,
   ConsoleHandler,
   WriterHandler,
--- a/log/mod.ts
+++ b/log/mod.ts
@@ -357,7 +357,7 @@
  */
 
 import { Logger } from "./logger.ts";
-import type { GenericFunction } from "./logger.ts";
+import type { GenericFunction, LogRecord } from "./logger.ts";
 import {
   BaseHandler,
   ConsoleHandler,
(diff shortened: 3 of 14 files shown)


yaml/schema/core.ts
```diff
@@ -9,6 +9,6 @@ import { json } from "./json.ts";
 
 // Standard YAML's Core schema.
 // http://www.yaml.org/spec/1.2/spec.html#id2804923
-export const core = new Schema({
+export const core: Schema = new Schema({
```
dsherret, line 12: Can't be easily inferred because `Schema` could have a type parameter.

log/mod.ts
```diff
@@ -431,15 +431,23 @@ const state = {
  * For examples check source code of {@linkcode FileHandler}`
  * and {@linkcode TestHandler}.
  */
-export const handlers = {
+export const handlers: {
+  BaseHandler: typeof BaseHandler;
+  ConsoleHandler: typeof ConsoleHandler;
+  WriterHandler: typeof WriterHandler;
+  FileHandler: typeof FileHandler;
+  RotatingFileHandler: typeof RotatingFileHandler;
```
dsherret, line 439: Slightly unfortunate, but this is not too bad.

bytes/last_index_of_needle.ts
```diff
@@ -20,7 +20,7 @@
 export function lastIndexOfNeedle(
   source: Uint8Array,
   needle: Uint8Array,
-  start = source.length - 1,
+  start: number = source.length - 1,
```
iuioiua, line 23: Is this required? `deno doc --lint bytes/last_index_of_needle.ts` reports no errors before this change.

io/slice_long_to_bytes.ts
```diff
@@ -10,7 +10,7 @@
  */
 export function sliceLongToBytes(
   d: number,
-  dest = Array.from<number>({ length: 8 }),
+  dest: number[] = Array.from<number>({ length: 8 }),
```
iuioiua, line 13: Ditto. Is this needed?

iuioiua (changes requested):
Explicit types for arguments with default values don't appear to be needed.

bytes/last_index_of_needle.ts
dsherret: I think there's a bug in `deno doc --lint` related to this. I'm running a separate analyzer (it is an optimization to type checking).

See here how the type is missing: https://deno.land/std@0.210.0/bytes/last_index_of_needle.ts?s=lastIndexOfNeedle

io/slice_long_to_bytes.ts
dsherret: See here how `dest` is unsupported. It's a bug in `deno doc --lint`:

https://deno.land/std@0.210.0/io/slice_long_to_bytes.ts?s=sliceLongToBytes

bytes/last_index_of_needle.ts
iuioiua: Oh, I see. Perhaps we should fix that bug instead of accounting for it. WDYT?

dsherret: I opened https://github.com/denoland/deno_doc/issues/435

Doing a big refactor of everything at the moment (which is what helped spot this), so I'll add a test and investigate in there later.

iuioiua: Perhaps let's continue to include these explicit types on arguments with default values. It'll fix the docs sooner. We can then come behind later on and clean these up. I can create an issue once this PR is merged.

log/mod.ts
```diff
@@ -431,15 +431,23 @@ const state = {
  * For examples check source code of {@linkcode FileHandler}`
  * and {@linkcode TestHandler}.
  */
-export const handlers = {
+export const handlers: {
```
dsherret, line 434: Actually, going to revert this. Deno doc understands it so I should be able to get the other thing to as well.

bytes/last_index_of_needle.ts
dsherret: I think this is all the public API that's affected. Everything else seems to be good. After this PR I have Deno type checking a subset of the code properly in the new registry.

iuioiua:
Towards #3764 and #3878.

iuioiua (approved):
LGTM. Thank you.

PR #4020, opened 2023-12-23
test(jsonc): do not use deno.lock in subcommand

### LINKED ISSUE #4015: bug: `[jsonc] use Object.defineProperty when setting object property` test is flaky
I've noticed this test flake a few times. I'm unfamiliar with the test or `Object.prototype.__proto__`, but perhaps this test can run the code directly instead of using `Deno.Command()`.

Source code: https://github.com/denoland/deno_std/blob/5ddf32a1eaeb04ce55d2680e32cac19e4dd8a8ca/jsonc/parse_test.ts#L176-L202

See https://github.com/denoland/deno_std/actions/runs/7292878535/job/19874783040

kt3k:
This PR tries to fix flaky test failure of `[jsonc] use Object.defineProperty when setting object property` case.

The test case sometimes fails with the below error message:

```
------- output -------
error: Unable to parse contents of lockfile. /home/runner/work/deno_std/deno_std/deno.lock
./jsonc/parse_test.ts => [jsonc] use Object.defineProperty when setting object property ...----- output end -----
./jsonc/parse_test.ts => [jsonc] use Object.defineProperty when setting object property ... FAILED (7ms)
```
https://github.com/denoland/deno_std/actions/runs/7292878535/job/19874783040

The command doesn't need to use lock file. So we can avoid the above flaky failure by passing `--no-lock` flag.

closes #4015

### DIFF
--- a/jsonc/parse_test.ts
+++ b/jsonc/parse_test.ts
@@ -194,7 +194,7 @@ Deno.test({
     const command = new Deno.Command(Deno.execPath(), {
       stdout: "inherit",
       stderr: "inherit",
-      args: ["eval", testCode],
+      args: ["eval", "--no-lock", testCode],
     });
     const { success } = await command.output();
     assert(success);


iuioiua (approved):
LGTM

kt3k:
My interpretation of the above error message is that some other deno process is in the middle of updating `deno.lock` when the above `Deno.Command` is invoked (which is possible because we use `--parallel` flag and many deno processes are running simultaneously). `deno.lock` is being updated in very small amount of time during deno process runs. The issue happens only when the period of updating `deno.lock` overlaps with the time of invoking the above `Deno.Command`

PR #4021, opened 2023-12-24
chore(log): rework `std/log` to work without deprecated `Deno.Writer`

syhol:
Attempting to solve https://github.com/denoland/deno_std/issues/3907

> std/log is the last of the sub-modules that uses the deprecated Writer (i.e. Deno.Writer) interface. It should use WritableStream instead.

Previous attempt of https://github.com/denoland/deno_std/pull/3969 was trying to use `std/streams/buffer.ts` and the streams API, but moving to async flushes resulted in potentially lost logs.

Now just solving with a `Uint8Array` and a `ArrayBuffer` and a number pointer.

### DIFF
--- a/log/handlers.ts
+++ b/log/handlers.ts
@@ -183,18 +183,22 @@ export class FileHandler extends WriterHandler {
       this.flush();
     }
     this._buf.set(bytes, this._pointer);
-    this._pointer += bytes.length;
+    this._pointer += bytes.byteLength;
   }
 
   flush() {
-    if (this._buf?.byteLength > 0 && this._file) {
-      this._file?.writeSync(this._buf.subarray(0, this._pointer));
+    if (this._pointer > 0 && this._file) {
+      let written = 0;
+      while (written < this._pointer) {
+        written += this._file.writeSync(
+          this._buf.subarray(written, this._pointer),
+        );
+      }
       this.resetBuffer();
     }
   }
 
   resetBuffer() {
-    this._buf.fill(0);
     this._pointer = 0;
   }
 
--- a/log/handlers.ts
+++ b/log/handlers.ts
@@ -163,7 +163,7 @@ export class FileHandler extends WriterHandler {
 
   override setup() {
     this._file = Deno.openSync(this._filename, this._openOptions);
-    this.resetBuffer();
+    this.#resetBuffer();
 
     addEventListener("unload", this.#unloadCallback);
   }
@@ -194,11 +194,11 @@ export class FileHandler extends WriterHandler {
           this._buf.subarray(written, this._pointer),
         );
       }
-      this.resetBuffer();
+      this.#resetBuffer();
     }
   }
 
-  resetBuffer() {
+  #resetBuffer() {
     this._pointer = 0;
   }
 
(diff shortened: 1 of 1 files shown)


log/handlers.ts
```diff
@@ -180,18 +178,30 @@ export class FileHandler extends WriterHandler {
...
+    this._pointer += bytes.byteLength;
   }
 
   flush() {
-    if (this._buf?.buffered() > 0) {
-      this._buf.flush();
+    if (this._pointer > 0 && this._file) {
+      let written = 0;
+      while (written < this._pointer) {
+        written += this._file.writeSync(
```
iuioiua, line 193: This is using the `WriterSync` interface, which is part of the deprecations in question. Are we able to use `file.writable` instead? Though, doing so would make `flush()` async.

iuioiua: > Are you sure? I can see that [std/io/types.d.ts WriterSync](https://deno.land/std@0.210.0/io/types.d.ts?s=WriterSync) is deprecated but neither [Deno.FsFile](https://deno.land/api@v1.39.1?s=Deno.FsFile) nor [Deno.FsFile.prototype.writeSync](https://deno.land/api@v1.39.1?s=Deno.FsFile&p=prototype.writeSync) appear to be. Do the docs need to be updated to mark `Deno.FsFile.prototype.writeSync` as deprecated?

Yes, I'm sure. I've opened https://github.com/denoland/deno/pull/21701.

We'll get back to you on the other points.

iuioiua: Actually, there are no plans to deprecate these `Deno.FsFile` methods. Let's continue with this PR 👍🏾

log/handlers.ts
```diff
@@ -139,7 +137,8 @@ interface FileHandlerOptions extends HandlerOptions {
  */
 export class FileHandler extends WriterHandler {
   protected _file: Deno.FsFile | undefined;
-  protected _buf!: BufWriterSync;
+  protected _buf: Uint8Array = new Uint8Array(new ArrayBuffer(PAGE_SIZE));
```
iuioiua, line 140:
```suggestion
  protected _buf: Uint8Array = new Uint8Array(PAGE_SIZE);
```

log/handlers.ts
```diff
@@ -180,18 +178,30 @@ export class FileHandler extends WriterHandler {
...
+      while (written < this._pointer) {
+        written += this._file.writeSync(
+          this._buf.subarray(written, this._pointer),
+        );
+      }
+      this.resetBuffer();
     }
   }
 
+  resetBuffer() {
```
iuioiua, line 201: Let's make this a private method. I don't think there's a need to introduce another public method to the API at this point.

iuioiua (changes requested):

syhol:
Sure, TS private keyword or JS hash prefix?

iuioiua:
Hash prefix please 🙂

log/handlers.ts
syhol: Done

syhol:
All suggestions applied 🙏  ready for another review 🎉

iuioiua (approved):
Great work! Thank you for this.

dvv:
Hi!

With this attempt to log circa 4096 octets message just dumps the process.

PR #4024, opened 2023-12-27
feat(semver): optional SemVer build and prerelease properties

timreichen:
makes `SemVer` `prerelease` and `build` properties optional to reduce empty array boilerplate.


semver/increment.ts
```diff
@@ -107,7 +107,7 @@ export function increment(
     // If the input is a non-prerelease version, this acts the same as
     // prepatch.
     case "prerelease":
-      if (version.prerelease.length === 0) {
+      if (version.prerelease?.length === 0) {
```
kt3k, line 110: When `prerelease` is `undefined`, that should be considered as the same as empty array. So this line seems introducing a bug. (this passes with prerelease = `[]`, but doesn't pass with prerelease = `undefined`)

Can you add test cases to `increment_test.ts` to exercise these patterns?

kt3k: Probably we need to translate `version.prerelease.length === 0` to `version.prerelease === undefiend || version.prerelease.length === 0`

semver/is_semver_test.ts
```diff
@@ -47,6 +45,8 @@ Deno.test({
   fn: async (t) => {
     let i = 0;
     const versions: [unknown][] = [
+      [{ major: 0, minor: 0, patch: 0, prerelease: [] }],
+      [{ major: 0, minor: 0, patch: 0, build: [] }],
```
kt3k, line 49: We also need to add test cases to check with both prerelease and build undefined

semver/increment.ts
```diff
@@ -134,7 +134,7 @@ export function increment(
       if (
         version.minor !== 0 ||
         version.patch !== 0 ||
-        version.prerelease.length === 0
+        version.prerelease?.length === 0
```
kt3k, line 137: ditto. This seems introducing a bug

semver/increment.ts
```diff
@@ -161,7 +161,7 @@ export function increment(
       // 1.2.1 bumps to 1.3.0
       if (
         version.patch !== 0 ||
-        version.prerelease.length === 0
+        version.prerelease?.length === 0
```
kt3k, line 164: Ditto. This seems introducing a bug. Please add test case to exercise this pattern with prerelease = `undefined`.

semver/increment.ts
```diff
@@ -186,7 +186,7 @@ export function increment(
       // If it is a pre-release it will bump up to the same patch version.
       // 1.2.0-5 patches to 1.2.0
       // 1.2.0 patches to 1.2.1
-      if (version.prerelease.length === 0) {
+      if (version.prerelease?.length === 0) {
```
kt3k, line 189: Ditto. This seems introducing a bug. Please add test case to exercise this pattern with prerelease = `undefined`.

semver/is_semver_test.ts
timreichen: done

semver/increment.ts
timreichen: done

semver/increment.ts
timreichen: fixed

semver/increment.ts
timreichen: fixed

semver/increment.ts
timreichen: done

kt3k (approved):
LGTM

PR #4025, opened 2023-12-27
deprecation(semver): deprecate cmp

timreichen:
deprecates `cmp`, because one can achieve the same with `compare` and operators or `eq`, `neq` etc. `cmp` even uses them internally.


iuioiua (approved):
This deprecation makes sense to me. Thank you.

kt3k (approved):
LGTM

PR #4026, opened 2023-12-27
refactor(semver): use capturing groups in `parseComparator()`

timreichen:
enhances `parseComparator` by using capturing groups instead of parsing string a second time.


semver/parse_comparator.ts
```diff
@@ -1,26 +1,51 @@
...
+    major,
+    minor,
+    patch,
+    prerelease,
+    buildmetadata,
+  } = groups as REGEXP_GROUPS;
+
+  const semver = major
+    ? {
+      major: parseInt(major),
```
kt3k, line 39: Now the major version is not checked against `Number.MAX_SAFE_INTEGER`, which was checked in `parse`

timreichen: I've fixed it with a centralized `parseNumber` function that parses and checks the number.

kt3k (approved):
LGTM

PR #4031, opened 2023-12-27
deprecation(semver): deprecate parse SemVer type

timreichen:
deprecates `SemVer` as type for `parse`.


iuioiua (commented):
This PR appears to have been created by mistake.

The `parse(version: SemVer)` overload is already deprecated (see above the modified line). Plus, the modified line deprecates the overload with a `string` argument, not a `SemVer` argument.

PR #4034, opened 2023-12-28
fix(expect): align toEqual to jest

### LINKED ISSUE #3947: Expect toEqual: undefined properties are required in the expected value
**Describe the bug**

When using `expect().toEqual`, undefined properties in the test value cause a failure when those properties don't exist in the expect value.

**Steps to Reproduce**

```typescript
import { expect as denoExpect } from "https://deno.land/std/expect/mod.ts";
import { expect as npmExpect } from "npm:expect";

// Passes
Deno.test("npm expect().toEqual ignores undefined values", () => {
  npmExpect({
    prop1: "string",
    prop2: undefined,
  }).toEqual({
    prop1: "string",
  });
});

// Fails
Deno.test("deno expect().toEqual ignores undefined values", () => {
  denoExpect({
    prop1: "string",
    prop2: undefined,
  }).toEqual({
    prop1: "string",
  });
});
```

**Expected behavior**

I expect `deno expect` to behave the same as `npm expect` and the test above should have passed. If there are undefined properties in the test value, those properties shouldn't need to exist in the expect value.

**Environment**

- OS: MacOS 13.6.3 ARM
- deno version: 1.38.5
- std version: latest main and stable (0.208.0)

kt3k:
This PR updates `expect().toEqual()` behavior. Now the difference of `undefined`-valued keys are ignored in objects.

closes #3947

### DIFF
--- a/expect/_matchers.ts
+++ b/expect/_matchers.ts
@@ -26,14 +26,37 @@ export function toBe(context: MatcherContext, expect: unknown): MatchResult {
   }
 }
 
+// deno-lint-ignore no-explicit-any
+function filterUndefined(obj: any): any {
+  if (typeof obj !== "object") return obj;
+  if (obj instanceof Date) return obj;
+  if (obj instanceof RegExp) return obj;
+  if (ArrayBuffer.isView(obj)) return obj;
+  if (obj === null) return obj;
+  if ("size" in obj) return obj;
+  if (Symbol.iterator in obj) return obj;
+  if (Array.isArray(obj)) return obj.map(filterUndefined);
+
+  // deno-lint-ignore no-explicit-any
+  const result = {} as any;
+  for (const key in obj) {
+    const val = obj[key];
+    if (val === undefined) continue;
+    result[key] = filterUndefined(val);
+  }
+  return result;
+}
+
 export function toEqual(
   context: MatcherContext,
   expected: unknown,
 ): MatchResult {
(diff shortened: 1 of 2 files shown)


expect/_matchers.ts
iuioiua, line 1: What about using https://deno.land/std/collections/mod.ts?s=filterValues ?

kt3k: `filterValues` only filters one level (immediate child values of the root object), but this needs to filter recursively.

expect/_matchers.ts
```diff
@@ -26,14 +26,37 @@ export function toBe(context: MatcherContext, expect: unknown): MatchResult {
...
+// deno-lint-ignore no-explicit-any
+function filterUndefined(obj: any): any {
+  if (typeof obj !== "object") return obj;
+  if (obj instanceof Date) return obj;
+  if (obj instanceof RegExp) return obj;
+  if (ArrayBuffer.isView(obj)) return obj;
+  if (obj === null) return obj;
+  if ("size" in obj) return obj;
+  if (Symbol.iterator in obj) return obj;
+  if (Array.isArray(obj)) return obj.map(filterUndefined);
```
iuioiua, line 38: Nit: Perhaps this logic could be factored out into its own `isPlainObject()` just so its purpose is made clear.

iuioiua (approved):

PR #4035, opened 2023-12-28
fix(expect): align toStrictEqual behavior to jest

kt3k:
This PR changes the behavior of `expect().toStrictEqual()` to make it aligned to jest's. This has been pointed in https://github.com/denoland/deno_std/issues/3947#issuecomment-1870858120


iuioiua (approved):
Nice

PR #4007, opened 2023-12-21
refactor(fs): trim internal utilities

iuioiua:
Closes #3915 as `std/fs` is the only sub-module with these "extraneous imports within the internal utilities" problem.

Before:
```
asher@Ashers-MacBook-Air deno_std % deno info fs/ensure_dir.ts
local: /Users/asher/GitHub/deno_std/fs/ensure_dir.ts
emit: /Users/asher/Library/Caches/deno/gen/file/Users/asher/GitHub/deno_std/fs/ensure_dir.ts.js
type: TypeScript
dependencies: 24 unique
size: 30.91KB

file:///Users/asher/GitHub/deno_std/fs/ensure_dir.ts (1.61KB)
└─┬ file:///Users/asher/GitHub/deno_std/fs/_util.ts (2.74KB)
  ├─┬ file:///Users/asher/GitHub/deno_std/path/resolve.ts (528B)
  │ ├── file:///Users/asher/GitHub/deno_std/path/_os.ts (705B)
  │ ├─┬ file:///Users/asher/GitHub/deno_std/path/posix/resolve.ts (1.55KB)
  │ │ ├─┬ file:///Users/asher/GitHub/deno_std/path/_common/normalize_string.ts (2.25KB)
  │ │ │ └── file:///Users/asher/GitHub/deno_std/path/_common/constants.ts (1.97KB)
  │ │ ├── file:///Users/asher/GitHub/deno_std/path/_common/assert_path.ts (306B)
  │ │ └─┬ file:///Users/asher/GitHub/deno_std/path/posix/_util.ts (391B)
  │ │   └── file:///Users/asher/GitHub/deno_std/path/_common/constants.ts *
  │ └─┬ file:///Users/asher/GitHub/deno_std/path/windows/resolve.ts (4.73KB)
  │   ├── file:///Users/asher/GitHub/deno_std/path/_common/constants.ts *
  │   ├── file:///Users/asher/GitHub/deno_std/path/_common/normalize_string.ts *
  │   ├── file:///Users/asher/GitHub/deno_std/path/_common/assert_path.ts *
  │   └─┬ file:///Users/asher/GitHub/deno_std/path/windows/_util.ts (828B)
  │     └── file:///Users/asher/GitHub/deno_std/path/_common/constants.ts *
  ├─┬ file:///Users/asher/GitHub/deno_std/path/separator.ts (273B)
  │ └── file:///Users/asher/GitHub/deno_std/path/_os.ts *
[...]

PR #4036, opened 2023-12-28
refactor(semver): move `isOperator` to separate file

timreichen:


iuioiua (changes requested):
Perhaps, this should be `_shared_test.ts`.

timreichen:
> Perhaps, this should be `_shared_test.ts`.

Maybe we can move `isOperator` to a separate file?

iuioiua:
SGTM. Maybe we should also review other internal dependencies in the same way for `std/semver`. See #3916 and #4007 for an idea of what I mean.

timreichen:
> SGTM. Maybe we should also review other internal dependencies in the same way for `std/semver`. See #3916 and #4007 for an idea of what I mean.

Agreed. There is definitely room for improvement. I'll take a look.

iuioiua (changes requested):
We should prefix these files with an underscore as `isOperator()` is part of the private API. Right?

timreichen:
> We should prefix these files with an underscore as `isOperator()` is part of the private API. Right?

done

iuioiua:
Hey Tim, upon reviewing this PR, I realised that the whole `isOperator()` doesn't seem needed. I'll close this in favour of #4038. Sorry, I should've realised this earlier. Either way, thanks for the PR. Please feel free to review the other PR.

PR #4037, opened 2023-12-28
deprecation(semver): deprecate `testComparator()`

timreichen:
deprecates `testComparator` in favour of using `compare`.

### DIFF
--- a/semver/test_comparator.ts
+++ b/semver/test_comparator.ts
@@ -7,6 +7,7 @@ import { cmp } from "./cmp.ts";
  * @param version The version to compare
  * @param comparator The comparator
  * @returns True if the version is within the comparators set otherwise false
+ * @deprecated (will be removed in 0.212.0) Use {@linkcode compare} instead.
  */
 export function testComparator(
   version: SemVer,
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

I now wonder if `Comparator` interface is useful/meaningful to the user at all

semver/test_comparator.ts
```diff
@@ -7,6 +7,7 @@ import { cmp } from "./cmp.ts";
  * @param version The version to compare
  * @param comparator The comparator
  * @returns True if the version is within the comparators set otherwise false
+ * @deprecated (will be removed in 0.212.0) Use {@linkcode compare} instead.
```
iuioiua, line 10: Nit: 0.212.0 is quite soon. It would mean the lifetime of the deprecation would be one minor version. I think we should at least push out to 0.213.0.

timreichen: done

iuioiua:
> I now wonder if `Comparator` interface is useful/meaningful to the user at all

Let's explore that idea if we haven't already. CC @timreichen

iuioiua (approved):
Thank you, Tim!

Issue #937, opened 2021-05-21
todo: enable `NoUncheckedIndexedAccess` TS compiler option


AZMCode:
I'm aware this is a controversial topic, however I would find it useful if Deno's std library was compatible with Typescript's new noUncheckedIndexedAccess flag, since I find my projects that use it severely handicapped. I believe dropping some assertions throughout the code (after patching assert itself), or using less indexes should be enough.

kitsonk:
My personal opinion is that I don't mind having `noUncheckedIndexedAccess` "clean" code for `std`, I just wouldn't want to make it the default for the Deno CLI.

iuioiua:
I think we should add this. PRs are welcome!

iuioiua:
Closing this in favour of #4040. Thank you for the input, guys!

PR #4039, opened 2023-12-30
chore: enable `noUncheckedIndexedAccess` TS compiler option

syhol:
Implementing as requested in the issue https://github.com/denoland/deno_std/issues/937

> I'm aware this is a controversial topic, however I would find it useful if Deno's std library was compatible with Typescript's new noUncheckedIndexedAccess flag, since I find my projects that use it severely handicapped [...]

### DIFF
--- a/path/_common/glob_to_reg_exp.ts
+++ b/path/_common/glob_to_reg_exp.ts
@@ -65,7 +65,7 @@ export function _globToRegExp(
 
   // Remove trailing separators.
   let newLength = glob.length;
-  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]); newLength--);
+  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]!); newLength--);
   glob = glob.slice(0, newLength);
 
   let regExpString = "";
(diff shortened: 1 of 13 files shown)


semver/parse.ts
```diff
@@ -41,19 +41,19 @@ export function parse(version: string | SemVer): SemVer {
   if (!groups) throw new TypeError(`Invalid Version: ${version}`);
 
   // these are actually numbers
-  const major = parseInt(groups.major);
-  const minor = parseInt(groups.minor);
-  const patch = parseInt(groups.patch);
+  const major = parseInt(groups.major!);
+  const minor = parseInt(groups.minor!);
+  const patch = parseInt(groups.patch!);
```
syhol, line 46: Ignoring the lint error with `!` as `parseInt(undefined)` will parse to `NaN` and I've added a new isNaN check below.

path/posix/resolve.ts
```diff
@@ -16,8 +16,9 @@ export function resolve(...pathSegments: string[]): string {
   for (let i = pathSegments.length - 1; i >= -1 && !resolvedAbsolute; i--) {
     let path: string;
 
-    if (i >= 0) path = pathSegments[i];
-    else {
+    if (i >= 0) {
+      path = pathSegments[i]!;
```
syhol, line 20: `i` starts at `pathSegments.length` and decrements down to -1, however this statement is protected with `i >= 0` so the `i` should always be inbounds.

path/_common/glob_to_reg_exp.ts
```diff
@@ -65,7 +65,7 @@ export function _globToRegExp(
 
   // Remove trailing separators.
   let newLength = glob.length;
-  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]); newLength--);
+  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]!); newLength--);
   glob = glob.slice(0, newLength);
```
syhol, line 69: It looks like there might be a bug here, the semicolon at the end of the for statement means line below is only ever run once. The for loop itself does nothing.

iuioiua:
Great to see this being worked on! Implementing this feature will require a bit of work across the entire codebase. We should create a tracking issue for this to break PRs up. Perhaps, this PR can focus on the sub-modules it already touches. WDYT?

syhol:
Never mind, there are many more issues than I first thought. Will spend some more time on this before I submit a PR.

iuioiua:
To elaborate, I think the tracking issue should list each sub-module as a to-do list and include instructions on how to work on the issue.

Issue #4040, opened 2023-12-30
tracking: enable `noUncheckedIndexedAccess`


iuioiua:
## Background

The [`noUncheckedIndexedAccess` TSConfig option](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess) makes access objects safer and improves object inference. Enabling this option will improve the overall code quality of the Standard Library. This tracking issue aims to shine a new light on the functionality first brought up in #937.

## Objectives

Modify the sub-modules of the Standard Library such that the TypeScript checker has no problems. The list of applicable sub-modules includes:
- [x] _tools
- [x] archive
- [x] assert
- [x] async
- [x] bytes
- [x] cli
- [x] collections
- [x] console
- [x] crypto
- [x] csv
- [x] data_structures
- [x] dotenv
- [x] encoding
- [x] expect
- [x] fmt
- [x] front_matter
- [x] fs
- [x] html
- [x] http
- [x] ini
- [x] io
- [x] json
- [x] jsonc
- [x] log
- [x] media_types
- [x] path
- [x] semver
- [x] streams
- [x] testing
- [x] text
- [x] toml
- [x] ulid
- [x] uuid
- [x] url
- [x] webgpu
- [x] yaml

## How to contribute
This assumes that the [Contributing Guidelines](https://github.com/denoland/deno_std/blob/7a0294a7f434919c67a807187dcea7bd5fe4947c/.github/CONTRIBUTING.md) are being followed. Also, please read [this](https://github.com/denoland/deno_std/pull/4275#pullrequestreview-1861556153) before proceeding with one of these PRs.
1. Choose a sub-module or small set of sub-modules you'd like to work on.
2. Temporarily enable the option in the [Deno configuration file](https://github.com/denoland/deno_std/blob/7a0294a7f434919c67a807187dcea7bd5fe4947c/deno.json):
    ```diff
    // deno.json
    {
      "compilerOptions": {
      "strict": true,
      "useUnknownInCatchVariables": true,
      "noImplicitOverride": true,
    + "noUncheckedIndexedAccess": true
      },
    ...
    }
    ```
3. Make the necessary code modifications to your given sub-module(s) until the type-checker is happy. E.g. for `std/collections`:
    ```bash
    deno check collections/mod.ts
    ```
4. Revert the option in the Deno configuration file.
    ```diff
    // deno.json
    {
      "compilerOptions": {
      "strict": true,
      "useUnknownInCatchVariables": true,
[...]

PR #4043, opened 2023-12-30
refactor(media_types): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the media_types module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/media_types/_util.ts
+++ b/media_types/_util.ts
@@ -40,8 +40,9 @@ export function consumeValue(v: string): [value: string, rest: string] {
     if (r === `"`) {
       return [value, v.slice(i + 1)];
     }
-    if (r === "\\" && i + 1 < v.length && isTSpecial(v[i + 1])) {
-      value += v[i + 1];
+    const next = v[i + 1];
+    if (r === "\\" && typeof next === "string" && isTSpecial(next)) {
+      value += next;
       i++;
       continue;
     }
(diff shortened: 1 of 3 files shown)


media_types/_util.ts
```diff
@@ -83,14 +84,15 @@ export function decode2331Encoding(v: string): string | undefined {
   if (sv.length !== 3) {
     return undefined;
   }
-  const charset = sv[0].toLowerCase();
+  const [sv0, , sv2] = sv as [string, string, string];
```
syhol, line 87: Since we check the string length is 3, we know that indices 0,1,2 will always be string.

media_types/parse_media_type.ts
```diff
@@ -42,7 +42,7 @@ import { consumeMediaParam, decode2331Encoding } from "./_util.ts";
 export function parseMediaType(
   v: string,
 ): [mediaType: string, params: Record<string, string> | undefined] {
-  const [base] = v.split(";");
+  const [base] = v.split(";") as [string];
```
syhol, line 45: A `string.split()` with a non-empty string as a parameter (e.g. `";"`) will always return an array with at least one string, so its safe to continue without runtime checks

media_types/_util.ts
```diff
@@ -133,7 +135,7 @@ function isTokenChar(r: string): boolean {
 }
 
 function isTSpecial(r: string): boolean {
-  return `()<>@,;:\\"/[]?=`.includes(r[0]);
+  return r[0] ? `()<>@,;:\\"/[]?=`.includes(r[0]) : false;
```
syhol, line 138: This handles the case of an empty string. Previously
```
`()<>@,;:\\"/[]?=`.includes(undefined)
```
would return `false`.

media_types/parse_media_type.ts
syhol:
```ts
"".split(";")
// [ "" ]
"1234".split(";")
// [ "1234" ]
";;".split(";")
// [ "", "", "" ]
```

iuioiua (approved):
Great! Thank you.

PR #4045, opened 2023-12-31
refactor(assert): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the assert module, tracked in https://github.com/denoland/deno_std/issues/4040


assert/_diff.ts
```diff
@@ -95,6 +95,14 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
     { length: size },
     () => ({ y: -1, id: -1 }),
   );
+  function getPoint(index: number) {
+    const point = fp[index];
+    if (!point) {
+      throw Error(`Missing FarthestPoint at index ${index}`);
+    }
+    return point;
+  }
```
syhol, line 104: Using this new function to make lookups to fp safe adds a 10% performance overhead. I'm not sure if this is acceptable for this module so I wanted to raise it:

BENCH
```typescript
Deno.bench("_diff", () => {
  diff(["abc", "c"], ["abc", "bcd", "c"]);
});
```
BEFORE
```shell
file:///Users/simon.holloway/Code/deno_std/assert/_diff.ts
_diff         198.63 ns/iter   5,034,517.5 (178.23 ns … 679.03 ns) 193.23 ns 662.74 ns 667.21 ns
```
AFTER
```shell
file:///Users/simon.holloway/Code/deno_std/assert/_diff.ts
_diff         219.05 ns/iter   4,565,144.2 (188.39 ns … 730.58 ns) 217.06 ns 680.21 ns  691.8 ns
```

assert/_diff.ts
```diff
@@ -119,28 +127,28 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
...
       } else if (type === ADDED) {
         result.unshift({
           type: swapped ? DiffType.added : DiffType.removed,
-          value: A[a],
+          value: A[a]!,
         });
         a -= 1;
       } else {
-        result.unshift({ type: DiffType.common, value: A[a] });
+        result.unshift({ type: DiffType.common, value: A[a]! });
```
syhol, line 151: I feel a bit unsure about this one. We can't say for certain if these indices are correct as it's all dependent on if `routes` was built correctly.

assert/_diff.ts
syhol: Maybe ignore this, this benchmark seems to fluctuate by ~10% between calls. Not sure if there is actually any impact here.

syhol: Oh, looks like I missed some tests, turns out throwing an error when fp[index] is undefined causes the tests to fail. I need to reconsider how to handle this case.

syhol, line 104: Oh, looks like I missed some tests, turns out throwing an error when fp[index] is undefined causes the tests to fail. I need to reconsider how to handle this case.

assert/_diff.ts
```diff
@@ -35,13 +35,13 @@ const COMMON = 2;
...
 function createCommon<T>(A: T[], B: T[], reverse?: boolean): T[] {
-  const common = [];
+  const common: T[] = [];
   if (A.length === 0 || B.length === 0) return [];
   for (let i = 0; i < Math.min(A.length, B.length); i += 1) {
     if (
       A[reverse ? A.length - i - 1 : i] === B[reverse ? B.length - i - 1 : i]
     ) {
-      common.push(A[reverse ? A.length - i - 1 : i]);
+      common.push(A[reverse ? A.length - i - 1 : i]!);
```
syhol, line 44: Since we do an equality check the line above, this should always be in bound.

syhol (commented):
Best to review this as 4 separate bits:
- **createCommon**: small and simple
- **backTrace**: small risk but ultimately ignoring with `!`
- **createFP / snake / ensureDefined / the bits where currentFP is used**: this was a messy situation and the types were lying, now marking `FarthestPoint` as being possibly undefined in more places, and needing to handle new cases of missing values.
- **diffstr (tokenize / createDetails)**: a few new `const`s and type tweaks, nothing crazy

assert/_diff.ts
```diff
@@ -274,31 +285,36 @@ export function diffstr(A: string, B: string) {
...
+          tokens[i] += tokenPlusTwo;
           tokens.splice(i + 1, 2);
           i--;
         }
       }
       return tokens.filter((token) => token);
     } else {
       // Split string on new lines symbols
-      const tokens = [], lines = string.split(/(\n|\r\n)/);
+      const tokens: string[] = [], lines = string.split(/(\n|\r\n)/);
```
iuioiua, line 305:
```suggestion
      const tokens: string[] = [];
      const lines = string.split(/(\n|\r\n)/);
```

assert/_diff.ts
```diff
@@ -151,37 +152,38 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
   }
 
   function createFP(
-    slide: FarthestPoint,
-    down: FarthestPoint,
+    slide: FarthestPoint | undefined,
+    down: FarthestPoint | undefined,
```
iuioiua, line 156:
```suggestion
    slide?: FarthestPoint,
    down?: FarthestPoint,
```

assert/_diff.ts
```diff
@@ -151,37 +152,38 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
...
+      throw new Error("Unexpected missing FarthestPoint");
     }
   }
 
   function snake<T>(
     k: number,
-    slide: FarthestPoint,
-    down: FarthestPoint,
+    slide: FarthestPoint | undefined,
+    down: FarthestPoint | undefined,
```
iuioiua, line 186:
```suggestion
    slide?: FarthestPoint,
    down?: FarthestPoint,
```

assert/_diff.ts
```diff
@@ -201,7 +203,15 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
     return fp;
   }
 
-  while (fp[delta + offset].y < N) {
+  function ensureDefined<T>(item: T | undefined): T {
+    if (!item) {
+      throw Error("Unexpected missing FarthestPoint");
+    }
+    return item;
+  }
```
iuioiua, line 211:
```suggestion
  function ensureDefined<T>(item?: T): T {
    if (item === undefined) {
      throw Error("Unexpected missing FarthestPoint");
    }
    return item;
  }
```
For strictness. Also, I don't see a reason for this to be a nested function. Can we please make it top-level?

assert/_diff.ts
```diff
@@ -201,7 +203,15 @@ export function diff<T>(A: T[], B: T[]): Array<DiffResult<T>> {
...
+
+  let currentFP: FarthestPoint = ensureDefined(fp[delta + offset]);
```
iuioiua, line 213:
```suggestion
  let currentFP = ensureDefined<FarthestPoint>(fp[delta + offset]);
```

iuioiua (changes requested):

assert/_diff.ts
syhol: I did this following convention in this file, the `diff` function has 3 nested functions. Happy to move it though.

syhol:
Changing
```typescript
function (
    slide: FarthestPoint | undefined,
    down: FarthestPoint | undefined,
) {}
```
to
```typescript
function (
    slide?: FarthestPoint,
    down?: FarthestPoint,
) {}
```
is causing the error:
```
A required parameter cannot follow an optional parameter.deno-ts(1016)
```
Option are:
1. Revert it back to `FarthestPoint | undefined`
2. Move the optional parameters to the end of the parameter list

Personally I prefer option 1 as we never expect to call the function without those parameters. I'll revert it for now but happy to change to option 2 if you wish.

assert/_diff.ts
syhol: Issues with this, see comment https://github.com/denoland/deno_std/pull/4045#issuecomment-1875372330

assert/_diff.ts
syhol: Issues with this, see comment https://github.com/denoland/deno_std/pull/4045#issuecomment-1875372330

syhol:
Waiting on https://github.com/denoland/deno_std/pull/4074 to get checks to pass

iuioiua:
> Changing
> 
> ```ts
> function (
>     slide: FarthestPoint | undefined,
>     down: FarthestPoint | undefined,
>     foo: any,
> ) {}
> ```
> 
> to
> 
> ```ts
> function (
>     slide?: FarthestPoint,
>     down?: FarthestPoint,
>     foo: any,
> ) {}
> ```
> 
> is causing the error:
> 
> ```
> A required parameter cannot follow an optional parameter.deno-ts(1016)
> ```
> 
> Options are:
> 
> 1. Revert it back to `FarthestPoint | undefined`
> 2. Move the optional parameters to the end of the parameter list
> 
> Personally I prefer option 1 as we never expect to call the function without those parameters. I'll revert it for now but happy to change to option 2 if you wish.

Let's do number 1.

assert/_diff.ts
iuioiua: I'm unsure what to do here too. @kt3k, WDYT?

syhol:
> Let's do number 1.

Great, thats already applied as of https://github.com/denoland/deno_std/pull/4045/commits/8aedb7c6f386922fa94fd0fbfb2bab0ab670e83f

assert/_diff.ts
kt3k: `diff` is hugely used across Deno ecosystem (via `assertEquals`). I don't think this includes such basic bugs. I think it's fine to use non null assertion here.

kt3k (approved):
LGTM

Issue #3948, opened 2023-12-13
cleanup semver


timreichen:
## Background
There are several inconsistencies in the semver mod:

### File names and exported functions
The semver mod is not well structured. The file names and functions are not consistent. `range_format.ts` and `comparator_format.ts` uses `[*_verb].ts` and `*Format()`, but `parse_range.ts` and `parse_comparator.ts` uses `[verb_*].ts` and `parse*()`.
Another inconsistency is: `parse_comparator.ts` and `parse_range.ts` but `is_semver_comparator.ts`  and `is_semver_range.ts`. The `_semver` in the name seems obsolete.

### eq, neq, gt, gte, lt, lte
These functuons are wrapper functions for `compare()` compare the result to 0 with different comparators:
`return compare(s0, s1) [===|!==|>|>=|<|<=] 0`. These seem too trivial.

### rcompare
This function calls `compare()` with switched arguments. This seems too trivial.

### File names and exported functions
Rename `[*_verb].ts` to `[verb_*].ts` and  `*Format()` to `format*()` for all filenames and functions and deprecate old functions and remove `_semver` from file names and function:
- [x] `rangeFormat()` => `formatRange()` https://github.com/denoland/deno_std/pull/4090
- [x] ~~`comparatorFormat()` => `formatComparator()`~~ deprecated
- [x] `isSemVerComparator()` => `isComparator()` https://github.com/denoland/deno_std/pull/3957
- [x] `isSemVerRange()` => `isRange()` https://github.com/denoland/deno_std/pull/4161
- [x] `testRange()` => `inRange()` or `rangeIncludes()` https://github.com/denoland/deno_std/pull/4364

### eq, neq, gt, gte, lt, lte
- [x] ~~deprecate all these functions https://github.com/denoland/deno_std/pull/4048~~
- [x] rename all these functions https://github.com/denoland/deno_std/pull/4083
- [x] ~~discuss if these functions should be put in `compare.ts`~~
### rcompare
- [x] deprecate this function https://github.com/denoland/deno_std/pull/3958
### Others
- [x] investigate `Comparator` necessity https://github.com/denoland/deno_std/issues/4047
- [x] investigate possible obsolete `OPERATORS` ("", "==", "===", "!==") https://github.com/denoland/deno_std/pull/4271
- [x] investigate deprecate `compareBuild()` https://github.com/denoland/deno_std/pull/4088
- [x] deprecate `outside()` https://github.com/denoland/deno_std/pull/4185

[...]

PR #4048, opened 2024-01-02
deprecation(semver): deprecate `eq()`, `neq()`, `lt()`, `lte()`, `gt()` and `gte()`

timreichen:
ref: https://github.com/denoland/deno_std/issues/3948#issuecomment-1873464725

- deprecates `eq()`, `neq()`, `lt()`, `lte()`, `gt()` and `gte()`
- adds common use cases to `compare()` jsdoc


semver/eq.ts
```diff
@@ -2,8 +2,12 @@
 import { compare } from "./compare.ts";
 import type { SemVer } from "./types.ts";
 
-/** Returns `true` if they're logically equivalent, even if they're not the exact
- * same version object. */
+/**
+ * Returns `true` if they're logically equivalent, even if they're not the exact same version object.
+ * @deprecated (will be removed in 0.213.0) use {@linkcode compare} instead.
```
iuioiua, line 7:
```suggestion
 * @deprecated (will be removed in 0.213.0) Use {@linkcode compare} instead.
```
Ditto for other instances.

semver/eq.ts
```diff
@@ -2,8 +2,12 @@
...
+ *
+ * This is equal to `compare(s0, s1) === 0`
```
iuioiua, line 9:
```suggestion
 * This is equal to `compare(s0, s1) === 0`.
```
Ditto for other instances.

semver/compare_test.ts
```diff
@@ -54,3 +54,171 @@ Deno.test({
...
+        const s0 = parse(v0);
+        const s1 = parse(v1);
+
+        const eq = compare(s0, s1);
+        const op = expected ? "==" : "!=";
+        assertEquals(eq, expected, `${s0} ${op} ${s1}`);
+      });
+    }
+  },
+});
```
iuioiua, line 101: This is the only test needed. It takes care of all of the other equality and inequality cases more precisely.

iuioiua (changes requested):
I agree with the overall idea of this PR. I have just a few nits.

It'd be good to hear other's opinions on this change.

kt3k:
I'd prefer to keep these methods as they are easier to remember than the meaning of return values of `compare`

timreichen:
> I'd prefer to keep these methods as they are easier to remember than the meaning of return values of `compare`

Ok, I will close this in favour of another PR.

PR #4049, opened 2024-01-02
deprecation(semver): deprecate `tryParse()` and `tryParseRange()`

timreichen:
- deprecates `tryParse()` and `tryParseRange()`


iuioiua (approved):
LGTM

kt3k:
I think these are useful in some cases.

Is this discussed somewhere?

timreichen:
> I think these are useful in some cases.
> 
> Is this discussed somewhere?
https://discord.com/channels/684898665143206084/775393009981849600/1190216234822807602

We have `canParse()` for `SemVer` and no other instances of wrapping a function in a try-catch block in std. There is no explanation for the purpose when it was introduced in https://github.com/denoland/deno_std/pull/3385/files.
 So I think this is something the user can do themselves, no?

kt3k:
`parse` originally return SemVer or null. ref. https://deno.land/std@0.191.0/semver/mod.ts?s=parse

But in the redesign at #3385, it's split into `parse` and `tryParse`. (It's shortly described in #3169) The purpose is to make the naming consistent. I think we should keep both.

timreichen:
> `parse` originally return SemVer or null. ref. https://deno.land/std@0.191.0/semver/mod.ts?s=parse
> 
> But in the redesign at #3385, it's split into `parse` and `tryParse`. (It's shortly described in #3169) The purpose is to make the naming consistent. I think we should keep both.

Yes, but we didn't have `canParse()` then.
The [JSON.tryParse()](https://github.com/Jack-Works/proposal-json-tryParse) proposal was rejected recently as
> specific a solution for a generalized language-wide problem

(ref: https://github.com/tc39/proposals/commit/9d4b241dbb677e24fcee712fbdfe1438ecb82307)

And the same can be achieved with

```ts
const result = canParse(input) ? parse(input) : undefined
```

So I think this isn't very useful. Implementing `canParseRange` (and `canParseComparator`) would make much more sense to me.

kt3k:
`tryParse(input)` looks convenient as a shorthand of `canParse(input) ? parse(input) : undefined`. I think it's good to have all of `parse`, `tryParse` and `canParse`.

iuioiua:
> `tryParse(input)` looks convenient as a shorthand of `canParse(input) ? parse(input) : undefined`. I think it's good to have all of `parse`, `tryParse` and `canParse`.

To me, this shows the amount of overlap between the two functions. There's enough overlap to justify choosing just one function. I prefer `canParse()` as it aligns with `URL.canParse()`. Nor do I think `tryParse()` will be missed. However, my opinion isn't strong. I'm happy to defer to your decision, Yoshiya.

timreichen:
> `tryParse(input)` looks convenient as a shorthand of `canParse(input) ? parse(input) : undefined`. I think it's good to have all of `parse`, `tryParse` and `canParse`.

I think this is too trivial to justify a function for it in the mod.

iuioiua:
Tim, can you please update this PR now that `tryParseComparator()` is deprecated already?

timreichen:
> Tim, can you please update this PR now that `tryParseComparator()` is deprecated already?

Done

iuioiua (approved):
I'm in favour of this. LGTM again.

kt3k:
In my view people usually don't prefer APIs which throw an error, but prefer APIs which return undefined/null for invalid inputs. I think `tryParse` would be the primary way to use `std/semver` in most cases.

iuioiua:
One could use the following:
```ts
const semver = semver.canParse(value) && semver.parse();
```

[`tryParseRange()` isn't used in the wild](https://github.com/search?q=semver%2Ftry_parse_range.ts&type=code&p=1).

kt3k:
We should support primary use case with a single function instead of the combination of 2 API calls with `??` opeartor

iuioiua:
I discussed this with Yoshiya offline. I now agree that these functions do hold value for those who don't want to throw. It's a valid use case, and the maintenance cost for such functions is low. I'll close this PR now. Either way, thank you for the suggestion, Tim. Any idea that genuinely aims to improve design is appreciated.

PR #4050, opened 2024-01-02
deprecation(semver): deprecate `format()` `style` argument

timreichen:
ref: https://github.com/denoland/deno_std/pull/3962#issuecomment-1864389806
- deprecates `format()` `style` argument as it is obsolete due to optional `SemVer` properties

### DIFF
--- a/semver/format.ts
+++ b/semver/format.ts
@@ -13,6 +13,10 @@ function formatNumber(value: number) {
 }
 
 /**
+ * @deprecated (will be removed in 0.213.0)
+ */
+export function format(semver: SemVer, style?: FormatStyle): string;
+/**
  * Format a SemVer object into a string.
  *
  * If any number is NaN then NaN will be printed.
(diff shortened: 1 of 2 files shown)


semver/format.ts
```diff
@@ -12,6 +12,10 @@ function formatNumber(value: number) {
   }
 }
 
+/**
+ * @deprecated (will be removed in 0.213.0)
+ */
+export function format(semver: SemVer, style?: FormatStyle): string;
```
iuioiua, line 18: Having an overload signature that's identical to the implementation signature doesn't make sense. Did you mean to create two overload signatures and deprecate the one that contains the `style` argument?

semver/format.ts
```diff
@@ -22,11 +26,13 @@ function formatNumber(value: number) {
...
  * @returns The string representation of a semantic version.
  */
-export function format(semver: SemVer, style: FormatStyle = "full"): string {
+export function format(semver: SemVer, style?: FormatStyle): string {
   if (semver === ANY) {
     return "*";
   }
 
+  style ??= "full";
+
```
iuioiua, line 35: IMO, these changes aren't needed.

iuioiua (changes requested):

iuioiua (approved):
I'm fine with this change. However, other's opinions would be appreciated.

iuioiua:
Can you please fix the type error?

timreichen:
> Can you please fix the type error?

done

kt3k (approved):
LGTM

PR #4052, opened 2024-01-02
refactor(semver): update internal variables and comments

timreichen:
- updates comments in `_shared` and add examples in jsdoc
- renames `MAIN_VERSION` to `VERSION_CORE` (according to semver bnf)
- renames `FULL_PLAIN` to `FULL_VERSION` (according to semver bnf)
- renames `XRANGE_PLAIN` to `XRANGE`

### DIFF
--- a/semver/_shared.ts
+++ b/semver/_shared.ts
@@ -107,9 +107,9 @@ const BUILD =
 /**
  * A version, followed optionally by a pre-release version and build metadata.
  */
-const SEMVER = `v?${VERSION_CORE}${PRERELEASE}?${BUILD}?`;
+const FULL_VERSION = `v?${VERSION_CORE}${PRERELEASE}?${BUILD}?`;
 
-export const FULL_REGEXP = new RegExp(`^${SEMVER}$`);
+export const FULL_REGEXP = new RegExp(`^${FULL_VERSION}$`);
 
 /**
  * A comparator.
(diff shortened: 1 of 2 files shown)


semver/_shared.ts
```diff
@@ -63,78 +63,89 @@ export function compareIdentifier(
...
-
-// Note that the only major, minor, patch, and pre-release sections of
-// the version string are capturing groups.  The build metadata is not a
-// capturing group, because it should not ever be used in version
-// comparison.
-const FULL_PLAIN = `v?${MAIN_VERSION}${PRERELEASE}?${BUILD}?`;
+/**
+ * A version, followed optionally by a pre-release version and build metadata.
+ */
+const SEMVER = `v?${VERSION_CORE}${PRERELEASE}?${BUILD}?`;
```
iuioiua, line 110: I think how this line was before was fine. All the other changes look good.

iuioiua (changes requested):

semver/_shared.ts
timreichen: Do you mean the comment about capturing groups? Because `BUILD` is using a capturing group and is used for parsing. It is not used for comparisons, but that logic is inside `compare()`.

iuioiua: Sorry, I just meant how the constant and its parts are named.

timreichen: Ah, I see. I don't know, I think `FULL_PLAIN` is not very descriptive. The comment even had to clarify: `## Full Version String` and `A main version, ...`. Maybe `SEMVER_VERSION` or `FULL_VERSION`?

iuioiua: `FULL_VERSION` SGTM

timreichen: Done

iuioiua (approved):
Thank you

PR #4054, opened 2024-01-02
refactor(semver): make `difference()` more readable

timreichen:


semver/difference.ts
```diff
@@ -1,32 +1,28 @@
...
-/** Returns difference between two versions by the release type, or
- * `undefined` if the versions are the same. */
+/**
+ * Returns difference between two versions by the release type, or `undefined` if the versions are the same.
+ */
 export function difference(
-  s0: SemVer,
-  s1: SemVer,
+  version1: SemVer,
+  version2: SemVer,
```
iuioiua, line 10: IMO, `s0` and `s1` were fine as they were. Was there something about them that was unclear?

timreichen: I think we should not have to abbreviate parameter names. While the code gets a bit bigger, full names are easier to comprehend in context imo.

kt3k: `s0` `s1` look fine to me too as the scope of them are enough short here. I think there's very little chance for them to be confused with something else.

timreichen: ok, reverted.

iuioiua (approved):
LGTM. Thank you.

PR #4055, opened 2024-01-02
BREAKING(crypto): remove stable `KeyStack()`

iuioiua:
`KeyStack()` is available in `crypto/unstable_keystack.ts`.


kt3k (approved):
LGTM. Thanks!

PR #4058, opened 2024-01-02
docs(json): complete documentation

sylc:
Towards https://github.com/denoland/deno_std/issues/3764

`deno doc --lint json/mod.ts` has no errors.


json/json_stringify_stream.ts
```diff
@@ -82,6 +82,9 @@ export interface StringifyStreamOptions {
  * ```
  */
 export class JsonStringifyStream extends TransformStream<unknown, string> {
+  /**
+   * Construct new instance
+   */
```
crowlKats, line 87: This should have documentaiton about the options that can be provided.

sylc: done

iuioiua: @crowlKats, `StringifyStreamOptions` is fully documented above. To clarify, did you mean to provide example snippets where the constructor is used in different ways? Is that required if we already have examples on the class above?

iuioiua (approved):
I've made a bunch of tweaks. PTAL, @sylc. These were done as I was figuring out how this should all be done. Regarding class constructors, their argument types were documented, which suffices. Examples are done on the classes themselves, rather than the constructors.

Thank you for this.

Issue #4047, opened 2024-01-01
Discussion(semver): Necessity of `Comparator`


timreichen:
## Background

ref: https://github.com/denoland/deno_std/pull/4037#pullrequestreview-1797976465, https://github.com/denoland/deno_std/pull/4037#issuecomment-1871622745

This issue explores, if `Comparator` is necessary in the `semver` module.

@kt3k What made you think that `Comparator` might not be necessary?

kt3k:
Comparator looks like an internal utility interface. It is very limited version of `SemVerRange`. I think the users are usually interested in working with `SemVerRange`s and not `Comparator`s.

`Comparator` still needs to be exposed as a public type because it's part of public `SemVerRange`, but I think the users are not interested in APIs which accepts/outputs comparators (e.g. `parseComparator`, `testComparator`)

timreichen:
I agree. `Comparator` is probably not very useful as part of the public api. I think we should remove it from the public api.
However, working with `SemVerRange` directly is kinda clunky because
- it has an obsolete `ranges` property
- it represents an array of arrays of `Comparators`
That might not be a problem when using `parseRange`, but it is when using objects. Creating a simple Range is a bit awkward:
```ts
const simpleRange = { ranges: [ [ { operator: "=", semver: { major: 1, minor: 2, patch: 3 } } ] ] }
```
We might need to find a good solution to simplify usage. (related: https://github.com/denoland/deno_std/issues/3967)

timreichen:
Ok, I will go forward and create a PR to deprecate all public `Comparator` functions and types.

iuioiua:
Completed in #4131.

PR #4059, opened 2024-01-02
deprecation(semver): deprecate `Comparator.semver` property

timreichen:
- deprecates `Comparator.semver` property in favour of `Comparator`extending `SemVer`.


iuioiua:
What is the reasoning behind this?

timreichen:
> What is the reasoning behind this?

Imo separating `Operator` and `SemVer` by individual properties is just unnecessary nesting.

iuioiua:
I'm not sure about this change. It'd require many changes to the `std/semver` API, and its benefits aren't apparent.

timreichen:
> I'm not sure about this change. It'd require many changes to the `std/semver` API, and its benefits aren't apparent.

I think it declutters the interface and. It also will simplify `SemVerRange`, and once we make `SemVerRange` redefine `SemVerRange` without the `ranges` property, it will look much more clean:

before:
```ts
const range = { ranges: [ [ { operator: "<", semver: { major: 1, minor: 2, patch: 3 } } ] ] }
```
after
```ts
const range = [ [ { operator: "<", major: 1, minor: 2, patch: 3 } ] ]
```
But we can wait with this until we decide about https://github.com/denoland/deno_std/issues/4047 and discuss this some more.

iuioiua:
> But we can wait with this until we decide about #4047 and discuss this some more.

Yeah, I think we should do that.

kt3k:
I think we should first decide on what to do about #4047 before changing anything about Comparator

timreichen:
> I think we should first decide on what to do about #4047 before changing anything about Comparator

This is about the type `Comparator` which will stay public either way because it is part of `SemVerRange`.

kt3k:
> This is about the type Comparator which will stay public either way because it is part of SemVerRange.

I think we can stop using `Comparator` in `SemVerRange` definition if we only want to change `SemVerRange`

timreichen:
> > This is about the type Comparator which will stay public either way because it is part of SemVerRange.
> 
> I think we can stop using `Comparator` in `SemVerRange` definition if we only want to change `SemVerRange`

What do you mean exactly?

kt3k:
For example, by replacing `Comparator` in `SemVerRangeAnd` like the below:

```ts
type SemVerRangeAnd = (SemVer & { operator: Operator })[];
```

iuioiua:
Does this PR still apply if `Comparator` is now deprecated?

timreichen:
> Does this PR still apply if `Comparator` is now deprecated?

I will push another PR where we can do that step when renaming `SemVerRange` to `Range` directly. Closing.

PR #4064, opened 2024-01-02
refactor(toml): format test names

timreichen:

### DIFF
--- a/toml/parse_test.ts
+++ b/toml/parse_test.ts
@@ -25,7 +25,7 @@ import {
 import { parse } from "./parse.ts";
 
 Deno.test({
-  name: "[TOML parser] Scanner",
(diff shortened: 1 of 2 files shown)


toml/parse_test.ts
```diff
@@ -25,7 +25,7 @@ import {
 import { parse } from "./parse.ts";
 
 Deno.test({
-  name: "[TOML parser] Scanner",
+  name: "Scanner()",
```
iuioiua, line 28:
```suggestion
  name: "Scanner",
```
I discussed this Yoshiya. Let's omit the brackets from class names. It being in PascalCase should be enough to signify it's a class.

iuioiua (approved):
LGTM. Thank you.

PR #4065, opened 2024-01-03
deprecation(semver): deprecate `compareBuild()` in favour of `compare()` matcher option

timreichen:
ref: https://github.com/denoland/deno_std/issues/3948#issuecomment-1874699487

- deprecates `compareBuild()`
- adds `CompareOptions` interface to `compare()` with `matcher` property. The name `matcher` is borrowed from [localematcher ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#localematcher)


timreichen:
closed in favour of https://github.com/denoland/deno_std/pull/4088

PR #4076, opened 2024-01-03
refactor(semver): remove unnecessary code

timreichen:
- removes unnecessary code and simplify statements


semver/is_semver.ts
```diff
@@ -25,22 +25,28 @@ export function isSemVer(value: unknown): value is SemVer {
...
+    major,
+    minor,
+    patch,
+    build = [],
+    prerelease = [],
+  } = value as Record<string, unknown>;
+  return (
+    typeof major === "number" && isValidNumber(major) &&
     typeof minor === "number" && isValidNumber(minor) &&
     typeof patch === "number" && isValidNumber(patch) &&
```
iuioiua, line 38: Do we need `typeof part === "number"` if we're calling `isValidNumber(part)` anyway?

semver/is_semver.ts
```diff
@@ -25,22 +25,28 @@ export function isSemVer(value: unknown): value is SemVer {
...
     Array.isArray(prerelease) &&
+    prerelease
+      .every((v) => typeof v === "string" || typeof v === "number") &&
+    prerelease
+      .filter((v) => typeof v === "string")
+      .every((v) => isValidString(v)) &&
```
iuioiua, line 44:
```suggestion
      .filter((v) => isValidString(v)).length &&
```
Nit: Can this be done here and for the one below? Does it cater for edge cases?

iuioiua (changes requested):

semver/is_semver.ts
timreichen: You are right, that is not necessary.

semver/is_semver.ts
timreichen: That wouldn't work, but it can be reduced to
```ts
prerelease.every((v) =>
  (typeof v === "string" && isValidString(v)) ||
  (typeof v === "number" && isValidNumber(v))
) 
```

timreichen: Actually:
```ts
prerelease.every((v) => isValidString(v)) || isValidNumber(v)) 
```

iuioiua (approved):
LGTM. Nice.

PR #4077, opened 2024-01-03
test(path): move tests away from using deprecated `path.posix` and `path.win32`

babiabeo:
Closes: #4073


iuioiua (approved):
I renamed the use of the `win32` alias to `windows`. Now, LGTM! Thank you for this.

PR #4081, opened 2024-01-03
fix(fs): fix `expandGlob()` test race condition

iuioiua:
Should fix https://github.com/denoland/deno_std/actions/runs/7402850608/job/20141568466


kt3k (approved):
LGTM 👍

PR #4082, opened 2024-01-03
feat(text): cases

timreichen:
This is the successor of https://github.com/denoland/deno_std/pull/3440. Big thanks to @Gustrb for the initial PR.

- implements the following cases:
  - `toCamelCase()`
  - `toKebabCase()`
  - `toPascalCase()`
  - `toScreamingSnakeCase()`
  - `toSentenceCase()`
  - `toSnakeCase()`
  - `toTitleCase()`
  
They are interchangeable with each other, so this works:
```ts
let input = "hello world";
input = input.toUpperCase();
input = toCamelCase(input);
input = toKebabCase(input);
input = toPascalCase(input);
input = toScreamingSnakeCase(input);
input = toSentenceCase(input);
input = toSnakeCase(input);
input = toTitleCase(input);
input = input.toLowerCase();
console.log(input) // "hello world"
```

### DIFF
--- a/text/_util.ts
+++ b/text/_util.ts
@@ -1,0 +1,16 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+export function split(input: string) {
+  input = input.trim();
+  if (input.includes(" ")) return input.split(/ +/).filter(Boolean);
+  if (input.includes("-")) return input.split(/-+/).filter(Boolean);
+  if (input.includes("_")) return input.split(/_+/).filter(Boolean);
+  return input.split(/(?=[A-Z])+/).filter(Boolean);
+}
+export function capitalizeWord(word: string): string {
+  if (!word) return word;
+  const firstChar = word[0];
+  const firstCharAsUpperCase = firstChar.toLocaleUpperCase();
+  const restOfTheString = word.slice(1).toLocaleLowerCase();
+  return `${firstCharAsUpperCase}${restOfTheString}`;
+}
--- a/text/case.ts
+++ b/text/case.ts
@@ -1,0 +1,130 @@
+import { capitalizeWord, split } from "./_util.ts";
+
+/**
+ * Converts a input into camelCase
+ *
+ * @example
+ * ```ts
+ * import { toCamelCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toCamelCase("deno is awesome"), "denoIsAwesome");
+ * ```
+ *
+ * @param input The string that is going to be converted into camelCase
+ * @returns The string as camelCase
+ */
+export function toCamelCase(input: string): string {
+  const [first = "", ...rest] = split(input);
+  return [first.toLocaleLowerCase(), ...rest.map(capitalizeWord)].join("");
+}
+
+/**
+ * Converts a input into kebab-case
+ *
+ * @example
+ * ```ts
+ * import { toKebabCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toKebabCase("deno is awesome"), "deno-is-awesome");
+ * ```
+ *
+ * @param input is the string that is going to be converted into kebab-case
+ * @returns The string as kebab-case
+ */
+export function toKebabCase(input: string): string {
+  return split(input).join("-").toLocaleLowerCase();
+}
+
+/**
+ * Converts a input into PascalCase
+ *
+ * @example
+ * ```ts
+ * import { toPascalCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toPascalCase("deno is awesome"), "DenoIsAwesome");
+ * ```
+ *
+ * @param input The string that is going to be converted into PascalCase
+ * @returns The string as PascalCase
+ */
+export function toPascalCase(input: string): string {
+  return split(input).map(capitalizeWord).join("");
+}
+
+/**
+ * Converts a input into SCREAMING_SNAKE_CASE
+ *
+ * @example
+ * ```ts
+ * import { toScreamingSnakeCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toScreamingSnakeCase("deno is awesome"), "DENO_IS_AWESOME");
+ * ```
+ *
+ * @param input is the string that is going to be converted into SCREAMING_SNAKE_CASE
+ * @returns The string as SCREAMING_SNAKE_CASE
+ */
+export function toScreamingSnakeCase(input: string): string {
+  return toSnakeCase(input).toLocaleUpperCase();
+}
+
+/**
+ * Converts a input into Sentence case
+ *
+ * @example
+ * ```ts
+ * import { toSentenceCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ * assertEquals(toSentenceCase("deno is awesome"), "Deno is awesome");
+ * ```
+ *
+ * @param input is the string that is going to be converted into Sentence case
+ * @returns The string as Sentence case
+ */
+export function toSentenceCase(input: string): string {
+  const [first = "", ...rest] = split(input);
+  return [
+    capitalizeWord(first),
+    ...rest.map((word) => word.toLocaleLowerCase()),
+  ].join(" ");
+}
+
+/**
+ * Converts a input into snake_case
+ *
+ * @example
+ * ```ts
+ * import { toSnakeCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ * assertEquals(toSnakeCase("deno is awesome"), "deno_is_awesome");
+ * ```
+ *
+ * @param input is the string that is going to be converted into snake_case
+ * @returns The string as snake_case
+ */
+export function toSnakeCase(input: string): string {
+  return split(input).join("_").toLocaleLowerCase();
+}
+
+/**
+ * Converts a input into Title Case
+ *
+ * @example
+ * ```ts
+ * import { toTitleCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toTitleCase("deno is awesome"), "Deno Is Awesome");
+ * ```
+ *
+ * @param input is the string that is going to be converted into Title Case
+ * @returns The string as Title Case
+ */
+export function toTitleCase(input: string): string {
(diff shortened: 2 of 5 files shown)


text/_util.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (input.includes("-")) return input.split(/-+/).filter(Boolean);
+  if (input.includes("_")) return input.split(/_+/).filter(Boolean);
+  return input.split(/(?=[A-Z])+/).filter(Boolean);
+}
+export function capitalizeWord(word: string): string {
+  if (!word) return word;
+  const firstChar = word[0];
+  const firstCharAsUpperCase = firstChar.toLocaleUpperCase();
+  const restOfTheString = word.slice(1).toLocaleLowerCase();
+  return `${firstCharAsUpperCase}${restOfTheString}`;
```
iuioiua, line 15:
```suggestion
  return word ? word[0].toLocaleUpperCase() + word.slice(1).toLocaleLowerCase() : word;
```

text/_util.ts
```diff
@@ -0,0 +1,16 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+export function split(input: string) {
+  input = input.trim();
+  if (input.includes(" ")) return input.split(/ +/).filter(Boolean);
+  if (input.includes("-")) return input.split(/-+/).filter(Boolean);
+  if (input.includes("_")) return input.split(/_+/).filter(Boolean);
```
iuioiua, line 7:
```suggestion
  if (/[\s-_]+/.test(input)) return input.split(/[\s-_]+/).filter(Boolean);
```
This should also take care of mixed delimiters if we want to do that.

text/case.ts
```diff
@@ -0,0 +1,132 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import { capitalizeWord, split } from "./_util.ts";
+
+/**
+ * Converts a input into camelCase
```
iuioiua, line 6:
```suggestion
 * Converts a string into camelCase.
```
Ditto for other instances.

text/case.ts
```diff
@@ -0,0 +1,132 @@
...
+ *
+ * @example
+ * ```ts
+ * import { toCamelCase } from "https://deno.land/std@$STD_VERSION/text/case.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/testing/asserts.ts";
+ *
+ * assertEquals(toCamelCase("deno is awesome"), "denoIsAwesome");
```
iuioiua, line 13:
```suggestion
 * toCamelCase("deno is awesome"); // "denoIsAwesome"
```
This means we can remove the `assertEquals()` import too. Ditto for other instances.

iuioiua (changes requested):
Nicely done! Just a few nits.

text/case.ts
timreichen: done

text/_util.ts
timreichen: The question is if we want to handle that. If we do, we should also consider handling punctuation as well as different languages chars, such as german umlaute.
I am not sure if we should tackle that or just leave it to the user to prepare the string before applying a case.

Different cases require different transformations, replacing either one or all delimiters. That would mean data loss between them:

```ts
// replacing all delimiters and punctuation
console.log(toCamelCase("I am up-to-date!")); // should be "iAmUpToDate"
console.log(toKebabCase("I am up-to-date!")); // should be "i-am-up-to-date"
console.log(toPascalCase("I am up-to-date!")); // should be "IAmUpToDate"
console.log(toScreamingSnakeCase("I am up-to-date!")); // should be "I_AM_UP_TO_DATE"
console.log(toSnakeCase("I am up-to-date!")); // should be "i_am_up_to_date"

// replacing one delimiter
console.log(toTitleCase("I am up-to-date!")); // should be "I Am Up-to-date!"
console.log(toSentenceCase("I AM UP-TO-DATE!")); // should be "I am up-to-date!"
```
So maybe let `split()` have an `splitAllDelimiters` option  which is true for all cases but `toSentenceCase()` and `toTitleCase()`
and also remove all `/\W/` if splitAllDelimiters is true?

timreichen:
I updated the PR so split takes `singleDelimiter` and/or `removeSpecialCharacters` options. It also no longers trims as `toSentenceCase` and `toTitleCase` want to keep whitespace.

text/_util.ts
iuioiua: No, please, let's keep it simple. Let's use your original implementation. It's good enough for the first pass. Let's discuss options in a separate follow-up issue.

text/_util.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+  if (singleDelimiter) {
+    if (/\s+/.test(input)) return input.split(/\s+/).filter(Boolean);
+    if (/-+/.test(input)) return input.split(/-+/).filter(Boolean);
+    if (/_+/.test(input)) return input.split(/_+/).filter(Boolean);
+  } else {
+    if (/[\s-_]+/.test(input)) return input.split(/[\s-_]+/).filter(Boolean);
+  }
+  return input.split(/(?=[A-Z])+/).filter(Boolean);
+}
```
iuioiua, line 23: Let's revert this to the original version. It's good for the first-pass.

timreichen: Are you sure? I think it is an improvement that handles a bunch of edge cases concerning delimiters, whitespaces and special characters.

iuioiua: Yes, for now. I know it won't be what's published upon release, but it allows us to explore ideas. Frankly, I'm not a fan of the `SplitOptions` argument here. My idea was to perhaps have a 2nd optional parameter, say `separator: string | RegExp`, that the user can define but has a reasonable default. Again, I'd want to play with the idea a little first.

iuioiua: The problem with this function is that there's a seemingly arbitrary order of precedence in the characters that the word is split by. I.e. the word is split by spaces if spaces are what's first detected. Then dashes. Then underscores. Why in that order?

I think we should just split by all of these delimiters in one hit, as I previously suggested. I understand that it'll also split words with delimiters of mixed types. But that seems like the correct behaviour to me and is behaviour that's easier to understand. Also, that's what lodash does.
```ts
import camelCase from "npm:lodash.camelcase";
console.log(camelCase("why hello-there")); // "whyHelloThere"
```

iuioiua (changes requested):

text/case_test.ts
```diff
@@ -0,0 +1,246 @@
...
+  const input = "she turned me into a newt";
+  const expected = "She Turned Me Into A Newt";
+  assertEquals(toTitleCase(input), expected);
+});
+
+Deno.test({
+  name: "toTitleCase() converts multiple delimiters",
+  fn() {
+    const result = toTitleCase("I am up-to-date!");
+    const expected = "I Am Up-to-date!";
```
kt3k, line 234: Does this need to be like this?

timreichen: Well, it should be `I Am Up-to-Date!`, but title case is not easy to do correctly without analyzing language and grammar. As pointed out on https://titlecaseconverter.com:
> I	is capitalized because it is the first word of the title.
Am	is capitalized because it is a verb, and all verbs are capitalized in title case. More information on [capitalizing am](https://titlecaseconverter.com/blog/is-am-capitalized/)
Up	is capitalized because it seems to be part of an adjective or noun. More information on [capitalizing up](https://titlecaseconverter.com/blog/is-up-capitalized/) • [capitalizing hyphenated words](https://titlecaseconverter.com/blog/capitalizing-hyphenated-words/)
to	is not capitalized because it is used as a preposition. More information on [capitalizing to](https://titlecaseconverter.com/blog/is-to-capitalized/) • [capitalizing hyphenated words](https://titlecaseconverter.com/blog/capitalizing-hyphenated-words/)
Date	is capitalized because it is the last word of the title or subtitle. More information on [capitalizing hyphenated words](https://titlecaseconverter.com/blog/capitalizing-hyphenated-words/)

timreichen:
@iuioiua I think we could solve that if we create a separate splitter for `toSentenceCase()` and `toTitleCase()` these two can contain whitespace. The order of delimiters doesn't matter for all others. I'lll do some experiments today.

iuioiua (changes requested):
PTAL at some of the suggestions in my previous review.

text/case.ts
```diff
@@ -0,0 +1,137 @@
...
+ *
+ * toScreamingSnakeCase("deno is awesome"); // "DENO_IS_AWESOME"
+ * ```
+ *
+ * @param input is the string that is going to be converted into SCREAMING_SNAKE_CASE
+ * @returns The string as SCREAMING_SNAKE_CASE
+ */
+export function toScreamingSnakeCase(input: string): string {
+  return toSnakeCase(input).toLocaleUpperCase();
+}
```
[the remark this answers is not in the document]
timreichen: done

text/_util.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+export interface SplitOptions {
+  singleDelimiter?: boolean;
+  removeSpecialCharacters?: boolean;
+}
+
+export function split(
```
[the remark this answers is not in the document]
timreichen: I am not sure if `normalize` describes well enough what the function is actually doing. How about `splitToWords` or `splitToTextFragments`?

text/_util.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+export interface SplitOptions {
```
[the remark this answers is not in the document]
timreichen: removing the export makes sense. See above about the rename.

text/_util.ts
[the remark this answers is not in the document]
iuioiua: I'm happy with any name that describes what the function does 🙂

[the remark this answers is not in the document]
iuioiua: `splitToWords()` might be fine.

iuioiua:
> * drop `toTitleCase()` for now because it needs to understand grammar for good output. (ref: [this](https://github.com/blakeembrey/change-case/blob/main/packages/title-case/src/index.ts) implementation has a basic grammar analysis but only in English)

I'm happy to add `toTitleCase()` later. I think we can have further discussions on how it can be done.

> * rewrite `split()` regexp to split without delimiter order preference as suggested.

👍🏾 

> * handle `toSentenceCase()` split separately with ` input.split(/\s/)`.

One question I have is why each function must split its words differently. IMO, each phrase should be split in the same way. It'd make understanding and working with all these functions much easier and still provide good DX. WDYT, @kt3k and @timreichen?

timreichen:
> One question I have is why each function must split its words differently. IMO, each phrase should be split in the same way. It'd make understanding and working with all these functions much easier and still provide good DX. WDYT, @kt3k and @timreichen?

What I  meant was that `toSentenceCase()` is different, because it keeps whitespaces and special chars while all other functions do not. We can use `splitToWords()` for all except `toSentenceCase()`.

iuioiua:
I had a chat with Yoshiya. He also thinks that `splitToWords()` is perhaps too complex. I'm still adamant that all input strings should be split in an identical manner, irrespective of the output case.

It'd be good to get this PR over the line. I suggest we do the following:
1. Remove `toSentenceCase()` for now. This can be tackled later.
2. Simplify `splitToWords()` and ensure all functions split words in the same way. I.e. remove options from it.

That should be good enough for the first pass. We can discuss and iterate after.

timreichen:
> I had a chat with Yoshiya. He also thinks that `splitToWords()` is perhaps too complex. I'm still adamant that all input strings should be split in an identical manner, irrespective of the output case.
> 
> It'd be good to get this PR over the line. I suggest we do the following:
> 
> 1. Remove `toSentenceCase()` for now. This can be tackled later.
> 2. Simplify `splitToWords()` and ensure all functions split words in the same way. I.e. remove options from it.
> 
> That should be good enough for the first pass. We can discuss and iterate after.

Ok, done.

kt3k (approved):
LGTM

Issue #4027, opened 2023-12-27
deprecation note for `assertEquals` is not displayed in IDE


kt3k:
When I overlay the API doc of 'deprecated' assertEquals of `testing/asserts.ts`, the deprecation note for it (`(will be removed after 1.0.0) Import from {@link https://deno.land/std/assert/assert_equals.ts} instead`) is not displayed in IDE. (The user can't get the information about the replacement in IDE)

<img width="841" alt="Screenshot 2023-12-27 at 23 01 40" src="https://github.com/denoland/deno_std/assets/613956/3fb8c1de-d77e-443f-8aee-8e335dbad12c">

iuioiua:
I tested locally and can reproduce the issue. It appears that deprecation notices on re-exports, like those done in `testing/asserts.ts`, aren't correctly displayed. However, deprecation notices display fine when one imports a dependency and then exports that as a constant.

Doesn't work:
```ts
export {
  /** @deprecated ... */
  foo
} from "./foo.ts";
```

Works:
```ts
import { foo as _foo } from "./foo.ts";

/** @deprecated ... */
export foo = _foo;
```

This appears to be a runtime or VSCode plugin issue. I'll open up a separate issue in the main repo shortly.

reggi:
I'm seeing this too, when I go to the source code there is no `@deprecation` comment. Is this a false flag because of the re-export?

<img width="895" alt="Screenshot 2023-12-29 at 10 53 28 AM" src="https://github.com/denoland/deno_std/assets/296798/f55bc5e5-17ef-4d79-a848-3a1adfe41cb0">

iuioiua:
@reggi, what you're seeing is expected behaviour.

This is the old, deprecation location for this function.
```ts
import { assertEquals } from "https://deno.land/std@0.210.0/testing/asserts.ts";
```

This is the new, updated location for this function.
```ts
import { assertEquals } from "https://deno.land/std@0.210.0/assert/mod.ts";
```

kt3k:
@iuioiua 

```ts
import { foo as _foo } from "./foo.ts";

/** @deprecated ... */
export const foo = _foo;
```

This pattern still doesn't seem working to me.

<img width="1078" alt="Screenshot 2023-12-31 at 21 14 29" src="https://github.com/denoland/deno_std/assets/613956/6de23926-2391-465f-9097-bd42efa0d844">

kt3k:
[...]

PR #4083, opened 2024-01-04
deprecation(semver): rename `eq()`, `neq()`, `lt()`, `lte()`, `gt()` and `gte()`

timreichen:
Successor of https://github.com/denoland/deno_std/pull/4048

Refs: https://github.com/denoland/deno_std/issues/3948#issuecomment-1869159287,  https://github.com/denoland/deno_std/pull/4048#issuecomment-1873922759

- renames comparator abbreviated functions to the full name
- deprecates abbreviated functions
- adds `less_or_equal_test.ts`


iuioiua (changes requested):

iuioiua (approved):
LGTM. Thank you!

semver/eq.ts
```diff
@@ -1,9 +1,9 @@
...
- * same version object. */
-export function eq(s0: SemVer, s1: SemVer): boolean {
-  return compare(s0, s1) === 0;
-}
+/**
+ * Returns `true` if they're logically equivalent, even if they're not the exact same version object.
+ *
+ * @deprecated (will be removed after 0.213.0) Use {@linkcode equals} instead.
+ */
+export const eq = equals;
```
kt3k, line 9: Can you do the deprecation like the below structure? This structure of deprecation has a problem in showing the `@deprecation` note in IDE (#4027).

```ts
/**
 * @deprecated (will be removed after 0.213.0) Use {@linkcode equals} instead.
 *
 * Returns `true` if they're logically equivalent, even if they're not the exact
 * same version object. 
 */
export function eq(s0: SemVer, s1: SemVer): boolean {
  return equals(s0, s1);
}
```

kt3k:
@iuioiua 

> Yes. In cases where functions don't incur any dependency costs on each other and share much overlap, it might make sense to have them in the same file. With equals() and friends, this might make sense to do. It also might not. E.g. if such a structure would be surprising to a newcomer, it might not make sense to do. WDYT, @kt3k?

Sounds interesting/reasonable idea to me, but not immediately obvious if it's good or not. I'd like to keep discussing that in somewhere else.

timreichen:
> @iuioiua
> 
> > Yes. In cases where functions don't incur any dependency costs on each other and share much overlap, it might make sense to have them in the same file. With equals() and friends, this might make sense to do. It also might not. E.g. if such a structure would be surprising to a newcomer, it might not make sense to do. WDYT, @kt3k?
> 
> Sounds interesting/reasonable idea to me, but not immediately obvious if it's good or not. I'd like to keep discussing that in somewhere else.

I'll add it to the list in https://github.com/denoland/deno_std/issues/3948

timreichen:
> Another concern about naming: We have `assertLess` `assertGreater` `assertGreaterOrEqual` `assertLessOrEqual` in `std/assert` #3496
> 
> So these names might look inconsistent with them. What do you think?

Thank you for bringing that up. I think we should align to assert names, I'll update the PR.

semver/eq.ts
timreichen: done

iuioiua:
> > Another concern about naming: We have `assertLess` `assertGreater` `assertGreaterOrEqual` `assertLessOrEqual` in `std/assert` #3496
> > So these names might look inconsistent with them. What do you think?
> 
> Thank you for bringing that up. I think we should align to assert names, I'll update the PR.

Sorry for coming in late. I don't see a reason to align these function names with those from `std/assert`. They're different APIs with different purposes. But more importantly, I think removing the "than" word from these function names decreases their clarity. E.g. the purpose of `less()` in `std/semver` isn't immediately obvious. However, `lessThan()` in `std/semver` is clearer. What do we think?

kt3k:
@iuioiua 
What do you think about the below?
- lessThan
- lessOrEqual
- greaterThan
- greaterOrEqual

iuioiua:
> @iuioiua What do you think about the below?
> 
> * lessThan
> * lessOrEqual
> * greaterThan
> * greaterOrEqual

SGTM. To me, `lessThanOrEqual()` and friends sound better. But that's a weak opinion and not something I'd block any PR over. I also understand that `lessThanOrEqual()` may be too long.

timreichen:
> SGTM. To me, `lessThanOrEqual()` and friends sound better. But that's a weak opinion and not something I'd block any PR over. I also understand that `lessThanOrEqual()` may be too long.

@kt3k how should we proceed?

kt3k:
> @kt3k how should we proceed?

The below pattern looks best among candidates to me.

- lessThan
- lessThanOrEqual
- greaterThan
- greaterThanOrEqual

timreichen:
> > @kt3k how should we proceed?
> 
> The below pattern looks best among candidates to me.
> 
> * lessThan
> * lessThanOrEqual
> * greaterThan
> * greaterThanOrEqual

done

iuioiua (approved):
I made a couple of tweaks:
1. Moved deprecation notices to after their respective JSDoc descriptions. This is needed for deprecation directives to show up properly in IDEs, etc.
2. Added an "s" to the "*OrEqual()` functions, to be consistent with `equals()`.

Now, LGTM! Great addition, Tim. Thank you.

kt3k:
Mostly looks good to me. Thanks for updating.

> Added an "s" to the `*OrEqual()` functions, to be consistent with `equals()`.

Does that sound better to you? If I search `lessThanOrEqual` and `lessThanOrEquals` in google, the former returns more results (`lessThanOrEqual` looks more common as an API name to me). I think `equal` here is more of an adjective rather than a verb..

iuioiua:
I just thought "a equals b".

kt3k:
- `a equals b` <- equals is a verb, so it has `s` suffix
- `a is less than or equal b` <- `less than` and `equal` are adjectives. equal doesn't have `s`

kt3k:
> `lessThanOrEqual` or `lessThanOrEquals`

Maybe not a big deal, but I'd like to hear some more from community members

iuioiua:
I know I said the names don't have to be consistent with `std/assert`, but now, the names are consistent with `std/assert` 😂

Either way, the names convey the purpose of the functions, which is the objective of a function name, with or without the "s".

Leokuma:
TBH I didn't mind the abbreviated names. But I don't mind the longer names either.

I vote for `lessThanOrEquals`. To be very precise grammatically, it should be `lessThanOrEqualTo`, which is too long. So, if we're going to be grammatically imprecise anyway, `lessThanOrEquals` is semantically closer to `lessThanOrEqualTo` than `lessThanOrEqual` is because it keeps a strong sense of direction (which means the order of the arguments in this case).

timreichen:
> TBH I didn't mind the abbreviated names. But I don't mind the longer names either.
> 
> 
> 
> I vote for `lessThanOrEquals`. To be very precise grammatically, it should be `lessThanOrEqualTo`, which is too long. So, if we're going to be grammatically imprecise anyway, `lessThanOrEquals` is semantically closer to `lessThanOrEqualTo` than `lessThanOrEqual` is because it keeps a strong sense of direction (which means the order of the arguments in this case).

Well, `lessThanOrEqualTo` is just one letter longer than `lessThanOrEquals`:) I think the length should not be the determining factor, because everybody is probably using autocomplete anyway.

Leokuma:
That's true. But if we go for `lessThanOrEqualTo`, maybe we should also rename `equals` to `equalTo`. It feels a little too verbose to me. And when I see the word `equals`, my brain parses it as a symbol rather than a "grammatical word", so it's actually faster for me to read `lessThanOrEquals` than `lessThanOrEqualTo`. So I'm still in favor `lessThanOrEquals`. Sorry if I'm overthinking this 😅.

iuioiua:
I now think it is worth simply aligning with `std/assert`. The names are short and simple and show their purpose. That's all that's needed.

I'm sorry for the indecisiveness. In hindsight, the PR was fine in its previous state, IMO.

timreichen:
> I now think it is worth simply aligning with `std/assert`. The names are short and simple and show their purpose. That's all that's needed. I'm sorry for the indecisiveness.

Done

kt3k:
Let's settle on `greaterOrEqual` and `lessOrEqual` for `gte` and `lte` replacements, which now have much support from core team (@mmastrac @satyarohith @lucacasonato @crowlKats @dsherret are in favor of them).

I changed `less` and `greater` to `lessThan` and `greaterThan`. They are a bit inconsistent with `assertLess` and `assertGreater` of `std/assert`, but `less` & `greater` look too less descriptive to me. I believe this change is a reasonable compromise between consistency and descriptiveness.

iuioiua (approved):
LGTM

kt3k (approved):
LGTM

PR #4084, opened 2024-01-04
refactor(testing): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the testing module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/testing/bdd_test.ts
+++ b/testing/bdd_test.ts
@@ -118,9 +118,9 @@ Deno.test("global", async (t) => {
   }
 
   await t.step("global hooks", async () => {
-    const test = stub(Deno, "test"),
-      fns = [spy(), spy()],
-      { beforeAllFn, afterAllFn, beforeEachFn, afterEachFn } = hookFns();
+    const test = stub(Deno, "test");
+    const fns = [spy(), spy()] as const;
+    const { beforeAllFn, afterAllFn, beforeEachFn, afterEachFn } = hookFns();
 
     const context = new TestContext("global");
     try {
--- a/testing/bdd_test.ts
+++ b/testing/bdd_test.ts
@@ -1633,7 +1633,7 @@ Deno.test("global", async (t) => {
         "nested",
         async () => {
           const test = stub(Deno, "test");
-          const fns = [spy(), spy()] as readonly [Spy, Spy];
+          const fns = [spy(), spy()] as const;
           const { beforeAllFn, afterAllFn, beforeEachFn, afterEachFn } =
             hookFns();
 
--- a/testing/bdd_test.ts
+++ b/testing/bdd_test.ts
@@ -130,15 +130,15 @@ Deno.test("global", async (t) => {
       beforeEach(beforeEachFn);
       afterEach(afterEachFn);
 
-      assertEquals(it({ name: "example 1", fn: fns[0] }), undefined);
-      assertEquals(it({ name: "example 2", fn: fns[1] }), undefined);
+      assertEquals(it({ name: "example 1", fn: fns[0]! }), undefined);
+      assertEquals(it({ name: "example 2", fn: fns[1]! }), undefined);
 
-      assertSpyCalls(fns[0], 0);
-      assertSpyCalls(fns[1], 0);
+      assertSpyCalls(fns[0]!, 0);
+      assertSpyCalls(fns[1]!, 0);
 
       assertSpyCall(test, 0);
       const call = test.calls[0];
-      const options = call.args[0] as Deno.TestDefinition;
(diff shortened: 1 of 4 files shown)


testing/mock.ts
```diff
@@ -1074,8 +1082,8 @@ export function returnsThis<
 // deno-lint-ignore no-explicit-any
 export function returnsArg<Arg, Self = any>(
   idx: number,
-): (this: Self, ...args: Arg[]) => Arg {
-  return function (...args: Arg[]): Arg {
+): (this: Self, ...args: Arg[]) => Arg | undefined {
+  return function (...args: Arg[]): Arg | undefined {
```
syhol, line 1086: This is a change to the public API.

Its always been the case that this could return `undefined` if the index is outside the array bounds, now the type just reflects reality. Alternativly I could add a check and throw which would keep the type consistent but change the runtime behaviour.

testing/bdd_test.ts
```diff
@@ -130,15 +130,15 @@ Deno.test("global", async (t) => {
       beforeEach(beforeEachFn);
       afterEach(afterEachFn);
 
-      assertEquals(it({ name: "example 1", fn: fns[0] }), undefined);
-      assertEquals(it({ name: "example 2", fn: fns[1] }), undefined);
+      assertEquals(it({ name: "example 1", fn: fns[0]! }), undefined);
```
syhol, line 133: There are loads of cases like this in this file. At one point I tried to be clever about it and define `fns` as `[Spy, Spy]` but it got messy very quickly. Ultimatly I decided that "worse is better" and just threw non-null assertions everywhere.

testing/mock.ts
iuioiua: I think throwing if the index is outside of the argument bounds is a better idea.

testing/mock.ts
```diff
@@ -349,7 +349,7 @@ const sessions: Set<Spy<any, any[], any>>[] = [];
 // deno-lint-ignore no-explicit-any
 function getSession(): Set<Spy<any, any[], any>> {
   if (sessions.length === 0) sessions.push(new Set());
-  return sessions[sessions.length - 1];
+  return sessions[sessions.length - 1]!;
```
iuioiua, line 352:
```suggestion
  return sessions.at(-1)!;
```
Unrelated but might as well do.

testing/bdd_test.ts
iuioiua: Check out what I did in https://github.com/denoland/deno_std/pull/4084/commits/d412a7c41aedaad0b43e45662498b7d9f7609537. It should remove the need of all these non-null assertions. `as const` is powerful.

iuioiua (changes requested):

testing/mock.ts
syhol: Done

testing/bdd_test.ts
syhol: Done. Updated all cases in mock_test.ts and bdd_test.ts to use this instead. Worth noting that bdd_test.ts uses a few callbacks, so those function types needed to be updated from `cb: (fns: Spy[]) => void,` to `cb: (fns: readonly [Spy, Spy, Spy]) => void`.

syhol:
All comments addressed, ready for another review 🙇

testing/mock_test.ts
```diff
@@ -1761,12 +1782,12 @@ Deno.test("returnsThis", () => {
...
   let callback = returnsArg(0);
-  assertEquals(callback(), undefined);
+  assertThrows(() => callback());
   assertEquals(callback("a"), "a");
   assertEquals(callback("b", "c"), "b");
   callback = returnsArg(1);
-  assertEquals(callback(), undefined);
-  assertEquals(callback("a"), undefined);
+  assertThrows(() => callback());
+  assertThrows(() => callback("a"));
```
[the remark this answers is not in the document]
syhol: Following on from this conversation https://github.com/denoland/deno_std/pull/4084#discussion_r1441182192 the choice was either changing the type or changing the runtime behaviour, you settled on changing the runtime behaviour so the tests needed to change to account for the new behaviour.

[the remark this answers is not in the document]
iuioiua: Oh! I see. Hm... Ok. I now better understand how it affected the public API. On second thought, it might be better to add `undefined` to the return type instead of error-checking. Sorry, are you able to do that?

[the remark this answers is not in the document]
syhol: No problem, I could have made it more clear. I'll push the change soon.

iuioiua (approved):
LGTM! Thank you for this!

PR #4086, opened 2024-01-04
refactor(assert): cleanup `_diff.ts`

iuioiua:
This aims to make some of the logic a little easier to understand.


assert/_diff.ts
```diff
@@ -16,18 +16,14 @@ interface FarthestPoint {
...
 
-export const DiffType = {
-  removed: "removed",
-  common: "common",
-  added: "added",
-} as const;
+export const DIFF_TYPES = ["added", "removed", "common"] as const;
 
-export type DiffType = keyof typeof DiffType;
+export type DiffType = typeof DIFF_TYPES[number];
```
kt3k, line 21: How about directly define this as `"added" | "removed" | "common"` and removing `DIFF_TYPES`?

iuioiua: Good idea. Done.

iuioiua:
Actually, closing as this undoes some of the work from #4121. I might approach this later.

PR #4089, opened 2024-01-04
fix(semver): remove `undefined` return type from `rangeMax()`

timreichen:
- removes obsolete `undefined` from `rangeMax()` return type


iuioiua (approved):
LGTM

PR #4090, opened 2024-01-04
deprecation(semver): rename `rangeFormat()` to `formatRange()`

timreichen:
- renames `rangeFormat()` to `formatRange()`

### DIFF
--- a/semver/range_format.ts
+++ b/semver/range_format.ts
@@ -1,14 +1,13 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
-import type { SemVerRange } from "./types.ts";
-import { comparatorFormat } from "./comparator_format.ts";
+
+import { formatRange } from "./format_range.ts";
 
 /**
  * Formats the range into a string
  * @example >=0.0.0 || <1.0.0
  * @param range The range to format
  * @returns A string representation of the range
+ *
+ * @deprecated (will be removed after 0.213.0) Use {@linkcode formatRange} instead.
  */
-export function rangeFormat(range: SemVerRange): string {
-  return range.ranges.map((c) => c.map((c) => comparatorFormat(c)).join(" "))
-    .join("||");
-}
+export const rangeFormat = formatRange;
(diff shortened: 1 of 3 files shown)


semver/range_test.ts
iuioiua, line 1: Should this file be renamed to `format_range_test.ts`?

iuioiua (changes requested):
Makes sense

kt3k:
I don't see the reason of the deprecation. Is this discussed somewhere?

timreichen:
> I don't see the reason of the deprecation. Is this discussed somewhere?

Javascript generally has a verb-first convention for functions. It is highlighted in the https://github.com/denoland/deno_std/issues/3948#issue-2039604547. description.

semver/range_format.ts
```diff
@@ -1,14 +1,13 @@
...
  * @param range The range to format
  * @returns A string representation of the range
+ *
+ * @deprecated (will be removed after 0.213.0) Use {@linkcode formatRange} instead.
  */
-export function rangeFormat(range: SemVerRange): string {
-  return range.ranges.map((c) => c.map((c) => comparatorFormat(c)).join(" "))
-    .join("||");
-}
+export const rangeFormat = formatRange;
```
kt3k, line 13: Please use the below structure to make sure deprecation note appear in IDE:

```ts
/**
 * Formats the range into a string
 * @example >=0.0.0 || <1.0.0
 * @param range The range to format
 * @returns A string representation of the range
 *
 * @deprecated (will be removed after 0.213.0) Use {@linkcode formatRange} instead.
 */
export function rangeFormat(range: SemVerRange): string {
  return formatRange(range);
}
```

kt3k:
> Javascript generally has a verb-first convention for functions. It is highlighted in the https://github.com/denoland/deno_std/issues/3948#issue-2039604547 description.

Fair enough.

semver/range_format.ts
timreichen: done

semver/range_test.ts
timreichen: It could be split into `test_range_test.ts` and the last test case would be `format_range_test.ts`. If so, let's do in another PR.

iuioiua: I think we should have a `format_range_test.ts` with the appropriate tests to accompany the new `format_range.ts`.

iuioiua (approved):
I've added a `format_range_test.ts` file. Now, LGMT. Thank you.

kt3k (approved):
LGTM. Thanks!

PR #4092, opened 2024-01-04
refactor(cli): make `Spinner.color` public

timreichen:
- makes `Spinner.color` prop public.

### DIFF
--- a/cli/spinner.ts
+++ b/cli/spinner.ts
@@ -91,9 +91,11 @@ export class Spinner {
     this.#interval = options?.interval ?? DEFAULT_INTERVAL;
     this.color = options?.color;
   }
+
   set color(value: Color | undefined) {
     this.#color = value ? COLORS[value] : undefined;
   }
+
   get color() {
     return this.#color;
   }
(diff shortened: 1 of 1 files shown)


cli/spinner.ts
```diff
@@ -83,7 +83,13 @@ export class Spinner {
...
     this.#message = options?.message ?? "";
     this.#interval = options?.interval ?? DEFAULT_INTERVAL;
-    this.#color = options?.color ? COLORS[options.color] : undefined;
+    this.color = options?.color;
+  }
+  set color(value: Color | undefined) {
+    this.#color = value ? COLORS[value] : undefined;
+  }
+  get color() {
+    return this.#color;
```
iuioiua, line 92:
```suggestion
  }
  
  set color(value?: Color) {
    this.#color = value ? COLORS[value] : undefined;
  }
  
  get color() {
    return this.#color;
```
Nits

iuioiua (changes requested):
Can we add a short message saying this can be changed on the fly, like the `message` property?

cli/spinner.ts
timreichen: A 'set' accessor cannot have an optional parameter.

iuioiua: Oh. Makes sense in hindsight...

iuioiua (approved):
LGTM

kt3k (approved):
LGTM

PR #4093, opened 2024-01-04
refactor(cli): make `Spinner.interval` public

timreichen:
- makes `Spinner.color` prop public and ensures the rendering is started again with the new interval while keeping the current frame index if spinner is active.

### DIFF
--- a/cli/spinner.ts
+++ b/cli/spinner.ts
@@ -68,6 +68,7 @@ export class Spinner {
   #color?: Color;
   #intervalId: number | undefined;
   #active = false;
+  #frameIndex = 0;
 
   /**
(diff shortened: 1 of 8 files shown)


iuioiua:
On second thought, I can't think of many instances of seeing a spinner change its speed while active... @kt3k, do you like the idea of this change?

kt3k:
I'm not in favor of the idea as I can't imagine the use case

iuioiua:
Alright. I'll close this without merging. Sorry we didn't think of this prior, Tim. Either way, thank you. We can use this PR if a feature request for this comes up in the future.

PR #4094, opened 2024-01-04
refactor(semver): make `rangeMin()` and `rangeMax()` more readable

timreichen:
- replaces .sort, .filter and .map with loops while making the code more readable.
- reduces overhead


iuioiua (approved):
LGTM. Thanks!

PR #4095, opened 2024-01-04
refactor(semver): make `outside()` more readable

timreichen:
- makes `outside()` more readable by replacing the algorithm with already existing functions.

As pointed out in https://github.com/denoland/deno_std/issues/3948#issuecomment-1876875415, this seems now to be a weird function to include in the semver mod.


iuioiua (commented):
Are you able to add the test case mentioned in https://github.com/denoland/deno_std/issues/3948#issuecomment-1876875415 ?

timreichen:
> Are you able to add the test case mentioned in [#3948 (comment)](https://github.com/denoland/deno_std/issues/3948#issuecomment-1876875415) ?

Done.

cli/parse_args.ts
iuioiua, line 1: I suspect these changes have been mistakenly added.

timreichen: Sorry, that wasn't supposed to be in that PR. Reverted

timreichen:
Actually, would it make sense to additionally deprecate `outside()` and link to `!testRange()`, `gtr` and `ltr`? These are the three use cases `outside()` handles.

iuioiua (approved):
LGTM, as this removes the dependency on `outside()`, which _might_ be worth deprecating.

Please see the comment on the new test that I've added. We should relocate, then tweak this test if the reported behaviour is agreed upon to be a bug (which both Tim and I already do).

kt3k (approved):
LGTM

PR #4097, opened 2024-01-04
refactor(semver): reduce `minSatisfying()` and `maxSatisfying()` overhead

timreichen:
- reduces `minSatisfying()` and `maxSatisfying()` array overhead.


iuioiua:
Can you please clarify what you mean by reducing overhead?

timreichen:
> Can you please clarify what you mean by reducing overhead?
I mean memory overhead. The original code unnecessarily creates an array with `.filter()` which then gets sorted and one element gets popped instead of just storing one element and comparing against that.

iuioiua (approved):
Thanks for clarifying. LGTM.

PR #4098, opened 2024-01-04
deprecation(semver): deprecate `sort()`

timreichen:
- deprecates `sort()`, which just calls `.sort(compare)` on the array. 

**Note**: This is not discussed but seemed obvious. The User should rather import `compare()` and call sort directly on the array.

### DIFF
--- a/semver/sort.ts
+++ b/semver/sort.ts
@@ -5,7 +5,7 @@ import { compare } from "./compare.ts";
 /**
  * Sorts a list of semantic versions in ascending order.
  *
- * @deprecated (will be removed after 0.213.0) Use `list.sort(compare)` with {@linkcode compare} instead.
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
I agree with this. LGTM.

kt3k (approved):
LGTM

semver/sort.ts
```diff
@@ -2,7 +2,11 @@
 import type { SemVer } from "./types.ts";
 import { compare } from "./compare.ts";
 
-/** Sorts a list of semantic versions in ascending order. */
+/**
+ * Sorts a list of semantic versions in ascending order.
+ *
+ * @deprecated (will be removed after 0.213.0) Use `list.sort(compare)` with {@linkcode compare} instead.
```
kt3k, line 8:
```suggestion
 * @deprecated (will be removed after 0.214.0) Use `list.sort(compare)` with {@linkcode compare} instead.
```

PR #4100, opened 2024-01-04
refactor(log): remove obsolete `Log.handler` property getter and setter

timreichen:
- removes obsolete `Log.handler` property getter and setter and make `#handler` property public.


iuioiua (approved):
LGTM. Thank you.

PR #4110, opened 2024-01-05
refactor(archive): prepare for noUncheckedIndexedAccess

syhol:
Handling all noUncheckedIndexedAccess issues for the archive module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/archive/untar.ts
+++ b/archive/untar.ts
@@ -34,6 +34,7 @@ import {
   HEADER_LENGTH,
   readBlock,
   type TarMeta,
+  UstarFields,
   ustarStructure,
 } from "./_common.ts";
 import { readAll } from "../streams/read_all.ts";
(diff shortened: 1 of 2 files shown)


archive/untar.ts
```diff
@@ -222,7 +223,7 @@ export class Untar {
         // Ignore checksum header
         continue;
       }
-      sum += header[i];
+      sum += header[i]!;
```
syhol, line 226: Since `i` gets no larger than HEADER_LENGTH and header is a Uint8Array with the size of HEADER_LENGTH this should never be out of bounds

archive/untar.ts
```diff
@@ -47,9 +48,9 @@ export interface TarMetaWithLinkName extends TarMeta {
   linkName?: string;
 }
 
-export interface TarHeader {
-  [key: string]: Uint8Array;
-}
+export type TarHeader = {
+  [key in UstarFields]: Uint8Array;
+};
```
syhol, line 53: If you don't want this public type to have this slight change, we could look at creating a new non-exported type and casting the header to this new type inside getMetadata: https://github.com/denoland/deno_std/blob/b64357605cf55edc2fdd278e057d7f0b9f665195/archive/untar.ts#L256-L293

iuioiua (approved):
LGTM! Thank you. @crowlKats, can you PTAL? This touches on a few points relating to the upcoming streams variations.

PR #4112, opened 2024-01-05
chore: ignore `/docs` folder

iuioiua:
This change ensures the `/docs` directory is ignored in appropriate checks, etc. This is so nothing is ever accidentally committed when working with `deno doc --html`.


kt3k (approved):
LGTM. Thanks!

PR #4114, opened 2024-01-05
deprecation(json): deprecate `writableStrategy` and `readableStrategy` constructor options

### LINKED ISSUE #4113: suggestion: deprecate `writableStrategy` and `readableStrategy` argument properties for `std/json` constructors
The `TransformStream`s in `std/json` are the only ones with `writableStrategy` and `readableStrategy` argument properties in their constructors. There are many other `TransformStream`s in the Standard Library, but none of them has these options, nor do they seem to need to. The `std/json` classes don't seem to be part of a special case either that justifies these properties. I think these options should be deprecated to trim needless moving parts.

iuioiua:
Closes #4113


iuioiua:
I think this PR should be merged, but I'll close for now until we decide on #4114.

Issue #4099, opened 2024-01-04
suggestion(log): Logger `warn()` method instead of `warning()`


timreichen:
Shouldn't the `warning()` method  in `Logger` be actually called `warn()` like in `console.warn()`? 

**Describe the solution you'd like**

I suggest deprecating `warning()` and rename it to `warn()` in order to align to `console`.

joehillen:
Yes, please. This has tripped me up as well.

iuioiua:
PRs are welcome.

timreichen:
That begs the question if the `WARNING` log level should be `WARN` instead too.

iuioiua:
Completed in #4117 and #4170. Thank you for the suggestion.

PR #4117, opened 2024-01-05
deprecation(log): rename `warning()` to `warn()`

timreichen:
ref: https://github.com/denoland/deno_std/issues/4099


log/test.ts
```diff
@@ -23,7 +23,7 @@ Deno.test("defaultHandlers", async function () {
   } = {
     DEBUG: log.debug,
     INFO: log.info,
-    WARNING: log.warning,
+    WARNING: log.warn,
```
joehillen, line 26: `WARNING` should be changed to `WARN`

log/logger.ts
```diff
@@ -179,11 +179,21 @@ export class Logger {
     return this.#_log(LogLevels.INFO, msg, ...args);
   }
 
+  /** @deprecated (will be removed after 0.214.0) Use {@linkcode warn} instead. */
   warning<T>(msg: () => T, ...args: unknown[]): T | undefined;
   warning<T>(msg: T extends GenericFunction ? never : T, ...args: unknown[]): T;
```
iuioiua, line 184:
```suggestion
  /** @deprecated (will be removed after 0.214.0) Use {@linkcode warn} instead. */
  warning<T>(msg: T extends GenericFunction ? never : T, ...args: unknown[]): T;
```
So both overloads show as deprecated.

log/logger.ts
```diff
@@ -179,11 +179,21 @@ export class Logger {
...
+    return this.warn(msg as () => T, ...args);
+  }
+
+  warn<T>(msg: () => T, ...args: unknown[]): T | undefined;
+  warn<T>(msg: T extends GenericFunction ? never : T, ...args: unknown[]): T;
+  warn<T>(
+    msg: (T extends GenericFunction ? never : T) | (() => T),
+    ...args: unknown[]
   ): T | undefined {
     return this.#_log(LogLevels.WARNING, msg, ...args);
```
iuioiua, line 198:
```suggestion
    return this.#_log(LogLevels.WARN, msg, ...args);
```

log/mod.ts
```diff
@@ -498,7 +498,11 @@ export function info<T>(
...
 }
 
-/** Log with warning level, using default logger. */
+/**
+ * @deprecated (will be removed after 0.214.0) Use {@linkcode warn} instead.
+ *
+ * Log with warning level, using default logger.
+ */
 export function warning<T>(msg: () => T, ...args: unknown[]): T | undefined;
 export function warning<T>(
```
iuioiua, line 507:
```suggestion
/**
 * @deprecated (will be removed after 0.214.0) Use {@linkcode warn} instead.
 *
 * Log with warning level, using default logger.
 */
export function warning<T>(
```

log/mod.ts
```diff
@@ -510,12 +514,30 @@ export function warning<T>(
...
+  if (msg instanceof Function) {
+    return getLogger("default").warn(msg, ...args);
   }
-  return getLogger("default").warning(msg, ...args);
+  return getLogger("default").warn(msg, ...args);
 }
 
-/** Log with error level, using default logger. */
+/** Log with e
+ * rror level, using default logger. */
```
iuioiua, line 540: Please revert

log/test.ts
iuioiua: I agree

iuioiua (changes requested):
Nice to see this done! Few nits.

log/test.ts
timreichen: done

log/mod.ts
timreichen: done

iuioiua (approved):
LGTM! Nice.

Am I correct in understanding that this change is non-breaking?

kt3k:
> Am I correct in understanding that this change is non-breaking?

Strictly speaking, if the users do any log analysis using `WARNING` symbol, then that analysis will break. Can we delay the change of output string from `WARNING` to `WARN` to the actual breaking version? (`after 0.214.0`)

timreichen:
> > Am I correct in understanding that this change is non-breaking?
> 
> Strictly speaking, if the users do any log analysis using `WARNING` symbol, then that analysis will break. Can we delay the change of output string from `WARNING` to `WARN` to the actual breaking version? (`after 0.214.0`)

Reverted

kt3k:
@timreichen I don't see well what part is reverted?

The below still seems printing `WARN hi`
```js
import * as log from "./log/mod.ts";
log.warning("hi");
```

timreichen:
> @timreichen I don't see well what part is reverted?
> 
> The below still seems printing `WARN hi`
> 
> ```js
> import * as log from "./log/mod.ts";
> log.warning("hi");
> ```

Sorry, seems like the commit wasn't pushed. Done now.

kt3k (approved):
LGTM

iuioiua (approved):
LGTM. Thank you.

PR #4119, opened 2024-01-05
refactor(bytes): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the bytes module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/bytes/equals_test.ts
+++ b/bytes/equals_test.ts
@@ -19,7 +19,7 @@ Deno.test("equals() handles randomized testing", () => {
     const arr3 = arr1.slice(0);
     // the chance of arr1 equaling arr2 is basically 0
     // but introduce an inequality at the end just in case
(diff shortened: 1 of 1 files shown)


bytes/equals_test.ts
```diff
@@ -19,7 +19,7 @@ Deno.test("equals() handles randomized testing", () => {
     const arr3 = arr1.slice(0);
     // the chance of arr1 equaling arr2 is basically 0
     // but introduce an inequality at the end just in case
-    arr2[arr2.length - 1] = arr1[arr1.length - 1] ^ 1;
+    arr2[arr2.length - 1] = arr1.at(-1)! ^ 1;
```
syhol, line 22: `arr1` is a `Uint8Array` with a length between 995 and 1004, so it will never be empty and `.at(-1)` will always be in bounds.

iuioiua (approved):
LGTM. Nice!

PR #4121, opened 2024-01-05
refactor(cli): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the cli module, tracked in https://github.com/denoland/deno_std/issues/4040


cli/parse_args.ts
```diff
@@ -567,7 +568,7 @@ export function parseArgs<
       o = get(o, key) as NestedMapping;
     });
 
-    const key = keys[keys.length - 1];
+    const key = keys.at(-1)!;
```
syhol, line 571: `keys` is the result of `name.split(".");` and that should never be empty, so can assume non-null

cli/parse_args.ts
```diff
@@ -660,50 +663,51 @@ export function parseArgs<
...
           broken = true;
           break;
         } else {
-          setArg(letters[j], get(flags.strings, letters[j]) ? "" : true, arg);
+          setArg(letter, get(flags.strings, letter) ? "" : true, arg);
         }
       }
 
-      const [key] = arg.slice(-1);
+      const key = arg.at(-1)!;
```
syhol, line 698: This code is run inside the check `/^-[^-]+/.test(arg)` which means arg needs to be a non-empty string.

cli/prompt_secret.ts
```diff
@@ -73,7 +73,7 @@ function readLineFromStdinSync(callback?: (n: number) => void): string {
     if (c[0] === BS || c[0] === DEL) {
       buf.pop();
     } else {
-      buf.push(c[0]);
+      buf.push(c[0]!);
```
syhol, line 76: c is a Uint8Array of size 1, so `c[0]` will always be non-null.

cli/parse_args.ts
```diff
@@ -371,8 +371,8 @@ function hasKey(obj: NestedMapping, keys: string[]): boolean {
     o = (get(o, key) ?? {}) as NestedMapping;
   });
 
-  const key = keys[keys.length - 1];
-  return hasOwn(o, key);
+  const key = keys.at(-1);
+  return typeof key !== "undefined" && hasOwn(o, key);
```
iuioiua, line 375:
```suggestion
  return key !== undefined && hasOwn(o, key);
```
Ditto for other instances of checking if a variable doesn't equal `undefined`.

cli/parse_args.ts
```diff
@@ -618,11 +619,13 @@ export function parseArgs<
 
   for (let i = 0; i < args.length; i++) {
     const arg = args[i];
+    assert(typeof arg !== "undefined");
```
iuioiua, line 622:
```suggestion
    assertEquals(arg, undefined);
```
As far as I can see, ditto to other instances of `assert()`.

iuioiua (changes requested):

cli/parse_args.ts
syhol: In order to do this and get proper type narrowing, I need to update the assert functions:

from:
```typescript
export function assertNotEquals<T>(actual: T, expected: T, msg?: string): void { /**/ }
```
to
```typescript
export function assertNotEquals<T, E>(
  actual: T,
  expected: E,
  msg?: string,
): asserts actual is Exclude<T, E> { /**/ }
```

Are you happy for me to make this change? Would you like it in this PR or another?

By the way, I could make E extend T (`<T, E extends T>`) but without noUncheckedIndexedAccess we would get errors when writing `assertNotEquals(m[1], undefined);` because undefined wouldn't be part of the type.

syhol: We could also consider an `assertDefined`:

```
export function assertDefined<T>(
  actual: T,
  msg?: string,
): asserts actual is Exclude<T, null | undefined> { /**/ }
```

Either checking for `null` and `undefined` or just `undefined`.

iuioiua: Wow. Crazy that `assertNotEquals()` isn't an assertion signature. Yes please, a separate PR for this would be great!

syhol: On further investigation, I don't think we need `assertDefined` as it seems we already have `assertExists` that checks for null and undefined. I'll use that for this PR.

As for adding assertion signatures to other assertions, I'll still create the PR but I've discovered that its required to do `<T, E extends T>`. This should be a backwards compatible change. Also it shouldn't be a problem for this PR because we can just use `assertExists` instead.

syhol: Updated this PR to use assertExists

iuioiua (approved):
I've made just a couple of tweaks. Now, LGTM! Thank you, Simon.

PR #4124, opened 2024-01-06
refactor(assert): remove `void` return types

iuioiua:
Deno sees `void` return types are unncessary boilerplate.


kt3k (approved):
LGTM

ref: https://github.com/denoland/deno/issues/8832

PR #4125, opened 2024-01-06
fix(http): punt removal version of `unstable_cookie_map.ts`

iuioiua:
This is so `cookie_map.ts` and `unstable_cookie_map.ts` can be removed at the same time. This should prevent any confusion that would otherwise arise from both files being removed at different times.


kt3k (approved):
LGTM

PR #4127, opened 2024-01-07
feat(collections): pass `key` to `mapValues()` transformer

ratson:
Pass `key` parameter to `mapValues()` transformer to support transformed values that depends on `key`.

### DIFF
--- a/collections/map_values.ts
+++ b/collections/map_values.ts
@@ -28,7 +28,7 @@
  */
 export function mapValues<T, O, K extends string>(
   record: Readonly<Record<K, T>>,
-  transformer: (value: T) => O,
+  transformer: (value: T, key: K) => O,
 ): Record<K, O>;
 /**
  * Applies the given transformer to all values in the given record and returns a
@@ -57,20 +57,19 @@ export function mapValues<T, O, K extends string>(
  */
 export function mapValues<T, O, K extends string>(
   record: Readonly<Partial<Record<K, T>>>,
-  transformer: (value: T) => O,
+  transformer: (value: T, key: K) => O,
 ): Partial<Record<K, O>>;
 export function mapValues<T, O, K extends string>(
   record: Record<K, T>,
-  transformer: (value: T) => O,
+  transformer: (value: T, key: K) => O,
   // deno-lint-ignore no-explicit-any
 ): any {
   // deno-lint-ignore no-explicit-any
   const ret: any = {};
-  const entries = Object.entries(record);
+  const entries = Object.entries<T>(record);
 
   for (const [key, value] of entries) {
-    // deno-lint-ignore no-explicit-any
-    const mappedValue = transformer(value as any);
+    const mappedValue = transformer(value, key as K);
 
     ret[key] = mappedValue;
   }
--- a/collections/map_values_test.ts
+++ b/collections/map_values_test.ts
@@ -139,3 +139,17 @@ Deno.test({
     assertEquals(actual, expected);
   },
 });
+
+Deno.test({
+  name: "mapValues() pass key to transformer",
+  fn() {
+    const key = "key";
+    const actual = mapValues(
+      { [key]: "value" },
+      (_, k) => k,
+    );
+    const expected = { [key]: key };
+
+    assertEquals(actual, expected);
+  },
+});


kt3k:
I'm not sure this is a useful addition as this can be covered by the below structure:
```ts
mapEntries(obj, ([key, val]) => [key, transform(val, key)]);
```

@ratson What do you think?

Also does anyone have any opinion on this?

ratson:
@kt3k The code works, but it is harder to grab the intent that only values are transformed and the keys is not going to change.

timreichen:
To me this looks like this is a very special use case. If I remember correctly transformers have generally one argument, which I like.
This particular use case is also solvable natively in one line:
```ts
Object.fromEntries(Object.entries(obj).map(([key, value]) => [key, key])
```
My opinion is not strong,  but if this was implemented, we would should check what else would need changes to be consistent, like  adding a `value` argument to the `mapKeys()` transformer for keys that depend on values.

iuioiua:
I'm not in favour of this addition too. As Yoshiya said, the same can be accomplished using [`mapEntries()`](https://deno.land/std@0.211.0/collections/mod.ts?s=mapEntries). Also, applying this change would set a precedent for other similar functions in the `std/collections` sub-module in the name of consistency. E.g. `filterValues()`.

ratson:
@timreichen From my experience, it is not uncommon to require accessing the `key` value when mapping over the values.
Is there any practical advantage to limiting only one argument?

@iuioiua I don't think `filterValues()` should follow the same signature, as it does not make sense to operate on keys for the function, there is `filterKeys()` for the purpose.

FYR, Kotlin use different signatures for them,

https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/map-values.html
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/filter-values.html

I have been using `lodash.mapValues` for long, and often find the `key` argument is useful.

@kt3k  Here is a benchmark result in case you wonder performance difference of different approach.

```bash
deno bench https://github.com/ratson/bug-report/raw/deno/map-values-bench/map_values_bench.ts
```

```
benchmark             time (avg)        iter/s             (min … max)       p75       p99      p995
---------------------------------------------------------------------- -----------------------------

group size = 1
mapValues             192.3 ns/iter   5,200,219.9 (187.75 ns … 215.19 ns) 193.89 ns 210.02 ns 212.36 ns
mapEntries           195.49 ns/iter   5,115,451.8 (191.17 ns … 231.24 ns)    197 ns 205.12 ns 207.22 ns
Object.entries       206.53 ns/iter   4,841,885.4 (200.44 ns … 231.75 ns) 208.27 ns 225.12 ns 228.71 ns
lodash.mapValues      194.7 ns/iter   5,136,088.6   (186.3 ns … 216.2 ns) 196.36 ns 206.67 ns 211.86 ns

summary
  mapValues
   1.01x faster than lodash.mapValues
   1.02x faster than mapEntries
   1.07x faster than Object.entries

group size = 10
mapValues            673.79 ns/iter   1,484,145.3 (661.87 ns … 710.51 ns) 676.69 ns 710.51 ns 710.51 ns
mapEntries           768.07 ns/iter   1,301,972.0  (748.6 ns … 825.77 ns) 778.03 ns 825.77 ns 825.77 ns
Object.entries         1.01 µs/iter     992,397.9   (949.85 ns … 1.84 µs)      1 µs   1.84 µs   1.84 µs
lodash.mapValues     473.73 ns/iter   2,110,907.8 (462.05 ns … 648.43 ns) 475.28 ns 517.83 ns 648.43 ns

summary
  mapValues
   1.42x slower than lodash.mapValues
   1.14x faster than mapEntries
   1.5x faster than Object.entries

group size = 100
mapValues              5.49 µs/iter     182,281.7     (5.46 µs … 5.56 µs)   5.49 µs   5.56 µs   5.56 µs
mapEntries             6.39 µs/iter     156,545.5     (6.09 µs … 6.74 µs)   6.49 µs   6.74 µs   6.74 µs
Object.entries         8.77 µs/iter     114,038.1   (7.78 µs … 171.22 µs)   8.73 µs  12.58 µs  20.07 µs
lodash.mapValues       3.44 µs/iter     291,054.1     (3.41 µs … 3.62 µs)   3.44 µs   3.62 µs   3.62 µs

summary
  mapValues
   1.6x slower than lodash.mapValues
   1.16x faster than mapEntries
   1.6x faster than Object.entries

group size = 1000
mapValues             53.68 µs/iter      18,627.5  (47.91 µs … 262.28 µs)  52.85 µs  74.45 µs 193.22 µs
mapEntries            62.49 µs/iter      16,002.0  (52.48 µs … 394.18 µs)   59.5 µs 197.75 µs 212.65 µs
Object.entries        87.95 µs/iter      11,369.5  (81.72 µs … 331.68 µs)  86.18 µs 242.81 µs  258.6 µs
lodash.mapValues      31.44 µs/iter      31,806.6   (28.7 µs … 227.23 µs)  31.16 µs  39.09 µs  126.3 µs

summary
  mapValues
   1.71x slower than lodash.mapValues
   1.16x faster than mapEntries
   1.64x faster than Object.entries

group size = 10000
mapValues            562.93 µs/iter       1,776.4     (494.3 µs … 1.1 ms) 525.39 µs   1.04 ms   1.06 ms
mapEntries           712.06 µs/iter       1,404.4   (558.35 µs … 1.22 ms)  975.1 µs   1.14 ms   1.16 ms
Object.entries       903.59 µs/iter       1,106.7   (819.82 µs … 1.33 ms) 868.71 µs   1.29 ms   1.29 ms
lodash.mapValues     315.17 µs/iter       3,172.9 (286.22 µs … 606.92 µs) 309.87 µs 541.41 µs 547.37 µs

summary
  mapValues
   1.79x slower than lodash.mapValues
   1.26x faster than mapEntries
   1.61x faster than Object.entries
```

timreichen:
> @timreichen From my experience, it is not uncommon to require accessing the `key` value when mapping over the values. Is there any practical advantage to limiting only one argument?

No, there is not. I was just cautious because native iterator callback functions  like `Iterator.prototype.map()` have `index` as their second argument and I was thinking this implementation would differ from common web api design.
But the `Map.prototype.forEach()` callback has `value` and then `key` as their arguments: `callbackfn: (value: any, key: any, map: Map<any, any>) => void`, so providing `key` after `value` seems pretty web api-ish.

kt3k:
@ratson Thanks for additional info and benchmarks.

I'm now in favor of accepting this because `key` is available in transform function in both kotlin's mapValues and lodash mapValues. Also the benchmark is better than the alternative constructions given above.

I'll keep this PR open for a while to see what community members think

iuioiua:
> @ratson Thanks for additional info and benchmarks.
> 
> I'm now in favor of accepting this because `key` is available in transform function in both kotlin's mapValues and lodash mapValues. Also the benchmark is better than the alternative constructions given above.
> 
> I'll keep this PR open for a while to see what community members think

I'm now in favour, too, for the same reasons.

iuioiua (approved):
LGTM

kt3k (approved):
LGTM. Thanks for your suggestion! @ratson

PR #4128, opened 2024-01-08
feat(io): re-introduce IO functions

### LINKED ISSUE #4120: suggestion: re-consider the deprecation of some IO-related APIs
`std/io` is being deprecated because the `Deno.Reader` interface and friends are being deprecated from the runtime. However, their `read()`, `write()`, and other methods on I/O-related APIs, such as `Deno.FsFile` and `Deno.Conn`, are here to stay. These methods staying around justify keeping some IO-related APIs within the Standard Library instead of deprecating them.

To keep the Standard Library and its maintenance lean, I suggest we keep APIs that allow converting `Reader`s and `Writer`s to streams. This will need some more review, but, at first glance, I think we should un-deprecate:
* `readAll()`
* `readableStreamFromReader()`
* `writeAll()`
* `writableStreamFromWriter()`

CC @bartlomieju, @piscisaureus and @kt3k

iuioiua:
This change re-introduces some of the previously deprecated and I/O-related functions and types. In short, these functions are moved from `std/streams` to `std/io`, which makes a lot more sense, in my opinion. These changes also include a few minor non-breaking (hopefully) tweaks.

The refreshed `std/io` API aims to be as basic and minimal as possible while still covering the basic I/O modalities such as reading (`readAll()`), writing (`writeAll()`) and transferring (`copy()`), as well as conversion utilities to the Streams API's interfaces for more advanced use. E.g. `TextLineStream()`. There's an argument to be made that storage (`Buffer`) should be included in these primary modalities, but users may be fine without it.

Closes #4120
Towards https://github.com/denoland/deno/issues/9795


archive/tar_test.ts
```diff
@@ -88,7 +88,7 @@ Deno.test("Tar() appends file with long name to tar archive", async function ():
   assert(result !== null);
   assert(!result.consumed);
   const untarText = new TextDecoder("utf-8").decode(await readAll(result));
-  assert(result.consumed);
+  // assert(result.consumed);
```
iuioiua, line 91: When enabled, this fails because `TarEntry.read()` doesn't change `TarEntry.#consumed` to `true` once fully read. I'm not sure why yet.

io/read_all.ts
```diff
@@ -0,0 +1,74 @@
...
+ * // Example from stdin
+ * const stdinContent = await readAll(Deno.stdin);
+ *
+ * // Example from file
+ * const file = await Deno.open("my_file.txt", {read: true});
+ * const myFileContent = await readAll(file);
+ * file.close();
+ * ```
+ */
+export async function readAll(reader: Reader): Promise<Uint8Array> {
```
iuioiua, line 25: This function differs from `readAll()` from `std/streams` in that it doesn't depend on `Buffer()` from `std/io`. It's simpler, only relying on `concat()`. This should be a non-breaking change.

io/read_all.ts
```diff
@@ -0,0 +1,74 @@
...
+import { DEFAULT_CHUNK_SIZE } from "./_constants.ts";
+import type { Reader, ReaderSync } from "./types.d.ts";
+
+/**
+ * Read {@linkcode Reader} `r` until EOF (`null`) and resolve to the content as
+ * {@linkcode Uint8Array}.
+ *
+ * @example
+ * ```ts
+ * import { readAll } from "https://deno.land/std@$STD_VERSION/io/read_all.ts";
```
iuioiua, line 14: Note: `Buffer()` is not imported here as it's deprecated.

io/read_all_test.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { assert, assertEquals } from "../assert/mod.ts";
+import { readAll, readAllSync } from "./read_all.ts";
+import { Buffer } from "./buffer.ts";
+import { init } from "./_test_common.ts";
+
+Deno.test("readAll()", async () => {
+  const testBytes = init();
+  assert(testBytes);
+  const reader = new Buffer(testBytes.buffer);
```
iuioiua, line 11: We rely on `Buffer()` here but should work on removing that dependency in the future (in a PR).

io/to_readable_stream.ts
```diff
@@ -0,0 +1,75 @@
...
+ *
+ * @example
+ * ```ts
+ * import { toReadableStream } from "https://deno.land/std@$STD_VERSION/io/to_readable_stream.ts";
+ *
+ * const file = await Deno.open("./file.txt", { read: true });
+ * const fileStream = toReadableStream(file);
+ * ```
+ */
+export function toReadableStream(
```
iuioiua, line 41: Note: renamed from `readableStreamfFromReader()`.

io/to_writable_stream.ts
```diff
@@ -0,0 +1,58 @@
...
+ * ```ts
+ * import { toWritableStream } from "https://deno.land/std@$STD_VERSION/io/to_writable_stream.ts";
+ *
+ * const file = await Deno.open("./file.txt", { create: true, write: true });
+ * await ReadableStream.from("Hello World")
+ *   .pipeThrough(new TextEncoderStream())
+ *   .pipeTo(toWritableStream(file));
+ * ```
+ */
+export function toWritableStream(
```
iuioiua, line 32: Note: renamed from `writableStreamFromWriter()`.

archive/tar_test.ts
kt3k: maybe we don't need to touch `std/archive` in this PR as it's re-designed completely in #1985?

iuioiua: We'll have to fix this if we're going to merge this PR before legacy `std/archive` is removed.

iuioiua: Fixed.

streams/copy.ts
```diff
@@ -22,8 +22,7 @@ import type { Reader, Writer } from "../io/types.ts";
  * @param dst The destination to copy to
  * @param options Can be used to tune size of the buffer. Default size is 32kB
  *
- * @deprecated (will be removed after 1.0.0) Use
- * {@linkcode ReadableStream.pipeTo} instead.
+ * @deprecated (will be removed in 0.214.0) Import from {@link https://deno.land/std/io/copy.ts} instead.
  */
 export async function copy(
```
iuioiua, line 27: Note: a `copySync()` could be implemented. But there's never been any demand for it. I also wonder whether there's demand for `readAllSync()` and `writeAllSync()`.

We could also name this `copyAll()` to align it with `readAll()` and `writeAll()`.

kt3k (approved):
LGTM

PR #4131, opened 2024-01-08
deprecation(semver): deprecate `Comparator` functions

timreichen:
ref: https://github.com/denoland/deno_std/issues/4047

This PR deprecates all public `Comparator` functions.

 **Note**: Some of these functions are used in range functions, which means they should be made private (for example `is_comparator.ts` -> `_is_comparator.ts`) after the deprecation deadline or combined into one `_comparator.ts` file.

### DIFF
--- a/semver/comparator_format.ts
+++ b/semver/comparator_format.ts
@@ -7,6 +7,8 @@ import { format } from "./format.ts";
  * @example >=0.0.0
  * @param comparator
  * @returns A string representation of the comparator
+ *
+ * @deprecated (will be removed in 0.213.0) Use {@linkcode formatRange} instead.
  */
 export function comparatorFormat(comparator: Comparator): string {
   const { semver, operator } = comparator;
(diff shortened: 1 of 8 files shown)


semver/comparator_format.ts
```diff
@@ -7,6 +7,8 @@ import { format } from "./format.ts";
  * @example >=0.0.0
  * @param comparator
  * @returns A string representation of the comparator
+ *
+ * @deprecated (will be removed in 0.213.0) Use {@linkcode formatRange} instead.
```
kt3k, line 11:
```suggestion
 * @deprecated (will be removed in 0.213.0) Use {@linkcode SemVerRange} instead of {@linkcode Comparator} and {@linkcode formatRange} for formatting it.
```

kt3k:
`in 0.213.0` looks too short notice. (The users only see the deprecation in `0.212.0`)

timreichen:
> `in 0.213.0` looks too short notice. (The users only see the deprecation in `0.212.0`)

True, I bumped the note to  `0.214.0`

iuioiua (approved):
LGTM!

One thing I like to do when doing these deprecation PRs is to do a dry run of the removal. I.e. remove the proposed functions locally and ensure they don't have any unintended side effects in other parts of the API.

kt3k (approved):
LGTM

PR #4132, opened 2024-01-08
fix(fs/ensure_dir): allow links to directories 

joehillen:
`lstat()` will not consider a symlink to a directory as a valid directory. Any use case of `ensureDir()` should consider a symlink to a directory as a directory.

Suppose a symlink to a directory needs to be replaced/handled differently. In that case, it should be handled by the caller outside of `ensureDir()` as any user of `ensureDir()` simply wants a valid place to put/find files.

I don't consider this a breaking change because the current behavior will throw an exception if a link exists, which would be unexpected behavior rather than intended behavior.


iuioiua:
Please run `deno fmt`.

kt3k:
> Suppose a symlink to a directory needs to be replaced/handled differently. In that case, it should be handled by the caller outside of ensureDir() as any user of ensureDir() simply wants a valid place to put/find files.

I think this depends on the use case. Some users might want actual dir instead of linked dir, and because the current `ensureDir` work like that, some users can be dependent on that assumption.

I think this should be implemented as option instead of changing the default behavior.

joehillen:
> Some users might want actual dir instead of linked dir

I think there is some ambiguity over who the "user" is. I'll use these definitions:

- Developer of the application (the one calling `ensureDir`), let's call the "dev".
- The application user who is running the application, let's call "user".

The next question is, who is the intended "dev" calling `ensureDir`? I argue that, in general, the dev using `std` wants to be able to use the library for simple tasks and not worry about the details. For example:

```ts
await ensureDir('foo');
await Deno.writeTextFile("foo/hello.txt", "Hello World");
```

This use-case is valid for both links to directories and directories. Unix is designed so that devs don't need to care, which is why `stat` is the preferred system call over the special case of `lstat`. The layout of the filesystem is generally under the control of the user, not the dev.

The value of `std` is that it is a simple library for devs who don't need to worry about the details. Applications that *do* care have `mkdir` and will likely want to handle edge cases differently.

> I think this should be implemented as option instead of changing the default behavior.

In this case, I think adding an option exposes an unnecessary detail that clutters the interface and reduces the value of `std`.

The default behavior of throwing an exception is undesirable to both the dev and the user. The dev because they didn't handle an exception they didn't care about, and the user because the application crashed with no recourse but to arbitrarily rearrange their filesystem to make the application happy. I often have to use bind mounts or [bindfs](https://bindfs.org/) to work around these applications.

PR #4133, opened 2024-01-08
deprecation(io): move types file

iuioiua:
Doing as a breaking change as the change is a simple rename.

See https://youtu.be/zu-EgnbmcLY?feature=shared
See https://github.com/microsoft/TypeScript/issues/52593#issuecomment-1419505081


kt3k:
I think we still should do deprecation first.

io/types.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @deprecate (will be removed in 1.0.0) Use the
+ * [Streams API]{@linkcode https://developer.mozilla.org/en-US/docs/Web/API/Streams_API}
+ * instead.
+ */
+
+/**
+ * An abstract interface which when implemented provides an interface to read bytes into an array buffer asynchronously.
+ *
+ * @deprecated (will be removed in 1.0.0) Use {@linkcode ReadableStream} instead.
```
iuioiua, line 15: Note to self: remove this if we bring back `std/io`.

kt3k (approved):
LGTM

PR #4134, opened 2024-01-08
chore: enable `skipLibCheck` option

iuioiua:
The functionality won't be needed once #4133 is merged. Performance optimisation.

See https://www.typescriptlang.org/tsconfig#skipLibCheck


iuioiua:
Oh. This option doesn't exist in the `compilerOptions` field in the Deno configuration file... I'll investigate and re-open later.

Issue #4139, opened 2024-01-10
todo: give `assertNotEquals()` assertion signature


iuioiua:
See https://github.com/denoland/deno_std/pull/4121#discussion_r1446042471
See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

PR #4140, opened 2024-01-10
feat(assert): assertion signatures on more of the assert modules

syhol:
Spawned from this conversation https://github.com/denoland/deno_std/pull/4121#discussion_r1444173340

Adding more assertion signatures to assert modules.

### Background

There are two ways assertion signatures are already implemented in assert/* and thats:

**actual unknown** - assertInstanceOf & assertIsError & assertStrictEquals:

```typescript
export function assert???<T>(
  actual: unknown,
  expected: T,
  msg?: string,
): asserts actual is T { /**/ }
```

This seems to be used for all positive asserts.

and

**two type params** - assertNotInstanceOf:

```typescript
export function assert???<A, T extends A>(
  actual: A,
  expected: T,
  msg?: string,
): asserts actual is T { /**/ }
```

This seems to be uses for all negative/not asserts.

I expect the reason that negative/not asserts use the **two type params** is because its the only way to do type narrowing by doing `asserts actual is Exclude<A, T>`.

### This PR

In this PR I've implemented assertion signatures on the remaining assert functions that make sense to have assertion signatures, following the pattern thats been previously established:
- assertEquals using the **actual unknown** style
- assertNotEquals and assertNotStrictEquals using the **two type params** style

### DIFF
--- a/assert/assert_equals.ts
+++ b/assert/assert_equals.ts
@@ -23,12 +23,12 @@ import { CAN_NOT_DISPLAY } from "./_constants.ts";
  *
  * Note: formatter option is experimental and may be removed in the future.
  */
-export function assertEquals<T>(
-  actual: T,
+export function assertEquals<A, T extends A>(
+  actual: A,
   expected: T,
   msg?: string,
   options: { formatter?: (value: unknown) => string } = {},
-) {
+): asserts actual is T {
   if (equal(actual, expected)) {
     return;
   }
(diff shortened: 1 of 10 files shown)


syhol:
Addressing this ticket: https://github.com/denoland/deno_std/issues/4139

kt3k:
Probably this is not a desired change. See the discussion in https://github.com/denoland/deno_std/pull/2203 for details. It has several benefits to give `actual` and `expected` the same generic type `T`, and this change looses it.

iuioiua:
Yes, I agree with Yoshiya. After seeing this PR, I realised my initial suggestion was wrong. We always want `actual` and `expected` to have the same type. You don't need to use either of the assertions if they're not the same type. I'll close this without merging. Either way, thank you, Simon.

PR #4146, opened 2024-01-10
refactor(data_structures): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the data_structures module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/data_structures/binary_heap.ts
+++ b/data_structures/binary_heap.ts
@@ -5,9 +5,9 @@ import { descend } from "./comparators.ts";
 
 /** Swaps the values at two indexes in an array. */
 function swap<T>(array: T[], a: number, b: number) {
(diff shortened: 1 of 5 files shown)


data_structures/binary_heap.ts
syhol, line 1: This file "data_structures/binary_heap.ts" might need a bit of a closer review. The implementation seems sound for now but this won't help us catch issues in future. Are you happy leaving them as non-null assertions for now?

iuioiua (approved):
Yes, I think the use of non-null assertions is valid in this case. Also, PTAL at the tweak I appplied, for future reference. Either way, LGTM! Thank you, Simon.

PR #4150, opened 2024-01-10
refactor(expect): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the expect module, tracked in https://github.com/denoland/deno_std/issues/4040

# BUG FIX INCLUDED

Looks like toHaveBeenNthCalledWith failure case was checking with the wrong variable, I'll call out where I've updated it to now check the correct variable.

### DIFF
--- a/expect/_matchers.ts
+++ b/expect/_matchers.ts
@@ -542,7 +542,7 @@ export function toHaveBeenLastCalledWith(
 ): MatchResult {
   const calls = getMockCalls(context.value);
   const hasBeenCalled = calls.length > 0 &&
-    equal(calls[calls.length - 1].args, expected);
+    equal(calls.at(-1)?.args, expected);
 
   if (context.isNot) {
     if (hasBeenCalled) {
@@ -583,7 +583,7 @@ export function toHaveBeenNthCalledWith(
   const calls = getMockCalls(context.value);
   const callIndex = nth - 1;
   const hasBeenCalled = calls.length > callIndex &&
-    equal(calls[callIndex].args, expected);
+    equal(calls[callIndex]?.args, expected);
 
   if (context.isNot) {
     if (hasBeenCalled) {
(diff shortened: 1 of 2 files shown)


expect/_matchers.ts
```diff
@@ -596,7 +596,7 @@ export function toHaveBeenNthCalledWith(
   } else {
     if (!hasBeenCalled) {
       const nthCall = calls[callIndex];
-      if (!nth) {
+      if (!nthCall) {
```
syhol, line 599: This is the bugfix. At the top of this function we check `if (nth < 1) throw Error ...` so this check here was redundant, but reading the error message below (601,602,603) shows that it thinks its checking for "but the n-th call does not exist" so it makes sense to instead check `nthCall` aka `calls[callIndex];`

Read the full function for full context, its only about 30 lines.

iuioiua: Nice. Should the two error throws below use `nthCall` instead of `nth` too?

iuioiua: Also, lines 599-603 aren't covered in tests (see [here](https://app.codecov.io/gh/denoland/deno_std/pull/4150/blob/expect/_matchers.ts#L599)). Would you be able to add one?

syhol: > Nice. Should the two error throws below use nthCall instead of nth too?

If you mean the interpolated string `(n=${nth})` then no, that looks good as it is. If you mean something else then you'll have to link it for me.

> Also, lines 599-603 aren't covered in tests (see here). Would you be able to add one?

My mistake, I've added a new test case following the standard set in https://github.com/denoland/deno_std/issues/3754 and speaking of that standard, expect is missing from the checklist.

I've also added some Given/When/Then comments in that new test, if you don't like them I can remove them, just let me know.

iuioiua: > If you mean the interpolated string `(n=${nth})` then no, that looks good as it is. If you mean something else then you'll have to link it for me.

👍🏾 

> My mistake, I've added a new test case following the standard set in #3754 and speaking of that standard, expect is missing from the checklist.

Cool. Thanks. I've added `expect` to the checklist.

> I've also added some Given/When/Then comments in that new test, if you don't like them I can remove them, just let me know.

All good. I removed them myself.

iuioiua (approved):
LGTM! Thank you.

PR #4154, opened 2024-01-10
chore(path): integrate zero length string test into other tests

lino-levan:
Closes https://github.com/denoland/deno_std/issues/4068.


iuioiua (approved):
LGTM! Thanks, Lino.

PR #4161, opened 2024-01-10
deprecation(semver): deprecate `SemVerRange`, introduce `Range`

timreichen:
This PR does a bunch of things. While some things could theoretically be split up into separate PRs (ref: https://github.com/denoland/deno_std/pull/4059), some stand-alone PRs might look redundant.

- deprecates `SemVerRange` type in favour of `Range`.
- deprecates `isSemVerRange()` in favour of `isRange()`
- makes `Comparator.semver` property optional in favour of `Comparator` extending `SemVer` that allows a simpler and less nested `Range` type.
**Note**: `Comparator` might become be deprecated (ref: https://github.com/denoland/deno_std/pull/4059#issuecomment-1882287030), but because `SemVerRange` is dependent on its structure, it needs to be changed in order to avoid breaking changes.


iuioiua (changes requested):
Sorry, Tim. Can you please resolve the merge conflicts? I also noticed that the removal version for some added deprecation notices is 0.213.0 and others 0.214.0. Can you please only have one version so removals can happen all at the same time?

semver/is_range.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param value The value to check if its a valid Range
+ * @returns True if its a valid Range otherwise false.
+ */
+export function isRange(value: unknown): value is Range {
+  if (value === null || value === undefined) return false;
+  if (!Array.isArray(value)) return false;
+  const range = value as Range;
+  return range.every((r) =>
+    Array.isArray(r) && r.every((c) => isComparator(c))
+  );
```
iuioiua, line 21:
```suggestion
  return Array.isArray(value) &&
    value.every((r) => Array.isArray(r) && r.every((c) => isComparator(c)));
```

iuioiua: Should we be using `isComparator()` if it's deprecated?

semver/comparator_format.ts
```diff
@@ -12,5 +12,5 @@ import { format } from "./format.ts";
  */
 export function comparatorFormat(comparator: Comparator): string {
   const { semver, operator } = comparator;
-  return `${operator}${format(semver)}`;
+  return `${operator}${format(semver ?? comparator)}`;
```
iuioiua, line 15: Nice. Can we please have added functionality? We want to make sure that both `SemVer` and `Comparator` inputs work fine on all these functions, including the others in this PR.

semver/parse_comparator.ts
```diff
@@ -47,5 +47,5 @@ export function parseComparator(comparator: string): Comparator {
     }
     : ANY;
 
-  return { operator, semver };
+  return { operator, ...semver, semver };
```
iuioiua, line 50: I forsee the output of this object being confusing to users during the deprecation period. But I don't see another viable way of doing this. Food for thought...

iuioiua (changes requested):
I mostly agree with this PR. The only part I'm slightly uncertain about is having `Comparator` extend `SemVer`. Either way. I left a few suggestions.

Also, I notice we still use some deprecated functions within other functions. We should look into switching to the updated functions in those cases. Probably best done after this PR, separately.

timreichen:
> I mostly agree with this PR. The only part I'm slightly uncertain about is having Comparator extend SemVer. Either way. I left a few suggestions.

i don't see a better way to remove the `semver` in `Comparator` prop without breaking changes. But I am open for suggestions.

> Also, I notice we still use some deprecated functions within other functions. We should look into switching to the updated functions in those cases. Probably best done after this PR, separately.

I'll have a look at that in a short while. I think we might be best of putting all comparator code into one `_comparator.ts`.

semver/is_range.ts
timreichen: see comment below.

semver/comparator_format.ts
timreichen: How do you mean exactly?

timreichen:
> Sorry, Tim. Can you please resolve the merge conflicts? I also noticed that the removal version for some added deprecation notices is 0.213.0 and others 0.214.0. Can you please only have one version so removals can happen all at the same time?

Done

kt3k:
This looks like an immediate breaking change to me. `parseRange(">=1.2.3 <2.0.0")` now returns:

```ts
[
  [
    {
      operator: ">=",
      major: 1,
      minor: 2,
      patch: 3,
      prerelease: [],
      build: [],
      semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] }
    },
    {
      operator: "<",
      major: 2,
      minor: 0,
      patch: 0,
      prerelease: [],
      build: [],
      semver: { major: 2, minor: 0, patch: 0, prerelease: [], build: [] }
    }
  ]
]
```

instead of:

```js
{
  ranges: [
    [
      {
        operator: ">=",
        semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] }
      },
      {
        operator: "<",
        semver: { major: 2, minor: 0, patch: 0, prerelease: [], build: [] }
      }
    ]
  ]
}
```

It's typed like both are supported, but actually only `Range` is supported.

kt3k:
I feel this kind of thing can't be done in a non-breaking way.

timreichen:
> This looks like an immediate breaking change to me. `parseRange(">=1.2.3 <2.0.0")` now returns:
> 
> ```ts
> [
>   [
>     {
>       operator: ">=",
>       major: 1,
>       minor: 2,
>       patch: 3,
>       prerelease: [],
>       build: [],
>       semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] }
>     },
>     {
>       operator: "<",
>       major: 2,
>       minor: 0,
>       patch: 0,
>       prerelease: [],
>       build: [],
>       semver: { major: 2, minor: 0, patch: 0, prerelease: [], build: [] }
>     }
>   ]
> ]
> ```
> 
> instead of:
> 
> ```js
> {
>   ranges: [
>     [
>       {
>         operator: ">=",
>         semver: { major: 1, minor: 2, patch: 3, prerelease: [], build: [] }
>       },
>       {
>         operator: "<",
>         semver: { major: 2, minor: 0, patch: 0, prerelease: [], build: [] }
>       }
>     ]
>   ]
> }
> ```
> 
> It's typed like both are supported, but actually only `Range` is supported.

`parseRange()` actually has the `ranges` property. it is not enummerated, but if you do `console.log((ranges as SemVerRange).range);` it will return properly. If you like we can make it enumerable: `Object.defineProperty(ranges, "ranges", { enumerable: true, value: ranges })` so also shows in the log.
`

semver/comparator_format.ts
timreichen: I added some test cases. PTAL.

kt3k:
Ah, ok. Sorry I missed `Object.defineProperty` part in `parseRange`..

semver/types.ts
```diff
@@ -55,15 +58,21 @@ export interface SemVer {
...
+ * inner array represents AND comparisons.
+ */
+export type Range = Comparator[][];
 
 /**
  * A type representing a semantic version range. The ranges consist of
  * a nested array, which represents a set of OR comparisons while the
  * inner array represents AND comparisons.
+ *
+ * @deprecated (will be removed in 0.214.0) Use {@linkcode Range} instead.
```
kt3k, line 73:
```suggestion
 * @deprecated (will be removed in 0.215.0) Use {@linkcode Range} instead.
```

semver/types.ts
```diff
@@ -39,9 +39,12 @@ export type FormatStyle =
  * The shape of a valid semantic version comparator
  * @example >=0.0.0
  */
-export interface Comparator {
+export interface Comparator extends SemVer {
   operator: Operator;
-  semver: SemVer;
+  /**
+   * @deprecated (will be removed in 0.214.0) {@linkcode Comparator} extends {@linkcode SemVer}. Use `major`, `minor`, `patch`, `prerelease`, and `build` properties instead.
```
kt3k, line 45:
```suggestion
   * @deprecated (will be removed in 0.215.0) {@linkcode Comparator} extends {@linkcode SemVer}. Use `major`, `minor`, `patch`, `prerelease`, and `build` properties instead.
```

semver/is_semver_range.ts
```diff
@@ -11,14 +11,16 @@ import { isComparator } from "./is_comparator.ts";
  * Adds a type assertion if true.
  * @param value The value to check if its a valid SemVerRange
  * @returns True if its a valid SemVerRange otherwise false.
+ *
+ * @deprecated (will be removed in 0.214.0) Use {@linkcode isRange} instead.
```
kt3k, line 15:
```suggestion
 * @deprecated (will be removed in 0.215.0) Use {@linkcode isRange} instead.
```

kt3k (approved):
LGTM

iuioiua (approved):
LGTM! Nice work, Tim! Thank you.

PR #4162, opened 2024-01-10
refactor(semver): remove `cmp()` dependency from `testComparator()`

timreichen:
Inlines `cmp()` code in `testComparator()` in preparation of the removal of `cmp()`.


iuioiua (approved):
Nice! Thanks, Tim.

PR #4165, opened 2024-01-11
docs(fmt): complete documentation

iuioiua:
Towards #3764 and #3878

`fmt/colors.ts` could probably do with further polish in the future. But the current state is good enough to make the doc linter happy.


kt3k (approved):
LGTM. Thanks!

PR #4168, opened 2024-01-11
fix: add `--no-lock` to spawned process args in tests

iuioiua:
I noticed that tests containing spawned processes that use the Deno binary tend to be flaky. There have been a few instances in the past that these processes fail because the runtime is unable to parse the contents of the lockfile. I.e. race condition.

This fix adds the `--no-lock` CLI flag to all such instances. Hopefully, we'll see a noticable reduction in test flakiness in CI.


iuioiua:
5x successful CI passes. I believe the flakiness is fixed 😎

kt3k (approved):
Great! Thanks!

PR #4179, opened 2024-01-12
BREAKING(http): remove `CookieMap`

iuioiua:


kt3k (approved):
LGTM

joehillen:
Why do y'all keep removing stuff without any justification or documentation?

crowlKats:
@joehillen this was marked deprecated for multiple releases with notes on replacement (being https://deno.land/std/http/cookie.ts), and then was removed.

joehillen:
Links to a GitHub issue for the discussion or justification in the release notes, PR, commit message, or deprecation notice would be appreciated.

joehillen:
And I'm not just referring to this specific issue. There is a general lack of transparency in this project.

iuioiua:
Hey @joehillen, we don't remove anything without any justification or documentation. In this case, I mistakenly didn't link to the corresponding issues concerning the rationale for the deprecation. My bad. I've now fixed the initial comment of this PR.

Regarding documentation, as Leo said, we mark the APIs as deprecated for multiple versions. However, from now on, we should open a GitHub issue for the lifetime of an API's deprecation. That issue should then be linked to within the deprecation notice. The issue would only be closed if the consensus remains in support of the deprecation. WDYT, @kt3k? If you agree, we can work to upgrade the current deprecation notices in the codebase.

A general lack of transparency is the opposite of what we aim for in this project. If there are any other holes like this, please let us know.

kt3k:
> However, from now on, we should open a GitHub issue for the lifetime of an API's deprecation. That issue should then be linked to within the deprecation notice. The issue would only be closed if the consensus remains in support of the deprecation.

That sounds a good idea 👍 Let's do it

sholladay:
It's really disappointing to see that `SecureCookieMap` was removed from `std`.

Every server framework should be implementing signed cookies, at a minimum, if not encrypted + signed cookies.

I was excited when `SecureCookieMap` was added, and I planned to use it in [Pogo](https://github.com/sholladay/pogo), but wanted to let the dust settle on Deno's server APIs before continuing work on it, as both `Deno.serve()` and `std/http/server.ts` were undergoing a lot of changes at the time. Now I come back to check on the state of things and they're both gone! 😄

I am mostly fine with `server.ts` going away, as `Deno.serve()` is a pretty solid replacement (although it does mean the version used is out of my control as a library author). However, the `signCookie()`, `verifySignedCookie()`, `parseSignedCookie()` replacements are just... not nearly as good. They don't support key rotation, they don't handle multiple cookies, and they don't align to [CookieStore](https://developer.mozilla.org/en-US/docs/Web/API/CookieStore) or any other standard, as `SecureCookieMap` did.

If Deno intends to be a platform for high quality servers, it needs awesome tools for secure cookies.

kt3k:
@sholladay 
`SignedCookieMap` looked like implementing very opinionated behavior to me. Is there any standard defining how it's signed, how the keys should be rotated/handled, etc? It seemed inspired by [npm:cookies](https://www.npmjs.com/package/cookies) package, but I'm not sure how much that package is 'standard'.

Also I don't find the section for 'signed cookies' in the docs of recently popular frameworks such as Next.js, Nuxt, Astro, etc. I'm not very sure signed cookie is still the best practice in 2024 if such popular frameworks don't have special support for it.

sholladay:
> `SignedCookieMap` looked like implementing very opinionated behavior to me.

It's high-level, to be sure, but the only part that I could see as opinionated is the format of the signature.

> Is there any standard defining how it's signed, how the keys should be rotated/handled, etc?

Sadly, no. According to [RFC 6265 section 8.3](https://www.rfc-editor.org/rfc/rfc6265#section-8.3), it is up to the server:

> Servers SHOULD encrypt and sign the contents of cookies **(using whatever format the server desires)** when transmitting them to the user agent (even when sending the cookies over a secure channel).

Some systems put the signature in a separate cookie, while others encode it into the cookie value. In my experience, the latter approach, as used in `SecureCookieMap`, is the most common. It's also the approach I would like to use, because it keeps the signature as close as possible to the data it is signing.

See: https://stackoverflow.com/questions/66067645/is-there-a-standard-format-for-hmac-in-cookies

> It seemed inspired by npm:cookies package, but I'm not sure how much that package is 'standard'.

I share this viewpoint, however `SecureCookieMap` improves on the npm `cookies` package by aligning with the `CookieStore` web standard. So it's only the key management and signing format that aren't standardized. The signing format could easily be an option if it's really too opinionated. The key management is about as simple and elegant as I can possibly imagine. It's literally just an array of keys. And the keys themselves are based on web standard crypto.

> Also I don't find the section for 'signed cookies' in the docs of recently popular frameworks such as Next.js, Nuxt, Astro, etc. I'm not very sure signed cookie is still the best practice in 2024 if such popular frameworks don't have special support for it.

This is... not a reasonable argument. Those frameworks are negligent. As is most of the Node.js ecosystem when it comes to security.

The hapi.js framework does use encrypted and signed cookies by default via `iron` and they have >600k weekly downloads and ~1M weekly downloads respectively.

Lots of popular frameworks in other languages also sign cookies, such as Django.

And as mentioned earlier, the HTTP spec recommends it, in addition to encryption which `SecureCookieMap` unfortunately does not implement.

PR #4184, opened 2024-01-12
feat(io): un-deprecate `Buffer`

iuioiua:
See https://github.com/denoland/deno_std/issues/4120#issuecomment-1888570028


kt3k (approved):
LGTM

PR #4185, opened 2024-01-12
deprecation(semver): deprecate `outside()`

timreichen:
- deprecates `outside()`
Reason: `outside()` just calls other functions present in the semver mod.


semver/outside.ts
```diff
@@ -13,6 +13,8 @@ import { rangeMin } from "./range_min.ts";
  * @param range The range of possible versions
  * @param hilo The operator for the comparison or both if undefined.
  * @returns True if the version is outside of the range based on the operator
+ *
+ * @deprecated (will be removed in 0.214.0) Use {@linkcode gtr}, {@linkcode ltr} or {@linkcode testRange} instead.
```
iuioiua, line 17:
```suggestion
 * @deprecated (will be removed in 0.215.0) Use {@linkcode gtr}, {@linkcode ltr} or {@linkcode testRange} instead.
```
0.214.0 is a little too soon.

iuioiua (changes requested):

kt3k (approved):
I agree with the deprecation. The semantics of `outside` feels very strange to me.

LGTM

iuioiua (approved):
LGTM. Thanks for the work put into leading up to this, Tim!

PR #4186, opened 2024-01-12
refactor(semver): replace imports of deprecated functions

timreichen:
- puts all deprecated but still needed `Comparator` functions into one private `_comparator.ts` file.
- replaces all imports of deprecated functions in the mod with imports of that file.


iuioiua:
I assume it's being done to make removal of `Comparator`-related APIs in the future easier. I'm not in favour of this change as it increases the dependency sizes of all `Comparator`-related APIs until removal with no benefit.

timreichen:
> I assume this was done to make the removal of `Comparator`-related APIs in the future easier. I'm not in favour of this change as it increases the dependency sizes of all `Comparator`-related APIs until removal with no benefit.

PR #4176, opened 2024-01-12
deprecation(log): move handlers to single-export files

iuioiua:
Closes #4174

PR #4188, opened 2024-01-13
refactor(log): replace deprecated imports

timreichen:
- replaces imports from deprecated exports with imports from non-deprecaed exports


log/mod.ts
```diff
@@ -358,16 +358,12 @@
...
-  FileHandler,
-  RotatingFileHandler,
-  WriterHandler,
-} from "./handlers.ts";
 import { assert } from "../assert/assert.ts";
 import type { LevelName } from "./levels.ts";
 import { jsonFormatter } from "./formatters.ts";
+import { ConsoleHandler } from "./console_handler.ts";
+import { BaseHandler } from "./base_handler.ts";
+import { FileHandler, RotatingFileHandler, WriterHandler } from "./handlers.ts";
```
kt3k, line 366: Probably we should also update the import path for `FileHandler`, `RotatingFileHandler`?

timreichen: I Agree. Done

iuioiua (approved):
I didn't make these changes in [my original PR](https://github.com/denoland/deno_std/pull/4176) as I thought some of them would be breaking changes. Looking at this PR, I now think it's fine to do.

I've updated some more imports, and now, LGTM! Thanks, Tim.

PR #4189, opened 2024-01-13
refactor(cli): cleanup `parse_args.ts`

timreichen:
This PR removes redundant code in `parse_args.ts` and makes code more readable and possibly improves performance by
- using a single RegExp for parsing
- replacing some objects with `Map`
- replacing some arrays with `Set`

The code can still be made more readable. There probably will be more PRs in the future to do so.


iuioiua (changes requested):

cli/parse_args.ts
```diff
@@ -329,52 +329,77 @@ export interface ParseOptions<
...
-  if (typeof x === "number") return true;
-  if (/^0x[0-9a-f]+$/i.test(String(x))) return true;
-  return /^[-+]?(?:\d+(?:\.\d*)?|\.\d+)(e[-+]?\d+)?$/.test(String(x));
+function aliasIsBoolean(
+  aliasMap: Map<string, Set<string>>,
+  booleanSet: Set<string>,
+  key: string,
+): boolean {
+  const set = aliasMap.get(key);
+  if (!set) return false;
```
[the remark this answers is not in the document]
timreichen: `aliasMap` values are either a `Set` or `undefined`. So I think comparing it to so it is not falsy should be enough, no?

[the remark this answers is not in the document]
iuioiua: Admittedly, this is more of a personal preference of mine - being as explicit as possible.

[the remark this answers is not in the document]
timreichen: Ok, I am fine either way.

iuioiua:
A bit is going on here. Can you add your own comments (in GitHub, not in the code), where appropriate, to ease our understanding and reviews?

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
     TDoubleDash
   > = {},
 ): Args<TArgs, TDoubleDash> {
-  const aliases: Record<string, string[]> = {};
-  const flags: Flags = {
```
timreichen, line 454: `flags` is only used internally. I've put all props as separate variables.

cli/parse_args.ts
```diff
@@ -329,52 +329,75 @@ export interface ParseOptions<
...
+}
 
-  const key = keys.at(-1);
-  return key !== undefined && hasOwn(o, key);
+function parseBooleanString(value: unknown) {
+  return value !== "false";
 }
 
+const FLAG_REGEXP =
+  /^(?:-(?:(?<doubleDash>-)(?<negated>no-)?)?)(?<key>.+?)(?:=(?<value>.+?))?$/s;
```
timreichen, line 399: Argument strings are now parsed with one regexp. Before there were multiple smaller regexps, but the argument string had to be processed multiple times.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
+    const negatableArgs = Array.isArray(negatable) ? negatable : [negatable];
     for (const key of negatableArgs.filter(Boolean)) {
-      flags.negatable[key] = true;
-      const alias = get(aliases, key);
-      if (alias) {
-        for (const al of alias) {
-          flags.negatable[al] = true;
-        }
-      }
+      negatableSet.add(key);
```
timreichen, line 528: This does the same as the removed code, but more readable with the use of `Set`.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
+      aliasMap.get(key)?.forEach((alias) => negatableSet.add(alias));
     }
   }
 
   const argv: Args = { _: [] };
 
-  function argDefined(key: string, arg: string): boolean {
```
timreichen, line 548: This is merged into `setArgument()`

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
     );
   }
 
   let notFlags: string[] = [];
 
   // all args after "--" are not parsed
-  if (args.includes("--")) {
-    notFlags = args.slice(args.indexOf("--") + 1);
-    args = args.slice(0, args.indexOf("--"));
+  const index = args.indexOf("--");
```
timreichen, line 564: Little performance improvement: `indexOf()` has only to be called once.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
+  if (index !== -1) {
+    notFlags = args.slice(index + 1);
+    args = args.slice(0, index);
   }
 
   for (let i = 0; i < args.length; i++) {
     const arg = args[i];
-    assertExists(arg);
```
timreichen, line 623: This block was rewritten and uses `FLAG_REGEXP` instead of smaller regexps.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
-        i++;
-      } else if (next !== undefined && (next === "true" || next === "false")) {
-        setArg(key, next === "true", arg);
-        i++;
-      } else {
-        setArg(key, get(flags.strings, key) ? "" : true, arg);
+
+    const groups = arg.match(FLAG_REGEXP)?.groups;
+
+    if (groups) {
```
timreichen, line 575: I tried to get rid of as many `else` clauses as possible and use early return or loop break or continuation instead.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
 
-  for (const key of Object.keys(flags.strings)) {
-    if (!hasKey(argv, key.split(".")) && get(flags.collect, key)) {
-      setKey(
-        argv,
-        key,
-        [],
-        false,
-      );
+  for (const key of booleanSet.keys()) {
```
timreichen, line 708: This does the same as the removed code, but more readable with the use of Set.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
 
-  for (const key of Object.keys(flags.bools)) {
-    if (!hasKey(argv, key.split("."))) {
-      const value = get(flags.collect, key) ? [] : false;
-      setKey(
-        argv,
-        key,
-        value,
-        false,
+  for (const [key, value] of Object.entries(defaults)) {
```
timreichen, line 698: This does the same as the removed code, but more readable with the use of Set.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
-        break;
-      }
+      continue;
     }
-  }
 
-  for (const [key, value] of Object.entries(defaults)) {
-    if (!hasKey(argv, key.split("."))) {
-      setKey(argv, key, value, false);
+    if (unknownFn?.(arg) !== false) {
```
timreichen, line 686: This does the same as the removed code, but more readable with the use of Set.

cli/parse_args.ts
```diff
@@ -450,316 +473,250 @@ export function parseArgs<
...
+
+        value = stringSet.has(key) ? "" : true;
+        setArgument(key, value, arg, true);
+        continue;
       }
-    } else if (/^-[^-]+/.test(arg)) {
       const letters = arg.slice(1, -1).split("");
 
       let broken = false;
-      for (const [j, letter] of letters.entries()) {
```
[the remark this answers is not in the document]
iuioiua: The previous code was actually better in that TypeScript knows that `letter` is defined. When refactoring for `noUncheckedIndexedAccess`, we increase our use of `.entries()` for this reason. Can you please revert?

cli/parse_args.ts
iuioiua: I don't follow. Is this comment correct?

iuioiua (changes requested):
Sorry for the delayed review.

cli/parse_args.ts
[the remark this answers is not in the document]
timreichen: done

cli/parse_args.ts
timreichen: Updated comment

iuioiua (approved):
LGTM! Thank you, Tim.

Issue #4187, opened 2024-01-13
bug(toml): `parse` duplicates the character next to reserved escape sequences


babiabeo:
**Describe the bug**

According to the [TOML spec](https://toml.io/en/v1.0.0#string), escape sequences that are not listed here are reserved:
```
\b         - backspace       (U+0008)
\t         - tab             (U+0009)
\n         - linefeed        (U+000A)
\f         - form feed       (U+000C)
\r         - carriage return (U+000D)
\"         - quote           (U+0022)
\\         - backslash       (U+005C)
\uXXXX     - unicode         (U+XXXX)
\UXXXXXXXX - unicode         (U+XXXXXXXX)
```

But when I parse a string that includes a reserved escape sequence, the parser duplicates the character next to it.

**Steps to Reproduce**

```ts
import { parse } from "https://deno.land/std@0.212.0/toml/mod.ts";

parse('a = "a\\0b"'); // { a: "abb" }
```

**Expected behavior**

There are 2 cases that I expected:

- Ignores the reserved escape sequence:
```ts
parse('a = "a\\0b"'); // { a: "ab" }
```

- Throws an error as specified in the spec:
> All other escape sequences not listed above are reserved; if they are used, TOML should produce an error.

**Environment**

- OS:
- deno version: 1.39.3
- std version: 0.212.0

PR #4192, opened 2024-01-14
fix(toml): `parse()` duplicates the character next to reserved escape sequences

babiabeo:
Fix: #4187


kt3k:
> There are 2 cases that I expected:
> 
> * Ignores the reserved escape sequence:
> 
> ```ts
> parse('a = "a\\0b"'); // { a: "ab" }
> ```
> 
> * Throws an error as specified in the spec:
> 

@babiabeo 
Do you think ignoring is more useful than throwing?

iuioiua:
I think we should stick to the spec and throw.

babiabeo:
> > There are 2 cases that I expected:
> > 
> > * Ignores the reserved escape sequence:
> > 
> > ```ts
> > parse('a = "a\\0b"'); // { a: "ab" }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * Throws an error as specified in the spec:
> 
> @babiabeo Do you think ignoring is more useful than throwing?

Currently, this PR solves the issue by ignoring reserved escape sequences.

However, I see that most TOML parser packages throw an error in this case

- [`js-toml`](https://www.npmjs.com/package/js-toml)
```ts
import { load } from "npm:js-toml";

console.log(load('a = "a\\0b"'));

// error: Uncaught (in promise) SyntaxParseError: Syntax error
// unexpected character: ->"<- at offset: 4, skipped 3 characters.
// unexpected character: ->b<- at offset: 8, skipped 2 characters.
```

- [`tomllib`](https://docs.python.org/3/library/tomllib.html)
```py
import tomllib

print(tomllib.loads('a = "a\\0b"'))

# File "D:\coding\programming\languages\python\Lib\tomllib\_parser.py", line 494, in parse_basic_str_escape
#     raise suffixed_err(src, pos, "Unescaped '\\' in a string") from None
# tomllib.TOMLDecodeError: Unescaped '\' in a string (at line 1, column 9)
```

- [Go `toml`](https://pkg.go.dev/github.com/BurntSushi/toml)
```go
package main

import (
	"fmt"

	"github.com/BurntSushi/toml"
)

type Data struct {
    A string `toml:"a"`
}

func main() {
    var data Data
    _, err := toml.Decode("a = \"a\\0b\"", &data)

    fmt.Println(err)

    // toml: line 1 (last key "a"): invalid escape in string '\0'
}
```

iuioiua (changes requested):
I think it best if we instead threw instead of ignoring, as to follow the spec. Are you able to rework this PR to do that?

iuioiua (commented):
@babiabeo, I've made some tweaks to improve the thrown error message and accuracy of the tests. Do these changes seem reasonable to you?

babiabeo:
> @babiabeo, I've made some tweaks to improve the thrown error message and accuracy of the tests. Do these changes seem reasonable to you?

Yes. I forgot to add `\\`. Thanks for adding it!

iuioiua (approved):
LGTM! Thank you, David.

Issue #4174, opened 2024-01-12
suggestion: re-structure `std/log` to adhere to singe-export files


iuioiua:
Each file should export only tightly-related symbols.

Background #2936

kt3k:
We recently sometimes choose non single-export file pattern (e.g. `std/text/cases`). What is the criteria to follow the pattern or divert from the pattern? Can we create an objective rule about it?

iuioiua:
I think it's reasonable to group multiple symbols in the same file if at least one of the following is true:
1. Their signatures and dependencies have near-total overlap. E.g. `std/text/cases`
2. The symbols are highly likely to be used in conjunction. E.g. encoding and decoding functions.

These criteria should justify cases with many symbols within a given file or just one.

I'd be happy to document this if this sounds reasonable.

kt3k:
Ok. Let's document them in README

eryue0220:
Can I have a tried for this one?

iuioiua:
Sure! I believe the next step is to factor out `setup()` into its own style.

iuioiua:
Completed in #4254.

PR #4194, opened 2024-01-14
docs: add "Minimal Exports" explainer

iuioiua:
Previously, we referred to this as the "single-export files" pattern. I think it makes more sense to use the term "minimal-export files".

See https://github.com/denoland/deno_std/issues/4174#issuecomment-1888435091


kt3k (approved):
LGTM

PR #4195, opened 2024-01-14
feat(log): make handlers disposable

iuioiua:
This change makes all `std/log` handlers disposable.

See https://github.com/tc39/proposal-explicit-resource-management


kt3k (approved):
LGTM. Nice!

PR #4197, opened 2024-01-14
deprecation(io): rework deprecations

iuioiua:
This change reworks `std/io` deprecations by bringing the removal version forward to v0.215.0 from v1, and providing alternative solutions for advanced I/O use.

It seems that the wide-use of `copy()`, `readAll()`, `writeAll()` and `Buffer` justified having a removal version of v1. But now that these APIs have been undeprecated, I think it's reasonable to remove the other, more advanced, but less-used `std/io` APIs sooner.

### DIFF
--- a/io/buf_reader.ts
+++ b/io/buf_reader.ts
@@ -3,7 +3,8 @@
 
 /**
  * @module
- * @deprecated (will be removed after 1.0.0) Use the [Web Streams API]{@link https://developer.mozilla.org/en-US/docs/Web/API/Streams_API} instead.
+ * @deprecated (will be removed in 0.215.0) Use
+ * {@linkcode https://deno.land/std/io/buffer.ts | Buffer} instead.
  */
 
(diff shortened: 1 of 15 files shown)


io/buf_reader.ts
```diff
@@ -40,15 +43,19 @@ export class PartialReadError extends Error {
...
  */
 export interface ReadLineResult {
   line: Uint8Array;
   more: boolean;
 }
 
 /**
- * @deprecated (will be removed after 1.0.0) Use the [Web Streams API]{@link https://developer.mozilla.org/en-US/docs/Web/API/Streams_API} instead.
+ * @deprecated (will be removed in 0.215.0) Use
+ * {@linkcode https://deno.land/std/io/buffer.ts | Buffer} instead.
```
kt3k, line 58: Is BufReader usage covered by Buffer?

iuioiua:
It appears this is no longer needed.

PR #4199, opened 2024-01-15
deprecation(streams): rework deprecations

iuioiua:
This change reworks `std/streams` deprecations for various reasons (see comments). I think it's reasonable to bring some of the removal versions forward because:
1. Some of these APIs aren't heavily used.
2. Their migrations aren't too involved.
3. These APIs are still available through import version pinning.


iuioiua:
I'll come back to this.

PR #4200, opened 2024-01-15
feat: print warning on use of deprecated API

iuioiua:
This function prints a friendly warning if the given deprecated API is being used. This implementation is based on https://github.com/denoland/deno/pull/21939, but with a few differences that make more sense for the Standard Library.

image


iuioiua:
I think this tool is good enough for first pass.

internal/warn_deprecated_api.ts
```diff
@@ -0,0 +1,47 @@
...
+export function warnDeprecatedApi(config: WarnDeprecatedApiConfig) {
+  const variable = "NO_DEPRECATION_WARNINGS";
+  if (
+    Deno?.permissions.querySync({ name: "env", variable }).state ===
+      "granted" && Deno?.env.get(variable) === "1"
+  ) return;
+
+  // Remove the prepending "Error\n" from the stack output.
+  const stack = config.stack.slice(6);
+  const key = config.name + stack;
```
andrewthauer, line 35: Thought - Could you just set a unique `key` on the `WarnDeprecatedApiConfig` type instead of building a unique key from the name & stack on very function call? That way you don't need to do string manipulation and concatenation on every function call (which can potential slow things down).

iuioiua: Doing so would require the maintainer to keep track of the unique keys they use. Using `new Error().stack` allows for each location of the deprecated API's use to be reported. Also, having a slight slow-down during the use of a deprecated API is even more reason to migrate to the updated API.

andrewthauer: Fair enough. That makes logically sense. I do worry about the lag / time for intermediate libraries being updated to the latest stdlib though. I've seen quite a few libraries using very outdated versions  of stdlib already. Hopefully end user / app consumers of these libraries can motivate upstream updates in reasonably amount of time to avoid being too much of a disruption.

iuioiua: Yeah, I hope that this change motivates maintainers to use the latest `deno_std` too 🙂

deno.json
```diff
@@ -12,7 +12,7 @@
     "automation/": "https://raw.githubusercontent.com/denoland/automation/0.10.0/"
   },
   "tasks": {
-    "test": "deno test --doc --unstable --allow-all --parallel --coverage --trace-ops",
+    "test": "NO_DEPRECATION_WARNINGS=1 deno test --doc --unstable --allow-all --parallel --coverage --trace-ops",
```
iuioiua, line 15: This is needed for cases where we must still continue to test deprecated APIs.

internal/mod.ts
iuioiua, line 1: Does having an `internal` folder interfere with workspaces functionality, @lucacasonato?

internal/warn_on_deprecated_api.ts
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * Prints a warning message to the console for the given deprecated API.
+ *
+ * These warnings can be disabled by setting `NO_DEPRECATION_WARNINGS=1`
+ * in the current process.
+ */
+export function warnOnDeprecatedApi(config: WarnDeprecatedApiConfig) {
+  const variable = "NO_DEPRECATION_WARNINGS";
+  if (
+    Deno?.permissions.querySync({ name: "env", variable }).state ===
```
kt3k, line 29: It seems Deploy doesn't have `querySync`.

Also how about checking this outside of `warnOnDeprecatedApi` to reduce the overhead?

lucacasonato: Please open a bug for the missing Deploy feature

lucacasonato (commented):
Yes, this interferes with workspaces.

iuioiua:
> Yes, this interferes with workspaces. 

How can we make internal tools and workspaces work? What are some options?

internal/warn_on_deprecated_api.ts
kt3k: Created

iuioiua: I've gone with `Deno.permissions.query()`, which is implemented in Deploy.

internal/warn_on_deprecated_api.ts
```diff
@@ -5,6 +5,10 @@
 const { Deno } = globalThis as any;
 
 const ALREADY_WARNED_DEPRECATED = new Set<string>();
+const ENV_VAR_KEY = "DENO_NO_DEPRECATION_WARNINGS";
+const shouldDisableDeprecatedApiWarning =
+  Deno?.permissions.querySync({ name: "env", variable: ENV_VAR_KEY })
```
kt3k:
```suggestion
  Deno?.permissions.querySync?.({ name: "env", variable: ENV_VAR_KEY })
```

internal/warn_on_deprecated_api.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+// deno-lint-ignore no-explicit-any
+const { Deno } = globalThis as any;
+
+const ALREADY_WARNED_DEPRECATED = new Set<string>();
+const ENV_VAR_KEY = "DENO_NO_DEPRECATION_WARNINGS";
+const shouldDisableDeprecatedApiWarning =
+  (await Deno?.permissions.query({ name: "env", variable: ENV_VAR_KEY }))
```
kt3k, line 10: Top-level-await caused many issues in the past (The order of module executions became unpredictable with it). I think we should avoid it

I'd suggest we should use `querySync` for CLI and we can skip permission check in Deploy (env access is always possible in Deploy).

iuioiua: Ah, good point. Fixed.

kt3k (approved):
LGTM

PR #4201, opened 2024-01-15
refactor: cleanup deprecated-related APIs in `std/io` and `std/streams`

iuioiua:
Various cleanups that we previously missed.


kt3k (approved):
LGTM

Issue #4190, opened 2024-01-13
bug(expect): The function signature of `toMatchObject` is incorrect


BlackGlory:
**Describe the bug**
The current `toMatchObject` cannot pass an array of objects, although it actually works.
https://github.com/denoland/deno_std/blob/4fe6be730d581c196fde7f33cc85b5dde8892244/expect/expect.ts#L77

```ts
expect(result).toMatchObject([])
// Argument of type 'never[]' is not assignable to parameter of type 'Record<PropertyKey, unknown>'.
// Index signature for type 'string' is missing in type 'never[]'.deno-ts(2345)
```

**Expected behavior**

It should be
```ts
toMatchObject<E extends {} | any[]>(expected: E): R;
```
(https://github.com/DefinitelyTyped/DefinitelyTyped/blob/15024fd054cd63fa11739f178a166d7c836b248e/types/jest/index.d.ts#L1078)

or

```ts
toMatchObject(
  expected: Record<string, unknown> | Array<Record<string, unknown>>,
): R;
```
(https://github.com/jestjs/jest/blob/2178fa2183cb7cb2ac3e262388499bafd032ef03/packages/expect/src/types.ts#L298-L300)

**Environment**
- std version: 0.212.0

kt3k:
Thanks for your report! Can you send a PR to fix it? (The both fixes of the above look fine to me)

PR #4202, opened 2024-01-15
fix(expect): fix the function signature of `toMatchObject()`

BlackGlory:
Fix https://github.com/denoland/deno_std/issues/4190

### DIFF
--- a/expect/_matchers.ts
+++ b/expect/_matchers.ts
@@ -432,7 +432,7 @@ export function toMatch(
 
 export function toMatchObject(
   context: MatcherContext,
-  expected: Record<PropertyKey, unknown>,
+  expected: Record<PropertyKey, unknown> | Array<Record<PropertyKey, unknown>>,
 ): MatchResult {
   if (context.isNot) {
     let objectMatch = false;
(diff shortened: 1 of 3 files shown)


expect/_matchers.ts
```diff
@@ -432,15 +432,15 @@ export function toMatch(
 
 export function toMatchObject(
   context: MatcherContext,
-  expected: Record<PropertyKey, unknown>,
+  expected: Record<PropertyKey, unknown> | Array<Record<PropertyKey, unknown>>,
```
iuioiua, line 435:
```suggestion
  expected: Record<PropertyKey, unknown> | Record<PropertyKey, unknown>[],
```
Nit

iuioiua (changes requested):
Thanks for your contribution! Can you please run `deno fmt`?

iuioiua (approved):
LGTM. Thank you. @kt3k, should we also update `assertObjectMatch()` in a similar way?

kt3k:
> LGTM. Thank you. @kt3k, should we also update assertObjectMatch() in a similar way?

I don't think so. `toMatchObject` needed to be aligned to jest, but `assertObjectMatch` doesn't need to be aligned to anything.

PR #4204, opened 2024-01-15
refactor(semver): replace `parseComparator()` with comparator objects

timreichen:


iuioiua (changes requested):
Mostly LGTM. Are you able to add return types to the affected functions? It'd make the purpose of this PR a little clearer.

timreichen:
> Mostly LGTM. Are you able to add return types to the affected functions? It'd make the purpose of this PR a little clearer.

done

iuioiua (approved):
LGTM. I notice that `parseRangeString()` below uses `parseComparator()`. Has that been left out deliberately too?

timreichen:
> LGTM. I notice that `parseRangeString()` below uses `parseComparator()`. Has that been left out deliberately too?

Yes, because if `OPERATOR_XRANGE_REGEXP` does not match, it needs to be matched with `COMPARATOR_REGEXP`. Maybe we can try to combine these two regexps in a future PR.

Issue #3888, opened 2023-12-01
discussion: to type-check or not to type-check


iuioiua:
Most functions/methods in the Standard Library do not check the type-check arguments at runtime. However, [some](https://github.com/denoland/deno_std/blob/451efa6243759031be85c55c927cbe5d04472afe/path/windows/extname.ts#L14) do. Aiming for consistency, this begs the question: should we type-check arguments at runtime or not throughout the Standard Library? This question is mostly important for transpilation of browser-compatible modules for the browser. E.g. Fresh islands. There are 3 prominent possible approaches and arguments.

**1. Type-check all modules**
The benefit is that all code, when transpiled, is type-guarded. This is nice but would incur a considerable engineering cost, especially if done codebase-wide. The risk is if this is done but not needed or demanded, that would mean a substantial ongoing engineering cost. This is what Node does. All code would also become dependent on `std/assert`.

**2. Don't type-check any modules (for now)**
It's possible that the demand for type-checking is zero or near zero. In that case, the main benefit of this approach is that it could avoid needless engineering costs. It's also possible that, in the future, there _will be_ a demand for type-checking expressed by the community. This approach would allow us to re-evaluate while being more informed in the future. We would have to clearly state, somewhere in the README, that no type-checking is done, and this fact should be considered for those transpiling code for the browser.

**3. Only type-check web-compatible modules**
This approach meets in the middle of the other two. An engineering cost would be incurred on browser-compatible modules but not on the remainder of the modules.

I'm most in favour of option 2. It's simple and avoids wasting any time on possibly needless type-checking, allowing us to make a more informed decision in the future. It'd require us to perform some cleanups around the codebase, but that wouldn't take too long.

Community, what do we think? Are there other unexplored options? Please let us know.

kt3k:
> It's possible that the demand for type-checking is zero or near zero.

I disagree with this. Runtime type-checking is not a kind of feature which users immediately need when they start using these APIs, but it's a feature which helps debugging very difficult bugs that were thrown from deep in the stack. The lack of demand for it can't be a reason for removing them in my view.

TypeScript is not strictly typed language. There always be chances for variables to be in wrong types in runtime even when the users completely follow the TS types. (Many builtin APIs return `any` type, typically json related APIs. Those values are not validated by TS typing system at all, and these can be passed to any APIs with absurd types. Runtime type checking helps preventing to throw very enigmatic errors messages when such situation happens)

jcayzac:
[...]

PR #4205, opened 2024-01-15
refactor(semver): remove runtime string type check in `parse()`

timreichen:
- removes runtime string type check in `parse()`
- refactors the function code


iuioiua:
Related #3888

iuioiua (approved):
LGTM and I agree with the change. Though, Yoshiya might not.

kt3k:
Now if you pass `undefined` to `parse`, the error message becomes `TypeError: Cannot read properties of undefined (reading 'length')` which is less intelligible than the current message `TypeError: version must be a string`. That is worse DX.

iuioiua:
> Now if you pass `undefined` to `parse`, the error message becomes `TypeError: Cannot read properties of undefined (reading 'length')` which is less intelligible than the current message `TypeError: version must be a string`. That is worse DX.

I'm not disagreeing but it's worth noting that this only happens when types are ignored, such as when transpiled to JavaScript in browsers.

timreichen:
> Now if you pass `undefined` to `parse`, the error message becomes `TypeError: Cannot read properties of undefined (reading 'length')` which is less intelligible than the current message `TypeError: version must be a string`. That is worse DX.

This particular check in this function is just random imo. There are lots of exported functions in semver that take strings, none of them have the runtime checks...
I think types should be enough. The type error states `Argument of type 'undefined' is not assignable to parameter of type 'string'`.

kt3k:
I don't think type check by TS is enough. Deno supports JS by default, and if you use .js, you don't get type check error. Also there are many builtin APIs that return `any` values (e.g. `JSON.parse`, `response.json()`), and those values can be passed to `parse` without type error.

timreichen:
> I don't think type check by TS is enough. Deno supports JS by default, and if you use .js, you don't get type check error. Also there are many builtin APIs that return `any` values (e.g. `JSON.parse`, `response.json()`), and those values can be passed to `parse` without type error.

Ok, I'll close this PR. 
I think we would profit greatly by having a std-wide solution for runtime checks. I could imagine a tool that checks all exported functions and adds missing runtime checks, at least for basic types. Something along the lines of:

```ts
export function someFunction(arg1: string, arg2?: number) {
 ...
}
```
=>
```ts
export function someFunction(arg1: string, arg2?: number) {
  runtimeTypeCheck("arg1", arg1, ["string"]);
  runtimeTypeCheck("arg2", arg2, ["number", "undefined"]);
}
```
WDYT?

iuioiua:
> > I don't think type check by TS is enough. Deno supports JS by default, and if you use .js, you don't get type check error. Also there are many builtin APIs that return `any` values (e.g. `JSON.parse`, `response.json()`), and those values can be passed to `parse` without type error.
> 
> Ok, I'll close this PR. I think we would profit greatly by having a std-wide solution for runtime checks. I could imagine a tool that checks all exported functions and adds missing runtime checks, at least for basic types. Something along the lines of:
> 
> ```ts
> export function someFunction(arg1: string, arg2?: number) {
>  ...
> }
> ```
> 
> =>
> 
> ```ts
> export function someFunction(arg1: string, arg2?: number) {
>   runtimeTypeCheck("arg1", arg1, ["string"]);
>   runtimeTypeCheck("arg2", arg2, ["number", "undefined"]);
> }
> ```
> 
> WDYT?

A point of discussion for #3888.

PR #4207, opened 2024-01-16
fix(flags): correct deprecation notices

### LINKED ISSUE #4203: `parse` of `std/flags` is not marked deprecated in API doc
https://deno.land/std@0.212.0/flags/mod.ts?s=parse

iuioiua:
Fixes #4203

Normally, we'd call `parseArgs()` within `parse()`. However, `parseArgs()` is so complex with all its type arguments, one cannot simply do so. I figured the next best thing to do is copy the entire `std/cli/parse_args.ts` implementation. I did this instead of reverting to the old `std/flags` code, as the new code has minor tweaks that enable the `noUncheckedIndexedAccess` compiler option.

### DIFF
--- a/flags/mod.ts
+++ b/flags/mod.ts
@@ -2,8 +2,6 @@
 // This module is browser compatible.
 
 /**
- * @deprecated (will be removed in 1.0.0) Import from `std/cli/parse_args.ts` instead
- *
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #4208, opened 2024-01-16
fix(fmt): correct `stripColor()` deprecation notice

iuioiua:
This change fixes how `stripColor()`'s signature is documented by calling `stripAnsiCode()` within it.

Before:
image

After:
image


kt3k (approved):
LGTM. Nice!

PR #4211, opened 2024-01-16
test(toml): improve test coverage

babiabeo:
Related: #3713

### DIFF
--- a/toml/_parser.ts
+++ b/toml/_parser.ts
@@ -894,7 +894,7 @@ export function ParserFactory<T>(parser: ParserComponent<T>) {
           if (count > line.length) {
             count -= line.length + 1;
           } else {
-            return count;
+            break;
           }
         }
         return count;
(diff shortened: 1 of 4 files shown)


toml/_parser.ts
```diff
@@ -894,7 +894,7 @@ export function ParserFactory<T>(parser: ParserComponent<T>) {
           if (count > line.length) {
             count -= line.length + 1;
           } else {
-            return count;
+            break;
```
babiabeo, line 897: I use `break` here because `subStr.split("\n")` always returns a non-empty array. Therefore, the `for` loop is always called, which makes the line 900 unreachable.

toml/test.ts
```diff
@@ -762,3 +763,19 @@ Deno.test({
     assertEquals(actual, expected);
   },
 });
+
+Deno.test({
+  name: "stringify throws on invalid value",
```
kt3k, line 768:
```suggestion
  name: "stringify() throws on invalid value",
```

PR #4212, opened 2024-01-16
refactor(console): rename `_rle` to `_run_length.ts`

timreichen:
This PR renames `_rle` to `_run_length.ts`.

I have no idea what these exported functions actually do. It might be a good idea to add jsdoc as well.


kt3k:
> I have no idea what these exported functions actually do. It might be a good idea to add jsdoc as well.

cc @lionel-rowe

kt3k (approved):
LGTM. Thanks

PR #4214, opened 2024-01-17
fix(log): make `flattenArgs()` private

### LINKED ISSUE #4175: suggestion: remove `flattenArgs()` from `log/formatters.ts` public API
This function doesn't seem useful as part of the public API.

iuioiua:
Closes #4175

Technically, this is a breaking change. However, I doubt anyone would be using this function.


iuioiua:
> Maybe let's tag this as `fix` by considering this export as a bug.

Done

Issue #4216, opened 2024-01-17
todo: cleanup deprecated `Deno.*` APIs


iuioiua:
The following APIs are either deprecated or soon-to-be-deprecated, and their use should be removed from the Standard Library:

- [x] Deno.copy
- [x] Deno.iter
- [x] Deno.iterSync
- [x] Deno.File
- [x] Deno.Buffer
- [x] Deno.readAll 
- [x] Deno.readAllSync
- [x] Deno.writeAll 
- [x] Deno.writeAllSync
- [x] Deno.metrics
- [x] Deno.FsWatcher.return
- [x] Deno.run
- [x] Deno.serveHttp
- [x] Deno.customInspect
- [x] Deno.Server
- [x] Deno.ListenTlsOptions.certFile
- [x] Deno.ListenTlsOptions.keyFile
- [x] Deno.ConnectTlsOptions.certFile
- [x] Deno.read 
- [x] Deno.readSync
- [x] Deno.write 
- [x] Deno.writeSync
- [x] Deno.seek
- [x] Deno.seekSync
- [x] Deno.fdatasync
- [x] Deno.fdatasyncSync
- [x] Deno.fstat
- [x] Deno.fstatSync
- [x] Deno.ftruncate
- [x] Deno.ftruncateSync
- [x] Deno.futime
- [x] Deno.futimeSync
- [x] Deno.close 
- [x] Deno.isatty
- [x] Deno.shutdown

CC @bartlomieju

iuioiua:
- `Deno.serveHttp()`: Used in `std/http`'s server implementation, which will soon be deprecated. This may be fine to leave as-is under the assumption that the server implementation is removed before `Deno.serveHttp()`. https://github.com/denoland/deno_std/blob/e3cb0cc3bfc03532b1168792f8be0a445a6ae969/http/server.ts#L426
- `Deno.close()`: Used in `std/webgpu` to cleanup resources linked to `GPUDevice` in tests. According to the comment, the WebGPU spec lacks a means of cleaning up these resources. I'm not sure if the situation has changed recently. @lucacasonato, anything to add? https://github.com/denoland/deno_std/blob/e3cb0cc3bfc03532b1168792f8be0a445a6ae969/webgpu/_test_util.ts#L39
[...]

PR #4217, opened 2024-01-17
docs(streams): remove `Deno.metrics()` use in example

iuioiua:
As `Deno.metrics()` is deprecated.

Towards #4216


kt3k (approved):
LGTM. Good catch!

PR #4218, opened 2024-01-17
feat(collections): Add `pick` and `omit`

lambdalisue:
This is like Lodash's [pick](https://lodash.com/docs/#pick) and [omit](https://lodash.com/docs/#omit), but it's a bit more restrictive. Specifically, it only accepts an array of strings (`string[]`) rather than just a single string.

### DIFF
--- a/collections/omit.ts
+++ b/collections/omit.ts
@@ -15,8 +15,8 @@
  * ```
  */
 export function omit<T extends object, K extends keyof T>(
-  obj: T,
-  keys: K[],
+  obj: Readonly<T>,
+  keys: readonly K[],
 ): Omit<T, K> {
   const excludes = new Set(keys);
   const has = excludes.has.bind(excludes);
--- a/collections/omit.ts
+++ b/collections/omit.ts
@@ -1,0 +1,26 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Creates a new object by excluding the specified keys from the provided object.
+ *
+ * @example
+ * ```ts
+ * import { omit } from "https://deno.land/std@$STD_VERSION/collections/omit.ts";
+ * import { assertEquals } from "https://deno.land/std@$STD_VERSION/assert/assert_equals.ts";
+ *
+ * const obj = { a: 5, b: 6, c: 7, d: 8 };
+ * const omitted = omit(obj, ["a", "c"]);
+ *
+ * assertEquals(omitted, { b: 6, d: 8 });
+ * ```
+ */
+export function omit<T extends object, K extends keyof T>(
+  obj: T,
+  keys: K[],
+): Omit<T, K> {
+  const excludes = new Set(keys);
+  const has = excludes.has.bind(excludes);
+  return Object.fromEntries(
+    Object.entries(obj).filter(([k, _]) => !has(k as K)),
+  ) as Omit<T, K>;
+}
--- a/collections/pick.ts
+++ b/collections/pick.ts
@@ -1,0 +1,22 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * Creates a new object by including the specified keys from the provided object.
+ *
+ * @example
+ * ```ts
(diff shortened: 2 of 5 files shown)


timreichen:
Isn't this very similar to `filterKeys()`?

kt3k:
> Isn't this very similar to filterKeys()?

Look like a bit restricted versions of `filterKeys` with better typing when `T` and `K` are more specific than just `Record<string, X>` and `string`

I'm slightly in favor of including this

lambdalisue:
Ah, I didn't notice that there is [`filterKeys()`](https://deno.land/std@0.212.0/collections/filter_keys.ts?s=filterKeys). Thanks for mention. However, I still prefer `pick` and `omit` while TypeScript natively provides `Pick` and `Omit` type functions so those pair functions are more natural to me.

collections/omit.ts
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * const obj = { a: 5, b: 6, c: 7, d: 8 };
+ * const omitted = omit(obj, ["a", "c"]);
+ *
+ * assertEquals(omitted, { b: 6, d: 8 });
+ * ```
+ */
+export function omit<T extends object, K extends keyof T>(
+  obj: T,
+  keys: K[],
```
petamoriken, line 19: Since this function does not perform mutable operations on its arguments, it would be better to explicitly type it.

```suggestion
  obj: Readonly<T>,
  keys: readonly K[],
```

collections/pick.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * const obj = { a: 5, b: 6, c: 7, d: 8 };
+ * const picked = pick(obj, ["a", "c"]);
+ *
+ * assertEquals(picked, { a: 5, c: 7 });
+ * ```
+ */
+export function pick<T extends object, K extends keyof T>(
+  obj: T,
+  keys: K[],
```
petamoriken, line 19: ditto

```suggestion
  obj: Readonly<T>,
  keys: readonly K[],
```

kt3k (approved):
Looks like we have more positive feedbacks to this.

LGTM

Leokuma:
I'm -1 for the following reasons:

1. What Omit and Pick do is too similar to `filterKeys`. We'll end up with 3 functions that do the same thing but have completely different naming criteria and usage.
2. While the names "Omit" and "Pick" align with TS, they disalign with JS' `Array.filter`. From that perspective, `filterKeys` sounds more consistent. Also, TS' Omit and Pick are used in the context of typing, not runtime. In the context of typing, the names "Pick" and "Omit" sound intuitive, but in runtime it's not so obvious what they do if you don't take TS into account, e.g. it's not clear that they deal with keys and not values.

Now that I'm thinking about it, I'm not even sure why we have `filterKeys` _and_ `filterValues` _and_ `filterEntries`. Couldn't we have only `filterEntries`, renamed to just `filter`? And its callback would take 2 args (key and value) instead of one array, just like `Array.filter`. Sorry if this has already been discussed. I think `std/collections` already has quite a lot of functions.

lambdalisue:
@Leokuma 

Thank you for providing your perspective. The statement below does not align with the discussed topic but reflects my personal view. Both filterKeys and filter offer a predicate function, leading to a non-deterministic return type. While Deno is undoubtedly a JavaScript runtime, it also functions as a TypeScript runtime, and I believe it should offer functions relevant to the TypeScript context. Consequently, I disagree with the notion that it can be replaced solely based on the existence of filterKeys-type functions.

Leokuma:
Sorry, I didn't notice the return type was different from `filterKeys`. In this case I'm in favor of this because it's quite cumbersome to type the return value of `filterKeys` in that use case. And it makes sense to align it with TS.

My question about `filterKeys`, `filterEntries` and `filterValues` still stands, but I can open another issue for that.

iuioiua (approved):
I like the idea! Cleanly done. LGTM.

kt3k:
Landing as there seem only supportive opinions from the community and the core team.

@lambdalisue Thanks for your contribution!

PR #4223, opened 2024-01-19
docs(toml): complete documentation

sylc:
towards https://github.com/denoland/deno_std/issues/3764

### DIFF
--- a/toml/stringify.ts
+++ b/toml/stringify.ts
@@ -21,7 +21,7 @@ type ArrayType =
   | "MIXED";
 
 /**
- * Formating Options for {@linkcode stringify}
+ * Formatting Options for {@linkcode stringify}
  */
 export interface FormatOptions {
   /** Define if the keys should be aligned or not */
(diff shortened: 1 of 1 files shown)


sylc:
The toml/mod module comment has an @example. But i'm not sure how useful that is from within intellisense. Should the example be duplicated on the 2 functions parse and stringify themselves ? 
(also the main module comment uses markdown. is that ok?)

kt3k:
> Should the example be duplicated on the 2 functions parse and stringify themselves ?

Duplicated example for parse and stringify functions are also useful in my view. Such examples will appear in html doc pages like this one https://deno.land/std@0.212.0/toml/mod.ts?s=parse

> also the main module comment uses markdown. is that ok?

Markdown comments are fine. They are rendered nicely in html doc pages like [this page](https://deno.land/std@0.212.0/toml/mod.ts).

kt3k (approved):
LGTM. Thanks!

PR #4228, opened 2024-01-23
deprecation(semver): deprecate `ltr()` and `gtr()`

timreichen:
- ~~renames `ltr()` to `lessThanRange()` and deprecates the old function~~
- ~~renames `gtr()` to `greaterThanRange()` and deprecates the old function~~
- deprecates `ltr()`
- deprecates `gtr()`


semver/greater_than_range.ts
```diff
@@ -0,0 +1,12 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { Range, SemVer, SemVerRange } from "./types.ts";
+import { rangeMax } from "./range_max.ts";
+import { greaterThan } from "./greater_than.ts";
+
+/** Checks to see if the version is greater than all possible versions of the range. */
```
iuioiua, line 6: If one asks, "Is this version greater than the range?" perhaps it's worth swapping the argument order. Ditto with the other new functions. What do we think?

semver/gtr.ts
```diff
@@ -1,12 +1,11 @@
...
-  version: SemVer,
-  range: SemVerRange | Range,
-): boolean {
-  return gt(version, rangeMax(range));
+/**
+ * Checks to see if the version is greater than all possible versions of the range.
+ * @deprecated (will be removed after 0.215.0) Use {@linkcode greaterThanRange} instead.
+ */
+export function gtr(version: SemVer, range: SemVerRange | Range): boolean {
+  return greaterThanRange(range, version);
```
iuioiua, line 10: Can you please use the new [`warnOnDeprecatedApi()`](https://github.com/denoland/deno_std/blob/0a32c8f5e82f939e8a46f8c027c1fb5b3040590a/internal/warn_on_deprecated_api.ts#L33)? Ditto with other deprecated functions.

iuioiua (changes requested):
I agree with the changes. Nice.

semver/greater_than_range.ts
timreichen: My train of thought was to pass the context element first, e.g. the range. But that makes a lot of sense.

semver/gtr.ts
timreichen: I'll have a look

kt3k:
I wonder what are the use cases of these functions. I guess most users are only interested in `testRange` for comparing a version and a range. In what cases, would people try checking these conditions?

timreichen:
> I wonder what are the use cases of these functions. I guess most users are only interested in `testRange` for comparing a version and a range. In what cases, would people try checking these conditions?

I have no idea. I think it is pretty trivial to get the min/max version of a range and compare it to another version. But I guess there must be a reason why it was ported over.

semver/gtr.ts
timreichen: I like that new API, but this seems like the addition of this should be automated to stay in sync with`@deprecated` tags.

iuioiua: I agree, and I think we should explore that idea in a separate issue. But for now, can we please use `warnOnDeprecatedApi()`?

timreichen: done

kt3k:
> But I guess there must be a reason why it was ported over.

I feel we blindly ported everything from `npm:semver` and these wasn't a discussion about utilities of these APIs.

I'd suggest we should just deprecate these without replacement APIs because they don't seem common operation and also the alternative ways to perform these are trivial as you point out. The deprecation note would be something like `use greaterThan(version, rangeMax(range)) instead`

timreichen:
> > But I guess there must be a reason why it was ported over.
> 
> I feel we blindly ported everything from `npm:semver` and these wasn't a discussion about utilities of these APIs.
> 
> I'd suggest we should just deprecate these without replacement APIs because they don't seem common operation and also the alternative ways to perform these are trivial as you point out. The deprecation note would be something like `use greaterThan(version, rangeMax(range)) instead`

done

semver/gtr.ts
```diff
@@ -1,12 +1,22 @@
...
 import type { Range, SemVer, SemVerRange } from "./types.ts";
 import { rangeMax } from "./range_max.ts";
-import { gt } from "./gt.ts";
+import { greaterThan } from "./greater_than.ts";
+import { warnOnDeprecatedApi } from "../internal/warn_on_deprecated_api.ts";
 
-/** Checks to see if the version is greater than all possible versions of the range. */
+/**
+ * Checks to see if the version is greater than all possible versions of the range.
+ * @deprecated (will be removed after 0.215.0) Use `greaterThan(version, rangeMax(range))` instead.
```
kt3k, line 9:
```suggestion
 * @deprecated (will be removed after 0.217.0) Use `greaterThan(version, rangeMax(range))` instead.
```

semver/ltr.ts
```diff
@@ -1,12 +1,22 @@
...
 import type { Range, SemVer, SemVerRange } from "./types.ts";
-import { lt } from "./lt.ts";
+import { lessThan } from "./less_than.ts";
 import { rangeMin } from "./range_min.ts";
+import { warnOnDeprecatedApi } from "../internal/warn_on_deprecated_api.ts";
 
-/** Less than range comparison */
+/**
+ *  Less than range comparison
+ * @deprecated (will be removed after 0.215.0) Use `lessThan(version, rangeMin(range))` instead.
```
kt3k, line 9:
```suggestion
 * @deprecated (will be removed after 0.217.0) Use `lessThan(version, rangeMin(range))` instead.
```

iuioiua:
Towards #4228

iuioiua (approved):
LGTM. I've made some modifications and opened a new issue (#4273) to track the deprecation. Does this new way of deprecation look good to us?

kt3k:
> LGTM. I've made some modifications and opened a new issue (https://github.com/denoland/deno_std/issues/4273) to track the deprecation. Does this new way of deprecation look good to us?

Thanks! #4273 Looks good to me.

kt3k (approved):
LGTM

PR #4233, opened 2024-01-24
fix: ignore linting rule for `Deno.resources()`

iuioiua:


iuioiua (commented):
Should fail on stable but pass on canary.

bartlomieju (approved):

PR #4238, opened 2024-01-25
BREAKING(streams): remove `readAll()`, `writeAll()` and `copy()`

iuioiua:
These functions now live in `std/io`. This includes synchronous versions.


kt3k (approved):
LGTM. Thanks!

PR #4240, opened 2024-01-25
chore(tools): add check flag to `browser_compat.ts`

timreichen:
- adds check flag to the `browser_compat.ts` tool to check missing comments in file. It will inject missing comments if the flag is not present.
This feature is borrowed from the `check_licence.ts` tool.


_tools/check_browser_compat.ts
```diff
@@ -26,24 +29,24 @@ function isBrowserCompatible(filePath: string): boolean {
...
+  if (isBrowserCompatible(path) && !hasBrowserCompatibleComment(source)) {
+    if (CHECK) {
+      console.log(
+        `${path} is likely browser-compatible and can have the "${DECLARATION}" comment added.`,
+      );
+    } else {
+      const index = source.indexOf(COPYRIGHT);
+      await Deno.writeTextFile(
+        path,
+        source.slice(0, index + COPYRIGHT.length) + "\n" + DECLARATION + "\n" +
```
kt3k, line 47: The second `"\n"` looks redundant to me.

_tools/check_browser_compat.ts
```diff
@@ -6,11 +6,14 @@
...
  * Run using: deno run --allow-read --allow-run _tools/check_browser_compat.ts
  */
-import { walkSync } from "../fs/walk.ts";
+
+import { walk } from "../fs/walk.ts";
+import { COPYRIGHT } from "./check_licence.ts";
 
 const ROOT = new URL("../", import.meta.url);
 const SKIP = [/(test|bench|\/_|\\_|testdata|version.ts)/];
 const DECLARATION = "// This module is browser compatible.";
```
kt3k, line 15: Not directly related to this PR, but some file (`data_structures/comparators.ts`) seems having `/** This module is browser compatible. */` comment

_tools/check_browser_compat.ts
iuioiua: This section makes me wonder whether we should integrate both copyright and web compatibility checks. Either way, an idea for another issue/PR.

iuioiua (commented):
Great idea!

_tools/check_browser_compat.ts
timreichen: > The second `"\n"` looks redundant to me.

removed

_tools/check_browser_compat.ts
timreichen: Ok, I guess we should replace that with a single line comment in another PR.

_tools/check_browser_compat.ts
timreichen: > This section makes me wonder whether we should integrate both copyright and web compatibility checks. Either way, an idea for another issue/PR.

I agree, this looks ugly and hacky. I had a try with the ts compiler API which would work without direct string manipulation, but the ts printer doesn't retain format and whitespaces. The same problem appeared on a draft of auto-inserting `warnOnDeprecatedApi()` calls for deprecated functions. Is there a simple ts parser that also retains format and whitespaces?

kt3k (approved):
LGTM

PR #4242, opened 2024-01-26
fix(expect): expect.toEqual need to return original obj

eryue0220:
If support `addEqualityTester`, this case will fail, because it return a new object.

```js
import { expect } from "https://deno.land/std/expect/mod.ts";

class Volume { /* specify logic*/ }

function createVolume() {
  return new Volume();
}

function areVolumeEqual(a: unknown, b unknown) {
   return a instacneof Volume && b instanceof Volume
}

Deno.test('custom equality tester', () => {
    expect.addEqualityTester([areVolumeEqual]);
    const volume1 = createVolume();
    const volume2 = createVolume();
    expect(volume1).toEqual(volume2);
});
```

And current ` expect(source1).toEqual(source2)` will return `false`.

### DIFF
--- a/expect/_matchers.ts
+++ b/expect/_matchers.ts
@@ -37,15 +37,16 @@ function filterUndefined(obj: any): any {
   if (Symbol.iterator in obj) return obj;
   if (Array.isArray(obj)) return obj.map(filterUndefined);
 
+  const result = Object.create(obj);
   for (const key in obj) {
     const val = obj[key];
     if (val === undefined) {
-      delete obj[key];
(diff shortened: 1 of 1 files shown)


kt3k:
This modifies the input object. That doesn't look a good idea.

eryue0220:
> This modifies the input object. That doesn't look a good idea.

Yeah, I'll consider other solutions.

eryue0220:
If the #4246 is merged, this pr will be solved as well. And it's no need any more.

Issue #4244, opened 2024-01-27
expect.toEquals() class instance check different from jest


marvinhagemeister:
**Describe the bug**

Asserting class instances is different in jest's `expect` library compared to ours.

**Steps to Reproduce**

```ts
import { expect } from "https://deno.land/std@0.213.0/expect/mod.ts";

function create() {
  class Person {
    constructor(public readonly name = "deno") {}
  }
  return new Person();
}

Deno.test("should work", () => {
  expect([create()]).toEqual([create()]);
});
```

**Expected behavior**

The test should pass like it does in `jest`.

**Environment**

- OS: macOS 14.2.1
- deno version: 1.40.2
- std version: 0.213.0

vinc3re:
The issue lies in comparing the constructors of the objects of both `Person` objects. I tested it with the following code:

```ts
import { assertEquals } from "std/assert";

function create() {
  class Person {
    constructor(public readonly name = "deno") {}
  }
  return new Person();
}

const a = [create()];
const b = [create()];

Deno.test("should work", () => {
  assertEquals(a, b);
});
```

Using `assertEquals()`.

`assertEquals()` calls `equals()`, so the `equals()` method is our target. In the first call of `equals`, both arrays are passed. The sequence is:

1. Successfully validates that they are objects at [line 54](https://github.com/denoland/deno_std/blob/aad99992653b45d9bd09ccba1f645accb22f5cf4/assert/equal.ts#L54) with `if (a && typeof a === "object" && b && typeof b === "object")`. This is true for both arrays.

2. Goes through all other `if` checks (with `false`) until it reaches [line 104](https://github.com/denoland/deno_std/blob/aad99992653b45d9bd09ccba1f645accb22f5cf4/assert/equal.ts#L104), where `compare()` is recursively called with the `Person` object on each side.

   ```ts
   //...
   const a = [create()];
   const b = [create()];

   console.log(a["0"]);
   console.log(b["0"]);

   //output...

   //Person { name: "deno" }
   //Person { name: "deno" }
   ```

   Before that, of course, the merge of both arrays is done.

   ```ts
   //...
   const a = [create()];
   const b = [create()];

   console.log({ ...a, ...b });

   //output...
   //{ "0": Person { name: "deno" } }
   ```

   And the merging of properties:

   ```ts
   const a = [create()];
   const b = [create()];
   const merge = { ...a, ...b };
   const propMerge = [
     ...Object.getOwnPropertyNames(merge),
[...]

PR #4246, opened 2024-01-28
fix(expect): align `toEqual` to jest

eryue0220:
This pr is to solve #4244

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -3,12 +3,6 @@ function isKeyedCollection(x: unknown): x is Set<unknown> {
   return [Symbol.iterator, "size"].every((k) => k in (x as Set<unknown>));
 }
 
-function constructorsEqual(a: object, b: object) {
-  return a.constructor === b.constructor ||
-    a.constructor === Object && !b.constructor ||
-    !a.constructor && b.constructor === Object;
-}
-
 /**
  * Deep equality comparison used in assertions
  * @param c actual value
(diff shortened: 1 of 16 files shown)


kt3k:
Thanks for trying this, but this is significant breaking change to `assertEquals` and we can't merge this.

Fundamental issue here is that we share the same logic for `assertEquals` and `expect().toEqual()` but they have many incompatibilities in edge cases.

I think it's time to fork `assertEquals` for `std/expect` to safely diverge these 2 different APIs. Can you copy the necessary part of `assertEquals` under `expect/` as internal utilities and apply these change to them?

eryue0220:
> Thanks for trying this, but this is significant breaking change to `assertEquals` and we can't merge this.
> 
> Fundamental issue here is that we share the same logic for `assertEquals` and `expect().toEqual()` but they have many incompatibilities in edge cases.
> 
> I think it's time to fork `assertEquals` for `std/expect` to safely diverge these 2 different APIs. Can you copy the necessary part of `assertEquals` under `expect/` as internal utilities and apply these change to them?

Sure, I'll update this.

eryue0220:
I‘ve update the `std/expect`, which will refers to `toEqual` and `toStrictEqual`. Besides fix the issues #4244 cases, this pr also align some other cases to jest.

kt3k:
Looks great! Thanks for these changes.

I'd like to point one another thing. The below doesn't pass in Jest, but it passes with this change.
```js
class A {}
class B {}
expect(new A()).toStrictEqual(new B());
```

I think we need to add constructor check if `strictCheck` is true.

eryue0220:
> Looks great! Thanks for these changes.
> 
> I'd like to point one another thing. The below doesn't pass in Jest, but it passes with this change.
> 
> ```js
> class A {}
> class B {}
> expect(new A()).toStrictEqual(new B());
> ```
> 
> I think we need to add constructor() check if `strictCheck` is true.

Thank you,

I've fixed this problem.

kt3k (approved):
LGTM. This looks great improvements towards jest compatibility. Thanks for working on this!

PR #4247, opened 2024-01-29
feat(io): re-introduce `iterate()` and `iterateSync()`

iuioiua:
While working on deprecation documentation, I realised that there's no new means for synchronous iteration for the `ReaderSync` interface. I thought we might as well un-deprecate this functionality, along with its asynchronous sibling.

Note: this hasn't yet been decided upon.

### DIFF
--- a/streams/iterate_reader.ts
+++ b/streams/iterate_reader.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { DEFAULT_BUFFER_SIZE } from "./_common.ts";
+import { iterate, iterateSync } from "../io/iterate.ts";
 import type { Reader, ReaderSync } from "../io/types.ts";
 
 export type { Reader, ReaderSync };
@@ -35,24 +35,15 @@ export type { Reader, ReaderSync };
  * }
  * ```
  *
- * @deprecated (will be removed after 1.0.0) Use {@linkcode ReadableStreamDefaultReader} instead.
+ * @deprecated (will be removed in 0.216.0) Use {@linkcode iterate} instead.
  */
-export async function* iterateReader(
+export function iterateReader(
   r: Reader,
   options?: {
     bufSize?: number;
   },
 ): AsyncIterableIterator<Uint8Array> {
-  const bufSize = options?.bufSize ?? DEFAULT_BUFFER_SIZE;
-  const b = new Uint8Array(bufSize);
-  while (true) {
-    const result = await r.read(b);
-    if (result === null) {
-      break;
-    }
-
-    yield b.slice(0, result);
-  }
+  return iterate(r, options);
 }
 
 /**
@@ -87,22 +78,13 @@ export async function* iterateReader(
  * responsibility to copy contents of the buffer if needed; otherwise the
  * next iteration will overwrite contents of previously returned chunk.
  *
- * @deprecated (will be removed after 1.0.0) Use {@linkcode ReadableStream} instead.
+ * @deprecated (will be removed in 0.216.0) Use {@linkcode iterateSync} instead.
  */
-export function* iterateReaderSync(
+export function iterateReaderSync(
   r: ReaderSync,
   options?: {
     bufSize?: number;
   },
 ): IterableIterator<Uint8Array> {
-  const bufSize = options?.bufSize ?? DEFAULT_BUFFER_SIZE;
-  const b = new Uint8Array(bufSize);
-  while (true) {
-    const result = r.readSync(b);
-    if (result === null) {
-      break;
-    }
-
-    yield b.slice(0, result);
-  }
+  return iterateSync(r, options);
 }
--- a/io/iterate.ts
+++ b/io/iterate.ts
@@ -1,0 +1,104 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+import { DEFAULT_BUFFER_SIZE } from "./_constants.ts";
+import type { Reader, ReaderSync } from "./types.ts";
+
+export type { Reader, ReaderSync };
+
+/**
+ * Turns a {@linkcode Reader} into an async iterator.
+ *
+ * @example
+ * ```ts
+ * import { iterate } from "https://deno.land/std@$STD_VERSION/io/iterate.ts";
+ *
+ * using file = await Deno.open("/etc/passwd");
+ * for await (const chunk of iterate(file)) {
+ *   console.log(chunk);
+ * }
+ * ```
+ *
+ * Second argument can be used to tune size of a buffer.
+ * Default size of the buffer is 32kB.
+ *
+ * @example
+ * ```ts
+ * import { iterate } from "https://deno.land/std@$STD_VERSION/io/iterate.ts";
+ *
+ * using file = await Deno.open("/etc/passwd");
+ * const iter = iterate(file, {
+ *   bufSize: 1024 * 1024
+ * });
+ * for await (const chunk of iter) {
+ *   console.log(chunk);
+ * }
+ * ```
+ */
+export async function* iterate(
+  reader: Reader,
+  options?: {
+    bufSize?: number;
+  },
+): AsyncIterableIterator<Uint8Array> {
+  const bufSize = options?.bufSize ?? DEFAULT_BUFFER_SIZE;
+  const b = new Uint8Array(bufSize);
+  while (true) {
+    const result = await reader.read(b);
+    if (result === null) {
+      break;
+    }
+
+    yield b.slice(0, result);
+  }
+}
+
+/**
+ * Turns a {@linkcode ReaderSync} into an iterator.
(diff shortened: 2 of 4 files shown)


kt3k:
These used to be called `iterateReader` and `iterateReaderSync` (https://github.com/denoland/deno_std/blob/aad99992653b45d9bd09ccba1f645accb22f5cf4/streams/iterate_reader.ts). Is it intentional to change these names?

iuioiua:
Yes. I slightly liked the name more. This is not a strong opinion. Happy to change it back if we'd like.

kt3k:
They were originally `Deno.iter` and `Deno.iterSync`, and then moved to std and renamed to `iterateReader` and `iterateReaderSync` (These names were suggested by @nayeemrmn IIRC).

I don't have strong opinions to them, but I'd like to hear more opinions from the community.

iuioiua:
I've now renamed them to `toIterator()` and `toIteratorSync()` to align it with the adjacent `toReadableStream()` and `toWritableStream()`.

Leokuma:
Is there a risk of ambiguity with `toIterator`? Like maybe there's another function that creates an iterator from something other than Reader. Would the other function also be called `toIterator`, just imported from a different module?

iuioiua:
No risk of ambiguity that I'm aware of.

nayeemrmn:
Almost any unary generator function can be called `toIterator()`. IMO it's better to name them based on what the body of the generator function is doing. Rather than a conversion function, it's a utility for 'iterating the content of' or 'iteratively reading' a `Reader`. So perhaps it should contain a `read` or `iterate` verb.

iuioiua:
But it is a conversion. I'm not too fussed. Which sounds better; `toReaderIterator()` or `iterateReader()`?

Leokuma:
+1 `toReaderIterator`, but maybe `readerToIterator` would be better so that it aligns with `readerFromIterable`?

`iterateReader` doesn't look so good when assigned to a variable. It might give the impression that it's doing something mutable in the `Reader` or that it receives a callback to be run onto `Reader`.

iuioiua:
Ok, I think I've settled for `toReaderIterator()`, which addresses the concerns mentioned above and indicates the purpose of the function.

iuioiua:
Are we happy with the name now?

kt3k:
'ReaderIterator' sounds a bit strange to me as I've never heard of that word combination in js context.

> So perhaps it should contain a `read` or `iterate` verb.

Also does `toReaderIterator` address @nayeemrmn's above comment?

Leokuma:
It could also be `iteratorFromReader`, in addition to my previous suggestion `readerToIterator`.

Personally I'd prefer to have the noun "iterator" over the verb "iterate" because the function doesn't really iterate: it produces an iterator, which can be iterated later or never. It's different from `Array.reduce` for example, which iterates immediately.

iuioiua:
> 'ReaderIterator' sounds a bit strange to me as I've never heard of that word combination in js context.
> 
> > So perhaps it should contain a `read` or `iterate` verb.
> 
> Also does `toReaderIterator` address @nayeemrmn's above comment?

No, but IMO, it shouldn't. The function neither reads nor iterates. @Leokuma makes this same point, which I agree with. The function converts a `Reader` to an iterator. Hence, the name `toReaderIterator()`, which, I feel, conveys its purpose to readers well. Otherwise, `readerToIterator()` might be another alternative.

nayeemrmn:
The problem is the methods on iterator also don't convey that IO is taking place -- they convey 'give me the next element'. That's why it's fine to say it in the generator function's name.

iuioiua:
My opinion on this isn't strong enough to block this PR. Would we be happy with just resorting to the original `iterateReader()`?

Leokuma:
My opinion isn't strong either. I'm fine with `iterateReader`.

iuioiua:
Name restored to the original `iterateReader[Sync]()`.

kt3k:
I'd like to explore some other ideas. How about the below candidates:

- readOn ([to continue reading](https://www.oxfordlearnersdictionaries.com/definition/english/read-on))
- readThrough ([to read (something) from beginning to end](https://www.merriam-webster.com/dictionary/read%20through))
- readEach

iuioiua:
> I'd like to explore some other ideas. How about the below candidates:
> 
> * readOn ([to continue reading](https://www.oxfordlearnersdictionaries.com/definition/english/read-on))
> * readThrough ([to read (something) from beginning to end](https://www.merriam-webster.com/dictionary/read%20through))
> * readEach

I agree with Nayeem that the function name should contain "iterate". The functionality of `readThrough()` could be confused with `readAll()` from `std/io`.

kt3k:
> I agree with Nayeem that the function name should contain "iterate".

Nayeem suggests it should include `read` _or_ `iterate` verb. https://github.com/denoland/deno_std/pull/4247#issuecomment-1919758634

> `readThrough()`'s functionality could be confused with that of `readAll()` from `std/io`.

I think what this API does is similar to `readAll`, but in a different form (Promise vs Async Iterable). So I think giving them similar names isn't too unnatural.

nayeemrmn:
Maybe we should have been comparing this to `readLines()` and `readDelim()` (now deprecated) in the sense that this is the base form of those that uses the raw delimitation. Just call it `readChunks()`.

It also occurred to me that we shouldn't bring back the non-sync version. Why wouldn't that one be deprecated when its aforementioned siblings are, and why do they need to be coupled? We should just bring back `readChunksSync()`.

By the way an old discussion from when `std/streams` was added is how `std/io` would be ugly, in the sense that the non-deprecated stuff there would be a small handful of random utils that barely aren't applicable to 'streams'. You can kinda see that: https://deno.land/std@0.214.0/io/mod.ts. It's even uglier now that `Reader` etc. have been deprecated from the CLI. I think we're only keeping it around to prevent breakages in the ecosystem.

Sorry for the scattered thoughts. I think we should:
- Deprecate `std/io` as a whole. Stop adding new things to it. The interfaces it revolves around are deprecated in the CLI so it has no blessed significance in the Deno ecosystem. We don't have to take the last step of removing it to protect dependents, but I doubt anyone can tell me what the vision or structure for this module is.
- Add `std/streams/sync.ts` for stuff like `readChunksSync()`, `readAllSync()`, `writeAllSync()`. Keep in mind `ReaderSync` and `WriterSync` _are_ 'streams', just not 'web streams'. I argued that to get `iterateReader()` moved there in the past.
- Make these sync methods accept `Deno.FsFile` as an argument instead of `ReaderSync`/`WriterSync`, since nothing else implements these in Deno proper. Maybe.

WDYT @iuioiua @kt3k

kt3k:
> Deprecate std/io as a whole. 

I don't think deprecation (again) of `std/io` is an option here (at least in the new future). We decide to bring `std/io` back in https://github.com/denoland/deno_std/issues/4120 https://github.com/denoland/deno_std/pull/4128 to keep supporting Reader, Writer, and related stuff.

>  I doubt anyone can tell me what the vision or structure for this module is.

The purpose of `std/io` is now to support Reader, Writer, and related operations, which is not primary way to perform I/O in Deno, but they are still supported for the cases where Web Streams are insufficient/inappropariate.

BTW I think this topic is far beyond this PR.

syhol:
Is there any interest in having this be part of the runtime API (say on FsFile) by implementing Symbol.iterator and Symbol.asyncIterator?

I feel this would be in line with the Deno philosophy: aligning with Web APIs and JS primitives. Personally I love working with iterators as I think they are a great abstraction and should be treated as first class.

syhol:
However, If it's out of the question then I would like to throw the following into the ring:
```typescript
iterateChunks();
```

Iterate chunks tells me 2 things I want to know:
1. I'm going to get an iterator from this function
2. That iterator is going to contain chunks

I liked the chunks from `readChunks()` but "read" is too ambiguous for me. I find "iterate" is much clearer.

I noticed a few options used "Reader" but I don't think the input type is too important in the function name because I would aim to optimise for reading and it should already be obvious what the input is e.g. `iterateChunks(file)`. This only becomes an issue if you intend to create a function that iterates chunks for a different type in the same directory. For example, if you wanted to iterate chunks on streams. However you could support both types in the same function.

Alternatively instead of chunks we could have:
- `iterateByteArrays`
- `iterateByteChunks`
- `iterateUint8Arrays`

streams/iterate_reader.ts
```diff
@@ -35,24 +38,15 @@ export type { Reader, ReaderSync };
  * }
  * ```
  *
- * @deprecated (will be removed after 1.0.0) Use {@linkcode ReadableStreamDefaultReader} instead.
+ * @deprecated (will be removed in 1.0.0) Use {@linkcode iterateReader} instead.
```
kt3k, line 41: Maybe we should mention about the new path to import?

iuioiua: Done

kt3k:
Thank you for the suggestions and opinions. We discussed this internally and decided to go with the original names `iterateReader` and `iterateReaderSync` to minimize the ecosystem impact.

kt3k (approved):
LGTM

Issue #4221, opened 2024-01-18
Deno std_assert lacks a method to validate if a native Error object has the same message attribute as another.


vinc3re:
I find myself in the following scenario: I want to validate if a native Error object returned from any function will have the expected `message` attribute. I noticed that Jest can validate this, but the native Deno testing library cannot. Here's an example with Jest:

1. Prepare the project:

   ```bash
   $ mkdir jest-error-assert
   $ cd jest-error-assert
   $ npm init -y
   $ npm i -D jest
   ```

2. Create the `index.js` file in the root:

   ```js
   function getError() {
     return new Error("missing param: name");
   }

   test('[test1] Should return error object with "missing param: name" message', () => {
     const expectErrObj = new Error("missing param: name");
     expect(getError()).toEqual(expectErrObj);
   });

   test('[test2] Should return error object with "missing param: email" message', () => {
     const expectErrObj = new Error("missing param: email");
     expect(getError()).toEqual(expectErrObj);
   });
   ```

3. Run Jest with `npx jest`:

Okay, notice that [test2] will fail because the Error object returned from the function does not have the expected `message`: _`missing param: email`_. Just modify `getError` to pass the test.

---

Now let's look at an example with Deno's standard assert:

1. Prepare the project:

   ```bash
   $ mkdir deno-error-assert
   $ cd deno-error-assert
   $ deno init
   ```

2. Remove everything from `main.ts` and `main_test.ts`, then add:

   ```ts
   // main.ts
   export function getError() {
     return new Error("missing param: name");
   }
   ```

   ```ts
   // main_test.ts
   import {
     assertEquals,
     equal,
   } from "https://deno.land/std@0.204.0/assert/mod.ts";
   import { getError } from "./main.ts";

   Deno.test(
     '[test1] Should return error object with "missing param: name" message',
     () => {
       const expectErrObj = new Error("missing param: name");
       assertEquals(getError(), expectErrObj);
     },
   );

   Deno.test(
     '[test2] Should return error object with "missing param: email" message',
     () => {
       const expectErrObj = new Error("missing param: email");
       equal(getError(), expectErrObj);
       assertEquals(getError(), expectErrObj);

       const anotherExpectErrObj = new Error("any message");
       equal(getError(), anotherExpectErrObj);
       assertEquals(getError(), anotherExpectErrObj);
     },
   );
   ```

3. Run the test with `deno test`:

[...]

PR #4248, opened 2024-01-29
fix(expect): add Error object equal check.

eryue0220:
This pr is to solved #4221, which is missed from the previous PR.

### DIFF
--- a/expect/_to_equal_test.ts
+++ b/expect/_to_equal_test.ts
@@ -11,6 +11,7 @@ import {
 import { assertThrows } from "../assert/assert_throws.ts";
 import { AssertionError } from "../assert/assertion_error.ts";
 import { expect } from "./expect.ts";
+import { assertFalse } from "https://deno.land/std@$STD_VERSION/assert/mod.ts";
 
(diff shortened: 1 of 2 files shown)


expect/_to_equal_test.ts
```diff
@@ -263,3 +262,16 @@ Deno.test("align to jest test cases", () => {
...
+  function getError() {
+    return new Error("missing param: name");
+  }
+
+  const expectErrObjectWithName = new Error("missing param: name");
+  expect(getError()).toEqual(expectErrObjectWithName);
+
+  const expectErrObjectWithEmail = new Error("missing param: email");
+  expect(getError()).not.toEqual(expectErrObjectWithEmail);
+  assertFalse(expect(getError()).not.toEqual(expectErrObjectWithEmail));
```
kt3k, line 276: This doesn't seem testing anything as `toEqual()` always _returns_ `undefined`

kt3k: Updated

kt3k (approved):
LGTM. Thanks!

Issue #2072, opened 2022-03-30
Path class similar to Rust's PathBuf


ry:
I've come to love Rust's `PathBuf` and `Path` abstractions. Is anyone interested in doing a class in std that works like these?

https://doc.rust-lang.org/stable/std/path/struct.PathBuf.html
https://doc.rust-lang.org/stable/std/path/struct.Path.html

I'm not sure how exactly this should look - but here's some example code of what I have in mind...

Example:
```ts
import { Path } from "https://deno.land/std/path/mod.ts";
import { assert } from "https://deno.land/std/testing/assert.ts";

const p = new Path("/Users/ry/src/deno/README.md");
for (let a in p.ancestors()) {
  assert(a.exists());
}
```

- You should be very familiar with the Rust `PathBuf` and `Path` abstractions if you undertake this.
- This should work on Windows.

nayeemrmn:
Duplicate of #678. We need to consider the fact that Rust's `PathBuf` and `Path` are built-in and how a non-native `Path` type would fragment APIs across the ecosystem which accept paths.

lucacasonato:
I agree with @nayeemrmn here. I am not sure I am in favour of a new `Path` wrapper class. Paths should be represented as strings, and path methods should be functions IMO.

ry:
Closed in favor of https://github.com/denoland/deno_std/issues/678

Issue #678, opened 2020-01-25
Consider adding a path class to the standard library


cjbassi:
Currently there is a path module in the Deno standard library, but it's not architectured in an OOP way and the functions are stringly typed. This is similar to how python used to do path operations with the `os` module, and then they replaced it with the pathlib module which works in a more object-oriented way with methods instead of functions and with a strongly typed path type. Rust also has a path module in the standard library and they ended up basing it on pythons pathlib module. Thanks!

bartlomieju:
I think this is still desirable. If anyone is courageous enough to undertake it, please let me know.

iuioiua:
By the lack of discussion in over a year, it appears as though this feature is not desired much in the community. Perhaps, this issue can be closed.

lino-levan:
I agree with @iuioiua on this one. It doesn't seem like anyone really wants to pick this up.

dsherret:
I wrote a path API for dax here: https://github.com/dsherret/dax#path-api (just released) -- we might consider moving this down to deno_std, but how we would get interop to work well between different versions of deno_std will be annoying with TypeScript. In dax this is ok because most people won't be using it in libraries and are end users writing scripts, but that's not the case for deno_std. Perhaps we will need to ensure the type is always assignable to past versions and that the interface remains quite stable—never do breaking changes on it and be really strict about what we include.

lino-levan:
After some soul searching, I've decided that I'd be pretty for adding a path class. Dax's path class looks great (and quite similar to the rust version). My thinking is that we don't allow path classes as input anywhere since it seems like it'll be a problem for forward and backwards compatibility. Just use `.toString()`!

jollytoad:
please don't turn a simple set of functions into a class, it's completely unnecessary, platforms can easily be separated using a module structure... https://github.com/denoland/deno_std/discussions/3553

iuioiua:
I agree. Simple functions, like we have now, are better than a possible class-based solution. I'm -1 on this. Do you think we can finally close this issue off, @kt3k?

kt3k:
Looks like there's no strong support on this. Let's close

dsherret:
I created a `Path` class that's built on top of `@std/path@1` and `@std/fs@1` (extracted out from dax). It's available here: https://jsr.io/@david/path

Please provide feedback here: https://github.com/dsherret/path (also, please consider contributing docs with examples) -- I'd like to stabilize this to 1.0 soon once I can get feedback.

PR #4250, opened 2024-01-30
feat(path): POC - Path class

dsherret:
This is a proof of concept of moving Dax's `Path` class into deno_std and simplifying the API a bit. Currently no tests, but I can port those over easily because I already have most of them written in Dax.

Once we have `import { Path } from "jsr:@std/path@1"` then this kind of API makes a lot more sense because interop between packages will be easier.

Related: https://github.com/denoland/deno_std/issues/678 https://github.com/denoland/deno_std/issues/2072

### DIFF
--- a/path/mod.ts
+++ b/path/mod.ts
@@ -75,3 +75,4 @@ export * from "./glob_to_regexp.ts";
 export * from "./is_glob.ts";
 export * from "./join_globs.ts";
 export * from "./normalize_glob.ts";
+export * from "./path.ts";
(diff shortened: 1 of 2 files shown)


iuioiua:
Frankly, I'd much rather use the functional approach that `std/path` currently uses. Either way, let's see what others think.

timreichen:
I don't see the point of wrapping these functions up into a class.
Classes are for stateful data imo, but this class has only a single `#path` property which is never altered.

dsherret:
Yeah, I think we can extract this out to a library. Will be easier and more feasible with JSR now

dsherret:
I created a repo and jsr package with this here: https://github.com/dsherret/path

Issue #4249, opened 2024-01-30
`io` module is marked deprecated in API doc


babiabeo:
**Describe the bug**

Visit https://deno.land/std@0.213.0/io/mod.ts

The `io` module is marked deprecated in API doc although there is no `@deprecated` tag in jsdoc in `mod.ts`.

![screenshot-1706619286174](https://github.com/denoland/deno_std/assets/95457148/c831d6b2-8b3b-4fc1-8353-246938c00041)

**Expected behavior**

It should not be marked deprecated as it was re-introduced in #4128.

**Environment**

N/A

iuioiua:
I agree. The `@deprecated` tag should be removed, and its stability status should be updated in the README.

babiabeo:
> I agree. The `@deprecated` tag should be removed, and its stability status should be updated in the README.

I think we should open an issue for updating modules' stability status.

PR #3556, opened 2023-08-21
BREAKING(io): deprecate io top level module

lino-levan:
Closes #3552.

PR #4252, opened 2024-01-31
docs(io): fix `io` module is marked deprecated

babiabeo:
I think this deprecation notice is unnecessary because all APIs in `buf_reader.ts` have already been marked deprecated.

Fix: #4249
Ref: #3556


iuioiua (approved):
LGTM. Thank you.

PR #4254, opened 2024-02-01
refactor(log): refactor log codebase to adhere to singe-export files

eryue0220:
This PR is to solve #4174.


log/mod.ts
```diff
@@ -418,183 +387,10 @@ const state = {
...
-}
-
-setup(DEFAULT_CONFIG);
+export { critical } from "./critical.ts";
+export { debug } from "./debug.ts";
+export { error } from "./error.ts";
+export { getLogger } from "./get_logger.ts";
+export { info } from "./info.ts";
+export { setup } from "./_state.ts";
+export { warn } from "./warn.ts";
```
iuioiua, line 396: Can you please group exports together? There's another group above. Also, please use `export * from ...` in such a way that the exports from `mod.ts` are identical to how they were before.

log/mod.ts
```diff
@@ -418,183 +387,10 @@ const state = {
...
-export function warning<T>(
-  msg: (T extends GenericFunction ? never : T) | (() => T),
-  ...args: unknown[]
-): T | undefined {
-  // Assist TS compiler with pass-through generic type
-  if (msg instanceof Function) {
-    return warn(msg, ...args);
-  }
-  return warn(msg, ...args);
-}
```
iuioiua, line 504: Where did these exports go? They should probably live in `log/warn.ts` until they're deprecated.

log/_state.ts
```diff
@@ -0,0 +1,56 @@
...
+
+import { BaseHandler } from "./base_handler.ts";
+import { DEFAULT_CONFIG, DEFAULT_LEVEL } from "./_config.ts";
+import { type LogConfig, Logger } from "./logger.ts";
+
+export const state = {
+  handlers: new Map<string, BaseHandler>(),
+  loggers: new Map<string, Logger>(),
+  config: DEFAULT_CONFIG,
+};
```
iuioiua, line 11: Can you please move this to a separate `_state.ts` file and rename this current file `setup.ts`?

iuioiua (changes requested):

log/mod.ts
eryue0220: Updated.

log/_state.ts
eryue0220: Updated.

log/mod.ts
eryue0220: I thought it can be removed together. 

Updated.

Issue #3964, opened 2023-12-15
tracking: complete `std/expct`


iuioiua:
The following APIs are currently not yet implemented in `std/expect`:

- [ ] `toMatchSnapShot`
- [ ] `toMatchInlineSnapShot`
- [ ] `toThrowErrorMatchingSnapShot`
- [ ] `toThrowErrorMatchingInlineSnapShot`
- [ ] `expect.anything`
- [ ] `expect.any`
- [ ] `expect.arrayContaining`
- [ ] `expect.not.arrayContaining`
- [ ] `expect.closedTo`
- [ ] `expect.objectContaining`
- [ ] `expect.not.objectContaining`
- [ ] `expect.stringContaining`
- [ ] `expect.not.stringContaining`
- [ ] `expect.stringMatching`
- [ ] `expect.not.stringMatching`
- [ ] `expect.assertions`
- [ ] `expect.hasAssertions`
- [ ] `expect.addEqualityTester`
- [ ] `expect.addSnapshotSerializer`
- [ ] `expect.extend`

For those wanting to contribute, please submit one PR per API.

raashidanwar:
Hey @iuioiua, 
I would love to contribute here.

iuioiua:
> Hey @iuioiua, 
> 
> I would love to contribute here.

Great! Let us know which ones you'd like to work on, so there's no overlap with other's work.

syhol:
As you start working on the async matchers, you might end up fixing https://github.com/denoland/deno_std/issues/3947 as a side effect.

raashidanwar:
@iuioiua I will be picking.
```
expect.objectContaining
expect.not.objectContaining
expect.stringContaining
expect.not.stringContaining
expect.stringMatching
expect.not.stringMatching
```

eryue0220:
Hi @iuioiua @raashidanwar 
Can I try for this too?

I'll pick

```ts
expect.assertions
expect.hasAssertions
expect.addSnapshotSerializer
expect.extend
```

iuioiua:
> Hi @iuioiua @raashidanwar Can I try for this too?
> 
> I'll pick
> 
> ```ts
> expect.assertions
> expect.hasAssertions
> expect.addSnapshotSerializer
> expect.extend
> ```

For sure! Thank you. Please let us know if you need any help with anything.

eryue0220:
Hi, @iuioiua 

I have a question about is there a way that I can know whether the current `Deno.test` is completed or not?

[...]

PR #4255, opened 2024-02-01
feat(expect): add `expect.addEqualityTesters` api.

eryue0220:
This pr is one of prs to solve #3964, which is to support `expect.addEqualityTester` api.


expect/_custom_equality_tester.ts
```diff
@@ -0,0 +1,19 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import { Tester } from "./_types.ts";
+
+const customEqualityTesters: Array<Tester> = [];
+
+export function addCustomEqualityTester(newTesters: Array<Tester>) {
```
babiabeo, line 7: I think we can use `Tester[]` instead of `Array<Tester>`.

expect/expect.ts
```diff
@@ -212,22 +167,4 @@ export function expect(value: unknown, customMessage?: string): Expected {
...
-};
-
-function isPromiseLike(value: unknown): value is PromiseLike<unknown> {
-  if (value == null) {
-    return false;
-  } else {
-    return typeof ((value as Record<string, unknown>).then) === "function";
-  }
-}
+expect.addEqualityTester = addCustomEqualityTester;
```
kt3k, line 170: Looks like this needs to be `addEqualityTesters` (plural `s`) https://jestjs.io/docs/expect#expectaddequalitytesterstesters

expect/_custom_equality_tester_test.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+
+Deno.test("Basic custom equality test", () => {
+  expect(duration1).toEqual(duration1);
+  expect(duration1).not.toEqual(duration3);
+  expect(duration1).not.toEqual(duration2);
+  expect(duration2).not.toBe(duration1);
+  expect(toIterator([duration1, duration2])).toEqual(
+    toIterator([duration2, duration1]),
+  );
```
kt3k, line 69: If I tested this on jest, it threw with the below error:
```
TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them
```
What does this case try to test? Also the inverted order of `duration1` and `duration2` intentional?

expect/expect.ts
eryue0220: Updated.

expect/_custom_equality_tester_test.ts
eryue0220: Oooh, sorry I carelessly ignored this case. 

This case should _**not**_ equal **actually**.

What the test wants to check includes:
* check yield value in array is pass or not (whatever the order);
* check type is pass or not;
* check the instance indeed value is pass or not;

`Jest` implements this through [iterableEquality](https://github.com/jestjs/jest/blob/main/packages/expect-utils/src/utils.ts#L173), and `vitest` reuses this methods. And currently I don't have many ideas about this, so I marked it `todo`.

Or we also can reused the `jest` implementations?

WDYT?

kt3k: Reusing iterableEquality sounds good to me.

BTW this PR looks enough useful at this point. I'm landing this as is.

PRs for the above addition are welcome

eryue0220: PR for this case: https://github.com/denoland/deno_std/pull/4286

PR #4170, opened 2024-01-12
deprecation(log): deprecate `LogLevels.WARNING`

iuioiua:
Related #4117

PTAL @timreichen

PR #4258, opened 2024-02-01
BREAKING(log): remove `WARNING` log level

iuioiua:
Users should use the `WARN` log level and `warn()` instead.

Deprecations were made in #4170 and #4117.
Original proposal in #4099.

CC @timreichen


kt3k (approved):
LGTM. Thanks!

PR #4152, opened 2024-01-10
chore(path): split up glob test into multiple files

lino-levan:
Closes https://github.com/denoland/deno_std/issues/4071.

PR #4259, opened 2024-02-01
BREAKING(path): remove `glob.ts` files

iuioiua:
Users should use the file corresponding to the glob-related function instead. E.g. `path/is_glob.ts` for `isGlob()`.

Note: the deprecation notices for `GlobOptions` and `GlobToRegExpOptions` were missed in the deprecation PR. However, their corresponding functions were. These types are now exported from the files that use them.

Deprecation was made in #4152.
Original proposal in #4071.

CC @lino-levan


lino-levan:
Seems reasonable to me. Thanks for the great work as usual Asher.

kt3k (approved):
LGTM

PR #4260, opened 2024-02-01
BREAKING(path): remove `separator.ts` files

iuioiua:
Users should use the constants provided in the `constants.ts` files within `std/path` instead.

Deprecations were made in #4153.
Original propsal in #4070.

CC @lino-levan


lino-levan:
Same feedback as other PR. LGTM.

kt3k (approved):
LGTM

PR #4262, opened 2024-02-01
BREAKING(semver): remove `SemVerRange` type

iuioiua:
This change removes the deprecated `SemVerRange` type from `std/semver`. This includes the removal of `isSemVerRange()` as well.

Users should migrate to using the [`Range`](https://deno.land/std/semver/types.ts?s=Range) type and [`isRange()`](https://deno.land/std/semver/is_range.ts?s=isRange) instead.

Original proposal: #4059
Deprecation PR: #4161

CC @timreichen


iuioiua:
On second thought, it's too early to be doing this. I'll push instead push the removal version back.

iuioiua:
Closing in favour of #4295.

PR #4264, opened 2024-02-02
BREAKING(semver): remove `sort()`

iuioiua:
This change removes the deprecated `sort()` function from `std/semver`.

To migrate, users should use the in-built `.sort()` method on `SemVer[]`.

Proposal and deprecation in #4098

CC @timreichen


kt3k (approved):
LGTM

PR #4265, opened 2024-02-02
BREAKING(semver): remove `rangeFormat()`

iuioiua:
This change removes the deprecated `rangeFormat()` function.

To migrate, users should use [`formatRange()`](https://deno.land/std/semver/format_range.ts?s=formatRange) instead, which is identical in functionality.

Proposal in https://github.com/denoland/deno_std/issues/3948#issue-2039604547
Deprecation in #4090

CC @timreichen


kt3k (approved):
LGTM. Thanks!

PR #4266, opened 2024-02-02
refactor(semver): cleanup comparison functions

iuioiua:
This change removes the internal use of the deprecated `eq()`, `lt()`, `lte()`, `neq()`, `gt()` and `gte()` functions.

This will make the future removal PR cleaner.


kt3k (approved):
LGTM. Nice clean ups!

PR #4267, opened 2024-02-02
chore(semver): postpone removal versions of some APIs

iuioiua:
This change postpones the removal version of the following APIs from 0.215.0 to 0.216.0: `eq()`, `gt()`, `gte()`, `neq()`, `lt()`, `lte()`, `SemVerRange` and `isSemVerRange()`.

A removal version of 0.215.0 is too early, considering they were deprecated in 0.213.0.

CC @timreichen


kt3k (approved):
Sounds reasonable. How about tagging this commit as `chore` as this actually doesn't deprecate anything new?

PR #4271, opened 2024-02-02
deprecation(semver): deprecate `==`, `===`, `!==`, and `""` operators

timreichen:
- deprecates `"=="`, `"==="`, `"!=="` operators, as they function the same as `"="` and `"!="` respectively. 
- makes `Comparator.operator` optional to replace `""` operator.


kt3k:
> removes "==", "===", "!==" operators, as they function the same as "=" and "!=" respectively.
> makes Comparator.operator optional to replace "" operator.

It looks like the both are breaking changes

iuioiua:
We should consider that stricter `==`, `===` and `!==` may exist because they are familiar to JavaScript devs. That being said, that may not justify having them in the API. I'm impartial.

kt3k:
> Would you agree though that these operators do not make sense and should be removed?

Not sure it's worth the breaking change.

Also not sure we are strongly in favor of removing `''` from operator kind. I guess nobody cares about it.

Also we don't do breaking change without deprecation period.

kt3k (approved):
This probably makes sense. I've never seen semver range expression using these operators. e.g. No `==1.0.0` or `===1.0.0`, it's always `=1.0.0` in documents out there.

LGTM.

iuioiua:
> * makes `Comparator.operator` optional to replace `""` operator.

Neither an optional `operator` nor `operator = ""` make sense to me. If `comparator.operator` is either `undefined` or `""`, it seizes to be a comparator and really is a `SemVer` at that point. Am I missing something?

kt3k:
> If comparator.operator is either undefined or "", it seizes to be a comparator and really is a SemVer at that point. Am I missing something?

I think that's still a valid comparator because the expression like "1.0.0" is a valid range specifier, which matches the exact version "1.0.0"

iuioiua:
A range that lacks an operator is surprising behaviour to me. It's also an exception to the rule: a comparator is a semver prefixed with an operator. Either way, it's a discussion for a separate issue. I won't block this PR.

kt3k:
A range without operator is just fine in npm. See https://docs.npmjs.com/cli/v10/configuring-npm/package-json#dependencies

I think we shouldn't change that convention here

semver/_constants.ts
```diff
@@ -1,5 +1,8 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
+/**
+ * @deprecated (will be removed in 0.217.0) `"=="`, `"==="`, `"!=="` and `""` operators are deprecated. Use `"="`, `"!="` or `undefined` instead.
```
kt3k, line 4:
```suggestion
 * @deprecated (will be removed in 0.219.0) `"=="`, `"==="`, `"!=="` and `""` operators are deprecated. Use `"="`, `"!="` or `undefined` instead.
```

semver/types.ts
```diff
@@ -18,6 +18,7 @@ export type ReleaseType =
 
 /**
  * SemVer comparison operators.
+ * @deprecated (will be removed in 0.217.0) `"=="`, `"==="`, `"!=="` and `""` operators are deprecated. Use `"="`, `"!="` or `undefined` instead.
```
kt3k, line 21:
```suggestion
 * @deprecated (will be removed in 0.219.0) `"=="`, `"==="`, `"!=="` and `""` operators are deprecated. Use `"="`, `"!="` or `undefined` instead.
```

PR #4276, opened 2024-02-04
refactor(fmt): prepare for `noUncheckedIndexedAccess`

eryue0220:
Ref: #4040

### DIFF
--- a/fmt/printf.ts
+++ b/fmt/printf.ts
@@ -256,7 +256,7 @@ class Printf {
     this.flags = new Flags();
     const flags = this.flags;
     for (; this.i < this.format.length; ++this.i) {
-      const c = this.format[this.i];
+      const c = this.format[this.i] as string;
       switch (this.state) {
         case State.PERCENT:
           switch (c) {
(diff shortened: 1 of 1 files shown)


iuioiua (commented):
Please see https://github.com/denoland/deno_std/pull/4275#pullrequestreview-1861556153

eryue0220:
@iuioiua 

Sorry for the late response, this pr is ready.

fmt/printf.ts
```diff
@@ -453,7 +453,7 @@ class Printf {
   /** Handle verb */
   handleVerb() {
     const verb = this.format[this.i];
-    this.verb = verb;
+    this.verb = verb || this.verb;
```
iuioiua, line 456: Can you please explain this change?

eryue0220: Because the  `verb` maybe undefined, So I think this can reusing the `this.verb`, which is initialized to the empty string.

iuioiua (approved):
LGTM! Thank you.

PR #4277, opened 2024-02-04
refactor(fs): prepare for `noUncheckedIndexedAccess`

eryue0220:
Ref: #4040


iuioiua (commented):
Nice use of `as const`! Can you please provide comments regarding the justifications of the remaining changes?

Related: https://github.com/denoland/deno_std/pull/4275#pullrequestreview-1861556153

fs/testdata/empty_dir.ts
```diff
@@ -1,6 +1,6 @@
 import { emptyDir } from "../empty_dir.ts";
 
-emptyDir(Deno.args[0])
```
eryue0220, line 3: As for this change, there is an indeed case that it may `undefined` for the `Deno.args[0]`, but the `empty` function has wrapped in `try/catch` block. On the other hand, it's hard to say exactly what type is it on the runtime, it may `string` or `URL`. Thus I think it's ok to use `!` operator.

fs/testdata/empty_dir_sync.ts
```diff
@@ -1,7 +1,7 @@
 import { emptyDirSync } from "../empty_dir.ts";
 
 try {
-  emptyDirSync(Deno.args[0]);
+  emptyDirSync(Deno.args[0]!);
```
eryue0220, line 4: Similar to the [previous comment](https://github.com/denoland/deno_std/pull/4277/files#r1477861640).

eryue0220:
Hi,

@iuioiua 

This PR is ready to review, and the other two I'll upgrade ASAP.

fs/expand_glob.ts
```diff
@@ -253,7 +253,7 @@ export function* expandGlobSync(
   let fixedRoot = isGlobAbsolute
     ? winRoot !== undefined ? winRoot : "/"
     : absRoot;
-  while (segments.length > 0 && !isGlob(segments[0])) {
+  while (segments.length > 0 && !isGlob(segments[0]!)) {
```
iuioiua, line 256: Note: this non-null assertion is justified because we know that `segments.length > 0`.

iuioiua (approved):
LGTM! Thank you.

fs/expand_glob.ts
iuioiua: Also, a non-null assertion is fine here because we already know `segments: string[]`.

PR #4278, opened 2024-02-04
BREAKING(semver): remove `Comparator` functions

iuioiua:
This change removes `Comparator` functions, as they don't appear to have much use in the public API.

Proposal in #4047
Deprecation in #4131

CC @timreichen


kt3k (approved):
LGTM. Thanks!

PR #4280, opened 2024-02-05
deprecation(semver): deprecate `reverseSort()`

timreichen:
Deprecates `reverseSort()` function

This function seems like a really special user case and trivial by using `compare()`.

ref: https://github.com/denoland/deno_std/issues/3948

### DIFF
--- a/semver/reverse_sort.ts
+++ b/semver/reverse_sort.ts
@@ -1,10 +1,20 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 import type { SemVer } from "./types.ts";
 import { compare } from "./compare.ts";
+import { warnOnDeprecatedApi } from "https://deno.land/std@$STD_VERSION/internal/warn_on_deprecated_api.ts";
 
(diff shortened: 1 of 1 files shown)


semver/reverse_sort.ts
```diff
@@ -1,10 +1,20 @@
...
+ */
 export function reverseSort(
   versions: SemVer[],
 ): SemVer[] {
+  warnOnDeprecatedApi({
+    apiName: "reverseSort",
+    stack: new Error().stack!,
+    removalVersion: "0.217.0",
+    suggestion: "Use `versions.sort((a, b) => compare(b, a))` instead.",
+  });
```
iuioiua, line 18: Can you please remove the use of `warnOnDeprecatedApi()`? We've stopped using it for now (yes, it was short-lived 😆). Note to self: remove or disable the function when I can.

timreichen: oh🤔 Done.

iuioiua (approved):
LGTM

PR #4283, opened 2024-02-07
docs(async): use `Array.fromAsync()` in `debounce()` example

### LINKED ISSUE #4282: suggestion(async): Update `debounce` JSDoc example with `Array.fromAsync`
There are current JSDoc comments.

https://github.com/denoland/deno_std/blob/52138c6f99fb3cf59d690722a07eb140c2203acc/async/debounce.ts#L27-L37

BTW, `Array.fromAsync` implemented [Deno 1.38.0 ](https://deno.com/blog/v1.38#v8-120).

So, it could be able to replace without type annotation like below. 

```typescript
Array.fromAsync(
  Deno.watchFs('./'),
  debounce((event) => {
    console.log('[%s] %s', event.kind, event.paths[0]);
  }, 200),
);
```

Hajime-san:
resolves #4282

### DIFF
--- a/async/debounce.ts
+++ b/async/debounce.ts
@@ -26,15 +26,12 @@ export interface DebouncedFunction<T extends Array<unknown>> {
  * ```
  * import { debounce } from "https://deno.land/std@$STD_VERSION/async/debounce.ts";
  *
- * const log = debounce(
- *   (event: Deno.FsEvent) =>
- *     console.log("[%s] %s", event.kind, event.paths[0]),
- *   200,
+ * Array.fromAsync(
+ *   Deno.watchFs('./'),
+ *   debounce((event) => {
+ *     console.log('[%s] %s', event.kind, event.paths[0]);
+ *   }, 200),
  * );
- *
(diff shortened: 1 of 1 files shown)


async/debounce.ts
```diff
@@ -26,15 +26,12 @@ export interface DebouncedFunction<T extends Array<unknown>> {
  * ```
  * import { debounce } from "https://deno.land/std@$STD_VERSION/async/debounce.ts";
  *
- * const log = debounce(
- *   (event: Deno.FsEvent) =>
- *     console.log("[%s] %s", event.kind, event.paths[0]),
- *   200,
+ * Array.fromAsync(
```
iuioiua, line 29:
```suggestion
 * await Array.fromAsync(
```

iuioiua (approved):
LGTM. Thank you!

kt3k:
I feel this is rather confusing as we don't use the result of `Array.fromAsync()`

iuioiua:
Perhaps. Though, we don't assign the result to a variable... Happy to revert if you prefer.

Issue #4284, opened 2024-02-07
Allow passing arguments to std/datetime/format()


paul-greenweb:
**Is your feature request related to a problem? Please describe.**

The problem is, it seems that the current implementation of datetime/format() doesn't allow to return a date in UTC. It is an important feature for logging purposes.

**Describe the solution you'd like**

I would like to be able to pass an options object to the formatter.format() function 

```
export function format(date: Date, formatString: string, **options**): string {
  const formatter = new DateTimeFormatter(formatString);
  return formatter.format(date, **options**);
}
```

Code : https://deno.land/std@0.214.0/datetime/format.ts?source=#L27

I see that the options.timeZone is already implemented in formatter.format()

```
format(date: Date, options: Options = {}): string {
    let string = "";

    const utc = options.timeZone === "UTC";
```

Code : https://deno.land/std@0.214.0/datetime/_common.ts?source#L259

**Describe alternatives you've considered**

I didn't find an easy solution

PR #4285, opened 2024-02-07
feat(datetime): `format()` options

babiabeo:
Fix: #4284

### DIFF
--- a/datetime/format.ts
+++ b/datetime/format.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { DateTimeFormatter, Options } from "./_common.ts";
+import { DateTimeFormatter } from "./_common.ts";
 
 /**
  * Takes an input `date` and a `formatString` to format to a `string`.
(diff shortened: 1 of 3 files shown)


iuioiua (changes requested):
Hey David, could we add the following to `datetime/format.ts` instead?
```ts
export interface FormatOptions {
  utc?: boolean;
}
```
We should also have accompanying documentation. WDYT?

datetime/format.ts
```diff
@@ -18,13 +18,29 @@ import { DateTimeFormatter } from "./_common.ts";
...
+): string {
   const formatter = new DateTimeFormatter(formatString);
-  return formatter.format(date);
+  return formatter.format(
+    date,
+    options.utc ? { timeZone: "UTC" } : undefined,
+  );
+}
+
+/** The options used for formatting date in {@linkcode format}. */
```
iuioiua, line 42:
```suggestion
/** Options for {@linkcode format}. */
```

iuioiua (approved):
LGTM! Thanks! I'll let Yoshiya review too.

kt3k (approved):
This addition makes sense to me. LGTM

PR #4286, opened 2024-02-08
fix(expect): fix custom equality test case

eryue0220:
This pr is to support iterable object test check and fix expect custom equality test case.

--------------------------
Background: https://github.com/denoland/deno_std/pull/4255#discussion_r1479303914


kt3k:
@eryue0220 
Thanks for the PR, but this seems including too much code for fixing that specific test case. Can you try minimize the code introduced?

eryue0220:
> Thanks for the PR, but this seems including too much code for fixing that specific test case. Can you try minimize the code introduced?

Ok, I'll try for this.

eryue0220:
@kt3k 

Sorry for the late response,

I've check the api and its usage again, and I restructure the codebase. But the codebase is not just for the specific case, it's also a public api in jest, which is exported from [jest/expect-utils](https://github.com/jestjs/jest/blob/main/packages/expect-utils/src/index.ts). 

And other expect apis, such as `arrayContaining` and `stringContaining` will depend on this method.

expect/_utils.ts
```diff
@@ -18,3 +18,82 @@ export function isPromiseLike(value: unknown): value is PromiseLike<unknown> {
     return typeof ((value as Record<string, unknown>).then) === "function";
   }
 }
+
+// SENTINEL constants are from https://github.com/facebook/immutable-js
+const IS_KEYED_SENTINEL = "@@__IMMUTABLE_KEYED__@@";
+const IS_SET_SENTINEL = "@@__IMMUTABLE_SET__@@";
+const IS_LIST_SENTINEL = "@@__IMMUTABLE_LIST__@@";
+const IS_ORDERED_SENTINEL = "@@__IMMUTABLE_ORDERED__@@";
+const IS_RECORD_SYMBOL = "@@__IMMUTABLE_RECORD__@@";
```
kt3k, line 27: Checks for these sentinel keys/values don't make much sense to me because nothing seems setting these keys.

I guess we can just remove these constants and checks related to these

eryue0220: For internal codebase, it is yes. But as for the apps which uses `immutable.js`, they will rely on these keys/value to test it, such as the following:

```ts
import { OrderedMap } from "npm:immutable@4.3.5";
import { expect, iterableEquality } from "https://deno.land/std@$STD_VERSION/expect/mod.ts";

Deno.test("immutable test", () => {
    const a = OrderedMap().set("saving", true);
    const b = OrderedMap().merge({saving: true});
    expect(iterableEquality(a, b)).toBe(true);
});
```

kt3k: Does jest support such cases natively?

eryue0220: Yes, the `iterableEquality` is exported from `expect-utils`, and the test codebase is [here](https://github.com/jestjs/jest/blob/main/packages/expect-utils/src/__tests__/utils.test.ts#L579)

expect/mod.ts
```diff
@@ -85,3 +85,4 @@
  */
 export { expect } from "./expect.ts";
 export { fn } from "./fn.ts";
+export { iterableEquality } from "./iterable_equality.ts";
```
kt3k, line 88: I don't think it's good idea to expose iterableEquality from here.

`iterableEquality` is exported from `@jest/expect-utils`, but the README of `@jest/expect-utils` seems discouraging importing things directly from it. ref. https://www.npmjs.com/package/@jest/expect-utils

I'll make `iterableEquality` private API in `std/expect`

eryue0220: I agree in some degree, because it doesn't indeed belong to `expect` module. But if it's private, how to use in public?

kt3k: I suggest we shouldn't provide direct import of `iterableEquality`. The purpose of `std/expect` is compatiblity to jest's expect, but jest doesn't document `iterableEquality` in their docs. Supporting `iterableEquality` is out of scope here in my view.

kt3k (approved):

kt3k:
I'm landing this PR at this point.

Let's discuss in another issue if you have further questions about the scope of `std/expect`

PR #4291, opened 2024-02-08
chore: fix publish of `std/crypto` to JSR

iuioiua:


kt3k (approved):
LGTM

kt3k:
BTW `crypto/_wasm/target` is specified in `.gitignore` at the top level. It might make sense to ignore such files automatically

PR #4293, opened 2024-02-08
BREAKING(semver): remove `outside()`

iuioiua:
This change removes `outside()` from `std/semver`.

To migrate, users should use [`greaterThan()`](https://deno.land/std/semver/greater_than.ts?s=greaterThan), [`lessThan()`](https://deno.land/std/semver/less_than.ts?s=lessThan) and [`testRange()`](https://deno.land/std/semver/test_range.ts?s=testRange) instead.

Deprecation happened in #4185


kt3k (approved):
LGTM. Thanks

PR #4294, opened 2024-02-09
BREAKING(semver): remove `eq()`, `gt()`, `gte()`, `lt()`, `lte()` and `neq()`

iuioiua:
This change deprecates `eq()`, `gt()`, `gte()`, `lt()`, `lte()` and `neq()`.

Deprecation happened in #4083


kt3k (approved):
LGTM

PR #4297, opened 2024-02-09
refactor(html): prepare for `noUncheckedIndexedAccess`

gabelluardo:
Ref: #4040

### DIFF
--- a/html/entities.ts
+++ b/html/entities.ts
@@ -93,7 +93,7 @@ export function unescape(
   }
 
   return str
-    .replaceAll(entityRe, (m) => entityList[m])
+    .replaceAll(entityRe, (m) => entityList[m]!)
     .replaceAll(RX_DEC_ENTITY, (_, dec) => codePointStrToChar(dec, 10))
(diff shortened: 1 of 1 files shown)


html/entities.ts
```diff
@@ -93,7 +93,7 @@ export function unescape(
   }
 
   return str
-    .replaceAll(entityRe, (m) => entityList[m])
+    .replaceAll(entityRe, (m) => entityList[m]!)
```
iuioiua, line 96: Can you please justify why `entityList[m]` is non-null?

iuioiua (changes requested):

html/entities.ts
gabelluardo: `entityList` is filled with the `defaultEntityList` and the values in `entityRe` are generated from the keys of `entityList`, so I think `entityList[m]` with `m` an element of `entityRe` is always non-null

iuioiua (approved):
LGTM. Thank you.

PR #4301, opened 2024-02-10
feat(uuid): add type predicate return type to validate (v4)

nnmrts:
I added this so that one could use `validate` to actually narrow down a value for further usage at places where a v4 UUID is expected, like for example:

```ts
import { validate } from "std/uuid/v4";

const getUserByName = (name: string) => { /* ... */ };

const getUserById1 = (id: ReturnType<typeof crypto.randomUUID>) => { /* ... */ };

const getUserById2 = (id: `${string}-${string}-${string}-${string}-${string}`) => { /* ... */ };

// stupid unrealistic function but I hope you get the point
const getUser = (idOrName: string) => {
	if (validate(idOrName)) {
		// Without the change in this PR, TS/Deno will complain here with
		// "Argument of type 'string' is not assignable to parameter of type
		// '`${string}-${string}-${string}-${string}-${string}`'",
		// because the `validate` check above doesn't actually narrow it down for TS
		return [getUserById1(idOrName), getUserById2(idOrName)];
	}
	
	return getUserByName(idOrName);
};

```

I'm not actually writing in TypeScript often so I'm already sorry if I did something wrong here.

I know this is still not "exact", but the best we can do without regex types, I guess.

### DIFF
--- a/uuid/v4.ts
+++ b/uuid/v4.ts
@@ -17,6 +17,8 @@ const UUID_RE =
  * validate("this-is-not-a-uuid"); // false
  * ```
  */
-export function validate(id: string): id is ReturnType<typeof crypto.randomUUID> {
+export function validate(
+  id: string,
+): id is ReturnType<typeof crypto.randomUUID> {
   return UUID_RE.test(id);
 }
--- a/uuid/v4.ts
+++ b/uuid/v4.ts
@@ -17,6 +17,6 @@ const UUID_RE =
  * validate("this-is-not-a-uuid"); // false
  * ```
  */
-export function validate(id: string): boolean {
+export function validate(id: string): id is ReturnType<typeof crypto.randomUUID> {
   return UUID_RE.test(id);
 }


iuioiua:
Hi Nano, I'll be closing this PR without merging, as the change doesn't seem to benefit the function. `ReturnType<typeof crypto.randomUUID>` evaluates to `string` (see [here](https://github.com/denoland/deno/blob/26d9b2f3174030138a13bdb60c6c76a38fa0df19/ext/crypto/lib.deno_crypto.d.ts#L393)). So the type signature you propose simplifies to:
```ts
export function validate(
  id: string,
): id is string {
  return UUID_RE.test(id);
}
```
In other words, this gives the function a type predicate signature when the type is already defined in the `id: string` parameter.

Either way, we appreciate your contribution 🙂

nnmrts:
@iuioiua Sorry for the oversight, but when someone has their libs defined as `["dom", "dom.iterable", "dom.asynciterable", "deno.ns"]`, which is the recommended config for targeting Deno and the browser (a very common use-case I assume), according to the [Deno docs](https://docs.deno.com/runtime/manual/advanced/typescript/configuration#targeting-deno-and-the-browser), `ReturnType<typeof crypto.randomUUID>` does resolve to this:

```ts
`${string}-${string}-${string}-${string}-${string}`
```

Couldn't we/you leave this PR open and I can work on "fixing" `lib.deno_crypto.d.ts`? Because as far as I know, the return type of `crypto.randomUUID` should be defined there as `${string}-${string}-${string}-${string}-${string}` as well.

Personally I think this PR could even be merged if there are no other issues with it, no? It seems to cause no issues even if it **currently** simplifies down to just `string`.

Please let me know if one of those options would be okay for you or if it's better if I ask this in an issue at the main deno repo.

kt3k:
> Couldn't we/you leave this PR open and I can work on "fixing" lib.deno_crypto.d.ts? Because as far as I know, the return type of crypto.randomUUID should be defined there as ${string}-${string}-${string}-${string}-${string} as well.

Sounds a good point. `lib.deno_crypto.d.ts` should be aligned to `lib.dom.d.ts` in my view.

babiabeo:
An issue has been opened in deno repo. 👍

https://github.com/denoland/deno/issues/22393

nnmrts:
@babiabeo Thanks, I created a PR for that issue here: https://github.com/denoland/deno/pull/22465

kt3k (approved):
LGTM

PR #4304, opened 2024-02-11
feat(http): add headers and methods to the http namespace

halvardssm:
Hi! 

I am submitting this PR as I see the usefulness of HTTP methods and headers to be available in this typed and shorthand matter. The headers and methods are crawled from MDN, and thus also linked. I am not sure how to deal with the copyright, so feedback is appreciated in case I did not do this correctly.

Please let me know if anything needs to be changed.

### DIFF
--- a/http/header.ts
+++ b/http/header.ts
@@ -1,0 +1,782 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+// This module is generated from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers.
+
+/**
+ * Authentication
+ *
+ * @link https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#authentication
+ */
+export const HttpHeaderAuthentication = {
+  /**
+   * Defines the authentication method that should be used to access a resource.
+   *
+   * @link https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate
+   */
+  WwwAuthenticate: "WWW-Authenticate",
+
+  /**
+   * Contains the credentials to authenticate a user-agent with a server.
+   *
+   * @link https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization
+   */
+  Authorization: "Authorization",
+
+  /**
+   * Defines the authentication method that should be used to access a resource behind a proxy server.
+   *
+   * @link https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Proxy-Authenticate
+   */
+  ProxyAuthenticate: "Proxy-Authenticate",
+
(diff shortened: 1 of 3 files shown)


kt3k:
We are probably not accepting `method.ts` as we recently removed a very similar module from `std/http` https://github.com/denoland/deno_std/pull/3951. The main reason for it was that we couldn't reach the consensus on what should be included in it.

I'm not sure about `header.ts`. Does this list include all headers standardized in RFC?

(Feedbacks on this PR from the community are welcome.)

halvardssm:
For the methods, I followed the [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) list, but after researching a bit more, it seems that the standard methods are maintained at [IANA](https://www.iana.org/assignments/http-methods/http-methods.xhtml#methods). Is it this list that did not find a consensus? It seems to me that there are two categories here, webdav and standard HTTP. I think it would at least be beneficial to have the standard headers present, but I wouldn't see an issue with having all of them either (we can just split them up).

For headers, I would probably change it to reflect this list at [IANA](https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names), but separate it by status.

kt3k:
> For the methods, I followed the [RFC 9110](https://www.rfc-editor.org/rfc/rfc9110.html) list, but after researching a bit more, it seems that the standard methods are maintained at [IANA](https://www.iana.org/assignments/http-methods/http-methods.xhtml#methods). Is it this list that did not find a consensus?

Some preferred to list only RFC 9110 ones. Some preferred RFC 9110 + IANA methods. Some preferred what Node.js lists in `require("http").METHODS` (which is different from IANA list). Some argued any string is valid http method (which is apparently what browsers doing)... So we gave up to define 'standard' set of HTTP methods

> For headers, I would probably change it to reflect this list at [IANA](https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names), but separate it by status.

Maybe a good idea, but can you split it to another PR? I think header.ts and method.ts require independent discussions.

halvardssm:
> Some preferred to list only RFC 9110 ones. Some preferred RFC 9110 + IANA methods. Some preferred what Node.js lists in require("http").METHODS (which is different from IANA list). Some argued any string is valid http method (which is apparently what browsers doing)... So we gave up to define 'standard' set of HTTP methods

Wouldn't this be solved by having two groups? One for RFC 9110 and one for IANA, and then have one export that combines the two (alternatively also have a node headers export for compatibility)? It somehow feels like removing them altogether is not the optimal solution.

> Maybe a good idea, but can you split it to another PR? I think header.ts and method.ts require independent discussions.

Sure, Ill submit another PR 💪🏻

kt3k:
Thanks for splitting PRs!

Closing this one in favor of #4317 and #4320

Issue #3966, opened 2023-12-15
investigate(testing): explicit resource management for `spy()` and `stub()`


iuioiua:
Currently, one must execute `.restore()` on a spied (?) or stubbed object to restore the original function/instance to its initial behaviour. It'd be great if `spy()` and `stub()` returned instances implementing the `[Symbol.dispose]` method that would automatically perform this restoring method once out of scope by using the `using` keyword. This might involve implementing new underlying classes and reworking the functions, moving them away from using `Object.defineProperties`.

The same goes for any other applicable APIs in the Standard Library.

Inspired by #3961

Further reading:
* https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-2.html#using-declarations-and-explicit-resource-management
* https://deno.com/blog/v1.38#using-with-deno-apis

syhol:
SWC make's this hard to do without breaking current functionality.

The spy in its current form needs to be a function, so you can use it like this:
```ts
import { spy } from "https://deno.land/std/testing/mock.ts";
const mySpy = spy(Date, "now");
mySpy();
```

SWC's `using` seems to have [this check](https://github.com/swc-project/swc/blob/d6cd2146faed9155ae6f2c625daa173ddc7c0064/packages/helpers/esm/_using.js#L5-L7):
```ts
if (typeof value !== "object") {
  throw new TypeError("using declarations can only be used with objects, null, or undefined.");
}
```

I don't know any way pass the `typeof value !== "object"` check and also support calling the returned spy.

Would you be open to a breaking change and having method spies being objects instead of functions?

Someone could ask SWC if this is something they could change. If not then I think it's something the deno typechecker should pick up on, because I got a runtime error rather than a static one.

iuioiua:
`spy()` can still be a function and return a disposable class instance from under the hood, no? I'm not sure I'm following.

syhol:
I'm just talking about the return value of `spy()`, thats `mySpy` in the example above. `spy` itself will always be a function.

You could make that return value be a disposable class instance, but then it couldn't _also_ be callable like a function, as it is in the example above.

It's important that the return value of spy is a function for the case of the FunctionSpy, when you would wrap a plain function rather than a object method. For example:

```ts
// FunctionSpy
const decodeSpy = spy(decodeURI);
// This is the only way to call the spy
decodeSpy();
```

whereas when you spy on a method (aka MethodSpy)

```ts
// MethodSpy
const randomSpy = spy(Math, "random");
[...]

PR #4306, opened 2024-02-11
feat(testing): explicit resource management for `spy()` and `stub()`

KyleJune:
Closes https://github.com/denoland/deno_std/issues/3966

This is a work in progress. It won't work until [swc is upgraded to v1.4.0](https://github.com/denoland/deno_ast/issues/210) because earlier versions do not support the `using` keyword on functions. Until that issue is resolved, the new test cases I added fail with the following errors.

image

Once that issue is resolved, I believe the new test cases in this PR will pass and we could go forward with shipping it.

Regarding documentation, I added a description to both the spy and stub sections to explain that they are disposable. I noticed that FakeTime was already made disposible and it's example uses the `using` keyword. I thought it would be better to explicitly explain that since `using` is a new feature to TypeScript. I think if all the examples were updated to just use the `using` keyword to automate cleanup, people could easily miss the importance of that keyword in ensuring that the tests are cleaned up properly. Feel free to change my descriptions though.


testing/time.ts
```diff
@@ -211,7 +211,7 @@ let dueTree: RedBlackTree<DueNode>;
  * });
  * ```
  */
-export class FakeTime {
+export class FakeTime implements Disposable {
```
KyleJune, line 214: This class was implementing Disposable already but didn't explicitly say that it was. I updated it to show that it implements Disposable to make it more clear that it is usable.

iuioiua: Good catch! Though, let's do this in a different PR. We could probably do PR that ensures that `[Async]Disposable` is implemented across the codebase where necessary.

iuioiua:
Note to self: run CI when the required functionality is available in canary. Tests should pass in canary but fail in stable.

iuioiua:
> Note to self: run CI when the required functionality is available in canary. Tests should pass in canary but fail in stable.

I didn't notice, but this is already the case! 😆 We've already upgraded to `swc@1.4.0`. I'll review this in time for the next release of `deno_std`.

iuioiua:
@KyleJune, would you also like to update any applicable spies and stubs across the codebase to use the `using` keyword, too?

babiabeo:
I think we should also add a note about the minimum Deno version required for it to work. 👍

KyleJune:
> @KyleJune, would you also like to update any applicable spies and stubs across the codebase to use the `using` keyword, too?

Done, the only ones outside of the context of testing `testing/mock_test.ts` and `testing/time_test.ts` were `testing/bdd_test.ts` and `http/file_server_test.ts`. I didn't update the mock and time tests since the restore abilities are part of what is being tested.

> I think we should also add a note about the minimum Deno version required for it to work. 👍

I don't think that is necessary and that wasn't done when `FakeTime` was updated to have explicit resource management. I don't know if the next version will be v1.40.5 or v1.41.0 so I'm not going to update it to say which version is required. If the core members want to update it to specify the minimum version required for it, they can make the change.

kt3k (approved):
LGTM. Looks a nice feature!

iuioiua (approved):
This is awesome. Thank you! LGTM.

PR #4314, opened 2024-02-12
fix(semver): return new object instance  in `comparatorMin()` and `comparatorMax()`

timreichen:
This fixes cases where `semver` might not be just a `SemVer` and have additional properties:

Affected private APIs
- `comparatorMin()`
- `comparatorMax()`
```ts
// bug
const comparator = { major: 1, minor: 2, patch: 3, operator: "=" }
const semver = comparatorMin(comparator) // semver === comparator and has an "operator" property.
```

Affected public APIs:
- `rangeMin()`
- `rangeMax()`
```ts
import { parseRange, rangeMin } from "https://deno.land/std@$STD_VERSION/semver/mod.ts";
console.log(rangeMin(parseRange(">=1.0.0"))); // { operator: ">=", major: 1, minor: 0, patch: 0, prerelease: [], build: [] }
```


kt3k:
These 2 APIs are now private utils. Can you create a similar example only using public APIs? (I think the commit message should mention affected public APIs)

timreichen:
> These 2 APIs are now private utils. Can you create a similar example only using public APIs? (I think the commit message should mention affected public APIs)

```ts
import { parseRange, rangeMin } from "https://deno.land/std@$STD_VERSION/semver/mod.ts";
console.log(rangeMin(parseRange(">=1.0.0"))); // { operator: ">=", major: 1, minor: 0, patch: 0, prerelease: [], build: [] }
```

kt3k (approved):
Thanks!

LGTM

PR #4316, opened 2024-02-12
refactor(semver): simplify args for `comparatorMin()` and `comparatorMax()`

timreichen:
This refactors `comparatorMin()` and `comparatorMax()` args so they take a `Comparator` instead of two separate args.


iuioiua (approved):
Thanks for this. LGTM.

PR #4317, opened 2024-02-12
feat(http/unstable): `headers` module

halvardssm:
Adds IANA Headers as a follow-up from #4304


http/header.ts
```diff
@@ -0,0 +1,1685 @@
...
+ * @see {@link https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names | IANA Hypertext Transfer Protocol (HTTP) Field Name Registry}
+ */
+export const HttpHeader = {
+  ...HttpHeaderPermanent,
+  ...HttpHeaderDeprecated,
+  ...HttpHeaderObsoleted,
+  ...HttpHeaderProvisional,
+};
+
+export type HttpHeader = typeof HttpHeader[keyof typeof HttpHeader];
```
iuioiua, line 1685:
```suggestion
export type Header = typeof HEADER[keyof typeof HEADER];
```

http/header.ts
```diff
@@ -0,0 +1,1685 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+// This module is generated from {@link https://www.iana.org/assignments/http-fields/http-fields.xhtml#field-names | IANA Hypertext Transfer Protocol (HTTP) Field Name Registry}
+
+/**
+ * HTTP Headers with status permanent
+ */
+export const HttpHeaderPermanent = {
```
iuioiua, line 8:
```suggestion
export const HEADER = {
```

iuioiua (changes requested):
Yoshiya and I would now like to have this addition.

Can you please remove `HttpHeaderDeprecated`, `HttpHeaderObsoleted` and `HttpHeaderProvisional` objects? What's now `HttpHeaderPermanent` seems to be the only object needed for now.

Can you please also add simple/minimal tests to ensure the plumbing works?

halvardssm:
@iuioiua Done! Let me know if there is anything else!

iuioiua (approved):
LGTM! Thank you for this.

PR #4323, opened 2024-02-13
fix(semver): fix `prerelease` handlings in range utils

### LINKED ISSUE #4300: semver: `rangeIntersects` ignores the intersection of prerelease versions
I'm trying to use the semver range utilities, and I the default handling of pre-releases in ranges different from how npm would do it. I _think_ I need to use the `includePrerelease` flag (documented in the "Prerelease Tags" section of https://deno.land/std@0.215.0/semver/mod.ts) to get the same behavior as npm, but I cannot figure out where to pass it given that the [testRange](https://deno.land/std@0.215.0/semver/mod.ts?s=testRange) function does not accept an option bag, and the [Range](https://deno.land/std@0.215.0/semver/mod.ts?s=Range) type does not include that property.

kt3k:
closes #4300

This fixes the case when the given ranges intersect at prerelease versions.

ex. `rangeIntersects(parseRange("<7.0.0-beta.20"), parseRange(">7.0.0-beta.0"))`

Current implementation tries to get min/max of comparators and compares them, but `comparatorMax` doesn't work with prerelease versions since it was first introduced in #3385 (It completely ignores prerelease versions)

`comparatorMax` seems wrong by design because the max of `<7.0.0-beta.20` is something like `7.0.0-beta.19.zzzz....`(infinite 'z's), and that is very inconvenient to represent in javascript.

This PR gives up the current approach for calculating the intersection. Instead this restores the original algorithm used for comparing comparators (see https://github.com/denoland/deno_std/blob/e1698647677d2e58348f3308ddeebb9c40cc0ca9/semver/mod.ts#L821-L874)


kt3k:
`comparatorMax` is inherently wrong. We should remove all usages of it. We also should remove `rangeMax` which doesn't (can't) return correct values.

semver/_comparator_intersects.ts
```diff
@@ -1,54 +1,61 @@
...
+
 /**
- * Returns true if the range of possible versions intersects with the other comparators set of possible versions
- * @param c0 The left side comparator
- * @param c1 The right side comparator
- * @returns True if any part of the comparators intersect
+ * Returns true if the range of possible versions intersects with the other c1arators set of possible versions
+ * @param c0 The left side c1arator
+ * @param c1 The right side c1arator
+ * @returns True if any part of the c1arators intersect
```
iuioiua, line 18:
```suggestion
 * Returns true if the range of possible versions intersects with the other comparators set of possible versions
 * @param c0 The left side comparators
 * @param c1 The right side comparators
 * @returns True if any part of the comparators intersect
```
I'm not sure why `comparators` was changed to `c1arators`. Is this deliberate?

semver/range_intersects_test.ts
```diff
@@ -69,6 +69,7 @@ Deno.test({
       ["1.x", "1.3.0 || <1.0.0 >2.0.0", true],
       ["*", "*", true],
       ["x", "", true],
+      ["<7.0.0-beta.20", ">7.0.0-beta.0", true],
```
iuioiua, line 72: Can you please add one or two more tests?

iuioiua (changes requested):

iuioiua:
> `comparatorMax` is inherently wrong. We should remove all usages of it. We also should remove `rangeMax` which doesn't (can't) return correct values.

In this PR, `comparatorIntersects()` now depends on `testRange()`, which depends on `comparatorMin()` and `comparatorMax()`. How should that then be reconciled?

kt3k:
> In this PR, comparatorIntersects() now depends on testRange(), which depends on comparatorMin() and comparatorMax(). How should that then be reconciled?

Thanks for pointing it. `testRange` also has a bug like `testRange(parse("1.0.0-3"), parseRange("<1.0.0-5")) === false` (should be true). We have to remove all such usages of `comparatorMax`

semver/range_test.ts
```diff
@@ -13,7 +13,6 @@ Deno.test("range", async (t) => {
     ["1.2.3-pre+asdf - 2.4.3-pre+asdf", "1.2.3"],
     ["1.2.3-pre+asdf - 2.4.3-pre+asdf", "1.2.3-pre.2"],
     ["1.2.3-pre+asdf - 2.4.3-pre+asdf", "2.4.3-alpha"],
-    ["1.2.3+asdf - 2.4.3+asdf", "2.4.3-alpha"],
```
kt3k, line 16: This case is incompatible `npm:semver`. The below prints false, but expected true here.

```js
import * as semver from "npm:semver";
const range = new semver.Range("1.2.3+asdf - 2.4.3+asdf");
console.log(range.test("2.4.3-alpha"));
```

semver/range_test.ts
```diff
@@ -71,7 +70,6 @@ Deno.test("range", async (t) => {
     ["1.2.3 >=1.2.1", "1.2.3"],
     [">=1.2.3 >=1.2.1", "1.2.3"],
     [">=1.2.1 >=1.2.3", "1.2.3"],
-    ["<=1.2.3", "1.2.3-beta"],
```
kt3k, line 74: This case is also incompatible `npm:semver`. The below prints false, but expected true here.

```js
import * as semver from "npm:semver";
const range = new semver.Range("<=1.2.3");
console.log(range.test("1.2.3-beta"));
```

semver/range_test.ts
```diff
@@ -185,28 +183,11 @@ Deno.test({
   },
 });
 
-Deno.test("unlockedPrereleaseRange", function () {
-  const versions: [string, string][] = [
-    ["*", "1.0.0-rc1"],
-    ["^1.0.0-0", "1.0.1-rc1"],
-    ["^1.0.0-rc2", "1.0.1-rc1"],
-    ["^1.0.0", "1.0.1-rc1"],
-    ["^1.0.0", "1.1.0-rc1"],
```
kt3k, line 194: These cases don't pass with `npm:semver`. For example:

```js
import * as semver from "npm:semver";
const range = new semver.Range("*");
console.log(range.test("1.0.0-rc"));
```

The above prints false, but expected true here.

semver/range_test.ts
iuioiua: Is this test from `npm:semver` present in `std/semver` somewhere?

semver/range_test.ts
iuioiua: Ditto. Is this `npm:semver` present somewhere?

semver/range_test.ts
iuioiua: Ditto. Do we have this `npm:server` test somewhere?

semver/range_test.ts
kt3k: I don't get the question well.. This case was tested with the opposite expectation (as `npm:semver`), and this PR removes it.

Do you suggest we should add this case with fixed expectation?

nicolo-ribaudo:
@kt3k Out of curiosity, do these cases behave differently from `npm:semver` also when you pass `{ includePrereleases: true }` to it?

semver/range_test.ts
iuioiua: > Do you suggest we should add this case with fixed expectation?

Yes. Ditto with the other similar comments.

kt3k: Added this case back to `testRange() returns false when the version is not in the range` with fixed expectation

semver/range_test.ts
kt3k: Added this case back to `testRange() returns false when the version is not in the range` with fixed expectation

semver/range_test.ts
kt3k: Added these cases back to `testRange() returns false when the version is not in the range` with fixed expectations

iuioiua (approved):

PR #4325, opened 2024-02-13
refactor: rename underscore packages to hyphens

crowlKats:
needs to add backwards compatible packages


timreichen:
Was this discussed somewhere?

crowlKats:
@timreichen jsr changed its naming policy to use hyphens instead of underscores.

kt3k:
How about doing these renames in `_tools/convert_to_workspace.ts`? That will only affects jsr version of the standard library.

timreichen:
That is strange. Is there an issue you can link to about why that change was made? Using hyphens instead of underscores directly violates the [deno style guide](https://docs.deno.com/runtime/manual/references/contributing/style_guide#use-underscores-not-dashes-in-filenames).

cknight:
> Was this discussed somewhere?

See https://discord.com/channels/684898665143206084/1203185670508515399/1206693236425490483

iuioiua:
Closing as this has been superseded by #4357.

PR #4328, opened 2024-02-14
docs(semver): fix arg names in jsdoc

kt3k:
This PR fixes the arg names in jsdoc of `compare`. Currently the jsdoc and the function signature use different names for arguments, and they look confusing.

This change makes the following change in VS Code overlay view.

BEFORE
Screenshot 2024-02-14 at 12 54 38

AFTER
Screenshot 2024-02-14 at 12 54 02

### DIFF
--- a/semver/compare.ts
+++ b/semver/compare.ts
@@ -9,7 +9,7 @@ import {
 /**
  * Compare two semantic version objects.
  *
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Good catch!

PR #4329, opened 2024-02-14
docs: update module stability

kt3k:
To my understanding, `encoding`, `front_matter`, `path`, and `streams` are stable. `io` is now undeprecated.


kt3k:
> Should we also update _tools/check_circular_submodule_dependencies.ts?

Yes, I'll do that in a following PR

PR #4334, opened 2024-02-15
chore(semver): align removal version of reverseSort to others

kt3k:
This PR suggests to move the removal version of `reverseSort` to `after 0.217.0` to align it to other items (By this change, the users are exposed to the breaking changes less often. That will cause less frictions to the existing users of semver)

### DIFF
--- a/semver/reverse_sort.ts
+++ b/semver/reverse_sort.ts
@@ -4,7 +4,7 @@ import { compare } from "./compare.ts";
 
 /**
  * Sorts a list of semantic versions in descending order.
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Should we do the same with [`Comparator.semver`](https://github.com/denoland/deno_std/blob/55c33e1e8a3f51747d7d02213cd7c5913bfba5a4/semver/types.ts#L30-L33)?

kt3k:
> Should we do the same with [Comparator.semver](https://github.com/denoland/deno_std/blob/55c33e1e8a3f51747d7d02213cd7c5913bfba5a4/semver/types.ts#L30-L33)?

Thanks for catching it. Yes, we should. I'll send another PR

PR #4337, opened 2024-02-15
doc: fix `std/path/posix` link

jersou:
Just the "path/" missing in the link...


iuioiua (approved):
Nice catch. Thank you.

PR #4343, opened 2024-02-16
feat(io): un-deprecate `readerFromStreamReader()`

### LINKED ISSUE #4336: Deprecation of readerFromStreamReader seems to have no direct replacement from Readable to Reader
**Describe the bug**

I'm unable to find a replacement for the deprecated `readerFromStreamReader` when trying to map a steams API `Readable` to a stdlib `Reader`

**Steps to Reproduce**

```ts
// old code
const fileReader = (await Deno.open(tarballPath, { read: true }))
  .readable
  .pipeThrough(new DecompressionStream('gzip'))
  .getReader();
const untar = new Untar(readerFromStreamReader(fileReader));

// trying to get this to work
const fileReader = (await Deno.open(tarballPath, { read: true }))
  .readable
  .pipeThrough(new DecompressionStream('gzip'))
  .getReader();
const untar = new Untar(fileReader);
//    is not compatible ^^^^^^^^^^ getReader.read() vs Reader.read(uint)
```

**Expected behavior**

I should be able to go from an `Deno.FsFile` while using a `pipeThrough` to a `Reader` in order to use some of the stdlib APIs like `tar`, etc.

If there is a way to convert this without writing custom code, this should be documented. Otherwise we should re-consider deprecating `readerFromStreamReader`.

**Environment**

- OS: macOS Ventura arm64
- deno version: 1.40.3
- std version: 0.215.0

iuioiua:
This change un-deprecates `readerFromStreamReader()` and moves it from `std/streams` to `std/io`.

Closes #4336

### DIFF
--- a/streams/reader_from_stream_reader.ts
+++ b/streams/reader_from_stream_reader.ts
@@ -19,7 +19,7 @@ import type { Reader } from "../io/types.ts";
  * await copy(reader, file);
  * ```
  *
- * @deprecated (will be removed in 0.220.0) Import from {@link https://deno.land/std/io/reader_from_stream_reader.ts} instead.
+ * @deprecated (will be removed in 1.0.0) Import from {@link https://deno.land/std/io/reader_from_stream_reader.ts} instead.
  */
 export function readerFromStreamReader(
   streamReader: ReadableStreamDefaultReader<Uint8Array>,
--- a/io/mod.ts
+++ b/io/mod.ts
@@ -24,6 +24,7 @@ export * from "./read_long.ts";
 export * from "./read_range.ts";
 export * from "./read_short.ts";
 export * from "./read_string_delim.ts";
+export * from "./reader_from_stream_reader.ts";
 export * from "./slice_long_to_bytes.ts";
 export * from "./string_reader.ts";
 export * from "./string_writer.ts";
--- a/io/reader_from_stream_reader.ts
+++ b/io/reader_from_stream_reader.ts
@@ -1,8 +1,9 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import { readerFromStreamReader as _readerFromStreamReader } from "../io/reader_from_stream_reader.ts";
-import type { Reader } from "../io/types.ts";
+import { Buffer } from "./buffer.ts";
+import { writeAll } from "./write_all.ts";
+import type { Reader } from "./types.ts";
 
 /**
  * Create a {@linkcode Reader} from a {@linkcode ReadableStreamDefaultReader}.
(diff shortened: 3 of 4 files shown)


streams/reader_from_stream_reader.ts
```diff
@@ -20,25 +19,10 @@ import type { Reader } from "../io/types.ts";
  * await copy(reader, file);
  * ```
  *
- * @deprecated (will be removed after 1.0.0) Use {@linkcode ReadableStreamDefaultReader} directly.
+ * @deprecated (will be removed in 0.220.0) Import from {@link https://deno.land/std/io/reader_from_stream_reader.ts} instead.
```
kt3k, line 22: I'd prefer to keep the original removal version

andrewthauer:
@kt3k - Added some discussion / curiosity in general around the Streams API vs Deno Reader/Writer compatibility in #4336. Wondering if you might have some insights here?

streams/reader_from_stream_reader.ts
andrewthauer: Do you mean deprecate the `io` version? Since `readerFromStreamReader` is the inverse of `toReadableStream` it make sense to have in `io` imo.

kt3k: My suggestion is un-deprecating `readerFromStreamReader()` in `std/io` (as suggested in this PR) and keep (deprecated) `readerFromStreamReader()` in `std/streams` until v1.0 instead of removing them in `0.220.0`

> TBH - I'd vote to have the names of both of these match a little more closely. It's not obvious by the names they are related.

What is your suggestion? BTW this API has been here for relatively long time. I think randomly changing it to a different name generally has a negative effect to the ecosystem.

andrewthauer: > My suggestion is un-deprecating readerFromStreamReader() in std/io (as suggested in this PR) and keep (deprecated) readerFromStreamReader() in std/streams until v1.0 instead of removing them in 0.220.0

Thanks for clarifying. That makes sense to me.

> What is your suggestion? BTW this API has been here for relatively long time. I think randomly changing it to a different name generally has a negative effect to the ecosystem.

It's probably better to keep the names as is, rather then changing or introducing an alternative, etc. Maybe since `readerFromStreamReader` is being added back to the `io` namespace alongside `toReadableStream` it will be more obvious. I guess the difference in terms `StreamReader` and `ReadableStream` doesn't immediately say, these are 2 are related and opposite to each other. Naming is hard 😂.

iuioiua: Removal version adjusted to v1.

kt3k (approved):
LGTM 👍

iuioiua:
Merging as I received support for this un-deprecation from @crowlKats offline.

PR #4345, opened 2024-02-16
test(semver): add test for parse_range

javihernant:


javihernant:
This is just a draft. I am not sure if test should be implemented like this. Can anyone confirm?

semver/parse_range_test.ts
```diff
@@ -0,0 +1,77 @@
...
+  }
+  for (let i = 0; i < cs0.length; i++) {
+    if (!equalComp(cs0[i], cs1[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+function equalRanges(
```
kt3k, line 34: I don't think we need this util. We can use `assertEquals` to compare nested objects.

javihernant:
This new commit contains one way to test `parseRange()`. The only problem I see here is, if parseRange() implementation is slightly changed so that the object returned is a bit different from that in the expected array (in `parse_args_test.ts`), then the tests are going to fail. For example, when I originally made the `parse_args_test.ts` file, all tests worked fine. However, today I have rechecked and one of them didn't work because one of the objects returned had changed. It changed from having the property "operator" as an empty string to an undefined value. How do you think we can tackle this issue? What changes can I make to this PR?

kt3k:
> The only problem I see here is, if parseRange() implementation is slightly changed so that the object returned is a bit different from that in the expected array (in parse_args_test.ts), then the tests are going to fail. For example, when I originally made the parse_args_test.ts file, all tests worked fine. However, today I have rechecked and one of them didn't work because one of the objects returned had changed. It changed from having the property "operator" as an empty string to an undefined value. How do you think we can tackle this issue?

I think that's just fine. Tests are for catching such changes. We'll update the test expectation in that cases if the change looks reasonable.

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+    ]],
+  ];
+
+  for (const [r, expected] of ranges) {
+    const range = parseRange(r);
+    assertEquals(range, expected, 'call to parseRange("' + r + '") failed');
+  }
+});
+
+Deno.test("parseHyphenRanges", () => {
```
kt3k, line 569: We have some rule for naming test cases: https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md#tests

This case probably should be something like:
```suggestion
Deno.test("parseRange() parses ranges with hyphens", () => {
```

Can you also update the other test cases alike?

javihernant: done :)

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: ">=", major: 1, minor: 2, patch: 3, prerelease: [] },
+        { operator: "<", major: 2, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["^1.2.3+build", [
+      [
+        { operator: ">=", major: 1, minor: 2, patch: 3, prerelease: [] },
+        { operator: "<", major: 2, minor: 0, patch: 0 },
+      ],
+    ]],
```
kt3k, line 60: This one seems the same as the above one?

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          operator: "<=",
+          major: 1,
+          minor: 2,
+          patch: 3,
+          prerelease: [],
+          build: [],
+        },
+      ],
+    ]],
+    ["^1.2.3", [
```
kt3k, line 106: This case seems appearing twice

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          patch: 3,
+          prerelease: [],
+          build: [],
+        },
+      ],
+    ]],
+    ["=1.2.3", [
+      [
+        {
+          operator: undefined,
```
kt3k, line 127: I wonder why this isn't `"="`? 🤔

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          operator: ">",
+          major: 1,
+          minor: 0,
+          patch: 0,
+          prerelease: [],
+          build: [],
+        },
+      ],
+    ]],
+    [">1.0.0", [
```
kt3k, line 203: Looks a duplicate

kt3k:
@javihernant Thanks for updating! Looks close to be merged, but I found several cases were repeated with completely the same input and expectations. Can you clean up them?

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          operator: "<=",
+          major: 2,
+          minor: 0,
+          patch: 0,
+          prerelease: [],
+          build: [],
+        },
+      ],
+    ]],
+    ["<=2.0.0", [
```
kt3k, line 227: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          operator: "<",
+          major: 0,
+          minor: 0,
+          patch: 1,
+          prerelease: [],
+          build: [],
+        },
+      ],
+    ]],
+    [">=0.2.3 || <0.0.1", [
```
kt3k, line 331: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+      [
+        { operator: ">=", major: 1, minor: 2, patch: 0 },
+        { operator: "<", major: 1, minor: 3, patch: 0 },
+      ],
+      [
+        { operator: ">=", major: 2, minor: 0, patch: 0 },
+        { operator: "<", major: 3, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["1.2.x || 2.x", [
```
kt3k, line 375: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: "<", major: 3, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["2.*.*", [
+      [
+        { operator: ">=", major: 2, minor: 0, patch: 0 },
+        { operator: "<", major: 3, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["2.*.*", [
```
kt3k, line 391: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+      [
+        { operator: ">=", major: 1, minor: 2, patch: 0 },
+        { operator: "<", major: 1, minor: 3, patch: 0 },
+      ],
+      [
+        { operator: ">=", major: 2, minor: 0, patch: 0 },
+        { operator: "<", major: 3, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["1.2.* || 2.*", [
```
kt3k, line 413: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: "<", major: 0, minor: 1, patch: 0 },
+      ],
+    ]],
+    ["~2.4", [
+      [
+        { operator: ">=", major: 2, minor: 4, patch: 0 },
+        { operator: "<", major: 2, minor: 5, patch: 0 },
+      ],
+    ]],
+    ["~2.4", [
```
kt3k, line 453: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: "<", major: 2, minor: 5, patch: 0 },
+      ],
+    ]],
+    ["~>3.2.1", [
+      [
+        { operator: ">=", major: 3, minor: 2, patch: 1, prerelease: [] },
+        { operator: "<", major: 3, minor: 3, patch: 0 },
+      ],
+    ]],
+    ["~>3.2.1", [
```
kt3k, line 465: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: "<", major: 3, minor: 3, patch: 0 },
+      ],
+    ]],
+    ["~1", [
+      [
+        { operator: ">=", major: 1, minor: 0, patch: 0 },
+        { operator: "<", major: 2, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["~>1", [
```
kt3k, line 477: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+    [">=0.7.x", [[{ operator: ">=", major: 0, minor: 7, patch: 0 }]]],
+    ["<0.7.x", [[{ operator: "<", major: 0, minor: 7, patch: 0 }]]],
+    [">1.2", [[{ operator: ">=", major: 1, minor: 3, patch: 0 }]]],
+    ["^0.0.1", [
+      [
+        { operator: ">=", major: 0, minor: 0, patch: 1, prerelease: [] },
+        { operator: "<", major: 0, minor: 0, patch: 2 },
+      ],
+    ]],
+    ["^0.0.1", [
```
kt3k, line 518: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+      [
+        { operator: ">=", major: 0, minor: 0, patch: 1, prerelease: [] },
+        { operator: "<", major: 0, minor: 0, patch: 2 },
+      ],
+    ]],
+    ["^1.2.3", [
```
kt3k, line 524: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+        { operator: "<", major: 2, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["^1.2.3", [
+      [
+        { operator: ">=", major: 1, minor: 2, patch: 3, prerelease: [] },
+        { operator: "<", major: 2, minor: 0, patch: 0 },
+      ],
+    ]],
+    ["^1.2", [
```
kt3k, line 536: seems a duplicate

semver/parse_range_test.ts
```diff
@@ -0,0 +1,869 @@
...
+          major: 0,
+          minor: 0,
+          patch: 0,
+          prerelease: [],
+          build: [],
+          semver: { major: 0, minor: 0, patch: 0, prerelease: [], build: [] },
+        },
+      ],
+    ]],
+    ["^1.2.3", [
```
kt3k, line 555: seems a duplicate

javihernant:
Updated the tests to remove all duplicates. I'm sorry I left so many!!!

kt3k (approved):
LGTM. Thanks for working on this!

PR #4347, opened 2024-02-17
docs(media_types): fix erroneous docs in extensions_by_type.ts

nberlette:
The function `extensionsByType` has conflicting information in its JSDoc comment on whether or not the returned extensions have a leading `.` in their names. They do not, in fact, have a leading dot. This PR updates the docs to accurately reflect that.


kt3k (approved):
Thanks for catching this! LGTM

Issue #4344, opened 2024-02-16
bug: `parseRange()` returns unexpected result


javihernant:
**Describe the bug**
`parseRange('>1.2')` produces the following `Range` object:
```
[ [ { operator: ">", major: 2, minor: 0, patch: 0 } ] ]
```

**Expected behavior**
It should instead produce (according to [node-semver](https://github.com/npm/node-semver?tab=readme-ov-file)):
```
[ [ { operator: ">=", major: 1, minor: 3, patch: 0 } ] ]
```
Otherwise, versions 1.x  wouldn't be included in the range.

I'd like to work on fixing this!

kt3k:
> **Expected behavior** It should instead produce (according to [node-semver](https://github.com/npm/node-semver?tab=readme-ov-file)):
> 
> ```
> [ [ { operator: ">=", major: 1, minor: 3, patch: 0 } ] ]
> ```

I tested with the below code:

```js
import * as semver from "npm:semver";
const range = new semver.Range(">1.2.0");
console.log(range);
```

and it printed:

```
Range {
  options: {},
  loose: false,
  includePrerelease: false,
  raw: ">1.2.0",
  set: [
    [
      Comparator {
        options: {},
        loose: false,
        operator: ">",
        semver: SemVer {
          options: {},
          loose: false,
          includePrerelease: false,
          raw: "1.2.0",
          major: 1,
          minor: 2,
          patch: 0,
          prerelease: [],
          build: [],
          version: "1.2.0"
        },
        value: ">1.2.0"
      }
    ]
  ],
  range: ">1.2.0"
}
```

So I think the expectation here is:
```
[ [ { operator: ">=", major: 1, minor: 2, patch: 0 } ] ]
```

kt3k:
fixed by #4350

PR #4350, opened 2024-02-17
fix(semver): fix parse_range for `>x.y` pattern

javihernant:
fix parse_range behaviour in ranges containing the greater than operator

fixes issue #4344


kt3k:
Thanks for the fix! Can you add test cases which this change fixes?

semver/parse_range.ts
```diff
@@ -187,14 +187,12 @@ function handleGreaterThanOperator(groups: RegExpGroups): Comparator[] {
...
   if (minorIsWildcard) {
-    if (patchIsWildcard) {
-      return [{ operator: ">=", major: major + 1, minor: 0, patch: 0 }];
-    }
-    return [{ operator: ">", major: major + 1, minor: 0, patch: 0 }];
+    return [{ operator: ">=", major: major + 1, minor: 0, patch: 0 }];
   }
   if (patchIsWildcard) {
-    return [{ operator: ">", major: major + 1, minor: 0, patch: 0 }];
+    return [{ operator: ">=", major, minor: minor + 1, patch: 0 }];
```
kt3k, line 195: See https://github.com/denoland/deno_std/issues/4344#issuecomment-1950969831

I don't think this should be `>=`

kt3k (approved):
LGTM. Thanks for the fix!

PR #4351, opened 2024-02-18
fix(media_types): load extensions when directly importing `extensionsByTypes()`

kt3k:
`extensionsByTypes` always returns `undefined` when imported from `media_types/extensions_by_type_test.ts`.

This PR fixes it by importing `_db.ts` (`_db.ts` sets values in `extensions` Map object when imported).

### DIFF
--- a/media_types/extensions_by_type.ts
+++ b/media_types/extensions_by_type.ts
@@ -3,6 +3,7 @@
 
 import { parseMediaType } from "./parse_media_type.ts";
 import { extensions } from "./_util.ts";
+import "./_db.ts";
 
 export { extensions };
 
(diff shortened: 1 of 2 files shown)


iuioiua:
> `extensionsByTypes` always returns `undefined` when imported from `media_types/extensions_by_type_test.ts`.

Do you mean when imported from `media_types/extensions_by_type.ts`?

> This PR fixes it by importing `_db.ts` (`_db.ts` sets values in `extensions` Map object when imported).

While I'm for this PR, this behaviour doesn't make sense to me. I think we should define and populate `extensions` in a single file.

iuioiua (approved):

kt3k:
> I think we should define and populate extensions in a single file - perhaps, _db.ts or a new _extensions.ts.

Sounds good to me.

PR #4354, opened 2024-02-19
refactor(semver): prepare for `noUncheckedIndexedAccess`

javihernant:
ref #4040

### DIFF
--- a/semver/parse_range.ts
+++ b/semver/parse_range.ts
@@ -5,7 +5,7 @@ import { OPERATOR_XRANGE_REGEXP, XRANGE } from "./_shared.ts";
 import { parseComparator } from "./_parse_comparator.ts";
 import { parseBuild, parsePrerelease } from "./_shared.ts";
 
-function isWildcard(id: string): boolean {
+function isWildcard(id?: string): boolean {
   return !id || id.toLowerCase() === "x" || id === "*";
 }
 
@@ -50,9 +50,9 @@ function parseHyphenRange(range: string) {
   if (isWildcard(rightGroups.major)) {
     to = "";
   } else if (isWildcard(rightGroups.minor)) {
-    to = `<${+rightGroups.major + 1}.0.0`;
+    to = `<${+rightGroups.major! + 1}.0.0`;
   } else if (isWildcard(rightGroups.patch)) {
-    to = `<${rightGroups.major}.${+rightGroups.minor + 1}.0`;
+    to = `<${rightGroups.major}.${+rightGroups.minor! + 1}.0`;
   } else if (rightGroups.prerelease) {
     to =
       `<=${rightGroups.major}.${rightGroups.minor}.${rightGroups.patch}-${rightGroups.prerelease}`;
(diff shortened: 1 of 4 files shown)


semver/parse_range.ts
```diff
@@ -5,8 +5,8 @@ import { OPERATOR_XRANGE_REGEXP, XRANGE } from "./_shared.ts";
 import { parseComparator } from "./_parse_comparator.ts";
 import { parseBuild, parsePrerelease } from "./_shared.ts";
 
-function isWildcard(id: string): boolean {
-  return !id || id.toLowerCase() === "x" || id === "*";
+function isWildcard(id: string | undefined): boolean {
```
iuioiua, line 8: Please use the optional `?` assertion for option arguments instead of `undefined` union types. Ditto to other instances.

semver/parse_range.ts
```diff
@@ -50,9 +50,9 @@ function parseHyphenRange(range: string) {
   if (isWildcard(rightGroups.major)) {
     to = "";
   } else if (isWildcard(rightGroups.minor)) {
-    to = `<${+rightGroups.major + 1}.0.0`;
+    to = `<${+rightGroups.major! + 1}.0.0`;
```
iuioiua, line 53: Can you please provide a justification for this non-null assertion? Ditto to other similar non-null assertions.

iuioiua (changes requested):

semver/range_intersects.ts
```diff
@@ -12,7 +12,7 @@ function rangesSatisfiable(ranges: Range[]): boolean {
 function comparatorsSatisfiable(comparators: Comparator[]): boolean {
   // Comparators are satisfiable if they all intersect with each other
   for (let i = 0; i < comparators.length - 1; i++) {
-    const c0 = comparators[i];
+    const c0 = comparators[i]!;
```
javihernant, line 15: comparators is an array of Comparators (`Comparator[]`). There is no possibility `c0` is going to be undefined.

semver/parse_range.ts
javihernant: ok, but `?` is only available for functions whose required parameters don't follow an optional one. Meaning, I can change it in `isWildcard()`, but not in `parseNumber()` (`_shared.ts`)

semver/parse_range.ts
javihernant: Also, I was looking at [](https://github.com/denoland/deno_std/pull/4159/commits/db1f140075f6162047385ac4930b0ba2890e0c3d) as reference

javihernant:
changed to using `?` whenever possible. Also, provided justification for code that uses non-null assertion (ie. `!`). Let me know if there's anything else needed

semver/_shared.ts
```diff
@@ -185,7 +185,10 @@ export function parseBuild(buildmetadata: string) {
   return buildmetadata.split(".").filter(Boolean);
 }
 
-export function parseNumber(input: string, errorMessage: string) {
+export function parseNumber(input: string | undefined, errorMessage: string) {
+  if (input === undefined) {
+    return 0;
+  }
   const number = Number(input);
```
iuioiua, line 192: This change seems wrong. Instead, I think we should use non-null assertions for the values we pass to the `input` argument. Even if the `input` is `null`, the function should fail, which is what we want.

iuioiua (changes requested):

iuioiua (approved):
LGTM! Thank you for this.

PR #4356, opened 2024-02-19
refactor(path): prepare for `noUncheckedIndexedAccess`

javihernant:


path/_common/glob_to_reg_exp.ts
```diff
@@ -65,7 +65,7 @@ export function _globToRegExp(
 
   // Remove trailing separators.
   let newLength = glob.length;
-  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]); newLength--);
+  for (; newLength > 1 && c.seps.includes(glob[newLength - 1]!); newLength--);
```
javihernant, line 68: `glob` is accessing their values using the `newLength` that is (initially) the length of `glob`. So, using `!` is justified as the elements accessed will always be defined

path/windows/to_file_url.ts
```diff
@@ -20,10 +20,12 @@ export function toFileUrl(path: string): URL {
...
   }
-
   const [, hostname, pathname] = path.match(
     /^(?:[/\\]{2}([^/\\]+)(?=[/\\](?:[^/\\]|$)))?(.*)/,
   )!;
+  if (!pathname) {
+    throw new TypeError("Invalid pathname.");
+  }
   const url = new URL("file:///");
   url.pathname = encodeWhitespace(pathname.replace(/%/g, "%25"));
```
javihernant, line 30: to use `pathname.replace()` first it's needed to check that pathname is not `undefined` (due to maybe a failed regex match)

iuioiua: Are you able to add a test for this new case?

iuioiua (approved):
LGTM! Thank you.

iuioiua:
Towards #4040

PR #4357, opened 2024-02-19
build: update _ to - in workspace converter script

lucacasonato:


crowlKats (approved):

iuioiua (changes requested):
Is it intentional that folder names still contain underscores? I think we should change that too.

lucacasonato:
Yes this is intentional for the time being. This can be addressed in another follow up

iuioiua (approved):

kt3k (approved):

PR #4359, opened 2024-02-20
chore(internal): remove `warnOnDeprecatedApi()` and `internal` folder

iuioiua:
It looks like we will not be using this function anytime in the future.


kt3k (approved):
LGTM.

BTW I think `internal` module could be useful for deduplicating shared private utils such as `assert/_diff.ts` and `expect/_diff.ts`

iuioiua:
Yep, I can do in a follow up PR.

PR #4360, opened 2024-02-20
chore(_tools): remove unneeded `--unstable` flag

iuioiua:
Also avoids an unnecessary warning being printed in the terminal.


kt3k (approved):
LGTM. Thanks!

Issue #4272, opened 2024-02-02
Command flag for `ParseOptions`


ngdangtu-vn:
**Is your feature request related to a problem? Please describe.**

I want to create a CLI with multiple command that flags are used differently. Something like the `deno` cli.

**Describe the solution you'd like**

A nest of `ParseOptions` would be ideal. Here is my purposal:

```ts
import { parseArgs } from 'https://deno.land/std@0.213.0/cli/parse_args.ts'

const option = {
   commandList: [
      { command: 'build' },
      { command: 'dev', alias: { watch: 'w' }, boolean: [ 'watch' ] },
      {
         command: 'start',
         commandList: [
            { command: 'the-sims', },
            { command: 'fallout', alias: { series: 's' }, string: [ 'series' ] },
         ],
         alias: { mode: 'm', vsync: 'v' },
         string: [ 'mode' ],
         boolean: [ 'vsync' ],
      }
   ],
   alias: { help: 'h', version: 'v' },
   boolean: [ 'help', 'version' ],
}

parseArgs([ 'build', '-w' ], option)
// parsedArgs: { command: [ 'build' ], watch: true, w: true, _: [] }

parseArgs([ '-w', 'build' ], option)
// parsedArgs: { watch: true, w: true, _: [ 'build' ] }

parseArgs([ 'start', 'fallout', '-v', '-m window' ], option)
// parsedArgs: { command: [ 'start', 'fallout' ], v: true, vsync:true, m:'window', mode:'window', _: [] }

parseArgs([ 'start', '-v', 'fallout', '--mode=window' ], option)
// parsedArgs: { command: [ 'start' ], v: true, vsync:true, m:'window', mode:'window', _: [ 'fallout' ] }
```

**Describe alternatives you've considered**

I don't know, but for now I will check the first and second item in `Deno.args` to decide if it is a command or not. It is fine but it doesn't offer a good context for alias flags.

timreichen:
I remember opening an issue a few years back with a similar request: https://github.com/denoland/deno/issues/6917. I would love to have such an addition in std!
But imo `parse_args()` might be suboptimal to do that task, because one passes an object with properties that contain a list of arguments. I think it would be better the other way around: The usage of Command and Option declarations.
For example:

```ts
interface Option {
  name: string;
  alias?: string;
  type?: BooleanConstructor | StringConstructor | NumberConstructor;
}

interface Command {
  name: string;
  options?: Option[];
  commands?: Command[];
}

interface Program {
[...]

PR #4362, opened 2024-02-20
feat(cli/unstable): add `parse()` with descriptive schema

timreichen:
ref: https://github.com/denoland/deno_std/issues/4272

This PR adds `parse()` that takes a descriptive schema for parsing args.
This is inspired  by rusts `clap`, `yargs` and `commanderjs`.

This needs more tests for edge cases and some discussions about features like
- `Option.value.accepted?: T[]`  for allowed values as a shortcut for
`Option.fn: (value) => { if (!accepted.includes(value)) { throw new Error("...")}
- `Option.value.min?: number` for a minimum of arguments taken
- `Option.value.max?: number` for a maximum of arguments taken 

The type generics also need some rework to imply correct types based on `Option.default` and `Option.type` for returned `result` and inside `Option.fn(value: T)`
Feedback and suggestions are very welcome. @iuioiua @ngdangtu-vn


timreichen:
> This is the 2nd set of API for parsing cli arguments. It feels confusing to have 2 different sets of APIs for the same purpose.

I agree, two wouldn't make too much sense, though the one currently implemented is based on minimist and is not capable of some use cases, especially comparing it to other poplar tools like cliffy, nor compatible with such a schema based approach.

> I'd recommend you develop this tool as a 3rd party tool. If that tool got popularity and adoption in the ecosystem, we would be able to consider adopting it in std

There are plenty of popular 3rd party tools out there (commander, yargs, cliffy, etc.) that have this declarative approach implemented and are far more popular than minimist. So I think publishing another 3rd party tool wouldn't make much sense to prove that point.

The declarative approach this implements avoids the function call chains and replaces them with native objects and arrays to make it less framework-ish, but is similar to the named modules.
Example:
```ts
...
.option("--foo")
.option("--bar")
```
 =>
```ts
{ ...
  options: [
    { name: "foo", },
    { name: "bar", },
  ]
}
```

andrewthauer:
Deno is fantastic tool for writing scripts and CLIs in general. I've been using it a lot lately and find that I can get a lot of mileage using just Deno + `@std`. The only other libraries I typically reach for are `dax` & `cliffy`.

For parsing args, the @std `parseArgs` works in a pinch for very basic usage. However, I don't find the API overly intuitive and quite simplistic for anything but very basic scripts where I don't feel the need to provide a nice DX on top.

The direction @timreichen is taking here would be much more preferred imo then the existing `parseArgs`. I believe it would help bridge the gap between very simplistic scripts/CLIs and a more advanced CLI which I'd typically reach for `cliffy` to implement parsing with.

kt3k:
> There are plenty of popular 3rd party tools out there (commander, yargs, cliffy, etc.) that have this declarative approach implemented and are far more popular than minimist.

How do you measure the popularity of packages? `yargs` has 72M weekly downloads, and `minimist` has 44M. I think these are similarly popular.

Also I don't see what `declarative approach` means. `minimist` (and therefore `parseArgs` of `std/cli`) accepts the entire parameter all at once as an option object. There's nothing procedural here. I would say minimist also has 'declarative' API.

> So I think publishing another 3rd party tool wouldn't make much sense to prove that point.

Without being tested as 3rd party library, how can we be convinced that this design is better than minimist?

timreichen:
### Sidenote: What I mean by declarative approach
I mean having all declarations in one object structured in a certain way:

#### Option properties
While `minimist` declares options in an object, it is done so by splitting the declaration by property.

property->objectName

```ts
parseArgs(Deno.args, {
  boolean: ["color"], // type is declared here
  default: { color: true }, // default value is declared here
  negatable: ["color"], // negatable is declared here
});
```

`commanderjs`, `yargs` and friends do it the other way around, where all properties of an option are gathered in one object (or in `commanderjs` case in one command).

object->properties

```ts
parse(Deno.args, {
  options: [
    { name: "color", type: Boolean, default: true, negatable: true } // all option properties declared in one object
  ]
});
```

#### Subcommands
As explored [here](https://github.com/denoland/deno_std/issues/4272), minimist does not support subcommand parsing and a possible implementation would lead to messy code due to the nature of property->objectName.

#### Named arguments
Since there is no *option object* but a collection of properties declared, it is not possible to have named arguments for options. This is a problem, when one wants to be able to print *help* for a cli. 

property->objectName

```ts
???
```


object->properties

```ts
parse(Deno.args, {
  options: [
    { name: "color", type: Boolean, default: true, negatable: true } // all option properties declared in one object
  ]
});
```
> How do you measure the popularity of packages? `yargs` has 72M weekly downloads, and `minimist` has 44M. I think these are similarly popular.

Yes, but `commanderjs` has 141M weekly downloads. So 

> Also I don't see what `declarative approach` means. `minimist` (and therefore `parseArgs` of `std/cli`) accepts the entire parameter all at once as an option object. There's nothing procedural here. I would say minimist also has 'declarative' API.

I mean that it is possible to handle subcommands as pointed out in https://github.com/denoland/deno_std/issues/4272 as well as named arguments for options.

One thing that seems very odd about minimist is that an option declaration is split by properties. This leads to repetition for options with lots of properties. Instead `yargs` and friends declare all properties in one object which seems more logical and js-api-ish to me (object->properties instead of properties->object).


> Without being tested as 3rd party library, how can we be convinced that this design is better than minimist?

I was talking about the fact that `commanderjs` and other libs that follow such an implementation are already used extensively and are more popular than `minimist` (by weekly downloads). So a separate 
Some functionality like subcommands are apparently not possible with minimist.

kt3k:
> Since there is no option object but a collection of properties declared, it is not possible to have named values for options. This is a problem, when one wants to be able to print help for a cli.

I don't see this argument very well. `--help` is usually done like the below with the current `parseArgs`:

```ts
import { parseArgs } from "jsr:@std/cli/parse-args";

const args = parseArgs(Deno.args, {
  boolean: ["help"],
});
// This becomes { _: [], help: true }, { _: [], help: false }, etc.

if (args.help) {
  console.log("Usage");
  Deno.exit(0);
}
```

Isn't `{ _: [], help: true }` named option?

Also I don't see well what `{ name: "VALUE" }` part does in your example..

> I was talking about the fact that commanderjs and other libs that follow such an implementation are already used extensively and are more popular than minimist (by weekly downloads).

But this suggested API design isn't similar to `commander` nor `yargs` at all. It's completely an invented design, which is not tested anywhere.

kt3k:
@timreichen BTW what do you think about [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) of Node.js, which is also available in Deno via `node:util` module

timreichen:
> @timreichen BTW what do you think about [`util.parseArgs`](https://nodejs.org/api/util.html#utilparseargsconfig) of Node.js, which is also available in Deno via `node:util` module

I think it is very bare-bone. But I like how one defines the options as an object with properties.

91pornaaaaa:
https://91porna.com/

cli/parse_test.ts
```diff
@@ -0,0 +1,725 @@
...
+
+  await t.step("strip single quoted value", () => {
+    const expected = { options: { foo: "bar" }, arguments: {} };
+    const actual = parse(["--foo", "'bar'"], {
+      options: [
+        { name: "foo", value: { name: "VALUE" } },
+      ],
+    });
+    assertEquals(actual, expected);
+  });
```
phaux, line 384: Why? Bash will already remove the quotes. You can just leave them if someone passes them anyway by escaping them.

It reminds me of PHP automatically escaping and unescaping random strings without my consent.

cli/parse_test.ts
```diff
@@ -0,0 +1,725 @@
...
+  await t.step("name without value property but present value", () => {
+    assertThrows(() => {
+      parse(["--foo=bar"], { options: [{ name: "foo" }] });
+    });
+  });
+  await t.step("name with multiple values", () => {
+    const expected = { options: { foo: ["bar", "baz"] }, arguments: {} };
+    const actual = parse(["--foo", "bar", "baz"], {
+      options: [{ name: "foo", value: { name: "VALUE", multiple: true } }],
+    });
```
phaux, line 146: I would expect to require passing `--foo bar --foo baz` here. The current `parseArgs` works that way. 

Same with all the other cases with `multiple`.

cli/parse_test.ts
```diff
@@ -0,0 +1,725 @@
...
+      ],
+    });
+    assertEquals(called, true);
+  });
+});
+
+Deno.test("parse() handles commands", async (t) => {
+  await t.step("fn() is called", () => {
+    let called = false;
+    parse(["run", "--foo"], {
```
phaux, line 622: I think parse with command should return which command was called.

This `parse` should return `{ command: ["run"], options: {...}, arguments: {...} }`.

Other examples:

- `[]` - no command was called.
- `["foo"]` - foo command was called.
- `["foo", "bar"]` - a nested command was called.

You would be able to discriminate on this property to get the correct types of options for a particular command.

cli/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  description?: string;
+  default?: T;
+  type?: OptionTypeConstructor<T>;
+  negatable?: boolean;
+  value?: {
+    name: string;
+    optional?: boolean;
+    multiple?: boolean;
+    requireEquals?: boolean;
+  };
```
phaux, line 20: I don't think this nesting is necessary and the value's name field can be removed (does it even do anything?)

cli/types.ts
```diff
@@ -0,0 +1,48 @@
...
+  fn?: (value: T) => T | void;
+}
+
+export interface Argument<T = unknown> {
+  name: string;
+  description?: string;
+  multiple?: boolean;
+  optional?: boolean;
+  fn?: (value: T) => T | void;
+}
```
phaux, line 30: 2 things:

1. Do the positional arguments need a name? I guess it's useful for generating help later, but they might as well be returned as a tuple. I think it's simpler.
2. Maybe instead of `multiple` and `optional` flags on arguments, have separate lists of `requiredArguments`, `optionalArguments`, and a single `restArgument` property on `Command`. That way it can be enforced that required arguments always come first and there's at most one rest (multiple) argument at the end. It doesn't make sense to have them in any other order.

cli/types.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface Command<T> {
+  name: string;
+  description?: string;
+  options?: ReadonlyArray<Option<T>>;
+  commands?: ReadonlyArray<Command<T>>;
+  arguments?: ReadonlyArray<Argument>;
+  fn?: (result: ParseResult) => void;
+}
```
phaux, line 39: Maybe consider `options` and `commands` to be a `Record` where keys are names?

cli/parse_test.ts
timreichen: Fair point.

cli/parse_test.ts
timreichen: Again, fair point.

cli/types.ts
timreichen: See below.

cli/types.ts
timreichen: I have considered that, but if an option doesn't have properties, They would need to be declared with an empty object, which seems strange:
```ts
parse(["--foo"], { options: { foo: {} } })
```
```ts
parse(["run", "--foo"], {
  commands: {
    run: { options: { foo: {} } }
  }
});
```
Or is there a better way?

cli/types.ts
timreichen: 1. That is exactly right, the name property as well as the nesting is intended for the (future) stringification of an option.
2. That gets more complicated, if one intends to extend the functionality with lets for example `standalone` options or `conflictingWith` options. So I think it is good to have these properties in the option object itself.

iuioiua:
I'll be closing this without merge. `parseArgs()` is one of the most widely used functions in the Standard Library. Swapping it out with a new implementation presents issues centered around disruption (I'm sure they're obvious, so I won't list them). We're currently focusing on Deno 2, including polishing the Standard Library up for it. So we need to be quite selective on major additions.

I think your PR is heading in the right direction, and the design seems superior to the current implementation. However, the case for the addition must be sufficiently strong for the Deno core team to accept. So I strongly recommend porting this implementation to your own package, and making the design, implementation, documentation, and testing as best as possible. Users should feel a tangible benefit from migrating too. It's probably best done accomplished with others. Then, once ready, present it at sometime in the future as a PR.

That all said, we greatly appreciate the thought and effort you've gone through with this, and other, PRs. Thank you very much.

Issue #4346, opened 2024-02-16
todo: fix tests after workspaces migration script is run


iuioiua:
I've yet to investigate this properly. I am putting this here in case someone can get to it before me. To fix:
1. Run the migration script using `deno run -A _tools/convert_to_workspace.ts`.
2. Run `deno task test`.
3. Fix any test failures.
4. Submit a PR of the fixes, excluding the migration script changes.

kt3k:
- [ ] ./assert/_format_test.ts
- [ ] ./crypto/crypto_test.ts:169:6
- [ ] digest() keeps memory usage reasonable with many calls => ./crypto/crypto_test.ts:251:6
- [ ] KeyStack() handles inspection => ./crypto/unstable_keystack_test.ts:225:6
- [ ] format() overrides default behaviours of Deno.inspect ... format() wraps Object with getters => ./expect/_format_test.ts:43:11
- [ ] format() doesn't truncate long strings in object => ./expect/_format_test.ts:89:6
- [ ] sprintf() handles formatV => ./fmt/printf_test.ts:620:6
- [ ] sprintf() handles lessThan flag => ./fmt/printf_test.ts:666:6
- [ ] sprintf() handles errors => ./fmt/printf_test.ts:677:6
- [ ] emptyDir() test ("testdata/testfolder") without --allow-read & without --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDirSync() test ("testdata/testfolder") without --allow-read & without --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDir() test ("testdata/testfolder") with --allow-read & without --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDirSync() test ("testdata/testfolder") with --allow-read & without --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDir() test ("testdata/testfolder") without --allow-read & with --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDirSync() test ("testdata/testfolder") without --allow-read & with --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDir() test ("testdata/testfolder") with --allow-read & with --allow-write permission => ./fs/empty_dir_test.ts:192:8
- [ ] emptyDirSync() test ("testdata/testfolder") with --allow-read & with --allow-write permission => ./fs/empty_dir_test.ts:192:8
[...]

PR #4367, opened 2024-02-21
chore(crypto): move test scripts to own files

iuioiua:
Fixes test cases in #4346 for `crypto/crypto_test.ts`. I can confirm that these tests pass after running the migration script.

These tests were previously failing because the migration script didn't factor in the inline code.

### DIFF
--- a/crypto/crypto_test.ts
+++ b/crypto/crypto_test.ts
@@ -2,11 +2,8 @@
 import { assert, assertEquals, assertInstanceOf, fail } from "../assert/mod.ts";
 import { crypto as stdCrypto } from "./mod.ts";
 import { repeat } from "../bytes/repeat.ts";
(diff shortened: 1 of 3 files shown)


crypto/crypto_test.ts
```diff
@@ -167,35 +164,9 @@ Deno.test("digest() handles length option", async () => {
...
-      largeDigest,
-      heapBytesAfterLarge,
-    }));
-  `;
-
   const command = new Deno.Command(Deno.execPath(), {
-    args: ["eval", "--no-lock", code],
-    cwd: moduleDir,
+    args: ["run", "--no-lock", "crypto/testdata/digest_large_inputs.ts"],
+    stderr: "inherit",
```
iuioiua, line 169: Helps with debugability.

kt3k (approved):
LGTM. Nice finding!

PR #4368, opened 2024-02-22
chore(http): fix spawned tests after migration script

iuioiua:
Adding `--config deno.json` seems to be needed for these tests to complete successfully after running the workspaces migration script. Shouldn't `deno.json` be automatically detected and loaded? This appears to be a bug.

I can confirm that these tests now pass after running the migration script.

Towards #4346

### DIFF
--- a/http/file_server_test.ts
+++ b/http/file_server_test.ts
@@ -344,10 +344,11 @@ Deno.test("serveDir() script prints help", async () => {
       "--no-check",
       "--quiet",
       "--no-lock",
-      "file_server.ts",
+      "--config",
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM 👍

kt3k:
> Shouldn't deno.json be automatically detected and loaded? This appears to be a bug.

We have `http/deno.json` after running the convert script. Is it possible that file affects the auto-discovery of deno.json at the top level?

iuioiua:
Note to self: test the affected files if/when a bug is confirmed and fixed.

Issue #4363, opened 2024-02-20
link replacement of fs.ensureSymlink()


ryota2357:
**Describe the bug**

I have two questions regarding the operation of invalid symlink files.
I thought it was a bug, but I apologize if it is a spec.

In the documentation, `ensureSymlink()` is:
> Ensures that the link exists, and points to a valid file.

When the link exists but the points are invalid, what should `ensureSymlink()` do?

**Steps to Reproduce**

Run the following script.

```ts
import * as fs from "https://deno.land/std@0.216.0/fs/mod.ts";
import * as path from "https://deno.land/std@0.216.0/path/mod.ts";

const cwd = Deno.cwd();

const originalAFile = path.join(cwd, "original_a.txt");
const originalBFile = path.join(cwd, "original_b.txt");
const linkFile = path.join(cwd, "linkfile.txt");

async function reproduce_replaceBroken() {
  await Deno.writeTextFile(originalAFile, "A");
  await Deno.writeTextFile(originalBFile, "B");

  await fs.ensureSymlink(originalAFile, linkFile);
  const expectA = await Deno.readTextFile(linkFile);
  console.log(expectA); // => A

  await Deno.remove(originalAFile);

  await fs.ensureSymlink(originalBFile, linkFile);
  const expectB = await Deno.readTextFile(linkFile); // cause error here.
  console.log(expectB);
}

async function cleanup() {
  await Deno.remove(originalAFile);
  await Deno.remove(originalBFile);
  await Deno.remove(linkFile);
}

await reproduce_replaceBroken();
// await cleanup();
```

**Expected behavior**

I firstly expected `linkFile` to be a valid link and it points to `originalBFile`, because I ensured that  'the link exists, and points to a valid file' with `fs.ensureSymlink(originalBFile, linkFile);`

However, I noticed that if `linkFile` is already a symlink file it does not replace to new link via the `ensureSymlink()` like the following scripts.

```ts
async function reproduce_replaceLink() {
  await Deno.writeTextFile(originalAFile, "A");
  await Deno.writeTextFile(originalBFile, "B");

  await fs.ensureSymlink(originalAFile, linkFile);
  const expectA = await Deno.readTextFile(linkFile);
  console.log(expectA); // => A

  await fs.ensureSymlink(originalBFile, linkFile);
  const expectB = await Deno.readTextFile(linkFile);
  console.log(expectB); // => A
}
```

Therefore, I do not know if this behavior is a bug or a specification.
[...]

PR #4371, opened 2024-02-22
fix(fs/ensure_symlink): check symlink is pointing the given target

ryota2357:
fix: #4363

> So the API doc seems correct, and I think we should check the symlink is actually pointing the given target file if `Deno.errors.AlreadyExists` is thrown

I implemented it.

### DIFF
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -19,6 +19,8 @@ function resolveSymlinkTarget(target: string | URL, linkName: string | URL) {
 /**
  * Ensures that the link exists, and points to a valid file.
  * If the directory structure does not exist, it is created.
+ * If the link already exists, it is not modified but error is thrown if it is not point to the given target.
+ * Requires the `--allow-read` and `--allow-write` flag.
  *
  * @param target the source file path
  * @param linkName the destination link path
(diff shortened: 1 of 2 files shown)


kt3k:
This solution replaces the link with a new link when the link target is not correct. I think this is one good solution, but I also wonder if it makes sense to throw some error when the existing link target is not correct. What do you think?

ryota2357:
> throw some error when the existing link target is not correct.
- I feels from "ensureSymlink(A, B)" word that "B always links to A"
- Anothoer ensure* function such that "ensureDir()" create directory at a given path.

From these, I think that there is no need to throw an error if the existing link target is not correct, only make the correct symlink.

However, I do not know enough about English or API design to know which is better.

kt3k:
> Anothoer ensure* function such that "ensureDir()" create directory at a given path.

Yes, they create new dirs or files, but they don't overwrite or replace existing things to my understanding. However this change replaces the existing symlink with a new one, and that symlink is silently lost in their system.

I personally think throwing some error with the message like `A symlink targeting to an undesired path already exists: ${linkName} -> ${linkRealPath}` would make more sense.

ryota2357:
I apologize for late response.

> they don't overwrite or replace existing things

I see, I think this is also clear and looks good to me.

I updated the implementation and added doc.
Could you please re-review it?

kt3k (approved):
LGTM. Thanks for the fix!

PR #4373, opened 2024-02-22
refactor(semver): rename `comparatorFormat()` to `formatComparator()`

timreichen:
**Changes**
- renames `comparatorFormat()` to `formatComparator()`

**Reasons**
- javascript generally has a verb-first convention
- this aligns with `parseComparator()`
- this function and file are private, so no breaking changes


iuioiua (approved):
Thanks! LGTM.

PR #4374, opened 2024-02-23
chore(semver): remove legacy `Range.ranges` object definition

timreichen:
This was missed after the removal of `SemVerRange`.

Ref: https://github.com/denoland/deno_std/pull/4161


iuioiua (approved):
Thank you.

PR #4375, opened 2024-02-23
refactor(semver): improve `parseRange()` logic

timreichen:
**Changes**
- splits `parseHyphenRange()` logic into separate functions.
- replaces string based returns with `Comparator[] | undefined` returns in `parseHyphenRange()` 

**Reasons**
- `parseHyphenRange()` should only parse hyphen ranges as the name implies. All other logic should be handled in separate functions.
- Splitting `handleLeftHyphenRangeGroups()` and `handleRightHyphenRangeGroups()` reduces if-else statements with early returns.
- returning `Comparator`s  instead of strings avoids string parsing after the function call.


semver/parse_range.ts
```diff
@@ -18,50 +23,105 @@ type RegExpGroups = {
   build?: string;
 };
 
-function parseHyphenRange(range: string) {
-  // remove spaces between comparator and groups
-  range = range.replace(/(?<=<|>|=) +/, "");
-
+function handleLeftHyphenRangeGroups(leftGroup: RegExpGroups) {
```
iuioiua, line 26:
```suggestion
function handleLeftHyphenRangeGroups(leftGroup: RegExpGroups): Comparator | undefined {
```

semver/parse_range.ts
```diff
@@ -18,50 +23,105 @@ type RegExpGroups = {
...
+    major: +leftGroup.major,
+    minor: +leftGroup.minor,
+    patch: +leftGroup.patch,
+    prerelease: leftGroup.prerelease
+      ? parsePrerelease(leftGroup.prerelease)
+      : [],
+    build: [],
+  };
+}
+function handleRightHyphenRangeGroups(rightGroups: RegExpGroups) {
```
iuioiua, line 59:
```suggestion
function handleRightHyphenRangeGroups(rightGroups: RegExpGroups): Comparator | undefined {
```

iuioiua (changes requested):
This seems like a good idea. To me, the readability improvement is worth the added LOC.

Can you please ensure these new functions include explicit return types?

iuioiua (approved):
LGTM! Thanks Tim.

PR #4376, opened 2024-02-23
chore: fix .editorconfig syntax

javihernant:
fix .editorconfig to make .js/.ts settings take effect.

Previously those settings were being ignored

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*.ts,*.js]
+[*.{ts,js}]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true


.editorconfig
```diff
@@ -1,6 +1,6 @@
 root = true
 
-[*.ts,*.js]
+[*{.ts,*.js}]
```
kt3k, line 3: Should this be the below?
```suggestion
[*{.ts,.js}]
```

kt3k: Probably the below would be better?
```suggestion
[*.{ts,js}]
```

javihernant:
updated commit.
From [editorconfig.org](https://editorconfig.org/)
```
# Matches multiple files with brace expansion notation
# Set default charset
[*.{js,py}]
charset = utf-8
```

kt3k (approved):
Thanks!

PR #4384, opened 2024-02-25
Allow for asserting throws on async methods 

mcgear:
We ran into a use case where we need to be able to assert throws from an async function. I was unable to figure out how to do this with existing assert throws, so added this implementation.

### DIFF
--- a/assert/assert_throws_async.ts
+++ b/assert/assert_throws_async.ts
@@ -3,15 +3,15 @@ import { assertIsError } from "./assert_is_error.ts";
 import { AssertionError } from "./assertion_error.ts";
 
 /**
- * Executes a function, expecting it to throw. If it does not, then it
+ * Executes an async function, expecting it to throw. If it does not, then it
  * throws.
  *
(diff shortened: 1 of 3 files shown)


aapoalas:
You're probably looking for `assertRejects`

iuioiua:
Aapo is right. See https://deno.land/std/assert/mod.ts?s=assertRejects

Perhaps we should mention the related asynchronous or synchronous function in each of the functions JSDocs. This isn't the first time someone hasn't realised one of these assertion functions exists.

iuioiua:
@mcgear, please see the comments above.

kt3k:
> Perhaps we should mention the related asynchronous or synchronous function in each of the functions JSDocs. This isn't the first time someone hasn't realised one of these assertion functions exists.

Sounds good. Let's mention `assertRejects` in the doc of `assertThrows`

mcgear:
This seems to have worked just fine, closing the PR.

PR #4385, opened 2024-02-25
docs: clarify underscore guidance in README

jsejcksn:


iuioiua (approved):
Good catch. Thank you.

PR #4387, opened 2024-02-25
chore(semver): move comparator functions into range files

timreichen:
Ref: https://github.com/denoland/deno_std/issues/3948#issuecomment-1960998276

**Changes**
- moves private comparator files into public range files.

**Reasons**
- all function are imported by just one file each.
- changes only concern private files and are non-breaking.
- function tests are covered by dependent range functions.


iuioiua (approved):
Nice cleanup! Thanks, Tim!

kt3k:
Many test cases (from `comparator_test.ts`, `_is_comparator_test.ts`, `_comparator_intersects_test.ts`) are lost in this change. Are we sure it's safe to remove those test cases?

iuioiua:
> Many test cases (from `comparator_test.ts`, `_is_comparator_test.ts`, `_comparator_intersects_test.ts`) are lost in this change. Are we sure it's safe to remove those test cases?

We can compare the test coverage differences by looking at https://app.codecov.io/gh/denoland/deno_std/pull/4387/tree/semver and https://app.codecov.io/gh/denoland/deno_std/tree/main/semver. I'll check this myself once I get a chance, but from a quick look, comparator code seems to be covered fine.

iuioiua:
@timreichen, I agree with Yoshiya. I don't think we should decrease test coverage because of this PR. Are you able to add relevant tests to the dependent range functions? To make this easier, here are the functions I think we should cover in tests as part of this PR:
1. [`isComparator()`](https://app.codecov.io/gh/denoland/deno_std/pull/4387/blob/semver/is_range.ts#L7)
2. [`parseComparator()`](https://app.codecov.io/gh/denoland/deno_std/pull/4387/blob/semver/parse_range.ts#L21)
3. [`comparatorMax()`](https://app.codecov.io/gh/denoland/deno_std/pull/4387/blob/semver/range_max.ts#L7)

semver/is_range.ts
```diff
@@ -1,6 +1,22 @@
...
+    typeof value !== "object"
+  ) return false;
+  if (value === NONE || value === ALL) return true;
+  const { operator, semver } = value as Comparator;
+  return (
+    (operator === undefined ||
+      OPERATORS.includes(operator)) &&
+    isSemVer(semver)
+  );
+}
```
iuioiua, line 19: Sidenote: I'm sure this function can be simplified (in a separate PR).

timreichen: I think once `Comparator.semver` is removed (deprecated now) this can be simplified a lot.

timreichen:
> Codecov still shows test coverage of these 3 affected functions still needs to be included. Are you able to target them? Then, we should be good to merge.

Sorry, I have no experience at all with codecov. I thought I ported all tests to the sibling range test files. What exactly is missing?

iuioiua:
Oh, sorry, yes, you did. Codecov can sometimes miss covered lines. Can you please rebase? @kt3k, any objections?

iuioiua:
Merging now as it seems there aren't any objections. This change is a soft blocker for other `std/semver` cleanups.

PR #4388, opened 2024-02-25
docs: ignore bad snippet examples

iuioiua:
These snippets are deliberately bad and should be ignored.

Fixes https://github.com/denoland/deno_std/actions/runs/8041176789/job/21960124528


kt3k (approved):
LGTM. Nice!

PR #4393, opened 2024-02-25
chore(log): sync `level` and `levelName` in BaseHandler

timreichen:
**Changes**
- makes `levelName` a getter and setter so it stays in sync with `level`.
- small cleanup with `options` in `constructor()`.

**Reasons**
- makes `levelName` and `level` should never get out of sync.
- avoids `||` operator by setting default formatter value.

### DIFF
--- a/log/base_handler.ts
+++ b/log/base_handler.ts
@@ -11,7 +11,7 @@ export interface BaseHandlerOptions {
 }
 
 export class BaseHandler {
-  #logLevel: LevelName;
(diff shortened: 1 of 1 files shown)


timreichen:
> This looks like causing the regression in the performance when accessing to handler.levelName.

I changed it to store the current value in a private property. This should ensure synchronisation while keeping performance.

kt3k (approved):
LGTM. Thanks

PR #4400, opened 2024-02-26
chore(http): format test names

timreichen:

### DIFF
--- a/http/server_test.ts
+++ b/http/server_test.ts
@@ -127,7 +127,7 @@ class MockListener implements Deno.Listener {
 }
 
 Deno.test(
-  "Server.addrs should expose the addresses the server is listening on",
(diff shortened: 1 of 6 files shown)


http/server_test.ts
```diff
@@ -127,7 +127,7 @@ class MockListener implements Deno.Listener {
 }
 
 Deno.test(
-  "Server.addrs should expose the addresses the server is listening on",
+  "Server() exposes the addresses the server is listening on as addrs property",
```
iuioiua, line 130: Can you please remove the round brackets on class names?

iuioiua (changes requested):

http/server_test.ts
timreichen: done

iuioiua (approved):
LGTM! Thanks again, Tim.

PR #4402, opened 2024-02-26
docs(jsonc): add module-level docs

iuioiua:


kt3k (approved):
LGTM. Nice!

jsonc/mod.ts
```diff
@@ -1,4 +1,22 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
+/**
+ * Provides tools for working with JSONC (JSON with comments).
+ *
+ * This module is browser compatible.
+ *
```
kt3k, line 8: Maybe it's also better to note that we currently only provide the parser of JSONC and serializer is not available

iuioiua: Added.

Issue #4404, opened 2024-02-27
suggestion: optional `reason` param for assert::unreachable


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

As we know, calling `unreachable()` doesn't actually guarantee that a code path is unreachable. It's an assertion that the writer _believes_ with a reasonable degree of certainty that the code path is unreachable, but they could still be wrong. Similarly to how calling `expect` with a message is usually preferable to `unwrap` in Rust, calling `unreachable` with a message would allow for self-documenting why the code path should never be hit in a way that would show up in error traces.

**Describe the solution you'd like**

Something like this:

```diff
# https://github.com/denoland/deno_std/blob/main/assert/unreachable.ts

- export function unreachable(): never {
-   throw new AssertionError("unreachable");
+ export function unreachable(reason?: string): never {
+   throw new AssertionError(reason ?? "unreachable");
  }
```

**Describe alternatives you've considered**

N/A

PR #4405, opened 2024-02-27
feat(assert): optional `reason` param for `unreachable` (#4404)

lionel-rowe:
Fixes https://github.com/denoland/deno_std/issues/4404

### DIFF
--- a/assert/unreachable.ts
+++ b/assert/unreachable.ts
@@ -11,6 +11,6 @@ import { AssertionError } from "./assertion_error.ts";
  * unreachable(); // Throws
  * ```
  */
-export function unreachable(): never {
-  throw new AssertionError("unreachable");
+export function unreachable(reason?: string): never {
+  throw new AssertionError(reason ?? "unreachable");
 }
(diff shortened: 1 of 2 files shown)


assert/unreachable.ts
```diff
@@ -11,6 +11,6 @@ import { AssertionError } from "./assertion_error.ts";
  * unreachable(); // Throws
  * ```
  */
-export function unreachable(): never {
-  throw new AssertionError("unreachable");
+export function unreachable(reason?: string): never {
```
iuioiua, line 14: We usually use the `msg` argument name in `std/assert`. Though, how this is now might be fine.

iuioiua (approved):
LGTM! I'm for the idea. I'll let Yoshiya review it, too. This does make me wonder whether we should introduce an `UnreachableError` class that extends `AssertionError`. Not a strong idea, though.

kt3k (approved):
The suggestion seems reasonable to me. LGTM

PR #4406, opened 2024-02-27
refactor: make the code work under `verbatimModuleSyntax`

dsherret:
I have an upcoming lint rule to enforce this. Was just testing it out on deno_std.

Read more here: https://www.typescriptlang.org/tsconfig#verbatimModuleSyntax


iuioiua (approved):
This is great! I've always wondered whether such a rule existed and resorted to trying to enforce this myself manually. Thanks David 👍🏾

PR #4407, opened 2024-02-27
fix(semver): fix `isComparator()` and `isRange()`

timreichen:
**Changes**
- fixes a bug where `isComparator()` returned `false` when comparator didn't contain a `semver` property.
- adds test cases.


kt3k:
Is there any behavior of public APIs which is affected by this fix?

timreichen:
> Is there any behavior of public APIs which is affected by this fix?

Yes, it affects `isRange()`. I added the tests to `is_range_test.ts` for clarity.

iuioiua (approved):
LGTM! Cheers Tim

iuioiua:
Interesting... The way `Comparator` is currently typed, the legacy way of defining it (i.e. with a `semver` property only) presents a type error. That means a breaking change was accidentally made in #4161 (0.213.0).

iuioiua:
Now, `comparator.semver` is due for removal. It's best we just remove it instead of applying this PR. Sorry for the delay, Tim.

timreichen:
> Now, `comparator.semver` is due for removal. It's best we just remove it instead of applying this PR. Sorry for the delay, Tim.

Ok, but I think we should add the new tests cases. Shall I update this PR to just do that?

iuioiua:
SGTM

timreichen:
Seems like these tests were already added when `is_comparator.ts` was merged with `is_range.ts`. Closing.

PR #4409, opened 2024-02-28
feat(tools): add jsdoc tags check

timreichen:
**Features**
- check for `@returns` instead of `@return`
  prints `⚠️ @returns tag is not allowed. Use @return tag instead: {path:line}`
- check for missing `@return` for functions with non-void return types
  prints `⚠️ @return tag is missing: {path:line}`
- check for missing `@example` for functions
  prints  `⚠️ @example tag is missing: {path:line}`
- check for empty `@param` without description
  prints `⚠️ @param tag comment is empty: {path:line}`

**Todos**
- check for missing jsdoc for exports


iuioiua:
I'll be closing this PR in favour of #4618, which uses a different approach. @timreichen, feel free to review that PR or help iterate on that JSDoc checker. Either way, thank you for making this suggestion.

PR #4412, opened 2024-02-28
feat(expect): support `expect.extend()` api

eryue0220:
This pr is one of prs to solve https://github.com/denoland/deno_std/issues/3964, which is to support `expect.extend` api.


expect/_extend_test.ts
```diff
@@ -0,0 +1,119 @@
...
+      return {
+        message: () =>
+          `expected ${actual} to be within range ${floor} - ${ceiling}`,
+        pass: false,
+      };
+    }
+  },
+});
+
+Deno.test("expect().extend() api test case", () => {
```
kt3k, line 117:
```suggestion
Deno.test("expect.extend() api test case", () => {
```

expect/expect.ts
```diff
@@ -94,7 +98,7 @@ const matchers: Record<MatcherKey, Matcher> = {
   toThrow,
 };
 
-export function expect(value: unknown, customMessage?: string): Expected {
+export const expect = (value: unknown, customMessage?: string): Expected => {
```
kt3k, line 101: We have a style guide about this. We prefer `function` over arrow syntax https://docs.deno.com/runtime/manual/references/contributing/style_guide#top-level-functions-should-not-use-arrow-syntax

kt3k (approved):
Nicely implemented! Thanks!

PR #4414, opened 2024-02-28
fix(path): fix common() results that depend on the order and/or number of input paths

jviide:
This pull request addresses (and adds tests for) cases where [`common()`](https://deno.land/std@0.217.0/path/mod.ts?s=common) returns different values depending on the order and/or number of the input paths, when different results may not be intended.

For example in this case it's probably expected that both cases return an empty string:

```ts
> common(["/", ""], "/")
'/'
> common(["", "/"], "/")
''
```

In the next case the result depends on the number of paths, even if the paths are all equal. The expected result in both cases is likely `'/foo/bar'`:

```ts
> common(["/foo/bar"], "/")
'/foo/'      // missing 'bar'
> common(["/foo/bar", "/foo/bar"], "/")
'/foo/bar/'  // extra slash at the end
```

This pull request adds a variable (`append`) to track whether a separator should be appended to the end of the result string. Early returns are removed to keep this tracking consistent. The performance doesn't seem to degrade in the following tested case:

```ts
Deno.bench("original common()", () => {
  commonOriginal([
    "./deno/std/path/mod.ts",
    "./deno/std/fs/mod.ts",
  ], "/")
});

Deno.bench("modified common()", () => {
  commonModified([
    "./deno/std/path/mod.ts",
    "./deno/std/fs/mod.ts",
  ], "/")
});
```

Results for Deno v1.41.0, MacOS 14.3.1:

```
benchmark              time (avg)        iter/s             (min … max)       p75       p99      p995
----------------------------------------------------------------------- -----------------------------
original common()     225.89 ns/iter   4,426,966.2 (219.21 ns … 255.68 ns) 227.8 ns 246.43 ns 247.09 ns
modified common()     196.07 ns/iter   5,100,328.9  (188.82 ns … 217.8 ns) 196.75 ns 216.05 ns 216.22 ns
```

### DIFF
--- a/path/_common/common.ts
+++ b/path/_common/common.ts
@@ -3,24 +3,24 @@
 
 export function _common(paths: string[], sep: string): string {
   const [first = "", ...remaining] = paths;
-  if (first === "" || remaining.length === 0) {
-    return first.substring(0, first.lastIndexOf(sep) + 1);
-  }
   const parts = first.split(sep);
 
   let endOfPrefix = parts.length;
+  let append = "";
   for (const path of remaining) {
     const compare = path.split(sep);
+    if (compare.length <= endOfPrefix) {
+      endOfPrefix = compare.length;
+      append = "";
+    }
+
     for (let i = 0; i < endOfPrefix; i++) {
       if (compare[i] !== parts[i]) {
         endOfPrefix = i;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
All suggested changes seem reasonable to me. Thanks for working on this!

Issue #4411, opened 2024-02-28
Log FileHandler chokes at messages > 4096


dvv:
**Describe the bug**
With #4021 it is not possible to rely on std logger.

**Steps to Reproduce**
[see here](https://github.com/denoland/deno_std/pull/4021#issuecomment-1968338173)

**Expected behavior**

Proper sinking of (rather) large messages.

**Environment**

- OS: [ArchLinux]
- deno version: 1.40.4
- std version: 0.217.0

kt3k:
cc @syhol Can you take a look?

syhol:
Sure. I'll spend some time on it this evening

syhol:
Resolved by https://github.com/denoland/deno_std/pull/4415 this should be safe to close.

dvv:
Thanks!

PR #4415, opened 2024-02-29
fix(log): don't discard bytes >4096 in `FileHandler`

syhol:
Fixes the bug https://github.com/denoland/deno_std/issues/4411
Bug introduced in https://github.com/denoland/deno_std/pull/4021

TLDR; Attempting to log 4096 or more bytes in a single call to the logger would throw an error and the log would not be written.

Note this PR appears to introduce a new dependency between std/log and std/io. If this is not acceptable then we can find a workaround but the solution will be more verbose.

### DIFF
--- a/log/file_handler.ts
+++ b/log/file_handler.ts
@@ -2,6 +2,7 @@
 import { type LevelName, LogLevels } from "./levels.ts";
 import type { LogRecord } from "./logger.ts";
 import { BaseHandler, type BaseHandlerOptions } from "./base_handler.ts";
+import { writeAllSync } from "../io/write_all.ts";
 
 const PAGE_SIZE = 4096;
 export type LogMode = "a" | "w" | "x";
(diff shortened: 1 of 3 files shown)


syhol:
I'm not sure why canary is failing, I don't think it's related to my change as its failing in ./testing/snapshot_test.ts and I don't see any link between that and the log module.

iuioiua (changes requested):

kt3k:
> Note this PR appears to introduce a new dependency between std/log and std/io. If this is not acceptable then we can find a workaround but the solution will be more verbose.

We recently slightly changed the policy about `std/io` (ref: #4128), and now it's not a deprecated module anymore. So the dependency to `std/io` is just fine.

syhol:
Looks like https://github.com/denoland/deno_std/pull/4416 fixed the canary issues, thanks for that @kt3k 🙏 

@iuioiua's suggestions have been merged and it's ready for another review.

kt3k (approved):
LGTM. Thanks!

iuioiua (approved):

iuioiua:
Thank you, @syhol!

PR #4416, opened 2024-02-29
chore(testing): update snapshot testing expectations

kt3k:
The string patterns for indicating output during test executions have been changed in https://github.com/denoland/deno/pull/22611, and snapshot test cases are affected by that change. This PR updates the expectations of these test cases to pass CI with Deno canary.

This PR also disables CI for v1.x because it can't pass with those versions.


iuioiua (approved):
🆒

PR #4421, opened 2024-03-01
0.218.2

kt3k:
This PR bumps patch version (+2) to publish 0.218.2 with provenance enabled to JSR.


iuioiua (approved):
Why not 0.218.1?

kt3k:
I already published 0.218.1 (https://jsr.io/@std/assert@0.218.1) to JSR by mistake, and we need to skip that version.

The purpose of this patch version is to enable provenance attestation in JSR, which can only be done via CI

PR #4424, opened 2024-03-01
fix: use hyphens for jsr package exports

dsherret:
Uses a consistent separator instead of mixing the two: `@std/data-structures/binary_heap` -> `@std/data-structures/binary-heap`


kt3k (dismissed):
LGTM

ry:
will this create any instability ?

iuioiua:
It is, but we did make a similar change in #4357. We can help by making migration instructions and release notes super-clear.

kt3k:
> will this create any instability ?

This doesn't affect `deno.land/std`. This will affect `jsr:@std` imports.

iuioiua (approved):

kt3k (approved):
LGTM

PR #4429, opened 2024-03-04
chore: use `--trace-leaks` instead of `--trace-ops`

iuioiua:
`--trace-ops` has been deprecated in favour of `--trace-leaks`. See https://github.com/denoland/deno/pull/22598.


kt3k (approved):
LGTM

PR #4430, opened 2024-03-04
chore: run `deno task test` after workspace conversion

iuioiua:
This is a more complete means of testing after the workspaces conversion script has run.

Depends on #4346 being closed.


kt3k (approved):
LGTM

PR #4431, opened 2024-03-04
chore(assert): add missing example and `@module` tags

iuioiua:
IIUC, this will result in a perfect [JSR score](https://jsr.io/@std/assert/score).

### DIFF
--- a/assert/assert.ts
+++ b/assert/assert.ts
@@ -11,6 +11,8 @@ import { AssertionError } from "./assertion_error.ts";
  * assert("hello".includes("ello")); // Doesn't throw
  * assert("hello".includes("world")); // Throws
  * ```
+ *
+ * @module
  */
 export function assert(expr: unknown, msg = ""): asserts expr {
   if (!expr) {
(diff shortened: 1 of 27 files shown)


assert/assert.ts
```diff
@@ -11,6 +11,8 @@ import { AssertionError } from "./assertion_error.ts";
  * assert("hello".includes("ello")); // Doesn't throw
  * assert("hello".includes("world")); // Throws
  * ```
+ *
+ * @module
```
crowlKats, line 15: then assert wont have any docs

iuioiua: image

I was following this suggestion. Am I missing something?

kt3k: Does this pattern work as 'module doc'? When I tested with my tool, it seemed that if a jsdoc recognized as function's doc, then it doesn't work as module doc.

iuioiua:
I now think this PR doesn't provide the fix it should. I'll close this for now.

PR #4432, opened 2024-03-04
chore(async): add missing code snippet tags

iuioiua:
Having these tags enables syntax highlighting in documentation.


kt3k (approved):
LGTM 👍

PR #4433, opened 2024-03-04
chore: exclude test files from being published to JSR

iuioiua:
This change prevents many unnecessary files from being published and reduces the number of warnings from `deno publish --dry-run`.


kt3k (approved):
LGTM. Let's try

timreichen:
`test.ts` without an underscore should also be excluded.
Maybe it might be worth to just exclude all test files with different suffixes as well (`.tsx`, `.js`, `.jsx`) to avoid possible future problems?

Issue #4391, opened 2024-02-25
question(log): remove `setup()` and `destroy()` from `BaseHandler`?


timreichen:
**Is your feature request related to a problem? Please describe.**

`BaseHandler.setup()` and `BaseHandler.destroy()` seem kinda odd as methods for `BaseHandler`, especially their names.
They do not have a purpose in the `BaseHandler` itself, except `destroy()` being called on `[Symbol.dispose]`.

**Describe the solution you'd like**

I think `BaseHandler` should remove `setup()` and `destroy()`, `FileHandler` should have methods called `open()` and `close()` instead of `setup()` and `destroy()` as that describes better what is happening inside the method.
`[Symbol.dispose]` should be implemented in `FileHandler` to call `close()`. `RotatingFileHandler` would inherit from `FileHandler` as it does now and could be overridden as needed.

**Describe alternatives you've considered**

Leave as is.

iuioiua:
+1

kt3k:
-1 to this. Doesn't sound like appropriate reasons for doing breaking changes.

> BaseHandler.setup() and BaseHandler.destroy() seem kinda odd as methods for BaseHandler, especially their names.

We shouldn't do breaking changes based on just preferences/tastes.

timreichen:
> -1 to this. Doesn't sound like appropriate reasons for doing breaking changes.
> 
> > BaseHandler.setup() and BaseHandler.destroy() seem kinda odd as methods for BaseHandler, especially their names.
> 
> We shouldn't do breaking changes based on just preferences/tastes.

It is not only about taste, web APIs give some guidance for naming methods and properties. We should orient ourselves towards that. For example `dispose()` would be better than `destroy()` because `Symbol.dispose` exists.

But in this particular case `BaseHandler.setup()` and `BaseHandler.destroy()` are empty methods and callbacks (similar to [customElements](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements) `connectedCallback` and `disconnectedCallback()`) for  subclasses like `FileHandler`.
These get called in `setup()`. But `FileHandler.open()` could be called manually or inside `FileHandler.handle()`, if the file is not open yet. Similarly, `FileHandler.close()` can be called manually and is called on the `unload` event.

iuioiua:
To be clear, I'm +1 for removing `BaseHandler.setup()` and `BaseHandler. destroy()`, as they're empty methods solely there to be overwritten. IMO that doesn't justify their existence.

[...]

PR #4435, opened 2024-03-04
refactor(log): Deprecate `BaseHandler.setup()` and `BaseHandler.destroy()`

timreichen:
Ref: https://github.com/denoland/deno_std/issues/4391

**Changes**

- Deprecates `BaseHandler.setup()`
- Deprecates `BaseHandler.destroy()`
- Introduces `FileHandler.open()` to replace `FileHandler.setup()`
- Introduces `FileHandler.close()`to replace `FileHandler.destroy()`
- Adds comments how to make `setup()` function work without `BaseHandler.setup()` and `BaseHandler.destroy()`

### DIFF
--- a/log/file_handler.ts
+++ b/log/file_handler.ts
@@ -38,9 +38,12 @@ export class FileHandler extends BaseHandler {
   protected _mode: LogMode;
   protected _openOptions: Deno.OpenOptions;
   protected _encoder: TextEncoder = new TextEncoder();
-  #unloadCallback = (() => {
-    this.destroy();
-  }).bind(this);
+  #isOpen: boolean;
+  #unloadCallback = (() => this.close()).bind(this);
+
+  [Symbol.dispose]() {
+    this.close();
+  }
 
   constructor(levelName: LevelName, options: FileHandlerOptions) {
     super(levelName, options);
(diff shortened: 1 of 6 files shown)


log/file_handler.ts
```diff
@@ -100,11 +114,4 @@ export class FileHandler extends BaseHandler {
   #resetBuffer() {
     this._pointer = 0;
   }
-
-  override destroy() {
```
iuioiua, line 104: This method should have a deprecation notice. Otherwise, this is a breaking change. We can also just call `.close()` within it now.

log/file_handler.ts
```diff
@@ -54,13 +57,24 @@ export class FileHandler extends BaseHandler {
       truncate: this._mode !== "a",
       write: true,
     };
+    this.#isOpen = false;
   }
 
-  override setup() {
+  open() {
+    if (this.#isOpen) return;
```
iuioiua, line 64: This isn't relevant to this PR. I might argue it's unnecessary, as we have the open/close status of `this._file` we can use instead.

log/file_handler_test.ts
iuioiua, line 1: Can you please revert these changes? This ensures the legacy and possible new methods all work.

log/rotating_file_handler_test.ts
iuioiua, line 1: Ditto. Can you please revert these changes?

iuioiua (changes requested):
I have a few suggestions. Also, please note that this deprecation is yet to be agreed upon. So it's possible we may not go ahead with it.

iuioiua (changes requested):
What if we re-orient this PR to just removing `BaseHandler.setup()` and `BaseHandler.destroy()`? That seems like a more agreeable design decision.

iuioiua:
@timreichen, I'm going to close this PR for now. I think we need to discuss and consider this more before proceeding. Either way, thanks for trying to push this forward.

PR #4441, opened 2024-03-05
chore: ignore more test files from being published

iuioiua:
This change:
1. Includes `**/*test.{ts,js}` files to be published, as there's https://github.com/denoland/deno_std/blob/main/front_matter/test.ts that is part of the non-testing, public API.
2. Adds further test files, like `io/_test_common.ts` to be excluded from publishing.

### DIFF
--- a/_tools/convert_to_workspace.ts
+++ b/_tools/convert_to_workspace.ts
@@ -233,8 +233,8 @@ for (const pkg of packages) {
     exclude: [
       "**/test.ts",
       "**/test.js",
-      "**/*_test.ts",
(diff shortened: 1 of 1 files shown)


lucacasonato:
Why did we do this? Publishing test files to JSR is not a bad thing. Also this broke publishing of `std/expect`: `_custom_equality_tester.ts` matches the ignore regexp.

kt3k:
Thanks for catching!

> Publishing test files to JSR is not a bad thing.

Ok. Let's include test files in published assets.

PR #4445, opened 2024-03-06
refactor(uuid): prepare for `noUncheckedIndexedAccess`

syhol:
Handling all noUncheckedIndexedAccess issues for the uuid module, tracked in https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/uuid/mod.ts
+++ b/uuid/mod.ts
@@ -76,5 +76,5 @@ export function version(uuid: string): number {
     throw new TypeError("Invalid UUID");
   }
 
-  return parseInt(uuid[14], 16);
+  return parseInt(uuid[14]!, 16);
 }
--- a/uuid/v1.ts
+++ b/uuid/v1.ts
@@ -136,6 +136,10 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
+  if (node.length < 6) {
+    throw new Error("Uuid node needs to have at least 6 numbers");
+  }
+
   _lastMSecs = msecs;
   _lastNSecs = nsecs;
   _clockseq = clockseq;
@@ -163,7 +167,7 @@ export function generate(
   b[i++] = clockseq & 0xff;
 
   for (let n = 0; n < 6; ++n) {
-    b[i + n] = node[n];
+    b[i + n] = node[n]!;
   }
 
   return buf ?? bytesToUuid(b);
--- a/uuid/v1.ts
+++ b/uuid/v1.ts
@@ -137,7 +137,9 @@ export function generate(
   }
 
   if (node.length !== 6) {
-    throw new Error("Can't create uuid. The node must be an array of 6 bytes");
+    throw new Error(
+      "Cannot create UUID. The node option must be an array of 6 bytes",
+    );
   }
 
   _lastMSecs = msecs;
--- a/uuid/v1.ts
+++ b/uuid/v1.ts
@@ -137,7 +137,7 @@ export function generate(
   }
 
   if (node.length < 6) {
-    throw new Error("Uuid node needs to have at least 6 numbers");
+    throw new Error("Can't create a uuid with a node of less than 6 numbers");
   }
 
   _lastMSecs = msecs;
--- a/uuid/v1.ts
+++ b/uuid/v1.ts
@@ -136,8 +136,8 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
-  if (node.length < 6) {
-    throw new Error("Can't create a uuid with a node of less than 6 numbers");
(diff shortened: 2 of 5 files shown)


uuid/mod.ts
```diff
@@ -76,5 +76,5 @@ export function version(uuid: string): number {
     throw new TypeError("Invalid UUID");
   }
 
-  return parseInt(uuid[14], 16);
+  return parseInt(uuid[14]!, 16);
```
syhol, line 79: The uuid validate function makes sure uuid has at least 30 chars.

uuid/v3.ts
```diff
@@ -54,8 +54,8 @@ export async function generate(
   const buffer = await crypto.subtle.digest("MD5", toHash);
   const bytes = new Uint8Array(buffer);
 
-  bytes[6] = (bytes[6] & 0x0f) | 0x30;
-  bytes[8] = (bytes[8] & 0x3f) | 0x80;
+  bytes[6] = (bytes[6]! & 0x0f) | 0x30;
+  bytes[8] = (bytes[8]! & 0x3f) | 0x80;
```
syhol, line 58: line 54 crypto.subtle.digest always returns an ArrayBuffer of length 20, so bytes 6 and 8 will be fine.

uuid/v5.ts
```diff
@@ -53,8 +53,8 @@ export async function generate(
   const buffer = await crypto.subtle.digest("sha-1", toHash);
   const bytes = new Uint8Array(buffer);
 
-  bytes[6] = (bytes[6] & 0x0f) | 0x50;
-  bytes[8] = (bytes[8] & 0x3f) | 0x80;
+  bytes[6] = (bytes[6]! & 0x0f) | 0x50;
+  bytes[8] = (bytes[8]! & 0x3f) | 0x80;
```
syhol, line 57: line 53 crypto.subtle.digest always returns an ArrayBuffer of length 20, so bytes 6 and 8 will be fine.

uuid/v1.ts
```diff
@@ -163,7 +167,7 @@ export function generate(
   b[i++] = clockseq & 0xff;
 
   for (let n = 0; n < 6; ++n) {
-    b[i + n] = node[n];
+    b[i + n] = node[n]!;
```
syhol, line 170: Since we now validate the node length above we can be sure 0-5 will always be a valid value

uuid/v1.ts
```diff
@@ -136,6 +136,10 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
+  if (node.length < 6) {
+    throw new Error("Uuid node needs to have at least 6 numbers");
+  }
```
syhol, line 141: `node` can be passed in as a parameter, but if empty it will be set to an array of 6 numbers. All the current tests call this function with ether no `node` or a `node` of 6 numbers. I think its safe to say that calling this function with a `node` of less than 6 numbers is undefined behaviour and we should be safe to throw.

syhol: https://github.com/denoland/deno_std/blob/f0401ad2a6535d7b47147ffd9322edd603bc780b/uuid/v1.ts#L34-L39

uuid/v1.ts
```diff
@@ -136,6 +136,10 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
+  if (node.length < 6) {
+    throw new Error("Uuid node needs to have at least 6 numbers");
```
syhol, line 140: Adding a test now

uuid/v1.ts
```diff
@@ -136,6 +136,10 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
+  if (node.length < 6) {
+    throw new Error("Can't create a uuid with a node of less than 6 numbers");
+  }
```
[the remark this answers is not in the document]
syhol: Done

uuid/v1.ts
```diff
@@ -136,6 +136,10 @@ export function generate(
     throw new Error("Can't create more than 10M uuids/sec");
   }
 
+  if (node.length !== 6) {
+    throw new Error("Can't create uuid. The node must be an array of 6 bytes");
```
iuioiua, line 140:
```suggestion
    throw new Error("Cannot create UUID. The node option must be an array of 6 bytes");
```

iuioiua (changes requested):
Just a couple more nits. Can you please tweak the test with the suggested output?

uuid/v1.ts
syhol: Updated. Ready for another round.

iuioiua (approved):
LGTM! Thank you.

Issue #4365, opened 2024-02-21
semver: deprecate `rangeMax` and `rangeMin`


kt3k:
`rangeMax` and `rangeMin` were introduced in #3385, but these 2 APIs seem having critical design flaws. I suggest we should deprecate these 2 APIs.

1. In many cases, `rangeMax` returns invalid semvers.

`rangeMax(parseRange("<1.0.0"))` returns 0.∞.∞ (0.Infinity.Infinity).
`rangeMax(parseRange("<1.1.0"))` returns 1.0.∞ (1.0.Infinity).
`rangeMax(parseRange("<0.0.0"))` returns -∞.∞.∞ (-Infinity.Infinity.Infinity).

These values are non semvers. Exposing them are very confusing.

2. `rangeMax` ignores prerelease

`rangeMax(parseRange("<1.2.3-a"))` returns 1.2.2 ignoring prerelease version.

Because of this behavior, the example below returns `false`, which seems a bug, but there's no clear way to fix this.

```js
greaterThan(rangeMax(parseRange("<1.2.3-beta.20")), rangeMin(parseRange(">1.2.3-beta.10")))
```

3. rangeMin ignores prerelease version only when the given range doesn't include prerelease.

`rangeMin(parseRange(">1.2.3"))` return 1.2.4, but actually 1.2.4-0 is smaller than that (and bigger than 1.2.3). On the other hand, `rangeMin(parseRange(">1.2.3-a"))` returns the correct value 1.2.3-a.0, but the switching of the algorithm depending on the input value seems confusing and arbitrary.

---
Note: There's no concern about the compatibility to [`npm:semver`](https://github.com/npm/node-semver) because these APIs are invented by us in #3385.

timreichen:
I agree, this seems very confusing.

Just a thought about 1. for the record:
Maybe this could  be fixed by using something like [MAX_SAFE_INTEGER ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) instead of Infinity. But this seems hacky.

timreichen:
In we deprecate these two functions, there is no way to check anymore if a version is bigger/less than a range, because https://github.com/denoland/deno_std/issues/4273 relies on these two as an alternative. Any thoughts?

kt3k:
[...]

PR #4450, opened 2024-03-07
[unstable] BREAKING(semver): remove `gtr()` and `ltr()`

kt3k:
Scheduled removal of semver/gtr and semver/ltr


iuioiua:
Didn't we suggest `rangeMin()` and `rangeMax()` were buggy in #4365?

kt3k:
Yes, they are buggy, but these 2 functions are also buggy in the same way. I think there's no reason to keep these.

iuioiua (approved):

PR #4452, opened 2024-03-07
chore: move typos action config

niklasmtj:
handle the `typos.toml` similar to `dependabot.yml` and `labeler.yml` since these are configuration files and not GH Action workflow definitions.

What do you think about keeping the configs in the `.github` folder and not the workflows' folder?


kt3k (approved):
Thanks for the suggestion. This makes more sense to me.

LGTM

PR #4457, opened 2024-03-09
refactor(yaml): prepare for `noUncheckedIndexedAccess`

gabelluardo:
Ref: https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/yaml/_dumper/dumper.ts
+++ b/yaml/_dumper/dumper.ts
@@ -135,7 +135,7 @@ function testImplicitResolving(state: DumperState, str: string): boolean {
     index < length;
     index += 1
   ) {
-    type = state.implicitTypes[index];
+    type = state.implicitTypes[index]!;
 
     if (type.resolve(str)) {
       return true;
@@ -379,7 +379,7 @@ function foldString(string: string, width: number): string {
   // tslint:disable-next-line:no-conditional-assignment
   while ((match = lineRe.exec(string))) {
     const prefix = match[1],
-      line = match[2];
+      line = match[2]!;
     moreIndented = line[0] === " ";
     result += prefix +
       (!prevMoreIndented && !moreIndented && line !== "" ? "\n" : "") +
@@ -582,7 +582,7 @@ function writeFlowMapping(
 
     if (index !== 0) pairBuffer += ", ";
 
-    objectKey = objectKeyList[index];
+    objectKey = objectKeyList[index]!;
     objectValue = object[objectKey];
 
     if (!writeNode(state, level, objectKey, false, false)) {
@@ -646,7 +646,7 @@ function writeBlockMapping(
       pairBuffer += generateNextLine(state, level);
     }
 
-    objectKey = objectKeyList[index];
+    objectKey = objectKeyList[index]!;
     objectValue = object[objectKey];
 
     if (!writeNode(state, level + 1, objectKey, true, true, true)) {
@@ -701,7 +701,7 @@ function detectType(
   let style: StyleVariant;
   let _result: string;
   for (let index = 0, length = typeList.length; index < length; index += 1) {
-    type = typeList[index];
+    type = typeList[index]!;
 
     if (
       (type.instanceOf || type.predicate) &&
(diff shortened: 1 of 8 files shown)


yaml/_dumper/dumper.ts
```diff
@@ -379,7 +379,7 @@ function foldString(string: string, width: number): string {
   // tslint:disable-next-line:no-conditional-assignment
   while ((match = lineRe.exec(string))) {
     const prefix = match[1],
-      line = match[2];
+      line = match[2]!;
```
iuioiua, line 382: Can you please justify this change?

yaml/_dumper/dumper.ts
```diff
@@ -582,7 +582,7 @@ function writeFlowMapping(
 
     if (index !== 0) pairBuffer += ", ";
 
-    objectKey = objectKeyList[index];
+    objectKey = objectKeyList[index]!;
```
iuioiua, line 585: Can we use `const [objectKey, objectValue] of objectKeyList` as the for-loop condition?

yaml/_dumper/dumper.ts
```diff
@@ -646,7 +646,7 @@ function writeBlockMapping(
       pairBuffer += generateNextLine(state, level);
     }
 
-    objectKey = objectKeyList[index];
+    objectKey = objectKeyList[index]!;
```
iuioiua, line 649: Ditto.

yaml/_dumper/dumper.ts
```diff
@@ -701,7 +701,7 @@ function detectType(
   let style: StyleVariant;
   let _result: string;
   for (let index = 0, length = typeList.length; index < length; index += 1) {
-    type = typeList[index];
+    type = typeList[index]!;
```
iuioiua, line 704: Ditto. I think we can instead change the for-loop condition.

yaml/_dumper/dumper_state.ts
```diff
@@ -21,12 +21,12 @@ function compileStyleMap(
   const keys = Object.keys(map);
   let tag: string, style: StyleVariant;
   for (let index = 0, length = keys.length; index < length; index += 1) {
-    tag = keys[index];
+    tag = keys[index]!;
```
iuioiua, line 24: Ditto. Can we tweak the for-loop condition to be `const tag of Object.keys(map)`?

yaml/_type/timestamp.ts
```diff
@@ -68,7 +68,7 @@ function constructYamlTimestamp(data: string): Date {
 
   let delta = null;
   if (match[9]) {
-    const tzHour = +match[10];
+    const tzHour = +match[10]!;
```
iuioiua, line 71: Can you please justify this non-null assertion? The other assertions look fine.

iuioiua (changes requested):
Can you please justify non-trivial non-null assertions? No comment is needed for trivial ones.

Also, I think we can omit some non-null assertions within loops by instead changing the loop conditions.

Sidenote: looking through this module makes me realise how much room for improvement there is.

yaml/_dumper/dumper.ts
gabelluardo: Seems this can't be done because `objectKeyList` contains only the keys of `object`, so I think we can do `const [index, objectKey] of objectKeyList.entries()` and leave the `objectValue = object[objectKey]` inside the loop. Furthermore, `index` is needed in `(index !== 0) pairBuffer += ", "`

yaml/_type/timestamp.ts
gabelluardo: I thought that the check of `match[9]` was enough to establish that a time zone is present, but maybe an explicit check on `match[10]` is better. I will change this.

yaml/_dumper/dumper.ts
gabelluardo: ref https://github.com/denoland/deno_std/pull/4457#discussion_r1519025322

yaml/_dumper/dumper.ts
gabelluardo: I will change this with `match[2] || ""`, the check for line make sense inside the `foldLine` function

gabelluardo:
I refactored a bit the for-loops to improve readability, removing some `let` variables that can be `const`

yaml/_dumper/dumper.ts
```diff
@@ -845,14 +814,8 @@ function inspectNode(
...
-
-        for (
-          let idx = 0, length = objectKeyList.length;
-          idx < length;
-          idx += 1
-        ) {
-          inspectNode(object[objectKeyList[idx]], objects, duplicatesIndexes);
+        for (const objectKey of Object.keys(object)) {
+          inspectNode(object[objectKey], objects, duplicatesIndexes);
         }
```
iuioiua, line 819:
```suggestion
        for (const object of Object.values(object)) {
          inspectNode(object, objects, duplicatesIndexes);
        }
```

iuioiua (changes requested):

iuioiua (approved):
Ignore my previous request for changes. LGTM! Thank you.

PR #4459, opened 2024-03-10
fix: revert #992

### LINKED ISSUE #4426: bug(http): missing docs and `setCookie()` allows `Max-age` to be 0
**Description**

Regarding this line: https://github.com/denoland/deno_std/blob/45f7a6b46018e9f4d330943918299c68ae80f1bb/http/cookie.ts#L91

While working on the cookie implementation in undici/node, I realized that the max-age implementation is slightly off.

This is what I wrote to the corresponding code piece in the undici PR

> Imho the spec rfc 6265 is not correctly implemented.
> 
> Either only positive integers excluding 0 are allowed (which the current implementation does not match) for MaxAge or all integer values are allowed (this PR)
> 
> I think the problem arises because imho https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1 talks about how the data in the useragent is stored while https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2 talks about the data which the useragent receives and how it should be processed.
> 
> So a server can send negative values for max-age, while a client has to drop all cookies with negative values immediatly. If the maxage is set to 0, then it would actually mean to drop the cookie according to the rfc, but it seems most useragents keep the cookie till the end of the session.
> 
> Anyhow, this means that we must allow negative integers for validateMaxAge.

https://github.com/nodejs/undici/pull/2888

**Expected behavior**

Negative numbers should be allowed

Uzlopak:
fix #4426

@iuioiua


kt3k:
Closing as there's no consensus on #4426

PR #4470, opened 2024-03-11
refactor(io): prepare for `noUncheckedIndexedAccess`

gabelluardo:
ref https://github.com/denoland/deno_std/issues/4040

### DIFF
--- a/io/buf_reader.ts
+++ b/io/buf_reader.ts
@@ -207,7 +207,7 @@ export class BufReader implements Reader {
       if (this.#eof) return null;
       await this.#fill(); // buffer is empty.
     }
-    const c = this.#buf[this.#r];
+    const c = this.#buf[this.#r]!;
     this.#r++;
     // this.lastByte = c;
     return c;
--- a/io/read_delim.ts
+++ b/io/read_delim.ts
@@ -19,7 +19,7 @@ function createLPS(pat: Uint8Array): Uint8Array {
       lps[i] = 0;
       i++;
     } else {
-      prefixEnd = lps[prefixEnd - 1];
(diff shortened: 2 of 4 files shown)


io/buf_reader.ts
```diff
@@ -207,7 +207,7 @@ export class BufReader implements Reader {
       if (this.#eof) return null;
       await this.#fill(); // buffer is empty.
     }
-    const c = this.#buf[this.#r];
+    const c = this.#buf[this.#r]!;
```
gabelluardo, line 210: The while-loop wait until `this.#buf` is filled, so this can't be null

io/read_delim.ts
```diff
@@ -19,7 +19,7 @@ function createLPS(pat: Uint8Array): Uint8Array {
       lps[i] = 0;
       i++;
     } else {
-      prefixEnd = lps[prefixEnd - 1];
+      prefixEnd = lps[prefixEnd - 1]!;
```
gabelluardo, line 22: `prefixEnd` is incremented inside the loop and `lsp` size is `pat.length` so this is non-null

io/string_writer_test.ts
```diff
@@ -33,6 +33,6 @@ Deno.test("ioStringWriterIsolationTest", async function () {
     const written = await w.write(c);
     assertEquals(written, 1);
   }
-  srcChunks[0][0] = 88;
+  // srcChunks[0][0] = 88;
```
gabelluardo, line 36: I think this can be deleted because the values in `srcChunks` aren't checked in the assertion

iuioiua (approved):
LGTM! Thank you.

Issue #4468, opened 2024-03-11
suggestion: enable `no-sync-fn-in-async-fn` linting rule


iuioiua:
**Is your feature request related to a problem? Please describe.**

"Using sync functions like Deno.readTextFileSync blocks the deno event loop so it's not recommended to use it inside of an async function, because it stops progress of all other async tasks." - [deno_lint documentation](https://lint.deno.land/rules/no-sync-fn-in-async-fn)

**Describe the solution you'd like**

1. Enable the [`no-sync-fn-in-async-fn`](https://lint.deno.land/rules/no-sync-fn-in-async-fn) lint rule by adding it to `lint.rules.include` in `deno.json`.
2. Find and fix linting violations by running `deno lint`.

PR #4471, opened 2024-03-11
chore: enable `no-sync-fn-in-async-fn` lint rule

gabelluardo:
ref: #4468

### DIFF
--- a/http/server.ts
+++ b/http/server.ts
@@ -773,8 +773,10 @@ export async function serveTls(
     once: true,
   });
 
-  const key = options.key || await Deno.readTextFile(options.keyFile!);
-  const cert = options.cert || await Deno.readTextFile(options.certFile!);
+  // deno-lint-ignore no-sync-fn-in-async-fn
+  const key = options.key || Deno.readTextFileSync(options.keyFile!);
+  // deno-lint-ignore no-sync-fn-in-async-fn
+  const cert = options.cert || Deno.readTextFileSync(options.certFile!);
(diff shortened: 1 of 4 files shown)


http/server.ts
```diff
@@ -773,8 +773,10 @@ export async function serveTls(
     once: true,
   });
 
-  const key = options.key || await Deno.readTextFile(options.keyFile!);
-  const cert = options.cert || await Deno.readTextFile(options.certFile!);
+  // deno-lint-ignore no-sync-fn-in-async-fn
```
gabelluardo, line 776: I think this is ok since the `serveTls` function is deprecated and will be removed soon

iuioiua: Usually, we shouldn't bother. But this is a small/simple change. Can you please use `await Deno.readTextFile()` here?

gabelluardo: I tried but it seems to break the test https://github.com/denoland/deno_std/actions/runs/8232529366/job/22510162053#step:4:5048

iuioiua: Oh, that's interesting. @bartlomieju, do you have any idea why this might be? Is this expected behaviour? This specific instance doesn't matter much; rather, the behaviour surprises me.

Either way, @gabelluardo, let's ignore 👍🏾

iuioiua:
I'll re-run CI a few times to ensure there aren't any issues.

iuioiua (approved):

PR #4472, opened 2024-03-11
feat(assert): add assertThrowsAsync()

halvardssm:
Adds an async alternative to assertThrows that can handle async functions


kt3k:
Thanks for the suggestion, but we already have [assertRejects](https://github.com/denoland/deno_std/blob/d63e97c735b6f3285099ceea73c5588b8dc1b6d8/assert/assert_rejects.ts), which does mostly the same thing

halvardssm:
You are very right, i completely missed this one. Thanks!

PR #4476, opened 2024-03-12
docs(crypto): add module example

iuioiua:
Towards a 100% JSR score.

### DIFF
--- a/crypto/mod.ts
+++ b/crypto/mod.ts
@@ -7,6 +7,16 @@
  * supporting additional encryption APIs, but also delegating to the built-in
  * APIs when possible.
  *
+ * ```ts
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

crypto/mod.ts
```diff
@@ -7,6 +7,16 @@
...
  * APIs when possible.
  *
+ * ```ts
+ * import { crypto } from "https://deno.land/std@$STD_VERSION/crypto/crypto.ts";
+ *
+ * const message = "Hello, Deno!";
+ * const encoder = new TextEncoder();
+ * const data = encoder.encode(message);
+ *
+ * await crypto.subtle.digest("BLAKE3", data);
```
kt3k, line 17: 👍

PR #4485, opened 2024-03-14
Revert "refactor(cli): cleanup `parse_args.ts` (#4189)"

### LINKED ISSUE #4481: parseArgs duplicates collected + aliased arguments
Version: Deno 1.41.2

The `parseArgs` function from https://deno.land/std/cli/parse_args.ts seems to duplicate `collect` arguments when an alias is part of the definition.

This affects the [file_server](https://docs.deno.com/runtime/tutorials/file_server) with the `--header` option; this results in HTTP headers being written twice.

Minimal test case to repro:

```typescript
import { parseArgs } from "https://deno.land/std/cli/parse_args.ts";
import { assertEquals } from "https://deno.land/std/assert/mod.ts";

Deno.test("collect with alias", () => {
  const args = ["--header", "abc", "--header", "def"];
  const parsed = parseArgs(args, {
    collect: ["header"],
    alias: {
      H: "header",
    },
  });
  assertEquals(parsed.header, ["abc", "def"])
})
```

Output:

```
$ deno test
running 1 test from ./duplicatedArgs.test.ts
collect with alias ... FAILED (4ms)

 ERRORS 

collect with alias => ./duplicatedArgs.test.ts:4:6


    [Diff] Actual / Expected


    [
      "abc",
-     "abc",
-     "def",
      "def",
    ]
```

Removing the alias makes things work as expected.

kt3k:
This reverts commit 8c022c4f6488accaca1f0ccea93d982ba0c503dc.

fixes #4481

### DIFF
--- a/cli/parse_args.ts
+++ b/cli/parse_args.ts
@@ -16,7 +16,7 @@
  *
  * @module
  */
-import { assert } from "../assert/assert.ts";
+import { assertExists } from "../assert/assert_exists.ts";
 
 /** Combines recursively all intersection types and returns a new single type. */
 type Id<TRecord> = TRecord extends Record<string, unknown>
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
@timreichen, please feel free to reland your cleanup changes along with a passing test for the issue outlined in #4481.

timreichen:
I think I found the bug. The aliasMap Set for key contains key itself. On that example header contains header as an alias.

Map(2) {
  "H" => Set(1) { "header" },
  "header" => Set(2) { "H", "header" }
}
This is fixed by filtering out the key.

before:

const set = new Set([key, ...aliases]);
aliases.forEach((alias) => aliasMap.set(alias, set));
after:

aliases.forEach((alias) => aliasMap.set(alias, new Set([key, ...aliases.filter((it) => it !== alias)])));

PR #4489, opened 2024-03-14
docs(path): add more details to `ParsedPath` and other examples

ooker777:


path/_interface.ts
```diff
@@ -3,14 +3,22 @@
 
 /**
  * A parsed path object generated by path.parse() or consumed by path.format().
+ * parse("c:\\path\\dir\\index.html")
```
kt3k, line 6: Can you use code block and `@example` directive here?

kt3k (commented):
Thanks for the PR. Left one comment

path/_interface.ts
ooker777: Thanks for the feedback. Is there any guideline or convention about this? The contributing guideline doesn't talk about this

https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md

kt3k (approved):
LGTM

iuioiua (approved):
I've made a few formatting nits to make it a little more consistent with the rest of the codebase. Either way, great work! Thank you for your contribution. LGTM.

PR #4495, opened 2024-03-16
docs: account for classes in test guideline

iuioiua:
So contributors are clear on how to write test names for classes.


kt3k (approved):
LGTM

PR #4496, opened 2024-03-16
docs(datetime): add module doc

iuioiua:
Towards a 100% JSR score.

### DIFF
--- a/datetime/mod.ts
+++ b/datetime/mod.ts
@@ -6,7 +6,7 @@
  *
  * This module is browser compatible.
  *
- * ```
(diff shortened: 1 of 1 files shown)


datetime/mod.ts
```diff
@@ -6,6 +6,18 @@
  *
  * This module is browser compatible.
  *
+ * ```
```
kt3k, line 9: Maybe better to add the language tag?

```suggestion
 * ```ts
```

kt3k (approved):
LGTM with nits

PR #4497, opened 2024-03-16
docs(encoding): polish documentation

iuioiua:
This file allow us to provide module docs and an entrypoint for the off-chance that one does want utilities for multiple formats.

Towards a 100% JSR score.


kt3k:
Now `encode` and `decode` methods exported from `@std/encoding` is for varint encoding. That feels a bit misleading to me

iuioiua:
This is true. However, this is more a result of `encode()` and `decode()` being inconsistent with other public function names within `std/encoding`. To be consistent, they should instead be `encodeVarint()` and `decodeVarint()`.

kt3k (approved):
LGTM 👍

PR #4499, opened 2024-03-16
refactor(front_matter): cleanup `_formats.ts`

timreichen:
**Changes**
- renames constants in `_formats.ts`, uses conventional name format.
- removes `Array` destructuring.
- renames `createRegExp()` to `createRegExps()` as it returns multiple regexps.
- removes `createRegExps()` spread arguments in favour of an array.


kt3k (approved):
LGTM

Issue #4299, opened 2024-02-09
Review auto-generated documentation and fill-in gaps


rbetts:
We want to review the auto-generated documentation for the standard library - correct anything that's confusing and complete anything that's missing. We'd like to complete this work by end of Feb. 2024.

iuioiua:
I'm closing this in favor of #3764, which tracks documentation oriented toward stabilization.

PR #4505, opened 2024-03-19
docs(front_matter): cleanup and improve docs

iuioiua:
Towards a 100% score, #4299, #3764 and #3878.


front_matter/mod.ts
```diff
@@ -133,24 +78,46 @@
...
- *   "is": "JSON"
- * }
+ * = yaml =
+ * as: yaml
+ * = yaml =
  * ```
  *
  * @module
  */
+import { extract as extractJson } from "./json.ts";
```
iuioiua, line 118: These are added so symbol linking on both JSR and /x work as expected.

kt3k (approved):
LGTM. Nice!

PR #4510, opened 2024-03-22
refactor(crypto): removed unused code from Wasm build

jeremyBanks:
The `std/crypto/_wasm` build currently exports several functions that are no longer called anywhere and are just bloat. This PR removes them and their supporting code. This reduces the size of `deno_std_wasm_crypto.generated.mjs` from 237,510B to 194,324B on disk (18% reduction) and from 88,596 B to 77,230B after `gzip --best` (13% reduction).


iuioiua (approved):
Very nice cleanup! LGTM! Thank you.

Issue #4511, opened 2024-03-22
Bug: expect comparing `RegExp` and `Set` is wrong


marvinhagemeister:
**Describe the bug**

The `expect` function compares class instances incorrectly.

**Steps to Reproduce**

```ts
import { expect } from "jsr:@std/expect";

Deno.test("expect bug", () => {
  // This should throw, but passes
  expect(/foo/).toEqual(new Set());
});
```

**Expected behavior**

The expect should throw because a `RegExp` instance isn't the same as a `Set`

**Environment**

- OS: macOS 14.2.1
- deno version: 1.41.3
- std version: 0.220.1

marvinhagemeister:
This works as expected in `jest` and with our own `assertEquals`. It seems like the custom testers passed inside `toEqual` are letting this through.

PR #4512, opened 2024-03-22
fix(expect/assert): missing constructor equality check

marvinhagemeister:
We didn't do any `constructor` checks when not in strict mode which is wrong.

Fixes https://github.com/denoland/deno_std/issues/4511

### DIFF
--- a/expect/_equal.ts
+++ b/expect/_equal.ts
@@ -73,11 +73,14 @@ export function equal(c: unknown, d: unknown, options?: EqualOptions): boolean {
     if (typeof a === "number" && typeof b === "number") {
       return Number.isNaN(a) && Number.isNaN(b) || a === b;
     }
+    if (a === null || b === null) {
+      return a === b;
+    }
     if (Object.is(a, b)) {
       return true;
     }
     if (a && typeof a === "object" && b && typeof b === "object") {
-      if (strictCheck && a && b && !constructorsEqual(a, b)) {
+      if (a && b && !constructorsEqual(a, b)) {
         return false;
       }
       if (a instanceof WeakMap || b instanceof WeakMap) {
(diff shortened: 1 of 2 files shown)


expect/_equal.ts
```diff
@@ -73,11 +73,14 @@ export function equal(c: unknown, d: unknown, options?: EqualOptions): boolean {
     if (typeof a === "number" && typeof b === "number") {
       return Number.isNaN(a) && Number.isNaN(b) || a === b;
     }
+    if (a === null || b === null) {
```
marvinhagemeister, line 76: Added this as a fast path. It allows us to bail out early.

archive/untar.ts
```diff
@@ -86,7 +86,6 @@ function parseHeader(buffer: Uint8Array): TarHeader {
 }
 
 /** Tar entry */
-// deno-lint-ignore no-empty-interface
```
marvinhagemeister, line 89: Canary complains about this, but stable wants this warning. Since our CI lints only with the canary release I had to remove this to get it green.

expect/_equal.ts
```diff
@@ -73,6 +73,13 @@ export function equal(c: unknown, d: unknown, options?: EqualOptions): boolean {
...
+      return a === b;
+    }
+    const className = Object.prototype.toString.call(a);
+    if (className != Object.prototype.toString.call(b)) {
```
iuioiua, line 80: Nit: should the inequality be stricter?

iuioiua (approved):
One nit. Otherwise, LGTM! Thank you.

expect/_equal.ts
marvinhagemeister: Good point, yeah that should've been a strict equality check.

PR #4516, opened 2024-03-24
chore(crypto): switch Wasm crate's sha-1 dependency to sha1 to reflect rename by maintainer

jeremyBanks:
The SHA-1 implementation from @RustCrypto used by our Wasm bundle used to be published as `sha-1`, but is now published as `sha1` (see https://crates.io/crates/sha_1 or https://github.com/RustCrypto/hashes/issues/438 for confirmation). This dependency was changed in denoland/deno at https://github.com/denoland/deno/pull/16327, but the change was never made in this repository. The old package is no longer receiving updates, so we're now a few patches behind.

I doubt this is consequential in practice, but we might as well switch to the correct crate, so this PR does so and updates to the latest version. There are no observable changes in performance or size.


iuioiua (approved):
Nice catch! Thank you.

PR #4518, opened 2024-03-25
docs(http): fix default algorithm for etag in docs

kt3k:
The default algorithm for etag is not FNV32A, but SHA-256. This PR fixes the doc for it.


iuioiua (approved):
Nice catch

PR #4524, opened 2024-03-26
feat(html): add `normalize` function for HTML entities (#4523)

### LINKED ISSUE #4523: Suggestion: Normalize HTML/XML entities

Often, when consuming HTML or XML files from external sources, it's desirable to normalize the entities. For example, I'm interacting with an API that produces XML where `encoding='UTF-8'` is correctly specified, but non-ASCII characters are unnecessarily encoded as numbered entities. This makes non-Latin-script text completely unreadable. I want to store these files in a format that's human-readable _as well as_ machine-readable.

Currently, [html/entities](https://github.com/denoland/deno_std/blob/main/html/entities.ts) exports `escape` and `unescape` functions. I suggest exporting a third function (tentatively named `normalize`) that normalizes all entities in a string of HTML or XML to a form that's valid, interoperable, and (mostly) human-readable:

```ts
normalize('<p>&#x4e24;&#x53ea;&#x5c0f;&#x871c;&#x8702;</p>') // '<p>两只小蜜蜂</p>'
normalize('a&b') // 'a&amp;b'
normalize('&#62;&#x3e;&gt;') // '&gt;&gt;&gt;'
normalize('&apos;') // '&#39;'
```

The logic could be something like this:

```ts
export function normalize(xml: string) {
    return xml
        .split(/([<>'"]+)/)
        .map((segment, i) => {
            return i % 2 ? segment : escape(unescape(segment))
        })
        .join('')
}
```

It might be worth having multiple normalized forms (which would likely also affect the API surface area of `escape`); for example, a "readability" form that converts `&#x4e24;&#x53ea;&#x5c0f;&#x871c;&#x8702;` to `两只小蜜蜂` vs a "compatibility" form that converts in the opposite direction. I don't currently have a use case for the "compatibility" form as any XML-consuming APIs I need to interact with either default to UTF-8 or respect UTF-8 where specified, but it might be useful for users needing to interact with legacy or poorly-designed APIs.

lionel-rowe:
Fixes #4523

Various bikeshedding things:
* Given its dual use for XML, should the top-level dir have its name changed from `html`? If so, to what?
  - `xml` suffers the same issue but in reverse
  - `html_and_xml` is pretty gross
  - `markup` seems overly vague
  - `sgml_like` might be technically correct, but would the average web developer think to look there?
* Is it worth exporting `escapeAllCharsAsHex`? Could be useful for users wanting finer-grained control than the 2 normalization forms, but it's a very simple function to replicate
* Is the name `NormalizationForm` OK or is it too confusing with Unicode normalization forms (NFC, NFD, etc?)

### DIFF
--- a/html/entities.ts
+++ b/html/entities.ts
@@ -48,7 +48,7 @@ export function escape(
   const escaped = str.replaceAll(rawRe, (m) => rawToEntity.get(m)!);
   return options.form === "compatibility"
     ? escapeAllNonAsciiPrintable(escaped)
-    : escaped;
+    : escapeAllXmlRestricted(escaped);
 }
 
 /** Options for {@linkcode unescape}. */
@@ -136,6 +136,8 @@ export type NormalizationOptions = {
  *
  * @example
  * ```ts
+ * import { normalize } from "https://deno.land/std@$STD_VERSION/html/entities.ts";
+ *
  * normalize("&#x3e;"); // "&gt;"
  * normalize("&apos;"); // "&#39;"
  * normalize("&#x4e24;&#x53ea;&#x5c0f;&#x871c;&#x8702;"); // "两只小蜜蜂"
@@ -159,14 +161,25 @@ export function normalize(
     .join("");
 }
 
-function escapeAllCharsAsHex(str: string) {
-  return [...str].map((c) => `&#x${c.codePointAt(0)!.toString(16)};`).join("");
+/**
+ * See https://en.wikipedia.org/wiki/Valid_characters_in_XML#Non-restricted_characters
+ */
+function escapeAllXmlRestricted(str: string) {
+  return str.replaceAll(
+    // deno-lint-ignore no-control-regex
+    /[^\x09\x0a\x0d\x20-\x7e\x85\xa0-\ud7ff\ue000-\ufdcf\ufdf0-\ufffd\u{10000}-\u{1fffd}\u{20000}-\u{2fffd}\u{30000}-\u{3fffd}\u{40000}-\u{4fffd}\u{50000}-\u{5fffd}\u{60000}-\u{6fffd}\u{70000}-\u{7fffd}\u{80000}-\u{8fffd}\u{90000}-\u{9fffd}\u{a0000}-\u{afffd}\u{b0000}-\u{bfffd}\u{c0000}-\u{cfffd}\u{d0000}-\u{dfffd}\u{e0000}-\u{efffd}\u{f0000}-\u{ffffd}\u{100000}-\u{10fffd}]+/gu,
+    (m) => escapeAllCharsAsHex(m),
+  );
 }
 
 function escapeAllNonAsciiPrintable(str: string) {
   return str.replaceAll(
     // deno-lint-ignore no-control-regex
-    /[\x00-\x08\x0b\x0c\x0e-\x1F\x7F-\u{10ffff}]+/gu,
+    /[^\x09\x0a\x0d\x20-\x7e]+/gu,
     (m) => escapeAllCharsAsHex(m),
   );
+}
+
+function escapeAllCharsAsHex(str: string) {
+  return [...str].map((c) => `&#x${c.codePointAt(0)!.toString(16)};`).join("");
 }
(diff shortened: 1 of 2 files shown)


html/entities.ts
```diff
@@ -103,3 +115,73 @@ function codePointStrToChar(codePointStr: string, radix: number) {
...
+    .join("");
+}
+
+/**
+ * See https://en.wikipedia.org/wiki/Valid_characters_in_XML#Non-restricted_characters
+ */
+function escapeXmlRestricted(str: string) {
+  return str.replaceAll(
+    // deno-lint-ignore no-control-regex
+    /[^\x09\x0a\x0d\x20-\x7e\x85\xa0-\ud7ff\ue000-\ufdcf\ufdf0-\ufffd\u{10000}-\u{1fffd}\u{20000}-\u{2fffd}\u{30000}-\u{3fffd}\u{40000}-\u{4fffd}\u{50000}-\u{5fffd}\u{60000}-\u{6fffd}\u{70000}-\u{7fffd}\u{80000}-\u{8fffd}\u{90000}-\u{9fffd}\u{a0000}-\u{afffd}\u{b0000}-\u{bfffd}\u{c0000}-\u{cfffd}\u{d0000}-\u{dfffd}\u{e0000}-\u{efffd}\u{f0000}-\u{ffffd}\u{100000}-\u{10fffd}]+/gu,
```
kt3k, line 172: Why do we start escaping these chars by default?

Also does HTML have the same restricted chars concept as XML?

html/entities.ts
```diff
@@ -34,15 +38,23 @@ const rawRe = new RegExp(`[${[...rawToEntity.keys()].join("")}]`, "g");
  * // Characters that don't need to be escaped will be left alone,
  * // even if named HTML entities exist for them.
  * escape("þð"); // "þð"
+ * // You can force non-ASCII chars to be escaped by setting the `form` option to `compatibility`:
+ * escape("þð", { form: "compatibility" }); // "&#xfe;&#xf0;"
```
kt3k, line 42: `form: "compatibility"` sounds confusing to me as I don't see what it's compatible with.

html/entities.ts
```diff
@@ -103,3 +115,73 @@ function codePointStrToChar(codePointStr: string, radix: number) {
...
+ * normalize("&#x3e;"); // "&gt;"
+ * normalize("&apos;"); // "&#39;"
+ * normalize("&#x4e24;&#x53ea;&#x5c0f;&#x871c;&#x8702;"); // "两只小蜜蜂"
+ *
+ * // other markup is left untouched
+ * normalize("<p class='foo'>&#x1f308;</p>") // "<p class='foo'>🌈</p>"
+ *
+ * // specifying a `form` option (default is `readability`):
+ * normalize("两只小蜜蜂", { form: "readability" }); // "两只小蜜蜂"
+ * normalize("两只小蜜蜂", { form: "compatibility" }); // "&#x4e24;&#x53ea;&#x5c0f;&#x871c;&#x8702;"
```
kt3k, line 152: `form: "compatibility"` feels defeating the purpose of `normalize` to me. How about not having this option for now?

html/entities.ts
lionel-rowe: Upon reflection, this PR needs some more thought. I'd initially thought a sort of "baseline-compatible with both HTML and XML" would be a sensible default, given that these characters are rare in practice yet could cause problems in XML. But it turns out that entities for certain C1 control-character codepoints have [different semantics in HTML](https://html.spec.whatwg.org/multipage/parsing.html#numeric-character-reference-end-state) than XML — for example, `&#x80;`:

```js
['application/xml', 'text/html'].map((contentType) => {
    const { literal, entity } = JSON.parse(
        new DOMParser().parseFromString(
            '<div>{"literal": "\x80", "entity": "&#x80;"}</div>',
            contentType,
        ).querySelector('div').textContent,
    )
    return { contentType, literal, entity }
})

// { "contentType": "application/xml", "literal": "\x80", "entity": "\x80" }
// { "contentType": "text/html", "literal": "\x80", "entity": "€" }
```

Also, I'm not sure converting those characters to entities is the right approach. Probably simply stripping them out would be more sensible in most cases.

iuioiua:
@lionel-rowe, to not keep stale PRs open, are you happy for us to close this PR for now? You can re-open the existing PR or a new one once it is ready.

lionel-rowe:
> @lionel-rowe, to not keep stale PRs open, are you happy for us to close this PR for now? You can re-open the existing PR or a new one once it is ready.

Sure, I'll close now.

PR #4525, opened 2024-03-26
deprecation(crypto): rename an export to match style guide, deprecating original and two other obsolete imports

jeremyBanks:
(Originally part of #4515, broken into separate PR as [suggested by iuioiua](https://github.com/denoland/deno_std/pull/4515#pullrequestreview-1956667727).) 

This PR renames `crypto.ts`'s exported `const` `wasmDigestAlgorithms` to `DIGEST_ALGORITHM_NAMES`, to match Deno's style guide and reflect the fact that it includes all algorithm names supported by the digest function, not just a subset. The old name is maintained as a deprecated alias for now. The exported `WasmDigestAlgorithm` type is deprecated, as it's redundant with the existing `DigestAlgorithmName` export (all algorithms are available in Wasm, and if that changes in the future it should probably just be an internal implementation detail, not part of the public interface). The `FNVAlgorithms` export is similarly deprecated (they were previously a special case that was not included in Wasm).

### DIFF
--- a/crypto/crypto.ts
+++ b/crypto/crypto.ts
@@ -297,21 +297,21 @@ const stdCrypto: StdCrypto = ((x) => x)({
 /**
  * A FNV (Fowler/Noll/Vo) digest algorithm name supported by std/crypto.
  *
- * @deprecated (will be removed in 0.224.0)
+ * @deprecated (will be removed in 1.0.0)
  */
 export type FNVAlgorithms = "FNV32" | "FNV32A" | "FNV64" | "FNV64A";
 
 /**
  * Digest algorithm names supported by std/crypto with a Wasm implementation.
  *
- * @deprecated (will be removed in 0.224.0) Consider using {@linkcode DIGEST_ALGORITHM_NAMES} instead.
+ * @deprecated (will be removed in 1.0.0) Consider using {@linkcode DIGEST_ALGORITHM_NAMES} instead.
  */
 export const wasmDigestAlgorithms = DIGEST_ALGORITHM_NAMES;
 
 /**
  * A digest algorithm name supported by std/crypto with a Wasm implementation.
  *
- * @deprecated (will be removed in 0.224.0) Consider using {@linkcode DigestAlgorithmName} instead.
+ * @deprecated (will be removed in 1.0.0) Consider using {@linkcode DigestAlgorithmName} instead.
  */
 export type WasmDigestAlgorithm = DigestAlgorithmName;
 
--- a/crypto/crypto.ts
+++ b/crypto/crypto.ts
@@ -107,12 +107,12 @@
  * @module
  */
 import {
-  DIGEST_ALGORITHM_NAMES as wasmDigestAlgorithms,
-  type DigestAlgorithmName as WasmDigestAlgorithm,
+  DIGEST_ALGORITHM_NAMES,
+  type DigestAlgorithmName,
   instantiateWasm,
 } from "./_wasm/mod.ts";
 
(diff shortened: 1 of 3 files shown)


crypto/crypto.ts
```diff
@@ -294,11 +294,26 @@ const stdCrypto: StdCrypto = ((x) => x)({
   },
 });
 
-/** FNV (Fowler/Noll/Vo) algorithms names. */
+/**
+ * A FNV (Fowler/Noll/Vo) digest algorithm name supported by std/crypto.
+ *
+ * @deprecated (will be removed in 0.224.0)
```
kt3k, line 300: I'd suggest we remove this at version 1.0.0.

jeremyBanks: Gotcha. For all three of them, right?

kt3k (approved):
LGTM

crypto/crypto.ts
```diff
@@ -294,11 +294,26 @@ const stdCrypto: StdCrypto = ((x) => x)({
...
 export type FNVAlgorithms = "FNV32" | "FNV32A" | "FNV64" | "FNV64A";
 
-/** Extended digest algorithm names. */
-export type DigestAlgorithmName = WasmDigestAlgorithm | FNVAlgorithms;
+/**
+ * Digest algorithm names supported by std/crypto with a Wasm implementation.
+ *
+ * @deprecated (will be removed in 1.0.0) Consider using {@linkcode DIGEST_ALGORITHM_NAMES} instead.
+ */
+export const wasmDigestAlgorithms = DIGEST_ALGORITHM_NAMES;
```
iuioiua, line 309: Ditto. See https://github.com/search?q=%22wasmDigestAlgorithms%22&type=code

crypto/crypto.ts
```diff
@@ -294,11 +294,26 @@ const stdCrypto: StdCrypto = ((x) => x)({
   },
 });
 
-/** FNV (Fowler/Noll/Vo) algorithms names. */
+/**
+ * A FNV (Fowler/Noll/Vo) digest algorithm name supported by std/crypto.
+ *
+ * @deprecated (will be removed in 1.0.0)
+ */
 export type FNVAlgorithms = "FNV32" | "FNV32A" | "FNV64" | "FNV64A";
```
iuioiua, line 302: The use of this export is minimal. A removal version of v1 instead of sooner is overkill and would be non-impactful to remove sooner. See https://github.com/search?q=%22FNVAlgorithms%22&type=code and notice they're almost entirely forks of std. WDYT, @kt3k?

crypto/crypto.ts
```diff
@@ -294,11 +294,26 @@ const stdCrypto: StdCrypto = ((x) => x)({
...
+ * @deprecated (will be removed in 1.0.0) Consider using {@linkcode DIGEST_ALGORITHM_NAMES} instead.
+ */
+export const wasmDigestAlgorithms = DIGEST_ALGORITHM_NAMES;
+
+/**
+ * A digest algorithm name supported by std/crypto with a Wasm implementation.
+ *
+ * @deprecated (will be removed in 1.0.0) Consider using {@linkcode DigestAlgorithmName} instead.
+ */
+export type WasmDigestAlgorithm = DigestAlgorithmName;
```
iuioiua, line 316: Ditto. See https://github.com/search?q=%22WasmDigestAlgorithm%22&type=code

crypto/crypto.ts
kt3k: There's no benefit of removing them earlier in my view.

Breaking at 1.0 is more obvious and less confusing. That is better than breaking at random 0.x version in my view

PR #4534, opened 2024-03-28
feat(semver): `greaterThanRange()` and `lessThanRange()`

kt3k:
towards #4365

This PR adds `greaterThanRange` and `lessThanRange` APIs without using currently broken `rangeMax` and `rangeMin`.

The test cases are borrowed from `node-semver` (https://github.com/npm/node-semver/tree/main/test/fixtures). They are attributed in the headers of test files


semver/less_than_range_test.ts
```diff
@@ -0,0 +1,171 @@
...
+    ["1.0.0 - 2.0.0", "2.0.0"],
+    ["^3.0.0", "4.0.0"],
+    ["^1.0.0 || ~2.0.1", "2.0.0"],
+    ["^0.1.0 || ~3.0.1 || 5.0.0", "3.2.0"],
+    ["^0.1.0 || ~3.0.1 || 5.0.0", "5.0.0-0"],
+    ["^0.1.0 || ~3.0.1 || >4 <=5.0.0", "3.5.0"],
+    ["^1.0.0-alpha", "1.0.0-beta"],
+    ["~1.0.0-alpha", "1.0.0-beta"],
+    ["=0.1.0", "1.0.0"],
+    // Note: Mayve worth considering to add support for includePrerelease option
```
iuioiua, line 159:
```suggestion
    // Note: maybe worth considering adding support for `includePrerelease` option
```

semver/less_than_range.ts
```diff
@@ -0,0 +1,56 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import type { Comparator, Range, SemVer } from "./types.ts";
+import { testComparatorSet } from "./_test_comparator_set.ts";
+import { isWildcardComparator } from "./_shared.ts";
+import { compare } from "./compare.ts";
+
+/** Check if the semver is less than the range. */
+export function lessThanRange(semver: SemVer, range: Range): boolean {
```
iuioiua, line 9: Optional for this PR, but full documentation would be nice.

semver/less_than_range.ts
```diff
@@ -0,0 +1,56 @@
...
+    // If the semver is greater than any of the comparator set, then it's not less than the range.
+    if (
+      comparatorSet.some((comparator) =>
+        greaterThanComparator(semver, comparator)
+      )
+    ) {
+      return false;
+    }
+  }
+  return true;
```
iuioiua, line 27:
```suggestion
  return !range.some((comparatorSet) =>
    comparatorSet.some(isWildcardComparator) ||
    // If the semver satisfies the comparator set, then it's not less than the range.
    testComparatorSet(semver, comparatorSet) ||
    // If the semver is greater than any of the comparator set, then it's not less than the range.
    comparatorSet.some((comparator) =>
      greaterThanComparator(semver, comparator)
    )
  );
```

semver/less_than_range.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+    case "<": {
+      return cmp >= 0;
+    }
+    case ">=": {
+      return false;
+    }
+    case "<=": {
+      return cmp > 0;
+    }
```
iuioiua, line 54:
```suggestion
    case "=":
    case undefined:
      return cmp > 0;
    case "!=":
    case ">":
    case ">=":
      return false;
    case "<":
    case "<=":
      return cmp >= 0;
```

semver/less_than_range.ts
```diff
@@ -0,0 +1,56 @@
...
+        greaterThanComparator(semver, comparator)
+      )
+    ) {
+      return false;
+    }
+  }
+  return true;
+}
+
+function greaterThanComparator(
```
iuioiua, line 30: Nit: This is more of a personal preference, but I like it when functions are defined before they're used. In other words, I'd prefer this function be moved before `lessThanRange()`.

semver/greater_than_range.ts
```diff
@@ -0,0 +1,51 @@
...
+      return false;
+    }
+    // If the semver is less than any of the comparator set, then it's not greater than the range.
+    if (
+      comparatorSet.some((comparator) => lessThanComparator(semver, comparator))
+    ) {
+      return false;
+    }
+  }
+  return true;
```
iuioiua, line 25:
```suggestion
    comparatorSet.some(isWildcardComparator) ||
    // If the semver satisfies the comparator set, then it's not greater than the range.
    testComparatorSet(semver, comparatorSet) ||
    // If the semver is less than any of the comparator set, then it's not greater than the range.
    comparatorSet.some((comparator) => lessThanComparator(semver, comparator))
  );
```

semver/greater_than_range.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+    case "<": {
+      return false;
+    }
+    case ">=": {
+      return cmp < 0;
+    }
+    case "<=": {
+      return false;
+    }
```
iuioiua, line 49:
```suggestion
    case "=":
    case undefined:
      return cmp < 0;
    case "!=":
      return false;
    case ">":
      return cmp <= 0;
    case "<":
      return false;
    case ">=":
      return cmp < 0;
    case "<=":
      return false;
    default:
      return false;
```

iuioiua (changes requested):
Some of my suggestions are optional.

semver/less_than_range.ts
kt3k: I extracted `lessThanComparatorSet` to make this part more readable

semver/greater_than_range.ts
kt3k: I extracted `greaterThanComparatorSet` instead

iuioiua (approved):

PR #4537, opened 2024-03-30
feat(expect): support `expect.addSnapshotSerialize()` api

eryue0220:
This pr is one of part to solve https://github.com/denoland/deno_std/issues/3964, which is to support `expect.addSnapshotSerialize` api. But there is still a lot of other work to support [serialize plugins](https://github.com/jestjs/jest/tree/main/packages/pretty-format/src/plugins). I think it can be divided into a separate PR or Issue for tracking.


kt3k (approved):
LGTM. Thanks!

PR #4539, opened 2024-04-02
chore: enable `verbatim-module-syntax` lint rule

dsherret:


front_matter/mod.ts
```diff
@@ -115,9 +115,9 @@
  *
  * @module
  */
-import { extract as extractJson } from "./json.ts";
-import { extract as extractToml } from "./toml.ts";
-import { extract as extractYaml } from "./yaml.ts";
+import type { extract as extractJson } from "./json.ts";
+import type { extract as extractToml } from "./toml.ts";
+import type { extract as extractYaml } from "./yaml.ts";
```
dsherret, line 120: Huh, interesting.

iuioiua: Omg. We import these symbols without exporting them. Lol. If these were exported, it'd break their functionality. Let's remove these imports for now. I'll fix this up in a separate PR.

dsherret: I think this is done for the module doc above? Not sure.

iuioiua: Oh... Yes. That might be right.

iuioiua (approved):
LGTM

Issue #4326, opened 2024-02-13
suggestion: restructure `std/console`, `std/cli`, `std/fmt` and `std/text`


Leokuma:
You guys probably have a vision for the future of those modules, but as they currently stand, I would like to propose these changes:

1. ✅ Move `std/console` into `std/cli` and deprecate `std/console`
2. Move `fmt/colors` to `std/cli`, and possibly rename from `colors.ts` to ~~`ansi.ts`~~ `styles.ts` as it also does bold and italic (I'm assuming the majority of people looking for ANSI color codes are targetting the terminal)
3. Move `fmt/printf/printf` to `std/cli` (`fmt/printf/sprintf` would stay in `fmt`)
4. ⬇️ Move `text/case` to `std/fmt` (see new idea below)
5. ❌ ~~Rename `std/text` to `std/text_search`. Not sure about this one~~. This would narrow down the module too much. Golang's String module has many functions that could be added to std/text in the future.

-------------

New idea for item 4: deprecate `std/fmt` entirely:

- Move `std/fmt/bytes.ts` to `std/bytes`
  - Questionable because `bytes.ts` formats for example 1337 → 1.34 kB, so it handles Number, not Uint8Array or ArrayBuffer. It might make more sense to have such a function in an std/number, but there's no std/number.
- Move `std/fmt/colors.ts` to `std/cli` (item 2)
- Move `std/fmt/duration.ts` to `std/datetime`
  - Maybe not worth the breaking change because both `std/fmt/duration.ts` and `std/datetime` might be deprecated once Temporal is stabilized
- Move `std/fmt/printf.ts` to `std/text` (impacts item 3)

iuioiua:
I agree with 1-3, am unsure about 4, and disagree with 5, as `std/text_search` narrows down the purpose of that module too much.

iuioiua:
I talked with Yoshiya, and we're in favour of merging `std/console` into `std/cli` and deprecating `std/console`. We'll make sure to get back to you about the other points.

iuioiua:
I'm in favour of moving `fmt/colors` to `std/cli`. WDYT, @kt3k?

Leokuma:
I've updated item 4 with a new suggestion to deprecate `std/fmt`.

iuioiua:
[...]

PR #4547, opened 2024-04-04
deprecation(console): copy `std/console` APIs into `std/cli` and deprecate `std/console`

iuioiua:
As part of #4326


console/mod.ts
```diff
@@ -14,6 +14,10 @@
  * unicodeWidth("天地玄黃宇宙洪荒"); // 16
  * ```
  *
+ * @deprecated Use {@link https://jsr.io/@std/cli | std/cli} instead. This
```
iuioiua, line 17: Are we sure about this?

kt3k (approved):
LGTM

PR #4549, opened 2024-04-05
fix(cli):  make the output of `promptSecret()` consistent with `prompt()`

roj1512:


iuioiua (approved):
Nice catch! Thank you.

PR #4550, opened 2024-04-05
fix(cli): make `promptSecret()` browser-compatible

roj1512:


kt3k:
This type of browser compatibility feels rather confusing to me as the main feature of the API is completely skipped.

iuioiua:
I agree. This change is misleading because it doesn't add the ability to prompt secret phrases in the browser, nor is it possible without the required low-level console APIs in the browser.

We can perhaps tweak this function to work in other non-browser environments by having `stdin` and `stdout` be arguments that use `read()` and `write()` methods if they're similar in design to `Deno.stdin.read()` and `Deno.stdout.write()`.

roj1512:
> This type of browser compatibility feels rather confusing to me as the main feature of the API is completely skipped.

The main feature of fmt/colors.ts is also skipped in browsers.

iuioiua:
> @iuioiua @kt3k The text at https://deno.land/std@0.221.0/cli/mod.ts is misleading where it says: This module is browser compatible.
> 
> And below it is the `promptSecret()` symbol.

In that case, we shouldn't classify that API as browser-compatible.

> > This type of browser compatibility feels rather confusing to me as the main feature of the API is completely skipped.
> 
> The main feature of fmt/colors.ts is also skipped in browsers.

Not exactly. Instead, it's the ability to "turn off" those functions via a `NO_COLOR` environment variable that is disabled. One can still print styled text in the console in browsers with those APIs.

roj1512:
> In that case, we shouldn't classify that API as browser-compatible.

Looking forward to that change.

PR #4551, opened 2024-04-05
tests(fmt): add `duration()` tests

niklasmtj:
This should up the [code coverage of `fmt/duration`](https://app.codecov.io/gh/denoland/deno_std/blob/main/fmt%2Fduration.ts) up to 100%. Not 100% about the second test's name so I'm open to suggestions since we're testing here the default `switch` case.

This referes to #3713


iuioiua (approved):
LGTM! Thank you.

PR #4555, opened 2024-04-08
deprecation(encoding): make exported function names consistent with `std/encoding`

iuioiua:
This PR deprecates:
1. `encode()` in favor of `encodeVarint()`
2. `decode()` in favor of `decodeVarint()`
3. `decode32()` in favor of `decodeVarint32()`

### DIFF
--- a/encoding/varint.ts
+++ b/encoding/varint.ts
@@ -60,7 +60,7 @@ const U64_VIEW = new BigUint64Array(AB);
  * decode(buf); // [ 300n, 2 ];
  * ```
  *
- * @deprecated (will be removed in 0.224.0) Use {@linkcode decodeVarint} instead.
+ * @deprecated (will be removed in 1.0.0) Use {@linkcode decodeVarint} instead.
  */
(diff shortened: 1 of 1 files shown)


encoding/varint.ts
```diff
@@ -57,11 +57,41 @@ const U64_VIEW = new BigUint64Array(AB);
  * import { decode } from "https://deno.land/std@$STD_VERSION/encoding/varint.ts";
  *
  * const buf = new Uint8Array([0x8E, 0x02]);
- * decode(buf);
- * // [ 300n, 2 ];
+ * decode(buf); // [ 300n, 2 ];
  * ```
+ *
+ * @deprecated (will be removed in 0.224.0) Use {@linkcode decodeVarint} instead.
```
kt3k, line 63: I'd suggest we should remove them in 1.0.0 as `std/encoding` is marked 'Stable' in README.

iuioiua: Fixed.

kt3k (approved):
LGTM 👍

PR #4557, opened 2024-04-08
ci: validate PR titles

iuioiua:
This allows us to more accurately follow [Conventional Commits](https://www.conventionalcommits.org/), which we already mostly do. The correctness of PR titles will become more important, as we plan to use PR titles in to determine release versions of packages.

### DIFF
--- a/.github/workflows/title.yml
+++ b/.github/workflows/title.yml
@@ -1,0 +1,72 @@
+name: "Lint PR title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+permissions:
+  pull-requests: read
+
+jobs:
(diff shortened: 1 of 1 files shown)


kt3k:
The way to express breaking changes also isn't aligned to our convention.

iuioiua:
We can use `!` for breaking changes. E.g. `feat(assert)!: assertAlmostEquals() sets useful tolerance automatically`.

iuioiua:
Maybe even just add `BREAKING` to the list of types.

iuioiua:
Updated to include `[unstable] BREAKING` and `BREAKING`.

kt3k (approved):
Ok. Let's try this

PR #4558, opened 2024-04-09
docs(log): remove string formatter description

wangbinyq:
update the documentation according to this pull request: https://github.com/denoland/deno_std/pull/4239


iuioiua (approved):
Good catch. Thank you.

PR #4560, opened 2024-04-09
chore: update browser-compatible files with declaration

iuioiua:
These files were missing `// This module is browser compatible.` declarations.


kt3k (approved):
LGTM

PR #4561, opened 2024-04-10
deprecation(semver): deprecate `rangeMax()` and `rangeMin()`

### LINKED ISSUE #4365: semver: deprecate `rangeMax` and `rangeMin`

kt3k:
closes #4365

### DIFF
--- a/semver/range_max.ts
+++ b/semver/range_max.ts
@@ -44,7 +44,7 @@ function comparatorMax(comparator: Comparator): SemVer {
 }
 
 /**
- * @deprecated (will be removed in 1.0.0) use greaterThanRange or lessThanRange for comparing ranges and semvers. The maximum version of a range is often not well defined, and therefore this API shouldn't be used. See https://github.com/denoland/deno_std/issues/4365 for details.
+ * @deprecated (will be removed in 1.0.0) Use {@linkcode greaterThanRange} or {@linkcode lessThanRange} for comparing ranges and semvers. The maximum version of a range is often not well defined, and therefore this API shouldn't be used. See https://github.com/denoland/deno_std/issues/4365 for details.
  *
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
I made a couple of tweaks. Now, LGTM!

PR #4565, opened 2024-04-10
test(testing): cleanup bdd tests

iuioiua:
This PR cleans up `testing/bdd_test.ts` by removing the unnecessary wrapping `global` test.


kt3k:
@KyleJune Does this change make sense to you? Or did you have plan to extend the test cases which need to be isolated from "global" testing?

KyleJune:
> @KyleJune Does this change make sense to you? Or did you have plan to extend the test cases which need to be isolated from "global" testing?

I can't recall the reason I had the top level global grouping for the test cases. But it looks like it works without that, so I think it's fine to remove.

I think it might have just been to signal that those were the tests for when those functions were used at the top/global level rather than nested.

kt3k (approved):
LGTM

PR #4566, opened 2024-04-10
deprecation(yaml): cleanup schema exports

iuioiua:
Previously, `core`, `def`, `extended`, `json` and `failsafe` constants were exported from their source files. However, the same symbols were exported as `CORE_SCHEMA`, `DEFAULT_SCHEMA`, `EXTENDED_SCHEMA`, `JSON_SCHEMA`, and `FAILSAFE_SCHEMA` were exported elsewhere, which were the same symbols, respectively. See https://jsr.io/@std/yaml/doc. This change uses the uppercase snake case names, which is the appropriate casing for constants.

### DIFF
--- a/yaml/schema/core.ts
+++ b/yaml/schema/core.ts
@@ -5,10 +5,22 @@
 // This module is browser compatible.
 
 import { Schema } from "../schema.ts";
-import { json } from "./json.ts";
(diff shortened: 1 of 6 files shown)


kt3k (approved):
Good catch! LGTM

PR #4567, opened 2024-04-10
test(async): improve test coverage

mbhrznr:
pushes code coverage of the `async` sub-module closer to 100%.

the two remaining files in the sub-module which aren't at 100% are:
- `mux_async_iterator.ts`: 
   resetting the errors via `this.#throws.length = 0` seems to be unreachable code.
   if we have a truthy value for `this.#throws.length` we would throw all of the errors, thus exiting `iterate`.
- `pool.ts`:
   the `catch` block as well as the error handling seems to be covered by tests already.


async/pool_test.ts
```diff
@@ -22,7 +22,7 @@ Deno.test("pooledMap()", async function () {
   assert(diff < 3000);
 });
 
-Deno.test("pooledMap() handles errors", async () => {
+Deno.test("pooledMap() handles errors", async function () {
```
iuioiua, line 25: Nit: we generally use the arrow syntax for anonymous functions. Please revert. Ditto elsewhere.

async/delay_test.ts
```diff
@@ -103,3 +105,35 @@ Deno.test("delay() handles already aborted signal", async function () {
...
+  using unrefTimer = stub(Deno, "unrefTimer", () => {
+    throw new ReferenceError();
+  });
+  await delay(100, { persistent: false });
+  assertSpyCalls(unrefTimer, 1);
+});
+
+Deno.test({
+  name: "delay() handles persistent option with error",
+  fn: async function () {
```
iuioiua, line 125:
```suggestion
  async fn() {
```
Nit

iuioiua (changes requested):
Nicely done! I have just a couple of nits. Let's also remove the unreachable code from `mux_async_iterator.ts`.

async/pool_test.ts
mbhrznr: understood! also took liberty to convert the other test cases in the async sub-module to make use of the arrow syntax as well.

iuioiua (approved):
LGTM! Good stuff. Thank you.

PR #4575, opened 2024-04-11
deprecation(permissions): update removal time for `std/permissions`

iuioiua:
This PR updates the removal version of `std/permissions` to be when `std` migrates to JSR, as `std/permissions` will never reach v1.


kt3k (approved):
LGTM

PR #4576, opened 2024-04-12
chore: update deprecation notices and check

iuioiua:
Having a deprecation notice containing "will be removed after..." can be a little confusing as it suggests that there's no definitive time/version when an API will be removed.


iuioiua:
Closing in favour of #4577.

PR #4577, opened 2024-04-12
chore: simplify deprecation notice check

iuioiua:
This change simplifies the deprecation notice check by making it just check whether a deprecation notice comes with a message. Deprecation notices have also been changed to no longer include brackets at the beginning of the sentence, which always seemed odd to me.

Previously, this check was stricter than it needed to be and has become a little outdated, given that the way we deprecate APIs is evolving.

The script execution time has reduced from ~5.6s to ~3.


kt3k:
I find the existing check of removal version numbers is useful. I don't see the reason to remove it.

iuioiua:
It did. But not all deprecation notices contain versions ([example](https://github.com/denoland/deno_std/blob/61183bcbccc70a54f95707414d6341429a98e015/console/mod.ts#L18-L20)). We currently compensate by [skipping checking this file](https://github.com/denoland/deno_std/blob/61183bcbccc70a54f95707414d6341429a98e015/_tools/check_deprecation.ts#L50) entirely, which means checking possibly valid deprecation notices.

kt3k (approved):
LGTM

iuioiua:
Note: This is a temporary change. Once we migrate to JSR, I'll add a check that ensures deprecation notices contain a semver string.

PR #4579, opened 2024-04-12
chore: enable Codecov comments in PRs

iuioiua:
This change enables Codecov comments. I find myself having to manually look up test coverage changes on commits in the Codecov UI. Having this presented within PR feeds would help a lot. We can re-evaluate if it becomes too noisy.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,4 +1,3 @@
-comment: false
 codecov:
   require_ci_to_pass: true
 coverage:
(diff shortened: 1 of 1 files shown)


.github/codecov.yml
```diff
@@ -9,5 +8,3 @@ coverage:
     patch:
       default:
         informational: true
-ignore:
```
iuioiua, line 12: Pass-by cleanup. `std/examples` no longer exists.

kt3k (approved):
LGTM

Issue #4546, opened 2024-04-03
csv - stringify with no options should not output an empty blank line before the data


dsherret:
**Describe the bug**

```
> import { stringify } from "jsr:@std/csv@0.221";
undefined
> stringify([[1, 2, 3]])
"\r\n1,2,3\r\n"
```

**Steps to Reproduce**

See above.

**Expected behavior**

Should output `1,2,3\n` (also, don't bother with `\r\n` on Windows. Make the world better and just use `\n`)

**Environment**

- std version: 0.221.0

PR #4581, opened 2024-04-13
BREAKING(csv): use `LF` instead of `CRLF`

fdemir:
Fixes https://github.com/denoland/deno_std/issues/4546#issue-2224115477

### DIFF
--- a/csv/stringify.ts
+++ b/csv/stringify.ts
@@ -122,7 +122,6 @@ export type StringifyOptions = {
 
 const QUOTE = '"';
 const LF = "\n";
-const CRLF = "\r\n";
 const BYTE_ORDER_MARK = "\ufeff";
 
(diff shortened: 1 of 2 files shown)


fdemir:
Removing CLRF and make the world bettter place.

iuioiua (changes requested):
I'm for this change. Modern OSs handle LF fine, and this is one less character. The use of LF also seems more prevalent. However, this is a breaking change of a stable API. This PR should be landed in 1.0.

Either way, please resolve the test failures.

kt3k:
@fdemir Can you split the fix for #4546 and breaking change of CRLF to LF into different PRs?

iuioiua:
If CRLF is what the spec defines, let's not make this specific change. The other suggestion in #4546 makes sense, though. @kt3k, please close if you agree.

PR #4592, opened 2024-04-16
refactor(assert,expect): use shared `format.ts` and `diff.ts`

### LINKED ISSUE #3912: move `_format.ts` into root

followup of https://github.com/denoland/deno_std/pull/3886#issuecomment-1835302006. currently `assert/_format.ts` and `expect/_format.ts` is identical but duplicated across modules.

create private top-level module, `_format.ts` and `_format_test.ts`

- symlinks would probably not work on web.
- keep the status quo, however this increases cognitive complexity and makes it harder to sync both files.

iuioiua:
Closes #3912


internal/mod.ts
kt3k, line 1: Is this file necessary?

kt3k:
This seem copying files from `assert` to `internal`, but test cases seem having better names in `expect`. How about copying these utils from `expect/` instead?

internal/mod.ts
iuioiua: Yes. Check out the file diff. Some files import from both `internal/format.ts` and `internal/diff.ts`.  For such cases, we can use `mod.ts` for now.

kt3k (approved):
LGTM

Issue #4545, opened 2024-04-03
csv - parse takes undefined opt


dsherret:
Kind of confusing that the `opt` parameter here is `opt?: undefined`:

https://github.com/denoland/deno_std/blob/74fc8ff920991c4c79fa4c7647d855e5780db64c/csv/parse.ts#L326

...shouldn't that just not exist given the other overloads?

fdemir:
Should we remove the opt?: undefined from this overload?

dsherret:
Yeah, I think it should just be `export function parse(input: string): string[][];`

Wencho8:
I can work on this. Removing the extra comments and the opt parameter.

iuioiua:
Completed in #4598

dandv:
Not sure what the status of this is, but I'm giving Deno a first try with a simple task: manipulate a CSV. Trying to figure out the parameters for `parse` in WebStorm is tricky, because the options object doesn't seem to be documented (only `input` is), but is used in the example:

![image](https://github.com/denoland/deno_std/assets/33569/afa4de9e-ad96-4f95-81d8-23680483e90a)

I'm using jsr:@std/csv@^0.224.1.

BlackAsLight:
> Not sure what the status of this is, but I'm giving Deno a first try with a simple task: manipulate a CSV. Trying to figure out the parameters for `parse` in WebStorm is tricky, because the options object doesn't seem to be documented (only `input` is), but is used in the example:
> 
[...]

PR #4598, opened 2024-04-16
fix(csv): remove unused `opt` parameter

Wencho8:
This was made according to the issue #4545 inside the file deno_std/csv/parse.ts


iuioiua (approved):
Nice! Thank you.

iuioiua:
Closes #4545

PR #4599, opened 2024-04-16
BREAKING(streams): replace rest params of `mergeReadableStreams()` with a single array param

roj1512:
- This makes it consistent with `concat()` of the bytes submodule.
- It is also slightly more performant.

### DIFF
--- a/streams/merge_readable_streams.ts
+++ b/streams/merge_readable_streams.ts
@@ -13,11 +13,11 @@
  * const stream2 = ReadableStream.from(["a", "b", "c"]);
  *
  * // ["2", "c", "a", "b", "3", "1"]
- * await Array.fromAsync(mergeReadableStreams(stream1, stream2));
+ * await Array.fromAsync(mergeReadableStreams([stream1, stream2]));
  * ```
  */
 export function mergeReadableStreams<T>(
-  ...streams: ReadableStream<T>[]
+  streams: ReadableStream<T>[]
 ): ReadableStream<T> {
   const resolvePromises = streams.map(() => Promise.withResolvers<void>());
   return new ReadableStream<T>({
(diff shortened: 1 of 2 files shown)


kt3k:
> This makes it consistent with concat() of the bytes submodule.
> It is also slightly more performant.

These don't sound enough reasons to make a breaking change. The API stability matters more in my view.

Also we don't land immediate breaking change like this. We always deprecate the existing API first.

roj1512:
> These don't sound enough reasons to make a breaking change.

Even the Web API’s types and functions take arrays, and not rest parameters.

crowlKats:
I completely agree with @kt3k, this change seems unecessary

kt3k:
`concat` and `mergeReadableStreams` belong to different sub modules, and they don't have particular special connection. I don't think people consider the current state as inconsistency.

roj1512:
I consider it as inconsistency. Am I not people? Are you discriminating me?

iuioiua:
@roj1512, there's nothing wrong with disagreeing, but reacting out-of-context or perhaps trolling like this is unwelcome (see our [code of conduct](https://github.com/denoland/deno_std/blob/main/.github/CODE_OF_CONDUCT.md)). Please remain respectful.

Either way, I see both sides of the argument. Can you share benchmarks to show the difference in performance before and after the refactor?

iuioiua:
Thanks for the benchmarks, @roj1512. For context, `concat()` migrated away from using the spread operator because there's a use case for concatenating a large number of buffers (see #3696). There are not many instances where a user could merge a large number of readable streams simply because they usually seem to be used in lower numbers.

For this reason, I'll be closing this PR without merging. Either way, thank you for the suggestion! In future, please let us know of these suggestions in issues, so we can discuss and come to a consensus first 🙂

PR #4603, opened 2024-04-17
fix(bytes): improve invalid count error message for `repeat()`

iuioiua:
This change simplifies the error logic and message in `repeat()` when the `count` parameter is negative or not an integer.


kt3k (approved):
The error message looks better to me.

LGTM

PR #4604, opened 2024-04-17
feat(http): file server prints local network address

### LINKED ISSUE #4597: [http/file_server] Show local network IP

When developing with a local file server, it's often useful to be able to connect to it from other devices (for example, to test how a website looks on mobile). `file_server ./path` only gives me the localhost address, and I have to check in ifconfig to find my IP address:

```
➜ file_server .
Listening on http://localhost:4507/
```

Vercel's `server` prints this output:
```
➜ npx serve .
(node:30004) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)

   ┌────────────────────────────────────────────┐
   │                                            │
   │   Serving!                                 │
   │                                            │
   │   - Local:    http://localhost:3000        │
   │   - Network:  http://192.168.178.40:3000   │
   │                                            │
   │   Copied local address to clipboard!       │
   │                                            │
   └────────────────────────────────────────────┘
```

It's be great if `file_server` showed the "Network" IP too.

kt3k:
closes #4597

This PR updates `onListen` hook of file_server to show the local network IP address. The scripts shows 2 urls like the below (Updated):

```shellsession
$ deno run -A http/file_server.ts 
Listening on:
- Local: http://localhost:4507
- Network: http://192.168.79.163:4507
```

Suggestions to the output format are welcome

---

Note: `npm:serve` chooses non internal IPv4 address as the network address. This implementation follows it. ref. https://github.com/vercel/serve/blob/1ea55b1b5004f468159b54775e4fb3090fedbb2b/source/utilities/http.ts#L40

Note: Deno's [NetworkInterfaceInfo](https://deno.land/api@v1.42.4?s=Deno.NetworkInterfaceInfo) object doesn't have `internal` field, but it's polyfilled as `address.startsWith("127")` in the node compat layer https://github.com/denoland/deno/blob/9acbf90b06bf79dd6e4cf2428b3566da009bed65/ext/node/polyfills/os.ts#L232-L234. This implementation follows it as well.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -807,7 +807,7 @@ function main() {
   const useTls = !!(keyFile && certFile);
 
   function onListen({ port }: { port: number }) {
-    const networkAddress = getNewtowrkAddress();
+    const networkAddress = getNetworkAddress();
     const protocol = useTls ? "https" : "http";
     let message =
       `Serving static files!\nlocal: ${protocol}://localhost:${port}`;
@@ -839,7 +839,7 @@ function main() {
  * inspired by the util of the same name in `npm:serve`
  * https://github.com/vercel/serve/blob/1ea55b1b5004f468159b54775e4fb3090fedbb2b/source/utilities/http.ts#L33
  */
-function getNewtowrkAddress() {
+function getNetworkAddress() {
   for (const { family, address } of Deno.networkInterfaces()) {
     if (family === "IPv4" && !address.startsWith("127.")) {
       return address;
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -810,7 +810,7 @@ function main() {
     const networkAddress = getNetworkAddress();
     const protocol = useTls ? "https" : "http";
     let message = `Listening on:\n- Local: ${protocol}://${hostname}:${port}`;
-    if (networkAddress) {
+    if (networkAddress && !DENO_DEPLOYMENT_ID) {
       message += `\n- Network: ${protocol}://${networkAddress}:${port}`;
     }
     console.log(message);
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -806,13 +806,12 @@ function main() {
 
   const useTls = !!(keyFile && certFile);
 
-  function onListen({ port }: { port: number }) {
+  function onListen({ port, hostname }: { port: number; hostname: string }) {
     const networkAddress = getNetworkAddress();
     const protocol = useTls ? "https" : "http";
-    let message =
-      `Serving static files!\nlocal: ${protocol}://localhost:${port}`;
+    let message = `Listening on:\n- local: ${protocol}://${hostname}:${port}`;
     if (networkAddress) {
-      message += `\nnetwork: ${protocol}://${networkAddress}:${port}`;
+      message += `\n- network: ${protocol}://${networkAddress}:${port}`;
     }
     console.log(message);
   }
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -806,10 +806,22 @@ function main() {
 
   const useTls = !!(keyFile && certFile);
 
+  function onListen({ port }: { port: number }) {
+    const networkAddress = getNewtowrkAddress();
+    const protocol = useTls ? "https" : "http";
+    let message =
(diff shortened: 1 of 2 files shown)


kt3k:
Side note: This will prints ip address that is not useful to the user in Deno Deploy like:

```
Serving static files!
local: http://localhost:80
network: http://10.0.0.10:80
```

http/file_server.ts
```diff
@@ -806,21 +806,47 @@ function main() {
 
   const useTls = !!(keyFile && certFile);
 
+  function onListen({ port }: { port: number }) {
```
iuioiua, line 809: My suggested format (staying close to default behaviour).
```
Listening on:
- Local: http://localhost:4507
- Network: http://192.168.0.104:4507
```

iuioiua: We should also use the `hostname` property.

iuioiua (changes requested):
Can you please add a test?

> Side note: This will prints ip address that is not useful to the user in Deno Deploy like:
> 
> ```
> Serving static files!
> local: http://localhost:80
> network: http://10.0.0.10:80
> ```

This doesn't seem to be the case, as the functionality added in this PR is only accessible when running the script directly via `deno run -A jsr:@std/http/file-server` because we're using `import.meta.main`.

kt3k:
> This doesn't seem to be the case. The functionality added in this PR is only accessible when running the script directly via deno run -A jsr:@std/http/file-server because we're using import.meta.main.

file_server is often used as the entrypoint script in Deploy when serving static site. `import.meta.main` is true in that case. ref. https://github.com/search?q=path%3Adeploy.yml%20file_server.ts&type=code

We even optimize etag generation for Deno Deploy. https://github.com/denoland/deno_std/pull/3186

kt3k:
Now this prints:

```shellsession
$ deno run -A http/file_server.ts 
Listening on:
- local: http://localhost:4507
- network: http://192.168.79.163:4507
```

http/file_server.ts
```diff
@@ -806,21 +806,46 @@ function main() {
 
   const useTls = !!(keyFile && certFile);
 
+  function onListen({ port, hostname }: { port: number; hostname: string }) {
+    const networkAddress = getNetworkAddress();
+    const protocol = useTls ? "https" : "http";
+    let message = `Listening on:\n- Local: ${protocol}://${hostname}:${port}`;
+    if (networkAddress) {
```
iuioiua, line 813:
```suggestion
    if (networkAddress && DENO_DEPLOYMENT_ID === undefined) {
```
So nothing is printed in Deno Deploy.

iuioiua (changes requested):
Testing of this script can be done another time. I created #4605 to track.

http/file_server_test.ts
```diff
@@ -970,3 +972,45 @@ Deno.test(
...
+  const port = await getAvailablePort();
+  const cmd = new Deno.Command(Deno.execPath(), {
+    args: [
+      "run",
+      "--no-lock",
+      "--quiet",
+      "--allow-net",
+      "--allow-read",
+      "--allow-sys=networkInterfaces",
+      "--config",
```
iuioiua, line 986: Reduces flakiness (ref #4168):
```suggestion
      "--config",
      "--no-lock",
```

iuioiua (changes requested):

http/file_server_test.ts
kt3k: `--no-lock` is at the above

http/file_server_test.ts
```diff
@@ -975,42 +975,87 @@ Deno.test(
...
+    "--allow-read",
+    "--allow-sys=networkInterfaces",
+    "http/file_server.ts",
+    "--port",
+    `${port}`,
+  ]);
+  const output = await readUntilMatch(process.stdout, "Network:");
+  const networkAdress = Deno.networkInterfaces().find((i) =>
+    i.family === "IPv4" && !i.address.startsWith("127")
+  )?.address;
```
iuioiua, line 989: It seems like this should be used as the implementation for `getNetworkAddress()`.

iuioiua (approved):

PR #4607, opened 2024-04-19
chore: add `lint:docs` task

iuioiua:
Definitively provides confirmation that `deno doc --lint <entrypoint>` passes.

This'll do for now. Later on we can come up with a more elaborate solution that intelligently checks the entire codebase.

Towards #3764


kt3k (approved):
LGTM

PR #4609, opened 2024-04-19
test(csv): fix type error in `csv/parse_test.ts`

### LINKED ISSUE #4606: `csv/parse_test.ts` has type errors, but they are not caught in CI
The file has type errors, but they are not caught in CI. (`deno task test` doesn't error at them)

```shellsession
$ deno check csv/parse_test.ts
Check file:///Users/kt3k/denoland/deno_std/csv/parse_test.ts
error: TS2345 [ERROR]: Argument of type 'undefined' is not assignable to parameter of type 'ParseOptions'.
      const parsed = parse("a\nb", undefined);
                                   ~~~~~~~~~
    at file:///Users/kt3k/denoland/deno_std/csv/parse_test.ts:828:36

TS2793 [ERROR]:     The call would have succeeded against this implementation, but implementation signatures of overloads are not externally visible.
    export function parse<const T extends ParseOptions>(
                    ~~~~~
        at file:///Users/kt3k/denoland/deno_std/csv/parse.ts:353:17

TS2344 [ERROR]: Type 'false' does not satisfy the constraint 'true'.
      type _ = AssertTrue<IsExact<typeof parsed, string[][]>>;
                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at file:///Users/kt3k/denoland/deno_std/csv/parse_test.ts:829:27

Found 2 errors.
```

kt3k:
This PR fixes the type error in `csv/parse_test.ts`. This also adds workaround in tasks. Now `deno task test` type check the codebase correctly, and we check the doc types with a separate task `deno task doc-test`.

closes #4606

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,9 @@ jobs:
           deno-version: ${{ matrix.deno }}
 
       - name: Run tests
-        run: deno task test
+        # TODO(kt3k): remove doc-test task when the below issue resolved
+        # https://github.com/denoland/deno/issues/23430
+        run: deno task test && deno task doc-test
 
       - name: Run tests (simulate Deno 2)
         if: matrix.deno == 'canary'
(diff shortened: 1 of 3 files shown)


iuioiua (approved):
Frankly, I think this is two PRs in one.

PR #4612, opened 2024-04-19
chore: reword to use "package"

iuioiua:
As the Standard Library moves towards being a JSR-based project, which uses the "package" instead of "sub-module" or "module" to describe a group of APIs.


kt3k (approved):
LGTM

PR #4613, opened 2024-04-19
chore: cleanup old git submodule settings

iuioiua:
This codebase no longer contains any git submodules since the Node compatibility layer moved to the runtime repo.


kt3k (approved):
LGTM

PR #4620, opened 2024-04-21
test(crypto): improve test coverage

mbhrznr:
pushes code coverage of the `crypto` sub-module closer to 100%.

currently the following files aren't at 100%:
- `_wasm`:
  - personally i'm just too unfamiliar w/ what's going on.
  - tried my best to add test cases but nothing seemed to work, so reverted what i had going.
- `crypto`:
  - due to the way `webCrypto` is bound to `globalThis.crypto` my usual strategy of `stub`bing didn't seem to work out.
  - would `webCrypto.subtle.digest` ever be falsy to begin with?
- `unstable_keystack`:
  - for `Symbol.for("nodejs.util.inspect.custom")` i tried to simply import `inspect` from `node:util` and compare the outcome, similar on how the test case for `Symbol.for("Deno.customInspect")` looked like. however this didn't change anything on cov.


crypto/crypto_test.ts
```diff
@@ -260,6 +266,35 @@ Deno.test("digest() keeps memory usage reasonable with many calls", async () =>
...
 
+Deno.test("digest() throws on invalid input", async () => {
+  const inputString = "taking the hobbits to isengard";
+  const inputBytes = new TextEncoder().encode(inputString);
+
+  await assertRejects(
+    () => stdCrypto.subtle.digest("BLAKE2B", {} as Iterable<never>),
+    TypeError,
+    "data must be a BufferSource or [Async]Iterable<BufferSource>",
+  );
```
iuioiua, line 277: Can you please add the same test for `stdCrypto.subtle.digestSync()`? See https://app.codecov.io/gh/denoland/deno_std/pull/4620/blob/crypto/crypto.ts#L290

crypto/crypto_test.ts
```diff
@@ -260,6 +266,35 @@ Deno.test("digest() keeps memory usage reasonable with many calls", async () =>
...
+        (async function* () {
+          yield undefined;
+        })() as AsyncIterable<BufferSource>,
+      ),
+    TypeError,
+    "data contained chunk of the wrong type",
+  );
+
+  await assertRejects(
+    () => stdCrypto.subtle.digest("BLAK" as DigestAlgorithmName, inputBytes),
```
iuioiua, line 292: Nit: ditto
```suggestion
    async () => await stdCrypto.subtle.digest("BLAK" as DigestAlgorithmName, inputBytes),
```

crypto/crypto_test.ts
```diff
@@ -260,6 +266,35 @@ Deno.test("digest() keeps memory usage reasonable with many calls", async () =>
   );
 });
 
+Deno.test("digest() throws on invalid input", async () => {
+  const inputString = "taking the hobbits to isengard";
+  const inputBytes = new TextEncoder().encode(inputString);
+
+  await assertRejects(
+    () => stdCrypto.subtle.digest("BLAKE2B", {} as Iterable<never>),
```
iuioiua, line 274: Nit: just to make it clearer to the reader that this is an async method.
```suggestion
    async () => await stdCrypto.subtle.digest("BLAKE2B", {} as Iterable<never>),
```

crypto/crypto_test.ts
iuioiua, line 1: Are you able to target https://app.codecov.io/gh/denoland/deno_std/pull/4620/blob/crypto/crypto.ts#L260 ?

crypto/unstable_keystack_test.ts
iuioiua, line 1: @kt3k, do you know how we could target https://app.codecov.io/gh/denoland/deno_std/pull/4620/blob/crypto/unstable_keystack.ts#L168 ?

iuioiua (changes requested):
Excellent work so far! I don't think we can do much about `crypto/_wasm/lib/deno_std_wasm_crypto.generated.mjs`. Maybe we should just ignore it in Codecov.

crypto/unstable_keystack_test.ts
kt3k: `util.inspect` of `node:util` should call that hook:

```js
import * as util from "node:util";
util.inspect(new KeyStack(["hi"]));
```

crypto/crypto_test.ts
mbhrznr: ah, yes. sorry about that. added cases for `stdCrypto.subtle.digestSync()`.

crypto/crypto_test.ts
mbhrznr: i've tried by stubbing `webCrypto` to be able to get into the remaining `else` block, but no luck so far.

crypto/unstable_keystack_test.ts
mbhrznr: thanks for the example!
interestingly using `node:util`'s `inspect` still seems to invoke `Deno`'s `inspect`.

crypto/crypto_test.ts
iuioiua: Actually, that last `else` clause might be redundant. I'll handle this in a separate PR.

iuioiua (approved):
Great stuff! Thank you.

PR #4623, opened 2024-04-22
chore: update Codecov settings

iuioiua:
Codecov isn't correctly updating commits (see [here](https://app.codecov.io/gh/denoland/deno_std/commits)). I've also noticed that the files changed are way off (see [here](https://app.codecov.io/gh/denoland/deno_std/pull/4620)). So Codecov may've been broken for 3 months. This change might fix that.


.github/workflows/ci.yml
```diff
@@ -60,9 +60,10 @@ jobs:
 
       - name: Upload coverage
         uses: codecov/codecov-action@v4
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
iuioiua, line 64: @kt3k, please ensure this token has been added ([documentation](https://docs.codecov.com/docs/adding-the-codecov-token#github-actions)).

.github/workflows/ci.yml
```diff
@@ -60,9 +60,10 @@ jobs:
...
         with:
           name: ${{ matrix.os }}-${{ matrix.deno }}
-          files: cov.lcov
```
iuioiua, line 65: Codecov targets this file by default anyway.

.github/workflows/ci.yml
kt3k: This is already set

kt3k (approved):
LGTM

PR #4626, opened 2024-04-23
refactor(crypto): cleanup `crypto.subtle.digest[Sync]()` logic

iuioiua:
This PR bumps test coverage of `crypto/crypto.ts` to 100% and aims to improve logic and readability.

This might interest you, @mbhrznr.

### DIFF
--- a/crypto/crypto.ts
+++ b/crypto/crypto.ts
@@ -114,6 +114,15 @@ import {
 
 export { DIGEST_ALGORITHM_NAMES, type DigestAlgorithmName };
 
+/** Digest algorithms supported by WebCrypto. */
+const WEB_CRYPTO_DIGEST_ALGORITHM_NAMES = [
+  "SHA-384",
+  "SHA-256",
+  "SHA-512",
+  // insecure (length-extendable and collidable):
+  "SHA-1",
+] as const;
+
 /**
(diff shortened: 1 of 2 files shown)


crypto/crypto.ts
```diff
@@ -204,92 +194,60 @@ const stdCrypto: StdCrypto = ((x) => x)({
...
-        (WEB_CRYPTO_DIGEST_ALGORITHM_NAMES as readonly string[]).includes(
-          name,
-        ) &&
-        // and the data is a single buffer,
-        bytes
-      ) {
-        return webCrypto.subtle.digest(algorithm, bytes);
-      } else if (DIGEST_ALGORITHM_NAMES.includes(name as DigestAlgorithmName)) {
-        if (bytes) {
+      if (DIGEST_ALGORITHM_NAMES.includes(name as DigestAlgorithmName)) {
```
kt3k, line 197: This seems to change the priority of the crypto implementation to be used.

Currently `digest` uses native Web Crypto `subtle.digest` if the given algorithm is one of 4 standard algorithms, but now it uses `stdCrypto.subtle.digestSync` (therefore `wasmCrypto.digest`). Is this change intentional?

iuioiua: Made some tweaks now.

kt3k (approved):
LGTM

PR #4630, opened 2024-04-23
fix(bytes): `equals()` works with subarray

### LINKED ISSUE #3603: `bytes/equals` incorrectly handles subarrays
```ts
import { assert, assertEquals } from "jsr:@std/assert"
import { equals } from "jsr:@std/bytes"

const a = new Uint8Array(1001).subarray(1)
const b = new Uint8Array(1000)

a[0] = 123
b[0] = 123

assertEquals(a, b) // ok
assert(equals(a, b)) // error; false negative
```
```ts
import { assert, assertNotEquals } from "jsr:@std/assert"
import { equals } from "jsr:@std/bytes"

const a = new Uint8Array(1001).subarray(1)
const b = new Uint8Array(1000)

a[999] = 123

assertNotEquals(a, b) // ok
assert(!equals(a, b)) // error; false positive
```

The `equals32Bit` implementation [does not account for subarrays]( https://github.com/denoland/deno_std/blob/main/bytes/equals.ts#L24-L25).

iuioiua:
Fixes #3603

### DIFF
--- a/bytes/equals.ts
+++ b/bytes/equals.ts
@@ -28,8 +28,8 @@ function equalsNaive(a: Uint8Array, b: Uint8Array): boolean {
 function equals32Bit(a: Uint8Array, b: Uint8Array): boolean {
   const len = a.length;
   const compressible = Math.floor(len / 4);
-  const compressedA = new Uint32Array(a.buffer, 0, compressible);
-  const compressedB = new Uint32Array(b.buffer, 0, compressible);
+  const compressedA = new Uint32Array(a, 0, compressible);
+  const compressedB = new Uint32Array(b, 0, compressible);
   for (let i = compressible * 4; i < len; i++) {
     if (a[i] !== b[i]) return false;
   }
--- a/bytes/equals_test.ts
+++ b/bytes/equals_test.ts
@@ -1,6 +1,6 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 import { equals } from "./equals.ts";
-import { assert } from "../assert/mod.ts";
+import { assert, assertEquals, assertNotEquals } from "../assert/mod.ts";
 
 Deno.test("equals()", () => {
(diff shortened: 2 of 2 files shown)


kt3k (approved):
LGTM 👍

tjjfvi:
This change fixes the bug but is really bad for performance:
```ts
import { equals } from "https://raw.githubusercontent.com/denoland/deno_std/0052a36/bytes/equals.ts"

const a999 = new Uint8Array(999);
const b999 = new Uint8Array(999);

const a1k = new Uint8Array(1000);
const b1k = new Uint8Array(1000);

Deno.bench("999", () => {
  equals(a999, b999)
})

Deno.bench("1k", () => {
  equals(a1k, b1k)
})
```
```
999           461.05 ns/iter   2,168,977.3 (454.72 ns … 511.36 ns) 460.35 ns 487.23 ns 511.36 ns
1k              1.37 µs/iter     728,273.7     (1.29 µs … 1.53 µs)   1.41 µs   1.53 µs   1.53 µs
```

The "optimized" `equals32Bit` function is 3x slower than the naive version.

As @0f-0b pointed out, the `new Uint32Array(a, ...)` is copying each element; `new Uint32Array(new Uint8Array([1, 2]))` is `new Uint32Array([1, 2])`. So, it's currently, taking two N-byte buffers, allocating two 4N-byte buffers, copying the N bytes to every fourth byte, and then checking if all 4N bytes are equivalent (75% of which are zeroes).

I don't think this function should be allocating memory at all (for one thing, it makes the function always O(N), even if the first byte differs). Instead, the `equals32Bit` path should be taken only when `a.byteOffset % 4 === b.byteOffset % 4`, and the uint32arrays should be constructed as subarrays.

iuioiua:
Thank you for the insight, @0f-0b and @tjjfvi. You guys are right. I'll submit a PR that reverts this change shortly.

iuioiua:
See #4632

PR #4632, opened 2024-04-23
revert(bytes): `equals()` works with subarray

iuioiua:
See https://github.com/denoland/deno_std/pull/4630#issuecomment-2072109408


iuioiua:
Closing in favour of #4633.

PR #4639, opened 2024-04-24
chore(http): remove `http/test.ts`

iuioiua:
This was added in https://github.com/denoland/deno/pull/7264 but it's premise is no longer true.


kt3k (approved):
LGTM

PR #4649, opened 2024-04-27
test(datetime): improve test coverage

mbhrznr:
pushes code coverage of the `datetime` sub-module closer to 100%.

currently only misses:
- a single branch within `dateTimeFormatter.format()`, which is tagged as `FIXME`
- a single branch within `dateTimeFormatter.partsToDate()`, where the local the format doesn't contain a day and the timezone isn't utc


iuioiua (approved):
Awesome work! Thank you.

Issue #4600, opened 2024-04-16
The Road to Stabilizing the Standard Library


iuioiua:
The Standard Library (`std`) aims to provide high-quality, reliable APIs for fearless use in JavaScript and TypeScript projects. For some time, how and when std would reach stability has been in question. Now, thanks to [JSR](https://jsr.io/) and [workspaces](https://github.com/denoland/deno/issues/4275) functionality in the Deno runtime, we have a path forward.

## What stabilization will look like

Currently, `std` acts as a single module with a single version, meaning a user may be inclined to upgrade, even if a release doesn't affect the sub-module they use.

After stabilization, new `std` packages will be exclusively hosted on JSR as separate packages with independent versions. Deno's workspaces functionality allows us to treat each top-level subdirectory as its own package and better adhere to [Semantic Versioning](https://semver.org/). Dependencies become more manageable by only requiring an upgrade for the packages you use, not the entire `std` module.

More importantly, independent versioning allows us to work on stabilizing packages asynchronously. We can stabilize more established packages while preparing other packages for stabilization. In other words, we can progressively deliver stable packages for the community sooner rather than later.

These changes will also mean we will release `std` packages independently of Deno runtime releases.

### Step 1 - Decouple `std` from the runtime (complete)

`deno` and `deno_std` are now effectively decoupled, so releasing them in tandem is not required. The remaining point of friction is `apiland_scripts`, which generates docs. We now have two separate GH Actions in `apiland_scripts` that can be run independently (`release CLI` and `release STD`). These actions can be run independently, and it's a matter of deciding if we do the "split". 

### Step 2 - Publish exclusively to JSR (complete)

At the moment, `std` is published to [deno.land/std](https://deno.land/std) as-is, then converted into a JSR-compatible form in CI and published to [JSR](https://jsr.io/@std) as separate packages under the `@std` scope. Once the previous step is complete, the codebase can be [permanently converted into its JSR-compatible form](https://github.com/denoland/deno_std/issues/4544), providing the granular package versioning for the next step.

Note: For the foreseeable future, [deno.land/std](https://deno.land/std) will remain online and accessible until the version before the JSR cutover is complete.

### Step 3 - Stabilize packages asynchronously (in progress)

Once `std` is entirely cut over to JSR and packages become more stable, we can begin stabilizing packages. A package will only be stabilized once strict criteria are met that guarantee its quality. These will require that each package is:

1. Reviewed and approved with sufficient consensus
2. Tested for maximum coverage
[...]

PR #4650, opened 2024-04-28
chore: switch to JSR-oriented codebase

### LINKED ISSUE #4544: Proposal: Migrate to JSR-oriented codebase
We currently [convert](https://github.com/denoland/deno_std/blob/e9a8db202d0b7ba0287f3c09b975bc3b1c21ff0e/_tools/convert_to_workspace.ts) the existing codebase to JSR oriented codebase and publish the exactly the same thing to both [JSR](https://jsr.io/@std) and https://deno.land/std. However this prevents us from doing split versioning of each submodule, and that is the blocker for [stabilization of std](https://github.com/denoland/deno_std/issues/3489) (We don't stabilize the entire standard libary, but we plan to stabilize it module by module gradually).

I think we should migrate to JSR-oriented code base in the near future, but there are some implications by doing that:
- We don't update deno.land/std actively anymore (except security fixes). New features go to `jsr.io/@std`.
  - This is because if we start the split versioning with JSR oriented codebase, it's difficult to convert it back to single-versioned std.
- The releases of CLI and STD are decoupled. CLI release and STD release happen independently.

I personally think we are fine with the above conditions, but does anyone have concerns or objections to it?

related #4600

kt3k:
closes #4544 

related #4600


badge.svg
iuioiua, line 1: This updates the link to https://jsr.io/@std.

deno.json
```diff
@@ -49,5 +47,46 @@
         "verbatim-module-syntax"
       ]
     }
-  }
+  },
+  "workspaces": [
```
iuioiua, line 51: These have been sorted alphabetically to be easier to follow. Ditto for `browser-compat.tsconfig.json`.

iuioiua (approved):
I've made some tweaks, including updating old references of https://deno.land/std.

Now, LGTM! Great to have this done! 🥳

PR #4651, opened 2024-04-29
feat(bytes): stabilize `@std/bytes`

iuioiua:


lucacasonato (approved):
I have looked through the docs, and they look good. All function signatures seem reasonable. No blockers from me.

kt3k:
One thing I noticed is that the 3rd params of `copy`, `includesNeedle`, `indexOfNeedle`, and `lastIndexOfNeedle` don't exactly follow [the style guide rule](https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object) (`Exported functions: max 2 args, put the rest into an options object.`).

They all use optional primitive type (`offset?: number`) as the 3rd param consistently. I personally think this is intentionally in this way, and we don't need to change them (instead we probably should update the style guide)

What do you think?

iuioiua:
Yeah, that's a great idea. I think options objects suite functions/methods with broader functionality and the potential to expand their functionality in the future. Non-object options suite functions/methods with narrower functionality and are unlikely to expand their functionality in the future. Perhaps this distinction should be noted.

lucacasonato:
Yes I agree. I don't think we should make them objects. Let's update the style guide

kt3k (approved):
LGTM

PR #4653, opened 2024-04-29
chore: update `check_circular_package_dependencies.ts` for workspaces

kt3k:
`_tools/check_circular_package_dependencies.ts` was disabled in #4650 

This PR re-enables it.

related #4544 #4600

### DIFF
--- a/_tools/check_circular_package_dependencies.ts
+++ b/_tools/check_circular_package_dependencies.ts
@@ -7,8 +7,6 @@ import {
 } from "@deno/graph";
 
 /**
- * TODO(kt3k): This script stopped working after JSR migration. Enable this script.
- *
  * Checks for circular dependencies in the std packages.
  *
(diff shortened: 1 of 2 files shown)


iuioiua:
I tested this locally by creating a two-way dependency between `@std/fs/copy` and `@std/path/basename` and got `Circular dependencies found: fs -> path -> fs`, which doesn't provide sufficient troubleshooting detail. Can we improve this a little?

kt3k:
What information would you suggest to add?

iuioiua:
Just a bit more specificity, such as the exact modules in question. E.g.
```
Circular dependencies found: fs/copy -> path/basename -> fs/copy
```

kt3k:
Sounds a good idea, but that requires fair amount of work. Sounds out of scope of this PR. Created a new issue https://github.com/denoland/deno_std/issues/4656

iuioiua (approved):

PR #4654, opened 2024-04-29
docs: update README for JSR

iuioiua:
This change trims some of the README, by relying more on JSR's usage instructions built-in to package pages. 

My thinking is that JSR already does a fine job at instructing the reader on how to use the package and further usage instructions simply aren't needed elsewhere. This should also act as more of a forcing function to improve our documentation.

I'd like to complete remove the packages table once https://github.com/jsr-io/jsr/issues/421 is merged, so there's no need to maintain such a list. We can instead have a note on each package's module documentation like so:
```
/**
 * This module does this and that.
 *
 * Note: this module is unstable and may change in the future.
 * @module
 */
```
and
```
/**
 * This module does this and that.
 *
 * Note: this module is settled and is unlikely to change in the future.
 * @module
 */
```

Supercedes #4652

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,58 +6,14 @@
 
 High-quality APIs for [Deno](https://deno.com/) and the web. Use fearlessly.
 
-> **NOTE:** The standard library is also available on JSR - check out the
-> [`@std` scope on JSR here](https://jsr.io/@std).
-
-## Get Started
-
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #4655, opened 2024-04-29
chore: exclude `testdata` directories from being published

iuioiua:
Avoids [some warnings](https://github.com/denoland/deno_std/actions/runs/8826423794/job/24238408524) we get when publishing.


kt3k:
I wonder if it make sense to publish test files without testdata..

iuioiua:
Maybe. Maybe not. Either way, I think we can explore that idea later.

PR #4657, opened 2024-04-29
chore: update `check_deprecation.ts`

kt3k:
We disabled `check_deprecation.ts` in #4650. This PR re-enables it by providing appropriate resolver to `@deno/doc`. Also shares the same resolver with `check_circular_package_dependencies.ts`.

### DIFF
--- a/_tools/utils.ts
+++ b/_tools/utils.ts
@@ -1,15 +1,13 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
-import { join } from "../path/join.ts";
-
 const workspaces = JSON.parse(await Deno.readTextFile("deno.json"))
   .workspaces as string[];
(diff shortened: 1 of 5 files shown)


_tools/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import { join } from "../path/join.ts";
+
+const workspaces = JSON.parse(await Deno.readTextFile("deno.json"))
+  .workspaces as string[];
+// deno-lint-ignore no-explicit-any
+const denoConfig = {} as Record<string, any>;
+for (const workspace of workspaces) {
+  const config = JSON.parse(
+    await Deno.readTextFile(join(workspace, "deno.json")),
```
iuioiua, line 11: Nit: you could dynamically import as JSON module.

iuioiua (approved):

PR #4660, opened 2024-04-30
chore: skip yanked versions

kt3k:
These versions are used & yanked in JSR by accident. Let's move forward the versions to skip them in the next release.

### DIFF
--- a/assert/deno.json
+++ b/assert/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "@std/assert",
-  "version": "0.224.0",
+  "version": "0.225.0",
   "exports": {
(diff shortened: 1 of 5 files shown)


iuioiua (approved):
Skipping versions with changes... Seems odd to do. I defer to your decision.

iuioiua (changes requested):
Wait, aren't these versions supposed to be skipped?

kt3k:
Setting these versions will result in skipping these versions. It's for the version updater script ( https://github.com/denoland/bump_workspaces ) to recognize these versions as latest versions

iuioiua (approved):

PR #4662, opened 2024-05-01
fix(cli): spinner support NO_COLOR

ryuapp:
This may change `fmt/colors` in the future, but please support this first.


kt3k (approved):
Confirmed this works locally.

Screenshot 2024-05-02 at 19 15 20

LGTM

Issue #4647, opened 2024-04-26
proposal: stabilize `std/collections`


iuioiua:
Package documentation: https://jsr.io/@std/collections/doc

Progress:

- [x] 100% documented, passing deno doc --lint checks and adhering to the [documentation guidelines](https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md#documentation).
- [x] Maximum possible test coverage
- [x] No open issues or pull requests that might lead to breaking changes. For example, issues that suggest new non-breaking features are fine to exist at stabilization.
- [x] Publish `1.0.0-rc.1`
- [ ] Wait 1 month for feedback
- [ ] Publish `1.0.0`

Blockers:
- [x] #4664
- [x] #3384
- [x] ~~#1203~~ (update: likely not a blocker)
- [x] #1126

PR #4664, opened 2024-05-02
docs(collections): complete documentation

iuioiua:
Towards #4647


collections/join_to_string.ts
```diff
@@ -3,22 +3,56 @@
...
+   */
   prefix?: string;
+  /**
+   * The string to use as a suffix for the resulting string.
+   *
+   * @default {""}
+   */
   suffix?: string;
+  /**
+   * The maximum number of elements to append.
```
kt3k, line 25: Let's describe what `-1` means here?

collections/min_of.ts
```diff
@@ -6,7 +6,15 @@
  * returns the min value of all elements. If an empty array is provided the
  * function will return undefined.
  *
- * @example
+ * @template T The type of the elements in the array.
+ *
+ * @param array The array to find the minimum element in.
+ * @param selector The function to get the value to compare from each element.
+ *
+ * @returns The first element that is the smallest value of the given function
```
kt3k, line 14: ditto. `The first element that is` part seems unnecessary

PR #4666, opened 2024-05-02
refactor(collections): improve implementation signatures

iuioiua:
This PR simplifies the implementation signatures of some functions within `std/collections`.

Previously, `selector` parameters were written in a manner that was a little more complicated than needed, also resulting in having to rely on `ReturnType<typeof selector>` within function bodies. The return type of `selector` parameters are now defined within a second type argument, `S`.

This change is non-breaking as it adds a new type argument. The aim of this change is to make implementation signatures of these functions easier to understand, and hopefully avoid the need for overload signatures for these functions.

### DIFF
--- a/collections/max_by.ts
+++ b/collections/max_by.ts
@@ -91,13 +91,9 @@ export function maxBy<T>(
   array: Iterable<T>,
   selector: (el: T) => Date,
 ): T | undefined;
-export function maxBy<T>(
+export function maxBy<T, S extends (number | string | bigint | Date)>(
   array: Iterable<T>,
-  selector:
-    | ((el: T) => number)
-    | ((el: T) => string)
-    | ((el: T) => bigint)
-    | ((el: T) => Date),
+  selector: (el: T) => S,
 ): T | undefined {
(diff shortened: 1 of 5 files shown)


collections/max_by.ts
```diff
@@ -91,13 +91,9 @@ export function maxBy<T>(
...
 ): T | undefined;
-export function maxBy<T>(
+export function maxBy<T, S extends (number | string | bigint | Date)>(
   array: Iterable<T>,
-  selector:
-    | ((el: T) => number)
-    | ((el: T) => string)
-    | ((el: T) => bigint)
-    | ((el: T) => Date),
+  selector: (el: T) => S,
```
kt3k, line 96: This allows mixed return type 'selector' like `() => Math.random() > 0.5 ? "abc" : 1`, which we don't want to allow.

kt3k: Other signatures below have the same problems

iuioiua: Hm... This is true. Closing.

PR #4667, opened 2024-05-02
chore: add `@template` JSDoc tag check

iuioiua:
This change adds the ability for the doc checker to ensure that all functions have a `@template` JSDoc tag for each type argument.


kt3k (approved):
LGTM

PR #4675, opened 2024-05-06
chore: remove doc test workaround

kt3k:
We added doc-test workaround in https://github.com/denoland/deno_std/issues/4606 because of the issue https://github.com/denoland/deno/issues/23430

This PR removes that workaround because the issue https://github.com/denoland/deno/issues/23430 is now resolved in main.


iuioiua (approved):
Nice!

PR #4677, opened 2024-05-06
feat(net): `getNetworkAddress()`

iuioiua:
This seems like a useful tool to have.


net/get_network_address.ts
```diff
@@ -0,0 +1,48 @@
...
+  family: Deno.NetworkInterfaceInfo["family"] = "IPv4",
+): string | undefined {
+  return Deno.networkInterfaces()
+    .find((i) =>
+      i.family === family &&
+      (family === "IPv4"
+        // Cannot lie within 127.0.0.0/8
+        ? !i.address.startsWith("127")
+        // Cannot lie within ::1/128 or fe80::/10
+        : (!i.address.startsWith("::1")) || !i.address.startsWith("fe80::"))
```
kt3k, line 45: Deno seems checking with the condition `i.address === "::1" || i.address === "fe80::1"` https://github.com/denoland/deno/blob/cbb78e138ffbfc92ca64f2cce56dd5c629e4f142/ext/node/polyfills/os.ts#L236 for simulating `internal`

Also according to `npm:network-address` module, it seems also necessary to check `scopeid === 0` for IPv6 family https://github.com/mafintosh/network-address/blob/a4704e359a9b90ea60a50076215e901b3b5c9c8e/index.js#L7

iuioiua: Further context: https://github.com/mafintosh/network-address/pull/3

net/get_network_address.ts
```diff
@@ -0,0 +1,49 @@
...
+  family: Deno.NetworkInterfaceInfo["family"] = "IPv4",
+): string | undefined {
+  return Deno.networkInterfaces()
+    .find((i) =>
+      i.family === family &&
+      (family === "IPv4"
+        // Cannot lie within 127.0.0.0/8
+        ? !i.address.startsWith("127")
+        // Cannot lie within ::1/128 or fe80::/10
+        : (!i.address.startsWith("::1")) || !i.address.startsWith("fe80::") &&
```
kt3k, line 45: If we follow the CLI compat layer, shouldn't this be `i.address !== "::1" && i.address !== "fe80::"` instead of `startWith` check?

I feel we need test cases with `stub`'ed `Deno.networkInterfaces` to check these algorithms are correct.

net/get_network_address.ts
```diff
@@ -0,0 +1,49 @@
...
+  family: Deno.NetworkInterfaceInfo["family"] = "IPv4",
+): string | undefined {
+  return Deno.networkInterfaces()
+    .find((i) =>
+      i.family === family &&
+      (family === "IPv4"
+        // Cannot lie within 127.0.0.0/8
+        ? !i.address.startsWith("127")
+        // Cannot be ::1 (loopback) or lie within fe80::/10 (link-local)
+        : i.address !== "::1" || !i.address.startsWith("fe80::") &&
```
kt3k, line 45:
```suggestion
        : i.address !== "::1" && !i.address !== "fe80::" &&
```

iuioiua:
Sorry, my understanding of some of the IPv6 addressing was incorrect. Corrected now, and scope ID isn't needed.

kt3k:
> and scope ID isn't needed.

Why is the checkc of scopeid unnecessary?

`npm:network-address` seems checking it? https://github.com/mafintosh/network-address/pull/3

kt3k (approved):
LGTM

PR #4680, opened 2024-05-07
feat(log): make `FileHandler` and `RotatingFileHandler` buffer size configurable

### LINKED ISSUE #4674: suggestion: make `RotatingFileHandler` buffer size configurable
By @fushihara in https://github.com/denoland/deno_std/issues/3676#issuecomment-2094898479:

> In the current source code, the buffer size seems to be fixed at 4096 bytes in PAGE_SIZE.
> My program has a low load and logs only one case every few seconds, so I would like to reduce the buffer size.
> I see no reason why this value cannot be specified in the constructor, so I hope it will be changeable.
> 
> But I can't wait, so I created my own inherited class.
> I ran it only once, but I don't think there is any problem.
> 
> ```ts
> class RotatingFileHandler2 extends RotatingFileHandler {
>   constructor(levelName: ConstructorParameters<typeof RotatingFileHandler>[0], options: ConstructorParameters<typeof RotatingFileHandler>[1]) {
>     super(levelName, options);
>     this._buf = new Uint8Array(1);
>   }
> }
> ```
> 
> @markthree 
> I understand that this is a unique situation,
> but when I am debugging with vscode and force quit the deno with the stop button, this logic does not seem to work.
>

iuioiua:
Closes #4674

### DIFF
--- a/log/file_handler.ts
+++ b/log/file_handler.ts
@@ -4,12 +4,17 @@ import type { LogRecord } from "./logger.ts";
 import { BaseHandler, type BaseHandlerOptions } from "./base_handler.ts";
 import { writeAllSync } from "@std/io/write-all";
 
-const PAGE_SIZE = 4096;
 export type LogMode = "a" | "w" | "x";
 
 export interface FileHandlerOptions extends BaseHandlerOptions {
   filename: string;
   mode?: LogMode;
+  /**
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #4681, opened 2024-05-07
refactor(collections): minor cleanups

iuioiua:
These are non-breaking but either improve implementation logic and/or readability.

### DIFF
--- a/collections/chunk.ts
+++ b/collections/chunk.ts
@@ -40,24 +40,17 @@
 export function chunk<T>(array: readonly T[], size: number): T[][] {
   if (size <= 0 || !Number.isInteger(size)) {
     throw new Error(
-      `Expected size to be an integer greater than 0 but found ${size}`,
+      `Expected size to be a positive integer, but found ${size}`,
     );
   }
 
-  if (array.length === 0) {
-    return [];
+  const result: T[][] = [];
+  let index = 0;
+
+  while (index < array.length) {
+    result.push(array.slice(index, index + size));
+    index += size;
   }
 
-  const ret = Array.from<T[]>({ length: Math.ceil(array.length / size) });
-  let readIndex = 0;
-  let writeIndex = 0;
-
-  while (readIndex < array.length) {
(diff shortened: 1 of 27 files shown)


collections/chunk.ts
```diff
@@ -40,24 +40,17 @@
...
-    ret[writeIndex] = array.slice(readIndex, readIndex + size);
+  const result: T[][] = [];
+  let index = 0;
 
-    writeIndex += 1;
-    readIndex += size;
+  while (index < array.length) {
+    result.push(array.slice(index, index + size));
+    index += size;
   }
```
kt3k, line 53: This changes how we allocate the result array. Are you sure this doesn't cause perf degradation?

kt3k:
Left one question. Other part looks good to me

collections/chunk.ts
iuioiua: The difference is negligible.
```ts
import { chunk } from "jsr:@std/collections@0.224.0/chunk";
import { chunk as chunk2 } from "./collections/chunk.ts";

function createArray(length: number) {
  return Array.from({ length }).fill(0);
}

Deno.bench("chunk()", (b) => {
  const array = createArray(1_000_000);
  b.start();
  chunk(array, 3);
  b.end();
});

Deno.bench("chunk2()", (b) => {
  const array = createArray(1_000_000);
  b.start();
  chunk2(array, 3);
  b.end();
});
```
```
asher@Ashers-MacBook-Air deno_std % deno bench bench.ts
Check file:///Users/asher/GitHub/deno_std/bench.ts
cpu: Apple M2
runtime: deno 1.43.1+4ab68df (aarch64-apple-darwin)

chunk()        20.43 ms/iter          48.9   (11.23 ms … 33.67 ms) 25.44 ms 33.67 ms 33.67 ms
chunk2()       20.19 ms/iter          49.5   (11.58 ms … 32.57 ms) 25.34 ms 32.57 ms 32.57 ms
```

kt3k (approved):
LGTM. Nice clean ups!

PR #4682, opened 2024-05-07
fix(collections): correct error class when `chunk()` throws

iuioiua:
This is a more fitting error class for when `size` is incorrectly defined.


kt3k (approved):
LGTM

I'd rather tag this `fix` as this includes a user-visible change

PR #4693, opened 2024-05-08
docs(bytes): trim module documentation

iuioiua:
I realise the previous documentation was just the symbol documentation, duplicated, and provided little extra value compared to just looking at the symbol documentation. This approach is also more viable to apply codebase-wide.


kt3k (approved):
Sounds reasonable to me. These docs and examples are given in each API docs. No need to duplicate them in `@module` docs.

PR #4694, opened 2024-05-08
deprecation: move `@std/fmt/colors` to `@std/cli/styles`

### LINKED ISSUE #4676: suggestion: move `std/fmt/colors` to `std/cli/styles`
I suggest moving/renaming `std/fmt/colors` to `std/cli/styles` for the following reasons:
1. The module contains functionality for more than just colors. I.e. `italic()` and `bold()`. Hence, the rename to `styles`.
2. The module's functionality is only concerned with console output. Hence, the move to `std/cli`.
This new name would make far more sense than it currently does and be much more accurate.

This is a continuation of the conversation happening in #4326.

iuioiua:
Closes #4676


iuioiua:
I'm going to take another approach with this move.

PR #4698, opened 2024-05-09
refactor(internal): cleanup and restructure

iuioiua:
This change simplifies the `DiffType` and moves functions out into their own respective files with the aim of improving maintainability, as `internal/diff.ts` is quite a busy file. Further improvements should come after this PR.


internal/diff_str.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { DiffResult } from "./_types.ts";
+import { _internals } from "./_internals.ts";
```
kt3k, line 3: Looks like this can be replaced by importing `diff` from `diff.ts`? (and we can remove `_internals.ts`)

internal/deno.json
```diff
@@ -3,7 +3,6 @@
   "version": "0.225.0",
   "exports": {
     ".": "./mod.ts",
-    "./diff": "./diff.ts",
```
kt3k, line 6: now `diff`, `diffstr`, and `buildMessage` are only available from `@std/internal` (root). I think that would be inconvenient in the future when we added completely unrelated APIs to `@std/internal`

internal/deno.json
```diff
@@ -3,7 +3,9 @@
...
+    "./build-message": "./build_message.ts",
+    "./diff-str": "./diff_str.ts",
+    "./diff.ts": "./diff.ts",
```
kt3k, line 8:
```suggestion
    "./diff": "./diff.ts",
```

internal/diff_str.ts
```diff
@@ -0,0 +1,144 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { DiffResult } from "./_types.ts";
+import { _internals } from "./diff.ts";
```
kt3k, line 3: Is `_internals` object necessary? It seems only used for accessing to `diff` function like `_internals.diff()`. Is it possible to simply import `diff` instead of `_internals`?

iuioiua: It's needed to mock a throw in `diff()` to cater for `[Cannot display]` output in tests. If #4700 and #4701 go ahead, it'll no longer be needed.

kt3k: Ah, ok. Then let's keep it for now

kt3k (approved):
LGTM

Issue #4699, opened 2024-05-09
todo: investigate use of `CAN_NOT_DISPLAY` constant


iuioiua:
We need to determine whether the use of this constant is actually warranted. It's been found in:
- [x] `assert/assert_equals.ts`
- [x] `assert/assert_not_equals.ts`
- [x] `assert/assert_strict_equals.ts`
- [x] `expect/_build_message.ts`

PR #4700, opened 2024-05-09
fix(expect): don't swallow the original error while creating assertion error

iuioiua:
This change removes lines that seem unreachable. This also eliminates a dependency on `fmt/colors.ts`.

Towards #4699

### DIFF
--- a/expect/_build_message.ts
+++ b/expect/_build_message.ts
@@ -1,7 +1,5 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
-import { red } from "@std/fmt/colors";
-import { CAN_NOT_DISPLAY } from "./_constants.ts";
 import { buildMessage, diff, diffstr, format } from "@std/internal";
 import type { EqualOptions } from "./_types.ts";
 
@@ -9,6 +7,10 @@ type EqualErrorMessageOptions = Pick<
   EqualOptions,
   "formatter" | "msg"
 >;
+
+function isString(value: unknown): value is string {
+  return typeof value === "string";
+}
 
 export function buildEqualErrorMessage<T>(
   actual: T,
@@ -22,17 +24,12 @@ export function buildEqualErrorMessage<T>(
 
   let message = `Values are not equal${msgSuffix}`;
 
-  try {
-    const stringDiff = (typeof actual === "string") &&
-      (typeof expected === "string");
-    const diffResult = stringDiff
-      ? diffstr(actual as string, expected as string)
-      : diff(actualString.split("\n"), expectedString.split("\n"));
-    const diffMsg = buildMessage(diffResult, { stringDiff }).join("\n");
-    message = `${message}\n${diffMsg}`;
-  } catch {
-    message = `${message}\n${red(CAN_NOT_DISPLAY)} + \n\n`;
-  }
+  const stringDiff = isString(actual) && isString(expected);
+  const diffResult = stringDiff
+    ? diffstr(actual, expected)
+    : diff(actualString.split("\n"), expectedString.split("\n"));
+  const diffMsg = buildMessage(diffResult, { stringDiff }).join("\n");
+  message = `${message}\n${diffMsg}`;
 
   return message;
 }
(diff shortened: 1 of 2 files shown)


expect/_build_message.ts
```diff
@@ -10,6 +8,10 @@ type EqualErrorMessageOptions = Pick<
   "formatter" | "msg"
 >;
 
+function isString(value: unknown): value is string {
```
iuioiua, line 11: Pass-by cleanup.

expect/_build_message.ts
```diff
@@ -22,17 +24,12 @@ export function buildEqualErrorMessage<T>(
...
-  try {
-    const stringDiff = (typeof actual === "string") &&
-      (typeof expected === "string");
-    const diffResult = stringDiff
-      ? diffstr(actual as string, expected as string)
-      : diff(actualString.split("\n"), expectedString.split("\n"));
-    const diffMsg = buildMessage(diffResult, { stringDiff }).join("\n");
-    message = `${message}\n${diffMsg}`;
-  } catch {
-    message = `${message}\n${red(CAN_NOT_DISPLAY)} + \n\n`;
```
iuioiua, line 34: It's probably best that the underlying functions, such as `diffstr()`, `diff()`, and `buildMessage()`, throw naturally. This catch might actually impede the ability to troubleshoot.

expect/_build_message.ts
```diff
@@ -43,19 +40,8 @@ export function buildNotEqualErrorMessage<T>(
...
-  } catch {
-    actualString = CAN_NOT_DISPLAY;
-  }
-  try {
-    expectedString = String(expected);
-  } catch {
-    expectedString = CAN_NOT_DISPLAY;
-  }
+  const actualString = String(actual);
+  const expectedString = String(expected);
```
iuioiua, line 44: [`String()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/String) as a function is extremely versatile. I can't think of any objects that would throw when passed as input to it unless it's something super contrived like:
```ts
const problematicObject = {
  toString() {
    throw new Error("Cannot convert to string");
  }
};
```

kt3k: > ```ts
> const problematicObject = {
>   toString() {
>     throw new Error("Cannot convert to string");
>   }
> };
> ```

So the `catch` block is not unreachable with the above example. I don't think we should remove it

iuioiua: `String()` would only fail if an object deliberately throws on `.toString()`. I don't know of any other cases in which it'd throw. If `String()` does throw, does changing the message to `[Cannot display]` help with troubleshooting?

kt3k: > `String()` would only fail if an object deliberately throws on `.toString()`.

I think that can be also caused by mistake. For example, some might define `toString` like:

```ts
  toString() {
    return JSON.stringify(this.someProperty);
  }
```

This throws when someProperty includes BigInt or circular reference ([MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify#exceptions))

kt3k: > If `String()` does throw, does changing the message to `[Cannot display]` help with troubleshooting?

`[Cannot display]` is actually not helpful. I think we should instead append the thrown error message after it like `Cannot display because of the error: ${originalErrorMessage}`

iuioiua: Does changing the message to `[Cannot display]` help in troubleshooting in the case you provided, or is it better to throw?

iuioiua: Why not just let it throw?

kt3k: Sounds a fair point. Probably it's a little better to let it throw as is than the current state (The current `cannot display` message only hides what the actual error is).

kt3k:
Updated the PR title. WDYT?

kt3k (approved):
LGTM 👍

iuioiua:
SGTM

PR #4701, opened 2024-05-09
fix(assert): don't swallow the original error while creating assertion error

iuioiua:
This change removes lines that seem unreachable and reduces the dependency on `fmt/colors.ts`. The exact reasons for the changes are the same as in #4700.

Towards #4699


kt3k (approved):
LGTM

PR #4710, opened 2024-05-10
feat(collections): add `invert()` and `invertBy()`

mbhrznr:
first feature on `deno_std` 🎉 

tackles https://github.com/denoland/deno_std/issues/4686 and creates first versions for the `invert` and `invertBy` functions.

the implementation is inspired by the [lodash](https://lodash.com)'s  [invert](https://lodash.com/docs/4.17.15#invert) and [invertBy](https://lodash.com/docs/4.17.15#invertBy) counterparts.

feedback is highly appreciated.

PR #4712, opened 2024-05-10
refactor(collections): sort exports in `mod.ts`

mbhrznr:
during the implementation of https://github.com/denoland/deno_std/pull/4710 i noticed that the exports were unsorted.
this pr aligns `mod.ts`' exports statements with the order of the `exports` property in the corresponding `deno.json` file.


kt3k (approved):
LGTM. Thanks!

PR #4715, opened 2024-05-11
fix(assert): handle `assertStrictEquals(-0, +0)` correctly

### LINKED ISSUE #4714: assertStrictEquals(-0, +0) should not throw
**Describe the Bug**
According to the _[Strict Equality Comparison Algorithm](https://262.ecma-international.org/6.0/#sec-strict-equality-comparison)_ of the Javascript standard, -0 and +0 are considered equal.

Assuming that `assertStrictEquals` is meant to assert equality in the sense of `actual === expected`, it should not throw when passing -0 and +0 for `actual` and `expected`.

**Steps to Reproduce**
```ts
import { assert, assertStrictEquals } from "https://deno.land/std@0.224.0/assert/mod.ts";

assert(-0 === +0); // passes
assertStrictEquals(-0, +0); // fails
```
**Expected Behavior**
`assertStrictEquals(-0, +0)` should not throw.

**Environment**
- OS: macOS
- deno version: 1.43.2
- std version: 0.224

rherrmann:
Closes #4714

### DIFF
--- a/assert/assert_strict_equals.ts
+++ b/assert/assert_strict_equals.ts
@@ -26,10 +26,15 @@ export function assertStrictEquals<T>(
   expected: T,
   msg?: string,
 ): asserts actual is T {
-  if (Object.is(actual, expected)) {
+  if (actual === expected) {
     return;
   }
-
+  if (
+    typeof actual === "number" && typeof expected === "number" &&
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Thanks!

PR #4716, opened 2024-05-13
refactor(assert): vendor code from `fmt/colors.ts`

iuioiua:
Towards #4717


iuioiua:
Frankly, I'd prefer we minimize our use of colors as much as possible.

kt3k:
Maybe this only makes sense when we reached a consensus on #4676

iuioiua:
How does that issue affect this PR?

kt3k:
If we don't move `fmt/colors` (and stabilize it as is), then we don't need to do this

iuioiua:
Stabilizing `std/fmt` blocks the stabilization of `std/internal` and `std/assert`, and stabilizing `std/fmt` is not yet on the radar. I think we should proceed with this PR.

kt3k:
Ok. Makes sense

kt3k (approved):
LGTM

PR #4718, opened 2024-05-13
chore(assert): revert "handle `assertStrictEquals(-0, +0)` correctly"

iuioiua:
See https://github.com/denoland/deno_std/issues/4714#issuecomment-2106423300


kt3k (approved):
LGTM

PR #4722, opened 2024-05-14
docs: "Frequently Asked Questions" page

iuioiua:
First pass. Other questions that should be answered:
1. Why is the Standard Library versioned differently to the Deno runtime?
2. What are the compatibility guarantees for the Standard Library?

@kt3k, are you able to handle these?

### DIFF
--- a/.github/FAQ.md
+++ b/.github/FAQ.md
@@ -48,13 +48,13 @@ range in the import specifier.
 E.g. If the latest major version of a package is 1, it should be imported as
 follows:
 
-```ts
+```ts, ignore
 import { bar } from "jsr:@std/foo@1/bar";
 ```
 
 The same goes if the latest major version of a package is 0:
 
-```ts
+```ts, ignore
 import { bar } from "jsr:@std/foo@0/bar";
 ```
 
@@ -89,7 +89,7 @@ by pinning the package version.
 E.g. Say `bar()` is deprecated in `@std/foo@0` and removed in `@std/foo@1`. Pin
 the import specifier to `@std/foo@0` to continued use:
 
-```ts
+```ts, ignore
 import { bar } from "jsr:@std/foo@0/bar";
 ```
 
--- a/.github/FAQ.md
+++ b/.github/FAQ.md
@@ -49,13 +49,13 @@ E.g. If the latest major version of a package is 1, it should be imported as
 follows:
 
 ```ts, ignore
-import { bar } from "jsr:@std/foo@1/bar";
+import { bar } from "jsr:@std/foo@^1";
 ```
 
 The same goes if the latest major version of a package is 0:
 
 ```ts, ignore
-import { bar } from "jsr:@std/foo@0/bar";
+import { bar } from "jsr:@std/foo@^0";
 ```
 
 This approach allows for bug fixes and new features, while avoiding breaking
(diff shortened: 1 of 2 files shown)


.github/FAQ.md
```diff
@@ -0,0 +1,105 @@
...
+We recommend using the latest version of a given package.
+
+Thanks to JSR, you can constrain dependency versions by defining the SemVer
+range in the import specifier.
+
+E.g. If the latest major version of a package is 1, it should be imported as
+follows:
+
+```ts, ignore
+import { bar } from "jsr:@std/foo@1/bar";
```
kt3k, line 52: `deno add <pkg>` adds import map entry with the form of `"@std/foo": "jsr:@std/foo@^1.2.3"`. The above example seems slightly inconsistent with it to me

.github/FAQ.md
```diff
@@ -0,0 +1,105 @@
...
+1. It's been moved to a more appropriate location. E.g. To another package.
+1. It's been renamed more appropriate. E.g. To a name that more clearly
+   describes its behavior.
+1. It's been deemed no longer fitting to the Standard Library.
+
+## When is a deprecated API removed?
+
+In almost all cases, a deprecated API is removed in the next major version after
+deprecation. Rarely, a deprecated API is removed in the a minor version as a
+small breaking change. This is to minimize breaking changes for users.
```
kt3k, line 82: > Rarely, a deprecated API is removed in the a minor version as a
small breaking change.

I personally doubt we would ever do this intentionally. Also I find the term 'small breaking change' strange. What makes a breaking change 'small'?

kt3k:
> .github/FAQ.md

I wonder this is a right place to document this information. How about adding these sections in "Standard Library" page of [deno-docs](https://github.com/denoland/deno-docs) https://docs.deno.com/runtime/manual/basics/standard_library ?

.github/FAQ.md
iuioiua: Small breaking changes are those with minimal impact. Either way, I've removed that section now.

.github/FAQ.md
iuioiua: Sorry, what are you suggesting? I changed it to `jsr:@std/foo@^1`. How's that?

iuioiua:
> > .github/FAQ.md
> 
> I wonder this is a right place to document this information. How about adding these sections in "Standard Library" page of [deno-docs](https://github.com/denoland/deno-docs) https://docs.deno.com/runtime/manual/basics/standard_library ?

Almost all information on the Standard Library should be contained within this repo. It doesn't sense to disperse this information elsewhere.

kt3k:
> Almost all information on the Standard Library should be contained within this repo. It doesn't sense to disperse this information elsewhere.

Maybe this topic is out of scope of this PR, but the issue with `.github/FAQ.md` is that this file doesn't appear anywhere in jsr.io. I think there's high chance that users miss this file.

kt3k (approved):
LGTM

iuioiua:
> > Almost all information on the Standard Library should be contained within this repo. It doesn't sense to disperse this information elsewhere.
> 
> Maybe this topic is out of scope of this PR, but the issue with `.github/FAQ.md` is that this file doesn't appear anywhere in jsr.io. I think there's high chance that users miss this file.

Hopfully, https://github.com/jsr-io/jsr/issues/507 fixes that. I would be happy to discuss it elsewhere.

Issue #4500, opened 2024-03-17
suggestion: `concatReadableStreams()`


mfulton26:
**Is your feature request related to a problem? Please describe.**

I want to combine multiple `ReadableStream` instances into a single `ReadableStream` instance where each stream is read fully in sequence.

**Describe the solution you'd like**

A function to go along with `earlyZipReadableStreams`, `mergeReadableStreams`, and `zipReadableStreams` but processes streams in order, one at a time.

e.g.

```ts
function concatReadableStreams<T>(
  ...streams: ReadableStream<T>[]
): ReadableStream<T> {
  return new ReadableStream({
    async start(controller) {
      try {
        for (const stream of streams) {
          for await (const chunk of stream) {
            controller.enqueue(chunk);
          }
        }
        controller.close();
      } catch (e) {
        controller.error(e);
      }
    },
  });
}
```

**Describe alternatives you've considered**

I wonder if this isn't already included because there's some reason not to do this with streams or that there's already a simple way to do this built into the JS Streams API that I have yet to find.

mfulton26:
While composing multiple readable streams into a single one I found that in some cases I need to generate the streams asynchronously. This lead me to a different implementation with a signature more like `ReadableStream.from()` instead of like `mergeReadableStreams`:

```ts
type AnyIterable<T> = AsyncIterable<T> | Iterable<T | PromiseLike<T>>;

function concatReadableStreams<T>(
  streams: AnyIterable<ReadableStream<T>>,
): ReadableStream<T> {
  return new ReadableStream({
    async start(controller) {
      try {
        if (Symbol.asyncIterator in streams) {
          for await (const stream of streams) {
            for await (const chunk of stream) {
              controller.enqueue(chunk);
            }
          }
        } else {
          // use `for...of` instead of `for await...of` to call sync generator `finally` blocks
          for (const stream of streams) {
            for await (const chunk of await stream) {
              controller.enqueue(chunk);
            }
          }
        }
        controller.close();
      } catch (e) {
        controller.error(e);
      }
    },
  });
}
```

```diff
@@ -1,12 +1,23 @@
+type AnyIterable<T> = AsyncIterable<T> | Iterable<T | PromiseLike<T>>;
+
 function concatReadableStreams<T>(
-  ...streams: ReadableStream<T>[]
+  streams: AnyIterable<ReadableStream<T>>,
 ): ReadableStream<T> {
   return new ReadableStream({
     async start(controller) {
       try {
[...]

PR #4723, opened 2024-05-14
feat(streams): `ConcatStreams()`

iuioiua:
Implements #4500

### DIFF
--- a/streams/concat_streams.ts
+++ b/streams/concat_streams.ts
@@ -1,5 +1,4 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
-// This module is browser compatible.
 
 /**
  * A transform stream that concatenates multiple readable streams into a single
@@ -32,7 +31,6 @@ export class ConcatStreams<I = any>
   constructor() {
     super({
       async transform(stream, controller) {
-        // @ts-ignore ReadableStream<I> does have a '[Symbol.asyncIterator]()' method that returns an async iterator.
         for await (const value of stream) {
           controller.enqueue(value);
         }
--- a/streams/concat_streams.ts
+++ b/streams/concat_streams.ts
@@ -3,7 +3,7 @@
 
 /**
  * A transform stream that concatenates multiple readable streams into a single
- * stream.
+ * stream, in order.
  *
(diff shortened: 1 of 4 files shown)


BlackAsLight:
There are some aspects with this one and the one I created in the issue mentioned and that's that this one will push the entire contents of a stream into the queue continuously as it loops around the event loop, where the other one worked on a pull method, only giving the next chunk when it was wanted. This could cause problems if there is a significant delay between piping streams in and asking for values out.

A pulling method would look like this:
```ts
// Browser compatible
export class ConcatStreams<T> {
	#readable: ReadableStream<T>
	#writable: WritableStream<ReadableStream<T>>
	constructor() {
		const { readable, writable } = new TransformStream<ReadableStream<T>, ReadableStream<T>>()
		const gen = async function* () {
			for await (const stream of readable)
				for await (const chunk of stream)
					yield chunk
		}()
		this.#readable = new ReadableStream<T>({
			async pull(controller) {
				const { done, value } = await gen.next()
				if (done)
					controller.close()
				else
					controller.enqueue(value)
			}
		})
		this.#writable = writable
	}

	get readable(): ReadableStream<T> {
		return this.#readable
	}

	get writable(): WritableStream<ReadableStream<T>> {
		return this.#writable
	}
}
```
or this:
```ts
// Not browser compatible
export class ConcatStreams<T> {
	#readable: ReadableStream<T>
	#writable: WritableStream<ReadableStream<T>>
	constructor() {
		const { readable, writable } = new TransformStream<ReadableStream<T>, ReadableStream<T>>()
		this.#readable = ReadableStream.from(async function* () {
			for await (const stream of readable)
				for await (const chunk of stream)
					yield chunk
		}())
		this.#writable = writable
	}

	get readable(): ReadableStream<T> {
		return this.#readable
	}

	get writable(): WritableStream<ReadableStream<T>> {
		return this.#writable
	}
}
```
I couldn't see a way to use a `extend TransformStream` method with a pulling instead of pushing method.

iuioiua:
I realise that other contributors already came up with implementations. I'll close this PR to allow others to make this contribution. Either way, thanks for the input, @BlackAsLight.

PR #4724, opened 2024-05-14
docs(assert): document `assertStrictEquals()` equality comparison behavior

### LINKED ISSUE #4714: assertStrictEquals(-0, +0) should not throw

iuioiua:
Closes #4714

### DIFF
--- a/assert/assert_strict_equals.ts
+++ b/assert/assert_strict_equals.ts
@@ -8,6 +8,12 @@ import { red } from "@std/fmt/colors";
  * Make an assertion that `actual` and `expected` are strictly equal. If
  * not then throw.
  *
+ * This function copies the behavior of Node's
+ * {@linkcode https://nodejs.org/api/assert.html#assertstrictequalactual-expected-message | assert.strictStrictEqual},
+ * which uses
+ * {@linkcode https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/is | Object.is}
+ * for equality comparison.
+ *
  * @example
  * ```ts
  * import { assertStrictEquals } from "@std/assert/assert-strict-equals";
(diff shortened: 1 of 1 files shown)


assert/assert_strict_equals.ts
```diff
@@ -8,6 +8,12 @@ import { red } from "@std/fmt/colors";
  * Make an assertion that `actual` and `expected` are strictly equal. If
  * not then throw.
  *
+ * This function copies the behavior of Node's
+ * {@linkcode https://nodejs.org/api/assert.html#assertstrictequalactual-expected-message | assert.strictStrictEqual},
```
kt3k, line 12: We also consider the compatibility to jest, vitest, etc. I don't think we need to mention this information here

assert/assert_strict_equals.ts
```diff
@@ -8,6 +8,12 @@ import { red } from "@std/fmt/colors";
  * Make an assertion that `actual` and `expected` are strictly equal. If
```
kt3k, line 8: > `Make an assertion that actual and expected are strictly equal`

I think this part needs to be updated as it's not technically 'strict equality'

kt3k (approved):
LGTM 👍

Issue #4608, opened 2024-04-19
suggestion: add `cache` package


lucsoft:
**Is your feature request related to a problem? Please describe.**

Often in code there is a need to run something like lazy, like connecting to a database. Seen also in Kotlin

**Describe the solution you'd like**

i would like to have a function that gets executed once and returns the results always.

```ts
const con = await lazy(async () => db.connect()) // async lazy
const firstHitDate = lazy(() => new Date()) // sync lazy
```

**Describe alternatives you've considered**

Doing it manually and more verbose.

**Example Implementation**
```ts
export const lazy = <T>(callback: () => T) => {
    let result: T = undefined!;
    return () => result = (result || callback());
};
```

lucsoft:
Thanks for your Comment @lionel-rowe! The scope of this feature is just to allow lazy initialization / lazy code execution. See the Kotlin std where the inspo is coming from.

Your large comment kinda shows how complex real memo is. Which maybe not beneficial when the proposal advances or as a first step going into the Memo topic for the Deno Std. The Lazy method is simple enough to not have any large drawbacks, as it's quite straight forward. if cache empty, run the function, cache the result, return the cache.

lionel-rowe:
> Your large comment kinda shows how complex real memo is. Which maybe not beneficial when the proposal advances

Proposals at stage 1 typically take several years to get to stage 4 and achieve widespread implementation, and there's no guarantee they'll get there at all — quite often the committee fails to reach consensus on some sticking point and they get delayed indefinitely or withdrawn. `std` routinely covers ground that overlaps with APIs advancing through the TC39 proposal process, with the relevant `std` APIs being deprecated and removed if and when those proposals achieve implementation. So I think overlap is a feature, not a bug.

The reason I mention memoization here is that I have a couple of use cases for it currently, and only the other day I was considering suggesting it for inclusion in `std`. Given that `lazy` is a special case of memoization, it'd make sense to roll them into one if feasible to do so (otherwise the rollout would go something like "release `lazy` → release `memoize` → alias `lazy` to `memoize` → deprecate the `lazy` alias → remove the `lazy` alias", which isn't a great story for API stability).

[...]

PR #4725, opened 2024-05-14
feat(cache): add memoize and LruCache

lionel-rowe:
Closes https://github.com/denoland/deno_std/issues/4608

Some questions:

* Which methods should count as accessing/using an entry stored in an `LruCache`? Currently, all three of `get`, `set`, and `has` are tracked. In other words, when deleting the "least-recently-used", it means deleting the entry that was least-recently touched by any of `get`, `set`, or `has`.
* Is "on-the-fly" usage even worth supporting? It adds a small amount of complexity to the implementation and allows `memoize(fn)(arg)` to be memoized, with the following gotcha:
```js
// ✅ reference to function
const fn1 = memoize(fn)
const result1 = fn1(42)

// ✅ function literal
const fn2 = memoize((n: number) => n + 1)
const result2 = fn1(42)

// ✅ on-the-fly with reference to function
const result3 = memoize(fn)(42)

// ⚠️ on-the-fly with function literal - fails to memoize
const result4 = memoize((n: number) => n + 1)(42)
```

### DIFF
--- a/cache/deno.json
+++ b/cache/deno.json
@@ -1,0 +1,9 @@
+{
+  "name": "@std/cache",
+  "version": "0.224.1",
+  "exports": {
+    ".": "./mod.ts",
+    "./lru-cache": "./lru_cache.ts",
+    "./memoize": "./memoize.ts"
+  }
+}
--- a/cache/lru_cache.ts
+++ b/cache/lru_cache.ts
@@ -1,0 +1,56 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { MemoizationCache } from "./memoize.ts";
+
+/**
+ * [Least-recently-used](
+ * 	https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU
+ * ) cache.
+ *
+ * Automatically removes entries above the max size based on when they were
+ * last accessed with `get`, `set`, or `has`.
+ */
+export class LruCache<K, V> extends Map<K, V>
+  implements MemoizationCache<K, V> {
+  constructor(public maxSize: number) {
+    super();
+  }
+
+  #setMostRecentlyUsed(key: K, value: V): void {
+    // delete then re-add to ensure most recently accessed elements are last
+    super.delete(key);
+    super.set(key, value);
+  }
+
+  #pruneToMaxSize(): void {
+    if (this.size > this.maxSize) {
+      this.delete(this.keys().next().value);
+    }
+  }
+
+  override has(key: K): boolean {
+    const exists = super.has(key);
+
+    if (exists) {
+      this.#setMostRecentlyUsed(key, super.get(key)!);
+    }
+
+    return exists;
+  }
+
+  override get(key: K): V | undefined {
+    if (super.has(key)) {
+      const value = super.get(key)!;
+      this.#setMostRecentlyUsed(key, value);
+      return value;
+    }
+
+    return undefined;
+  }
+
+  override set(key: K, value: V): this {
+    this.#setMostRecentlyUsed(key, value);
+    this.#pruneToMaxSize();
+
+    return this;
+  }
+}
--- a/cache/memoize.ts
+++ b/cache/memoize.ts
@@ -1,0 +1,203 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import { LruCache } from "./lru_cache.ts";
+
+export type MemoizationCache<K, V> = {
+  has: (key: K) => boolean;
+  get: (key: K) => V | undefined;
+  set: (key: K, val: V) => unknown;
+  delete: (key: K) => unknown;
+};
+
+export type MemoizationOptions<Fn extends (...args: never[]) => unknown> = {
+  /**
+   * Provide a custom cache for getting previous results. By default, a new
+   * `Map` object is instantiated upon memoization and used as a cache.
+   */
+  cache?: MemoizationCache<unknown, ReturnType<Fn>>;
+  /**
+   * Function to get a unique cache key from the function's arguments. By
+   * default, a composite key is created from all the arguments plus the `this`
+   * value, using reference equality to check for equivalence.
+   *
+   * @example
+   * ```ts
+   * import { memoize } from "@std/cache";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const fn = memoize(({ value }: { cacheKey: number; value: number }) => {
+   *   return value;
+   * }, { getKey: ({ cacheKey }) => cacheKey });
+   *
+   * assertEquals(fn({ cacheKey: 1, value: 2 }), 2);
+   * assertEquals(fn({ cacheKey: 1, value: 99 }), 2);
+   * assertEquals(fn({ cacheKey: 2, value: 99 }), 99);
+   * ```
+   */
+  getKey?: (this: ThisParameterType<Fn>, ...args: Parameters<Fn>) => unknown;
+  /**
+   * Only use args as cache keys up to the `length` property of the function.
+   * Useful for passing unary functions as array callbacks, but should be
+   * avoided for functions with variable argument length (`...rest` or default
+   * params)
+   *
+   * @default false
+   */
+  truncateArgs?: boolean;
+  /**
+   * By default, promises are automatically removed from the cache upon
+   * rejection. If `cacheRejectedPromises` is set to `true`, promises will be
+   * retained in the cache even if rejected.
+   *
+   * @default false
+   */
+  cacheRejectedPromises?: boolean;
+};
+
+// used for memoizing the `memoize` function itself to enable on-the-fly usage
+const _cache: LruCache<unknown, ReturnType<typeof memoize>> = new LruCache(100);
+const _getKey: ReturnType<typeof _serializeArgList> = _serializeArgList(_cache);
+
+/**
+ * Cache the results of a function based on its arguments.
+ *
+ * @param fn - The function to memoize
+ * @param options - Options for memoization
+ *
+ * @example
+ * ```ts
+ * import { memoize } from "@std/cache";
+ * import { assertEquals } from "@std/assert";
+ *
+ * // fibonacci function, which is very slow for n > ~30 if not memoized
+ * const fib = memoize((n: bigint): bigint => {
+ *   return n <= 2n ? 1n : fib(n - 1n) + fib(n - 2n);
+ * });
+ *
+ * assertEquals(fib(100n), 354224848179261915075n);
+ * ```
+ */
+const memoize_: typeof memoize = memoize(memoize, {
+  cache: _cache,
+  getKey(fn, options) {
+    const optionVals = Object.entries(options ?? {})
+      .sort(([a], [b]) => a > b ? 1 : a < b ? -1 : 0)
+      .map(([, v]) => v);
+    return _getKey(fn, ...optionVals);
+  },
+});
+
+export { memoize_ as memoize };
+
+function memoize<Fn extends (...args: never[]) => unknown>(
+  fn: Fn,
+  options?: NoInfer<Partial<MemoizationOptions<Fn>>>,
+): Fn & {
+  cache: MemoizationCache<unknown, ReturnType<Fn>>;
+  getKey: (this: ThisParameterType<Fn>, ...args: Parameters<Fn>) => unknown;
+} {
+  const cache = options?.cache ?? new Map();
+  const getKey = options?.getKey ?? _serializeArgList(cache);
+  const truncateArgs = options?.truncateArgs ?? false;
+  const cacheRejectedPromises = options?.cacheRejectedPromises ?? false;
+
+  const memoized = function (
+    this: ThisParameterType<Fn>,
+    ...args: Parameters<Fn>
+  ): ReturnType<Fn> {
+    if (truncateArgs) args = args.slice(0, fn.length) as Parameters<Fn>;
+
+    const key = getKey.apply(this, args);
+
+    if (cache.has(key)) {
+      return cache.get(key)!;
+    }
+
+    let val = fn.apply(this, args) as ReturnType<Fn>;
+
+    if (val instanceof Promise && !cacheRejectedPromises) {
+      val = val.catch((reason) => {
+        cache.delete(key);
+        throw reason;
+      }) as typeof val;
+    }
+
+    cache.set(key, val);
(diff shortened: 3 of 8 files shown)


cache/memoize.ts
```diff
@@ -0,0 +1,203 @@
...
+   * By default, promises are automatically removed from the cache upon
+   * rejection. If `cacheRejectedPromises` is set to `true`, promises will be
+   * retained in the cache even if rejected.
+   *
+   * @default false
+   */
+  cacheRejectedPromises?: boolean;
+};
+
+// used for memoizing the `memoize` function itself to enable on-the-fly usage
```
kt3k, line 56: What is the `on-the-fly` usage?

memoize function being memoized feels surprising to me.

cache/memoize.ts
```diff
@@ -0,0 +1,203 @@
...
+      if (typeof arg === "undefined") return "undefined";
+      if (typeof arg === "bigint") return `${arg}n`;
+
+      if (
+        arg === null ||
+        ["string", "boolean", "number", "record", "tuple"].includes(typeof arg)
+      ) {
+        // null, string, boolean, number, or one of the upcoming
+        // [record/tuple](https://github.com/tc39/proposal-record-tuple)
+        // ECMAScript types
```
kt3k, line 162: I think it's too early to include "record" and "tuple" here.

cache/memoize.ts
```diff
@@ -0,0 +1,203 @@
...
+  const weakKeyToKeySegmentCache = new WeakMap<WeakKey, string>();
+  const weakKeySegmentToKeyCache = new Map<string, string[]>();
+  let i = 0;
+
+  const registry = new FinalizationRegistry<string>((keySegment) => {
+    for (const key of weakKeySegmentToKeyCache.get(keySegment) ?? []) {
+      cache.delete(key);
+    }
+    weakKeySegmentToKeyCache.delete(keySegment);
+  });
```
kt3k, line 148: Seems like a nice way to keep the cache minimal 👍

cache/memoize.ts
```diff
@@ -0,0 +1,203 @@
...
+   */
+  cacheRejectedPromises?: boolean;
+};
+
+// used for memoizing the `memoize` function itself to enable on-the-fly usage
+const _cache: LruCache<unknown, ReturnType<typeof memoize>> = new LruCache(100);
+const _getKey: ReturnType<typeof _serializeArgList> = _serializeArgList(_cache);
+
+/**
+ * Cache the results of a function based on its arguments.
```
kt3k, line 61: The keying implementation looks great to me, but can you document what types of arguments work with this? Does this work with any types of arguments? Is there any limitation with argument types?

lionel-rowe: Works with any type of arguments using value equality for primitives and reference equality otherwise, similar to `useMemo` in React. I'll add some more explanation in the comment.

cache/memoize.ts
lionel-rowe: An alternative that would avoid explicitly handling record/tuple yet would remain forward-compatible for them would be to negate the condition. Something like:

```diff
- if (
-    arg === null ||
-    ["string", "boolean", "number", "record", "tuple"].includes(typeof arg)
- ) 
+ if (
+    typeof arg !== 'function' && (arg === null || typeof arg !== 'object')
+ )
```

Intention is less clear that way though.

cache/memoize.ts
kt3k: Ah, thanks (I didn't read the PR description 😓 

I personally don't expect `memoize(fn)(42)` works in a memoized way. Do you know any prior example which supports that pattern, or is this just invented?

lionel-rowe: Not aware of a prior example, it just seemed like a nifty feature to have. But I think on balance the small bump in complexity coupled with the gotcha are reason enough to remove it.

cache/memoize.ts
kt3k: Is that safe to do? I slightly prefer only including `null`, `string`, `boolean`, and `number`, which feels easier to read.

One reason why I'd like to drop record and tuple is that I heard the rumor that proposal is not going forward anymore

lionel-rowe: > Is that safe to do? I slightly prefer only including null, string, boolean, and number, which feels easier to read.

Makes sense, I'll change to that along with a comment about forward-compatibility.

> I heard the rumor that proposal is not going forward

Sad if true 😔

cache/deno.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@std/cache",
+  "version": "0.224.1",
```
kt3k, line 3: I think this should be 0.1.0.
```suggestion
  "version": "0.1.0",
```

cache/memoize.ts
```diff
@@ -0,0 +1,197 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+export type MemoizationCache<K, V> = {
+  has: (key: K) => boolean;
+  get: (key: K) => V | undefined;
+  set: (key: K, val: V) => unknown;
+  delete: (key: K) => unknown;
```
kt3k, line 7: How about aligning these types with `Map<K, V>`?
```suggestion
  set: (key: K, val: V) => this;
  delete: (key: K) => unknown;
```

kt3k: Or maybe how about using Map compatible type for `cache` option, like `Map<unknown, ReturnType<Fn>>`? Then the users can get cache size like `memoized.cache.size`

cache/memoize.ts
```diff
@@ -0,0 +1,197 @@
...
+};
+
+export type MemoizationOptions<Fn extends (...args: never[]) => unknown> = {
+  /**
+   * Provide a custom cache for getting previous results. By default, a new
+   * `Map` object is instantiated upon memoization and used as a cache.
```
kt3k, line 13: Let's mention that `LruCache` can be a useful option for this.

cache/memoize.ts
```diff
@@ -0,0 +1,197 @@
...
+ * import { assertEquals } from "@std/assert";
+ *
+ * // fibonacci function, which is very slow for n > ~30 if not memoized
+ * const fib = memoize((n: bigint): bigint => {
+ *   return n <= 2n ? 1n : fib(n - 1n) + fib(n - 2n);
+ * });
+ *
+ * assertEquals(fib(100n), 354224848179261915075n);
+ * ```
+ */
```
kt3k, line 73: Can you also add the example of introspecting the cache contents using `memoized.cache` and `memoized.getKey`?

cache/memoize.ts
lionel-rowe: > Or maybe how about using Map compatible type for `cache` option, like `Map<unknown, ReturnType<Fn>>`? Then the users can get cache size like `memoized.cache.size`

`Map<unknown, ReturnType<Fn>>` is probably too restrictive, as I can't think of any reason to require things like `forEach`, `clear`, etc. to be implemented given they're not called internally, nor for implementation details like `set` returning `this`. It'd also place an unnecessary burden on consumers using non-Map-inherited cache implementations to implement new methods if the spec/TS types change.

Good point about `size` though, I'll try adding another type param to `memoize` to allow for accessing `size` and other props/methods where Map-inherited caches are used.

kt3k: Fair enough. `MemoizationCache` type makes sense to me now

lionel-rowe: Types for `memoizedFn.cache` are now correct, but the internal type logic is kinda disgusting, and it's now compulsory to explicitly supply type params for `LruCache`, which is less-than-ideal:

```ts
const fn = (x: 1) => x;
// good (cache is Map<string, 1>)
const memoized1 = memoize(fn);

// good (cache is Map<string, 1>)
const memoized2 = memoize(fn, {});

// ok-ish but cache is now Map<any, any>
const memoized3 = memoize(fn, {
    cache: new Map(),
});

// TS error "Type 'LruCache<unknown, unknown>' is not assignable to type 'MemoizationCache<string, 1>'."
const memoized4 = memoize(fn, {
  cache: new LruCache(99),
});

// ok but more verbose
const memoized5 = memoize(fn, {
  cache: new LruCache<string, 1>(99),
});
```

Any ideas on how to fix this without breaking type inference elsewhere would be much appreciated. Fixing it is like whack-a-mole — everything I've tried seems to either result in `Key` type becoming `string` where it should be something else (i.e. `getKey`'s return type is something other than string, yet `Key` is inferred as `string`), types getting widened to `any`, `unknown`s refusing to narrow to something more specific, `LruCache<K, V>`s getting widened to `MemoizationCache<K, V>`, etc.

kt3k (approved):
LGTM

cache/lru_cache.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * [Least-recently-used](
+ * 	https://en.wikipedia.org/wiki/Cache_replacement_policies#LRU
+ * ) cache.
+ *
+ * Automatically removes entries above the max size based on when they were
+ * last accessed with `get`, `set`, or `has`.
+ */
+export class LruCache<K, V> extends Map<K, V>
```
iuioiua, line 12: Please add `@example`

cache/mod.ts
```diff
@@ -0,0 +1,3 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+export * from "./memoize.ts";
```
iuioiua, line 2: Please add module docs. Example: https://github.com/denoland/std/blob/714fccba66e3ac52cb8d16f36d77bb2dcb10d73d/uuid/mod.ts#L4-L27

cache/memoize.ts
```diff
@@ -0,0 +1,227 @@
...
+};
+
+export type MemoizationOptions<
+  Fn extends (...args: never[]) => unknown,
+  Key,
+  Cache extends MemoizationCache<Key, ReturnType<Fn>>,
+> = {
+  /**
+   * Provide a custom cache for getting previous results. By default, a new
+   * `Map` object is instantiated upon memoization and used as a cache, with no
```
iuioiua, line 20:
```suggestion
   * {@linkcode Map} object is instantiated upon memoization and used as a cache, with no
```

cache/lru_cache.ts
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  override get(key: K): V | undefined {
+    if (super.has(key)) {
+      const value = super.get(key)!;
+      this.#setMostRecentlyUsed(key, value);
+      return value;
+    }
+
+    return undefined;
```
iuioiua, line 47: Can you please add a test for this line?

cache/memoize.ts
```diff
@@ -0,0 +1,227 @@
...
+        return JSON.stringify(arg);
+      }
+
+      try {
+        assertWeakKey(arg);
+      } catch (e) {
+        if (typeof arg === "symbol") {
+          return `Symbol.for(${JSON.stringify(arg.description)})`;
+        }
+        throw e;
```
iuioiua, line 198: Non-blocking: Can you please add a test for this line?

iuioiua (changes requested):
Great work so far - thorough tests and well thought out. I have few requests for changes and have updated our tools and checks to include `@std/cache`. Please let us know if you need any help getting those done.

However, my main concern with the package in it's current state is that it does too much, especially for given this is the initial version. I think we should start simple and add features as we see demand. I hope the costs to the maintainer and user for not doing so are obvious. On that same vain, some thoughts:
1. `options.cache` property looks useful, but it might be better as a private property.
2. Is `options.getKey` really needed, except for some edge cases?
3. Is `options.truncateArgs` needed if it mainly only avoids a small/trivial number of keystrokes in array mapper callbacks (and similar)? Might a simple example suffice instead?
4. Might `options.cacheRejectedPromises` be fine to omit if we have a reasonable default? We can add in the future if there's demand.

lionel-rowe:
@iuioiua 

> 1. `options.cache` property looks useful, but it might be better as a private property.

Do you mean it can be supplied as an option but not publicly available as a property of the returned function? That would still allow manual manipulation if a reference was maintained, just not to manipulate it from a reference to the function alone. If so, that sounds reasonable.

```ts
const myCache = ...
const myFunc = memoize((arg: any) => { /* ... */ }, { cache: myCache })
export { myFunc }
// `myCache` is is no longer available via `myFunc.cache`
export { myCache }
// now it's available via explicit reference
```

> 2. Is `options.getKey` really needed, except for some edge cases?

I think `getKey` is essential, otherwise caching with non-primitive arguments is always by reference (the default).

```ts
const user: User = { id: 1 }
const sameUser: User = { id: 1 }
const getUserDataByReference = memoize((user: User) => { /* ... */ })

getUserDataByReference(user) // computed once
getUserDataByReference(user) // still computed once
getUserDataByReference(sameUser) // now computed twice

const getUserDataById = memoize((user: User) => { /* ... */ }, { getKey: ({ id }) => id)

getUserDataById(user) // computed once
getUserDataById(sameUser) // still computed once
```

> 3. Is `options.truncateArgs` needed if it mainly only avoids a small/trivial number of keystrokes in array mapper callbacks (and similar)? Might a simple example suffice instead?

Fair point, I'll remove that and add an example.

> 4. Might `options.cacheRejectedPromises` be fine to omit if we have a reasonable default? We can add in the future if there's demand.

Probably the most sensible default is the current one, i.e. to remove promises from the cache upon rejection (after all, if the function threw synchronously, that result obviously wouldn't be cached either). And it can be worked around easily enough by wrapping the result/rejection of the inner function in something akin to `Promise.allSettled`'s `PromiseSettledResult<T>`.

iuioiua:
I see. So based on your feedback, I suggest we do the following:
1. Keep the `cache` option, but remove it from the returned function.
2. Keep the `getKey` option.
3. Remove the `truncateArgs` option, keeping its default behavior, and instead add an example for said use case.
4. Remove the `cacheRejectedPromises`, keeping its default behavior, and instead add an example for said use case.

Does this all sound reasonable? Again, if we see demand for these functionalities in the future, we can add them in a non-breaking manner. WDYT?

cache/memoize.ts
lionel-rowe: Replacing the re-throw with `unreachable` as this line should never be hit.

iuioiua (changes requested):
I just realise I hadn't clicked "Submit review" from like a week ago 🤦🏾‍♂️

kt3k (approved):
LGTM!

PR #4731, opened 2024-05-15
docs(media-types): polish documentation

iuioiua:
Towards #4730

### DIFF
--- a/media_types/get_charset.ts
+++ b/media_types/get_charset.ts
@@ -9,14 +9,20 @@ import { db, type KeyOfDb } from "./_db.ts";
  * Given a media type or header value, identify the encoding charset. If the
  * charset cannot be determined, the function returns `undefined`.
  *
- * @example
+ * @param type The media type or header value to get the charset for.
+ *
+ * @returns The charset for the given media type or header value, or `undefined`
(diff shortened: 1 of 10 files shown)


media_types/get_charset.ts
```diff
@@ -9,14 +9,20 @@ import { db, type KeyOfDb } from "./_db.ts";
...
+ * @example Usage
  * ```ts
  * import { getCharset } from "@std/media-types/get-charset";
+ * import { assertEquals } from "@std/assert/assert-equals";
  *
- * getCharset("text/plain"); // "UTF-8"
- * getCharset("application/foo"); // undefined
- * getCharset("application/news-checkgroups"); // "US-ASCII"
- * getCharset("application/news-checkgroups; charset=UTF-8"); // "UTF-8"
+ * assertEquals(getCharset("text/plain"), "UTF-8");
```
iuioiua, line 22: Honestly, I wonder whether having assertions within examples is the best solution for what examples aim to achieve. The pro is that it guarantees that examples are correct. The cons are a slight reduction in readability and an extra dependency. I'm happy to hear other's thoughts.

kt3k (approved):
LGTM 👍

PR #4736, opened 2024-05-15
chore: update `title.yml`

kt3k:
- Changed snake_case names to kebab-case for scope names. [The version upgrade tool](https://github.com/denoland/bump-workspaces) assumes these match with JSR package name.
- Removed `_tools` scope as it's not jsr package. I think we should use non-scoped `chore:` for tool updates.
- Added `cache` to make #4725 CI green.

### DIFF
--- a/.github/workflows/title.yml
+++ b/.github/workflows/title.yml
@@ -33,7 +33,7 @@ jobs:
             ci
             chore
             revert
-          # This should be kept up-to-date with the packages in the Standard
+          # This should be kept up-to-date with the current packages list
           scopes: |
             archive
             assert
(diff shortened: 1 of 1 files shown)


.github/workflows/title.yml
```diff
@@ -34,24 +34,23 @@ jobs:
             chore
             revert
           # This should be kept up-to-date with the packages in the Standard
```
iuioiua, line 36:
```suggestion
          # This should be kept up-to-date with the current packages list
```

iuioiua (changes requested):

iuioiua (approved):
Nice!

PR #4749, opened 2024-05-16
refactor(uuid): factor out common symbols into `common` module

iuioiua:
This PR moves `NIL_UUID` into `uuid/constants.ts` and the remaining symbols implemented in `mod.ts` into `uuid/common.ts`.

Towards #4748


kt3k (approved):
It makes sense to export common utils from `@std/uuid/common`. LGTM

Issue #4659, opened 2024-04-29
`expect().toContain()` error message doesn't show expected string


marvinhagemeister:
**Describe the bug**

When an `expect().toContain()` assertion failed we get a very generic error message:

```txt
error: AssertionError: The value doesn't contain the expected item
      throw new AssertionError("The value doesn't contain the expected item");
```

This makes it difficult to know what exactly went wrong.

**Steps to Reproduce**

```ts
import { expect } from "jsr:@std/expect";

Deno.test("foo", () => {
  expect("foobar").toContain("baz");
});
```

**Expected behavior**

Print something like `Input string did not contain "baz"`. Not sure what to do for long strings though.

**Environment**

- OS: macOS 14.4.1
- deno version: 1.42.4
- std version: 0.224.0

JakeAve:
I've never used `toContainEqual` but it looks like we could do the same thing there. Should I include that in the PR?

marvinhagemeister:
Good idea, makes sense to add this to `toContainEqual` too

kt3k:
fixed in #4750

PR #4750, opened 2024-05-16
fix(expect): updated error message for toContain

JakeAve:
See issue: #4659 

Updated error message `toContain` to include the values from the failed test.

~~Should we also add this to `toContainEqual`?~~

For `toContainEqual` we need to come to an agreement for how to format that.

### DIFF
--- a/expect/_matchers.ts
+++ b/expect/_matchers.ts
@@ -417,13 +417,17 @@ export function toContainEqual(
   if (context.isNot) {
     if (doesContain) {
       throw new AssertionError(
-        `The value ${fmtValue} contains the expected item ${fmtExpected}`,
+        `The value contains the expected item.
+Value: ${fmtValue}
+Expected: ${fmtExpected}`,
       );
     }
   } else {
     if (!doesContain) {
       throw new AssertionError(
-        `The value ${fmtValue} doesn't contain the expected item ${fmtExpected}`,
+        `The value doesn't contain the expected item.
+Value: ${fmtValue}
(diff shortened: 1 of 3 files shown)


expect/_matchers.ts
```diff
@@ -375,11 +375,15 @@ export function toContain(
 
   if (context.isNot) {
     if (doesContain) {
-      throw new AssertionError("The value contains the expected item");
+      throw new AssertionError(
+        `The value "${context.value}" contains the expected item "${expected}"`,
```
kt3k, line 379: Can you use `format` function instead? `format` handles the case when the value has `"` in it.

expect/_to_contain_test.ts
```diff
@@ -15,9 +15,20 @@ Deno.test("expect().toContain()", () => {
...
   }, AssertionError);
-  assertThrows(() => {
-    expect("foobarbaz").toContain("qux");
-  }, AssertionError);
+  assertThrows(
+    () => {
+      expect("foobarbaz").toContain("qux");
+    },
+    AssertionError,
+    "foobarbaz",
```
kt3k, line 23: I think it's better to assert the entire error message.

kt3k:
Thanks for your contribution!

> Should we also add this to toContainEqual?

That's also welcome!

kt3k:
I updated the message format a little for `toContainEqual`.

kt3k (approved):
LGTM. Thanks!

Issue #3087, opened 2023-01-03
Add a transform stream version for base64 encoding/decoding


davbrito:
**Is your feature request related to a problem? Please describe.**

I'd like to encode to base64 a stream of binary data without having to load it all on memory.

**Describe the solution you'd like**

A streamed version of base64 encoder & decoder. Eg:

```ts
import { EncoderStream } from "https://deno.land/std@$STD_VERSION/encoding/base64.ts";

source.pipeThrough(new EncoderStream()).pipeTo(dest)
```

**Describe alternatives you've considered**

I've found this library that works with Node streams https://github.com/mazira/base64-stream
I haven't tried if it works on Deno yet, and I don't know if there is a way to convert a Node Transform stream into a web TransformStream.

iuioiua:
Are you able to elaborate on your use case?

PR #4756, opened 2024-05-16
feat(streams): add `Base64EncoderStream` and `Base64DecoderStream`

babiabeo:
Close: #3087


streams/base64_decoder_stream.ts
```diff
@@ -0,0 +1,53 @@
...
+ * ```
+ */
+export class Base64DecoderStream extends TransformStream<string, Uint8Array> {
+  /** The remain data from the previous chunk. */
+  #remain: string = "";
+
+  constructor() {
+    super({
+      transform: (chunk, controller) => {
+        chunk = this.#remain + chunk.replaceAll(/(\r\n|\r|\n)/g, "");
```
kt3k, line 30: I'm not sure why we need to replace line-breaks here

babiabeo: There will be some cases where the encoded data is split into lines, for example, when working with MIME

kt3k: hmm maybe it's useful in that case, but it could be unnecessary perf penalty for the case when there's no line-breaks in input.. I wonder if it's worth making this replacement optional.

BlackAsLight: What type of MIME type would insert line breaks in what is meant to be one continuous base64 string?

babiabeo: The MIME mentioned here is a standard format of email messages. According to [RFC 2045](https://datatracker.ietf.org/doc/html/rfc2045#section-6.8):
> The encoded output stream must be represented in lines of no more than 76 characters each. All line breaks or other characters not found in Table 1 must be ignored by decoding software.

BlackAsLight: That requirement of 76 characters per line isn't within the Base64 encoding spec https://datatracker.ietf.org/doc/html/rfc2045#section-6.7 so arguable such behaviour should be excluded from it and handled either before or after when necessary.

babiabeo:
As suggested by @BlackAsLight, I have removed line breaks handling from both encoder and decoder.

kt3k:
> As suggested by @BlackAsLight, I have removed line breaks handling from both encoder and decoder.

I think this also works. In this case, we can remove/insert line-breaks before/after the decoding/encoding.

kt3k:
I wonder if it makes more sense to put these APIs next to the `encodeBase64`/`decodeBase64` APIs in `@std/encoding/base64`. WDYT? cc @iuioiua @crowlKats

iuioiua:
> I wonder if it makes more sense to put these APIs next to the `encodeBase64`/`decodeBase64` APIs in `@std/encoding/base64`. WDYT? cc @iuioiua @crowlKats

Yep.

iuioiua:
Before merging this in, I'd like to know more about any justifications for encoding/decoding streams. Adding these APIs will likely set a precedent for other streaming modes like base 32, etc.

kt3k:
> I'd like to know more about any justifications for encoding/decoding streams.

It's used for encoding/decoding a huge data, the size of which exceeds the memory limit of your program.

kt3k:
> We recently agreed to not have encoding/decoding + streaming APIs in the Standard Library as we thought there aren't many use cases.

Can you link to the discussion?

iuioiua:
The agreement was made in a previous weekly `deno_std` meeting. I sent you the details.

To be clear, I'm happy with having `Base64EncoderStream` in the Standard Library; we should be aware that doing so will set a precedent and justify having the others in there, too. And at that point, maybe we should just include [`@doctor/encoding-stream`](https://jsr.io/@doctor/encoding-stream) in the Standard Library (with permission from and proper attribution for @BlackAsLight, of course).

BlackAsLight:
> And at that point, maybe we should just include [`@doctor/encoding-stream`](https://jsr.io/@doctor/encoding-stream) in the Standard Library (with permission from and proper attribution for @BlackAsLight, of course).

Sure

BlackAsLight:
> This PR only implements the base64 encoder and decoder based on [mazira/base64-stream](https://github.com/mazira/base64-stream) which may not be the same as [@doctor/encoding-stream/base64](https://jsr.io/@doctor/encoding-stream/doc/base64/~).

Your implementation is more or less the same as mine. I would characterise mine as more tidy though.

iuioiua:
Yes, let's close this PR and open a new one, adding `@doctor/encoding-stream` to `std/encoding` in new files. I.e., `encoding/base32_stream.ts`, etc.

PR #4758, opened 2024-05-16
refactor(assert,internal): rename `diffstr()` to `diffStr()`

iuioiua:
There's no reason for this function to not adhere to the function naming convention.

Prerequisite #4757

### DIFF
--- a/assert/assert_equals.ts
+++ b/assert/assert_equals.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 import { equal } from "./equal.ts";
-import { buildMessage, diff, diffstr, format } from "@std/internal";
+import { buildMessage, diff, diffStr, format } from "@std/internal";
 import { AssertionError } from "./assertion_error.ts";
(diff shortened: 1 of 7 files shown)


kt3k:
`@std/assert@0.225.2` will be broken by this change.

iuioiua:
How?

kt3k:
`@std/assert@0.225.2` depends on `@std/internal@^0.225.1`. The dependency automatically upgrades to the next version of `@std/internal`, but it still imports `diffstr` and that import doesn't work

iuioiua:
But won't both packages be upgraded at the same time?

kt3k:
The next version of `@std/assert` will work with the next version of `@std/internal`, but we can't modify `@std/assert@0.225.2` at this point, and it still depends on the future (patch-upgraded) version of `@std/internal`

kt3k:
I think the solution would be upgrading `@std/internal` version (probably major?) in this PR or another PR.

iuioiua:
Ok. I can upgrade it in a follow-up PR.

kt3k (approved):
LGTM

PR #4763, opened 2024-05-17
docs(collections): fix typos and improve example

kt3k:


collections/associate_with.ts
```diff
@@ -2,10 +2,10 @@
 // This module is browser compatible.
 
 /**
- * Associates each element of an array with a value returned by a selector
+ * Associates each string element of an array with a value returned by a selector
  * function.
  *
- * If any of two pairs would have the same value the latest on will be used
+ * If any of two pairs would have the same value the latest one will be used
```
iuioiua, line 8:
```suggestion
 * If any of two pairs would have the same value, the latest one will be used
```

iuioiua (changes requested):

iuioiua (approved):
Nice!

PR #4764, opened 2024-05-17
fix(msgpack): encode maps with null prototype

sanekb:


iuioiua (changes requested):
Please run `deno fmt`.

sanekb:
> Please run `deno fmt`.

thanks for remind

iuioiua (commented):
Can you provide some context behind why this change is needed? Is there a related issue?

kt3k:
This 'is-plain-object' check seems mostly aligned with the implementation of `npm:is-plain-obj@2.1.0` https://github.com/sindresorhus/is-plain-obj/blob/v2.1.0/index.js and looks reasonable to me.

sanekb:
> Is there a related issue?
it is not related to existed issue
1. but in my case, objects are validated using `jsr:@oxi/schema` and then piped to the `encode()`, which throws an error because the validated objects are recreated by calling `Object.create(null)`
2. logically, objects without a prototype are the most "plain", isn't it?

iuioiua:
> > Is there a related issue?
> 
> it is not related to existed issue
> 
> 1. but in my case, objects are validated using `jsr:@oxi/schema` and then piped to the `encode()`, which throws an error because the validated objects are recreated by calling `Object.create(null)`
> 2. logically, objects without a prototype are the most "plain", isn't it?

Fair justifications. For future reference, please be sure to include some context in the initial comment for PRs 🙂

PR #4767, opened 2024-05-17
perf(bytes): skip doing extra work in some scenarios

dsherret:
Not sure if this is worth it (or if I have introduced some bugs with this change), but I think we can avoid doing some work here.


iuioiua (approved):
Nice

PR #4771, opened 2024-05-20
docs: improve documentation on how to write documentation

iuioiua:
In preparation for the upcoming documentation hackathon.

PR #4772, opened 2024-05-20
docs(collections,internal,media-types): use `@typeParam` tag instead of `@template`

iuioiua:
To align with https://deno.com/blog/document-javascript-package, and therefore, #4771. Plus, the purpose of `@typeParam` is clearer than `@template`.


kt3k (approved):
`@typeParam` sounds more straightforward to me. LGTM

PR #4744, opened 2024-05-15
chore(media-types): `media-types@1.0.0-rc.1`

iuioiua:
Package documentation: https://jsr.io/@std/media-types/doc

Towards #4730

Pre-requisiste #4730

PR #4776, opened 2024-05-20
fix(media-types): do not expose vendored json file as public API

kt3k:
Currently [vendor/mime-db.v1.52.0](https://jsr.io/@std/media-types/doc/vendor/mime-db.v1.52.0/~/) is exposed as a public API, but I believe this is an error.

related https://github.com/denoland/deno_std/pull/4744/ https://github.com/denoland/deno_std/issues/4730


iuioiua (approved):
Oh, good catch!

PR #4777, opened 2024-05-20
docs(csv,encoding,http,json,media-types,uuid): use `rfc-editor.org` for RFC links

kt3k:
`www.rfc-editor.org` seems the most official place for hosting RFCs. I think we should always use `www.rfc-editor.org` urls for the references of RFCs

ref:
- https://en.wikipedia.org/wiki/Request_for_Comments#Obtaining_RFCs
- github discussion discussing various RFC urls https://github.com/ietf-tools/datatracker/discussions/7259


iuioiua (approved):
I've added more cases. I suggest checking the codebase once more for the remaining old URLs. LGTM and good idea.

PR #4779, opened 2024-05-20
test(internal): improve test coverage

mbhrznr:
pushes code coverage of the `internal` sub-module closer to 100%.

the only file missing is `internal/styles.ts`, which is vendored from `fmt/colors.ts`.
added a test case for `noColor`, yet that doesn't seem to be sufficient to appease codecov.


internal/diff.ts
```diff
@@ -147,22 +152,17 @@ function createFp(
  */
 export function diff<T>(A: T[], B: T[]): DiffResult<T>[] {
   const prefixCommon = createCommon(A, B);
-  const suffixCommon = createCommon(
```
mbhrznr, line 150: apparently this became unreachable with the most recent refactoring in https://github.com/denoland/deno_std/pull/4704 due to the lack of `reverse` logic.

internal/diff.ts
```diff
@@ -187,7 +187,6 @@ export function diff<T>(A: T[], B: T[]): DiffResult<T>[] {
   ): FarthestPoint {
     const M = A.length;
     const N = B.length;
-    if (k < -N || M < k) return { y: -1, id: -1 };
```
mbhrznr, line 190: yet another unreachable piece of code, which got elimiated.

internal/diff_str.ts
```diff
@@ -38,22 +36,9 @@ const EXT_LATIN_CHARS =
  */
 function tokenize(string: string, wordDiff = false): string[] {
   if (wordDiff) {
-    const tokens = string.split(WHITESPACE_SYMBOLS).filter((token) => token);
```
mbhrznr, line 41: the check for `tokens[i + 1]` would've always returned `true`, as we're filtering for truthy values.

mbhrznr (commented):
@iuioiua i would be really keen to get your feedback on this one.

internal/build_message.ts
```diff
@@ -112,3 +111,9 @@ export function buildMessage(
   messages.push(...(stringDiff ? [diffMessages.join("")] : diffMessages), "");
   return messages;
 }
+
+/** Used internally for testing. */
+export const _internals = {
```
iuioiua, line 116: Let's export these functions instead of putting them in these `_internals` objects. It'll be much cleaner. Ditto for the other files in this package. This is fine to do here as this is purely an internal package. You'll have to add `@example` tags to the newly exported symbols to make the doc checker (`deno task lint:docs`) too.

iuioiua (changes requested):
Excellent work so far. This is great.

internal/build_message.ts
mbhrznr: gotcha!
for some reason i was unsure if we would ever want to export the other functions in this package.
implemented as suggested.
love the way how the doc checker works bth!

iuioiua: Are you able to fix the functions in this file up too?

internal/diff.ts
```diff
@@ -37,13 +51,62 @@ function createCommon<T>(A: T[], B: T[]): T[] {
...
 
-function backTrace<T>(
+/**
+ * Creates an array of backtraced differences.
+ *
+ * @typeParam T The type of elements in the arrays.
+ *
+ * @param A The first array.
+ * @param B The second array.
+ * @param current The current {@linkcode FarthestPoint}y.
```
iuioiua, line 91:
```suggestion
 * @param current The current {@linkcode FarthestPoint}.
```

iuioiua (changes requested):
Looking better.

internal/build_message.ts
mbhrznr: oh... i'm so sorry.
cleanup for `build_message` is done as well.

iuioiua (approved):
Thank you very much for this. We should discuss and explore options for hitting that colors code.

PR #4782, opened 2024-05-21
chore: improve doc checker

iuioiua:
This PR:
1. Swaps the command order in the `lint:docs` task so `document.jsDoc` is known to be defined in the doc checker.
2. Simplifies some of the functions in the doc checker by just passing in `document`.
3. Adds the ability for the doc checker to check if the JSDocs for exported variables contain at least one example.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -19,7 +19,7 @@
     "lint:circular": "deno run --allow-env --allow-read --allow-net=deno.land,jsr.io ./_tools/check_circular_package_dependencies.ts",
     "lint:mod-exports": "deno run --allow-env --allow-read ./_tools/check_mod_exports.ts",
     "lint:tools-types": "deno check _tools/*.ts",
-    "lint:docs": "deno run -A _tools/check_docs.ts && deno doc --lint collections/mod.ts bytes/mod.ts datetime/mod.ts internal/mod.ts media_types/mod.ts url/mod.ts",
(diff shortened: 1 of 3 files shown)


deno.json
```diff
@@ -19,7 +19,7 @@
     "lint:circular": "deno run --allow-env --allow-read --allow-net=deno.land,jsr.io ./_tools/check_circular_package_dependencies.ts",
     "lint:mod-exports": "deno run --allow-env --allow-read ./_tools/check_mod_exports.ts",
     "lint:tools-types": "deno check _tools/*.ts",
-    "lint:docs": "deno run -A _tools/check_docs.ts && deno doc --lint collections/mod.ts bytes/mod.ts datetime/mod.ts internal/mod.ts media_types/mod.ts url/mod.ts",
+    "lint:docs": "deno doc --lint collections/mod.ts bytes/mod.ts datetime/mod.ts internal/mod.ts media_types/mod.ts url/mod.ts && deno run -A _tools/check_docs.ts",
```
iuioiua, line 22: By swapping the order of commands, we know that `document.jsDoc` is defined.

media_types/extensions_by_type.ts
```diff
@@ -4,8 +4,6 @@
 import { parseMediaType } from "./parse_media_type.ts";
 import { extensions } from "./_db.ts";
 
-export { extensions };
```
iuioiua, line 7: Prerequisite for this PR: #4783

iuioiua:
On second thought, do we want to check that even variables have examples?

kt3k:
> On second thought, do we want to check that even variables have examples?

hmm, I don't think every variable needs examples..

kt3k (approved):
LGTM

PR #4783, opened 2024-05-21
fix(media-types): remove `extensions` export

iuioiua:
Like #4776, it looks like this variable is exported accidently.


kt3k:
This doesn't seem accidentally exported to me. It's documented and exported from the beginning of the package https://github.com/denoland/deno_std/pull/2286/files#diff-92e0917a6de42a1319f4f29e026ba39b73bf1dc70dececdcbced2d732b018d99R30-R31

PR #4785, opened 2024-05-21
docs(cli): improve spinner.message document

andrewthauer:
This makes the `Spinner` in `@std/cli` expose a `message` property to be consistent with the existing `color` property. It also adds doc comments and an examples for both `message` and `color `

_NOTE: I noticed another issue while making this change. The `Color` type also includes `Ansi` which is defined as a string. However, the actual code in the set `color` property will never let you set this to anything other then a valid `COLORS` value._

### DIFF
--- a/cli/spinner.ts
+++ b/cli/spinner.ts
@@ -98,7 +98,7 @@ export class Spinner {
    * }
    *
    * spinner.stop();
-   * spinner.message = "Done!";
+   * console.log("Done!");
    * ```
    */
   message: string = "";
--- a/cli/spinner.ts
+++ b/cli/spinner.ts
@@ -74,12 +74,34 @@ export interface SpinnerOptions {
  */
 export class Spinner {
   #spinner: string[];
-  /** The message to display next to the spinner. */
-  message: string;
   #interval: number;
   #color?: Color;
   #intervalId: number | undefined;
   #active = false;
+
+  /**
+   * The message to display next to the spinner.
+   * This can be changed while the spinner is active.
+   *
+   * @example
+   * ```ts
+   * import { Spinner } from "@std/cli/spinner";
+   *
+   * const spinner = new Spinner({ message: "Working..." });
+   * spinner.start();
+   *
+   * for (let step = 0; step < 5; step++) {
+   *   // do some work
+   *   await new Promise((resolve) => setTimeout(resolve, 1000));
+   *
+   *   spinner.message = `Finished Step #${step}`;
+   * }
+   *
+   * spinner.stop();
+   * spinner.message = "Done!";
+   * ```
+   */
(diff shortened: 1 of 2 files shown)


kt3k:
Changing message seems already supported? See https://github.com/denoland/deno_std/pull/4079

andrewthauer:
> Changing message seems already supported? See #4079

That's correct. The feature is already available. Could be a me thing, but I did not realize you could change `message` initially which led me to look at the code and realized it was a public field and not a property like `color` is. The [documentation in jsr](https://jsr.io/@std/cli/doc/spinner/~/Spinner) isn't as clearly noticeable atm (separate issue). So I figured it would make more sense to use a formal property here so it's both consistent with `color` and has some documentation. I've also added docs for `color` too.

kt3k:
This sounds rather doc issue, or UI issue of JSR. `message` property doc appears in JSR https://jsr.io/@std/cli/doc/spinner/~/Spinner#property_message . So changing message to getter/setter seems like increasing unnecessary boilerplate.

How about updating the API doc of `message` property (Probably adding the `@example` of the current `set message` doc)?

andrewthauer:
> This sounds rather doc issue, or UI issue of JSR. `message` property doc appears in JSR https://jsr.io/@std/cli/doc/spinner/~/Spinner#property_message . So changing message to getter/setter seems like increasing unnecessary boilerplate.
>
> How about updating the API doc of `message` property (Probably adding the `@example` of the current `set message` doc)?

To me this is more about consistency and readability of the class. The act of creating a property screams this is meant to be public. Where, a field could just be something you accidentally forgot to make private explicitly. Also, fields tend to be located at the top which then makes interlacing larger comments between fields a bit unwieldy imo.
 
Another thing that is potentially an issue is usage of `constructor(public someField: string)`. It doesn't appear `someField` is easily document-able as far as I can tell. This might be something for consideration.

Anyways, I looked around and see there is a fair bit or precedence for making fields public already. I'm happy to comment the field versus making is a property. If any of my notes resonate let me know and I'm happy to adjust.

cli/spinner.ts
```diff
@@ -74,13 +74,35 @@ export interface SpinnerOptions {
...
+   *
+   * for (let step = 0; step < 5; step++) {
+   *   // do some work
+   *   await new Promise((resolve) => setTimeout(resolve, 1000));
+   *
+   *   spinner.message = `Finished Step #${step}`;
+   * }
+   *
+   * spinner.stop();
+   * spinner.message = "Done!";
```
andrewthauer, line 101: Just realized this should be `console.log` since setting the message after `stop` won't do anything. I'll fixl

kt3k (approved):
LGTM again. Thanks for your contribution!

PR #4790, opened 2024-05-21
docs(uuid): update module docs

marvinhagemeister:
- Some wording changes
- Show an example of how to use this module on the main page on jsr.io


uuid/mod.ts
```diff
@@ -2,17 +2,26 @@
...
- * Based on {@linkcode https://www.npmjs.com/package/uuid | npm:uuid}, which is
- * based on {@link https://www.rfc-editor.org/rfc/rfc4122.html | RFC 4122}.
- *
- * Support for RFC4122 version 1, 3, 4, and 5 UUIDs
+ * Based on {@linkcode https://www.npmjs.com/package/uuid | npm:uuid}.
  *
  * This module is browser compatible.
  *
+ * @example
+ * ```js
```
iuioiua, line 16: Nit: for consistency with the rest of the codebase.
```suggestion
 * ```ts
```

uuid/mod.ts
```diff
@@ -2,17 +2,26 @@
...
  *
- * Based on {@linkcode https://www.npmjs.com/package/uuid | npm:uuid}, which is
- * based on {@link https://www.rfc-editor.org/rfc/rfc4122.html | RFC 4122}.
- *
- * Support for RFC4122 version 1, 3, 4, and 5 UUIDs
+ * Based on {@linkcode https://www.npmjs.com/package/uuid | npm:uuid}.
  *
  * This module is browser compatible.
  *
+ * @example
```
iuioiua, line 15: Nit: So the code snippet title isn't "Example 1" on JSR.
```suggestion
 * @example Usage
```

iuioiua (changes requested):
PTAL at https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md#documentation

It contains steps such as adding `uuid/mod.ts` to the `lint:docs` task.

crowlKats:
Some of the functions are missing @param and @return doc tags

_tools/check_docs.ts
```diff
@@ -47,6 +47,12 @@ const ENTRY_POINTS = [
   "../streams/mod.ts",
   "../text/mod.ts",
   "../ulid/mod.ts",
+  "../uuid/common.ts",
+  "../uuid/constants.ts",
+  "../uuid/v1.ts",
+  "../uuid/v3.ts",
+  "../uuid/v4.ts",
+  "../uuid/v5.ts",
```
iuioiua, line 55: Files are done individually because `deno doc --lint uuid/mod.ts` doesn't work.

uuid/mod.ts
```diff
@@ -2,23 +2,87 @@
...
+ * ```ts
+ * import { v5, NAMESPACE_DNS } from "@std/uuid";
+ * import { assert } from "@std/assert/assert";
+ *
+ * const data = new TextEncoder().encode("deno.land");
+ * const uuid = await v5.generate(NAMESPACE_DNS, data);
+ * assert(v5.validate(uuid));
+ * ```
+ */
 export * as v5 from "./v5.ts";
```
iuioiua, line 88: These changes actually don't register as docs for `v1`, etc., and I'm not sure how to make it do so. This seems like a bug. Any ideas @crowlKats or @dsherret?

kt3k: How about the below workaround?

```ts
import * as v1_ from "./v1.ts";
/**
 * ...Docs...
 */
export const v1 = v1_;
```

kt3k: Looks like the below is even better?:

```ts
import { generate as generateV1, validate as validateV1 } from "./v1.ts";
/**
 * ...docs
 */
export const v1 = {
  generate: generateV1,
  validate: validateV1,
};
```

kt3k (approved):
LGTM

iuioiua (approved):

uuid/mod.ts
iuioiua: Awesome. Thanks!

PR #4794, opened 2024-05-21
BREAKING(data-structures): hide private internals

lucacasonato:
This commit makes the internals of
`BinarySearchTree` properly private, so that
`BinarySearchNode` is not leaked in public types anymore.


kt3k (approved):
LGTM. Thanks!

PR #4798, opened 2024-05-21
docs(json): lint `@std/json` docs

littledivy:


iuioiua:
FYI, we just updated the doc checker to check classes. Hence, the lint failure(s).

json/concatenated_json_parse_stream.ts
```diff
@@ -12,32 +12,90 @@ const primitives = new Map(
...
+ *   { foo: "bar" },
+ *   { baz: 100 },
+ * ]);
  * ```
  */
 export class ConcatenatedJsonParseStream
   implements TransformStream<string, JsonValue> {
-  /** A writable stream of byte data. */
+  // TODO(iuioiua): Investigate why this class is implemented differently to the other JSON streams.
+  /**
```
iuioiua, line 38: I went with this for now. I didn't bother with a proper example because this, as a public property, seems like a mistake/fault, and may get removed in the future.

iuioiua:
PTAL @kt3k

json/concatenated_json_parse_stream.ts
```diff
@@ -12,32 +12,90 @@ const primitives = new Map(
...
+ * assertEquals(await Array.fromAsync(stream), [
+ *   { foo: "bar" },
+ *   { baz: 100 },
+ * ]);
  * ```
  */
 export class ConcatenatedJsonParseStream
   implements TransformStream<string, JsonValue> {
-  /** A writable stream of byte data. */
+  // TODO(iuioiua): Investigate why this class is implemented differently to the other JSON streams.
```
kt3k, line 37: `JSONParseStream` supposes every chunk is valid JSON string which represents a single JSON object. `ConcatenatedJsonParseStream` handles stream of strings where each chunk can be incomplete fragments of JSON strings.

This difference exists because in NDJSON or JSONLines, each line contains single JSON object. So we can use `TextLineStream` to split them and pass the result to `JSONParseStream`. On the other hand, in Concatenated JSON, there's no explicit delimiter of JSONs. So the class needs to maintain the internal parsing state of JSONs

json/json_parse_stream.ts
```diff
@@ -17,48 +17,44 @@ function isBrankString(str: string) {
...
- * import { TextLineStream } from "@std/streams/text-line-stream";
- * import { JsonParseStream } from "@std/json/json-parse-stream";
- *
- * const url = "@std/json/testdata/test.jsonl";
- * const { body } = await fetch(url);
- *
- * const readable = body!
- *   .pipeThrough(new TextDecoderStream())  // convert Uint8Array to string
- *   .pipeThrough(new TextLineStream()) // transform into a stream where each chunk is divided by a newline
- *   .pipeThrough(new JsonParseStream()); // parse each chunk as JSON
```
kt3k, line 32: Can we keep one of this type of example with `no-eval`? This seems typical real world usage of this class to me. It helps the users to know how to read JSON Lines response from servers

kt3k (approved):
LGTM

json/concatenated_json_parse_stream.ts
iuioiua: I just meant how this implements `TransformStream` rather than extending it.

PR #4800, opened 2024-05-21
ci: handle classes in check_docs.ts

lucacasonato:


_tools/check_docs.ts
```diff
@@ -4,14 +4,21 @@
  * This script checks that all public symbols documentation aligns with the
  * {@link ./CONTRIBUTING.md#documentation | documentation guidelines}.
  *
- * TODO(iuioiua): Add support for classes and methods.
+ * TODO(lucacasonato): Add support for variables, interfaces, namespaces, and type aliases.
```
iuioiua, line 7: What ideas did you have for these other symbols? I was thinking of classes, methods, and class properties to be as far as we should go...

iuioiua (approved):
Dope. Thank you for this.

_tools/check_docs.ts
lucacasonato: Interfaces methods and type params should have names, same for type params on aliases. For variables there should be an `@example`, and for namespaces we should apply the rules to all nodes in the namespace.

PR #4803, opened 2024-05-21
docs(async): improve docs for stabilization

crowlKats:


async/abortable.ts
```diff
@@ -78,7 +101,11 @@ export function abortable<T>(
  * const p = abortablePromise(request, c.signal);
  *
  * // The below throws if the request didn't resolve in 100ms
- * await p;
+ * try {
+ *   await p;
+ * } catch (e) {
+ *   console.error(e);
+ * }
```
dsherret, line 108: Not sure if these try/catches are worth doing because they add a bit of noise.

dsherret (approved):
LGTM

async/abortable.ts
crowlKats: @dsherret its either doing those, or adding `no-eval`, as else the doc test task will fail. I don't mind either way

crowlKats: discussed with @iuioiua, will go with no-eval

async/abortable.ts
```diff
@@ -4,10 +4,15 @@
...
 /**
- * Make {@linkcode Promise} abortable with the given signal.
+ * Make a {@linkcode Promise} abortable with the given signal.
+ *
+ * @typeParam T The type of the provided and returned promise.
+ * @param p The promise to make abortable.
+ * @param signal The signal to abort the promise with.
+ * @returns A promise that can be aborted.
  *
  * @example
```
iuioiua, line 14: Nit: It would've been nice to have a snippet title to avoid having it as "Example 1" in JSR. Ditto for other `@example` code snippets.

async/debounce.ts
```diff
@@ -36,8 +36,10 @@ export interface DebouncedFunction<T extends Array<unknown>> {
  * // output: Function debounced after 200ms with baz
  * ```
  *
+ * @typeParam T The arguments of the provided function.
```
iuioiua, line 39: Nit: descriptions don't have to have their starting columns aligned.

webgpu/create_capture.ts
```diff
@@ -20,7 +20,7 @@ export interface CreateCapture {
  * Creates a texture and buffer to use as a capture.
  *
  * @example
- * ```ts, no-eval
+ * ```ts no-eval
```
iuioiua, line 23: For context: https://github.com/denoland/deno_doc/issues/583

crowlKats: already fixed in #4813

async/debounce.ts
crowlKats: most parts of the codebase don't have that, only saw that in a single place. some even use `-` as a separator between argument and description.

async/abortable.ts
crowlKats: i feel like if its a single example, it being `Example 1` is fine. most scenarios have only really one example that makes sense to have, and describing it will seem a bit too much. though definitively should be enforced for multiple examples

iuioiua: We've now added example titles to the list of checks for the doc checker. Please feel free to review #4841.

PR #4805, opened 2024-05-21
chore: re-add workspace level `imports`

lucacasonato:
We are going to revert the behaviour that injected this automatically. In the future we can use a special `@workspace` version.

@kt3k Please confirm that this will work correctly with the release process. My understanding is that it will.


kt3k:
> @kt3k Please confirm that this will work correctly with the release process. My understanding is that it will.

Yes, the version upgrade tool can handle that. https://github.com/denoland/bump-workspaces/blob/20a5e674e196869ce820821f109d962160c2474c/util.ts#L364-L367

kt3k (approved):
LGTM

PR #4810, opened 2024-05-21
refactor(ini): various changes

dsherret:
I've collected all these changes in a draft pr to get feedback. Let me know if I should split this up into multiple PRs.

### DIFF
--- a/ini/ini_map.ts
+++ b/ini/ini_map.ts
@@ -185,10 +185,6 @@ export class IniMap {
     return size;
   }
 
-  get formatting(): Formatting {
-    return this.#formatting;
-  }
-
   /** Manage comments in the INI. */
   get comments(): Comments {
     return this.#comments;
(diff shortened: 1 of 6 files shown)


ini/ini_map.ts
```diff
@@ -22,18 +22,10 @@ type Formatting = Omit<FormattingOptions, "lineBreak" | "commentChar"> & {
 
 /** Options for parsing INI strings. */
 export interface ParseOptions {
-  /** The character used to assign a value to a key; defaults to '='. */
-  assignment?: FormattingOptions["assignment"];
```
dsherret, line 26: Not used.

ini/ini_map.ts
```diff
@@ -22,18 +22,10 @@ type Formatting = Omit<FormattingOptions, "lineBreak" | "commentChar"> & {
...
   /** Provide custom parsing of the value in a key/value pair. */
   reviver?: ReviverFunction;
 }
 
-/** Options for constructing INI strings. */
-export interface StringifyOptions extends FormattingOptions {
-  /** Provide custom string conversion for the value in a key/value pair. */
-  replacer?: ReplacerFunction;
-}
```
dsherret, line 35: Moved to `stringify.ts` because only used there.

ini/ini_map.ts
```diff
@@ -88,25 +80,25 @@ export class IniMap {
       }
       return false;
     },
-    getAtLine: (line: number): string | undefined => {
+    getTextAtLine: (line: number): string | undefined => {
```
dsherret, line 83: Recommend naming these with `Text` in the name in order to possibly allow having a `getAtLine` in the future that returns a comment object.

ini/ini_map.ts
```diff
@@ -185,10 +177,6 @@ export class IniMap {
     return size;
   }
 
-  get formatting(): Formatting {
-    return this.#formatting;
-  }
```
dsherret, line 190: This seemed exposed for testing purposes and I don't believe it made much sense otherwise. It seems we can remove it.

ini/ini_map.ts
```diff
@@ -130,48 +122,48 @@ export class IniMap {
...
       }
-      return this.comments;
     },
-    setAtKey: (
+    setTextAtKey: (
       keyOrSection: string,
       textOrKey: string,
       noneOrText?: string,
-    ): Comments => {
+    ) => {
```
dsherret, line 130: Removed the `Comments` return type because it keeps this open to possibly return back different values in the future, but let me know if I should revert this.

ini/ini_map.ts
```diff
@@ -388,34 +380,23 @@ export class IniMap {
...
-            lineBreakLength = 0;
+        if (char === "\r" && text[i + 1] === "\n") {
+          i++;
+          if (!this.#formatting.lineBreak) {
+            this.#formatting.lineBreak = "\r\n";
           }
+        } else if (!this.#formatting.lineBreak) {
+          this.#formatting.lineBreak = char;
         }
-        i += lineBreakLength;
```
dsherret, line 418: This code above was previously incorrectly identifying stuff like `\n\r` as a single line break when it should be two.

PR #4812, opened 2024-05-21
docs(text): add module doc

dsherret:
Reviewed (looks good) then added a module doc for the package.


littledivy (commented):
Can you also add `text/mod.ts` to `_tools/check_docs.ts`?

lucacasonato (approved):
LGTM

PR #4816, opened 2024-05-21
fix(ini): remove unused `ParseOptions.assignment` property

dsherret:
This property is not used.

### DIFF
--- a/ini/ini_map.ts
+++ b/ini/ini_map.ts
@@ -22,8 +22,6 @@ type Formatting = Omit<FormattingOptions, "lineBreak" | "commentChar"> & {
 
 /** Options for parsing INI strings. */
 export interface ParseOptions {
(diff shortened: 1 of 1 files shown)


iuioiua:
Is it not used in the following places?
1. https://github.com/dsherret/deno_std/blob/3637df660968bbf142c4168db47aa7fdaf9c64b0/ini/ini_map.ts#L477
2. https://github.com/dsherret/deno_std/blob/3637df660968bbf142c4168db47aa7fdaf9c64b0/ini/ini_map.ts#L513

dsherret:
@iuioiua no, those are formatting options. This is parse options.

iuioiua (approved):
Oh, my misunderstanding. This can be marked as a fix. It shouldn't break anything.

PR #4818, opened 2024-05-21
BREAKING(ini): remove internal `Formatting` type

dsherret:
Seems like we should just use `FormattingOptions` here?


kt3k (approved):
LGTM

iuioiua (approved):
Nice cleanup. As this is a minor breaking change, the next version of this package should be a minor release, right?

kt3k:
> As this is a minor breaking change, the next version of this package should be a minor release, right?

Sounds correct. Ok, let's tag this as BREAKING

PR #4820, opened 2024-05-21
chore: Add a root module (`mod.ts`) to `@std/encoding`

nathanwhit:
Currently the `@std/encoding` package is one of the few without a root module that re-exports the public symbols.

This PR adds a `mod.ts` file that re-exports the submodules in `encoding`. I diverged a bit from convention by not re-exporting _all_ public symbols, leaving out the `encode` and `decode` functions from `@std/encoding/varint`. Those functions are deprecated, and it seemed confusing to have them exported from the root.

### DIFF
--- a/encoding/README.md
+++ b/encoding/README.md
@@ -1,9 +1,0 @@
-Utilities for encoding and decoding common formats like hex, base64, and varint.
-
-This module is browser compatible.
-
-```ts
-import { encodeBase64 } from "@std/encoding/base64";
-
-encodeBase64("foobar"); // "Zm9vYmFy"
-```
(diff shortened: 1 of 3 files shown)


kt3k:
> I diverged a bit from convention by not re-exporting all public symbols, leaving out the encode and decode functions from @std/encoding/varint. Those functions are deprecated, and it seemed confusing to have them exported from the root.

This is totally reasonable. Thanks for taking care of this!

kt3k (approved):
LGTM. Thanks!

iuioiua (commented):
Awesome! Thank you.

Issue #4826, opened 2024-05-22
to-do: remove use of `protected`, `private` and `public` from `@std/data-structures`


iuioiua:
We should use `#private` or regular properties instead of these access modifiers. This work was essentially done in https://github.com/denoland/deno_std/pull/4793/commits/5b86a65872ecdfcb7454646416577cab4c7811b7 but was then later reverted.

PR #4827, opened 2024-05-22
BREAKING(data-structures): hide private internals

iuioiua:
This commit makes the internals of
`BinarySearchTree` is properly private so that
`BinarySearchNode` is not leaked in public types anymore.

Relands #4794
Towards #4826

### DIFF
--- a/data_structures/binary_search_tree.ts
+++ b/data_structures/binary_search_tree.ts
@@ -241,7 +241,7 @@ export class BinarySearchTree<T> implements Iterable<T> {
    */
   static from<T, U, V = undefined>(
     collection: ArrayLike<T> | Iterable<T> | BinarySearchTree<T>,
-    options?: {
+    options: {
       compare?: (a: U, b: U) => number;
       map: (value: T, index: number) => U;
       thisArg?: V;
--- a/data_structures/binary_search_tree.ts
+++ b/data_structures/binary_search_tree.ts
@@ -116,7 +116,8 @@ export class BinarySearchTree<T> implements Iterable<T> {
    * To create a binary search tree from an array like, an iterable object, or an
    * existing binary search tree, use the {@link BinarySearchTree.from} method.
    *
-   * @param compare A custom comparison function to sort the values in the tree. By default, the values are sorted in ascending order.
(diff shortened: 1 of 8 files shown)


iuioiua:
Tests are failing due to:
1. https://github.com/denoland/deno_std/blob/d934617a98e18baae103afbe369f032861588ffd/data_structures/binary_search_tree_test.ts#L553
2. https://github.com/denoland/deno_std/blob/d934617a98e18baae103afbe369f032861588ffd/data_structures/red_black_tree.ts#L220-L221

data_structures/binary_search_tree.ts
```diff
@@ -205,7 +206,7 @@ export class BinarySearchTree<T> implements Iterable<T> {
    */
   static from<T, U, V = undefined>(
     collection: ArrayLike<T> | Iterable<T> | BinarySearchTree<T>,
-    options: {
```
lucacasonato, line 208: This is wrong. When this overload is used, `options` should be specified. The overload that matches when `options` is not  specified is the first one.

data_structures/red_black_tree.ts
```diff
@@ -219,7 +217,7 @@ export class RedBlackTree<T> extends BinarySearchTree<T> {
     let unmappedValues: ArrayLike<T> | Iterable<T> = [];
     if (collection instanceof RedBlackTree) {
       result = new RedBlackTree(
-        options?.compare ?? (collection as unknown as RedBlackTree<U>).compare,
```
lucacasonato, line 222: This doesn't work? You can not reference private fields in the superclass from a subclass. That would defeat the whole point of private fields.

The static block thing I did is the proper way to "leak" private field or methods for trusted consumers.

data_structures/binary_search_tree.ts
iuioiua: Apologies. Mistake.

data_structures/red_black_tree.ts
iuioiua: Another mistake. Fixed.

lucacasonato (approved):
LGTM

PR #4828, opened 2024-05-22
fix(html): remove `named-entity-list.json` entrypoint

iuioiua:
This entrypoint was added by mistake.

### DIFF
--- a/html/deno.json
+++ b/html/deno.json
@@ -3,7 +3,6 @@
   "version": "0.224.0",
   "exports": {
     ".": "./mod.ts",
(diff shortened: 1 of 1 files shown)


iuioiua:
Hm... It looks like this is exported deliberately and used in https://github.com/denoland/deno_std/blob/20432696d0625da3c51c5898e6495569caec5b8e/html/entities.ts#L69.

kt3k:
Oh, ok. Then probably it's fine to keep..

Note: the original PR had that example https://github.com/denoland/deno_std/pull/3335/files

PR #4829, opened 2024-05-22
docs(fmt): improve API docs

kt3k:
part of #3764


fmt/colors.ts
```diff
@@ -75,8 +75,24 @@ export interface Rgb {
 let enabled = !noColor;
 
 /**
- * Set changing text color to enabled or disabled
- * @param value
+ * Set text color enable or disable.
```
lucacasonato, line 78:
```suggestion
 * Enable or disable text color when styling.
```

fmt/colors.ts
```diff
@@ -86,7 +102,17 @@ export function setColorEnabled(value: boolean) {
...
 
-/** Get whether text color change is enabled or disabled. */
+/**
+ * Get whether text color change is enabled or disabled.
+ *
+ * @example Usage
+ * ```ts
+ * import { getColorEnabled } from "@std/fmt/colors";
+ *
+ * console.log(getColorEnabled());
```
lucacasonato, line 112:
```suggestion
 * console.log(getColorEnabled()); // true if enabled, false if disabled
```

fmt/printf.ts
```diff
@@ -916,11 +916,23 @@ class Printf {
 }
 
 /**
- * Converts and format a variable number of `args` as is specified by `format`.
+ * Converts and formats a variable number of `args` as is specified by `format`.
```
lucacasonato, line 919: Mention that users should look at the `printf` module level doc to view what format strings are supported.

@crowlKats we should really have some way to link to module level docs with `@link`

crowlKats: @lucacasonato I can look into it on friday

fmt/printf.ts
```diff
@@ -916,11 +916,25 @@ class Printf {
...
+ *
+ * assertEquals(sprintf("%d", 9), "9");
+ * assertEquals(sprintf("%o", 9), "11");
+ * assertEquals(sprintf("%f", 4), "4.000000");
+ * assertEquals(sprintf("%.3f", 0.9999), "1.000");
+ * ```
+ *
+ * @param format The format string to use
+ * @param args The arguments to format
+ * @returns formatted string
```
magurotuna, line 937: nit: in other places of `@returns` tag the sentences start with `The`, but here it doesn't

fmt/printf.ts
magurotuna: Do we want to wait until we can link to the module level doc before this PR lands? I think other than that the PR looks good to me

iuioiua (approved):
Some nits. Otherwise, LGTM! Good to see this done.

PR #4834, opened 2024-05-22
chore: improve check_docs.ts

dsherret:
1. Collect all diagnostics and output them in one shot.
1. Output when a snippet takes more than 10s to run.


iuioiua (commented):
Awesome! Thank you.

PR #4835, opened 2024-05-22
fix(cli): reduce flicker in spinner render function

dsherret:


cli/spinner.ts
```diff
@@ -136,13 +136,16 @@ export class Spinner {
...
         noColor
           ? this.#spinner[i] + " " + this.message
           : color + this.#spinner[i] + COLOR_RESET + " " + this.message,
       );
-      Deno.stdout.writeSync(frame);
+      // call writeSync once to reduce flickering
+      const writeData = new Uint8Array(LINE_CLEAR.length + frame.length);
+      writeData.set(LINE_CLEAR);
+      writeData.set(frame, LINE_CLEAR.length);
+      Deno.stdout.writeSync(writeData);
```
iuioiua, line 148: An alternative using [`concat()`](https://jsr.io/@std/bytes@1.0.0-rc.3/doc/concat/~/concat):
```suggestion
      Deno.stdout.writeSync(concat([LINE_CLEAR, frame]));
```

kt3k (approved):
LGTM

cli/spinner.ts
dsherret: Oh, I missed that this was in a separate package. I'm not sure it's worth causing another package load to occur when importing this module.

iuioiua (approved):
It's no big deal omitting `concat()` for just two `Uint8Array`s. LGTM.

PR #4837, opened 2024-05-22
docs(text): pass docs check

dsherret:


_tools/check_docs.ts
```diff
@@ -84,7 +85,7 @@ function assertHasReturnTag(document: { jsDoc: JsDoc; location: Location }) {
   const tag = document.jsDoc.tags?.find((tag) => tag.kind === "return");
   if (tag === undefined) {
     diagnostics.push(
-      new DocumentError("Symbol must have a @return tag", document),
+      new DocumentError("Symbol must have a @returns tag", document),
```
iuioiua, line 88: Nit: this is the most true.
```suggestion
      new DocumentError("Symbol must have a @return or @returns tag", document),
```

iuioiua (approved):

_tools/check_docs.ts
dsherret: Maybe we should standardize on just `@returns`, which is mostly used?

PR #4838, opened 2024-05-22
docs(encoding): Cleanup and fix doc lints

nathanwhit:
Fixes doc lints for the `encoding` package to prepare for stabilization.

While I was there, I also updated examples to use assertions (and caught a couple of incorrect examples in the process).


encoding/varint.ts
```diff
@@ -2,16 +2,24 @@
 // Copyright 2020 Keith Cirkel. All rights reserved. MIT license.
 // Copyright 2023 Skye "MierenManz". All rights reserved. MIT license.
 /**
- * Functions for encoding typed integers in array buffers.
+ * Utilities for {@link https://protobuf.dev/programming-guides/encoding/#varints VarInt} encoding
```
kt3k, line 5: Thanks for adding the reference!

encoding/varint.ts
```diff
@@ -20,8 +28,19 @@
 // This implementation is a port of https://deno.land/x/varint@v2.0.0 by @keithamus
 // This module is browser compatible.
 
+/**
+ * The maximum value of an unsigned 64-bit integer.
+ * Equivalent to `2n**64n - 1n`
```
kt3k, line 33: 👍

kt3k (approved):
LGTM

PR #4839, opened 2024-05-22
doc(path): Fix doc lints in `@std/path` to prepare for stabilization

nathanwhit:


path/is_absolute.ts
```diff
@@ -7,7 +7,22 @@ import { isAbsolute as windowsIsAbsolute } from "./windows/is_absolute.ts";
...
+ *
  * @param path to be verified as absolute
+ * @returns `true` if path is absolute, `false` otherwise
+ *
+ * @example Usage
+ * ```ts
+ * import { isAbsolute } from "@std/path/is-absolute";
+ *
+ * // posix
+ * isAbsolute("/home/foo"); // true
```
nathanwhit, line 19: Ideally we could use assertions in these examples, but I couldn't figure out a nice way to make it work due to the os-dependent behavior.

I suppose we could do something like
```ts
if (Deno.build.os === "windows") {
  // windows specific examples
} else {
  // posix
}
```

but I'm not sure if it's worth the clutter

path/glob_to_regexp.ts
```diff
@@ -9,9 +9,17 @@ import {
   globToRegExp as windowsGlobToRegExp,
 } from "./windows/glob_to_regexp.ts";
 
-export type { GlobOptions };
+export type { GlobOptions, OSType };
```
nathanwhit, line 12: Needed to resolve the lint that `OsType` is private despite appearing in a public type (`GlobToRegExpOptions`).

Alternatively we could inline the definition of `OSType` instead of exporting it, but that sort of defeats the purpose of the type.

path/is_absolute.ts
kt3k: I think the readability matters more here. I'd suggest adding `no-eval` directive when it's difficult to make it work in cross-platform way

path/glob_to_regexp.ts
kt3k: Exporting the type makes sense to me, but I think it should be renamed to `OsType`. The Deno style guide has the below section:

> When the names are in camelCase or PascalCase, always follow the rules of them even when the parts of them are acronyms.
>
> Note: Web APIs use uppercase acronyms (JSON, URL, URL.createObjectURL() etc.). Deno Standard Library does not follow this convention.

https://docs.deno.com/runtime/manual/references/contributing/style_guide#naming-convention

Issue #4802, opened 2024-05-21
multiple entrypoint doesn't make sense for `regexp`, and `jsonc`


kt3k:
These ~~3~~ 2 packages only have single API, but have multiple entrypoint `.` and `<API_name>`. Having 2 entrypoint doesn't make much sense to them.

iuioiua:
I believe that's for the likely event that APIs are added to these packages—we don't need to factor out the existing APIs into separate files, as they're already there. However, I see your point, too.

kt3k:
We decided to keep the current structure as is. This structure makes the future contribution easier.

PR #4842, opened 2024-05-22
BREAKING(regexp): remove `@std/regexp/escape` entrypoint

iuioiua:
This package has only one API: `escape()`. There's currently no reason to have another entry point that's functionally identical to the root entry point.

In other words, the `@std/regexp/escape` entry point is identical to `@std/regexp`, so we removed `@std/regexp/escape`.

Towards #4802


kt3k:
How about keeping the current file structure, but only remove `escape` entry in `deno.json`? I think that will achieve that same thing and that also prepare better for the future contribution of another API

iuioiua:
With the same reasoning, we could also just keep it as is, which I'm fine with too 😆

kt3k:
Ok. I changed my mind. Let's keep the current structure as is.

Thanks for working on this anyway

PR #4843, opened 2024-05-23
BREAKING(jsonc): remove `@std/jsonc/parse` entrypoint

iuioiua:
This package has only one API: `parse()`. There's currently no reason to have another entry point that's functionally identical to the root entry point.

Put differently, the `@std/jsonc/parse` entry point is identical to `@std/jsonc`, so we removed `@std/jsonc/parse`.

Towards #4802


iuioiua:
We've decided against this change. Either way, the Super Linter failure is interesting and will need to be solved at some point.

PR #4844, opened 2024-05-23
test(cli): reduce the flakiness of `Spinner` test cases

### LINKED ISSUE #4775: bug: `Spinner constructor accepts interval` test is flaky
See https://github.com/denoland/deno_std/actions/runs/9153952807/job/25163712071

Perhaps we should re-assess our approach to these tests.

kt3k:
closes #4775

### DIFF
--- a/cli/spinner_test.ts
+++ b/cli/spinner_test.ts
@@ -60,7 +60,7 @@ Deno.test("Spinner constructor accepts interval", async () => {
   const actual2 = normalizeString(text2);
 
   // give setInterval a good buffer to avoid needlessly failing
-  assertGreater(actual2.length, 100);
+  assertGreater(actual2.length, 50);
   assertLess(actual2.length, 300);
 });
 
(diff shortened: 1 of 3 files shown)


cli/spinner_test.ts
```diff
@@ -132,11 +132,8 @@ Deno.test("Spinner constructor accepts each color", async () => {
 Deno.test("Spinner.color can set each color", async () => {
   const text = await spawnDeno(["cli/testdata/spinner_cases/set_color.ts"]);
 
-  const expectedStr = `${LINE_CLEAR}\u001b[30m⠋${COLOR_RESET} ` + // Black
-    `${LINE_CLEAR}\u001b[31m⠙${COLOR_RESET} ` + // Red
-    `${LINE_CLEAR}\u001b[32m⠹${COLOR_RESET} `; // Green
-
-  assertStringIncludes(text, expectedStr);
+  assertStringIncludes(text, `${LINE_CLEAR}\u001b[30m⠋${COLOR_RESET} `); // includes black spinner
```
iuioiua, line 135: We could use `black()` from `std/fmt/colors`, right?

iuioiua (approved):

cli/spinner_test.ts
kt3k: Good point, but it looks like `std/fmt/colors` and `std/cli/spinner` use different codes (`0` vs `39`) for resetting the colors. I'm not sure it's fine to align them. Leaving this as is for now.

iuioiua: That's strange. Can we create an issue for that so we can investigate further?

PR #4846, opened 2024-05-23
docs(semver): improve docs

kt3k:
part of #3764 

This improves the API docs of `@std/semver`

### DIFF
--- a/semver/can_parse.ts
+++ b/semver/can_parse.ts
@@ -3,7 +3,7 @@
 import { parse } from "./parse.ts";
 
 /**
- * Returns true if the string can be parsed as semver.
+ * Returns true if the string can be parsed as SemVer.
  *
  * @example Usage
  * ```ts
@@ -15,7 +15,7 @@ import { parse } from "./parse.ts";
  * ```
  *
  * @param version The version string to check
- * @returns `true` if the string can be parsed as semver, `false` otherwise
+ * @returns `true` if the string can be parsed as SemVer, `false` otherwise
  */
 export function canParse(version: string): boolean {
   try {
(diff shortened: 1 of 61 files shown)


semver/can_parse.ts
```diff
@@ -2,6 +2,21 @@
 // This module is browser compatible.
 import { parse } from "./parse.ts";
 
+/**
+ * Returns true if the string can be parsed as semver.
+ *
+ * @example Usage
+ * ```ts
+ * import { canParse } from "@std/semver";
```
magurotuna, line 10: I'm wondering if there is a general guideline that we can follow on whether we should put module-level import (i.e. `from "@std/semver"`) or one deeper one (i.e. `from "@std/semver/can-parse"`). Do you have any thought?

kt3k: Currently we don't have explicit policy about it

iuioiua: We should use the most specific import specifier. It encourages keeping dependencies leaner and aligns with JSR.

semver/range_max.ts
```diff
@@ -44,15 +44,21 @@ function comparatorMax(comparator: Comparator): SemVer {
 }
 
 /**
+ * The maximum valid SemVer for a given range or INVALID
+ *
+ * @example Usage
+ * ```ts
+ * // This API is deprecated
```
iuioiua, line 51: Perhaps, we can have two snippets in this example; one for the deprecated usage and one for the new API's usage, like a migration guide.

iuioiua (changes requested):
Great to have this done! My main nit is the proper casing of "SemVer".

iuioiua (approved):
Nice!

PR #4847, opened 2024-05-23
docs(data-structures): enables doc lint of data-structures

kt3k:
part of #3764

### DIFF
--- a/data_structures/red_black_tree.ts
+++ b/data_structures/red_black_tree.ts
@@ -34,7 +34,7 @@ const {
  * | min()         | O(log n)     | O(log n)   |
  * | max()         | O(log n)     | O(log n)   |
  *
- * @example
+ * @example Usage
  * ```ts
  * import {
  *   ascend,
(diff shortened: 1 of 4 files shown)


data_structures/red_black_tree.ts
```diff
@@ -101,8 +101,6 @@ const {
  * @typeparam T The type of the values being stored in the tree.
  */
 export class RedBlackTree<T> extends BinarySearchTree<T> {
-  declare protected root: RedBlackNode<T> | null;
```
iuioiua, line 104: Is this breaking?

iuioiua (approved):

data_structures/red_black_tree.ts
kt3k: Technically yes, but it's leftover from https://github.com/denoland/deno_std/pull/4827 and breakingness of this change is covered by #4827 because this one and #4827 belong to the same release. So I avoid tag this commit as 'BREAKING'

PR #4851, opened 2024-05-24
test(yaml): test handling of omap

kt3k:
This PR adds test cases of `yaml`, which exercises the handling `omap` type in yaml string.

related: #3713

### DIFF
--- a/yaml/_type/omap.ts
+++ b/yaml/_type/omap.ts
@@ -36,7 +36,7 @@ function resolveYamlOmap(data: Any): boolean {
 }
 
 function constructYamlOmap(data: Any): Any {
-  return data !== null ? data : [];
+  return data;
 }
 
 export const omap = new Type("tag:yaml.org,2002:omap", {
--- a/yaml/parse_test.ts
+++ b/yaml/parse_test.ts
@@ -240,3 +240,31 @@ Deno.test({
     });
   },
 });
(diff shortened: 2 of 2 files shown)


yaml/_type/omap.ts
```diff
@@ -36,7 +36,7 @@ function resolveYamlOmap(data: Any): boolean {
 }
 
 function constructYamlOmap(data: Any): Any {
-  return data !== null ? data : [];
+  return data;
 }
```
iuioiua, line 40: This function seems to do nothing. Can we remove it?

yaml/parse_test.ts
```diff
@@ -240,3 +240,31 @@ Deno.test({
...
+`;
+    assertEquals(parse(yaml), [
+      { "Mark McGwire": 65 },
+      { "Sammy Sosa": 63 },
+      { "Ken Griffey": 58 },
+    ]);
+
+    // Invalid omap
+    // map entry is not an object
+    assertThrows(() => parse("--- !!omap\n"));
```
iuioiua, line 260: Can you please add the error class and message arguments for `assertThrows()`? Ditto for the other instances.

iuioiua (changes requested):

iuioiua (approved):
Nice

PR #4861, opened 2024-05-27
BREAKING(encoding): replace `Ascii85Options` with `EncodeAscii85Options` and `DecodeAscii85Options`

iuioiua:
This change deprecates `Ascii85Options` in favor of `EncodeAscii85Options` and `DecodeAscii85Options`. Previously, these were used for both `encodeAscii85()` and `decodeAscii85()`, but the latter didn't use the `delimiter` property.


kt3k:
Looks like technically a breaking change. Let's land together with `1.0.0-rc.1` PR to not break the user's program.

iuioiua:
What does this break?

kt3k:
Doesn't `decodeAscii85(str, { delimiter: true })` cause type error now?

kt3k (approved):
LGTM

kt3k (approved):
LGTM again!

PR #4866, opened 2024-05-29
chore: implement custom resolver for Super Linter

iuioiua:
Previously, inter-package module imports broke the Super Linter. This change fixes that.

Pre-requisite for #4790


kt3k (approved):
LGTM

PR #4870, opened 2024-05-29
chore: draw only direct dependencies in dep diagram

kt3k:
Currently the tool draws the arrows for direct and transitive dependencies. This PR updates it to only draw arrows for the direct dependencies.

This makes the diagram a bit easier to read.


aapoalas (approved):

iuioiua (approved):
Nice!

PR #4872, opened 2024-05-29
deprecation(uuid): rename `V1Options` to `GenerateOptions`

iuioiua:
The convention in this codebase is for options interfaces to be named `XOptions`, where `X` is the name of the symbol that the options interface belongs to.


iuioiua:
I realize we don't export this interface from `uuid/mod.ts`. How do we do that again?

kt3k (approved):
Good catch. Makes sense to me. LGTM

PR #4873, opened 2024-05-29
chore: don't require `@returns` tag when return type is type assertion

kt3k:
When the return type of a function is type assertion like `asserts expr`, the actual return value is undefined. We should skip requiring `@returns` doc in that case.

Example:
https://github.com/denoland/deno_std/blob/0b95598ec109a35f679b78cd92eb9573aa34ad27/assert/assert.ts#L16


iuioiua (approved):
Good catch.

Issue #4865, opened 2024-05-28
suggestion: don't use `std/assert` in implementation code


iuioiua:
Using assertions from `std/assert` within implementation code in the codebase has a few drawbacks:
1. In cases where the user needs to error handle an API that uses assertions from `std/assert`, they need to import and use `AssertionError`.
2. While error messages provide the majority of the context behind an error, the constructor name does, too. I.e. `RangeError`. `AssertionError` is generic and doesn't add any context to the error.
3. It introduces an added dependency. In most cases, this doesn't matter. But for some, it does.

For these reasons, I suggest that we remove the use of assertions from `std/assert` within the implementation code in the Standard Library and instead use [native error classes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) where possible. In cases where these native error classes do not fit, we can create and export custom error classes that extend `Error`. Doing so requires minimal code changes and can be more explicit than calling a generic assertion.

- [x] #4950 
- [x] #4948
- [x] #4966 
- [x] #4968
- [x] #4973
- [x] #4974
- [x] #4975
- [x] #4976
- [x] #4977
- [x] #4978
- [x] #4979
- [x] #4980
- [ ] Document the design decision

PR #4874, opened 2024-05-29
fix(uuid): validate namespace UUIDs in `v3.generate()` and `v5.generate()`

iuioiua:
Previously, these `generate()` functions only checked if the namespace UUIDs were the correct length and nothing more. In other words, providing a string with the proper length but incorrect UUID format would go uncaught. It also threw an `AssertionError`, meaning that handling that error, if using these functions, would require importing `AssertionError` from `@std/assert.`

These functions now check that the namespace UUID is a valid UUID and throw a `TypeError` if it is not, making error handling a little easier, as it's a native JS error.

This is a breaking fix, as the class and message of the thrown error have changed.

Related #4865

### DIFF
--- a/uuid/_common.ts
+++ b/uuid/_common.ts
@@ -2,8 +2,11 @@
 // This module is browser compatible.
 
 /**
- * Converts the byte array to a UUID string
- * @param bytes Used to convert Byte to Hex
+ * Converts a hex byte array to a UUID string.
+ *
+ * @param bytes Byte array of the UUID.
+ *
+ * @returns UUID string.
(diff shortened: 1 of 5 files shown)


kt3k:
> It also threw an AssertionError, meaning that handling that error, if using these functions, would require importing AssertionError from @std/assert

I think this `assert` usage was incorrect. `assert` shouldn't be used for checking the user's input. So I'm in favor of replacing it with TypeError

kt3k (approved):
LGTM

PR #4875, opened 2024-05-29
test(uuid): add tests for invalid namespace UUID

iuioiua:
Notice: These `generate()` functions only check the length of the namespace UUID and not its format. This behavior will be improved in #4874.


kt3k (approved):
LGTM

PR #4876, opened 2024-05-29
docs(assert): improve docs

kt3k:
This PR improves the API docs of `assert` package. Now the `@param` `@returns` `@typeParam` are fully documented. It passes doc-linter check.

part of #4600 #3764

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -26,6 +26,7 @@ type DocNodeWithJsDoc<T = DocNodeBase> = T & {
 };
 
 const ENTRY_POINTS = [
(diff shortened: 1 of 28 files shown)


iuioiua (changes requested):
Instead of not evaluating the example code snippets, in most cases, we could instead remove LOC that throw and keep LOC that don't throw.

kt3k:
> Instead of not evaluating the example code snippets, in most cases, we could instead remove LOC that throw and keep LOC that don't throw.

I think it's good to give explicit example of the case where it throws.

iuioiua (approved):
Nice!

PR #4877, opened 2024-05-29
BREAKING(uuid): remove `v1.generate()` signature with `buf` and `offset` parameters and `number[]` return type

iuioiua:
This PR removes the optional `buf` and `offset` arguments and `number[]` return type from `v1.generate()`. These settings are unneeded in almost all cases and gave the function a shape that differed from the other `generate()` functions within `@std/uuid`.


kt3k:
This probably makes sense. Also optional 3 arguments breaks the style guide of the standard library (There should be only one optional argument at most https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object

kt3k (approved):
LGTM

PR #4879, opened 2024-05-29
fix(net): check `scopeid === 0` in `getNetworkAddress()`

kt3k:
According to https://github.com/mafintosh/network-address/pull/3 the scopeid needs to be 0 for the address to be reachable if the family is `IPv6`.

(This is a PR targeting the feature branch, not main)


iuioiua (approved):
Thanks for getting to this before me. I added tests that ensure that the return network address is listenable.

PR #4880, opened 2024-05-29
deprecation(uuid): deprecate `v1.generate()` signature with `buf` and `offset` parameters

iuioiua:
This PR deprecates the `v1.generate()` signature with the optional `buf` and `offset` arguments and `number[]` return type. These settings are unneeded in almost all cases and gave the function a shape that differed from the other `generate()` functions within `@std/uuid`. The function also went against Deno's Style Guide, which required functions to have only one optional argument (see https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object).

Pre-requisite for #4877

### DIFF
--- a/uuid/v1.ts
+++ b/uuid/v1.ts
@@ -103,7 +103,41 @@ export interface GenerateOptions {
  * const uuid = generate(options) as string;
  * assert(validate(uuid));
  * ```
+ *
+ * @deprecated This will be removed in 1.0.0. Use the other overload instead.
  */
+export function generate(
+  options?: GenerateOptions,
+  buf?: number[],
+  offset?: number,
+): string | number[];
+/**
+ * Generates a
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

kt3k:
BTW I think this can be skipped if the next version of `@std/uuid` is `1.0.0-rc.1`

iuioiua:
We should merge this and other non-breaking changes for `@std/uuid@0.225.0`. Then we can do `@std/uuid@1.0.0-rc.1`, including the breaking changes. WDYT?

kt3k:
> We should merge this and other non-breaking changes for @std/uuid@0.225.0. Then we can do @std/uuid@1.0.0-rc.1, including the breaking changes. WDYT?

That works too. Ok. Let's do that.

PR #4886, opened 2024-05-29
chore: exclude generated script from coverage report

kt3k:
This PR excludes the generated script (by wasmbuild) from coverage report.

We don't develop these generated files. Measuring coverage of these files doesn't make much sense.

This improves the coverage of `std/crypto`.


iuioiua (approved):
Good idea

iuioiua:
Can't this instead be done in some ignore property in `deno.json`?

kt3k:
> Can't this instead be done in some ignore property in deno.json?

I don't think it's possible right now

PR #4887, opened 2024-05-29
test(crypto): test node.js custom inspection of `KeyStack`

kt3k:
This PR updates the test case of Node.js custom inspection.

Deno uses Deno's custom inspection if `Symbol.for("Deno.customInspect")` is defined. We need to overwrite it with undefined to invoke the custom Node.js inspection. ( https://github.com/denoland/deno/blob/4f9b23b3664578c2bf48415db246fb21e49abddb/ext/console/01_console.js#L451-L465 )

This improves the code coverage of `crypto` package.

### DIFF
--- a/crypto/unstable_keystack_test.ts
+++ b/crypto/unstable_keystack_test.ts
@@ -248,9 +248,25 @@ Deno.test({
   async fn() {
     const { inspect } = await import("node:util");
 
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Nice!

PR #4890, opened 2024-05-29
docs(http): added `--allow-sys` to file server example

sant123:
With this new [change](https://github.com/denoland/deno_std/pull/4604), running the following script asks the user to approve Deno.networkInterfaces.

`deno run --allow-net --allow-read jsr:@std/http/file-server`

image

Certainly they are required so not having access ends the application:

image

Having --allow-sys fixes the prompt

image

### DIFF
--- a/http/mod.ts
+++ b/http/mod.ts
@@ -8,7 +8,7 @@
  * A small program for serving local files over HTTP.
  *
  * ```sh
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Thank you!

PR #4893, opened 2024-05-30
BREAKING(assert,testing): remove `formatter` option from `assertEquals()`

### LINKED ISSUE #4891: to-do: investigate the viability of `formatter` option for `assertEquals()`
We need to decide whether this option is to be kept before achieving stabilization. See [documentation](https://jsr.io/@std/assert/doc/~/assertEquals).

One clear point against it is that it's a fourth argument for `assertEquals()`, which goes against the Deno Style Guide of having two arguments and a third option argument, max (see https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object). If we are to keep it, we should merge `msg` and `formatter` into a third options argument.

iuioiua:
It seems this option is used in testing `@std/testing/snapshot`. I'm not sure how we should proceed, @kt3k.

Closes #4891


kt3k:
I think https://github.com/denoland/deno_std/pull/4930 unblocks this PR

testing/asserts.ts
```diff
@@ -92,18 +92,15 @@ export function assertArrayIncludes<T>(
...
  * @deprecated This will be removed in 1.0.0. Import from
  * {@link https://jsr.io/@std/assert | @std/assert} instead.
  */
 export function assertEquals<T>(
   actual: T,
   expected: T,
   msg?: string,
-  options: { formatter?: (value: unknown) => string } = {},
 ) {
-  asserts.assertEquals<T>(actual, expected, msg, options);
```
kt3k, line 106: How about leaving this unchanged because this API is already deprecated and removed at 1.0.0-rc.1

iuioiua: I think we should keep the APIs between `@std/assert` and `@std/testing/asserts` synchronized, as it makes any migration seamless. Also, keeping this unchanged would mean I'd have to re-implement `assertEquals()` here in its previous state.

kt3k: Ah ok. It depends on `@std/assert`.. I missed it.

kt3k (approved):
LGTM

PR #4900, opened 2024-05-30
docs(path): improve API docs

kt3k:
part of #4600 #3764

supercedes #4839


path/basename.ts
```diff
@@ -20,6 +20,7 @@ import { basename as windowsBasename } from "./windows/basename.ts";
  *
  * @param path - path to extract the name from.
  * @param [suffix] - suffix to remove from extracted name.
```
iuioiua, line 22: These nits appear throughout the package.
```suggestion
 * @param path Path to extract the name from.
 * @param suffice Suffix to remove from extracted name.
```

path/basename.ts
```diff
@@ -20,6 +20,7 @@ import { basename as windowsBasename } from "./windows/basename.ts";
...
+ * @returns basename of the path.
```
iuioiua, line 23: Nit: proper casing
```suggestion
 * @returns The basename of the path.
```

iuioiua (changes requested):
I see quite a few instances of improper casing and partial sentences. Can you please fix those?

path/basename.ts
```diff
@@ -6,20 +6,27 @@ import { basename as posixBasename } from "./posix/basename.ts";
...
  *
- * basename("/home/user/Documents/"); // "Documents"
- * basename("C:\\user\\Documents\\image.png"); // "image.png"
- * basename("/home/user/Documents/image.png", ".png"); // "image"
+ * if (Deno.build.os === "windows") {
+ *   assertEquals(basename("C:\\user\\Documents\\image.png"), "image.png");
+ * } else {
+ *   assertEquals(basename("/home/user/Documents/image.png"), "image.png");
+ * }
  * ```
```
iuioiua, line 24: What if we wrote the OS-dependent functions like this? It clearly shows the differences in behaviors between difference OS systems.

iuioiua (changes requested):
I've fixed quite a few nits and added a few improvements. Though, I think there's a bit more work to do here:
1. Mention the differences in behavior for OS-dependent APIs.
2. Add `../path/posix/mod.ts` and `../path/windows/mod.ts` to the Super Linter, and document everything in those directories. We should get top-level documentation to a good state, then just copy the relevant details over.
3. Improve the example code snippets (see suggestion).

path/basename.ts
kt3k: I'm not sure it's a good idea to use Deno specific API in examples as we specify this package as cross-runtime.

iuioiua: Does `isWindows` evaluate to the correct OS on Node?

kt3k: Currently no. But I think we should add that support.

iuioiua: I agree; we should add that support. But in the meantime, this package is only partially cross-platform and we should remove the other runtimes from the displayed list on JSR until it's truly cross-platform. I think it'd be good to have the assertions in these top-level functions, for now, until that's solved, as they guarantee correctness. My opinion isn't strong enough to block this PR. Up to you.

iuioiua: See #4914

kt3k: Removed Node and Bun compatibility on JSR.

Added assertions to examples.

91pornaaaaa:
https://91porna.com/

iuioiua (approved):
Big win! Great work!

PR #4903, opened 2024-05-30
chore: exclude 1st level items in `path` package from coverage report

kt3k:
These files are all in the below form:

```ts
export function basename(path: string, suffix = ""): string {
  return isWindows
    ? windowsBasename(path, suffix)
    : posixBasename(path, suffix);
}
```

It's difficult to cover this function in testing because it depends on the platform. Also it's not important to track the coverage of these files because they don't have particular logic in it except branching by platform.

I think we should exclude these files from coverage report.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -65,7 +65,7 @@
     "typos": "typos -c ./.github/workflows/typos.toml",
     "build:crypto": "deno task --cwd crypto/_wasm wasmbuild",
     "wasmbuild": "deno run -A jsr:@deno/wasmbuild@0.17.1 --js-ext mjs --sync",
-    "cov": "deno coverage --ignore=\"**/*.generated.mjs\"",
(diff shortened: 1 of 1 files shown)


iuioiua (commented):
I know it's a bit of work, but I think we should mock `Deno.build.os` instead.

kt3k:
You can't mock `isWindows` as it's defined in the top level of the module https://github.com/denoland/deno_std/blob/f8f5444478ee5aef67d68ef939f6b011854671cb/path/_os.ts#L31

iuioiua:
Allow me to investigate an alternative. I don't think we should do this. That OS-switching logic should be tested rather than ignored.

91pornaaaaa:
https://91porna.com

iuioiua:
I'm closing this as I don't see a good reason to ignore the coverage of these lines. It's fine if we don't have 100% test coverage. We can work on a solution later.

PR #4904, opened 2024-05-30
test(path): add test cases of `isAbsolute()`, `joinGlobs()`, and `common()`

kt3k:
part of #3713

This improves the test coverage of `isAboslute`, `joinGlobs`, and `common` in `path` package.


iuioiua (approved):
Nice

PR #4905, opened 2024-05-30
chore: remove `RUST_BACKTRACE=1` from `test` task

iuioiua:
I realise this actually makes troubleshooting harder because diagnostics are difficult to understand. We can instead temporarily enable this environment variable when needed.


kt3k (approved):
LGTM

PR #4906, opened 2024-05-30
chore: enable `deno test --clean`

iuioiua:
This cleans the coverage directory everytime `deno task test` is run. Contributors who haven't manually deleted this directory in a while can expect this to take some time to execute on the first run.


kt3k:
How about splitting task to `test` (without coverage) and `test-cov` (with coverage)? It's unnecessary overhead to generate coverage every time when the contributor doesn't care about the coverage.

iuioiua:
I'd rather keep tasks simple. Having coverage doesn't hinder anyone.

kt3k:
Fair enough

PR #4907, opened 2024-05-30
refactor(crypto): cleanup redundant assertion in `timingSafeEqual()`

iuioiua:
The assertion wasn't needed as `a` and `b` were coerced into `DataView`s anyway. This removes the dependency on `@std/assert` (see graph).


kt3k (approved):
LGTM. Nice clean up!

PR #4911, opened 2024-05-31
chore: ensure code examples use assertions in Doc Linter

iuioiua:
This change adds the ability for the Super Doc Linter to check whether code snippets contain assertions. This change can be skipped for a given code snippet by adding the `no-assert` directive to the starting code snippet delimiter. For example:
````
```ts no-assert
// Code snippet
```
````
Docs across the entire codebase have been re-written to satisfy this check.


deno.json
iuioiua, line 1: @kt3k, these changes fix the `lint:docs` flakiness. Does this break releases or anything related?

kt3k (approved):
LGTM

Issue #4885, opened 2024-05-29
stabilize `@std/crypto`


kt3k:
part of #4600 

- [x] fully documented #4884
- [x] fully tested
- [x] no ongoing breaking change discussion
- [x] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

---
related issues:
- none

PR #4912, opened 2024-05-31
chore(crypto): release `crypto@1.0.0-rc.1`

iuioiua:
Towards #4885


kt3k (approved):
Looked through all public APIs. They looked reasonable to me.

LGTM

PR #4915, opened 2024-05-31
feat(encoding/unstable): adds streaming versions for hex, base32, base32hex, base64 and base64url

BlackAsLight:
Jumping from https://github.com/denoland/deno_std/pull/4756, closes https://github.com/denoland/deno_std/issues/3087

Essentially merging `@doctor/encoding-stream` into `@std/encoding`.


iuioiua (changes requested):
Tidily done so far! Can you please add these modules to new `<encoding>-stream` exports? E.g., `@std/encoding/hex-stream`. The reasoning is mostly for greater visibility.

BlackAsLight:
> Tidily done so far! Can you please add these modules to new `<encoding>-stream` exports? E.g., `@std/encoding/hex-stream`. The reasoning is mostly for greater visibility.

I definitely can but it looked like here https://github.com/denoland/deno_std/pull/4756#issuecomment-2121787656 that you'd agreed it should be in the same module.

iuioiua:
Ah, to clarify, I meant within the `@std/encoding` package. Sorry for the confusion.

BlackAsLight:
I pulled them out into their own module, but it does seem a bit harder to maintain since so many of the files start with the same name `base64`

iuioiua:
FYI @BlackAsLight, in case you're unaware, we haven't forgotten about this PR. We will take a closer look once `@std/encoding@1.0.0` is released.

BlackAsLight:
I assumed as much, I think it might be good to wait until after https://github.com/denoland/deno_std/pull/4931 is merged in as well though so I can then also add a `Base32HexStream`.

iuioiua:
FYI #4931 is now merged.

iuioiua (approved):
I cleaned up the examples and tests, and added unstable notices. Now, LGTM! Thank you for this work! Much appreciated.

PR #4928, opened 2024-06-02
BREAKING(path): remove `GlobToRegExpOptions.os` and `OSType` export

iuioiua:
### What's changed

This change removes the `os` property from the `GlobToRegExpOptions` interface and the `options` interface for `globToRegExp()` and also removes `OSType` from the public API.

### Why this change was made

The `os` property controlled which OS `globToRegExp()` worked for. More specifically, it controlled how it delimited paths, as path delimiters change between OSs. This was the only function in `@std/path` that had this functionality, as other functions instead detected the OS of the current machine and then used the appropriate OS-specific logic. Since the `os` property had the `OSType` type, it's now been removed from the public API.

### Migration guide

If you'd like `globToRegExp()` to automatically work for the current machine's OS and didn't use the `os` option, there's no need to change your code.

However, if you'd like to use have `globToRegExp()` work for a specific OS, import the OS-specific version of the function. For POSIX systems:

```diff
- import { globToRegExp } from "@std/path/glob-to-regexp";
+ import { globToRegExp } from "@std/path/posix/glob-to-regexp";

- globToRegExp("*.js", { os: "posix" });
+ globToRegExp("*.js");
```

For Windows systems:

```diff
- import { globToRegExp } from "@std/path/glob-to-regexp";
+ import { globToRegExp } from "@std/path/windows/glob-to-regexp";

- globToRegExp("*.js", { os: "windows" });
+ globToRegExp("*.js");
```


kt3k (approved):
LGTM

PR #4930, opened 2024-06-02
refactor(testing): do not use assertEquals in snapshot.ts

kt3k:
Dependency to `assertEquals` from `testing/snapshot` was introduced in https://github.com/denoland/deno_std/pull/3563 to avoid using non public API of `assert` from `testing` (That was necessary for a part of JSR migration).

Now those internal APIs `diff`, `diffStr`, and `buildMessage` are available from `std/internal`. This PR (partially) reverts #3563 and make `testing/snapshot` directly use those APIs from `std/internal`.

This unblocks #4893


iuioiua (approved):
Nice

Issue #4917, opened 2024-05-31
suggestion: add base32hex to encoding


BlackAsLight:
Since the [spec](https://www.rfc-editor.org/rfc/rfc4648.html) for [`@std/encoding`](https://jsr.io/@std/encoding) specifies a [base32hex](https://www.rfc-editor.org/rfc/rfc4648.html#section-7) format, I think it would be reasonable to also include this variation in the [`@std/encoding`](https://jsr.io/@std/encoding) lib. The only difference between [base32hex](https://www.rfc-editor.org/rfc/rfc4648.html#section-7) and [base32](https://www.rfc-editor.org/rfc/rfc4648.html#section-6) is the alphabet used.

91pornaaaaa:
https://91porna.com/

BenMcLean981:
I think I have this working in my forked copy. Will create a PR soon.

PR #4931, opened 2024-06-02
feat(encoding/unstable): `decodeBase32Hex()` and `encodeBase32Hex()`

BenMcLean981:
As requested in #4917 I have added base32hex encoding and decoding.

I first created new tests from the base32 test cases and using this tool:
https://tomeko.net/online_tools/base32hex.php?lang=en

Once those tests passed, I extracted the common base32 and base32 hex into a `_utils`module with a new parameter for a lookup table.

This is my first time contributing to an open source project and using deno. I am having trouble running the `task ok`. It is failing on my documentation examples with the error below. I think I'm missing a step though, because it's executing with the published module from jsr rather than the local version. This means it makes sense that my module is not there, because it won't be in version 0.224.3. If there's any step I'm missing here I'd really appreciate any feedback.

```
[error] Failed to execute snippet: 
import { decodeBase32Hex } from "@std/encoding/base32hex";
import { assertEquals } from "@std/assert/assert-equals";

assertEquals(
  decodeBase32Hex("GZRTMMDDGA======"),
  new TextEncoder().encode("6c60c0"),
);
error: Unknown export './base32hex' for '@std/encoding@0.224.3'.
  Package exports:
 * .
 * ./ascii85
 * ./base32
 * ./base58
 * ./base64
 * ./base64url
 * ./hex
 * ./varint
 at file:///home/benmclean/projects/deno/deno_std/encoding/base32hex.ts:54
[error] Failed to execute snippet: 
import { encodeBase32Hex } from "@std/encoding/base32hex";
import { assertEquals } from "@std/assert/assert-equals";

assertEquals(encodeBase32Hex("6c60c0"), "GZRTMMDDGA======");
error: Unknown export './base32hex' for '@std/encoding@0.224.3'.
  Package exports:
 * .
 * ./ascii85
 * ./base32
 * ./base58
 * ./base64
 * ./base64url
 * ./hex
 * ./varint
 at file:///home/benmclean/projects/deno/deno_std/encoding/base32hex.ts:74
2 errors found
```

Closes #4931


assert/assert_almost_equals.ts
```diff
@@ -38,7 +38,7 @@ export function assertAlmostEquals(
   }
 
   const msgSuffix = msg ? `: ${msg}` : ".";
-  const f = (n: number) => Number.isInteger(n) ? n : n.toExponential();
+  const f = (n: number) => (Number.isInteger(n) ? n : n.toExponential());
```
iuioiua, line 41: Please revert this. This is unrelated to the objective of this PR.

iuioiua (commented):
Thank you for this. `@std/encoding is about to be in its release candidate phase, so it'll be in a change freeze (except for documentation improvements and bug fixes). This change will be merged once that's complete.

iuioiua (changes requested):
Can you please add the new module to the exports in `encoding/deno.json`?

BenMcLean981:
> Can you please add the new module to the exports in `encoding/deno.json`?

I actually already did, I didn't add the `_utils.ts` module though because it should be private.

BlackAsLight:
The only error left seems to be that you forgot to update the assertEquals in the jsdoc examples when you copied it over.

BenMcLean981:
> The only error left seems to be that you forgot to update the assertEquals in the jsdoc examples when you copied it over.

Good catch, this has been fixed, thanks.

iuioiua:
Hey there. Sorry for the delay. On second thought, with the only difference being the alphabet used, I think we should add an `options` argument to the base32 functions, instead of creating new symbols and modules. It could look like this.
```ts
interface Base32Options {
  /**
   * Base32 format to use.
   *
   * @default {"standard"}
   */
  format?: "standard" | "hex";
}
```
What do y'all think?

iuioiua:
Actually, we can have separate functions but have them all in the existing `base32` module, since there's so much overlap.

BlackAsLight:
I think the best way to do it would be the same way base64 and base64url are done.

iuioiua:
base32 and base32Hex functions have almost complete overlap - there's no benefit to separating them. Arguably, the same could be said for base32Url, but that's a separate conversation.

BlackAsLight:
I don't have an opinion one way or the other on how it should be done, but I think consistency is best.

iuioiua (approved):
I chatted with Yoshiya, and on second thought, we're happy to proceed with this format. I also made a couple of tweaks. Now, LGTM! Thank you very much.

PR #4933, opened 2024-06-03
docs(cli,csv,datetime,fmt,fs,http,ini,semver,testing): assert optional properties on types/interfaces have `@default` tag

iuioiua:
This change adds support for asserting that all optional interface properties have a `@default` JSDoc tag defined to the doc linter.

It seems that the most appropriate tag for some of these properties is `@default {undefined}`. However, I consider this boilerplate. So, I'm thinking of removing these lines across the codebase, then disabling this new assertion. We can then re-enable the assertion whenever we'd like to update these `@default` JSDoc tags again. WDYT? Happy to hear any ideas on how we can ignore this rule.


iuioiua:
PTAL @kt3k

fs/expand_glob.ts
```diff
@@ -19,9 +19,17 @@ const isWindows = Deno.build.os === "windows";
 
 /** Options for {@linkcode expandGlob} and {@linkcode expandGlobSync}. */
 export interface ExpandGlobOptions extends Omit<GlobOptions, "os"> {
-  /** File path where to expand from. */
+  /**
+   * File path where to expand from.
+   *
+   * @default {Deno.cwd}
```
kt3k, line 25:
```suggestion
   * @default {Deno.cwd()}
```

kt3k (approved):
LGTM

async/deadline.ts
```diff
@@ -8,7 +8,11 @@ import { abortable } from "./abortable.ts";
 
 /** Options for {@linkcode deadline}. */
 export interface DeadlineOptions {
-  /** Signal used to abort the deadline. */
+  /**
+   * Signal used to abort the deadline.
+   *
+   * @default {undefined}
```
iuioiua, line 14: `undefined` will now show up in docs as the default value in quite a few of these cases. I consider that boilerplate and wanted to remove those instances before merging this. Does that all sound good, @kt3k?

kt3k: Sounds reasonable, but doesn't that cause a problem with the linter?

kt3k: Sounds good

> I'm not yet sure how else we can direct the linter to ignore a property.

Let's revisit that later

Issue #4717, opened 2024-05-13
proposal: stabilize `std/assert`


iuioiua:
Requirements:

- [x] 100% documented, passing deno doc --lint checks and adhering to the [documentation guidelines](https://github.com/denoland/deno_std/blob/main/.github/CONTRIBUTING.md#documentation).
- [x] Maximum possible test coverage
- [x] No open issues or pull requests that might lead to breaking changes. For example, issues that suggest new non-breaking features are fine to exist at stabilization.
- [x] Publish `1.0.0-rc.1`
- [ ] Wait 1 month for feedbacks
- [ ] Publish `1.0.0`

Blockers:
- [x] #4716
- [ ] #3771
- [x] #3513

Towards #4600

kt3k:
I don't think #3771 is a blocker as it's a bug, not a design issue.

Issue #3771, opened 2023-11-06
assertObjectMatch doesn't print whole object in assertion error


Industrial:
**Describe the bug**
assertObjectMatch doesn't print whole object in assertion error

**Steps to Reproduce**
```ts
import { assertObjectMatch } from "https://deno.land/std@0.207.0/assert/mod.ts";
const x = {
  command: 'error',
  payload: {
    message: 'NodeNotFound',
  },
  protocol: 'graph',
};
const y = {
  protocol: 'graph',
  command: 'addgroup',
  payload: {
    graph: 'foo',
    metadata: {
      description: 'foo',
    },
    name: 'somegroup',
    nodes: [
      'somenode',
      'someothernode',
    ]
  }
};
assertObjectMatch(x, y);
```
```
error: AssertionError: Values are not equal.


    [Diff] Actual / Expected


    {
+     command: "addgroup",
+     payload: {
+       graph: "foo",
+       metadata: {
+         description: "foo",
+       },
+       name: "somegroup",
+       nodes: {
+         "0": "somenode",
+         "1": "someothernode",
+       },
+     },
-     command: "error",
-     payload: {},
      protocol: "graph",
    }

  throw new AssertionError(message);
        ^
    at assertEquals (https://deno.land/std@0.204.0/assert/assert_equals.ts:53:9)
    at assertObjectMatch (https://deno.land/std@0.204.0/assert/assert_object_match.ts:83:10)
    at assertOutputMatchesExpected (file:///home/tom/Code/code9/deno/c9-fbp/src/runtime.test.ts:101:3)
    at eventLoopTick (ext:core/01_core.js:183:11)
    at async Object.<anonymous> (file:///home/tom/Code/code9/deno/c9-fbp/src/runtime.test.ts:522:13)
    at async Function.runTest (https://deno.land/std@0.204.0/testing/_test_suite.ts:358:7)
    at async Function.runTest (https://deno.land/std@0.204.0/testing/_test_suite.ts:346:9)
    at async Function.runTest (https://deno.land/std@0.204.0/testing/_test_suite.ts:346:9)
    at async Function.runTest (https://deno.land/std@0.204.0/testing/_test_suite.ts:346:9)
    at async Function.runTest (https://deno.land/std@0.204.0/testing/_test_suite.ts:346:9)
```

**Expected behavior**
Payload in the error to be filled.

**Environment**

- OS: NixOS
- deno version:
deno 1.37.2 (release, x86_64-unknown-linux-gnu)
[...]

PR #4934, opened 2024-06-03
chore(assert): release `assert@1.0.0-rc.1`

iuioiua:
Towards #4717


deno.json
```diff
@@ -13,7 +13,7 @@
     "graphviz": "npm:node-graphviz@^0.1.1",
 
     "@std/archive": "jsr:@std/archive@^0.224.0",
-    "@std/assert": "jsr:@std/assert@^0.225.3",
+    "@std/assert": "jsr:@std/assert@1.0.0-rc.1",
```
kt3k, line 16: Thanks for updating this as well!

kt3k (approved):
Went through all APIs. All looked reasonable to me. LGTM.

iuioiua:
I am being nitpicky, but I have some thoughts while reviewing this package:
1. Error messages end in periods ("."), when the `msg` argument is undefined. Why? This seems strange. Is this done in the Deno runtime or Web API?
2. `msg` arguments might be better named `msgSuffix`. WDYT?
3. Arguments could be better named. E.g., for `assertArrayIncludes()`, arguments could be named `array` and `subArray`, instead of `actual` and `expected`, respectively. WDYT?
4. I wonder whether extending `AssertionError` in some cases and providing more properties that add context to the error thrown is worthwhile. E.g., for `assertNotEquals()`:
    ```ts
    export class AssertNotEqualsError<T> extends AssertionError {
      actual: T;
      expected: T;

      constructor(actual: T, expected: T, msg?: string) {
        super(
          msg
            ? `Expected actual: ${actual} not to be: ${expected}: ${msg}`
            : `Expected actual: ${actual} not to be: ${expected}`,
        );
        this.name = this.constructor.name;
        this.actual = actual;
        this.expected = expected;
      }
    }
    ```
    This can be done in a backward-compatible manner while making the error thrown more ergonomic. WDYT?

kt3k:
> Error messages end in periods ("."), when the msg argument is undefined. Why? This seems strange. Is this done in the Deno runtime or Web API?

Error messages ending with "." seem just common as well as the error message without "." (Try seach for "new Error" in std). This doesn't seem the issue for `std/assert` to me. (This should be discussed in broader scope)

> msg arguments might be better named msgSuffix. WDYT?

`msg` looks enough descriptive to me. `msgSuffix` looks rather confusing to me.

> Arguments could be better named. E.g., for assertArrayIncludes(), arguments could be named array and subArray, instead of actual and expected, respectively. WDYT?

`actual` and `expected` are conventional in this usage. These look just fine to me.

> I wonder whether extending AssertionError in some cases and providing more properties that add context to the error thrown is worthwhile. E.g., for assertNotEquals():

Sounds a good idea, but doesn't look like a blocker for 1.0.0-rc.1.

iuioiua:
> Error messages ending with "." seem just common as well as the error message without "." (Try seach for "new Error" in std). This doesn't seem the issue for `std/assert` to me. (This should be discussed in broader scope)

There are 100+ `new Error`s in the Standard Library. 14 of them contain a period. So, periods in error messages are part of a minority. Either way, on second thought, let's not worry about this. It's inconsistent but inconsequential.

> `msg` looks enough descriptive to me. `msgSuffix` looks rather confusing to me.

It's appended to the error message. Either way, not blocking.

> `actual` and `expected` are conventional in this usage. These look just fine to me.

Yes, it's conventional, but it doesn't completely make sense. Either way, not blocking.

> Sounds a good idea, but doesn't look like a blocker for 1.0.0-rc.1. Sounds like a good enhancement candidate after the stabilization

I'll open a new issue when I can.

iuioiua:
One last thing - should we consider #4934 a blocker?

kt3k:
> One last thing - should we consider https://github.com/denoland/deno_std/pull/4934 a blocker?

#4934 is this issue

kt3k:
If you mean #3771, then no. It can be landed as a fix after the stabilization.

iuioiua:
My bad - typo. Ok.

PR #4941, opened 2024-06-03
refactor(assert): minor cleanups

iuioiua:


assert/assert_is_error.ts
```diff
@@ -22,14 +22,14 @@ import { stripAnsiCode } from "@std/internal/styles";
  *
  * @typeParam E The type of the error to assert.
  * @param error The error to assert.
- * @param ErrorClass The optional error class to assert.
```
kt3k, line 25: I slightly prefer capital `ErrorClass` as it indicates this is a class constructor instead of a class instance.

assert/assert_match.ts
```diff
@@ -23,9 +23,7 @@ export function assertMatch(
...
   msg?: string,
 ) {
-  if (!expected.test(actual)) {
-    const msgSuffix = msg ? `: ${msg}` : ".";
-    msg = `Expected actual: "${actual}" to match: "${expected}"${msgSuffix}`;
-    throw new AssertionError(msg);
-  }
+  const msgSuffix = msg ? `: ${msg}` : ".";
+  msg = `Expected actual: "${actual}" to match: "${expected}"${msgSuffix}`;
+  assert(expected.test(actual), msg);
```
kt3k, line 28: This causes slight perf degradation as it always constructs error messages

assert/assert_not_match.ts
```diff
@@ -23,10 +23,7 @@ export function assertNotMatch(
...
 ) {
-  if (expected.test(actual)) {
-    const msgSuffix = msg ? `: ${msg}` : ".";
-    msg =
-      `Expected actual: "${actual}" to not match: "${expected}"${msgSuffix}`;
-    throw new AssertionError(msg);
-  }
+  const msgSuffix = msg ? `: ${msg}` : ".";
+  msg = `Expected actual: "${actual}" to not match: "${expected}"${msgSuffix}`;
+  assert(!expected.test(actual), msg);
```
kt3k, line 28: Ditto. This causes perf degradation

PR #4943, opened 2024-06-03
BREAKING(assert): make `unreachable()` consistent with `@std/assert`

iuioiua:
### What's changed

This change changes the format of the error message thrown in `unreachable()` to consistent with the rest of `@std/assert`. It also capitalizes the first letter of the error message.

### Why this change was made

Previously, the format of the error message `unreachable()` threw didn't have the message suffix, given by the `msg` argument, at the end of the message string, which is what the rest of `@std/assert`. This change was made for consistency.

### Migration guide

To migrate, update the logic that handles the error message:

```diff
import { unreachable } from "@std/assert/unreachable";

try {
  unreachable("Hello");
} catch (error) {
- if (error.message === "Hello") {
+ if (error.message === "Unreachable: Hello") {
    foo();
  }
}
```


kt3k (approved):
LGTM

PR #4945, opened 2024-06-03
docs(cli,encoding,fmt,jsonc,path,semver,testing,text,toml,url): remove "This module is browser compatible" note

iuioiua:
This is already stated in JSR.


kt3k:
Note: This only removes `This module is browser compatible.` from the `@module` jsdoc. The directives for type-checking browser compatibility still remain at the top of each file.

kt3k (approved):
LGTM

PR #4946, opened 2024-06-04
refactor(path): cleanup `join()` implementations

iuioiua:
This change also removes the `@std/assert` dependency.


path/posix/join.ts
```diff
@@ -20,17 +19,6 @@ import { normalize } from "./normalize.ts";
  * @returns The joined path.
  */
 export function join(...paths: string[]): string {
-  if (paths.length === 0) return ".";
-
-  let joined: string | undefined;
-  for (let i = 0; i < paths.length; ++i) {
-    const path = paths[i]!;
-    assertPath(path);
```
kt3k, line 28: Why remove assertion?

path/posix/join.ts
```diff
@@ -20,17 +19,6 @@ import { normalize } from "./normalize.ts";
  * @returns The joined path.
  */
 export function join(...paths: string[]): string {
-  if (paths.length === 0) return ".";
```
kt3k, line 23: I prefer to keep this fast path. This should be good for perf

path/posix/join.ts
iuioiua: It asserts that an element of an array of strings is a string.

kt3k: Now the input like `[null]` doesn't throw

kt3k: TypeScript doesn't guarantee the runtime type

iuioiua: Is there a reason why `@std/path` should be inconsistent with the rest of the Standard Library and perform runtime type checking?

kt3k: `std/encoding`, `std/crypto also do runtime type checking. It's not inconsistent

iuioiua: You're referring to instances where variables with multiple possible types are coerced into single types or where the types of variables are checked to change their logic paths. Those cases make sense and are different from what I'm referring to.

I'm referring to simple runtime type-checking of arguments so that their runtime type matches their inferred type.

So that's one out of 40 packages with this behavior, which has never been problematic. Yet, for some reason, we need to keep this behavior here. I don't understand. Why?

kt3k: At least I think this shouldn't be done as part of 'cleanup' as this definitely change the policy about runtime behavior.

iuioiua: I can do this package-wide in a separate PR. Sound good?

iuioiua:
I've re-added `assertPath()` for now.

bartlomieju (approved):

PR #4947, opened 2024-06-04
BREAKING(path): remove separator argument from `common()`

iuioiua:
### What's changed

This change removes the separator argument, `sep`, from `posix.common()`, `windows.common()` and `common()`.

### Why this change was made

This change was made so these functions are made to be consistent with other APIs within `@std/path`, which don't have the ability to set the path separator via an argument. Instead, the user should use the OS-specific variant of `common()` if OS-specific behavior is required and the OS-agnostic `common()` so behavior is oriented for the current machine's OS.

### Migration guide

To migrate, either use the OS-specific or OS-agnostic variant of `common()`, for your use case:

For the current machine's OS (automatically detected):
```ts
import { common } from "@std/path/common";

// No need to define a second argument
common(["C:\\foo\\bar", "C:\\foo\\baz"]);
```

For Windows systems:
```diff
- import { common } from "@std/path/common";
+ import { common } from "@std/path/windows/common";

- common(["C:\\foo\\bar", "C:\\foo\\baz"], "\\");
+ common(["C:\\foo\\bar", "C:\\foo\\baz"]);
```

For POSIX systems:
```diff
- import { common } from "@std/path/common";
+ import { common } from "@std/path/posix/common";

- common(["C:\\foo\\bar", "C:\\foo\\baz"], "/");
+ common(["./deno/std/path/mod.ts", "./deno/std/fs/mod.ts"]);
```

### DIFF
--- a/path/common_test.ts
+++ b/path/common_test.ts
@@ -7,13 +7,12 @@ import * as windows from "./windows/mod.ts";
 Deno.test({
   name: "common() returns shared path",
   fn() {
-    const actual = common(
+    const actual = posix.common(
       [
         "file://deno/cli/js/deno.ts",
         "file://deno/std/path/mod.ts",
         "file://deno/cli/js/main.ts",
       ],
-      "/",
     );
     assertEquals(actual, "file://deno/");
   },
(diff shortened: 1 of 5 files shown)


path/common_test.ts
```diff
@@ -81,27 +78,11 @@ Deno.test({
...
       ],
-      "/",
     );
     assertEquals(actual, "./deno/std/path/mod.ts");
   },
 });
 
-Deno.test({
-  name: "windows.common() returns shared path",
-  fn() {
```
iuioiua, line 92: This seemed incorrect and failed once the separator argument was removed. That's fine, as the correct corresponding test is above (name: "common() checks windows separator").

kt3k (approved):
LGTM

PR #4948, opened 2024-06-04
refactor(fs): cleanup use of `@std/assert`

iuioiua:
Towards #4865


fs/copy.ts
```diff
@@ -34,6 +33,12 @@ interface InternalCopyOptions extends CopyOptions {
   isFolder?: boolean;
 }
 
+function assertIsDate(date: Date | null, name: string): asserts date is Date {
+  if (date === null) {
+    throw new ReferenceError(`${name} is unavailable`);
```
kt3k, line 38: This usage of `ReferenceError` seems strange to me (It's not about variable being unavailable). I think this should be `TypeError` or `Error`.

fs/expand_glob.ts
```diff
@@ -158,7 +157,9 @@ export async function* expandGlob(
     : absRoot;
   while (segments.length > 0 && !isGlob(segments[0]!)) {
     const seg = segments.shift();
-    assert(seg !== undefined);
+    if (seg === undefined) {
+      throw new ReferenceError("Unexpected undefined segment");
```
kt3k, line 161: Ditto. I think this should be `TypeError` or `Error`

fs/expand_glob.ts
```diff
@@ -316,7 +317,9 @@ export function* expandGlobSync(
...
```
kt3k, line 321: Ditto. I think this should be `TypeError` or `Error`

fs/copy.ts
iuioiua: I was going to use `TypeError` originally. But we pass [`Deno.FileInfo.atime`](https://deno.land/api?s=Deno.FileInfo#prop_atime) and [`Deno.FileInfo.mtime`](https://deno.land/api?s=Deno.FileInfo#prop_mtime) to this function, which can only be `null` on some platforms (see documentation). So I went with `ReferenceError`. WDYT?

kt3k: `ReferenceError` is thrown for the case where undeclared 'variable' used like the below:

```ts
const foo = a; // a is not declared
```

There seems no undeclared variable involved in this case. So I think this case doesn't fit to `ReferenceError`

iuioiua: Ah, yeah, fair point. I'll go with `Error`.

fs/expand_glob.ts
iuioiua: I went with `TypeError`.

fs/copy.ts
iuioiua: I went with `Error`.

kt3k (approved):
LGTM. Nice!

PR #4954, opened 2024-06-04
BREAKING(path): remove runtime type-checking

iuioiua:
### What's changed

This change removes `assertPath()` and its use from `@std/path`.

### Why was this change made?

`assertPath()` was an internal function that checked whether a provided string was a string (see implementation). In other words, it could catch an incorrectly typed variable when either not using TypeScript or bypassing the type-checker. In most environments, this function wasn't useful. Hence, the removal.

While the topic of runtime type-checking is still being discussed in #3888, this change was made to be consistent with the rest of the Standard Library in the meantime, which doesn't perform initial type-checking of arguments.

### Migration guide

If using an environment that doesn't use TypeScript or bypasses the type-checker, and if needed, check that variables passed to `@std/path` API are the correct type manually.

### Related

- #3888
- https://github.com/denoland/deno_std/pull/4946#discussion_r1625357674


kt3k:
There seems no consensus reached in #3888. I'd suggest we should discuss this after the stabilization.

iuioiua:
We should be consistent in the meantime. We can come to a decision for the codebase as a whole later, right?

kt3k:
As I commented in https://github.com/denoland/deno_std/pull/4946#discussion_r1625461291, there isn't such consistency in std.

PR #4957, opened 2024-06-04
fix(testing): cast `globalThis` to any

joscha:
This casts `globalThis` to `any` for adding the event listener, following other examples, like this one: 

https://github.com/denoland/deno_std/blob/e48f7cf0611a79d132fa37230cd77bdf2f2d4938/path/_os.ts#L27

## Reason
I encountered:
```
src/deps/jsr.io/@std/testing/0.225.0/snapshot.ts:443:18 - error TS7017: Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.

443       globalThis.addEventListener("unload", this.#teardown);
                     ~~~~~~~~~~~~~~~~

error: Uncaught (in promise) Error: Had 1 diagnostics.
          throw new Error(`Had ${diagnostics.length} diagnostics.`);
                ^
    at getProgramAndMaybeTypeCheck (https://jsr.io/@deno/dnt/0.41.2/mod.ts:468:17)
    at build (https://jsr.io/@deno/dnt/0.41.2/mod.ts:354:17)
    at eventLoopTick (ext:core/01_core.js:168:7)
    at async file:///Users/joscha/dev/affinity-node/scripts/build_npm.ts:10:1
```
whilst using https://github.com/denoland/dnt/

### DIFF
--- a/testing/snapshot.ts
+++ b/testing/snapshot.ts
@@ -438,7 +438,8 @@ class AssertSnapshotContext {
    */
   public registerTeardown() {
     if (!this.#teardownRegistered) {
-      globalThis.addEventListener("unload", this.#teardown);
+      // deno-lint-ignore no-explicit-any
+      (globalThis as any).addEventListener("unload", this.#teardown);
       this.#teardownRegistered = true;
     }
   }


kt3k:
hmm, Deno has `addEventListener` defined in `globalThis`. So this shouldn't throw type error with normal type configuration in Deno.

I'm also not familiar with the vendored path of that pattern `src/deps/jsr.io/@std/testing/0.225.0/snapshot.ts` . Is that path generated by `dnt`?

joscha:
> hmm, Deno has `addEventListener` defined in `globalThis`. So this shouldn't throw type error with normal type configuration in Deno.

I did wonder how that passed here in the first place. 

> I'm also not familiar with the vendored path of that pattern `src/deps/jsr.io/@std/testing/0.225.0/snapshot.ts` . Is that path generated by `dnt`?

I believe so. It's the compat layer that is generated:

Notification_Center

(`npm` is the folder that `dnt` generates into)

Where are the type definitions for `globalThis` in this repo? Is it possible that we just need to teach `dnt` to pick them up when vendoring the dependency? @dsherret ?

kt3k:
@joscha BTW I guess you're trying to support both Deno and Node in your project. Have you considered using JSR for supporting both runtimes? I think JSR is now more officially recommended way to do it

joscha:
> @joscha BTW I guess you're trying to support both Deno and Node in your project. Have you considered using JSR for supporting both runtimes? I think JSR is now more officially recommended way to do it

yes, agreed. I have a runtime environment which doesn't have JSR support yet, unfortunately.

dsherret:
@joscha I think it's because these are considered DOM types. You can ignore the diagnostic using the `filterDiagnostic` hook on the build options or specify:

```ts
  compilerOptions:{
    "lib": ["DOM", "ESNext"],
  },
```

joscha:
>     "lib": ["DOM", "ESNext"],

this doesn't seem to make a difference.

```typescript
    filterDiagnostic(diagnostic) {
        if (
          diagnostic.file?.fileName.endsWith('src/deps/jsr.io/@std/testing/0.225.0/snapshot.ts')
        ) {
            // see https://github.com/denoland/deno_std/pull/4957
          return false; // ignore all diagnostics in this file
        }
    },
```
works.

I couldn't get anything useful regarding the local versioned file path from `import.meta.resolve("@std/testing/snapshot")` unfortunately, as it yields a JSR reference with a semver range (`jsr:/@std/testing@^0.225.0/snapshot`).

joscha:
Update: `filterDiagnostic` is not needed anymore since version 1.x

PR #4964, opened 2024-06-05
chore(toml): minor cleanup

iuioiua:


toml/mod.ts
```diff
@@ -43,21 +44,6 @@
  * Because local time does not exist in JavaScript, the local time is stored as a
  * string.
  *
- * #### Inline Table
```
iuioiua, line 46: This was removed because it provided no added value to the checkmark in the list above.

PR #4968, opened 2024-06-05
refactor(jsonc): cleanup use of `@std/assert`

iuioiua:
Towards #4865


jsonc/parse.ts
```diff
@@ -78,8 +75,6 @@ type Token = {
   position: number;
 };
 
-const originalJSONParse = globalThis.JSON.parse;
```
iuioiua, line 81: Pass-by cleanup

kt3k (approved):
LGTM 👍

PR #4970, opened 2024-06-05
chore: improve circular dependencies checker

iuioiua:
This change improves the circular dependency checker to automatically show green on packages at 1.x (including release candidates) and cleans up some old logic and visual noise.


kt3k (approved):
LGTM

PR #4974, opened 2024-06-05
refactor(http): cleanup use of `@std/assert`

iuioiua:
Towards #4865


kt3k (approved):
These thrown errors make more sense than `AssertionError`. Looks Nice improvements to me 👍

PR #4979, opened 2024-06-05
refactor(io): cleanup use of `@std/assert`

iuioiua:
Toward #4865


io/buf_reader.ts
```diff
@@ -95,7 +94,6 @@ export class BufReader implements Reader {
         this.#eof = true;
         return;
       }
-      assert(rr >= 0, "negative read");
```
iuioiua, line 98: I removed these negative read checks as it seems impossible.

io/buf_reader.ts
```diff
@@ -262,10 +257,11 @@ export class BufReader implements Reader {
       let partial;
       if (err instanceof PartialReadError) {
         partial = err.partial;
-        assert(
-          partial instanceof Uint8Array,
-          "bufio: caught error from `readSlice()` without `partial` property",
-        );
+        if (partial === undefined) {
```
bartlomieju, line 260: Maybe 
```suggestion
        if (!(partial instanceof Uint8Array)) {
```

bartlomieju (approved):

Issue #4967, opened 2024-06-05
suggestion: have `url.dirname()` return `string` instead of `URL`


iuioiua:
It currently returns a `URL` instead of a `string`. I find this behavior surprising, as it requires the user to call `.href` in most cases. Do you have any thoughts, @marvinhagemeister?

marvinhagemeister:
Doesn't it already return strings https://jsr.io/@std/url/doc/~/basename ?

iuioiua:
(sighs) I meant [`dirname()`](https://jsr.io/@std/url/doc/dirname/~/dirname) 😅

kt3k:
APIs in `url` consistently return URL instance when the returned value can be expressed as URL (`dirname`, `join`, `normalize`). `extname` and `basename` return string as the returned values can't be expressed as URLs

iuioiua:
It makes sense for `join()` (join URLs) and `normalize()` (normalize a URL) to return `URL`s. I'd expect `dirname()` to behave the same way `basename()` and `extname()`, as their names imply.

marvinhagemeister:
Both points of views make sense to me and are valid options. So it comes more down to what we expect developers to use it for. I'm trying to make a case from that perspective. In apps I've written the `dirname()` function on paths is typically used to traverse upwards, like finding the nearest `deno.json` for example. A similar use case seems to apply to URLs as well. If the type changes to a `string` you'd constantly have to cast it back to a `URL` again. For that reason it seems to make more  sense to me that `dirname()` returns a `URL` instance. You can always get back the stringified variant by doing `url.pathname` I guess.

iuioiua:
Alright. I'm not sure I fully agree, but let's leave it as-is.

kt3k:
There was also a small discussion about `dirname` when it was first landed https://github.com/denoland/deno_std/pull/3527#issuecomment-1702850650

Issue #4924, opened 2024-05-31
stabilize `@std/url`


kt3k:
part of #4600 

- [x] fully documented
- [x] fully tested
- [x] no ongoing breaking change discussion
- [ ] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

91pornaaaaa:
https://91porna.com/

iuioiua:
Blocker: #4967

kt3k:
Probably we are not stabilizing this package. ref #4992 

Instead we explore alternative solution to solve the issues which `std/url` was supposed to solve

PR #4981, opened 2024-06-05
chore(url): release `url@1.0.0-rc.1`

iuioiua:
Frankly, I still think `url.dirname()` should probably return a `string` rather than `URL` (#4967). Nevertheless, I won't keep that as a blocker. This package looks ready to me.

Towards #4924


kt3k:
> I still think url.dirname() should probably return a string rather than URL (https://github.com/denoland/deno_std/issues/4967)

`url.parent()` might be a better name..?

kt3k (approved):
LGTM

Let's discuss `url.dirname` issue during RC

PR #4982, opened 2024-06-05
docs(net): complete documentation

iuioiua:
Towards #3764

### DIFF
--- a/net/mod.ts
+++ b/net/mod.ts
@@ -3,6 +3,13 @@
 /**
  * Network utilities.
  *
+ * ```ts
+ * import { getNetworkAddress } from "@std/net";
+ * import { assertNotEquals } from "@std/assert/assert-not-equals";
+ *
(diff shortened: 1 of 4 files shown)


net/mod.ts
```diff
@@ -3,6 +3,15 @@
 /**
  * Network utilities.
  *
+ * ```ts no-assert no-eval
+ * import { getNetworkAddress, getAvailablePort } from "@std/net";
+ *
+ * const port = getAvailablePort();
+ * const hostname = getNetworkAddress();
+ *
+ * Deno.serve({ port, hostname }, () => new Response("Hello, world!"));
```
kt3k, line 12: This usage of `hostname` feels a bit misleading to me. Do we recommend using `getNetworkAddress()` in this way?

iuioiua: Good point. Fixed.

kt3k (approved):
LGTM

PR #4983, opened 2024-06-06
refactor(data-structures): remove use of `public` keyword

### LINKED ISSUE #4826: to-do: remove use of `protected`, `private` and `public` from `@std/data-structures`

iuioiua:
Closes #4826


kt3k (approved):
LGTM

PR #4992, opened 2024-06-06
chore(url): release `url@1.0.0`

### LINKED ISSUE #4924: stabilize `@std/url`

iuioiua:
To be merged July 7.

Closes #4924


kt3k (approved):
LGTM

thisisjofrank (approved):

lucacasonato (changes requested):
I don't think the `basename` behaviour makes sense - we should not stabilize it.

Namely the fact that it is really
```js
import { basename } from "@std/path/posix";

if (url.path !== "/") {
  return basename(url.path);
} else {
  return url.hostname;
}
```

Additionally, the connotation that `url.*` methods operates on a URL is not really correct. It operates on URLs that have posix formatted paths. This is weird, because not all URLs have posix formatted paths. Take for example data URLs:

```
> new URL("data:application/javascript,1+1;")
URL {
  href: "data:application/javascript,1+1;",
  origin: "null",
  protocol: "data:",
  username: "",
  password: "",
  host: "",
  hostname: "",
  port: "",
  pathname: "application/javascript,1+1;",
  hash: "",
  search: ""
}
```

This is not a posix formatted path.

Because of this, I think that really we shouldn't be doing this - we should just document that people can use `join`, `basename`, and friends from `@std/path/posix`.

iuioiua:
Frankly, the purpose of `@std/url` has always been a little unclear to me because the APIs are such thin wrappers around `@std/path` APIs, as you point out. I'll create a PR which aims to improve `@std/path` documentation by including equivalent examples for `@std/url`. Then, we can see make a call on what to do with `@std/url`.

iuioiua:
Perhaps, one of the only things `@std/url` offers over just using `@std/path/posix` APIs is that it strips [`URL.hash`](https://developer.mozilla.org/en-US/docs/Web/API/URL/hash) and [`URL.search`](https://developer.mozilla.org/en-US/docs/Web/API/URL/search) properties before processing the values. I'm not sure whether that provides sufficient justification for keeping the package...

iuioiua:
Closing as we're instead removing `@std/url`. See #5530 for details.

PR #4995, opened 2024-06-09
feat(streams/unstable): `FixedChunkStream`

BlackAsLight:
This pull request adds a new TransformStream to the `@std/streams` module, that resizes the incoming `Uint8Array` chunks to the specified size, with the exception of the last chunk.
## Example
```ts
import { ResizeStream } from "@std/streams/resize-stream";

const readable = ReadableStream.from(function* () {
  for (let i = 0; i < 100; ++i) {
    yield new Uint8Array(Math.floor(Math.random() * 1000));
  }
}())
  .pipeThrough(new ResizeStream(512))
  .pipeThrough(new TransformStream({
      controller.enqueue(chunk.length.toString() + '\n')
    }
  }))
  .pipeTo(Deno.stdout.writable, { preventClose: true })
```
## Possible Output
```
512
512
512
512
512
512
512
320
```


kt3k:
Could you elaborate on the use cases?

BlackAsLight:
The use cases is really just if one wants to read an incoming stream in a fixed size chunks then this would make it easy for them to do that, instead of writing their own method to slice up and merge the chunks.
```ts
for (const chunk of readable.pipeThrough(new ResizeStream(512))
  // consume chunk
```
One place I know this would be useful is in the [Tar implementation](https://github.com/denoland/deno_std/pull/4548/files#diff-1ff925dd4b61263a39e33308919663ba92d70ba444b7d98fcbc5bc2c08a19ed5R106-R131) which reads tar files in 512 size chunks.

iuioiua:
Ah, yes. When I was looking through that `TarStream` implementation, I also thought this might be worth doing. However, I think it could be better to use a better name. I am open to suggestions.

BlackAsLight:
> Ah, yes. When I was looking through that `TarStream` implementation, I also thought this might be worth doing. However, I think it could be better to use a better name. I am open to suggestions.

- `ResizeChunkStream`
- `FixedChunkStream`

iuioiua:
I like `FixedChunkStream`. Any objections?

streams/fixed_chunk_stream_test.ts
```diff
@@ -0,0 +1,20 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "@std/assert";
+import { FixedChunkStream } from "./fixed_chunk_stream.ts";
+
+Deno.test("ResizeStream", async () => {
```
kt3k, line 6:
```suggestion
Deno.test("FixedChunkStream", async () => {
```

streams/fixed_chunk_stream.ts
```diff
@@ -0,0 +1,100 @@
...
+ *   }))
+ * ```
+ *
+ * > [!WARNING]
+ * > **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @experimental
+ */
+export class FixedChunkStream<T extends Fixable<T>>
+  extends TransformStream<T, T> {
```
iuioiua, line 68: I understand that you'd like to have the input be very flexible. However, it would make this the only API in `@std/streams` of this kind. Should demand for this functionality arise in the future, it should be something that applied to all of `@std/streams`, where applicable, and can be done in a non-breaking manner. Let's also remove `Fixable`.
```suggestion
export class FixedChunkStream extends TransformStream<Uint8Array, Uint8Array> {
```

iuioiua (changes requested):
Let's get this landed for #4548.

crowlKats (approved):
LGTM

iuioiua (approved):
I've made the adjustments to make this accept `Uint8Array`. Now, LGTM! Thanks for this!

PR #5022, opened 2024-06-11
refactor(ulid): improve error types and messages

iuioiua:


ulid/mod.ts
```diff
@@ -84,12 +85,14 @@ export function decodeTime(ulid: string): number {
     .reduce((carry, char, index) => {
       const encodingIndex = ENCODING.indexOf(char);
       if (encodingIndex === -1) {
-        throw new Error("invalid character found: " + char);
+        throw new SyntaxError(`Invalid ULID character found: ${char}`);
```
kt3k, line 88: I've never seen a library throwing a SyntaxError for non-javascript syntax errors. I think this should be TypeError or Error.

iuioiua: Reasoning based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/JSON_bad_parse

kt3k: Why the above one being an `Error` and this one being `SyntaxError`?

iuioiua: Hm... Yeah, they should both be the same error type... Changed to `Error`.

kt3k (approved):
LGTM

PR #5033, opened 2024-06-12
docs(testing): improve the docs of `@std/testing`

kt3k:
part of #3764


_tools/check_docs.ts
```diff
@@ -59,6 +59,11 @@ const ENTRY_POINTS = [
   "../semver/mod.ts",
   "../streams/mod.ts",
   "../text/mod.ts",
+  "../testing/bdd.ts",
+  "../testing/mock.ts",
+  "../testing/snapshot.ts",
+  "../testing/time.ts",
+  "../testing/types.ts",
```
kt3k, line 66: `../testing/asserts.ts` is excluded because they're going to be removed at 1.0.0-rc.1

_tools/check_docs.ts
```diff
@@ -67,7 +72,8 @@ const ENTRY_POINTS = [
 ] as const;
 
 const TS_SNIPPET = /```ts[\s\S]*?```/g;
-const ASSERTION_IMPORT = /import \{.*\} from "@std\/assert(?:\/.*)?";/gm;
+const ASSERTION_IMPORT =
+  /from "@std\/(assert(\/[a-z-]+)?|testing\/(mock|snapshot|types))"/g;
```
kt3k, line 76: Removed `import \{.*\} ` part as this doesn't work if there are line breaks between `import` and `from`.

Also added `testing\/(mock|snapshot|types)` part as these files also include assertion functions.

iuioiua: Having assertions spread across multiple packages is confusing. What if we moved assertions (only) to `@std/assert`? WDYT?

testing/snapshot.ts
```diff
@@ -152,8 +152,10 @@ import { buildMessage } from "@std/internal/build-message";
 const SNAPSHOT_DIR = "__snapshots__";
 const SNAPSHOT_EXT = "snap";
 
+/** The mode of snapshot testing. "assert" or "update". */
```
iuioiua, line 155:
```suggestion
/** The mode of snapshot testing. */
```

iuioiua (changes requested):
First pass review

_tools/check_docs.ts
kt3k: That makes `@std/assert` a huge package. In the current layout, assertions are organized by the category (mock / snapshot / types). This makes more sense to me.

testing/types.ts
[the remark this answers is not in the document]
kt3k: Added examples

iuioiua (approved):
Great work! Good to see this done.

PR #5041, opened 2024-06-14
BREAKING(async): stop exporting ERROR_WHILE_MAPPING_MESSAGE

kt3k:
part of #5001

We usually don't export error message string as string constant value. Also exporting this doesn't look useful. I suggest we should remove this export.


iuioiua (approved):
Nice!

PR #5044, opened 2024-06-14
feat(http): support partitioned cookies

### LINKED ISSUE #5043: Support Cookies Having Independent Partitioned State (CHIPS)/Partioned cookies in http/cookie.ts

We'd like to emit our cookies with the `Partitioned` attribute, to allow using third-party cookies in iframes on supported browsers.

Extend the Cookie interface with a `partitioned` property, similar to the `secure` property and emit this in cookie.ts's toString function.

Using `unparsed` works, but isn't as type-safe.

KlaasSkelteAtEnatom:
closes #5043

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -49,7 +49,17 @@ export interface Cookie {
   secure?: boolean;
   /** The cookie's `HTTPOnly` attribute. If `true`, the cookie cannot be accessed via JavaScript. */
   httpOnly?: boolean;
-  /** The cookie's `Partitioned` attribute. If `true`, the cookie will be only be included in the `Cookie` request header if the domain it is embedded by matches the domain the cookie was originally set from.  */
+  /**
+   * The cookie's `Partitioned` attribute.
+   * If `true`, the cookie will be only be included in the `Cookie` request header if
+   * the domain it is embedded by matches the domain the cookie was originally set from.
+   *
+   * Warning: This is an attribute that has not been fully standardized yet.
+   * It may change in the future without following the semver semantics of the package.
+   * Clients may ignore the attribute until they understand it.
+   *
+   * @default {false}
+   */
   partitioned?: boolean;
   /**
    * Allows servers to assert that a cookie ought not to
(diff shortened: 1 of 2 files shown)


kt3k:
Note:

Browser support:
- Only Chrome (and Edge) implements it.
- Firefox doesn't support it, but its position is positive https://github.com/mozilla/standards-positions/issues/678
- Safari doesn't support it, but the position seems positive now? https://github.com/WebKit/standards-positions/issues/50

Library support:
- Go supports it https://github.com/golang/go/commit/2bf686dfe902c900bb0c9a140b49e5697d451944
- Fastify supports it with warning about possible future change https://github.com/fastify/fastify-cookie/commit/9e4652ff303c9b1c4581c36a86d37b026a767bc4
- python is planning the addition in 3.13 https://github.com/python/cpython/issues/112713

I think it's probably fine to add this now.

Does anyone have any opinion on this?

http/cookie.ts
```diff
@@ -49,6 +49,8 @@ export interface Cookie {
   secure?: boolean;
   /** The cookie's `HTTPOnly` attribute. If `true`, the cookie cannot be accessed via JavaScript. */
   httpOnly?: boolean;
+  /** The cookie's `Partitioned` attribute. If `true`, the cookie will be only be included in the `Cookie` request header if the domain it is embedded by matches the domain the cookie was originally set from.  */
```
kt3k, line 52: Can you add some warning message that this isn't standardized yet, and this option may change in the future (without following the semver semantics of the package)

ref: https://github.com/fastify/fastify-cookie/commit/9e4652ff303c9b1c4581c36a86d37b026a767bc4#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R131

iuioiua: Can you please add `@default {false}` to this?

iuioiua (approved):
I'm fine with this change. LGTM.

http/cookie.ts
KlaasSkelteAtEnatom: Addressed both in 802c56a, hopefully phrased properly.

iuioiua (approved):
Thanks!

PR #5050, opened 2024-06-16
refactor(testing): remove `FakeTime.start` setter

iuioiua:
See https://github.com/denoland/deno_std/pull/5033#discussion_r1641593658

### DIFF
--- a/testing/time.ts
+++ b/testing/time.ts
@@ -238,6 +238,9 @@ let dueTree: RedBlackTree<DueNode>;
  * Overrides the real Date object and timer functions with fake ones that can be
  * controlled through the fake time instance.
  *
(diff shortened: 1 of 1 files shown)


kt3k:
By this change, the error message changes from:

```
error: Uncaught (in promise) Error: cannot change start time after initialization
```

To:

```
error: Uncaught (in promise) TypeError: Cannot set property start of #<FakeTime> which has only 
```

I slightly prefer the 1st one because it explicitly says it's intentional by design. The latter sounds like the case is not considered by the author. WDYT?

iuioiua:
I see having a method that is not useful as having poor design. I've added a note to counteract any possible confusion.

kt3k (approved):
LGTM

PR #5051, opened 2024-06-16
refactor(testing): remove use of `public` keyword

iuioiua:
As we don't use the access modifiers in the Standard Library.


kt3k (approved):
LGTM

PR #5053, opened 2024-06-16
refactor(fs): use `isSamePath()` in `isSubdir()`

iuioiua:
This is a more robust means of checking whether the source directory is the same as the destination directory.

### DIFF
--- a/fs/_is_subdir.ts
+++ b/fs/_is_subdir.ts
@@ -3,6 +3,7 @@
 
 import { SEPARATOR } from "@std/path/constants";
 import { toPathString } from "./_to_path_string.ts";
+import { isSamePath } from "./_is_same_path.ts";
 
 /**
  * Checks whether `src` is a sub-directory of `dest`.
@@ -19,7 +20,7 @@ export function isSubdir(
   dest: string | URL,
   sep = SEPARATOR,
 ): boolean {
-  if (src === dest) {
+  if (isSamePath(src, dest)) {
     return false;
   }
   src = toPathString(src);


kt3k:
> This is a more robust means of checking whether the source directory is the same as the destination directory.

Can you add test cases that this change fixes?

fs/_is_subdir.ts
```diff
@@ -19,7 +20,7 @@ export function isSubdir(
   dest: string | URL,
   sep = SEPARATOR,
 ): boolean {
-  if (src === dest) {
+  if (isSamePath(src, dest)) {
```
kt3k, line 23: `isSamePath` internally calls `toPathString` for `src` and `dest`, which is also called for `src` and `dest` in `isSubdir`. I think we should avoid the these duplicated calls

babiabeo: Perhaps we can use `resolve()` directly instead of using `isSamePath()`:

```ts
export function isSubdir(
  src: string | URL,
  dest: string | URL,
  sep = SEPARATOR,
): boolean {
  src = toPathString(src);
  dest = toPathString(dest);

  if (resolve(src) === resolve(dest)) {
    return false;
  }

  const srcArray = src.split(sep);
  const destArray = dest.split(sep);
  return srcArray.every((current, i) => destArray[i] === current);
}
```

kt3k: @babiabeo Looks good to me. Could you make a PR with a test case?

iuioiua: Cool. I'll close this PR then 👍🏾

PR #5054, opened 2024-06-16
refactor(ulid): remove `len` argument from `encodeTime()` and `encodeRandom()`

iuioiua:
There's no reason why this value shouldn't be hardcoded. Also, includes a small cleanup of the functions.


kt3k (approved):
Good catch! LGTM

PR #5058, opened 2024-06-17
BREAKING(async): simplify `deadline()` logic, remove `DeadlineError` and improve errors

iuioiua:
### What's changed

This PR:
1. Ensures that all thrown/rejected errors within [`@std/async`](https://jsr.io/@std/async) APIs are [`DOMException`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException)s, with the appropriate [`AbortError`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#aborterror) or [`TimeoutError`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#timeouterror) error names.
2. Simplifies [`deadline()`](https://jsr.io/@std/async/doc/~/deadline) by relying on [`abortable()`](https://jsr.io/@std/async/doc/~/abortable).
3. Removes `DeadlineError` for `deadline()`, as it's no longer needed.
4. Clarifies the error message when one of these functions is called and the `AbortSignal` is already aborted.

### Why was this change been made?

These changes were made to:
1. Ensure this package's APIs are consistent in how they throw errors.
2. Reduce the number of moving parts.
3. Reduce the need to import an additional custom error, `DeadlineError`.
4. Improve the clarity of errors and therefore help with troubleshooting.

### Migration guide

If using `deadline()` compare the error with `DOMException` instead of `DeadlineError` when handling errors.

```diff
- import { deadline, type DeadlineError } from "@std/async";
+ import { deadline } from "@std/async";

  try {
    await deadline(promise, 1_000);
  } catch (error) {
-   if (error instanceof DeadlineError && error.message === "Deadline") {
+   if (error instanceof DOMException && error.name === "TimeoutError") {
      // Handle error
    }
  }
```

### Related

Depends on https://github.com/denoland/deno/pull/23842


kt3k:
> The main consequences of these changes are that the error type has changed from DeadlineError to DOMException and the error message from Deadline to The signal has been aborted. Deadline error has been removed as it's no longer used.

I don't understand why we want to make this change. Currently the user can distinguish whether the promise is aborted by signal or aborted by deadline by catching the error. Now they can't do it.

iuioiua:
You can compare using `error.name === "AbortError"` or `error.name === "TimeoutError"`, similar to `AbortSignal.timeout()`. I've clarified my initial comment.

kt3k:
> You can compare using error.name === "AbortError" or error.name === "TimeoutError", similar to AbortSignal.timeout().

Oh, this sounds nice.

> I've clarified my initial comment. Perhaps, we can update the documentation as well.

Sounds a good idea. Can you update the doc examples and test case to demonstrate this?

async/deadline_test.ts
```diff
@@ -14,15 +14,19 @@ Deno.test("deadline() returns fulfilled promise", async () => {
...
-  await assertRejects(async () => {
-    await deadline(p, 100);
-  }, DeadlineError);
+  await assertRejects(
+    async () => {
+      await deadline(p, 100);
+    },
+    DOMException,
+    "Signal timed out.",
+  );
```
kt3k, line 29: Maybe we should also assert `err.name === "TimeoutError"`?

async/deadline_test.ts
```diff
@@ -77,8 +87,13 @@ Deno.test("deadline() handles already aborted signal", async () => {
...
   abort.abort();
-  await assertRejects(async () => {
-    await deadline(p, 100, { signal: abort.signal });
-  }, DeadlineError);
+  const error = await assertRejects(
+    async () => {
+      await deadline(p, 100, { signal: abort.signal });
+    },
+    DOMException,
+    "The signal has been aborted",
```
iuioiua, line 95: Hm... The discrepancy between the error message in this test and the one above seems odd.

iuioiua:
I've decided to revert the use of `signal.throwIfAborted()`, as it throws the reason, which is a string, instead of an error with the message as the reason.

iuioiua:
I've tweaked the direction of this PR and re-written the descript. @kt3k, please take a look once more.

0f-0b:
> I've decided to revert the use of `signal.throwIfAborted()`, as it throws the reason, which is a string, instead of an error with the message as the reason.

I think it's better to always just throw the reason itself, to be consistent with Web APIs like `fetch` and [`navigator.locks.request`](https://w3c.github.io/web-locks/#dom-lockmanager-request).

iuioiua:
> I think it's better to always just throw the reason itself, to be consistent with Web APIs like `fetch` and [`navigator.locks.request`](https://w3c.github.io/web-locks/#dom-lockmanager-request).

This code does, wrapped in a `DOMException`. Do you see any points where it doesn't?

async/mod.ts
```diff
@@ -1,5 +1,9 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
-// This module is browser compatible.
+// TODO(iuioiua): Add web-compatible declaration once TypeScript 5.5 is released
+// and in the Deno runtime. See https://github.com/microsoft/TypeScript/pull/58211
+//
+// Note: this code is still compatible with recent
+// web browsers. See https://caniuse.com/?search=AbortSignal.any
```
kt3k, line 6: This seems no longer necessary?

async/delay.ts
```diff
@@ -15,6 +15,7 @@ export interface DelayOptions {
 /**
  * Resolve a {@linkcode Promise} after a given amount of milliseconds.
  *
+ * @throws {DOMException} If the optional signal is already aborted.
```
kt3k, line 18:
```suggestion
 * @throws {DOMException} If the optional signal is aborted before the delay duration.
```

async/mod.ts
iuioiua: It is necessary because `deadline()` uses `AbortSignal.any()`.

0f-0b:
> This code does, wrapped in a `DOMException`. Do you see any spots where it doesn't?

Here the reason (which is usually [already an `"AbortError"` `DOMException`](https://developer.mozilla.org/docs/Web/API/AbortController/abort#reason)) is implicitly converted to a string and used as the message of another exception. That is no longer the reason itself.

kt3k:
Maybe `ac.abort()` method suppose the reason should be an error object instead of string?

iuioiua:
Oh, I see. Thank you. I've made some adjustments, which made things even cleaner. PTAL.

async/abortable.ts
```diff
@@ -77,11 +125,9 @@ function abortablePromise<T>(
   p: Promise<T>,
   signal: AbortSignal,
 ): Promise<T> {
-  if (signal.aborted) {
-    return Promise.reject(createAbortError(signal.reason));
-  }
+  if (signal.aborted) return Promise.reject(signal.reason);
```
iuioiua, line 128: On 2nd thought. Maybe this should throw. Thoughts?

async/abortable.ts
```diff
@@ -1,61 +1,109 @@
...
+ * ```ts
+ * import { abortable, delay } from "@std/async";
+ * import { assertRejects, assertEquals } from "@std/assert";
  *
- * const p = delay(1000);
- * const c = new AbortController();
- * setTimeout(() => c.abort(), 100);
+ * const promise = delay(1_000);
+ * const controller = new AbortController();
+ * controller.abort(new Error("This is my reason"));
```
kt3k, line 36: 👍

async/mod.ts
kt3k: Ah, ok

async/abortable.ts
kt3k: This looks fine to me

kt3k (approved):
LGTM

I think the removal of wrapping of abort reason is a good move towards Web standard alignment. Thanks @0f-0b for the ideas!

PR #5061, opened 2024-06-17
test(streams): improve `Buffer` test

kt3k:
This PR adds test cases for `Buffer` class of `std/streams`, and also removes one dead if-block.

This PR improves the line coverage of `Buffer` class from 72.54% to 100%.

### DIFF
--- a/streams/buffer.ts
+++ b/streams/buffer.ts
@@ -375,9 +375,6 @@ export class Buffer {
   }
 
   #reslice(len: number) {
-    if (len > this.#buf.buffer.byteLength) {
-      throw new RangeError("Length is greater than buffer capacity");
-    }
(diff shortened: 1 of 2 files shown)


crowlKats (approved):
LGTM

streams/buffer.ts
```diff
@@ -375,9 +375,6 @@ export class Buffer {
   }
 
   #reslice(len: number) {
-    if (len > this.#buf.buffer.byteLength) {
```
iuioiua, line 378: Is this condition unsatisfiable? If so, how?

kt3k: It's guarded by the condition before calling #reslice(). It seems not possible to make `len` exceed `this.#buf.buffer.byteLength`

iuioiua (approved):

PR #5064, opened 2024-06-17
chore: remove `--table` functionality from graph checker

iuioiua:
This functionality is no longer used and won't be needed after https://github.com/jsr-io/jsr/pull/528 is merged.


kt3k (approved):
LGTM

PR #5073, opened 2024-06-18
feat(internal): `assertStrictThrows()` and `assertStrictRejects()`

### LINKED ISSUE #5072: suggestion: `assertStrictIsError()` and friends
I suggest adding `assertStrictIsError()`, `assertStrictThrows()` and `assertStrictReject()`. These functions would behave differently than their non-strict counterparts by making their arguments required and changing `msgIncludes` to `msgEquals` (with `===` equality). These functions would holistically define the exact properties of a thrown/rejected error. Having these functions defined in `@std/internal` would be best, then moving to `@std/assert` once 1.0.0 is released.

iuioiua:
Once merged, we can track the updating of tests to use these new functions.

Closes #5072

### DIFF
--- a/internal/assert_strict.ts
+++ b/internal/assert_strict.ts
@@ -36,8 +36,8 @@ export class StrictAssertionError extends Error {
  * Asserts that the error is an instance of the given error class and that the
  * error message is a given string.
  *
- * @throws {import("@std/assert/assertion-error").AssertionError} If the error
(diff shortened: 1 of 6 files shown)


iuioiua:
Closing. See https://github.com/denoland/deno_std/issues/5072#issuecomment-2177483593.

PR #5076, opened 2024-06-18
refactor(fs): resolve directories when checking whether they are the same in `isSubdir()`

babiabeo:
Jumping from: #5053


iuioiua (approved):
Thank you, David 👍🏾

PR #5077, opened 2024-06-18
refactor(archive,expect,io,log,toml,yaml): remove `private` and `public` access modifiers

### LINKED ISSUE #5066: to-do: remove access modifiers across the codebase
The `public` keyword should be removed, and the `private` keyword should be replaced with the `#` prefix. These changes should be made within both documentation and implementation code, for consistency.

MadGanGithub:
Closes #5066

Fixes:
Removed public keyword
Added separate declaration and constructor assignment

Replaced private key with # keyword

### DIFF
--- a/io/_test_common.ts
+++ b/io/_test_common.ts
@@ -3,7 +3,7 @@
 import type { Reader } from "./types.ts";
 
 export const MIN_READ_BUFFER_SIZE = 16;
-export const bufsizes: number[] = [
+export const bufsizes = [
   0,
   MIN_READ_BUFFER_SIZE,
   23,
@@ -20,7 +20,7 @@ export class BinaryReader implements Reader {
   index = 0;
   #bytes: Uint8Array;
 
-  constructor(bytes: Uint8Array = new Uint8Array(0)) {
+  constructor(bytes = new Uint8Array(0)) {
     this.#bytes = bytes;
   }
 
--- a/io/_test_common.ts
+++ b/io/_test_common.ts
@@ -18,10 +18,10 @@ export const bufsizes: number[] = [
 
 export class BinaryReader implements Reader {
   index = 0;
-  #bytes:Uint8Array;
+  #bytes: Uint8Array;
 
   constructor(bytes: Uint8Array = new Uint8Array(0)) {
-    this.#bytes=bytes;
+    this.#bytes = bytes;
   }
 
   read(p: Uint8Array): Promise<number | null> {
--- a/io/_test_common.ts
+++ b/io/_test_common.ts
@@ -18,11 +18,14 @@ export const bufsizes: number[] = [
 
 export class BinaryReader implements Reader {
   index = 0;
+  #bytes:Uint8Array;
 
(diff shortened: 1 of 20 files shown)


io/_test_common.ts
```diff
@@ -18,11 +18,14 @@ export const bufsizes: number[] = [
 
 export class BinaryReader implements Reader {
   index = 0;
+  #bytes:Uint8Array;
 
-  constructor(private bytes: Uint8Array = new Uint8Array(0)) {}
+  constructor(bytes: Uint8Array = new Uint8Array(0)) {
```
iuioiua, line 23: Nit: the type can be inferred from the variable's default value. Type annotation is not needed.
```suggestion
  constructor(bytes = new Uint8Array(0)) {
```

io/buf_reader.ts
```diff
@@ -50,8 +52,8 @@ export class BufReader implements Reader {
   #r = 0; // buf read position.
   #w = 0; // buf write position.
   #eof = false;
-  // private lastByte: number;
-  // private lastCharSize: number;
+  // #lastByte: number;
+  // #lastCharSize: number;
```
iuioiua, line 56: Nit: these aren't used. Safe to remove.
```suggestion
```

io/string_writer.ts
```diff
@@ -41,11 +41,13 @@ export class StringWriter implements Writer, WriterSync {
   #chunks: Uint8Array[] = [];
   #byteLength = 0;
   #cache: string | undefined;
+  #base:string;
 
-  constructor(private base: string = "") {
+  constructor(base: string = "") {
```
iuioiua, line 46: Ditto. Type annotation not needed.
```suggestion
  constructor(base = "") {
```

yaml/_error.ts
```diff
@@ -14,7 +14,7 @@ export class YAMLError extends Error {
     this.name = this.constructor.name;
   }
 
-  public override toString(_compact: boolean): string {
+  override toString(_compact: boolean): string {
```
iuioiua, line 17: Nit: let's remove this variable since it's not used.
```suggestion
  override toString(): string {
```

yaml/_loader/loader_state.ts
```diff
@@ -26,31 +26,31 @@ export interface LoaderStateOptions {
...
+  tagMap?: ArrayObject;
+  anchorMap?: ArrayObject;
+  tag?: string | null;
+  anchor?: string | null;
+  kind?: string | null;
+  result: ResultType | null = "";
 
   constructor(
-    public input: string,
+    input: string,
```
iuioiua, line 53: I believe we need to define `this.input` to fix test failures.

iuioiua (changes requested):
It looks mostly good. There are a few nits that need to be addressed. Can you please run `deno fmt` to format the code?

kt3k:
Also please sign the CLA from the above @CLAassistant's comment.

iuioiua (approved):
I've taken the liberty of formatting and applying the suggested changes to take this over the line. Thank you for your contribution.

PR #5078, opened 2024-06-19
test(streams): improve test coverage

iuioiua:
Tests for [`mergeReadableStreams()`](https://app.codecov.io/gh/denoland/deno_std/blob/main/streams%2Fmerge_readable_streams.ts) are all that remain once this is merged.

### DIFF
--- a/streams/delimiter_stream.ts
+++ b/streams/delimiter_stream.ts
@@ -195,6 +195,8 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             chunkStart = disposition === "prefix"
               ? delimitedChunkEnd
               : inspectIndex;
+          } else {
+            throw new Error("unreachable");
           }
         }
       } else if (matchIndex === 0) {
@@ -286,6 +288,8 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             ? delimitedChunkEnd
             : inspectIndex;
           controller.enqueue(result);
+        } else {
+          throw new Error("unreachable");
         }
       } else {
         inspectIndex++;
--- a/streams/delimiter_stream.ts
+++ b/streams/delimiter_stream.ts
@@ -195,8 +195,6 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             chunkStart = disposition === "prefix"
               ? delimitedChunkEnd
               : inspectIndex;
-          } else {
-            throw new Error("unreachable");
           }
         }
       } else if (matchIndex === 0) {
@@ -288,8 +286,6 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             ? delimitedChunkEnd
             : inspectIndex;
           controller.enqueue(result);
-        } else {
-          throw new Error("unreachable");
         }
       } else {
         inspectIndex++;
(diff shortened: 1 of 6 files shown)


iuioiua:
> Tests for [`mergeReadableStreams()`](https://app.codecov.io/gh/denoland/deno_std/blob/main/streams%2Fmerge_readable_streams.ts) are all that remain once this is merged.

@BlackAsLight, might this be something you'd be interested in doing?

streams/early_zip_readable_streams.ts
```diff
@@ -88,21 +88,20 @@ export function earlyZipReadableStreams<T>(
...
         while (true) {
           for (const reader of readers) {
             const { value, done } = await reader.read();
             if (!done) {
               controller.enqueue(value!);
             } else {
               await Promise.all(readers.map((reader) => reader.cancel()));
-              break loop;
+              controller.close();
+              break;
```
kt3k, line 99: Looks like this doesn't break from outer while-loop anymore?

streams/early_zip_readable_streams.ts
```diff
@@ -88,21 +88,20 @@ export function earlyZipReadableStreams<T>(
...
             }
           }
         }
-        controller.close();
-      } catch (e) {
-        controller.error(e);
+      } catch (error) {
+        controller.error(error);
       }
     },
```
BlackAsLight, line 106: How is this not an infinite loop? The break keyword is breaking out of the for loop only to go around again in while loop, to maybe throw an error.

Could this readable stream not be better done with the pull method? That way it's not continually filling up the queue, in the stance the queue isn't being emptied.
```ts
async pull(controller) {
  for (const reader of readers) {
    const { done, value } = await reader.read()
    if (done) {
      await Promise.all(readers.map(reader => reader.cancel()))
      return controller.close()
    }
    controller.enqueue(value)
  }
}
```

iuioiua: I made a mistake - reverted now. Would you like to do this in a separate performance improvement PR?

streams/early_zip_readable_streams.ts
iuioiua: Reverted. My bad.

streams/delimiter_stream.ts
```diff
@@ -195,8 +195,6 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             chunkStart = disposition === "prefix"
               ? delimitedChunkEnd
               : inspectIndex;
-          } else {
-            throw new Error("unreachable");
```
kt3k, line 199: I prefer to keep this line. If the user his this line, this error helps us debugging the issue.

streams/delimiter_stream.ts
```diff
@@ -288,8 +286,6 @@ export class DelimiterStream extends TransformStream<Uint8Array, Uint8Array> {
             ? delimitedChunkEnd
             : inspectIndex;
           controller.enqueue(result);
-        } else {
-          throw new Error("unreachable");
```
kt3k, line 292: Ditto. I'd prefer to keep this

streams/delimiter_stream.ts
iuioiua: Reverted for now, but created #5090 for further investigation.

kt3k (approved):
LGTM 👍

Issue #3587, opened 2023-08-30
streams: add APIs from stream-consumers


kt3k:
ref: https://github.com/jimmywarting/stream-consumers

These 4 APIs look very handy for handling readable streams. How about adding these to `std/streams`?

```js
import { arrayBuffer, blob, json, text } from 'stream-consumers'
```

0f-0b:
I don't think these functions provide much value. One can already write, for example, `new Response(readable).json()`, as [noted in the readme](https://github.com/jimmywarting/stream-consumers#alternative-native-methods). An async iterable or a Node.js `Readable` can first be converted with `ReadableStream.from`.

kt3k:
> new Response(readable).json(), as [noted in the readme](https://github.com/jimmywarting/stream-consumers#alternative-native-methods).

That works, but as described in that readme, it feels still 'hack' and 'abuse' especially when you are not working on http server or client

PR #5079, opened 2024-06-19
BREAKING(streams): have `toJson()` and `toText()` accept `ReadableStream<Uint8Array>`

iuioiua:
### What's changed

Previously, [`toJson()`](https://jsr.io/@std/streams/doc/to-json/~/toJson) and [`toText()`](https://jsr.io/@std/streams/doc/to-text/~/toText) accepted `ReadableStream`, without the type of stream being defined. Now, these functions accept `ReadableStream<Uint8Array>`.

### Why this change was made

This change was made to align `toJson()` to [`Request.json()`](https://developer.mozilla.org/en-US/docs/Web/API/Request/json) and [`Response.json()`](https://developer.mozilla.org/en-US/docs/Web/API/Response/json), and `toText()` to [`Request.text()`](https://developer.mozilla.org/en-US/docs/Web/API/Request/text) and [`Response.text()`](https://developer.mozilla.org/en-US/docs/Web/API/Response/text), which seems to be the most common use case for these functions. The change also aligned these two functions with the similarly-oriented [`toBlob()`](https://jsr.io/@std/streams/doc/to-blob/~/toBlob), which did correctly accept `ReadableStream<Uint8Array>`.

### Migration guide

To migrate, ensure that a `ReadableStream` of `Uint8Array`s is being passed to these functions:
```diff
  import { toText } from "@std/streams/to-json";

- const stream = ReadableStream(["Hello, ", "world!"]);
+ const stream = ReadableStream(["Hello, ", "world!"]).pipeThrough(new TextEncoderStream());
  await toText(stream);
```

In most cases, adding `.pipeThrough(new TextEncoderStream())` for existing use cases that use `ReadableStream<string>` is all that's needed to migrate.

### Discussion

Alternatively, all 3 functions could be modified to accept `ReadbleStream<Uint8Array> | ReadableStream<string>`. However, this wasn't done in order to align with the `Request` and `Response` objects and slightly improve performance while still catering to the majority of use cases.

### DIFF
--- a/streams/mod.ts
+++ b/streams/mod.ts
@@ -9,7 +9,8 @@
  * import { toText } from "@std/streams";
  * import { assertEquals } from "@std/assert/assert-equals";
  *
- * const stream = ReadableStream.from("Hello, world!");
+ * const stream = ReadableStream.from(["Hello, ", "world!"])
+ *   .pipeThrough(new TextEncoderStream());
  * const text = await toText(stream);
  *
  * assertEquals(text, "Hello, world!");
--- a/streams/to_blob.ts
+++ b/streams/to_blob.ts
@@ -3,7 +3,8 @@
 
 /**
  * Converts a {@linkcode ReadableStream} of {@linkcode Uint8Array}s to a
- * {@linkcode Blob}. Works the same as {@linkcode Response.blob}.
+ * {@linkcode Blob}. Works the same as {@linkcode Request.blob} and
+ * {@linkcode Response.blob}.
  *
  * @param stream A `ReadableStream` of `Uint8Array`s to convert into a `Blob`.
  * @returns A `Promise` that resolves to the `Blob`.
--- a/streams/to_json.ts
+++ b/streams/to_json.ts
@@ -4,7 +4,7 @@
 import { toText } from "./to_text.ts";
 
 /**
- * Converts a JSON-formatted {@linkcode ReadableSteam} of strings or
+ * Converts a JSON-formatted {@linkcode ReadableSteam} of
  * {@linkcode Uint8Array}s to an object. Works the same as
  * {@linkcode Response.json}.
  *
@@ -20,13 +20,13 @@ import { toText } from "./to_text.ts";
  *   "[1, true",
  *   ', [], {}, "hello',
  *   '", null]',
- * ]);
+ * ]).pipeThrough(new TextEncoderStream());
  * const json = await toJson(stream);
  * assertEquals(json, [1, true, [], {}, "hello", null]);
  * ```
  */
 export function toJson(
-  readableStream: ReadableStream,
+  readableStream: ReadableStream<Uint8Array>,
 ): Promise<unknown> {
   return toText(readableStream).then(JSON.parse);
 }
--- a/streams/to_json.ts
+++ b/streams/to_json.ts
@@ -6,7 +6,7 @@ import { toText } from "./to_text.ts";
 /**
  * Converts a JSON-formatted {@linkcode ReadableSteam} of
  * {@linkcode Uint8Array}s to an object. Works the same as
- * {@linkcode Response.json}.
+ * {@linkcode Request.json} and {@linkcode Response.json}.
  *
  * @param readableStream A `ReadableStream` whose chunks compose a JSON.
  * @returns A promise that resolves to the parsed JSON.
--- a/streams/to_json_test.ts
+++ b/streams/to_json_test.ts
@@ -6,18 +6,5 @@ import { toJson } from "./to_json.ts";
 Deno.test("toJson()", async () => {
   const byteStream = ReadableStream.from(["[", "1, 2, 3, 4", "]"])
     .pipeThrough(new TextEncoderStream());
-
   assertEquals(await toJson(byteStream), [1, 2, 3, 4]);
-
-  const stringStream = ReadableStream.from([
-    '{ "a": 2,',
-    ' "b": 3,',
-    ' "c": 4 }',
-  ]);
-
-  assertEquals(await toJson(stringStream), {
-    a: 2,
-    b: 3,
-    c: 4,
-  });
 });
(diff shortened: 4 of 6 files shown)


kt3k:
> This change was made to align toJson() to [Request.json()](https://developer.mozilla.org/en-US/docs/Web/API/Request/json) and [Response.json()](https://developer.mozilla.org/en-US/docs/Web/API/Response/json), and toText() to [Request.text()](https://developer.mozilla.org/en-US/docs/Web/API/Request/text) and [Response.text()](https://developer.mozilla.org/en-US/docs/Web/API/Response/text),

I don't understand why we need to align them to Response methods. This seems unnecessarily reducing the use cases.

iuioiua:
That's what it originally said in the descriptions of these functions. Alternatively, we can go ahead with what I described in the Discussion section of the initial comment, but I'd prefer not to.

kt3k:
The original inspiration for these APIs are [stream/consumers](https://nodejs.org/api/webstreams.html#webstreams_utility_consumers) APIs of Node.js (see https://github.com/denoland/deno_std/issues/3587 ), and they accept streams of strings

iuioiua:
Then why don't users use `node:stream/consumers`? It's implemented cleanly in Deno for server-side use cases. For most browser-based use cases, one would likely just need the `Request` and `Response` methods.

kt3k:
> Then why don't users use node:stream/consumers? It's implemented cleanly in Deno for server-side use cases. For most browser-based use cases, one would likely just need the Request and Response methods.

I don't see the point well. Is this argument related to this PR? Or do you suggest removing these APIs?

iuioiua:
Yes, I suggest removing them. Frankly, I've never understood the reasons for adding them in the first place 😅

kt3k:
Maybe let's keep discussing that as RC feedback to `streams`

iuioiua:
This almost slipped past us before stabilizing `@std/streams`. What would you like to do, @kt3k? I suggest proceeding with this PR, as it seems to support almost all use cases ([GH search](https://github.com/search?q=%2Fstd.*%5C%2Fstreams%2F%20%22toText%22&type=code)).

kt3k:
I still don't understand why excluding string (and other type) input is desirable. Why do we need to align the behavior to `Response.prototype.toJson`? What's the benefit?

iuioiua:
I think a better question is why should we support `ReadableStream<string>` when `ReadableStream<Uint8Array>` is the primary use case? These 3 functions were deliberately created to align with `Response` methods (it's in the their descriptions), but extending them to also support `ReadableStream<string>` seems like it was an arbitrary decision. As a Standard Library, we should be consistent and not do more than we need to, across the board.

kt3k:
> why should we support ReadableStream<string> when ReadableStream<Uint8Array> is the primary use case?

Because they are exposed as public API and we should avoid unnecessary breaking changes as far as possible.

> These 3 functions were deliberately created to align with Response methods (it's in the their descriptions)

It's also explicitly documented that they accept string inputs. So this is intentional design, not a mistake.

kt3k:
Also there seem no downside of accepting `ReadableStream<string>`. Isn't it just sensible/predictable enhancement to `Response.prototype.json`? We sometimes do such enhancement to Web spec inspired APIs. e.g. `crypto.digest` of `std/crypto` accepts `Iterable` and `AsyncIterable` data sources in addition to ArrayBuffer(View)s

iuioiua:
> Also there seem no downside of accepting `ReadableStream<string>`.

If we say that enough, we end up with bloated implementations.



> Isn't it just sensible/predictable enhancement to `Response.prototype.json`? We sometimes do such enhancement to Web spec inspired APIs. e.g. `crypto.digest` of `std/crypto` accepts `Iterable` and `AsyncIterable` data sources in addition to ArrayBuffer(View)s

Yes, but that should be vetted on a case-by-case basis, and within reason. There seems to be no good reason to do this for these functions, as most are using `ReadableStream<Uint8Array>`. I think the Standard Library should cater to majority of use cases. Current users of these functions would also enjoy a small performance boost.

```
import { toJson } from "./streams/to_json.ts";

async function toText(stream: ReadableStream<Uint8Array>): Promise<string> {
  const textDecoder = new TextDecoder();
  const reader = stream.getReader();
  let result = "";

  while (true) {
    const { done, value } = await reader.read();

    if (done) {
      break;
    }

    result += textDecoder.decode(value, { stream: true });
  }
  result += textDecoder.decode();
  return result;
}

function toJson2(stream: ReadableStream<Uint8Array>): Promise<unknown> {
  return toText(stream).then(JSON.parse);
}

function createStream() {
  const chunks = [
    "[1, true",
    ', [], {}, "hello',
    '", null]',
  ];
  return ReadableStream.from(chunks).pipeThrough(new TextEncoderStream());
}

const group = "x";

Deno.bench("current", { group, baseline: true }, async () => {
  await toJson(createStream());
});

Deno.bench("new", { group }, async () => {
  await toJson2(createStream());
});
```
```
cpu: Apple M2
runtime: deno 1.45.2 (aarch64-apple-darwin)

file:///Users/asher/GitHub/std/bench.ts

group x
current        18.36 µs/iter      54,463.3  (13.83 µs … 389.46 µs) 16.83 µs 46.58 µs 96.5 µs
new            17.38 µs/iter      57,527.5   (15.5 µs … 222.79 µs) 16.21 µs 37.42 µs 99.5 µs

summary
  current
   1.06x slower than new
```

kt3k:
> as most are using ReadableStream<Uint8Array>.

I disagree with this observation. There are several std APIs that return `ReadableStream<string>`:
- `CsvStringifyStream`
- `JsonStringifyStream`
- `TextDelimiterStream`
- `TextLineStream`

Also this change causes perf penalty when the user has `ReadableStream<string>`. With this change the user once needs to encode it into `ReadableStream<Uint8Array>`, then decode it to text again using `toText`.

iuioiua:
Ok. I'll submit a superseding PR that clarifies these functions and unblocks `streams@1.0.0`.

PR #5082, opened 2024-06-19
fix(streams): prevent `earlyZipReadableStreams()` from possibly using excessive memory

BlackAsLight:
Jumping from: https://github.com/denoland/deno_std/pull/5078#discussion_r1645446809

This pull request changes the structure of the readable stream to use a pull method instead of the start method, stopping it from the possibility of using excessive memory, possibly overflowing it if not consumed fast enough, or at all. This alternative also offers a way to cancel the provided reason streams from the returned stream.

## Current Version - overloaded the memory
Screenshot 2024-06-19 at 16 20 10

## Alterative Version
Screenshot 2024-06-19 at 16 20 30

## Benchmark
This refactor did turn out to be slightly slower than the existing one in a benchmark.
Screenshot 2024-06-19 at 16 32 06

### DIFF
--- a/streams/early_zip_readable_streams.ts
+++ b/streams/early_zip_readable_streams.ts
@@ -91,7 +91,9 @@ export function earlyZipReadableStreams<T>(
         const { done, value } = await readers[i]!.read();
         if (done) {
           await Promise.all(
-            readers.map((reader) => reader.cancel(`Stream ${i} ended`)),
+            readers.map((reader) =>
+              reader.cancel(`Stream at index ${i} ended`)
+            ),
           );
           controller.close();
           return;
--- a/streams/early_zip_readable_streams.ts
+++ b/streams/early_zip_readable_streams.ts
@@ -93,7 +93,8 @@ export function earlyZipReadableStreams<T>(
           await Promise.all(
             readers.map((reader) => reader.cancel("early_zip_ended")), // A better cancel message should maybe be put here?
           );
-          return controller.close();
+          controller.close();
+          return;
         }
         controller.enqueue(value);
       }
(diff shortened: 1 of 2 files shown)


streams/early_zip_readable_streams_test.ts
```diff
@@ -60,3 +60,20 @@ Deno.test("earlyZipReadableStreams() can zip three streams", async () => {
     "3",
   ]);
 });
+
+Deno.test("earlyZipReadableStreams() forwards cancel()", async () => {
```
iuioiua, line 64: Can you please use an assertion somewhere here?

iuioiua (changes requested):
This is great. Thank you for the solid troubleshooting.

streams/early_zip_readable_streams_test.ts
BlackAsLight: I used one in the cancel method

iuioiua: Sorry, missed that.

streams/early_zip_readable_streams.ts
```diff
@@ -84,26 +84,22 @@
...
-              await Promise.all(readers.map((reader) => reader.cancel()));
-              break loop;
-            }
-          }
+    async pull(controller) {
+      for (const reader of readers) {
+        const { done, value } = await reader.read();
+        if (done) {
+          await Promise.all(
+            readers.map((reader) => reader.cancel("early_zip_ended")), // A better cancel message should maybe be put here?
```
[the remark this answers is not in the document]
BlackAsLight: Is it correct to say it ended early though since if all the streams are of the same length then the first stream will be detected as finished, with all streams getting this message even though they're all empty.

Maybe something like \`Early Zip Stream ended at index: ${i}`?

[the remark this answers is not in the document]
iuioiua: Good point. `Stream ended at index ${i}` SGTM—no need to reiterate the class name.

kt3k (approved):
LGTM

iuioiua (approved):
I made a final tweak to the cancel reason. Thank you very much, @BlackAsLight. Great work.

PR #5086, opened 2024-06-19
test(fs): check `ensureDir()` is not racy

kt3k:
This PR adds a check for #3233


iuioiua:
Nice! Are there other APIs we should do this for?

PR #5089, opened 2024-06-19
refactor(fs): improve `expandGlob()` implementation and testing

kt3k:
This PR improves the line coverage of `fs/expand_glob.ts`.

part of #3713

### DIFF
--- a/fs/expand_glob.ts
+++ b/fs/expand_glob.ts
@@ -290,14 +290,9 @@ export async function* expandGlob(
   const shouldInclude = (path: string): boolean =>
     !excludePatterns.some((p: RegExp): boolean => !!path.match(p));
 
-  let fixedRoot = isGlobAbsolute
-    ? winRoot !== undefined ? winRoot : "/"
-    : absRoot;
+  let fixedRoot = isGlobAbsolute ? winRoot ?? "/" : absRoot;
   while (segments.length > 0 && !isGlob(segments[0]!)) {
-    const seg = segments.shift();
-    if (seg === undefined) {
-      throw new TypeError("Unexpected undefined segment");
-    }
+    const seg = segments.shift()!;
     fixedRoot = joinGlobs([fixedRoot, seg], globOptions);
   }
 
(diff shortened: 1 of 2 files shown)


fs/expand_glob.ts
```diff
@@ -316,12 +311,8 @@ export async function* expandGlob(
...
     } else if (globSegment === "..") {
       const parentPath = joinGlobs([walkInfo.path, ".."], globOptions);
-      try {
-        if (shouldInclude(parentPath)) {
-          return yield await createWalkEntry(parentPath);
-        }
-      } catch (error) {
-        throwUnlessNotFound(error);
+      if (shouldInclude(parentPath)) {
+        return yield await createWalkEntry(parentPath);
```
iuioiua, line 315: Can you please explain this change?

fs/expand_glob_test.ts
```diff
@@ -78,6 +84,42 @@ Deno.test("expandGlobSync() with wildcard input returns all test data", function
...
+    "subdir",
+  ]);
+});
+
+Deno.test("expandGlob() returns empty array if path doesn't exist", async function () {
+  assertEquals(await expandGlobArray("nonexistent", EG_OPTIONS), []);
+  assertEquals(expandGlobSyncArray("nonexistent", EG_OPTIONS), []);
+});
+
+Deno.test(
```
iuioiua, line 110: Could you please define the full expected error message in the assertions here?

fs/expand_glob.ts
iuioiua, line 1: I think we should move `advanceMatch()` generator functions out of their respective `expandGlob[Sync]()` functions as `advanceMatch()` and `advanceMatchSync()`.

iuioiua (changes requested):

fs/expand_glob.ts
kt3k: This try-catch block tries to handle `Deno.errors.NotFound` thrown from `Deno.stat` (called in `createWalkEntry`), and re-throw other errors as is (see `throwUnlessNotFound` for details).

However, `Deno.errors.NotFound` is not possible to  be thrown here because the target path (`parentPath`) is created by `joinGlobs([walkInfo.path, ".."])` which is parent path of existing path (`walkInfo` is only created for existing path). (e.g. If `foo/bar/baz` exists, `foo/bar` needs to exist.)

If `Deno.errors.NotFound` is not possible to be thrown, this try-catch block does nothing.

iuioiua: Cool. As I thought. Thanks!

iuioiua (approved):
Nice. Perhaps we can move `advanceMatch()` out of these functions in a follow-up PR.

PR #5093, opened 2024-06-20
test(jsonc): remove dead code and improve testing

kt3k:
This improves the test coverage of `jsonc` from 85.26% to 97.42% (missed lines are reduced from  46 to 8)

part of #3713

### DIFF
--- a/jsonc/parse_test.ts
+++ b/jsonc/parse_test.ts
@@ -8,6 +8,10 @@ import {
   assertThrows,
 } from "@std/assert";
 
+import "./testdata/JSONTestSuite/test.ts";
+import "./testdata/node-jsonc-parser/test.ts";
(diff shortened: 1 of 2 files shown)


jsonc/parse_test.ts
```diff
@@ -8,6 +8,10 @@ import {
   assertThrows,
 } from "@std/assert";
 
+import "./testdata/JSONTestSuite/test.ts";
+import "./testdata/node-jsonc-parser/test.ts";
+import "./testdata/test262/test.ts";
```
kt3k, line 13: I think these test cases used to run without these imports, but currently they are not discovered by `deno test` command.

iuioiua (approved):
Nice

Issue #5090, opened 2024-06-19
to-do: investigate possibly unreachable LOC in `@std/streams/delimiter-stream`


iuioiua:
We need to confirm whether the following LOC are indeed unreachable. If they are, with sufficient validation, they can be removed.
https://github.com/denoland/deno_std/blob/c52c6ee057cfd95ecc7a489dc1d0f99cc1d2e0c6/streams/delimiter_stream.ts#L199
https://github.com/denoland/deno_std/blob/c52c6ee057cfd95ecc7a489dc1d0f99cc1d2e0c6/streams/delimiter_stream.ts#L292

PR #5094, opened 2024-06-20
refactor(streams): simplify `DelimiterStream`

BlackAsLight:
When exploring https://github.com/denoland/deno_std/issues/5090, the code looked quite overcomplicated so I tried making my own version, which passes with existing tests, and is much shorter and simpler.

When benching the code to the existing one, it had 3-10x worse results when benching in the deno_std repo, but very comparable results when both abstracted to a separate module. Idk why there was such a weird difference.
## Benchmark
Screenshot_2024-06-20_at_18 48 14

### DIFF
--- a/.github/dependency_graph.svg
+++ b/.github/dependency_graph.svg
@@ -66,7 +66,7 @@
 <!-- cli -->
 <g id="node7" class="node">
 <title>cli</title>
-<ellipse fill="none" stroke="black" cx="171" cy="-144" rx="36" ry="36"/>
+<ellipse fill="lightgreen" stroke="black" cx="171" cy="-144" rx="36" ry="36"/>
 <text text-anchor="middle" x="171" y="-139.8" font-family="Times,serif" font-size="14.00">cli</text>
 </g>
 <!-- collections -->
@@ -90,7 +90,7 @@
 <!-- streams -->
 <g id="node11" class="node">
 <title>streams</title>
-<ellipse fill="none" stroke="black" cx="81" cy="-144" rx="36" ry="36"/>
+<ellipse fill="lightgreen" stroke="black" cx="81" cy="-144" rx="36" ry="36"/>
 <text text-anchor="middle" x="81" y="-139.8" font-family="Times,serif" font-size="14.00">streams</text>
 </g>
 <!-- csv&#45;&gt;streams -->
@@ -194,7 +194,7 @@
 <!-- fs -->
 <g id="node21" class="node">
 <title>fs</title>
-<ellipse fill="none" stroke="black" cx="1019" cy="-144" rx="36" ry="36"/>
+<ellipse fill="lightgreen" stroke="black" cx="1019" cy="-144" rx="36" ry="36"/>
 <text text-anchor="middle" x="1019" y="-139.8" font-family="Times,serif" font-size="14.00">fs</text>
 </g>
 <!-- path -->
--- a/deno.json
+++ b/deno.json
@@ -27,7 +27,7 @@
     "@std/expect": "jsr:@std/expect@^0.224.4",
     "@std/fmt": "jsr:@std/fmt@^0.225.4",
     "@std/front-matter": "jsr:@std/front-matter@^0.224.2",
-    "@std/fs": "jsr:@std/fs@^0.229.3",
+    "@std/fs": "jsr:@std/fs@^1.0.0-rc.1",
     "@std/html": "jsr:@std/html@^1.0.0-rc.1",
     "@std/http": "jsr:@std/http@^0.224.5",
     "@std/ini": "jsr:@std/ini@^0.225.1",
--- a/fs/copy_test.ts
+++ b/fs/copy_test.ts
@@ -122,26 +122,54 @@ testCopy(
 testCopy(
   "copy() copies with preserve timestamps",
   async (tempDir: string) => {
-    const srcFile = path.join(testdataDir, "copy_file.txt");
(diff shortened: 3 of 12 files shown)


kt3k:
> it had 3-10x worse results when benching in the deno_std repo,

Can you elaborate on this? In what setup did you compare it?

BlackAsLight:
> Can you elaborate on this? In what setup did you compare it?

It was more or less the same as the above image, but importing the local current version instead of the one on JSR. The local one took around 2.5ms while my version took around 6-8ms on average sometimes as much as 30ms. But when doing it in a separate repo, like in the image, both dramatically improved in speed.

BlackAsLight:
I figured out what I was doing wrong and now am getting consistent results again. In the image above, I set the chunk size to zero instead of 1_000 so the chunks were always empty. Filling it in and I am getting 2.5ms (current) to 8ms (pull request).

## Bench Source Code
```ts
import { DelimiterStream as CurrentStream } from "./streams/delimiter_stream.ts";

const x = Array.from(function* () {
  for (let i = 0; i < 1_000; ++i) {
    yield new Uint8Array(1_000).map(() => Math.floor(Math.random() * 256));
  }
}());

const delimiter = new Uint8Array(2).map(() => Math.floor(Math.random() * 256));

Deno.bench(async function current() {
  await Array.fromAsync(
    ReadableStream
      .from(x.slice())
      .pipeThrough(new CurrentStream(delimiter)),
  );
});

Deno.bench(async function alter() {
  await Array.fromAsync(
    ReadableStream
      .from(x.slice())
      .pipeThrough(new AlternativeStream(delimiter)),
  );
});

export interface DelimiterStreamOptions {
  disposition: "suffix" | "prefix" | "discard";
}

export class AlternativeStream extends TransformStream<Uint8Array, Uint8Array> {
  constructor(
    delimiter: Uint8Array,
    options: Partial<DelimiterStreamOptions> = { disposition: "discard" },
  ) {
    let buffer = new Uint8Array(0);
    let chunkIndex = 0;
    let delimiterIndex = 0;
    super({
        const concat = new Uint8Array(buffer.length + chunk.length);
        concat.set(buffer);
        concat.set(chunk, buffer.length);
        chunk = concat;

        while (chunkIndex < chunk.length) {
          if (chunk[chunkIndex] === delimiter[delimiterIndex]) {
            ++delimiterIndex;
            ++chunkIndex;
          } else {
            chunkIndex -= delimiterIndex - 1;
            delimiterIndex = 0;
          }
          if (delimiter.length == delimiterIndex) {
            // If suffix, don't exclude the end delimiter from enqueuing.
            controller.enqueue(
              chunk.slice(
                0,
                chunkIndex -
                  (options.disposition === "suffix" ? 0 : delimiterIndex),
              ),
            );
            // If prefix, include the end delimiter in the remaining chunk.
            chunk = chunk.slice(
              chunkIndex -
                (options.disposition === "prefix" ? delimiterIndex : 0),
            );
            // If prefix, skip checking the first delimiter.length to avoid an infinite loop of enqueuing `new Uint8Array(0)`.
            chunkIndex = options.disposition === "prefix" ? delimiterIndex : 0;
            delimiterIndex = 0;
          }
        }
        buffer = chunk;
      },
      flush(controller) {
        controller.enqueue(buffer);
      },
    });
  }
}
```
## Bench Results
```
soul@SilverMini deno_std % deno bench bench.ts
Check file:///Users/soul/Projects/deno_std/bench.ts
cpu: Apple M1
runtime: deno 1.44.4 (aarch64-apple-darwin)

file:///Users/soul/Projects/deno_std/bench.ts
current         2.56 ms/iter         391.3     (2.33 ms … 4.21 ms) 2.6 ms 3.41 ms 3.47 ms
alter           7.99 ms/iter         125.1      (7.5 ms … 9.93 ms) 8.1 ms 9.93 ms 9.93 ms
```

On another note, there is a difference in behavour between this version and the existing version when the delimiter is of length zero. The existing version returns one large Uint8Array, where this version slices all chunks into lengths of one. There doesn't seem to be a test case to dictate what the correct behaviour should be when the delimiter is empty, but I think throwing an error in such a case might be appropriate as there are better ways to achieve either result.

kt3k:
Can you identify the reason of perf degradation? 2x perf degradation doesn't sound great.

> There doesn't seem to be a test case to dictate what the correct behaviour should be when the delimiter is empty, but I think throwing an error in such a case might be appropriate as there are better ways to achieve either result.

Throwing error for empty delimiter sounds good to me.

BlackAsLight:
> Can you identify the reason of perf degradation? 2x perf degradation doesn't sound great.
> 
> > There doesn't seem to be a test case to dictate what the correct behaviour should be when the delimiter is empty, but I think throwing an error in such a case might be appropriate as there are better ways to achieve either result.
> 
> Throwing error for empty delimiter sounds good to me.

If I was to guess, it's because this version is merging the chunks on each call to transform, therefore needing to move memory around, where I believe the current version only merges them once it has identified a match. I haven't tested this theory as of yet though.

BlackAsLight:
After exploring the theory previously mentioned I was only able to get slightly better results with a delimiter of length 2. I also learned that when a delimiter of length 3+ pushed the time to take ~40 ms/iter instead of the ~8 ms/iter, while this new theory seems to have the inverse effect in that a delimiter of length 1 would be ~20 ms/iter while a delimiter of length 2+ is ~8 ms/iter.

The current implementation throughout these different delimiter lengths had a consistent ~2 ms/iter with length 2+ and ~3 ms/iter with length 1. Based off that I don't think this is a good pull request any more.

PR #5097, opened 2024-06-20
BREAKING(json): remove `writableStrategy` and `readableStrategy` options

### LINKED ISSUE #4113: suggestion: deprecate `writableStrategy` and `readableStrategy` argument properties for `std/json` constructors

iuioiua:
### What's changed

The `writableStrategy` and `readableStrategy` options have been removed from the constructors from `ConcatenatedJsonParseStream`, `JsonParseStream` and `JsonStringifyStream`.

### Why this change was made

These options were removed because they aren't used in most use cases. For this reason, none of the other `TransformStream` implementations in the Standard Library have these options.

### Who this affects

Only those who use these options.

### Migration guide

If you require the use of these options, please implement a custom class that uses them.

```
class CustomJsonStringifyStream extends TransformStream<unknown, string> {
  constructor(
    writableStrategy: QueuingStrategy<unknown>,
    readableStrategy: QueuingStrategy<string>,
  ) {
    super(
      {
          controller.enqueue(JSON.stringify(chunk));
        },
      },
      writableStrategy,
      readableStrategy,
    );
  }
}
```

### Related

Closes #4113

### DIFF
--- a/json/_test_common.ts
+++ b/json/_test_common.ts
@@ -2,16 +2,14 @@
 import { assertEquals, assertRejects } from "@std/assert";
 import type { ConcatenatedJsonParseStream } from "./concatenated_json_parse_stream.ts";
 import type { JsonParseStream } from "./json_parse_stream.ts";
-import type { ParseStreamOptions } from "./common.ts";
 
 export async function assertValidParse(
   transform: typeof ConcatenatedJsonParseStream | typeof JsonParseStream,
   chunks: string[],
   expect: unknown[],
-  options?: ParseStreamOptions,
 ) {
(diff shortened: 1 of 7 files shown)


kt3k (approved):
LGTM

PR #5099, opened 2024-06-21
BREAKING(testing): remove deprecated `asserts` module

iuioiua:
### What's changed

The deprecated module, [`@std/testing/asserts`](https://jsr.io/@std/testing/doc/asserts/~), has been removed in favor of its own package, [`@std/assert`](https://jsr.io/@std/assert).

### Why this change was made

This module was moved because it was sufficiently large to be its own package. The move also prevents users from having to import the entirety of all assertion code and instead only import the symbols they require (see below).

### Migration guide

To migrate, import from `@std/assert` instead of `@std/testing/asserts`.

```diff
- import { assertEquals, assertInstanceOf } from "@std/testing/asserts";
+ import { assertEquals, assertInstanceOf } from "@std/assert";

// ...
```

```diff
- import { assertEquals } from "@std/testing/asserts";
+ import { assertEquals } from "@std/assert/assert-equals";

// ...
```


kt3k:
LGTM, but let's land after the next release (I plan to do it on Tuesday)

kt3k (approved):
As we already landed one breaking change to testing #5130. Let's land this as well.

LGTM

PR #5101, opened 2024-06-21
BREAKING(http): move `unstable-signed-cookie` to `signed-cookie`

iuioiua:
### What's changed

The [`@std/http/unstable-signed-cookie`](https://jsr.io/@std/http/doc/unstable-signed-cookie/~/) module has been renamed `@std/http/signed-cookie`. This module is still unstable, and the documentation states this using a new description preface and `@experimental` JSDoc tag.

### Why this change was made

This change aligns the module with other unstable modules in the Standard Library, which uses JSDocs to reflect instability rather than the `unstable—*` module name prefix.

### Migration guide

To migrate, import from `@std/http/signed-cookie` instead of `@std/http/unstable-signed-cookie`.

```diff
- import { parseSignedCookie, signCookie, verifyCookie } from "@std/http/unstable-signed-cookie";
+ import { parseSignedCookie, signCookie, verifyCookie } from "@std/http/signed-cookie";

// ...
```


kt3k (approved):
LGTM

PR #5109, opened 2024-06-22
docs(http): remove outdated mention of `CookieMap`

elisee:
It was removed in #4179


iuioiua (approved):
Good catch! Thank you.

Issue #4999, opened 2024-06-10
stabilize `@std/text`


kt3k:
part of #4600 

- [x] fully documented
- [x] fully tested
- [x] no ongoing breaking change discussion
- [x] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

PR #5110, opened 2024-06-22
feat(text/unstable): add `toConstantCase()`

guy-borderless:
per discussion at https://github.com/denoland/deno_std/issues/3125#issuecomment-2180354086


iuioiua (approved):
Nice! Thank you. I prefer this name more. Let's see what others think.

kt3k:
Looks good to me. `@std/text` is now in the process of stabilization #4999. Let's land after 1.0.0 of `@std/text` published.

guy-borderless:
The current implementation of splitToWords() only supports Latin letters. This commit adds a unicode-compatible implementation + support for WordSpliting according to case. I can open it in a separate PR if you prefer.

iuioiua:
> The current implementation of splitToWords() only supports Latin letters. This commit adds a unicode-compatible implementation + support for WordSpliting according to case. I can open it in a separate PR if you prefer.

Nice! Thank you. Yes, could you please open a separate PR?

guy-borderless:
> Nice! Thank you. Yes, could you please open a separate PR?

[moved](https://github.com/denoland/std/pull/5447)

iuioiua (changes requested):
Since this PR was created, we split up the word utility files in #5428. Can you please rebase?

iuioiua (approved):

PR #5112, opened 2024-06-23
feat(text): add `distanceFn` option to string comparison functions

### LINKED ISSUE #5111: text: Add `algorithm` option to string comparison functions

Currently, the string comparison functions in the `text` module (such as `wordSimilaritySort`) are hard coded to use Levenshtein distance, with a note saying "this metric may change in future versions", which could potentially break projects expecting consistent results.

An added option (similar to `caseSensitive`) in these functions for the user to provide their own algorithm, which could also be passed the already exported `levenshteinDistance` function. These functions could, like before, default to whatever algorithm Deno recommends (currently Levenshtein).

```ts
/** Options for {@linkcode compareSimilarity}. */
export interface CompareSimilarityOptions {
  /**
   * Whether the distance should include case.
   *
   * @default {false}
   */
  caseSensitive?: boolean;
  /**
  * Function used to compare two strings.
  * Lower number means closer match.
  * 0 means numbers match.
  * 
  * @default {@linkcode levenshteinDistance}
  */
  algorithm?: (str1: string, str2: string) => number;
}
```

RuiNtD:
Adds an `algorithm` option to the string comparison functions. Allows Deno to change the default metric in the future, while allowing users to continue using these functions with levenshteinDistance for consistency. 

Also adds an export for `getWordDistance`, which was previously defined in 2 different files, so that Deno's "recommended" metric at the time can accessed by the user. Literally just exports `levenshteinDistance`. If this is not desired, I can undo this.

No tests were included, since `text` doesn't include any other metrics to test with. If anyone could give me a dummy metric to use, I could add the tests.

Resolves #5111


text/get_word_distance.ts
```diff
@@ -0,0 +1,12 @@
...
+/**
+ * Compare two strings by similarity.
+
+ * Currently exports {@linkcode levenshteinDistance}.
+ * Metric used may be swapped / improved in the future.
+ *
+ * @returns The difference in distance. Lower number means closer match.
+ * 0 means numbers match.
+ */
+export const getWordDistance = levenshteinDistance;
```
kt3k, line 12: This export feels redundant to me.

text/compare_similarity.ts
```diff
@@ -13,6 +10,14 @@ export interface CompareSimilarityOptions {
...
    */
   caseSensitive?: boolean;
+  /**
+   * Function used to compare two strings.
+   * Lower number means closer match.
+   * 0 means numbers match.
+   *
+   * @default {getWordDistance}
+   */
+  algorithm?: (str1: string, str2: string) => number;
```
kt3k, line 20: `algorithm` sounds a bit too generic name for this option to me. How about `distance`?

iuioiua: Distance sounds like a variable rather than a function. How about `distanceCalculator`?

kt3k: I feel it's also common to call such function as 'distance' in CS context. For example 'distance' in levenshtein distance, hamming distance, etc

text/closest_string.ts
```diff
@@ -33,16 +34,14 @@ const getWordDistance = levenshteinDistance;
...
+  options?: ClosestStringOptions
 ): string {
   if (possibleWords.length === 0) {
     throw new TypeError(
-      "When using closestString(), the possibleWords array must contain at least one word",
+      "When using closestString(), the possibleWords array must contain at least one word"
     );
   }
-  const { caseSensitive } = { ...options };
+  const { caseSensitive, algorithm = getWordDistance } = { ...options };
```
timreichen, line 44: What is the reason that `options` is spread? Couldn't it just be 
```ts
const { caseSensitive, algorithm = getWordDistance } = options;
```

text/get_word_distance.ts
RuiNtD: As mentioned in the PR description:
> Also adds an export for getWordDistance, which was previously defined in 2 different files, so that Deno's "recommended" metric at the time can accessed by the user. Literally just exports levenshteinDistance. If this is not desired, I can undo this.

Without this export, the user has no way to get the current "recommended" Metric. Again, I can undo this if desired, and put it back into the two different files it came from.

Also, I re-exported the import instead of just using `export from` so that I could change the documentation.

text/compare_similarity.ts
RuiNtD: Other options could include:
- `distanceAlgorithm`
- `distanceMetric`
- `distanceFunction`
- `distanceFn`

I'm feeling more `distanceFn`. Let me know if this is fine. I'm also not sure if this would be consistent with other the other `std` libraries.

text/closest_string.ts
RuiNtD: Because `options` can be undefined.
Code_lqLHnLYpEC

text/get_word_distance.ts
iuioiua: I agree with Yoshiya—we don't need this exported. The default algorithm can be obtained by reading the documentation for the new `distanceFn` option.

iuioiua (changes requested):

text/closest_string.ts
```diff
@@ -33,16 +34,14 @@ const getWordDistance = levenshteinDistance;
...
+  options?: ClosestStringOptions
 ): string {
   if (possibleWords.length === 0) {
     throw new TypeError(
-      "When using closestString(), the possibleWords array must contain at least one word",
+      "When using closestString(), the possibleWords array must contain at least one word"
     );
   }
-  const { caseSensitive } = { ...options };
+  const { caseSensitive, distanceFn = getWordDistance } = { ...options };
```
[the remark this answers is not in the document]
RuiNtD: https://github.com/denoland/deno_std/pull/5112#discussion_r1650116537

text/compare_similarity.ts
```diff
@@ -45,14 +50,16 @@ export interface CompareSimilarityOptions {
  */
 export function compareSimilarity(
   givenWord: string,
-  options?: CompareSimilarityOptions,
+  options?: CompareSimilarityOptions
 ): (a: string, b: string) => number {
-  if (options?.caseSensitive) {
+  const { caseSensitive, distanceFn = getWordDistance } = { ...options };
```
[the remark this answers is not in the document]
RuiNtD: https://github.com/denoland/deno_std/pull/5112#discussion_r1650116537

text/compare_similarity.ts
```diff
@@ -13,6 +13,14 @@ export interface CompareSimilarityOptions {
    * @default {false}
    */
   caseSensitive?: boolean;
+  /**
+   * Function used to compare two strings.
+   * Lower number means closer match.
+   * 0 means numbers match.
+   *
+   * @default {getWordDistance}
```
kt3k, line 21:
```suggestion
   * @default {levenshteinDistance}
```

text/compare_similarity.ts
kt3k: `distanceFn` sounds fine to me

iuioiua (approved):
I removed the `getWordDistance` constant as it didn't do anything. Now, LGTM! Excellent idea and work, @RuiNtD! We'll merge this once `@std/text@1.0.0` is released.

iuioiua:
Sincere apologies, @RuiNtD, this PR completely slipped my mind and I completed #5448, which has the same idea as this one. Thank you for your contribution, and apologies I didn't merge this one first.

PR #5113, opened 2024-06-24
chore: `import_map.json` generator and checker

iuioiua:


kt3k:
I'm not in favor of having this file in std. Because this file is only needed by CLI repo, it seems confusing to have this file here.

iuioiua:
Yeah, it would be confusing... I've moved the script to https://gist.github.com/iuioiua/10f89676aee55c9f35e7d8b71cdbb9a4.

PR #5119, opened 2024-06-24
BREAKING(jsonc): remove `allowTrailingComma` option

iuioiua:
### What's changed

This change removes the [`ParseOptions.allowTrailingComma`](https://jsr.io/@std/jsonc/doc/~/ParseOptions.allowTrailingComma) option, and therefore the [`ParseOptions`](https://jsr.io/@std/jsonc/doc/~/ParseOptions) interface. For now on, trailing commas are allowed (the previous default) without the ability to change.

### Why this change was made

This change was made because setting `ParseOptions.allowTrailingComma` to `false` seemed to lack any useful use case. Also, having this added control also added some logic overhead. By removing this option, we slightly simplify the API surface of [`@std/jsonc`](https://jsr.io/@std/jsonc) while still remaining compatible with the majority of existing use cases.

### Migration guide

No further action is needed for users that didn't use the `options` argument of `parse()` or set `options.allowTrailingComma` to `true`.

For users that set `options.allowTrailingComma` to `false`, `parse()` now no longer throws a [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) when it encounters a trailing comma.


kt3k (approved):
LGTM

PR #5120, opened 2024-06-24
refactor(http): remove dead code and improve `UserAgent` testing

kt3k:
This PR removes some dead code and improves typings in `http/user_agent.ts`, and also adds test cases which exercise uncovered lines.

towards #3713

### DIFF
--- a/http/user_agent.ts
+++ b/http/user_agent.ts
@@ -146,7 +146,7 @@ const windowsVersionMap = new Map<string, string | string[]>([
 ]);
 
 function has(str1: string, str2: string): boolean {
-  return lowerize(str2).indexOf(lowerize(str1)) !== -1;
+  return lowerize(str2).includes(lowerize(str1));
 }
 
 function mapWinVer(str: string) {
--- a/http/user_agent.ts
+++ b/http/user_agent.ts
@@ -53,12 +53,11 @@ const ZEBRA = "Zebra";
 type ProcessingFn = (value: string) => string | undefined;
 
 type MatchingTuple = [matchers: [RegExp, ...RegExp[]], processors: (
-  string | [string, string] | [string, ProcessingFn] | [
-    string,
-    RegExp,
-    string,
-    ProcessingFn?,
-  ]
+  | string
+  | [string, string]
+  | [string, ProcessingFn]
+  | [string, RegExp, string]
+  | [string, RegExp, string, ProcessingFn]
 )[]];
 
 interface Matchers {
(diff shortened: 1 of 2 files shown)


http/user_agent.ts
```diff
@@ -146,15 +145,7 @@ const windowsVersionMap = new Map<string, string | string[]>([
...
 
-function has(str1: string | string[], str2: string): boolean {
-  if (Array.isArray(str1)) {
-    for (const el of str1) {
-      if (lowerize(el) === lowerize(str2)) {
-        return true;
-      }
-    }
-    return false;
-  }
```
kt3k, line 157: `str1` is always string with all `has` usages. This if-block is dead code.

http/user_agent.ts
```diff
@@ -208,9 +196,6 @@ function mapper(
               target[prop] = match ? match.replace(re, value) : undefined;
             } else {
               const [prop, re, value, fn] = processor;
-              if (!fn) {
-                throw new TypeError("Function must be defined in processor");
-              }
```
kt3k, line 213: If `processor` has 4 elements, the 4th element needs to be a function by its typing. So this if-block does nothing.

http/user_agent.ts
```diff
@@ -146,15 +145,7 @@ const windowsVersionMap = new Map<string, string | string[]>([
...
-  if (Array.isArray(str1)) {
-    for (const el of str1) {
-      if (lowerize(el) === lowerize(str2)) {
-        return true;
-      }
-    }
-    return false;
-  }
+function has(str1: string, str2: string): boolean {
   return lowerize(str2).indexOf(lowerize(str1)) !== -1;
```
iuioiua, line 149: How about `.includes()`?

http/user_agent.ts
```diff
@@ -184,21 +175,18 @@ function mapper(
...
     let k = 0;
     while (j < matchers.length && !matches) {
-      if (!matchers[j]) {
-        break;
-      }
       matches = matchers[j++]!.exec(ua);
 
       if (matches) {
         for (const processor of processors) {
-          const match = matches[++k]!;
```
kt3k, line 194: There were cases where `match` is actually `undefined`. So appending `!` here is not ideal.

http/user_agent.ts
iuioiua, line 1: Can we remove `lowerize()` and `trim()`? Are there any others we can remove?

iuioiua (changes requested):
Few nits

http/user_agent.ts
kt3k: lowerize and trim is used as `ProcessingFn` in processors. I think it's difficult to remove them

iuioiua (approved):

PR #5121, opened 2024-06-24
feat(streams/unstable): `toLines()`

### LINKED ISSUE #5071: feature request(streams):  lines(readable): AsyncIterable<string>
Idea from some discussion: A quick helper for getting an async iterable (or maybe another stream) of lines from a ReadableStream.

Example signature: `lines(readable: ReadableStream<Uint8Array>|ReadableStream<string>): AsyncIterable<string>`

```ts
import { lines } from "@std/streams";

const file = await Deno.open(filename);
for await (const line of lines(file.readable)) {
  console.log(line);
}
```

BlackAsLight:
Closes  #5071

Signature: `toLines(readable: ReadableStream<Uint8Array>): ReadableStream<string>`

## Usage
```ts
import { toLines } from '@std/streams'

for await (const line of toLines((await Deno.open('./deno.json')).readable))
  console.log('-', line)
```


streams/to_lines.ts
```diff
@@ -0,0 +1,76 @@
...
+ * }
+ *
+ * assertEquals(
+ *   people,
+ *   [{ "name": "Alice", "age": 30 }, { "name": "Bob", "age": 25 }],
+ * );
+ * ```
+ */
+export async function* toLines(
+  readable: ReadableStream<Uint8Array> | ReadableStream<string>,
```
iuioiua, line 36: I think it should just handle `ReadableStream<Uint8Array>`. One could easily just use `readable.pipeThrough(new TextLineStream()` for `ReadableStream<string>`.

streams/to_lines.ts
```diff
@@ -0,0 +1,76 @@
...
+console.log(
+  await Array.fromAsync(toLines(ReadableStream.from([
+    "qwertzu",
+    "iopasd\r\nmnbvc",
+    "xylk\rjhgfds\napoiuzt\r",
+    "qwr\r09ei\rqwrjiowqr\r",
+    "\nrewq0987\n\n654321",
+    "\nrewq0987\r\n\r\n654321\r",
+  ]))),
+);
```
iuioiua, line 76:
```suggestion
```

streams/to_lines.ts
```diff
@@ -0,0 +1,76 @@
...
+          ++index;
+        }
+      }
+    }
+    if (buffer.length) {
+      yield buffer;
+    }
+  } catch (reason) {
+    await reader.cancel(reason);
+  }
```
iuioiua, line 64: If we proceed with the above suggestion, we can just do this, using [`TextLineStream`](https://jsr.io/@std/streams/doc/text-line-stream/~/TextLineStream).
```suggestion
): ReadableStream<string> {
  return readable
    .pipeThrough(new TextDecoderStream())
    .pipeThrough(new TextLineStream());
```

iuioiua (changes requested):
Thank you for your contribution! However, I'm still skeptical about whether this provides sufficient value to justify having it in the Standard Library (weak opinion).

jsejcksn:
Since this is effectively a functional `TransformStream` pipe operation, should it accept a pipe options argument (ref: [Deno](https://deno.land/api@v1.44.4?s=PipeOptions), [MDN](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeThrough#options))? Refer to https://github.com/denoland/deno_std/issues/5071#issuecomment-2187833043 for additional context.

streams/to_lines.ts
```diff
@@ -0,0 +1,76 @@
...
+          buffer = buffer.slice(index + 1);
+          index = 0;
+        } else {
+          ++index;
+        }
+      }
+    }
+    if (buffer.length) {
+      yield buffer;
+    }
```
babiabeo, line 61:
```suggestion
    }
    reader.releaseLock();
```

I think we should release lock on the stream so it can be reused

babiabeo (changes requested):

streams/to_lines.ts
BlackAsLight: At that point in time the stream will be empty, so how can it be reused?

babiabeo: I see. You are right 👍

streams/to_lines.ts
BlackAsLight: The original issue suggested it should accept both types of streams, and I think it accepting both would give it more reason to exist

iuioiua: The Standard Library already has `TextLineStream` for `ReadableStream<string>`.

iuioiua (approved):
I made a few tweaks. Now, LGTM.

PR #5122, opened 2024-06-24
fix(testing): function call of `Date` constructor is not correctly faked

kt3k:
`Date()` (call without `new`) isn't faked correctly with the current implementation. This PR fixes it.


iuioiua (approved):
Nice catch!

PR #5123, opened 2024-06-24
test(testing): improve `FakeTime` testing

kt3k:
This PR improves test line coverage of `testing/time.ts` from 85.22% to 93.08% (The missed lines are reduced from 57 to 22)

part of #3713


testing/time_test.ts
```diff
@@ -632,3 +634,88 @@ Deno.test("Date from FakeTime is structured cloneable", () => {
...
+
+Deno.test("FakeTime can be called multiple times (later call is effective)", () => {
+  const now = Date.now();
+  try {
+    const _time1 = new FakeTime(1000);
+    const _time2 = new FakeTime(2000);
+    assertEquals(Date.now(), 2000);
+  } finally {
+    FakeTime.restore();
+  }
```
iuioiua, line 673: This should work once #5125 is merged.
```suggestion
  {
    using _time1 = new FakeTime(1000);
    using _time2 = new FakeTime(2000);
    assertEquals(Date.now(), 2000);
  }
```

testing/time_test.ts
```diff
@@ -632,3 +634,88 @@ Deno.test("Date from FakeTime is structured cloneable", () => {
...
+  try {
+    const _time1 = new FakeTime(1000);
+    assertEquals(Date.now(), 1000);
+    const _time2 = new FakeTime(new Date(2000));
+    assertEquals(Date.now(), 2000);
+    const _time3 = new FakeTime("Thu Jan 01 1970 00:00:03 GMT+0000");
+    assertEquals(Date.now(), 3000);
+  } finally {
+    FakeTime.restore();
+  }
```
iuioiua, line 687: Ditto. We can use `using` once #5125 is merged.

testing/time_test.ts
```diff
@@ -632,3 +634,88 @@ Deno.test("Date from FakeTime is structured cloneable", () => {
...
+  const _time = new FakeTime();
+  FakeTime.restore();
+  assertThrows(() => FakeTime.restore(), TimeError, "time already restored");
+});
+
+Deno.test("time.restore() throws when the time is already restored", () => {
+  const time = new FakeTime();
+  time.restore();
+  assertThrows(() => time.restore(), TimeError, "time already restored");
+});
```
iuioiua, line 704: This won't be needed once #5125 is merged.

testing/time_test.ts
```diff
@@ -632,3 +634,88 @@ Deno.test("Date from FakeTime is structured cloneable", () => {
...
+
+Deno.test("time.now = N throws when N < time.now", () => {
+  using time = new FakeTime(1000);
+  assertThrows(
+    () => {
+      time.now = 999;
+    },
+    Error,
```
iuioiua, line 712: Not for this PR, but we should throw `RangeError`.

iuioiua (changes requested):
Great work! Some nits. This also makes me realize we can improve the error types and messages thrown in `@std/testing/time`. However, that's best handled in a follow-up PR.

testing/time_test.ts
kt3k: This case has been removed

testing/time_test.ts
kt3k: I think this case now makes sense as now we prohibit the double creations of FakeTime #5130

iuioiua (approved):

Issue #5117, opened 2024-06-24
proposal: do not expose implementation details of yaml schemas


kt3k:
Currently YAML package exposes implementation details of schemas as public APIs such as `Schema` class or `Type` class. These seem too much details to be provided as public API. I propose we should make these APIs private.

iuioiua:
Can you please clarify which symbols you'd like to make private and which you'd like to keep public?

kt3k:
Keep public
- stringify
- parse
- parseAll
- StyleVariant
- DumpOption
- ParseOption

Make private
- CORE_SCHEMA
- DEFAULT_SCHEMA
- EXTENDED_SCHEMA
- FAILSAFE_SCHEMA
- JSON_SCHEMA
- KindType
- RepresentFn
- Type

iuioiua:
Nice! Related thoughts:
1. Is it worth keeping [`parseAll()`](https://jsr.io/@std/yaml/doc/~/parseAll)? Is there something special about YAML that justifies having the ability to parse multiple YAML documents within a single body of text? Neither TOML or JSON(C) have this ability. If the answer to these questions is "no", I suggest removing `parseAll()`.
2. Is [`DEFAULT_SCHEMA`](https://jsr.io/@std/yaml/doc/~/DEFAULT_SCHEMA) the correct default schema? According to the [YAML spec](https://yaml.org/spec/1.2.2/#103-core-schema), [`CORE_SCHEMA`](https://jsr.io/@std/yaml/doc/~/CORE_SCHEMA) should be the default schema.
WDYT?

kt3k:
> Is there something special about YAML that justifies having the ability to parse multiple YAML documents within a single body of text?

YAML spec defines the special syntax for storing multiple documents in single file. `parseAll` supports parsing it. ref https://yaml.org/spec/1.2.0/#id2600795

> Is [DEFAULT_SCHEMA](https://jsr.io/@std/yaml/doc/~/DEFAULT_SCHEMA) the correct default schema? According to the [YAML spec](https://yaml.org/spec/1.2.2/#103-core-schema), [CORE_SCHEMA](https://jsr.io/@std/yaml/doc/~/CORE_SCHEMA) should be the default schema.

The default schema is extended from core schema. It seems adding `binary`, `omap`, `pairs`, `set`, `timestamp`, and `merge` types.

[...]

Issue #4908, opened 2024-05-30
stop doing deprecation process for 0.x version packages


kt3k:
Thanks to the migration to JSR we now follow semver semantics about breaking changes (In 0.x versions we follow the rule suggested in https://github.com/semver/semver/pull/923 (something like `O.MAJOR.MINOR`)).

In previous `/std/` there's no way to expect when the breaking changes happened, and that was why we had to do deprecation process to not break the users program unexpectedly. Now because of the semver semantics, breaking changes can be predicted easily by the version number difference. I think we don't need to do deprecation process for 0.x versions anymore.

iuioiua:
Is anything else needed here?

kt3k:
Let's close

PR #5127, opened 2024-06-25
docs(yaml): improve `yaml` document

kt3k:
part of #3764 

This PR documents `parse` and `stringify` of `yaml` package. I skipped documenting other items (such as `Schema`, `Type`, etc) as they are planned to be excluded from public APIs #5117

### DIFF
--- a/yaml/parse.ts
+++ b/yaml/parse.ts
@@ -31,10 +31,10 @@ export interface ParseOptions {
 }
 
 /**
- * Parses `content` as single YAML document.
+ * Parse `content` as single YAML document, and return it.
  *
- * Returns a JavaScript object or throws `YAMLError` on error.
- * By default, does not support regexps, functions and undefined. This method is safe for untrusted data.
+ * This function does not support regexps, functions, and undefined by default.
+ * This method is safe for parsing untrusted data.
  *
  * @example Usage
  * ```ts
@@ -49,6 +49,7 @@ export interface ParseOptions {
  * assertEquals(data, { id: 1, name: "Alice" });
  * ```
  *
+ * @throws {YAMLError} Throws error on invalid YAML.
  * @param content YAML string to parse.
  * @param options Parsing options.
  * @returns Parsed document.
--- a/yaml/parse.ts
+++ b/yaml/parse.ts
@@ -4,8 +4,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import type { Schema } from "./schema.ts";
-import { type CbFunction, load, loadAll } from "./_loader/loader.ts";
+import { load, loadAll } from "./_loader/loader.ts";
 import { replaceSchemaNameWithSchemaClass } from "./mod.ts";
 
 /**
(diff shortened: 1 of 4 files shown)


yaml/parse.ts
```diff
@@ -23,8 +22,10 @@ export interface ParseOptions {
    * Specifies a schema to use.
    *
    * Schema class or its name.
+   *
+   * Passing Schema class is deprecated. Use schema name instead.
    */
-  schema?: string | Schema;
```
kt3k, line 27: I changed `Schema` to `unknown` here for now. This is required to pass the doc linter check. If we keep `Schema` here, we need to document all properties of `Schema` and `Type` to pass the doc linter check, which involves huge amount of work. However we plan to hide these from public API soon #5117

yaml/parse.ts
```diff
@@ -36,6 +37,23 @@ export interface ParseOptions {
  *
  * Returns a JavaScript object or throws `YAMLError` on error.
```
iuioiua, line 38: Are you able to replace this line with a [`@throws`](https://jsdoc.app/tags-throws) tag?

yaml/parse.ts
```diff
@@ -61,20 +80,54 @@ export function parse(content: string, options?: ParseOptions): unknown {
...
  * id: 3
  * name: Eve
- * `);
- * console.log(data);
+ * `, (doc: any) => {
+ *   assertEquals(typeof doc, "object");
+ *   assertEquals(typeof doc.id, "number");
+ *   assertEquals(typeof doc.name, "string");
+ * });
  * // => [ { id: 1, name: "Alice" }, { id: 2, name: "Bob" }, { id: 3, name: "Eve" } ]
```
iuioiua, line 88:
```suggestion
 * assertEquals(data, [{ id: 1, name: "Alice" }, { id: 2, name: "Bob" }, { id: 3, name: "Eve" }]);
```

yaml/schema/mod.ts
```diff
@@ -19,7 +18,7 @@ export {
 };
 
 export function replaceSchemaNameWithSchemaClass(
-  options?: { schema?: string | Schema },
+  options?: { schema?: string | unknown },
```
iuioiua, line 21: Ditto elsewhere where this type is used.
```suggestion
  options?: { schema?: "core" | "default" | "failsafe" | "json" | "extended" | unknown },
```

yaml/parse.ts
```diff
@@ -23,8 +22,10 @@ export interface ParseOptions {
    * Specifies a schema to use.
    *
    * Schema class or its name.
+   *
+   * Passing Schema class is deprecated. Use schema name instead.
```
iuioiua, line 26: I thought we were no longer doing deprecations. Are we making an exception here? #4908

iuioiua (changes requested):

yaml/parse.ts
kt3k: Ok. Let's skip that

yaml/schema/mod.ts
iuioiua, line 21:
```suggestion
  options?: { schema?: "core" | "default" | "failsafe" | "json" | "extended" | unknown },
```

iuioiua (approved):

PR #5129, opened 2024-06-25
BREAKING(http): improve thrown errors in `cookie` module

iuioiua:
### What's changed

[`getCookies()`](https://jsr.io/@std/http/doc/cookie/~/getCookies) or [`setCookie()`](https://jsr.io/@std/http/doc/cookie/~/setCookie) now throw [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError)s if they encounter invalid characters for their respective implementations. Previously, they threw a mix of [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error)s and [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError)s.

### Why this change was made

This change was made because `SyntaxError` is typically used in JavaScript to represent the encountering of invalid characters.

### Migration guide

When error-handling either of these functions, compare against `SyntaxError`, rather than `Error` or `TypeError`.

```diff
import { setCookie } from "@std/http/cookie";

const headers = new Headers();

try {
  setCookie(headers, {
    name: "location",
    value: "United Kingdom",
  });
} catch (error) {
- if (error instanceof Error) {
+ if (error instanceof SyntaxError) {
    // ...
  }
}
```

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -137,7 +137,7 @@ function toString(cookie: Cookie): string {
  */
 function validateName(name: string | undefined | null) {
   if (name && !FIELD_CONTENT_REGEXP.test(name)) {
-    throw new TypeError(`Invalid cookie name: "${name}".`);
+    throw new SyntaxError(`Invalid cookie name: "${name}".`);
   }
 }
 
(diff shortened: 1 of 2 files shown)


kt3k:
Should we tag this as `BREAKING`?

kt3k:
Note: Currently `ini` and `jsonc` uses `SyntaxError` for parse errors. I think probably it's fine to use `SyntaxError` for these cases.

kt3k (approved):
LGTM

PR #5133, opened 2024-06-25
chore: ignore test utils in coverage reporting

kt3k:
This PR unifies the naming convention of test utility files to `_test_utils.ts` and ignore them from coverage reporting.


iuioiua (approved):
Very nice

PR #5135, opened 2024-06-25
refactor(http): change error thrown in `ServerSentEventStream` to `SyntaxError`

iuioiua:
### What's changed

This PR changes the error thrown in [`ServerSentEventStream`](https://jsr.io/@std/http/doc/server-sent-event-stream/~/ServerSentEventStream) from [`RangeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError) to [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) if the `comment`, `event` or `id` properties of the chunk contained a newline.

### Why this change was made

This change was made as `SyntaxError` is more appropriate for invalid characters, than `RangeError`.

### Migration guide

To migrate, when error-handling, compare against `SyntaxError` instead of `RangeError`.
```diff
import { ServerSentEventStream } from "@std/http/server-sent-event-stream";

const stream = ReadableStream
  .from<ServerSentEventMessage>([{ event: "a\nb" }])
  .pipeThrough(new ServerSentEventStream());

try {
  await Array.fromAsync(stream);
} catch (error) {
- if (error instanceof RangeError) {
+ if (error instanceof SyntaxError) {
    // ...
  }
}
```

### DIFF
--- a/http/server_sent_event_stream.ts
+++ b/http/server_sent_event_stream.ts
@@ -24,7 +24,7 @@ export interface ServerSentEventMessage {
 
 function assertHasNoNewline(value: string, varName: string) {
   if (value.match(NEWLINE_REGEXP) !== null) {
-    throw new RangeError(`${varName} cannot contain a newline`);
+    throw new SyntaxError(`${varName} cannot contain a newline`);
   }
 }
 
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #5136, opened 2024-06-25
test(testing): improve bdd testing

kt3k:
This PR improves test coverage of `testing/bdd.ts`.

part of #3713

### DIFF
--- a/testing/bdd_test.ts
+++ b/testing/bdd_test.ts
@@ -3,11 +3,15 @@ import {
   assert,
   assertEquals,
   assertObjectMatch,
+  assertRejects,
   assertStrictEquals,
+  assertThrows,
 } from "@std/assert";
 import {
+  after,
   afterAll,
   afterEach,
+  before,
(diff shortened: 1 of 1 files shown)


testing/bdd_test.ts
```diff
@@ -543,6 +547,14 @@ Deno.test("it()", async (t) => {
           }),
       );
 
+      await t.step(
+        "minimum options (skip)",
+        async () =>
+          await assertMinimumOptions((fn) => {
+            assertEquals(it.skip({ name: "example", fn }), undefined);
```
iuioiua, line 554: I don't understand using `assertEquals()` on methods that always return `void`. Wouldn't it be fine to just call `it.skip({ name: "example", fn })`? Ditto for the other changes in this file.

kt3k: Probably it's fine to remove them from here as they are already tested in other places.

iuioiua (approved):

PR #5137, opened 2024-06-25
refactor(testing): remove dead code and improve test of `testing/mock.ts`

kt3k:
This PR improves testing of `testing/mock.ts`, and also removes some dead code in it.

part of #3713

### DIFF
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -1413,11 +1413,6 @@ export async function assertSpyCallAsync<
           "do not expect error and return, only one should be expected",
         );
       }
-      if (call.error) {
-        throw new AssertionError(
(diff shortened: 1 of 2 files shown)


testing/mock.ts
```diff
@@ -1413,11 +1413,6 @@ export async function assertSpyCallAsync<
           "do not expect error and return, only one should be expected",
         );
       }
-      if (call.error) {
-        throw new AssertionError(
-          "spy call did not return expected value, an error was thrown.",
-        );
-      }
```
kt3k, line 1420: When `call.error` is truthy, `assertSpyCallAsync` throws earlier before reaching here.

PR #5139, opened 2024-06-25
fix(testing): correctly throw in constructor with `spy()`

kt3k:
Currently thrown error from constructor is not captured by `spy`. This PR enables spying thrown errors from constructor.


testing/mock.ts
iuioiua, line 1: The added test seems to pass without these changes.

kt3k: Thanks for catching it! I forgot to add another assertion.

iuioiua (approved):
Nice

PR #5144, opened 2024-06-26
BREAKING(http): rename `Entity` to `ETagSource` and `calculate()` to `eTag()`

### LINKED ISSUE #5140: proposal(http): rename `calculate` to `eTag`
`calculate` sounds like an overly generic name, and it can be confusing in `@std/http` documentation. (It's fine in `@std/http/etag` documentation though).

I suggest we should rename it to `eTag`.

### LINKED ISSUE #5141: proposal(http): rename `Entity` to `ETagSource`
Currently the input type of `calculate` function of `@std/http/etag` module is named `Entity`, but it sounds too generic, and it can be confusing in `@std/http` docs. I suggest we should rename this to `ETagSource` which makes the purpose of this type clearer.

iuioiua:
### What's changed

The `Entity` interface has been renamed `ETagSource`, and `calculate()` has been renamed `eTag()`.

### Why these changes were made

These changes were made to better describe the purposes of these symbols. Previously, their names were too generic for their purposes to be immediately clear.

### Migration guide

To migrate, use the `ETagSource` interface instead of `Entity` and `eTag()` instead of `calculate()`.

```diff
- import { calculate, type Entity } from "@std/http/etag";
+ import { eTag, type ETagSource } from "@std/http/etag";

- const body: Entity = "hello deno!";
+ const body: ETagSource = "hello deno!";

- const etag = await calculate(body);
+ const etag = await eTag(body);

  const res = new Response(body, { headers: { etag } });
```

### Related

Closes #5141
Closes #5140


timreichen:
Web APIs have the convention of using a verb first for functions. `eTag()` breaks that convention. Maybe it would be better to call it `calculateETag()` or something similar.

kt3k:
We have many examples of functions which have the noun as its name:
- https://jsr.io/@std/media-types/doc/~/contentType
- https://jsr.io/@std/media-types/doc/~/extension
- https://jsr.io/@std/cli/doc/~/unicodeWidth
- https://jsr.io/@std/collections/doc/~/maxOf
- https://jsr.io/@std/collections/doc/~/permutations
- https://jsr.io/@std/datetime/doc/~/difference

> Web APIs have the convention of using a verb first for functions. eTag() breaks that convention.

I don't think we have this convention in std.

timreichen:
Valid point, though I think just calling it `eTag()` doesn't provide any information on what the function actually does except it has to do something with etag. Maybe `createETag()` borrowed from `Deno.createKv()` or `toETagString()` borrowed from `toString()` would reflect better what the function does?

kt3k:
> though I think just calling it eTag() doesn't provide any information on what the function actually does except it has to do something with etag.

It feels common to name the function as its returned thing. The above std functions are the examples of such naming convention.

I can also find many Web APIs named in that way:
- https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID
- https://developer.mozilla.org/en-US/docs/Web/API/Response/json
- https://developer.mozilla.org/en-US/docs/Web/API/Response/text
- https://developer.mozilla.org/en-US/docs/Web/API/Response/blob
- Temporal.Now.instant()
- Temporal.Now.timeZoneId()

kt3k (approved):
LGTM

PR #5146, opened 2024-06-26
refactor(testing): cleanup use of `assertIsError()`

iuioiua:


kt3k:
This seems making the error message a bit unclear.

example
```ts
import { spy, assertSpyCalls } from "./testing/mock.ts";
const fn = spy()
assertSpyCalls(fn, 1);
```

BEFORE
```
error: Uncaught (in promise) AssertionError: spy not called as much as expected:


    [Diff] Actual / Expected


-   0
+   1

```

AFTER
```
error: Uncaught (in promise) AssertionError: Values are not equal: spy not called as much as expected


    [Diff] Actual / Expected


-   0
+   1

```

`Values are not equal:` part doesn't make much sense in `AFTER`

iuioiua:
I'll close this, but I think this highlights a flaw in our assertion functions: They require these hacks to present a legible error message. If these functions' `msg` arguments replaced the entire assertion message, we wouldn't have these problems.

kt3k:
`msg` arg used to replace the entire assertion message, but there were requests from the community that 'actual' and 'expected' values also should be shown together with `msg` #982 #2461, and we changed it to append `msg` in #3253

I personally don't think this is a typical usage of assertion functions.

iuioiua:
In this context, with `assertEquals()`, I mean replacing the entire message before the diff, not the diff itself.

PR #5148, opened 2024-06-26
BREAKING(yaml): remove `iterator` argument from `parseAll()`

### LINKED ISSUE #5128: suggestion: remove `iterator` argument from `parseAll()` in `@std/yaml`
I suggest removing the [`iterator`](https://jsr.io/@std/yaml/doc/parse/~/parseAll#function_parseAll_0_parameters_iterator) argument from [`parseAll()`](https://jsr.io/@std/yaml/doc/parse/~/parseAll).

It provides no added value to the function as it doesn't affect the return array and is only called on each YAML document (element) once all documents are already loaded onto memory. If a user really needs this functionality, they can use [`Array.prototype.forEach()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach).

https://github.com/denoland/deno_std/blob/a4d4e51cfad1fd79eae5c70bdb8b0d6f92a591d7/yaml/_loader/loader.ts#L1773-L1789

iuioiua:
Closes #5128

### DIFF
--- a/yaml/_loader/loader.ts
+++ b/yaml/_loader/loader.ts
@@ -1765,27 +1765,11 @@ function loadDocuments(input: string, options?: LoaderStateOptions): unknown[] {
   return state.documents;
 }
 
-export type CbFunction = (doc: unknown) => void;
-function isCbFunction(fn: unknown): fn is CbFunction {
-  return typeof fn === "function";
-}
-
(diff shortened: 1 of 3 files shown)


kt3k (approved):
There seems no special scenario supported only by the 1st signature. This signature seem just confusing. Also that usage wasn't in the docs until recently.

LGTM

PR #5149, opened 2024-06-26
refactor(yaml): rename `YAMLError` to `YamlError`

iuioiua:
This aligns the class name with our naming convention.


kt3k:
`YAMLError` doesn't seem exposed as a public API. Is this a breaking change? https://jsr.io/@std/yaml/doc

iuioiua:
Good catch! Changed to refactor. I'll make this public in another PR. It should be public for error-handling purposes.

kt3k (approved):
LGTM

PR #5151, opened 2024-06-26
refactor(csv): remove dead code and improve tests

kt3k:
part of #3713 

This improves the test coverage of `csv/parse.ts`.

### DIFF
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -50,35 +50,29 @@ class Parser {
   #readLine(): string | null {
     if (this.#isEOF()) return null;
 
-    if (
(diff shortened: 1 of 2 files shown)


csv/parse.ts
```diff
@@ -50,35 +50,29 @@ class Parser {
   #readLine(): string | null {
     if (this.#isEOF()) return null;
 
-    if (
-      !this.#input.startsWith("\r\n", this.#cursor) ||
-      !this.#input.startsWith("\n", this.#cursor)
```
kt3k, line 55: This condition can't be `false` as the same string can't start with both `\r\n` and `\n`.

iuioiua (approved):
Nice

PR #5158, opened 2024-06-27
refactor(csv): minor cleanups

iuioiua:

PR #5165, opened 2024-06-27
refactor(streams): throw in `zipReadableStreams()` naturally

iuioiua:
This has the same effect as the previous code. Just a little tidier.


iuioiua:
I realise this is likely wrong due to the errors that #5158 caused.

PR #5171, opened 2024-06-27
BREAKING(yaml): rename `DumpOptions` to `StringifyOptions`

kt3k:
`DumpOptions` is the options for `stringify`. This should be named `StringifyOptions`

# What's changed

`DumpOptions` has been renamed to `StringifyOptions`

# Motivation

The current name isn't aligned to the naming convention of the option types because it's options for `stringify`. There's no public API called `dump`.

# Migration

```diff
-import type { DumpOptions } from "@std/yaml`;
+import type { StringifyOptions as DumpOptions } from "@std/yaml`;
```

### DIFF
--- a/yaml/stringify.ts
+++ b/yaml/stringify.ts
@@ -10,7 +10,7 @@ import { replaceSchemaNameWithSchemaClass } from "./mod.ts";
 /**
  * The option for strinigfy.
  */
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Nice!

PR #5173, opened 2024-06-27
BREAKING(json): remove `json` from module names

### LINKED ISSUE #5177: proposal(json,csv): avoid repetition in entrypoint name if it has the same word as the pacakge name in the beginning
In addition to #5162, the core team suggests the below entrypoint name changes to make the rule more consistent across std.

- `std/csv/csv-parse-stream` -> `std/csv/parse-stream`
- `std/csv/csv-stringify-stream` -> `std/csv/stringify-stream`
- `std/json/json-parse-stream` -> `std/json/parse-stream`
- `std/json/json-stringify-stream` -> `std/json/stringify-stream`

What do you think?

iuioiua:
### What's changed

`json` has been removed from module names in `@std/json`. In other words:
1. `@std/json/json-parse-stream` has moved to `@std/json/parse-stream`.
2. `@std/json/json-stringify-stream` has moved to `@std/json/stringify-stream`.

### Motivations

These changes were made to shorten these module specifiers while still retaining their context; that they're JSON-oriented streams.

### Migration guide

To migrate:
1. Import from `@std/json/parse-stream` instead of `@std/json/json-parse-stream`.

   ```diff
   - import { JsonParseStream } from "@std/json/json-parse-stream";
   + import { JsonParseStream } from "@std/json/parse-stream";

   // ...
   ```
2. Import from `@std/json/stringify-stream` instead of `@std/json/json-stringify-stream`.

   ```diff
   - import { JsonStringifyStream } from "@std/json/json-stringify-stream";
   + import { JsonStringifyStream } from "@std/json/stringify-stream";

   // ...
   ```

---

Closes #5177.


json/deno.json
```diff
@@ -4,8 +4,8 @@
   "exports": {
     ".": "./mod.ts",
     "./types": "./types.ts",
-    "./concatenated-json-parse-stream": "./concatenated_json_parse_stream.ts",
```
kt3k, line 7: Should we do this as well?

iuioiua: Yes, I think so.

kt3k: If we do this rename, I feel the rule becomes a bit obscure.

The rule here might be something like: `if the package name is included anywhere in module export name, then that part is removed from the export name`.

In that case the below items also includes package names:
- `@std/semver/is-semver`
- `@std/async/mux-async-iterator`

But I think we probably don't want to rename the above items.

iuioiua: I'm mostly impartial, as I don't think such small details make much of a difference, either way. As long as the module name achieves its objective of signaling its purpose through its name, I think it's good. Happy to defer to you or hear other opinions.

iuioiua: From some internal chats and a [small Discord survey](https://discord.com/channels/684898665143206084/775393009981849600/1257227741879865406), it seems people would rather keep `@std/json/concatenated-json-parse-stream`, so I reverted that change.

kt3k (approved):
LGTM

PR #5175, opened 2024-06-27
BREAKING(yaml): make `Schema` and `Type` private and stop supporting user-defined schemas

### LINKED ISSUE #5117: proposal: do not expose implementation details of yaml schemas

kt3k:
### What's changed

`Type` and `Schema` classes have been made private, and user-defined schemas are no longer supported.

### Motivation

`Type` and `Schema` expose too many implementation details to be part of the public API.

### Migration

If you use one of the pre-defined schemas, you can specify them by its name.

```diff
- import { EXTENDED_SCHEMA, parse } from "@std/yaml";
+ import { parse } from "@std/yaml";

- parse(yamlText, { schema: EXTENDED_SCHEMA });
+ parse(yamlText, { schema: "extended" });
```

If you require a user-defined schema, use [`npm:js-yaml` ](https://www.npmjs.com/package/js-yaml), which is what `@std/yaml` was initially based on. It still supports user-defined schemas.

---
Closes #5117


iuioiua (approved):
This significantly simplifies the `@std/yaml` package - great work!

PR #5176, opened 2024-06-27
BREAKING(assert): remove `assert` from module names

### LINKED ISSUE #5162: proposal(assert): change the entrypoint pattern from "assert/assert-foo" to "assert/foo"
The core team proposes the change to the entrypoint pattern in `@std/assert` like the below:

```diff
-import { assertEquals } from "@std/assert/assert-equals";
+import { assertEquals } from "@std/assert/equals";
```

iuioiua:
Closes #5162


assert/deno.json
```diff
@@ -25,7 +25,7 @@
     "./assert-strict-equals": "./assert_strict_equals.ts",
     "./assert-string-includes": "./assert_string_includes.ts",
     "./assert-throws": "./assert_throws.ts",
-    "./assertion-error": "./assertion_error.ts",
```
kt3k, line 28: Should `assertion-error` also be moved to `/error`? I guessed it would keep the current location

kt3k:
I think this change can't be done in a single PR. I think the relative import of `./assert/*` from the outside would break `deno publish --dry-run`

iuioiua:
This is not yet complete. I should be doable in a single PR.

kt3k:
If the commit tag of this PR is `BREAKING(assert)`, then only `assert` package is upgraded by this. But all depending packages need to have `refactor(package)` commit to follow `assert` change. We need those 2 different kinds of grades, but they can't be expressed in a single PR

assert/deno.json
iuioiua: Having an `@std/assert/assertion-error` among the other modules would look weird. I think this is fine.

iuioiua:
`@std/assert` imports in `@std/testing` and `@std/expect` will be updated to their most-specific module specifiers in a follow-up PR.

assert/mod.ts
```diff
@@ -18,29 +18,29 @@
...
+export * from "./not_strict_equals.ts";
+export * from "./object_match.ts";
+export * from "./rejects.ts";
+export * from "./strict_equals.ts";
+export * from "./string_includes.ts";
+export * from "./throws.ts";
 export * from "./assert.ts";
-export * from "./assertion_error.ts";
+export * from "./error.ts";
 export * from "./equal.ts";
```
iuioiua, line 44: What do we want to do with `@std/assert/equal`?

assert/deno.json
kt3k: This case doesn't exactly look like a repetition of the same word to me...

assert/mod.ts
kt3k: I think it's fine as is at least in this PR

assert/deno.json
kt3k: Can we discuss this case independently?

iuioiua: I've reverted the change.

kt3k (approved):
LGTM

PR #5179, opened 2024-06-28
BREAKING(semver): remove `canParse()`

iuioiua:
### What's changed

`canParse()` has been removed from [`@std/semver`](https://jsr.io/@std/semver) in favor of [`tryParse()`](https://jsr.io/@std/semver/doc/~/tryParse).

### Motivation

This change was made because `tryParse()` provides extended functionality over `canParse()`, as it allows the user to determine whether a string is able to be parsed into a SemVer version, but returns that SemVer object too.

### Migration

To migrate, use `tryParse()` instead of `canParse()`, and compare the return value to `null`.

```diff
- import { canParse, parse } from "@std/semver/can-parse";
+ import { tryParse } from "@std/semver/try-parse";

  const str = "1.2.3";
- if (canParse(str)) {
- const semver = parse(str);
+ const semver = tryParse(str);
+ if (semver !== null) {
    console.log(`The SemVer is {semver}`);
  } else {
    console.log("The string is not a valid SemVer");
  }
```

---

CC @timreichen


timreichen:
I think it would be better the other way around. `canParse()` is borrowed from web APIs like [URL.prototype.canParsw](https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static) while `tryParse()` was rejected as a web API (ref https://github.com/denoland/deno_std/pull/4049)

iuioiua:
The motivation for this PR (see initial comment), I think, is a good counterargument for doing that. This is what Yoshiya previously explained. I didn't agree with it then, but I do now.

kt3k:
now URL has `URL.parse` which returns `URL | null` https://github.com/whatwg/url/pull/825 that is mostly aligned to `semver.tryParse`.

kt3k:
Some deno projects ([deployctl](https://github.com/denoland/deployctl), [publish-on-tag](https://jsr.io/@david/publish-on-tag)) seem using `canParse`.

https://github.com/search?q=semver+canParse+language%3ATypeScript&type=code&l=TypeScript

Do we need to remove this?

iuioiua:
My opinion on this isn't strong. Existing use is a sufficient reason to abandon the idea. Fine with me. Closing.

PR #5183, opened 2024-06-28
refactor(yaml): simplify map initializations

kt3k:
This PR makes `tagMap` and `anchorMap` properties non-nullable. These 2 properties are assigned with `Object.create(null)` nearly the beginning of their usages, but not initialized in the constructor. That causes them nullable types, and makes the logic unnecessarily complex. This PR removes those complexities.


iuioiua (approved):
Nice catch

PR #5185, opened 2024-06-28
fix(yaml): fix merge (<<) type handling in `parse()`

kt3k:
[merge type](https://yaml.org/type/merge) (`<<`) handling seems broken (ref #4457). This PR fixes it and add test cases to check it.


iuioiua (approved):
Good catch

PR #5192, opened 2024-06-28
fix(yaml): fix `!!pairs` parsing

kt3k:
This PR fixes the handling of empty pairs, and also adds the test cases for parsing various `!!pairs` patterns.


yaml/_type/pairs.ts
```diff
@@ -9,6 +9,8 @@ import type { Any } from "../_utils.ts";
 const _toString = Object.prototype.toString;
 
 function resolveYamlPairs(data: Any[][]): boolean {
+  if (data === null) return true;
```
kt3k, line 12: This check exists in the original (js-yaml). ref: https://github.com/nodeca/js-yaml/blob/0d3ca7a27b03a6c974790a30a89e456007d62976/lib/type/pairs.js#L8

iuioiua (approved):

Issue #2348, opened 2022-06-15
proposal(media_types): Improve contentType type definition


ayame113:
The return type of the `contentType()` of the media_types module is `string | undefined`.
Therefore, if I set the return value of the `contentType()` to the Content-Type of the Response object as it is, a type error will occur. (`Type '{ "Content-Type": string | undefined; }' is not assignable to type 'HeadersInit | undefined'.`)

```ts
new Response("<b>hello</b>", {
  headers: { // Type '{ "Content-Type": string | undefined; }' is not assignable to type 'HeadersInit | undefined'.
    "Content-Type": contentType(".html"),
  },
});
```

However, for well-known extensions like `contentType(".html")`, I think that the return value can be narrowed down to string by improving the type definition. (And it will be possible to avoid the above type error.)

```ts
type ExtensionOrType = ...;

export function contentType<T extends string>(
  extensionOrType: T,
): Lowercase<T> extends ExtensionOrType ? string : string | undefined {
...
```

If an unknown extension is passed, the return type will continue to be `string | undefined`.

**Describe alternatives you've considered**

The problem described above can be easily avoided with a Non-null assertion(`!`). So it may not be worth introducing this.

```ts
new Response("<b>hello</b>", {
  headers: {
    "Content-Type": contentType(".html")!,
  },
});
```

ayame113:
I opened #2357.

PR #5193, opened 2024-06-29
fix(media-types): update `db.ts`

### LINKED ISSUE #5186: media_types: incorrect mime for javascript - does not match mime-db
The documentation for [media_type](https://deno.land/std@0.224.0/media_types/vendor/mime-db.v1.52.0.ts?source) claims the mime database comes from mime-db v1.52, yet in deno ".mjs" and ".js" map to "application/javascript" where as in the actual mime-db v1.52 [db.json](https://github.com/jshttp/mime-db/blob/master/db.json) it appears that they map to the correct "text/javascript".

iuioiua:
This PR creates a new script, `media_types/vendor/update.ts`, that updates `media_types/vendor/db.json` directly from the [`mime-db`](https://github.com/jshttp/mime-db) repo.

Closes #5186


deno.json
```diff
@@ -36,7 +36,7 @@
     "@std/json": "jsr:@std/json@^1.0.0-rc.1",
     "@std/jsonc": "jsr:@std/jsonc@^1.0.0-rc.1",
     "@std/log": "jsr:@std/log@^0.224.3",
-    "@std/media-types": "jsr:@std/media-types@^1.0.0-rc.1",
```
iuioiua, line 39: Notice: This was previously incorrect. I've raised this issue in #5194 to ensure this doesn't happen again.

kt3k: Can you create another PR for fixing this?

media_types/content_type_test.ts
```diff
@@ -33,13 +33,9 @@ Deno.test({
     _str = contentType("text/html");
     _str = contentType("txt");
 
-    // @ts-expect-error: string | undefined is not assignable to string
```
kt3k, line 36: These type errors are expected. See #2348 and #2357 for details

iuioiua: Not anymore. Having them there now produces a `Unused '@ts-expect-error' directive.` error.

kt3k: Currently `contentType` returns `string` type if the input is well known type, and it returns `string | undefined` if the input is not well known type (`contentType("foo")` actually returns `undefined`. No type error here means it's typed incorrectly).

These changes of the type expectations means this function loses that feature.

iuioiua: Type inference seems to have broken because we're using a `.json` file instead of a `.ts` file. I'll fix it.

media_types/vendor/update.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+const response = await fetch(
+  "https://raw.githubusercontent.com/jshttp/mime-db/master/db.json",
+);
+
+const db = await response.text();
+const result =
+  `// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+// deno-fmt-ignore-file
```
kt3k, line 10: Can we run `deno fmt` to this file instead of ignoring?

kt3k (approved):
LGTM

PR #5196, opened 2024-06-30
test(archive): fix typo in `tar_test.ts`

familyboat:
According to this line https://github.com/denoland/deno_std/blob/9c543c9f36deb6b96bfe389034783400407f04e8/archive/tar_test.ts#L74, the byte length should be 163.


iuioiua (approved):
Nice catch. Thank you.

PR #5199, opened 2024-06-30
refactor(assert,async,bytes,cli,collections,crypto,csv,data-structures,datetime,dotenv,encoding,expect,fmt,front-matter,fs,html,http,ini,internal,io,json,jsonc,log,media-types,msgpack,net,path,semver,streams,testing,text,toml,ulid,url,uuid,webgpu,yaml): import from `@std/assert`

iuioiua:
Unblocks #5176 (will result in a cleaner diff). All these imports will be updated after #5176 is merged.


kt3k (approved):
LGTM

PR #5201, opened 2024-06-30
fix(testing): Throw error eagerly when insufficient permissions are granted to write to snapshot file in update mode

### LINKED ISSUE #5155: Deno test fails with inscrutable error when `assertSnapshot` is used in update mode but the `--allow-write` flag is missing
**Describe the bug**

Deno test fails with inscrutable error when `assertSnapshot` is used in update mode but the `--allow-write` flag is missing. This is hard to debug without knowledge of the inner workings of `assertSnapshot`.

**Steps to Reproduce**

1. In a new folder, create the file `main_test.ts` with the following contents:

```ts
// main_test.ts
import { assertSnapshot } from "@std/testing/snapshot";

Deno.test("should match the snapshot", async (context) => {
  await assertSnapshot(context, "hello world!");
});
```

2. Run `deno test --allow-read -- -u` and observe the output:

```sh
$ deno test --allow-read -- -u
running 1 test from ./main_test.ts
should match the snapshot ... ok (2ms)
Uncaught error from ./main_test.ts FAILED

 ERRORS 

./main_test.ts (uncaught error)
error: null
This error was not caught from a test and caused the test runner to fail on the referenced module.
It most likely originated from a dangling promise, event/timeout handler or top-level code.

 FAILURES 

./main_test.ts (uncaught error)

FAILED | 1 passed | 1 failed (5ms)

error: Test failed
```

**Expected behavior**

Ideally, a sensible error would be logged, such as:

```sh
$ deno test --allow-read -- -u
running 1 test from ./main_test.ts
should match the snapshot ... FAILED (0ms)

 ERRORS 

should match the snapshot => ./main_test.ts:4:6
error: AssertionError: The --allow-write flag must be passed when updating snapshots.
          ^
    at assert (https://jsr.io/@std/assert/1.0.0-rc.2/assert.ts:21:11)
    at ...

 FAILURES 

should match the snapshot => ./main_test.ts:4:6

FAILED | 0 passed | 1 failed (1ms)

error: Test failed
```

**Environment**

- OS: [e.g. Ubuntu 20.04, MacOS 11] MacOS Ventrura 13.6.6
- deno version:
```
deno 1.44.4 (release, aarch64-apple-darwin)
v8 12.6.228.9
typescript 5.4.5
```
- std version: 0.225.2

bcheidemann:
Fixes #5155

### DIFF
--- a/testing/__snapshots__/snapshot_test.ts.snap
+++ b/testing/__snapshots__/snapshot_test.ts.snap
@@ -470,3 +470,26 @@ snapshot[`createAssertSnapshot() - options object - custom name 1`] = `This gree
 snapshot[`createAssertSnapshot() > message 1`] = `"This snapshot has failed as expected"`;
 
 snapshot[`createAssertSnapshot() - composite - custom Name 1`] = `This green text has had it's colours stripped`;
+
+snapshot[`assertSnapshot() - regression #5155 1`] = `
+"running 1 test from <tempDir>/test.ts
+Snapshot Test ... FAILED (--ms)
+
+ ERRORS 
+
+Snapshot Test => ../../../..<tempDir>/test.ts:4:12
+error: PermissionDenied: Missing write access to snapshot file (file://<path>). This is required because assertSnapshot was called in update mode. Please pass the --allow-write flag.
+        throw new Deno.errors.PermissionDenied(
+              ^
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM. Thanks!

PR #5209, opened 2024-07-01
chore(text): release `text@1.0.0`

### LINKED ISSUE #4999: stabilize `@std/text`

iuioiua:
To be merged July 11.

The latest API docs: https://jsr.io/@std/text@1.0.0-rc.2

Closes #4999


magurotuna:
### closestString

https://github.com/denoland/deno_std/blob/22d3bda488f145b725fc1eaeee16922a97d88add/text/closest_string.ts#L27-L31

This note sounds concerning to me when it comes to stabilization. I'd suggest that we pin the algorithm that this function internally uses to calculate similarity, or make it configurable by an optional parameter and default to Levenshtein if none is specified by a user.

### compareSimilarity

https://github.com/denoland/deno_std/blob/22d3bda488f145b725fc1eaeee16922a97d88add/text/compare_similarity.ts#L41-L44

ditto

### wordSimilaritySOrt

https://github.com/denoland/deno_std/blob/22d3bda488f145b725fc1eaeee16922a97d88add/text/word_similarity_sort.ts#L48

ditto

### toCamelCase

This is defined in `case.ts` not `to_camel_case.ts`, so we can't import it with `import { toCamelCase } from "jsr:@std/text/to-camel-case"` but instead should do like `jsr:@std/text/case`. This would be surprising to users because usually one public function has one module that allows us to import only that function but `toCamelCase` doesn't follow this convention.

The same goes for `toKebabCase`, `toPascalCase`, and `toSnakeCase`

kt3k:
> This is defined in case.ts not to_camel_case.ts, so we can't import it with import { toCamelCase } from "jsr:@std/text/to-camel-case"

Maybe it's better to align to that rule in this case. @iuioiua Do you have any opinions?

BTW there are multiple places where we break that rule. The examples are:
- html/entities
- http/*
- testing/*
- fmt/*
- encoding/*
- uuid/*

kt3k:
> make it configurable by an optional parameter and default to Levenshtein if none is specified by a user.

This sounds reasonable to me. Thanks for the suggestion!

iuioiua:
All good suggestions. Happy to proceed with them.

kt3k (approved):
LGTM

magurotuna (approved):
LGTM 🚀

PR #5212, opened 2024-07-01
chore(cli): release `cli@1.0.0`

### LINKED ISSUE #5002: stabilize `@std/cli`
part of #4600 

- [x] fully documented
- [x] fully tested
- [ ] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

---
related issue #4362

iuioiua:
To be released July 18.

The latest API docs: https://jsr.io/@std/cli@1.0.0-rc.5/doc

Closes #5002


nathanwhit (approved):
LGTM, just one minor comment on the docs for `parseArgs`:

On [the two paragraphs starting here](https://jsr.io/@std/cli/1.0.0-rc.5/parse_args.ts#L438)

> By default, the flags module tries to determine the type of all arguments automatically and the return type of the parseArgs method will have an index signature with any as value ({ [x: string]: any }).
>
> If the string, boolean or collect option is set, the return value of the parseArgs method will be fully typed and the index signature of the return type will change to { [x: string]: unknown }.

It refers to the "flags" module, which is incorrect, and it should probably use "this function" instead of "the parseArgs method"

iuioiua:
> LGTM, just one minor comment on the docs for `parseArgs`:
> 
> On [the two paragraphs starting here](https://jsr.io/@std/cli/1.0.0-rc.5/parse_args.ts#L438)
> 
> > By default, the flags module tries to determine the type of all arguments automatically and the return type of the parseArgs method will have an index signature with any as value ({ [x: string]: any }).
> > If the string, boolean or collect option is set, the return value of the parseArgs method will be fully typed and the index signature of the return type will change to { [x: string]: unknown }.
> 
> It refers to the "flags" module, which is incorrect, and it should probably use "this function" instead of "the parseArgs method"

Done in #5458

kt3k (approved):
LGTM

PR #5217, opened 2024-07-01
chore(http): release `http@1.0.0`

### LINKED ISSUE #5007: stabilize `@std/http`
part of #4600 

- [x] fully documented
- [ ] fully tested
- [ ] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

iuioiua:
To be merged July 25.

The latest API docs: https://jsr.io/@std/http@1.0.0-rc.6/doc

Closes #5007


bartlomieju:
Do we expect to stabilize `parseSignedCookie` and `signCookie` before merging this one?

iuioiua:
No, we're keeping them unstable until they've lived around long enough and until #5257 is resolved.

bartlomieju (approved):

iuioiua:
@kt3k, please provide final review.

kt3k:
`eTag` return type can be `Promise<string>` for `string` and `Uint8Array` inputs. Not sure worth the breaking change though..

iuioiua:
Can you create PR so we can get a better idea?

iuioiua:
Good to merge now, @kt3k?

kt3k (approved):
LGTM

kt3k:
but I'll also suggest removing `ETagResource` symbol, as it's not referenced from public API anymore.

iuioiua:
SGTM

iuioiua:
I suggest renaming [`STATUS_CODE`](https://github.com/denoland/std/blob/577fd9a6e85a8b58d513a8db5aae81500409d7b0/http/status.ts#L34) to `STATUS_CODES` (with an "s" suffix) and [`STATUS_TEXT`](https://github.com/denoland/std/blob/577fd9a6e85a8b58d513a8db5aae81500409d7b0/http/status.ts#L169) to `STATUS_TEXTS` (also with an "s" suffix).

kt3k:
> I suggest renaming [`STATUS_CODE`](https://github.com/denoland/std/blob/577fd9a6e85a8b58d513a8db5aae81500409d7b0/http/status.ts#L34) to `STATUS_CODES` (with an "s" suffix) and [`STATUS_TEXT`](https://github.com/denoland/std/blob/577fd9a6e85a8b58d513a8db5aae81500409d7b0/http/status.ts#L169) to `STATUS_TEXTS` (also with an "s" suffix).

I'm not in favor of that change. Why do we need it?

iuioiua:
Well, the name doesn't make sense. The object is not a "status code". Rather, it's an object whose entries are status codes. Same for "status text". The only reason we'd keep it is to avoid disruption, but we'd be settling for something that's incorrect.

iuioiua:
For context, `node:http` has [`STATUS_CODES`](https://nodejs.org/api/http.html#httpstatus_codes) and [`METHODS`](https://nodejs.org/api/http.html#httpmethods), which are also objects. Changing to the wrong name was made in #3781 by me.

iuioiua:
That seems like an obscure reason for the name. Again, it's an object whose entries are status codes. It itself, is not actually a status code.

PR #5219, opened 2024-07-01
chore(csv): release `csv@1.0.0`

### LINKED ISSUE #5012: stabilize `@std/csv`
part of #4600 

- [x] fully documented
- [x] fully tested
- [ ] 1.0.0-rc.1
- [ ] wait 1 month
- [ ] 1.0.0 (stabilization)

iuioiua:
To be merged July 29.

The latest API doc: https://jsr.io/@std/csv@1.0.0-rc.5/doc

Closes #5012


kt3k (approved):
LGTM

magurotuna:
https://github.com/denoland/std/blob/bfa56fce2d847486df809ddfca3f38e6a6eb2069/csv/parse.ts#L309-L310

Is this comment correct? In particular, the second line mentions "auto/custom mapper for columns" but I didn't understand what this mapper means.

kt3k:
> Is this comment correct? In particular, the second line mentions "auto/custom mapper for columns" but I didn't understand what this mapper means.

Good point. The mapping of columns to fields is only available in the 2nd overload signature. I'll clarify that description.

magurotuna:
https://github.com/denoland/std/blob/89770fc76a98a20198162055bebe795283ede8f9/csv/parse_stream.ts#L105-L106

What about this comment on `CsvParseStraem`? Is this also worth clarification?

magurotuna:
https://github.com/denoland/std/blob/89770fc76a98a20198162055bebe795283ede8f9/csv/stringify_stream.ts#L21

`CsvStringifyStream` takes `columns: Array<string>` while `stringify` takes `columns: Column[]` where `Column` is much more powerful than a string array. https://jsr.io/@std/csv@1.0.0-rc.5/doc/stringify/~/Column

There are other differences between `CsvStringifyStream` and `stringify`, e.g. BOM support, all of which makes `CsvStringifyStream` have strictly less capability than `stringify`. I think it's better to make `CsvStringStream` support all the options that `stringify` supports before v1.0.

iuioiua:
> https://github.com/denoland/std/blob/89770fc76a98a20198162055bebe795283ede8f9/csv/parse_stream.ts#L105-L106
> 
> What about this comment on `CsvParseStraem`? Is this also worth clarification?

Would you be able to address that in #5605?

> https://github.com/denoland/std/blob/89770fc76a98a20198162055bebe795283ede8f9/csv/stringify_stream.ts#L21
> 
> `CsvStringifyStream` takes `columns: Array<string>` while `stringify` takes `columns: Column[]` where `Column` is much more powerful than a string array. https://jsr.io/@std/csv@1.0.0-rc.5/doc/stringify/~/Column
> 
> There are other differences between `CsvStringifyStream` and `stringify`, e.g. BOM support, all of which makes `CsvStringifyStream` have strictly less capability than `stringify`. I think it's better to make `CsvStringStream` support all the options that `stringify` supports before v1.0.

All good points. I agree with all the suggestions.

magurotuna:
A couple more things I noticed:

- `stringify` and `CsvStringifyStream` both use CRLF as line breaks, with no other option (i.e. LF) provided. For comparison, [Go's CSV writer provides an option whether to use CRLF or LF, which defaults to LF](https://github.com/golang/go/blob/8e1fdea8316d840fd07e9d6e026048e53290948b/src/encoding/csv/writer.go#L34). I'm not sure if we should change our default to LF, but I think at least we may want to give an option to users. (This could be done after 1.0 though)
- It feels asymmetric that the `stringify` family has the `bom` option whereas the `parse` family doesn't.

iuioiua:
> * `stringify` and `CsvStringifyStream` both use CRLF as line breaks, with no other option (i.e. LF) provided. For comparison, [Go's CSV writer provides an option whether to use CRLF or LF, which defaults to LF](https://github.com/golang/go/blob/8e1fdea8316d840fd07e9d6e026048e53290948b/src/encoding/csv/writer.go#L34). I'm not sure if we should change our default to LF, but I think at least we may want to give an option to users. (This could be done after 1.0 though)

Ah, good catch. Yes, I think it should use LF as default. Yeah, maybe we can explore adding an option in a later PR.

> * It feels asymmetric that the `stringify` family has the `bom` option whereas the `parse` family doesn't.

Yeah, we should add it to `parse()`.

We're making a few changes to `@std/csv`. I suggest we postpone stabilization of it, like what we're doing with `@std/ini`. What do you guys think?

magurotuna:
> We're making a few changes to @std/csv. I suggest we postpone stabilization of it, like what we're doing with @std/ini. What do you guys think?

Yes, I happened to bring up too many things that we'd probably like to fix before stabilization. It sounds like a good idea to put it off.

kt3k:
The issues here seem mostly about bugs or doc issues, not design issue. I'm still in favor of stabilizing `std/csv`

kt3k:
> stringify and CsvStringifyStream both use CRLF as line breaks, with no other option (i.e. LF) provided.

I think this is fine and sensible choice as our implementation is based on [RFC4180](https://www.rfc-editor.org/rfc/rfc4180.html) (as mentioned [here](https://github.com/denoland/std/blob/12d6097289bdd7fef031d0db3c8258e219f8f904/csv/mod.ts#L19)), which defines CRLF as the default (and probably only) line break.

kt3k:
Another reason we want to land this is that `std/csv` has relatively large number of usages. If we postpone it, it's has more negative effect/disappointment than the case of `std/ini` (`std/ini` is a very young and less used package)

iuioiua:
> I think this is fine and sensible choice as our implementation is based on [RFC4180](https://www.rfc-editor.org/rfc/rfc4180.html) (as mentioned [here](https://github.com/denoland/std/blob/12d6097289bdd7fef031d0db3c8258e219f8f904/csv/mod.ts#L19)), which defines CRLF as the default (and probably only) line break.

Fair reason to keep as-is.

> Another reason we want to land this is that `std/csv` has relatively large number of usages. If we postpone it, it's has more negative effect/disappointment than the case of `std/ini` (`std/ini` is a very young and less used package)

Ok. Happy to stabilize then.

magurotuna:
So to summarize what I have brought up, the following things are possible improvements to make the csv module more consistent and ergonomic:

- Add `bom` option to `parse` and `CsvParseStream`
- Add `headers` and `bom` option to `CsvStringifyStream`
- Add `useLf` or `useCrLf` (name TBD) to `stringify` and `CsvStringifyStream`
- Change the type of `columns` of `CsvStringifyStreamOptions` from `readonly string[]` to `readonly Column[]`

None of these will be a breaking change even if we do so after v1.0, by carefully choosing the default value (for instance, `headers` option in `CsvStringifyStream` needs to be `true` by default, since its current behavior is as if `headers` is true).

magurotuna (approved):
We can go now and we'll be able to address the remaining improvements later.

PR #5221, opened 2024-07-01
test(expect): improve `expect` test cases

kt3k:
This PR adds test cases of:
- `expect().resolves`
- `expect().rejects`
- Throwing cases of custom matchers.

This improves the code coverage of `expect` about 2%.

part of #3713

### DIFF
--- a/expect/test.ts
+++ b/expect/test.ts
@@ -1,12 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
 import { expect } from "./expect.ts";
-import {
-  assert,
-  AssertionError,
-  assertRejects,
-  assertThrows,
-} from "@std/assert";
+import { AssertionError, assertRejects, assertThrows } from "@std/assert";
 
 Deno.test("expect().resolves.toEqual()", async () => {
   await expect(Promise.resolve(42)).resolves.toEqual(42);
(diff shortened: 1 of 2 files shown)


expect/test.ts
```diff
@@ -0,0 +1,51 @@
...
+Deno.test("expect().resolves.toEqual()", async () => {
+  await expect(Promise.resolve(42)).resolves.toEqual(42);
+  await expect(Promise.resolve(42)).resolves.not.toEqual(0);
+
+  assertThrows(
+    () => expect(42).resolves.toEqual(42),
+    AssertionError,
+    "expected value must be Promiselike",
+  );
+  assertThrows(
```
iuioiua, line 15: Just checking that we want this to throw rather than reject, like Jest?

iuioiua (approved):

expect/test.ts
kt3k: Yes. I checked with the below jest snippets just in case. They throw instead of returning rejected promise.

```ts
describe("mod", () => {
  it("func", () => {
    try {
      expect(null).resolves.toEqual(null);
    } catch (e) {
      console.log("thrown", e);
    }
    try {
      expect(42).resolves.toEqual(42);
    } catch (e) {
      console.log("thrown", e);
    }
  });
});
```

PR #5223, opened 2024-07-01
BREAKING(yaml): remove `filename` option from `parse()` and `stringify()`

iuioiua:
### What's changed

The `ParseOptions.filename` and `StringifyOptions.filename` options have been removed.

### Motivation

These options were removed as, in most cases, their functionality is redundant - stating the filename in question, if defined. By default, this value is undefined.

### Migration guide

If you require this feature, use [`@std/yaml@0.224.3`](https://jsr.io/@std/yaml@0.224.3), as this is the last version containing this feature.


kt3k (approved):
Probably a redundant option to keep as our API only works with input strings, not input files. Let's remove.

LGTM

PR #5224, opened 2024-07-01
BREAKING(yaml): remove `listener` option from `parse()` and `stringify()`

iuioiua:


kt3k (approved):
This option exposes the entire LoaderState structure to the users via `listener` callback, which is something we don't want to consider as a public API. This removal of the option makes sense to me.

LGTM

PR #5227, opened 2024-07-01
fix(expect): fix validation of nth param in `toHaveBeenNthCalledWith` matcher

kt3k:
This PR fixes the validation of `nth` argument of `toHaveBeenNthCalledWith` matcher.


iuioiua (approved):
Nice catch

PR #5229, opened 2024-07-01
BREAKING(yaml): remove `ParseOptions.legacy` option

iuioiua:
### What's changed

The `ParseOptions.legacy` option has been removed.

### Motivation

Setting this option to true allowed documents adhering to versions <1.2 of the YAML specification. v1.2 of the YAML spec was published in 2009 (15 years ago). It's fair to say that most, if not, all, YAML use these days adheres to >=v1.2 of the spec, making this option redundant.

### Migration guide

This option only affected those who have the option set to `true` and are adhering to YAML spec version <1.2. If you require this feature, use [`@std/yaml@0.224.3`](https://jsr.io/@std/yaml@0.224.3), as this is the last version containing this feature. Otherwise, we recommend updating your document to use the latest version of the YAML spec.

### DIFF
--- a/yaml/_loader/loader.ts
+++ b/yaml/_loader/loader.ts
@@ -1593,7 +1593,7 @@ function readDocument(state: LoaderState) {
   let ch: number;
 
   state.version = null;
-  state.checkLineBreaks = state.legacy;
+  state.checkLineBreaks = false;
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

PR #5230, opened 2024-07-01
test(yaml): test `stringify()` case when string encoded into hex sequences

kt3k:
part of #3713 

This PR adds the test cases of `stringify` when the output is hex encoded strings.


iuioiua (approved):
Cute cat. LGTM.

Issue #5124, opened 2024-06-24
yaml: Rename negative property names in options


timreichen:
I think some properties of function options are poorly named. Options should be formulated positively imo and the default value adjusted accordingly.

https://github.com/denoland/deno_std/blob/3fac8a2e8655150181835acf83fd8a9071d5cd0e/yaml/_dumper/dumper_state.ts#L45
Proposal: Rename to `indentArrays` or something similar.

https://github.com/denoland/deno_std/blob/3fac8a2e8655150181835acf83fd8a9071d5cd0e/yaml/_dumper/dumper_state.ts#L74
The comment states: `if true, don't convert duplicate objects into references (default: false)`. Is it just me or is this confusing?
Proposal: Rename to `createRefs` or something similar.

https://github.com/denoland/deno_std/blob/3fac8a2e8655150181835acf83fd8a9071d5cd0e/yaml/_dumper/dumper_state.ts#L80
Proposal: Rename to `compatMode` or something similar.

WDYT?

iuioiua:
I agree. Negative tenses are far more ripe for confusion. The proposed names seem mostly fine. Happy to take a look at a PR.

timreichen:
How should that be done without breaking changes?

iuioiua:
We'd allow the breaking changes for this one, at this point in time.

PR #5232, opened 2024-07-01
BREAKING(yaml): rename negative option property names

timreichen:
Ref: https://github.com/denoland/deno_std/issues/5124


yaml/stringify.ts
```diff
@@ -41,15 +41,15 @@ export type StringifyOptions = {
...
   /**
    * If true, don't convert duplicate objects
-   * into references (default: false)
+   * into references (default: true)
    */
-  noRefs?: boolean;
+  createRefs?: boolean;
   /**
    * If true don't try to be compatible with older yaml versions.
    * Currently: don't quote "yes", "no" and so on,
```
kt3k, line 49: Now this describes the option in the opposite way

yaml/stringify.ts
```diff
@@ -41,15 +41,15 @@ export type StringifyOptions = {
   lineWidth?: number;
   /**
    * If true, don't convert duplicate objects
```
kt3k, line 43: Ditto. The description needs to be reversed as well

yaml/stringify.ts
```diff
@@ -14,7 +14,7 @@ export type StringifyOptions = {
   /** Indentation width to use (in spaces). */
   indent?: number;
   /** When true, will not add an indentation level to array elements */
```
kt3k, line 16: Ditto. The description needs to be reversed as well

PR #5233, opened 2024-07-01
BREAKING(yaml): remove `StringifyOptions.noCompatMode`

iuioiua:
### What's changed

The `StringifyOptions.noCompatMode` option has been removed. [`stringify()`](https://jsr.io/@std/yaml/doc/stringify/~/stringify) now behaves as though the option were set to `true`, by no longer wrapping [deprecated booleans](https://yaml.org/type/bool.html), such as "no" and "yes", in quotes.

### Motivation

This change was made to simplify the functionality of `stringify()` and have more reasonable defaults, by no longer supporting YAML 1.1 syntax, which was superceded by YAML 1.2 in 2009. This option was also difficult to understand due to it being negatively tensed.

### Migration guide

[Deprecated booleans](https://yaml.org/type/bool.html) will no longer be wrapped in quotes. Keep this in mind when processing stringified YAML output.


kt3k:
Oh, this is the default behavior change... I guess probably some users depend on this feature.

kt3k:
`yes` and `no` seem used in many places, for example https://github.com/search?q=yes+no+language%3Ayaml&type=code

kt3k:
`yes` and `no` seem used as boolean still today https://github.com/search?q=yes+no+language%3Ayaml&type=code

I think it's still sensible to quote these values in yaml output

iuioiua:
Discussed this with Yoshiya offline. We'll instead rename this option `compateMode`.

PR #5236, opened 2024-07-01
refactor(yaml): simplify schema

timreichen:
## Changes 
- removes `SchemaDefinition` and replaces it with `Schema`
- introduces `SCHEMA_MAP` and removes `replaceSchemaNameWithSchemaClass`
- ~uses `keyof typeof SCHEMA_MAP`  instead of hardcoded strings~
- replaces optional `options` with default empty option object for `stringify()`, `parse()` and `parseAll()`


yaml/stringify.ts
```diff
@@ -28,7 +28,7 @@ export type StringifyOptions = {
   /** Each tag may have own set of styles.	- "tag" => "style" map. */
   styles?: Record<string, "lowercase" | "uppercase" | "camelcase" | "decimal">;
   /** Name of the schema to use. */
-  schema?: "core" | "default" | "failsafe" | "json" | "extended";
+  schema?: keyof typeof SCHEMA_MAP;
```
kt3k, line 31: A public API can't reference private type to be able to publish to jsr (that causes lint error). I think this need to be literal type.
```suggestion
  schema?: "core" | "default" | "failsafe" | "json" | "extended";
```

iuioiua (approved):
Nice cleanup. Thank you.

PR #5239, opened 2024-07-02
refactor(yaml): minor cleanups

iuioiua:


yaml/_dumper/dumper.ts
```diff
@@ -726,13 +723,11 @@ function writeNode(
...
   }
 
-  const type = _toString.call(state.dump);
-
   if (block) {
     block = state.flowLevel < 0 || state.flowLevel > level;
   }
 
-  const objectOrArray = type === "[object Object]" || type === "[object Array]";
+  const objectOrArray = common.isObject(object) || Array.isArray(object);
```
kt3k, line 730:
```suggestion
  const objectOrArray = common.isObject(state.dump) || Array.isArray(state.dump);
```

yaml/_dumper/dumper.ts
```diff
@@ -780,13 +775,15 @@ function writeNode(
...
-    } else if (type === "[object String]") {
+    } else if (typeof state.dump === "string") {
       if (state.tag !== "?") {
         writeScalar(state, state.dump, level, iskey);
       }
     } else {
       if (state.skipInvalid) return false;
-      throw new YamlError(`unacceptable kind of an object to dump ${type}`);
+      throw new YamlError(
+        `unacceptable kind of an object to dump ${state.dump}`,
```
kt3k, line 785: This doesn't show the type of `state.dump` with this change
```suggestion
        `unacceptable kind of an object to dump ${_toString.call(state.dump)}`,
```

yaml/_loader/loader.ts
```diff
@@ -323,17 +277,12 @@ function storeMappingPair(
   // We need to convert keyNode to a string, but doing so can hang the process
   // (deeply nested arrays that explode exponentially using aliases).
   if (Array.isArray(keyNode)) {
-    keyNode = Array.prototype.slice.call(keyNode);
```
kt3k, line 326: This line seems for avoiding to modify the source array below

yaml/_dumper/dumper.ts
iuioiua: It should. Can you elaborate?

kt3k: If `state.dump` is, for example, a number, `_toString.call(state.dump)` is `[object Number]`, but `${state.dump}` is something like `42` (the number itself)

iuioiua: Oh, right. Would `typeof` be better?

yaml/_type/bool.ts
```diff
@@ -7,12 +7,7 @@ import { Type } from "../_type.ts";
 import { isBoolean } from "../_utils.ts";
 
 function resolveYamlBoolean(data: string): boolean {
-  const max = data.length;
-
-  return (
-    (max === 4 && (data === "true" || data === "True" || data === "TRUE")) ||
-    (max === 5 && (data === "false" || data === "False" || data === "FALSE"))
-  );
+  return ["true", "false"].includes(data.toLowerCase());
```
kt3k, line 10: This allows `tRuE` to be resolved to boolean. We want to only allow 3 forms `true` `True` `TRUE`

https://yaml.org/spec/1.2.2/#103-core-schema

yaml/_dumper/dumper.ts
kt3k: `typeof` works for most cases, but it doesn't work for `null` and arrays (both becomes `object`)

iuioiua: Ah, true. Will revert.

iuioiua: Actually, `Object.prototype.toString(null)` and `Object.prototype.toString([])` both evaluate to `[object Object]`. I'll go with `typeof` as it's a little cleaner.

kt3k: > Actually, Object.prototype.toString(null) and Object.prototype.toString([]) both evaluate to [object Object]

No, existing code is equivalent of `Object.prototype.toString.call(null)` and `Object.prototype.toString.call([])`. They evaluate to `[object Null]` and `[object Array]`

iuioiua: Ok. Thanks for clarifying. TIL.

yaml/_loader/loader.ts
```diff
@@ -330,10 +284,7 @@ function storeMappingPair(
         return throwError(state, "nested arrays are not supported inside keys");
       }
 
-      if (
-        typeof keyNode === "object" &&
-        _class(keyNode[index]) === "[object Object]"
-      ) {
+      if (typeof keyNode === "object" && common.isObject(keyNode[index])) {
```
kt3k, line 287: This part is also safe to keep as `Object.prototype.toString.call(keyNode[index]) === "[object Object]"`. `common.isObject` is true for arrays, functions, etc, but it's not with the existing code.

yaml/_type/set.ts
```diff
@@ -6,15 +6,11 @@
 import { Type } from "../_type.ts";
 import type { Any } from "../_utils.ts";
 
-const { hasOwn } = Object;
-
 function resolveYamlSet(data: Any): boolean {
   if (data === null) return true;
 
   for (const key in data) {
-    if (hasOwn(data, key)) {
```
kt3k, line 15: This `hasOwn` is for avoiding to be affected by Object.prototype extension. If someone extend `Object.prototype`, then unrelated `key` appears here.

yaml/_type/pairs.ts
```diff
@@ -6,15 +6,13 @@
...
-const _toString = Object.prototype.toString;
-
 function resolveYamlPairs(data: Any[][]): boolean {
   if (data === null) return true;
 
   const result = Array.from({ length: data.length });
 
   for (const [index, pair] of data.entries()) {
-    if (_toString.call(pair) !== "[object Object]") return false;
+    if (pair.toString() !== "[object Object]") return false;
```
kt3k, line 15: This will throw when `pair` is `null` or `undefined`. The existing code doesn't

yaml/_type/omap.ts
```diff
@@ -17,18 +15,18 @@ function resolveYamlOmap(data: Any): boolean {
   for (const pair of data) {
     pairHasKey = false;
 
-    if (_toString.call(pair) !== "[object Object]") return false;
+    if (!isObject(pair)) return false;
```
kt3k, line 18: This isn't equivalent change. `isObject` is true for arrays and functions, but the existing code is not

kt3k (approved):
LGTM

iuioiua:
Apologies for the silly mistakes in this one. Will do better next time.

PR #5240, opened 2024-07-02
test(yaml): add `parse()` test for duplicate keys

iuioiua:
Unblocks #5240

### DIFF
--- a/yaml/parse_test.ts
+++ b/yaml/parse_test.ts
@@ -784,3 +784,16 @@ Deno.test("parse() throws with empty mapping key", () => {
     "incomplete explicit mapping pair; a key node is missed; or followed by a non-tabulated empty line at line 1, column 3:\n    ? : 1\n      ^",
   );
 });
+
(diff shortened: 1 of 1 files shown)


yaml/parse_test.ts
iuioiua, line 1: Current documentation says that if `ParseOptions.json` is `true`, then duplicate keys cause `parse()` to throw. By default, `ParseOptions.json` is `false`, and duplicate keys cause `parse()` to throw. In other words, `ParseOptions.json` currently does the exact opposite of what it should 😆

kt3k: `JSON.parse()` doesn't seem throwing with duplicated keys?

```
> JSON.parse('{"a":1,"a":2}')
{ a: 2 }
```

kt3k (approved):
LGTM

yaml/parse_test.ts
iuioiua: Ok - then my understanding of `JSON.parse()` was wrong 😆

PR #5241, opened 2024-07-02
BREAKING(yaml): remove `ParseOptions.json`

iuioiua:


iuioiua:
Discussed this with Yoshiya. We'll instead rename this option to `allowDuplicateKeys`.

PR #5242, opened 2024-07-02
refactor(expect,testing): update `@std/assert` imports

iuioiua:
This cleans up dependencies following #5176.


kt3k (approved):
LGTM. Nice!

PR #5243, opened 2024-07-02
chore(yaml): simplify loader

timreichen:
## Changes
- merges `loader/_loader.ts` and `loader/_loader_state.ts` into `_parser.ts`
- renames `LoaderState` to `Parser` and `LoaderStateOptions` to `ParserOptions`
- inlines lots of functions into `Parser` class to avoid passing instance as an argument
- renames `loadDocument()` to `parseDocument()`
- renames `loadDocuments()` to `parseDocuments()`
- removes `DirectiveHandler` type and makes directiveHandlers standalone functions


iuioiua (approved):
This structure makes a lot more sense to me. WYDT, @kt3k?

kt3k:
This seems including too many structure changes.

YAML is least tested module in std. This structure change at this point feels too risky to me.

kt3k:
Also we can't track the change with this huge diffs as @iuioiua points.

Closing without merge

PR #5245, opened 2024-07-02
test(yaml): check handling of binary type

kt3k:
part of #3713 

This PR adds the test cases which check the handling of `!!binary` type.


iuioiua (approved):
Nice

PR #5248, opened 2024-07-02
refactor(yaml): add `sanitizeInput()` function

timreichen:
- moves code into a dedicated `sanitizeInput()` function

### DIFF
--- a/yaml/_loader/loader.ts
+++ b/yaml/_loader/loader.ts
@@ -1656,12 +1656,8 @@ function readDocument(state: LoaderState) {
   }
 }
 
-export function loadDocuments(
-  input: string,
-  options?: LoaderStateOptions,
-): unknown[] {
+function sanitizeInput(input: string) {
   input = String(input);
-  options = options || {};
 
   if (input.length !== 0) {
     // Add tailing `\n` if not exists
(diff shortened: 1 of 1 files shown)


kt3k:
`sanitizeInput` is only used once. I don't see this change reasonable.

iuioiua:
I don't see the purpose of this change either.

timreichen:
> `sanitizeInput` is only used once. I don't see this change reasonable.

I will create a second PR that uses generator functions for documents in `loader.ts`. Therefore `load()` and `loadDocuments()` will need to call `sanitizeInput()` individually. The gain will be to throw early if there are more than one document in `load()` instead of parsing all documents and then throw.

yaml/_loader/loader.ts
```diff
@@ -1690,10 +1687,21 @@ export function loadDocuments(
...
+  return input;
+}
+
+export function loadDocuments(
+  input: string,
+  options?: LoaderStateOptions,
+): unknown[] {
+  input = sanitizeInput(input);
+
+  options = options || {};
```
iuioiua, line 1702:
```suggestion
export function loadDocuments(
  input: string,
  options: LoaderStateOptions = {},
): unknown[] {
  input = sanitizeInput(input);
```

iuioiua (changes requested):

kt3k (approved):
LGTM

iuioiua (approved):

PR #5249, opened 2024-07-02
refactor(yaml): add default options object

timreichen:
- replaces optional options with default object


iuioiua:
Can you please elaborate on why this is needed?

timreichen:
> Can you please elaborate on why this is needed?

It  replaces `options = options || {};`.

iuioiua (approved):

PR #5256, opened 2024-07-02
fix(yaml): fix `!!int` style variation handling in `stringfiy`

kt3k:
This PR fixes the handling of style variaions for `!!int` type.

### DIFF
--- a/yaml/_type.ts
+++ b/yaml/_type.ts
@@ -7,7 +7,14 @@
 import type { Any, ArrayObject } from "./_utils.ts";
 
 export type KindType = "sequence" | "scalar" | "mapping";
-export type StyleVariant = "lowercase" | "uppercase" | "camelcase" | "decimal";
+export type StyleVariant =
+  | "lowercase"
+  | "uppercase"
+  | "camelcase"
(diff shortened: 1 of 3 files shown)


yaml/_type.ts
```diff
@@ -7,7 +7,25 @@
...
+  | "binary"
+  | "bin"
+  | 2
+  | "octal"
+  | "oct"
+  | 8
+  | "hexadecimal"
+  | "hex"
+  | 16;
 export type RepresentFn = (data: Any, style?: StyleVariant) => Any;
```
iuioiua, line 29: Why do we have mutiple variants that mean the same thing? Is this behavior defined in the spec? Having multiple ways to do that same thing might confuse users.

iuioiua (changes requested):

yaml/_type.ts
kt3k: It's already implemented in int type. See `yaml/_type/int.ts`

iuioiua: Oh, I see.

iuioiua (approved):

PR #5259, opened 2024-07-02
refactor(yaml): merge `yaml/_loader/` files

timreichen:
- merges `_loader/loader.ts` and `_loader/loader_state.ts` into  `_loader/loader.ts`


iuioiua (approved):
LGTM. Thank you.

PR #5262, opened 2024-07-02
refactor(yaml): replace `Type` class with interface

timreichen:
- removes `Type` class and replaced with a `Type` interface.


iuioiua (approved):
I like this and was going to do this myself at some point. LGTM.

yaml/_type.ts
```diff
@@ -10,44 +10,15 @@ export type KindType = "sequence" | "scalar" | "mapping";
...
-      this.instanceOf = options.instanceOf;
-      this.predicate = options.predicate;
-      this.represent = options.represent;
-      this.defaultStyle = options.defaultStyle;
-      this.styleAliases = options.styleAliases;
-    }
-  }
-  resolve: (data?: Any) => boolean = (): boolean => true;
-  construct: (data?: Any) => Any = (data): Any => data;
+  resolve?: (data?: Any) => boolean;
```
kt3k, line 22: Making resolve optional feels strange to me

yaml/_type.ts
```diff
@@ -10,44 +10,15 @@ export type KindType = "sequence" | "scalar" | "mapping";
...
+  construct?: (data?: Any) => Any;
```
kt3k, line 23: I don't think this should be optional

kt3k:
I'm not in favor of this change. There are many changes in optionality of fields. I'm very unsure this doesn't introduce any bugs.

iuioiua (commented):
I think this change is an improvement in design. It also seems safe enough to do.

However, I understand Yoshiya's sentiment. In #5239, I made changes which I thought were harmless, as tests passed. But rather testing is not yet at a level for this package to instil such confidence.

@kt3k, if you feel strongly about this, let's close this for now and revisit once testing is at an adequate level.

yaml/_loader/loader.ts
```diff
@@ -1545,14 +1547,16 @@ function composeNode(
         );
       }
 
-      if (!type.resolve(state.result)) {
+      if (!type.resolve?.(state.result) ?? true) {
```
kt3k, line 1550: This condition is reversed when `resolve` is a default implementation

yaml/_type.ts
kt3k: > An alternative would be to require resolve and construct methods in Type and add them to types that are missing them.

I think that is more correct way to perform this transition.

kt3k: this line https://github.com/denoland/deno_std/pull/5262/files#r1663457908 seems a bug caused by the optionality of the field

yaml/_loader/loader.ts
iuioiua: Nice catch! Yeah, if we're to go through with this PR, let's make `Type.resolve()` and `Type.construct()` required.

yaml/_type.ts
timreichen: Reverted.

yaml/_loader/loader.ts
timreichen: Done.

kt3k (approved):
LGTM

PR #5266, opened 2024-07-03
BREAKING(front-matter): remove `createExtractor()`

### LINKED ISSUE #5265: suggestion: make `createExtractor` module private in `@std/front-matter`
This module contains implementation details that are extraneous to most, if not all, users. Doing so would make `@std/front-matter` easier to understand.

iuioiua:
### What's changed

[`@std/front-matter/create-extractor`](https://jsr.io/@std/front-matter@0.224.3/doc/create-extractor/~/) has been made private. It is still used in the underlying package implementation.

### Motivation

This module was removed because it exposed implementation details of `@std/front-matter`'s `extract()` functions and wasn't used (correctly) according to [GitHub search](https://github.com/search?q=%22%40std%2Ffront-matter%22+%22createExtractor%22&type=code). Making this code private allows us to more easily make improvements in the backend without breaking user code.

Closes #5265


kt3k (approved):
Note: `createExtractor` was for supporting arbitrary combination for front-matter language, but we couldn't find any correct usage in github open source search.

LGTM

PR #5267, opened 2024-07-03
test(assert): add "`assert()` throws if expr is falsy" test

iuioiua:
Mainly so `@std/assert` reaches 100% coverage.


kt3k (approved):
LGTM. Nice!

PR #5269, opened 2024-07-03
refactor(csv): remove dead code

iuioiua:
Brings `@std/csv` coverage to 100%.

Towards #3713

### DIFF
--- a/csv/_io.ts
+++ b/csv/_io.ts
@@ -175,20 +175,6 @@ export async function parseRecord(
             break parseField;
           }
           recordBuffer += "\n"; // preserve line feed (This is because TextProtoReader removes it.)
-        } else {
-          // Abrupt end of file (EOF on error).
-          if (!opt.lazyQuotes) {
-            const col = runeCount(fullLine);
-            quoteError = new ParseError(
-              startLine + 1,
-              lineIndex,
-              col,
-              ERR_QUOTE,
-            );
-            break parseField;
-          }
-          fieldIndexes.push(recordBuffer.length);
-          break parseField;
         }
       }
     }


csv/_io.ts
```diff
@@ -175,20 +175,6 @@ export async function parseRecord(
...
-            quoteError = new ParseError(
-              startLine + 1,
-              lineIndex,
-              col,
-              ERR_QUOTE,
-            );
-            break parseField;
-          }
-          fieldIndexes.push(recordBuffer.length);
-          break parseField;
```
iuioiua, line 191: This block is unreachable, as `line.length === 0` and `line.length > 0` are both involved in logical OR expressions above. It's impossible for `line.length < 0`.

kt3k: I can't confirm your reasoning. This else-block seems a part of the below structure. Is `line.length === 0` involved in these conditions?

```ts
if (i >= 0) {
  // ...
} else if (line.length > 0 || !reader.isEOF()) {
  // ...
} else {
  // ...
}
```

iuioiua: https://github.com/denoland/deno_std/blob/769cc02d7199e766a20036b9c6136dfa0fbd7c7e/csv/_io.ts#L87

kt3k: `line` variable is modified in the below part. https://github.com/denoland/deno_std/blob/b968c2fe21063030f972d71f846930cb9a98ceb8/csv/_io.ts#L119

I don't think that reasoning works if `line` is modified in the middle

timreichen: Maybe https://github.com/denoland/deno_std/pull/5277 might help for more clarity

iuioiua: I'll close this until CSV simplification changes are made.

PR #5270, opened 2024-07-03
refactor(yaml): move loader and dumper files

timreichen:
- moves `_loader/loader.ts` to `_loader.ts`
- moves `_dumper/dumper.ts` to `_dumper.ts`


iuioiua (approved):
Nice

PR #5273, opened 2024-07-03
refactor(yaml): inline error functions as methods

timreichen:
- inlines `throwError()` and `throwWarning()` in `LoaderState` to avoid passing the state as an argument
- replaces `call(null ...)` in `throwWarning()` with normal function call


kt3k (approved):
LGTM

PR #5274, opened 2024-07-03
refactor(yaml): simplify `sanitizeInput()`

timreichen:
- uses existing `isEOL()` instead of char check
- replace `length !== 0` with `length > 0`


timreichen:
Is `input = String(input);` really needed if input is of type `string`?

kt3k (approved):
LGTM

PR #5275, opened 2024-07-03
refactor(yaml): remove `State` class

timreichen:
- removes `State` and inlines schema property into `LoaderState` and `DumperState`


kt3k (approved):
LGTM

PR #5277, opened 2024-07-03
refactor(csv): simplify `parseRecord()`

timreichen:
- removes nesting for more clarity


iuioiua (commented):
Are you able to break this up into smaller PRs?

timreichen:
> Are you able to break this up into smaller PRs?

https://github.com/denoland/deno_std/pull/5297

iuioiua (approved):
Nice. @kt3k, PTAL.

kt3k:
I don't understand what's improved in this change.

iuioiua:
Mostly, it reduces removes unnecessary logic for when `line.length === 0` and slightly reduces nesting.

timreichen:
> I don't understand what's improved in this change. This introduces another label for indicating inner loop, and introduce several `continue label` statements. That doesn't feel 'simplification' to me

I actually think it is more readable and clearer to understand. I imagine that this can help extracting some logic into that can be shared between the two files in another PR.

kt3k:
Nesting level seems exactly the same. Looks like just swapping the if-clause and else-clause randomly. Also increased number of lines seem concerning. I think we should avoid this kind of random change.

Closing without merge.

PR #5278, opened 2024-07-03
test(fmt): add test cases for `printf()`

kt3k:
part of #3713 

This improves the test coverage of `fmt/printf.ts` from 94.58% to 96.93% (reduces 16 uncovered lines)


iuioiua (approved):
Nice

PR #5279, opened 2024-07-03
fix(fmt): fix the case when mantissa exceeds 10 by rounding and exponent is negative

kt3k:
Currently `e` formatting with rounding has a bug in some edge cases. For example `sprintf("%.3e", 0.000099999)` becomes `1.000e--04` (which has double minus sign). This PR fixes it.


iuioiua (approved):
Good catch

Issue #5195, opened 2024-06-29
to-do: investigate the viability of `ParseOptions` and `StringifyOptions` in `@std/yaml`


iuioiua:
These interfaces have a lot of options, and each added option means more engineering overhead for both users and maintainers. We should go over each option and only keep the ones that have sufficient justifications for existing.

[`ParseOptions`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions):
- [x] [`legacy`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.legacy) (removed in #5229)
- [x] [`listener`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.listener) (removed in #5224)
- [x] [`filename`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.filename) (removed in #5223)
- [x] [`schema`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.schema)
- [ ] [`json`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.json) (renamed to `allowDuplicateKeys` in #5282)
- [x] [`onWarning`](https://jsr.io/@std/yaml/doc/parse/~/ParseOptions.onWarning) (reworked in #5284)

[`StringifyOptions`](https://jsr.io/@std/yaml/doc/~/DumpOptions) (currently `DumpOptions` but to be updated in the next release):
- [ ] [`indent`](https://jsr.io/@std/yaml/doc/~/DumpOptions.indent)
- [ ] [`noArrayIndent`](https://jsr.io/@std/yaml/doc/~/DumpOptions.noArrayIndent)
- [ ] [`skipInvalid`](https://jsr.io/@std/yaml/doc/~/DumpOptions.skipInvalid)
- [ ] [`flowLevel`](https://jsr.io/@std/yaml/doc/~/DumpOptions.flowLevel)
- [ ] [`styles`](https://jsr.io/@std/yaml/doc/~/DumpOptions.styles)
- [x] [`schema`](https://jsr.io/@std/yaml/doc/~/DumpOptions.schema)
- [ ] [`sortKeys`](https://jsr.io/@std/yaml/doc/~/DumpOptions.sortKeys)
- [ ] [`lineWidth`](https://jsr.io/@std/yaml/doc/~/DumpOptions.lineWidth)
- [ ] [`noRefs`](https://jsr.io/@std/yaml/doc/~/DumpOptions.noRefs)
[...]

PR #5282, opened 2024-07-03
BREAKING(yaml): rename `ParseOptions.json` to `ParseOptions.allowDuplicateKeys`

iuioiua:
### What's changed

[`ParseOptions.json`](https://jsr.io/@std/yaml@0.224.3/doc/parse/~/ParseOptions.json) has been renamed `ParseOptions.allowDuplicateKeys`. This change only affects those that previously set `ParseOptions.json` to `true`. No changes in behavior have been made.

### Motivation

This name more clearly describes how the option affects the behavior of [`parse()`](https://jsr.io/@std/yaml/doc/~/parse).

### Migration guide

To migrate, use the `allowDuplicateKeys` option instead of the `json` option for `parse()`.

```diff
import { parse } from "@std/yaml/parse";

const yamlString = `name: John Doe
age: 30
name: Jane Doe`;

- parse(yamlString, { json: true });
+ parse(yamlString, { allowDuplicateKeys: true });
```

### Related

Towards #5195

### DIFF
--- a/yaml/parse.ts
+++ b/yaml/parse.ts
@@ -14,7 +14,7 @@ export interface ParseOptions {
   /** Name of the schema to use.*/
   schema?: "core" | "default" | "failsafe" | "json" | "extended";
   /** compatibility with JSON.parse behaviour. */
-  json?: boolean;
+  allowDuplicateKeys?: boolean;
   /** function to call on warning messages. */
   onWarning?(this: null, e?: Error): void;
 }
--- a/yaml/_loader.ts
+++ b/yaml/_loader.ts
@@ -63,7 +63,7 @@ interface LoaderStateOptions {
   /** specifies a schema to use. */
   schema?: Schema;
   /** compatibility with JSON.parse behaviour. */
-  json?: boolean;
+  allowDuplicateKeys?: boolean;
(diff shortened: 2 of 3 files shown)


kt3k:
I'm in favor of this change. The option name `json` doesn't describe what it changes.

I slightly feel `allowDuplicatedKeys` is a bit more natural. What do you think?

iuioiua:
> I slightly feel `allowDuplicatedKeys` is a bit more natural. What do you think?

I don't think it's any more natural than what it is now.

iuioiua:
I did a quick search - "duplicate keys" seems more correct.

yaml/parse.ts
```diff
@@ -14,7 +14,7 @@ export interface ParseOptions {
   /** Name of the schema to use.*/
   schema?: "core" | "default" | "failsafe" | "json" | "extended";
   /** compatibility with JSON.parse behaviour. */
```
kt3k, line 16: Let's also change the description

iuioiua: I'll be reworking docs for all of `@std/yaml` once the breaking changes are made.

kt3k (approved):
LGTM

PR #5286, opened 2024-07-04
BREAKING(yaml): rename `ParseOptions.noArrayIndent` to `ParseOptions.arrayIndent`

iuioiua:
### What's changed

The `ParseOptions.noArrayIndent` option has been renamed to `ParseOptions.arrayIndent`. As a result, the polarity of the option has been swapped. In other words, `true` behavior has been changed to `false` behavior, and visa-versa. This only affects users who currently use `ParseOptions.noArrayIndent`.

### Motivation

### Migration guide

Use `ParseOptions.arrayIndent` instead of `ParseOptions.noArrayIndent` and swap the polarity of the option.

```diff
import { parse } from "@std/yaml/parse";

const yamlString = `fruits:
  - Apple
  - Banana
  - Cherry`;

- parse(yamlString, { noArrayIndent: false });
+ parse(yamlString, { arrayIndent: true });
```

### Related

Towards #5124

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -99,8 +99,8 @@ function compileStyleMap(
 export interface DumperStateOptions {
   /** indentation width to use (in spaces). */
   indent?: number;
-  /** when true, will not add an indentation level to array elements */
-  noArrayIndent?: boolean;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #5287, opened 2024-07-04
BREAKING(yaml): rename `StringifyOptions.noCompatMode` to `StringifyOptions.compatMode`

iuioiua:
### What's changed

The `ParseOptions.noCompatMode` option has been renamed to `ParseOptions.compatMode`. As a result, the polarity of the option has been swapped. In other words, `true` behavior has been changed to `false` behavior, and visa-versa. This only affects users who currently use `ParseOptions.noCompatMode`.

### Motivation

### Migration guide

Use `ParseOptions.compatMode` instead of `ParseOptions.noCompatMode` and swap the polarity of the option.

```diff
import { parse } from "@std/yaml/parse";

const yamlString = `fruits:
  - Apple
  - Banana
  - Cherry`;

- parse(yamlString, { noCompatMode: false });
+ parse(yamlString, { compatMode: true });
```

### Related

Towards #5124

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -131,11 +131,11 @@ export interface DumperStateOptions {
    */
   noRefs?: boolean;
   /**
-   * if true don't try to be compatible with older yaml versions.
+   * if false don't try to be compatible with older yaml versions.
    * Currently: don't quote "yes", "no" and so on,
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

PR #5288, opened 2024-07-04
BREAKING(yaml): rename `StringifyOptions.noRefs` to `StringifyOptions.useAnchors`

### LINKED ISSUE #5124: yaml: Rename negative property names in options

iuioiua:
### What's changed

The `ParseOptions.noRefs` option has been renamed to `ParseOptions.useAnchors`. As a result, the polarity of the option has been swapped. In other words, `true` behavior has been changed to `false` behavior, and visa-versa. This only affects users who currently use `ParseOptions.noRefs`.

### Motivation

### Migration guide

Use `ParseOptions.useAnchors` instead of `ParseOptions.noRefs` and swap the polarity of the option.

```diff
import { parse } from "@std/yaml/parse";

const yamlString = `fruits:
  - Apple
  - Banana
  - Cherry`;

- parse(yamlString, { noRefs: false });
+ parse(yamlString, { useAnchors: true });
```

### Related

Closes #5124

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -129,7 +129,7 @@ export interface DumperStateOptions {
    * if false, don't convert duplicate objects
    * into references (default: true)
    */
-  createRefs?: boolean;
+  useAnchors?: boolean;
   /**
    * if false don't try to be compatible with older yaml versions.
    * Currently: don't quote "yes", "no" and so on,
@@ -153,7 +153,7 @@ export class DumperState {
   flowLevel: number;
   sortKeys: boolean | ((a: Any, b: Any) => number);
   lineWidth: number;
-  createRefs: boolean;
+  useAnchors: boolean;
   compatMode: boolean;
   condenseFlow: boolean;
   implicitTypes: Type[];
@@ -174,7 +174,7 @@ export class DumperState {
     styles = null,
     sortKeys = false,
     lineWidth = 80,
-    createRefs = true,
+    useAnchors = true,
     compatMode = true,
     condenseFlow = false,
   }: DumperStateOptions) {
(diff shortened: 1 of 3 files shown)


kt3k:
I tested `noRefs` option, but it doesn't seem working in `std/yaml`.

```ts
import { stringify } from "./yaml/mod.ts";
import * as yaml from "npm:js-yaml";

const b = { baz: 42 };
const a = {
  foo: {
    foo: {
      foo: {
      },
    },
  },
};
console.log(stringify([a, a], { noRefs: false }));
console.log(stringify([a, a], { noRefs: true }));
console.log(yaml.dump([a, a], { noRefs: false }));
console.log(yaml.dump([a, a], { noRefs: true }));
```

This prints:
```
- foo:
    foo:
      foo: {}
- foo:
    foo:
      foo: {}

- foo:
    foo:
      foo: {}
- foo:
    foo:
      foo: {}

- &ref_0
  foo:
    foo:
      foo: {}
- *ref_0

- foo:
    foo:
      foo: {}
- foo:
    foo:
      foo: {}
```

I suggest we should rather remove this option. What do you think?

iuioiua:
Oh, yes, let's remove then.

kt3k:
Ah, wait. The feature seems working with 0.224.3, but it's gone in 1.0.0-rc.1. It seems broken somewhere between 0.224.3 - 1.0.0-rc.1

kt3k:
How about `useAnchors` instead? 'reference' is not an official yaml term for it. Also 'create' sounds a bit strange to me. It reduces the duplicate occurrence of the same objects, instead of creating something

iuioiua:
SGTM

kt3k (approved):
LGTM

PR #5291, opened 2024-07-04
BREAKING(yaml): remove `StringifyOptions.noRefs`

iuioiua:


kt3k:
Sorry `noRefs` was actually working until recently. I'll fix `noRefs` first.

https://github.com/denoland/deno_std/pull/5288#issuecomment-2208148575

PR #5295, opened 2024-07-04
docs(assert): improve `assertNotStrictEquals` example

kt3k:
This PR improves the example of `assertNotStrictEquals` doc.

### DIFF
--- a/assert/not_strict_equals.ts
+++ b/assert/not_strict_equals.ts
@@ -12,8 +12,11 @@ import { format } from "@std/internal/format";
  * ```ts no-eval
  * import { assertNotStrictEquals } from "@std/assert";
  *
- * assertNotStrictEquals(1, 1); // Doesn't throw
- * assertNotStrictEquals(1, 2); // Throws
+ * assertNotStrictEquals(1, 1); // Throws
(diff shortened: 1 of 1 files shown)


assert/not_strict_equals.ts
```diff
@@ -12,8 +12,11 @@ import { format } from "@std/internal/format";
  * ```ts no-eval
  * import { assertNotStrictEquals } from "@std/assert";
  *
- * assertNotStrictEquals(1, 1); // Doesn't throw
- * assertNotStrictEquals(1, 2); // Throws
+ * assertNotStrictEquals(1, 1); // Throws
+ * assertNotStrictEquals(1, 2); // Doesn't throw
```
kt3k, line 16: The inline comments were wrong

assert/not_strict_equals.ts
```diff
@@ -12,8 +12,11 @@ import { format } from "@std/internal/format";
...
+ *
+ * assertNotStrictEquals(0, 0); // Throws
+ * assertNotStrictEquals(0, -0); // Doesn't throw
```
kt3k, line 19: These examples show that this check uses `Object.is` internally

iuioiua (approved):
Good catch

PR #5296, opened 2024-07-04
docs(assert): improve `assertObjectMatch` docs

kt3k:
This PR improves the docs of `assertObjectMatch`.

### DIFF
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -3,7 +3,7 @@
 import { assertEquals } from "./equals.ts";
 
 /**
- * Make an assertion that `actual` object is a superset of `expected` object,
+ * Make an assertion that `expected` object is a subset of `actual` object,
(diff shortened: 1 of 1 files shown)


assert/object_match.ts
```diff
@@ -3,7 +3,7 @@
 import { assertEquals } from "./equals.ts";
 
 /**
- * Make an assertion that `actual` object is a subset of `expected` object,
+ * Make an assertion that `actual` object is a superset of `expected` object,
```
kt3k, line 6: `actual` needs to include all of `expected` fields. I think `actual` is called `superset` in this case.

iuioiua: I think it'd be a little easier to understand if we say that `expected` is a subset of `actual`.

iuioiua (approved):

PR #5297, opened 2024-07-04
refactor(csv): rename arguments, variables and loop

timreichen:
- renames `opt` to `options` argument
- ~renames `fullLine` to `line` and `line` to `currentLine`~
- ~renames `parseField` to `currentLineLoop`~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*.{ts,js,yml}]
+[*.{ts,js,yml,mjs}]
 charset = utf-8
 end_of_line = lf
 insert_final_newline = true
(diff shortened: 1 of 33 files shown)


kt3k:
> renames fullLine to line and line to currentLine
> renames parseField to currentLineLoop

I don't see the point of these renames. Can you revert these changes?

timreichen:
> > renames fullLine to line and line to currentLine
> > renames parseField to currentLineLoop
> 
> I don't see the point of these renames. Can you revert these changes?

I think it is weird to have an assignment which then is stored in another variable because the former is mutated.
How about
```ts
let fullLine = this.#readLine();
...
let line = fullLine;
```

kt3k (approved):
Looks better now. LGTM

PR #5298, opened 2024-07-04
refactor(csv): remove `runeCount()` function

timreichen:
- replaces `runeCount()` with length of strings


iuioiua (approved):

kt3k:
This makes the count of characters broken for surrogate pair characters

```
> "🐱".length
2
> Array.from("🐱").length
1
```

timreichen:
> This makes the count of characters broken for surrogate pair characters
> 
> ```
> > "🐱".length
> 2
> > Array.from("🐱").length
> 1
> ```

Ah damn, sorry for the oversight. I think we should add a test for that.

PR #5301, opened 2024-07-04
refactor(yaml): rename `throwWarning()` to `dispatchWarning()`

timreichen:
- renames `throwWarning()` to  `dispatchWarning()` as the function does not throw an error


iuioiua (approved):
Fair point. Thanks!

PR #5302, opened 2024-07-04
refactor(yaml): add `readIndent()` method

timreichen:


kt3k:
The method name sounds strange to me. How about `readIndent`?

kt3k (approved):
LGTM

PR #5305, opened 2024-07-04
refactor(yaml): simplify `encodeHex()`

timreichen:


yaml/_dumper.ts
```diff
@@ -195,25 +195,20 @@ export class DumperState {
...
-    handle = "u";
-    length = 4;
-  } else if (character <= 0xffffffff) {
-    handle = "U";
-    length = 8;
-  } else {
-    throw new YamlError(
-      "code point within a string may not be greater than 0xFFFFFFFF",
-    );
+    return `\\x${"0".repeat(2 - length)}${string}`;
```
kt3k, line 201: Looks like this can be simplified further with `.padStart` method?

timreichen: You are right, I overlooked that one.

PR #5306, opened 2024-07-04
refactor(yaml): remove `Any` type

timreichen:
- replaces `Any` type with `any`
- adds lint ignore rules
- replaces any with unknown in type assertion functions

Note: The point of this PR is to expose any as a suboptimal type, which should be replaced where possible.


kt3k (approved):
LGTM. Thanks!

PR #5308, opened 2024-07-04
fix(expect): make `.not` aware of whether is it is in async context

lowlighter:
Makes it possible to use `resolves.not` without hinting error (else you need to use `not.resolves` to ditch out this warning but it feels more awkard)


iuioiua (approved):
Nice! Thank you.

PR #5309, opened 2024-07-04
fix(expect): improve `expect` type to make it work better with `expect.extend`

lowlighter:
Looking at how you need to use `expect.extend`, the return type `ExtendMatchResult` is expected but it is not exported publicly:
https://github.com/denoland/deno_std/blob/81943c34c7ba662ee44446d4fa49790e7138ee23/expect/_extend_test.ts#L7-L11

Workaround is to use `any` but it's not ideal
```ts
declare module "@std/expect" {
  interface Expected {
    toMatchDescriptor: (key: PropertyKey, expected:PropertyDescriptor) => any;
  }
}
```
```diff
-    toMatchDescriptor: (key: PropertyKey, expected:PropertyDescriptor) => any;
+    toMatchDescriptor: (key: PropertyKey, expected:PropertyDescriptor) => ReturnType<Matcher>;

### DIFF
--- a/expect/_extend_test.ts
+++ b/expect/_extend_test.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
 import { expect } from "./expect.ts";
-import type { MatcherContext, Tester } from "./_types.ts";
+import type { Async, Expected, MatcherContext, Tester } from "./_types.ts";
 import { AssertionError, assertThrows } from "@std/assert";
 
 declare module "./_types.ts" {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Review by iuioiua — CHANGES_REQUESTED** [review]
It seems we need to export more of the underlying types, like `MatchResult`, `ExtendMatchResult`, `MatcherContext`, etc.

Also, it seems this PR targets this TODO:
https://github.com/denoland/deno_std/blob/a92e1b6060347bbf9f4c39aeb3120adb812f4c7e/expect/expect.ts#L290

**Comment by lowlighter** [comment]
@iuioiua 
Yeah I noticed with the errors of the ci, there's a lot more to export but I'm not so sure what's the best anymore

It feels weird to export so many types which aren't necessarly "needed", because TS seems to be able to infer a lot of stuff.
Also turns out that passing `unknown` as return type is acceptable and doesn't really affects the code despite being a bit "less precise":

```ts
declare module "@std/expect" {
  interface Expected {
    toMatchDescriptor: (key: PropertyKey, expected:PropertyDescriptor) => unknown;
  }
}

expect(Object.defineProperties({}, {foo:{value:"bar", writable:false}}))
  .toMatchDescriptor("foo", {writable:false}) // no typing issue nor runtime errors
```

So tbh I'm not sure this is worth implementing, thoughts ?

iuioiua:
Any preferences, @kt3k?

lowlighter:
I'm not sure whether the way I did it is the "official way" to do it since https://jsr.io/@std/expect/doc/~/expect.extend is currently undocumented (I looked up the unit test and wasn't even aware such a thing existed previously)

But if that's seem the intended way, maybe I could change the scope of this PR to document `expect.extend` instead ?

kt3k:
> But if that's seem the intended way, maybe I could change the scope of this PR to document expect.extend instead ?

Sounds great!

lowlighter:
Hum, there's a small catch though

Extending `Expected` interface based on how it's done in the unit test render the package unpublishable on jsr.io :
```bash
error[unsupported-ambient-module]: found an ambient module, which is a global augmentation, which are not unsupported
 --> C:\...\libs\testing\_expect.ts:8:16
  |
8 | declare module "@std/expect" {
  |                ^^^^^^^^^^^^^
  = hint: remove the ambient module declaration

  info: ambient modules are not supported because they can modify the types of a module from outside of that module
  docs: https://jsr.io/go/slow-type-unsupported-ambient-module
```

I think maybe it can be worked around by changing the signature of `expect():Expected` to `expect<T extends Expected = Expected>():T` so it's possible to extends the interface instead ? (Not sure if it works though, not tested)

lowlighter:
@kt3k @iuioiua 

After experimenting a bit for [`@libs/testing/expect`](https://jsr.io/@libs/testing/2.1.0/expect.ts#L18), it seems like the workaround suggested to just make the `expect` generic to accept and extended `Extended` interface works good enough.

There's no need to export additional typings it seems (just using unknown as return type is fine), however there's still the fact you need to manually override the `not`/`resolves`/`rejects` since it's not possible to create interface with recursive type parameters.
```ts
interface ExtendedExpected<IsAsync = false> extends Expected<IsAsync> {
  not: IsAsync extends true ? Async<ExtendedExpected<true>> : ExtendedExpected<false>;
  resolves: Async<ExtendedExpected<true>>;
  rejects: Async<ExtendedExpected<true>>;
}
```

I've put together an exhaustive example of what I've gathered while using `expect.extend` and changed the scope of this PR to reflect more what's the intended goal

iuioiua:
PTAL @kt3k

expect/_extend_test.ts
```diff
@@ -110,3 +110,51 @@ Deno.test("expect.extend() api test case", () => {
...
+
+  // Alias expect to avoid having to pass the generic typing argument each time
+  // This is probably what you want to export and reuse across your tests
+  const myexpect = expect<ExtendedExpected>;
+
+  // Perform some tests
+  myexpect("🦕").toBeDinosaur();
+  myexpect("🦧").not.toBeDinosaur();
+  await myexpect(Promise.resolve("🦕")).resolves.toBeDinosaur();
+  await myexpect(Promise.resolve("🦧")).resolves.not.toBeDinosaur();
```
kt3k, line 155: Confirmed these are correctly type checked in `deno test` and in the editor. Nice work!

Screenshot 2024-07-16 at 12 59 46

kt3k (approved):
LGTM

PR #5310, opened 2024-07-04
refactor(yaml): rename util functions

timreichen:
- renames `_class()` to `getObjectTypeString()`
- renames `encodeHex()` to `charCodeToHexString()`
- renames `fromDecimalCode()` to `decimalCharCodeToNumber()`
- renames `fromHexCode()` to `hexCharCodeToNumber()`

~All these functions are moved into `_util.ts` to clean up `_loader.ts` and `_dumper.ts`~


yaml/_dumper.ts
```diff
@@ -193,17 +195,6 @@ export class DumperState {
   }
 }
 
-function encodeHex(character: number): string {
```
iuioiua, line 196: Let's avoid moving non-shared symbols to `./_utils.ts`. This'd otherwise make sense to do if these symbols were shared between various files.

iuioiua (changes requested):

yaml/_dumper.ts
timreichen: I would agree if the file was called something like `_shared.ts`. But it is `_utils.ts` and these really are util functions, no? It also includes `isBoolean()` for example which is not shared.

iuioiua: What is the purpose of moving symbols away from where they're used, unless they're shared?

timreichen: Tbh I think `_loader.ts` and `_dumper.ts` are really bloated. These are functions that don't take a state as an argument. I also noticed that some similar code is used in other mods (` const className = Object.prototype.toString.call(a);` in `expect/_equal.ts`, `isA()` is similar in `expect/_utils.ts`). Maybe we could even add a global utils file for these?

iuioiua: I agree - these files are bloated. But moving them elsewhere doesn't solve any problems. Also, while I agree similar code is used across packages, I think having as global utilities is more trouble that what it's worth, since they're small, more-trivial functions. Different story for less trivial functions, like those in `@std/internal`.

Either way, let's keep this PR focused on utility renames. They're good because they provide clearer descriptions of the function's purposes.

iuioiua (approved):
Nice! Thank you.

PR #5311, opened 2024-07-04
refactor(yaml): cleanup `inspectNode()`

timreichen:
- replaces `duplicatesIndexes` array with Set of objects
- reduces nesting and adds early returns


iuioiua (approved):
Nice cleanup. Thank you.

PR #5313, opened 2024-07-04
refactor(yaml): cleanup `detectType()`

timreichen:
Note: The if statement conditions seemed very odd. I am not sure, if the new conditions reflect the original. Please check.


kt3k (approved):
LGTM

kt3k:
`instanceOf` seems used from only 1 type (timestamp). Maybe we can replace it with `predicate`

PR #5314, opened 2024-07-04
refactor(yaml): cleanup `mergeMappings()`

timreichen:


iuioiua (approved):
Thank you.

iuioiua (approved):
Actually, I missed that we shouldn't switch to property assignment: https://github.com/denoland/deno_std/blob/f0057a87b26309588de51b3cf1bd623775c1cf18/yaml/parse_test.ts#L137-L142

Thanks for catching and reverting that.

timreichen:
> Thank you.

Turns out `Object.defineProperty` is used because sometimes a prototype is passed...

PR #5203, opened 2024-07-01
chore(path): release `path@1.0.0`

iuioiua:
To be merged July 4.

The latest API docs: https://jsr.io/@std/path@1.0.0-rc.3/doc

Closes #4922

PR #5320, opened 2024-07-05
BREAKING(path): remove `GlobToRegExpOptions`

iuioiua:
### What's changed

The `GlobToRegExpOptions` interface has been replaced with `GlobOptions`, which is identical in behavior. This change only affects users who explicitly import `GlobToRegExpOptions`.

### Motiviation

This change was made because `GlobToRegExpOptions` is identical to `GlobOptions`, and `GlobOptions` is already used in `normalizeGlob()`, `joinGlobs()`, and `expandGlob()` in `@std/fs`. This removal means a decrease in the API surface of `@std/path` without impeding functionality.

### Migration guide

To migrate, use `GlobOptions` instead of `GlobToRegExpOptions`. This is only needed if you already explicitly import `GlobToRegExpOptions`.

```diff
- import type { GlobToRegExpOptions } from "@std/path/glob-to-regexp";
+ import type { GlobOptions } from "@std/path/glob-to-regexp";

- const options: GlobToRegExpOptions = {
+ const options: GlobOptions = {
    extended: false,
    globstar: false,
    caseInsensitive: true,
  };

  // ..
```

### Related
Towards https://github.com/denoland/deno_std/pull/5203#issuecomment-2208846773

### DIFF
--- a/path/_common/glob_to_reg_exp.ts
+++ b/path/_common/glob_to_reg_exp.ts
@@ -1,7 +1,10 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-/** Options for {@linkcode globToRegExp}. */
+/**
+ * Options for {@linkcode globToRegExp}, {@linkcode joinGlobs},
+ * {@linkcode normalizeGlob} and {@linkcode expandGlob}.
+ */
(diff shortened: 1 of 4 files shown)


kt3k (approved):
LGTM. Nice clean up!

PR #5321, opened 2024-07-05
BREAKING(path): remove `FormatInputPathObject`

iuioiua:
### What's changed

`FormatInputPathObject` has been removed in favor of `Partial<ParsedPath>`, which is identical in behavior. This also introduces `@std/path/types`, where `ParsedPath` can be accessed.

### Motivation

This interface was removed because it's the same as `Partial<ParsedPath>`. By just using `Partial<ParsedPath>`, we can reduce the API surface area of `@std/path` without affecting behavior.

### Migration guide

To migrate, use `Partial<ParsedPath>` instead of `FormatInputPathObject`. This only affects users that explictly use `FormatInputPathObject`.

```diff
- import { FormatInputPathObject } from "@std/path";
+ import { ParsedPath } from "@std/path/types";

- const parsedPath: FormatInputPathObject = {
+ const parsedPath: Partial<ParsedPath> = {
    base: "file.txt",
    ext: ".txt",
    name: "file"
  }

  // ...
```

### Related

Towards https://github.com/denoland/deno_std/pull/5203#issuecomment-2208846773

### DIFF
--- a/path/_common/format.ts
+++ b/path/_common/format.ts
@@ -1,11 +1,11 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import type { FormatInputPathObject } from "../_interface.ts";
+import type { ParsedPath } from "../types.ts";
 
 export function _format(
   sep: string,
-  pathObject: FormatInputPathObject,
+  pathObject: Partial<ParsedPath>,
 ): string {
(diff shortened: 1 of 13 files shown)


iuioiua:
Might it also be worth renaming `ParsedPath` to `Path`? It's an interface, so I'd think it's obvious that it's parsed.

kt3k (approved):
Looks reasonable to me.

PR #5324, opened 2024-07-05
docs(yaml): improve documentation

iuioiua:


yaml/mod.ts
```diff
@@ -37,6 +32,11 @@
  * `);
  * ```
  *
+ * ## Limitations
+ * - `binary` type is currently not stable.
+ *
+ * For further examples see https://github.com/nodeca/js-yaml/tree/master/examples.
```
kt3k, line 38: Should we keep this reference?

kt3k (approved):
LGTM with nits

yaml/mod.ts
iuioiua: Good point. Removed.

PR #5325, opened 2024-07-05
fix(front-matter): improve `extract` types

### LINKED ISSUE #5323: Re-export the type for @std/front-matter
For example, in [@std/front-matter/yaml.ts](https://jsr.io/@std/front-matter/0.224.3/yaml.ts "@std/front-matter - JSR"), I'd like to import the type `Extract`, so that I can use it in my Fresh route's handler and component. In my understanding, async function shouldn't be used in component. As the `extract` function is an async one, I should put it in the handler, not component. Therefore I need a way to explicitly type the data. I can of course copy the type, but having them exported by default is more convenient.

kt3k:
Currently in `std/front-matter` package `extract` functions are typed as `Extractor` variable, and therefore it doesn't show type information of `@params`, `@returns`, etc.

This PR changes them to functions and improves the docs.

BEFORE
Screenshot 2024-07-05 at 13 42 35

AFTER
Screenshot 2024-07-05 at 13 43 45

This PR also exports `Extract` type (which is return type of `extract`) from each language file. This resolves #5323

### DIFF
--- a/front_matter/any.ts
+++ b/front_matter/any.ts
@@ -3,9 +3,15 @@
 import { createExtractor, type Parser } from "./_create_extractor.ts";
 import { parse as parseYaml } from "@std/yaml/parse";
 import { parse as parseToml } from "@std/toml/parse";
-import type { Extractor } from "./types.ts";
+import type { Extract, Extractor } from "./types.ts";
 
-export type { Extractor };
+export type { Extract, Extractor };
+
+const _extractor = createExtractor({
+  yaml: parseYaml as Parser,
+  toml: parseToml as Parser,
+  json: JSON.parse as Parser,
+});
 
 /**
  * Extracts and parses {@link https://yaml.org | YAML}, {@link https://toml.io |
@@ -28,9 +34,11 @@ export type { Extractor };
  * result.body; // "Hello, world!"
  * result.attrs; // { title: "Three dashes marks the spot" }
  * ```
+ *
+ * @typeParam T The type of the parsed front matter.
+ * @param text The text to extract front matter from.
+ * @returns The extracted front matter and body content.
  */
-export const extract: Extractor = createExtractor({
-  yaml: parseYaml as Parser,
-  toml: parseToml as Parser,
-  json: JSON.parse as Parser,
-});
+export function extract<T>(text: string): Extract<T> {
+  return _extractor(text);
+}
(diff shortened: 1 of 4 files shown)


front_matter/any.ts
```diff
@@ -28,9 +34,11 @@ export type { Extractor };
...
+ * @param text The text to extract front matter from.
+ * @returns The extracted front matter and body content.
  */
-export const extract: Extractor = createExtractor({
-  yaml: parseYaml as Parser,
-  toml: parseToml as Parser,
-  json: JSON.parse as Parser,
-});
+export function extract<T>(text: string): Extract<T> {
+  return _extractor(text);
```
iuioiua, line 43:
```suggestion
  return createExtractor({
    yaml: parseYaml as Parser,
    toml: parseToml as Parser,
    json: JSON.parse as Parser,
  })(text);
```

front_matter/any.ts
```diff
@@ -3,9 +3,15 @@
 import { createExtractor, type Parser } from "./_create_extractor.ts";
 import { parse as parseYaml } from "@std/yaml/parse";
 import { parse as parseToml } from "@std/toml/parse";
-import type { Extractor } from "./types.ts";
+import type { Extract, Extractor } from "./types.ts";
```
iuioiua, line 6: There no longer seems a reason to export `Extractor`.
```suggestion
import type { Extract } from "./types.ts";
```

front_matter/any.ts
```diff
@@ -3,9 +3,15 @@
...
 
-export type { Extractor };
+export type { Extract, Extractor };
```
iuioiua, line 8:
```suggestion
export type { Extract };
```

front_matter/any.ts
```diff
@@ -3,9 +3,15 @@
...
+
+const _extractor = createExtractor({
+  yaml: parseYaml as Parser,
+  toml: parseToml as Parser,
+  json: JSON.parse as Parser,
+});
```
iuioiua, line 14:
```suggestion
```

front_matter/any.ts
iuioiua, line 1: Ditto for the other files

iuioiua (changes requested):
IIUC, `Extractor` should no longer be public.

front_matter/any.ts
kt3k: `createExtractor` is relatively expensive. I don't think it's good idea to call it each time.

kt3k:
> Extractor should no longer be public.

I'll do that in a later PR with BREAKING tag.

front_matter/any.ts
iuioiua: Good point. Never mind.

iuioiua (approved):

PR #5327, opened 2024-07-05
refactor(yaml): cleanup dead code in `Mark` class

iuioiua:


yaml/_mark.ts
```diff
@@ -20,59 +24,42 @@ export class Mark {
...
-    if (!compact) {
-      snippet = this.getSnippet();
-
-      if (snippet) {
-        where += `:\n${snippet}`;
-      }
-    }
-
-    return where;
+  toString(): string {
```
iuioiua, line 61: The `compact` argument is never used.

yaml/_mark.ts
```diff
@@ -20,59 +24,42 @@ export class Mark {
     this.column = column;
   }
 
-  getSnippet(indent = 4, maxLength = 75): string | null {
-    if (!this.buffer) return null;
```
kt3k, line 24: Is it safe to remove this check?

iuioiua: Realistically, this is safe to remove, unless the user disables or misuses TypeScript. Either way, I've reverted this for now.

yaml/_mark.ts
```diff
@@ -20,59 +24,43 @@ export class Mark {
...
-  toString(compact?: boolean): string {
-    let snippet;
-    let where = "";
-
-    where += `at line ${this.line + 1}, column ${this.column + 1}`;
-
-    if (!compact) {
-      snippet = this.getSnippet();
-
-      if (snippet) {
```
kt3k, line 71: I think now we need to keep this condition

timreichen:
I kinda think we could just make `Mark` an interface inside `_error.ts` and combine `getSnippet()` and `toString()` into one `stringifyMark()` function. The mark class is not use elsewhere and the instance is only used once inside the error constructor...

iuioiua:
> I kinda think we could just make `Mark` an interface inside `_error.ts` and combine `getSnippet()` and `toString()` into one `stringifyMark()` function. The mark class is not use elsewhere and the instance is only used once inside the error constructor...

Yeah, let's do that in a following PR.

kt3k (approved):
LGTM

PR #5328, opened 2024-07-05
refactor(yaml): simplify `YamlError`

iuioiua:
Pre-requisite work for exposing `YamlError` in the public API.

### DIFF
--- a/yaml/_error.ts
+++ b/yaml/_error.ts
@@ -7,7 +7,7 @@ import type { Mark } from "./_mark.ts";
 
 export class YamlError extends Error {
   constructor(message: string, mark?: Mark) {
-    super(`${message} ${mark}`);
+    super(mark ? `${message} ${mark}` : message);
     this.name = this.constructor.name;
   }
 }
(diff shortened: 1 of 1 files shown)


kt3k:
> Pre-requisite work for exposing YamlError in the public API.

What do you think about `Mark` type dependency of `YamlError`? Probably we don't want to expose `Mark` public?

iuioiua:
Yeah, best we keep `Mark` private. Either way, that can be figured out independently of this PR.

timreichen:
Should we split `YamlError` into `ParseError extends SyntaxError` and `StringifyError extends Error` like in csv?

yaml/_error.ts
```diff
@@ -6,15 +6,8 @@
 import type { Mark } from "./_mark.ts";
 
 export class YamlError extends Error {
-  constructor(
-    message = "(unknown reason)",
-    protected mark: Mark | string = "",
-  ) {
+  constructor(message: string, mark?: Mark) {
     super(`${message} ${mark}`);
```
kt3k, line 10: When mark is not passed, now the error message becomes something like `YamlError: message undefined` instead of `YamlError: abc`

iuioiua:
> Should we split `YamlError` into `ParseError extends SyntaxError` and `StringifyError extends Error` like in csv?

Let's discuss in #5340.

kt3k (approved):
LGTM

PR #5332, opened 2024-07-05
refactor(yaml): move `getObjectTypeString()`

timreichen:
- moves `getObjectTypeString()` to `_utils.ts` as it can be shared.


kt3k:
Can you also replace the usages in `yaml/_type/omap.ts` and `yaml/_type/pairs.ts`?

iuioiua (approved):
Thanks again, Tim.

PR #5333, opened 2024-07-05
refactor(yaml): cleanup `writeScalar()`

timreichen:
- removes nested function in favour of string returning function.
- renames argument `iskey` to `isKey`
- general simplifications

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -523,77 +523,63 @@ function blockHeader(string: string, indentPerLevel: number): string {
 //    • No ending newline => unaffected; already using strip "-" chomping.
 //    • Ending newline    => removed then restored.
 //  Importantly, this keeps the "+" chomp indicator from gaining an extra line.
-function writeScalar(
+function stringifyScalar(
   state: DumperState,
   string: string,
   level: number,
-  iskey: boolean,
+  isKey: boolean,
 ) {
-  state.dump = ((): string => {
-    if (string.length === 0) {
-      return "''";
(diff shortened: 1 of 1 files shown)


yaml/_dumper.ts
```diff
@@ -523,77 +523,63 @@ function blockHeader(string: string, indentPerLevel: number): string {
 //    • No ending newline => unaffected; already using strip "-" chomping.
 //    • Ending newline    => removed then restored.
 //  Importantly, this keeps the "+" chomp indicator from gaining an extra line.
-function writeScalar(
```
kt3k, line 526: These functions are consistently named `write<Thing>`. I don't think it's a good idea to break that consistency

kt3k: Also this change breaks the practice to assign to .dump field in each `write*` function. I'm not in favor of breaking that either.

timreichen: Yes, I am aware of that. I think we might want to change more with other prs, because `state.dump` is not very well defined. The type changes in different functions, with type any it is hard to follow and debug.

kt3k: > The type changes in different functions, with type any it is hard to follow and debug.

I think that is the reason to not change that until we write more robust testing because it's too risky to change, and also too difficult to validate that change is not breaking. `_dumper.ts` currently has 262 lines uncovered by testing. Can you first contribute to improve the test coverage?

timreichen:
closing as not planned as of now.

PR #5335, opened 2024-07-05
refactor(yaml): simplify `indentString()`

timreichen:
- replaces char loop with `split().map().join()`
- adds tests


iuioiua (changes requested):
Can you please instead test this functionality by testing `parse()`? We want to make sure everything works through the call stack. Either way, great work so far. Thank you.

yaml/stringify_test.ts
iuioiua, line 1: Can confirm that these tests pass with the previous implementation also.

yaml/_dumper_test.ts
iuioiua, line 1: This change is desirable, but should be part of a different PR. Please note for next time.

iuioiua (approved):
Thank you.

PR #5336, opened 2024-07-05
refactor(yaml): simplify and rename `dropEndingNewline()`

timreichen:
- renames `dropEndingNewline()` to `trimTrailingNewline()`
- uses `.at(-1)`
- adds tests

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -431,9 +431,8 @@ function foldLine(line: string, width: number): string {
   return result.slice(1); // drop extra \n joiner
 }
 
-// (See the note for writeScalar.)
-function dropEndingNewline(string: string): string {
-  return string[string.length - 1] === "\n" ? string.slice(0, -1) : string;
+function trimTrailingNewline(string: string) {
+  return string.at(-1) ? string.slice(0, -1) : string;
 }
 
 // Note: a long line without a suitable break point will exceed the width limit.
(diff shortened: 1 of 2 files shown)


yaml/_dumper.ts
```diff
@@ -578,13 +577,11 @@ function writeScalar(
         return `'${string.replace(/'/g, "''")}'`;
       case STYLE_LITERAL:
         return `|${blockHeader(string, state.indent)}${
-          dropEndingNewline(
-            indentString(string, indent),
-          )
+          trimTrailingNewline(indentString(string, indent))
```
iuioiua, line 580: Anything wrong with this?
```suggestion
          indentString(string, indent).trimEnd()
```

timreichen: I think this will also trim whitespaces. Not sure if that is a problem though.

iuioiua: Yeah, it'll trim all whitespaces, and I think that's fine. @kt3k, see any issues with that?

kt3k: That will break the case when the string has white spaces at the end.

(The 2nd case of https://github.com/denoland/deno_std/pull/5341)

kt3k (approved):

yaml/_dumper.ts
iuioiua: I know, but in what cases is that an issue? Genuine question.

kt3k: When the user's input has trailing whitespace, then that part needs to be serialized as is. Trimming that whitespace means the user loses some data while serializing.

iuioiua: Oh, yes, that's right. Obvious once you say it 😆

PR #5337, opened 2024-07-06
feat(yaml): add tests for `foldLine()`

timreichen:


iuioiua (changes requested):
Again, can we please test this functionality by testing `stringify()`? And again, great work so far. Thank you.

timreichen:
> Again, can we please test this functionality by testing `stringify()`? And again, great work so far. Thank you.

I think we should test separate functions like in toml with combinators. Tests can be more on point no?

iuioiua:
We should test functions how users would use them. It guarentees not just that that underlying functionality works, but that said functionality is plumbed correctly.

timreichen:
Closing in favour of already merged tests

PR #5343, opened 2024-07-07
refactor(toml): remove `Patterns` object

timreichen:
- replaces Pattern object with regexp constants


iuioiua (approved):
Nice. Thank you.

PR #5344, opened 2024-07-07
perf(encoding): fix loop times in `encodeHex()`

zsokami:


kt3k (approved):
LGTM. Thanks!

kt3k:
I changed the commit tag to `perf` as the current behavior doesn't cause any bug, but it caused unnecessary perf overhead, which this PR fixes.

PR #5345, opened 2024-07-08
fix(fs): accept `exts` without leading period in `walk[Sync]()`

iuioiua:
Previously, `walk[Sync]()` would return files and folders ending with strings, as defined in the `exts` option. E.g. It'd return the `asserts` folder when setting `exts: ["ts"]`. This is unexpected behaviour, as one would not expect folders to be turned when setting file extensions.

This change adds the leading period, `.`, to any extensions defined in `exts`, if missing, as there doesn't appear to be any good reason to require the user to have this leading period.

### DIFF
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -136,7 +136,9 @@ export interface WalkOptions {
    * If specified, entries without the file extension specified by this option
    * are excluded.
    *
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

Issue #5340, opened 2024-07-07
suggestion: replace `YamlError` with native errors


iuioiua:
Should we split `YamlError` into `ParseError extends SyntaxError` and `StringifyError extends Error` like in csv?

_Originally posted by @timreichen in https://github.com/denoland/deno_std/issues/5328#issuecomment-2212350787_

timreichen:
On that topic: There are two standards now: Throwing native errors (msgpack, jsonc, ini, csv, encoding, semver), and custom errors (dotenv, toml, yaml). 
Maybe we could have all std parse/stringify mods behave the same and therefore more predictably. Shall we get rid of `TOMLError` and `YamlError` completely and replace them with native errors?

iuioiua:
Well, we throw custom errors in `assert`, `async`, `csv`, `dotenv`, `fs`, `io`, `testing`, `toml` and `yaml`. It's a good suggestion, but I'm on the fence. Can you please open an issue with the suggestion? It'd be good to hear other opinions.

iuioiua:
I'm not yet certain that we should use native errors in all cases, but I think we should at least align with [`JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) and [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) in their exceptions, for packages that parse and stringify. These exceptions are [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) and [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError), respectively.

PR #5346, opened 2024-07-08
BREAKING(toml): throw `SyntaxError` in `parse()` instead of `TOMLParseError`

iuioiua:
### What's changed

`TOMLParseError` has been removed in favor of [`SynaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) as the error thrown in [`parse()`](https://jsr.io/@std/toml/doc/~/parse).

### Motivation

This change aligns `parse()` with [`JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse) in its exceptions. The benefit of doing this is that a standard error class is used that fits the use case without having an extraneous custom error class.

### Migration guide

To migrate, compare against `SyntaxError` instead of `TOMLParseError`, when error-handling `parse()`.

```diff
- import { parse, TOMLParseError } from "@std/toml/parse";
+ import { parse } from "@std/toml/parse";

try {
  parse("[1, 2, 3");
} catch (error) {
- if (error instanceof TOMLParseError) {
+ if (error instanceof SyntaxError) {
    // ...
  }
}
```

### Related

Towards https://github.com/denoland/deno_std/issues/5340#issuecomment-2212437577


kt3k:
If there's no strong concern/objection, we'll probably land this before the stabilization.

lucacasonato (approved):
LGTM

PR #5349, opened 2024-07-08
chore(net): release `net@1.0.0-rc.1`

kt3k:
towards #5006


iuioiua (approved):

iuioiua:
Now that `getAvailablePort()` has been found to have valid use cases, stabilizing `@std/net` with it seems fine.

PR #5354, opened 2024-07-08
fix(http): allow deleting cookie with `secure`, `httpOnly` and `partitioned` attributes

### LINKED ISSUE #5049: deleteCookie doesn't allow deleting cookies with the SameSite, Secure or unparsed attributes set
**Describe the bug**
When setting a cookie with the attributes "Secure", "SameSite", or any other attribute that would scope the cookie to a narrower scope, `deleteCookie` will generate a cookie that doesn't delete that cookie.

**Steps to Reproduce**
1. Set a cookie with the attributes "SameSite" (set to "Strict"), "Partitioned" and "Secure". (I don't remember the specific attribute that has this effect; Partitioned is what I'm concerned about and it doesn't matter for this bug report)
2. Attempt to delete the cookie using `deleteCookie`, noticing that you can only specify the attributes `path` and `domain`.
3. Observe that the cookie sent by Deno is missing the attributes "SameSite", "Secure" and "Partitioned".
4. Observe that the browser refuses to delete the cookie.

**Expected behavior**

Deno allows deleting cookies with "SameSite", "Secure", etc. specified.

**Environment**

This isn't environment specific (although some browsers might be more forgiving when validating cookies).

- OS: Debian 12
- deno version: 1.44.2
- std version: 0.216.0

kt3k:
closes #5049

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -299,7 +299,10 @@ export function setCookie(headers: Headers, cookie: Cookie) {
 export function deleteCookie(
   headers: Headers,
   name: string,
-  attributes?: Omit<Cookie, "name" | "value" | "sameSite">,
+  attributes?: Pick<
+    Cookie,
+    "path" | "domain" | "secure" | "httpOnly" | "partitioned"
+  >,
 ) {
   setCookie(headers, {
     name: name,
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -273,7 +273,8 @@ export function setCookie(headers: Headers, cookie: Cookie) {
 /**
  * Set the cookie header with empty value in the headers to delete it.
  *
- * The attributes (path, domain, secure, httpOnly, partitioned) need to match the values when the cookie was set.
+ * The attributes (`path`, `domain`, `secure`, `httpOnly`, `partitioned`) need
+ * to match the values when the cookie was set.
  *
  * > Note: Deleting a `Cookie` will set its expiration date before now. Forcing
  * > the browser to delete it.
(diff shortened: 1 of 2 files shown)


iuioiua (approved):

http/cookie.ts
```diff
@@ -296,7 +299,7 @@ export function setCookie(headers: Headers, cookie: Cookie) {
 export function deleteCookie(
   headers: Headers,
   name: string,
-  attributes?: { path?: string; domain?: string },
+  attributes?: Omit<Cookie, "name" | "value">,
```
kt3k, line 302: I don't think this is correct. As far as I tested, `sameSite` attribute didn't have effect in `deleteCookie` (You can delete sameSite=Lax cookie without specifying `sameSite=Lax` when deleting it)

iuioiua: Fixed. PTAL.

kt3k: LGTM

likonefokou:
Still don't delete cookies 😅

http/cookie.ts
likonefokou: > I don't think this is correct. As far as I tested, `sameSite` attribute didn't have effect in `deleteCookie` (You can delete sameSite=Lax cookie without specifying `sameSite=Lax` when deleting it)

It work on localhost, but don't when deployed. Using Deno Deploy

kt3k:
What version of `std/http` do you use? This fix is only available in the latest RC version (`@std/http@1.0.0-rc.3`)

likonefokou:
II was using version 0.224.0, i've changed to the version you indicated and it's now working. Thanks !

Issue #5034, opened 2024-06-12
`describe` should throw if passed a function that returns a promise


lucacasonato:
`describe` is used to group related tests. Actual test code should be put in an `it` block or a `before*`/`after*` hook. For example, you can initialize `node` in a `beforeAll` hook:

```js
import { beforeAll, describe, it } from "jsr:@std/testing@^0.220.1/bdd";
import { assertEquals } from "jsr:@std/assert@^0.220.1/assert_equals";

describe("DOMNode", () => {
  let node;

  beforeAll(async () => {
    node = await xmlFromString(…);
  });

  it("in root namespace", () => assertEquals(…));
  it("in leaf namespace", () => assertEquals(…));
});
```

<details><summary><code>Deno.test</code> equivalent</summary>

```js
import { assertEquals } from "jsr:@std/assert@^0.220.1/assert_equals";

Deno.test("DOMNode", async (t) => {
  let node = await xmlFromString(…);

  await t.step("in root namespace", () => assertEquals(…));
  await t.step("in leaf namespace", () => assertEquals(…));
});
```

</details>

The error message could definitely be improved. [Jest throws a helpful error](https://github.com/jestjs/jest/blob/d4d1f2b8004b1eb4bf0cf862698dd1142e13278f/packages/jest-circus/src/index.ts#L72-L82) if the callback passed to `describe` returns a promise or anything other than `undefined`.

_Originally posted by @0f-0b in https://github.com/denoland/deno/issues/22991#issuecomment-2010379940_

kt3k:
Hey @KyleJune, can you take a look at this?

KyleJune:
> Hey @KyleJune, can you take a look at this?

All the call signatures already only accept callbacks that return void. I believe the issue would have been caught if they were using TypeScript but I think it would be fine to add a check to the code so that JavaScript users will get an error if they try having it return anything.

kt3k:
`jest` seems throwing with async describe function, but, on the other hand, `vitest` seems supporting async describe definition. Probably the issue author of https://github.com/denoland/deno/issues/22991 expected vitest-compatible behavior here?

Also there is another request to make it vite-compatible #4634

I wonder if there's any downside to support vite-compatible behavior.

> All the call signatures already only accept callbacks that return void.

That's true in type definitions, but assigning `() => Promise<void>` to `() => void` doesn't seem causing type error, and therefore `describe("", async () => {})` doesn't cause type error. Is there any way to make this a type error?

[...]

PR #5355, opened 2024-07-08
fix(testing): cause type error for async function as describe body

kt3k:
Partially addresses #5034 

This changes uses `() => void | undefined` type for describe body type instead of `() => void`, which causes type error if the user assigned async functions to it.

This will prevents the confusion like https://github.com/denoland/deno/issues/22991


iuioiua (approved):
Nice

PR #5356, opened 2024-07-08
test(yaml): test parsing of single quoted scalars

kt3k:
part of #3713 

This PR add test cases that exercises the parsing of single quoted strings.

This improves the coverage of `yaml/_loader.ts` from 82.51% to 84.76%.


iuioiua (approved):
Nice

PR #5367, opened 2024-07-09
BREAKING(front-matter): remove `unknown` from `Format` union type

timreichen:
- removes `unknown` as a format and makes handles all invalid formats the same.


iuioiua (approved):
Nice catch! Thank you.

kt3k (approved):
LGTM

PR #5370, opened 2024-07-09
refactor(front-matter): replace regexp factory with regexp literals

timreichen:
- replaces `createRegExps` with regexp literals


kt3k:
This looks harder to maintain to me.

timreichen:
I think the building step is more complex and harder to understand than simple regexes, as they are very straight forward. If you'd like I could add names to capturing groups.

front_matter/_formats.ts
```diff
@@ -1,60 +1,25 @@
...
 
-export const EXTRACT_REGEXP_MAP = {
+const EXTRACT_YAML_REGEXP =
+  /^\ufeff?(?:---yaml|= yaml =|---)$([\s\S]+?)^(?:---|= yaml =)\s*$\r?\n?/im;
+const EXTRACT_TOML_REGEXP =
+  /^\ufeff?(?:---toml|\+\+\+|= toml =)$([\s\S]+?)^(?:---|\+\+\+|= toml =)\s*$\r?\n?/im;
+const EXTRACT_JSON_REGEXP =
+  /^\ufeff?(?:---json|= json =)$([\s\S]+?)^(?:---|= json =)\s*$\r?\n?/im;
+
+export const EXTRACT_REGEXP_MAP: Record<Format, RegExp> = {
```
iuioiua, line 22: Sidenote: we should change these to `Map`s.

iuioiua (approved):
While I wouldn't call the `EXTRACT_*` constants "simple regexes", this is simpler and easier to understand to me, compared to regex factories.

kt3k:
Repetition of the same pattern (such as `---toml|\+\+\+|= toml =`) in multiple places doesn't look great to me. It feels like a error-prone.

front_matter/_formats.ts
```diff
@@ -1,60 +1,25 @@
...
-    "= toml =",
-  ],
-);
-const [RECOGNIZE_JSON_REGEXP, EXTRACT_JSON_REGEXP] = createRegExps(
-  [
-    ["---json", "---"],
-    "= json =",
-  ],
-);
+const RECOGNIZE_YAML_REGEXP = /^---yaml|= yaml =|---$/im;
```
kt3k, line 5: I think these regexes need parens.
```suggestion
const RECOGNIZE_YAML_REGEXP = /^(?:---yaml|= yaml =|---)$/im;
```

timreichen:
> Repetition of the same pattern (such as `---toml|\+\+\+|= toml =`) in multiple places doesn't look great to me. It feels like a error-prone.

I think the repetition is justified because they are two different regexps, one for recognizing and one for extracting.

Btw, I think the extract regexes could be written without the multiline flag:

```ts
const EXTRACT_YAML_REGEXP =
  /^\ufeff?(?:---yaml|= yaml =|---)([\s\S]+?\r?\n)(?:---|= yaml =)\s*\r?\n?/i;
const EXTRACT_TOML_REGEXP =
  /^\ufeff?(?:---toml|\+\+\+|= toml =)([\s\S]+?\r?\n)(?:---|\+\+\+|= toml =)\s*\r?\n?/i;
const EXTRACT_JSON_REGEXP =
  /^\ufeff?(?:---json|= json =)([\s\S]+?\r?\n)(?:---|= json =)\s*\r?\n?/i;
  ```

kt3k:
> I think the repetition is justified because they are two different regexps, one for recognizing and one for extracting.

I don't understand this reasoning. Why is it justified because of 2 regexps? Current code reuses the same pattern for 2 regexps. Isn't that an ideal state?

> `/^\ufeff?(?:---yaml|= yaml =|---)([\s\S]+?\r?\n)(?:---|= yaml =)\s*\r?\n?/i;`

This example has error in closing part. Feels like proving it's an error-prone.

timreichen:
Ok, what about creating the regexps with string literals like we do in semver?

iuioiua:
> Ok, what about creating the regexps with string literals like we do in semver?

Let's try that.

iuioiua (changes requested):
I've added a temporary test (for the lifetime of this PR) that is a proof that the old constants are equivalent to the new ones. It's currently failing. Please ensure this test passes.

iuioiua (approved):
LGTM, as these constants seem easier to follow, compared to the previous functions, and have been shown to be equivalent in https://github.com/denoland/std/pull/5370/commits/11b5f3edc8668917a5ccb8c2e2ae3f848340939e.

@timreichen, frankly, we'll be reluctant to accept any further simplifications to these regexes. Getting regexes are hard, and can very subtle mistakes can lead to bugs. Nevertheless, we can take a look at a PR, but no guarentees in it being merged.

Either way, thank you.

PR #5371, opened 2024-07-10
docs(path): add examples to `@std/path/posix` examples

iuioiua:
This documentation aims at providing sufficient guidance to justify removing `@std/url`.

Pre-requisite for #5530
Addresses https://github.com/denoland/std/pull/4992#pullrequestreview-2166480173


kt3k (approved):
LGTM

kt3k:
I think we should also explore the idea (from @lucacasonato ) of making some of `path` APIs directly accept file URL object and file URL strings like some of FS APIs in Deno CLI:

```ts
dirname("file:///path/to/dir/file"); //=> returns `/path/to/dir` on linux
dirname("file:///C:/path/to/dir/file"); // => returns `C:\path\to\dir` on windows
```

I think this mostly covers the motivation of `std/url`

iuioiua:
Let's explore in a separate issue 🙂

kt3k:
created #5537

PR #5375, opened 2024-07-10
chore: use new format for workspaces and re-enable using Deno canary and v1.x

kt3k:


fs/empty_dir_test.ts
```diff
@@ -207,8 +207,6 @@ for (const s of scenes) {
           "--no-lock",
           "--quiet",
           "--no-prompt",
-          "--config",
```
iuioiua, line 210: Nice to see these no longer needed.

iuioiua (approved):
Nice!

iuioiua:
Merging as to unblock failures on `main`.

Issue #5194, opened 2024-06-29
to-do: automate checking of package versions


iuioiua:
The package version defined in each package's `deno.json` file should match that of the one defined in the root `deno.json`. We should write a script that automates this check.

PR #5377, opened 2024-07-10
chore: fix path version range in import map

kt3k:


iuioiua (approved):
Good catch. We really need to automate checking these values (#5194).

PR #5380, opened 2024-07-10
chore: check import map range specifier

### LINKED ISSUE #5194: to-do: automate checking of package versions

kt3k:
closes #5194

This PR adds `lint:import-map` task, which check that the version ranges of std specifiers are in the form of ^x.y.z. (It errors with `@1.0.0` for example.)


iuioiua (approved):
Nice

PR #5381, opened 2024-07-10
fix(streams): toText() incorrect with multibyte strings in different chunks

leonco:
toText() should return complete UTF-8 string.


kt3k (approved):
Good catch! Thank you.

The usage of `TextDecoder` was wrong.

PR #5382, opened 2024-07-10
BREAKING(csv): rename `bom` option to `byteOrderMark`

timreichen:
- rename `bom` to `byteOrderMark` in `StringifyOptions`

### DIFF
--- a/csv/stringify.ts
+++ b/csv/stringify.ts
@@ -119,7 +119,7 @@ export type StringifyOptions = {
    *
    * @default {false}
    */
-  bom?: boolean;
+  byteOrderMark?: boolean;
 };
 
 const QUOTE = '"';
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Makes sense to me. Though, I'd understand an argument against this change that "bom" for "byte order mark" might be well-understood. E.g. There's the [`ignoreBom`](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#ignorebom) option for `TextDecoder`. Happy to go with what Yoshiya decides.

kt3k:
> There's the [ignoreBom](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder/TextDecoder#ignorebom) option for TextDecoder

Good point. `BOM` acronym seems just fine. Let's not do this.

PR #5385, opened 2024-07-10
fix(testing): throw error when async func is passed to describe

### LINKED ISSUE #5034: `describe` should throw if passed a function that returns a promise

kt3k:
closes #5034

This PR updates the behavior of `describe` in `testing/bdd`. Now `describe` throws if an async function is passed to it.

### DIFF
--- a/testing/_test_suite.ts
+++ b/testing/_test_suite.ts
@@ -97,7 +97,13 @@ export class TestSuiteInternal<T> implements TestSuite<T> {
       const temp = TestSuiteInternal.current;
       TestSuiteInternal.current = this;
       try {
-        fn();
+        // deno-lint-ignore no-explicit-any
+        const value = fn() as any;
+        if (value instanceof Promise) {
+          throw new Error(
+            'Returning a Promise from "describe" is not supported. Tests must be defined synchronously.',
+          );
+        }
       } finally {
         TestSuiteInternal.current = temp;
       }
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Nice

PR #5388, opened 2024-07-10
refactor(yaml): move state functions inside class

timreichen:
- moves all state modifying functions inside state class
- modifies `generateNextLine()` and `testImplicitResolving()` arguments

Note: The goal is to separate functions that modify the state and functions that do not.


iuioiua (approved):
Thanks

PR #5389, opened 2024-07-10
refactor(yaml): remove nested functions

timreichen:


yaml/_dumper.ts
```diff
@@ -533,7 +533,7 @@ function writeScalar(
   level: number,
   isKey: boolean,
 ) {
-  state.dump = ((): string => {
+  const createDump = () => {
```
iuioiua, line 536: We typically reserve the arrow function syntax for anonymous functions. Changing...

iuioiua:
Please rebase

yaml/_dumper.ts
timreichen: Yes, but since this functions is now a method of DumperState, you would need to bind `this` with a regular function, right?

iuioiua (approved):

PR #5390, opened 2024-07-10
refactor(yaml): replace `indexOf()` tests with `includes()`

timreichen:


iuioiua (approved):
Nice. Thanks. Are you also able to include similar changes for `yaml/_type/int.ts`?

iuioiua (approved):

PR #5394, opened 2024-07-10
refactor(yaml): make `Type` generic

timreichen:
- adds generic to `Type`
- removes some `any` types


iuioiua (changes requested):
Great idea! Couple of nits.

yaml/_type/set.ts
```diff
@@ -5,8 +5,7 @@
 
 import type { Type } from "../_type.ts";
 
-// deno-lint-ignore no-explicit-any
-function resolveYamlSet(data: any): boolean {
+function resolveYamlSet(data: Record<PropertyKey, unknown>): boolean {
```
kt3k, line 8:
```suggestion
function resolveYamlSet(data: ResultType | null): boolean {
```

timreichen: Actually, that doesn't work, because result type could also be an object or string.

yaml/_type.ts
```diff
@@ -18,20 +18,22 @@ export type StyleVariant =
...
-// deno-lint-ignore no-explicit-any
-export type RepresentFn = (data: any, style?: StyleVariant) => any;
 
-export interface Type {
+export type RepresentFn<D> = (data: D, style?: StyleVariant) => string;
+
+type Constructor<T> = new (...args: unknown[]) => T;
+
+// deno-lint-ignore no-explicit-any
+export interface Type<D = any> {
```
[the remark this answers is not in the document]
timreichen: There are a lot of properties in `Schema` that use Type[]

yaml/_type/set.ts
kt3k: I think we should revert this to any as we don't know what types are actually passed here.

timreichen: It must be an object or `for (const key in data) {...}` would throw no?

yaml/_type.ts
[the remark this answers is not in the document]
iuioiua: Whoops. Alright.

iuioiua (approved):

PR #5395, opened 2024-07-10
refactor(yaml): fix `onWarning` arg type

timreichen:


iuioiua (approved):
Good spot. Sidenote: this reminds me that we need to solve #5340.

iuioiua (changes requested):
Actually, can you please also make the same change to `ParseOptions.onWarning`?

iuioiua (approved):
Thank you

PR #5400, opened 2024-07-10
refactor(yaml): replace writeFlowSequence any type

timreichen:


yaml/_dumper.ts
```diff
@@ -601,8 +601,7 @@ function writeScalar(
 function writeFlowSequence(
   state: DumperState,
   level: number,
-  // deno-lint-ignore no-explicit-any
-  object: any,
+  object: unknown[],
```
iuioiua, line 604: Now that this is a `DumperState` method, and `state.dump` is the only value ever passed as `object`, why not just remove this last argument and use `this.dump` within this method?

timreichen:
closing in favour of https://github.com/denoland/deno_std/pull/5413

PR #5401, opened 2024-07-10
refactor(yaml): replace writeBlockSequence any type

timreichen:


yaml/_dumper.ts
```diff
@@ -622,8 +622,7 @@ function writeFlowSequence(
 function writeBlockSequence(
   state: DumperState,
   level: number,
-  // deno-lint-ignore no-explicit-any
-  object: any,
+  object: unknown[],
```
iuioiua, line 625: Again, this argument could be removed and we could instead use `this.dump` within the method.

timreichen:
closing in favour of https://github.com/denoland/deno_std/pull/5413

PR #5405, opened 2024-07-10
BREAKING(csv): remove `ParseError`

timreichen:
### What's changed

`ParseError`, which is thrown in `parse()`, has been removed and replaced in favor of [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError). The `startLine`, `line` and `column` properties have been removed also. Error messages have remained unchanged.

### Motivation

This change was made to align `parse()` with other `parse()` functions in the Standard Library, and [`JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse), which throw `SyntaxError`, and without any special properties. This all helps with providing consistent behavior across the Standard Library.

### Migration guide

To migrate, compare against `SyntaxError` instead of `ParseError` when error-handling `parse()` or `CsvParseStream`.

```diff
- import { parse, ParseError } from "@std/csv/parse";
+ import { parse } from "@std/csv/parse";

try {
  parse(`a,b,🐱"`);
} catch (error) {
- if (error instanceof ParseError) {
+ if (error instanceof SyntaxError) {
    // ...
  }
}
```


csv/parse_stream_test.ts
```diff
@@ -49,11 +48,11 @@ Deno.test({
...
     const error = await assertRejects(() => reader.read());
-    assert(error instanceof ParseError);
-    assertEquals(error.startLine, 4);
-    assertEquals(error.line, 5);
-    assertEquals(error.column, 0);
-    assertStringIncludes(error.message, ERR_QUOTE);
+    assert(error instanceof SyntaxError);
+    assertStringIncludes(
+      error.message,
+      `record on line 4; parse error on line 5, column 0: extraneous or missing " in quoted-field`,
```
iuioiua, line 54: The error class and message can be checked in the first and second arguments in the `assertRejects` above.

iuioiua (approved):
Just found a nit. Otherwise, LGTM! Thanks for this. @kt3k, PTAL.

iuioiua:
Also, it does seem like some of these changes could've been done in a separate PR. Not a deal big enough to change now or block this PR, but more for future reference.

csv/parse_stream_test.ts
```diff
@@ -48,12 +42,11 @@ Deno.test({
     const reader = readable.getReader();
     assertEquals(await reader.read(), { done: false, value: ["id", "name"] });
     assertEquals(await reader.read(), { done: false, value: ["1", "foo"] });
-    const error = await assertRejects(() => reader.read());
-    assert(error instanceof ParseError);
-    assertEquals(error.startLine, 4);
-    assertEquals(error.line, 5);
-    assertEquals(error.column, 0);
```
kt3k, line 55: These information will be unavailable. Is that fine?

WDYT? cc @iuioiua

iuioiua: Ah, true. We can store this data within [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).

kt3k: > We can store this data within [Error.cause](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).

Sounds a reasonable alternative to me. That use case is recommended in the 2nd example https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#providing_structured_data_as_the_error_cause

But I'm a bit concerned about the unavailability of the type. `err.cause` seems always `unknown` type. The user need to check type and properties of `cause` before accessing it. On the other hand, if it's kept `ParseError`, these properties can be accessed after the user checked `err instanceof ParseError`, which feels a bit easier and clearer to me. What do you think?

iuioiua: Good point. Actually, now, I'm wondering whether it's worth having those extra details at all. `JSON.parse` doesn't do this. E.g. `JSON.parse("[1,2,3,")` throws an error with `.cause` undefined. We might be fine without. I suggest removing these properties. If it's a sufficiently significant issue in the future, we can revisit. WDYT?

kt3k: Fair enough. Let's try this and see the community reactions.

kt3k (approved):
LGTM

iuioiua (approved):
LGTM once the PR description is complete with what's changed, motivations, and migration guide.

Issue #3472, opened 2023-06-30
`assertObjectMatch` error message diff prints arrays as objects


vwkd:
**Describe the bug**

The error message diff in `assertObjectMatch` prints arrays as objects.

This happens if the value has potentially nested property with an array or if the value itself is an array.

**Steps to Reproduce**

```js
import { assertObjectMatch } from "https://deno.land/std@0.191.0/testing/asserts.ts";

Deno.test("foo", () => {
  const act = { foo: [] };

  const exp = {
    foo: [
      "bar",
    ],
  };

  assertObjectMatch(act, exp);
});
```

**Expected behavior**

```txt
    [Diff] Actual / Expected


    {
+     foo: [
+       "bar",
+     ],
-     foo: [],
    }
```

**Actual behavior**

```txt
    [Diff] Actual / Expected


    {
+     foo: {
+       "0": "bar",
+     },
-     foo: {},
    }
```

**Environment**

- OS: macOS 12.6.6
- deno version: deno 1.34.1
- std version: 0.191.0

hasundue:
Happened to me too.

```
deno 1.36.0 (release, x86_64-unknown-linux-gnu)
v8 11.6.189.12
typescript 5.1.6
```

PR #5408, opened 2024-07-11
fix(assert): `assertObjectMatch()` does not print objects correctly

babiabeo:
Fixes: #3472

- Currently, when filtering arrays, `assertObjectMatch()` uses `fn({ ...value }, { ...subset })`, which converts arrays into objects. Consequently, arrays are displayed as objects. This PR adds `filterArray()` to handle arrays.

Fixes: #3771

- When filtering objects, if `actual` and `expected` don't have the same keys or symbols, the `fn()` currently returns an empty object. It leads to missing properties if both are not empty. This PR adds logic so that it returns the keys and symbols in `actual` instead of an empty one. (also rename `fn()` to `filterObject()`)


iuioiua:
Are you able to split this into separate PRs? There are a few moving parts and I don't want to miss anything.

babiabeo:
Sure 👍

PR #5409, opened 2024-07-11
refactor(yaml): add writeNode options

timreichen:
- replaces boolean params with options object for more clarity.
- removes default values


kt3k (approved):
This seems better for readability to me.

LGTM

PR #5413, opened 2024-07-11
refactor(yaml): remove passing `this.dump` as method param

timreichen:
Note: We might want to re-introduce param passing for type security again at a later point, since this shifts the problem to the class property `dump`, which type is `any` atm.


iuioiua (approved):
Nice cleanup. Thanks!

PR #5414, opened 2024-07-11
test(testing): fix flakiness of snapshot test

kt3k:
This PR fixes the flakiness of `assertSnapshot() - different directory` test case.

In the test case, 2 snapshot updates happen, but the order of them seems racy now.

This change masks the count of these 2 updates, instead checks them independently.


iuioiua (approved):
Nice

PR #5415, opened 2024-07-11
refactor(collections): inline `randomInteger` utility function

timreichen:
- inlines `randomInteger()` as it is not used elsewhere.


iuioiua (approved):
Good spot. Thanks!

PR #5417, opened 2024-07-11
refactor(collections): use `Set.prototype.intersection()` method in `intersect()`

timreichen:
- reduces code by using `Set.prototype.intersection()`.


iuioiua:
I like this PR. However, it begs a few questions around browser compatibility and breaking changes, now that `@std/collections` is stable. Please allow me and Yoshiya to discuss this and get back to you on this one.

iuioiua (approved):
Talked with Yoshiya - this seems fine to do as there's sufficient browser support. LGTM.

PR #5304, opened 2024-07-04
test(csv): add grapheme length test

timreichen:
- reverts https://github.com/denoland/deno_std/pull/5298#issuecomment-2208355438
- adds test

PR #5421, opened 2024-07-12
chore(csv): Rename `graphemeLength` to `codePointLength`

lionel-rowe:
Further to https://github.com/denoland/deno_std/pull/5304:

* The previous name, `runeCount`, was indeed confusing as "rune" has no standardized meaning in JS (it may have been carried over from Go, where it essentially means the same as code point). Possible confusion with [Runic (Unicode block)](https://en.wikipedia.org/wiki/Runic_(Unicode_block)).
* The existing name, `graphemeLength`, is confusing as grapheme refers to ["the smallest functional unit of a writing system"](https://en.wikipedia.org/wiki/Grapheme). Unicode has the concept of ["grapheme clusters"](https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries), which can be accessed in JS via `Intl.Segmenter` with `granularity: 'grapheme'`:
   ```js
   const str = 'நி';
   str.length; // 2
   Array.from(str).length; // 2
   [...new Intl.Segmenter(undefined, { granularity: 'grapheme' }).segment(str)].length; // 1
   ```
   While a grapheme count may be considered more accurate or intuitive measure of string length for some use cases, it's too high-level of a concept to be suitable for giving a "column" number of a syntax error.
* `codePointLength` is clear and unambiguous as `Array.from` splits a string into characters, with each character corresponding to a single Unicode code point.

See also [What's the difference between a character, a code point, a glyph and a grapheme? - Stack Overflow](https://stackoverflow.com/questions/27331885/whats-the-difference-between-a-character-a-code-point-a-glyph-and-a-grapheme)


kt3k (approved):
Thanks for the good references!

LGTM

PR #5425, opened 2024-07-12
docs(text): refine documents in text module

magurotuna:
This commit refines the document in the `text` module. Namely,

- Move the note about word distance calculation possibly to change in the future to more visible place
- Fix the wrong doc comments (e.g. `possibleWords` description on `closestString`)
- Change `possibleWords` type from `string[]` to `ReadonlyArray<string>` to communicate to readers that the passed array won't be mutated
- Use `toSorted` instead of `sort` to ensure immutability

### DIFF
--- a/text/closest_string.ts
+++ b/text/closest_string.ts
@@ -6,7 +6,12 @@ import { levenshteinDistance } from "./levenshtein_distance.ts";
 const getWordDistance = levenshteinDistance;
 
 /**
- * The the most similar string from an array of strings.
+ * Finds the most similar string from an array of strings.
+ *
+ * Note: the ordering of words may change with version-updates
+ * E.g. word-distance metric may change (improve)
+ * use a named-distance (e.g. levenshteinDistance) to
+ * guarantee a particular ordering.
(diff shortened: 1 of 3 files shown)


text/closest_string.ts
```diff
@@ -20,19 +25,14 @@ const getWordDistance = levenshteinDistance;
...
- * @param possibleWords The string-array that will be sorted
+ * @param possibleWords The string-array to pick the closest string from
  * @param options An options bag containing a `caseSensitive` flag indicating
  * whether the distance should include case. Default is false.
- * @returns A sorted copy of possibleWords
- * @note
- * the ordering of words may change with version-updates
- * e.g. word-distance metric may change (improve)
- * use a named-distance (e.g. levenshteinDistance) to
- * guarantee a particular ordering
```
magurotuna, line 31: I see `@note` not rendered at all https://jsr.io/@std/text@1.0.0-rc.2/doc/~/closestString

kt3k (approved):
LGTM. Nice!

Issue #3757, opened 2023-11-01
Deprecate "deferred" API


bartlomieju:
With Deno v1.38 a new [`Promise.withResolvers()`](https://github.com/tc39/proposal-promise-with-resolvers) API will be available which has the same purpose as `deferred`. We should deprecate the latter and guide users to use the new built-in API.

irbull:
I wonder if there is a way to document these things on the std library page. I was looking for deferred semantics, and "deno deferred" brought me to https://deno.land/std@0.50.0/async/deferred.ts. (not sure why 0.50.0). If I select "go to latest" then I just get nothing. It took me more time than it should have to find this issue.

Moving the std library forward is a good idea, and trimming the fat is a great objective. But I wonder if others also struggle with (seemingly) sudden disappearances of standard library functionality and if there is a way this can be made a bit easier for others.

iuioiua:
Hi @irbull, valid point. We always do our best to document breaking changes and deprecations within release notes, so that's the best place to look. I created #5435 to guide users on how to find out about changed APIs. Please feel free to review.

Specifically regarding `deferred`, I was able to find track it by [searching "deferred"](https://github.com/denoland/deno_std/releases?q=deferred&expanded=true) in the releases page, as per the PR I created. The search term is highlighted.

Hopefully, this all helps.

irbull:
Thanks! The `Find a release` worked much better than I thought. It's not really clear that `find a release` means you can search for contents of the release notes, but that seems like a GitHub problem.

This is good solution. 

It would be nice if "Go To Latest" took you to a page with the deprecation notice, explaining what happened to the API. But this would likely be a lot of work, and maybe not worth the effort.

Thanks @iuioiua for a solid solution to this.

iuioiua:
> It would be nice if "Go To Latest" took you to a page with the deprecation notice, explaining what happened to the API. But this would likely be a lot of work, and maybe not worth the effort.

This seems like an interesting idea. Would you be able to [create an issue](https://github.com/jsr-io/jsr/issues/new) in the JSR repo?

PR #5435, opened 2024-07-14
chore: document how to find changed API

iuioiua:
Addresses https://github.com/denoland/deno_std/issues/3757#issuecomment-2223547790


irbull:
+1. The message is good and the `Find a release` worked better than I expected.

kt3k (approved):
LGTM

PR #5438, opened 2024-07-14
docs(jsonc): remove docs for removed `options` parameter

nikeee:
Options were removed in 255b7994ee20e8318016009693f0be00ab55416d


iuioiua (approved):
Good catch. Thank you.

PR #5440, opened 2024-07-15
chore(testing): add `.` export

iuioiua:
This change adds the `.` export for `@std/testing`. While the package is fine in it's current state without this export, it's likely better to have the `.` export rather than not, for a few reasons:
1. It allows users to import multiple symbols from `@std/testing` using a single import specifier. Such use cases exist in the Standard Library itself, and have been updated as part of this PR.
2. It allows the doc checker to check `testing/mod.ts`. This previously wasn't the case for `testing/README.md`.
3. It's consistent with the 36 out of 37 other packages across the Standard Library (except `@std/fmt`), which include `mod.ts` files. AFAIK, there's nothing unique about `@std/testing` that sufficiently justifies having it be treated different.
4. It removes the special treatment of the package in scripts in `_tools`.

I'm not aware of any sufficiently strong reasons against having a `.` export in this package.


kt3k:
I'd prefer to keep the current structure. Because `@std/testing` have more variety of purposes than other std packages, the list of APIs exported from `@std/testing` seem a bit too chaotic to me with this change.

I think this issue is motivated by the discord thread https://discord.com/channels/684898665143206084/1258496628923564053, but this seems cause by CLI bug, and it's not a std design issue.

iuioiua:
This PR enables multiple symbols within `@std/testing` in a single import. But it does not somehow inhibit users from importing from individual modules. Also, while that Discord thread triggered me to rethink about this package, but it is not the primary reason behind the change. That's why I deliberately didn't reference it in my opening comment.

I can't convince myself to share your concern. It seems minuscule compared to the benefits outlined in the initial comment.

kt3k:
I only see the maintainer's benefits in this change. Also this was never asked or suggested by the users.

BDD testing, mocking, snapshot testing, asserting types are quite different categories of test utilities, and also they export relatively large number of APIs (especially `bdd`, `mock` and `types`). Exporting all of them from single endpoint feels chaotic to me.

iuioiua:
There are plenty of examples of us making changes without user request. Yes, this mainly benefits the maintainer, but costs the user nothing. Nothing is stopping them from importing from the most-specific location. As seen by the changes within our own codebase in this PR, it's not impossible that someone wants to import more than one module from `@std/testing`.

kt3k:
> but costs the user nothing.

In my view, the cost here is a bit confusing and chaotic `default` namespace in API docs, which list all different kinds of APIs in alphabetical order. In my view that namespace is better not  existing. So this change looks having slight negative effect in doc readability to me

iuioiua:
Fair point. To summarise, these are the pros and cons.

Pros:
1. Allows users to import from multiple `@std/testing` modules in a single import statement. Users are still able to import from modules directly.
2. Allows doc checker, in it's current state, to check `testing/mod.ts`.
3. Makes the package consistent with 36 out of 37 other packages in the codebase.
4. Removes special handling of the package with scripts in `_tools`.

Cons:
1. Introduces a large `default` module.

Closing, as I don't think this is worth the hassle.

KyleJune:
The downside that makes me agree with @kt3k is that people may use asserts in non test code, and having a single import for all will likely result in people unintentionally importing a bunch of test modules they don't need into their applications.

iuioiua:
The same could be said for the root (`default`) module of any package in the Standard Library.

KyleJune:
> The same could be said for the root (`default`) module of any package in the Standard Library.

The only difference is that other modules in @std/testing besides assert would never normally be used outside the context of tests so importing all of them is guaranteed to be unnecessary. While you could have a reason to import all of some other modules, like if you know your application is going to be using most of the modules included.

KyleJune:
I had taken a break for a while from Deno, I just noticed that assert was moved to it's own top level std module. That kind of nulls the point I was making about how people might be importing @std/testing just for the assertion functions. I think the only reason people would import @std/testing now is for use in tests, so actually I am now in favor of their being a main entrypoint for importing all of @std/testing.

PR #5442, opened 2024-07-15
refactor(yaml): make `Type.kind` required

iuioiua:
All `Type` objects already have this property defined. Users don't have control over this value.

Towards #5340


kt3k (approved):
Nice finding!

PR #5445, opened 2024-07-15
refactor(yaml): remove dead code in `Schema` constructor

iuioiua:
This code checked that all `Type` objects passed to `Schema.implicit` had their `kind` properties set to `scalar`. I can confirm this is true for in all cases:
- Timestamp
- Merge
- Nil
- Bool
- Int
- Float

This code is not controlled by the user.


kt3k (approved):
LGTM

PR #5446, opened 2024-07-15
BREAKING(yaml): replace `YamlError` with `SyntaxError` in `parse()`

iuioiua:
### What's changed

[`parse()`](https://jsr.io/@std/yaml@1.0.0-rc.3/doc/~/parse) now throws [`SyntaxError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SyntaxError) instead of `YamlError`. No changes in behavior have been made.

### Motivation

This change was made to align `@std/yaml` with other `parse()` functions in the Standard Library, such as in `@std/csv` and `@std/toml`, and [`JSON.parse()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse), which all throw `SyntaxError`. This will also result in a slightly smaller API surface and allows the user to more easily differentiate between `parse()` and `stringify()` errors (`stringify()` will soon throw `TypeError`).

### Migration guide

To migrate, compare against `SyntaxError` instead of `YamlError` when error-handling `parse()`.
```diff
- import { parse, YamlError } from "@std/yaml/parse";
+ import { parse } from "@std/yaml/parse";

try {
  parse(`"`);
} catch (error) {
- if (error instanceof YamlError) {
+ if (error instanceof SyntaxError) {
    // ...
  }
}
```

### Related

Towards #5340

### DIFF
--- a/yaml/_loader.ts
+++ b/yaml/_loader.ts
@@ -35,7 +35,6 @@ import {
   SPACE,
   VERTICAL_LINE,
 } from "./_chars.ts";
-import { YamlError } from "./_error.ts";
 import { Mark } from "./_mark.ts";
 import { DEFAULT_SCHEMA, type Schema, type TypeMap } from "./_schema.ts";
 import type { Type } from "./_type.ts";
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

PR #5447, opened 2024-07-16
fix(text): unicode support and word splitting according to case

guy-borderless:
The current implementation of splitToWords() only supports Latin letters. This PR adds a unicode-compatible implementation + support for WordSpliting according to case.


text/_util.ts
```diff
@@ -1,9 +1,25 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+function splitByCase(input: string): string[] {
+  const acronym = /\p{Lu}{2,}/gu; // e.g. ID, URL
+  const capitalizedWord = /\p{Lu}\p{Ll}+/gu; // e.g. Apple
+  const lowercasedWord = /\P{Lu}+/gu; // e.g. apple, this will also match a sequence of unicode letters in languages without a concept of upper/lower case
+  const digits = /\p{N}+/gu; // e.g. 123
```
iuioiua, line 6: Nit: can you please move these out of the function? Otherwise, they're declared every time `splitByCase()` is called. Also, nice comments.

text/_util.ts
```diff
@@ -1,9 +1,25 @@
...
+}
 
 export function splitToWords(input: string) {
-  input = input.replaceAll(/[^a-zA-Z0-9\s-_]/g, "");
-  if (/[\s-_]+/.test(input)) return input.split(/[\s-_]+/);
-  return input.split(/(?=[A-Z])+/);
+  const words = input.split(/[^\p{L}\p{N}]+/gu).filter(Boolean).flatMap(
+    splitByCase,
+  );
+  return words;
```
iuioiua, line 22: Nit: can you please also move the regex out into a variable?
```suggestion
  return input
    .split(/[^\p{L}\p{N}]+/gu)
    .filter(Boolean)
    .flatMap(splitByCase);
```

text/case.ts
iuioiua, line 1: This file was split up into separate files in #5428. Please rebase.

iuioiua (changes requested):
Nice. Just a few nits on the first pass. Thank you.

text/_util.ts
```diff
@@ -1,9 +1,22 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
+const RE_ACRONYM = /\p{Lu}{2,}/gu; // e.g. ID, URL
```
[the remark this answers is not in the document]
kt3k: Looks like the flag does nothing in this case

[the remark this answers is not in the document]
kt3k: Oh, `u` flag seems necessary to avoid the type error.

kt3k:
By this change, the result of `toSnakeCase("Camel1Case")` becomes `camel_1_case` (It is  `camel1_case` in main), while the result of `toSnakeCase("camel1Case")` is kept as `camel1_case`. Is this change reasonable/acceptable?

guy-borderless:
> By this change, the result of `toSnakeCase("Camel1Case")` becomes `camel_1_case` (It is `camel1_case` in main), while the result of `toSnakeCase("camel1Case")` is kept as `camel1_case`. Is this change reasonable/acceptable?

Looking at a semantically reasonable example like `iAte2Cookies` the change does seem reasonable to me.

kt3k:
> Looking at a more semantic example like iAte2Cookies the change does seem reasonable to me.

I agree with this.

In that case, probably we should return `has_2_cookies` for `toSnakeCase("has2Cookies")` for consistency (It currently returns `has2_cookies`). I'll work on this.

guy-borderless:
I made some changes to better handle Acronyms followed by a capitalized word, like in `URLPattern` (before it would have been `["URLP", "aattern"]`, now return `["URL", "Pattern"]`.
But went ahead and simplified the implementation to only use one `input.match()` call

iuioiua (approved):

text/_util.ts
```diff
@@ -1,9 +1,18 @@
...
+const WORD_OR_NUMBER_REGEXP = new RegExp(
+  `${CAPITALIZED_WORD_REGEXP.source}|${ACRONYM_REGEXP.source}|${LOWERCASED_WORD_REGEXP.source}|${ANY_LETTERS.source}|${DIGITS_REGEXP.source}`,
+  "gu",
+);
+
 export function splitToWords(input: string) {
-  input = input.replaceAll(/[^a-zA-Z0-9\s-_]/g, "");
-  if (/[\s-_]+/.test(input)) return input.split(/[\s-_]+/);
-  return input.split(/(?=[A-Z])+/);
+  return input.match(WORD_OR_NUMBER_REGEXP) || [];
```
kt3k, line 15: @guy-borderless Can you give example that this regexp doesn't match? Is `[]` appropriate fallback in that case?

guy-borderless: "no words" may be an expected happy path response so why not allow it to fit in the normal path of the code, if it's logical. I think an empty array here is ok and less annoying than null. We can do null though. Anyway, added a test for that https://github.com/guy-borderless/deno_std/blob/dfd42c715e36fda7a94de287ecf99dd14fe11ba4/text/_util_test.ts#L14

kt3k: Thanks! Looks nice test cases to illustrate what happens to non-alphanumeric

kt3k (approved):
LGTM

kt3k:
> I made some changes to better handle Acronyms followed by a capitalized word, like in URLPattern (before it would have been ["URLP", "aattern"], now return ["URL", "Pattern"].

This looks a great improvement to me. Nice!

PR #5450, opened 2024-07-16
BREAKING(dotenv): remove `examplePath` option from `load[Sync]()`

### LINKED ISSUE #5449: proposal: remove `examplePath` option from `@std/dotenv/load`
[This option](https://jsr.io/@std/dotenv/doc/~/LoadOptions.examplePath) has two main fundamental design flaws:
1. It's magical, using another `.env` file to define required environment variable values in the main `.env` file. Such configuration should be handled in-code (see #4019).
2. The behavior that the option controls isn't clearly described by the name (see #3561).

For these reasons, it's likely best to remove this option.

iuioiua:
### What's changed

The `examplePath` option for `load[Sync]()` from `@std/dotenv` has been removed. The `allowEmptyValues` option and `MissingEnvVarsError` have also been removed, as the basis for their existence was dependent on the `examplePath` option.

### Motivation

See details in #5449.

### Migration guide

To migrate, if environment variables need to be validated, do so in-code instead of using the `examplePath` option.

```env
# .env.example
# Notice: missing `QUX` environment variable
FOO=bar
```

```diff
import "@std/dotenv/load";

+ if (!Deno.env.has("QUX") {
+   // ...
+ }
```

### Related

Closes #5449

### DIFF
--- a/dotenv/mod.ts
+++ b/dotenv/mod.ts
@@ -152,32 +152,7 @@ export function loadSync(
  * |File|Purpose|
  * |----|-------|
  * |.env|primary file for storing key-value environment entries
- * |.env.example|this file does not set any values, but specifies env variables which must be present in the configuration object or process environment after loading dotenv
  * |.env.defaults|specify default values for env variables to be used when there is no entry in the `.env` file
- *
- * ### Example file
- *
- * The purpose of the example file is to provide a list of environment
- * variables which must be set or already present in the process environment
- * or an exception will be thrown.  These
- * variables may be set externally or loaded via the `.env` or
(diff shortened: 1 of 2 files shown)


kt3k (approved):
`.env.example` is a common file name for showing the examples of what should be in `.env` file. https://github.com/search?q=path%3A.env.example&type=code

However using that file as the validator of `.env` file feels like an opinionated interpretation of it, and doesn't seem fitting to the policy of std. This removal seems reasonable to me.

iuioiua:
I'll merge in a few days if no strong objections arise.

PR #5451, opened 2024-07-16
BREAKING(dotenv): remove `defaultPath` option from `load[Sync]()`

### LINKED ISSUE #5464: proposal: remove `defaultPath` from `@std/dotenv/load`
[This option](https://jsr.io/@std/dotenv/doc/~/LoadOptions.defaultsPath) uses a `.env` file to define the default values of in a `.env` file, rather than in-code. It also doesn't align with Deno's philosophy of providing control in an explicit manner, instead of magically.

For these reasons, and similarly to #5449, I suggest we remove this option.

iuioiua:
### What's changed

The `defaultPath` option has been removed from `@std/dotenv/load`.

### Motivation

See details in #5464

### Migration

To migrate, if you want to define default values for environment variables, do so in-code.

```diff
import "@std/dotenv/load";

+ if (!Deno.env.has("FOO")) {
+   Deno.env.set("FOO", "bar");
+ }
```

### Related

Closes #5464


kt3k (approved):
Similarly to #5450, this handling of `.env.default` file seems opinionated, and it doesn't seem very common in the ecosystem.

I think it's reasonable to remove this feature from `std/dotenv`.

PutziSan:
Noticed a small typo in the PR description: it states `defaultPath` was removed, but the actual name is `defaultsPath.` Just noting here in case anyone else searches for this change in the future.

kt3k:
Thanks! Updated the PR title and description

PR #5452, opened 2024-07-16
BREAKING(yaml): replace `YamlError` with `TypeError` in `stringify()`

### LINKED ISSUE #5340: suggestion: replace `YamlError` with native errors

iuioiua:
### What's changed

[`stringify()`](https://jsr.io/@std/yaml@1.0.0-rc.3/doc/~/stringify) now throws [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) instead of `YamlError`. No changes in behavior have been made.

### Motivation

This change was made to align `@std/yaml` with other `stringify()` functions in the Standard Library, such as in `@std/csv` and `@std/toml`, and [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify), which all throw `TypeError`. This will also result in a slightly smaller API surface and allows the user to more easily differentiate between `parse()` and `stringify()` errors.

### Migration guide

To migrate, compare against `SyntaxError` instead of `YamlError` when error-handling `parse()`.
```diff
- import { stringify, YamlError } from "@std/yaml/stringify";
+ import { stringify } from "@std/yaml/stringify";

try {
  stringify({ undefined: undefined });
} catch (error) {
- if (error instanceof YamlError) {
+ if (error instanceof TypeError) {
    // ...
  }
}
```

### Related

Closes #5340


kt3k (approved):
LGTM

PR #5453, opened 2024-07-16
refactor(yaml): define ScalarType type and use it for implicit types

kt3k:
We removed the check of non-scalar types in implicit list of schema in #5445

This PR checks that constraint by defining `ScalarType` type and only allowing it for `implicit` types of the schemas.


iuioiua:
What about adding a type argument to `Type` instead? That's what I was gong to do after #5445.

kt3k:
Can you create a PR?

PR #5456, opened 2024-07-16
feat(html/unstable): add `isValidCustomElementName()`

### LINKED ISSUE #5437: feat req (html): add a validator for custom name elements ?

I don't know if it's out of scope for the html package, but would it be possible to add a [custom element name](https://developer.mozilla.org/en-US/docs/Web/API/Web_components/Using_custom_elements#name) validator ?

It is defined by the spec https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name as

> PotentialCustomElementName ::= 
> - [a-z] ([PCENChar](https://html.spec.whatwg.org/multipage/custom-elements.html#prod-pcenchar))* '-' ([PCENChar](https://html.spec.whatwg.org/multipage/custom-elements.html#prod-pcenchar))*
>
> PCENChar ::=  
> - "-" | "." | [0-9] | "_" | [a-z] | #xB7 | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x203F-#x2040] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
>
> name must not be any of the following:
> -   annotation-xml
> -   color-profile
> -   font-face
> -   font-face-src
> -   font-face-uri
> -   font-face-format
> -   font-face-name
> -   missing-glyph

Since the spec is unlikely to change often, this is low maintenance and it's something standard.

The gain is that the regex is pretty troublesome to define so people might tempted to implement it themselves and end up with a non-conforming.

Something similar to https://github.com/mathiasbynens/is-potential-custom-element-name/blob/master/index.js I guess,
but it's missing the list of forbidden names.

luk3skyw4lker:
These changes add a isValidCustomElement method to validate a custom element created in the HTML document. It is implemented according to the HTML spec that can be found at: https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name

I have also added tests for all the cases I could think of, if any case is missing please let me know and I can add it.

Closes #5437


timreichen:
I was wondering about the name staring with `is`. I am not sure if that is a common name prefix for a function instead of a property. Maybe we could borrow from [regexp test](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/test) or `check` from the [element api](https://developer.mozilla.org/en-US/docs/Web/API/Element/checkVisibility)?

luk3skyw4lker:
@timreichen hmmmm, I'm more inclined to do`'testCustomElementIsValid()`. Would that be okay? I don't want to leave the name too open to interpretations. The is<someting> pattern is pretty common in Ruby tho, I'm not against keeping it, although in JS/TS it's more common for properties as you said.

luk3skyw4lker:
`testValidCustomElement` also works for me, I think it's a very descriptive and nice name.

timreichen:
I would put `name` also in the function name. How about `testCustomElementName()` or `checkCustomElementNameValidity()`?

html/entities.ts
```diff
@@ -12,6 +12,20 @@ const rawToEntityEntries = [
   ["'", "&#39;"],
 ] as const;
 
+const forbiddenCustomElementNames: string[] = [
```
iuioiua, line 15: Few things:
1. Please uppercase, as this variable doesn't change.
2. `string[]` isn't needed.
3. Please link to WHATWG documentation for this list here.

html/entities.ts
```diff
@@ -12,6 +12,20 @@ const rawToEntityEntries = [
...
+  "color-profile",
+  "font-face",
+  "font-face-src",
+  "font-face-uri",
+  "font-face-format",
+  "font-face-name",
+  "missing-glyph",
+] as const;
+
+const potentialCustomElementsNameChars =
```
iuioiua, line 26: Ditto

html/entities.ts
```diff
@@ -115,6 +129,52 @@ export function unescape(
...
+/**
+ * A valid custom element name is a sequence of characters name that meets all of the following requirements:
+ *
+ * name must match the PotentialCustomElementName production:
+ *
+ * PotentialCustomElementName ::= [a-z] (PCENChar)* '-' (PCENChar)*
+ *
+ * PCENChar ::= "-" | "." | [0-9] | "_" | [a-z] | #xB7 | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x203F-#x2040] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]
+ *
+ * This uses the EBNF notation from the XML specification. [XML]
```
iuioiua, line 141: What is "this"? Also, is there supposed to be a link to the XML spec?

html/entities.ts
```diff
@@ -115,6 +129,52 @@ export function unescape(
...
+ * name must not be any of the following:
+ *
+ * annotation-xml
+ * color-profile
+ * font-face
+ * font-face-src
+ * font-face-uri
+ * font-face-format
+ * font-face-name
+ * missing-glyph
```
iuioiua, line 152:
```suggestion
 * The element name must not be any of the following:
 * - `annotation-xml`
 * - `color-profile`
 * - `font-face`
 * - `font-face-src`
 * - `font-face-uri`
 * - `font-face-format`
 * - `font-face-name`
 * - `missing-glyph`
```

html/entities.ts
```diff
@@ -115,6 +129,52 @@ export function unescape(
...
+ *
+ * @example Basic usage
+ *
+ * Using a valid custom element name
+ *
+ * ```ts
+ * import { isValidCustomElement } from '@std/html/entities'
+ * import { assertEquals } from "@std/assert";
+ *
+ * assertEquals(isValidCustomElement("custom-element"), true)
```
iuioiua, line 162: Please add a case that evaluates to `false`.

html/entities.ts
```diff
@@ -115,6 +129,52 @@ export function unescape(
...
+ * ```
+ *
+ * Reference on invalid names can be found [here](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name)
+ *
+ * @param {string} elementName The element name to be validate
+ * @returns {boolean} A boolean value indicating if the custom element name is valid or not
```
iuioiua, line 168: JSDoc `@param` types aren't need as we're using TypeScript.
```suggestion
 * @param elementName The element name to be validate
 * @returns A boolean value indicating if the custom element name is valid or not
```

html/entities.ts
iuioiua, line 141: These details may seem overwhelming to the reader. I think it's best that we just link to the relevant spec section for those that want to find out more.

iuioiua (changes requested):
Nice! Just a few nits.

kt3k:
I'd prefer the name `isValidCustomElementName`, which clearly shows the function is a predicate, returning boolean value.

> I was wondering about the name staring with is. I am not sure if that is a common name prefix for a function instead of a property.

The function name starting with `is` is common in std. There are many examples:
- [isLeap](https://jsr.io/@std/datetime/doc/~/isLeap)
- [isUtcLeap](https://jsr.io/@std/datetime/doc/~/isUtcLeap)
- [isStatus](https://jsr.io/@std/http/doc/~/isStatus)
- [isInformationalStatus](https://jsr.io/@std/http/doc/~/isInformationalStatus)
- [isSuccessfulStatus](https://jsr.io/@std/http/doc/~/isSuccessfulStatus)
- [isRedirectStatus](https://jsr.io/@std/http/doc/~/isRedirectStatus)
- [isClientErrorStatus](https://jsr.io/@std/http/doc/~/isClientErrorStatus)
- [isServerErrorStatus](https://jsr.io/@std/http/doc/~/isServerErrorStatus)
- [isErrorStatus](https://jsr.io/@std/http/doc/~/isErrorStatus)
- [isAbsolute](https://jsr.io/@std/path/doc/~/isAbsolute)
- [isGlob](https://jsr.io/@std/path/doc/~/isGlob)
- [isRange](https://jsr.io/@std/semver/doc/~/isRange)
- [isSemVer](https://jsr.io/@std/semver/doc/~/isSemVer)
- [isNil](https://jsr.io/@std/uuid/doc/~/isNil)

kt3k:
The function name starting with `test` has issues as discussed in https://github.com/denoland/std/pull/4364. I think we should avoid `testXxx` naming convention in new APIs.

html/entities.ts
```diff
@@ -115,6 +129,52 @@ export function unescape(
...
+ */
+export function isValidCustomElement(elementName: string): boolean {
```
kt3k, line 170: This API needs to be placed in its own file like `is_valid_custom_element.ts` as it's not related to HTML entities.

html/entities.ts
iuioiua: Actually, 2 and 3 are fine to leave.

iuioiua (approved):
I've made changes to align with the codebase. Now, LGTM. Thanks!

luk3skyw4lker:
Sorry to take long to answer guys!

html/is_valid_custom_element_name_test.ts
```diff
@@ -0,0 +1,46 @@
...
+  await t.step("handles custom names with especial chars", () => {
+    assertEquals(isValidCustomElementName("custom-element@"), false);
+  });
+
+  await t.step("handles custom names with numbers", () => {
+    assertEquals(isValidCustomElementName("custom-1-element"), true);
+  });
+
+  await t.step("handles custom names with underscores", () => {
+    assertEquals(isValidCustomElementName("custom_element"), true);
```
kt3k, line 36: A custom element name needs to include at least one `-`. This shouldn't be a valid name.

html/is_valid_custom_element_name_test.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  await t.step("handles custom names with points", () => {
+    assertEquals(isValidCustomElementName("custom.element"), true);
```
kt3k, line 40: Ditto. This isn't a valid name.

html/is_valid_custom_element_name_test.ts
luk3skyw4lker: Hmmmm, you're right, just took a look at the HTML reference and got it. I'll fix it tomorrow afternoon

iuioiua (approved):
I've fixed the previously mentioned faults and improved tests. LGTM now.

kt3k (approved):
I also cleaned up the regexp a bit.

LGTM

PR #5460, opened 2024-07-17
refactor(yaml): improve `Type.predicate` behavior

iuioiua:
Adds type guards and makes the purpose of the method clearer.


kt3k (approved):
LGTM

PR #5461, opened 2024-07-17
refactor(yaml): add `KindType` generic type argument to `Type`

iuioiua:
I'm mostly certain this is the right approach.

Supercedes #5453


kt3k (approved):
LGTM

PR #5462, opened 2024-07-17
refactor(yaml): remove `instanceOf` field of `Type`

kt3k:
This PR removes instanceOf field of `Type` type, which is only used by `timestamp` type and can be easily replaced by `object instanceof Date` check.

This simplifies the `detectType` logic in `yaml/_dumper.ts`


iuioiua (approved):
Nice! I noticed this too.

iuioiua (approved):

PR #5467, opened 2024-07-17
fix(text): bail early on timeout for `levenshteinDistance()` (#5465)

### LINKED ISSUE #5465: Performance: `levenshteinDistance` becomes extremely slow for large edit distances
**Describe the bug**

[`levenshteinDistance`](https://github.com/denoland/std/blob/c6998992a332f07cf223c89031eaa554ca6a4e67/text/levenshtein_distance.ts) becomes extremely slow for large edit distances.

**Steps to Reproduce**

```ts
const fmt = (n: number) => n.toLocaleString('en-US').padStart(6)

for (const n of [1_000, 2_000, 3_000, 10_000, 20_000, 30_000]) {
    const start = Date.now()
    levenshteinDistance('a'.repeat(n), 'b'.repeat(n))
    console.log(`Length = ${fmt(n)}: ${fmt(Date.now() - start)} ms`)
}
```

Sample results:

```
Length =  1,000:     39 ms
Length =  2,000:     80 ms
Length =  3,000:    177 ms
Length = 10,000:  2,044 ms
Length = 20,000:  7,865 ms
Length = 30,000: 20,883 ms
```

These performance characteristics are expected given that Levenshtein distance is `O(n * m)` where `n` and `m` are the lengths of the strings, but some way of mitigating the problem is desirable, as it could cause issues such as denial-of-service (whether unintentional or malicious).

**Expected behavior**

Possible options:
1. No change, maybe add a notice to the JSDoc comment for `levenshteinDistance` and the other functions that depend on it.
2. Throw on long inputs. Breaking change and probably not ideal as slow calculation may not be an issue in all use cases.
3. Return a special value for long inputs, such as `NaN`, `Infinity`, `-1`. Of these, `Infinity` is probably the least likely to cause problems, but it'd still be a breaking change... plus it's still less than ideal as `JSON.stringify(Infinity)` gives `'null'`.
4. Return `Math.max(str1.length, str2.length)` for long inputs. Simple and unlikely to cause show-stopping problems, but highly inaccurate for typical inputs.
5. Periodically check time elapsed while the function is executing, then return some value or throw if time elapsed exceeds some threshold. The returned value could be `Math.max(str1.length, str2.length)` or maybe something more accurate using the partial calculation so far as a heuristic. Would probably cause a slight degradation in performance for shorter inputs due to repeated `Date.now()` calls, but hopefully this wouldn't be significant.
6. Use a reduced granularity of edit difference (words? lines?) for longer inputs. Probably not a great option as it's both overly complex and insufficient to prevent malicious denial-of-service — any input that's split into a massive number of very short words/lines would still have the same performance problems.

Currently I'm favoring 5 or 4, maybe with a configurable option for max time elapsed/input length.

**Environment**

- OS: Ubuntu 20.04, WSL
- deno version: 1.45.2
- std version: text@0.224.3

lionel-rowe:
Fixes #5465.

This commit adds a `maxTimeElapsed` option to the `levenshteinDistance` function. If the calculation takes longer than this, the result will be approximated. The default is 100ms.

### DIFF
--- a/text/levenshtein_distance.ts
+++ b/text/levenshtein_distance.ts
@@ -2,6 +2,23 @@
 // This module is browser compatible.
 
 /**
+ * Options for Levenshtein distance calculation.
+ */
+export type LevenshteinOptions = {
+  /**
+   * The maximum calculation time to allow in milliseconds. If the calculation
+   * takes longer than this, the result will be approximated.
+   *
+   * @default 100
+   */
+  maxTimeElapsed: number;
+};
+
+const defaultOptions: LevenshteinOptions = {
+  maxTimeElapsed: 100,
+};
+
+/**
  * Calculates the
  * {@link https://en.wikipedia.org/wiki/Levenshtein_distance | Levenshtein distance}
  * between two strings.
@@ -15,9 +32,27 @@
  * ```
  * @param str1 The first string.
  * @param str2 The second string.
+ * @param options The options for the calculation.
  * @returns The Levenshtein distance between the two strings.
  */
-export function levenshteinDistance(str1: string, str2: string): number {
+export function levenshteinDistance(
+  str1: string,
+  str2: string,
+  options?: Partial<LevenshteinOptions>,
+): number {
+  if (str1 === str2) {
+    return 0;
+  }
+  if (!str1) {
+    return str2.length;
+  }
+  if (!str2) {
+    return str1.length;
+  }
+
+  const { maxTimeElapsed } = { ...defaultOptions, ...options };
+  const startTime = Date.now();
+
   if (str1.length > str2.length) {
     [str1, str2] = [str2, str1];
   }
@@ -45,6 +80,9 @@ export function levenshteinDistance(str1: string, str2: string): number {
       }
     }
     distances = tempDistances;
+    if (Date.now() - startTime > maxTimeElapsed) {
+      break;
+    }
   }
   return distances.at(-1)!;
 }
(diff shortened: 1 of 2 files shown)


timreichen:
I am not sure that this is a fix. Couldn't this potentially lead to bugs where the function returns a wrong value if it takes a bit longer?
I think the user should handle abortions themselves if needed since this is not really a algorithm problem.

lionel-rowe:
> Couldn't this potentially lead to bugs where the function returns a wrong value if it takes a bit longer

It depends if you consider a less accurate result to be more or less of a bug than crashes/denial of service for long inputs. 

> I think the user should handle abortions themselves if needed since this is not really a algorithm problem.

It's not possible to abort synchronous code if no hook is provided to do so. Another alternative would be providing the hook via a `maxTimeElapsed` option but defaulting it to `Infinity`.

I've mentioned some other alternatives in the attached issue btw, this isn't the only possible "fix" (although it's my current preferred one for the reasons I laid out there).

timreichen:
> It depends if you consider a less accurate result to be more or less of a bug than crashes/denial of service for long inputs. 

Possible Dos attacks should be handled by the user by validating input and limiting its lengths.

I would definitely consider a wrong result more of a bug than long calculation time.

Crashing the system is different, though your examples show long calculation times, not crashes. Is there an input that crashes the runtime? If so, I would consider that a bug.

lionel-rowe:
> Crashing the system is different, though your examples show long calculation times, not crashes. Is there an input that crashes the runtime? If so, I would consider that a bug.

I just mean that sufficiently long execution on the main thread can easily crash browser tabs or lock up a server instance. Or on Deno deploy, it's liable to knock down the server above the limit of 50ms of execution time per request (in practice it seems like longer requests are tolerated, at least if they're infrequent, but multiple seconds per request is definitely going to cause problems).

lionel-rowe:
Closing in favor of https://github.com/denoland/std/pull/5472

PR #5471, opened 2024-07-17
BREAKING(semver): replace `prerelease` and `buildmetadata` arguments with options object

### LINKED ISSUE #4420: suggestion(semver): remove redundant `increment()` arguments

`increment()` has four arguments (two required, two optional). That is more than the [style guide](https://docs.deno.com/runtime/manual/references/contributing/style_guide#exported-functions-max-2-args-put-the-rest-into-an-options-object) suggests.
- `buildmetadata` just seem to replace the semver build property.
- `prerelease` as a string kinda breaks the established pattern of prerelease being an `(string|number)[]` array and needs to be parsed inside `increment()`. It doesn't seem very intuitive, how that argument works in combination with `ReturnType` and existing `version.prerelease`.  This is why the majority of the jsdoc description tries to explain that.

- Remove `buildmetadata` argument, as `version.build` can be set outside the `increment()` function.
- Remove `prerelease ` argument, as `version.prerelease` can be set outside the `increment()` function.

Leave as is.

iuioiua:
### What's changed

The `prerelease` and `buildmetadata` arguments for [`increment()`](https://jsr.io/@std/semver/doc/increment/~/increment) into a new optional `IncrementOptions` interface.

### Motivation

This change was made to ensure the symbol adheres to the [Deno Style Guide](https://docs.deno.com/runtime/manual/references/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object.) which states that optional arguments should go into an options object.

### Migration

To migrate, use the options argument, instead of the positional arguments in `increment()`.

```diff
import { increment, parse } from "./semver/mod.ts";

- increment(parse("1.2.3"), "prerelease", "alpha", "0");
+ increment(parse("1.2.3"), "prerelease", { prerelease: "alpha", build: "0" });
```

### Related

Closes #4420

### DIFF
--- a/semver/increment.ts
+++ b/semver/increment.ts
@@ -35,6 +35,14 @@ function bumpPrerelease(
   return values;
 }
 
+/** Options for {@linkcode increment}. */
+export interface IncrementOptions {
+  /** The pre-release metadata of the new version. */
+  prerelease?: string;
+  /** The build metadata of the new version. */
+  build?: string;
+}
+
 /**
  * Returns the new SemVer resulting from an increment by release type.
  *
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

PR #5474, opened 2024-07-18
chore: allow `scope/unstable` notation in commit title

kt3k:
This allows using scope of the form `foo/unstable` such as `feat(text/unstable)` which expresses the commit only changes the unstable part of that package, and which doesn't affect major or minor version of the package.


kt3k:
tested in https://github.com/kt3k/unstable-test/commit/58008fa1b8564bcad9a333e4b3e7f6b664fcac71 and https://github.com/kt3k/unstable-test/pull/4 and

iuioiua (approved):

PR #5475, opened 2024-07-18
fix(media-types): return `video/mp4` for `.mp4` extension

### LINKED ISSUE #5433: http/file-server, media_types: new addition `application/mp4` to mime db overrides `video/mp4` and breaks some net players
**Describe the bug**
New addition [`application/mp4`](https://github.com/denoland/deno_std/blame/f54eab4355c1def5baa78533fb852bde1c6b761d/media_types/vendor/db.ts#L1049) to `media_types/vendor/db.ts` (#2286) overrides [`video/mp4`](https://github.com/denoland/deno_std/blob/f54eab4355c1def5baa78533fb852bde1c6b761d/media_types/vendor/db.ts#L8597) for the `contentType` function. As `file-server` specifies `content-type` [by that function and the db](https://github.com/denoland/deno_std/blob/f54eab4355c1def5baa78533fb852bde1c6b761d/http/file_server.ts#L234), it breaks some network-capable video players which expecting the mimetype to be `video/*`, because the received mimetype `application/mp4` is unexpected, they won't play the video anymore

**Steps to Reproduce**

1. Start newest `file-server` inside a folder with some mp4
```
$ deno run --allow-read --allow-net --allow-sys jsr:@std/http@1.0.0-rc.4/file-server
Listening on:
- Local: http://localhost:8000
- Network: http://192.168.2.103:8000
```
2. Issue a HEAD request to any video:
```
$ curl --head http://localhost:8000/1.mp4
HTTP/1.1 200 OK
...
content-type: application/mp4
...
```
3. The `content-type` is the new addition `application/mp4`, as it wins `video/mp4`. This breaks some network players as stated, as they don't identify `application/mp4`; they only know about `video/mp4`

**Expected behavior**

1. Start an older version of `file-server`, without the mimetype update
```
$ deno run --allow-read --allow-net --allow-sys jsr:@std/http@0.224.0/file-server
Listening on:
- Local: http://localhost:4507
- Network: http://192.168.2.103:4507
```
2. Issue a HEAD request to any video:
```
$ curl --head http://localhost:4507/1.mp4
HTTP/1.1 200 OK
...
content-type: video/mp4
...
```
3. as the mimetype `video/mp4` is by far the most preferred, it always worked for most (if not all) network players

**Environment**

- OS: any
- deno version: 1.45.2
- std version:
  - `@std/media_types` after the commit 6d58a57a011527c453dc79afc21453be1beb449e
  - `@std/http` 0.224.1 and above

**P.S.**

I think we need a smarter `contentType` function or mimetypes db that has preference over the "best" mimetype for certain file extensions. I see a similar problem had happened on `application/javascript` over `text/javascript` (#5186), probably there are far more media types are affected.

Likely that most players affected by this are proprietary, so fixing all the players is not feasible.

kt3k:
closes #5433

### DIFF
--- a/media_types/_db.ts
+++ b/media_types/_db.ts
@@ -32,6 +32,7 @@ for (const type of Object.keys(db) as KeyOfDb[]) {
 
       if (
         current !== "application/octet-stream" &&
+        current !== "application/mp4" &&
         (from > to ||
           // @ts-ignore work around https://github.com/denoland/dnt/issues/148
           (from === to && current.startsWith("application/")))
--- a/media_types/type_by_extension_test.ts
+++ b/media_types/type_by_extension_test.ts
@@ -16,6 +16,7 @@ Deno.test({
       ["file.json", undefined],
       ["foo", undefined],
       [".foo", undefined],
+      [".mp4", "video/mp4"],
     ] as const;
     for (const [fixture, expected] of fixtures) {
       assertEquals(typeByExtension(fixture), expected);


iuioiua (approved):
This makes me wonder if there are any extensions that are incorrect. E.g. `.mpg4` should return `video/mp4`, but it also returns `video/mp4`. We should probably come up with some better way to calculate preference in a follow-up PR.

This is a quick and dirty script to help find which extensions have more than one content-type:
```ts
import db from "./media_types/vendor/db.ts";

const result: Record<string, string[]> = {};

for (const [key, value] of Object.entries(db)) {
  if (Array.isArray(value.extensions)) {
    for (const ext of value.extensions) {
      if (!result[ext]) {
        result[ext] = [];
      }
      result[ext].push(key);
    }
  }
}

for (const [ext, contentTypes] of Object.entries(result)) {
  if (contentTypes.length > 1) console.log([ext, contentTypes]);
}
```

Issue #5476, opened 2024-07-18
to-do: review custom error classes


iuioiua:
The need to ensure any thrown errors are providing value.
- [x] `RetryError`
- [x] `MissingEnvVarsError` (removed in #5450)
- [x] `SubdirectoryMoveError` (removed in #5532)
- [x] `WalkError` (removed in #5477)
- [x] `BufferFullError`
- [x] `PartialReadError`
- [x] `MockError`
- [x] `TimeError` (removed in #5550)

Note: custom error classes from `@std/assert` and `@std/expect` have been omitted on purpose.

jsejcksn:
Unless there's a specific reason to omit the [`options`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error#options) argument — I think that any [`Error`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error) subclass in `std` (e.g. [`AssertionError`](https://github.com/denoland/std/blob/main/assert/assertion_error.ts), etc.) should also support `options` in its constructor.

There have been many projects I've worked on where I would have loved to use `AssertionError`, but couldn't because it's not currently able to capture a `cause` at construction in order to maintain context during exception-handling.

Similar to:

- https://github.com/denoland/deno/issues/24712

iuioiua:
I'm happy to look at a PR that adds the `options` parameter to `AssertionError`.

timreichen:
Is the error grammar in all std mods the same? I think to remember different ones (`missing xyz: something`, `xyz is missing in something`, `you didn't put xyz in something`). Is this also handled in these PRs?

iuioiua:
Let's look into error messages in a separate issue 🙂

PR #5477, opened 2024-07-18
BREAKING(fs): throw `Deno.errors.NotFound` instead of `WalkError` in `walk[Sync]()`

iuioiua:
### What's changed

[`walk()`](https://jsr.io/@std/fs/doc/walk/~/walk) and [`walkSync()`](https://jsr.io/@std/fs/doc/walk/~/walkSync) now naturally throw [`Deno.errors.NotFound`](https://docs.deno.com/api/deno/~/Deno.errors#namespace_deno_errors_notfound) instead of throwing `WalkError`. The `root` property has also been removed and the `message` property no longer includes the normalized path.

### Motivation

This change was made to allow these functions to delegate error throwing functionality to the Deno runtime, which provides sufficient context naturally. This also removes a dependency on `@std/path/normalize` and slightly reduces the API surface area of `@std/fs/walk`.

### Migration guide

To migrate, compare against `Deno.errors.NotFound` instead of `WalkError` when error-handling `walk()`.

```diff
- import { walk, WalkError } from "@std/fs/walk";
+ import { walk } from "@std/fs/walk";

const root = "../my-dir";

try {
  await Array.fromAsync(
    walk(root),
    async () => await Deno.remove(root, { recursive: true }),
  );
} catch (error) {
- if (error instanceof WalkError) {
+ if (error instanceof Deno.errors.NotFound) {
    // ...
  }
}
```

### Related

Towards #5476


kt3k:
`WalkError.path` seems to have a usage https://github.com/denoland/std/issues/3023#issuecomment-1387458640

iuioiua:
That suggestion was made as a passing remark rather than being triggered from a reported need from a user. After recent work on errors in the codebase, I'm becoming more convinced that throwing naturally is sufficient in almost all cases. If it's good enough for the runtime, I think it's good enough for the Standard Library.

iuioiua:
Furthermore, `Deno.errors.NotFound` naturally provides clearer context for the error.

kt3k:
> Furthermore, Deno.errors.NotFound naturally provides clearer context for the error.

`WalkError.prototype.root` gives the root path of `walk`. On the other hand, `Deno.error.NotFound` only shows the exact path, which was not found. The same information is not available with this change.

kt3k:
@lowlighter and @lambdalisue are the users of WalkError in open source https://github.com/search?q=WalkError+std%2Ffs&type=code

What do you think about this change?

kt3k:
Also I think @binyamin uses this in some projects https://github.com/denoland/std/issues/3023#issuecomment-1387458640

@binyamin Do you still use `WalkError` in your projects? What do you think about this change?

lambdalisue:
If the change is clearly mentioned in the release notes, I'm fine with it. In fact, I prefer `Deno.errors.NotFound` over `WalkError` because users of `walk()` can wrap the error themselves if they need `root` information. I believe that certain responsibilities should lie with the users rather than the providers.

marvinhagemeister:
+1 on going with `Deno.errors.NotFound`. It feels more consistent with the rest of Deno to me as other file system APIs throw that too.

lowlighter:
I concur with what have been already said

kt3k (approved):
Thanks for the feedbacks/comments!

LGTM

Issue #5481, opened 2024-07-18
to-do: fix unstable notices


iuioiua:
We need to use a consistent pattern for unstable API docs and ensure notices aren't the first thing in the description. Otherwise, we get [this](https://jsr.io/@std/cli@1.0.0-rc.5/doc/spinner/~).

PR #5482, opened 2024-07-18
docs(cli): improve unstable API notices

iuioiua:
Towards #5481

### DIFF
--- a/cli/spinner.ts
+++ b/cli/spinner.ts
@@ -19,9 +19,10 @@ const DEFAULT_SPINNER = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧",
 export type Ansi = string & {};
 
 /**
- * **UNSTABLE**: New API, yet to be vetted.
- *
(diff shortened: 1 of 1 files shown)


cli/spinner.ts
```diff
@@ -19,10 +19,11 @@ const DEFAULT_SPINNER = ["⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧",
 export type Ansi = string & {};
 
 /**
- * **UNSTABLE**: New API, yet to be vetted.
- *
  * Color options for {@linkcode SpinnerOptions.color}.
  *
+ * > [!WARNING]
+ * > **UNSTABLE**: New API, yet to be vetted.
```
iuioiua, line 25: Note: this doesn't look the best when hovered-on in VSCode, but I think that's fixable, and a nit, compared to how prominent these notices are [in JSR](https://jsr.io/@std/net@1.0.0-rc.1/doc/~/getNetworkAddress).

image

kt3k (approved):
LGTM

PR #5485, opened 2024-07-19
chore: remove requirement for `@example`s in constructors

iuioiua:
These aren't needed as JSR doesn't have a dedicated page for class constructors. Also, VSCode merges the documentation for classes and their constructors. In other words, class examples are shown, even if there are no constructor examples, when one hovers over a class.

I'll remove the redundant examples in follow-up PRs.


kt3k (approved):
Fair enough

LGTM

PR #5486, opened 2024-07-19
docs(cli): fix options arguments display

iuioiua:
This ensures that the argument is displayed as `options` instead of [`unnamed 1`](https://jsr.io/@std/cli@1.0.0-rc.5/doc/~/parseArgs) in documentation.


kt3k (approved):
LGTM

PR #5488, opened 2024-07-19
docs(http): fix options argument display

iuioiua:
This ensures that the argument is displayed as `options` instead of `unnamed 1` in documentation.


kt3k (approved):
LGTM

PR #5492, opened 2024-07-19
feat(streams): new ExactReader

BlackAsLight:
This pull request adds a new class to `@std/streams` that takes in a `ReadableStream<Uint8Array>` and returns a `ReadableStreamBYOBReader` that guarantees the buffer provided will always be fully filled when it resolves, with the exception of the last value returned.

This class is useful for if you want to consume the stream in changing but exact sizes. Say you want to consume the first byte then the next x bytes, etc.
```ts
import { ExactReader } from '@std/streams'
import { assertEquals } from '@std/assert'

const reader = new ExactReader((await Deno.open('./deno.json')).readable)
let x = await reader.read(1)
assertEquals(x.value!.length, 1)
reader.releaseLock()
```

### DIFF
--- a/streams/deno.json
+++ b/streams/deno.json
@@ -8,6 +8,7 @@
     "./concat-readable-streams": "./concat_readable_streams.ts",
     "./delimiter-stream": "./delimiter_stream.ts",
     "./early-zip-readable-streams": "./early_zip_readable_streams.ts",
+    "./exact-reader": "./exact_reader.ts",
     "./limited-bytes-transform-stream": "./limited_bytes_transform_stream.ts",
     "./limited-transform-stream": "./limited_transform_stream.ts",
     "./merge-readable-streams": "./merge_readable_streams.ts",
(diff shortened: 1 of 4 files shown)


kt3k:
I feel the same thing can be achieved by using `min` option of `ReadableStreamBYOBReader.read`

ref: https://github.com/denoland/deno/pull/20849 and https://github.com/whatwg/streams/pull/1145

```js
const readable = (await Deno.open('./deno.json')).readable
const reader = readable.getReader({ mode: "byob" });
const u8 = new Uint8Array(1024);
const result = await reader.read(u8, { min: 1024 });
console.log(result.value.length)
```

BlackAsLight:
I did not even know this was a thing. I can't find any info on a second argument for the read method on MDN

BlackAsLight:
The MDN docs makes no mention of the second argument: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStreamBYOBReader/read

But looking at the spec it lists, it does show a min argument: https://streams.spec.whatwg.org/#ref-for-byob-reader-read%E2%91%A2

Issue #5495, opened 2024-07-19
`@std/cli`: move informational output to `stderr` (instead of `stdout`)


jsejcksn:
Refer to the comments beginning here for context: https://github.com/denoland/std/issues/5002#issuecomment-2181247050

It appears that some modules (e.g. [`spinner.ts`](https://github.com/denoland/std/blob/release-2024.07.19/cli/spinner.ts) and [`prompt_secret.ts`](https://github.com/denoland/std/blob/release-2024.07.19/cli/prompt_secret.ts), but there may be more) are still writing informational output to `stdout`. Is this really the intended behavior?

iuioiua:
This sounds reasonable to me. I've created #5497 to play with the idea with `Spinner`. The errors need fixing.

This might be a little trickier to do with `promptSecret()`, as it's now a stable API (as of yesterday 😅).

WDYT, @kt3k?

kt3k:
How about making the output destination optional (like, `output: "stderr"`) instead of changing the default behavior?

jsejcksn:
> How about making the output destination optional (like, `output: "stderr"`) instead of changing the default behavior?

@kt3k In the linked issue that took place prior to stabilization, I explained why `stdout` should not be used for informational output. In short: writing to `stdout` by default will corrupt any expected structured output from a process (e.g. streams, piped data, structured logs, etc.)

kt3k:
We can do that for Spinner. But we'd like to avoid changing it for `promptSecret` as it's already stabilized.

> will corrupt any expected structured output from a process (e.g. streams, piped data, structured logs, etc.)

I think this generally doesn't apply to `promptSecret` because when `promptSecret` is used, then stdin/stdout should be the terminal

jsejcksn:
> But we'd like to avoid changing it for promptSecret as it's already stabilized.

I think this is unfortunate because these concerns were raised prior to stabilization, but no maintainer engaged them.

iuioiua:
[...]

PR #5497, opened 2024-07-20
BREAKING(cli/unstable): write `Spinner` output to stderr

iuioiua:
Experimenting with #5495


cli/spinner.ts
```diff
@@ -81,6 +83,12 @@ export interface SpinnerOptions {
    * This can be changed while the spinner is active.
    */
   color?: Color;
+  /**
+   * Stream to write the spinner to.
+   *
+   * @default {Deno.stderr}
+   */
+  stream?: WriterSync & { writable: WritableStream<Uint8Array> };
```
kt3k, line 91: How about just accepting `"stderr" | "stdout"`? Dependency to `std/io` doesn't seem ideal.

iuioiua: I was thinking of instead changing it to `writable?: WritableStream<Uint8Array>` with a default value of `Deno.stderr.writable`. That way, that'd make it much more flexible and drop the `@std/io` dependency.

iuioiua: While my above suggestion doesn't completely eliminate the tie to the Deno runtime (we use `Deno.noColor`), it at least reduces it, making future compatibility with other runtimes a little easier.

PR #5498, opened 2024-07-20
fix(assert): `assertObjectMatch` doesn't print whole object

babiabeo:
Separated from #5408
Fix: #3771

When filtering two objects, the `fn()` function returns an empty object if they don't have the same keys or symbols. This PR adds the logic so that it returns the entries of the first object in this case. (also rename `fn` to 	`filterObject`)

### DIFF
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -32,75 +32,6 @@ export function assertObjectMatch(
   expected: Record<PropertyKey, unknown>,
   msg?: string,
 ): void {
-  type loose = Record<PropertyKey, unknown>;
-
-  function filter(a: loose, b: loose) {
-    const seen = new WeakMap();
-    return fn(a, b);
-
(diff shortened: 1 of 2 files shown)


assert/object_match.ts
```diff
@@ -32,82 +32,107 @@ export function assertObjectMatch(
...
-            filtered[key] = fn({ ...value }, { ...subset });
-            continue;
-          }
-        } // On regexp references, keep value as it to avoid loosing pattern and flags
-        else if (value instanceof RegExp) {
-          filtered[key] = value;
+
+        // When both operands are set, build a filtered set with common values
+        if ((value instanceof Set) && (subset instanceof Set)) {
+          filtered[key] = value.intersection(subset);
```
iuioiua, line 125: Nice

iuioiua (approved):
Thanks!

PR #5501, opened 2024-07-21
docs(expect): minor doc tweaks

iuioiua:


expect/fn.ts
```diff
@@ -19,14 +17,16 @@
  *   expect(mockFn).toHaveBeenCalledTimes(1);
  * });
  * ```
+ *
+ * @module
```
iuioiua, line 21: Fixes module docs not being displayed.

expect/mod.ts
```diff
@@ -5,6 +5,16 @@
 /**
  * This module provides Jest compatible expect assertion functionality.
  *
+ * ```ts no-assert
+ * import { expect } from "@std/expect";
+ *
+ * const x = 6 * 7;
+ * expect(x).toEqual(42);
+ * expect(x).not.toEqual(0);
+ *
```
iuioiua, line 14: Aligns with other packages which usually have an example first.

expect/fn.ts
kt3k: Good catch

kt3k (approved):
LGTM

PR #5455, opened 2024-07-16
chore: release `expect@1.0.0`

kt3k:
To be merged July 22

The latest API docs: https://jsr.io/@std/expect@1.0.0-rc.3/doc

Closes #5014

PR #5502, opened 2024-07-22
docs(expect): link to matcher docs

kt3k:
related #5455

### DIFF
--- a/expect/mod.ts
+++ b/expect/mod.ts
@@ -20,7 +20,7 @@
  *   - {@linkcode Expected.toBeFalsy | toBeFalsy}
  *   - {@linkcode Expected.toContain | toContain}
  *   - {@linkcode Expected.toContainEqual | toContainEqual}
- *   - {@linkcode Expected.toHveLength | toHaveLength}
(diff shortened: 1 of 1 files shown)


expect/mod.ts
```diff
@@ -7,53 +7,53 @@
...
- *   - `toBeGreaterThan`
- *   - `toBeGreaterThanOrEqual`
- *   - `toBeLessThan`
- *   - `toBeLessThanOrEqual`
- *   - `toBeCloseTo`
- *   - `toBeInstanceOf`
- *   - `toThrow`
- *   - `toHaveProperty`
- *   - `toHaveLength`
+ *   - [`toBe`](./doc/~/Expected.toBe)
```
iuioiua, line 10:
```suggestion
 *   - {@linkcode Expected.toBe}
```
or
```
 *   - {@linkcode Expected.toBe | toBe}
```

iuioiua (changes requested):

iuioiua (approved):

iuioiua:
Actually, we still don't document how use the API (e.g. `expect.toBe`). We should do that before publishing `expect@1.0.0`.

PR #5503, opened 2024-07-22
fix(assert): fix `assertObjectMatch()` prints arrays as objects

babiabeo:
Separated from #5408
Fix: #3472

Currently, when filtering arrays, `assertObjectMatch()` uses `filterObject({ ...value }, { ...subset })` which converts arrays into objects. This PR adds `filterArray()` function to handle arrays.


iuioiua (approved):
Nice! Thank you.

kt3k:
Nice work!

PR #5505, opened 2024-07-23
fix(crypto): improve handling of `length` option

### LINKED ISSUE #1429: docs: document about which algorithms support `length` option for stdCrypto.digest
related: https://github.com/denoland/deno_std/pull/1386

iuioiua:
Closes #1429


iuioiua:
PTAL, @kt3k

crypto/crypto.ts
```diff
@@ -33,7 +36,7 @@
  * - `BLAKE2B-256`
  * - `BLAKE2B-384`
  * - `BLAKE2S`
- * - `BLAKE3`
+ * - `BLAKE3` (length-adjustable)
```
kt3k, line 39: What's the difference of `(length-adjustable)` and `(length-entendable)`?

crypto/crypto.ts
```diff
@@ -215,6 +218,11 @@ const stdCrypto: StdCrypto = ((x) => x)({
         // and the data is a single buffer,
         isBufferSource(data)
       ) {
+        if (length !== undefined) {
+          throw new TypeError(
```
kt3k, line 222: The native `crypto.subtle.digest` seems ignoring `length` for the case where `length` is irrelevant. Throwing here feels inconsistent with it.

iuioiua: Good point. I've reverted and instead added a note saying that `length` will be ignored in some cases.

crypto/crypto.ts
iuioiua: > "length-extendable" refers to a specific vulnerability in some hash functions, while "length-adjustable" is not a standard term in cryptography but could refer to the general capability of hash functions to process inputs of different lengths.

-- ChatGPT

littledivy (approved):
LGTM

PR #5516, opened 2024-07-23
chore: document design decision to only use `@std/assert` within testing code

### LINKED ISSUE #4865: suggestion: don't use `std/assert` in implementation code

iuioiua:
Closes #4865

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -54,6 +54,14 @@ and [architecture guide](./ARCHITECTURE.md) before contributing.
    deprecation(<package>): <symbol>
    ```
 
+## Implementations
+
+### Assertions
+
+Do not use assertions from [`@std/assert`](https://jsr.io/@std/assert) to assert
+values in implementation code. Instead, check values inline. `@std/assert`
+functions should only be used in testing. See #4865 for details.
+
 ## Tests
 
 ### Test names


kt3k (approved):
LGTM

PR #5517, opened 2024-07-23
fix(streams): strictly define `toJson()` and `toText()` input

iuioiua:
Includes documentation clarifications and test cleanups. See discussion in #5079.


kt3k (approved):
LGTM

PR #5523, opened 2024-07-23
test(yaml): test sortKeys option behavior

kt3k:
This PR adds test cases that check the behavior of `sortKeys` option of `stringify`.


iuioiua (commented):
I don't see a good reason for YAML to handle this functionality. Rather, this seems like something that should be done to the data before stringification. In other words, I suggest removing this option. WDYT?

kt3k:
The key order of Object is not easily controllable in JavaScript. See the discussion in https://stackoverflow.com/questions/5525795/does-javascript-guarantee-object-property-order

iuioiua (approved):

PR #5525, opened 2024-07-23
test(yaml): improve coverage for nil `Type`

iuioiua:


yaml/_type/nil.ts
```diff
@@ -29,9 +29,6 @@ export const nil: Type<"scalar", null> = {
   kind: "scalar",
   predicate: isNull,
   represent: {
-    canonical(): string {
```
iuioiua, line 32: Seems this isn't called anywhere.

kt3k (approved):
LGTM

PR #5526, opened 2024-07-23
refactor(yaml): cleanup dead code in float `Type` implementation

iuioiua:
According to the spec, it seems that the float type shouldn't contain a `:` (see https://yaml.org/spec/1.2.2/#10214-floating-point).

### DIFF
--- a/yaml/_type/float.ts
+++ b/yaml/_type/float.ts
@@ -36,7 +36,6 @@ function resolveYamlFloat(data: string): boolean {
 function constructYamlFloat(data: string): number {
   let value = data.replace(/_/g, "").toLowerCase();
   const sign = value[0] === "-" ? -1 : 1;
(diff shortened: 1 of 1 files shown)


kt3k:
colon separated numbers are sexagesimals (`1:00` equals `60` as int, also `1:00.0` equals `60.0` as float). It was part of spec in YAML 1.1. ref: https://yaml.org/type/float.html, but it seems to have been dropped in 1.2. Probably it's fine to remove as we mainly target 1.2

BTW we already tests sexagesimals for int types. Probably we should remove them as well for consistency

kt3k (approved):
LGTM

iuioiua:
> BTW we already tests sexagesimals for int types. Probably we should remove them as well for consistency

Can you please clarify?

PR #5528, opened 2024-07-23
test(text): add more testcases for `levenshteinDistance`

0f-0b:
https://github.com/denoland/std/pull/5527#pullrequestreview-2193725090.


iuioiua (approved):
Thanks! Note: `assertStrictlyEquals()` isn't really needed here.

PR #5529, opened 2024-07-24
docs(path): cleanup `@std/path/posix` and `@std/path/windows` module documentation

iuioiua:
This is clearer and more accurate.


kt3k (approved):
LGTM

PR #5532, opened 2024-07-24
BREAKING(fs): throw `Deno.errors.NotSupported` instead of `SubdirectoryMoveError` in `move[Sync]()`

iuioiua:
### What's changed

Previously, [`@std/fs/move`](https://jsr.io/@std/fs/doc/move/~/) threw `SubdirectoryMoveError`. Now, it throws [`Deno.errors.NotSupported`](https://docs.deno.com/api/deno/~/Deno.errors.NotSupported).

### Motivation

This change was made to reduce the API surface area by using an appropriate error class that's already built into the Deno runtime.

### Migration guide

To migrate, compare against `Deno.error.NotSupported` instead of `SubdirectoryMoveError` when error-handling `move()` or `moveSync()`.

```diff
- import { move, SubdirectoryMoveError } from "@std/fs/move";
+ import { move } from "@std/fs/move";

try {
  await move("./path/to/folder/subdir", "./path/to/folder");
} catch (error) {
- if (error instanceof SubdirectoryMoveError) {
+ if (error instanceof Deno.errors.NotSupported) {
    // ...
  }
}
```

### Related

Towards #5476

### DIFF
--- a/fs/move.ts
+++ b/fs/move.ts
@@ -4,38 +4,6 @@ import { isSamePath } from "./_is_same_path.ts";
 
 const EXISTS_ERROR = new Deno.errors.AlreadyExists("dest already exists.");
 
-/**
- * Error thrown in {@linkcode move} or {@linkcode moveSync} when the destination
- * is a subdirectory of the source.
- *
- * @example Usage
- * ```ts no-eval
- * import { move, SubdirectoryMoveError } from "@std/fs/move";
- *
- * try {
- *   await move("./foo", "./foo/bar");
- * } catch (error) {
- *   if (error instanceof SubdirectoryMoveError) {
- *     console.error(error.message);
- *   }
- * }
- * ```
- */
-export class SubdirectoryMoveError extends Error {
-  /**
-   * Constructs a new instance.
-   *
-   * @param src The source file or directory as a string or URL.
-   * @param dest The destination file or directory as a string or URL.
-   */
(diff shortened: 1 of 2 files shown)


kt3k:
`InvalidData` sounds a bit strange for describing this situation to me.

Is this worth the breaking change?

iuioiua:
> `InvalidData` sounds a bit strange for describing this situation to me.

How about `Deno.errors.NotSupported` if `src` is a sub-directory of `dst` and `Deno.errors.AlreadyExists` if `src` equals `dst`?

> Is this worth the breaking change?

Yes. It's one less extraneous symbol that the user has to import.

kt3k:
> It's one less extraneous symbol that the user has to import.

This doesn't sound a good reason for breaking change. There are many symbols we can remove if we optimize for fewer number of symbols.

iuioiua:
Perhaps, we're asking the wrong question. I think the question we should be asking is: why should we have a custom `SubdirectoryMoveError` class if there already exists an extensive collection of built-in error classes that accomplish the same thing? This is the only custom error class in `@std/fs`. Why?

kt3k:
Ok. I changed my mind. Let's do this change, but let's use `Deno.errors.NotSupported`, which seems more reasonable to me.

iuioiua:
Great! Updated.

kt3k (approved):
LGTM

PR #5538, opened 2024-07-24
BREAKING(yaml): remove `default` schema type

iuioiua:
### What's changed

Previously, the `default` schema type was set as the default schema for `parse()` and `stringify()`. Now, this has been changed to `core`, and the configuration of the `default` schema type have been merged with the `extended` schema type`.

### Motivation

This change was made to align with the [YAML spec](https://yaml.org/spec/1.2.2/#103-core-schema), which recommends the `core` schema be the default schema for YAML processors.

> The Core schema is an extension of the [JSON schema](https://yaml.org/spec/1.2.2/#json-schema), allowing for more human-readable [presentation](https://yaml.org/spec/1.2.2/#presentation-stream) of the same types. This is the recommended default [schema](https://yaml.org/spec/1.2.2/#recommended-schemas) that YAML [processor](https://yaml.org/spec/1.2.2/#processes-and-models) should use unless instructed otherwise.

Doing so also provides the benefit of removing one possible value of `SchemaType`, and therefore simplifying configuration for `parse()` and `stringify()`. Use cases which don't require the `extended` schema will also see a slight performance boost.

### Migration guide

To migrate, if parsing or stringifying one of the following types, ensure you're using `option.schema = "extended"`:
1. Binary,
2. Ordered map (omap)
3. Pairs
4. Set
5. Regular expressions
6. Undefined types
7. Anchors and aliases

```diff
import { parse } from "@std/yaml/parse";

- parse('message: !!binary "SGVsbG8=");
+ parse('message: !!binary "SGVsbG8="', { schema: "extended" });
```
```diff
import { stringify } from "@std/yaml/stringify";

- stringify(new Uint8Array([1, 2, 3]));
+ stringify(new Uint8Array([1, 2, 3]), { schema: "extended" });
```

### DIFF
--- a/yaml/_dumper.ts
+++ b/yaml/_dumper.ts
@@ -26,7 +26,7 @@ import {
   SINGLE_QUOTE,
   VERTICAL_LINE,
 } from "./_chars.ts";
-import { DEFAULT_SCHEMA, type Schema } from "./_schema.ts";
+import { CORE_SCHEMA, type Schema } from "./_schema.ts";
 import type { KindType, StyleVariant, Type } from "./_type.ts";
 import { type ArrayObject, getObjectTypeString, isObject } from "./_utils.ts";
 
@@ -486,7 +486,7 @@ export class DumperState {
   dump: any;
 
   constructor({
-    schema = DEFAULT_SCHEMA,
+    schema = CORE_SCHEMA,
     indent = 2,
     arrayIndent = true,
     skipInvalid = false,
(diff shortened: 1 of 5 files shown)


kt3k:
I'm strongly against this change. The use of `merge` type `<<` is very common in the ecosystem. https://github.com/search?q=%3C%3C%3A+path%3A*.yml&type=code

Removing that feature from default behavior seems too disruptive.

PR #5539, opened 2024-07-24
refactor(yaml): remove deprecated sexagesimal functionality for integer types

iuioiua:
This functionality is part of the YAML 1.1 spec. We mainly support YAML 1.2.

Addresses https://github.com/denoland/std/pull/5526#issuecomment-2247018521


yaml/parse_test.ts
```diff
@@ -333,9 +333,9 @@ Deno.test({
       -Infinity,
       NaN,
       12000,
-      75,
-      4520,
-      -4520,
+      1,
+      1,
+      -1,
```
kt3k, line 338: These need to be:
```
"1:15",
"1:15:20",
"-1:15:20",
```

(all need to be resolved as strings) if we drop the sexagesimal integer parsing.

kt3k (approved):
Nice! LGTM

PR #5540, opened 2024-07-24
BREAKING(http): make printing of network IP address opt-in for file server

iuioiua:
### What's changed

Previously, `@std/http/file-server` printed the current machine's network IP address upon listen, by default. Now, this behavior is opt-in; set by either the `printNetworkIp` option if calling `serveDir()`, or the corresponding `--ip` flag if running as a script.

### Motivation

This change was made to remove the need to set `--allow-sys` permissions.

### Migration guide

If calling `serveDir()`, set the `printNetworkIp` option to `true` to print the network IP address of the current machine on listen.
```diff
import { serveDir } from "@std/http/file-server";

- Deno.serve((req) => serveDir(req));
+ Deno.serve((req) => serveDir(req, { printNetworkIp: true }));
```

If running `@std/http/file-server` as a script, set the `--ip` flag and `--allow-sys` permission flag to print the network IP address of the current machine on listen.
```diff
- $ deno run --allow-net --allow-read --allow-sys jsr:@std/http/file-server
+ $ deno run --allow-net --allow-read --allow-sys jsr:@std/http/file-server --ip
```

If running `@std/http/file-server` as a script, and you don't want to print the network IP address of the current machine on listen, you can remove the `--allow-sys` permission flag.

```diff
- $ deno run --allow-net --allow-read --allow-sys jsr:@std/http/file-server
+ $ deno run --allow-net --allow-read jsr:@std/http/file-server
```

PR #5547, opened 2024-07-25
fix(file_server): only show LAN address when `--allow-sys` is provided

lucacasonato:
This commit changes the behaviour of the file server to only print the LAN address if `--allow-sys=networkInterfaces` is provided. If it is not provided, no permission prompt for `--allow-sys` will be raised, and the LAN address will not be printed. You could think of this as progressive enhancement :)

Alternate for #5540

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -830,7 +830,13 @@ function main() {
   const useTls = !!(keyFile && certFile);
 
   function onListen({ port, hostname }: { port: number; hostname: string }) {
-    const networkAddress = getNetworkAddress();
+    let networkAddress: string | undefined = undefined;
+    if (
+      Deno.permissions.querySync({ name: "sys", kind: "networkInterfaces" })
+        .state === "granted"
+    ) {
+      networkAddress = getNetworkAddress();
+    }
     const protocol = useTls ? "https" : "http";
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Nice alternative 👍🏾

iuioiua:
Can you please add a test for when `--allow-sys` isn't set?

iuioiua (changes requested):
Can you please add a test for when `--allow-sys` isn't set?

iuioiua (approved):
Great! Thank you.

http/file_server.ts
```diff
@@ -830,7 +830,13 @@ function main() {
   const useTls = !!(keyFile && certFile);
 
   function onListen({ port, hostname }: { port: number; hostname: string }) {
-    const networkAddress = getNetworkAddress();
+    let networkAddress: string | undefined = undefined;
+    if (
+      Deno.permissions.querySync({ name: "sys", kind: "networkInterfaces" })
+        .state === "granted"
```
mxdvl, line 836: I think this will fail in Deno Deploy – proposal to fix in #6033

kt3k: Good catch. Thanks!

PR #5548, opened 2024-07-25
docs: use caption for doc example titles

nayeemrmn:
The titles as written before would mess up the formatting in VSCode. This puts them in `<caption>` tags as sort of suggested in https://jsdoc.app/tags-example (the way it's written there also doesn't work, though). It also needs an empty line after in the case the caption spans multiple lines.

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -76,6 +76,7 @@ const ENTRY_POINTS = [
 ] as const;
 
 const TS_SNIPPET = /```ts[\s\S]*?```/g;
+const CAPTION_TAG = /^<caption>.+<\/caption>/s;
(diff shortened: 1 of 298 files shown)


nayeemrmn:
We're probably gonna replace the caption formatting in the language server instead

iuioiua:
Are you able to provide before/after shots? I can't seem to see where formatting in VSCode is getting messed up.

kt3k:
I don't find problem with existing rendering either.

Also always requiring `<caption></caption>` doesn't feel like ideal DX.

iuioiua:
To clarify Nayeem's previous comment, we'll likely not proceed with this change. But rather, file an issue with VSCode.

PR #5550, opened 2024-07-26
BREAKING(testing): replace `TimeError` exception in favor of built-in error classes in some cases

iuioiua:
### What's changed

Previously, all exceptions within [`FakeTime`](https://jsr.io/@std/testing@1.0.0-rc.4/doc/time/~/FakeTime) threw `TimeError`s. Now, some exceptions throw [`RangeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RangeError)s, and [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError)s, depending on the nature of the error. [`@throws`](https://jsdoc.app/tags-throws) JSDoc tags have also been added to the documentation.

### Motivation

This change was made for two main reasons:
1. Reduce the number of imported symbols required to error-handle `FakeTime`.
2. Provide more context behind the nature of the errors, via the error class names, and therefore allow errors to be more easily differentiated within `FakeTime`.

### Migration guide

To migrate, when error-handling `FakeTime`, use the appropriate error class for the given error message (see table below for reference).

```diff
- import { FakeTime, TimeError } from "@std/testing/time";
+ import { FakeTime } from "@std/testing/time";

try {
  new FakeTime.restore(NaN);
} catch (error) {
- if (error instanceof TimeError && error.message === "Invalid start") {
+ if (error instanceof TypeError && error.message === "Invalid start time") {
    // ...
  }
}
```

| Error message                | Corresponding error class |
|------------------------------|---------------------------|
| `"Invalid start time"`            | `TypeError`               |
| `"Time cannot go backwards"` | `RangeError`              |

### Related

Towards #5476


testing/time.ts
```diff
@@ -131,12 +107,12 @@ function fakeSetTimeout(
   // deno-lint-ignore no-explicit-any
   ...args: any[]
 ): number {
-  if (!time) throw new TimeError("Time is not faked");
+  if (!time) throw new ReferenceError("Time is not faked");
```
iuioiua, line 110: Note: made to be similar to when a non-existent variable is referenced ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Not_defined)).

testing/time.ts
```diff
@@ -288,7 +264,7 @@ export class FakeTime {
     start?: number | string | Date | null,
     options?: FakeTimeOptions,
   ) {
-    if (time) throw new TimeError("Time is already faked");
+    if (time) throw new TypeError("Time is already faked");
```
iuioiua, line 270: Note: made to be similar to when a constant tried to get re-assigned ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_const_assignment)).

testing/time.ts
```diff
@@ -373,7 +349,7 @@ export class FakeTime {
    * ```
    */
   static restore() {
-    if (!time) throw new TimeError("Time is already restored");
+    if (!time) throw new TypeError("Time is already restored");
```
iuioiua, line 357: Note: made to be similar to when a constant tried to get re-assigned ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_const_assignment)).

iuioiua:
@KyleJune, does this change seem reasonable to you?

KyleJune:
> @KyleJune, does this change seem reasonable to you?

Seems reasonable to me. I don't have any issues with this change.

testing/time.ts
kt3k: I don't see the similarity with that situation. How about `Error`?

testing/time.ts
kt3k: `TypeError` looks strange to me here because this is more related to 'state' of fake time, not 'type'.

testing/time.ts
kt3k: Ditto. This feels like error about the state, not the type. I would use 'Error' if we need to use native error types.

iuioiua: Fair point. Changed to `Error`.

testing/time.ts
iuioiua: Updated to `Error`.

iuioiua: Well, another way to think about it is you're trying to change `time` to be `undefined`, when it can only be changed to be defined. Does that make sense?

kt3k: `time` variable is implementation detail and not visible to the users.

testing/time.ts
kt3k: Similarly to the case of restore, `time` variable referenced here is implementation detail, and not visible to the users.

iuioiua: Fair point. Ok, what if we used a [`NotSupportedError`](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#notsupportederror) `DOMException` instead for these cases ("Time is not faked", "Time is already faked" and "Time is already restored")? Alternatively, we could keep `TimeError` for these cases, as standard JS errors don't seem to cover this use case. WDYT?

kt3k: `TimeError` sounds reasonable to me

kt3k (approved):
LGTM

PR #5551, opened 2024-07-26
test(cli): stop spinner without error

kt3k:
Currently most spinner test scripts end the execution with errors (because `setTimeout(spinner.stop, num)` throws instead of correctly stopping the spinner).

This PR fixes them by correctly stopping the spinners.

(This probably fixes https://github.com/denoland/std/issues/5541 )

### DIFF
--- a/cli/testdata/spinner_cases/change_message.ts
+++ b/cli/testdata/spinner_cases/change_message.ts
@@ -7,4 +7,4 @@ spinner.start();
 setTimeout(() => (spinner.message = "One dino 🦕"), 125); // 150
 setTimeout(() => (spinner.message = "Two dinos 🦕🦕"), 200); // 225
 
-setTimeout(spinner.stop, 500);
+setTimeout(() => spinner.stop(), 500);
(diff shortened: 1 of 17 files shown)


cli/testdata/spinner_cases/change_message.ts
```diff
@@ -7,4 +7,4 @@ spinner.start();
 setTimeout(() => (spinner.message = "One dino 🦕"), 125); // 150
 setTimeout(() => (spinner.message = "Two dinos 🦕🦕"), 200); // 225
 
-setTimeout(spinner.stop, 500);
+setTimeout(() => spinner.stop(), 500);
```
iuioiua, line 10: Isn't this the same?

kt3k: No. The existing call doesn't have access to `this` inside the method, and errors at `this.#intervalId`

iuioiua (approved):

cli/testdata/spinner_cases/change_message.ts
iuioiua: Interesting... That's what's blocking #5497.

Issue #5557, opened 2024-07-26
fs: `walk` on Windows does not return an absolute path


ryuapp:
**Describe the bug**

walk function should return absolute paths, but it on Windows does not return that.

**Steps to Reproduce**

```sh
# Dirctory
C:.
    deno.json
    deno.lock
    main.ts
```

```ts
// main.ts
import { walk } from "@std/fs/walk";
console.log(await Array.fromAsync(walk(".")))
```

```ts
// Result
[
  {
    path: ".",
    name: ".",
    isFile: false,
    isDirectory: true,
    isSymlink: false
  },
  {
    path: "deno.json",
    name: "deno.json",
    isFile: true,
    isDirectory: false,
    isSymlink: false
  },
  {
    path: "deno.lock",
    name: "deno.lock",
    isFile: true,
    isDirectory: false,
    isSymlink: false
  },
  {
    path: "main.ts",
    name: "main.ts",
    isFile: true,
    isDirectory: false,
    isSymlink: false
  }
]
```

**Expected behavior**

Return absolute paths.

**Environment**

- OS: Win10
- deno version: 1.45.3
- std version: `@std/fs@1.0.0`

jsejcksn:
It is by design that `walk` returns paths relative to the root dir (first parameter). For context, see:

- https://github.com/denoland/std/issues/3456

If you'd like absolute paths for all entries, you can write a simple abstraction that uses [`Deno.realPath`](https://docs.deno.com/api/deno/~/Deno.realPath)/[`Deno.realPathSync`](https://docs.deno.com/api/deno/~/Deno.realPathSync) like this:

```ts
// main.ts

import { walk } from "@std/fs/walk";

const walkAbsolute: typeof walk = async function* (...args) {
  for await (const entry of walk(...args)) {
    entry.path = await Deno.realPath(entry.path);
    yield entry;
  }
};

console.log(await Array.fromAsync(walkAbsolute(".")));
```

ryuapp:
@jsejcksn 
Thank you. It works well.
The reason I mistook this for a bug was because the JSDocs had examples of it returning an absolute path.

jsejcksn:
> The reason I mistook this for a bug was because the JSDocs had examples of it returning an absolute path.

[...]

PR #5559, opened 2024-07-26
docs(fs): fix incorrect examples for walk

ryuapp:
Related: #5557 

Even though a walk function originally returns relative paths, it gave examples of absolute paths, so I misunderstood.
This PR fixes it.

### DIFF
--- a/fs/walk.ts
+++ b/fs/walk.ts
@@ -97,7 +97,7 @@ export type { WalkEntry };
  * Recursively walks through a directory and yields information about each file
  * and directory encountered.
  *
- * The file paths are absolute paths. The root directory is included in the
+ * The file paths are relative paths. The root directory is included in the
  * yielded entries.
  *
  * Requires `--allow-read` permission.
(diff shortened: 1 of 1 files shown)


fs/walk.ts
```diff
@@ -97,7 +97,7 @@ export type { WalkEntry };
  * Recursively walks through a directory and yields information about each file
  * and directory encountered.
  *
- * The file paths are absolute paths. The root directory is included in the
+ * The file paths are relative paths. The root directory is included in the
```
kt3k, line 100: Looks like it depends on the input. If you pass an absolute path to `walk`, then it seems returning absolute paths.

```ts
import { walk } from "@std/fs";
for await (const item of walk("/Users/kt3k/denoland/std/regexp")) {
  console.log(item);
}
```

```
{
  path: "/Users/kt3k/denoland/std/regexp",
  name: "regexp",
  isFile: false,
  isDirectory: true,
  isSymlink: false
}
{
  path: "/Users/kt3k/denoland/std/regexp/escape_test.ts",
  name: "escape_test.ts",
  isFile: true,
  isDirectory: false,
  isSymlink: false
}
...
```

ryuapp: You are right. I changed the expression as following:
> The root path determines whether the file paths is relative or absolute.

fs/walk.ts
```diff
@@ -97,8 +97,8 @@ export type { WalkEntry };
  * Recursively walks through a directory and yields information about each file
  * and directory encountered.
  *
- * The file paths are absolute paths. The root directory is included in the
- * yielded entries.
+ * The root path determines whether the file paths is relative or absolute.
```
kt3k, line 100:
```suggestion
 * The root path determines whether the file paths are relative or absolute.
```

kt3k (approved):
LGTM. Thanks!

fs/walk.ts
```diff
@@ -520,8 +520,8 @@ export async function* walk(
...
```
jsejcksn, line 523:
```suggestion
 * The root path determines whether the file paths are relative or absolute.
```

jsejcksn (commented):
Thanks for making these documentation updates. This is the only typo I saw that still needs to be changed.

fs/walk.ts
ryuapp: This mistake was fixed before merging :D 
-> https://github.com/denoland/std/pull/5559/commits/a434f65f551cdfde4d4b74c075e11d785c0d93e7

jsejcksn: @ryuapp You're absolutely right — I'm not sure why I wasn't shown the changes from that commit when I clicked to review… looks good!

PR #5560, opened 2024-07-27
fix(async): `abortableAsyncIterable()` should call `.return()` on the generator once aborted

raaymax:
Fixes https://github.com/denoland/std/issues/3518


iuioiua (approved):
Apologies for the delay, and thank you for contributing! LGTM.

iuioiua:
I realise that `return()` should be called in a `finally` block, and `throw()` should be called in a `catch` block.

raaymax:
Ach yes, there is missing part but it's not just that. I created new PR that handles this one.

PR #5561, opened 2024-07-28
feat(assert): add `options` parameter to `AssertionError` constructor

jsejcksn:
Ref:

- https://github.com/denoland/std/issues/5476#issuecomment-2254261250

### DIFF
--- a/assert/assertion_error.ts
+++ b/assert/assertion_error.ts
@@ -8,21 +8,13 @@
  * ```ts no-eval
  * import { AssertionError } from "@std/assert";
  *
- * // Throw an AssertionError:
- * throw new AssertionError("Assertion failed");
- *
- * // Throw an AssertionError with a specified cause:
- * const originalError = new Error("foo");
  * try {
- *   try {
- *     throw originalError;
- *   } catch (cause) {
- *     throw new AssertionError("bar", { cause });
+ *   throw new AssertionError("foo", { cause: "bar" });
+ * } catch (error) {
+ *   if (error instanceof AssertionError) {
+ *     error.message === "foo"; // true
+ *     error.cause === "bar"; // true
  *   }
- * } catch (cause) {
- *   cause instanceof AssertionError && // true
- *     cause.message === "bar" && // true
- *     cause.cause === originalError; // true
  * }
  * ```
  */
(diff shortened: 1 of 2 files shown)


iuioiua (changes requested):
Are you able to add a test the ensures that the `cause` correctly propagates? Also, can you please add an example showing this behavior?

jsejcksn:
@iuioiua

> Are you able to add a test the ensures that the `cause` correctly propagates?

I didn't see a test module for this class, but I'm happy to create one. I'll add an equality test for the error `message` property parameter as well.

> Also, can you please add an example showing this behavior?

I'm not sure what you have in mind. Do you essentially want me to copy the body of the test? If not, can you make a suggestion?

jsejcksn:
@iuioiua I added tests using actual control flow. As a contrived but simpler alternative, it could look like this:

```ts
Deno.test("AssertionError", () => {
  const originalError = new Error("foo");
  const assertionError = new AssertionError("bar", { cause: originalError });
  assertStrictEquals(assertionError.message, "bar");
  assertStrictEquals(assertionError.cause, originalError);
});

```

…but that doesn't represent intended usage in exception-handling control flow. With that juxtaposition available, what are your current thoughts about the following?

1. the structure/style of the tests
2. the content of the example documentation

jsejcksn:
Aside: I'm seeing ci failures in the lint steps ([example](https://github.com/denoland/std/actions/runs/10131143544/job/28013460563?pr=5561)). They are for the rule [`no-unreachable`](https://lint.deno.land/rules/no-unreachable), and I think they shouldn't be happening (the tests prove that the lint diagnostic is wrong). Perhaps a new issue should be raised in that repository…

iuioiua (approved):
I've cleaned up the example and test (they don't have to be too thorough, as this error is used to a large extend throughout testing in the codebase). Now, LGTM! Thanks!

jsejcksn:
> they don't have to be too thorough

@iuioiua Ok, sounds good. I did make one more change: using strict equality comparison for the `error.cause` because I think it's important to illustrate that they aren't just the same shapes, but actually the same value.

Do you have any thoughts about https://github.com/denoland/std/pull/5561#issuecomment-2254477232?

iuioiua:
> 29bb954

Yes please.

jsejcksn:
> > [29bb954](https://github.com/denoland/std/commit/29bb95420bbb54a29c10723e2bae02b9de1cb4fd)
> 
> Yes please.

Ref:

- https://github.com/denoland/deno_lint/issues/1303

PR #5563, opened 2024-07-29
fix(http): update localhost strings in tests

iuioiua:
Fixes https://github.com/denoland/std/actions/runs/10136063768/job/28024339317?pr=5561


kt3k (approved):
LGTM

PR #5564, opened 2024-07-29
fix(semver): correctly remove spaces between operators and versions in `parseRange()`

iuioiua:
Missed in #4375

### DIFF
--- a/semver/parse_range.ts
+++ b/semver/parse_range.ts
@@ -406,7 +406,7 @@ function parseOperatorRanges(string: string): Comparator[] {
 export function parseRange(range: string): Range {
   return range
     // remove spaces between operators and versions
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM

semver/parse_range.ts
```diff
@@ -406,7 +406,7 @@ function parseOperatorRanges(string: string): Comparator[] {
 export function parseRange(range: string): Range {
   return range
     // remove spaces between operators and versions
-    .replaceAll(/(?<=<|>|=) +/g, "")
+    .replaceAll(/(?<=<|>|=|~) +/g, "")
```
kt3k, line 409: Do we also need to take care of `^`?

PR #5570, opened 2024-07-29
BREAKING(testing): replace `MockError` exception in favor of built-in error classes in some cases

iuioiua:
### What's changed

Previously, [`spy()`](https://jsr.io/@std/testing@1.0.0-rc.5/doc/mock/~/spy) and [`stub()`](https://jsr.io/@std/testing@1.0.0-rc.5/doc/mock/~/stub) threw [`MockError`](https://jsr.io/@std/testing@1.0.0-rc.5/doc/mock/~/MockError) when the property for the given class was not a method or not configurable. Now, in these cases, [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) is thrown. The purpose of `MockError` has also been clarified by updating the description.

### Motivation

This change was made to provide clearer and more accurate context on the error being thrown.

### Migration guide

To migrate, when error-handling `spy()` or `stub()`, compare against `TypeError` rather than `MockError` when checking if the instance methods are correct.

```diff
- import { spy, MockError } from "@std/testing/mock";
+ import { spy } from "@std/testing/mock";

const obj = {};

try {
  spy(obj as any, "fn");
} catch (error) {
- if (error instanceof MockError) {
+ if (error instanceof TypeError) {
    console.error("Property is not an instance method");
  }
}
```

```diff
- import { spy, MockError } from "@std/testing/mock";
+ import { spy } from "@std/testing/mock";

const obj = { fn() {} };
Object.defineProperty(obj, "fn", { configurable: false })

try {
  spy(obj, "fn");
} catch (error) {
- if (error instanceof MockError) {
+ if (error instanceof TypeError) {
    console.error("Cannot spy on non-configurable instance method");
  }
}
```

### Related

Towards #5476

### DIFF
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -333,9 +333,19 @@ import { AssertionError } from "@std/assert/assertion-error";
  * import { MockError, spy } from "@std/testing/mock";
  * import { assertThrows } from "@std/assert";
  *
- * assertThrows(() => {
- *   spy({} as any, "no-such-method");
- * }, MockError);
+ * class Point {
+ *   x: number;
+ *   y: number;
+ *
+ *   constructor(x: number, y: number) {
+ *     this.x = x;
+ *     this.y = y;
+ *   }
+ * }
+ *
+ * const PointSpy = spy(Point);
+ *
+ * assertThrows(() => PointSpy.restore(), MockError, "Constructor cannot be restored");
  * ```
  */
 export class MockError extends Error {
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -325,7 +325,8 @@ import { assertRejects } from "@std/assert/rejects";
 import { AssertionError } from "@std/assert/assertion-error";
 
 /**
- * An error related to spying on a function or instance method.
+ * Represents an error when trying to execute an invalid mocking or spying
+ * operation.
  *
  * @example Usage
  * ```ts
--- a/testing/mock.ts
+++ b/testing/mock.ts
@@ -643,7 +643,7 @@ function methodSpy<
   Return,
 >(self: Self, property: keyof Self): MethodSpy<Self, Args, Return> {
   if (typeof self[property] !== "function") {
-    throw new MockError("Property is not an instance method");
+    throw new TypeError("Property is not an instance method");
   }
   if (isSpy(self[property])) {
     throw new MockError("Already spying on instance method");
@@ -651,7 +651,7 @@ function methodSpy<
 
   const propertyDescriptor = Object.getOwnPropertyDescriptor(self, property);
   if (propertyDescriptor && !propertyDescriptor.configurable) {
-    throw new MockError("Cannot spy on non-configurable instance method");
+    throw new TypeError("Cannot spy on non-configurable instance method");
   }
 
   const original = self[property] as unknown as (
@@ -948,6 +948,8 @@ export function spy<
  * @param self The instance to spy.
  * @param property The property of the method to spy.
  * @returns The spy function.
+ * @throws {TypeError} If the property is not an instance method or is not
+ * configurable.
  */
 export function spy<
   Self,
@@ -1060,6 +1062,8 @@ export function stub<
  * @param property The property of the instance to replace.
  * @param func The fake implementation of the function.
  * @returns The stub function which replaced the original.
+ * @throws {TypeError} If the property is not an instance method or is not
+ * configurable.
  */
 export function stub<
   Self,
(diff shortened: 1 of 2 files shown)


kt3k:
It throws sometimes TypeError and sometimes MockError. This feels confusing to me.

The current implementation consistently throws MockError for any wrong usage of mock APIs. That feels more easier to remember and easier to understand to me.

Is this change worth it?

iuioiua:
APIs that throw different types of errors is a fairly normal thing. E.g. [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#exceptions). I think it's important that we throw errors in a consistent and smart manner across the codebase, according to the error at-hand. Recent work on errors has improved this situation a fair amount and this is the last spot remaining.

testing/mock.ts
```diff
@@ -643,15 +654,15 @@ function methodSpy<
   Return,
 >(self: Self, property: keyof Self): MethodSpy<Self, Args, Return> {
   if (typeof self[property] !== "function") {
-    throw new MockError("Property is not an instance method");
+    throw new TypeError("Property is not an instance method");
```
kt3k, line 657: This looks like trying an invalid spying. Why isn't this MockError?

testing/mock.ts
```diff
@@ -643,15 +654,15 @@ function methodSpy<
...
   }
   if (isSpy(self[property])) {
     throw new MockError("Already spying on instance method");
   }
 
   const propertyDescriptor = Object.getOwnPropertyDescriptor(self, property);
   if (propertyDescriptor && !propertyDescriptor.configurable) {
-    throw new MockError("Cannot spy on non-configurable instance method");
+    throw new TypeError("Cannot spy on non-configurable instance method");
```
kt3k, line 665: Similarly, why isn't this MockError?

testing/mock.ts
```diff
@@ -1082,15 +1097,15 @@ export function stub<
   func?: (this: Self, ...args: Args) => Return,
 ): Stub<Self, Args, Return> {
   if (self[property] !== undefined && typeof self[property] !== "function") {
-    throw new MockError("Property is not an instance method");
+    throw new TypeError("Property is not an instance method");
```
kt3k, line 1100: Ditto. This looks like an error on attempting invalid stubbing. Why isn't this MockError?

testing/mock.ts
```diff
@@ -1082,15 +1097,15 @@ export function stub<
...
   }
   if (isSpy(self[property])) {
     throw new MockError("Already spying on instance method");
   }
 
   const propertyDescriptor = Object.getOwnPropertyDescriptor(self, property);
   if (propertyDescriptor && !propertyDescriptor.configurable) {
-    throw new MockError("Cannot spy on non-configurable instance method");
+    throw new TypeError("Cannot spy on non-configurable instance method");
```
kt3k, line 1108: Ditto. This looks like an error on attempting invalid stubbing. Why isn't this MockError?

kt3k: Another point. The error message probably should use the verb `stub`?
```suggestion
    throw new TypeError("Cannot stub on non-configurable instance method");
```

testing/mock.ts
iuioiua: The property is the wrong type; it should be a function.

testing/mock.ts
iuioiua: Actually, yes, this could be reverted.

iuioiua: [`TypeError`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError) documentation:
> A `TypeError` may be thrown when:
> - an operand or argument passed to a function is incompatible with the type expected by that operator or function; or
> - when attempting to modify a value that cannot be changed; or
> - when attempting to use a value in an inappropriate way.

testing/mock.ts
kt3k: The choice between TypeError and MockError seems arbitrary and confusing to me.

iuioiua: Mocking fails because the property is the wrong type. There's overlap, but I feel to see what's confusing.

iuioiua:
Are we able to get one more opinion on this? Happy to drop if this if I'm the only one that thinks this is a good idea. CC @magurotuna @KyleJune

testing/mock.ts
iuioiua: Let me address the error message in a separate PR.

KyleJune:
I think like in https://github.com/denoland/std/pull/5550, the errors that ended up being changed could have gone either way. It's pretty arbitrary whether they are a TypeError or MockError. I have no strong opinions on keeping it the same or changing it. The only reason I created TimeError and MockError was to tell them apart from other errors and to make it more clear what they are related to.

iuioiua:
I think by not merging this, we're a little inconsistent. However, it seems there isn't sufficient support for this change.

PR #5571, opened 2024-07-29
fix(http): better `eTag` return type for `string` and `Uint8Array` inputs

kt3k:
Not completely sure this is a breaking change, but this PR improves the return type of `eTag` function. Now it returns `Promise<string>` if the input is `string` or `Uint8Array`.

related #5217

### DIFF
--- a/http/etag.ts
+++ b/http/etag.ts
@@ -95,7 +95,7 @@ async function calcFileInfo(
 }
 
 /**
- * Calculate an ETag for string or Uint8Array entities. This returns a strong tag.
+ * Calculate an ETag for string or `Uint8Array` entities. This returns a strong tag.
  *
  * @example Usage
  * ```ts
--- a/http/etag.ts
+++ b/http/etag.ts
@@ -95,9 +95,7 @@ async function calcFileInfo(
 }
 
 /**
- * Calculate an ETag for an entity. When the entity is a specific set of data
- * it will be fingerprinted as a "strong" tag, otherwise if it is just file
- * information, it will be calculated as a weak tag.
+ * Calculate an ETag for string or Uint8Array entities. This returns a strong tag.
  *
  * @example Usage
  * ```ts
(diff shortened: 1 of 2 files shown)


http/etag_test.ts
iuioiua, line 1: Good to see `@std/testing/types` being used to test overload signatures. We should probably do this more throughout the codebase.

http/etag.ts
```diff
@@ -116,6 +116,38 @@ async function calcFileInfo(
...
  * @returns The calculated ETag.
  */
+export async function eTag(
+  entity: string | Uint8Array,
+  options?: ETagOptions,
+): Promise<string>;
+/**
+ * Calculate an ETag for an entity. When the entity is a specific set of data
+ * it will be fingerprinted as a "strong" tag, otherwise if it is just file
+ * information, it will be calculated as a weak tag.
```
iuioiua, line 126: We should update the descriptions to orient themselves to their respective overloads.

iuioiua (changes requested):

iuioiua:
I'd also change the title to something like "Improve overloads for `eTag()`". And I don't think this is a breaking change.

http/etag.ts
```diff
@@ -95,9 +95,7 @@ async function calcFileInfo(
 }
 
 /**
- * Calculate an ETag for an entity. When the entity is a specific set of data
- * it will be fingerprinted as a "strong" tag, otherwise if it is just file
- * information, it will be calculated as a weak tag.
+ * Calculate an ETag for string or Uint8Array entities. This returns a strong tag.
```
iuioiua, line 98: Can you please clarify what you mean by "strong tag"?

iuioiua, line 98:
```suggestion
 * Calculate an ETag for string or `Uint8Array` entities. This returns a strong tag.
```

http/etag.ts
```diff
@@ -116,6 +114,36 @@ async function calcFileInfo(
  * @param options Various additional options.
  * @returns The calculated ETag.
  */
+export async function eTag(
+  entity: string | Uint8Array,
+  options?: ETagOptions,
+): Promise<string>;
+/**
+ * Calculate an ETag for file information entity. This returns a weak tag.
```
iuioiua, line 122: Ditto for "weak tag".

iuioiua (approved):
LGTM once these last nits are addressed.

iuioiua (approved):

PR #5562, opened 2024-07-29
chore(ini): release `ini@1.0.0`

iuioiua:
To be merged July 31.

Documentation: https://jsr.io/@std/ini@1.0.0-rc.2

Closes #5017

PR #5572, opened 2024-07-29
BREAKING(ini): reduce options for `stringify`, make `FormattingOptions` type private

kt3k:
This PR removes `commentChar`, `assignment`, and `deduplicate` options from `stringfiy` options.

- `deduplicate` is no longer relevant as we dropped the support of `IniMap`, which was only meaningful with `IniMap`
- `commentChar` is no longer relevant as stringify doesn't support comments in output.
- Dropped `assignment` option as `=` seems the only common character for assignment in `INI`.

This PR also make `FormattingOptions` private as it's no longer reasonable to keep it public.

related #5562

### DIFF
--- a/ini/stringify.ts
+++ b/ini/stringify.ts
@@ -12,7 +12,7 @@ export interface StringifyOptions {
    */
   lineBreak?: "\n" | "\r\n" | "\r";
   /**
-   * Use a plain assignment char or pad with spaces. Ignored on parse.
+   * Use a plain assignment char or pad with spaces.
    *
    * @default {false}
    */
--- a/ini/stringify.ts
+++ b/ini/stringify.ts
@@ -1,14 +1,10 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 
-import {
-  type FormattingOptions,
-  IniMap,
-  type ReplacerFunction,
-} from "./_ini_map.ts";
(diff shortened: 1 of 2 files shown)


ini/stringify.ts
```diff
@@ -1,14 +1,22 @@
...
-  IniMap,
-  type ReplacerFunction,
-} from "./_ini_map.ts";
+import { IniMap, type ReplacerFunction } from "./_ini_map.ts";
 
 /** Options for {@linkcode stringify}. */
-export interface StringifyOptions extends FormattingOptions {
+export interface StringifyOptions {
+  /**
+   * Character(s) used to break lines in the config file. Ignored on parse.
```
iuioiua, line 9:
```suggestion
   * Character(s) used to break lines in the config file.
```

ini/stringify.ts
```diff
@@ -1,14 +1,22 @@
...
+   *
+   * @default {"\n"}
+   */
+  lineBreak?: "\n" | "\r\n" | "\r";
+  /**
+   * Use a plain assignment char or pad with spaces. Ignored on parse.
```
iuioiua, line 15:
```suggestion
   * Use a plain assignment char or pad with spaces.
```

iuioiua (approved):
Nice. We should clean up any logic in `IniMap`, as a result of these changes, either in this PR or a follow-up.

PR #5573, opened 2024-07-29
chore(assert): mark `options` argument of `AssertionError` constructor unstable

kt3k:


iuioiua (approved):
IMO, this PR doesn't achieve anything, given how trivial this addition is and it's unlikeliness to change in the future. Nevertheless, approved.

PR #5575, opened 2024-07-30
fix(testing): correct `stub()` error message

iuioiua:
Addresses https://github.com/denoland/std/pull/5570#discussion_r1696174755


kt3k (approved):
LGTM

PR #5576, opened 2024-07-30
refactor(ini): cleanup dead code

iuioiua:
A lot of this code is no longer used after a few of the `stringify()` options were removed in #5572 and `IniMap` was made private in #5351. #5593 moves the code out of `ini/_ini_map.ts`, making the file obsolete. Testing should continue to be improved within `ini/parse_test.ts` and `ini/stringify_test.ts`.


kt3k (approved):
LGTM

PR #5580, opened 2024-07-30
refactor(io): use `writeAll()` within `copy()`

jsejcksn:
- deduplicate logic by importing `writeAll`
- add JSDoc description of returned value
- add test step for returned value


iuioiua (approved):
Nice! Thank you.

PR #5581, opened 2024-07-30
chore: clarify the policy about new features

kt3k:


iuioiua (approved):
I think it's good that this **doesn't** include timelines. LGTM.

iuioiua:
I think we can make exceptions for APIs that we are highly confident will not yield unexpected behavior. E.g. #5573 is fine to add as stable.

ry (commented):
Do we have a mechanism to mark APIs as unstable? Is there a special jsdoc markup?

kt3k:
> Do we have a mechanism to mark APIs as unstable? Is there a special jsdoc markup?

Yes for symbols https://jsr.io/@std/cli/doc/~/Spinner

But I'm not sure about the way to mark specific parameter in function as unstable. (which is necessary for the case like #5573 )

crowlKats:
@kt3k sadly both jsdoc and tsdoc dont have a way to mark unstable/experimental parameters. the only way i can think of is having overloads, where the overload with an unstable parameter is marked as unstable

kt3k:
@crowlKats Ah, ok. Thanks for the input! Then we probably need to do overloads when an API is partially unstable.

PR #5583, opened 2024-07-31
fix(semver): do not throw in `canParse`

kt3k:
This addressed the concern raised in https://github.com/denoland/std/pull/5220#pullrequestreview-2208098580


iuioiua (approved):
Another good catch

PR #5584, opened 2024-07-31
BREAKING(semver): do not accept undefined input in `tryParse`

kt3k:
# What's changed

The argument of `tryParse` becomes `string` instead of `string | undefined`.

# Motivation

`canParse` and `parse` only accepts `string`. This change aligns `tryParse` to these similar APis.

# Migration guide

```diff
-tryParse(version);
+typeof version === "string" ? tryParse(version) : undefined;
```

### DIFF
--- a/semver/try_parse.ts
+++ b/semver/try_parse.ts
@@ -21,10 +21,7 @@ import { parse } from "./parse.ts";
  * @param version The version string to parse
  * @returns A valid SemVer or `undefined`
  */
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Good catch. Weird that we did this to begin with.

iuioiua:
Addresses https://github.com/denoland/std/pull/5220#pullrequestreview-2208098580

PR #5586, opened 2024-07-31
docs(semver): clarify compare docs

kt3k:
This PR clarifies the description of `compare` docs. The expression `s0 === s1` is misleading as the equality of semvers can't be checked with `===` operator.


iuioiua (approved):
Nice

PR #5587, opened 2024-07-31
BREAKING(semver): remove the handling of non-standard SemVers in format function

kt3k:
# What's changed

`format` currently handles NaN semvers (`{ major: NaN, minor: NaN, patch: NaN }`) or Infinity semvers (such as `{ major: Infinity, minor: Infinity, patch: Infinity }`) in a special way.

This change stops those special handlings.

# Motivation

The infinity SemVers or NaN SemVers don't represent valid semantic versions, and they are not part of the semver spec. It's not reasonable to handle them in a special way, but rather we should consider them invalid versions.

# Migration

Do not use these non-standard semvers as sentinel values. Instead use comparison functions such as `compare`, `greaterThan`, `lessThan`, `greaterThanRange`, `lessThanRange`, etc for checking semver conditions.

### DIFF
--- a/semver/format.ts
+++ b/semver/format.ts
@@ -1,16 +1,9 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 import type { SemVer } from "./types.ts";
-import { isWildcardComparator } from "./_shared.ts";
 
 function formatNumber(value: number) {
-  if (value === Number.POSITIVE_INFINITY) {
-    return "∞";
-  } else if (value === Number.NEGATIVE_INFINITY) {
-    return "⧞";
-  } else {
-    return value.toFixed(0);
-  }
+  return value.toFixed(0);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


semver/format_test.ts
```diff
@@ -26,19 +24,4 @@ Deno.test("format()", async (t) => {
       },
     });
   }
-
-  const constantSemVers: [SemVer, string][] = [
-    [MAX, "∞.∞.∞"],
-    [MIN, "0.0.0"],
```
iuioiua, line 32: Should we maybe still test `MIN` by adding it to the `versions` array above?

semver/format.ts
```diff
@@ -37,10 +26,6 @@ function formatNumber(value: number) {
  * @returns The string representation of a semantic version.
  */
 export function format(semver: SemVer): string {
-  if (isWildcardComparator(semver)) {
-    return "*";
-  }
-
```
iuioiua, line 43: Why was this change made?

iuioiua (approved):
Nice. I agree with this.

semver/format.ts
kt3k: This is also handling of non-standard semvers. If a semver is `{ major: NaN, minor: NaN, patch: NaN }`, then it returns `*`, which is not documented, nor tested. I don't think this is a reasonable behavior.

iuioiua: SGTM

PR #5589, opened 2024-07-31
BREAKING(ini): move `reviver` to 2nd argument in `parse()`

iuioiua:
### What's changed

This PR replaces [`ParseOptions`](https://jsr.io/@std/ini@1.0.0-rc.2/doc/~/ParseOptions) with `reviver` (the function previously defined as [`ParseOptions.reviver`](https://jsr.io/@std/ini@1.0.0-rc.2/doc/~/ParseOptions.reviver)). As a result, the `ParseOptions` interface has been removed as it's no longer used anywhere.

### Motivation

This change was made to simplify working with [`parse()`](https://jsr.io/@std/ini@1.0.0-rc.2/doc/parse/~/parse), as having a whole `options` object was extraneous, given it only included one property.

### Migration guide

To migrate, move the `reviver` function from `options.reviver` to the 2nd argument of `parse()`.

```diff
import { parse } from "@std/ini/parse";
import { assertEquals } from "@std/assert";

function reviver(
  key: string,
  value: string,
  section?: string,
): number | Date | string {
  if (section === "section Foo") {
    if (key === "date") {
      return new Date(value);
    } else if (key === "amount") {
      return +value;
    }
  }
  return value;
}

const iniText = `
[section Foo]
date = 2012-10-10
amount = 12345
`;

- const parsed = parse(iniText, { reviver });
+ const parsed = parse(iniText, reviver);

assertEquals(parsed, {
  "section Foo": {
    date: new Date("2012-10-10"),
    amount: 12345,
  },
});
```

### DIFF
--- a/ini/parse.ts
+++ b/ini/parse.ts
@@ -2,17 +2,7 @@
 // This module is browser compatible.
 
 import { IniMap, type ReviverFunction } from "./_ini_map.ts";
-export type { ParseOptions, ReviverFunction };
-
-/** Options for {@linkcode parse}. */
-interface ParseOptions {
-  /**
-   * Provide custom parsing of the value in a key/value pair. Similar to the
-   * {@link https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse#reviver | reviver}
-   * function in {@linkcode JSON.parse}.
-   */
-  reviver?: ReviverFunction;
-}
+export type { ReviverFunction };
 
 /**
  * Parse an INI config string into an object.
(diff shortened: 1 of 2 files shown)


kt3k:
This doesn't seem a reasonable change to me. The current form is already conformant to the style guide.

This change suggests using the exceptional pattern when additional options don't seem possible in the future, as described below.

https://docs.deno.com/runtime/manual/references/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object.

> An optional parameter that's not in an options object might be acceptable if there is only one, and it seems inconceivable that we would add more optional parameters in the future.

However `parse` is relatively complex function. I think it's very possible for us to add options in the future. The current form prepares for such future development. This change makes such future change difficult.

PR #5592, opened 2024-07-31
fix(ini): correctly handle quoted values in `parse()`

### LINKED ISSUE #4808: ini file value parsed with quotes contains quotes in value
```
> parse('value="value"')
{ value: '"value"' }
```

Should be:

```
{ value: "value" }
```

std 0.224.0

iuioiua:
Fixes #4808


kt3k (approved):
LGTM

Let's keep iterating on other edge cases

PR #5595, opened 2024-07-31
refactor(cli): make regexps constants

timreichen:
- moves regexps outside function and makes them constants.
- adds named capturing group to `VALUE_REGEXP` instead of testing and splitting value afterwards.
- reduces code block with the same regexp check.


iuioiua (approved):
LGTM

PR #5596, opened 2024-08-01
test(async): fix `abortable.AsyncIterable() calls return before throwing` test

iuioiua:
Fixes https://github.com/denoland/std/actions/runs/10189680039/job/28188160624

### DIFF
--- a/async/abortable_test.ts
+++ b/async/abortable_test.ts
@@ -137,9 +137,13 @@ Deno.test("abortable.AsyncIterable() calls return before throwing", async () => 
   const iterable: AsyncIterable<string> = {
     [Symbol.asyncIterator]: () => ({
       next: () =>
-        new Promise((resolve) =>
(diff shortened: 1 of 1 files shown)


async/abortable_test.ts
```diff
@@ -137,9 +137,13 @@ Deno.test("abortable.AsyncIterable() calls return before throwing", async () =>
...
       next: () =>
-        new Promise((resolve) =>
-          setTimeout(() => resolve({ value: "Hello", done: false }), 1)
-        ),
+        new Promise((resolve) => {
+          const timeoutId = setTimeout(
+            () => resolve({ value: "Hello", done: false }),
+            1,
+          );
+          clearTimeout(timeoutId);
```
kt3k, line 145: These 4 lines look equivalent of nop (timer is created and immediately deleted). I wonder if there's a better fix

kt3k (approved):
LGTM

PR #5598, opened 2024-08-01
refactor(cli): use `for of` instead of `for in` loop

timreichen:
- replaces `for in` loop with `for of` loop by using `Object.entries()`.


kt3k (approved):
LGTM

iuioiua (approved):

PR #5599, opened 2024-08-01
chore(cli): simplify `setNested()` and `hasNested()`

timreichen:


iuioiua (approved):
Thanks

cli/parse_args.ts
```diff
@@ -373,12 +373,11 @@ function setNested(
   value: unknown,
   collect = false,
 ) {
-  keys.slice(0, -1).forEach((key) => {
-    object[key] ??= {};
-    object = object[key] as NestedMapping;
-  });
+  keys = [...keys];
+  const key = keys.pop();
+  if (!key) throw new Error(`'keys' cannot be an empty array.`);
```
kt3k, line 378: This check seems always false

iuioiua: Ah, yes. Are you suggesting we just use a non-null assertion?

kt3k: ah, wait, it might be fine. Let me check

kt3k: > Ah, yes. Are you suggesting we just use a non-null assertion?

Sounds good. That was what the previous code was doing at `const key = keys.at(-1)!`

PR #5600, opened 2024-08-01
feat(csv): support `fieldsPerRecord` in `CsvParseStream`

magurotuna:
Although the constructor of `CsvParseStream` accepts `fieldsPerRecord` option (see https://jsr.io/@std/csv@1.0.0-rc.5/doc/~/CsvParseStreamOptions) that ensures that every record has the specified (or inferred from the first row) number of fields, this option has no effect at all in the current implementation.
To fix this issue, this patch implements the `fieldsPerRecord` logic in `CsvParseStream` together with sufficient amount of test cases.

### DIFF
--- a/csv/parse_stream_test.ts
+++ b/csv/parse_stream_test.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 import { CsvParseStream } from "./parse_stream.ts";
 import type { CsvParseStreamOptions } from "./parse_stream.ts";
-import { assertEquals, assertRejects } from "@std/assert";
+import { assert, assertEquals, assertRejects } from "@std/assert";
 import type { AssertTrue, IsExact } from "@std/testing/types";
 import { fromFileUrl, join } from "@std/path";
 import { delay } from "@std/async/delay";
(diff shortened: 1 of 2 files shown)


csv/parse_stream_test.ts
```diff
@@ -341,18 +394,21 @@ x,,,
...
         }
+        if ("fieldsPerRecord" in testCase) {
+          options.fieldsPerRecord = testCase.fieldsPerRecord;
+        }
+        if ("skipFirstRow" in testCase) {
+          options.skipFirstRow = testCase.skipFirstRow;
+        }
+        if ("columns" in testCase) {
+          options.columns = testCase.columns;
+        }
```
magurotuna, line 411: This adds `fieldsPerRecord`, and also reorders `skipFirstRow` and `columns` to match the order of properties declared in `CsvParseStreamOptions` so that it will be easy for us to notice the missing field in the future

csv/parse_stream_test.ts
```diff
@@ -312,18 +354,29 @@ x,,,
...
       },
       {
         name: "bad quote in quoted field",
         input: `"wo"rd",1,2,3`,
-        errorMessage: "Error line: 1\nBad quoting",
+        error: {
+          klass: SyntaxError,
+          msg:
+            'record on line 1; parse error on line 0, column 3: extraneous or missing " in quoted-field',
+        },
```
magurotuna, line 379: In the previous test the expected error messages were not asserted. I fixed this and now the error class and message are checked - this reveals that these actual error messages feel a little bit unnatural, in the sense that the first sentence says "line 1" but the second says "parse error on line 0". I think this is inconsistent indexing.
This is not really related to what I'd like to cover in this PR, but maybe we'd like to address it before v1 too?

kt3k (approved):
Oh, good catch and nice fix! Thanks!

PR #5602, opened 2024-08-01
refactor(cli): remove `broken` variable in favour of loop label

timreichen:
- replaces `broken` boolean and just continues the loop directly by adding loop label.


iuioiua (approved):
Thank you

PR #5603, opened 2024-08-01
BREAKING(http): rename `STATUS_CODE` to `STATUS_CODES` and `STATUS_TEXT` to `STATUS_TEXTS`

iuioiua:
### What's changed

The `STATUS_CODE` constant has been renamed to `STATUS_CODES`, and the `STATUS_TEXT` constant has been renamed to `STATUS_TEXTS`.

### Motivation

The names of these constants were renamed to provide names that correctly describe what these symbols are. `STATUS_CODE` is not a "status code". Rather, it is an object whose entries are status codes. Similarly with `STATUS_TEXT`.

Also, [`node:http`](https://nodejs.org/api/http.htm) has [`STATUS_CODES`](https://nodejs.org/api/http.html#httpstatus_codes) and [`METHODS`](https://nodejs.org/api/http.html#httpmethods), which are also objects. Changing to the wrong name was made in #3781 by me.

### Migration guide

To migrate, use `STATUS_CODES` instead of `STATUS_CODE` and `STATUS_TEXTS` instead of `STATUS_TEXT`.

```diff
- import { STATUS_CODE, STATUS_TEXT } from "@std/http/status";
+ import { STATUS_CODES, STATUS_TEXTS } from "@std/http/status";

- STATUS_TEXT[STATUS_CODE.NotFound]; // "Not Found"
+ STATUS_TEXTS[STATUS_CODES.NotFound]; // "Not Found"
```

### Related

Addresses https://github.com/denoland/std/pull/5217#issuecomment-2262783646

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -43,8 +43,8 @@ import { contentType } from "@std/media-types/content-type";
 import { eTag, ifNoneMatch } from "./etag.ts";
 import {
   isRedirectStatus,
-  STATUS_CODE,
-  STATUS_TEXT,
+  STATUS_CODES,
+  STATUS_TEXTS,
   type StatusCode,
 } from "./status.ts";
 import { ByteSliceStream } from "@std/streams/byte-slice-stream";
@@ -90,7 +90,7 @@ function modeToString(isDir: boolean, maybeMode: number | null): string {
 }
 
 function createStandardResponse(status: StatusCode, init?: ResponseInit) {
-  const statusText = STATUS_TEXT[status];
+  const statusText = STATUS_TEXTS[status];
   return new Response(statusText, { status, statusText, ...init });
 }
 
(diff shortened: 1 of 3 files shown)


kt3k:
I'm not in favor of this change. The reasons are:
- Singular form for enum like object looks just fine to me.
- It's very late to start discussing this kind of idea (one day before the stabilization)

bartlomieju (changes requested):
I'm strongly against this change

BlackAsLight:
I'm against this change. It makes sense for enums to be singular. When you get a value from it, it is only a single value in the form of dot notation. If it was more in the form of an array then I would agree it should be plural as it is an object containing status codes, but as an enum it is a status code of a value yet to be decided.

iuioiua:
Ok. That's a sufficient amount of sentiment against. Dropping. Thanks for your input, guys.

PR #5604, opened 2024-08-01
fix(csv): show 1-based line and column numbers in error messages

magurotuna:


iuioiua (approved):
Great work! Thank you

PR #5605, opened 2024-08-01
docs(csv): more examples for `parse` and `CsvParseStream`

magurotuna:
This commit adds more examples to `parse` function and `CsvParseStream` class to cover all the provided options.

Also fixes a few other things:

- Replace stale description `ParseError` with the correct `SyntaxError`.
- Fix the default value of `comment` property. The old comment says the default value is `#`, but this was wrong.
- Get negative value in `fieldsPerRecord` option working in `parse` as documented

### DIFF
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -261,7 +261,7 @@ export interface ParseOptions {
    * are ignored. With leading whitespace the comment character becomes part of
    * the field, even you provide `trimLeadingSpace: true`.
    *
-   * @default {"#"}
+   * By default, no character is considered to be a start of a comment.
    */
   comment?: string;
   /** Flag to trim the leading space of the value.
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -306,8 +306,7 @@ export interface ParseOptions {
 }
 
 /**
- * Csv parse helper to manipulate data.
- * Provides an auto/custom mapper for columns.
+ * Parses CSV string into an array of arrays of strings.
  *
  * @example Usage
  * ```ts
@@ -324,8 +323,10 @@ export interface ParseOptions {
  */
 export function parse(input: string): string[][];
 /**
- * Csv parse helper to manipulate data.
- * Provides an auto/custom mapper for columns.
+ * Parses CSV string into an array of objects or an array of arrays of strings.
+ *
+ * If `column` or `skipFirstRow` option is provided, it returns an array of
+ * objects, otherwise it returns an array of arrays of string.
  *
  * @example Usage
  * ```ts
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -505,9 +505,9 @@ export function parse<const T extends ParseOptions>(
       headers = options.columns;
     }
 
-    const firstLineIndex = options.skipFirstRow ? 1 : 0;
+    const zeroBasedFirstLineIndex = options.skipFirstRow ? 1 : 0;
     return r.map((row, i) => {
-      return convertRowToObject(row, headers, firstLineIndex + i);
+      return convertRowToObject(row, headers, zeroBasedFirstLineIndex + i);
     }) as ParseResult<ParseOptions, T>;
   }
   return r as ParseResult<ParseOptions, T>;
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -208,7 +208,7 @@ class Parser {
     this.#input = input.startsWith(BYTE_ORDER_MARK) ? input.slice(1) : input;
     this.#cursor = 0;
     const result: string[][] = [];
-    let _nbFields: number | undefined;
+
     let lineResult: string[];
     let first = true;
     let lineIndex = 0;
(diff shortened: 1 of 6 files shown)


csv/parse.ts
```diff
@@ -333,26 +347,137 @@ export function parse(input: string): string[][];
...
+ * ```ts
+ * import { parse } from "@std/csv/parse";
+ * import { assertEquals } from "@std/assert/equals";
+ * import { assertType, IsExact } from "@std/testing/types"
+ *
+ * const string = "a,b,c\nd,e,f";
+ * const result = parse(string, { skipFirstRow: true });
+ *
+ * assertEquals(result, [{ a: "d", b: "e", c: "f" }]);
+ * assertType<IsExact<typeof result, Record<string, string | undefined>[]>>(true);
```
magurotuna, line 376: I added `assertType` here because I think this is a very good document for users given that the return type varies depending on the option we pass in.

While I was writing these assertions, though, I was unsure in what cases the value in the return value becomes `undefined`. Does anyone know why we have `Record<string, string | undefined>` instead of `Record<string, string>` here?

csv/parse.ts
```diff
@@ -333,26 +347,139 @@ export function parse(input: string): string[][];
...
+ * import { parse } from "@std/csv/parse";
+ * import { assertEquals } from "@std/assert/equals";
+ *
+ * const string = "a\tb\tc\nd\te\tf";
+ * const result = parse(string, { separator: "\t" });
+ *
+ * assertEquals(result, [["a", "b", "c"], ["d", "e", "f"]]);
+ * ```
+ *
+ * @example Trim leading space with `trimLeadingSpace: true`
```
iuioiua, line 417: Is there a reason there's an option to trim only leading spaces? Why not just trim the end too using `String.prototype.trim()`?

iuioiua (approved):
Great work

csv/parse.ts
magurotuna: I'm not sure either, but according to the head comment of parse.ts, many parts of the implementation was ported from Go which has the same option.
https://github.com/golang/go/blob/go1.12.5/src/encoding/csv/reader.go#L136

iuioiua: I can't think of a reason to only trim the start... I suggest changing to `trim: boolean`. Not a strong opinion. What do we think?

csv/parse.ts
magurotuna: I tried the following three test cases and these pass. I couldn't find a case where the parsed value becomes `undefined`

```ts
    await t.step({
      name: "BlankField2",
      fn() {
        const input = "a,b,c\nd,,f";
        assertEquals(
          parse(input, { skipFirstRow: true }),
          [{ a: "d", b: "", c: "f" }],
        );
      },
    });

    await t.step({
      name: "Exessive fields with skipFirstRow: true",
      fn() {
        const input = "a,b,c\nd,,f,g";
        assertThrows(
          () => parse(input, { skipFirstRow: true }),
          Error,
          "Error number of fields line: 1\nNumber of fields found: 3\nExpected number of fields: 4",
        );
      },
    });

    await t.step({
      name: "Insufficient fields with skipFirstRow: true",
      fn() {
        const input = "a,b,c\nd,e";
        assertThrows(
          () => parse(input, { skipFirstRow: true }),
          Error,
          "Error number of fields line: 1\nNumber of fields found: 3\nExpected number of fields: 2",
        );
      },
    });
```

csv/parse.ts
magurotuna: Asked ChatGPT, it answered some application may add leading whitespaces when exporting to CSV. I don't know if this is true at all 😄 Anyway `trim: boolean` should be useful in some scenarios and I think we could add that after 1.0 without breaking things?

csv/parse.ts
magurotuna: Added (or fixed the existing) tests on what happens if the number of fields in records doesn't match that of the header. This fix contains the change in how we display the line number is error messages (which was 0-based but now changed to 1-based).

a91806b

csv/parse.ts
kt3k: I think trailing spaces are usually considered as a part of value of the last column (For example, excel handles it in that way)

kt3k (approved):
LGTM

csv/parse.ts
kt3k: I couldn't find the case when this returns `Record<string, string | undefined>`, but I  also noticed that `fieldsPerRecord` option doesn't seem working as described.

It says:
> If negative, no check is made and records may have a variable number of fields.

But `parse` doesn't seem allowing wrong number of fields even when I passed negative number to this option. Maybe this current state is related to it?

magurotuna: > But parse doesn't seem allowing wrong number of fields even when I passed negative number to this option. Maybe this current state is related to it?

Oh that's a good find. I'll add a test case and fix it

PR #5607, opened 2024-08-01
refactor(cli): `isNumber()`

timreichen:
- combines `isNumber()` regexps to a single `IS_NUMBER_REGEXP` constant
- removes obsolete string casting inside `isNumber()`


cli/parse_args.ts
```diff
@@ -362,9 +362,20 @@ interface NestedMapping {
...
+const NUMBER_REGEXP = /-?\d+(\.\d*)?(e-?\d+)?$/;
+const HYPHEN_REGEXP = /^(-|--)[^-]/;
+const VALUE_REGEXP = /=(?<value>.+)/;
+const FLAG_NAME_REGEXP = /^--[^=]+$/;
+const IS_FLAG_REGEXP = /^-/;
+const SPECIAL_CHAR_REGEXP = /\W/;
+const IS_NUMBER_REGEXP =
+  /^(?:[-+]?(?:\d+(?:\.\d*)?|\.\d+)(?:e[-+]?\d+)?|0x[0-9a-f]+)$/i;
+
+function isNumber(string: string): boolean {
```
iuioiua, line 377: We should just us `isNan(Number(string))`. No need to deal with regexes.

cli/parse_args.ts
```diff
@@ -362,9 +362,20 @@ interface NestedMapping {
...
+  return IS_NUMBER_REGEXP.test(string);
```
kt3k, line 378: This change drops the hex number handling

cli/parse_args.ts
kt3k: Are you sure that check is equivalent of the regexp check?

iuioiua: The current `isNumber()` handles integers, decimals, negatives, decimals, hexadecimals, exponents, and whitespaces like `\t`. If we modify the function to be the following, the check aligns better with how values are actually used in the implementation (see [here](https://github.com/timreichen/deno_std/blob/c41bcfdfd659e3c7f752545dc7415081ff98ad73/cli/parse_args.ts#L610) and [here](https://github.com/timreichen/deno_std/blob/c41bcfdfd659e3c7f752545dc7415081ff98ad73/cli/parse_args.ts#L742)), which use `Number()`.

```ts
function isNumber(string: string): boolean {
  return string.trim() !== "" && !Number.isNaN(Number(string));
}
```

The checking of these values can be validated with the following tests. They pass on the current `isNumber()` and my above suggestion.

```ts

Deno.test("isNumber should return true for valid numbers", () => {
  assertEquals(isNumber("123"), true);
  assertEquals(isNumber("-456"), true);
  assertEquals(isNumber("0"), true);
  assertEquals(isNumber("3.14"), true);
  assertEquals(isNumber("-2.5"), true);
  assertEquals(isNumber("1e3"), true);
  assertEquals(isNumber("-5e-2"), true);
  assertEquals(isNumber("0x10"), true);
});

Deno.test("isNumber should return false for invalid numbers", () => {
  assertEquals(isNumber("abc"), false);
  assertEquals(isNumber("123abc"), false);
  assertEquals(isNumber("12.34.56"), false);
  assertEquals(isNumber("1e"), false);
  assertEquals(isNumber("e3"), false);
  assertEquals(isNumber("0xg"), false);
  assertEquals(isNumber(""), false);
  assertEquals(isNumber("\t"), false);
});
```

kt3k: It seems we also need to exclude `"Infinity"` and `"-Infinity"`. Maybe we can use `Number.isFinite` for it

kt3k (approved):
LGTM

PR #5609, opened 2024-08-01
fix(async): abortable should not change original outputs

raaymax:
Related to this PR: https://github.com/denoland/std/pull/5560
And this issue https://github.com/denoland/std/issues/3518


async/abortable_test.ts
```diff
@@ -166,3 +166,43 @@ Deno.test("abortable.AsyncIterable() calls return before throwing", async () =>
...
+  const abortController = new AbortController();
+  const abortableIterator = abortable(gen(), abortController.signal);
+
+  assertEquals(
+    await abortableIterator.next(123),
+    await normalIterator.next(123),
+  );
+  assertEquals(
+    await abortableIterator.return(321),
+    await normalIterator.return(321),
```
raaymax, line 205: There is still small difference in types of `return` function argument between `normalIterator` and `abortableIterator`. 
But this can't be fixed without changing `AsyncIterable<T>` type.
https://github.com/denoland/deno/issues/24837

iuioiua (changes requested):
Nice work. Can you please also move the `return()` from the `catch` block to a `finally` block? I believe that should be the correct way to do it.

raaymax:
As you wish but this way, the function will be called twice if no exception occurs.

iuioiua:
In that case, we'd likely need to at a `let returned = false` before the `try`/`catch`, set it do `true` when `done` is truthy, then check if that value is `true` in the `finally` block. We should also have a test case for that too.

kt3k:
> In that case, we'd likely need to at a let returned = false before the try/catch, set it do true when done is truthy, then check if that value is true in the finally block.

Sounds like the same thing can be achieved by calling `it?.return()` in catch block?

raaymax:
I thought so too and thats exactly what I did in the first place. 
In my opinion, using a boolean flag for this purpose is just ugly, but it's not my decision to make.
You need to decide on one solution here.

PR #5611, opened 2024-08-02
test(async): improve test speed of `async/pool_test.ts`

iuioiua:
It seems that these tests didn't need to be so slow.

On M2 laptop:
- Before: ~8s
- After: ~0.7s


kt3k (approved):
LGTM

PR #5618, opened 2024-08-02
refactor(cli): use non-null assertion in `parseArgs()` logic

iuioiua:
Addresses https://github.com/denoland/std/pull/5599#discussion_r1701269351


kt3k (approved):
LGTM

PR #5624, opened 2024-08-02
docs(semver): remove diagram line overlaps

timreichen:
**Changes**
This PR removes line overlaps in the semver diagram.

**Reasoning**
Event though not all names are aligned the same anymore, I think it makes it easier to read. It also visually distinguishes between semver "identifiers" (build, pre, patch, minor, major) and  combinations (full, release, primary).

On another note: Some of these names differ from the [spec names](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions) (e.g. `primary` instead of `core`, `pre` instead of `prerelease`). Is this intentional?


kt3k (approved):
Interesting. LGTM

PR #5625, opened 2024-08-02
refactor(yaml): move functions to class methods

timreichen:
**Changes**
- moves `readDocument()` inside `LoaderState` class
- moves `readDocuments()` inside `LoaderState` class

**Reasoning**
Both functions take the state as an argument, which is not necessary with methods.


iuioiua (approved):
Thanks

PR #5626, opened 2024-08-03
feat(random/unstable): basic randomization functions

lionel-rowe:
Closes https://github.com/denoland/std/issues/4848

Currently includes the following:

- `@std/random/between`: `randomBetween` function
- `@std/random/integer-between`: `randomIntegerBetween` function
- `@std/random/sample`: enhanced version of (to-be-deprecated?) `@std/collections/sample`'s `sample` function, with new `weights` and `random` options
- `@std/random/seeded`: `randomSeeded` function implementing the [PGC32](https://en.wikipedia.org/wiki/Permuted_congruential_generator) algorithm
- `@std/random/shuffle`: `shuffle` function implementing [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher–Yates_shuffle)


timreichen:
> * `@std/random/pick`: `pick` and `pickWeighted` functions

Note that we already have `sample()` in `@std/collections`. So if we introduced a weighted `sample()` function, it should be added to `@std/collections` or even as an option to the current sample, for example a weight option:
```ts
const numbers = [1, 2, 3, 4];
const random = sample(numbers, { weights: [1, 9999, 2, 8888] });
```

> * `@std/random/seeded`: `SeededPrng` class implementing the [Wichman-Hill algorithm](https://en.wikipedia.org/wiki/Wichmann%E2%80%93Hill)

Is there a particular reason why this algorithm is added?

> * `@std/random/shuffle`: `shuffle` function implementing [Fisher–Yates shuffle](https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle)

I think this would be a cool addition to `@std/collections`.

lionel-rowe:
@timreichen
> > * `@std/random/seeded`: `SeededPrng` class implementing the [Wichman-Hill algorithm](https://en.wikipedia.org/wiki/Wichmann%E2%80%93Hill)
> 
> Is there a particular reason why this algorithm is added?

Only that it was the algorithm suggested in the linked issue, it has a period high enough for any sensible use case, and it [seems to give a good distribution of results](https://github.com/denoland/std/blob/19b6ed7e17fc361f9e8825bff34e53da9018fad2/random/seeded_test.ts#L39-L90). One other plus is that arbitrarily chosen positive integer seeds still seem to give good entropy, even if they're apparently "low entropy" numbers like `1` or `12` (i.e. they don't need to be primes or anything fancy). I don't have the requisite expertise to verify its randomness properties beyond "seems to give good results" though.

lionel-rowe:
> > * `@std/random/pick`: `pick` and `pickWeighted` functions
> 
> Note that we already have `sample()` in `@std/collections`. So if we introduced a weighted `sample()` function, it should be added to `@std/collections` or even as an option to the current `sample()` function:
> 
> ```ts
> const numbers = [1, 2, 3, 4];
> const random = sample(numbers, { weights: [1, 9999, 2, 8888] });
> ```

I quite like the idea of making it an option. It does require "unzipping" the values from their weights, but it can still be pretty versatile in terms of usage:

```ts
const weighted = new Map([["a", 5], ["b", 3], ["c", 2]]);
const result = sample([...weighted.keys()], { weights: [...weighted.values()] });
```

I'll keep the `collections` stuff within the scope of this PR for now as some of it relies on the `random` stuff (e.g. I'll use `SeededPrng` for tests), but happy to split it out before merging if that's preferable.

collections/shuffle.ts
```diff
@@ -0,0 +1,48 @@
...
+  // https://en.wikipedia.org/wiki/Fisher–Yates_shuffle#The_modern_algorithm
+  // -- To shuffle an array a of n elements (indices 0..n-1):
+  // for i from n−1 down to 1 do
+  for (let i = result.length - 1; i >= 1; --i) {
+    // j ← random integer such that 0 ≤ j ≤ i
+    const j = randomIntegerBetween(0, i, opts);
+    // exchange a[j] and a[i]
+    const temp = result[i];
+    result[i] = result[j]!;
+    result[j] = temp!;
```
babiabeo, line 44: Perhaps we can use `[result[i], result[j]] = [result[j], result[i]]` for short?

kt3k:
@lionel-rowe Can you move `shuffle` and this version of `sample` to `std/random` (leaving the existing `collections/sample` untouched)?

Also I'd prefer to see `SeededPrng` exported from `@std/random/seeded-prng`, `randomBetween` from `@std/random/between`, `randomIntegerBetween` from `@std/random/integer-between` for consistency of the rest of std.

collections/sample.ts
```diff
@@ -25,12 +46,55 @@ function randomInteger(lower: number, upper: number): number {
...
+  options?: Partial<SampleOptions>,
+): T | undefined {
+  const { random, weights } = { ...defaultOptions, ...options };
+
+  if (weights) {
+    assertEquals(
+      weights.length,
+      array.length,
+      "The length of the weights array must match the length of the input array",
+    );
```
kt3k, line 74: We now have policy of not using `@std/assert` for checking user's input. https://github.com/denoland/std/blob/4432ba1c1bc02abc30e2e2238bc50e0127000fb9/.github/CONTRIBUTING.md#assertions

Can you replace this with if-statement check and throw RangeError in this case?

Could you also do the same for the other usages of `@std/assert`?

random/between_test.ts
```diff
@@ -0,0 +1,120 @@
...
+});
+
+Deno.test("randomBetween() throws if min or max are NaN", () => {
+  assertThrows(() => randomBetween(NaN, 1), AssertionError);
+  assertThrows(() => randomBetween(1, NaN), AssertionError);
+});
+
+Deno.test("randomBetween() throws if max is less than min", () => {
+  assertThrows(() => randomBetween(10, 1), AssertionError);
+});
```
kt3k, line 40: Nice checks 👍

timreichen:
> @lionel-rowe Can you move `shuffle` and this version of `sample` to `std/random` (leaving the existing `collections/sample` untouched)?

@kt3k Wouldn't that be confusing to have two `sample()` functions? Or do you mean to deprecate collections/sample in another PR?

> Also I'd prefer to see `SeededPrng` exported from `@std/random/seeded-prng`, `randomBetween` from `@std/random/between`, `randomIntegerBetween` from `@std/random/integer-between` for consistency of the rest of std.

`Prng` looks weird to me. Maybe it should it be written out as `SeededPseudoRandomNumberGenerator`?

kt3k:
> @kt3k Wouldn't that be confusing to have two sample() functions? Or do you mean to deprecate collections/sample in another PR?

I think we can deprecate `collections/sample` when we stabilize `random/sample`.

> > Also I'd prefer to see SeededPrng exported from @std/random/seeded-prng, randomBetween from @std/random/between, randomIntegerBetween from @std/random/integer-between for consistency of the rest of std.
>
> Prng looks weird to me. Maybe it should it be written out as SeededPseudoRandomNumberGenerator?

Or how about the name like `SeededRandom`? There is a similarly named npm package https://www.npmjs.com/package/seedrandom

random/_types.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Options for random number generation.
+ */
+export type RandomOptions = {
+  /**
+   * A function returning a random number between 0 and 1, used for
+   * randomization.
+   * @default {Math.random}
+   */
+  random: () => number;
```
timreichen, line 12: I think this option is not very web api-ish. I think it would be better to remove it and add some kind of `algorithm` option if needed,  that calls different functions internally. Doing `customRandomFunction() * (max - min) + min` seems very trivial.

random/seeded.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  /**
+   * Convert a user-provided scalar seed value to a state tuple suitable for
+   * usage by `SeededPrng`.
+   *
+   * Aims to provide good results for both well-randomized seeds (such as
+   * 64-bit outputs from `crypto.getRandomValues`) and low-entropy seeds (such
+   * as 1, 2, 3, keyboard mashing, etc.).
+   */
+  static #seedToState(seed: bigint): State {
```
timreichen, line 183: Couldn't that be just a separate function? I think a static private method is probably not very useful.

random/seeded.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.#inc = -1n;
+    this.#state = -1n;
+
+    this.state = typeof seed === "bigint"
+      ? SeededPrng.#seedToState(seed)
+      : seed;
+
+    // For convenience, allowing destructuring and direct usage in callbacks
+    // equivalently to Math.random()
+    this.random = this.random.bind(this);
```
timreichen, line 84: I think destructuring can be done without this.

random/seeded.ts
```diff
@@ -0,0 +1,220 @@
...
+   * @example Usage
+   * ```ts
+   * import { SeededPrng } from "@std/random";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const prng = new SeededPrng({ seed: 14614327452668470620n });
+   * assertEquals(prng.state, [13062938915293834817n, 10846994826184652623n]);
+   * ```
+   */
+  get state(): Readonly<State> {
```
timreichen, line 100: Does `state` need to be accessible outside the class? I only see it for testing purposes.

random/seeded.ts
lionel-rowe: Yeah probably. I'm still not settled on how best to handle seeding tbh — upon looking into it further it seems like the C/C++ reference implementation have a good way of initializing via seed data, but that seed consists of 2 parts (`initstate` and `initseq`), both of which are uint64. So just accepting 16 bytes of state as a seed could work, even though it's no longer a scalar value (unless you conceptualize it as a single 128-bit int, but JS doesn't support 128-bit ints via any web APIs etc)

random/_types.ts
lionel-rowe: > I think this option is not very web api-ish. I think it would be better to remove it and add some kind of `algorithm` option if needed, that calls different functions internally. Doing `customRandomFunction() * (max - min) + min` seems very trivial.

Callbacks seem pretty web API-ish to me, and they're more versatile than specifying a name of an algorithm — with a callback, it's easy to use with third-party randomization functions/sources. But perhaps more importantly, specifying an algorithm name would surely make any and all of the all implemented algorithms into non-tree-shakeable dependencies of any function taking such options.

random/seeded.ts
lionel-rowe: > I think destructuring can be done without this.

Destructuring loses `this` context of unbound methods (as does passing by reference in a callback).

random/seeded.ts
lionel-rowe: I think `state` should definitely be gettable, as it allows persisting a PRNg's current state for resumption later on. It's arguable whether or not it needs to be settable though, perhaps setting it should only be possible via the constructor or a `from` static method.

As with seeding, I'm still not too settled on how this should be exposed, maybe it should have explicit `state` and `inc` properties (and probably an `algorithm` property too, even though there's only one currently implemented).

collections/sample.ts
lionel-rowe: I'll remove those. What about `unreachable` though? The rationale from https://github.com/denoland/std/issues/4865 mentions three reasons not to use asserts in implementation code, of which two don't apply to `unreachable` (there should never be any need to handle `unreachable` errors, and an `unreachable` error fully captures its semantics as it's very literally an assertion). The third reason, introduction of a dependency, does still apply.

lionel-rowe:
How best to represent `SeededRandom`'s seed and state values is giving me a bit of a headache.

| Type | Pros | Cons |
|------|------|------|
| `BigInt` | - A single scalar value can hold a 64-bit uint<br>- Visually compact | - Can also hold negative values and values bigger than uint64, and validity can only be checked at runtime<br>- `JSON.stringify` throws |
| `BigUint64Array` | - Probably the most semantically "correct" | - The fact that the seed represents a 64-bit uint is an implementation detail that's not relevant for typical consumers of the library<br>- `JSON.stringify` throws<br>- Subject to endianness issues e.g. if created directly from underlying binary data. That seems like a deal-breaker to me. |
| `Uint8Array` | - Still semantically OK - bytes are bytes are bytes<br>- Easy to extract uint64 values without worrying about endiannness using `DataView#getBigUint64` | - Length can only be checked at runtime<br>- `JSON.stringify` gives an object with numeric keys but no `length` property, i.e. it can't be ergonomically converted back to a `Uint8Array`<br>- The bytes really represent 64-bit uints, not 8-bit uints |
| Hex string | - Ergonomic to write<br>- Visually compact | - Strings give very weak compile-time type checking, and TS's template literal types can't represent more than a couple of hex digits before running out of space |
| Standard array of `Byte`s (union of number literals 0..255) | - Can be easily converted to and from `Uint8Array`s<br>- Hard-coded values can be fully type-checked at compile time! You can check both length and the values of elements, e.g. allowing `[1, 1, 1, 1, 1, 1, 1, 1]` but disallowing `[1, 1, 1]` or `[1, 1, 1, 1, 1, 1, 1, 256]`. | - Not very "webby"<br>- Requires `as` type assertions for values that aren't hard-coded or directly emitted from the library |

One possible idea: accept both `Uint8Array` and standard arrays of numbers as inputs (maximum ergonomics for both hard-coded and non-hard-coded use), while emitting arrays of numbers in the output (maximum portability — easy JSON serialization etc). Proof of concept: https://github.com/lionel-rowe/deno_std/commit/11961938e305f2b32f6a22562bf9510ea94aacfa

kt3k:
> One possible idea: accept both Uint8Array and standard arrays of Bytes as inputs (maximum ergonomics for both hard-coded and non-hard-coded use), while emitting arrays of Bytes in the output (maximum portability — easy JSON serialization etc)

This sounds good to me. Also this is going to be unstable package for a while. We can iterate later.

random/_types.ts
timreichen: @kt3k do we have a guide on how to name option functions? I thought thy have a suffix of `Fn` -> `randomFn`. Not sure though.

kt3k (approved):
LGTM

random/sample.ts
iuioiua, line 1: I feel like a prng option makes the `weight` option redundant. Like, you could create a psuedo-random number generate that expresses weights, no? Removing an unneeded option would be a design win.

random/sample.ts
```diff
@@ -0,0 +1,102 @@
...
+ * assertArrayIncludes(numbers, [sampled]);
+ * ```
+ *
+ * @example Using `weights` option
+ * ```ts no-assert
+ * import { sample } from "@std/random/sample";
+ *
+ * const weighted = new Map([["a", 5], ["b", 3], ["c", 2]]);
+ * const result = sample([...weighted.keys()], { weights: [...weighted.values()] });
+ * // gives "a" 50% of the time, "b" 30% of the time, and "c" 20% of the time
```
iuioiua, line 58: Can we make this a little easier on the eyes by simply having two arrays; `array` and `weighted`? The comment should make this snippet understandable.

random/seeded_random.ts
```diff
@@ -0,0 +1,301 @@
...
+ * required, as it will always generate the same results.
+ *
+ * May be suitable for testing, but not recommended for production use.
+ */
+// deno-fmt-ignore
+export const PCG32_INITIALIZER: Readonly<SeededRandomState> = {
+  algorithm: "pcg32",
+  state: [0x85, 0x3c, 0x49, 0xe6, 0x74, 0x8f, 0xea, 0x9b],
+  inc: [0xda, 0x3e, 0x39, 0xcb, 0x94, 0xb9, 0x5b, 0xdb],
+}
```
iuioiua, line 16: I'm not so sure this should be part of the public API.

random/seeded_random.ts
```diff
@@ -0,0 +1,301 @@
...
+   * import { SeededRandom } from "@std/random";
+   * const prng = new SeededRandom({
+   *  seed: [
+   *    0x19, 0x57, 0xc5, 0x20, 0x6d, 0x56, 0xb3, 0xf4,
+   *    0x6e, 0xa1, 0x3c, 0x00, 0xb4, 0x6b, 0xcf, 0xa9,
+   *  ],
+   * });
+   * ```
+   */
+  constructor({ seed }: SeededRandomConstructorOptions) {
```
iuioiua, line 124: `seed` will always be required for this constructor.
```suggestion
  constructor(seed: <type>) {
```

random/seeded_random.ts
```diff
@@ -0,0 +1,301 @@
...
+const U32_CEIL = 2n ** 32n;
+
+/**
+ * Options for {@linkcode SeededRandom}.
+ */
+type SeededRandomConstructorOptions = {
+  /**
+   * The seed for the random number generator. Must be 16 bytes long.
+   */
+  seed: Uint8Array | (Byte[] & { length: 16 });
```
iuioiua, line 36: Why is this so complicated? Genuine question.

random/seeded_random.ts
iuioiua, line 1: Frankly, at first glance, this class seems to be far more complicated than it needs to be. Like, couldn't this just be a function that returns a function? Disclaimer: My opinion isn't super-strong, so please allow me to do a little more reading.

random/shuffle.ts
```diff
@@ -0,0 +1,46 @@
...
+import { defaultOptions, type RandomOptions } from "./_types.ts";
+import { randomIntegerBetween } from "./integer_between.ts";
+export type { RandomOptions };
+
+/**
+ * Shuffles the provided array, returning a copy and without modifying the original array.
+ *
+ * @typeParam T - The type of the items in the array
+ * @param items - The items to shuffle
+ * @param options - The options for the random number generator
```
iuioiua, line 11:
```suggestion
 * @typeParam T The type of the items in the array
 * @param items The items to shuffle
 * @param options The options for the random number generator
```

random/_benches/bench.ts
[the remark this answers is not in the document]
lionel-rowe: Will remove.

random/between.ts
```diff
@@ -0,0 +1,41 @@
...
+ * randomBetween(1, 10); // 3.6267118101712006
+ * randomBetween(1, 10); // 7.853320239013774
+ * ```
+ */
+export function randomBetween(
+  min: number,
+  max: number,
+  options?: RandomOptions,
+): number {
+  if (!isFinite(min)) {
```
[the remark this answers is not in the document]
lionel-rowe: I can't honestly remember why I used `isFinite` but it might have been due to how TS types them (`Number.isFinite(number: unknown): boolean` vs `isFinite(number: number): boolean`). But I'll swap for `Number.isFinite` as the runtime safety seems more useful

random/seeded_random.ts
lionel-rowe: From https://github.com/denoland/std/pull/5626#issuecomment-2275691117
> One possible idea: Accept both `Uint8Array` and standard arrays of `Byte`s as inputs (maximum ergonomics for both hard-coded and non-hard-coded use), while emitting arrays of `Byte`s in the output (maximum portability — easy JSON serialization etc). Proof of concept: [lionel-rowe@1196193](https://github.com/lionel-rowe/deno_std/commit/11961938e305f2b32f6a22562bf9510ea94aacfa)

I think it's probably true that it's overcomplicated in terms of surface area. I've since found that the Rust `rand` crate contains a nice function for converting a uint64 into a seed of any length with decent entropy, which would certainly be a nicer experience than needing to care about the number of bytes required by a given algorithm. It'd also remove the need for anything like `PCG32_INITIALIZER`, as seeding from `1n` would be perfectly acceptable for testing use.

random/integer_between.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import { defaultOptions, type RandomOptions } from "./_types.ts";
+import { randomBetween } from "./between.ts";
+export type { RandomOptions };
+
+/**
+ * Generates a random integer between the provided minimum and maximum values.
+ *
+ * The number is in the range `[min, max]`, i.e. both `min` and `max` are included.
```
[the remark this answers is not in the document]
lionel-rowe: Here I think `[min, max]` is a more intuitive default, as when you say "think of a number between 1 and 10" you typically mean 1 and 10 are both acceptable choices. Also, if you _round_ `Math.random()` to an integer, the results obviously include both 0 and 1.

random/integer_between.ts
```diff
@@ -0,0 +1,36 @@
...
+ * randomIntegerBetween(1, 10); // 9
+ * randomIntegerBetween(1, 10); // 2
+ * ```
+ */
+export function randomIntegerBetween(
+  min: number,
+  max: number,
+  options?: RandomOptions,
+): number {
+  if (!Number.isInteger(min) || !Number.isInteger(max)) {
```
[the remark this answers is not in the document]
lionel-rowe: Maybe it's unnecessary, I don't have a strong opinion either way

random/sample.ts
lionel-rowe: Something like this?

```ts
const weighted = new Map([["a", 5], ["b", 3], ["c", 2]]);
const result = sample([...weighted.keys()], { weights: [...weighted.values()] });

sample([...weighted.keys()], { random: weightedRandomFrom([...weighted.values()]) });
```

random/sample.ts
lionel-rowe: I wanted the example to use "zipped" values as a starting point, then only "unzip" them on calling the function, as I think starting with unzipped values isn't a best practice and leads to brittle code.

E.g.
```ts
const items = [
    'some long string that causes the code formatter to line wrap',
    'something else',
    'third thing',
];
const weights = [5, 3, 2];
```

It's a lot less easy to see which weight goes with which value, harder to refactor, and easier to make mistakes (e.g. pushing to `items` without pushing to `weights`, calling `sort` on one or both arrays causing them to get out-of-sync, etc).

random/seeded_random.ts
lionel-rowe: How about this for v0.1 of `SeededRandom`: remove all surface area except for ctor and `random()`; implement Rust `rand` crate's seeding logic; ctor's `seed` parameter then takes a bigint, which internally gets wrapped to fit in a uint64, before being passed to the seeding logic. Leave it as a class so that it's easy to expand in the future, plus I think that helps conceptually given that people tend to think "classes = oop = state" whereas "functions = fp = pure", which a seeded prng can't be unless the calling code manually passes the state each time it's called.

random/_types.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * A function implementing the same contract as `Math.random`, i.e. returning a
+ * random number in the range [0, 1).
+ */
+export type Prng = typeof Math.random;
```
iuioiua, line 7: Otherwise, `Prng`-typed functions won't allow for arguments, as `Math.random()` doesn't have any arguments.
```suggestion
export type Prng = (...args: number[]) => number;
```

random/_types.ts
```diff
@@ -0,0 +1,17 @@
...
+   * A function returning a random number between 0 and 1, used for
+   * randomization.
+   * @default {Math.random}
+   */
+  random: () => number;
+};
+
+export const defaultOptions: RandomOptions = {
+  random: Math.random,
+};
```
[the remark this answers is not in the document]
iuioiua: Please remove. We can just use `Math.random()` in implementations, as it's more explicit.

random/integer_between.ts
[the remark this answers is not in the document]
iuioiua: I'd be for keeping these constraints only if there was a good reason. Though, I can't think of a good reason, so I suggest we remove them.

random/_types.ts
```diff
@@ -0,0 +1,23 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * A function implementing the same contract as `Math.random`, i.e. returning a
```
iuioiua, line 4: We should have the term "pseudo-random number generator" somewhere in here. It's a common programming term that should make things more familiar.

random/_types.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * Options for random number generation.
+ */
+export type RandomOptions = {
+  /**
+   * A function returning a random number between 0 and 1, used for
+   * randomization.
+   * @default {Math.random}
+   */
+  random: Prng;
```
iuioiua, line 18: Same reason as above. It's a common programming term, and here, it seems more accurate.
```suggestion
  prng: Prng;
```

random/sample.ts
iuioiua: Yeah, maybe. Happy to hear other opinions.

random/sample.ts
iuioiua: I still think simple arrays incur less cognitive load. Not a strong opinion so happy to defer to your decision.

random/seeded_random.ts
iuioiua: We use 2nd-order functions that manage state in [`monotonicUlid()`](https://github.com/denoland/std/blob/d279c0a49a9e75a1641fbb514d06788fbfa89490/ulid/monotonic_ulid.ts#L52), which I think is fine. I think it should just return one thing: the prng. How about we implement `randomSeeded()` in `@std/random/seeded` and compare?

iuioiua (changes requested):

random/_types.ts
lionel-rowe: I'm not sure they _should_ allow for arguments, and I'm not even sure what those arguments would represent? The idea of passing it as an option is that it's a drop-in replacement for `Math.random`. If further customization is needed then it should be a different named type (which also makes me wonder whether `Prng` is the right name for this type after all).

random/integer_between.ts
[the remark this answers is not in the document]
lionel-rowe: Just to confirm: My intuition is that `randomIntegerBetween(-0.001, 1.999)` gives `0` 50% of the time and `1` 50% of the time. Does that chime with your intuition? In descending order of plausibility (to me), other possible interpretations could be that the result would be heavily biased toward `1`; that `2` is a possible result; that `-1` is a possible result; or that `-0.001` and `1.999` are possible results.

random/_types.ts
iuioiua: For psuedo-random number generators, arguments would represent parameters for distributions. E.g. we could one day have the following added to the package.
```ts
export function randomExponential(lambda: number): number {
  if (lambda <= 0) {
    throw new RangeError(
      `Cannot compute exponential: lambda must be greater than 0, got ${lambda}`,
    );
  }

  return -Math.log(1 - Math.random()) / lambda;
}
```
Imagine similar functions like `randomBinomial()`, etc. I guess we can wait until more of these are implemented to see if `...args: number[]` makes most sense and just use `...args: any[]` in the meantime.

import_map.json
```diff
@@ -36,6 +36,14 @@
...
     "@std/path": "jsr:@std/path@^1.0.3",
     "@std/regexp": "jsr:@std/regexp@^1.0.0",
+
+    "@std/random": "./random/mod.ts",
+    "@std/random/between": "./random/between.ts",
+    "@std/random/integer-between": "./random/integer_between.ts",
+    "@std/random/sample": "./random/sample.ts",
+    "@std/random/seeded-random": "./random/seeded_random.ts",
+    "@std/random/shuffle": "./random/shuffle.ts",
+
```
iuioiua, line 46:
```suggestion
    "@std/random": "jsr:@std/random@^0.1.0",
```

iuioiua (changes requested):

random/_types.ts
lionel-rowe: Those can just be function-returning-functions, same as proposed seeded random (or class methods, same as current implementation). Also `n > 0` doesn't fulfil the contract as it doesn't guarantee results in `[0, 1)`, so it wouldn't be suitable as a drop-in replacement for `Math.random` (`n >= 1` would work though).

```ts
type Prng = typeof Math.random;

function randomExponential(lambda: number): Prng {
  if (lambda < 1) {
    throw new RangeError(`lambda must be at least 1, got ${lambda}`);
  }

  return () => {
    return -Math.log(1 - Math.random()) / lambda;
  };
}
```

iuioiua: 2nd order functions make sense when the returned function relies on state. I.e. `randomSeeded()`. However, something like `randomExponential()` wouldn't rely on state. So there's no real need for it to be a 2nd order function.

lionel-rowe: In that case I think `{ prng: () => randomExponential(lambda) }` works fine. I'm strongly against the idea of `Prng` being `(...args) => number` because I don't understand the point of a type that doesn't represent a clear contract. Even if it was `(arg?: number) => number` it's not clear, as that single `number` argument could mean different things to different prng functions. And how does the calling function decide what that argument should be? Adding an additional `argsToPassToPrng` parameter just complicates things more.

random/seeded_random.ts
lionel-rowe: In that case I'll just convert to the the HOF version and remove the class for now. Given the reduced surface area, refactoring back to a class should be easy enough if needed.

random/_types.ts
0f-0b: In C++,

* A [uniform random bit generator](https://en.cppreference.com/w/cpp/named_req/UniformRandomBitGenerator) generates uniformly distributed numbers.
* A [distribution](https://en.cppreference.com/w/cpp/named_req/RandomNumberDistribution) samples a value according to an associated probability function using a given uniform random bit generator.

The two concepts are conflated into one `Prng` type here, which seems to be the source of confusion in this discussion?

lionel-rowe: Hmm... of the two, I wonder if there's really a use case for supplying a distribution as an option? It might make sense for some functions but it's less generally applicable than supplying a uniformly-distributed prng (which is generally useful for allowing seeding, and I guess cryptographic security too but with the usual caveats around that).

random/sample.ts
lionel-rowe: On second thoughts I don't think this makes sense (taking into account @0f-0b 's comment about distribution vs generation). My original implementation was `pickWeighted` (== `sampleWeighted`) as a separate function, which is another option.

random/_constants.ts
```diff
@@ -0,0 +1,13 @@
...
+ * The maximum value of a 64-bit unsigned integer plus one.
+ * `n % U64_CEIL` simulates uint64 overflow.
+ */
+export const U64_CEIL = 2n ** 64n;
+
+/**
+ * The maximum value of a 32-bit unsigned integer plus one.
+ * `n % U32_CEIL` simulates uint32 overflow.
+ */
+export const U32_CEIL = 2n ** 32n;
```
0f-0b, line 13: For simulating overflow there's a builtin function [`BigInt.asUintN`](https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/BigInt/asUintN).

random/integer_between.ts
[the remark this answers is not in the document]
lionel-rowe: FWIW, in Python `random.randomint` is `[min, max]` and throws on non-integer min/max, while `np.random.randomint` is `[min, max)` and silently truncates non-integer min/max. The latter leads to extremely counter-intuitive results, e.g. `np.random.randint(0.99,1.99)` always returns `0`.

random/seeded.ts
iuioiua, line 1: This seems a lot nicer. Right now, we hardcode the PCG32 algorithm. Should this instead be configurable? In other words, what if we allowed for BYO PRNG algorithm? As to not delay this PR, we can explore this in a follow-up issue/PR.

random/between.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { RandomOptions } from "./_types.ts";
+
+/**
+ * Generates a random number between the provided minimum and maximum values.
+ *
```
iuioiua, line 6: Please ensure that all public symbols follow [this guideline](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#notices-for-unstable-apis).

random/between.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
```
iuioiua, line 1: Please add where applicable.
```suggestion
// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
// This module is browser compatible.
```

iuioiua (changes requested):
Let's try get this landed soon.

random/sample.ts
lionel-rowe: > Sorry, I don't follow.

From the following three options:
1. Keep the weights option as-is
2. Separately expose the weighted sampling functionality with a `sampleWeighted` function, with a signature something like `function sampleWeighted<T>(items: { item: T, weight: number }[]): T`
3. Keep a single `sample` function, remove the `weights` option, and instead allow weighted functionality to be implemented supplying a biased `prng` (formerly known as `random`) option

3 no longer makes sense to me, as it conflates two different concepts and ultimately complicates things conceptually rather than simplifying. Of the other two, 1 would be my preference, but it's not a strong preference.

random/_types.ts
lionel-rowe: > PRNGs are purely mathematical. Hence, them only accepting 0 to N optional arguments of type number. Allowing this option provides lots of flexibility while not complicating implementations.

I think this will be clearer with an example.

`randomBetween` takes a `Prng` as an option, which it calls internally with zero arguments, expecting the result to be in `[0, 1)`. That's a simple contract and still provides plenty of flexibility. For example,  our current `seeded` function requires a seed (a numeric argument, albeit a `bigint`), but that's provided to the HOF, not the returned function, so the returned function can easily be provided as a `Prng` option, as seen in the tests and doc examples.

If you have another, user-supplied PRNG function that takes 3 numeric arguments, you just need to supply those arguments prior to passing as an option. For example, with `function customRandom(a: number, b: number, c: number): number`, you provide that as `{ prng: () => customRandom(1, -65, 99) }`. If those numbers vary each time, perhaps it'll be more like `{ prng: () => customRandom(nextA(), nextB('x', 'y'), nextC({ prop: true })) }`. The interface is just as versatile as it needs to be for any given function. `randomBetween` doesn't need to care about or understand any of those additional arguments.

If, on the other hand, the `Prng` type accepts one or more arguments, _even if those are only numerical arguments_, where will those arguments be supplied from when `randomBetween` calls it? And what does `randomBetween` assume them to mean? A seed value? A number of steps to advance the PRNG each time? And how does it decide how many arguments to pass? Should it read `prng.length` to find out? What if `prng` is variable-length? What if it uses `arguments` internally so its `length` can't be relied on? And does _every_ function that takes a `Prng` option need to re-implement those same details? What if different calling functions make different assumptions? It's not just that it's nonsensical as a contract, it's not even clear what it _might_ mean or how that could be implemented in code.

Restricting the argument types to `number` does nothing to solve the above problems, yet it's still too restrictive for many possible implementations (e.g. you may want to supply a Uint8Array of seed/state bytes, a string to generate a seed from, a file path like `/dev/urandom` to read synchronously from, etc)... But as noted above, these can easily be supplied as config prior to passing as an option, without changing the `Prng` type at all.

random/seeded.ts
lionel-rowe: Yeah there should be plenty of room to add other options in future. Possibly moving back to a class-based model for the algorithm itself will be desirable (but keep `seeded` as a HOF). Then roughly, algorithm classes compatible with `seeded` must implement an interface providing static methods like `fromSeed`, instance methods like `nextUint`, and static properties like `numSeedBytes` and `nextUintBitSize`.

random/sample.ts
lionel-rowe: On second thoughts I think you're right. Changed the example 👍

random/_types.ts
iuioiua: Ah, ok. I understand now. Yes, we only want functions that use `options.prng` to call `options.prng()` and nothing more. Let's keep as is then. Thanks for clearing up.

iuioiua (changes requested):
One last thing. Can you please go through all exceptions and ensure they following the [contributing guidelines](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md)? See https://github.com/denoland/std/blob/main/archive/tar_stream.ts for an example.

random/between.ts
```diff
@@ -0,0 +1,43 @@
...
+    throw new RangeError(`min cannot be ${min}`);
+  }
+  if (!Number.isFinite(max)) {
+    throw new RangeError(`max cannot be ${max}`);
+  }
+  if (max < min) {
+    throw new RangeError("max must be greater than or equal to min");
+  }
+
+  return (options?.prng ?? Math.random)() * (max - min) + min;
```
0f-0b, line 42: This algorithm has a few problems.

- `max - min` can overflow (such as when `min` is `-Number.MAX_VALUE` and `max` is `Number.MAX_VALUE`), making the result either `Infinity` or `NaN` depending on whether the RNG returns `0`.
- Rounding errors can cause the result to sometimes equal `max` (such as when `min` is `1`, `max` is `2` and the RNG returns `0.9999999999999999`) and the distribution to be non-uniform.

A comparably simple alternative (`min * (1 - x) + max * x`) never overflows, but is still prone to rounding errors. [The algorithm used in PHP](https://github.com/php/php-src/blob/407bc092db1729e30eec73147dd0c2a37579ab09/ext/random/gammasection.c#L30-L100) solves both problems but is much more complicated.

Should we change the algorithm or keep the current one?

random/sample.ts
0f-0b: I like 1 as well.

For one of my use cases though I would like a function that takes a list of weights (as an array-like object, specifically a typed array) and returns the position of the chosen item (to index into another typed array). Something like this:

```ts
function randomWeighted(weights: ArrayLike<number>, options?: RandomOptions): number;
```

Do you think this should be added?

random/between.ts
lionel-rowe: I'm inclined to say this should be fixed even for v0.1 if possible (except for the non-uniform thing, that seems less of a show-stopper as long as the non-uniformity won't cause noticeable bias). But I wonder if there isn't some other way to cheat and get the same effect, e.g. subtracting some multiple of `Number.EPSILON`?

```ts
const x = (options?.prng ?? Math.random)();
return (min * (1 - x) + max * x) - ((max - min) * Number.EPSILON);
```

0f-0b: Maybe just map any out-of-bound values to `min`?

```ts
const x = (options?.prng ?? Math.random)();
const y = min * (1 - x) + max * x;
return y >= min && y < max ? y : min;
```

lionel-rowe:
> One last thing. Can you please go through all exceptions and ensure they following the [contributing guidelines](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md)? See https://github.com/denoland/std/blob/main/archive/tar_stream.ts for an example.

@iuioiua I think they already do, with the exception that error messages beginning with a variable name don't capitalize the variable name:

```ts
throw new RangeError("max must be greater than or equal to min");
```

How should those be handled? `` `max` ``/`` `min` `` with backticks would be another option.

random/between.ts
```diff
@@ -0,0 +1,45 @@
...
+ * randomBetween(1, 10); // 7.853320239013774
+ * ```
+ */
+export function randomBetween(
+  min: number,
+  max: number,
+  options?: RandomOptions,
+): number {
+  if (!Number.isFinite(min)) {
+    throw new RangeError(`min cannot be ${min}`);
```
kt3k, line 33:
```suggestion
    throw new RangeError(`Cannot generate a random number: min cannot be ${min}`);
```

random/between.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+  if (!Number.isFinite(max)) {
+    throw new RangeError(`max cannot be ${max}`);
```
kt3k, line 36:
```suggestion
    throw new RangeError(`Cannot generate a random number: max cannot be ${max}`);
```

random/between.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+  if (max < min) {
+    throw new RangeError("max must be greater than or equal to min");
```
kt3k, line 39:
```suggestion
    throw new RangeError(`Cannot generate a random number as max must be greater than or equal to min: max=${max}, min=${min}`);
```

random/integer_between.ts
```diff
@@ -0,0 +1,40 @@
...
+ * randomIntegerBetween(1, 10); // 2
+ * ```
+ */
+export function randomIntegerBetween(
+  min: number,
+  max: number,
+  options?: RandomOptions,
+): number {
+  if (!Number.isFinite(min) || !Number.isFinite(max)) {
+    throw new RangeError("min and max must be finite");
```
kt3k, line 34:
```suggestion
    throw new RangeError("Cannot generate a random number: min and max must be finite");
```

random/sample.ts
```diff
@@ -0,0 +1,97 @@
...
+export function sample<T>(
+  array: readonly T[],
+  options?: SampleOptions,
+): T | undefined {
+  const { weights } = { ...options };
+
+  if (weights) {
+    if (weights.length !== array.length) {
+      throw new RangeError(
+        "The length of the weights array must match the length of the input array",
```
kt3k, line 65:
```suggestion
        "Cannot sample an item: The length of the weights array must match the length of the input array",
```

random/sample.ts
```diff
@@ -0,0 +1,97 @@
...
+      );
+    }
+
+    if (!array.length) return undefined;
+
+    const total = Object.values(weights).reduce((sum, n) => sum + n, 0);
+
+    if (total <= 0) {
+      throw new RangeError("Total weight must be greater than 0");
```
kt3k, line 74:
```suggestion
      throw new RangeError("Cannot sample an item: Total weight must be greater than 0");
```

kt3k (commented):
Suggested some updates in error messages based on the style guide at https://github.com/denoland/std/blob/1432018f0080422453103994513c782aaf180959/.github/CONTRIBUTING.md#error-messages

random/sample.ts
0f-0b: I guess that means I can also do this?

```ts
const values = Uint16Array.of(5687, 5720, 5730);
const weights = Uint32Array.of(4, 1, 1);
const value = sample(values, { weights });
```

If so that would work.

iuioiua (approved):
LGTM! Excellent work. Thank you, and thank you to the contributors that chimmed in. Let's track any possible points of improvements in a tracking issue for discussions, etc.

kt3k (approved):
LGTM!

PR #5530, opened 2024-07-24
deprecation(url): deprecate `@std/url`

iuioiua:
### What's changed

[`@std/url`](https://jsr.io/@std/url) has been deprecated in favor of the following [`@std/path/posix`](https://jsr.io/@std/path/doc/posix/~/) corresponding functions:
1. [`basename()`](https://jsr.io/@std/path/doc/posix/~/basename)
2. [`dirname()`](https://jsr.io/@std/path/doc/posix/~/dirname)
3. [`extname()`](https://jsr.io/@std/path/doc/posix/~/extname)
4. [`join()`](https://jsr.io/@std/path/doc/posix/~/join)
5. [`normalize()`](https://jsr.io/@std/path/doc/posix/~/normalize)

### Motivation

This change was made as `@std/path/posix` functions essentially do the same jobs as those in `@std/url`. `@std/url` will be soon be removed, but as JSR is immutable, the last version of the package to live on with all functionality intact, but noted as deprecated.

### Migration guide

To migrate, please see the latest documentation for the following functions for examples on how to use [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL)s:
1. [`basename()`](https://jsr.io/@std/path/doc/posix/~/basename)
2. [`dirname()`](https://jsr.io/@std/path/doc/posix/~/dirname)
3. [`extname()`](https://jsr.io/@std/path/doc/posix/~/extname)
4. [`join()`](https://jsr.io/@std/path/doc/posix/~/join)
5. [`normalize()`](https://jsr.io/@std/path/doc/posix/~/normalize)

### Related

Addresses https://github.com/denoland/std/pull/4992#pullrequestreview-2166480173

Issue #5537, opened 2024-07-24
path: make some `@std/path` functions work with file URL object


kt3k:
As FS APIs in Deno works with file URL object, it would make sense to support file URL input with `path.dirname`.

```ts
dirname(new URL("file:///path/to/dir/file")) // returns `/path/to/dir` on unix
dirname(new URL("file:///C:/path/to/dir/file")) // returns `C:\path\to\dir` on windows
```

This probably would throw with non-file URLs.

I think this mostly addresses the motivation we added `std/url`.

- [x] `dirname()`
- [x] `extname()`
- [x] `join()`
- [x] `basename`
- [x] `normalize`

iuioiua:
Just `dirname()` or the other functions too?

kt3k:
I think this also makes sense for other functions such as `basename`, `extname`, `common`, `join`, `relative`, and `resolve`

kt3k:
I don't think we need to limit this to `std/path/posix` as file URL object is platform independent

kt3k:
I'd suggest we would target `basename` and `normalize` instead of `resolve` and `relative` for the remaining items as they are something exposed as `std/url` APIs.

We can revisit `resolve` and `relative` as they are requested

iuioiua:
SGTM

kt3k:
done

PR #5630, opened 2024-08-05
BREAKING(url): remove `@std/url`

iuioiua:
See #5530 for background.


kt3k:
I'd like to land this after we addressed https://github.com/denoland/std/issues/5537

If we address #5537, then the deprecation message doesn't need to mention `path/poxix`

iuioiua:
Closing until we make the appropriate changes in `@std/path`.

PR #5637, opened 2024-08-06
fix(archive): make `block` and `reader` properties in `Untar` private

iuioiua:
These properties are not documented or tested anywhere. It seems they were included in the public API by mistake.


kt3k (approved):
LGTM

PR #5643, opened 2024-08-06
fix(http): handle wrong request method correctly

iuioiua:
Previously, the `file-server` module handled all requests with non-`GET` and non-`HEAD` methods the same as `GET` and `HEAD` methods, when they shouldn't have. Now, such methods cause a HTTP 405 status to be served.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -173,7 +173,7 @@ export async function serveFile(
   filePath: string,
   options?: ServeFileOptions,
 ): Promise<Response> {
-  if (req.method !== "GET" && req.method !== "HEAD") {
+  if (req.method !== "GET") {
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -173,6 +173,10 @@ export async function serveFile(
   filePath: string,
   options?: ServeFileOptions,
 ): Promise<Response> {
+  if (req.method !== "GET" && req.method !== "HEAD") {
```
kt3k, line 176: I'm unsure if `HEAD` is supported

iuioiua: Removed

kt3k (approved):
LGTM

PR #5645, opened 2024-08-06
fix(archive): make `data` property private in `Tar`

iuioiua:
This property is not documented or tested anywhere. It seems it was included in the public API by mistake.


kt3k (approved):
LGTM

PR #5646, opened 2024-08-06
feat(text/unstable): add `slugify()` function

timreichen:
Ref: https://github.com/denoland/std/issues/5627


timreichen:
@marvinhagemeister @kt3k The current draft implements the char replacement map from `npm:slugify`. Some of these replacements seem obsolete, because they are removed by `.replaceAll(/[^a-zA-Z0-9\s-]/g, "")` later. Should I remove all entries that have non-char values?

iuioiua:
PTAL @marvinhagemeister @kt3k

text/slugify.ts
```diff
@@ -0,0 +1,441 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+const CHAR_MAP = new Map([
```
kt3k, line 3: This list looks random and arbitrarily chosen. How about starting with something very minimal? For example:

```
const CHAR_MAP = new Map([
  ["$", "dollar"],
  ["%", "percent"],
  ["&", "and"],
]);
```

timreichen: This is the list taken from `npm:slugify`. I even wonder if we should implement char replacement at all. There exists no standard and any list will always be arbitrary. Maybe replacements should be done by the user before calling `slugify()` and all non-valid chars then are removed?

iuioiua: I agree with your suggestion, Tim.

kt3k: No char replacement makes more sense to me

kt3k (approved):
LGTM

text/slugify.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+/**
+ * **UNSTABLE**: New API, yet to be vetted.
```
iuioiua, line 4: For future reference, this being the first paragraph means it's used as the summary for this symbol. Please ensure the description is always first for symbols.

kt3k: This is the same as what CLI does. https://docs.deno.com/api/deno/all_symbols

If you suggest something different, I think you should also suggest the same to CLI docs.

iuioiua: https://github.com/denoland/deno/issues/25103

PR #5647, opened 2024-08-07
BREAKING(datetime): replace `utc` with `timeZone` option

timreichen:
**Changes**
This PR changes the `utc` boolean option to `timeZone` with explicit `"UTC"` value option.

**Reasoning**
This makes the option extendable if needed and aligns the api to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleTimeString#using_options

### DIFF
--- a/datetime/format.ts
+++ b/datetime/format.ts
@@ -6,11 +6,11 @@ import { DateTimeFormatter } from "./_date_time_formatter.ts";
 /** Options for {@linkcode format}. */
 export interface FormatOptions {
   /**
-   * Whether returns the formatted date in UTC instead of local time.
+   * The timezone the formatted date should be in.
    *
-   * @default {false}
+   * @default {undefined}
    */
-  utc?: boolean;
+  timeZone?: "UTC";
 }
 
 /**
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Seems fine to me.

kt3k:
Our formatter seems only supporting utc and local timezones? (because we use `Date` under the hood). ref. https://github.com/denoland/std/blob/96d70203cd85c9598477755e196f5431fba380c8/datetime/_date_time_formatter.ts#L154

Doesn't this give the false sense of other timezones being supported?

timreichen:
> Our formatter seems only supporting utc and local timezones? (because we use `Date` under the hood). ref.
> 
> https://github.com/denoland/std/blob/96d70203cd85c9598477755e196f5431fba380c8/datetime/_date_time_formatter.ts#L154
> 
> Doesn't this give the false sense of other timezones being supported?

I think it is clear that only `UTC` is supported, because it is the only valid value. If you want we can add a check and throw an error for invalid timezone values (other than `UTC` here, or maybe even better case insensitive) like in [Intl.DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat).

iuioiua:
My take: it doesn't matter too much, either way.

kt3k (approved):
LGTM

PR #5652, opened 2024-08-07
fix(http/unstable): make `info` parameter optional

iuioiua:
This makes the parameter compatible with `deno serve`, which was the original intention of `route()`.


kt3k (approved):
LGTM

PR #5656, opened 2024-08-08
docs(io): document `std/io`

kt3k:
part of #3764


kt3k:
@iuioiua PTAL

iuioiua (approved):
Nice! Once we trim the package of its deprecated APIs, we should revisit some symbols. Namely, options which don't have their own interfaces don't display their properties' JSDocs correctly.

PR #5657, opened 2024-08-08
refactor(io): cleanup `BufferFullError` and `PartialReadError` logic

iuioiua:


io/buf_reader.ts
```diff
@@ -14,10 +14,10 @@ const LF = "\n".charCodeAt(0);
  * @deprecated This will be removed in 1.0.0. Use the {@link https://developer.mozilla.org/en-US/docs/Web/API/Streams_API | Web Streams API} instead.
  */
 export class BufferFullError extends Error {
-  override name = "BufferFullError";
```
iuioiua, line 17: Removes the need to document this property.

io/buf_reader.ts
```diff
@@ -26,10 +26,12 @@ export class BufferFullError extends Error {
  * @deprecated This will be removed in 1.0.0. Use the {@link https://developer.mozilla.org/en-US/docs/Web/API/Streams_API | Web Streams API} instead.
  */
 export class PartialReadError extends Error {
-  override name = "PartialReadError";
```
iuioiua, line 29: Again, removes the need to document this property.

io/buf_reader.ts
```diff
@@ -26,10 +26,12 @@ export class BufferFullError extends Error {
...
-  partial?: Uint8Array;
-  constructor() {
+  partial: Uint8Array;
+
+  constructor(partial: Uint8Array) {
```
iuioiua, line 31: Previously, this property would be set within `try`/`catch` blocks. It should instead just be set during construction (see below).

kt3k (approved):
LGTM

PR #5659, opened 2024-08-08
refactor(front-matter): remove `Extractor` type

timreichen:
**Changes**
This PR removes the `Extractor` type and moves `Format` type to `_formats.ts`.

**Reasoning**
`Extractor` is not used anywhere.


iuioiua (approved):
Good catch!

PR #5661, opened 2024-08-08
feat(tools): add export check

timreichen:
This PR adds a check a file name against its exports. This will ensure that exports are named equal to the file name (different cases).

There are some exceptions, some files are excluded: `**/mod.ts`, `_*.ts` `test.ts`, `*_test.ts`, `**/types.ts`, `**/constants.ts`. It would probably be good to declare in the docs that `types.ts` and `constants.ts` are special cases (or remove the special treatment).

There might be some mod specific files that must not excluded after discussion, e.g. files in `fmt`, `encoding`, `testing`, `http`, `uuid`.


iuioiua:
Closing this for now. Let's discuss and investigate more before pulling the trigger.

PR #5662, opened 2024-08-09
feat(cache/unstable): `TtlCache`

lionel-rowe:
Closes https://github.com/denoland/std/issues/5669


iuioiua:
Please see https://github.com/denoland/std/issues/4608#issuecomment-2277606460

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import type { MemoizationCache } from "./memoize.ts";
+import { delay } from "@std/async/delay";
+
+/**
+ * Time-to-live cache.
+ *
+ * Automatically removes entries once the configured amount of time elapses. If
+ * the values themselves are promises, the countdown starts from when they
+ * resolve.
```
iuioiua, line 10: Hm... I'm not sure that promises should get any special treatment... @oles, do you have any preferences?

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * @typeParam K The type of the cache keys.
+ * @typeParam V The type of the cache values.
+ *
+ * @example Usage
+ * ```ts
+ * import { TtlCache } from "@std/cache";
```
iuioiua, line 17:
```suggestion
 * import { TtlCache } from "@std/cache/ttl-cache";
```

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
...
+ * import { assertEquals } from "@std/assert";
```
iuioiua, line 18:
```suggestion
 * import { assertEquals } from "@std/assert/equals";
```

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
...
+   *
+   * @example Usage
+   * ```ts
+   * import { TtlCache } from "@std/cache";
+   * import { assertEquals } from "@std/assert";
+   * const cache = new TtlCache(1000);
+   * assertEquals(cache.defaultTtl, 1000);
+   * ```
+   */
+  defaultTtl: number;
```
iuioiua, line 42: I think it's best that we don't expose any properties unless we have a real good reason to. I think it's fine if this is private.
```suggestion
  #defaultTtl: number;
```

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
...
+   * import { TtlCache } from "@std/cache";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const cache = new TtlCache<string, number>(1000);
+   *
+   * cache.set("a", 1);
+   * assertEquals(cache.get("a"), 1);
+   * ```
+   */
+  override set(key: K, value: V, customTtl?: number): this {
```
iuioiua, line 74:
```suggestion
  override set(key: K, value: V, customTtl = this.#defaultTimeout): this {
```

cache/ttl_cache.ts
```diff
@@ -0,0 +1,97 @@
...
+    this.#deleters.get(key)?.ac.abort();
+    const ac = new AbortController();
+
+    this.#deleters.set(key, {
+      promise: (async () => {
+        if (value instanceof Promise) {
+          await value;
+        }
+
+        await delay(customTtl ?? this.defaultTtl);
```
iuioiua, line 86: I wonder whether we really need this `delay()` dependency. Might we be just fine using [`AbortSignal.timeout()`](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static)?

cache/ttl_cache_test.ts
iuioiua, line 1: Can you please sprinkle in some use of `.get()`, so we know that the returned values are correct?

iuioiua (changes requested):

cache/ttl_cache.ts
kt3k: That sounds a good point. I don't see the rationale for the special handling of `Promise` very well.

Also this seems causing uncaught error when the promise is rejected. That is probably inconvenient for most users.

cache/ttl_cache.ts
```diff
@@ -0,0 +1,72 @@
...
+ * assertEquals(cache.size, 0);
+ * ```
+ */
+export class TtlCache<K, V> extends Map<K, V>
+  implements MemoizationCache<K, V> {
+  #defaultTtl: number;
+  #timeouts = new Map<K, number>();
+
+  /**
+   * Construct a new `TtlCache`.
```
iuioiua, line 32:
```suggestion
   * Constructs a new instance.
```

iuioiua (changes requested):
PTAL, @kt3k.

cache/ttl_cache_test.ts
```diff
@@ -0,0 +1,69 @@
...
+  await t.step("default TTL (passed in constructor)", () => {
+    using time = new FakeTime(0);
+
+    const cache = new TtlCache<number, string>(10);
+
+    cache.set(1, "one");
+    cache.set(2, "two");
+
+    time.now = 1;
+    assertEntries(cache, [[1, "one"], [2, "two"]]);
```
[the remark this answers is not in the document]
lionel-rowe: I added `assertEntries` and `UNSET` to make sure `has`, `get`, etc. are hit as well as just the iterable entries themselves (`UNSET` checks `has` is false)

iuioiua (approved):
LGTM! Thank you.

cache/ttl_cache.ts
```diff
@@ -0,0 +1,149 @@
...
+   *
+   * @example Usage
+   * ```ts no-assert
+   * import { TtlCache } from "@std/cache/ttl-cache";
+   * console.log(new TtlCache<string, number>(1000)); // TtlCache
+   * ```
+   */
+  override get [Symbol.toStringTag](): string {
+    return this.constructor.name;
+  }
```
iuioiua, line 43: Please add a test

kt3k (approved):
LGTM when Asher's point addressed

iuioiua (approved):
On second thought, I removed `[Symbol.toStringTag]()`. We don't implement this anywhere else in `std`. Either way, thank you for this PR!

jsejcksn (commented):
It might be useful to document the valid range for TTL values in this implementation. Because it defers to `setTimeout`, this will be implementation-dependent (see link to spec in previous comment), but testing can be performed to inform a suggestion for Deno at least. Here is an example demonstrating that the current build of V8 in Deno only properly handles unsigned 32-bit integers:

`settimeout_range.test.ts`:
```lang-ts
import { AssertionError } from "jsr:@std/assert@1.0.3/assertion-error";

console.log("Deno.version", Deno.version);

async function assertIntervalIsScheduledAfter1Millisecond(
  interval: number,
): Promise<void> {
  let timestamp: number | undefined;
  const expectedTimeStamp = Date.now() + interval;

  const timerId = setTimeout(() => {
    timestamp = Date.now();
  }, interval);

  await new Promise((resolve) => setTimeout(resolve, 1));

  if (timestamp) {
    throw new AssertionError(
      `The callback completed ${
        expectedTimeStamp - Date.now()
      } ms earlier than expected.`,
    );
  }

  clearTimeout(timerId);
}

Deno.test("setTimeout(): Schedules intervals…", async (t) => {
  await t.step(
    "less than 2 ** 31",
    () => assertIntervalIsScheduledAfter1Millisecond(2 ** 31 - 1),
  );

  await t.step(
    "greater than or equal to 2 ** 31",
    () => assertIntervalIsScheduledAfter1Millisecond(2 ** 31),
  );
});
```

Test output:

```txt
% deno test settimeout_range.test.ts
------- pre-test output -------
Deno.version { deno: "1.46.1", v8: "12.9.202.2-rusty", typescript: "5.5.2" }
----- pre-test output end -----
running 1 test from ./settimeout_range.test.ts
setTimeout(): Schedules intervals… ...
  less than 2 ** 31 ... ok (3ms)
  greater than or equal to 2 ** 31 ... FAILED (1ms)
setTimeout(): Schedules intervals… ... FAILED (due to 1 failed step) (4ms)

 ERRORS 

setTimeout(): Schedules intervals… ... greater than or equal to 2 ** 31 => ./settimeout_range.test.ts:34:10
error: AssertionError: The callback completed 2147483647 ms earlier than expected.
		throw new AssertionError(
		      ^
    at assertIntervalIsScheduledAfter1Millisecond (file:///Users/deno/settimeout_range.test.ts:18:9)
    at eventLoopTick (ext:core/01_core.js:213:9)
    at async innerWrapped (ext:cli/40_test.js:191:5)
    at async exitSanitizer (ext:cli/40_test.js:107:27)
    at async Object.outerWrapped [as fn] (ext:cli/40_test.js:134:14)
    at async TestContext.step (ext:cli/40_test.js:492:22)
    at async file:///Users/deno/settimeout_range.test.ts:34:2

 FAILURES 

setTimeout(): Schedules intervals… ... greater than or equal to 2 ** 31 => ./settimeout_range.test.ts:34:10

FAILED | 0 passed (1 step) | 1 failed (1 step) (5ms)

error: Test failed
```

iuioiua:
@jsejcksn, would you be able to submit a PR containing your suggestions?

jsejcksn:
> @jsejcksn, would you be able to submit a PR containing your suggestions?

@iuioiua Created: https://github.com/denoland/std/pull/5834

PR #5664, opened 2024-08-09
fix(cache/unstable): fix flaky `memoize()` test with `FakeTime`

lionel-rowe:
Fixes https://github.com/denoland/std/issues/5663


iuioiua (approved):
Great! Thank you.

Issue #5574, opened 2024-07-30
Normalize error message grammar


jsejcksn:
> Is the error grammar in all std mods the same? I think to remember different ones (`missing xyz: something`, `xyz is missing in something`, `you didn't put xyz in something`).

(Originally posted by @timreichen in https://github.com/denoland/std/issues/5476#issuecomment-2255599242)

jsejcksn:
There's been past discussion about messaging grammar in various comments in past issues/PRs, but most were somewhat tangent to the associated contexts. I created this issue as a dedicated place to potentially find alignment.

timreichen:
Some thoughts to keep this going:

- What grammar structure should be used for errors? Can we establish some rules for guidance?
  - noun-first: `input is invalid`
  - noun-first, but avoid *negative* adjectives: `input is not valid`
  - adjective-first: `invalid input`

- Should there be a dot at the end of an error message?
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/front_matter/_shared.ts#L43
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/encoding/base32.ts#L38

- Should error messages start uppercase or lowercase?*
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/front_matter/_shared.ts#L43
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/crypto/_fnv/mod.ts#L22

- Should error messages have an explanation, on how to avoid the error?
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/encoding/base32.ts#L38
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/encoding/base32.ts#L47

- Should an error include information, where the error occurred?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/JSON_bad_parse
https://github.com/denoland/std/blob/fab93bb71decf3f103bb26de88c37a9b15c26d11/datetime/_date_time_formatter.ts#L70-L72

- When to use `colons`, `commas`, `dots`*:
  - `TypeError: Cannot define property x, object is not extensible (V8-based)`
  - `TypeError: can't define property "x": Object is not extensible (Firefox)`
  

[...]

PR #5671, opened 2024-08-10
refactor(io): create a style guide for error messages, align error messages to the style guide

irbull:
Creates an initial style guide for user-facing error messages. The guide is based on the style guide from V8.


.github/CONTRIBUTING.md
```diff
@@ -167,3 +167,71 @@ after the starting description.
...
+```
+
+2. Messages should not end with a period:
+
+```
+Bad: Cannot parse input.
+Good: Cannot parse input
+```
+
+3. Message should not use quotes for values:
```
kt3k, line 196: I wonder if this rule is reasonable. If the value is a string, using quotes around it doesn't seem like a problem to me.

How about not having this rule?

.github/CONTRIBUTING.md
```diff
@@ -167,3 +167,71 @@ after the starting description.
...
+Bad: Can't parse input x
+Good: Cannot parse input x
+```
+
+7. Messages should use a colon when providing additional information:
+
+```
+Bad: Cannot parse input x. value is empty
+Bad: Cannot parse input x: value is empty
+Good: Cannot parse input x, value is empty
```
kt3k, line 229: The example part seems having an error?
```suggestion
Good: Cannot parse input x: value is empty
```

.github/CONTRIBUTING.md
irbull: IMHO I like the quotes too, so +1 from me. I just used what was suggested on the other PR and moved it here. I can update the PR if you'd like, however I'm at the lake with extremely bad internet this week, so I don't know if I'll be able to push the change. I'm back to civilization on Sunday, so I can do it then.

.github/CONTRIBUTING.md
irbull: Opps. From the original suggestion it was "Should the message use a colon or comma when providing additional information". Do you prefer colon, comma, or either?

irbull:
I'd be happy to take a first pass at the std to see if it conforms to these rules. I'd start by looking at places an `error` is raised. Let me know if you think this would help?

.github/CONTRIBUTING.md
iuioiua: We prefer colon.

irbull: I've updated the guide to mention that colon is preferred.

.github/CONTRIBUTING.md
irbull: I've updated the style guide to indicate that quotes should be used for values.

irbull:
I've updated the `CONTRIBUTING.md` with the suggestions that @kt3k had. I also took a pass at the `io/` folder in a second commit on this PR. If you like this approach, I can do some of the other folders too. Or if you would prefer I can file them in different PRs.

io/buffer.ts
```diff
@@ -352,7 +352,9 @@ export class Buffer implements Writer, WriterSync, Reader, ReaderSync {
       // don't spend all our time copying.
       copy(this.#buf.subarray(this.#off), this.#buf);
     } else if (c + n > MAX_SIZE) {
-      throw new Error("The buffer cannot be grown beyond the maximum size.");
+      throw new Error(
+        'The buffer cannot be grown beyond the maximum size of "${MAX_SIZE}"',
```
kt3k, line 356:
```suggestion
        `The buffer cannot be grown beyond the maximum size of "${MAX_SIZE}"`,
```

.github/CONTRIBUTING.md
```diff
@@ -167,3 +167,71 @@ after the starting description.
...
+Good: Cannot parse input
+```
+
+3. Message should use quotes for values:
+
+```
+Bad: Cannot parse input x
+Good: Cannot parse input "x"
+```
+
```
kt3k, line 202: Thanks for updating. But I think this rule needs more consideration. For example, using quotes makes sense when the value is a string, but it probably makes more sense to not use quotes when the value is a number. I guess there are edge cases where it's difficult to decide which pattern to use (nested object, etc).

For now, I'd suggest removing this rule (until we come up with a good way to formulate this)

```suggestion
```

kt3k:
Most other part looks good to me

io/read_range.ts
```diff
@@ -46,24 +46,24 @@ export async function readRange(
 ): Promise<Uint8Array> {
   // byte ranges are inclusive, so we have to add one to the end
   let length = range.end - range.start + 1;
-  if (length <= 0) throw new RangeError("Invalid byte range was passed.");
+  if (length <= 0) throw new RangeError("Invalid byte range was passed");
```
timreichen, line 49: According to the guide line shouldn't this be something like `Byte range start cannot be larger than end`?

io/read_range.ts
```diff
@@ -97,24 +97,24 @@ export function readRangeSync(
 ): Uint8Array {
   // byte ranges are inclusive, so we have to add one to the end
   let length = range.end - range.start + 1;
-  if (length <= 0) throw new RangeError("Invalid byte range was passed.");
+  if (length <= 0) throw new RangeError("Invalid byte range was passed");
```
timreichen, line 100: According to the guide line shouldn't this be something like `Byte range start cannot be larger than end`?

.github/CONTRIBUTING.md
jsejcksn: I think quoting strings is sensible… and I've often used direct JSON serializations (e.g. ``const message = "…found `${JSON.stringify(str)}`…";``) in my own error messages to get quotes for free (and representations that can be copy+pasted back into some JS source). The downside to that is when it's desirable to print literals which are replaced by escaped sequences in the output (e.g. line breaks).

jsejcksn:
Linking https://github.com/denoland/std/issues/5574

.github/CONTRIBUTING.md
irbull: This is why I attempted to updated the error messages in a folder, I think this gives us some good concrete examples to talk about and see how they _feel_.

I'm happy to remove this rule, or we could also use this:

3. Message should use quotes for values for strings:

```
Bad: Cannot parse input hello, world
Good: Cannot parse input "hello, world"
Good: Attempting to grow buffer by 10 bytes, which would exceed the maximum size of 100 bytes
```

What do you think, do you want to keep this rule for Strings?

io/buffer.ts
irbull: Yikes, sorry about that. I'll update the PR.

irbull:
> Linking:
> 
> * [Normalize error message grammar #5574](https://github.com/denoland/std/issues/5574)

Thanks for linking! I've added a table to the issue that lists all the packages. I'll take a stab at the other ones once we're happy with the style and I return to some more stable internet.

.github/CONTRIBUTING.md
kt3k: > 3. Message should use quotes for values for strings:
> 
> ```
> Bad: Cannot parse input hello, world
> Good: Cannot parse input "hello, world"
> Good: Attempting to grow buffer by 10 bytes, which would exceed the maximum size of 100 bytes
> ```

This looks good to me

irbull:
While updating a few more error messages, I think having either comma `,` or colon `:` for additional information makes sense. In some cases, it flows much better with commas. For example:

```
Messagepack decode came across an invalid type for a key of a map, keys must be a number or a string
```
vs
```
Messagepack decode came across an invalid type for a key of a map: keys must be a number or a string
```

irbull:
I've filed another PR with a few more folders (toml, msgpack & crypto). I'm not sure if you want all the commits here, or if it's easier to review / merge if we split them up.

kt3k:
To maintainers (@iuioiua, @bartlomieju): What commit tag should we be using for these style changes of error messages? `fix`, `docs`, or `refactor`?

iuioiua:
The docs changes in `.github/CONTRIBUTING.md` should be it's own PR with the enforcing of these rules in other PRs.

iuioiua:
Historically, the changing of these error messages have been refactors.

irbull:
> The docs changes in `.github/CONTRIBUTING.md` should be it's own PR with the enforcing of these rules in other PRs.
That's fair. I created separate commits so that should be easy. I wanted to include a few changes so we could see how these error messages look. I'll separate the initial commit into its own PR.

> Historically, the changing of these error messages have been refactors.
I can update the commit message to use `refactor` instead of `fix`. The first commit should be `doc` I guess, as it's the documentation change.

bartlomieju:
> To maintainers (@iuioiua, @bartlomieju): What commit tag should we be using for these style changes of error messages? `fix`, `docs`, or `refactor`?

I'm fine with `refactor`

bartlomieju:
Ooops, sorry, closed by accident 😓

irbull:
Would you like me to pull out the changes to the error messages in the `io` folder and leave this PR simply for creating the style guide? I typically `--amend` commits and make each commit (and message) the way I like it and then FF merge instead of squashing all the commits in a PR when merging. This way I can see the way each commit will look before merging, it's a bit more work, but it ensures that the commit & commit message are reviewed together.

But if you would prefer, I can pull out the `io` commit from this PR, create a separate PR for that and then we can just focus on the style guide here? Let me know.

kt3k (approved):
Let's just land this. Thanks for your contribution!

PR #5672, opened 2024-08-11
docs(archive,assert,cache,cli,encoding,html,http,net,streams,text): remove unstable Markdown alert

iuioiua:


kt3k:
> This text isn't needed. @experimental suffices.

This doesn't seem aligned with CLI's policy.

kt3k:
I agree with removing the `[!WARNING]` format for readability in IDE

iuioiua:
It's true that that the CLI uses the unstable note. But I still think it's not needed. Hints in the IDE show the `@experimental` tag, and JSR shows an unstable button ([example](https://jsr.io/@std/archive/doc/~/Tar)). This unstable notice doesn't seem to provide any added value.

iuioiua:
Furthermore, while the CLI employs this convention, it's not in the [Deno Style Guide](https://docs.deno.com/runtime/manual/references/contributing/style_guide/). So, I think these reasons provide sufficient justification to deviate.

kt3k:
> But I still think it's not needed. Hints in the IDE show the @experimental tag, and JSR shows an unstable button ([example](https://jsr.io/@std/archive/doc/~/Tar)).

I disagree with this. JSR rendering of `@experimental` tag is fine, but `@experimental` tag in IDE is not visible enough.

The Spinner docs in IDE after this change.

Screenshot 2024-08-19 at 20 31 24

It's at the last line, and it doesn't seem enough alerting. It's likely the users miss this.

Screenshot 2024-08-19 at 20 32 49

iuioiua:
That's because `@experimental` is the last tag. We should make `@experimental` always come after the initial description.

We can do:

```
/**
 * <description>
 *
 */
```

WDYT?

kt3k:
That looks better to me, but I think it's still important for CLI and STD to have the same policy (formatting) for this to make things less confusing.

@bartlomieju What do you think about the above suggested formatting?

kt3k:
> ```
> /**
>  * <description>
>  *
>  * @experimental **UNSTABLE**: New API, yet to be vetted.
>  */
> ```

This now looks a good compromise between consistency with CLI and better rendering in IDE and JSR. Let's do this

kt3k (approved):
LGTM

PR #5676, opened 2024-08-13
fix(testing): fix false positive in `IsExact`

Milly:
## Problem

`IsExact` has many false positives.

## My fix

- Add test cases for array or tuple types. These work correctly in the original `IsExact` implementation before this PR, but give false positives in [the alternative implementation](https://github.com/type-challenges/type-challenges/blob/e77262dba62e9254451f661cb4fe5517ffd1d933/utils/index.d.ts#L7-L9), so I am added it.
- Add test cases for function or class, and fix false positive.
  - This fix resulted in the following false negative.
- Add test cases for intersection, and fix false negative. In other words, I fixed the problem that "correct equality may still be stricter than some expect."

## Refs

https://github.com/denoland/std/issues/3196#issuecomment-1430099273
It was closed by arguments against including utility types.
However, `IsExact` already exists in `@std` so I fixed the bug.

### DIFF
--- a/testing/types.ts
+++ b/testing/types.ts
@@ -177,12 +177,14 @@ export type IsNullable<T> = Extract<T, null | undefined> extends never ? false
  * @typeParam T The type to check if it exactly matches type `U`.
  * @typeParam U The type to check if it exactly matches type `T`.
  */
-export type IsExact<T, U> = TupleMatches<AnyToBrand<T>, AnyToBrand<U>> extends
-  true
-  ? TupleMatches<DeepPrepareIsExact<T>, DeepPrepareIsExact<U>> extends true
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Thanks for looking into this.

This new version looks capturing `IsExact` concept more strictly without breaking any existing test case.

Now the cases like:
- `IsExact<{ prop: any }, { readonly prop: any }>`
- `IsExact<() => any, () => unknown>`
- `IsExact<(arg: any) => void, (arg: unknown) => void>`
- etc
are evaluated as `false`. These changes seem reasonable to me.

LGTM

iuioiua (commented):
Great work!

PR #5683, opened 2024-08-15
refactor(yaml): remove `dump` property

timreichen:
**Changes**
This PR removes the `dump` property of the `DumperState`. This is achieved by passing data directly into the former `write*()` functions. 
Because they do not have to set the `dump` property anymore, they can return a string instead and have been renamed `stringify*()` to reflect the return type.
There have been minor adjustments with early returns of said functions and a reduction of nested if/else statements.

**Reasoning**
The `dump` property was a global property used as a placeholder for various data passed between methods. Therefore it had the type `any` and was not predictable. It was also hard to follow what the dump was, because of the recursiveness of the functions.
This makes it possible to simplify the code in future PRs.


iuioiua (changes requested):
I just want to emphasise that a PR like this should only do one thing - remove the `dump` property. There seems to be other refactors sprinkled in this PR and it makes it more difficult to be certain that behavior isn't being broken.

timreichen:
> I just want to emphasise that a PR like this should only do one thing - remove the `dump` property. There seems to be other refactors sprinkled in this PR and it makes it more difficult to be certain that behavior isn't being broken.

I reduced the changes. I will follow up this PR with the others, because some method and variable names etc. don't make sense anymore and some code is now unnecessarily convoluted.

iuioiua (approved):

PR #5684, opened 2024-08-15
refactor(yaml): remove `result` property

timreichen:
**Changes**
This PR removes the `result` property of the `DumperState`.

**Reasoning**
It is not used anywhere.


kt3k (approved):
LGTM

PR #5686, opened 2024-08-16
refactor(yaml): remove `tag` property

timreichen:
⚠️This is a subsequent PR of https://github.com/denoland/std/pull/5683

**Changes**
This PR removes the `tag` property in `DumperState`.

**Reasoning**
`tag` is only used in `stringifyBlockMapping()`, so it can be passed as an argument instead.


timreichen:
closed in favour of https://github.com/denoland/std/pull/5689

PR #5688, opened 2024-08-16
chore: support `expect-error` directive in code snippets evaluator

magurotuna:
This commit adds support for `expect-error` directive in code snippets evaluator, which allows example snippets to have ones that are expected to fail either at compile time or runtime.

This is inspired by Rust where we can add doc tests that are supposed to panic, or to fail to compile, e.g.

```rust
/// ```should_panic
/// assert!(false);
/// ```
///
/// ```compile_fail
/// let x = 5;
/// x += 2;
/// ```
```

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -148,6 +148,17 @@ delimiter. E.g.
  */
 ````
 
+You can also write code snippets that are expected to fail either at compile
+time or at runtime by adding `expect-error`. E.g.
+
+````ts
+/**
+ * ```ts expect-error
+ * (code snippet to fail)
+ * ```
+ */
+````
+
 ### Notices for unstable APIs
 
 Each unstable API must have the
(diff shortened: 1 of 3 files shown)


kt3k (approved):
Looks a nice enhancement of doc linter. LGTM

iuioiua:
This is kind of overlaps with the pre-existing `no-eval` directive... I'm unsure this was worth adding.

magurotuna:
@iuioiua When adding `no-eval` we don't test that snippet at all, whereas `expect-error` allows that snippet to be run and to ensure that that snippet fails in an expected way.

My inclination is that we should use either no attribute (i.e. normal evaluation) or `expect-error` attribute, and avoid using `no-eval` as much as possible. In other words, we only want to use `no-eval` in very limited scenarios, such as where a code example needs to interact with external resources so we can't run within test, or where a code example triggers side effects that are not preferable in tests.

For instance, the following snippet, which is currently annotated `no-eval`, can be better described by separating it into two snippets, where one contains "doesn't throw" cases, and the other one contains "throws" cases with `expect-error` attribute.

https://github.com/denoland/std/blob/f3a22c05535f549d68282496e3241ba8f6ffd97d/assert/almost_equals.ts#L15-L24

iuioiua:
I see. Ok. Good explanation. Thank you. I've created #5728 to track to move towards using `expect-error`.

PR #5689, opened 2024-08-16
refactor(yaml): remove `tag` property

timreichen:
**Changes**
This PR removes the `tag` property in `DumperState`.

**Reasoning**
`tag` is only used in `writeBlockMapping()`, so it can be passed as an argument instead.


iuioiua (changes requested):
Can you please only include changes that directly pertain to removing the `tag` property? We can handle any other refactors in a separate PR.

timreichen:
> Can you please only include changes that directly pertain to removing the `tag` property? We can handle any other refactors in a separate PR.

Alright, I undo the `detectType` changes

yaml/_dumper_state.ts
```diff
@@ -766,39 +758,38 @@ export class DumperState {
       _result += pairBuffer;
     }
 
-    this.tag = _tag;
     this.dump = _result || "{}"; // Empty mapping if no valid pairs.
   }
 
-  detectType(explicit: boolean): boolean {
+  detectType(explicit: boolean) {
```
iuioiua, line 764: The return type for this method has changed from explicitly `boolean` to now being implicitly `string | null`. Why? Return types should be explicit and not change.

iuioiua (changes requested):

yaml/_dumper_state.ts
timreichen: Since the `tag` property is removed and was set inside this function, it has to be returned instead. It returns `string`(former `true`) and `null` (former `false`).

yaml/_dumper_state.ts
```diff
@@ -812,12 +803,9 @@ export class DumperState {
       isKey: boolean;
     },
   ): boolean {
-    this.tag = null;
     this.dump = object;
 
-    if (!this.detectType(false)) {
-      this.detectType(true);
-    }
+    const tag = this.detectType(false) ?? this.detectType(true) ?? null;
```
iuioiua, line 808:
```suggestion
    const tag = this.detectType(false) ?? this.detectType(true);
```

iuioiua (changes requested):

iuioiua (approved):
LGTM. Thank you.

PR #5690, opened 2024-08-16
feat(assert/unstable): add `assertNever`

magurotuna:
This commit adds a new function `assertNever` to the `assert` submodule.

This function is particularly useful when we want to check whether we properly handle all the variants of a discriminated union.
Ref: https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#union-exhaustiveness-checking

### DIFF
--- a/assert/never.ts
+++ b/assert/never.ts
@@ -27,11 +27,10 @@ import { AssertionError } from "./assertion_error.ts";
  */
 
 /**
+ * **UNSTABLE**: New API, yet to be vetted.
+ *
  * Make an assertion that `x` is of type `never`.
  * If not then throw.
- *
- * > [!WARNING]
- * > **UNSTABLE**: New API, yet to be vetted.
  *
  * @experimental
  *
--- a/assert/never.ts
+++ b/assert/never.ts
@@ -30,6 +30,11 @@ import { AssertionError } from "./assertion_error.ts";
  * Make an assertion that `x` is of type `never`.
  * If not then throw.
  *
+ * > [!WARNING]
+ * > **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @experimental
+ *
  * @example Exhaustivenss check
  * ```ts
  * import { assertNever } from "@std/assert/never";
--- a/assert/never.ts
+++ b/assert/never.ts
@@ -1,0 +1,99 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+import { format } from "@std/internal/format";
+import { AssertionError } from "./assertion_error.ts";
+
+/*!
+ * Ported and modified from: https://github.com/microsoft/TypeScript-Website/blob/v2/packages/documentation/copy/en/handbook-v1/Unions%20and%20Intersections.md#union-exhaustiveness-checking
+ * licensed as:
+ *
+ * The MIT License (MIT)
+ * Copyright (c) Microsoft Corporation
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of this software and
+ * associated documentation files (the "Software"), to deal in the Software without restriction,
+ * including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ * and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ *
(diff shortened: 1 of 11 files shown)


assert/never.ts
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT
+ * NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ * IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+ * WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+ * SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+ */
+
+/**
+ * Make an assertion that `x` is of type `never`.
```
kt3k, line 30: We now have a policy of accepting all new features as unstable API. Could you add `**UNSTABLE**: New API, yet to be vetted.` note at the beginning of jsdoc and also add `@experimental` jsdoc tag? https://github.com/denoland/std/blob/010784cd03d0b3ac7e913d82d3468d293adc13c8/.github/CONTRIBUTING.md#notices-for-unstable-apis

assert/never.ts
```diff
@@ -0,0 +1,104 @@
...
+ * If not then throw.
+ *
+ * > [!WARNING]
+ * > **UNSTABLE**: New API, yet to be vetted.
```
kt3k, line 34: I'll update the format a bit. (We're removing markdown `!WARNING` directive as it's not rendered well in VSCode https://github.com/denoland/std/pull/5672)

```suggestion
 * **UNSTABLE**: New API, yet to be vetted.
 *
 * Make an assertion that `x` is of type `never`.
 * If not then throw.
```

kt3k (approved):
LGTM

magurotuna:
@kt3k 
Minor question, should we fix [the instruction](https://github.com/denoland/std/blob/8d1d2d773990b875a7d7c4f7a4516d3e3051b47a/.github/CONTRIBUTING.md#notices-for-unstable-apis) I referred to so we'll place `<description>` after the UNSTABLE warning and remove `[!WARNING]`?

kt3k:
I think so, but @iuioiua might have a different idea. ref https://github.com/denoland/std/pull/5672

magurotuna:
Got it. I'll watch that issue 👍

iuioiua:
This is the same as [`unreachable()`](https://jsr.io/@std/assert/doc/unreachable/~/unreachable).

magurotuna:
@iuioiua I thought that way, but actually `unreachable` does not do any compile-time checking while `assertNever` does. 

This playground illustrates this:
https://www.typescriptlang.org/play/?#code/GYVwdgxgLglg9mABAQwM6oKYCcoDkMBu2AFAB4BciYh2AlJdUVogN4BQiiUAFlnAO5UMggKJY+WYgCJG2KbQDcbAL5s2oSLASJwWDMgjdkAIwA2GYgFtUAc0qooWGGBv0hTVhy68BQ0eLhJAANdfUMTc0QAEhZrG2UgxRU1KABPAAcMRAAVZBtEAF5EKQBBKUQAH2KAISkldXBoeCQQTBL0bDwaSSg8ylzXT05UfhgoQ0RiXsH2Tk4INCzSqXIvOcQAeg3EABM4RFQ4SwweZxs1ueM9ZABrJXWdjGBkEFMoVfXOLcRgQJs4KBcfZGMA7SJSWoAOhy3BgqEQelQryg8OcKC4GSy2AkkIunDQmBw+CYUzySU4qlUDU0zR0mAAqmBruEzBZpv0yUMDqNxtxJtNaFz5otimUPp9vnsDkcTrCXBdHs9keL1t9flh-oCoMDkKDwVCYXDdnAMPCIEd0jBzAAaFCg7x8fjw5CArDgWDHXGfHRMsJGVnSOBwdKoeT3RCUoA

PR #5691, opened 2024-08-16
refactor(yaml): Make `version` property required

timreichen:
**Changes**
This PR makes the `version` property in `DumperState` required.

**Reasoning**
The version is checked against `null`, so `undefined` should not be possible.


kt3k (approved):
LGTM

PR #5692, opened 2024-08-16
refactor(yaml): remove `schema` property

timreichen:
**Changes**
This PR removes the `schema` property in `DumperState`.

**Reasoning**
`schema` is only used in the constructor to extract values which are stored in properties themselves.


kt3k (approved):
LGTM

PR #5693, opened 2024-08-16
refactor(yaml): redefine `include` as optional `Schema`

timreichen:
**Changes**
This PR redefines the `include` property in `Schema` as a single optional `Schema`.

**Reasoning**
Every used Schema only has one included `Schema`.


kt3k (approved):
LGTM

PR #5694, opened 2024-08-16
refactor(yaml): use `Map`s for `TypeMap`

timreichen:
**Changes**
This PR changes the `TypeMap` type so it uses `Map`.


kt3k (approved):
LGTM

PR #5696, opened 2024-08-16
refactor(yaml): use `Map` for `tagMap` and `anchorMap`

timreichen:
**Changes**
This PR changes the type of `tagMap` and `anchorMap` to `Map`

**Reasoning**
The current implementation uses must use `Object.create(null)` to set the value.
This seems like legacy code that can be achieved with `Map`.


kt3k (approved):
LGTM

PR #5697, opened 2024-08-16
refactor(yaml): use `Set` for `overridableKeys`

timreichen:
**Changes**
This PR sets the type of `overridableKeys` to `Set<string>`

**Reasoning**
The current implementation seems to be legacy code.


kt3k (approved):
LGTM

iuioiua (commented):
Nice! I wonder if other similar changes can be made in the aim of eventually removing `ArrayObject` from this package.

PR #5698, opened 2024-08-16
refactor(json): remove branks function

timreichen:
**Changes**
This PR removes the regex check for non-whitespace data with `trim().length`.


kt3k:
I don't think this is a desirable change. `chunk.trim()` allocates another string, which is unnecessary and something we should avoid

timreichen:
> I don't think this is a desirable change. `chunk.trim()` allocates another string, which is unnecessary and something we should avoid

I see. @std has a few places where `trim()` is used for a comparison.

https://github.com/denoland/std/blob/f3a22c05535f549d68282496e3241ba8f6ffd97d/http/etag.ts#L205

https://github.com/denoland/std/blob/f3a22c05535f549d68282496e3241ba8f6ffd97d/internal/diff_str.ts#L188

https://github.com/denoland/std/blob/f3a22c05535f549d68282496e3241ba8f6ffd97d/cli/parse_args.ts#L374

Should we change the check there so @std is consistent?

kt3k:
I think `trim`ing only for checking should be generally avoided.

PR #5700, opened 2024-08-16
refactor(semver): unify param names

timreichen:
**Changes**
This PR unifies names of public functions.
- `value` if the `argument` is a string to be parsed
- `version` if the argument is a `SemVer`
- `versionN...` if multiple arguments are `SemVer`s
- `Range` if the argument is a `Range`

**Reasoning**
The docs are automatically created and use the jsdoc param names.
It makes the docs more esthetically pleasing and easier to understand.


kt3k (approved):
Nice improvement in docs consistency! Thanks!

PR #5701, opened 2024-08-16
BREAKING(log): change `setup` and `destroy`

timreichen:
Ref: https://github.com/denoland/std/issues/4391

**Changes**
This PR remove the `setup()` and `destroy()` methods from `BaseHandler`.
It removes the `setup()` and `destroy()` method calls in the `setup()` function.
~It makes `BaseHandler` and `log()` abstract.~
It adds a check whether the file is open when `log()` is called. If not, it opens it.


kt3k:
I'm not in favor of this change. This doesn't solve anything, but changes things in breaking way, based on aesthetic reason/naming preference.

timreichen:
> I'm not in favor of this change. This doesn't solve anything, but changes things in breaking way, based on aesthetic reason/naming preference.

Well, it removes the necessity of having the `setup` and `destroy` functions at all and thus simplifies the api.
Renaming them in the `FileHandler` to `open()` and `close()` describes what they do better imo.

Are you not in favour of both or just one of these changes?

kt3k:
I'm not in favor of either.

> it removes the necessity of having the setup and destroy functions at all and thus simplifies the api.

I don't understand this. It gives the default (empty) implementation. The inherited classes don't need to implement them.

timreichen:
> I'm not in favor of either.
> 
> 
> 
> > it removes the necessity of having the setup and destroy functions at all and thus simplifies the api.
> 
> 
> 
> I don't understand this. It gives the default (empty) implementation. The inherited classes don't need to implement them.

What I mean is that users can use the `constructor()` as a `setup()` function and `dispose()` as a `destroy()` function. There is no need for api specific functions.

iuioiua:
I understand both the reasons for and against this change. However, I'll close this without merge as I think the method names are fine as-is.

PR #5702, opened 2024-08-16
BREAKING(log): rename `BaseHandler` to `Handler` and make it abstract

timreichen:
Ref: https://github.com/denoland/std/issues/5620

**Changes**
This PR renames `BaseHandler` to `Handler` and `BaseHandlerOptions` to `HandlerOptions` and makes it abstract.


kt3k:
I'm not in favor of this change. This looks like the renaming for the sake of renaming.

timreichen:
> I'm not in favor of this change. This looks like the renaming for the sake of renaming.

Would you agree to that change if the class was made abstract?
I just feel that `Base` just means "do not create an instance of me, extend me".

kt3k:
Do you still suggest the rename? I'm in favor of making it abstract, but not in favor of the remame.

timreichen:
> Do you still suggest the rename? I'm in favor of making it abstract, but not in favor of the remame.

I am, but I can open another PR for adding the abstract.
`Base` just doesn't make sense to me, all handlers have a prefix that describe what they handle (`ConsoleHandler`, `FileHandler`. `BaseHandler` is just odd.

kt3k:
> Base just doesn't make sense to me, all handlers have a prefix that describe what they handle (ConsoleHandler, FileHandler. BaseHandler is just odd.

Sounds like a personal preference, which doesn't give enough reason for a breaking change.

iuioiua:
I think the current name is fine too.

iuioiua:
Closing as we'd like to keep the `BaseHandler` name for now.

Issue #5682, opened 2024-08-15
Suggestion: extend `assertObjectMatch` to support regexp


kevin-zou:
**Is your feature request related to a problem? Please describe.**

`assertObjectMatch` will only pass if the actual object is a subset of the expected object. If you have outputs that are non-deterministic/unknown at runtime then it's difficult to determine (eg. randomly generated values).

For example:
```ts
const expected = {
  vegetable: 'carrot',
  fruit: 'apple',
}

const outputOne = {
  vegetable: 'carrot',
  fruit: <some value we cannot determine>,
}

assertObjectMatch(actual, expected) // fails
```

**Describe the solution you'd like**

For `assertObjectMatch` support an expected object such as:
```ts
const expected = {
  vegetable: 'carrot',
  fruit: /[a-zA-Z]+/,
}

const outputOne = {
  vegetable: 'carrot',
  fruit: 'mango',
}

const outputTwo = {
  vegetable: 'carrot',
  fruit: 'peach',
}

assertObjectMatch(actual, outputOne) // passes
assertObjectMatch(actual, outputTwo) // passes
```

Open to discussion/feedback!

**Describe alternatives you've considered**

You could write a custom function that:
1. Iterates through the key/values of the expected object
2. If a regexp value is found, use `.test()` against the actual value. Otherwise the behaviour is the same

Alternatively you could make an assertion on the specific property you're looking at, then delete it from the object.

kt3k:
`expected` in `assertObjectMatch` doesn't need to have all properties of actual object. The above case can be expressed like the below using `assertObjectMatch` and `assertMatch`:

```ts
import { assertMatch, assertObjectMatch } from "@std/assert";

const expected = {
  vegetable: "carrot",
};

const actual = {
  vegetable: "carrot",
  fruit: Math.random() > 0.5 ? "mango" : "peach",
};

assertObjectMatch(actual, expected);
assertMatch(actual.fruit, /^[a-zA-Z]+$/);
```

kevin-zou:
> Make an assertion that `expected` object is a subset of `actual` object, deeply. If not, then throw.

I'm realizing now that I mixed up `expected` and `actual` in the docs. When I initially read it I thought `actual` had to be a subset of `expected`.

Thanks for your help, this is great! Do you think the docs would benefit from an example like this?

kt3k:
> Do you think the docs would benefit from an example like this?

I think so. Contributions to docs are very welcome

PR #5703, opened 2024-08-17
docs(assert): Add additional example in assertObjectMatch docs

kevin-zou:
As title describes, add an example in usage of `assertObjectMatch` to show that the `expected` object can be a subset of the `actual` object but the inverse will result in an assertion failure.

Discussed in #5682


kt3k (approved):
LGTM. Nice! Thanks

PR #5704, opened 2024-08-17
test(yaml): add edge case test for `stringify`

timreichen:
**Changes**
This PR adds a test for the edge case when `state.writeNode()` returns false in `stringify()`


kt3k (approved):
Confirmed this improves the line coverage of `stringify` method of `DumperState`. Nice!

LGTM

iuioiua:
For future reference, @timreichen, please use the `test` prefix in the title for such PRs that only affect test coverage.

PR #5708, opened 2024-08-18
fix(streams/unstable): fix type error in `to_lines.ts`

uki00a:
This PR changes the type of the `options` argument of `toLines()` from `StreamPipeOptions` to `PipeOptions` to resolve type error.


uki00a:
Oh sorry, type error was caused by my local Deno version being a bit old :pray:

PR #5710, opened 2024-08-18
chore: fix workspace package versions

kt3k:


iuioiua (approved):
I thought we had a script that checked for such errors 😓

kt3k:
The script only checks `^` symbol after `@` unfortunately

PR #5716, opened 2024-08-18
refactor(crypto): remove old unused `_fnv` folder

timreichen:


kt3k:
This change itself makes sense, but it looks like `crypto/_fnv/mod.ts` itself doesn't seem being used now (Probably it became unused after https://github.com/denoland/std/pull/4515 ). Can you instead remove `crypto/_fnv` dir instead?

timreichen:
> This change itself makes sense, but it looks like `crypto/_fnv/mod.ts` itself doesn't seem being used now (Probably it became unused after #4515 ). Can you instead remove `crypto/_fnv` dir instead?

done

iuioiua (approved):
Nice spot, Yoshiya and Tim! Surprised I didn't notice this myself.

PR #5718, opened 2024-08-19
refactor(path,streams): align additional error messages

irbull:
Aligns the error messages in the `streams` and `path` folders.


timreichen:
I think all error messages with `...must not be...` should reflect the current state, aka `...is...`

streams/byte_slice_stream.ts
```diff
@@ -48,7 +48,9 @@ export class ByteSliceStream extends TransformStream<Uint8Array, Uint8Array> {
     super({
       start: () => {
         if (start < 0) {
-          throw new RangeError("`start` must be greater than 0");
+          throw new RangeError(
+            "Invalid argument: `start` must not be negative",
```
irbull, line 52:
```suggestion
            "Invalid argument: `start` is negative",
```

irbull:
> I think all error messages with `...must not be...` should reflect the current state, aka `...is...`

There are two cases we commonly see:

1. **must not be** where we see things like: `Invalid argument: "start" must not be negative`, so this would change to `Invalid argument: "start" is negative`

But wouldn't this be even better if we had the resolution here? `Invalid argument: `start` must be positive`?

2. **must be** where we see things like: `Path must be absolute, path value "${path}"`. In this case we are saying the current state (path value ${path}) but we are also saying what MUST happen, it must be absolute. The alternative is `Cannot convert to file URL, path value is ${path}` which doesn't really help explain what I need to do.

streams/byte_slice_stream.ts
irbull:
```suggestion
            `Invalid argument, 'start' must >= 0: current value ${start}` ,
```

path/_common/from_file_url.ts
```diff
@@ -4,7 +4,7 @@
 export function assertArg(url: URL | string) {
   url = url instanceof URL ? url : new URL(url);
   if (url.protocol !== "file:") {
-    throw new TypeError("Must be a file URL.");
+    throw new TypeError("Must be a file URL");
```
irbull, line 7:
```suggestion
    throw new TypeError(`URL must be a file URL: URL protocol is ${url.protocol}`);
```

path/from_file_url_test.ts
```diff
@@ -17,12 +17,12 @@ Deno.test("posix.fromFileUrl()", function () {
   assertThrows(
     () => posix.fromFileUrl("http://localhost/foo"),
     TypeError,
-    "Must be a file URL.",
+    "Must be a file URL",
```
irbull, line 20:
```suggestion
    "URL must be a file URL: URL protocol is http",
```

kt3k:
> 8. Additional information should describe the current state, if possible, it should also describe the desired state in an affirmative voice:
> 
> ```
> Bad: Cannot compute the square root for x, value must not be negative
> Better: Cannot compute the square root for x: current value is ${x}
> Best: Cannot compute the square root for x, value must be positive: current value is ${x}
> ```

This generally sounds a good idea to me. Can you also send a PR to the style guide?

A minor point: I'm a bit unsure about `,` usage in 'Best' example. Is this comma grammatically fine? What other runtime do for this situation?

irbull:
> This generally sounds a good idea to me. Can you also send a PR to the style guide?
> 
> A minor point: I'm a bit unsure about `,` usage in 'Best' example. Is this comma grammatically fine? What other runtime do for this situation?

Done https://github.com/denoland/std/pull/5757

streams/byte_slice_stream.ts
irbull: Updated to `'start' must be >= 0: received ${start}`

path/_common/from_file_url.ts
irbull: Updated to `URL must be a file URL: received "${url.protocol}"`

streams/buffer.ts
```diff
@@ -315,7 +315,9 @@ export class Buffer {
       return;
     }
     if (n < 0 || n > this.length) {
-      throw Error("bytes.Buffer: truncation out of range");
+      throw Error(
+        `Error truncating buffer, truncation value "${n}" out of range: value must be between 0 and ${this.length}`,
```
kt3k, line 319:
```suggestion
        `Cannot truncate the buffer as the truncation value "${n}" is out of range: value must be between 0 and ${this.length}`,
```

iuioiua: Seems a little long, no?

kt3k: But that is what the style guide now recommend

timreichen: How about `Buffer truncation value "${n}" is out of range: value must be between 0 and ${this.length}`

streams/byte_slice_stream.ts
irbull: Settled on `Cannot construct ByteSliceStream as start must be >= 0: received ${start}`

streams/buffer.ts
irbull: I applied this

kt3k (approved):
LGTM

Issue #5612, opened 2024-08-02
to-do: enable `no-console` lint rule


iuioiua:
We should ensure that [`console`](https://developer.mozilla.org/en-US/docs/Web/API/console) methods aren't used incorrectly across the codebase. We can enforce this by enabling the [`no-console`](https://lint.deno.land/rules/no-console) rule. Any instances where it should be kept should have valid justifications.

This should probably broken up into separate PRs.

Inspired by #5610

PR #5723, opened 2024-08-19
chore: remove obsolete log

timreichen:
**Changes**
This PR removes success logs at the end of the check.

**Reasoning**
Other tools only log if they fail.


iuioiua:
Does this deliberately target #5612?

iuioiua (commented):
I'm impartial to this change. I'll defer to Yoshiya's review.

kt3k:
I'm not in favor. These logs are useful.

timreichen:
@kt3k should all tools log a success message if zhey do not fail?

kt3k:
I don't think we need strict consistency with tools. We should rather spend time on polishing user-facing APIs (in my opinion)

PR #5724, opened 2024-08-19
refactor(log): replace `protected` properties with `Symbol`s

timreichen:
**Changes**
This PR replaces `protected` properties with `Symbol`s. 

**Reasoning**
Either the properties should be public and not have an underscore or private and thus not accessible.


kt3k (approved):
LGTM

PR #5726, opened 2024-08-19
chore: avoid deno.lock incompatibility in ci

kt3k:


iuioiua (approved):
Shouldn't we just set `lock` to `false` in the config file?

PR #5732, opened 2024-08-20
docs(datetime): add constants to module docs

iuioiua:
Addresses https://github.com/denoland/std/issues/5695#issuecomment-2296582737


kt3k (approved):
LGTM

PR #5737, opened 2024-08-20
refactor(log): make `BaseHandler` abstract

timreichen:
Ref: https://github.com/denoland/std/pull/5702
**Changes**
This PR makes the `BaseHandler` and `log()` method abstract.
It removes the obsolete `override` for `log()` in `FileHandler` and `ConsoleHandler`.


kt3k (approved):
LGTM

PR #5738, opened 2024-08-20
refactor(yaml): rename `DumperState` methods

timreichen:
**Changes**
This PR renames `DumperState` methods fro `write*()` to `stringify*()`

**Reasoning**
https://github.com/denoland/std/pull/5683 removed the `dump` property. All these functions now turn passed data into a string and return it.


iuioiua (changes requested):
Can you please add explicit return types?

iuioiua (approved):

PR #5739, opened 2024-08-20
refactor(yaml): cleanup `stringifyNode()`

timreichen:
**Changes**
This PR removes `else` block due to early return.
It simplifies `isObject(object)` and `Array.isArray(object)` nesting.
It moves `duplicate` logic up to avoid early `compact` check.
~It moves `level` inside the options object.~


iuioiua (commented):
Just using `isObject()` instead of `isObject() || Array.isArray()` makes sense. But frankly, the rest of this PR moves code around with no clear indication that the the method has indeed been "cleaned up". I think we should only focus on efforts on cleaning up code if the benefit is obvious. So I'll close this PR. Either way, thank you for the PR.

PR #5740, opened 2024-08-20
refactor(yaml): move options to options object

timreichen:
**Changes**
This PR moves options to an options object in all `DumperState`.
**Reasoning**
All methods are now structured the same, taking a value as first argument and options as the second argument.


kt3k (approved):
LGTM

PR #5743, opened 2024-08-20
refactor(yaml): cleanup `stringifyFlowMapping()`

timreichen:


iuioiua (commented):
This method is not yet fully tested. I'd only feel completely comfortable merging this PR once it's fully tested. See https://app.codecov.io/gh/denoland/std/blob/main/yaml%2F_dumper_state.ts#L627

yaml/_dumper_state.ts
```diff
@@ -628,52 +628,41 @@ export class DumperState {
...
-        continue; // Skip this pair because of invalid value.
-      }
+      let pairBuffer = "";
 
+      pairBuffer += quote;
+      if (keyString.length > 1024) pairBuffer += "? ";
+      pairBuffer += keyString;
+      pairBuffer += quote;
+      pairBuffer += separator;
       pairBuffer += valueString;
```
iuioiua, line 660: Can be done in one line.

iuioiua (changes requested):
Frankly, I don't really see the purpose behind this PR.

timreichen:
> Frankly, I don't really see the purpose behind this PR.

There are several (tiny) performance benefits:
- `quote` and `separator` are take out of the loop, therefore only need one check
- `stringifyNode()`for key and value is called at the beginning of the loop, thus continues the loop before `pairBuffer` is even initialized.

It also is more readable imo.

iuioiua (approved):
Thank you

PR #5745, opened 2024-08-20
fix(dotenv): show warning when there is invalid key in `.env` file

### LINKED ISSUE #5619: to-do(dotenv): document key restrictions

It seems that the current dotenv implementation restricts certain strings from being used as keys. For example, keys like `'1a'` and `'process.env.xxx'` are not allowed.

I'd like to see these restrictions removed so that these keys can be used.

I'd appreciate an explanation of the rationale behind these restrictions. Or is it possible that I just haven't found the right way to do this?

I found npm dotenv package allowing this.

kt3k:
Currently `parse` of `dotenv` silently ignores the invalid key entries in `.env` file. This PR adds the warning like `Ignored the key "1INVALID" as it is not a valid identifier: The key need to match the pattern /^[a-zA-Z_][a-zA-Z0-9_]*$/.` when `.env` file has invalid key entry.

closes #5619

### DIFF
--- a/dotenv/parse.ts
+++ b/dotenv/parse.ts
@@ -10,7 +10,9 @@ type LineParseResult = {
 type CharactersMap = { [key: string]: string };
 
 const RE_KEY_VALUE =
-  /^\s*(?:export\s+)?(?<key>[a-zA-Z_]+[a-zA-Z0-9_]*?)\s*=[\ \t]*('\n?(?<notInterpolated>(.|\n)*?)\n?'|"\n?(?<interpolated>(.|\n)*?)\n?"|(?<unquoted>[^\n#]*)) *#*.*$/gm;
+  /^\s*(?:export\s+)?(?<key>[^\s=#]+?)\s*=[\ \t]*('\n?(?<notInterpolated>(.|\n)*?)\n?'|"\n?(?<interpolated>(.|\n)*?)\n?"|(?<unquoted>[^\n#]*)) *#*.*$/gm;
+
+const RE_VALID_KEY = /^[a-zA-Z_][a-zA-Z0-9_]*$/;
 
 const RE_EXPAND_VALUE =
   /(\${(?<inBrackets>.+?)(\:-(?<inBracketsDefault>.+))?}|(?<!\\)\$(?<notInBrackets>\w+)(\:-(?<notInBracketsDefault>.+))?)/g;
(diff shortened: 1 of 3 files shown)


timreichen:
I wonder if it should throw instead of warn. Doesn't deno throw on invalid keys like this?

kt3k:
`deno run --env` warns for invalid key entries:

```shellsession
$ deno run --env a.js
Warning Parsing failed within the specified environment file: .env at index: 0 of the value: 1bar=FOOBAR
1
```

dotenv/parse.ts
```diff
@@ -10,7 +10,9 @@ type LineParseResult = {
 type CharactersMap = { [key: string]: string };
 
 const RE_KEY_VALUE =
-  /^\s*(?:export\s+)?(?<key>[a-zA-Z_]+[a-zA-Z0-9_]*?)\s*=[\ \t]*('\n?(?<notInterpolated>(.|\n)*?)\n?'|"\n?(?<interpolated>(.|\n)*?)\n?"|(?<unquoted>[^\n#]*)) *#*.*$/gm;
+  /^\s*(?:export\s+)?(?<key>[^\s=#]+?)\s*=[\ \t]*('\n?(?<notInterpolated>(.|\n)*?)\n?'|"\n?(?<interpolated>(.|\n)*?)\n?"|(?<unquoted>[^\n#]*)) *#*.*$/gm;
```
iuioiua, line 13: Can you please explain this regex change?

dotenv/parse_test.ts
```diff
@@ -3,11 +3,14 @@
 import { assertEquals } from "@std/assert";
 import { parse } from "./parse.ts";
 import * as path from "@std/path";
+import { assertSpyCall, spy } from "@std/testing/mock";
 
 const moduleDir = path.dirname(path.fromFileUrl(import.meta.url));
 const testdataDir = path.resolve(moduleDir, "testdata");
 
 Deno.test("parse()", () => {
+  using consoleWarnSpy = spy(console, "warn");
```
iuioiua, line 12: Wow - great idea! I wonder where else in the codebase we can spy on `console` instead of spawning new processes and reading the output.

iuioiua (approved):
Nice

dotenv/parse.ts
kt3k: The key part `(?<key>[a-zA-Z_]+[a-zA-Z0-9_]*?)` has been changed to `(?<key>[^\s=#]+?)`.

Now this regexp matches even with invalid key entries (like `1A=FOO`), and the key validation is performed later inside the while loop. That makes us possible to warn at invalid keys. Previously the lines with invalid keys were completely skipped and there was no room to warn with it.

PR #5746, opened 2024-08-20
chore: check import map versions more strictly

kt3k:
This PR makes the check of import map stricter.

Now the dependency of `@std/foo` needs to be in the form of `@std/foo@^x.y.z` where `x.y.z` part needs to match the version number in `./foo/deno.json`

This prevents the error like #5710


iuioiua (approved):
Nice

PR #5747, opened 2024-08-20
feat(path/unstable): support `URL` input in `dirname()`

kt3k:
part of #5537

This PR adds `dirname(url: URL): string` overload to dirname function.

If a user have a file URL string or object, currently they need to convert it to a path string first to perform path operation on it:

```ts
import { dirname, fromFileUrl } from "@std/path";

dirname(fromFileUrl(new URL(import.meta.url)));
```

This change makes it simpler by automatically interpret the file URL as path

```ts
import { dirname } from "@std/path";

dirname(new URL(import.meta.url));
```

### DIFF
--- a/path/dirname.ts
+++ b/path/dirname.ts
@@ -23,6 +23,29 @@ import { dirname as windowsDirname } from "./windows/dirname.ts";
  * @param path Path to extract the directory from.
  * @returns The directory path.
  */
-export function dirname(path: string): string {
-  return isWindows ? windowsDirname(path) : posixDirname(path);
+export function dirname(path: string): string;
+/**
+ * Return the directory path of a file URL.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @example Usage
+ * ```ts
+ * import { dirname } from "@std/path/dirname";
+ * import { assertEquals } from "@std/assert";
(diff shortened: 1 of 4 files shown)


iuioiua (approved):
LGTM once `assertThrows()` comments are addressed.

This PR does make me think; what if someone does want to get the parent path of a HTTP URL? E.g. I want to get `/documents` from `https://example.com/documents/image.png`. How do I do that? Is that regarded as the directory in a HTTP URL? I think it's a valid use case and we should cater to it.

iuioiua:
Sorry, I want to understand; is there a reason for only allowing `dirname()` to only work on file URLs?

> `pathname` part of URL object is usually a posix path. I think we can recommend using `posix.dirname` to it:
> 
> ```
> > posix.dirname(new URL("https://example.com/documents/image.png").pathname)
> "/documents"
> ```

I think your example should work without having to call `.pathname`.

kt3k:
If we simply allow http urls, what happens with the call like `posix.dirname(new URL("https://example.com"))`?

iuioiua:
We could throw or return null.

kt3k:
http URLs are not paths while file URLs are equivalent of paths. I think that kind of handling is out of scope of `@std/path`

PR #5748, opened 2024-08-20
feat(front-matter/unstable): add pass-through `ParseOptions` argument for YAML parser

### LINKED ISSUE #5677: front-matter: allow YAML parsing to be reconfigured (or re-expose generic extraction)
Pre-1.0, `front-matter` exposed a `createExtractor` function to create extractors with custom parsers. I used this to use a YAML parser with a different schema. Since `front-matter` 1.0 only exposes the extractor functions and they have no configuration knobs, it is impossible to change the parser options (or use a different parser) with the public API.

Given that the `extractAndParse` function requires a format-specific regex, it looks like the best solution for my particular use case (configuring the yaml parser) would be to add an `options: ParserOptions` to `extractYaml`. Re-exposing generic front-matter extraction that returns plain text for the client to parse would also be useful for other extensions, but looks more difficult from my understanding of the current code and is not necessary for my immediate problem.

The alternatives I have considered are sticking with the last 0.2XX release of `front-matter` or importing `extractAndParse` directly from `_shared.ts`, but the latter looks impractical. I have tried seeing if I can get away with the standard schema, but due to legacy content the standard Yaml schema parses dates in incorrect time zones in my data.

mdekstrand:
This adds an `options` argument to `front_matter/yaml/extract` to pass options to the underlying YAML parser, such as schema selection, pursuant to #5677.

Closes #5677

### DIFF
--- a/front_matter/yaml.ts
+++ b/front_matter/yaml.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
 import { extractAndParse, type Parser } from "./_shared.ts";
-import { parse, ParseOptions } from "@std/yaml/parse";
+import { parse, type ParseOptions } from "@std/yaml/parse";
 import type { Extract } from "./types.ts";
 import { EXTRACT_YAML_REGEXP } from "./_formats.ts";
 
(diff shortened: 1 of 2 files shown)


kt3k:
This looks reasonable to me. I'll work on this a bit to mark the new options are unstable feature.

front_matter/yaml.ts
iuioiua, line 1: Can you please add one `@example` showcasing the effects of the `options` argument?

iuioiua (changes requested):

kt3k (approved):
LGTM

iuioiua (approved):

PR #5749, opened 2024-08-20
refactor(yaml): move state functions into `LoaderState` class

timreichen:
**Changes**
This PR moves state functions into the `LoaderState` class as methods.

**Reasoning**
It removes the need to pass `state` as an argument.
This change also allows for the removal of obsolete code and optimizations of some functions in other PRs.

**Note**
The diff of this PR looks like there are lots of changes. The code inside the functions is untouched except using `this` instead of the `state`.

### DIFF
--- a/yaml/_loader_state.ts
+++ b/yaml/_loader_state.ts
@@ -210,6 +210,1369 @@ export class LoaderState {
     this.onWarning?.(error);
   }
 
+  yamlDirectiveHandler(...args: string[]) {
+    if (this.version !== null) {
+      return this.throwError("duplication of %YAML directive");
+    }
+
+    if (args.length !== 1) {
+      return this.throwError("YAML directive accepts exactly one argument");
+    }
+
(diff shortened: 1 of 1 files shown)


yaml/_loader_state.ts
```diff
@@ -210,1069 +210,898 @@ export class LoaderState {
     this.onWarning?.(error);
   }
 
-  readDocument() {
```
iuioiua, line 213: Why was `readDocument()` moved?

iuioiua (commented):
Frankly, this seems like a lot of trouble just to avoid having to pass in an extra argument. And it's hard to discern that this only moves functions into `LoaderState`. Like, this would be far more understandable if it were broken up into smaller PRs.

yaml/_loader_state.ts
timreichen: It wasn't, `readDocument()` and `readDocuments()` were at the end of the class where they still are.

timreichen:
I'll close this PR and move every function individually.

PR #5750, opened 2024-08-20
refactor(yaml): replace `chooseScalarType()` callback with `implicitTypes ` argument

timreichen:
**Changes**
This PR removes the callback of `chooseScalarType()` and passes `implicitTypes` instead.

**Reasoning**
This allows to the inline the code of the utility function `testImplicitResolving()`, which is clearer to read and understand.


iuioiua (approved):
Thanks!

PR #5751, opened 2024-08-20
perf(cli,http,ini,internal,media-types): replace `trim()` comparisons with regexps

timreichen:
**Changes**
This replaces comparisons done with `trim()` with `RegExp.test()`

**Reasoning**
`trim()` creates a new string which is only used for the comparison and thus not memory efficient.


kt3k (approved):
LGTM. Nice!

PR #5752, opened 2024-08-20
fix(front-matter): remove os specific newline

timreichen:
**Changes**
This PR removes the os specific check for `\r?` before `\n`.

**Reasoning**
This is the only occurrence where a os check is done. Even the `recognize()` function in the same module doesn't do it:
https://github.com/denoland/std/blob/99200dcd9039eea515f3f850b2f13cf7be585d03/front_matter/_shared.ts#L36

### DIFF
--- a/front_matter/_formats.ts
+++ b/front_matter/_formats.ts
@@ -21,9 +21,7 @@ const JSON_HEADER = `(---json|${JSON_DELIMITER})\\s*`;
 const JSON_FOOTER = `(?:---|${JSON_DELIMITER})`;
 
 const DATA = "([\\s\\S]+?)";
-const NEWLINE = `${
-  globalThis?.Deno?.build?.os === "windows" ? "\\r?" : ""
-}(?:\\n)?`;
+const NEWLINE = "\\r?\\n?";
 
 export const RECOGNIZE_YAML_REGEXP = new RegExp(`^${YAML_HEADER}$`, "im");
 export const RECOGNIZE_TOML_REGEXP = new RegExp(`^${TOML_HEADER}$`, "im");


iuioiua (approved):
The reasoning makes sense to me. But this is definitely not a chore. It's either a breaking change or refactor.

front_matter/_formats.ts
```diff
@@ -21,9 +21,7 @@ const JSON_HEADER = `(---json|${JSON_DELIMITER})\\s*`;
 const JSON_FOOTER = `(?:---|${JSON_DELIMITER})`;
 
 const DATA = "([\\s\\S]+?)";
-const NEWLINE = `${
-  globalThis?.Deno?.build?.os === "windows" ? "\\r?" : ""
-}(?:\\n)?`;
+const NEWLINE = "\\r?\\n?";
```
iuioiua, line 24:
```suggestion
const NEWLINE = "\\r?(?:\\n)?";
```

iuioiua (changes requested):

front_matter/_formats.ts
timreichen: I think these brackets don't do anything.

kt3k:
I don't think this breaks any existing practical usage as the condition is now more permissive (windows-style new line is now allowed in linux and mac).

How about tagging this as 'fix'?

iuioiua:
It's more permissive but changes expected behavior, no?

kt3k:
If we change the parsing behavior depending on the platform, that means the users can't share the exact same markdown files between different platforms (e.g. If windows user creates a markdown file with front-matter with CRLF, then that file can't be parsed in mac/linux). I don't think that meets the expectation of the users.

front_matter/_formats.ts
iuioiua: Let's merge as-is but keep this suggestion in mind in case anyone runs into any issues.

iuioiua (approved):

PR #5753, opened 2024-08-20
docs: specifically disallow periods in error messages

irbull:
Error messages should not include a period, as it will result in multiple sentences. Colons are the preferred punctuation but other punctuation may make sense in some cases.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -245,11 +245,11 @@ Bad: Can't parse input x
 Good: Cannot parse input x
 ```
 
-7. Messages should use a colon when providing additional information:
+7. Messages should use a colon when providing additional information. Periods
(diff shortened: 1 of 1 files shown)


.github/CONTRIBUTING.md
```diff
@@ -245,11 +245,10 @@ Bad: Can't parse input x
 Good: Cannot parse input x
 ```
 
-7. Messages should use a colon when providing additional information:
+7. Messages should not use a period when providing additional information:
```
kt3k, line 248: I feel this obscures that we prefer colon. Can we mention both colon and period?

irbull: Updated it to read:

```
7. Messages should use a colon when providing additional information. Periods
   should never be used. Other punctuation may be used as needed:
```

kt3k (approved):
LGTM

PR #5756, opened 2024-08-21
chore: fix `lint:docs` task concurrency

iuioiua:
This should fix concurrency issues with the `lint:docs` task. See https://github.com/denoland/std/actions/runs/10484008177/job/29037626131


kt3k (approved):
LGTM

PR #5763, opened 2024-08-21
refactor(yaml): inline `yamlDirectiveHandler()`

timreichen:
Ref: https://github.com/denoland/std/pull/5749


iuioiua (approved):
LGTM. Thank you.

PR #5782, opened 2024-08-22
refactor(yaml): remove `any` types

timreichen:


iuioiua (changes requested):
Can you please ensure all of these functions have explicit return types?

iuioiua (approved):
Thanks!

PR #5784, opened 2024-08-22
refactor(assert): align additional error messages

irbull:
Aligns the error messages in the `assert` folder to match the style guide.

### DIFF
--- a/assert/almost_equals.ts
+++ b/assert/almost_equals.ts
@@ -46,7 +46,7 @@ export function assertAlmostEquals(
     return;
   }
 
-  const msgSuffix = msg ? `: ${msg}` : ".";
+  const msgSuffix = msg ? `: ${msg}` : "";
   const f = (n: number) => Number.isInteger(n) ? n : n.toExponential();
   throw new AssertionError(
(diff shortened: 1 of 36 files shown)


kt3k:
This will break some of Deno CLI's snapshot testing as they asserts the entire assertion error output (like this one: https://github.com/denoland/deno/blob/d54d29662f30c0fa5e1f048fdce4835e51248682/tests/specs/test/hide_stacktraces/dot.out )

I'm not sure at this point how this can be disruptive to the ecosystem in a similar way as Deno CLI repo.

Maybe this needs more discussion/investigation before landing.

irbull:
@kt3k Good catch, and that's fair. While I like the idea of consistent error messages I think we need to be pragmatic too. I don't mind looking through those assertions and trying to update them, but that doesn't help the broader community if they too rely on specific assertions in their tests.

irbull:
Here is an alternative way of handling this: https://github.com/denoland/std/pull/5797

kt3k:
#5797 makes sense to me. Closing this one for now

PR #5787, opened 2024-08-23
docs: update badge URL

iuioiua:
This change deletes the local badge file and instead calls directly to `shields.io`. It'll break only a few spots that use the current badge, but the note should help them figure out how to update.


badge.svg
kt3k, line 1: I'm personally not in favor of removing this file as this breaks many existing usages:

- https://github.com/fm-elpac/pmbb
- https://github.com/catuhana/denoh
- https://github.com/eser/stack
- https://github.com/babiabeo/deko
- https://github.com/stSoftwareAU/NEAT-AI
- https://github.com/Thesephi/oak-routing-ctrl
- https://github.com/thomas3577/tano
- https://github.com/420bytes/web-core
- https://github.com/SamTherapy/deno-deploy

README.md
```diff
@@ -78,21 +78,25 @@ package versions <1.0.0 follow
 
 ## Badge
 
-[![Built with the Deno Standard Library](./badge.svg)](https://jsr.io/@std)
+> [!NOTE]
+> Previously, this repo hosted the badge SVG file. Now, the badge is retrieved
+> directly from [Shields.io](https://shields.io/).
+
+[![Built with the Deno Standard Library](https://img.shields.io/badge/Built_with_deno__std-blue?logo=deno)](https://jsr.io/@std)
```
kt3k, line 85: Do we still use `deno_std` name? Or should we rename it to `std`?

iuioiua: `std` seems fine to me

kt3k (approved):
LGTM

PR #5795, opened 2024-08-23
refactor(yaml): remove type mod file

timreichen:
**Changes**
This PR removes the `_types/mod.ts` file

**Reasoning**
There seems no advantage over importing types from their file directly.


kt3k (approved):
LGTM

PR #5797, opened 2024-08-23
docs: add a note about error message format exception

irbull:
The assertion package uses a period at the end of the error message. While this doesn't conform to the standard error message format, it would be a breaking change to adjust this now. This change-set adds a note to the CONTRIBUTING.md file to explain this exception.


kt3k:
Inline comment on every file feels a bit excessive. Can you only add `Exceptions` section in contributing guide?

irbull:
> Inline comment on every file feels a bit excessive. Can you only add `Exceptions` section in contributing guide?

Sounds good. I've updated the PR and rebased it on main as well.

kt3k (approved):
LGTM

PR #5801, opened 2024-08-24
refactor(log): align additional error messages

irbull:
Aligns the error messages in the `log` folder to match the style guide.

### DIFF
--- a/log/rotating_file_handler.ts
+++ b/log/rotating_file_handler.ts
@@ -67,11 +67,13 @@ export class RotatingFileHandler extends FileHandler {
   override setup() {
     if (this.#maxBytes < 1) {
       this.destroy();
-      throw new Error("maxBytes cannot be less than 1");
+      throw new Error(`"maxBytes" must be >= 1: received ${this.#maxBytes}`);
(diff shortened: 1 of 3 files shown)


log/rotating_file_handler.ts
```diff
@@ -67,11 +67,13 @@ export class RotatingFileHandler extends FileHandler {
...
       this.destroy();
-      throw new Error("maxBytes cannot be less than 1");
+      throw new Error(`"maxBytes" must be >= 1: received ${this.#maxBytes}`);
     }
     if (this.#maxBackupCount < 1) {
       this.destroy();
-      throw new Error("maxBackupCount cannot be less than 1");
+      throw new Error(
+        `"maxBackupCount" must be >= 1: received ${this.#maxBackupCount}`,
+      );
```
irbull, line 76: Would it be better if these checks were moved to the constructor and we could fast fail then? Or is there a reason someone would construct the object and not want to call `setup`, and thus want this invalid state?

kt3k (approved):
LGTM

PR #5805, opened 2024-08-24
refactor(yaml): remove `Mark` class

timreichen:
**Changes**
This PR removes the `Mark` class and replaces it with functions.

**Reasoning**
`Mark` is only used in `#createError()`. It seems inefficient to instantiate a class instead of creating and returning the string directly in a function.


iuioiua (approved):
Nice catch. Thank you.

PR #5806, opened 2024-08-24
refactor(yaml): align additional error messages

irbull:
Aligns the error messages in the `yaml` folder to match the style guide.


yaml/_dumper_state.ts
```diff
@@ -870,9 +874,7 @@ export class DumperState {
       } else {
         if (this.skipInvalid) return null;
         throw new TypeError(
-          `unacceptable kind of an object to dump ${
-            getObjectTypeString(value)
-          }`,
+          `Cannot dump object of type: ${getObjectTypeString(value)}`,
```
kt3k, line 877: The verb 'dump' is only used internally. The public API is named `stringify`. So let's use 'stringify' here.

```suggestion
          `Cannot stringify object of type: ${getObjectTypeString(value)}`,
```

yaml/_loader_state.ts
```diff
@@ -1661,7 +1697,7 @@ function composeNode(
         }
       }
     } else {
-      return state.throwError(`unknown tag !<${state.tag}>`);
+      return state.throwError(`Unknown tag !<${state.tag}>`);
```
kt3k, line 1700:
```suggestion
      return state.throwError(`Cannot resolve unknown tag !<${state.tag}>`);
```

kt3k:
Some messages use `Error 'verb'-ing` form and some uses `Cannot 'verb'` form. I think these 2 forms are interchangeable. Should we align these 2 forms into the 1 form, or should we just allow using either of them?

irbull:
> Some messages use `Error 'verb'-ing` form and some uses `Cannot 'verb'` form. I think these 2 forms are interchangeable. Should we align these 2 forms into the 1 form, or should we just allow using either of them?

I think consistency would be good here. I think we've used `Cannot 'verb'` more often, I'll take a pass and align them. We've come this far.

kt3k (approved):
LGTM. Thanks!

PR #5807, opened 2024-08-24
test(text): move case tests to separate files

timreichen:
**Changes**
This PR moves all text case tests to separate files.

**Reasoning**
Case functions lived in the same file when they were [introduced](https://github.com/denoland/std/pull/4082/files) and are now split. This single test file is a leftover.


kt3k (approved):
LGTM

PR #5808, opened 2024-08-24
test(front-matter): remove duplicate tests

timreichen:
**Changes**
This PR removes duplicate tests. All removed tests already exists in `front_matter/test_test.ts`


kt3k (approved):
> All removed tests already exists in front_matter/test_test.ts

This seems true. Thanks for catching this

PR #5809, opened 2024-08-24
fix(front-matter): extend generic `Extract` type

timreichen:
**Changes**
This PR makes `Extract` generic extend `Record<string, unknown>` as `Parser` returns the same type.


kt3k:
I think `T` can be `string`, `number`, `string[]`, etc.

In the below example T is `string[]`

```ts
import { extract } from "@std/front-matter/yaml";

const x = extract<string[]>(`---
- foo
- bar
- baz
---
body
`);

console.log(x);
```

timreichen:
> I think `T` can be `string`, `number`, `string[]`, etc.
> 
> 
> 
> In the below example T is `string[]`
> 
> 
> 
> ```ts
> 
> import { extract } from "@std/front-matter/yaml";
> 
> 
> 
> const x = extract<string[]>(`---
> 
> - foo
> 
> - bar
> 
> - baz
> 
> ---
> 
> body
> 
> `);
> 
> 
> 
> console.log(x);
> 
> ```

In that case `Parser` probably has a wrong return type. I'll investigate and update this PR.

kt3k:
I don't see the reason to introduce type parameters to `toml.parse` and `yaml.parse`. These changes seem undesirable and also unnecessary to me

kt3k:
> This PR makes Extract generic extend Record<string, unknown> as Parser returns the same type.

What's the purpose of limiting that type? I don't see the reason well.

kt3k:
The result of `parse` can vary depending on the config language. I think it is difficult to type these correctly, and also I don't see much value in this.

iuioiua:
I tend to agree with Yoshiya. I think these changes aren't needed.

iuioiua:
Closing this for the reasons above. Either way, thank you for the contribution.

PR #5815, opened 2024-08-25
test(yaml): handles array entry that starts with newline

timreichen:


iuioiua:
It seems this scenario is already tested in https://github.com/denoland/std/blob/b91817f29d36915269a111aba0d8c44c31c385f3/yaml/stringify_test.ts#L419-L420. Either way, thanks for the PR. Closing.

PR #5816, opened 2024-08-25
test(yaml): add `undefined` object entry test with `skipInvalid` option

timreichen:


iuioiua:
This test already exists in https://github.com/denoland/std/blob/b91817f29d36915269a111aba0d8c44c31c385f3/yaml/stringify_test.ts#L191-L194. Perhaps for the future, you can check Codecov to see if test coverage has increased as an indication as to whether a test already exists. Either way, thanks for the PR! I'll close this now.

PR #5818, opened 2024-08-26
feat(path/unstable): support `URL` in `extname()`

iuioiua:
Towards #5537

### DIFF
--- a/path/extname.ts
+++ b/path/extname.ts
@@ -22,6 +22,29 @@ import { extname as windowsExtname } from "./windows/extname.ts";
  * @param path Path with extension.
  * @returns The file extension. E.g. returns `.ts` for `file.ts`.
  */
-export function extname(path: string): string {
-  return isWindows ? windowsExtname(path) : posixExtname(path);
+export function extname(path: string): string;
+/**
+ * Return the extension of the path with leading period (".").
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @example Usage
+ * ```ts
+ * import { extname } from "@std/path/extname";
+ * import { assertEquals } from "@std/assert";
+ *
+ * if (Deno.build.os === "windows") {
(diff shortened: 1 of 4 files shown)


path/posix/extname.ts
```diff
@@ -18,26 +18,37 @@ import { isPosixPathSeparator } from "./_util.ts";
...
  * ```
  *
  * @param path The path to get the extension from.
  * @returns The extension (ex. for `file.ts` returns `.ts`).
  */
-export function extname(path: string): string {
+export function extname(path: URL): string;
+export function extname(path: string | URL): string {
+  if (path instanceof URL) {
+    path = path.pathname;
```
kt3k, line 49: Can you use `fromFileUrl` instead? This allows other random urls such as `http:`, `data:`, `jsr:`, `node:`, etc, which is not aligned to how `dirname` works with those url types.

iuioiua: What's the harm of using `.pathname`?

kt3k: In `data:` scheme, the pathname doesn't represent the path, but it represent arbitrary data. The return value of `extname` can be any random value in that case.

path/extname.ts
```diff
@@ -22,6 +22,29 @@ import { extname as windowsExtname } from "./windows/extname.ts";
...
+ *   assertEquals(extname(new URL("file:///home/user/Documents/image.png")), ".png");
+ * }
+ * ```
+ *
+ * @param path Path with extension.
+ * @returns The file extension. E.g. returns `.ts` for `file.ts`.
+ */
+export function extname(path: URL): string;
+export function extname(path: string | URL): string {
+  // deno-lint-ignore no-explicit-any
```
timreichen, line 48: Couldn't `any` be avoided with `toString()` or `String(path)`?

kt3k: We like to do `path instanceof URL` check in each implementation. `String(path)` makes that check not working.

kt3k (approved):
LGTM

PR #5821, opened 2024-08-26
test(cli): reduce flakiness of Spinner test

### LINKED ISSUE #5817: bug: `Spinner.message can be updated` test is flaky
It seems that these tests are often flaky. I think we need to rethink how these tests are done.

https://github.com/denoland/std/actions/runs/10549292100/job/29224157622?pr=5816

kt3k:
closes #5817

### DIFF
--- a/cli/testdata/spinner_cases/change_message.ts
+++ b/cli/testdata/spinner_cases/change_message.ts
@@ -4,7 +4,7 @@ const spinner = new Spinner();
 
 spinner.start();
 
-setTimeout(() => (spinner.message = "One dino 🦕"), 125); // 150
-setTimeout(() => (spinner.message = "Two dinos 🦕🦕"), 200); // 225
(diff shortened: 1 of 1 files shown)


cli/testdata/spinner_cases/change_message.ts
iuioiua, line 1: Can you clarify this change? You reduced the time to change to "One dino 🦕" to 10 ms to allow more time between changing to "Two dinos 🦕". Is that right?

kt3k: Ah, maybe it's also better to delay updating to `"Two dinos 🦕"` because updating `"Two dinos 🦕"` (200ms) somehow happens before the first rendering of spinner (75ms) according to the flaky failure.

iuioiua (approved):

PR #5822, opened 2024-08-26
refactor(yaml): use `isObject()`

timreichen:
**Changes**
This PR removes `isArray()` in combination with `isObject()`, because array is an object.


iuioiua (approved):
Thanks

PR #5826, opened 2024-08-26
chore: specify import map path in version_bump workflow

kt3k:
This PR specifies import map path in version_bump workflow using https://github.com/denoland/bump-workspaces/pull/31

This is now necessary as we use custom (non-`deno.json`) import map path.


iuioiua (approved):
Cool

PR #5829, opened 2024-08-26
test(yaml): add undefined object entry with skipInvalid and flowLevel option test 

timreichen:


kt3k (approved):
Confirmed this reduces 8 missed lines in the coverage report of `_dumper_state.ts`

LGTM

PR #5836, opened 2024-08-27
refactor(yaml): move `duplicate` and `duplicateIndex`

timreichen:
**Changes**
This PR moves `duplicate` and `duplicateIndex` code inside `isObject()` block, to avoid double check.

**Note** One line `(tag !== null && tag !== "?")` is not covered. Seems like tag is always null if the value is an object.

### DIFF
--- a/yaml/_dumper_state.ts
+++ b/yaml/_dumper_state.ts
@@ -785,16 +785,16 @@ export class DumperState {
       const duplicateIndex = this.duplicates.indexOf(value);
       const duplicate = duplicateIndex !== -1;
 
+      if (duplicate) {
+        if (this.usedDuplicates.has(value)) return `*ref_${duplicateIndex}`;
+        this.usedDuplicates.add(value);
+      }
       if (
         (tag !== null && tag !== "?") ||
         duplicate ||
         (this.indent !== 2 && level > 0)
       ) {
         compact = false;
-      }
-      if (duplicate) {
-        if (this.usedDuplicates.has(value)) return `*ref_${duplicateIndex}`;
-        this.usedDuplicates.add(value);
       }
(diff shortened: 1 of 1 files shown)


yaml/_dumper_state.ts
```diff
@@ -781,69 +781,61 @@ export class DumperState {
...
-              value = `&ref_${duplicateIndex}${value}`;
-            }
-          } else {
-            value = this.stringifyFlowMapping(value, { level });
-            if (duplicate) {
-              value = `&ref_${duplicateIndex} ${value}`;
-            }
+      if (duplicate) {
+        if (this.usedDuplicates.has(value)) return `*ref_${duplicateIndex}`;
+        this.usedDuplicates.add(value);
```
iuioiua, line 790: This doesn't seem to pertain to the main objective of this PR.

yaml/_dumper_state.ts
```diff
@@ -781,69 +781,61 @@ export class DumperState {
...
+      }
+      if (
+        (tag !== null && tag !== "?") ||
+        duplicate ||
+        (this.indent !== 2 && level > 0)
+      ) {
+        compact = false;
+      }
```
iuioiua, line 798: Why was this moved?

timreichen: Because compact only needs to be evaluated after `if (this.usedDuplicates.has(value)) return ``*ref_${duplicateIndex}``;` doesn't return early.

yaml/_dumper_state.ts
timreichen: `if (duplicate) {...}` is would be checked twice otherwise, no?

iuioiua (approved):
Thank you

PR #5839, opened 2024-08-27
refactor(yaml): change error message in `stringifyNode()`

timreichen:
**Changes**
This PR changes the error message in `stringifyNode()`.
**Reasoning**
The error will not throw if the value is an object. so `object` should be dropped from the message.
For the same reason, `typeof` should suffice.


iuioiua (approved):
Nice. Thank you.

PR #5844, opened 2024-08-27
refactor(yaml): simplify pair `construct()` function

timreichen:
**Changes**
This PR simplifies the pair `construct()` function and adds correct types.

### DIFF
--- a/yaml/_type/pairs.ts
+++ b/yaml/_type/pairs.ts
@@ -27,16 +27,7 @@ function resolveYamlPairs(data: unknown[][]): boolean {
 export const pairs: Type<"sequence"> = {
   tag: "tag:yaml.org,2002:pairs",
   construct(data: Record<string, unknown>[] | null): [string, unknown][] {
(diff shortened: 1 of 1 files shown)


yaml/_type/pairs.ts
```diff
@@ -25,19 +25,16 @@ function resolveYamlPairs(data: unknown[][]): boolean {
 
   return true;
 }
-function constructYamlPairs(data: string) {
+function constructYamlPairs(
+  data: Record<string, unknown>[] | null,
+): [string, unknown][] {
   if (data === null) return [];
```
iuioiua, line 31:
```suggestion
  return data !== null ? data.flatMap(Object.entries) : [];
```

timreichen: Cool reduction, though it is not immediately obvious what is going on. I think I'll add an example doc.

iuioiua (approved):
Thank you

PR #5846, opened 2024-08-27
fix(dotenv): handle single-quotes in values in `stringify()`

emersonlaurentino:
Currently, stringify has a bug to parse JSON string.

Here is an example:
```ts
input { FOO: "{'key': 'value'}" }
input_stringify FOO='{\'key\': \'value\'}'
parsed_input { FOO: "{\\" }
```

After fix
```ts
input { FOO: "{'key': 'value'}" }
input_stringify FOO="{'key': 'value'}"
parsed_input { FOO: "{'key': 'value'}" }
```

Codebase to test
```ts
const input = { FOO: "{'key': 'value'}" };
const input_stringify = dotenv.stringify(input);
const parsed_input = dotenv.parse(input_stringify);

console.log("input", input);
console.log("input_stringify", input_stringify);
console.log("parsed_input", parsed_input);
```

### DIFF
--- a/dotenv/stringify.ts
+++ b/dotenv/stringify.ts
@@ -27,7 +27,7 @@ export function stringify(object: Record<string, string>): string {
         `key starts with a '#' indicates a comment and is ignored: '${key}'`,
       );
       continue;
-    } else if (escapedValue.includes("\n")) {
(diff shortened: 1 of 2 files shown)


iuioiua (approved):
Nice. Thank you for your contribution!

PR #5850, opened 2024-08-28
fix(http): less restrictive arguments for `accepts*()` functions

kitsonk:
Currently the http negotiation methods are arbitrarily restrictive on what objects can be passed to the methods. It requires a Fetch API request object, when in fact it only requires a very specific type in the code.  Setting overly restrictive types reduces compatibility and should be avoided.

Prior to 1.0 of std/http this was a case, but was changed for some reason causing downstream regressions in dependent code. Please consider expressing minimally restrictive types that actually reflect what the code is dependent upon.

### DIFF
--- a/http/negotiation.ts
+++ b/http/negotiation.ts
@@ -13,13 +13,6 @@ import { preferredLanguages } from "./_negotiation/language.ts";
 import { preferredMediaTypes } from "./_negotiation/media_type.ts";
 
 /**
- * The part of the {@linkcode Request} interface that the methods requires,
- * which allows for the working with objects which may not be strictly `Request`
- * objects.
- */
-export type Headered = { headers: { get(name: string): string | null } };
-
-/**
  * Returns an array of media types accepted by the request, in order of
  * preference. If there are no media types supplied in the request, then any
  * media type selector will be returned.
@@ -48,7 +41,7 @@ export type Headered = { headers: { get(name: string): string | null } };
  * @param request The request to get the acceptable media types for.
  * @returns An array of acceptable media types.
  */
-export function accepts(request: Headered): string[];
+export function accepts(request: Pick<Request, "headers">): string[];
 /**
  * For a given set of media types, return the best match accepted in the
  * request. If no media type matches, then the function returns `undefined`.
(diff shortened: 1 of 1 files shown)


timreichen:
> Currently the http negotiation methods are arbitrarily restrictive on what objects can be passed to the methods. It requires a Fetch API request object, when in fact it only requires a very specific type in the code. Setting overly restrictive types reduces compatibility and should be avoided.
> 
> Prior to 1.0 of std/http this was a case, but was changed for some reason causing downstream regressions in dependent code. Please consider expressing minimally restrictive types that actually reflect what the code is dependent upon.

Great point, but shouldn't it be `Headers` then that should be passed instead of wrapping them in an object?

iuioiua:
Yeah, it'd make much more sense if these functions simply accepted `Headers`.

kitsonk:
That is a breaking API change with little ergonomic upside.

Prior to 1.0 it accepted an object with a headers property, now it is restrictive. The least disruptive change would be to revert to the behaviour prior to 1.0.

http/negotiation.ts
```diff
@@ -12,6 +12,13 @@ import { preferredEncodings } from "./_negotiation/encoding.ts";
 import { preferredLanguages } from "./_negotiation/language.ts";
 import { preferredMediaTypes } from "./_negotiation/media_type.ts";
 
+/**
+ * The part of the {@linkcode Request} interface that the methods requires,
+ * which allows for the working with objects which may not be strictly `Request`
+ * objects.
+ */
+export type Headered = { headers: { get(name: string): string | null } };
```
kt3k, line 20: @kitsonk 
A question about some details. Can this be `{ headers: Headers }` in your use case?

iuioiua: Yes, I'd prefer this be `Headers` too.

kt3k: It feels a bit undesirable to expose this type as a public API. If it's ok to be `{ headers: Headers }`, maybe we can use `Pick<Request, "headers">` type instead of `Headered` for function parameter types

iuioiua (approved):
While I think these functions should've accepted `headers: Headers` from the start, it's true that we can't make such changes post-1.0, for the sake of stability. I've changed this to go with Yoshiya's suggestion of using `Pick<Request, "headers">`.

Either way, thank you, Kitson.

timreichen:
I think we also could do
```ts
export function acceptsLanguages(headers: Request | Headers, ...langs: string[]): string | string[] | undefined {
  headers = headers instanceof Request ? headers.headers : headers
  ...
}
```
to allow `Headers` but not break the code.

iuioiua:
> I think we also could do
> 
> ```ts
> export function acceptsLanguages(headers: Request | Headers, ...langs: string[]): string | string[] | undefined {
>   headers = headers instanceof Request ? headers.headers : headers
>   ...
> }
> ```
> 
> to allow `Headers` but not break the code.

I think the path of least action is what this PR currently does. It's ok how it is now.

Issue #5674, opened 2024-08-12
feature: add structured data in @str/collections


PandaWorker:
In programming, different data structures are very often needed and it would be good to add them to the package.

Some of the most popular are: Queue, Deque of the main ones based on the Linked List and more (Tree, Matrix, )

Very often I see how developers use arrays for queues, which is not correct. 

Therefore, I suggest adding all possible popular data structures to the package at once.

What do you think about this?

PandaWorker:
I made a mistake with the package, I need to add a single/double linked list (Queue, Deque) to @std/data-structures

jsejcksn:
I think these classic structures would be a useful addition if they are implemented using linked nodes. Such implementations would likely offer performance improvements over arrays[^1], but benchmarks of code in a PR will verify that.

[^1]: In some precursory benchmarks using implementation code from past programming puzzles, I saw performance improvements for stack `push`/`pop` while using a singly-linked node list implementation when compared with a native JavaScript array.

jsejcksn:
Some references from Wikipedia:
- [Linked list](https://en.wikipedia.org/wiki/Linked_list)
- [Doubly linked list](https://en.wikipedia.org/wiki/Doubly_linked_list)
- [Stack](https://en.wikipedia.org/wiki/Stack_(abstract_data_type))
- [Queue](https://en.wikipedia.org/wiki/Queue_(abstract_data_type))
- [Double-ended queue](https://en.wikipedia.org/wiki/Double-ended_queue)

…and some Rust structures for reference in consideration of API naming:
- [LinkedList in std::collections](https://doc.rust-lang.org/std/collections/struct.LinkedList.html)
- [VecDeque in std::collections](https://doc.rust-lang.org/std/collections/struct.VecDeque.html)

PR #5851, opened 2024-08-28
feat(data-structures/unstable): add `DoublyLinkedList`

babiabeo:
Towards: #5674

### DIFF
--- a/data_structures/linked_list.ts
+++ b/data_structures/linked_list.ts
@@ -70,7 +70,20 @@ export class LinkedList<T> implements Iterable<T> {
     this.#tail = null;
   }
 
-  /** The number of nodes in the linked list. */
+  /**
+   * The number of nodes in the linked list.
+   *
+   * @returns The number of nodes.
+   *
+   * @example Usage
+   * ```ts
+   * import { LinkedList } from "@std/data-structures";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const list = LinkedList.from([1, 2, 3, 4]);
+   * assertEquals(list.length, 4);
+   * ```
+   */
   get length(): number {
     return this.#length;
   }
@@ -89,6 +102,8 @@ export class LinkedList<T> implements Iterable<T> {
    * const list = LinkedList.from([1, 2, 3]);
    * assertEquals([...list], [1, 2, 3]);
    * ```
+   *
+   * @typeparam T The type of the values being stored in the list.
    */
   static from<T>(
     iterable: ArrayLike<T> | Iterable<T> | LinkedList<T>,
@@ -140,7 +155,7 @@ export class LinkedList<T> implements Iterable<T> {
    *
    * The complexity of this operation is O(1).
    *
-   * @returns The removed node's value, `undefined` if this list is empty.
+   * @returns The removed node's value, `undefined` if the list is empty.
    *
    * @example Usage
    * ```ts
@@ -216,7 +231,7 @@ export class LinkedList<T> implements Iterable<T> {
    *
    * The complexity of this operation is O(1).
    *
-   * @returns The removed node's value, `undefined` if this list is empty.
+   * @returns The removed node's value, `undefined` if the list is empty.
    *
    * @example Usage
    * ```ts
@@ -252,7 +267,7 @@ export class LinkedList<T> implements Iterable<T> {
   }
 
   /**
-   * Inserts the given value to the specified index.
+   * Inserts the given value to the given index.
    *
    * The complexity of this operation is linear O(n).
    *
--- a/data_structures/linked_list.ts
+++ b/data_structures/linked_list.ts
@@ -12,7 +12,7 @@ export interface LinkedNode<T> {
 }
 
 /**
- * A doublely linked list is a linked data structures that consists
+ * A doubly linked list is a linked data structures that consists
  * of a set of nodes. Each node contains a value and two link fields,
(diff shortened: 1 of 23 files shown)


data_structures/linked_list.ts
```diff
@@ -0,0 +1,475 @@
...
+   *
+   * list.push(1);
+   * list.push(2);
+   * list.push(3);
+   *
+   * assertEquals(list.length, 3);
+   * assertEquals([...list], [1, 2, 3]);
+   * ```
+   */
+  push(value: T): void {
```
timreichen, line 140: Could we make `push(...values: T[])`?

data_structures/linked_list.ts
```diff
@@ -0,0 +1,475 @@
...
+   *
+   * list.unshift(1);
+   * list.unshift(2);
+   * list.unshift(3);
+   *
+   * assertEquals(list.length, 3);
+   * assertEquals([...list], [3, 2, 1]);
+   * ```
+   */
+  unshift(value: T): void {
```
timreichen, line 216: Could we make `unshift(...values: T[])`?

data_structures/linked_list.ts
```diff
@@ -0,0 +1,475 @@
...
+   * import { LinkedList } from "@std/data-structures";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const list = LinkedList.from([2, 4, 6, 8]);
+   *
+   * assertEquals(list.has(4), true);
+   * assertEquals(list.has(0), false);
+   * ```
+   */
+  has(value: T): boolean {
```
timreichen, line 422: If we want to follow the array convention this probably should be called `includes(value: T)`

data_structures/linked_list.ts
```diff
@@ -0,0 +1,475 @@
...
+   * assertEquals([...list], [2, 4, 6, 8]);
+   *
+   * list.remove(1);
+   * assertEquals([...list], [2, 6, 8]);
+   *
+   * list.remove(2);
+   * assertEquals([...list], [2, 6]);
+   * ```
+   */
+  remove(index: number): T | undefined {
```
timreichen, line 365: If we want to follow the array convention this probably should be called `splice`

data_structures/linked_list.ts
```diff
@@ -0,0 +1,475 @@
...
+   * assertEquals([...list], [1, 2, 3, 4, 6, 8]);
+   *
+   * list.insert(5, 4);
+   * assertEquals([...list], [1, 2, 3, 4, 5, 6, 8]);
+   *
+   * list.insert(7, 6);
+   * assertEquals([...list], [1, 2, 3, 4, 5, 6, 7, 8]);
+   * ```
+   */
+  insert(value: T, index: number): void {
```
timreichen, line 299: If we want to follow the array convention this probably should be combined with `remove` to `splice`

data_structures/linked_list.ts
```diff
@@ -0,0 +1,490 @@
...
+ *
+ * assertEquals([...list], ["x", "z", "a", "b", "d"]);
+ * assertEquals(list.remove(2), "a");
+ * assertEquals(list.has("d"), true);
+ * assertEquals(list.has("a"), false);
+ * ```
+ *
+ * @typeparam T The type of the values being stored in the list.
+ */
+export class LinkedList<T> implements Iterable<T> {
```
kt3k, line 59: Why not `DoublyLinkedList`?

data_structures/linked_list.ts
babiabeo: I'm not sure about this because some npm packages implementing linked lists also use `remove` and `insert` instead of `splice`.

timreichen: True, but [this](https://www.npmjs.com/package/yallist) implements `splice()` and has like 125mil d/w.
I just think the implementation should be as web-api-ish as possible.

babiabeo: I think it's better to keep `insert` and `remove` instead of combining them to `splice` because `splice` seems non-specific compared to its use and some people may not be familiar with it

timreichen: Sorry for my persistence, I just think we should avoid mixing thermologies from different web apis and deviate from established names. In that case, we probably should orient this implementation not on `array` but on `FormData`:

- rename `remove()` to `delete()`
- rename `includes()` to `has()` (reverse prior change)
- rename `push()` to `append()` (reverse prior change of accepting multiple values)
- rename `unshift()` to `prepend()` (reverse prior change of accepting multiple values)
- remove `shift()` (use `delete(0)`)
- rename `pop()` to `deleteLast()`
- rename `length` to `size` 

**Reasoning**
`prepend()` does not exist in `FormData`, but is derived from `append()` 
`deleteLast` does not exist in any web api, but can be derived as a combination of `delete()` and `findLast()`

WDYT?

babiabeo: Based on your suggestion, here is my opinion:
- `push()` -> `append()`
- `unshift()` -> `prepend()`
- `pop()` -> `removeLast()`
- `shift()` -> `removeFirst()`
- `length` -> `size`
- `includes` -> `has` or `contains`

I also intend to add 2 versions of `remove()`, one for removing element based on the given index and another for removing element based on the given value.

Perhaps use `delete` and `remove` for their names? (This doesn't seems to be a good idea)

WDYT?

babiabeo:
Closed. Need more discussion about methods’ names

Issue #5854, opened 2024-08-28
Inconsistency between `throw` and `throw new`


irbull:
**Describe the bug**

In most cases, the Standard Library uses `throw new` but in some cases it directly throws the object. According to [1], it doesn't matter, but consistency in the codebase is good.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error

irbull:
[throw_no_new.txt](https://github.com/user-attachments/files/16785173/throw_no_new.txt)

Here is a quick and dirty analysis of where `throw` doesn't include `new`. Many of these are re-throws, so they don't need it.

iuioiua:
Just checking that the search parameters are correct here. I think we should be searching for the regex `throw (?!new)`. This should also include cases such as `throw SyntaxError()`, etc.

irbull:
I actually did the following:

`find . -type f -not -path "./coverage/*" -exec grep "throw" -H '{}' \; > ~/thrown.txt`

If you just use `grep`, I had problems with some of the test files that include a named pipe and grep just got hungup.

I then removed all the lines with `throw new`:
`grep -v "throw new" thrown.txt > throw_no_new.txt`

From here I scanned through the list and removed `.md`, `.mjs` `/testdata/`. It left a few hundred that didn't take too long to filter. There were a few common patters:
- One that were in comments. This was easy to spot because the line starts with a `*`
- `return this.throwError(`
- Text like "should not throw error"
- `_test.ts` files

I'm pretty quick with multi-cursors in VSCode, so I just filtered those common cases by hand. It left me with things you'd expect `throw Error` and stuff like `throw err`. There were maybe 10 files, and I just went through them and checked that the `throw err` was really just a re-throw and not some variable that was constructed earlier. If we assume that all Errors start with a capital, then `grep "throw [A-Z]"` now produces:

```
./crypto/_wasm/lib/deno_std_wasm_crypto.generated.mjs:      throw Error("TextDecoder not available");
./crypto/_wasm/lib/deno_std_wasm_crypto.generated.mjs:      throw Error("TextEncoder not available");
./Releases.md:- fix(encoding/base64url): throw TypeError in case of invalid base64url string
./fs/move.ts:      throw EXISTS_ERROR;
```
The first 2 are in a generated file. 3rd one is in a README, and the last one throws a constant which is defined as:

```typescript
const EXISTS_ERROR = new Deno.errors.AlreadyExists("dest already exists.");
[...]

PR #5855, opened 2024-08-28
refactor(assert,datetime,fmt,io,streams): consistently `throw new Error()` instead of `throw Error()`

irbull:
…nstead of `throw`

To ensure consistency across the codebase, this commit refactors the codebase to use `throw new Error` instead of `throw` for throwing errors.

https://github.com/denoland/std/issues/5854


iuioiua (approved):
Great! Thank you.

iuioiua:
Does this close #5854?

PR #5860, opened 2024-08-28
refactor(yaml): simplify regexp type

timreichen:


iuioiua (changes requested):
Any chance you'd be able to first write tests for this code? It's currently not covered.

iuioiua (approved):
LGTM now that we have tests. Thank you.

PR #5863, opened 2024-08-29
feat(path/unstable): support `URL` in first arg of `join()`

kt3k:
part of #5537

This PR adds support of URL as the first parameter of `join`

### DIFF
--- a/path/join.ts
+++ b/path/join.ts
@@ -27,6 +27,8 @@ export function join(...paths: string[]): string;
 /**
  * Join all given a sequence of `paths`,then normalizes the resulting path.
  *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
  * @example Usage
  * ```ts
  * import { join } from "@std/path/posix/join";
--- a/path/join.ts
+++ b/path/join.ts
@@ -23,6 +23,24 @@ import { join as windowsJoin } from "./windows/join.ts";
  * @param paths Paths to be joined and normalized.
  * @returns The joined and normalized path.
  */
-export function join(...paths: string[]): string {
-  return isWindows ? windowsJoin(...paths) : posixJoin(...paths);
+export function join(...paths: string[]): string;
+/**
+ * Join all given a sequence of `paths`,then normalizes the resulting path.
+ *
+ * @example Usage
+ * ```ts
+ * import { join } from "@std/path/posix/join";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const path = join(new URL("file:///foo"), "bar", "baz/asdf", "quux", "..");
+ * assertEquals(path, "/foo/bar/baz/asdf");
+ * ```
+ *
+ * @param path The path to join. This can be string or file URL.
+ * @param paths The paths to join.
+ * @returns The joined path.
+ */
+export function join(path?: URL | string, ...paths: string[]): string;
+export function join(path?: URL | string, ...paths: string[]): string {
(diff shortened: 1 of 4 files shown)


path/join.ts
```diff
@@ -23,6 +23,26 @@ import { join as windowsJoin } from "./windows/join.ts";
  * @param paths Paths to be joined and normalized.
  * @returns The joined and normalized path.
  */
-export function join(...paths: string[]): string {
-  return isWindows ? windowsJoin(...paths) : posixJoin(...paths);
+export function join(...paths: string[]): string;
+/**
+ * Join all given a sequence of `paths`,then normalizes the resulting path.
```
iuioiua, line 28:
```suggestion
 * Join all given a sequence of `paths`, then normalizes the resulting path.
```

path/join_test.ts
```diff
@@ -56,10 +58,14 @@ const joinTests =
     [["/", "", "/foo"], "/foo"],
     [["", "/", "foo"], "/foo"],
     [["", "/", "/foo"], "/foo"],
+
+    // URLs
+    [[new URL("file:///"), "x/b", "..", "/b/c.js"], "/x/b/c.js"],
+    [[new URL("file:///foo"), "../../../bar"], "/bar"],
```
iuioiua, line 64: The first test case looks good. But I'd expect the result of the 2nd test case to be `"/../../bar"`.

path/posix/join.ts
```diff
@@ -16,24 +17,33 @@ import { normalize } from "./normalize.ts";
  * assertEquals(path, "/foo/bar/baz/asdf");
  * ```
  *
- * @example Working with URLs
+ * @param paths The paths to join.
+ * @returns The joined path.
+ */
+export function join(...paths: string[]): string;
+/**
+ * Join all given a sequence of `paths`,then normalizes the resulting path.
```
iuioiua, line 25:
```suggestion
 * Join all given a sequence of `paths`, then normalizes the resulting path.
```

path/windows/join.ts
```diff
@@ -20,7 +21,29 @@ import { normalize } from "./normalize.ts";
  * @param paths The paths to join.
  * @returns The joined path.
  */
-export function join(...paths: string[]): string {
+export function join(...paths: string[]): string;
+/**
+ * Join all given a sequence of `paths`,then normalizes the resulting path.
```
iuioiua, line 26:
```suggestion
 * Join all given a sequence of `paths`, then normalizes the resulting path.
```

path/join_test.ts
iuioiua, line 1: Current tests seem a little bare. Could we please add the tests from the examples?

iuioiua (changes requested):

path/join_test.ts
kt3k: `/bar` is the normalized form of `/../../bar`. There's no parent directory of the root directory

iuioiua: Oh, I tested with `node:path` and that does appear to be the case. That doesn't really make sense to me but I guess is fine 🤷🏾‍♂️

iuioiua (approved):

path/join_test.ts
kt3k: How did you test it? The below prints `/bar` both in Node.js and Deno

```js
import { join } from "node:path";
console.log(join("/foo", "../../../bar"));
```

BTW `join("file:///foo", "../../../bar")` becomes `../bar` because `"file:///foo"` is considered relative path `file:/bar` (`file:` part is considered as directory name instead of scheme name)

PR #5872, opened 2024-08-30
fix(cache/unstable): fix flaky fibonacci test

lionel-rowe:
Fixes this: https://github.com/denoland/std/actions/runs/10625661936/job/29455981411?pr=5626

```
 ERRORS 

memoize() is performant for expensive fibonacci function => ./cache/memoize_test.ts:65:6
error: AssertionError: Expected actual: "1724984137590" to be close to "1724984137573": delta "17" is greater than "10".
  throw new AssertionError(
        ^
    at assertAlmostEquals (file:///D:/a/std/std/assert/almost_equals.ts:51:9)
    at file:///D:/a/std/std/cache/memoize_test.ts:73:3

 FAILURES 
```

### DIFF
--- a/cache/memoize_test.ts
+++ b/cache/memoize_test.ts
@@ -3,6 +3,7 @@ import {
   assert,
   assertAlmostEquals,
   assertEquals,
+  assertLess,
   assertRejects,
 } from "@std/assert";
 import { memoize } from "./memoize.ts";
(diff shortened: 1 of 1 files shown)


lionel-rowe:
Something weird going on here, attempt 1 at a fix was apparently timing out after a full 10 seconds (successful fix logged <=2ms to complete in each environment). Are functions like `performance.now`, `Date.now` etc now being globally overridden in CI or something?

https://github.com/denoland/std/actions/runs/10625999673/job/29456871384

iuioiua (approved):
LGTM. @kt3k, any idea what could've caused the weird behavior as mentioned in https://github.com/denoland/std/pull/5872#issuecomment-2319837139?

kt3k:
I don't think we override `performance.now` or `Date.now` in CI.

`FakeTime` overrides Date class, but that shouldn't affect this test case (Test cases in a single file should be executed sequentially even with `--parallel` flag)

So I have no idea what caused the above error

PR #5873, opened 2024-08-30
fix(yaml): fix `sortKey` error message

timreichen:
**Changes**
This PR fixes the `sortKey` error message with template string and adds a test.

⚠️ **Note**
Instead of throwing an error, we might want to remove this type check completely, because it should be covered by typescript.
```ts
if (typeof this.sortKeys === "function") {
  objectKeyList.sort(this.sortKeys);
} else if (this.sortKeys) {
  objectKeyList.sort();
}
```
WDYT?


iuioiua (approved):
Wow - what a funny/strange bug. Thanks! We're currently in an undecided state regarding runtime type checking of values, so let's keep the current behavior, even though I think it should go.

PR #5874, opened 2024-08-30
test(yaml): add invalid `represent` type test

timreichen:
**Changes**
This PR adds test for invalid represent type value.
It also changes the function code to throw early.


iuioiua (approved):
Nice! Thank you.

PR #5876, opened 2024-08-30
fix(yaml): fix Boolean object handling

### LINKED ISSUE #5857: `yaml` bool type inconsistent boolean instance evaluation
**Describe the bug**
In `yaml/_types/bool.ts`, booleans are evaluated as
```ts
typeof value === "boolean" || value instanceof Boolean
```

This means `Boolean` instances are handled as booleans in yaml but the representation is evaluated as
```ts
represent: {
    lowercase(object: boolean): string {
      return object ? "true" : "false";
    },
    uppercase(object: boolean): string {
      return object ? "TRUE" : "FALSE";
    },
    camelcase(object: boolean): string {
      return object ? "True" : "False";
    },
}
```
This will always stringify `Boolean` instances to `"true"`, which  is probably not the intended behavior.

**Steps to Reproduce**
```ts
import { stringify } from "@std/yaml";
assertEquals(stringify(new Boolean(false)), "true\n");
```

**Expected behavior**
`new Boolean(false)` should evaluate to `"false"` **OR**  `Boolean` instances should not be special treated and `value instanceof Boolean`  removed so they are handled as objects.

I am leaning torwards the removal. WDYT?

timreichen:
Closes #5857


iuioiua (approved):
@kt3k, PTAL at the related issue. Do you agree this is a fix? I think it is as it fixes expected behavior.

kt3k:
This changes the result of `stringify({ foo: new Boolean(true) })` to `foo: {}` instead of `foo: true`. I don't think this matches the expectation of the users.

I think we should fix `represent` handler of bool type instead. I guess something like `Boolean(object) ? "true" : "false"` should fix the issue.

timreichen:
> This changes the result of `stringify({ foo: new Boolean(true) })` to `foo: {}` instead of `foo: true`. I don't think this matches the expectation of the users.
> 
> I think we should fix `represent` handler of bool type instead. I guess something like `Boolean(object) ? "true" : "false"` should fix the issue.

The problem is that this special treats the `Boolean` object instance. We have the same problem with
```ts
stringify({ foo: new String("foo") })
```
which returns
```sh
foo:
  '0': f
  '1': o
  '2': o
  ```
and 
```ts
stringify({ foo: new Number(1) })
```
which returns 
```ts
foo: {}
```

timreichen:
Tbh, I don't know if we should special treat these cases as they should be discouraged anyway.

kt3k:
> The problem is that this special treats the `Boolean` object instance. We have the same problem with
> 
> ```ts
> stringify({ foo: new String("foo") })
> ```
> 
> which returns
> 
> ```shell
> foo:
>   '0': f
>   '1': o
>   '2': o
> ```
> 
> and
> 
> ```ts
> stringify({ foo: new Number(1) })
> ```
> 
> which returns
> 
> ```ts
> foo: {}
> ```

Those examples seem bugs to me.

`std/yaml` is originally ported from `js-yaml` and `js-yaml` handles those cases as `foo: foo` and `foo: 1`.

```js
import * as yaml from "npm:js-yaml";

console.log(yaml.dump({ foo: new String("foo") }));
console.log(yaml.dump({ foo: new Number(1) }));
```

kt3k:
Looks like those handlings of `new String` and `new Number` are regression in recent versions. It used to be handled as string and number:

```js
import { stringify } from "https://deno.land/std@0.224.0/yaml/mod.ts";

console.log(stringify({ foo: new String("foo") }));
console.log(stringify({ foo: new Number(1) }));
```

This prints:
```
foo: foo
foo: 1
```

timreichen:
Closing in favour of https://github.com/denoland/std/pull/5894

Issue #5830, opened 2024-08-26
`text/slugify` gives empty results for non-Latin alphabets


lionel-rowe:
**Describe the bug**

`slugify` gives readable results for only a subset of languages, namely those that use the Latin alphabet (English, Spanish, Vietnamese...). For all other languages (Chinese, Arabic, Russian...), it just returns an empty string.

**Steps to Reproduce**

```ts
slugify('三人行，必有我师焉') // returns ""
````

**Expected behavior**

`slugify` to return non-Latin-alphabet words unchanged (but still removing diacritics, stripping start/end punctuation, and replacing medial punctuation with dashes).

Something like the following:

```ts
function intlFriendlySlugify(str: string) {
    return str
        .toLowerCase()
        .normalize("NFD")
        .replaceAll(/\p{M}/gu, "")
        .split(/[^\p{L}\p{N}]+/u)
        .filter((x) => x !== '')
        .join('-')
}
```

IMO the `slugify` function should _not_ handle percent encoding, instead leaving that up to the calling code (e.g. passing to `URL` constructor or `URLSearchParams`):

```ts
const slug = intlFriendlySlugify('三人行，必有我师焉') // "三人行-必有我师焉"
const origin = 'https://example.com'
const url1 = new URL(origin)
url1.pathname = join('blog', slug)
const url2 = new URL(origin)
url2.searchParams.set('q', slug)
const url3 = new URL(origin)
url3.hostname = `${slug}.example.com`
```

* `url1`: https://example.com/blog/%E4%B8%89%E4%BA%BA%E8%A1%8C-%E5%BF%85%E6%9C%89%E6%88%91%E5%B8%88%E7%84%89
* `url2`: 
https://example.com/?q=%E4%B8%89%E4%BA%BA%E8%A1%8C-%E5%BF%85%E6%9C%89%E6%88%91%E5%B8%88%E7%84%89.
* `url3`: https://xn----io6a1vz30ct6b22enxlhzue11c.example.com/

While it's true they're unreadable in the percent-encoded/punycode form, modern browsers display the human-readable form automatically on hover, or in the address bar when you open them. At some point it might also be worth adding a `prettifyUrl` function to allow similar conversion in userland, but I'm leaving that out of scope for this issue.

**Environment**

OS: Ubuntu 20.04, WSL
deno version: 1.46.0
std version: text@1.0.4

Wanderx13:
Download 
[...]

PR #5880, opened 2024-08-31
fix(text/unstable): handle non-Latin-script text in `slugify`

lionel-rowe:
Fixes https://github.com/denoland/std/issues/5830

Changed tests:
* "déjà vu" is now `déjà-vu`, not `deja-vu` (and same for other diacritic tests). This was the most common handling of Latin-text diacritics in the sites I checked in https://github.com/denoland/std/issues/5830#issuecomment-2316741157.
* "$33,000" is now `33-000`, not `33000`. The Unicode category of `,` is "Punctuation (Other)", which is very broad and IMO should generally replaced with `"-"` rather than `""`. I don't think it's probably desirable to special-case individual characters beyond Unicode category as it just ends up with a load of hard-coded and largely arbitrary chars (with the exception of straight quote marks, which are sort-of in the "wrong" category as quote marks typically fall in "Punctuation (initial)" or "Punctuation (final)"). Also, while there isn't much to choose between `33-000` vs `33000` (`33000` is more natural, but `33-000` more readable due to retaining the thousands-place delimitation) `,` is used as a decimal point in many languages, and `la-valeur-de-pi-est-3-14` is clearly preferable to `la-valeur-de-pi-est-314`.


timreichen:
See https://github.com/denoland/std/issues/5830#issuecomment-2322802653
I am against this change, because this use case is better handled with `@std/text/to-kebab-case`
```ts
import { toKebabCase } from "@std/text/to-kebab-case";
console.log(toKebabCase("déjà vu")); // "déjà-vu"
console.log(toKebabCase("$33,000")); // "33-000"
```

iuioiua:
@lionel-rowe, I think we should first come to an agreement on the solution in #5830 before opening a PR. Are you alright to close this for now?

PR #5881, opened 2024-08-31
test(yaml): add positive timezone timestamp test

timreichen:
**Note** this tests doesn't cover more lines, it prevents unintentional breaks in future PRs.


kt3k (approved):
LGTM

PR #5882, opened 2024-08-31
refactor(yaml): add `isEof()` to `LoaderState`

timreichen:
**Changes**
This PR replaces char === 0 checks with `isEof()`.


yaml/_loader_state.ts
```diff
@@ -529,15 +534,15 @@ export class LoaderState {
     let lineBreaks = 0;
     let ch = this.peek();
 
-    while (ch !== 0) {
+    while (!this.#isEof()) {
```
iuioiua, line 537: So now we're getting `this.peek()` twice. This appears to be the case for the other changed lines also.

iuioiua (commented):
Idk, this change seems kinda unnecessary. I liked the verbosity of the previous code.

iuioiua:
I'm going to close this. The current code is fine and uses an already defined variable, `ch`, in a more verbose way. Thanks for the PR, either way.

PR #5885, opened 2024-09-01
refactor(yaml): simplify `stringifyBlockMapping()`

timreichen:
**Changes**
This PR simplifies the `stringifyBlockMapping()` function by using an array and joining it at the end like in `stringifyBlockSequence()` and moves variable assignments and function calls down so they will not be executed if the loop continues early.


iuioiua (approved):
Thanks

PR #5887, opened 2024-09-01
feat(uuid/unstable): `@std/uuid/v7`

Liam-Tait:
Adds [UUID Version 7](https://www.rfc-editor.org/rfc/rfc9562.html#name-uuid-version-7) support.

This variant provides  solution for scenarios requiring both uniqueness and chronological ordering.

The same pattern as other uuid generation has been used.
```ts
import { v7 } from "@std/uuid";
import { assert } from "@std/assert";
const uuid = v7.generate();
// or
const uuid2 = v7.generate({ timestamp: 1000 }));
```

I have not included the optional features for sub-millisecond timestamp fraction or  carefully seeded counter


uuid/v7.ts
```diff
@@ -0,0 +1,99 @@
...
+  // Unix timestamp in milliseconds (truncated to 48 bits)
+  const timestamp = BigInt(options.timestamp ?? Date.now());
+  view.setUint32(0, Number(timestamp >> 16n) & 0xffffffff);
+  view.setUint16(4, Number(timestamp & 0xffffn));
+  // Version (4 bits) & Random (12 bits)
+  const randA = options.random?.a ?? crypto.getRandomValues(new Uint8Array(2));
+  view.setUint8(6, randA[0]! & 0x0f | version << 4);
+  view.setUint8(7, randA[1]!);
+  // Variant (2 bits) & Random (62 bits)
+  const randB = options.random?.b ?? crypto.getRandomValues(new Uint8Array(8));
```
kt3k, line 89: Can these 2 `crypto.getRandomValues` calls be a single call for `new Uint8Array(10)`? (I think that is preferable for performance). Or is this required by the spec?

Liam-Tait: I think is fine based on the spec.

From [Section 5.7](https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7)
> UUIDv7 values are created by allocating a Unix timestamp in milliseconds in the most significant 48 bits and filling the remaining 74 bits, excluding the required version and variant bits, with random bits for each new UUIDv7 generated to provide uniqueness as per [Section 6.9](https://www.rfc-editor.org/rfc/rfc9562.html#unguessability)

Consolidating to a single `Uint8Array` value `options.random` adds a bit more convention to, but I think that is ok, as that option should only be used for testing.

Another option is to initially fill `bytes` with random data and then write timestamp, ver, and variant `crypto.getRandomValues(bytes)`.
That might simplify the code, I do not know the performance difference though.

kt3k: > Another option is to initially fill bytes with random data and then write timestamp, version, and variant crypto.getRandomValues(bytes).
That might simplify the code, I do not know the performance difference though.

I think that works too, and I think that is better than 2 `getRandomValues` calls

Liam-Tait: I will make the change tonight

babiabeo:
Should we consider supporting [monotonicity](https://www.rfc-editor.org/rfc/rfc9562.html#name-monotonicity-and-counters)?

I have also created a [UUID v7](https://github.com/babiabeo/uuid-v7) package, which you may find useful as a reference 👍

uuid/v7.ts
```diff
@@ -0,0 +1,95 @@
...
+  view.setBigUint64(0, timestamp << 16n);
+  if (options.random) {
+    bytes.set(options.random, 6);
+  } else {
+    crypto.getRandomValues(bytes.subarray(6));
+  }
+  // Version (4 bits) Occupies bits 48 through 51 of octet 6.
+  view.setUint8(6, (view.getUint8(6) & 0b00001111) | (version << 4));
+  // Variant (2 bits) Occupies bits 64 through 65 of octet 8.
+  view.setUint8(8, (view.getUint8(8) & 0b00111111) | (variant << 6));
```
babiabeo, line 93:
```suggestion
  view.setUint8(6, (view.getUint8(6) & 0b00001111) | 0b11110000);
  // Variant (2 bits) Occupies bits 64 through 65 of octet 8.
  view.setUint8(8, (view.getUint8(8) & 0b00111111) | 0b10000000);
```

uuid/v7.ts
```diff
@@ -0,0 +1,95 @@
...
+  // Unix timestamp in milliseconds (truncated to 48 bits)
+  if (options.timestamp !== undefined) {
+    if (!Number.isInteger(options.timestamp) || options.timestamp < 0) {
+      throw new Error(
+        `Cannot generate UUID as timestamp must be a non-negative integer: timestamp ${options.timestamp}`,
+      );
+    }
+  }
+  const timestamp = BigInt(options.timestamp ?? Date.now());
+  view.setBigUint64(0, timestamp << 16n);
```
iuioiua, line 84: Why must we deal with a `BigInt` here? Genuine question.

iuioiua (changes requested):
Cleanly done so far! Nice. Few concerns.

uuid/v7.ts
```diff
@@ -0,0 +1,95 @@
...
+   * @default {Date.now()}
+   */
+  timestamp?: number;
+  /**
+   * 10 bytes of random value to use in the UUID.
+   * Generally you should not need to set these, but it can be useful for testing.
+   *
+   * @internal
+   */
+  random?: Uint8Array;
```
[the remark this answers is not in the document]
Liam-Tait: https://github.com/denoland/std/blob/8abff456f81503f23ef2dd859e37659538990df4/uuid/v1.ts#L65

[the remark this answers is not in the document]
iuioiua: Oh... Nice find. We probably shouldn't be doing that 🤣

uuid/v7.ts
Liam-Tait: Convenience only, `BigInt` is not required.

Alternatives could be something along the lines of:
```ts
const shiftedTimestamp = (timestamp & 0xFFFFFFFFFFFF) << 16;
view.setUint32(0, shiftedTimestamp / 0x100000000); // High 32 bits
view.setUint32(2, shiftedTimestamp & 0xFFFFFFFF); // Low 32 bits
```
or
```
data[0] = (timestamp / 0x10000000000) & 0xff;
data[1] = (timestamp / 0x100000000) & 0xff;
data[2] = (timestamp / 0x1000000) & 0xff;
data[3] = (timestamp / 0x10000) & 0xff;
data[4] = (timestamp / 0x100) & 0xff;
data[5] = timestamp & 0xff;
```

Would need test the difference in performance

iuioiua: Ok. Let's keep as is then.

iuioiua: To clarify, the simplicity of what we have now looks good. But if there's a sufficient performance difference, then we should re-consider.

Liam-Tait: Thanks for the clarification. I will do some naive local testing tonight to check if performance is significantly different and report back.

uuid/v7.ts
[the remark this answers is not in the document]
Liam-Tait: We could remove the timestamp option too and test through fake time there too. 
Though it feels as though it has utility, not for me, but for someone.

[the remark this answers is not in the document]
iuioiua: The timestamp option is useful and worth keeping as one would not always want to generate a UUIDv7 for now. They may want to generate them for the past or future. In fact, we should also have a way for users to decode the timestamp for such cases too.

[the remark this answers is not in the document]
Liam-Tait: > have a way for users to decode the timestamp

That makes sense. Usage something like:

```typescript
import { generate, extractTimestamp } from "@std/uuid/v7";

const uuid: string = generate()
const timestamp: number = extractTimestamp(uuid)
```

Will add.

uuid/v7.ts
Liam-Tait: I wrote naive benchmarks on [this branch](https://github.com/Liam-Tait/std/tree/liam/uuid-v7-view-array-benchmark)

[bench](https://github.com/Liam-Tait/std/blob/34e42926b11bf931be2c334aa7f0d6f9b5284b47/uuid/v7_bench.ts)
[changes](https://github.com/Liam-Tait/std/blob/74edfcc02affcb7d5a2a709324fd3daa2e0e0672/uuid/v7.ts)

I am not getting results that gives me confidence either way. Sometimes faster, sometimes slower. Let me know if you get something more consistent.
Happy to go one way or the other.

uuid/v7.ts
```diff
@@ -0,0 +1,95 @@
...
+ * const uuid = generate();
+ * assert(validate(uuid));
+ * ```
+ */
+export function generate(options: GenerateOptions = {}): string {
+  const bytes = new Uint8Array(16);
+  const view = new DataView(bytes.buffer);
+  // Unix timestamp in milliseconds (truncated to 48 bits)
+  if (options.timestamp !== undefined) {
+    if (!Number.isInteger(options.timestamp) || options.timestamp < 0) {
```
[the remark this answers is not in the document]
Liam-Tait: [Done](https://github.com/denoland/std/pull/5887/commits/ae96347076a906240491829af1e24430906a9195)

uuid/v7.ts
Liam-Tait: [Done](https://github.com/denoland/std/pull/5887/commits/9d2620b06d3f4f5bb6b82655e424f05f986e226b)

Liam-Tait:
> Should we consider supporting [monotonicity](https://www.rfc-editor.org/rfc/rfc9562.html#name-monotonicity-and-counters)?

I intentionally left out the optional features, these could be added in a seperate pr as an option, for this pr I would like to start with the basic implementation

> implementations MAY fill the 74 bits, jointly, with a combination of the following subfields, in this order from the most significant bits to the least, to guarantee additional monotonicity within a millisecond:
>
> - An OPTIONAL sub-millisecond timestamp fraction (12 bits at maximum) as per [Section 6.2](https://www.rfc-editor.org/rfc/rfc9562.html#monotonicity_counters) (Method 3).
> - An OPTIONAL carefully seeded counter as per [Section 6.2](https://www.rfc-editor.org/rfc/rfc9562.html#monotonicity_counters) (Method 1 or 2).
> - Random data for each new UUIDv7 generated for any remaining space.

[link](https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7-2)

uuid/v7.ts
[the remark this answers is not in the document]
Liam-Tait: Added [here](https://github.com/denoland/std/pull/5887/commits/53b320c9daaec1433680150f4d5156cc4c1dcecd)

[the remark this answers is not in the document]
Liam-Tait: Removed `random` option and mocked `crypto.getRandomValues` [here](https://github.com/denoland/std/pull/5887/commits/79b7e63ac2f8f1d1ccfc5020cb59d9a4822abc1b).

uuid/v7.ts
iuioiua: Ok. Let's stick with what we have now then.

uuid/v7.ts
```diff
@@ -0,0 +1,115 @@
...
+ * ```ts
+ * import { extractTimestamp } from "@std/uuid/v7";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const uuid = "017f22e2-79b0-7cc3-98c4-dc0c0c07398f";
+ * const timestamp = extractTimestamp(uuid);
+ * assertEquals(timestamp, 1645557742000);
+ * ```
+ */
+export function extractTimestamp(uuid: string): number {
```
iuioiua, line 107: @kt3k, how do you feel about this name? `@std/ulid` has `decodeTime()`.

uuid/v7.ts
babiabeo: According to my brief local benchmark, your implementation seems to be ~1.5x-2x faster than `npm:uuid`, but I'm not sure if other machines will yield the same results.

uuid/v7.ts
kt3k: This sounds fine to me

uuid/v7.ts
```diff
@@ -0,0 +1,115 @@
...
+  if (!validate(uuid)) {
+    throw new TypeError(
+      "Could not extract timestamp because the UUID was not a valid UUIDv7",
```
kt3k, line 110: We have [the style guide](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#error-messages) of the error messages. I think the below form follows the guide better:

```suggestion
      `Cannot extract timestamp because the UUID (${uuid}) is not a valid UUIDv7`,
```

Liam-Tait: How about the following?

```suggestion
     `Cannot extract timestamp because the UUID is not a valid UUIDv7: uuid is "${uuid}"`
```

Liam-Tait: Update [here](https://github.com/denoland/std/pull/5887/commits/38c23954fd64806dec09f46dec216cfef3ffc3f1)

uuid/mod.ts
```diff
@@ -106,3 +111,22 @@ export const v5 = {
   generate: generateV5,
   validate: validateV5,
 };
+
+/**
+ * Generator and validator for
+ * {@link https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7 | UUIDv7}.
+ *
```
babiabeo, line 118: nit: Could you add the `@experimental` tag to this one?

uuid/v7.ts
```diff
@@ -0,0 +1,133 @@
...
+ * @example
+ * ```ts
+ * import { generate, validate, extractTimestamp } from "@std/uuid/v7";
+ * import { assert, assertEquals } from "@std/assert";
+ *
+ * const uuid = generate();
+ * assert(validate(uuid));
+ * assertEquals(extractTimestamp("017f22e2-79b0-7cc3-98c4-dc0c0c07398f"), 1645557742000);
+ * ```
+ *
```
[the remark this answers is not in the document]
Liam-Tait: added [here](https://github.com/denoland/std/pull/5887/commits/4875b4a2e28152df4ac4700ee1e43faa97d5b3e4)

uuid/v7.ts
```diff
@@ -0,0 +1,135 @@
...
+ * @example Usage
+ * ```ts
+ * import { generate, validate } from "@std/uuid/v7";
+ * import { assert } from "@std/assert";
+ *
+ * const uuid = generate();
+ * assert(validate(uuid));
+ * ```
+ */
+export function generate(options: GenerateOptions = {}): string {
```
iuioiua, line 84: I think it's safe to do this. WDYT?
```suggestion
export function generate(timestamp = Date.now()): string {
```

iuioiua (approved):
LGTM! Only one idea for `generate()`.

uuid/v7.ts
Liam-Tait: Keeping GenerateOptions allows adding the optional features for monotonicity approach be chosen  in the future. If that is desired, if monotonicity is not an option and is always the same, having just timestamp if works.
 
e.g something like the following
```ts
generate({ monotonicity: "counter" })
generate({ nano: true, counter: true })
```
[optional features](https://www.rfc-editor.org/rfc/rfc9562.html#section-5.7-3.1)

Liam-Tait: No consistency issues either way
- v1 `generate(options: GenerateOptions = {}): string`
- v3 `generate(namespace: string, data: Uint8Array): Promise<string>`
- v4 `crypto.randomUUID()`
- v5 `generate(namespace: string, data: Uint8Array): Promise<string>`

iuioiua: It's possible we'll add more options. It's also possible we won't. This is an experimental API. If we add more options, we can introduce a `GenerateOptions`. I think let's just go with a single `timestamp = Date.now()` argument for now 🙂

Liam-Tait: Sweet, I will update to a use a single timestamp argument then

Liam-Tait: Done [here](https://github.com/denoland/std/pull/5887/commits/7dfa002d2fdd3dd9faac2ce4a9ee37d597a58d4b).

As a side effect of this change, even when the default `Date.now()` is used, the timestamp is still validated to ensure it’s a non-negative integer. 
This differs from the previous behaviour, where validation was skipped if the default `Date.now()` was used.

iuioiua (approved):
LGTM again. Again, good work!

PR #5888, opened 2024-09-01
feat(http): add `accepts` param for routes

### LINKED ISSUE #5883: feat_req(http/routes): add a accept filter

Current implementation is supporting http method discriminant probably covers most use-case, but there's also one more use-case which I feel like is pretty common is Accept header discriminant.

Sometimes, same route path may serve different content based on Accept headers (for example a nice page for users when `text/html` is present, and a json object for developers when `application/json` is passed.

Example: sending different format based on requested accept types:
```ts
const routes: Route[] = [
  {
    pattern: new URLPattern({ pathname: "/api/foo" }),
    accept: ["application/xml"]
    handler: () => new Response("<foo>bar</foo>"),
  },
  {
    pattern: new URLPattern({ pathname: "/api/foo" }),
    accept: ["application/json"]
    handler: () => new Response('{"foo":{"bar"}}'),
  }
];
```

Behaviour could be:
- If `accept` is not specified, keep current behaviour, ie: execute handler regardless of client header
- If `accept` is specified, only execute handler if one of the requested types is matching the specified array

Which also means that it's possible to specify the same route multiple time with different accept headers, and possible specify one without if you want a default handler for this route

Not using `route` or handle in default handler but less elegant

lowlighter:
Closes #5883

### DIFF
--- a/http/route.ts
+++ b/http/route.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+import { accepts } from "./negotiation.ts";
 
 /**
  * Request handler for {@linkcode Route}.
@@ -34,6 +35,10 @@ export interface Route {
    */
   method?: string;
   /**
+   * Accept header.
+   */
+  accepts?: string[];
+  /**
    * Request handler.
    */
   handler: Handler;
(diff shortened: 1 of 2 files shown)


iuioiua:
Thank you for the PR, @lowlighter. But I'll be closing this without merge as we haven't yet reached consensus and I think it's not needed (see my latest comment in the related issue). Either way, thank you.

PR #5897, opened 2024-09-03
fix(yaml): handle string instances

timreichen:
Ref: https://github.com/denoland/std/issues/5857

**Note**
It seems weird to handle special treat strings in `stringifyNode()`. Maybe there is a way to move the evaluation inside `Type.predicate` like other types in a future PR.


iuioiua (approved):
Thank you

PR #5901, opened 2024-09-04
fix(expect): support `expect.hasAssertions()`

eryue0220:
This pr is one of part to solve https://github.com/denoland/std/issues/3964, which is to support expect.hasAssertion api.


testing/_bdd_utils.ts
iuioiua, line 1: This file doesn't seem needed. We should just simply throw `AssertionError` in `testing/bdd.ts`.

eryue0220: Updated.

expect/_assertion.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import { getAssertionState } from "@std/internal/assertion-state";
+
+const assertionState = getAssertionState();
+
+export function hasAssertion() {
```
kt3k, line 7:
```suggestion
export function hasAssertions() {
```

eryue0220: Updated.

expect/expect.ts
```diff
@@ -488,3 +491,21 @@ expect.stringContaining = asymmetricMatchers.stringContaining as (
 expect.stringMatching = asymmetricMatchers.stringMatching as (
   pattern: string | RegExp,
 ) => ReturnType<typeof asymmetricMatchers.stringMatching>;
+/**
+ * `expect.hasAssertions` verifies that a certain number of assertions are called during a test.
```
kt3k, line 495:
```suggestion
 * `expect.hasAssertions` verifies that at least one assertion is called during a test.
```

kt3k:
This seems only works for the first use of `hasAssertions` as we don't create a new assertionState for each `it` call.

The below script doesn't throw with the 2nd case, which misses assertions:
```js
import { expect } from "@std/expect";
import { test } from "@std/testing/bdd";

test("1", () => {
  expect.hasAssertions();
  expect(1).toBe(1);
});
test("2", () => {
  expect.hasAssertions();
});
```

eryue0220:
> This seems only works for the first use of `hasAssertions` as we don't create a new assertionState for each `it` call.
> 
> The below script doesn't throw with the 2nd case, which misses assertions:
> 
> ```js
> import { expect } from "@std/expect";
> import { test } from "@std/testing/bdd";
> 
> test("1", () => {
>   expect.hasAssertions();
>   expect(1).toBe(1);
> });
> test("2", () => {
>   expect.hasAssertions();
> });
> ```

Yes, the above code had some bugs. and I've updated. 

Thanks.

testing/_test_suite.ts
```diff
@@ -383,5 +389,11 @@ export class TestSuiteInternal<T> implements TestSuite<T> {
     } else {
       await fn.call(context, t);
     }
+
+    if (assertionState.checkAssertionErrorStateAndReset()) {
```
kt3k, line 393: I feel this check is unnecessary. If I removed this check, the test cases still pass.

I'm removing this block.

kt3k:
This API has been documented as 'still not available'. I think we can consider this addition as 'fix' as this is already expected to be added.

testing/_test_suite.ts
eryue0220: There is a little tricky about this, I think you run the test inside `Deno.test` and then it will never throw error. But if you remove it, then it works.

PR #5902, opened 2024-09-04
feat(path/unstable): support file `URL` arg in `normalize()`

kt3k:
This PR support file URL as 1st arg of `normalize` function.

related #5537

### DIFF
--- a/path/normalize.ts
+++ b/path/normalize.ts
@@ -26,6 +26,32 @@ import { normalize as windowsNormalize } from "./windows/normalize.ts";
  * @param path Path to be normalized
  * @returns The normalized path.
  */
-export function normalize(path: string): string {
+export function normalize(path: string): string;
+/**
+ * Normalize the path, resolving `'..'` and `'.'` segments.
+ *
+ * Note: Resolving these segments does not necessarily mean that all will be
+ * eliminated. A `'..'` at the top-level will be preserved, and an empty path is
+ * canonically `'.'`.
+ *
+ * @example Usage
+ * ```ts
+ * import { normalize } from "@std/path/normalize";
+ * import { assertEquals } from "@std/assert";
+ *
+ * if (Deno.build.os === "windows") {
+ *   assertEquals(normalize("C:\\foo\\bar\\..\\baz\\quux"), "C:\\foo\\baz\\quux");
+ *   assertEquals(normalize(new URL("file:///C:/foo/bar/../baz/quux")), "C:\\foo\\baz\\quux");
+ * } else {
+ *   assertEquals(normalize("/foo/bar/../baz/quux"), "/foo/baz/quux");
+ *   assertEquals(normalize(new URL("file:///foo/bar/../baz/quux")), "/foo/baz/quux");
+ * }
+ * ```
+ *
+ * @param path Path to be normalized. Path can be a string or a file URL object.
+ * @returns The normalized path.
+ */
+export function normalize(path: string | URL): string;
+export function normalize(path: string | URL): string {
   return isWindows ? windowsNormalize(path) : posixNormalize(path);
 }
--- a/path/normalize.ts
+++ b/path/normalize.ts
@@ -30,6 +30,8 @@ export function normalize(path: string): string;
 /**
  * Normalize the path, resolving `'..'` and `'.'` segments.
  *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
  * Note: Resolving these segments does not necessarily mean that all will be
  * eliminated. A `'..'` at the top-level will be preserved, and an empty path is
  * canonically `'.'`.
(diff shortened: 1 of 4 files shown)


path/normalize.ts
```diff
@@ -26,6 +26,34 @@ import { normalize as windowsNormalize } from "./windows/normalize.ts";
...
+ * } else {
+ *   assertEquals(normalize("/foo/bar/../baz/quux"), "/foo/baz/quux");
+ *   assertEquals(normalize(new URL("file:///foo/bar/../baz/quux")), "/foo/baz/quux");
+ * }
+ * ```
+ *
+ * @param path Path to be normalized. Path can be a string or a file URL object.
+ * @returns The normalized path.
+ */
+export function normalize(path: string | URL): string;
```
iuioiua, line 56:
```suggestion
export function normalize(path: string): string;
```

path/posix/normalize.ts
```diff
@@ -19,28 +20,41 @@ import { isPosixPathSeparator } from "./_util.ts";
...
+ * const path = normalize(new URL("file:///foo/bar//baz/asdf/quux/.."));
+ * assertEquals(path, "/foo/bar/baz/asdf/");
  * ```
  *
- * @param path The path to normalize.
+ * @param path The path to normalize. Path can be a string or a file URL object.
  * @returns The normalized path.
  */
-export function normalize(path: string): string {
+export function normalize(path: string | URL): string;
```
iuioiua, line 55:
```suggestion
export function normalize(path: string): string;
```

path/windows/normalize.ts
```diff
@@ -23,7 +24,38 @@ import { isPathSeparator, isWindowsDeviceRoot } from "./_util.ts";
...
+ * import { assertEquals } from "@std/assert";
+ *
+ * const normalized = normalize(new URL("file:///C:/foo/../bar"));
+ * assertEquals(normalized, "C:\\bar");
+ * ```
+ *
+ * @param path The path to normalize. Path can be a string or a file URL object.
+ * @returns The normalized path
+ */
+export function normalize(path: string | URL): string;
```
iuioiua, line 56:
```suggestion
export function normalize(path: string): string;
```

path/windows/normalize.ts
```diff
@@ -23,7 +24,38 @@ import { isPathSeparator, isWindowsDeviceRoot } from "./_util.ts";
...
+ * ```
+ *
+ * @example Working with URLs
+ * ```ts
+ * import { normalize } from "@std/path/windows/normalize";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const normalized = normalize(new URL("file:///C:/foo/../bar"));
+ * assertEquals(normalized, "C:\\bar");
+ * ```
```
iuioiua, line 51:
```suggestion
 * @example Usage
 * ```ts
 * import { normalize } from "@std/path/windows/normalize";
 * import { assertEquals } from "@std/assert";
 *
 * const normalized = normalize(new URL("file:///C:/foo/../bar"));
 * assertEquals(normalized, "C:\\bar");
 * ```
```

path/posix/normalize.ts
```diff
@@ -19,28 +20,41 @@ import { isPosixPathSeparator } from "./_util.ts";
...
- * url.pathname = normalize("std/assert/../async/retry.ts");
- * assertEquals(url.href, "https://deno.land/std/async/retry.ts");
+ * @example Working with URLs
+ * ```ts
+ * import { normalize } from "@std/path/posix/normalize";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const path = normalize(new URL("file:///foo/bar//baz/asdf/quux/.."));
+ * assertEquals(path, "/foo/bar/baz/asdf/");
  * ```
```
iuioiua, line 50: This should be one example snippet with the "Usage" title.

iuioiua (changes requested):

path/normalize.ts
kt3k: That is the same as the 1st overload

iuioiua: Sorry, I mean `URL`. Ditto elsewhere.

kt3k: I think `string | URL` is better for this type. When we stabilize URL input, this is going to be the only one signature of this function as is. This prepares for it better.

Also we don't need to use `as any` in `path/normalize.ts` by using this type.

iuioiua: That's not how we did the other PRs.

kt3k: I'd suggest we should fix others

iuioiua: Ok. Are you able to fix the others?

kt3k: Sure

iuioiua (approved):

PR #5905, opened 2024-09-04
feat(tar/unstable): `@std/tar`

### LINKED ISSUE #5733: Proposal: change @std/archive to @std/tar
As it stands, @std/archive only contains one type of archive format, Tar. This proposal suggests that it should be renamed to its own @std module for the following reasons:
1. The std currently supports many file formats and all of them, with the exception of Tar, have their own module. They are named after said file format instead of by category of format. It makes no sense for Tar to be treated differently than these other file formats. 
    - @std/csv 
    - @std/dotenv 
    - @std/ini
    - @std/json 
    - @std/jsonc 
    - @std/msgpack 
    - @std/toml 
    - @std/yaml 
1. While there are many archive formats, many of them also double as file compression formats, and some file formats only offer compression. Assuming both @std/archive and @std/compress existed, how would a file format that supported both be categorised?

iuioiua:
This is non-breaking for current streaming APIs within `@std/archive`, as those APIs haven't yet been published.

Closes #5733


kt3k (approved):
LGTM

PR #5909, opened 2024-09-04
feat(cbor/unstable): Introduce `@std/cbor`

BlackAsLight:
Closes https://github.com/denoland/std/issues/5479

This pull request introduces a CBOR implementation based off the [RFC 8949](https://datatracker.ietf.org/doc/html/rfc8949) spec from scratch. At the moment it only introduces `encodeCbor`/`decodeCbor` functions (which operate in a sync manner and therefore is limited to JavaScript's limitations on memory usage around variable sizes) and Byte, Text, Array, Map, and Sequence Encoder Streams for larger amounts of data or if wanting to work with indefinite length types. The decoding stream methods are yet to be committed to this pull request.

## Example
```ts
import { assert, assertEquals } from "@std/assert";
import { decodeCbor, encodeCbor } from "@std/cbor";

const rawMessage = [
  "Hello World",
  35,
  0.5,
  false,
  -1,
  null,
  Uint8Array.from([0, 1, 2, 3]),
];

const encodedMessage = encodeCbor(rawMessage);
const decodedMessage = decodeCbor(encodedMessage);

assert(decodedMessage instanceof Array);
assertEquals(decodedMessage, rawMessage);
```


kt3k:
I wonder if these need to be classes when they both don't have any instance states.

How about making them just `deocde` and `encode` functions? That mirrors the structure of `std/msgpack`

BlackAsLight:
> I wonder if these need to be classes when they both don't have any instance states.
> 
> How about making them just `deocde` and `encode` functions? That mirrors the structure of `std/msgpack`

Sure. I was trying to mirror the structure of `TextEncoder`/`TextDecoder`, but this other structure seems more common in the std.

kt3k:
Side note: `TextDecoder` stores some chunk from the previous decode when `stream` option is specified. Probably that is why it's implemented as a class

iuioiua:
@BlackAsLight, could you please draft this PR and undraft once ready for us to review?

BlackAsLight:
Is there a way I can get the Sentry bot to update or create a new report on what lines are missing?

cbor/deno.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "@std/cbor",
+  "version": "0.0.0",
```
kt3k, line 3: Let's specify `0.1.0`

cbor/deno.json
```diff
@@ -0,0 +1,10 @@
...
+  "exports": {
+    ".": "./mod.ts",
+    "./decode": "./decode.ts",
+    "./encode": "./encode.ts",
+    "./encode-stream": "./encode_stream.ts"
```
kt3k, line 8: `./decode-stream` seems missing?

kt3k:
@BlackAsLight 

Coverage diff of this PR should be visible in this page https://app.codecov.io/gh/denoland/std/pull/5909

> CborSequenceDecoderStream
> CborByteDecodedStream
> CborTextDecodedStream
> CborArrayDecodedStream
> CborMapDecodedStream

There are mixed naming styles, `FooDecoderStream` and `FooDecodedStream`. Are these differences intentional?

---

We'd like to export a single API from a single file. Can you split files based on exported APIs

- `encode_cbor.ts` exports `encodeCbor`
- `encode_cbor_sequence.ts` exports `encodeCborSequence`
- `tag.ts` exports `CborTag`
- `sequence_encoder.ts` exports `CborSequenceEncoderStream`
- `byte_encoder_stream.ts` exports `CborByteEncoderStream`
- `text_encoder_stream.ts` exports `CborTextEncoderStream`
- `array_encoder_stream.ts` exports `CborArrayEncoderStream`
- `map_encoder_stream.ts` exports `CborMapEncoderStream`
- `decode_cbor.ts` exports `decodeCbor`
- `decode_cbor_sequence.ts` exports `decodeCborSequence`
- `sequence_decoder_stream.ts` exports `CborSequenceDecoderStream`
- `byte_decoded_stream.ts` exports `CborByteDecodedStream`
- `text_decoded_stream` exports `CborTextDecodedStream`
- `array_decoded_stream` exports `CborArrayDecodedStream`
- `map_decoded_stream` exports `CborMapDecodedStream`

BlackAsLight:
> @BlackAsLight
> 
> Coverage diff of this PR should be visible in this page https://app.codecov.io/gh/denoland/std/pull/5909

That page hasn't gotten updated in like 5 days now, which is why I asked.
 
> > CborSequenceDecoderStream
> > CborByteDecodedStream
> > CborTextDecodedStream
> > CborArrayDecodedStream
> > CborMapDecodedStream
> 
> There are mixed naming styles, `FooDecoderStream` and `FooDecodedStream`. Are these differences intentional?

The difference in naming style is intentional. The "Encoder" and "Decoder" ones are TransformStreams that actually do the work of converting, while the "Decoded" ones are ReadableStreams that act as a simple wrapper so the user of the lib is able to know what type of chunks to expect.

I did explore the idea of having the "Decoded" ones also be TransformStreams and handling the logic of conversion, but the logic there seemed too complicated as you don't know how far to go until you've actually decoded it.

I also explored the idea of having the other "Encoder" ones (i.e. `CborByteEncodedStream`) act as a mere wrapper for `CborSequenceEncoderStream`, but that then seemed redundant if all you wanted to send for example a byte string, as more checks would essentially need to be done for no reason.

> We'd like to export a single API from a single file. Can you split files based on exported APIs
> 
>     * `encode_cbor.ts` exports `encodeCbor`
> 
>     * `encode_cbor_sequence.ts` exports `encodeCborSequence`
> 
>     * `tag.ts` exports `CborTag`
> 
>     * `sequence_encoder.ts` exports `CborSequenceEncoderStream`
> 
>     * `byte_encoder_stream.ts` exports `CborByteEncoderStream`
> 
>     * `text_encoder_stream.ts` exports `CborTextEncoderStream`
> 
>     * `array_encoder_stream.ts` exports `CborArrayEncoderStream`
> 
>     * `map_encoder_stream.ts` exports `CborMapEncoderStream`
> 
>     * `decode_cbor.ts` exports `decodeCbor`
> 
>     * `decode_cbor_sequence.ts` exports `decodeCborSequence`
> 
>     * `sequence_decoder_stream.ts` exports `CborSequenceDecoderStream`
> 
>     * `byte_decoded_stream.ts` exports `CborByteDecodedStream`
> 
>     * `text_decoded_stream` exports `CborTextDecodedStream`
> 
>     * `array_decoded_stream` exports `CborArrayDecodedStream`
> 
>     * `map_decoded_stream` exports `CborMapDecodedStream`

That is a lot of files, but I can do that.

cbor/types.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export type CborType = CborPrimitiveType | CborTag<CborType> | CborType[] | {
+  [k: string]: CborType;
+};
+
+/**
+ * Specifies the encodable value types for the {@link CborSequenceEncoderStream}
+ * and {@link CborArrayEncoderStream}.
+ */
+export type CborInputStream =
```
kt3k, line 40: How about renaming this to `CborStreamInput` as this itself is not always a stream, but an input for the stream?

The same suggestion could also apply to:
- `CborMapInputStream` -> `CborMapStreamInput`
- `CborOutputStream` -> `CborStreamOutput`
- `CborMapOutputStream` -> `CborMapStreamOutput`

kt3k:
I'll move the files `*_decoded_stream.ts` to the private paths (`_*_decoded_stream.ts`) and remove them from `exports` entries in `cbor/deno.json` as they are not intended to be used directly from the users. Let me know if you disagree with this change.

BlackAsLight:
> I'll move the files `*_decoded_stream.ts` to the private paths (`_*_decoded_stream.ts`) and remove them from `exports` entries in `cbor/deno.json` as they are not intended to be used directly from the users. Let me know if you disagree with this change.

The user isn't meant to create instances of the decoded streams themselves, but they do need access to them to be able to figure out what type they have. For example with an `entry instanceof ByteDecodedStream`.

kt3k:
> The user isn't meant to create instances of the decoded streams themselves, but they do need access to them to be able to figure out what type they have. For example with an entry instanceof ByteDecodedStream.

`instanceof` check makes sense. Currently they are still exported from `mod.ts`. Maybe it's also good to export them from relevant `decoder` streams' endpoints.

cbor/types.ts
kt3k: I'll also work on this change

kt3k:
@BlackAsLight 
I added the exports of `CborArrayDecodedStream`, `CborByteDecodedStream`, `CborMapDecodedStream`, `CborTextDecodedStream` from `cbor/sequence-decoder-stream`. What do you think?

BlackAsLight:
> @BlackAsLight 
> 
> I added the exports of `CborArrayDecodedStream`, `CborByteDecodedStream`, `CborMapDecodedStream`, `CborTextDecodedStream` from `cbor/sequence-decoder-stream`. What do you think?

Looks good to me

kt3k (approved):
LGTM

PR #5910, opened 2024-09-04
feat(data-structures/unstable): `@std/data-structures/bidirectional-map`

### LINKED ISSUE #5736: proposal: `BidirectionalMap`
I suggest a new `BidirectionalMap` class. It'd have all the methods/properties of [`Map`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) but with `getReverse()`, `hasReverse()`, etc. A contributor could use this #4985 as a baseline (closed for other higher priorities).

Liam-Tait:
Adds `BidirectionalMap` class to data structures. An extension of `Map` that supports two-way lookups between keys and values. All methods of `Map` are supported.

Automatically updates entries when a key or value already exists, ensuring no duplicates 
for either.

Closes #5736

### DIFF
--- a/data_structures/bidirectional_map.ts
+++ b/data_structures/bidirectional_map.ts
@@ -215,12 +215,12 @@ export class BidirectionalMap<K, V> extends Map<K, V> {
   /**
    * A String value that is used in the creation of the default string description of an object.
    * Called by the built-in method `Object.prototype.toString`.
-   * 
+   *
    * @example Usage
    * ```ts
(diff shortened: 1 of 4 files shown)


iuioiua (approved):
LGTM! Cleanly done and great tests. Thank you.

iuioiua:
Let's create a tracking issue to use this new class in `@std/encoding`. A few of the modules there could do with using a bidirectional map.

Issue #5470, opened 2024-07-17
tracking: support iterables in `@std/collections` APIs


iuioiua:
Following on from #4671, it seems there are APIs within `@std/collections` that could support `Iterables` rather than just arrays. From a quick scan, these symbols might be able to be tweaked to support iterables:
- [x] `chunk()`
- [x] `dropLastWhile()`
- [x] `dropWhile()`
- [x] `intersect()`
- [x] `sample()`
- [x] `slidingWindows()`
- [x] `sortBy()`
- [x] `takeLastWhile()`
- [x] `takeWhile()`
- [x] `withoutAll()`

See #3390 for an example of how with was done for `groupBy()`.

Liam-Tait:
A larger change to consider. Should collections use generators? This would be breaking change.

Generators give lazy evaluation, memory efficiency, and sometimes better performance

Example:
```typescript
function* takeWhile<T>(
  iterable: Iterable<T>,
  predicate: (el: T) => boolean,
): Generator<T> {
  for (const element of iterable) {
    console.log(`takeWhile(): ${element}`);
    if (!predicate(element)) {
      break;
    }
    yield element;
  }
}

function* map<T, U>(
  iterable: Iterable<T>,
  transform: (el: T) => U,
): Generator<U> {
  for (const element of iterable) {
    console.log(`map(): ${element}`);
    yield transform(element);
  }
}

const arr = [1, 2, 3, 4, 5, 6, 7, 8, 9];
const result = takeWhile(map(arr, (n) => n * 100), n => n < 500)
console.log([...result])
```

iuioiua:
I think we should stick to the current approach unless the performance benefits are sufficient. We'd have to measurements before coming to that decision.

Liam-Tait:
The move to iterables means that generators that never return are valid arguments.
```ts
  function* counter() {
    let i = 0;
    while (true) {
      yield i++;
    }
  }
```
Many of these functions would hang

Examples:
```ts
chunk(counter(), 10)
dropLastWhile(counter, v => v > 30)
dropWhile(counter, v => v === 0)
```
Here are some options:
1. Don't handle the case, onus is on the user to know the generator is infinite. Likely case is  throwing an `Uncaught RangeError: Invalid array length`
2. Catch the error an rethrow?
3. Have a max iterations option available

Liam-Tait:
@iuioiua Would you like to assign this to me, I am happy to commit myself to completing all of these changes

timreichen:
[...]

PR #5911, opened 2024-09-04
feat(collections/unstable): support `Iterable` argument in `takeWhile()`

Liam-Tait:
Support `Iterable` for collections `takeWhile`

#5470


iuioiua (approved):
Nice! LGTM. @kt3k, how do you prefer we handle this, in terms of stability?

kt3k:
> how do you prefer we handle this, in terms of stability?

I'd prefer to avoid overloading with this.

How about considering this as 'refactor' or 'fix'?

kt3k:
Maybe we can just land this as `feat(collections/unstable)` without overloading, but with note of something like `Iterable input is unstable`. WDYT?

iuioiua:
> Or, maybe we can just land this as `feat(collections/unstable)` without overloading, but with note of something like `Iterable input is unstable`. WDYT?

SGTM

Liam-Tait:
A larger change to consider. Should collections use generators? This would be breaking change.

Generators give lazy evaluation, memory efficiency, and sometimes better performance

Example:
```typescript
function* takeWhile<T>(
  iterable: Iterable<T>,
  predicate: (el: T) => boolean,
): Generator<T> {
  for (const element of iterable) {
    console.log(`takeWhile(): ${element}`);
    if (!predicate(element)) {
      break;
    }
    yield element;
  }
}

function* map<T, U>(
  iterable: Iterable<T>,
  transform: (el: T) => U,
): Generator<U> {
  for (const element of iterable) {
    console.log(`map(): ${element}`);
    yield transform(element);
  }
}

const arr = [1, 2, 3, 4, 5, 6, 7, 8, 9];
const result = takeWhile(map(arr, (n) => n * 100), n => n < 500)
console.log([...result])
```

iuioiua:
Sorry for the delay. We're deciding how to best handle these changes that should probably be considered unstable.

Liam-Tait:
No problem. Let me know if you need anything from me

kt3k:
We decided to separate unstable feature from stable features more clearly (as described in #5920 ) to make it clear which part is stable and which part is not.

We are going to move this PR's implementation to `unstable_take_while.ts` (which will be available from JSR with the path `jsr:@std/collections/unstable-take-while` ) before merging.

iuioiua (approved):
Sorry for the delay. We've made a few tweaks, and now, LGTM! Thank you.

kt3k (approved):
LGTM

PR #5913, opened 2024-09-04
perf(assert): add fast path for primitive keyed collections in `equal()`

lionel-rowe:
Towards https://github.com/denoland/std/issues/5903

Haven't yet made corresponding changes for `expect`, I'll add those once the rest of the PR is finalized as it looks like it'll just be a straightforward copy-paste job.


iuioiua:
Can you please add benchmark code and results (using `deno bench`)?

iuioiua:
Also, is it possible to separate this into two PRs tackling the performance improvements and complex case fix separately?

lionel-rowe:
Benchmark code:

```ts
import { assert, assertFalse } from "jsr:@std/assert";
import { equal as equalOld } from "jsr:@std/assert@1.0.4/equal";
import { equal as equalNew } from "https://raw.githubusercontent.com/lionel-rowe/deno_std/assert-keyed-collection-fast-path/assert/equal.ts";

for (const size of [10, 100, 1000, 2500]) {
  for (const [name, equal] of Object.entries({ equalNew, equalOld })) {
    const arr = Array.from({ length: size }, (_, i) => i);

    const set1 = makeSet(arr);
    const set2 = makeSet(arr);

    const map1 = makeMap(arr);
    const map2 = makeMap(arr);

    const set3 = makeSet(arr);
    const set4 = makeSet(arr.with(-1, -1));

    const map3 = makeMap(arr);
    const map4 = makeMap(arr.with(-1, -1));

    const group = `size ${size.toLocaleString("en-US")}`;

    Deno.bench({
      group,
      name,
      baseline: name === "equalNew",
      fn() {
        assert(equal(set1, set2));
        assert(equal(map1, map2));
        assertFalse(equal(set3, set4));
        assertFalse(equal(map3, map4));
      },
    });
  }
}

function makeSet(arr: number[]) {
  return new Set(arr);
}
function makeMap(arr: number[]) {
  return new Map(arr.map((v, i) => [i, v]));
}
```

Results:

[asciicast](https://asciinema.org/a/M3RHskRmCaXScSHzycJmMI0i0)

assert/equal_test.ts
```diff
@@ -276,3 +276,30 @@ Deno.test("equal() WeakMap, WeakRef and WeakSet", () => {
...
 });
+
+Deno.test("equal() fast path for primitive keyed collections", () => {
+  const arr = Array.from({ length: 10_000 }, (_, i) => i);
+
+  function makeSet(arr: number[]) {
+    return new Set(arr);
+  }
+  function makeMap(arr: number[]) {
+    return new Map(arr.map((v, i) => [i, v]));
```
iuioiua, line 287:
```suggestion
    return new Map(arr.map((v) => [v, v]));
```

iuioiua (changes requested):

assert/equal_test.ts
lionel-rowe: That will mean the final key of the `with(-1, -1)` versions differs as well as the value. I guess probably best to include both cases.

iuioiua (approved):
LGTM! Thanks! Great improvement.

PR #5914, opened 2024-09-05
fix(assert): value-equal complex keys

lionel-rowe:
Split from https://github.com/denoland/std/pull/5913. As with that PR, I haven't yet made any corresponding changes for `expect`.

Fixes (?) https://github.com/denoland/std/issues/5912 (not 100% sure it will be fixed for every possible edge case though)

### DIFF
--- a/assert/equal_test.ts
+++ b/assert/equal_test.ts
@@ -304,3 +304,43 @@ Deno.test("equal() fast path for primitive keyed collections", () => {
   const map6 = new Map(arr.with(-1, -1).map((v, i) => [i, v]));
   assertFalse(equal(map5, map6));
 });
+
+Deno.test("equal() keyed collection edge cases", () => {
+  assert(equal(new Set([{ b: 2 }, { a: 1 }]), new Set([{ a: 1 }, { b: 2 }])));
+  assertFalse(
+    equal(new Set([{ b: 2 }, { a: 1 }]), new Set([{ a: 1 }, { b: 3 }])),
+  );
+
+  const sym = Symbol();
+  assert(equal(new Set([sym]), new Set([sym])));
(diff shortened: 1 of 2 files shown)


iuioiua:
Can you please rebase?

assert/equal_test.ts
```diff
@@ -304,3 +304,43 @@ Deno.test("equal() fast path for primitive keyed collections", () => {
...
+  const sym = Symbol();
+  assert(equal(new Set([sym]), new Set([sym])));
+  assert(equal(new Set([sym, "a"]), new Set(["a", sym])));
+  assertFalse(equal(new Set([sym, "a"]), new Set(["b", sym])));
+  assertFalse(equal(new Set([Symbol()]), new Set([Symbol()])));
+  assert(equal(new Set([Symbol.for("x")]), new Set([Symbol.for("x")])));
+
+  assert(equal(
+    new Map([[{ b: 2 }, 2], [{ a: 1 }, 1]]),
+    new Map([[{ a: 1 }, 1], [{ b: 2 }, 2]]),
```
iuioiua, line 323: So we don't care about insertion order. I think that's fine. Do others agree?

iuioiua (approved):
LGTM. Thank you.

kt3k (approved):
LGTM

assert/equal_test.ts
lionel-rowe: > So we don't care about insertion order. I think that's fine. Do others agree?

IMO the existing behavior of not caring about it makes the most sense. There _is_ a valid (if not compelling) case to be made for a caring about it, given that sets/maps are explicitly ordered, but changing that behavior would be a significant breaking change. All this PR aims to is do is make reference equality work the same as value equality in this regard.

PR #5916, opened 2024-09-05
feat(collections): support `Iterable` argument in `slidingWindows()`

Liam-Tait:
Support `Iterable` for collections `slidingWindows`

Maintains the current sliding windows for Arrays for the empty Array case. Using the iterable version would change the output from an empty array causing the case is covered by the "slidingWindows() handles empty Array" test to fail.

Example:
```typescript
slidingWindows(new Array(3), 1)

// The difference is 
// [ [ <1 empty item> ], [ <1 empty item> ], [ <1 empty item> ] ]
// vs
// [ [ undefined ], [ undefined ], [ undefined ] ]
```

If the array specific implementation was removed, this would become a breaking change

With an infinite iterator this function would never return. 
I added a `maxWindows` option that limits the number of windows created which would be useful to prevent that case. (maybe it should throw instead?)

However I can remove if we think a seperate `take` function is better
or the onus is on the user
 e.g
```ts
function* take<T>(iterable: Iterable<T>, count: number): Iterable<T> {
  let i = 0;
  for (const item of iterable) {
    if (i++ >= count) {
      break;
    }
    yield item;
  }
}
function* infiniteNumbers() {
  let i = 0;
  while (true) {
    yield i++;
  }
}

slidingWindows(
  take(infiniteNumbers(), 5),
  3,
); // [ [ 0, 1, 2 ], [ 1, 2, 3 ], [ 2, 3, 4 ] ]
```

#5470

### DIFF
--- a/collections/sliding_windows.ts
+++ b/collections/sliding_windows.ts
@@ -87,8 +87,8 @@ export function slidingWindows<T>(
     throw new RangeError("Both size and step must be positive integer.");
   }
   if (
-    options.maxWindows &&
-    (Number.isInteger(options.maxWindows) && options.maxWindows <= 0)
+    options.maxWindows !== undefined &&
+    (!Number.isInteger(options.maxWindows) || options.maxWindows <= 0)
   ) {
     throw new RangeError("maxWindows must be a positive integer");
   }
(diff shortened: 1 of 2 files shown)


Liam-Tait:
Failing tests seem unrelated to this change

Liam-Tait:
On second thought, `maxWindows` is not really needed

Liam-Tait:
Will make a new pr instead of modifying this one with the new approach to adding unstable apis

PR #5921, opened 2024-09-07
refactor(assert): remove unnecessary `getValFromKeyedCollection()`

lionel-rowe:
Realized it's no longer necessary due to using the `symmetricDifference` extra-fast-path for sets.

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -2,13 +2,8 @@
 // This module is browser compatible.
 
 type KeyedCollection = Set<unknown> | Map<unknown, unknown>;
-
 function isKeyedCollection(x: unknown): x is KeyedCollection {
   return x instanceof Set || x instanceof Map;
-}
-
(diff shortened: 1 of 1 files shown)


assert/equal.ts
```diff
@@ -101,16 +96,13 @@ export function equal(c: unknown, d: unknown): boolean {
         if (primitiveKeysFastPath) {
           if (a instanceof Set) {
             return a.symmetricDifference(b).size === 0;
+          } else if (b instanceof Set) {
+            // Already checked constructorsEqual
+            throw new Error("Unreachable");
```
iuioiua, line 101: Can you please remove this `else if`?

lionel-rowe: It's necessary to narrow the type of `b`. What would be preferable? Just cast with `as`?

iuioiua: Yes please. Let's cast with `as`.

iuioiua (approved):
Nice! Thank you.

PR #5923, opened 2024-09-09
refactor(assert,expect): import internal APIs from more specific paths

kt3k:
This PR changes import specifiers for `@std/internal` more specific to avoid unnecessary dependencies included after #5901 landed.

#5901 introduces a new `@std/internal` API which are not necessary for these APIs.


iuioiua (approved):
Nice

PR #5924, opened 2024-09-09
fix(http): invalid ipv6 hostname printed to console

marvinhagemeister:
Before (invalid):

```sh
http://::1:8000
```

After:

```sh
http://[::1]:8000
```


kt3k (approved):
LGTM. Thanks!

PR #5925, opened 2024-09-09
feat(collections/unstable): `Iterable` argument in `chunk()`

Liam-Tait:
Create an unstable version of collections `chunk()`

This version supports both arrays and iterables. The existing implementation for arrays is essentially a branch in this new api for two reasons. 
- Chunk previously counted empty array values as an part of the chunk, removing this and chunking by non-empty values would make this a breaking change. There is a test case for this.
- The array implementation is generally faster (did some benchmarking locally) which IMO is likely the common use of this api.

#5470


kt3k (approved):
Thanks for updating! LGTM

PR #5926, opened 2024-09-09
fix(collections): ensure `pick` doesn't generate missing properties as `undefined`

### LINKED ISSUE #5922: `pick` shouldn't pick missing keys
https://github.com/denoland/std/blob/582a9f2af1d33fc5adb4cc2301f77df444bb4ce7/collections/pick.ts#L18-L23

This way, keys not present in the original object will be added with the value being `undefined` in the returned object. Is this intended behavior?

yuhr:
Fixes #5922.

### DIFF
--- a/collections/pick.ts
+++ b/collections/pick.ts
@@ -28,7 +28,7 @@ export function pick<T extends object, K extends keyof T>(
   obj: Readonly<T>,
   keys: readonly K[],
 ): Pick<T, K> {
-  return Object.fromEntries(
-    keys.flatMap((k) => (k in obj ? [[k, obj[k]]] : [])),
-  ) as Pick<T, K>;
+  const result = {} as Pick<T, K>;
+  for (const key of keys) if (key in obj) result[key] = obj[key];
+  return result;
 }
--- a/collections/pick.ts
+++ b/collections/pick.ts
@@ -28,5 +28,7 @@ export function pick<T extends object, K extends keyof T>(
   obj: Readonly<T>,
   keys: readonly K[],
 ): Pick<T, K> {
-  return Object.fromEntries(keys.map((k) => [k, obj[k]])) as Pick<T, K>;
+  return Object.fromEntries(
+    keys.flatMap((k) => (Object.hasOwn(obj, k) ? [[k, obj[k]]] : []))
+  ) as Pick<T, K>;
 }
--- a/collections/pick.ts
+++ b/collections/pick.ts
@@ -29,6 +29,6 @@ export function pick<T extends object, K extends keyof T>(
   keys: readonly K[],
 ): Pick<T, K> {
   return Object.fromEntries(
-    keys.flatMap((k) => (Object.hasOwn(obj, k) ? [[k, obj[k]]] : [])),
+    keys.flatMap((k) => (k in obj ? [[k, obj[k]]] : [])),
   ) as Pick<T, K>;
 }
--- a/collections/pick.ts
+++ b/collections/pick.ts
@@ -29,6 +29,6 @@ export function pick<T extends object, K extends keyof T>(
   keys: readonly K[],
 ): Pick<T, K> {
   return Object.fromEntries(
-    keys.flatMap((k) => (Object.hasOwn(obj, k) ? [[k, obj[k]]] : []))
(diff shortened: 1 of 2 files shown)


yuhr:
Oops, for now it should respect non-own properties as it used to.

collections/pick.ts
```diff
@@ -28,5 +28,7 @@ export function pick<T extends object, K extends keyof T>(
   obj: Readonly<T>,
   keys: readonly K[],
 ): Pick<T, K> {
-  return Object.fromEntries(keys.map((k) => [k, obj[k]])) as Pick<T, K>;
+  return Object.fromEntries(
+    keys.flatMap((k) => (k in obj ? [[k, obj[k]]] : [])),
```
lambdalisue, line 32: I have not confirmed but using `flatMap` here would impact the performance a lot for a large object.

collections/pick_test.ts
```diff
@@ -37,3 +50,18 @@ Deno.test({
...
+  name:
+    "pick() returns a new object with own properties that is non-own in the original object)",
+  fn() {
+    // deno-lint-ignore no-explicit-any
+    const obj = { a: 5, b: 6, c: 7, d: 8 } as any;
+    const picked: { toString: typeof Object.prototype.toString } = pick(obj, [
+      "toString",
+    ]);
+
+    assertEquals(picked, { toString: Object.prototype.toString });
```
kt3k, line 64: I personally don't think this is intentionally in this way.. (The check with `Object.hasOwn` meets the users' expectation better in my view)

@lambdalisue What do you think?

lambdalisue: Considering the original purpose, I hadn't envisioned a use case where `toString` would be picked. Therefore, I think @kt3k's suggestion is a good one. However, since the implementation currently references `obj[k]`, changing it to `Object.hasOwn` would be a **breaking change**. Additionally, in terms of type expression, it's written as `K in keyof T`, which means it allows something like `toString` (the following code does not result in a type error).

```ts
import { pick } from "jsr:@std/collections/pick";
import { omit } from "jsr:@std/collections/omit";

class A {
  constructor(public a: string, public b: string, public c: string) {}

  foo(): string {
    return "hello";
  }
}

const obj = new A("a", "b", "c");

const p = pick(obj, ["a", "c", "foo"]);
const o = omit(obj, ["b"]);

console.log(obj.foo());
console.log(p.foo());
console.log(o.foo());
```

This might be beyond the scope of this PR, but I realized that `o.foo()` does not work in the code above. Thus, currently there is an inconsistency between `pick` and `omit`. Given this situation, we need to choose between the following options:

1. Restrict the targets of `pick` and `omit` to `Object.hasOwn` (and adjust the type so it expresses this correctly).
2. Fix the implementation of `omit` so that `o.foo()` does not result in a runtime error.

Personally, I prefer option 1, but I’m not sure how to express this with types.

kt3k: Ah, ok. The current check (`key in obj`) works good for the case of the above `class A` example (`foo` method should be picked in that case).

I'm now in favor of landing this fix.

However, I'm still not sure we should include this particular example (`pick(obj, ["toString"])`) as a test case, as this causes type error if we don't type cast `obj` to any. I think we can consider this particular behavior something like undefined/not decided behavior.

kt3k: How about just removing this test case and revisit later when someone comes up with something better?

yuhr: I'm not opposed to it, but just curious, is the concept of “undefined behavior” common in the standard library? If we decide to make this UB, there has to be an explicit notice in the JSDoc comment.

kt3k: It's not common to make something explicitly undefined behavior in the standard library, but I think we are not ready to decide about this behavior, and also this looks out of scope of this PR.

In my view, in the below example, `pick`ing `foo` looks fine, but `pick`ing `toString` looks strange as it doesn't match its type (it's actually causes a type error).

```ts
import { pick } from "@std/collections";
class A {
  foo() {
  }
  bar() {
  }
}
const a = new A();
pick(a, ["foo"]);
pick(a, ["toString"]);
```

kt3k (approved):
LGTM. Thanks!

lambdalisue:
I've create an issue for `o.foo()` things
https://github.com/denoland/std/issues/5966

PR #5930, opened 2024-09-10
feat(io): add `Seeker[Sync]` interfaces

iuioiua:
After an offline chat with Bartek, we'll be removing `Deno.Seeker[Sync]`, which should've been a part of https://github.com/denoland/deno/issues/9795.


kt3k (approved):
LGTM

PR #5934, opened 2024-09-11
perf(expect): add fast path for primitive keyed collections in `equal()`

lionel-rowe:
Same as https://github.com/denoland/std/pull/5913/ (and https://github.com/denoland/std/pull/5921) but for `expect`


iuioiua (approved):
Nice. Thank you.

PR #5935, opened 2024-09-11
test: use `Deno.errors.NotCapable` where appropriate

iuioiua:
As a result of https://github.com/denoland/deno/pull/25431


kt3k (approved):
LGTM

PR #5936, opened 2024-09-11
fix(expect): value-equal complex keys in `expect.equal()`

lionel-rowe:
Same as https://github.com/denoland/std/pull/5914/ but for `expect`


iuioiua (approved):
Great! Thank you.

Issue #5920, opened 2024-09-06
Unstable APIs in stable packages should be more clearly marked as unstable


kt3k:
Currently we add the note `@experimental **UNSTABLE**: New API, yet to be vetted.` to the docs of unstable APIs. However there are feedbacks that this note is not enough for making the users aware of the unstableness of those APIs, and there are too high chance that users unintentionally start using them.

We currently consider moving unstable APIs under `unstable-` prefixed export paths. For example, (unstable) `route` API in `@std/http` becomes:

```diff
-import { route } from "@std/http/route";
+import { route } from "@std/http/unstable-route";
```

---

- [x] `std/assert`
  - `assertNever` function
- [x] `std/cli` #5946
  - `Spinner` class
  - `SpinnerOptions` type
  - `Color` type (part of `SpinnerOptions`)
- [x] `std/data-structures`
  - `BidirectionalMap` class
- [x] `std/encoding`
  - `Base32EncoderStream` class
  - `Base32DecoderStream` class
  - `Base32HexEncoderStream` class
  - `Base32HexDecoderStream` class
  - `Base64EncoderStream` class
  - `Base64DecoderStream` class
  - `Base64UrlEncoderStream` class
  - `Base64UrlDecoderStream` class
  - `HexEncoderStream` class
  - `HexDecoderStream` class
  - `encodeBase32Hex` function
  - `decodeBase32Hex` function
- [x] `std/front-matter/yaml` #5968 
  - `extract` function (the overload with option)
- [x] `std/html`
  - `isValidCustomElementName` function
- [x] `std/http` #5938
  - `HEADER` value
  - `Header` type
  - `METHOD` value
  - `Method` type
  - `Handler` type
  - `Route` type
  - `route` function #5939 
  - `signCookie` function
  - `verifySignedCookie` function
  - `parseSignedCookie` function
- [x] `std/net`
  - `getNetowrkAddress` function
- [x] `std/path`
  - `basename` function (URL overload) #5957
  - `dirname` function (URL overload) #5954
  - `extname` function (URL overload) #5962
  - `join` function (URL overload) #5964
  - `normalize` function (URL overload) #5967
- [x] `std/path/posix`
  - `basename` function (URL overload) #5957
  - `dirname` function (URL overload) #5954
  - `extname` function (URL overload) #5962
  - `join` function (URL overload) #5964
  - `normalize` function (URL overload) #5967
- [x] `std/path/windows`
  - `basename` function (URL overload) #5957
[...]

PR #5937, opened 2024-09-11
refactor(uuid): move UUID v7 APIs to `./unstable-v7`

kt3k:
part of #5920

This PR makes UUID v7 APIs only available from `@std/uuid/unstable-v7`


iuioiua (approved):
LGTM! We should document this new approach to unstable modules.

iuioiua:
Can you please add this file to the super doc linter as now it's not included in `uuid/mod.ts`?

iuioiua:
We should also probably change the title of the PR to be easier to spot for users. Like "move x to unstable-x"

PR #5939, opened 2024-09-11
BREAKING(http/unstable): move `route` to `./unstable-route`

kt3k:
part of #5920 

This PR moves `route` to `./unstable-route`.

We usually do not re-export unstable APIs from `mod.ts`, but in this case I suggest we should keep it (at least for a while) as it's used in `deno init --serve` output.


iuioiua (approved):

iuioiua:
Shouldn't this be marked as "BREAKING(http/unstable)"?

kt3k:
Ok. Let's use `BREAKING(pkg/unstable)` pattern

PR #5943, opened 2024-09-11
refactor(archive,io): use `Seeker[Sync]` from `@std/io/types`

iuioiua:
Once a new version of `std` is published, unblocks https://github.com/denoland/deno/pull/25551.


kt3k (approved):
LGTM

PR #5948, opened 2024-09-12
BREAKING(html/unstable): move `is-valid-custom-element-name` module to `unstable-is-valid-custom-element-name`

iuioiua:
Towards #5920


kt3k (approved):
LGTM

PR #5949, opened 2024-09-12
BREAKING(net/unstable): move `get-network-address` module to `unstable-get-network-address`

iuioiua:
Towards #5920


http/file_server.ts
```diff
@@ -51,7 +51,7 @@ import { ByteSliceStream } from "@std/streams/byte-slice-stream";
 import { parseArgs } from "@std/cli/parse-args";
 import denoConfig from "./deno.json" with { type: "json" };
 import { format as formatBytes } from "@std/fmt/bytes";
-import { getNetworkAddress } from "@std/net/get-network-address";
+import { getNetworkAddress } from "../net/unstable_get_network_address.ts";
```
kt3k, line 54: I think this causes `Module not found` error when published to JSR

kt3k (approved):
LGTM

PR #5956, opened 2024-09-12
BREAKING(encoding/unstable): move `base32hex-stream` module to `unstable-base32hex-stream`

iuioiua:
Towards #5920


kt3k (approved):
LGTM

kt3k:
Not directly related to this change, but I wonder if the export path might be `unstable-base32-hex-encoder-stream` (dash between `base32` and `hex`)...

iuioiua:
We'd probably have to change `@std/encoding/base32hex`, and that's stable. Perhaps, best to leave at this point.

kt3k:
> We'd probably have to change `@std/encoding/base32hex`, and that's stable. Perhaps, best to leave at this point.

It was missed in the list of #5920, but I just realized `@std/encoding/base32hex` is not stable yet. (`@std/encoding/base64url` is stable though)

iuioiua:
Oh, I didn't notice. Hm... I think it'd look nicer as `base32-hex`, but then it'd look inconsistent with `base64url`.

PR #5959, opened 2024-09-12
BREAKING(encoding/unstable): move `base64url-stream` module to `unstable-base64url-stream`

iuioiua:
Towards #5920


encoding/unstable_base64url_stream.ts
```diff
@@ -4,6 +4,19 @@
 /**
  * Utilities for encoding and decoding to and from base64url in a streaming manner.
  *
+ * ```ts
+ * import { assertEquals } from "@std/assert";
+ * import { encodeBase64Url } from "@std/encoding/base64url";
```
kt3k, line 9: This doesn't seem used?

kt3k (commented):
Left one comment

encoding/unstable_base64url_stream.ts
iuioiua: Fixed

kt3k (approved):
LGTM

PR #5962, opened 2024-09-12
BREAKING(path/unstable): move unstable overload of `extname` to `unstable-extname`

kt3k:
part of #5920 

This PR moves the unstable overload of `extname` function to `./unstable-extname`


iuioiua (approved):
Nice

PR #5941, opened 2024-09-11
fix(net): skip empty mac address for `getNetworkAddress()`

ddavid67:
hi

dont know why but on my machine I have a loopback IP starting with `10.`
which is wrongly found as my local IP when using `getNetworkAddress()` 
loopback link usually don't have any physical interface so we can safely skip empty mac address `00:00:00:00:00:00`

Fixes #5933

PR #5963, opened 2024-09-12
test(net): add test of `getNetworkAddress()`

kt3k:
This PR adds a test fixture to exercise the fix #5941


iuioiua (approved):
Ah, good catch. I missed this.

PR #5970, opened 2024-09-12
chore(net): add back `get-network-address` module to fix `http@1.0.5/file-server`

kt3k:
`http@1.0.5/file-server` is now broken because it depends on `net@^1.0.2/get-network-address` which now resolves to `net@1.0.3/get-network-address`, but that path doesn't exist in `net@1.0.3`. This PR fixes it.

### DIFF
--- a/net/deno.json
+++ b/net/deno.json
@@ -4,6 +4,7 @@
   "exports": {
     ".": "./mod.ts",
     "./get-available-port": "./get_available_port.ts",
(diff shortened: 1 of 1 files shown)


dsherret (approved):
LGTM. It would be nice to have a lint in deno std to detect stable code depending on unstable code and ban that.

kt3k:
> It would be nice to have a lint in deno std to detect stable code depending on unstable code and ban that.

Sounds a good idea 👍

PR #5974, opened 2024-09-13
fix(log): ensure consistent behavior with `@std/log`

### LINKED ISSUE #5972: bug(log): `warn()` and others behave different depending on import
Executing [`warn()`](https://jsr.io/@std/log/doc/~/warn), if imported from [`@std/log`](https://jsr.io/@std/log), prints to the console. However, when imported from `@std/log/warn`, it doesn't print anything. I think this is because [`setup()`](https://jsr.io/@std/log/doc/setup/~/setup) is executed, as it's included in `@std/log`, but not executed as it's excluded from `@std/log/warn`. I believe the fix is to add `import "@std/log/setup"` to `warn()` and its sibling functions, like `info()`, etc.

iuioiua:
Fixes #5972


log/debug_test.ts
iuioiua, line 1: This test is failing because `debug()` doesn't actually print to the console for some reason. Need to figure out why...

kt3k: I think you need to call setup and override the default log level to DEBUG like the below:

```ts
import * as log from "@std/log";
log.setup({
  handlers: {
    default: new log.ConsoleHandler("DEBUG"),
  },  
  loggers: {
    default: {
      level: "DEBUG",
      handlers: ["default"],
    }   
  }
});
log.debug("debug");
```

The default log level is set to "INFO" here https://github.com/denoland/std/blob/c6d1ab2d68ea124966f59c42e81501930629c038/log/_config.ts#L6

iuioiua: Are you suggesting we add that to `log/debug.ts`? What is the expected behavior of the following script?

```ts
import { debug } from "@std/log/debug";

debug("Hello, world!");
```

kt3k: I think the expected behavior is noop because the default log level is not low enough to log DEBUG log.

iuioiua: I have one more thing to check to see if that's true. But if it is true, we should remove the function.

kt3k: It's noop by default, however the behavior can be changed if the user want to see the debug logs. I personally don't think the removal is a good idea.

iuioiua: Yeah, actually, I'll document it behavior then.

iuioiua:
@kt3k PTAL

kt3k (approved):
LGTM

PR #5975, opened 2024-09-13
chore: enable `no-console` lint rule

### LINKED ISSUE #5612: to-do: enable `no-console` lint rule

iuioiua:
Closes #5612


kt3k (approved):
LGTM

PR #5978, opened 2024-09-14
fix(assert): accept abstract classes

0f-0b:
Makes this work:

```ts
import { assertInstanceOf } from "jsr:@std/assert/instance-of";

assertInstanceOf([].values(), Iterator);
```

### DIFF
--- a/assert/instance_of.ts
+++ b/assert/instance_of.ts
@@ -6,9 +6,7 @@ import { AssertionError } from "./assertion_error.ts";
 // deno-lint-ignore no-explicit-any
 export type AnyConstructor = new (...args: any[]) => any;
 /** Gets constructor type */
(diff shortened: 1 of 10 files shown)


assert/instance_of.ts
```diff
@@ -6,9 +6,7 @@ import { AssertionError } from "./assertion_error.ts";
 // deno-lint-ignore no-explicit-any
 export type AnyConstructor = new (...args: any[]) => any;
 /** Gets constructor type */
-export type GetConstructorType<T extends AnyConstructor> = T extends // deno-lint-ignore no-explicit-any
-new (...args: any) => infer C ? C
-  : never;
+export type GetConstructorType<T extends AnyConstructor> = InstanceType<T>;
```
0f-0b, line 9: These types are now unused and should probably be deprecated.

kt3k (approved):
LGTM. Nice!

PR #5983, opened 2024-09-16
docs(fs): clarify permissions requirements for `exists[Sync]()`

### LINKED ISSUE #5977: `@std/fs` `exists[Sync]` Does Not Require `--allow-sys` Any More?
**Describe the bug**
Not using `--allow-sys` does not change anything for those functions, but they're mentioned in the documentation that it is required.

**Steps to Reproduce**
*None*

**Expected behavior**
*Not relevant*

**Environment**
- std version: `@std/fs/1.0.3`

iuioiua:
`--allow-sys` is required as [`Deno.uid()`](https://docs.deno.com/api/deno/~/Deno.uid) and [`Deno.gid()`](https://docs.deno.com/api/deno/~/Deno.gid) are called when `options.isReadable = true`, which both require `--allow-sys` permissions.

Closes #5977

### DIFF
--- a/fs/exists.ts
+++ b/fs/exists.ts
@@ -36,7 +36,8 @@ export interface ExistsOptions {
  * @see {@link https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use} for
  * more information on the time-of-check to time-of-use bug.
  *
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

PR #5987, opened 2024-09-16
feat(tar/unstable): add `type` field to `TarStreamEntry`

iuioiua:
Makes the `untar-stream` module consistent with `tar-stream`. Does this via two new interfaces: `TarStreamEntryFile` and `TarStreamEntryDir`.

CC @BlackAsLight

### DIFF
--- a/tar/untar_stream.ts
+++ b/tar/untar_stream.ts
@@ -114,7 +114,11 @@ export interface PosixUstarFormat {
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  */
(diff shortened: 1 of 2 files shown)


BlackAsLight:
While TarStream only accepts files and directories, a Tar archive may have other types inside it. Like hard links, symbolic links, character/block device node, and FIFO node. Under this current change these types would have `type: "directory"`.

iuioiua:
I didn't even consider that. Thank you for pointing that out.

Issue #5986, opened 2024-09-16
to-do: deprecate and remove `@std/archive`


iuioiua:
We now have the streams-based `@std/tar`. In order to deprecate and remove the old `@std/archive`, we must first migrate:

- [x] [jsr:@deno/wasmbuild](https://jsr.io/@deno/wasmbuild)
- [x] [jsr:@libs/bundle](https://jsr.io/@libs/bundle)
- [x] [jsr:@libs/std](https://jsr.io/@libs/std)
- [x] [jsr:@polyseam/inflate-response](https://jsr.io/@polyseam/inflate-response)

PR #5988, opened 2024-09-16
deprecation(archive/unstable): deprecate `@std/archive`

iuioiua:
This package is being superceded by `@std/tar`.

Towards #5986


kt3k (approved):
LGTM

Issue #5989, opened 2024-09-16
to-do: trim `@std/io` APIs


iuioiua:
[`@std/io`](https://jsr.io/@std/io) contains quite a few deprecated APIs. We should review each one, and if still relevant, provide a migration guide, deprecate, then remove. Doing so pushes users to use [Streams APIs](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) and [`@std/streams`](https://jsr.io/@std/streams), and reduces the number of API choices they need to make for their projects.

| Symbol               | Deprecated | Removal PR | Removed |
|----------------------|------------|------------|---------|
| `BufReader`          | ✅          | #6056           | ✅        |
| `BufWriter`          | ✅          | #6057           | ✅        |
| `copyN()`            | ✅          | #5999      | ✅        |
| `LimitedReader`      | ✅          | #6058           |  ✅        |
| `MultiReader`        | ✅          | #6059           | ✅       |
| `readDelim()`        | ✅          | #6052           | ✅        |
| `readInt()`          | ✅          | #6048           | ✅        |
| `readLines()`        | ✅          | #5991      | ✅        |
| `readLong()`         | ✅          | #6047           | ✅        |
| `readRange[Sync]()`        | ✅          | #6049           | ✅        |
| `readShort()`        | ✅          | #6050           | ✅        |
| `readStringDelim()`  | ✅          | #6001      | ✅        |
| `sliceLongToBytes()` | ✅          | #6005      | ✅        |
| `StringReader`       | ✅          | #6062           |  ✅       |
| `StringWriter`       | ✅          | #6061           | ✅        |

PR #5991, opened 2024-09-16
BREAKING(io): remove `readLines()`

iuioiua:
### What's changed

`readLines()` has been removed from `@std/io`.

### Motivation

### Migration guide

To migrate, pass the readable stream to [`toLines()`](https://jsr.io/@std/streams/doc/unstable-to-lines/~/toLines).

```diff
- import { readLines } from "@std/io/read-lines";
+ import { toLines } from "@std/streams/unstable-to-lines";
  import { assert } from "@std/assert/assert"

  using file = await Deno.open("README.md");

- for await (const line of readLines(file)) {
+ for await (const line of toLines(file.readable)) {
    assert(typeof line === "string");
  }
```

### Related

Depends on #5990
Towards #5989


BlackAsLight:
In the migration guide here, you're using the `using` keyword to automatically close the `file` at the end of the scope, but that isn't needed since you're consuming the `.readable` property which will also automatically close the file once it hits the end of its stream. 

Another person the other day on discord had a bad resource ID error thrown when doing something very similar to this, but that might have been because they weren't consuming the entire readable stream inside the same scope as the `file` variable.

iuioiua:
I believe not. Consuming the `.readable` doesn't automatically close the file. Yeah, the issue you point to seems to be something different.

kt3k (approved):
LGTM

PR #5992, opened 2024-09-16
deprecation(io/unstable): deprecate `copyN()`

iuioiua:
Towards #5989

### DIFF
--- a/io/copy_n.ts
+++ b/io/copy_n.ts
@@ -24,7 +24,9 @@ const DEFAULT_BUFFER_SIZE = 32 * 1024;
  * @param size Read size
  * @returns Number of bytes copied
  *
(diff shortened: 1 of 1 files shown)


io/copy_n.ts
```diff
@@ -24,7 +24,9 @@ const DEFAULT_BUFFER_SIZE = 32 * 1024;
  * @param size Read size
  * @returns Number of bytes copied
  *
- * @deprecated This will be removed in 1.0.0. Use the {@link https://developer.mozilla.org/en-US/docs/Web/API/Streams_API | Web Streams API} instead.
+ * @deprecated Pipe through
+ * {@linkcode https://jsr.io/@std/streams/doc/~/ByteSliceStream | ByteSliceStream}
```
kt3k, line 28: Similarly to the case of #5990, I think it would be better to mention how to convert Reader/Writer to ReadableStream/WritableStream. Otherwise some users would be stuck with type errors.

kt3k (approved):
LGTM

PR #5996, opened 2024-09-17
fix(http): have `getCookies()` return `Record<string, string | undefined>`

### LINKED ISSUE #5979: Return type of `http/getCookies` is `Record<string, string>` instead of `Record<string, string | undefined>`

Because the return type of `getCookies` in the `http` package is `Record<string, string>`, any cookie you try to access has the type `string`, even if it is not actually present.

Change the return type of `getCookies` to `Record<string, string | undefined>`. (I understand that this would be a breaking change, and may not be feasible in the near-future).

I can obviously validate the cookie values in my code. But it would be nice if the type were `string | undefined` so that TypeScript would complain if I tried to use it as a string without first checking that it is undefined.

canac:
Make the return type of `getCookies` `Record<string, string | undefined>` so that TypeScript will force you to check that the cookie is actually present before using it.

Note that this is not a breaking code change, but it is a breaking type change, i.e. code that used to type check successfully may not anymore after updating to a version that includes this change.

Fixes #5979

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -251,10 +251,12 @@ function validateDomain(domain: string) {
  * @param headers The headers instance to get cookies from
  * @return Object with cookie names as keys
  */
-export function getCookies(headers: Headers): Record<string, string> {
+export function getCookies(
+  headers: Headers,
+): Record<string, string | undefined> {
   const cookie = headers.get("Cookie");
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
LGTM. Thank you.

PR #5999, opened 2024-09-17
BREAKING(io): remove `copyN()`

iuioiua:
### What's changed

`copyN()` has been removed from `@std/io`.

### Motivation

### Migration guide

To migrate, use [`ByteSliceStream`](https://jsr.io/@std/streams/doc/~/ByteSliceStream), setting `start` to 0 and `end` to 1 less than that of `size` in `concatN()`, then pipe to the destination writable stream.

```diff
- import { copyN } from "@std/io/copy-n";
+ import { ByteSliceStream } from "@std/streams/byte-slice-stream";

  const source = await Deno.open("README.md");

- await copyN(source, Deno.stdout, 10);
+ await source.readable
+   .pipeThrough(new ByteSliceStream(0, 9))
+   .pipeTo(Deno.stdout.writable);
```

### Related

Towards #5989


kt3k (approved):
LGTM

kt3k:
The user might need `ReadableStreamBYOBReader#read` for the replacement if they need more accurate translation. `readable.pipeThrough(new ByteSliceStream(0, 9))` consumes the entire readable stream, while copyN doesn't consume the entire reader. (The reader is paused at the end of copied bytes, and can be resumed from there)

PR #6003, opened 2024-09-17
feat(http/unstable): add support for multiple request methods on route

### LINKED ISSUE #5993: `@std/http/route` is doesn't automatically route `HEAD` requests
It's difficult to handle the HEAD request method on `route()`. For example:

```js
import { route } from '@std/http/route';

const routes = [
	{
		pattern: new URLPattern({ pathname: '/' }),
		handler: homepageHandler,
	},
	{
		pattern: new URLPattern({ pathname: '/favicon.ico' }),
		handler: faviconHandler,
	},
	{
		pattern: new URLPattern({ pathname: '/:iconSlug/:color?/:darkModeColor?' }),
		handler: iconHandler,
	},
];
```

The route configuration above only supports `GET` request method. You have to add three extra routes with `{ method: "HEAD", ... }`.

What I expect is that all configured methods allow HEAD requests by default. And add an extra option to let users choose strict mode that excludes HEAD.

Or we can change it `method` option to `methods`, so that users can define multiple handlers for one handler:

```diff
// @std/http/route.ts

return (request: Request, info?: Deno.ServeHandlerInfo) => {
  for (const route of routes) {
    const match = route.pattern.exec(request.url);
-   if (match && request.method === (route.method ?? "GET")) {
+   if (match && (route.methods ?? ["HEAD", "GET"]).includes(request.method)) {
      return route.handler(request, info, match);
    }
  }
  return defaultHandler(request, info);
};
```

or use add option `allowHead`:

```diff
// @std/http/route.ts

return (request: Request, info?: Deno.ServeHandlerInfo) => {
  for (const route of routes) {
    const match = route.pattern.exec(request.url);
-   if (match && request.method === (route.method ?? "GET")) {
+   if (match && [route.allowHead && 'HEAD' , route.method ?? 'GET'].filter(Boolean).includes(request.method)) {
      return route.handler(request, info, match);
    }
  }
  return defaultHandler(request, info);
};
```

Oak's `route.get()` handles the `HEAD` method as expected. The same is true for other frameworks, and we may need to consider polishing this part of the experience for developers.

LitoMore:
This is an approach for https://github.com/denoland/std/issues/5993#issuecomment-2355224277.

It allows the `method` option to accept a `string | string[]` value type. It's more convenient for some similar handlers.

Here are some examples:

```ts
const route: Route[] = [
	{
		method: ["GET", "HEAD"],
		pattern: new URLPattern({ pathname: '/status' }),
		handler: (req: Request) => new Response(req.method === 'HEAD' ? null : 'ok'),
	},
]
```

```ts
const route: Route[] = [
	{
		method: ["POST", "UPDATE", "DELETE", ...somethingElse],
		pattern: new URLPattern({ pathname: '/users/:id' }),
		handler: (req: Request) => {
			prepareSomething();
			switch (req.method) {
				case "POST":
					createUser();
					break;
				case "UPDATE":
					updateUser();
					break;
				case "DELETE":
					deleteUser();
					break;
				default:
					runSomethingElse();
					break;
			}
			const result = getConclusion();
			return new Response(result);
		},
	},
]
```

Closes #5993

### DIFF
--- a/http/unstable_route.ts
+++ b/http/unstable_route.ts
@@ -28,11 +28,11 @@ export interface Route {
    */
   pattern: URLPattern;
   /**
-   * Request method.
+   * Request method. This can be a string or an array of strings.
    *
    * @default {"GET"}
    */
-  method?: string;
+  method?: string | string[];
   /**
    * Request handler.
    */
@@ -61,7 +61,12 @@ export interface Route {
  *   {
  *     pattern: new URLPattern({ pathname: "/static/*" }),
  *     handler: (req: Request) => serveDir(req)
- *   }
+ *   },
+ *   {
+ *     pattern: new URLPattern({ pathname: "/api" }),
+ *     method: ["GET", "HEAD"],
+ *     handler: (req: Request) => new Response(req.method === 'HEAD' ? null : 'ok'),
+ *   },
  * ];
  *
  * function defaultHandler(_req: Request) {
@@ -91,7 +96,12 @@ export function route(
   return (request: Request, info?: Deno.ServeHandlerInfo) => {
     for (const route of routes) {
       const match = route.pattern.exec(request.url);
-      if (match && request.method === (route.method ?? "GET")) {
+      if (
+        match &&
+        (Array.isArray(route.method)
+          ? route.method.includes(request.method)
+          : request.method === (route.method ?? "GET"))
+      ) {
         return route.handler(request, info, match);
       }
     }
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Thanks for the PR. This makes sense to me. LGTM

kt3k:
Does anyone have any concern to this addition?

iuioiua (approved):
Actually, this seems fine. Thank you.

PR #6005, opened 2024-09-17
BREAKING(io/unstable): remove `sliceLongToBytes()`

iuioiua:
### What's changed

`sliceLongToBytes()` is being removed from `@std/io`.

### Motivation

### Related

Towards #5989


kt3k (approved):
LGTM

PR #6006, opened 2024-09-18
docs(log): document public interfaces

kt3k:
This PR fixes the errors from the command `deno doc --lint log/mod.ts`

### DIFF
--- a/log/base_handler.ts
+++ b/log/base_handler.ts
@@ -26,7 +26,7 @@ export abstract class BaseHandler {
   /** The log formatters */
   formatter: FormatterFunction;
 
-  /** Construct a {@linkcode BaseHandler} */
+  /** Constructs a new instance. */
   constructor(
     levelName: LevelName,
     options?: BaseHandlerOptions,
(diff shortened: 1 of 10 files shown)


log/base_handler.ts
```diff
@@ -8,19 +8,25 @@ import {
...
 }
 
+/** The base class of handlers */
 export abstract class BaseHandler {
   #levelName: LevelName;
   #level: LogLevel;
+  /** The log formatters */
   formatter: FormatterFunction;
 
+  /** Construct a {@linkcode BaseHandler} */
```
iuioiua, line 29: Ditto for the other constructors
```suggestion
  /** Constructs a new instance. */
```

iuioiua (changes requested):
Maybe we should clarify the title, as this PR only adds descriptions to public APIs.

iuioiua:
Also, I feel like these descriptions aren't particularly informative 😬

kt3k:
> Also, I feel like these descriptions aren't particularly informative 😬

At least this reduces lots of lint errors

PR #6008, opened 2024-09-18
deprecation(io/unstable): deprecate `readShort()`

iuioiua:
This will be removed without replacement.

Towards #5989


kt3k (approved):
LGTM

Replacement would be `DataView.prototype.getInt16` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getInt16

PR #6009, opened 2024-09-18
deprecation(io/unstable): deprecate `readInt()`

iuioiua:
This will be removed without replacement.

Towards #5989.


kt3k (approved):
LGTM

Could be replaced with `DataView.prototype.getInt32` https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView/getInt32

PR #6014, opened 2024-09-18
fix(text): handle code points > U+FFFF in `levenshteinDistance`

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6013

### DIFF
--- a/text/levenshtein_distance.ts
+++ b/text/levenshtein_distance.ts
@@ -6,21 +6,18 @@ const { ceil } = Math;
 // https://dl.acm.org/doi/pdf/10.1145/316542.316550
 const peq = new Uint32Array(0x110000);
 
-function myers32(t: string, p: string): number {
-  const n = unicodeStrLen(t);
-  const m = unicodeStrLen(p);
-  for (let i = 0; i < m;) {
-    const cp = p.codePointAt(i)!;
-    peq[cp]! |= 1 << i;
-    i += cp > 0xffff ? 2 : 1;
+function myers32(t: string[], p: string[]): number {
+  const n = t.length;
+  const m = p.length;
+  for (let i = 0; i < m; i++) {
+    peq[p[i]!.codePointAt(0)!]! |= 1 << i;
   }
   const last = m - 1;
   let pv = -1;
   let mv = 0;
   let score = m;
-  for (let j = 0; j < n;) {
-    const cp = t.codePointAt(j)!;
-    const eq = peq[cp]!;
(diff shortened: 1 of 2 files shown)


lionel-rowe:
Benchmarks:

```ts

import { levenshteinDistance as current } from "jsr:@std/text@1.0.6/levenshtein-distance";
import { levenshteinDistance as next } from "https://raw.githubusercontent.com/lionel-rowe/std/levenshtein-unicode/text/levenshtein_distance.ts";

for (const [name, fn] of Object.entries({ current, next })) {
  Deno.bench(`${name} (ASCII)`, () => {
    fn("a".repeat(10), "b".repeat(10));
    fn("a".repeat(100), "b".repeat(100));

    fn("a".repeat(10), "");
    fn("a".repeat(100), "");
    fn("", "b".repeat(10));
    fn("", "b".repeat(100));

    fn(
      "a".repeat(100) + "b".repeat(100) + "a".repeat(100),
      "b".repeat(100) + "a".repeat(100) + "b".repeat(100),
    );
  });
}

for (const [name, fn] of Object.entries({ current, next })) {
  // will give wrong result with `current`, but just testing perf here
  Deno.bench(`${name} (with emoji)`, () => {
    fn(
      "a".repeat(100) + "💩".repeat(100) + "a".repeat(100),
      "💩".repeat(100) + "a".repeat(100) + "💩".repeat(100),
    );
  });
}
```

Performance is almost identical on my machine, typical run:

image

text/levenshtein_distance.ts
```diff
@@ -4,20 +4,23 @@ const { ceil } = Math;
 
 // This implements Myers' bit-vector algorithm as described here:
 // https://dl.acm.org/doi/pdf/10.1145/316542.316550
-const peq = new Uint32Array(0x10000);
+const peq = new Uint32Array(0x10ffff);
```
0f-0b, line 7:
```suggestion
const peq = new Uint32Array(0x110000);
```

Maybe add a test for `levenshteinDistance("\u{10FFFF}", "\u{10FFFF}")`?

lionel-rowe:
Previous fix was still buggy as it's tough to keep the various increments and length measurements in sync, given that sometimes you want the code point length and sometimes the code unit length. As a result I just switched to using `[...str]` char arrays, which surprisingly still has no measurable impact on perf (updated benchmarks above).

kt3k (approved):
LGTM. Thanks!

PR #6021, opened 2024-09-19
docs(io): update deprecation notices

iuioiua:


io/copy_n.ts
```diff
@@ -26,7 +26,7 @@ const DEFAULT_BUFFER_SIZE = 32 * 1024;
  *
  * @deprecated Pipe the readable stream through a new
  * {@linkcode https://jsr.io/@std/streams/doc/~/ByteSliceStream | ByteSliceStream}
- * instead. This will be removed in the future.
+ * instead. This will be removed in 0.226.0.
```
kt3k, line 29: Do you suppose the next minor version of `std/io` will be 0.226.0? or will it be 0.225.0?

kt3k (approved):

PR #6029, opened 2024-09-20
fix(http): use non-locale-sensitive string methods for comparison

lionel-rowe:
Towards https://github.com/denoland/std/issues/6016

Doesn't include a fix for the various `text` functions as it's less obvious there what the correct fix should be (e.g. should a configurable `locale` option be added?)


kt3k (approved):
Good catch! Thanks

PR #6032, opened 2024-09-22
fix(expect,internal,testing): support `expect.assertions`

eryue0220:
Part of #3964, support `expect.assertions` api


internal/assertion_state.ts
```diff
@@ -69,26 +142,56 @@ export class AssertionState {
...
+   *
+   * const assertionState = new AssertionState();
+   * if (assertionState.checkAssertionCountSatisfied()) {
+   *   // throw AssertionError("");
+   * }
+   * ```
+   */
+  checkAssertionCountSatisfied(): boolean {
+    return this.#state.assertionCount !== undefined &&
+      this.#state.assertionCount > this.#state.assertionTriggeredCount;
```
kt3k, line 193: It seems in jest, `expect.assertions(N)` call expects exactly `N` assertions are made. For example, the below test file fails with the error message `Expected one assertion to be called but received two assertion calls.`:

```js
test("test", () => {
  expect.assertions(1);
  expect(1).toBe(1);
  expect(2).toBe(2);
});
```

eryue0220: I'll check this later.

eryue0220: Thanks, I've updated.

kt3k (approved):

PR #6033, opened 2024-09-22
fix(http): make `file-server` work on Deno Deploy

mxdvl:
## Why?

Because `Deno.permissions.querySync` is not a defined function in the deploy environment.

## What?

Conditionally call `Deno.permissions.querySync` and nullish coalesce to “granted” for network access.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -65,6 +65,9 @@ interface EntryInfo {
 const ENV_PERM_STATUS =
   Deno.permissions.querySync?.({ name: "env", variable: "DENO_DEPLOYMENT_ID" })
     .state ?? "granted"; // for deno deploy
+const NET_PERM_STATUS =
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Thank you!

mxdvl:
Thanks for making Deno so easy/fun to contribute to!

Any idea when `@st/http@1.0.7` will be released?

iuioiua:
We should be cutting a release within the next couple of days. Glad to hear you enjoy contributing 😄

kt3k:
@mxdvl We published [@std/http@1.0.7](https://jsr.io/@std/http@1.0.7) with this fix. Thanks for your contribution!

jakeg:
I presume this won't fix this issue though? https://github.com/denoland/deploy_feedback/issues/713

PR #6034, opened 2024-09-23
docs(archive): clarify deprecation notices

iuioiua:


archive/_common.ts
```diff
@@ -7,7 +7,7 @@ import type { Reader } from "@std/io/types";
  * Base interface for {@linkcode TarMeta}.
  *
  * @deprecated Use {@linkcode https://jsr.io/@std/tar | @std/tar} instead.
- * `@std/archive` will be removed after 0.225.4.
+ * `@std/archive` will be removed, with 0.225.4 being the last published version.
```
kt3k, line 10: I think there's no guarantee that 0.225.4 will be the last version. (For example, it might be required by some CLI updates etc.

iuioiua: If we upgrade `archive` in the next release, I believe it'll be 0.225.4. After that, the package will be deleted from the codebase and no longer published to JSR. I think that deprecation notice reflects that correctly. Do you have another suggestion?

To clarify, are you saying that if a bug is found with `archive`, we might re-create the package and publish the bug fix? If so, I think that's too cumbersome a process and we should just urge users to use `tar` and iterate on that.

kt3k: I think `@std/tar` is not ready yet to replace `@std/archive`. There seem some limitation that `@std/tar` has, but `@std/archive` doesn't have. ref: https://github.com/denoland/std/issues/6019

iuioiua: Hm... Fair point. I'll change this to just say "in the future". We can re-visit later.

kt3k (approved):
LGTM

PR #6036, opened 2024-09-23
feat(collections/unstable): support `Iterable` argument in `intersect()`

Liam-Tait:
Create a backwards compatible unstable version of `intersect()` that accepts iterables instead of arrays.

#5470


kt3k (approved):
LGTM. Thanks!

PR #6045, opened 2024-09-24
docs(http/unstable): add example for multiple request methods on route

LitoMore:
This adds an example of the multiple request methods on route definition.


kt3k (approved):
LGTM. Thanks!

PR #6046, opened 2024-09-25
feat(streams/unstable): `toByteStream()`

BlackAsLight:
This pull request introduces a small function to `@std/streams` that allows someone to easily upgrade a `ReadableStream<Uint8Array>` to support BYOB if it doesn't already. 

## Example
```ts
import { upgradeReadable } from "@std/streams/unstable-upgrade-readable-stream";

const reader = upgradeReadable(ReadableStream.from([
  new Uint8Array(100),
  new Uint8Array(78),
  new Uint8Array(125),
]))
  .getReader({ mode: "byob" });

reader.releaseLock()
```


kt3k:
This might be useful in some cases, but I'm not sure if it's common to call this operation/util 'upgrade'...('upgrade' reminds me of upgrading http connection to websocket connection). How about calling this 'toReadableByteStream'?

@crowlKats @lucacasonato What do you think about this?

BlackAsLight:
`toByteStream` is a slightly shorter name. 

I use a similar function a lot in the cbor pull request and thought it's unique enough and useful enough to offer as a utility. 

I also found if you don't end up enqueuing or responding in each pull then you end up with that promise left in the event loop error.

kt3k:
`toByteStream` sounds good to me

BlackAsLight:
I'd just like to say that getting this function right was quite hard as there always seemed like another hiccup. 

I've found that when using the `min` property, `value` will always be a `Uint8Array`, even when `done` is `true` and TypeScript infers it _might_ be `undefined`. `value` is still a `Uint8Array`, even if that length is zero. `value` when `done` is `true` will contain the amount it was able to get that's less than `min`.

PR #6047, opened 2024-09-25
BREAKING(io): remove `readLong()`

iuioiua:
### What's changed

`readLong()` is being removed from `@std/io`.

### Motivation

### Related

Towards #5989
Prerequisite #6005


kt3k (approved):
LGTM

PR #6050, opened 2024-09-25
BREAKING(io): remove `readShort()`

iuioiua:
### What's changed

`readShort()` is being removed from `@std/io`.

### Motivation

### Related

Towards #5989
Pre-requisite #6048


kt3k (approved):
LGTM

PR #6056, opened 2024-09-25
BREAKING(io): remove `BufReader`

iuioiua:
### What's changed

`BufReader` has been removed from `@std/io`.

### Motivation

### Migration guide

To migrate, use [`Buffer`](https://jsr.io/@std/streams/doc/buffer/~/Buffer) from [`@std/streams`](https://jsr.io/@std/streams) instead. If you have any use cases which lack a clear means of migration, [open an issue](https://github.com/denoland/std/issues/new/choose).

Before:
```ts
import { Buffer, BufReader } from "@std/io";
import { assertEquals } from "@std/assert/equals";

const reader = new BufReader(new Buffer(encoder.encode("Hello, world!")));
const buffer = new Uint8Array(11);
await reader.read(buffer);
const result = decoder.decode(buffer);

assertEquals(result, "Hello, world!");
```

After:
```ts
import { Buffer } from "@std/streams/buffer";
import { toText } from "@std/streams/to-text";

const buffer = new Buffer(encoder.encode("Hello, world!"));
const result = await toText(buffer.readable);

assertEquals(result, "Hello, world!");
```

### Related

Towards #5989


kt3k (approved):
LGTM

PR #6059, opened 2024-09-25
BREAKING(io): remove `MultiReader`

iuioiua:
### What's changed

`MultiReader` has been removed from `@std/io`.

### Motivation

This change was made to help push users towards using the [Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) and [`@std/streams`](https://jsr.io/@std/streams), and make design decisions about which APIs to use easier.

### Migration guide

To migrate, combine the corresponding readable streams using [`mergeReadableStreams`](https://jsr.io/@std/streams/doc/merge-readable-streams/~/mergeReadableStreams).

Before:
```ts
import { MultiReader } from "@std/io/multi-reader";
import { StringReader } from "@std/io/string-reader";
import { readAll } from "@std/io/read-all";

const r1 = new StringReader("hello");
const r2 = new StringReader("world");

const reader = new MultiReader([r1, r2]);
const bytes = await readAll(reader);

console.log(new TextDecoder().decode(bytes)); // helloworld
```

After:
```ts
import { mergeReadableStreams } from "@std/streams/merge-readable-streams";
import { toText } from "@std/streams/to-text";

const r1 = ReadableStream.from(["hello"]);
const r2 = ReadableStream.from(["world"]);

const mergedStream = mergeReadableStreams(r1, r2);

console.log(await toText(mergedStream)); // helloworld
```

### Related

Towards #5989


kt3k (approved):
LGTM

PR #6064, opened 2024-09-26
fix(tar): `UntarStream` bug

BlackAsLight:
Fixes: https://github.com/denoland/wasmbuild/pull/143

This pull request fixes a bug with the `UntarStream` class where it incorrectly assumes the contents of a stream will contain a tar file. This is incorrect as anything after a tar file can be appended and a valid tar decoder should ignore everything after it receives its 1024 bytes of `0b00` when checking for a header.

### DIFF
--- a/tar/untar_stream.ts
+++ b/tar/untar_stream.ts
@@ -176,7 +176,9 @@ export class UntarStream
   implements TransformStream<Uint8Array, TarStreamEntry> {
   #readable: ReadableStream<TarStreamEntry>;
   #writable: WritableStream<Uint8Array>;
-  #gen: AsyncGenerator<Uint8Array>;
+  #reader: ReadableStreamDefaultReader<Uint8Array>;
+  #buffer: Uint8Array[] = [];
+  // #gen: AsyncGenerator<Uint8Array>;
   #lock = false;
   constructor() {
     const { readable, writable } = new TransformStream<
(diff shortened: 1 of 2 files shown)


tar/untar_stream.ts
```diff
@@ -185,43 +187,50 @@ export class UntarStream
...
     while (true) {
       while (this.#lock) {
         await new Promise((resolve) => setTimeout(resolve, 0));
       }
 
-      const { done, value } = await this.#gen.next();
-      if (done) break;
+      // Check for premature ending
+      if (this.#buffer.every((value) => value.every((x) => x === 0))) {
+        await this.#reader.cancel("Tar Stream Finished");
```
iuioiua, line 223:
```suggestion
        await this.#reader.cancel(new RangeError("Tar stream finished prematurely"));
```

tar/untar_stream_test.ts
```diff
@@ -133,35 +133,31 @@ Deno.test("UntarStream() with size equals to multiple of 512", async () => {
...
-      size: 512,
-      readable: ReadableStream.from([new Uint8Array(512).fill(97)]),
-    },
-  ])
-    .pipeThrough(new TarStream())
-    .pipeThrough(
-      new TransformStream<Uint8Array, Uint8Array>({
-        flush(controller) {
-          controller.enqueue(new Uint8Array(100));
+  const bytes = concat(
```
iuioiua, line 136: We can now use [`toBytes()`](https://jsr.io/@std/streams/doc/unstable-to-bytes/~/toBytes) instead of `concat()` and `Array.fromAsync()`.

iuioiua (changes requested):

tar/untar_stream.ts
BlackAsLight: This isn't meant to be an error. It's perfectly fine for it to finish with data still in the stream

iuioiua: Fair enough. Let's go with `Tar stream finished prematurely` - it shouldn't be in title case and "prematurely" might provide needed context.

iuioiua (approved):
Thank you!

kt3k (approved):
LGTM

PR #6070, opened 2024-09-27
chore: re-enable `test:browser` task in CI

iuioiua:


kt3k:
I think this is blocked by https://github.com/denoland/deno/issues/25847

It currently passes because we published `@std/internal@1.0.4` and the difference between published package and local packages was removed. I think the same error will happen when someone added a new entrypoint in any package.

kt3k (approved):
The above issue fixed by https://github.com/denoland/deno/pull/25988

Confirmed non-published local change of exported items works without issue with Deno Canary.

LGTM

PR #6072, opened 2024-09-27
fix(log): remove leaky `ConsoleHandler.prototype.applyColors()`

iuioiua:
There doesn't appear to be a reason for this to be a method on the `ConsoleHandler` class. This move avoids the need to document it as a method.


kt3k (approved):
This change looks reasonable to me. `applyColors` seems like an internal utility instead of a public method.

LGTM

PR #6075, opened 2024-09-28
fix(expect): register extended matchers after native matchers

### LINKED ISSUE #6074: bug @std/expect: extended matchers are registered before native matchers
**Describe the bug**

`extendMatchers` are registered before `matchers` which means it's not possible to "extend" native matchers.

https://github.com/denoland/std/blob/e9716720b690e72e6ceaba308b665f2002f7b117/expect/expect.ts#L183-L186

There shouldn't be any restriction on this.
Since `ExtendedExpected` extends `Expected` anyways, "augmented" native matchers still needs to fulfill the initial signature.

**Steps to Reproduce**

Try to override a native matcher, like `toThrow()`

**Expected behavior**

Possible to override a native matcher with additional functionalities.
For example add a second argument to `toThrow()` to be on par with `assertIsError()`

lowlighter:
Close #6074

### DIFF
--- a/expect/expect.ts
+++ b/expect/expect.ts
@@ -181,8 +181,8 @@ export function expect<T extends Expected = Expected>(
 
       const extendMatchers: Matchers = getExtendMatchers();
       const allMatchers = {
-        ...extendMatchers,
         ...matchers,
+        ...extendMatchers,
       };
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM. Thanks for the fix!

PR #6076, opened 2024-09-28
feat(collections/unstable): support `Iterable` argument in `dropLastWhile()`

Liam-Tait:
Create a backwards compatible unstable version of `dropLastWhile()` that accepts an iterable.

Slight modification to the loop to avoid the `offset - 1` calculation each loop.

#5470

- [x] Remove the test benchmark before merging


kt3k (approved):
LGTM. Thanks!

PR #6077, opened 2024-09-28
docs(log): correct the examples of `debug()`

babiabeo:
This PR fixes the error that has occurred since https://github.com/denoland/std/commit/827fe95620fda2e3426189f80e754fac67e09898.


iuioiua (approved):
Thank you

PR #6078, opened 2024-09-28
BREAKING(log): remove `NOTSET` `LogLevel`

timreichen:
**Changes**
This PR removes the `NOTSET` `LogLevel`.

**Reasoning**
Instead of `NOTSET` loggers should be initialized with the default set `LogLevel` (which is `INFO`).


timreichen:
The log module is not stable, so breaking changes for v1 should not be a dealbreaker imo. 
To me this seems to be a question of principal: Does `NOTSET` as a loglevel make sense in the ts/js context? I would argue no. WDYT?

log/get_logger.ts
```diff
@@ -73,7 +74,7 @@ export function getLogger(name?: string): Logger {
   }
   const result = state.loggers.get(name);
   if (!result) {
-    const logger = new Logger(name, "NOTSET", { handlers: [] });
+    const logger = new Logger(name, DEFAULT_LEVEL, { handlers: [] });
```
kt3k, line 77: I wonder if throwing here would make more sense. It might be confusing to return default-leveled logger for invalid logger name

kt3k: Nevermind. I now don't think throwing makes sense after reading the `For module authors` section of the module document ( https://jsr.io/@std/log ). The module authors seem supposed to call `getLogger` with non-defined logger name.

kt3k:
The created logger doesn't have any handler. It means the logger doesn't do anything. I wonder if this change would impact the user in any meaningful way..

PR #6088, opened 2024-10-04
feat(collections/unstable): support `Iterable` argument in `dropWhile()`

Liam-Tait:
Create a backwards compatible unstable version of `dropWhile()` that accepts an iterable instead of arrays.

The array version is significantly more performant, so add the special case for when the iterable is an array.

Array path use `Array.prototype.findIndex()` instead tracking `offset` index  with a while loop. Testing locally using findIndex has a slight performance improvement, pretty negligible. IMO it is easier to understand though. I do not feel strongly about it, so let me know if the existing way approach is preferred.

#5470


kt3k:
> Array path use Array.prototype.findIndex() instead tracking offset index with a while loop. Testing locally using findIndex has a slight performance improvement, pretty negligible. IMO it is easier to understand though.

The use of findIndex just makes sense to me.

kt3k (approved):
LGTM

PR #6094, opened 2024-10-05
BREAKING(http/unstable): switch `params` and `info` args in `Handler` in `route()` for more conveniency

### LINKED ISSUE #5884: qol(http/route): typings and handler signature
The `Handler` interface is currently defined as:
https://github.com/denoland/std/blob/94a7e1b34be4b2f18af1238d7b62c1a0cc700f41/http/route.ts#L14-L18

But I feel like people users using this API would most likely prefer to have `URLPatternResult` as second argument and `Deno.ServeHandlerInfo` as third one.

Reasoning is that people using this API will almost always want to retrieve their route params, but the usage of `Deno.ServeHandlerInfo` is less likely. Also since the url/method is already checked by the `route()` function, it is also common not have have use for `Request`, so you may end up with:

```ts
handler: (_, __, result) => return new Response(result.pathname.groups.name)
```

Which is less elegant than:
```ts
handler: (_, result) => return new Response(result.pathname.groups.name)
```

I understand it was done to be compatible with regular handlers, but maybe `route()` could support an union of both.

___

Also the typing `params?: URLPatternResult | null,` seems incorrect (at least for non-default handler), because the result is checked before being passed down to handler:
https://github.com/denoland/std/blob/94a7e1b34be4b2f18af1238d7b62c1a0cc700f41/http/route.ts#L93-L96

So users need to add an uneeded `!.` since to avoid typing issues

lowlighter:
Closes  #5884

Not sure if it qualifies for "breaking change" since the API is still unstable, but it does break apps relying on current implementation `route()`

As explained in the linked issue, since `params: URLPatternResult` is way more likely to be used than `info: Deno.ServeHandlerInfo` in non-default handlers, it makes sense to switch them for improved DX

### DIFF
--- a/http/mod.ts
+++ b/http/mod.ts
@@ -104,6 +104,3 @@ export * from "./negotiation.ts";
 export * from "./server_sent_event_stream.ts";
 export * from "./user_agent.ts";
 export * from "./file_server.ts";
-// We keep this re-export as an exception for now as it's used in
-// `deno init --serve` output
-export * from "./unstable_route.ts";
--- a/http/unstable_route.ts
+++ b/http/unstable_route.ts
@@ -13,8 +13,8 @@
  */
 export type Handler = (
   request: Request,
+  params?: URLPatternResult,
   info?: Deno.ServeHandlerInfo,
-  params?: URLPatternResult | null,
 ) => Response | Promise<Response>;
 
 /**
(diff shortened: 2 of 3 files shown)


kt3k (approved):
LGTM.

There should be more usages for `params` than `info` in general. This change makes sense.

kt3k:
We recently removed the dependency to (unstable) `route` in main of CLI (in `deno init --serve`). ref. https://github.com/denoland/deno/pull/26595

I think we still need to wait for a while (at least a few versions of CLI releases) to land this PR to avoid the troubles.

kt3k:
We've released 3 versions of CLI (2.1.0, 2.1.1, 2.1.1) which don't use unstable `route` function in `deno init --serve` output.

I think it's time we can land this.

PR #6095, opened 2024-10-07
feat(collections/unstable): support `Iterable` argument in `slidingWindows`

Liam-Tait:
Support Iterable for `slidingWindows`

While working here I have made the errors a bit more granular.

With light benchmarking I have also made the array use case around ~2x faster (on my machine).

- [x] Remove the benchmark before merging (causing lint error as it is not exported)

#5470


kt3k (approved):
LGTM. Thanks!

PR #6104, opened 2024-10-10
refactor(json): use trim for blank space check

softwarespot:
Using `String.prototype.trim` to determine if a chunk is "blank", checks for more white space characters, than was in the previous regular expression.

### DIFF
--- a/json/parse_stream.ts
+++ b/json/parse_stream.ts
@@ -4,18 +4,14 @@
 import type { JsonValue } from "./types.ts";
 import { parse } from "./_common.ts";
 
-const blanks = /^[ \t\r\n]*$/;
-function isBlankSpace(str: string) {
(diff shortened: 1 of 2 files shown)


kt3k:
To my understanding, `chunk.trim()` allocates a new string, while `regexp.test(chunk)` doesn't. I think this has negative effects in runtime performance.

softwarespot:
You make a fair point, and I could be wrong, don't regular expressions also have a negative affect on runtime performance when the length of the chunk gets big?

timreichen:
Ref: https://github.com/denoland/std/pull/5751

softwarespot:
@timreichen I missed that PR, so thanks for bringing it to my attention. I will close this PR in that case. Thanks everyone

PR #6105, opened 2024-10-10
docs(path): re-add URL examples to `@std/path/posix` examples

devmattrick:
I'm not sure if this was intentional, but the documentation added regarding URL manipulation with the `@std/path/posix` module in #5371 was inadvertently removed in #5900 and #5904.

It seems like the methods that accept a URL object are still in unstable and the [current docs](https://jsr.io/@std/url) for `@std/url` link to the `@std/path/posix` module, so I think it might be a good idea to re-add these examples.


kt3k (approved):
Sounds reasonable to me. Thanks for the suggestion.

PR #6106, opened 2024-10-11
test(cbor): empty string being excluded from expected result

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6102

The expected result was filtering out empty strings provided in the input text, while `CborTextEncoderStream` doesn't remove them. The test appeared flaky because of the randomness of the test. In the event the first random chose zero (which is a 1 in 24 chance), an empty string would be one of the chunks. An empty string is encoded as byte 96. 

Lowering the size of the strings had no effect on the flakiness of the tests, but I am assuming that's why they were lowered in the first place.

### DIFF
--- a/cbor/text_encoder_stream_test.ts
+++ b/cbor/text_encoder_stream_test.ts
@@ -13,12 +13,13 @@ Deno.test(
     const strings = [
       "a".repeat(random(0, 24)),
       "a".repeat(random(24, 2 ** 8)),
-      "a".repeat(random(2 ** 8, 2 ** 10)),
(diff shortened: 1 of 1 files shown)


kt3k:
> The expected result was filtering out empty strings provided in the input text, while CborTextEncoderStream doesn't remove them. The test appeared flaky because of the randomness of the test. In the event the first random chose zero (which is a 1 in 24 chance), an empty string would be one of the chunks. An empty string is encoded as byte 96.

Thanks for looking into this!

kt3k (approved):
LGTM

PR #6107, opened 2024-10-11
refactor(cbor): replace `toByteStream` function in common with import from `@std/streams`

BlackAsLight:
This pull request replaces the `toByteStream` function that existed in `./common.ts` with the identical one that was recently merged into `@std/streams`.

This pull request also sorts all the imports alphabetically.


kt3k (approved):
LGTM. Thanks!

PR #6113, opened 2024-10-13
docs(tar): fix example in creating directories

BlackAsLight:
This pull request fixes a highly probable bug in the JSDocs example where if the directory already exists, it throws an error, or if the directory's directory doesn't exist, it throws an error.


kt3k (approved):
LGTM. Thanks!

PR #6115, opened 2024-10-13
fix(cbor): number precision error in decoding test

BlackAsLight:
Fixes: https://github.com/denoland/std/actions/runs/11310597614/job/31455830417#step:4:1

Due to the way JavaScript numbers work, numbers past `2 ** 53` lose precision. Weirdly though sometimes when converting them to BigInts they may gain more precision than they originally had.


kt3k (approved):
LGTM. Thanks!

kt3k:
BTW is it a good idea to show warnings or something when the input of `encodeCbor` is a number > `Number.MAX_SAFE_INTEGER`?

PR #6117, opened 2024-10-14
test(fmt): handle missing group separator for 1000.1 in some locales

IgorM867:
In some locales there is no group separator in parts which cause error during tests.
```ts
const parts = new Intl.NumberFormat().formatToParts(1000.1);
console.log(parts);
const group = parts.find(({ type }) => type === "group")!.value;
```
```
[
  { type: "integer", value: "1000" },
  { type: "decimal", value: "," },   
  { type: "fraction", value: "1" }   
]
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'value')
const group = parts.find(({ type }) => type === "group")!.value;
```
Change to 10000.1 fix this issue.
```typescript
const parts = new Intl.NumberFormat().formatToParts(10000.1);
console.log(parts);
const group = parts.find(({ type }) => type === "group")!.value;
```
```ts
[
  { type: "integer", value: "10" },
  { type: "group", value: " " },
  { type: "integer", value: "000" },
  { type: "decimal", value: "," },
  { type: "fraction", value: "1" }
]
```

### DIFF
--- a/fmt/bytes_test.ts
+++ b/fmt/bytes_test.ts
@@ -4,10 +4,10 @@
 
 import { format } from "./bytes.ts";
 import { assertEquals, assertThrows } from "@std/assert";
+import { stub } from "@std/testing/mock";
 
 const parts = new Intl.NumberFormat().formatToParts(10000.1);
 const decimal = parts.find(({ type }) => type === "decimal")!.value;
-const group = parts.find(({ type }) => type === "group")!.value;
 
 Deno.test("format() throws on invalid input", () => {
   // deno-lint-ignore no-explicit-any
@@ -76,10 +76,6 @@ Deno.test("format() handles locale option", () => {
   assertEquals(format(0.4, { locale: true }), `0${decimal}4 B`);
   assertEquals(format(1001, { locale: true }), "1 kB");
   assertEquals(format(10.1, { locale: true }), `10${decimal}1 B`);
-  assertEquals(
-    format(1e30, { locale: true }),
-    `1${group}000${group}000 YB`,
-  );
 
   assertEquals(format(-0.4, { locale: false }), "-0.4 B");
   assertEquals(format(0.4, { locale: false }), "0.4 B");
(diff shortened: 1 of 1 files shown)


lionel-rowe:
This still won't make the tests fully locale independent — at a minimum, it'll fail for locales like `hi-IN` that use [lakhs and crores](https://en.wikipedia.org/wiki/Indian_numbering_system).

There's currently only one test that uses `group`, so something like this could work:

```ts
import { format } from "./bytes.ts";

function mockNumberFormatDefaultLocale(defaultLocale: string) {
  const IntlNumberFormat = Intl.NumberFormat;
  // @ts-expect-error the original Intl.NumberFormat can be called without `new` (but our mock can't)
  Intl.NumberFormat = class extends IntlNumberFormat {
    constructor(
      locale?: Intl.LocalesArgument,
      opts?: Intl.NumberFormatOptions,
    ) {
      super(locale ?? defaultLocale, opts);
    }
  };

  const NumberPrototypeToLocaleString = Number.prototype.toLocaleString;

  Number.prototype.toLocaleString = function (locale, opts) {
    return NumberPrototypeToLocaleString.call(
      this,
      locale ?? defaultLocale,
      opts,
    );
  };

  return {
    [Symbol.dispose]() {
      Intl.NumberFormat = IntlNumberFormat;
      Number.prototype.toLocaleString = NumberPrototypeToLocaleString;
    },
  };
}

Deno.test("format() handles group separators", async (t) => {
  await t.step("thousands", async (t) => {
    for (const locale of ["en-US", "fr-FR", "es-ES"]) {
      await t.step(locale, () => {
        using _ = mockNumberFormatDefaultLocale(locale);

        const parts = new Intl.NumberFormat().formatToParts(10_000.1);
        const group = parts.find(({ type }) => type === "group")!.value;

        assertEquals(
          format(1e30, { locale: true }),
          `1${group}000${group}000 YB`,
        );
      });
    }
  });
  await t.step("lakhs and crores", async (t) => {
    for (const locale of ["hi-IN", "gu-IN"]) {
      await t.step(locale, () => {
        using _ = mockNumberFormatDefaultLocale(locale);

        const parts = new Intl.NumberFormat().formatToParts(10_000.1);
        const group = parts.find(({ type }) => type === "group")!.value;

        assertEquals(
          format(1e30, { locale: true }),
          `10${group}00${group}000 YB`,
        );
      });
    }
  });
});
```

kt3k (approved):
LGTM. Thanks!

PR #6121, opened 2024-10-16
BREAKING(ini): parse understands booleans, undefined, null and numbers

### LINKED ISSUE #4806: ini file parsing considers all values as strings
**Describe the bug**

**Steps to Reproduce**

```
> import { parse } from "jsr:@std/ini"
undefined
> parse("value=undefined")
{ value: "undefined" }
> parse("value=null")
{ value: "null" }
> parse("value=true")
{ value: "true" }
> parse("value=false")
{ value: "false" }
> parse("value=123")
{ value: "123" }
```

**Expected behavior**

Should understand booleans, undefined, null, and numbers.

**Environment**

- OS: [e.g. Ubuntu 20.04, MacOS 11]
- deno version:
- std version: 0.224.0

WasixXD:
Closes #4806

### DIFF
--- a/ini/_ini_map.ts
+++ b/ini/_ini_map.ts
@@ -302,7 +302,13 @@ export class IniMap<T = any> {
     }
     const reviverFunc: ReviverFunction = typeof reviver === "function"
       ? reviver
-      : (_key, value, _section) => value;
+      : (_key, value, _section) => {
+        if (value === "undefined") return undefined;
+        if (!isNaN(+value) && !value.includes('"')) return parseInt(value);
+        if (value === "null") return null;
+        if (value === "true" || value === "false") return value === "true";
+        return trimQuotes(value);
+      };
     let lineNumber = 1;
     let currentSection: LineSection | undefined;
 
@@ -355,7 +361,7 @@ export class IniMap<T = any> {
         }
 
         const key = leftHand.trim();
-        const value = trimQuotes(rightHand.trim());
+        const value = rightHand.trim();
 
         if (currentSection) {
           const lineValue: LineValue = {
--- a/ini/_ini_map.ts
+++ b/ini/_ini_map.ts
@@ -304,7 +304,7 @@ export class IniMap<T = any> {
       ? reviver
       : (_key, value, _section) => {
         if (value === "undefined") return undefined;
-        if (!isNaN(+value) && !value.includes('"')) return parseInt(value);
+        if (!isNaN(+value) && !value.includes('"')) return +value;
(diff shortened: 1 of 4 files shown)


ini/_ini_map.ts
```diff
@@ -302,7 +302,13 @@ export class IniMap<T = any> {
     }
     const reviverFunc: ReviverFunction = typeof reviver === "function"
       ? reviver
-      : (_key, value, _section) => value;
+      : (_key, value, _section) => {
+        if (value === "undefined") return undefined;
+        if (!isNaN(+value) && !value.includes('"')) return parseInt(value);
```
babiabeo, line 307: Why not just use `!isNaN(value)`?

WasixXD: There's no particular reason, I just put it that way out of habit

kt3k: This allows expression like `1e3` as parsed as number, and it results `1`. Is that intentional?

kt3k:
Can you also work on `stringify`?

WasixXD:
yeah sure no problem

ini/_ini_map.ts
babiabeo: > Edit: I just checked and isNaN expects a number, so there's no way to use `!isNaN(value)`

Oh you're right. I forgot we're using typescript not javascript lol

WasixXD: If its a number but has a `"` the user wants as a string and not as a number

babiabeo: Fair enough

kt3k (approved):
I think these value interpretations are fine because they are very natural in JavaScript, and I guess they probably gave least surprise to the users. LGTM

Note: There's no official specifications for INI format, and major parsers are very incompatible with each other. So we need to choose one way of parsing them, and I find this one is enough reasonable.

iuioiua:
AFAIK, this functionality is all that remains for stabilising `ini`.

kt3k:
Discussed this internally and we'd like to drop the interpretation of `undefined` literal as there's no other known parsers which parse `"undefined"` to `undefined`. (I pushed the commit https://github.com/denoland/std/pull/6121/commits/915ec9425112b23ecf5744a6140ead8d22710b96 to this branch)

What do you think?

WasixXD:
This comes closer to what `JSON.parse` do, but then shouldn't `stringify({ value: undefined })` return `{}`  or nothing at all?

kt3k:
Ok. Let's go with this

Issue #5848, opened 2024-08-28
testing/mock.ts: Add support for stubbing properties


KyleJune:
**Is your feature request related to a problem? Please describe.**

Currently the mocking tool doesn't support stubbing getter/setter properties. It only allows stubbing functions. Other mocking tools like sinon do support it.

It's probably not  a common need and would just be a nice to have. I'm creating this issue based on a comment someone made in the discord's #dev-std channel where they were requesting a way to stub a getter.

**Describe the solution you'd like**

The ideal solution to this problem to me would be to add an additional call signature to the stub function, that allows you to pass a property descriptor instead of a function as the third argument.

```js
stub<Self, Prop extends keyof Self>(self: Self, property: Prop, descriptor: Omit<PropertyDescriptor, 'configurable'>)
```

It would need to omit the configurable property since a property must be configurable for it to be able to be restored to it's original value.

If a get or set function are set on the descriptor, they should be wrapped in a spy so that you can make assertions about them being called. You would just need a way to access them. Maybe the value returned by the stub function could have get and set properties on it so that people can access the spys by doing `myStub.get` or `myStub.set` where `myStub` is the value returned by the stub function when you call it with a property descriptor instead of a function.

For more information about property descriptors, see [Object.defineProperty](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) documentation on mdn.

**Describe alternatives you've considered**

With sinon.js, you can mock properties by creating a stub then calling the get, set, or value functions on the stub object. That can be seen at the end of the following page. I'm guessing it's at the end because it's uncommon to use it, but  it would be good to have for when people want to do this.

https://sinonjs.org/releases/latest/stubs/

Here is an example of stubbing a getter with sinon.

```js
const sinon = require("sinon");
const referee = require("@sinonjs/referee");
const assert = referee.assert;

describe("stub", function () {
    it("should replace getter", function () {
        const myObj = {
            prop: "foo",
        };

        sinon.stub(myObj, "prop").get(function getterFn() {
            return "bar";
        });

        assert.equals(myObj.prop, "bar");
    });
});
```

PR #6128, opened 2024-10-19
feat(testing/unstable): support for stubbing properties

IgorM867:
This PR adds support for stubbing properties using stub() function which is mentioned in #5848. 
Example:
```ts
const obj = {
  prop: "foo",
};

const getterStub = stub(obj, "prop", {
  get: function () {
    return "bar";
  },
});

assertEquals(obj.prop, "bar");
assertSpyCalls(getterStub.get, 1);
```
Note:
I am new to open source and eager to contribute and learn. Apologies in advance if this PR wastes you time.


kt3k:
Looks like a good idea to me, but can you move this `stub` to `unstable_stub.ts` (and `unstable_stub_test.ts`) and existing `mock.ts` untouched?

IgorM867:
Sorry for the many commits. Just to confirm: did you mean for me to move the new stub implementation to unstable_stub.ts and keep the old version in mock.ts?

I also encountered a lint error: exported symbol is missing JSDoc documentation. This happens because I need to export registerMock and unregisterMock to use them in unstable_stub.ts. Should I add JSDoc comments for these functions? Currently, they look like this:
```ts
export function registerMock(spy: Spy<any, any[], any>) {
  const session = getSession();
  session.add(spy);
}
export function unregisterMock(spy: Spy<any, any[], any>) {
  const session = getSession();
  session.delete(spy);
}
```

kt3k:
> Sorry for the many commits. Just to confirm: did you mean for me to move the new stub implementation to unstable_stub.ts and keep the old version in mock.ts?

Yes, that is correct. The current structure looks mostly good to me.

> I also encountered a lint error: exported symbol is missing JSDoc documentation. This happens because I need to export registerMock and unregisterMock to use them in unstable_stub.ts. Should I add JSDoc comments for these functions? Currently, they look like this:

We don't want to expose them as public APIs. Can you move necessary items to some new file like `_mock_utils.ts` and importing them from both `mock.ts` (existing one) and `unstable_stub.ts`?

IgorM867:
One check failed, but it’s due to `unstable_throttle.ts`, so I don’t think it’s related to my changes.

kt3k (approved):
LGTM. Thanks for updating!

testing/unstable_stub_test.ts
```diff
@@ -0,0 +1,394 @@
...
+  assertSpyCalls(func.get, 2);
+
+  assertEquals(func.restored, false);
+  func.restore();
+  assertEquals(func.restored, true);
+  assertEquals(point.x, 5);
+  assertThrows(
+    () => func.restore(),
+    MockError,
+    "Cannot restore: instance method already restored",
```
KyleJune, line 121: One minor comment, this says an instance method was restored when it was actually a property in this case. We could probably determine which it is by looking at the type of the value. Also I'd suggest renaming func to prop in this test case since it is stubbing a property rather than a func.

testing/unstable_stub_test.ts
```diff
@@ -0,0 +1,394 @@
...
+    MockError,
+    "Cannot stub: already spying on instance method",
+  );
+});
+
+Deno.test("stub() throws when neither setter not getter is defined", () => {
+  const obj = { prop: "foo" };
+
+  assertThrows(
+    () => stub(obj, "prop", {}),
```
KyleJune, line 380: I think it would make sense to support an empty property descriptor for when you want it to start with the current value. I think we should also support passing a value option instead of a getter/setter in cases where they want to both spy on and override the value. Without this suggested change, you'd have to create getters and setters like this, where x could be the current value or the value you want to override it with.

Without this suggestion:

```ts
const point = new Point(5, 6);
let x = point.x;
const prop = stub(point, "x", {
  get: function () {
    return x;
  },
  set: function (value: number) {
    x = value;
  },
});
```

With this suggestion when you want to spy without changing the initial value you would do this:

```ts
const point = new Point(5, 6);
const prop = stub(point, "x", {});
```

If possible to get the typing right, it would be even cooler if you could stub a property like you can stub a function, without the third arg.

```ts
const point = new Point(5, 6);
const prop = stub(point, "x");
```

With this suggestion when you want to spy while also changing the initial value:

```ts
const point = new Point(5, 6);
const prop = stub(point, "x", { value: 3 });
```

With this suggestion, all these code samples would be functionally equivalent, where a fake getter/setter are created for the property, where you can make assertions about them. When restored, the property descriptor would be changed back to what it was originally.

testing/unstable_stub_test.ts
```diff
@@ -0,0 +1,394 @@
...
+  assertEquals(func.restored, true);
+  assertEquals(point.action, Point.prototype.action);
+  assertThrows(
+    () => func.restore(),
+    MockError,
+    "Cannot restore: instance method already restored",
+  );
+  assertEquals(func.restored, true);
+});
+Deno.test("stub() works on getter", () => {
```
KyleJune, line 87: Technically, on the point class, both x and y are values. We are overriding the property descriptor with one that has a getter instead of a value. Same with the other tests. I would describe this, "stub() property value with a getter".

We should get a test case that shows stubbing a property that originally had a getter, setter, or both works too. In all cases, restore should change the property descriptor to what it was before we stubbed it.

KyleJune: For that last part, I think the test cases should be updated to first get a copy of the property descriptor for point.x, then at the end after restoring, assert that the property descriptor for point.x matches what it was originally.

testing/unstable_stub_test.ts
KyleJune: On this point, I think another improvement that would be cool is to support spying on a property. Where the difference would basically be that at the end, if the original property descriptor had a value instead of a getter/setter, we would be changing the property descriptor to having the value be whatever we last set it to while it was being spied on. While a stub just always returns it back to it's original value.

PR #6135, opened 2024-10-22
refactor(fs): uncaught FS errors in web

### LINKED ISSUE #6132: fs.ensure_symlink throwing error in browser
**Describe the bug**

On [line 8 of ensure_symlink.ts](https://github.com/denoland/std/blob/a541fb468ccf67787cf28e3fd954ebcc2035fc46/fs/ensure_symlink.ts#L8) on-import the script checks if the user is on windows (`const isWindows = Deno.build.os === "windows";`). Deno however is undefined in browser throwing a fatal error.
image

**Steps to Reproduce**

1. Build project with:
```
import * as esbuild from "npm:esbuild";
await esbuild.build({ plugins: [...denoPlugins()], entryPoints: ["./src/in.ts"], outfile: "./dist/out.js", bundle: true, format: "esm", platform: "browser" });
```
2. Import in web:
```
    <script type="module" src="hydrafiles-web.esm.js"></script>
    <script type="module">
    import Hydrafiles from './hydrafiles-web.esm.js';
    </script>
```
3. Error in browser logs

**Expected behavior**

A check should be done before `Deno.build.os === "windows"` to first check if user is in Deno environment. For example:
```
const isWindows = window !== undefined && Deno.build.os === "windows";
```

**Environment**

- OS: N/A
- deno version: 2.0
- std version: 0.221.0

QuixThe2nd:
Fixes #6132

### DIFF
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -5,7 +5,7 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, type PathType } from "./_get_file_info_type.ts";
 import { toPathString } from "./_to_path_string.ts";
 
-const isWindows = Deno.build.os === "windows";
+const isWindows = typeof window === "undefined" && Deno.build.os === "windows";
 
 function resolveSymlinkTarget(target: string | URL, linkName: string | URL) {
   if (typeof target !== "string") return target; // URL is always absolute path
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -5,7 +5,7 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, type PathType } from "./_get_file_info_type.ts";
 import { toPathString } from "./_to_path_string.ts";
 
-const isWindows = typeof window === "undefined" && Deno.build.os === "windows";
+const isWindows = (globalThis as any).Deno?.build.os === "windows";
 
 function resolveSymlinkTarget(target: string | URL, linkName: string | URL) {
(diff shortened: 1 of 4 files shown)


fs/ensure_symlink.ts
```diff
@@ -5,7 +5,7 @@ import { ensureDir, ensureDirSync } from "./ensure_dir.ts";
 import { getFileInfoType, type PathType } from "./_get_file_info_type.ts";
 import { toPathString } from "./_to_path_string.ts";
 
-const isWindows = Deno.build.os === "windows";
+const isWindows = typeof window === "undefined" && Deno.build.os === "windows";
```
kt3k, line 8: `window` object is defined in Deno 1.x, and I think it's probably too early to stop supporting Deno v1.x.

Let's do something like `(globalThis as any).Deno?.build.os === "windows"` instead, similarly to `@std/path` package. ref. https://github.com/denoland/std/blob/ae7048f9ca983f038174aca4cd89c120945aef23/path/_os.ts#L7C3-L7C51

QuixThe2nd: Done

QuixThe2nd: Is `as any` required? It's causing tests to fail.

kt3k:
@QuixThe2nd Can you sign the CLA from the above link?

kt3k (approved):
LGTM. Thanks!

The current CI (lint) failure is irrelevant to this change. The maintainers will fix them and land this PR after that.

QuixThe2nd:
> @QuixThe2nd Can you sign the CLA from the above link?

Done! Thanks for dealing with this so fast :)

QuixThe2nd:
I noticed a few other files had the same issue.

PR #6148, opened 2024-10-27
feat(kv): add utilities for Deno KV

### LINKED ISSUE #6123: Proposal: Add @std/kv

While Deno KV is great, there are likely some additional capabilities that wouldn't be integrated into the runtime, but would be generally useful for those working with Deno KV.

I have been making [kv-toolbox](https://jsr.io/@kitsonk/kv-toolbox) available for over a year and I think there are parts of it that are suitable for becoming part of the Deno standard library.

I propose that `@std/kv` be created, initially with the following areas of capability:

- [kv-toolbox/json](https://jsr.io/@kitsonk/kv-toolbox/doc/json) - a set of utilities and types that are able to serialize and deserialize all of the keys and values that are supported by Deno KV in JSON. This is very useful when having to represent Deno KV data in a browser or transmit Deno KV data "over the wire" while being able to accurately preserve the item.
- [kv-toolbox/size_of](https://jsr.io/@kitsonk/kv-toolbox/doc/size_of) - a function that estimates the byte size of a serialized value in Deno KV. Currently the [documentation](https://docs.deno.com/api/deno/~/Deno.Kv) suggests to use `JSON.stringify()` to estimate the size of values. The problem is that is totally ineffective at estimating complex objects which can be stored in Deno KV but do not serialize properly to JSON (e.g. typed arrays, RegExp, Set, Map, Error, Date, bigint, etc.). While taking the byte length of V8 serialize is more accurate it isn't available in Deploy and is 10x slower than `sizeOf()`.

These capabilities would be fairly straightforward contributions, are fairly well documented and have pretty extensive test coverage.

There is a lot more to `kv-toolbox`, but my opinion is that aren't as generally applicable as the above areas, work around potential future features of Deno KV, or promote patterns that might come with risks, although I would be glad to contribute any part of `kv-toolbox` to Deno.

Just continuing to maintain `kv-toolbox`, but limiting the usefulness to a wider audience.

kitsonk:
This provides `@std/kv/json` and `@std/kv/size_of` which are utilities which enhances the utility of Deno KV.

Closes #6123

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,6 +1,6 @@
 {
   "deno.enable": true,
-  "deno.unstable": true,
+  "deno.unstable": ["kv"],
   "editor.formatOnSave": true,
   "editor.defaultFormatter": "denoland.vscode-deno"
 }
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -71,6 +71,7 @@ const ENTRY_POINTS = [
   "../io/mod.ts",
   "../json/mod.ts",
   "../jsonc/mod.ts",
+  "../kv/mod.ts",
   "../log/base_handler.ts",
   "../log/warn.ts",
   "../log/critical.ts",
--- a/browser-compat.tsconfig.json
+++ b/browser-compat.tsconfig.json
@@ -33,6 +33,7 @@
     "./io",
     "./json",
     "./jsonc",
+    "./kv",
     "./log",
     "./media_types",
     "./msgpack",
--- a/deno.json
+++ b/deno.json
@@ -8,7 +8,7 @@
   },
   "importMap": "./import_map.json",
   "tasks": {
-    "test": "deno test --unstable-http --unstable-webgpu --doc --allow-all --parallel --coverage --trace-leaks --clean",
+    "test": "deno test --unstable-http --unstable-kv --unstable-webgpu --doc --allow-all --parallel --coverage --trace-leaks --clean",
     "test:browser": "git grep --name-only \"This module is browser compatible.\" | grep -v deno.json | grep -v .github/workflows | grep -v _tools | grep -v encoding/README.md | grep -v media_types/vendor/update.ts | xargs deno check --config browser-compat.tsconfig.json",
     "test:node": "node --import ./_tools/node_test_runner/register_deno_shim.mjs ./_tools/node_test_runner/run_test.mjs",
     "fmt:licence-headers": "deno run --allow-read --allow-write ./_tools/check_licence.ts",
@@ -75,6 +75,7 @@
     "./io",
     "./json",
     "./jsonc",
+    "./kv",
     "./log",
     "./media_types",
     "./msgpack",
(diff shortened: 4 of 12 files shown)


ryuapp:
It's good that these utilities are managed by Deno organization.
However, wouldn't it be more reasonable to manage it with `@deno/kv`?

`@std` is the Deno standard libraries and except for packages related to File I/O, `@std` can be said to be web standards.
However the kv is Deno-only feature. Since `@deno/kv-oauth` is also managed by `@deno`, wouldn't it be better to use `@deno` scope?

kitsonk:
These libraries run perfectly fine in a browser or other runtimes.

Specifically the JSON module is how I have built tooling to display and manage Deno KV stores.

You can connect to Deno KV instances from Node.js using the [`@deno/kv`](https://www.npmjs.com/package/@deno/kv) npm package, of which this package would work with as well. So to say it is "Deno only" isn't accurate about Deno KV.

kt3k:
Discussed this internally with KV team, and we'd prefer to have `json` part of this PR in `@deno/`-namespaced package. While it would be useful for exchanging data across runtimes, the serialized JSON format is not officially specified one, but specific to this implementation.

Could you separate `json` part of this PR and send it to this repository? https://github.com/denoland/kv-utils

kt3k:
For `sizeOf` part, we don't understand the valid/effective usage of it yet. By using `sizeOf` utility, the users will face the limitation of the input like "You can write text until your data hit the Deno KV limit". We found this kind of limitation unreasonable as it's difficult for the users to know how much their data cost in Deno KV. The users should rather see the limitation like "You can write text until 50KB". To our current understanding, `sizeOf` feels like promoting a usage of KV in an undesirable way. (But we might be missing some other valid usage)

kitsonk:
@kt3k for those creating derivative libraries or capabilities on Deno KV it becomes more important. For example in kv-toolbox, I use `sizeOf` to be able to create batches of atomic transactions.

While it maybe logical to provide a user with something like "you can upload 50k of text", how big is an object, `Set`, `Map` or `Array` when stored? `sizeOf()` can give you a fairly accurate estimate. Specifically:

> We found this kind of limitation unreasonable as it's difficult for the users to know how much their data cost in Deno KV.

Is exactly what `sizeOf()` solves, a very low cost way to get an accurate estimate of the data cost, which can be either incorporated in code or used to provide feedback to application users in a browser for example.

I originally just used it internally in kv-toolbox, but realised it might be useful to other library others and exposed it. I do agree it may not reach the level of a standard library, but it is invaluable to know ahead of time the data size of a value or more importantly, the size of a transaction to ensure it is infallible. If it was decided not to include it, that is fine.

kt3k:
Ok. We'd accept it, but we'd like to change the API name to `estimateSize` as the serialization format is not officially part of V8 specification (It could be changed by upstream decision).

Also we'd like to accept this PR in `@deno`-namespaced package instead of `@std` package (because both APIs are not based on open spec). Could you move this PR to this repo? https://github.com/denoland/kv-utils (which is currently empty, but we'll publish it to JSR after accepting the PR).

kitsonk:
@kt3k excellent and will do!

PR #6155, opened 2024-10-28
docs(streams): rest arguments not being asserted in docs

efekrskl:
Currently, rest arguments are not being asserted when `lint:docs` is run. This PR fixes the issue and adds missing `@param`   field to docs.


kt3k (approved):
Good find! Thanks!

Issue #5439, opened 2024-07-14
to-do: add doc check for JSDoc tags that aren't needed


iuioiua:
I.e. If there's a `@param` tag for an argument, the [doc checker](https://github.com/denoland/deno_std/blob/main/_tools/check_docs.ts) should flag that. #5438 gave me this idea.

PR #6158, opened 2024-10-29
refactor(path): always name the parameters (add param definition check in doc linter)

efekrskl:
Implements #5439

This implementation doesn't go well with unnamed parameters (see the other changes in `path/*`). IMO, this is ok. As named parameters and Jsdoc param names go hand in hand, especially considering these are public APIs. Also, immediately on implementation, this helped me uncover some unused fields in the same functions. I'm curious on your opinions on this.

image

### DIFF
--- a/path/posix/join_globs.ts
+++ b/path/posix/join_globs.ts
@@ -26,8 +26,9 @@ export type { GlobOptions };
  */
 export function joinGlobs(
   globs: string[],
-  { extended = true, globstar = false }: GlobOptions = {},
+  options: Pick<GlobOptions, "globstar"> = {},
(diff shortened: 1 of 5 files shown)


kt3k (approved):
LGTM. Nice! Thanks

path/posix/join_globs.ts
```diff
@@ -26,8 +26,9 @@ export type { GlobOptions };
  */
 export function joinGlobs(
   globs: string[],
-  { extended = true, globstar = false }: GlobOptions = {},
```
kt3k, line 29: Good to find unused parameter 👍

Issue #6152, opened 2024-10-28
@std/expect toStrictEqual() on objects mis-compute their length


jerome-benoit:
**Steps to Reproduce**

Migration to @std/expect with the failures: 
https://github.com/poolifier/poolifier-web-worker/actions/runs/11553093927/job/32153534099

**Expected behavior**

Objects strict comparaison should work. 

**Environment**

N/A

kt3k:
@eryue0220 Can you take a look?

eryue0220:
> @eryue0220 Can you take a look?

Sure, I'll check this later.

Issue #6111, opened 2024-10-12
toMatchObject doesn't support matchers but only exact values


NaveenNeti:
Thank you to all for the work on this project.

The JSR page for @std/expect states "Jest compatible `expect` assertion functions"

While trying to use `toMatchObject` I noticed it doesn't seem to support asymmetric matchers similar to what jest would allow.

e.g. within Jest

```typescript
  expect({ position: { x: 0, y: 0 } }).toMatchObject({
    position: {
      x: expect.any(Number),
      y: expect.any(Number)
    }
  });
```

Running the same test in Deno:

```typescript
Deno.test("CustomMatcher tests", () => {
    expect({ position: { x: 0, y: 0 } }).toMatchObject({
        position: {
            x: expect.any(Number),
            y: expect.any(Number),
        },
    });
});
```

Results in the output:
```
error: AssertionError: Values are not equal.
    [Diff] Actual / Expected
    {
      position: {
+       x: {
+         value: [Function: Number],
+       },
+       y: {
+         value: [Function: Number],
+       },
-       x: 0,
-       y: 0,
      },
    }
```

**Describe the solution you'd like**

- Would be great if the `toMatchObject` would have consistent behavior with jest and allow for asymmetric matchers
- From what I can see it would require edits to:
https://jsr.io/@std/expect/1.0.5/_matchers.ts#L461
https://jsr.io/@std/expect/1.0.5/_matchers.ts#L487

Would be keen to help provide an MR to add support for it assuming the project owners see value.

**Versions**
```
deno: 2.0.0
@std/assert/1.0.6
```

eryue0220:
I'll check this later.

NaveenNeti:
Thank you @eryue0220 I will test this tomorrow 🙏

NaveenNeti:
Confirmed this is fixed @eryue0220 Thank you.

PR #6160, opened 2024-10-30
fix(expect): re-align `expect.toMatchObject` api

eryue0220:
This pr is to re implement `expect.toMatchObject` api.

Close: #6152 and #6111


kt3k (approved):
LGTM. Thank you! @eryue0220

PR #6163, opened 2024-10-31
test(bytes): document the cases being tested for equals/startsWith/endsWith

lionel-rowe:
Current tests are kinda confusing and unclear what's being tested. It could help resolve confusion like https://github.com/denoland/std/issues/6146 if there was a specific test case (`` startsWith() ... `true` where `source` starts with `prefix` ``) that could be pointed to as failing upon the proposed change.


kt3k (approved):
LGTM. Thanks!

PR #6169, opened 2024-11-02
feat(media-types): update media-types from latest mime-db

akmittal:
This change updates media_types by running `deno run media_types/vendor/update.ts`

This helps adding new mimetypes (I was specially looking for jxl image mimetype)


kt3k (approved):
LGTM. Thanks!

PR #6175, opened 2024-11-05
docs(log): document FileHandler

efekrskl:
Documents `FileHandler` for #3764

### DIFF
--- a/log/file_handler.ts
+++ b/log/file_handler.ts
@@ -185,8 +185,10 @@ export class FileHandler extends BaseHandler {
    * @example Usage
    * ```ts no-assert
    * import { FileHandler } from "@std/log/file-handler";
+   *
    * const handler = new FileHandler("INFO", { filename: "./logs.txt" });
    * handler.setup(); // Opens the file and prepares the handler for logging.
+   * handler.destroy();
    * ```
    */
   override setup() {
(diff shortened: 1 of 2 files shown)


log/file_handler.ts
```diff
@@ -30,35 +44,126 @@ export interface FileHandlerOptions extends BaseHandlerOptions {
...
+   * import { FileHandler } from "@std/log/file-handler";
+   *
+   * const handler = new FileHandler("INFO", { filename: "./logs.txt" });
+   * handler.setup();
+   * handler.log('Hello, world!'); // Buffers the message, or writes it to the file depending on buffer state
+   * handler.flush(); // Manually flushes the buffer
+   * handler.destroy(); // Closes the file and removes listeners
+   * ```
+   * **/
   [encoderSymbol]: TextEncoder = new TextEncoder();
```
efekrskl, line 156: Some of these symbols should probably be private properties, but since they are not, linting requires them to be documented still. I'm not sure how to handle this better.

kt3k (approved):
LGTM. Thanks!

log/file_handler.ts
kt3k: Probably that's the issue of linter (and it could be also an issue of doc generator), but I think it's fine for now for this PR

PR #6176, opened 2024-11-06
docs(log): improve logger.ts docs

efekrskl:
Gets us closer to #3764 

The file was already very well documented, it was just missing a few things

### DIFF
--- a/log/logger.ts
+++ b/log/logger.ts
@@ -22,15 +22,98 @@ export interface LogRecordOptions {
   loggerName: string;
 }
 
+/**
+ * Configuration options for a logger instance.
+ *
+ * @example Usage
+ *
+ * ```ts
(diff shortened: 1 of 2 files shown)


log/logger.ts
```diff
@@ -22,15 +22,98 @@ export interface LogRecordOptions {
...
+ *
+ * @example Usage
+ *
+ * ```ts
+ * import { ConsoleHandler, getLogger, setup, type LogConfig} from "@std/log";
+ * import { assert } from "@std/assert";
+ *
+ * const handler = new ConsoleHandler("INFO");
+ * const logConfig: LogConfig = {
+ *     handlers: {
```
kt3k, line 36: We generally prefer 2-space indent here, but not a big deal

kt3k (approved):
LGTM. Thanks!

PR #6183, opened 2024-11-13
chore(log): do not exec file handler examples in CI

kt3k:
`std/log/file-handler` examples generate `logs.txt` and it makes the repository modified state. That leads to the publishing failure https://github.com/denoland/std/actions/runs/11815536146/job/32916891866

This PR fixes the situation by ignoring examples in doc testing.


satyarohith (approved):
👍

PR #6184, opened 2024-11-13
chore: check publish --dry-run after testing

kt3k:
This PR updates the CI workflow to check `publish --dry-run` after test runs.

This prevents the case like 2024.11.13 release ( https://github.com/denoland/std/actions/runs/11815536146/job/32916891866 ), which failed because the git repository state became dirty after the test run and CI didn't detect it


satyarohith (approved):
Good precautionary measure.

LGTM.

PR #6190, opened 2024-11-16
feat(cli/unstable): add `promptSelect()`

timreichen:
Ref: https://github.com/denoland/std/issues/4678

**Changes**
Adds `promptSelect()` cli function.


kt3k:
- Can you rename the file to `unstable_prompt_select.ts` as we want to accept every new feature as unstable feature? ref: https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100
- Can you try to write test cases by spying/stubing methods of `Deno.stdin` and `Deno.stdout`? Mocking utilities are available from `@std/testing/mock`

timreichen:
> * Can you rename the file to `unstable_prompt_select.ts` as we want to accept every new feature as unstable feature? ref: https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100
> * Can you try to write test cases by spying/stubing methods of `Deno.stdin` and `Deno.stdout`? Mocking utilities are available from `@std/testing/mock`

I've added some tests but need help resolving the current errors. The tests ran successfully locally. I don't know, why they occur and how to fix them.

kt3k:
> The tests ran successfully locally. I don't know, why they occur and how to fix them.

I think you also need to stub `Deno.stdin.setRaw` (with empty function) to avoid error. (The error in CI seems happening because the stdin is not terminal in CI)

cli/unstable_prompt_select_test.ts
```diff
@@ -0,0 +1,359 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+
+import { assertEquals } from "../assert/equals.ts";
+import { promptSelect } from "./unstable_prompt_select.ts";
+
+const encoder = new TextEncoder();
+const decoder = new TextDecoder();
+
+function spyFn<T, K extends keyof T>(
```
kt3k, line 9: Why create original spy util here? What was wrong with `spy` in `@std/testing/mock`?

timreichen: `spy` didn't work for that purpose, but `@std/testing/mock`'s `stub()` did it 👍

Kacaii:
Im really excited for this one! Its exactly what I've been looking for <3

cli/unstable_prompt_select.ts
```diff
@@ -0,0 +1,76 @@
...
+    for (const [index, value] of values.entries()) {
+      const start = index === selectedIndex ? INDICATOR : PADDING;
+      Deno.stdout.writeSync(encoder.encode(`${start} ${value}\r\n`));
+    }
+    const n = Deno.stdin.readSync(buffer);
+    if (n === null || n === 0) break;
+    const input = decoder.decode(buffer.slice(0, n));
+    switch (input) {
+      case ETX:
+        return Deno.exit(0);
```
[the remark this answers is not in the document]
timreichen: This allows to exit the prompt with Ctrl + C.

cli/unstable_prompt_select.ts
```diff
@@ -0,0 +1,76 @@
...
+        selectedIndex = (selectedIndex + 1) % length;
+        break;
+      case CR:
+        break loop;
+    }
+    Deno.stdout.writeSync(encoder.encode(`\x1b[1A${CLR}`.repeat(length)));
+  }
+  if (clear) {
+    Deno.stdout.writeSync(encoder.encode(`\x1b[1A${CLR}`.repeat(length + 1))); // clear values and message
+  }
```
kt3k, line 73: I personally don't remember any tool which doesn't clear the selections in this context. How about always clearing values and message?

cli/unstable_prompt_select.ts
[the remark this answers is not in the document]
kt3k: Ah, makes sense!

cli/unstable_prompt_select.ts
timreichen: the built in `prompt()`doesn't clear.
This option is taken over from `promptSecret()`. I think if it is there, we should provide it here also.

kt3k: I mean I've never seen a tool which doesn't clear the choices part of the prompt. For example, `npm init vite` prompts several select-type prompts, but choices areas are always cleared after the selection. Leaving choices as is doesn't look like sensible default behavior to me.

kt3k: If we default to leave something on the screen after the selection, then I think the output should become something like:

```
Please select a browser: safari
```

instead of leaving the entire selection section as is, like:

```
Please select a browser:
❯ safari
  chrome
  firefox
```

(for the call `promptSelect("Please select a browser:", ["safari", "chrome", "firefox"])`

timreichen: I see. In that case I don't know what the best default is. I like the idea of not clearing the lines by default so the history remains and aligns with `prompt()` and `promptSecret()` but it doesn't make much sense to print a different line imo...

kt3k: hmm, ok. Let's keep this as is

kt3k (approved):
The tests seem nicely written. LGTM

cli/unstable_prompt_select.ts
timreichen: 👍 since it is unstable, we can adjust according to user feedback.

PR #6195, opened 2024-11-20
docs: simplify README

iuioiua:
This PR simplifies the README document by having resources be listed in a "Resources" section. It also removes the package table, as now, we can directly link to JSR, which contains package version information.


kt3k (approved):
Looks reasonable to me. Thanks!

PR #6198, opened 2024-11-20
fix(fs): improve the error message when the symlink target does not exist

### LINKED ISSUE #6191: fs.ensureSymlink does not work!
**Describe the bug**
`ensureSymlink` method does not work! Well... Mostly...
I'm new to using Deno and it's std library and I was trying out @std/fs and that's when I encountered an error with my very basic code...

```typescript
await fs.ensureSymlink("./dir/test.txt", "./dir/test.link.txt");
```
Error:
```
error: Uncaught (in promise) NotFound: No such file or directory (os error 2): lstat '/home/arb/Programming/Javascript/Practice/deno-hello-world/dir/dir/test.txt'
  const srcStatInfo = await Deno.lstat(targetRealPath);
                      ^
    at async Object.lstat (ext:deno_fs/30_fs.js:397:15)
    at async Module.ensureSymlink (https://jsr.io/@std/fs/1.0.5/ensure_symlink.ts:56:23)
    at async file:///home/arb/Programming/Javascript/Practice/deno-hello-world/main.ts:6:1
```

As you can see the target file exists and i'm trying to create the symlink inside the same directory as the target file:
Screenshot From 2024-11-17 21-38-41

I tried removing the `./` from the paths, but it didn't work, and I got the same error again.
The funny part is that if the target file and the path for creating symlink are both at the root of the working directory, The method work! So this line of code works with no problem:

```typescript
await fs.ensureSymlink("test.txt", "test.link.txt");
```

One more thing that i tried that also did not work was this:
```typescript
await fs.ensureSymlink("test.txt", "new-dir/test.link.txt");
```

I even made the `new-dir` directory myself but i still got the same error.

**Steps to Reproduce**

I'm not doing anything special!
1. Create a new project using `deno init`
2. Change content of `main.ts` to this:
```typescript
import * as fs from "@std/fs";

await fs.ensureSymlink("./dir/test.txt", "./dir/test.link.txt");
``` 
3. Create a file with this path: `./dir/test.txt`

**Expected behavior**

Creating a new symlink?

**Environment**

- OS: Fedora 41
- deno version: 2.0.6 (Stable)
- std version: @std/fs 1.0.5

kt3k:
This PR improves the error message of `fs.ensureSymlink` when the link target doesn't exist.

closes #6191

### DIFF
--- a/fs/ensure_symlink.ts
+++ b/fs/ensure_symlink.ts
@@ -53,7 +53,17 @@ export async function ensureSymlink(
   linkName: string | URL,
 ) {
   const targetRealPath = resolveSymlinkTarget(target, linkName);
-  const srcStatInfo = await Deno.lstat(targetRealPath);
+  let srcStatInfo;
+  try {
+    srcStatInfo = await Deno.lstat(targetRealPath);
+  } catch (error) {
+    if (error instanceof Deno.errors.NotFound) {
+      throw new Deno.errors.NotFound(
+        `Cannot ensure symlink as the target path does not exist: ${targetRealPath}`,
+      );
+    }
+    throw error;
+  }
   const srcFilePathType = getFileInfoType(srcStatInfo);
 
   await ensureDir(dirname(toPathString(linkName)));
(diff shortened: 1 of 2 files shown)


satyarohith (approved):
LGTM

Nit: update `target` param description that it should be relative to dirname of `linkName` when it's a string?

PR #6199, opened 2024-11-21
fix(tar): untar checksum calculation for the pax format

BlackAsLight:
Fixes: https://github.com/denoland/std/issues/6083

After further investigation it seems the problem tarball was using 7 octal digits to encode the checksum instead of 6, like the spec this implementation is based off says to use. Other [tar specs](https://www.gnu.org/software/tar/manual/html_node/Standard.html) don't seem to make any comment on only using 6 octal digits. So for more flexibility this spec just parses the entire 8 bytes instead of the first 6.


kt3k:
Could you add a test case for this fix?

kt3k (approved):
I also confirmed the example archives https://github.com/polyseam/cndi/releases/download/v2.20.0/cndi-mac.tar.gz and https://github.com/themesberg/flowbite-icons/archive/refs/tags/v1.3.0.tar.gz can be extracted with this change. Nice work!

PR #6201, opened 2024-11-22
feat(testing): add context to beforeEach and afterEach hooks

### LINKED ISSUE #6189: Testing: add Deno.TestContext as a parameter to beforeEach and afterEach
I have some shared preparation before each test and it requires the test name and test file. This is available in the Deno.TestContext object passed to each test.

It would be very helpful if this object would also be available in the `beforeEach` and `afterEach` hooks.

Mr-Duda:
Fixes #6189

This allows the `beforeEach` and `afterEach` methods to use the data provided in each test context object. 

I wanna build a small decorator based test framework on top of deno testing. The usecase is the following: 
- before each test `/path/TestCaseTest/testFileNotExistsThrowsError/{in,out}` folders are created if needed
- `it`
  - reads some files from `.../testFileNotExistsThrowsError/in` 
  -  creates some files in `.../testFileNotExistsThrowsError/out` using a custom context wrapper
- after each test the `.../testFileNotExistsThrowsError/out` is cleaned up to prevent unwanted outputs

```js

@TestSuite
class TestCaseTest extends TestCase {

    @Test
    public testFileNotExistsThrowsError(context: iTestContext) {
        // Read files...
        context.createFile("./dummy.txt", "dummy")
        return this.assertThrows(() => this.assertFileNotExists("./dummy.txt", context))
    }
}
```

Currently the `beforeEach` and `afterEach` don't have access to the test name and it's somehow difficult to get this information.
The test source file would also be needed at some point and i would't use `__filename` since it's also provided in the test context.

### DIFF
--- a/testing/_test_suite.ts
+++ b/testing/_test_suite.ts
@@ -25,12 +25,12 @@ export interface DescribeDefinition<T> extends Omit<Deno.TestDefinition, "fn"> {
     | ((this: T) => void | Promise<void>)[];
   /** Run some shared setup before each test in the suite. */
   beforeEach?:
-    | ((this: T) => void | Promise<void>)
-    | ((this: T) => void | Promise<void>)[];
+    | ((this: T, context: Deno.TestContext) => void | Promise<void>)
+    | ((this: T, context: Deno.TestContext) => void | Promise<void>)[];
   /** Run some shared teardown after each test in the suite. */
   afterEach?:
-    | ((this: T) => void | Promise<void>)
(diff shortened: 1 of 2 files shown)


kt3k:
Why not just use native `Deno.test()` and `t.step()`? These APIs should be more flexible than `@std/testing/bdd` and they should be more suitable for building testing framework on top of

Mr-Duda:
I shitched to using the Deno namespace and it works fine.

PR #6204, opened 2024-11-22
fix(text): use locale-independent letter case methods (#6016)

### LINKED ISSUE #6016: Use of locale-sensitive methods with `undefined` locale may cause environment-sensitive bugs
**Describe the bug**

I've mentioned this before tangentially in a couple of issues ([ex](https://github.com/denoland/std/issues/5424#issuecomment-2232291988)), but it's probably worth its own issue.

Use of locale-sensitive methods, such as `toLocaleLowerCase`, with an unspecified (`undefined`) locale may cause environment-sensitive bugs, because the default locale is host defined.

**Steps to Reproduce**

For example, in `http/cookie.ts`, which is marked browser-compatible:

https://github.com/denoland/std/blob/6a4eb6cb9185b0b9e0816a227c04b5bf039b99ef/http/cookie.ts#L361

1. Open Firefox
2. Go to https://codepen.io/lionel-rowe/pen/rNEXmoG?editors=0010:
    ```js
    import { getSetCookies } from 'https://esm.sh/jsr/@std/http@1.0.6/cookie'
    
    const setCookie = 'a=b; EXPIRES=Thu, 19 Sep 2024 07:47:28 GMT'
    const headers = new Headers({ 'set-cookie': setCookie })
    const cookie = getSetCookies(headers)[0]
                                
    document.querySelector('#target').textContent = JSON.stringify(cookie, null, 4)
    ```
4. Observe expected output (assuming your locale isn't Turkish):
    ```
    {
        "name": "a",
        "value": "b",
        "expires": "2024-09-19T07:47:28.000Z"
    }
    ```
5. Go to `about:preferences`
6. Set language to Turkish ("Türkçe"; Turkish for "language" is "dil" if you need to find the setting again)
7. Go back to or refresh https://codepen.io/lionel-rowe/pen/rNEXmoG?editors=0010 page
8. Observe output:
    ```
    {
        "name": "a",
        "value": "b",
        "unparsed": [
            "EXPIRES=Thu, 19 Sep 2024 07:47:28 GMT"
        ]
    }
    ```

**Expected behavior**

Consistent behavior in all locales

**Environment**

- OS: Windows 11
- deno version: N/A
- std version: @std/http@1.0.6

---

- [x] `http/cookie`
- [x] `http/negotiation`
- [ ] `text`

lionel-rowe:
Closes #6016

### DIFF
--- a/text/_util.ts
+++ b/text/_util.ts
@@ -16,5 +16,5 @@ export function splitToWords(input: string) {
 }
 
 export function capitalizeWord(word: string): string {
-  return word ? word?.[0]?.toUpperCase() + word.slice(1).toLowerCase() : word;
+  return word ? word[0]!.toUpperCase() + word.slice(1).toLowerCase() : word;
 }
--- a/text/_util.ts
+++ b/text/_util.ts
@@ -16,7 +16,5 @@ export function splitToWords(input: string) {
 }
 
 export function capitalizeWord(word: string): string {
-  return word
-    ? word?.[0]?.toLocaleUpperCase() + word.slice(1).toLocaleLowerCase()
-    : word;
+  return word ? word?.[0]?.toUpperCase() + word.slice(1).toLowerCase() : word;
 }
--- a/text/_util_test.ts
+++ b/text/_util_test.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 
 import { assertEquals } from "@std/assert";
-import { splitToWords } from "./_util.ts";
+import { capitalizeWord, splitToWords } from "./_util.ts";
 
 Deno.test({
   name: "split() returns an empty array for an empty string",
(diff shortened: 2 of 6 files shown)


kt3k (approved):
LGTM

kt3k:
Could this change/fix any output with Turkish locale?

lionel-rowe:
> Could this change/fix any output with Turkish locale?

@kt3k Sorry, only just noticed this comment after the merge. It would change such output to use "default" letter case conversions (which I think for all intents and purposes is the same as `toLocaleXCase('en-US')`).

It's possible there could be cases where the locale-sensitive version with the host locale is desirable, but I suspect such cases are rare, as it's usually better to explicitly specify a locale. For example, a Turkish-language-only website should consistently use Turkish letter case conventions even for users' whose browser/OS is in another language; meanwhile, a multilingual website supporting Turkish should use the locale of the current content. In cases where the locale is unknown, perhaps the host locale could be used as a fallback, with the usual caveats about potential bugs (string comparison etc).

kt3k:
No worries. I agree this is generally a good convention. Thanks for your contribution.

I was just wondering if we should tag this PR `refactor` if there's no actual fixed case, but that was not a big deal.

PR #6208, opened 2024-11-23
fix(regexp): escape leading digit

### LINKED ISSUE #3656: regexp.escape needs to escape leading digits
The [escape function](https://github.com/lionel-rowe/deno_std/blob/c8673a43e8dbdf09ef8406539bd527bf41cc8b32/regexp/escape.ts) neglects to escape leading digits, which means it doesn't actually work in all contexts:

```js
// ten characters, followed by the first character again, followed by the user input
let getMatch = input => new RegExp(`^(.)(.)(.)(.)(.)(.)(.)(.)(.)(.)\\1${escape(input)}$`);

// seems to work!
console.log(getMatch('x').test('abcdefghijax')); // true

// unless the user input starts with a 0...
console.log(getMatch('0').test('abcdefghija0')); // false
console.log(getMatch('0').test('abcdefghijj')); // true
```

kt3k:
This PR updates `escape` of `@std/regexp`. Now `escape` escapes the leading digit as recommended in #3656. This behavior is aligned with regexp-escape proposal ( https://github.com/tc39/proposal-regex-escaping, polyfill: https://npm.im/regexp.escape )

closes #3656

### DIFF
--- a/regexp/escape.ts
+++ b/regexp/escape.ts
@@ -93,5 +93,5 @@ export function escape(str: string): string {
   return str.replaceAll(
     RX_REGEXP_ESCAPE,
     (m) => RESERVED_CHARS[m as keyof typeof RESERVED_CHARS],
-  ).replace(/^\d/, (m) => `\\x${m.codePointAt(0)!.toString(16)}`);
+  ).replace(/^[0-9a-zA-Z]/, (m) => `\\x${m.codePointAt(0)!.toString(16)}`);
 }
--- a/regexp/escape.ts
+++ b/regexp/escape.ts
@@ -93,5 +93,5 @@ export function escape(str: string): string {
   return str.replaceAll(
     RX_REGEXP_ESCAPE,
     (m) => RESERVED_CHARS[m as keyof typeof RESERVED_CHARS],
-  );
+  ).replace(/^\d/, (m) => `\\x${m.codePointAt(0)!.toString(16)}`);
 }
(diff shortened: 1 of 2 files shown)


regexp/escape.ts
```diff
@@ -93,5 +93,5 @@ export function escape(str: string): string {
   return str.replaceAll(
     RX_REGEXP_ESCAPE,
     (m) => RESERVED_CHARS[m as keyof typeof RESERVED_CHARS],
-  );
+  ).replace(/^\d/, (m) => `\\x${m.codePointAt(0)!.toString(16)}`);
```
lionel-rowe, line 96:
```suggestion
  ).replace(/^[0-9a-zA-Z]/, (m) => `\\x${m.codePointAt(0)!.toString(16)}`);
```

https://tc39.es/proposal-regex-escaping/#sec-regexp.escape behavior and rationale:

> a. If escaped is the empty String and c is matched by either [DecimalDigit](https://tc39.es/ecma262/#prod-DecimalDigit) or [AsciiLetter](https://tc39.es/ecma262/#prod-AsciiLetter), then
>       i. NOTE: Escaping a leading digit ensures that output corresponds with pattern text which may be used after a \0 character escape or a [DecimalEscape](https://tc39.es/ecma262/#prod-DecimalEscape) such as \1 and still match S rather than be interpreted as an extension of the preceding escape sequence. Escaping a leading ASCII letter does the same for the context after \c.

kt3k: Good catch! Thanks

bartlomieju (approved):

PR #6210, opened 2024-11-25
docs(assert): add note and example for Blob comparison

### LINKED ISSUE #6202: assertEquals doesn't correctly compare Blobs
**Describe the bug**

`assertEquals` doesn't consider value-equal `Blob`s to be equal.

**Steps to Reproduce**

Run the following code:

```javascript
import { assertEquals } from "jsr:@std/assert";
assertEquals(new Blob(["foo"]), new Blob(["foo"]));
```

**Expected behavior**

This should not throw an error. Instead, it throws `AssertionError: Values are not equal.`

**Environment**

- OS: macOS 15
- deno version: 2.1.0
- std version: 1.0.8

IgorM867:
This PR improves the documentation for the `assertEquals` function by addressing how to compare `Blob` objects.

Fixes #6202

### DIFF
--- a/assert/equals.ts
+++ b/assert/equals.ts
@@ -15,12 +15,25 @@ import { AssertionError } from "./assertion_error.ts";
  * Type parameter can be specified to ensure values under comparison have the
  * same type.
  *
+ * Note: When comparing `Blob` objects, you should first convert them to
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM. Thanks!

PR #6212, opened 2024-11-26
feat(cli/unstable): export `/unstable-prompt-select`

uki00a:
`promptSelect()` has been added in #6190, but it does not seem to be exported. This PR exports it.


kt3k (approved):
Good catch! Thanks!

PR #6216, opened 2024-11-28
fix(dotenv): handle multiline variables on Windows

### LINKED ISSUE #6170: [@std/dotenv] can't handle multi line variable
**Describe the bug**

the `@std/dotenv` library can't handle multi line variable correctly, it only return the first line of the string.

**Steps to Reproduce**

.env
```env
KEY="one line string
second line string"
```

main.ts
```ts
import { load } from "@std/dotenv";
await load({ export: true });
console.log(Deno.env.get("KEY"));
```
`deno run -A main.ts` get output `one line string`

But if don't use `@std/dotenv`, only `deno run -A --env-file main.ts` 
main.ts
```ts
console.log(Deno.env.get("KEY"));
```
will get correct output `one line string\nsecond line string`

**Expected behavior**

get multiline output `one line string\nsecond line string` while use `@std/dotenv`

**Environment**

- OS: Window 10
- deno version: 2.0.4
- std/dotenv version: 0.225.2

IgorM867:
Current env parser do not support `.env` files with `\r\n` line break.

fixes #6170

### DIFF
--- a/dotenv/parse.ts
+++ b/dotenv/parse.ts
@@ -10,7 +10,7 @@ type LineParseResult = {
 type CharactersMap = { [key: string]: string };
 
 const RE_KEY_VALUE =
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM. Thanks for finding this out!

Issue #6206, opened 2024-11-22
expect's customMessage not working for all matchers


lcrespilho:
**Describe the bug**
The _customMessage_ parameter of **expect** module is not working for all matchers. For instance, it doesn't work for _toBeGreaterThan_ matcher.

**Steps to Reproduce**
1. Create a file named demo_test.ts with these contents:
```typescript
import { expect } from 'jsr:@std/expect'

Deno.test('mytest', () => {
  expect(1, 'My Custom Message').toBeGreaterThan(2)
})
```
2. Execute this command on the terminal to run the test:
`deno test`

3. The output contains:
```
mytest => ./demo_test.ts:3:6
error: AssertionError: Expected 1 to be greater than 2
      throw new AssertionError(
            ^
```

**Expected behavior**

The output should include the _customMessage_ parameter, like:
```
mytest => ./demo_test.ts:3:6
error: AssertionError: Expected 1 to be greater than 2: My Custom Message
      throw new AssertionError(
            ^
```

**Environment**

- OS: MacOS 15.1.1 Sequoia
- deno version: 2.1.1
- @std/assert version: 1.0.8

PR #6217, opened 2024-11-28
fix(expect): always show custom message

eryue0220:
This pr is to solve #6206


expect/_matchers.ts
```diff
@@ -89,13 +89,15 @@ export function toBeCloseTo(
   if (context.isNot) {
     if (pass) {
       throw new AssertionError(
-        `Expected the value not to be close to ${expected} (using ${numDigits} digits), but it is`,
+        context.customMessage ??
+          `Expected the value not to be close to ${expected} (using ${numDigits} digits), but it is`,
```
kt3k, line 93: I think this part should be something like:

```ts
const defaultMessage = `Expected the value not to be close to ${expected} (using ${numDigits} digits), but it is`
throw new AssertionError(
  context.customMessage ? `${context.customMessage}: ${defaultMessage}` : defaultMessage
);
```

Vitest, for example, shows the custom message like the above.

In Vitest:
```ts
expect(42, "custom message").toEqual(43);
```
This assertion error shows the message like:
```
AssertionError: custom message: expected 42 to deeply equal 43

- Expected
+ Received

- 43
+ 42
```

eryue0220: Updated

kt3k: Oh, I missed this message. Sorry for the delay!

kt3k:
I vendored `assertIsError` to `expect/` because the suggested change was not desired for `@std/assert/is-error`.

kt3k (approved):
LGTM. Thanks!

PR #6221, opened 2024-12-01
fix(cli/unstable): hide cursor while showing the selection with `promptSelect()`

timreichen:
This PR moves the cursor to the selected line instead of always staying at the bottom.

### DIFF
--- a/cli/unstable_prompt_select.ts
+++ b/cli/unstable_prompt_select.ts
@@ -13,10 +13,10 @@ const CR = "\r";
 const INDICATOR = "❯";
 const PADDING = " ".repeat(INDICATOR.length);
 
-const CLR_ALL = "\x1b[J"; // Clear all lines after cursor
-
 const encoder = new TextEncoder();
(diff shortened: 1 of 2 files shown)


kt3k:
The tools like [npm:inquirer](https://www.npmjs.com/package/inquirer) or [npm:prompts](https://www.npmjs.com/package/prompts) seem hiding the cursor during selection prompts. That behavior makes more sense to me than moving cursor along with the marker because the users can't input text while selecting. What do you think?

timreichen:
> The tools like [npm:inquirer](https://www.npmjs.com/package/inquirer) or [npm:prompts](https://www.npmjs.com/package/prompts) seem hiding the cursor during selection prompts. That behavior makes more sense to me than moving cursor along with the marker because the users can't input text while selecting. What do you think?

Agree, this makes more sense. Updated.

kt3k (approved):
Confirmed the cursor is now hidden. LGTM

PR #6222, opened 2024-12-01
fix(testing): add missing methods to `test` alias of `it`

### LINKED ISSUE #6220: `test` is not exactly an alias for `it`
**Describe the bug**

https://jsr.io/@std/testing/doc/bdd/~/test states that `test` is an "Alias of [`it`](https://jsr.io/@std/testing/doc/bdd/~/it)". But `test` does not have the `only`, `ignore`, and `skip` methods from `it`. This makes it incompatible with Jest test suites, for example.

**Steps to Reproduce**

`test.only(..)` will produce

> Property only does not exist on type <T>(...args: ItArgs<T>) => void

IgorM867:
`test` is documented as an alias for `it`, but it does not currently implement methods like `only`, `ignore`, and `skip`. This PR fixes the issue by ensuring `test` correctly mirrors the behavior and methods of `it`.

Fixes #6220


testing/bdd.ts
```diff
@@ -741,6 +741,88 @@ export function test<T>(...args: ItArgs<T>) {
...
+ * ```
+ *
+ * @param args The test case
+ */
+test.only = function itOnly<T>(...args: ItArgs<T>): void {
+  const options = itDefinition(...args);
+  return it({
+    ...options,
+    only: true,
+  });
```
kt3k, line 770: Can we use `it.only`?

```suggestion
  it.only(...args);
```

testing/bdd.ts
```diff
@@ -741,6 +741,88 @@ export function test<T>(...args: ItArgs<T>) {
...
+ * ```
+ *
+ * @param args The test case
+ */
+test.ignore = function itIgnore<T>(...args: ItArgs<T>): void {
+  const options = itDefinition(...args);
+  return it({
+    ...options,
+    ignore: true,
+  });
```
kt3k, line 799: Can we use `it.ignore`?
```suggestion
  it.ignore(...args);
```

kt3k (commented):
Mostly looks good to me. Thanks for working on this. Left a few questions.

IgorM867:
You’re right. I made the requested changes, and I also added some tests. However, I’m not completely confident about these tests because I copied an existing `it` test, which led to a lot of duplicate code. Let me know what you think.

kt3k:
I don't think we need new test cases for these as they only have single line implementations, and no new logics are introduced.

kt3k (approved):
LGTM

Issue #6213, opened 2024-11-26
Contributing "waitFor" function to async module, desired?


acrodrig:
Very often when writing integration tests I need to wait for a condition to be true to continue (for example the database record is there, or an SMS message has arrived). I use the function below repeatedly and I would love if it can be part of `@std/async`. If you see value in it, would contribute as PR. Please let know.

```typescript
/**
 * Wait for a condition to be true, jumping every `jump` milliseconds until 
 * `timeout` is reached. Can be used as:
 *
 * ```typescript
 * let i = 0;
 * const id = setInterval(() => i++, 42);
 * await waitFor(() => i > 10);
 * clearInterval(id);
 * assert(i > 10);
 * ```
 *
 * @param condition - predicate returning a (potentially promised) boolean
 * @param timeout - maximum wait time (default 20s, 0 for infinite)
 * @param jump - the time to wait between checks (default 100ms)
 */
function waitFor(condition: () => boolean | Promise<boolean>, timeout = 20000, jump = 100) {
  // Time we start the checks so that we can
  const start = Date.now();

  return new Promise(function (resolve, reject) {
    // If condition is "truthy" resolve, if timeout reject, otherwise keep waiting
    const check = async function (): Promise<void> {
      const value = await condition();
      if (value) return resolve(value);
      const time = Date.now() - start;
      if (time > timeout) return reject(new Error("Timeout (" + timeout + "ms) waiting for condition!"));
      setTimeout(check, jump);
    };

    // Start checking
    setTimeout(check);
  });
}
```

kt3k:
The suggestion looks reasonable to me.

Please add this as unstable feature first. https://github.com/denoland/std/blob/9daae0547ff646dddcd0628afe355b61eace78d5/.github/CONTRIBUTING.md#suggesting-a-new-feature

The error class might better to be `DOMException` with `TimeoutError` name, that is what `AbortSignal.timeout()` and `@std/async/deadline` throw

dandv:
Looks useful.

kt3k:
fixed by #6230

PR #6230, opened 2024-12-04
feat(async/unstable): add `waitFor` function to wait for condition to be true (#6213)

acrodrig:
Submitting PR to contribute `waitFor` function (see #6213). Fixed small type in delay tests as well.

### DIFF
--- a/async/unstable_wait_for.ts
+++ b/async/unstable_wait_for.ts
@@ -1,0 +1,63 @@
+// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// This module is browser compatible.
+
+import { deadline } from "./deadline.ts";
+
+/** Options for {@linkcode waitFor}. */
+export interface WaitForOptions {
+  /** Signal used to abort the waitFor. */
+  signal?: AbortSignal;
+  /** Indicates the step jump in time to wait for the predicate to be true.
+   *
+   * @default {100}
+   */
+  step?: number;
+}
+
+/**
+ * Resolve a {@linkcode Promise} after a given predicate becomes true or the
+ * timeout amount of milliseconds ahs been reached.
+ *
+ * @throws {DOMException} If signal is aborted before either the waitFor
+ * predicate is true or the timeout duration was reached, and `signal.reason`
+ * is undefined.
+ * @param predicate a Nullary (no arguments) function returning a boolean
+ * @param ms Duration in milliseconds for how long the waitFor should last.
+ * @param options Additional options.
+ *
+ * @example Basic usage
+ * ```ts no-assert
+ * import { waitFor } from "@std/async/unstable-wait-for";
+ *
+ * // Deno server to acknowledge reception of request/webhook
+ * let requestReceived = false;
+ * Deno.serve((_req) => {
+ *   requestReceived = true;
+ *   return new Response("Hello, world");
+ * });
+ *
+ * // ...
+ * waitFor(() => requestReceived, 10000);
+ * // If less than 10 seconds pass, the requestReceived flag will be true
+ * // assert(requestReceived);
+ * // ...
+ * ```
+ */
+export function waitFor(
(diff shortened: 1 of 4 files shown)


async/unstable_wait_for.ts
```diff
@@ -0,0 +1,63 @@
...
+  /** Indicates the step jump in time to wait for the predicate to be true.
+   *
+   * @default {100}
+   */
+  step?: number;
+}
+
+/**
+ * Resolve a {@linkcode Promise} after a given predicate becomes true or the
+ * timeout amount of milliseconds ahs been reached.
```
kt3k, line 19:
```suggestion
 * timeout amount of milliseconds has been reached.
```

async/unstable_wait_for_test.ts
```diff
@@ -0,0 +1,27 @@
...
+  setTimeout(() => flag = true, 100);
+  const start = Date.now();
+  await waitFor(() => flag === true, 1000);
+  // Expects the promise to be resolved after 100ms
+  assertAlmostEquals(Date.now() - start, 100, 10);
+});
+
+Deno.test("waitFor() throws DOMException on timeout", async () => {
+  let flag = false;
+  setTimeout(() => flag = true, 1000);
```
kt3k, line 18: This timer needs to be cleared at the end of the test case to avoid op leak error

async/unstable_wait_for_test.ts
```diff
@@ -0,0 +1,27 @@
...
+  const start = Date.now();
+  const error = await assertRejects(
+    () => waitFor(() => flag === true, 100),
+    DOMException,
+    "Signal timed out.",
+  );
+  assertAlmostEquals(Date.now() - start, 100, 10);
+  assertEquals(error.name, "TimeoutError");
```
kt3k, line 26: 👍

async/unstable_wait_for_test.ts
acrodrig: Will do, makes sense. I do not know why I did not get the leak when running locally.

kt3k (dismissed):
LGTM

kt3k:
I just remembered another rule about designing this type of function.

We tend to put the function-type argument at the last of parameters. The examples are [Deno.serve](https://docs.deno.com/api/deno/~/Deno.serve), [Deno.test](https://docs.deno.com/api/deno/~/Deno.test), [Deno.addSignalListener](https://docs.deno.com/api/deno/~/Deno.addSignalListener).

(It is noted in our [style guide](https://docs.deno.com/runtime/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object))

> Note: When one of the arguments is a function, you can adjust the order flexibly. See examples like [Deno.serve](https://docs.deno.com/api/deno/~/Deno.serve), [Deno.test](https://docs.deno.com/api/deno/~/Deno.test), [Deno.addSignalListener](https://docs.deno.com/api/deno/~/Deno.addSignalListener). See also [this post](https://twitter.com/jaffathecake/status/1646798390355697664).

What do you think about changing this to `waitFor(timeout, [options,] predicate)`? (We need to use type overload to achieve optional 2nd argument).

acrodrig:
> What do you think about changing this to `waitFor(timeout, [options,] predicate)`? (We need to use type overload to achieve optional 2nd argument).

Sure, makes sense, let me give it a try. I was following the [deadline](https://jsr.io/@std/async/doc/~/deadline) layout which has the `Promise` as the first argument.

kt3k:
> What do you think about changing this to waitFor(timeout, [options,] predicate)? (We need to use type overload to achieve optional 2nd argument).

I reconsidered this and I now think this is just fine (`debounce` and `throttle` also accept function as the first argument. So we are not necessarily consistent about this aspect). Let's land this as is

kt3k (approved):

acrodrig:
Awesome, much appreciated. One more thing I do not need my personal library for.

PR #6231, opened 2024-12-04
feat(http/unstable): add `cleanUrls` option to `serveDir`

### LINKED ISSUE #6232: proposal(http): add `cleanUrls` option to `serveDir`

Nowadays it's pretty common to trim html page extensions from urls to make them nice and rememberable.
(e.g. deno.com/blog, github.com/denoland/std, ..., some hosters such as Vercel also propose a [similar feature](https://vercel.com/docs/projects/project-configuration#cleanurls))

The reason for why I'd like it to be natively in the std function is:
- it avoids to implementing routing for simple use-cases (SSG websites)
  - `serveDir` is often also used as a fallback for `route()`, but it makes it kind of ugly to use it as-is if you want the clean urls behavior (see workaround below) 
- workaround exists but not practical
  - change your file structure (can cause interoperability issues with other modules) to use directories instead, rename all your file as `index.html` and use the index option
  - `serveDir` once, check if response status is 404, then rewrite the request (but kind of troublesome since `Request` are immutable, and also may already have been consumed) and retry with the extension. while it works, feels hacky and probably poorly memory efficient as you need to create multiple up to 4 instances of objects to serve a single file

The patch is pretty small to implement, as you only need to add a second resolution check if the option is enabled, so the footprint would be pretty low

I feel like the use-case is common enough, and makes it easier to deploy smaller-scales app / docs website / SSG without having to code anything, as you'd be able to directly `deno run -A jsr:@std/http/file_server` , which will be a nice QOL

If this get agreed on, #6231 can be reopened, here's the reference:

```diff
-  const fsPath = join(target, normalizedPath);
+  // Resolve path
+  // If cleanUrls is enabled, automatically append ".html" if not present
+  // and it does not shadow another existing file or directory
+  let fsPath = join(target, normalizedPath);
+  if (
+    cleanUrls && !fsPath.endsWith(".html") &&
+    !(await Deno.stat(fsPath).then(() => true).catch(() => false))
+  ) {
+    fsPath += ".html";
+  }
  const fileInfo = await Deno.stat(fsPath);
```

One of the workaround listed above

lowlighter:
This PR introduces a new `cleanUrls` option to `serveDir` function

It is inspired from a common pattern where you don't have to specify the html extension to serve a page.
(e.g. deno.com/blog, github.com/denoland/std, ..., some hosters such as Vercel also propose a [similar feature](https://vercel.com/docs/projects/project-configuration#cleanurls))

Considerations:
- Only `.html` is handled by this new option, for more advanced usage user would need to use their own routing
  - The reasoning for what it'd be supported natively by std is that it's a really common use case, and it's actually not that easy to get a simple workaround (e.g. if you use `route(request)`, as `Request` are immutable you'll need to do some shenanigans to trim extensions, catch the 404 to retry with it, etc.)
  - It'd be an extremely nice QOL and since it's gated behind an option would not be a breaking change and would not hinder performances of existing apps 
- If the path without extension "shadows" another file or directory, the latter takes precedences over the otherwise cleaned url aliased file

Closes #6232


lowlighter:
I'm closing it for now, but I still think this patch would be a nice QOL, especially since it doesn't have that much maintenance/overhead, but this may need further discussion as pointed on the discord server since opinions seems to be diverging.
___
Feedback from the discord:

> It's common as a "framework feature", but for file server (like apache) you usually need either rewriting urls or set the default mime type to `text/html`

> A workaround is to change the file structure to use folder + index.html instead but although it works it makes you change the way you organize file rather than your code (which shouldn't be, the file structure may be required to be a certain way depending on your use case)

iuioiua:
I've drafted #6668 to wait until we see what we do with https://github.com/denoland/deno/issues/29447.

kt3k:
I think this is independent from #6668 (#6668 changes the entrypoint and this changes `serveDir` API)

> I could support the new options from the cli tools too (should it be enabled by default ?)

Let's discuss that in an independent issue.

lowlighter:
Ok then I wasn't sure if you wanted to avoid conflicts on existing work then

The build fail is i think unrelated to my pr, it's failing due to rename/renameSync (both unstable api, maybe flaky tests ?)

kt3k (approved):
LGTM

kt3k:
The CI failure seems caused by CLI canary change (probably this one https://github.com/denoland/deno/commit/cb23193f74b2cf34cf06111f9ebbf0599d5ec24d ). Fixing it in https://github.com/denoland/std/pull/6695

PR #6236, opened 2024-12-05
test(fs): do not write files in source tree during test

### LINKED ISSUE #5358: test(fs): do not write files in source tree during test
`std/fs` test cases directly write files in source tree during test. This makes debugging of test cases difficult when they failed in the middle. These test cases should use temp directory instead.

jbronder:
This pull request addresses the issue of files and directories being created and written into the source tree while testing functions in the `fs` package. Creation of temporary directories were incorporated into the existing tests where applicable.    

Fixes #5358 .

### DIFF
--- a/fs/empty_dir_test.ts
+++ b/fs/empty_dir_test.ts
@@ -8,10 +8,9 @@ import {
 import * as path from "@std/path";
 import { emptyDir, emptyDirSync } from "./empty_dir.ts";
 
-const testdataDir = path.join(import.meta.dirname!, "testdata");
-
 Deno.test("emptyDir() creates a new dir if it does not exist", async function () {
-  const testDir = path.join(testdataDir, "empty_dir_test_1");
+  const tempDirPath = await Deno.makeTempDir({ prefix: "deno_std_empty_dir_" });
+  const testDir = path.join(tempDirPath, "empty_dir_test_1");
   const testNestDir = path.join(testDir, "nest");
   // empty a dir which not exist. then it will create new one
   await emptyDir(testNestDir);
@@ -21,13 +20,16 @@ Deno.test("emptyDir() creates a new dir if it does not exist", async function ()
     const stat = await Deno.stat(testNestDir);
     assertEquals(stat.isDirectory, true);
   } finally {
-    // remove the test dir
-    await Deno.remove(testDir, { recursive: true });
+    // Cleanup and remove test directories.
+    await Deno.remove(tempDirPath, { recursive: true });
   }
 });
 
 Deno.test("emptyDirSync() creates a new dir if it does not exist", function () {
-  const testDir = path.join(testdataDir, "empty_dir_test_2");
+  const tempDirPath = Deno.makeTempDirSync({
+    prefix: "deno_std_empty_dir_sync_",
+  });
+  const testDir = path.join(tempDirPath, "empty_dir_test_2");
   const testNestDir = path.join(testDir, "nest");
   // empty a dir which does not exist, then it will a create new one.
   emptyDirSync(testNestDir);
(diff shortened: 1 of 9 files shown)


fs/empty_dir_test.ts
```diff
@@ -8,10 +8,9 @@ import {
 import * as path from "@std/path";
 import { emptyDir, emptyDirSync } from "./empty_dir.ts";
```
iuioiua, line 10: I would've guessed that this is more or less all that's needed.
```suggestion
const testdataDir = await Deno.makeTempDir({ prefix: "deno_std_empty_dir_" });
```

jbronder: It was definitely my first thought and I probably should've asked before starting. I guess I was overthinking it and thought I should be keeping the tests completely asynchronous or synchronous for setup and tear-down. I'll try and clean it up some more.

On a separate note, would you happen to know off hand if the Windows CI tests are configured differently? The permissions being denied on the Windows build, but not on the other OS's seems OS specific.

iuioiua: Hm... Tests are being run with `--allow-all`, which should mitigate `PermissionDenied` errors, AFAIK. Not sure. Any idea, @kt3k? In the meantime, I suggest applying my suggestion.

kt3k: The error from deno permission system has been changed to `NotCapable` https://github.com/denoland/deno/pull/25431

So `PermissionDenied` should be related to the operating system's permission control.. But I'm not sure what causes this particular permission error..

jbronder: Apologies for the long bit of text for something that, on balance, sounds simple. I did think a top-level `await` would work well with the `Deno.test()` declarations. 

After trying to implement a top-level temporary directory, the one drawback I'm running into is that no clean up of the top-level temp directory can take place. This allows tests to write to the temp directory, but if I included a top-level `await Deno.remove(tempDirPath)` in the `_test.ts` files (even somewhere after all the `Deno.test(...)` tests have been declared), the removal of the temporary directory would be invoked _before_ any of the tests could run. It's not a desirable solution, but since it would affect developers running these tests on their local machines, I'm not sure if this is ideal, but if this is something worth moving forward on, please let me know.

There's currently two files that run their tests in temporary directories: `copy_test.ts` and `exists_test.ts`. `copy_test.ts` uses wrapper functions `testCopy()` and `testCopySync()` that use the test definition as a callback and creates temporary directories to be passed in as arguments to the callback. The other file, `exists_test.ts`, uses the approach that I similarly applied to all the other `_test.ts` files in the `fs` package, creating a temp directory at the beginning of the test and then cleaning up at the end. These solutions work, but if there's a less invasive way to modify the test, I'm not sure that a top-level `await` would be the approach.

kt3k: @jbronder Thanks for looking into it. Sounds like the current structure is better than the use of top-level await. Let's keep the structure of this PR as is.

I think the only remaining issue with this change is the CI error on windows. Do you have any idea about it?

IgorM867: @kt3k This error on Windows is caused by an invalid symlink created by the `copy()` function. PR #6278 should solve this issue.

jbronder: @kt3k  I took a look at PR #6278 and I managed to get the tests working, including windows, with this additional change to PR #6278:
- Using `Deno.realPath` and `Deno.realPathSync` instead of `Deno.readLink` and `Deno.readLinkSync` when assigning the `originSrcFilePath` variable.

I've included this change as a suggestion in PR #6278.  Apologies in advance if that's too forward. `Deno.readLink` does report back the target path from the given symlink, but it does not always resolve to an absolute path. Rather, I think it resolves to the target path given when the symlink was initially created. `Deno.realPath` returns the absolute path of the target that is pointed to by the symlink.

The use of `Deno.realPath` did create another issue that could be addressed in PR #6278 or it could be addressed in an another issue, if it makes more sense. The tests in `copy_test.ts`, entitled `copy() copies a symlink file` and
`copySync() copies symlink file`, produce an error indicating that the target path pointed to by the symlink can't be found. This is because the target path `./fs/testdata/copy_dir/0.txt` pointed to by symlink in `./fs/testdata/copy_dir_link_file/0.txt` doesn't actually point to the intended file. I could update the symlink to point to `../copy_dir/0.txt` in a separate PR, or it could be updated in PR #6278.

kt3k: I don't think changing that to `realPath` is correct. By changing it, the copied symlink points to the path in the original dir. I think we should keep the relative path instead.

I haven't figured out the reason of permission error on windows, but I feel it might be related to the difference of temp directory, which is on C: drive, but the source tree is on D: drive (On windows a relative path could be prefixed by the drive https://github.com/denoland/std/blob/ed52423921e6ad91e32a0b6a9a49365f5cc03caa/path/join_test.ts#L120 )

kt3k: BTW the specific test cases don't seem creating new files during test run. I think they don't absolutely need to use temp directory for testing. I'll revert them back to the original code for a temporary solution.

kt3k (approved):
LGTM. Thanks for your contribution! Also thanks for looking into the errors on windows

PR #6237, opened 2024-12-05
fix(testing): stop code execution in `describe.ignore`

### LINKED ISSUE #6233: describe.skip does not skip assertions in block
I expect that skipped `describe` blocks don't run any code inside, and especially not assertions.

**Steps to Reproduce**

1. Create `denobug.test.ts`
```ts
import { describe, it } from '@std/testing/bdd';
import { assertExists } from 'jsr:@std/assert';

let skipTestingThisVariable: string;

describe.skip('skip everything in here', () => {
  assertExists(skipTestingThisVariable);  // <-- this line should be skipped
  it('should work', () => {
    assertExists(true);
  });
});
```
2. Run `deno test denobug.test.ts`

```
Uncaught error from ./denobug.test.ts FAILED

 ERRORS 

./denobug.test.ts (uncaught error)
error: (in promise) AssertionError: Expected actual: "undefined" to not be null or undefined.
          ^
    at assertExists (https://jsr.io/@std/assert/1.0.8/exists.ts:29:11)
    at file:///~/deno/denobug.test.ts:7:3
    at new TestSuiteInternal (https://jsr.io/@std/testing/1.0.5/_test_suite.ts:91:23)
    at describe (https://jsr.io/@std/testing/1.0.5/bdd.ts:1111:22)
    at Function.describeIgnore [as ignore] (https://jsr.io/@std/testing/1.0.5/bdd.ts:1176:10)
    at Function.describeSkip [as skip] (https://jsr.io/@std/testing/1.0.5/bdd.ts:1208:19)
    at file:///~/deno/denobug.test.ts:6:10
This error was not caught from a test and caused the test runner to fail on the referenced module.
It most likely originated from a dangling promise, event/timeout handler or top-level code.

 FAILURES 

./denobug.test.ts (uncaught error)

FAILED | 0 passed | 1 failed (0ms)

error: Test failed
```

**Expected behavior**

The test should be ignored. If the `assertExists(skipTestingThisVariable);` is moved into the `it`, then it's correctly ignored

```ts
import { describe, it } from '@std/testing/bdd';
import { assertExists } from 'jsr:@std/assert';

let skipTestingThisVariable: string;

describe.skip('skip everything in here', () => {
  it('should work', () => {
    assertExists(skipTestingThisVariable);  // test is ignored now
    assertExists(true);
  });
});
```

**Environment**

- deno version: 2.1.2
- std version: "jsr:@std/expect@*": "1.0.8"

IgorM867:
In the current implementation, code inside a `describe` block is executed in every case, even when `describe.ignore` or `describe.skip` is used. This behavior is reflected in the following example and is currently tested as expected behavior:

Code:

```ts
describe.ignore("should ignore", () => {
  assert(false);
});
```

Error:

```sh
Uncaught error from ./main.ts FAILED

 ERRORS 

./main.ts (uncaught error)
error: (in promise) AssertionError
          ^
```

This happens because the describe block is treated as a list of steps rather than as a single step itself. In my implementation, all code inside a describe.ignore block is ignored.

Fixes #6233

### DIFF
--- a/testing/_test_suite.ts
+++ b/testing/_test_suite.ts
@@ -82,7 +82,40 @@ export class TestSuiteInternal<T> implements TestSuite<T> {
     this.symbol = Symbol();
     TestSuiteInternal.suites.set(this.symbol, this);
 
-    const { fn } = describe;
+    const { fn, ignore } = describe;
+    if (ignore) {
+      const {
+        name,
+        only,
+        permissions,
+        sanitizeExit,
+        sanitizeOps,
+        sanitizeResources,
+      } = describe;
+      const options: Deno.TestDefinition = {
+        name,
+        fn: async () => {},
+        ignore: true,
+      };
+      if (only !== undefined) {
+        options.only = only;
+      }
+      if (permissions !== undefined) {
+        options.permissions = permissions;
+      }
+      if (sanitizeExit !== undefined) {
+        options.sanitizeExit = sanitizeExit;
+      }
+      if (sanitizeOps !== undefined) {
+        options.sanitizeOps = sanitizeOps;
+      }
+      if (sanitizeResources !== undefined) {
+        options.sanitizeResources = sanitizeResources;
+      }
+      TestSuiteInternal.registerTest(options);
+      return;
+    }
+
     if (fn) {
       const temp = TestSuiteInternal.current;
       TestSuiteInternal.current = this;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Confirmed the case in #6233 is fixed by this change. Thanks!

dandv:
@IgorM867: thanks for the quick fix! thoughts on the Jest and docs issues I [mentioned in the issue](https://github.com/denoland/std/issues/6233#issuecomment-2522779087)?

PR #6245, opened 2024-12-09
feat(assert): type narrowing for assertLess, assertLessOrEqual, asserGreater, assertGreaterOrEqual

### LINKED ISSUE #6244: assertGreater, assertAlmostEquals, assertLess etc. don't do type narrowing
**Steps to Reproduce**

```ts
import { assertExists, assertGreater, assertAlmostEquals, assertLess } from '@std/assert';

const x = Math.random() > 0.5 ? 10 : undefined;
const y = Math.random() > 0.5 ? 10 : undefined;
const z = Math.random() > 0.5 ? 10 : undefined;
const t = Math.random() > 0.5 ? 10 : undefined;

assertExists(x);
console.log(x + 1);  // no warning, because assertExists did type narrowing
assertGreater(y, 0);
assertAlmostEquals(t, 5);  // "Argument of type number | undefined is not assignable to parameter of type number"
assertLess(z, 20);  // no complaints about z being "possibly undefined"
console.log(x - y - z - t);  // y, z, t are "possibly undefined", but x is not
```

**Expected behavior**

All asserts, not just `assertExists`, should narrow `undefined` out of the type if the assertion passes. That includes (not pictured here) `assertGreaterOrEqual` etc.

In addition, `assertAlmostEquals` should behave consistently with the other asserts, and accept an undefined.

**Environment**

- deno version: 2.1.3
- std version: 1.0.9

dandv:
- disallow undefined actual values because any comparison with `undefined` returns false
- disallow null expected values
- narrow the returned type to exclude undefined
- drive-by typo fix and relative path imports

Fixes #6244.

### DIFF
--- a/assert/array_includes.ts
+++ b/assert/array_includes.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 import { equal } from "./equal.ts";
-import { format } from "@std/internal/format";
+import { format } from "../internal/format.ts";
 import { AssertionError } from "./assertion_error.ts";
 
 /** An array-like object (`Array`, `Uint8Array`, `NodeList`, etc.) that is not a string */
--- a/assert/almost_equals_test.ts
+++ b/assert/almost_equals_test.ts
@@ -37,7 +37,7 @@ Deno.test("assertAlmostEquals() throws values outside higher precision range", (
   );
 });
 
-Deno.test("assertAlmostEquals() matches infinity with inifinity", () => {
+Deno.test("assertAlmostEquals() matches infinity with infinity", () => {
   assertAlmostEquals(Infinity, Infinity);
 });
 
--- a/assert/equals.ts
+++ b/assert/equals.ts
@@ -1,10 +1,10 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 import { equal } from "./equal.ts";
-import { buildMessage } from "@std/internal/build-message";
-import { diff } from "@std/internal/diff";
-import { diffStr } from "@std/internal/diff-str";
-import { format } from "@std/internal/format";
+import { buildMessage } from "../internal/build_message.ts";
+import { diff } from "../internal/diff.ts";
+import { diffStr } from "../internal/diff_str.ts";
+import { format } from "../internal/format.ts";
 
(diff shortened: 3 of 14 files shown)


dandv:
On a second thought (after my #6244 issue), `assertLess` & friends should not accept undefined, because any comparison with undefined returns false.

If x can be undefined, `console.log(x < 1)` produces a Deno error that `x` is possibly undefined:

```ts
const x = Math.random() > 1.0 ? 10 : undefined;

console.log(x < 3);  // error: TS18048 [ERROR]: 'x' is possibly 'undefined'.
```

So I've added [typing for these assertions to also expect `actual` to be any type except `undefined`](https://github.com/denoland/std/pull/6245/files#diff-d274e815c23383fb2c267957feca269fb2f4d82730b7653603b5f9e7a7069a05L24).

```ts
actual: Exclude<T, undefined> | null,
```

The error isn't pretty, but it's better than nothing:

> Argument of type 10 | undefined is not assignable to parameter of type 0 | 10 | null
> Type undefined is not assignable to type 0 | 10 | null

The `| null` is there for explicitness, because comparison with null are valid (null gets coerced to zero). Removing it leads to slightly simpler errors:

> Argument of type 10 | undefined is not assignable to parameter of type 0 | 10
> Type undefined is not assignable to type 0 | 10

assert/array_includes.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
 // This module is browser compatible.
 import { equal } from "./equal.ts";
-import { format } from "@std/internal/format";
+import { format } from "../internal/format.ts";
```
kt3k, line 4: This relative import doesn't work when published to JSR as each package is published independently

dandv: I made that change because the `@std` import led to this error when using the repo with `file:///` imports from 3rd party code:

`Relative import path "@std/internal/build-message" not prefixed with / or ./ or ../ and not in import map`

Is there a syntax that would work in both cases, or should I revert the change from the first commit of this PR?

kt3k:
This is a breaking change in the current format (The users will see the type error if the input type include `undefined` in its type), and we are not planning to make major version upgrade in the near future.

Also I don't see much value to this narrowing of types as the same effect can be achieved by calling `assertExists(value)` next line to these assertions.

kt3k:
We first need to discuss what can/should be done about this in #6244 

Closing for now

PR #6252, opened 2024-12-11
feat(cbor): add encoding/decoding for `new Map()` instance

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6249

These changes adds support for accepting a [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) instance in `encodeCbor` and returning it again from `decodeCbor`. It achieves this via the [259 tag number](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml) ([Spec](https://github.com/shanewholloway/js-cbor-codec/blob/master/docs/CBOR-259-spec--explicit-maps.md)). 

While the code is mostly the same for encoding and decoding Objects, it differs in the need for limiting keys to only strings. It also guarantees that the order isn't at risk of being changed due to encoding/decoding.

This pull request does not add support for accepting Map instances via the streaming methods of this library. The logistics for decoding that is a bit more trickier to figure out.

```ts
import { type CborType, decodeCbor, encodeCbor } from "@std/cbor";

assertEquals(
  decodeCbor(encodeCbor(new Map([[ 1, 2 ], [ "3", 4 ], [ [5], { a: 6 } ]]))),
  new Map([[ 1, 2 ], [ "3", 4 ], [ [5], { a: 6 } ]])
);
```

### DIFF
--- a/cbor/_common_decode.ts
+++ b/cbor/_common_decode.ts
@@ -234,29 +234,34 @@ export function decodeFive(
 export function decodeSix(
   source: number[],
   aI: number,
-): Date | CborTag<CborType> {
+): Date | Map<CborType, CborType> | CborTag<CborType> {
   if (aI > 27) {
     throw new RangeError(
       `Cannot decode value (0b110_${aI.toString(2).padStart(5, "0")})`,
     );
   }
   const tagNumber = decodeZero(source, aI);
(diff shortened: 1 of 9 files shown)


kt3k (approved):
LGTM

PR #6253, opened 2024-12-11
test(cli/unstable): cleanup `promptSelect()` tests

timreichen:
This PR renames some tests ~and uses an array to store outputs that are compared with the expected outputs instead of comparing each output~.


kt3k:
This branch seems to have started from the old state of main before #6221. Can you fix the conflicts?

kt3k (approved):
LGTM

PR #6254, opened 2024-12-11
test(cli/unstable): rewrite `Spinner` tests using stub

timreichen:
closes https://github.com/denoland/std/issues/6157
closes https://github.com/denoland/std/issues/6211


kt3k (approved):
LGTM. Nice work! Thanks

I also updated the test case. Now they are surrounded by try block and `restore` is called in `finally` block (By this change, `restore` is always called even when there's assertion errors in the middle)

PR #6256, opened 2024-12-12
test(cli): add tests for `promptSecret()`

timreichen:
Seems like `promptSecret()` is stable but not covered by any tests. I think that should be avoided in the future.

### DIFF
--- a/cli/prompt_secret.ts
+++ b/cli/prompt_secret.ts
@@ -48,6 +48,10 @@ export function promptSecret(
 ): string | null {
   const { mask = "*", clear } = options ?? {};
 
+  if (!input.isTerminal()) {
+    return null;
+  }
+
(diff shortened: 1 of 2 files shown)


cli/prompt_secret.ts
```diff
@@ -48,10 +48,6 @@ export function promptSecret(
 ): string | null {
   const { mask = "*", clear } = options ?? {};
 
-  if (!input.isTerminal()) {
-    return null;
-  }
-
```
kt3k, line 54: I don't think this should be removed. This is for avoid prompting in non terminal

kt3k: You can stub `isTerminal` to avoid errors in CI

kt3k (approved):
Nice! Thanks

PR #6257, opened 2024-12-12
chore(cli): align prompt functions

timreichen:
This PR aligns all cli prompt functions to look and work the same internally.

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -202,6 +202,10 @@ function assertHasParamDefinition(
       paramDoc.kind === "assign" && paramDoc.left.kind === "identifier"
     ) {
       return paramDoc.left.name === param.name;
+    } else if (
+      paramDoc.kind === "assign" && paramDoc.left.kind === "object"
+    ) {
+      return true;
     }
     return false;
   });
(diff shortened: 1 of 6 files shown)


_tools/check_docs.ts
```diff
@@ -202,6 +202,10 @@ function assertHasParamDefinition(
       paramDoc.kind === "assign" && paramDoc.left.kind === "identifier"
     ) {
       return paramDoc.left.name === param.name;
+    } else if (
+      paramDoc.kind === "assign" && paramDoc.left.kind === "object"
+    ) {
+      return true;
```
kt3k, line 208: This change has undesirable effect in doc rendering.

By using destructuring in parameter, there's no longer `options` param in arguments, and `@param options` directive becomes unclear about what it documents.

timreichen: I see. Why is it though that `deno task lint:docs` allows https://github.com/denoland/std/blob/cc9a7c006d3587c4eee04f4924dfe8c2cc928962/cli/unstable_prompt_select.ts#L41 but will throw here https://github.com/denoland/std/blob/cc9a7c006d3587c4eee04f4924dfe8c2cc928962/cli/unstable_prompt_multiple_select.ts#L44

timreichen:
I will close this in favour of multiple small PRs.

PR #6264, opened 2024-12-13
fix(cli/unstable): `promptSelect()` add `isTerminal()` check

timreichen:
ref https://github.com/denoland/std/pull/6257


kt3k (approved):
LGTM

PR #6268, opened 2024-12-14
docs(cli): remove `parseArgs()` console logs from examples

timreichen:
This PR removes all `parseArgs()` console logs and adds a constant declaration instead. This is how it is done in other examples and also prevents logs when calling `deno task test`.


kt3k (approved):
LGTM

PR #6269, opened 2024-12-14
docs(log): fix level name assertion in LogRecord example

ryunix:
Hello.

This pull request fixes the assertion in the `LogRecord` class test to correctly check the level name. The original assertion was checking `loggerName`, but I believe it should verify `levelName` instead.


kt3k (approved):
Good catch! Thanks

PR #6271, opened 2024-12-15
fix(expect): Fix `expect.arrayContaining` bug

eryue0220:
This pr is to fix https://github.com/denoland/std/issues/6243.


kt3k (approved):
LGTM. Thanks @eryue0220 !

PR #6272, opened 2024-12-16
fix(data-structures): ensure size consistency in BinarySearchTree.from

jiangshengdev:
Previously, when cloning an existing BinarySearchTree using the static from method, the new tree’s #size field was not updated, leading to a mismatch between the reported size and the actual number of nodes. This commit addresses the issue by directly assigning the original tree’s #size to the new instance after copying all nodes.

Additional tests have been added to verify that size remains accurate and consistent following the cloning process.


kt3k (approved):
Good catch! Thanks

kt3k:
Note: Confirmed the added assertions fail with the current implementation

PR #6275, opened 2024-12-17
fix(fs): use `stat` function in `stat` example

IgorM867:
The code example in the stat function imports the `stat` function but uses `Deno.stat` instead of `stat`.


kt3k (approved):
Good catch! Thanks

PR #6278, opened 2024-12-18
fix(fs): fix copying of symlink directories on Windows

IgorM867:
This PR fixes the handling of symlink directories in the `copy` function on Windows. Previously, copying a directory that was a symlink resulted in creating a symlink of type `"file"` instead of `"dir"`.

The test case for copying symlink directories passed because `Deno.lstat()` (which does not follow symlinks) was used. However, when `Deno.stat()` (which follows symlinks) was used on this type of symlink, the system would throw a `PermissionDenied` error.

### DIFF
--- a/fs/copy.ts
+++ b/fs/copy.ts
@@ -137,7 +137,7 @@ async function copySymLink(
 ) {
   await ensureValidCopy(src, dest, options);
   const originSrcFilePath = await Deno.readLink(src);
-  const type = getFileInfoType(await Deno.lstat(src));
+  const type = getFileInfoType(await Deno.lstat(originSrcFilePath));
   if (isWindows) {
     await Deno.symlink(originSrcFilePath, dest, {
       type: type === "dir" ? "dir" : "file",
@@ -161,7 +161,7 @@ function copySymlinkSync(
 ) {
   ensureValidCopySync(src, dest, options);
   const originSrcFilePath = Deno.readLinkSync(src);
-  const type = getFileInfoType(Deno.lstatSync(src));
+  const type = getFileInfoType(Deno.lstatSync(originSrcFilePath));
   if (isWindows) {
     Deno.symlinkSync(originSrcFilePath, dest, {
       type: type === "dir" ? "dir" : "file",


kt3k:
Thanks @IgorM867 for working on this!

I merged #6236 and this branch at #6280, but the CI shows some errors. Do you see the reason for those errors?

IgorM867:
@kt3k The reason is that in the `copySymLink()` function, when copying a symlink, `await Deno.readLink(src);` returns only a relative path ("dir"), not the full path. I believe this is due to how the symlinks were initially created. When I recreated them on Windows, all the tests passed.

fs/copy.ts
```diff
@@ -137,7 +137,7 @@ async function copySymLink(
 ) {
   await ensureValidCopy(src, dest, options);
   const originSrcFilePath = await Deno.readLink(src);
```
jbronder, line 139: Using `Deno.realPath` should return the absolute path of the symlink target. Additional explanation for suggestion in PR #6236:
```suggestion
  const originSrcFilePath = await Deno.realPath(src);
```

fs/copy.ts
```diff
@@ -161,7 +161,7 @@ function copySymlinkSync(
 ) {
   ensureValidCopySync(src, dest, options);
   const originSrcFilePath = Deno.readLinkSync(src);
```
jbronder, line 163: Using `Deno.realPathSync` should return the absolute path of the symlink target. Additional explanation for suggestion in PR #6236:
```suggestion
  const originSrcFilePath = Deno.realPathSync(src);
```

kt3k:
@IgorM867 

> @kt3k The reason is that in the `copySymLink()` function, when copying a symlink, `await Deno.readLink(src);` returns only a relative path ("dir"), not the full path. I believe this is due to how the symlinks were initially created. 

I think the symlink targeting relative path is fine, and I think we should keep the relative relationship between them while copying.

> When I recreated them on Windows, all the tests passed.

Did you create a symlink targeting absolute path? I don't think that works as absolute path vary in every environment

fs/copy.ts
kt3k: I think changing this to `realPath` causes another problem. I think we should keep this work for relative symlinks.

fs/copy.ts
```diff
@@ -137,7 +137,7 @@ async function copySymLink(
 ) {
   await ensureValidCopy(src, dest, options);
   const originSrcFilePath = await Deno.readLink(src);
-  const type = getFileInfoType(await Deno.lstat(src));
+  const type = getFileInfoType(await Deno.lstat(originSrcFilePath));
```
kt3k, line 140: I think this only works when `originSrcFilePath` is an absolute path or relative path accidentally points to path accessible from cwd. `originSrcFilePath` could be relative path from link path to target, which could be inaccessible from `cwd`, and in that case this `lstat` call would fail.

I think this should `Deno.stat(src)` instead with fallback to 'file' type in case `Deno.stat(src)` throws NotFound. What do you think?

PR #6279, opened 2024-12-19
chore: update title check CI

kt3k:


.github/workflows/title.yml
```diff
@@ -52,14 +52,14 @@ jobs:
             fmt(/unstable)?
             front-matter(/unstable)?
             fs(/unstable)?
+            functions(/unstable)?
```
kt3k, line 55: This prepares for #6143

PR #6286, opened 2024-12-20
feat(text/unstable): add `replaceStart`, `replaceEnd`, `replaceBoth` functions (#6265)

### LINKED ISSUE #6265: Suggestion: stripPrefix/stripSuffix functions

The Rust standard library has [strip_prefix](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix)/[strip_suffix](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_suffix) functions to remove a substring from the start or end of a string if it exists, or return `None`/`null` if the string doesn't have that substring at the start/end. I find myself wanting that function often when manipulating URLs or CLI output. Would that be a good fit for inclusion in `@std`? Maybe under `@std/text`? I would be happy to submit a PR.

```js
import { stripPrefix, stripSuffix } from 'jsr:@std/text';

assert(stripPrefix('string', 'str') === 'ing');
assert(stripPrefix('string', 'abc') === null);

assert(stripSuffix('string', 'ing') === 'str');
assert(stripSuffix('string', 'abc') === null);
```

Just writing a helper function whenever I need it in each project. I think it would be a generally useful feature though.

lionel-rowe:
Resolves #6265.

Roughly speaking, these replicate the combined functionality of [`trim_matches`](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_matches), [`trim_start_matches`](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_start_matches), [`trim_end_matches`](https://doc.rust-lang.org/std/string/struct.String.html#method.trim_end_matches), [`strip_prefix`](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_prefix), and [`strip_suffix`](https://doc.rust-lang.org/std/string/struct.String.html#method.strip_suffix) (plus a hypothetical `strip_prefix_and_suffix`) from the Rust standard library.


text/unstable_strip.ts
```diff
@@ -0,0 +1,161 @@
...
+): string {
+  let prev = str;
+
+  const count = options?.count ?? Infinity;
+
+  for (let i = 0; i < count; ++i) {
+    str = str.replace(regExp, "");
+    if (str === prev) break;
+    prev = str;
+  }
```
babiabeo, line 147: Why `Infinity`? If the string doesn't include the pattern to strip, it will loop forever, won't it?

lionel-rowe: > Why `Infinity`? If the string doesn't include the pattern to strip, it will loop forever, won't it?

It'll break on the `str === prev` condition as soon as there's nothing left to strip.

```ts
const aaa = 'a'.repeat(1000)
const aab = aaa + 'b'
const bbb = 'b'.repeat(1000)

stripStart(aaa, 'a') === ''
stripStart(aab, 'a') === 'b'
stripStart(bbb, 'a') === bbb
```

timreichen:
I think the `count` option should not be part of this api. Instead, the user can always call the function repeatedly or in a loop themselves. In that sense, I think `stripStart()` and `stripEnd()` should be the only exported functions, and maybe a `stripAll()` to reflect `replaceAll()`.

lionel-rowe:
> I think the count option should not be part of this api. Instead, the user can always call the function repeatedly or in a loop themselves. In that sense, I think stripStart() and stripEnd() should be the only exported functions, and maybe a stripAll() to reflect replaceAll().

@timreichen Probably worth getting some more use cases, but if I grep my code for functions named `*strip*`, mostly they simply strip out _all_ occurrences (usually based on `replaceAll` or similar, no need for a library function), and in a significant minority of cases they remove all prefixes and suffixes, like this:

```ts
// strip(str, /[^\p{L}\p{M}]+/gu, { count: Infinity })
function stripNonLetter(str: string) {
    return str.replaceAll(/^[^\p{L}\p{M}]+|[^\p{L}\p{M}]+$/gu, '')
}

// strip(str, /[_\- ]+/g, { count: Infinity })
const strip = (str: string) => str.replaceAll(/^[_\- ]+|[_\- ]+$/g, '')
```

There are also a few cases where they strip a single instance at the start or end and then stop there, and it sounds like @canac 's use case fits here:

```ts
// strip(str, '\ufeff', { count: 1 })
function stripBom(str: string) {
   return str.replace(/^\ufeff/, "")
}
```

I guess the "strip up to `n` occurrences" case is rare when `1 < n < Infinity`, but very common (possibly the most common) when `n == Infinity`, and also reasonably common where `n == 1`. So maybe a better API is `options?: { all?: boolean }` or `options?: { single?: boolean }`, depending on what the chosen default is (no strong opinion either way on the default).

I'm not keen on the idea of doubling the API surface area from `strip/stripStart/stripEnd` to `strip/stripStart/stripEnd/stripAll/stripAllStart/stripAllEnd`, even though that's essentially what Rust std library does. Seems like overkill in JS/TS.

Another way of supporting single/all, but without a specific option, would be reading the `g` flag if the supplied pattern is a regex. I'm also not keen on this option as it doesn't support string patterns, and it's also not really the correct semantics for the `g` flag. Currently I just ignore `g` and `y` flags, but maybe throwing would be better if `g` doesn't match all-ness (and always throwing if `y` supplied).

timreichen:
> @timreichen Probably worth getting some more use cases, but if I grep my code for functions named `*strip*`, mostly they simply strip out _all_ occurrences (usually based on `replaceAll` or similar, no need for a library function), and in a significant minority of cases they remove all prefixes and suffixes, like this:
> 
> ```ts
> // strip(str, /[^\p{L}\p{M}]+/gu, { count: Infinity })
> function stripNonLetter(str: string) {
>     return str.replaceAll(/^[^\p{L}\p{M}]+|[^\p{L}\p{M}]+$/gu, '')
> }
> 
> // strip(str, /[_\- ]+/g, { count: Infinity })
> const strip = (str: string) => str.replaceAll(/^[_\- ]+|[_\- ]+$/g, '')
> ```
> 
> There are also a few cases where they strip a single instance at the start or end and then stop there, and it sounds like @canac 's use case fits here:
> 
> ```ts
> // stripStart(str, '\ufeff', { count: 1 })
> function stripBom(str: string) {
>    return str.replace(/^\ufeff/, '')
> }
> ```
> 
> I guess the "strip up to `n` occurrences" case is rare when `1 < n < Infinity`, but very common (possibly the most common) when `n == Infinity`, and also reasonably common where `n == 1`. So maybe a better API is `options?: { all?: boolean }` or `options?: { single?: boolean }`, depending on what the chosen default is (no strong opinion either way on the default).

I agree, `strip` as a name is confusing as it doesn't indicate single or all occurrences.

> I'm not keen on the idea of doubling the API surface area from `strip/stripStart/stripEnd` to `strip/stripStart/stripEnd/stripAll/stripAllStart/stripAllEnd`, even though that's essentially what Rust std library does. Seems like overkill in JS/TS.

I wouldn't like that either. I think the problem here is that the functions are explicit in name but are n

> Another way of supporting single/all, but without a specific option, would be reading the `g` flag if the supplied pattern is a regex. I'm also not keen on this option as it doesn't support string patterns, and it's also not really the correct semantics for the `g` flag. Currently I just ignore `g` and `y` flags, but maybe throwing would be better if `g` doesn't match all-ness (and always throwing if `y` supplied).

Come to think of it, this api is tries to do too much combining all these rust functions into one imo. There is [String.prototype.replace](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace), [String.prototype.replaceAll](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll), which already cover a lot of use cases.

This api should be reduced to `replaceStart()` and `replaceEnd()`, which makes them actually more flexible, not only stripping but allow for any replacement:

strip_prefix => `replaceStart("foo:bar", "foo:", "") // "bar"`
strip_suffix => `replaceEnd("bar:foo", ":foo", "") // "bar"`
trim_start_matches => `replaceStart("11foo1bar", /1+/, "") // "foo1bar"`
trim_end_matches => `replaceEnd( "foo1bar11", /1+/, "") // "foo1bar"`
trim_matches => `"11foo1bar11".replace(/^1+|1+$/, "") // "foo1bar"` or call both `replaceStart()` and `replaceEnd()` 

N replacements can be handled by the regexp, using a prepared string or calling the function repeatedly:
`replaceEnd("barfoofoofoo", /(foo){2}/, "") // "barfoo"` or 
`replaceEnd("barfoofoofoo", "foo".repeat(2), "") // "barfoo"` or 
`for (let i = 0; i < 2; i++) replaceEnd("barfoofoofoo", "foo", "") // "barfoo"`

lionel-rowe:
> N replacements can be handled by the regexp, using a prepared string or calling the function repeatedly:
> `replaceEnd("barfoofoofoo", /(foo){2}/, "") // "barfoo"` or
> `replaceEnd("barfoofoofoo", "foo".repeat(2), "") // "barfoo"` or
> `for (let i = 0; i < 2; i++) replaceEnd("barfoofoofoo", "foo", "") // "barfoo"`

@timreichen None of these are really satisfactory for the case where you want to dynamically strip arbitrarily many occurrences of an arbitrary string, but none of the use cases I've found involve doing that anyway — use cases with unlimited occurrences typically seem to be more suited to regex, where you can trivially use a non-capturing group along with `*` or `+`.

As for `replace` vs `strip`:

```ts
replaceEnd(url.pathname, /\/*/, '/') // pathname always ends with slash
stripEnd(url.pathname, /\/*/) + '/' // same result
```

But it could be useful for cases like this:

```ts
// only prepend 'https:' when replacing existing 'http:' protocol
replaceStart(url.href, 'http:', 'https:')

// doesn't work with `strip` version as the protocol might be something else
// e.g. you might end up with `https:file:///...`
'https:' + stripStart(url.href, 'http:')
```

Using the name `replace*` along with allowing regexes weakly implies that the string-replacement syntax will be supported, which would add a moderate amount of implementation complexity — something like [this](https://github.com/lionel-rowe/irregex/blob/7a16c1dc5675ac6f0c809c694ed2f17763694429/src/irregex.ts#L131-L181).

```ts
'abcxyzabcxyz'.replace(/[abc]+/, '<$&>') // '<abc>xyz<abc>xyz'

// naive result would be `<$&>xyzabcxyz`
// but seems reasonable to expect result to be `<abc>xyzabcxyz`
replaceStart('abcxyzabcxyz', /[abc]+/, '<$&>')
```

Conversely, supporting callback replacements would be trivial, as you can just conditionalize the replacement based on whether the index lies at the start/end.

```ts
replaceStart('abcxyzabcxyz', /[abc]+/, (m) => `<${m}>`)
```

Also, the name `replace` for replace-both-ends would be confusing due to `String#replace`. Maybe `replaceEnds` or something. Requiring calling 2 functions just to strip/replace both ends is suboptimal as it requires either declaring an extra variable or repeating code.

timreichen:
How about just `replaceStart()` and `replaceEnd()` just supporting strings without regexp support?
I feel if one wants to do a complex strip/replacement with a regexp, prepending `^` or appending `$` and using the built in `replace()` function is more suited.
I also don't think we should provide a `replaceBoth()` function. The name is kinda weird and it is just a small convenience function which can be done by the user with one line.
If there are requests for more functionality by multiple users, we still can add to that.

kt3k:
I'm not in favor of adding `replaceStart`, `replaceEnd`, or `replaceBoth`. They look overlapping too much with `String.prototype.replace` and `String.prototype.replaceAll`. What problems do they solve?

Also why did we switch from `strip*` APIs to `replace*` APIs in this PR? I don't think this PR resolve #6265 anymore

lionel-rowe:
> I'm not in favor of adding replaceStart, replaceEnd, or replaceBoth. They look overlapping too much with String.prototype.replace and String.prototype.replaceAll. What problems do they solve?

I've tried to stick to practical `@example`s in the JSDoc documentation, which show realistic cases where each function would be useful.

> Also why did we switch from strip* APIs to replace* APIs in this PR? I don't think this PR resolve https://github.com/denoland/std/issues/6265 anymore

Seems like a minimal increase in complexity/surface area that has some handy use cases, and `replace*(str, x, '')` is easily understandable as `strip*(str, x)` for anyone familiar with the common `str.replace(x, '')` paradigm.

Possibly the problem space is too poorly defined at the moment, as I think canac, timreichen, you, and I all have slightly different ideas about what's useful here. That's partly why I've tried to stick to practical examples in the documentation, so we can discuss concretely about which of them make sense. E.g. I think `replaceStart(url.href, 'http:', 'https:')` is a decently compelling case for replace > strip, and `replaceBoth(str, /[^\p{L}\p{M}\p{N}]+/u, "")` is a strongly compelling case for `*Both` and regex support. But maybe there's no compelling case for the _combination_ of `*Both`, regex support, and replacement.

kt3k:
I don't see much support/interest from the community to this feature. Let's not do this.

Thanks for your contribution anyways

PR #6292, opened 2024-12-22
docs(expect): clarify the tolerance in `toBeCloseTo`

dandv:
I was looking for a more pedestrian matcher, that would assert a number being [within](https://jest-extended.jestcommunity.dev/docs/matchers/Number#tobewithinstart-end) a range, or ±X of the expected value. Clarified `toBeCloseTo` as useful for avoiding floating point precision errors instead.

### DIFF
--- a/expect/_types.ts
+++ b/expect/_types.ts
@@ -178,7 +178,7 @@ export interface Expected<IsAsync = false> {
 
   /**
    * Asserts that a given numerical value is approximately equal to an
-   * expected number within a within a certain margin of error
+   * expected number within a certain margin of error
    * (tolerance). Useful when comparing floating-point numbers, which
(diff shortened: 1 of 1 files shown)


expect/_types.ts
```diff
@@ -177,7 +177,10 @@ export interface Expected<IsAsync = false> {
   toBeCalledWith(...expected: unknown[]): void;
 
   /**
-   * Asserts that the value is close to the specified number within a tolerance.
+   * Asserts that a given numerical value is approximately equal to an
+   * expected number within a within a certain margin of error
```
kt3k, line 181:
```suggestion
   * expected number within a certain margin of error
```

kt3k (approved):
LGTM. Nice docs!

PR #6295, opened 2024-12-22
fix(fmt): add correct pluralization to fmt/duration when using style: full

### LINKED ISSUE #6293: @std/fmt/duration singularize words when the value === 1
Currently the full unit words (seconds, minutes, hours, etc) are plural in all cases, even when the number is 1. It seems fairly simple to only pluralize when the number != 1, which would produce more idiomatic, grammatically correct English phrases ( "1 minute" vs "1 minutes"). Since I believe that's the point of the `style: "full"` setting, this seems like a worthwhile update.

Perhaps there's some complexity there I haven't considered, but I've created a quick PR to show how it might be done. Happy to discuss.

GriffinJohnston:
fix #6293

### DIFF
--- a/fmt/duration.ts
+++ b/fmt/duration.ts
@@ -14,7 +14,7 @@
  *
  * assertEquals(format(99674, { ignoreZero: true }), "1m 39s 674ms");
  *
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Looks reasonable to me. Thanks for your PR!

PR #6301, opened 2024-12-24
docs(datetime): remove outdated mention of `utc` option

yibe:
Fixes a trivial oversight in the docs that was missed in 4ec7dd4be99049e6721d4bf41b5a4d3ded7770b2 (#5647)


kt3k (approved):
LGTM. Good catch! Thanks

PR #6305, opened 2024-12-25
BREAKING(http/unstable): add capability to attach handlers by methods, implement sensible defaults

zkerezsi:
Hi everyone!

I would like to propose a new implementation of the `route()` function that adds the capability to define handlers on a per-method basis (in a style similar to [Fresh](https://fresh.deno.dev/docs/concepts/routes)) or to use a single handler that handles all methods.

I envision the `route()` function being used only once when we pass it to `Deno.serve()`. In the case of a REST API, one would:

- Export an array of routes in `user.ts` with URL patterns such as `/api/users` and `/api/users/:id`.
- Export an array of routes in `products.ts` with URL patterns such as `/api/products` and `/api/products/:id`.
- Import these in `main.ts`, then call `Deno.serve(route([...userRoutes, ...productRoutes, notFoundRoute], errorHandler))`.

The current `defaultHandler()` does not seem very useful, as a wildcard route can provide the same functionality (see tests). I believe an optional `errorHandler()` would be more important.

While my PR adds some logic to the `route()` function, it conforms to the relevant specifications (RFC2616) and practices regarding the "Method Not Allowed" response, error handling, and the default "Not Found" response, at least to my understanding. Furthermore, all of the new default behavior remains customizable by the user.

Accepting this PR in its current form would essentially mean that the `URLPattern` would be the **only factor** determining whether a Route can handle the request or not.

### DIFF
--- a/http/unstable_route.ts
+++ b/http/unstable_route.ts
@@ -5,7 +5,7 @@
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
- * Extends {@linkcode Deno.ServeHandlerInfo} by adding adding a `params` argument.
+ * Extends {@linkcode Deno.ServeHandlerInfo} by adding a `params` argument.
  *
  * @param request Request
  * @param info Request info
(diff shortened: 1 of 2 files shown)


http/unstable_route.ts
```diff
@@ -96,15 +169,32 @@ export function route(
...
+
+      try {
+        return handler(request, match, info);
+      } catch (error) {
+        return errorHandler(error, request, match, info);
       }
     }
-    return defaultHandler(request, info);
+
+    return new Response("Not Found", { status: 404 });
```
kt3k, line 198: I don't think it's a good idea to remove the capability to customize the 404 page

http/unstable_route.ts
```diff
@@ -96,15 +169,32 @@ export function route(
...
+      if (match === null) {
+        continue;
+      }
+
+      let handler: Handler;
+      if (route.handler) {
+        handler = route.handler;
+      } else if (!(request.method in route.handlers)) {
+        /**
+         * @see {@link https://www.iana.org/go/rfc2616 | RFC2616, Section 14.7}
```
kt3k, line 181: I think this should reference rfc9110 instead (rfc2616 is now obsolete

kt3k:
Thanks for the suggestion.

The addition of errorHandler option seems interesting to me. But we probably don't need the default implementation as `Deno.serve` already has default error handling which returns the equivalent of `new Response("Internal Server Error", { status: 500 })`.

> I would like to propose a new implementation of the route() function that adds the capability to define handlers on a per-method basis (in a style similar to [Fresh](https://fresh.deno.dev/docs/concepts/routes)) or to use a single handler that handles all methods.

I don't see much benefit of this change. This design seems rather confusing to me as `handler` and `handlers` are similar in names, but have different effects.

http/unstable_route.ts
kerezsiz42: I think I mentioned it in the PR but did not emphasize it enough. So putting a wildcard route that matches any URLPattern at the end of the Route[] achieves exactly the same result as the current defaultHandler:

```ts
const wildcardRoute: Route = {
  pattern: new URLPattern({ pathname: "/*" }),
  handler: (request: Request) => {
    return new Response(new URL(request.url).pathname, { status: 404 });
  },
};
```

For this reason I think that the defaultHandler is unnecessary, and the second parameter of route could be deleted or changed to something else without losing features.

kerezsiz42:
> Thanks for the suggestion.
> 
> The addition of errorHandler option seems interesting to me. But we probably don't need the default implementation as `Deno.serve` already has default error handling which returns the equivalent of `new Response("Internal Server Error", { status: 500 })`.
> 
> > I would like to propose a new implementation of the route() function that adds the capability to define handlers on a per-method basis (in a style similar to [Fresh](https://fresh.deno.dev/docs/concepts/routes)) or to use a single handler that handles all methods.
> 
> I don't see much benefit of this change. This design seems rather confusing to me as `handler` and `handlers` are similar in names, but have different effects.

Thanks for the code review!

You did not criticize what I thought you would. So if I understand you correctly, only the naming of `handler` and `handlers` is problematic, and you don't like the redundancy that the `errorHandler()` parameter's default value provides. How should they be named in your opinion? If I remove the default value of `errorHandler()`, will it be acceptable?

I understand that this module of the std library is not on agenda currently, but we, as library developers should be concentrating on what the users of the standard library want. And a high quality HTTP toolkit is of utmost importance as we see it in case of Go. For new JS/TS projects the default option for new users should be `@std/http`, so we should strive to be better in ergonomics and ease of use compared to Hono, Oak and others while keeping the solution as bare-bones as possible.

I am new to open source development, so I might not be familiar with how this kind of development really works. I am keen to hearing your advice and opinion!

PR #6307, opened 2024-12-26
fix(data-structures): ensure size consistency in RedBlackTree.from

jiangshengdev:
Previously, when cloning an existing RedBlackTree using the static from method, the new tree’s #size field was not updated, leading to a mismatch between the reported size and the actual number of nodes. This commit addresses the issue by directly assigning the original tree’s #size to the new instance after copying all nodes.

Additional tests have been added to verify that size remains accurate and consistent following the cloning process.


kt3k (approved):
LGTM. Thanks!

PR #6312, opened 2024-12-28
fix(async): `abortable` should prevent uncaught error when promise is rejected

Milly:
In `abortable`:
- If the signal is aborted with delay and the promise is rejected, an uncaught error does not occur. This is OK.
- If the signal is already aborted and the promise is rejected, an uncaught error occurs.

Uncaught errors should not occur in either case.

Also, I changed the tests to use `delay` instead of `clearTimeout` to detect leaks.

### DIFF
--- a/async/abortable.ts
+++ b/async/abortable.ts
@@ -127,9 +127,9 @@ function abortablePromise<T>(
   p: Promise<T>,
   signal: AbortSignal,
 ): Promise<T> {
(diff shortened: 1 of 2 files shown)


kt3k:
The fix looks good to me.

> Also, I changed the tests to use delay instead of clearTimeout to detect leaks.

This change makes the test cases flaky like this one https://github.com/denoland/std/actions/runs/12528301271/job/34942765393

Can you revert these changes?

Issue #6289, opened 2024-12-22
@std/expect unify internals with @std/testing/mock


anion155:
**Is your feature request related to a problem? Please describe.**

For now, there is a misalignment between `@std/expect` and `@std/testing/mock`, which causes inability to use `@std/expect/fn` in `@std/testing/mock`'s `assertSpyCalls` api, and `@std/testing/mock` in `@std/expect`'s `expect().toHaveBeenCalledTimes()` api. Which and prevents us from using both of them at the same time (totally possible for most of the other's apis of both modules).

This seems highly unreasonable, taking into account that both modules uses `@std/internal` and also `@std/expect` using `@std/assert` for some of it's functionalities like:
- same `AssertError` class
- `toBe/toBeDefined/toBeUndefined/toBeNull` uses `assertNotStrictEquals` and `assertStrictEquals`
- `toBeInstanceOf` uses `assertInstanceOf` and `assertNotInstanceOf`
- etc.

**Describe the solution you'd like**

Implement some core mock internals functionality in `@std/internal` and use it in both modules, agnostic to specifics of `@std/assert` api and `@std/expect` jest compatibility, like:
- detection if function is mock function
- gathering of internal calls state

**Describe alternatives you've considered**

Using compatible between modules storage of mocked function calls, like in `@std/expect` storing calls in `.calls` field array, or the other way around in `@std/testing/mock` storing calls in `[Symbol.for("@MOCK")]` field

**PS**

I'm willing to start working on this issue my self in my free time, if that's ok

**PS**

Just wanted to mention issue about [`@std/expect/fn` not being 'Jest compatible'](https://github.com/denoland/std/issues/6247)

PR #6317, opened 2024-12-30
feat(internal,expect,testing): `@std/expect` jest compatible fn and unified mock experience with `@std/testing/mock`

anion155:
## Summary

Made an implementation of jest compatible `@std/expect`'s `fn` utility.
Implementation made with side eyeing latest version of this api in jest https://jestjs.io/docs/mock-function-api .

```ts
interface ExpectMockInstance<Args extends unknown[], Return> {
  /**
   * Sets current implementation.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn().mockImplementation((a: number, b: number) => a + b);
   *   expect(mockFn()).toEqual(3);
   * });
   * ```
   */
  mockImplementation(stub: (...args: Args) => Return): this;

  /**
   * Adds one time stub implementation.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockImplementationOnce((a, b) => a - b);
   *   expect(mockFn(1, 2)).toEqual(-1);
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  mockImplementationOnce(stub: (...args: Args) => Return): this;

  /**
   * Sets current implementation's return value.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn().mockReturnValue(5);
   *   expect(mockFn(1, 2)).toEqual(5);
   * });
   * ```
   */
  mockReturnValue(value: Return): this;

  /**
   * Adds one time stub implementation that returns provided value.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockReturnValueOnce(5);
   *   expect(mockFn(1, 2)).toEqual(5);
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  mockReturnValueOnce(value: Return): this;

  /**
   * Sets current implementation's that returns promise resolved to value.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", async () => {
   *   const mockFn = fn();
   *   mockFn.mockResolvedValue(5);
   *   await expect(mockFn(1, 2)).resolves.toEqual(5);
   *   expect(mockFn(3, 2)).toEqual(5);
   * });
   * ```
   */
  mockResolvedValue(value: Awaited<Return> | Return): this;

  /**
   * Adds one time stub implementation that returns promise resolved to value.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", async () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockResolvedValueOnce(5);
   *   await expect(mockFn(1, 2)).resolves.toEqual(5);
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  mockResolvedValueOnce(value: Awaited<Return> | Return): this;

  /**
   * Sets current implementation's that returns promise rejects with reason.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", async () => {
   *   const mockFn = fn();
   *   mockFn.mockRejectedValue(new Error("test error"));
   *   await expect(mockFn(1, 2)).rejects.toThrow("test error");
   *   expect(mockFn(3, 2)).toEqual(5);
   * });
   * ```
   */
  mockRejectedValue(reason?: unknown): this;

  /**
   * Adds one time stub implementation that returns promise rejects with reason.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", async () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockRejectedValueOnce(new Error("test error"));
   *   await expect(mockFn(1, 2)).rejects.toThrow("test error");
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  mockRejectedValueOnce(reason?: unknown): this;

  /**
   * Changes current implementation to provided stub.
   * Returns disposable resource that restores previous setup of stubs on dispose.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockReturnValueOnce(5);
   *   {
   *     using withMock = mockFn.withImplementation((a, b) => a - b);
   *     expect(mockFn(1, 2)).toEqual(-1);
   *   }
   *   expect(mockFn(1, 2)).toEqual(5);
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  withImplementation(stub: (...args: Args) => Return): Disposable;

  /**
   * Changes current implementation to provided stub.
   * Runs scope function and after it is final restores previous setup of stubs.
   * Also detects if scope function returns a promise and waits for it to resolve.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", async () => {
   *   const mockFn = fn((a: number, b: number) => a + b);
   *   mockFn.mockReturnValueOnce(5);
   *   await mockFn.withImplementation(async (a, b) => a - b, async () => {
   *     await expect(mockFn(1, 2)).resolves.toEqual(-1);
   *   });
   *   expect(mockFn(1, 2)).toEqual(5);
   *   expect(mockFn(1, 2)).toEqual(3);
   * });
   * ```
   */
  withImplementation<ScopeResult>(
    stub: (...args: Args) => Return,
    scope: () => ScopeResult
  ): ScopeResult;

  /**
   * Restores original implementation and discards one time stubs.
   * In case no original implementation was provided, the mock will be reset to an empty function.
   *
   * @example Usage
   * ```ts
   * import { fn, expect } from "@std/expect";
   *
   * Deno.test("example", () => {
   *   const mockFn = fn((a: number, b: number) => a + b).mockReturnValue(5).mockReturnValueOnce(1);
   *   mockFn.mockRestore();
   *   expect(mockFn()).toEqual(3);
   *   expect(mockFn()).toEqual(3);
   * });
   * ```
   */
  mockRestore(): void;
}
```

**Issues that this pr covers:**
- https://github.com/denoland/std/issues/6290
- https://github.com/denoland/std/issues/6289
- https://github.com/denoland/std/issues/6247


anion155:
@kt3k Hi! Maybe you will have any thoughts on this?

kt3k:

Addition of Jest compatible methods are welcome, but it doesn't seem desirable to make `@std/expect`'s `fn` and `@std/testing/mock` compatible. Such compatibility was never intended and never asked by users before #6289. That unification seems introducing unnecessary interdependency between them and making the code base harder to work with (The contributors of `@std/expect` and `@std/testing/mock` are very separate. If we land this PR, they now need to understand both packages. This feels like introducing contribution barrier).

andrewthauer:
From my perspective as a consumer of `@std` first, `fn` isn't overly useful by itself without the ability to mock and stub. The fact the only way to stub/spy on in the other package and not compatible is an issue. I've wanted these to be compatible and I've had co-workers confused as well why they do not work together either.

anion155:
> Sorry for the delay in review.
> 
> Addition of Jest compatible methods are welcome, but it doesn't seem desirable to make `@std/expect`'s `fn` and `@std/testing/mock` compatible. Such compatibility was never intended and never asked by users before #6289. That unification seems introducing unnecessary interdependency between them and making the code base harder to work with (The contributors of `@std/expect` and `@std/testing/mock` are very separate. If we land this PR, they now need to understand both packages. This feels like introducing contribution barrier).

I think compatibility between `fn` and `spy` is essential for the smooth transition of old codebases from jest tests to Deno tests in the future.

I don't agree that there is interdependency between these modules in current. All the common types and principals are moved to the `@std/internal/mock` which will let to create common standard interface for any `mock` module, and will basically guarantee that any `mock` module and assertion library that uses it will be interchangeable without creating dependencies between them

anion155:
> some adapter

About adapters. Event though it will be much simpler in implementation and less in existing code changes, this approach will introduce the exact issue of dependencies between modules @kt3k was talking about. Every `mock` library will need to be aware of every other `mock` library.

andrewthauer:
> > some adapter
> 
> 
> 
> About adapters. Event though it will be much simpler in implementation and less in existing code changes, this approach will introduce the exact issue of dependencies between modules @kt3k was talking about. Every `mock` library will need to be aware of every other `mock` library.

Possibly. I honestly didn't really think through how this would work. However, I completely agree this that first class mock/spy support for expect/fn are essential for a good experience.

PR #6318, opened 2024-12-30
fix(cli): Handle overflow in `promptSecret`

cosmastech:
To close https://github.com/denoland/std/issues/6306

There is an issue where backspacing characters from multiple lines will not move the cursor back to the original lines. I welcome any input on how to improve this, or if it's necessary.

https://github.com/user-attachments/assets/56c58805-4f9e-4150-9c50-fd2939505706


kt3k:
Thanks for the PR. This looks much better than the current state!

However if I filled the first 2 lines and then deleted the entire 2nd line, then the initial prompt part was duplicated again.. Could you take a look at this point as well?

(I guess we can use 'Cursor Control' codes to remove the entire prompt + masks https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#cursor-controls)

cosmastech:
> Thanks for the PR. This looks much better than the current state!
> 
> 
> 
> However if I filled the first 2 lines and then deleted the entire 2nd line, then the initial prompt part was duplicated again.. Could you take a look at this point as well?
> 
> 
> 
> (I guess we can use 'Cursor Control' codes to remove the entire prompt + masks https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#cursor-controls)

Absolutely!

Two thoughts:

1. Would it be beneficial to create a "cliWindow" utility that sets a height/width and handles overflow? Seems like it could be useful more generally. I _think_ it may be necessary to achieve this, because
2. How do we handle screen resizes? We have to know the former width so we can figure out how many lines to move back and replace. (This is more of a "I think," as I haven't played around yet)

cosmastech:
@kt3k I believe your issue is addressed. I've updated the description to include a screen recording of the current implementation.

There's still an issue where resizing the screen after prompting causes undesirable behavior. In order to fix this, we need to keep track of the console size and check it before writing anything to the terminal. Is calling Deno.consoleSize() an expensive operation, or is it safe to do on each character press?

An example of what happens when resizing a terminal window:

https://github.com/user-attachments/assets/203594d8-6647-44b0-8b76-a345b3b712f5

cli/prompt_secret_test.ts
```diff
@@ -526,3 +556,193 @@ Deno.test("promptSecret() handles empty readSync", () => {
...
+  );
+
+  const password = promptSecret("?");
+
+  assertEquals(password, "deno rules");
+  assertEquals(expectedOutput, actualOutput);
+  restore();
+});
+
+Deno.test("promptSecret() returns to previous line when deleting characters", () => {
```
kt3k, line 636: Nice test case!

cli/prompt_secret.ts
```diff
@@ -52,12 +53,41 @@ export function promptSecret(
...
+      );
+    }
+
+    // If the user has deleted a character
+    if (currentLength < previousLength) {
+      // Then clear the current line.
+      output.writeSync(CLR);
+      if (charsPastLineLength === 0) {
+        // And if there's no characters on the current line, return to previous line.
+        output.writeSync(MOVE_LINE_UP);
```
kt3k, line 77: Looks nicely commented 👍

kt3k (approved):
Confirmed this addresses the issue, and also they seem unit tested well. Thanks for your contribution!

Issue #6330, opened 2025-01-04
_tools/check_mod_exports.ts: `modFilePath` extracts file paths that are not strictly `mod.ts` files


jbronder:
**Describe the bug**
In the process of adding a file named `unstable_chmod.ts` to the `@std/fs` package to add towards #6255, running `deno task lint:mod-exports`the `_tools/check_mod_exports.ts` script will interpret `unstable_chmod.ts` as a `mod.ts` file and will emit warnings that fail the lint task.

**Steps to Reproduce**
- Create a file that ends in `*mod.ts`, (e.g. `fs/unstable_chmod.ts` or `fs/another_file_mod.ts`) in the `@std/fs` package. 
- Run `deno task lint:mod-exports`.

```bash
❯ ls fs/
_create_walk_entry.ts       empty_dir_test.ts           mod.ts
_get_file_info_type.ts      ensure_dir.ts               move.ts
_get_file_info_type_test.ts ensure_dir_test.ts          move_test.ts
_is_same_path.ts            ensure_file.ts              testdata
_is_same_path_test.ts       ensure_file_test.ts         unstable_chmod.ts  # <-- newly created file
_is_subdir.ts               ensure_link.ts              unstable_chmod_test.ts
_is_subdir_test.ts          ensure_link_test.ts         unstable_errors.d.ts
_map_error.ts               ensure_symlink.ts           unstable_errors.js
_to_file_info.ts            ensure_symlink_test.ts      unstable_lstat.ts
_to_path_string.ts          eol.ts                      unstable_lstat_test.ts
_utils.ts                   eol_test.ts                 unstable_stat.ts
copy.ts                     exists.ts                   unstable_stat_test.ts
copy_test.ts                exists_test.ts              unstable_types.ts
deno.json                   expand_glob.ts              walk.ts
empty_dir.ts                expand_glob_test.ts         walk_test.ts

❯ deno task lint:mod-exports # NOTE: absolute path not shown for brevity
Warn /Users/... .../deno_std/fs/unstable_chmod.ts does not export './eol.ts'.
Warn /Users/... .../deno_std/fs/unstable_chmod.ts does not export './empty_dir.ts'.
Warn /Users/... .../deno_std/fs/unstable_chmod.ts does not export './ensure_file.ts'.
Warn /Users/... .../deno_std/fs/unstable_chmod.ts does not export './ensure_dir.ts'.
[...]

PR #6331, opened 2025-01-05
chore: fix regex to match mod.ts files (#6330)

jbronder:
This PR is to address issue #6330. A change to the `match` property is proposed such that only `mod.ts` files are checked in the lint task of `deno task lint:mod-exports`.  This change is intended to prevent including files that are not `mod.ts` files, but for the purposes of naming files (e.g. `unstable_chmod.ts`) which happen to end with a `*mod.ts` in the file name.


_tools/check_mod_exports.ts
```diff
@@ -16,7 +16,7 @@ for await (
   const { path: modFilePath } of walk(ROOT, {
     includeDirs: true,
     exts: ["ts"],
-    match: [/mod\.ts$/],
+    match: [/\/mod\.ts$/],
```
kt3k, line 19:
```suggestion
    match: [/(\/|\\)mod\.ts$/],
```

kt3k (approved):
Good catch! Thanks

jbronder:
Oh yeah, I missed the double backslash.  Thank you for looking it over!

Issue #6335, opened 2025-01-06
`uniqBy`


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

cf https://lodash.com/docs/4.17.15#uniqBy (and equivalents in many other popular libraries/languages).

`uniq` probably isn't useful as as `[...new Set(arr)]` is trivial, but `uniqBy` would be a handy addition to `@std/collections`, as cases such as deduplicating API responses on a field such as ID are very common.

**Describe the solution you'd like**

```ts
const users = [{ name: 'Alice', id: 88 }, { name: 'Bob', id: 99 }, { name: 'Alice_2', id: 88 }]
uniqBy(users, (user) => user.id)
// [{ name: 'Alice', id: 88 }, { name: 'Bob', id: 99 }]
```

**Describe alternatives you've considered**

Possibly also allow shorthand similar to lodash:

```ts
uniqBy(users, 'id')
// [{ name: 'Alice', id: 88 }, { name: 'Bob', id: 99 }]
```

lionel-rowe:
nvm, looks like this is called `distinctBy`. I'll open a docs PR with the word "unique" to make it greppable.

PR #6336, opened 2025-01-06
docs(collections): add the word "unique" to `distinctBy` JSDoc for grepping purposes (#6335)

lionel-rowe:
See https://github.com/denoland/std/issues/6335

Currently, grepping for "uniq" or "unique" yield zero results:

image

This PR is mainly to fix that problem, along with a note about how equality is determined.

Also:
* Changed name of callback param from `selector` to `discriminator` — "selector" is confusing as it gets the respective key from each item, rather than selecting items
* Changed example to something more representative of typical usage


collections/distinct.ts
```diff
@@ -5,6 +5,8 @@
  * Returns all distinct elements in the given array, preserving order by first
  * occurrence.
  *
+ * Uniqueness is determined by same-value-zero equality.
```
kt3k, line 8:
```suggestion
 * Uniqueness is determined by [same-value-zero](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness#same-value-zero_equality) equality.
```

kt3k (approved):
All suggestions look reasonable to me. Thanks for the PR!

Issue #6255, opened 2024-12-12
Port File System APIs in Deno namespace to `@std/fs`


kt3k:
We aim to make `@std/fs` cross-runtime (ref #4313). To achieve that we need to port FS APIs from Deno namespace (with Node.js compatibility)

https://docs.deno.com/api/deno/file-system

### Classes
- [ ] FsFile

### Functions
- [ ] chmod
- [ ] chmodSync
- [ ] chown
- [ ] chownSync
- [ ] copyFile
- [ ] copyFileSync
- [ ] create
- [ ] createSync
- [ ] link
- [ ] linkSync
- [ ] lstat
- [ ] lstatSync
- [ ] makeTempDir
- [ ] makeTempDirSync
- [ ] makeTempFile
- [ ] makeTempFileSync
- [ ] mkdir
- [ ] mkdirSync
- [ ] open
- [ ] openSync
- [ ] readDir
- [ ] readDirSync
- [ ] readFile
- [ ] readFileSync
- [ ] readLink
- [ ] readLinkSync
- [ ] readTextFile
- [ ] readTextFileSync
- [ ] realPath
- [ ] realPathSync
- [ ] remove
- [ ] removeSync
- [ ] rename
- [ ] renameSync
- [ ] stat
- [ ] statSync
- [ ] symlink
- [ ] symlinkSync
- [ ] truncate
- [ ] truncateSync
- [ ] umask
- [ ] utime
- [ ] utimeSync
- [ ] watchFs
- [ ] writeFile
- [ ] writeFileSync
- [ ] writeTextFile
- [ ] writeTextFileSync

`shim-deno` package already implements Node.js compatible Deno APIs. We can borrow the code from there for the Node.js compatibility part https://github.com/denoland/node_shims/tree/main/packages/shim-deno

eryue0220:
Are `unlink` and `unlinkSync` ignored?

kt3k:
It looks like Deno doesn't have `unlink` and `unlinkSync`. Instead it has `remove` and `removeSync`. This list tries to mirror Deno FS APIs.

eryue0220:
Considering compatibility with Node, should `fs.unlink` and `fs.unlinkSync` also be provided?

kt3k:
Do you suggest `unlink` would be an alias to `remove`?

eryue0220:
Yes, except for compatibility, `link` and `unlink` is a pair, which is make more sense and more understandable than `remove`. WDYT?

[...]

PR #6338, opened 2025-01-08
feat(fs/unstable): add readDir

jbronder:
This PR adds the `readDir` API to `@std/fs`, which is a function intended to mirror `Deno.readDir`.

Towards #6255.


fs/unstable_types.ts
```diff
@@ -80,3 +80,18 @@ export interface FileInfo {
    * _Linux/Mac OS only._ */
   isSocket: boolean | null;
 }
+
+export interface DirEntry {
```
kt3k, line 84:
```suggestion
/**
 * Information about a directory entry returned from {@linkcode readDir}
 * and {@linkcode readDirSync}.
 */
export interface DirEntry {
```

kt3k (approved):
LGTM. Thank you!

PR #6339, opened 2025-01-10
chore: use native type stripping of Node.js in node testing

kt3k:
This PR starts using native type strip feature of Node.js to run tests in Node, and it fixes the CI on main.

This PR also does some clean ups.
- You can now directly execute `deno task test:node` without running `npm install` in `_tools/node_test_runner` dir beforehand.


satyarohith (approved):
LGTM

PR #6342, opened 2025-01-10
fix(assert): handle `__proto__` correctly in `assertObjectMatch`

### LINKED ISSUE #6341: `assertObjectMatch` is not able to test target object's own `__proto__`
**Describe the bug**

As the title says. `assertEquals` does work expectedly but `assertObjectMatch` does not.

**Steps to Reproduce**

1. Create a file named `test.ts` with the content being:
	
	```typescript
	import { assertObjectMatch, assertEquals } from "jsr:@std/assert@1.0.10"

	Deno.test("assertEquals", async t => {
		await t.step("should be able to test target object's own `__proto__`", async t => {
			const objectA = { ["__proto__"]: { polluted: true } }
			const objectB = { ["__proto__"]: { polluted: true } }
			assertEquals(objectA, objectB)
		})
	})

	Deno.test("assertObjectMatch", async t => {
		await t.step("should be able to test target object's own `__proto__`", async t => {
			const objectA = { ["__proto__"]: { polluted: true } }
			const objectB = { ["__proto__"]: { polluted: true } }
			assertObjectMatch(objectA, objectB)
		})
	})
	```
2. Execute the command `deno test --unstable-unsafe-proto test.ts`
3. See error is like this:
	
	```
	assertEquals ...
	  should be able to test target object's own `__proto__` ... ok (0ms)
	assertEquals ... ok (0ms)
	assertObjectMatch ...
	  should be able to test target object's own `__proto__` ... FAILED (1ms)
	assertObjectMatch ... FAILED (due to 1 failed step) (2ms)
	```
	
	And the diff is even unhelpful:
	
	```


	    [Diff] Actual / Expected


	    {}

	        ^
	```

**Expected behavior**

The test case successfully passes.

For normal users the infamous `__proto__` should be disabled by default, but as a library author I want to test our APIs are immune to the behavior, but currently I can't use `assignObjectMatch` for it due to this bug.

**Environment**

- OS: macOS Sequoia 15.0
- deno version: deno 2.1.4 (stable, release, aarch64-apple-darwin)
- std version: @std/assert@1.0.10

yuhr:
Fixes #6341.

### DIFF
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -137,7 +137,11 @@ function filter(a: loose, b: loose): loose {
           continue;
         }
 
-        defineProperty(filtered, key, filterObject(value as loose, subset as loose));
+        defineProperty(
+          filtered,
+          key,
+          filterObject(value as loose, subset as loose),
+        );
         continue;
       }
 
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -62,7 +62,7 @@ function defineProperty(target: object, key: PropertyKey, value: unknown) {
 }
 
 function filter(a: loose, b: loose): loose {
-  const seen = new WeakMap();
+  const seen = new WeakMap<object, object>();
   return filterObject(a, b);
 
   function filterObject(a: loose, b: loose): loose {
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -62,7 +62,7 @@ function defineProperty(target: object, key: PropertyKey, value: unknown) {
 }
 
 function filter(a: loose, b: loose): loose {
-  const seen = new WeakMap<object, object>();
+  const seen = new WeakMap<loose | unknown[], loose | unknown[]>();
   return filterObject(a, b);
 
   function filterObject(a: loose, b: loose): loose {
--- a/assert/object_match.ts
+++ b/assert/object_match.ts
@@ -52,15 +52,23 @@ function isObject(val: unknown): boolean {
   return typeof val === "object" && val !== null;
 }
 
+function defineProperty(target: object, key: PropertyKey, value: unknown) {
+  return Object.defineProperty(target, key, {
+    value,
+    configurable: true,
+    enumerable: true,
+    writable: true,
+  });
+}
+
 function filter(a: loose, b: loose): loose {
   const seen = new WeakMap();
   return filterObject(a, b);
 
   function filterObject(a: loose, b: loose): loose {
     // Prevent infinite loop with circular references with same filter
-    if ((seen.has(a)) && (seen.get(a) === b)) {
-      return a;
-    }
+    const memo = seen.get(a);
+    if (memo && (memo === b)) return a;
 
     try {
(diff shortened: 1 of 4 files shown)


deno.json
```diff
@@ -8,7 +8,7 @@
   },
   "importMap": "./import_map.json",
   "tasks": {
-    "test": "deno test --unstable-http --unstable-webgpu --doc --allow-all --parallel --coverage --trace-leaks --clean",
+    "test": "deno test --unstable-http --unstable-webgpu --unstable-unsafe-proto --doc --allow-all --parallel --coverage --trace-leaks --clean",
```
kt3k, line 11: I don't think we should change this for the entire test run.

Can you add another task which only run `deno test --unstabel-unsafe-proto assert/object_match_test.ts`?

assert/object_match.ts
```diff
@@ -52,15 +52,23 @@ function isObject(val: unknown): boolean {
...
   const seen = new WeakMap();
   return filterObject(a, b);
 
   function filterObject(a: loose, b: loose): loose {
     // Prevent infinite loop with circular references with same filter
-    if ((seen.has(a)) && (seen.get(a) === b)) {
-      return a;
-    }
+    const memo = seen.get(a);
+    if (memo && (memo === b)) return a;
```
kt3k, line 71: Is this change necessary? What happens if memo is the value `false`?

deno.json
yuhr: Is there any reason not to enable this option for the entire tests? Deno std library might be used everywhere out of control of Deno, such as browsers. If such runtimes are included in targets of the std, disabling unsafe proto is meaningless.

IMHO I even think this option should be enabled by default for `deno test` CLI. Without this option, users can't notice there might be potential bugs and vulnerabilities that could happen only when run in other runtimes. This would be quite hard to debug.

assert/object_match.ts
yuhr: Just performance thing. If you read the context of the changed lines, you'll realize that will never happen.

yuhr: See these types: [cfbff35](https://github.com/denoland/std/pull/6342/commits/cfbff35b1f911ed32d1720a754f81d57f4ff1ebc)

If you say this change should be done in another PR, I understand it and revert this part, but I'd say such a trivial and small change can be okay to come along with...

deno.json
kt3k: > Is there any reason not to enable this option for the entire tests? Deno std library might be used everywhere out of control of Deno, such as browsers. If such runtimes are included in targets of the std, disabling unsafe proto is meaningless.

Deno is currently the primary target of Deno std. The test cases should pass with the default state of Deno.

I think the compatibility with other runtimes should be checked by running test cases in those runtime, not simulating them in Deno.

> IMHO I even think this option should be enabled by default for deno test CLI. 

Open an issue in denoland/deno repository.

kt3k (approved):
LGTM. Thanks!

PR #6343, opened 2025-01-11
feat(fs/unstable): add chmod and chmodSync

jbronder:
This PR adds the `chmod` and `chmodSync` APIs to `@std/fs`, which are functions intended to mirror `Deno.chmod` and `Deno.chmodSync`, respectively.

Towards #6255.


kt3k:
We can't use Deno specific features in testing as these test are also run in Node.js (also we can't use `using` as it's not supported yet in Node).

One way to work around it is to use `node:fs` APIs in testing which is supported by both Node and Deno. Another way would be first porting those necessary APIs to `std/fs` and use them in testing

jbronder:
Ahh, I see.  I'll start updating the tests.  I appreciate the guidance.

kt3k (approved):
Thanks for updating the test. Now this looks perfect!

PR #6344, opened 2025-01-12
fix(cbor): incorrect decoding with subarrays

BlackAsLight:
This pull request fixes a simple bug in the event a subarray is passed into `decodeCbor` which would cause data to be incorrectly decoded and/or throw an error.


kt3k:
The fix looks good, but how about passing byteOffset and byteLength to DataView constructor instead?

BlackAsLight:
That is probably better. I didn't even realise that was an option.

kt3k (approved):
LGTM. Thanks!

PR #6350, opened 2025-01-18
feat(cli/unstable): support stderr on spinner

andrewthauer:
Related to #5495

Adds support to the `@std/cli/unstable_spinner` to write to `stderr` instead of the default `stdout`.

The current Spinner is unusable when piping to another command (e.g. `deno_cli | less` for paging). and will cause garbled data to the pipe/console. Using `stderr` is the standard approach for informational text since by default it is not passed along to the piped commands stdin.

NOTE: I personally would be happy to make this the default. I'm curious if folks would be open to this considering it's currently unstable.

An alternative could be to check if the terminal is interactive or not and change the default based on this.

### DIFF
--- a/cli/unstable_spinner.ts
+++ b/cli/unstable_spinner.ts
@@ -80,7 +80,7 @@ export interface SpinnerOptions {
    *
    * @default {Deno.stdout}
    */
-  stream?: typeof Deno.stderr | typeof Deno.stdout;
+  output?: typeof Deno.stderr | typeof Deno.stdout;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Looks reasonable addition to me.

Can you also add an example of passing `Deno.stderr` in Spinner class jsdoc?

cli/unstable_spinner.ts
```diff
@@ -75,9 +75,16 @@ export interface SpinnerOptions {
...
+   * The stream to write the spinner to.
+   *
+   * @default {Deno.stdout}
+   */
+  stream?: typeof Deno.stderr | typeof Deno.stdout;
 }
 
-/**
+/**nnnn
+ * that
```
kt3k, line 87:
```suggestion
/**
```

kt3k:
> I'm curious if you have an opinions or there is a standard name to use for the stream option?

I don't think there's standard name for it, but how about `output` as it doesn't accept arbitrary stream?

> Also any thoughts on making this the default vs stdout? From my perspective there aren't many or any downsides and only upsides to having this as the default. I've noticed other libraries already conform to the pattern of using stderr for informational things (e.g. dax, cliffy).

That sounds reasonable to me. Let's continue on that change after this one.

kt3k (approved):
LGTM again

PR #6351, opened 2025-01-19
Make debounce work with asynchronous function

Antman261:
## Motivation

The async package provides helper functions for asynchronous operations, yet the debounce function does not work with asynchronous functions. Since these functions are commonly the most valuable functions to debounce, It makes sense to design debounce to support asynchrony.

This also allows us to return a value from debounce.

## Solution

* Simplified the types for debuts: it's now able to infer the arguments and return type of the wrapped function.
* Every function call saves a resolver in an array: Every call to a debounced function will resolve the same result at the same time.

**Caveat:** I wrote this in the GitHub editor quickly before bed, it probably has a type error I missed somewhere -- will fix it tomorrow, but worth a general review in the meantime so I can include the feedback when I look at it tomorrow evening (Australian timezone).

### DIFF
--- a/async/debounce.ts
+++ b/async/debounce.ts
@@ -17,6 +17,8 @@ export interface DebouncedFunction<Fn extends (...args: any[]) => any> {
   readonly pending: boolean;
 }
 
+type Resolver<T> = (value: PromiseLike<T> | T) => void;
+
 /**
  * Creates a debounced function that delays the given `func`
  * by a given `wait` time in milliseconds. If the method is called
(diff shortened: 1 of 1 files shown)


Antman261:
Woke up this morning and decided to approach asynchronously debouncing in a separate function -- less of a breaking change. I would still like to improve the type definition of the original function, but it may still be too breaking for very little ROI? Happy to roll that one back if you like.

I'll write tests for debounceAsync tonight

kt3k:
> yet the debounce function does not work with asynchronous functions.

I don't understand this part well. The current `debounce` function allow debouncing async functions (You can assign `(...) => Promise<X>` to `(...) => void`)

```ts
debounce(() => fetch("https://example.com"), 1000);
```

kt3k:
Closing due to inactivity

PR #6352, opened 2025-01-20
feat(fs/unstable): add symlink and symlinkSync

jbronder:
This PR adds the `symlink` and `symlinkSync` APIs to `@std/fs`, which are functions intended to mirror `Deno.symlink` and `Deno.symlinkSync`, respectively.  The PR also adds the associated `SymlinkOptions` type.

Towards #6255.


kt3k (approved):
LGTM. Nice work! Thanks

PR #6355, opened 2025-01-22
fix(io): fix readAllSync for the case when read source is slow

### LINKED ISSUE #6346: `@std/io`: `readAllSync` only reads a single line from `Deno.stdin` when stdin is interactive
https://jsr.io/@std/io/doc/~/readAllSync

**Steps to Reproduce**

```ts
#!/usr/bin/env deno run

import { readAllSync } from "@std/io"

const content = readAllSync(Deno.stdin)

console.log()
console.log(content.length)
```

**Expected behavior**

When running the above script in a shell session, I'd expect to be able to do the following (typing 3 lines then pressing Ctrl-D (i.e. EOF) to finish):

```shell
$ ./repro.ts
one
two
three
^D

14
$
```

That is how all the other languages I've tried work.

**Describe the bug**

But `readAllSync` returns after I type the first line.

```shell
$ ./repro.ts
one

4
$
```

**Environment**

- OS: macOS 12.7.6
- deno version: 2.1.5
- std version: 0.225.0

kt3k:
This PR fixes `readAllSync` when the source stream outputs data slowly.

The bug was introduced in #4128 when we rewrite it without using `Buffer` class.

closes #6346

### DIFF
--- a/async/unstable_wait_for.ts
+++ b/async/unstable_wait_for.ts
@@ -1,4 +1,4 @@
-// Copyright 2018-2024 the Deno authors. All rights reserved. MIT license.
+// Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
 
 import { deadline } from "./deadline.ts";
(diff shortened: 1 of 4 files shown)


dsherret (approved):
LGTM

PR #6357, opened 2025-01-22
docs(expect): fix typo on objectContaining example

rogerpoliver:
Hi, I was reading the documentation for `expect.objectContaining` and noticed a mistake in the example.

The first expectation is incorrect because the expected value for the key bar does not match the actual value. I've updated the example to reflect the correct behavior.


kt3k (approved):
Good catch! Thanks

PR #6360, opened 2025-01-24
fix(fmt): make printf working with colors.

keelii:
`printf` is not working well with `colors` package.

```bash
➜  ent-std git:(master) ✗ cat test.ts  
import { printf } from "jsr:@std/fmt/printf";
import {underline} from "jsr:@std/fmt/colors";

printf("a: %7s|\n", underline("foo"))
printf("b: %7s|\n", underline("foo bar"))
➜  ent-std git:(master) ✗ deno run -A test.ts
a: foo|
b: foo bar|
```

Add `stripAnsiCode` in `Printf.fmtString` maybe solve this problem.

```bash
➜  std git:(feat/make_printf_working_with_colors) ✗ cat test.ts  
import {printf} from "./fmt/printf.ts";
import {underline} from "./fmt/colors.ts";

printf("a: %7s|\n", underline("foo"))
printf("b: %7s|\n", underline("foo bar"))
➜  std git:(feat/make_printf_working_with_colors) ✗ deno run -A test.ts
a:     foo|
b: foo bar|
```

### DIFF
--- a/fmt/printf.ts
+++ b/fmt/printf.ts
@@ -1,5 +1,7 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
+import { stripAnsiCode } from "./colors.ts";
+
 /**
  * {@linkcode sprintf} and {@linkcode printf} for printing formatted strings to
  * stdout.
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Looks reasonable to me

PR #6362, opened 2025-01-24
refactor(semver): clean up `parseRange`, add missing tests

timreichen:


semver/greater_than_range_test.ts
```diff
@@ -167,3 +168,41 @@ Deno.test("greaterThanRange() checks if the semver is greater than the range", a
...
+  };
+  const range = [[{
+    operator: "!=" as unknown as Operator,
+    major: 1,
+    minor: 0,
+    patch: 0,
+    prerelease: [],
+    build: [],
+  }]];
+  assertEquals(greaterThanRange(version, range), true);
```
kt3k, line 207: I wonder if this result is correct

I also just realized `!=` is not a valid operator in `npm:semver` (`semver.gtr("1.0.0", "!=1.0.0")` throws with invalid operator error, and docs doesn't mention `!=` https://github.com/npm/node-semver?tab=readme-ov-file#ranges

I'd suggest we should skip this particular test case for now.

semver/less_than_range_test.ts
```diff
@@ -169,3 +170,22 @@ Deno.test("lessThanRange() checks if the SemVer is less than the range", async (
...
+  };
+  const range = [[{
+    operator: "!=" as unknown as Operator,
+    major: 1,
+    minor: 0,
+    patch: 0,
+    prerelease: [],
+    build: [],
+  }]];
+  assertEquals(lessThanRange(version, range), true);
```
kt3k, line 190: This result feels wrong to me. I'd suggest skipping this test case for now

semver/parse_range_test.ts
```diff
@@ -666,3 +666,24 @@ Deno.test("parseRange() throws on invalid range", () => {
...
   );
 });
+
+Deno.test("parseRange() handles wildcards", () => {
+  assertEquals(parseRange("<1.*"), [
+    [{ operator: "<", major: 1, minor: 0, patch: 0 }],
+  ]);
+  assertEquals(parseRange("<1.*.0"), [
+    [{ operator: "<", major: 1, minor: NaN, patch: 0 }],
+  ]);
```
kt3k, line 676: The difference of these 2 results seem strange to me. Is this intentionally in this way?

timreichen: This seems indeed wrong. Imo `<1.*.0` should be an invalid range and throw.

semver/less_than_range_test.ts
timreichen: I think we should not skip test cases if they yield unexpected results but fix them instead. What would the proper result be?

semver/parse_range_test.ts
timreichen: I took a closer look at this, `satisfies()` tests allow this: 
https://github.com/denoland/std/blob/3b75ee7c1925388e01dd69540d51d9387e26dcb0/semver/satisfies_test.ts#L75C5-L75C25

semver/less_than_range_test.ts
timreichen: Would it make sense to create a PR to remove the `!=` straight away as a bug fix before landing this PR? This probably solves lots of headaches with intermediate patches.

kt3k: That will be a breaking change and we currently don't have a plan for it.

semver/greater_than_range_test.ts
```diff
@@ -167,3 +168,43 @@ Deno.test("greaterThanRange() checks if the semver is greater than the range", a
...
+    major: 1,
+    minor: 0,
+    patch: 0,
+    prerelease: [],
+    build: [],
+  }]];
+  assertEquals(greaterThanRange(version, range), false);
+});
+
+// BUG `!=` operator type does not exist in semver and should be removed.
```
kt3k, line 191: I don't consider `!=` itself as bug. It's already shipped as stable API. I think we shouldn't remove it at least during `@std/semver@1.x` versions.

If we try to remove it, we first need to discuss it with more people (included both members from core team and community) and need to make consensus to make that breaking change at `@std/semver@2.0.0`

semver/parse_range_test.ts
kt3k: I think the range expression `<1.*.0` is fine, but shouldn't the parse result be the same as `<1.x` as they seem expressing the same range?

kt3k (approved):
LGTM.

I changed the commit type to `refactor` as this PR modifies `parseRange` implementation.

semver/less_than_range_test.ts
timreichen: I think this bug could be fixed by modifying the `!=` case to `return cmp === 0;` instead of `return false;` in `greaterThanComparator()` and `lessThanComparator()`. Then it will return the expected `false`.

semver/compare_test.ts
```diff
@@ -54,3 +54,50 @@ Deno.test({
...
+          patch: 0,
+        }),
+      Error,
+      "Cannot compare against non-numbers",
+    );
+  },
+});
+
+Deno.test({
+  name: "compare() handles undefined in prerelease",
```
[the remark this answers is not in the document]
timreichen: If this is undefined behavior we probably should remove `if (a === undefined && b === undefined) return 0;` from `compareIdentifier()` as this is specifically handled there.

PR #6363, opened 2025-01-24
fix(csv): add `columns` options type for array items

timreichen:
Adds number type only for `columns` option when items are arrays.

Related: https://github.com/denoland/std/issues/2660


csv/unstable_stringify.ts
```diff
@@ -280,55 +308,15 @@ export type StringifyOptions = {
...
- * ```
- *
  * @param data The source data to stringify. It's an array of items which are
  * plain objects or arrays.
  * @param options Options for the stringification.
  * @returns A CSV string.
  */
+export function stringify(
+  data: readonly ObjectItem[],
+  options?: ObjectStringifyOptions,
```
kt3k, line 318: This only expose ObjectStringifyOptions to the users. (ArrayStringifyOptions is only used in implementation signature, and not visible from the user)

kt3k: I think we also need:

```
export function stringify(
  data: readonly ArrayItem[],
  options?: ArrayStringifyOptions,
): string;
```

BTW If we overload a function, the doc linter requires each overload (except implementation signature) having its own jsdoc.

timreichen: done

kt3k:
Sorry for the delay in review, but I don't think we can merge this change as this doesn't pass many of existing test cases in `stringify_test.ts`.

We plan to replace `stringify` with `unstable-stringify` in the future with some stabilization process. This makes unstable-stringify incompatible with stable stringify, and that makes future stabilization impossible.

Issue #6124, opened 2024-10-17
Remove logging framework


vwkd:
I'd suggest to remove the log module.

It isn't cross platform compatible, has the lowest JSR score of all `@std` packages with 88%, and is still unstable. The API has not evolved since originally porting it from Go, the API doesn't really fit JavaScript with magical global state that's influenced by the `setup` function, which the user must take care to always call first before any implementing library.

It's the only framework in the standard library. There are different ways to do logging and not one clear solution. There are various third-party logging frameworks already, often better maintained and cross platform. Similar arguments why a HTTP middleware framework hasn't been added yet #1295.

iuioiua:
I agree. I very much dislike `@std/log`. Discussions with Yoshiya (when I maintained this project) aligned with replacing the entire package with something better.

halvardssm:
I partially agree, I don’t mind the logger changing, but I am finding it very useful to have a standard logger library that can log to the console, a file and whatever pluggable custom logger I would find or make.

BlackAsLight:
I am in favour of it being changed instead of removed. I'm not 100% familiar with how it works, but things I'd like to see in a logging framework is:
- The ability for libraries to implement it in their code that doesn't add additional computational strait if the main module isn't using it.
  - This could maybe be archived though accepting simple types in `log.debug()` or any type that implements a custom symbol like `Symbol.Log`, where calling the symbol on the object will have it handle how to stringify the object.
- For the main module to decide completely how the output of the logger is displayed instead of opting for JSON.
- A cron like expression that triggers to allow rotation of where the logs are going. For example if it's going to a file, lets one switch to a new file after x hours.
- Many of the settings for the main module be configured via environment variables. This can remove people's disdain for the need to call `log.setup()` Like:
  - Default level being logged.
  - Library specific level being logged.
  - Location the logs are logged: "stdout", "stderr", a directory/file name, "null", or let runtime code decide.

chyzwar:
What is the best alternative? Last time I tried to use pino it was not working.
As user, I would want to know what is recommended logging framework in deno.

kt3k:
LogTape seems relatively popular to me https://github.com/dahlia/logtape
It officially supports Deno, and is available from JSR https://jsr.io/@logtape/logtape

chyzwar:
Logtape is not even a year old and only have one maintainer. It only has 510 stars. For me, this is the definition of niche and very early stage. Is there is a way see download stats in JSR?

chyzwar:
[...]

PR #6364, opened 2025-01-24
docs(log): mention potential deprecation of @std/log

dandv:
The description on JSR could be updated as well, to mention considering Open Telemetry per #6124.

image


kt3k (approved):
LGTM

This direction is not completely set in stone, but it's worth mentioning this in README to see the community's reaction.

PR #6366, opened 2025-01-24
feat(fs/unstable): add realPath and realPathSync

jbronder:
This PR adds the `realPath`and `realPathSync` APIs to `@std/fs`, which are functions intended to mirror `Deno.realPath` and `Deno.realPathSync`, respectively.

Towards #6255.


kt3k (approved):
LGTM. Thanks!

PR #6372, opened 2025-01-29
refactor(bytes,crypto,encoding,http,io,msgpack,streams,tar): make Uint8Array usages compatible with TS 5.7

kt3k:
related https://github.com/denoland/deno/pull/27857


dsherret (approved):
LGTM! Great idea with the Uint8Array_ type.

PR #6373, opened 2025-01-31
feat(fs/unstable): add readLink and readLinkSync

jbronder:
This PR adds the `readLink` and `readLinkSync` APIs to `@std/fs`, which are functions intended to mirror `Deno.readLink` and `Deno.readLinkSync`, respectively.

Towards #6255.


kt3k (approved):
LGTM. Nice!

PR #6378, opened 2025-02-03
feat(cli/unstable): introduce `new ProgressBar()` & `new ProgressBarStream()`

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6374

This pull request introduces two classes called `ProgressBar` and `ProgressBarStream`. They write and update the progress to the writable stream passed to them.

## Example
```ts
import { ProgressBarStream } from "@std/cli/unstable_progress_bar_stream";

const readable = ReadableStream
  .from<Uint8Array>(async function* (): AsyncGenerator<Uint8Array> {
    for (let i = 0; i < 100; ++i) {
      yield new Uint8Array(1000).fill(97);
      await new Promise((a) => setTimeout(a, (Math.random() * 200) | 0));
    }
  }())
  .pipeThrough(
    new ProgressBarStream(Deno.stdout.writable, {
      max: 1000 * 100,
    }),
  );

// deno-lint-ignore no-empty
for await (const _ of readable) {}
```
### stdout: 
```
[00:04] [###################-------------------------------] [38.09/97.66 KiB]
```


cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * The length that the progress bar should be, in characters.
+   * @default {50}
+   */
+  barLength?: number;
+  /**
+   * The character to fill the progress bar up with as it makes progress.
+   * @default {'#'}
+   */
+  fullChar?: string;
```
kt3k, line 26: `fillChar`?

kt3k:
I agree that showing the progress of byte number count is a common task. But can we also support the progress of abstract counter, like number of packages, number of files, etc.

BlackAsLight:
> I agree that showing the progress of byte number count is a common task. But can we also support the progress of abstract counter, like number of packages, number of files, etc.

Should the amount just offer to display no unit for that instance or ask for custom unit?

kt3k:
> Should the amount just offer to display no unit for that instance or ask for custom unit?

I think It'll probably be confusing to just omit the unit. If `10.0/50.0` is displayed, the user don't know if it's 50.0K or 50.0M.

How about having `bytes: boolean` option, for example? When it's true, it does the current behavior. If it's false, then it shows numbers like `30`, `20K`, `10M`, etc without `B` suffix.

BlackAsLight:
That could work. I was more thinking of offering a custom unit option and if set then the rate calculation would be 1 so they could see something like `[4/500,000 files]`

timreichen:
Maybe we should add a `formatter` option instead of many different styling options.
This is more flexible and also avoids the problem of localization.

Something according to:
```ts
import { format as formatDuration } from "@std/fmt/duration.ts"
import { format as formatBytes } from "@std/fmt/bytes.ts"
new ProgressBarStream(Deno.stdout.writable, { formatter: ({ time, progressBar, value, max }) => `[${formatDuration(time)}] ${progressBar} [${formatBytes(value)}/${formatBytes(max)}]` })
```

cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,189 @@
...
+/**
+ * The options that are provided to a {@link createProgressBar} or
+ * {@link ProgressBarStream}.
+ */
+export interface ProgressBarOptions {
+  /**
+   * The offset size of the input if progress is resuming part way through.
+   * @default {0}
+   */
+  currentSize?: number;
```
timreichen, line 12: Maybe let's call it `value` like in [progress  HTMLElement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress)?

cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * The total size expected to receive.
+   */
+  totalSize: number;
```
timreichen, line 16: Maybe let's call it `max` like in [progress  HTMLElement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress)?

cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * The character the progress bar starts out with.
+   * @default {'-'}
+   */
+  emptyChar?: string;
+  /**
+   * Whether the progress bar should be removed after completion.
+   * @default {false}
+   */
+  clearBar?: boolean;
```
timreichen, line 36: Maybe let's call it just `clear` as in https://github.com/denoland/std/blob/3b75ee7c1925388e01dd69540d51d9387e26dcb0/cli/prompt_secret.ts#L24 ?

cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,189 @@
...
+   * displayed.
+   * @default {true}
+   */
+  includeAmount?: boolean;
+  /**
+   * Whether or not the {@link WritableStream} argument should be closed upon
+   * completion.
+   * @default {true}
+   */
+  preventClose?: boolean;
```
timreichen, line 58: I think options should be formulated positively rather than negatively, something like `close`, `closeStream` or `closeStreamWhenCompleted`

BlackAsLight:
> Maybe we should add a `formatter` option instead of many different styling options. This is more flexible and also avoids the problem of localization.
> 
> Something according to:
> 
> ```ts
> import { format as formatDuration } from "@std/fmt/duration.ts"
> import { format as formatBytes } from "@std/fmt/bytes.ts"
> new ProgressBarStream(Deno.stdout.writable, { formatter: ({ time, progressBar, value, max }) => `[${formatDuration(time)}] ${progressBar} [${formatBytes(value)}/${formatBytes(max)}]` })
> ```

I added this and a bit more to it, making it optional as I doubt everyone who wants a progress bar wants to go through the effort of styling it.

BlackAsLight:
The only error left seems to be one unrelated to this pull request

cli/unstable_progress_bar.ts
```diff
@@ -0,0 +1,216 @@
...
+  let lastTime = startTime;
+  let lastValue = options.value!;
+
+  return addProgress;
+  /**
+   * @param addSize The amount of bytes of progressed made since last call.
+   * @param done Whether or not you're done reporting progress.
+   */
+  function addProgress(addSize: number, done?: false): void;
+  function addProgress(addSize: number, done: true): Promise<void>;
```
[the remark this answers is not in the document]
BlackAsLight: I can do this, although I'm not a great fan of it being a class

kt3k (approved):
Thanks for updating to class-based design. I believe this is more reasonable as it feels aligned to `Spinner` class.

LGTM.

timreichen:
Just a tiny nit pick: maybe we can make use of `@std/bytes/format` for formatting units?

BlackAsLight:
> Just a tiny nit pick: maybe we can make use of `@std/bytes/format` for formatting units?

I believe you meant `@std/fmt/bytes`. Looking at that module it doesn't look like one would be able to get the same default style here so I'm opposed it on that front, but it can be used in a custom format if the user chooses to provide one

timreichen:
> > Just a tiny nit pick: maybe we can make use of `@std/bytes/format` for formatting units?
> 
> I believe you meant `@std/fmt/bytes`. Looking at that module it doesn't look like one would be able to get the same default style here so I'm opposed it on that front, but it can be used in a custom format if the user chooses to provide one

Ok, never mind👍

PR #6379, opened 2025-02-04
feat(fs/unstable): add rename

jbronder:
This PR adds the `rename` API to `@std/fs`, which is intended to mirror `Deno.rename` function.

Towards #6255.


kt3k (approved):
LGTM. Thanks @jbronder !

PR #6381, opened 2025-02-05
feat(fs/unstable): add readDirSync

jbronder:
This PR adds the `readDirSync` API to `@std/fs`, which is intended to mirror the `Deno.readDirSync` function.  The PR also adds additional `@param` and `@returns` documentation to the `readDir` API.

Towards #6255 .


kt3k:
CI is failing with Deno v1.x because `IteratorObject` type is not available in Deno 1.x.

One workaround is to use `Iterable` instead, but that prevents us using iterator helper method on it.

Does anyone have an idea to best handle this situation?

jbronder:
I've only tried testing it on Deno v1.46.0, the function signatures are indeed different for `readDirSync`:

```ts
// In the 1.x Deno namespace:
function readDirSync(path: string | URL): Iterable<DirEntry>

// In the 2.x Deno namespace:
function readDirSync(path: string | URL): IteratorObject<DirEntry>
```

I suppose there's no real benefit to having `readDirSync` use a return type of `Iterable<DirEntry>` for the `@std/fs` package for both 1.x and 2.x other than to satisfy the CI, unless that is indeed the goal.

After modifying the return type to `Iterable<DirType>`, the iterable returned by Deno 1.x `readDirSync` does allow access to Iterator Helper methods after testing it out in a small script in the project root:
```ts
import { readDirSync } from "./fs/unstable_read_dir.ts";

const dir = readDirSync("fs/testdata");
const partialList = dir.take(5).toArray();
console.log("partialList:", partialList);
```

It probably doesn't help from a TypeScript perspective since the function signature does not semantically convey what the function does because of the return type differences.

I can leave this on the back burner for now and make progress on other APIs and return to it.

fs/unstable_read_dir.ts
```diff
@@ -38,3 +41,40 @@ export async function* readDir(path: string | URL): AsyncIterable<DirEntry> {
     }
   }
 }
+
```
petamoriken, line 44: How about defining the type of `IteratorObject` here?

```suggestion

interface IteratorObject<T> extends Iterable<T> {};

```

In Deno 1 (TypeScript 5.3 or lower), it should be treated as just a new type definition, and in Deno 2 (TypeScript 5.4 or higher), where `IteratorObject` type exists, it should be treated as a merge of interface types and nothing should happen.

sigmaSd:
Is there on update on this ?

jbronder:
I'll take another look at this soon.  While not necessarily related to the `readDirSync`, I'm currently working on a forthcoming PR for `open`, `openSync, and `FsFile` at the moment.

jbronder:
I appreciate your patience, @kt3k.  This PR is ready for review at your convenience.

kt3k (approved):
LGTM. Thanks!

PR #6382, opened 2025-02-05
feat(collections): add index arg to methods

### LINKED ISSUE #6288: [collections] Pass index to partition callback

I tried to use `partition` and expected to be able to access the element index like normal array methods, but it isn't passed to the callback.

Pass the element index to the `partition` callback. If the maintainers think this is a good idea, there are other methods in `collections` that could be passed the index as well. Maybe we should even consider passing in the original array as the third argument to match the signature of most built-in array method callbacks.

Call `.filter` twice to partition the array instead.

canac:
Adds an `index` arg to collections' predicate/selector/transformer/discriminator method. I added it only to collection helpers that operate on iterables because collection helpers that operate on records don't seem as likely to need access to the key index. Please let me know if I missed any.

Closes #6288

### DIFF
--- a/collections/distinct_by.ts
+++ b/collections/distinct_by.ts
@@ -29,12 +29,13 @@
  */
 export function distinctBy<T, D>(
   array: Iterable<T>,
-  discriminator: (el: T) => D,
+  discriminator: (el: T, index: number) => D,
 ): T[] {
   const keys = new Set<D>();
   const result: T[] = [];
+  let index = 0;
   for (const element of array) {
-    const key = discriminator(element);
+    const key = discriminator(element, index++);
     if (!keys.has(key)) {
       keys.add(key);
       result.push(element);
--- a/collections/distinct_by_test.ts
+++ b/collections/distinct_by_test.ts
@@ -5,7 +5,7 @@ import { distinctBy } from "./distinct_by.ts";
 
 function distinctByTest<I>(
   array: Array<I>,
-  selector: (element: I) => unknown,
+  selector: (element: I, index: number) => unknown,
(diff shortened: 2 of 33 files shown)


kt3k:
I'm generally in favor of the idea. It looks useful in several places, and also this is aligned to the design of builtin map methods (such as `map`, `filter`, `forEach`, etc)

What do people think?

---
Note: Currently we have [policy](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100) of accepting any new feature in `unstable_` prefixed path. That requires all these APIs duplicated in `unstable_foo.ts`, but that feels too unwieldy in this case. I'll discuss with core team about how to handle this particular case.

timreichen:
> What do people think?

I like it.

iuioiua:
I'm in favour.

iuioiua:
@kt3k should we add this in?

canac:
I'm more than happy to rebase this if we want to move forward with it. Just let me know!

iuioiua:
I suggest rebasing anyway 🙂

iuioiua:
@kt3k, it seems like this is a good idea to add.

tomas-zijdemans:
I support adding this!

kt3k:
(Unrelated side note: I initially thought it might be better to change [our new feature policy](https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100) (all feature should be unstable first) to land this PR directly, but somehow failed to invent a new policy. That delayed the review process/decision...)

PR #6390, opened 2025-02-09
refactor(front-matter): call `extract()` functions in `any.ts`

timreichen:
This PR replaces `parser` calls with `extract()` calls.
It also adds a missing type in `front_matter/toml.ts`


kt3k (approved):
LGTM

PR #6391, opened 2025-02-09
feat(fs/unstable): add makeTempDir and makeTempDirSync

jbronder:
This PR adds the `makeTempDir` and `makeTempDirSync` APIs to the `@std/fs` package which intends to mirror the `Deno.makeTempDir` and `Deno.makeTempDirSync` functions. The PR also adds the `MakeTempOptions` type and helper functions `getNodeOs` and `getNodePath` in the `_utils.ts` file in preparation for use in future implementations of `@std/fs` APIs: `makeTempFile` and `makeTempFileSync`.

To achieve cross-runtime compatibility with Node.js and satisfy the Deno FS API requirement, the `makeTempDir` and `makeTempDirSync` APIs with `MakeTempOptions` extends functionality for creating temporary directories in Node.js. By adding an optional interface to augment the naming of temporary directory paths, these APIs exist as an alternative to Node-specific APIs that pass a prefix path string argument to `mkdtemp` or `mkdtempSync` to create temporary directories.

Towards #6255.

### DIFF
--- a/_tools/node_test_runner/run_test.mjs
+++ b/_tools/node_test_runner/run_test.mjs
@@ -50,6 +50,7 @@ import "../../collections/unzip_test.ts";
 import "../../collections/without_all_test.ts";
 import "../../collections/zip_test.ts";
 import "../../fs/unstable_link_test.ts";
+import "../../fs/unstable_make_temp_dir_test.ts";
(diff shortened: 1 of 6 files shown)


kt3k (approved):
LGTM. Nice work!

PR #6393, opened 2025-02-09
perf(front-matter): remove regexp `m` flags, reduce unnecessary string operations

timreichen:
This PR restructures regexps by removing  `m` flags and adding capturing groups for `data` and `body`. This makes match post processing in `extractAndParse()` and pre processing in `recognize()` obsolete.


kt3k (approved):
Looks reasonable perf improvements to me. Thanks!

PR #6394, opened 2025-02-11
feat(fs/unstable): add readFile and readFileSync

jbronder:
This PR adds the `readFile` and `readFileSync` APIs to the `@std/fs` package which intends to mirror `Deno.readFile` and `Deno.readFileSync` functions. The PR also adds the `ReadFileOptions` type in preparation for adding the
`readTextFile` API to the `@std/fs` package.

A question on the documentation of `Deno.readFile` and `Deno.readFileSync`: Is the documentation slightly out of date? Specifically, this line:

> "Reading a directory returns an empty data array."

When attempting to do this on my local machine (MacOS), invoking functions `Deno.readFile` and `Deno.readFileSync` rejected/threw with an `Error` and an error code of `EISDIR` (os errno 21). This is true on Deno v1.46.0, v2.x, and
the latest Node.js AFAIK. For the purposes of the PR, I've assumed this is the intended behavior and not returning an empty `Uint8Array` object (`Uint8Array(0)[]`). The `Error` behavior seems to be consistent with `Deno.readTextFile` and `Deno.readTextFileSync`. Apologies if the assumption is too forward.

I can always add the behavior that the documentation states if needed with accompanying unit tests, but I have currently removed the aforementioned "empty array line" from the documentation in this PR. I also did not test for this behavior in the unit tests since all current runtimes (at least on MacOS) currently throw/reject with an `EISDIR` `errno 21` `Error`.

Towards #6255.

### DIFF
--- a/_tools/node_test_runner/run_test.mjs
+++ b/_tools/node_test_runner/run_test.mjs
@@ -51,6 +51,7 @@ import "../../collections/without_all_test.ts";
 import "../../collections/zip_test.ts";
 import "../../fs/unstable_link_test.ts";
 import "../../fs/unstable_read_dir_test.ts";
+import "../../fs/unstable_read_file_test.ts";
 import "../../fs/unstable_read_link_test.ts";
 import "../../fs/unstable_real_path_test.ts";
 import "../../fs/unstable_rename_test.ts";
--- a/fs/deno.json
+++ b/fs/deno.json
@@ -17,6 +17,7 @@
     "./unstable-link": "./unstable_link.ts",
     "./unstable-lstat": "./unstable_lstat.ts",
     "./unstable-read-dir": "./unstable_read_dir.ts",
+    "./unstable-read-file": "./unstable_read_file.ts",
     "./unstable-read-link": "./unstable_read_link.ts",
     "./unstable-real-path": "./unstable_real_path.ts",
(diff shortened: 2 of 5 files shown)


kt3k:
> A question on the documentation of `Deno.readFile` and `Deno.readFileSync`: Is the documentation slightly out of date? Specifically, this line:
> 
> > "Reading a directory returns an empty data array."

Looks like the behavior was changed in a really old version (1.1.0):

```
$ deno1.0.5
Deno 1.0.5
exit using ctrl+d or close()
> Deno.readFileSync(".")
Uint8Array(0) []
> 
$ deno1.1.0
Deno 1.1.0
exit using ctrl+d or close()
> Deno.readFileSync(".")
Uncaught Error: Is a directory (os error 21)
    at unwrapResponse ($deno$/ops/dispatch_minimal.ts:63:11)
    at Object.sendSyncMinimal ($deno$/ops/dispatch_minimal.ts:118:10)
```

Because this behavior exists for long time (more than 4 years), I think the assumption of this PR is reasonable.

Thanks for pointing this out!

kt3k (approved):
LGTM. Nice!

jbronder:
Sure thing! Thanks for writing the follow-up PR!

PR #6395, opened 2025-02-12
chore: add more coverage tasks to `deno.json`

BlackAsLight:
Nobody really asked for this type of change, but I have found it quite useful so hope it gets merged.

Essentially this change adds three new tasks to `deno.json`.
- `deno task cov:mac`
- `deno task cov:lin`
- `deno task cov:win`

Calling the one for your respective system will go through the process of generating the coverage reports and then open it with whatever your default application is for html files. Most likely your browser.

This pull request also makes the change of removing doc examples from being included in the coverage seen as, from my point of view, the doc examples aren't meant to be proving the validity of the code.


deno.json
```diff
@@ -24,10 +24,13 @@
     "typos": "typos -c ./.github/workflows/typos.toml",
     "build:crypto": "deno task --cwd crypto/_wasm wasmbuild",
     "wasmbuild": "deno run -A jsr:@deno/wasmbuild@0.17.1 --js-ext mjs --sync",
-    "cov": "deno coverage --ignore=\"**/*.generated.mjs,**/_test_utils.ts,.github\"",
+    "cov": "deno task test --coverage --clean --reporter=dot && deno coverage --ignore=\"**/*.generated.mjs,**/_test_utils.ts,.github\"",
```
kt3k, line 27: This seems causing the entire test cases run multiple times in CI. Can you also modify ci.yml to fix that?

kt3k (approved):
This also reduces unnecessary coverage output in publish CI. Nice!

LGTM

PR #6396, opened 2025-02-14
feat(fs/unstable): add renameSync

jbronder:
This PR adds the `renameSync` API to `@std/fs` which is intended to mirror the `Deno.renameSync` function.

Towards #6255.


kt3k (approved):
LGTM. Thanks!

PR #6397, opened 2025-02-14
chore: ignore jsonc/testdata from typo check

kt3k:
CI started failing at typo in `jsonc/testdata/test262/JSON/parse/15.12.1.1-g2-3.js`. https://github.com/denoland/std/actions/runs/13319999611/job/37206312707

The file is test fixture, and we can skip checking typo in there


satyarohith (approved):
👍

PR #6402, opened 2025-02-14
fix(cli): update interval in ProgressBar

BlackAsLight:
This pull request fixes the simple issue of the interval having the wrong delay of 200ms instead of 1000ms.


kt3k (approved):
LGTM

PR #6404, opened 2025-02-14
test(cli): fix flaky spinner test

timreichen:
cli `Spinner` test is flaky when testing the `interval` property. This is due to a too small time padding in `delay()`.


kt3k (approved):
LGTM

I also removed unnecessary stubs of `Deno.stdin.setRaw`, and introduced `FakeTime` in `Spinner constructor accepts interval` test case. (Without FakeTime, the case was still flaky)

PR #6406, opened 2025-02-15
BREAKING(cli/unstable): rename `ProgressBar` `end()` method to `stop()`

timreichen:
This PR renames `ProgressBar` `end()` method to `stop()` to align with [Spinner](https://github.com/denoland/std/blob/2829d6d8ec9a9070cf1cee06622e78e163b12444/cli/unstable_spinner.ts#L262).


kt3k:
@BlackAsLight What do you think?

kt3k (approved):
LGTM

BlackAsLight:
I think you forgot to merge this pull request?

kt3k:
I'd prefer to keep this and other breaking changes to ProgressBar open for a while to see if there's further feedbacks, and also to not break the usages for every release (ideally I like to merge all breaking changes to ProgressBar at once to minimize the impact of breakage)

PR #6407, opened 2025-02-15
refactor(cli/unstable): make `ProgressBar` `options` optional

timreichen:
This PR  make `ProgressBar` `options` an optional argument. This is done by setting `options.max` to `1` by default. This default value is taken from [HTMLProgressElement](https://developer.mozilla.org/de/docs/Web/HTML/Element/progress#attribute).

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -55,7 +55,7 @@ export interface ProgressBarOptions {
   /**
    * The total size expected to receive.
    */
-  max: number;
+  max?: number;
   /**
    * The length that the progress bar should be, in characters.
    * @default {50}
@@ -151,11 +151,11 @@ export class ProgressBar {
    */
   constructor(
     writable: WritableStream<Uint8Array>,
-    options: ProgressBarOptions,
+    options: ProgressBarOptions = {},
   ) {
     this.#options = {
       value: options.value ?? 0,
-      max: options.max,
+      max: options.max ?? 1,
       barLength: options.barLength ?? 50,
       fillChar: options.fillChar ?? "#",
       emptyChar: options.emptyChar ?? "-",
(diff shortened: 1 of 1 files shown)


BlackAsLight:
I'm not a fan of this change as, I imagine you're always going to want to set a `max` value, and forgetting to set one where the `value` goes over `max` will throw an error

timreichen:
> I'm not a fan of this change as, I imagine you're always going to want to set a `max` value, and forgetting to set one where the `value` goes over `max` will throw an error

I don't think `value > max` should throw. This might lead to unintended bugs with floating-point errors.
So we might want to change that which would pave the way for a default `max` (as in [HTMLProgressElement](https://developer.mozilla.org/de/docs/Web/HTML/Element/progress)).

WDYT?

BlackAsLight:
> > I'm not a fan of this change as, I imagine you're always going to want to set a `max` value, and forgetting to set one where the `value` goes over `max` will throw an error
> 
> 
> 
> I don't think `value > max` should throw. This might lead to unintended bugs with floating-point errors.
> 
> So we might want to change that which would pave the way for a default `max` (as in [HTMLProgressElement](https://developer.mozilla.org/de/docs/Web/HTML/Element/progress)).
> 
> 
> 
> WDYT?

It will throw when the maths has `"-".repeat(-1)` in the print function. As it doesn't seem to like negative numbers. It could be wrapped in a Math.max to stop this with the end result be that the progress bar just grows past its bar length.

timreichen:
> > > I'm not a fan of this change as, I imagine you're always going to want to set a `max` value, and forgetting to set one where the `value` goes over `max` will throw an error
> > 
> > 
> > I don't think `value > max` should throw. This might lead to unintended bugs with floating-point errors.
> > So we might want to change that which would pave the way for a default `max` (as in [HTMLProgressElement](https://developer.mozilla.org/de/docs/Web/HTML/Element/progress)).
> > WDYT?
> 
> It will throw when the maths has `"-".repeat(-1)` in the print function. As it doesn't seem to like negative numbers. It could be wrapped in a Math.max to stop this with the end result be that the progress bar just grows past its bar length.

I'll have a look at that. clamping these values seems like a good idea anyway.

kt3k:
Because this module formats the value as byte numbers by default, setting max=1 as default feels strange to me.

> I'm not a fan of this change as, I imagine you're always going to want to set a max value

If we consider `max` as required argument, then we probably should set `max` as positional argument instead of option

timreichen:
> Because this module formats the value as byte numbers by default, setting max=1 as default feels strange to me.
> 
> > I'm not a fan of this change as, I imagine you're always going to want to set a max value
> 
> If we consider `max` as required argument, then we probably should set `max` as positional argument instead of option

I was thinking about this a bit. `max = 1` just seems right for a progress indication as a default.
This causes a problem now because the current implementation is geared towards data/downloads which is not the only progress indication type.
So I was wondering:
Should we split up the `ProgressBar` and the `ProgressBarStream` implementation, making the `ProgressBar` sync like `Spinner` and remove the unit implementation and indicate percentages instead and let `ProgressBarStream` work like it does now for a data driven indication?

Another option would be to remove the custom `fmt` property from `ProgressBar` and instead introduced a `style` property that determine the format. Something like:
```ts
type ProgressBarType = "percentage" | "download"
new ProgressBar({ type: "percentage" }) // [###-------] [30%]
new ProgressBar({ type: "download" }) // [mm:ss] [###-------] [0.40/97.66 KiB]
```
This would be less customizable but simplify the api. I am aware that I proposed the format property instead of style options.

timreichen:
stale

PR #6408, opened 2025-02-15
feat(cli/unstable): add `ProgressBar` `start()` method

timreichen:
This PR adds `start()` method to `ProgressBar`. This in order to align to [Spinner](https://github.com/denoland/std/blob/2829d6d8ec9a9070cf1cee06622e78e163b12444/cli/unstable_spinner.ts#L218) and to allow instantiation of `ProgressBar` before usage.

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -139,10 +139,11 @@ export class ProgressBar {
   #unit: string;
   #rate: number;
   #writer: WritableStreamDefaultWriter;
-  #id: number;
-  #startTime: number;
-  #lastTime: number;
-  #lastValue: number;
+  #interval?: number;
+  #startTime: number = 0;
+  #lastTime: number = 0;
+  #lastValue: number = 0;
+
   /**
    * Constructs a new instance.
    *
@@ -186,12 +187,8 @@ export class ProgressBar {
     const stream = new TextEncoderStream();
     stream.readable
       .pipeTo(writable, { preventClose: this.#options.keepOpen })
-      .catch(() => clearInterval(this.#id));
+      .catch(() => clearInterval(this.#interval));
     this.#writer = stream.writable.getWriter();
-    this.#id = setInterval(() => this.#print(), 200);
-    this.#startTime = performance.now();
-    this.#lastTime = this.#startTime;
-    this.#lastValue = this.#options.value;
   }
 
   async #print(): Promise<void> {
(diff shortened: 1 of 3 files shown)


BlackAsLight:
The start method should allow for the options to be optionally provided again in case the user wants to change values, like the max.

This allows it to be reused later down the line.

timreichen:
> The start method should allow for the options to be optionally provided again in case the user wants to change values, like the max.
> 
> This allows it to be reused later down the line.

Instead of allowing `start()` to pass another options object, I think it is more *web api-ish* to have all properties public that can be changed while progressBar is running or when it is reused. Something along the lines of

```ts
const progressBar = new ProgressBar()
progressBar.start()
...
progressBar.stop()

...
progressBar.max = 100
progressBar.start()
...
```

The other benefit of that is that a public `value` property makes `add()` obsolete because it can be done with `progressBar.value += increment`. It is also more flexible be cause the value can be set directly `progressBar.value = newValue`.

WDYT?

BlackAsLight:
I think some properties like value, max, barLength and fmt could make sense, but other properties like clear and keepOpen could break things/ have no effect. 

Looking at the current state of the code even though one can call start and stop multiple times, the progress bar can't be reused. The interval ID isn't cleared for a second round and the writer is closed when stop is called, possibly closing the writable passed in.

kt3k:
I'm not in favor of this change as the added complexity is not small, but this doesn't support any new practical scenario.

kt3k:
As @BlackAsLight pointed, this introduces a new bugs related to start method. Because `end` method doesn't reset `this.#intervalId` to undefined, the second `start` call doesn't print the progress anymore. Also if the user passed `keepOption: false`, then the `#writer.close()` call in `end()` closes the writable, and the user gets the uncaught error in `#print` function. These need to be fixed with new test cases. I would see them fair amount of added complexity. If we don't support `start` method, those scenarios don't exist.

timreichen:
> As @BlackAsLight pointed, this introduces a new bugs related to start method. Because `end` method doesn't reset `this.#intervalId` to undefined, the second `start` call doesn't print the progress anymore. Also if the user passed `keepOption: false`, then the `#writer.close()` call in `end()` closes the writable, and the user gets the uncaught error in `#print` function after the 2nd `start` call. These need to be fixed with new test cases. I would see them fair amount of added complexity. If we don't support `start` method, those scenarios don't exist.

I'll have another look after some of the other PRs have landed. I imagine some of the mentioned issues are add not too much complexity, some are even already implemented in `Spinner`. I think to have a `start()` method is worth it not least because lots of web apis use the convention of class initialization as a setup and a method to start the process ([MutationObserver.observe()](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe), [SpeechRecognition.start()](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition/start), etc.)

kt3k:
> lots of web apis use the convention of class initialization as a setup and a method to start the process ([MutationObserver.observe()](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe), [SpeechRecognition.start()](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition/start), etc.)

There are also APIs which start the main processing at the constructor such as `WebSocket` `Worker`. I don't think having `start` is particularly aligned better with Web platform APIs.

kt3k:
Also alignment to the web platform api shouldn't be the goal in my view. It should be only supplementary for design decisions when there are equally possible designs.

If the alignment to web apis causes additional unnecessary complexity, then I think we should avoid it.

timreichen:
stale

PR #6409, opened 2025-02-15
BREAKING(cli/unstable): make `ProgressBar` `writable` optional

timreichen:
This PR removes the `ProgressBar` `writable` argument and adds it as an optional property in `options` with `Deno.stderr.writable` as the default.


cli/unstable_progress_bar.ts
```diff
@@ -146,14 +151,11 @@ export class ProgressBar {
...
-   * @param writable The {@link WritableStream} that will receive the progress bar reports.
    * @param options The options to configure various settings of the progress bar.
    */
-  constructor(
-    writable: WritableStream<Uint8Array>,
-    options: ProgressBarOptions,
-  ) {
+  constructor(options: ProgressBarOptions) {
     this.#options = {
+      writable: options.writable ??= Deno.stdout.writable,
```
andrewthauer, line 158: Can we make this `stderr` by default based on #5495? I started a thread with @kt3k about changing `spinner` to use `stderr` by default in #6350 as well.

BlackAsLight: I'm not a big fan of there being a default, but if there is one, based off what Andrew is saying here, it makes sense for it to be stderr

andrewthauer: @BlackAsLight - I'm curious your reasons for not wanting a default here?

BlackAsLight: I just think it's better for people to make the conscious choice about where it's going than perceive it as magic.

timreichen: Done

kt3k (approved):
LGTM

PR #6410, opened 2025-02-16
feat(text/unstable): add `reverse` function

scarf005:
Adds Unicode-aware string reverse function from https://github.com/mathiasbynens/esrever. Unicode handling can be disabled for extra performance.

Compared to `input.split().reverse().join()`:
- 1.69x faster (with Unicode handling)
- 2.91x faster (without Unicode handling)

<details><summary>Benchmark</summary>

```
scarf@fedora ~/r/d/deno_std (feat/text-unstable/reverse)> deno bench text/unstable_reverse_bench.ts
Check file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts
    CPU | AMD Ryzen 5 5600G with Radeon Graphics
Runtime | Deno 2.1.10+56f67b5 (x86_64-unknown-linux-gnu)

file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts

benchmark              time/iter (avg)        iter/s      (min … max)           p75      p99     p995
---------------------- ----------------------------- --------------------- --------------------------

group reverseString
splitReverseJoin                7.4 ms         135.7 (  6.8 ms …   8.7 ms)   7.5 ms   8.7 ms   8.7 ms
forOf                           4.0 ms         249.7 (  3.3 ms …   6.7 ms)   4.1 ms   6.7 ms   6.7 ms
reduce                          3.1 ms         323.4 (  2.5 ms …   5.8 ms)   3.2 ms   5.1 ms   5.8 ms
spreadReverseJoin               8.9 ms         112.3 (  8.5 ms …  10.5 ms)   9.0 ms  10.5 ms  10.5 ms
forLoop                         2.7 ms         373.8 (  2.4 ms …   4.7 ms)   2.7 ms   4.2 ms   4.7 ms
esrever                         4.2 ms         235.9 (  3.9 ms …   5.6 ms)   4.3 ms   5.2 ms   5.6 ms
esrever (no unicode)            2.6 ms         378.1 (  2.4 ms …   3.1 ms)   2.7 ms   3.0 ms   3.0 ms

summary
  esrever (no unicode)
     1.01x faster than forLoop
     1.17x faster than reduce
     1.51x faster than forOf
     1.60x faster than esrever
     2.79x faster than splitReverseJoin
     3.37x faster than spreadReverseJoin
```

</details> 

didn't apply changes from https://github.com/mathiasbynens/esrever/pull/12 as V8 seemed to to small string concatenation optimization and the push-based solution was 8x slower. 

<details><summary>Benchmark</summary>

```
scarf@fedora ~/r/d/deno_std (feat/text-unstable/reverse) [SIGINT]> deno bench text/unstable_reverse_bench.ts
Check file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts
    CPU | AMD Ryzen 5 5600G with Radeon Graphics
Runtime | Deno 2.1.10+56f67b5 (x86_64-unknown-linux-gnu)

file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts

benchmark                    time/iter (avg)        iter/s      (min … max)           p75      p99     p995
---------------------------- ----------------------------- --------------------- --------------------------

group reverseString
esrever                             185.4 ms           5.4 (174.9 ms … 206.4 ms) 189.4 ms 206.4 ms 206.4 ms
esrever (no unicode)                184.2 ms           5.4 (180.0 ms … 191.6 ms) 187.5 ms 191.6 ms 191.6 ms
esrever (old)                        22.2 ms          45.0 ( 21.3 ms …  23.7 ms)  22.6 ms  23.7 ms  23.7 ms
esrever (no unicode) (old)           22.3 ms          44.8 ( 21.0 ms …  29.8 ms)  22.4 ms  29.8 ms  29.8 ms

summary
  esrever (old)
     1.01x faster than esrever (no unicode) (old)
     8.29x faster than esrever (no unicode)
     8.35x faster than esrever
```

</details>

### DIFF
--- a/text/unstable_reverse.ts
+++ b/text/unstable_reverse.ts
@@ -1,0 +1,89 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+// This module is browser compatible.
+
+// Copyright Mathias Bynens <https://mathiasbynens.be/>
+//
+// Permission is hereby granted, free of charge, to any person obtaining
+// a copy of this software and associated documentation files (the
+// "Software"), to deal in the Software without restriction, including
+// without limitation the rights to use, copy, modify, merge, publish,
+// distribute, sublicense, and/or sell copies of the Software, and to
+// permit persons to whom the Software is furnished to do so, subject to
+// the following conditions:
+//
+// The above copyright notice and this permission notice shall be
+// included in all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+const REGEX_SYMBOL_WITH_COMBINING_MARKS = /(\P{M})(\p{M}+)/gu;
+const REGEX_SURROGATE_PAIR = /([\uD800-\uDBFF])([\uDC00-\uDFFF])/g;
+
+/** Options for {@linkcode reverse}  */
+export type ReverseOptions = {
+  /**
+   * Whether to handle Unicode symbols such as 🦕 at the cost of ~60% slowdown.
+   *
+   * Check {@link ./unstable_reverse_bench.ts} for performance comparison.
+   *
+   * @default {true}
+   */
+  handleUnicode: boolean;
+};
+
+/**
+ * Performs a Unicode-aware string reversal.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @param input - The input string to be reversed.
+ * @param options The options for the reverse function.
+ * @returns The reversed string.
+ *
+ * @example Standard usage
+ * ```ts
+ * import { reverse } from "@std/text/unstable-reverse";
+ * import { assertEquals } from "@std/assert";
+ *
+ * assertEquals(reverse("Hello, world!"), "!dlrow ,olleH");
+ * assertEquals(reverse("🦕Deno♥"), "♥oneD🦕");
+ * ```
+ *
+ * @example Performance optimization with disabled Unicode handling
+ * ```ts
+ * import { reverse } from "@std/text/unstable-reverse";
+ * import { assertEquals } from "@std/assert";
+ *
+ * assertEquals(reverse("Hello, world!", { handleUnicode: false }), "!dlrow ,olleH");
+ * ```
+ */
+export function reverse(
+  input: string,
+  options?: Partial<ReverseOptions>,
+): string {
+  if (options?.handleUnicode !== false) {
+    // Step 1: deal with combining marks and astral symbols (surrogate pairs)
+    input = input
+      // Swap symbols with their combining marks so the combining marks go first
+      .replace(REGEX_SYMBOL_WITH_COMBINING_MARKS, (_, $1, $2) => {
+        // Reverse the combining marks so they will end up in the same order
+        // later on (after another round of reversing)
+        return reverse($2) + $1;
+      })
+      // Swap high and low surrogates so the low surrogates go first
+      .replace(REGEX_SURROGATE_PAIR, "$2$1");
+  }
+
+  // Step 2: reverse the code units in the string
+  let result = "";
+  for (let index = input.length; index--;) {
+    result += input.charAt(index);
+  }
+  return result;
+}
--- a/text/_test_util.ts
+++ b/text/_test_util.ts
@@ -1,0 +1,6 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+export function generateRandomString(min: number, max: number): string {
+  return Array.from({ length: Math.floor(Math.random() * (max - min) + min) })
+    .map(() => String.fromCharCode(Math.floor(Math.random() * 26) + 97))
+    .join("");
+}
--- a/text/_test_util_test.ts
+++ b/text/_test_util_test.ts
@@ -1,0 +1,11 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+import { assertEquals } from "../assert/equals.ts";
+import { generateRandomString } from "./_test_util.ts";
+
+Deno.test({
+  name: "generateRandomString() generates a string of the correct length",
+  fn() {
+    assertEquals(generateRandomString(0, 0), "");
+    assertEquals(generateRandomString(10, 10).length, 10);
+  },
+});
--- a/text/deno.json
+++ b/text/deno.json
@@ -6,6 +6,7 @@
     "./closest-string": "./closest_string.ts",
     "./compare-similarity": "./compare_similarity.ts",
     "./levenshtein-distance": "./levenshtein_distance.ts",
+    "./unstable-reverse": "./unstable_reverse.ts",
     "./unstable-slugify": "./unstable_slugify.ts",
     "./to-camel-case": "./to_camel_case.ts",
     "./unstable-to-constant-case": "./unstable_to_constant_case.ts",
--- a/text/unstable_reverse_bench.ts
+++ b/text/unstable_reverse_bench.ts
@@ -1,0 +1,104 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+import { generateRandomString } from "./_test_util.ts";
+import { reverse } from "./unstable_reverse.ts";
+
+function splitReverseJoin(str: string) {
+  return str.split("").reverse().join("");
+}
+
+function forOf(str: string) {
+  let reversed = "";
+  for (const character of str) {
+    reversed = character + reversed;
+  }
+  return reversed;
+}
+
+function reduce(str: string) {
+  return str.split("").reduce(
+    (reversed, character) => character + reversed,
+    "",
+  );
+}
+
+function spreadReverseJoin(str: string) {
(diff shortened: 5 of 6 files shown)


text/_util.ts
```diff
@@ -18,3 +18,9 @@ export function splitToWords(input: string) {
 export function capitalizeWord(word: string): string {
   return word ? word[0]!.toUpperCase() + word.slice(1).toLowerCase() : word;
 }
+
+export function generateRandomString(min: number, max: number): string {
```
kt3k, line 22: Because this util is only used in benchmark, can you move this to somewhere else? (like `_test_util.ts`?)

text/unstable_reverse.ts
```diff
@@ -0,0 +1,90 @@
...
+// EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+// MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+// NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+// LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+// OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+// WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
+const REGEX_SYMBOL_WITH_COMBINING_MARKS =
+  /([\0-\u02FF\u0370-\u1AAF\u1B00-\u1DBF\u1E00-\u20CF\u2100-\uD7FF\uE000-\uFE1F\uFE30-\uFFFF]|[\uD800-\uDBFF][\uDC00-\uDFFF]|[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?:[^\uD800-\uDBFF]|^)[\uDC00-\uDFFF])([\u0300-\u036F\u1AB0-\u1AFF\u1DC0-\u1DFF\u20D0-\u20FF\uFE20-\uFE2F]+)/g;
+const REGEX_SURROGATE_PAIR = /([\uD800-\uDBFF])([\uDC00-\uDFFF])/g;
```
kt3k, line 27: I wonder if there's a way to do this using unicode character classes..

cc @lionel-rowe

text/_util.ts
scarf005: @timreichen sorry, but could you elaborate? i don't follow how shuffle could be used to generate random string.

scarf005: @kt3k applied and force-pushed in https://github.com/denoland/std/commit/ed38858d32ccae418dbff4a195e2c435c61f36f4

text/unstable_reverse.ts
scarf005: @kt3k I've simplified `REGEX_SYMBOL_WITH_COMBINING_MARKS` in https://github.com/denoland/std/commit/b6e44717db535b0fff170aae31ac6d999881e602 with help of [chatgpt](https://chatgpt.com/share/67b2fc1b-f8d0-8002-becc-a30cfc48b32b), but i wasn't able to simplify `REGEX_SURROGATE_PAIR`.

lionel-rowe: > I wonder if there's a way to do this using unicode character classes..
> 
> cc @lionel-rowe

There's `\p{Surrogate}` for an orphan surrogate. For a well-formed surrogate pair, `[\u{10000}-\u{10ffff}]` will match it in `u`/`v`-mode regex.

But I'm not convinced having this in `std` provides much additional utility over `Intl.Segmenter` with `granularity: 'grapheme'`, which didn't exist when `esrever` was published:

```ts
function reverse(str: string) {
    return [...new Intl.Segmenter('en-US', { granularity: 'grapheme' }).segment(str)]
        .map((x) => x.segment).reverse().join('')
}

reverse('mañana 🌈'.normalize('NFD'))
// '🌈 anañam'
```

> The correct (well, the most correct) way to do this is to implement [text segmentation as per TR29](http://www.unicode.org/reports/tr29/#Grapheme_Cluster_Boundaries) and then reverse each grapheme cluster (as well as swapping surrogate pairs) before further processing the string as usual.
>  

 _Originally posted by @mathiasbynens in [#5](https://github.com/mathiasbynens/esrever/issues/5#issuecomment-41997022) (back in 2014, before `Intl.Segmenter`)_

scarf005:
```
scarf@fedora /r/m/h/s/r/d/deno_std (feat/text-unstable/reverse)> deno bench text/unstable_reverse_bench.ts
Check file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts
    CPU | AMD Ryzen 5 5600G with Radeon Graphics
Runtime | Deno 2.1.10+56f67b5 (x86_64-unknown-linux-gnu)

file:///run/media/home/scarf/repo/deno/deno_std/text/unstable_reverse_bench.ts

benchmark              time/iter (avg)        iter/s      (min … max)           p75      p99     p995
---------------------- ----------------------------- --------------------- --------------------------

group reverseString
splitReverseJoin                6.8 ms         147.2 (  6.4 ms …   7.9 ms)   6.9 ms   7.9 ms   7.9 ms
forOf                           3.1 ms         320.1 (  3.0 ms …   3.3 ms)   3.2 ms   3.3 ms   3.3 ms
reduce                          2.5 ms         392.7 (  2.3 ms …   2.8 ms)   2.6 ms   2.8 ms   2.8 ms
spreadReverseJoin               8.3 ms         120.2 (  8.0 ms …   9.0 ms)   8.6 ms   9.0 ms   9.0 ms
forLoop                         2.4 ms         421.0 (  2.2 ms …   4.8 ms)   2.4 ms   3.7 ms   4.0 ms
esrever                         3.9 ms         254.0 (  3.7 ms …   5.7 ms)   4.0 ms   4.3 ms   5.7 ms
esrever (no unicode)            2.4 ms         423.3 (  2.2 ms …   3.7 ms)   2.4 ms   2.6 ms   3.7 ms
Intl.Segmenter                143.5 ms           7.0 (139.1 ms … 149.6 ms) 145.3 ms 149.6 ms 149.6 ms

summary
  Intl.Segmenter
    60.74x slower than esrever (no unicode)
    60.41x slower than forLoop
    56.35x slower than reduce
    45.93x slower than forOf
    36.44x slower than esrever
    21.12x slower than splitReverseJoin
    17.24x slower than spreadReverseJoin
```
while it handles hebrew correctly, it seems to have subpar performance in general. maybe we could add an 'mode' to choose between ascii, regex and Intl.Segmenter?

<details><summary>function used</summary>

```ts
function segmenter(str: string) {
  let x = "";
  for (
    const { segment } of new Intl.Segmenter("en-US", {
      granularity: "grapheme",
    }).segment(str)
  ) {
    x = segment + x;
  }
  return x;
}
```

</details>

lionel-rowe: @scarf005 I think some plausible use cases would help to clear up what functionality/performance characteristics are useful. I do a lot of text processing, but I can't recall the last time I needed to reverse a string, much less reverse a very long string or thousands of short strings. And the use cases would also help determine whether per-grapheme handling is even useful at all, versus per-UTF-8 codepoint, or (probably less commonly) per-UTF-16 code unit.

scarf005: @lionel-rowe 

> I think some plausible use cases would help to clear up what functionality/performance characteristics are useful. I do a lot of text processing, but I can't recall the last time I needed to reverse a string,

For example, palindrome checking and sorting in suffix order. https://old.reddit.com/r/AskComputerScience/comments/k49xmv/what_is_the_applicationfunction_of_reversed/ge8ku7p/

> I do a lot of text processing, but I can't recall the last time I needed to reverse a string, much less reverse a very long string or thousands of short strings.

However, it would be better if the function was performant regardless of length and amount.

> And the use cases would also help determine whether per-grapheme handling is even useful at all, versus per-UTF-8 codepoint, or (probably less commonly) per-UTF-16 code unit.

Per-grapheme handling seems to be a niche use case. I want to keep Unicode support since it affects my use case (CJK Unicode handling).

0f-0b: Even with per-grapheme handling, there are still characters that aren't easily reversible, such as [U+070F](https://en.wikipedia.org/wiki/Syriac_Abbreviation_Mark) which adds an overline to all subsequent letters until the first non-Syriac character.

lionel-rowe: > Per-grapheme handling seems to be a niche use case. I want to keep Unicode support since it affects my use case (CJK Unicode handling).

By "Unicode support" _as opposed to grapheme support_ I assume you mean per-Unicode code point[^1], in which case any of `[...spread]`, `Array.from`, `for...of` etc. will work. But the answers would intuitively be "wrong" for examples like `'mañana'.normalize('NFD') => 'anãnam'` (versus `'mañana'.normalize('NFC') => 'anañam'`)[^2]. On the other hand, that might be the desired result, if the unit you care about is Unicode code points.

To complicate things further, many people would understand "Unicode support" to _include_ grapheme support. Or perhaps more often, they wouldn't think about it too much and just assume Unicode support means it will "do the right thing" for multilingual text in any arbitrary language. The latter expectation is impossible to fulfill, because Unicode and its associated projects like CLDR and ICU can never capture the variability and evolving nature of natural language, and there will always be edge cases that aren't covered like @0f-0b 's example above.

Suffice to say, "grapheme", "code point", and "UTF-16 code unit" all have standardized meanings that are widely understood and relatively stable over time, so I'd suggest using one or all of those as primitives. And of those, UTF-16 code unit is only really useful for interoperability or where preserving numbered indexes in JS is desirable, so seems unlikely to be useful here.

[^1]: As an aside, generally speaking, all _common_ CJK characters fall within the BMP anyway, so per-UTF-16 code unit (`str.split('')` etc.) would be sufficient for them. Only rare characters like 𰻝 lie outside the BMP. But CJK has a loooooong tail of very rare characters.

[^2]: Normalizing to `'NFC'` before reversing would fix most (but not all) of these intuitively wrong results, but the invariant `reverse(str).length === str.length` would no longer always hold true.

kt3k (approved):
LGTM

kt3k:
Is there strong objection to the current state?

timreichen:
I would like the tests better if they had static values rather than generated ones.

scarf005:
> I would like the tests/benchs better if they had static values rather than generated ones.

- tests use static values.
- benchmark uses generated strings, but i'm unsure switching to static values would be beneficial.

PR #6412, opened 2025-02-16
test(async): fix flaky `pooledMap()` test

timreichen:
ref: https://github.com/denoland/std/issues/6287

This PR increases `setTimeout` intervals and therefore adds more leeway to checks.


kt3k (approved):
LGTM

PR #6413, opened 2025-02-16
fix(async/unstable): fix flaky `waitFor()` test

timreichen:
ref: https://github.com/denoland/std/actions/runs/13358554953/job/37304692671?pr=6412

This PR increases the duration of `waitFor()` to add more leeway.

### DIFF
--- a/async/unstable_wait_for_test.ts
+++ b/async/unstable_wait_for_test.ts
@@ -18,11 +18,11 @@ Deno.test("waitFor() throws DOMException on timeout", async () => {
   const id = setTimeout(() => flag = true, 1000);
   const start = Date.now();
   const error = await assertRejects(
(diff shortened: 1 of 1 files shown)


kt3k:
I sometimes see greater diff in flaky failure. Can we use `FakeTime` from `@std/testing/time` instead?

timreichen:
Does the 

> I sometimes see greater diff in flaky failure. Can we use `FakeTime` from `@std/testing/time` instead?

Actually, I think we can fix this by using `assertLess()` instead of `assertAlmostEquals()`. The times should always be less than 1000` which would trigger a different outcome on both tests. WDYT?

kt3k (approved):
Ah, that also sounds reasonable to me. LGTM. Thanks!

PR #6416, opened 2025-02-18
feat(fs/unstable): add truncate and truncateSync

jbronder:
This PR adds `truncate` and `truncateSync` APIs to the `@std/fs` package which are intended to mirror `Deno.truncate` and `Deno.truncateSync` functions.

Towards #6255.


kt3k:
Mostly looks good to me.

We started adding unstable fs APIs to doc checker (`_tools/check_docs.ts`) since #6400. Can you also add this file to that check?

kt3k (approved):
LGTM. Thanks!

PR #6418, opened 2025-02-18
refactor(front-matter): cleanup `_shared.ts`

timreichen:
This PR
- `recognize()` to `any.ts` as it is only used there.
- renames `data` capturing group to `frontMatter` so it is the same name as in `Extract`.
- rewrites `extractAndParse` as `extractFrontMatter` and do the parsing in the appropriate functions. This removes the need of the `Parser` type.


kt3k (approved):
LGTM. Nice clean ups!

PR #6419, opened 2025-02-18
BREAKING(cache/unstable): configurable cache ejection of thrown or rejected values, change cache type (#6417)

### LINKED ISSUE #6417: (unstable) cache/memoize: Make caching of rejected promises (and maybe thrown errors?) configurable

Current behavior:

> Memoization will not cache thrown errors and will eject promises from the cache upon rejection. If you want to retain errors or rejected promises in the cache, you will need to catch and return them.

However, I'm finding that I frequently want to configure this behavior when it comes to rejected promises from API responses, similar to https://github.com/denoland/std/issues/6196 for `async/retry`.

Something like this:

```ts
const fetchPostById = memoize(async (postId) => {
    const res = await fetch(`https://jsonplaceholder.typicode.com/posts/${postId}`)
    if (!res.ok) throw new HttpError(res.status)
    return res
}, {
    isRetriable: (e) => e instanceof HttpError && (e.status === 429 || e.status >= 500),
})
```

Then the promise would only be ejected from the cache if the rejection reason returned `true` for `isRetriable` (open to bikeshedding over option name; `isRetriable` is the name used for the equivalent `async/retry` option).

To fulfil the common user expectation that promise rejections are equivalent to thrown errors, synchronously thrown values returning `false` for `isRetriable` should probably also be cached. That could also be useful in cases where the error is only thrown after expensive synchronous calculation has been carried out.

Should subsequent throws/rejections for the same inputs:
1. Re-use the exact same thrown value?
2. Attempt to clone it in some way?
3. Use a custom error type?
4. Wrap it in an `AggregateError`?
5. Something else?

For simplicity's sake, I'm leaning toward 1.

Also, should the default behavior be changed while `cache/memoize` is still unstable?

lionel-rowe:
Closes #6417.

I don't love how verbose this makes passing an `LruCache` instance as a `memoize` option. Ideally it'd be possible to just pass it as a literal `{ cache: new LruCache(100) }` with the type params being inferred as what they're supposed to be, but I don't think it's possible to do that in TypeScript (???). Passing `{ cache: new Map() }` works, but only because `new Map()` is `Map<any, any>`, whereas `new LruCache(n)` is `LruCache<unknown, unknown>`.


kt3k (approved):
LGTM

kt3k:
I'll keep this PR open for a while to see if there are more feedbacks

PR #6420, opened 2025-02-18
refactor(cli/unstable): clean up `start` end `stop` methods of `Spinner`, add test cases for multiple start/stop calls

timreichen:
This PR adds test for multiple `start()` and `stop()` calls.
In addition, it removes the `#active` property and replaces it with `this.#intervalId === null` check. This should work identically.

### DIFF
--- a/cli/unstable_spinner.ts
+++ b/cli/unstable_spinner.ts
@@ -139,8 +139,7 @@ export class Spinner {
 
   #interval: number;
   #color: Color | undefined;
-  #intervalId: number | undefined;
-  #active = false;
+  #intervalId: number | null = null;
   #output: typeof Deno.stdout | typeof Deno.stderr;
 
   /**
@@ -216,11 +215,10 @@ export class Spinner {
    * ```
    */
   start() {
-    if (this.#active || this.#output.writable.locked) {
+    if (this.#intervalId !== null || this.#output.writable.locked) {
       return;
     }
 
-    this.#active = true;
     let i = 0;
     const noColor = Deno.noColor;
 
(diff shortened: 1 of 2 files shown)


cli/unstable_spinner.ts
```diff
@@ -139,8 +139,7 @@ export class Spinner {
 
   #interval: number;
   #color: Color | undefined;
-  #intervalId: number | undefined;
-  #active = false;
+  #intervalId: number | null = null;
```
kt3k, line 142: nit: why do you change this type?

timreichen: Because `this.#intervalId` is unset  in `stop()` . I think `null` shows more clearly that it is intentionally unset than `undefined`. Also because of the nature how js works, `this.#intervalId` cannot be checked with `if (!this.#intervalId)` anymore since the value `0` would pass (even though  I think `setInterval` would probably never return 0). So `(this.#intervalId == null)` seemed nicer.

BlackAsLight: If my understanding of what the spec says on setTimeout and setInterval, the ID returned from them has to be greater than zero.

kt3k: > the ID returned from them has to be greater than zero.

That sounds correct

https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#timer-initialisation-steps

> If previousId was given, let id be previousId; otherwise, let id be an [implementation-defined](https://infra.spec.whatwg.org/#implementation-defined) integer that is greater than zero and does not already [exist](https://infra.spec.whatwg.org/#map-exists) in global's [map of setTimeout and setInterval IDs](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#map-of-settimeout-and-setinterval-ids).

kt3k:
Changing the commit type to `refactor` and aligning the title.

kt3k (approved):
LGTM

PR #6424, opened 2025-02-19
docs(fmt): fix `printf` module doc absense

0x009922:
`@std/fmt/printf` module docs are currently absent: https://jsr.io/@std/fmt/doc/printf

It seems that the reason is that the `@module` JSDoc section happens to be after an import. So, I move the import below and it fixes the issue (checked locally with `deno doc`).

Anyway, I think `deno doc` should probably consider this case in some way - either force to put the module doc as the first item in the module, or to allow putting the module doc anywhere in the module.

### DIFF
--- a/fmt/printf.ts
+++ b/fmt/printf.ts
@@ -1,7 +1,5 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
-import { stripAnsiCode } from "./colors.ts";
-
 /**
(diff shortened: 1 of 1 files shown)


kt3k (approved):
Good findings! LGTM

> Anyway, I think deno doc should probably consider this case in some way - either force to put the module doc as the first item in the module, or to allow putting the module doc anywhere in the module.

This sounds like a reasonable suggestion.

cc @crowlKats

PR #6426, opened 2025-02-19
test(async/unstable): add `retry()` test with default options

timreichen:
This PR adds a test that doesn't pass any options to `retry()`.

### DIFF
--- a/async/retry.ts
+++ b/async/retry.ts
@@ -119,7 +119,7 @@ export interface RetryOptions {
  */
 export async function retry<T>(
   fn: (() => Promise<T>) | (() => T),
-  options?: RetryOptions,
+  options: RetryOptions = {},
 ): Promise<T> {
   const {
     multiplier = 2,
@@ -127,7 +127,7 @@ export async function retry<T>(
     maxAttempts = 5,
     minTimeout = 1000,
     jitter = 1,
-  } = options ?? {};
+  } = options;
 
   if (maxTimeout <= 0) {
     throw new TypeError(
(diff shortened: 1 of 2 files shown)


async/retry.ts
```diff
@@ -119,15 +119,15 @@ export interface RetryOptions {
  */
 export async function retry<T>(
   fn: (() => Promise<T>) | (() => T),
-  options?: RetryOptions,
+  options: RetryOptions = {},
```
kt3k, line 122: Why do we need this change?

BTW there's an occurrence of the same structure in `unstable_retry.ts`

timreichen: For some reason the options setup lines were not marked as covered in `codecov` in a previous PR before this change. Might have been a bug though.

kt3k: `retry` seems covered 100% in codecov including option setup line https://app.codecov.io/gh/denoland/std/blob/main/async%2Fretry.ts

async/retry_test.ts
```diff
@@ -151,3 +151,9 @@ Deno.test("retry() checks backoff function timings", async (t) => {
 
   Math.random = originalMathRandom;
 });
+
+Deno.test("retry() with default options", async () => {
```
kt3k, line 155: The default option behavior is tested in `"retry() waits four times by default"` case

kt3k:
This seems only adding duplicated test case which is already tested in the case `retry() waits four times by default`

Closing without merge

timreichen:
> This seems only adding duplicated test case which is already tested in the case `retry() waits four times by default`
> 
> 
> 
> Closing without merge

I think there was a problem in `codecov` where some lines weren't marked as covered when in fact they were. Alright👍

Issue #6427, opened 2025-02-19
`assertArrayIncludes` should fail type checking if elements of `expected` aren't assignable to the element type of `actual`


lionel-rowe:
**Describe the bug**

`assertArrayIncludes` currently passes type checking if elements of `expected` aren't assignable to the element type of `actual`.

**Steps to Reproduce**

```ts
declare const ones: 1[]
declare const twos: 2[]

Array.includes(ones, twos[0]) // fails type checking, as expected
assertArrayIncludes(ones, twos) // passes type checking
```

**Expected behavior**

`assertArrayIncludes(ones, twos)` should fail type checking.

Wrapping the `expected` param in the `NoInfer` utility type seems to fix it.

**Environment**

- OS: Ubuntu
- deno version: 2.1.10
- std version: assert@1.0.11

PR #6428, opened 2025-02-19
fix(assert): fail type check for `assertArrayIncludes` when element type of `expected` not assignable to `actual` (#6427)

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6427

### DIFF
--- a/collections/mod.ts
+++ b/collections/mod.ts
@@ -18,7 +18,7 @@
  *
  * assertEquals(intersect(lisaInterests, kimInterests), ["Cooking", "Music"]);
  *
- * assertArrayIncludes(lisaInterests, [sample(lisaInterests)]);
+ * assertArrayIncludes(lisaInterests, [sample(lisaInterests)!]);
  *
  * const cat = { name: "Lulu", age: 3, breed: "Ragdoll" };
  *
--- a/assert/array_includes.ts
+++ b/assert/array_includes.ts
@@ -29,7 +29,7 @@ export type ArrayLikeArg<T> = ArrayLike<T> & object;
  */
 export function assertArrayIncludes<T>(
   actual: ArrayLikeArg<T>,
-  expected: ArrayLikeArg<T>,
+  expected: NoInfer<ArrayLikeArg<T>>,
   msg?: string,
 ) {
   const missing: unknown[] = [];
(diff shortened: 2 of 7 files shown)


collections/mod.ts
```diff
@@ -18,7 +18,7 @@
  *
  * assertEquals(intersect(lisaInterests, kimInterests), ["Cooking", "Music"]);
  *
- * assertArrayIncludes(lisaInterests, [sample(lisaInterests)]);
+ * assertArrayIncludes(lisaInterests, [sample(lisaInterests)!]);
```
kt3k, line 21: hmm this change seems concerning to me (I think this PR will break actual users code in a similar way)

Can we keep this case working without change somehow?

kt3k:
While I think this change is aligned to the direction we chose in #2203, I'm also very concerned with the breaking nature of this PR.

cc @bartlomieju @ry

collections/mod.ts
lionel-rowe: > hmm this change seems concerning to me (I think this PR will break actual users code in a similar way)
> 
> Can we keep this case working without change somehow?

I don't think so, without explicitly allowing `undefined`, which seems wrong.

For comparison, again with `Array#includes`:

```ts
declare let xs: string[]
declare let x: string | undefined

// Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
xs.includes(x)
```

But on second thoughts, maybe the strict behavior rejecting legitimate uses is more annoying than the sloppy behavior accepting illegitimate ones/lacking autocomplete... it's not uncommon that I have to write code like this:

```ts
stringLiteralUnionArr.includes(arbitraryString as typeof stringLiteralUnionArr[number])
```

bartlomieju:
I tend to agree. Maybe try using this PR in the Deno repo submodule and see how bad it's gonna be?

kt3k:
It turned out that denoland/deno repo doesn't include any usage of `assertArrayIncludes`

```
~/denoland/deno $ git grep assertArrayIncludes | wc -l
       0
```

There seem relatively small number of usages in github search https://github.com/search?q=assertArrayIncludes+language%3ATypeScript&type=code&l=TypeScript

Maybe we can land this as 'fix'?

kt3k:
https://github.com/denoland/std/pull/6428#discussion_r1963230649

> But on second thoughts, maybe the strict behavior rejecting legitimate uses is more annoying than the sloppy behavior accepting illegitimate ones/lacking autocomplete... it's not uncommon that I have to write code like this:

I now agree with this opinion. The current behavior looks more ergonomic than the proposed behavior. Let's not do this

Thanks for your suggestion anyway @lionel-rowe

PR #6430, opened 2025-02-19
BREAKING(cli/unstable): make `ProgressBar` `value` and `max` properties public

timreichen:
Ref: https://github.com/denoland/std/pull/6408#issuecomment-2661396298

This PR makes the `ProgressBar` `value` and `max` properties public.
This is more flexible as the progress bar `max` can be adjusted after instantiation and allows to set the `value` directly (`progressBar.value = newValue`) instead of only adding increments.
This makes the `add()` method obsolete as the same can be achieved with `progressBar.value += x`.


kt3k (approved):
It makes sense to me to make `max` modifiable as I sometimes see such progress indicator (e.g. Deno's download indicator)

It also makes sense to make `value` public property as it now accepts the total `value` number in addition to the difference. (It should be convenient when the progress source only reports the total current number.)

LGTM

kt3k:
@timreichen Thanks for updating!

PR #6431, opened 2025-02-20
test(fs/unstable): add test case for reading text file with BOM

### LINKED ISSUE #6422: Add test case of reading a file with byte order mark to `fs/unstable_read_text_file_test.ts`
ref https://github.com/denoland/std/pull/6405#issuecomment-2660843043

eryue0220:
Fixed #6422


fs/testdata/copy_file_bom.txt
```diff
@@ -0,0 +1,2 @@
+0x00, 0x00, 0xFE, 0xFF
```
kt3k, line 1: I couldn't see BOM in this file (or am I missing something..?)

Screenshot 2025-02-20 at 15 10 09

You can create a text file with BOM like:

```js
Deno.writeTextFileSync("hello.txt", "\ufeffhello");
```

eryue0220: Update. I made a mistake about the BOM.

fs/testdata/copy_file_bom.txt
```diff
@@ -0,0 +1 @@
+﻿hello
```
kt3k, line 1: Updated this file a bit. This file now has the below data (in binary):

Screenshot 2025-02-21 at 11 04 30

kt3k (approved):
I just realized `Deno.readTextFile` doesn't have special handling for BOM. Not sure this is a very meaningful test case, but anyway this shouldn't be harmful

LGTM

fs/testdata/copy_file_bom.txt
eryue0220: Thank you

PR #6432, opened 2025-02-20
fix(uuid): handle `rng` correctly in uuid v1

kt3k:
Currently if the user specifies `rng` option, `generate` returns a broken string like `0b4e9750-ef65-11ef-8000-01undefinedundefinedundefinedundefinedundefined` because `rng` is not called as a function. This PR fixes it.


kt3k:
The initial version already has this behavior https://github.com/denoland/deno/pull/4758/files

bartlomieju (approved):

PR #6433, opened 2025-02-20
refactor(cli/unstable): Declare `ProgressBar` options as separate properties

timreichen:
Ref: https://github.com/denoland/std/pull/6408#issuecomment-2661534263

This PR removes `ProgressBar` `#options` property in favour of separate private properties.

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -135,7 +135,15 @@ export interface ProgressBarOptions {
  * bar.end();
  */
 export class ProgressBar {
-  #options: Required<ProgressBarOptions>;
+  #value: number;
+  #max: number;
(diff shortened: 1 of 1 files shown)


kt3k:
What is resolved by this change? I can't correlate this change with the linked comment

timreichen:
> What is resolved by this change? I can't correlate this change with the linked comment

This paves the way for PRs that make some props like `max` public.
Also, wrapping these props in an object is not necessary in general because the object is not used or passed anywhere.

BlackAsLight:
I think it would make more sense for the pull request to actually be apart of the other one where `value` is exposed?

timreichen:
Closing in favour of https://github.com/denoland/std/pull/6430

PR #6436, opened 2025-02-20
feat(fs/unstable): add mkdir and mkdirSync

jbronder:
This PR adds `mkdir` and `mkdirSync` APIs to the `@std/fs` package which are intended to mirror `Deno.mkdir` and `Deno.mkdirSync` functions.

Towards #6255.


kt3k (approved):
LGTM. Nice!

PR #6440, opened 2025-02-22
BREAKING(tar/unstable): fix handling of mode, uid, and gid

BlackAsLight:
Closes: https://github.com/denoland/std/pull/6376

This pull request makes the changes discussed in the other linked pull request.

It changes mode, uid, and gid to accept and decode numbers in the octal format `0o775` instead of decimal format `775`. It also changes slightly the decoding segment of these three to be more compatible with other tar specs.

### DIFF
--- a/tar/tar_stream.ts
+++ b/tar/tar_stream.ts
@@ -68,17 +68,17 @@ type TarStreamInputInternal =
 export interface TarStreamOptions {
   /**
    * An octal literal.
-   * Defaults to 755 for directories and 644 for files.
+   * Defaults to 0o755 for directories and 0o644 for files.
    */
   mode?: number;
   /**
    * An octal literal.
-   * @default {0}
+   * @default {0o0}
    */
   uid?: number;
   /**
    * An octal literal.
-   * @default {0}
+   * @default {0o0}
    */
   gid?: number;
   /**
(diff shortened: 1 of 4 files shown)


tar/tar_stream.ts
```diff
@@ -222,9 +222,9 @@ export class TarStream implements TransformStream<TarStreamInput, Uint8Array> {
         header.set(name); // name
         header.set(
           encoder.encode(
-            options.mode.toString().padStart(6, "0") + " \0" + // mode
-              options.uid.toString().padStart(6, "0") + " \0" + //uid
-              options.gid.toString().padStart(6, "0") + " \0" + // gid
+            options.mode.toString(8).padStart(6, "0") + " \0" + // mode
+              options.uid.toString(8).padStart(6, "0") + " \0" + //uid
+              options.gid.toString(8).padStart(6, "0") + " \0" + // gid
```
kt3k, line 227: I missed this in #6376, but I don't think UID and GID are generally handled as octals. See docs of `Deno.chown` for example. https://docs.deno.com/api/deno/~/Deno.chown

Can we do these changes only for 'mode'?

BlackAsLight: We can but the spec is only expecting octal digits to be written

0f-0b: I agree that this change should be done for UID and GID as well. Currently, to set the UID to, for example, 501, the user would have to write `uid: 765` (the octal representation of 501 disguised as a decimal literal), which doesn't make sense to me.

```ts
import { TarStream, type TarStreamInput } from "jsr:@std/tar@0.1.5/tar-stream";

const file = new Blob(["hello world\n"]);
await ReadableStream.from<TarStreamInput>([
  {
    type: "file",
    path: "hello.txt",
    size: file.size,
    readable: file.stream(),
    options: { uid: 765, gid: 24, mtime: 0 },
  },
])
  .pipeThrough(new TarStream())
  .pipeTo(Deno.stdout.writable);
```

```shellsession
$ deno run a.ts | tar tv
-rw-r--r--  0 501    20         12 Jan  1  1970 hello.txt
```

kt3k (approved):
LGTM. Thanks for the fix!

PR #6442, opened 2025-02-23
refactor(encoding): adding consistency and performace

BlackAsLight:
related: https://github.com/denoland/std/issues/5944

This pull request rewrites the unstable modules of `@std/encoding` bringing performance and consistency in how the code is structured, executed, as well as a few new benefits. This pull request also brings the performance benefits to the stable modules without breaking anything.

The unstable modules have been simplified and merged. There was a lot of files before with many the only difference being the encoding format. This pull request merges those and asks for a simple string argument to specify what one you're after when decoding and encoding. For example:
```ts
encodeBase64("Hello World", "Base64");
encodeBase64("Hello World", "Base64Url");
```

Something to note is that when passing a `Uint8Array` or an `ArrayBuffer` to `encodeBase64` etc, in the unstable modules, the buffer is detached and returned with the encoded or decoded output. The reason for this is because most users wouldn't be using the buffer passed in afterwards and for the few that do, they can simply force a copy with `.slice()`. This means that less memory is created and copied around for the GC. 

This pull request also adds some new functions (`encodeRawBase64`) around encoding/ decoding in a raw manner for those who want to only transform part of a buffer in place while still maintaining the information before it. Allowing a more advanced and low level function to the existing ones that are also zero copy. A new util function is also available called `calcMax` to help users easily calculate how much space they'll actually need up front.

Lastly this pull request should bring 100% coverage to the unstable modules, but testing locally the coverage report didn't seem to hit lines I made tests, and verified, to hit those lines.


BlackAsLight:
Benchmarking these changes produce some interesting results.

With very small buffers (2 ** 5), the current encoding seems to out perform these new changes by 5x. Pushing the buffer to 2 ** 10 causes these changes to beat the current encoding by 2x, and moving the buffers to 2 ** 15, the changes beat the current encoding by 10x. These changes also seem capable of encoding the deno binary `encodeBase64(await Deno.readFile(Deno.execPath()), "Base64");` in less than 200ms where the current encoding would crash due to running out of memory.

When decoding the atob seems to always beat these new changes. T_T

BlackAsLight:
This pull request is getting split up into four. Once those three land, one more is expected to bring the performance benefits to the stable versions of the package

PR #6444, opened 2025-02-25
feat(fs/unstable): add writeFile and writeFileSync

jbronder:
This PR add `writeFile` and `writeFileSync` APIs to the `@std/fs` package which are functions intended to mirror `Deno.writeFile` and `Deno.writeFileSync`.  The PR also adds the `WriteFileOptions` type as well as `getFsFlag` and `getNodeStream` functions to prepare for implementing subsequent APIs such as `open(Sync)` and `writeTextFile(Sync)`. 

Towards #6255.


fs/unstable_write_file.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @param data The content in bytes or a stream of bytes to be written.
+ * @param options Options to write files.
+ */
+export async function writeFile(
+  path: string | URL,
+  data: Uint8Array | ReadableStream<Uint8Array>,
+  options?: WriteFileOptions,
+): Promise<void> {
+  if (isDeno) {
+    return Deno.writeFile(path, data, { ...options });
```
eryue0220, line 38: I think here is no need return and just use `await`.

fs/unstable_write_file.ts
```diff
@@ -0,0 +1,132 @@
...
+ * @param data The content in bytes to be written.
+ * @param options Options to write files.
+ */
+export function writeFileSync(
+  path: string | URL,
+  data: Uint8Array,
+  options?: WriteFileOptions,
+): void {
+  if (isDeno) {
+    return Deno.writeFileSync(path, data, { ...options });
```
eryue0220, line 105: It's no need to`return`.

fs/unstable_write_file.ts
jbronder: Thank you! Appreciate it.

fs/unstable_write_file.ts
jbronder: Thank you!

fs/unstable_write_file.ts
```diff
@@ -0,0 +1,132 @@
...
+      append,
+      create,
+      createNew,
+      truncate,
+      write: true,
+    });
+    try {
+      if (data instanceof ReadableStream) {
+        const { Readable } = getNodeStream();
+        await getNodeFs().promises.writeFile(path, Readable.fromWeb(data), {
```
kt3k, line 59: I don't think this `Readable.fromWeb` call is necessary as `fs.promises.writeFile` supports `AsyncIterable` input and ReadableStream in Node.js should implement `AsyncIterable`

fs/unstable_write_file.ts
```diff
@@ -0,0 +1,132 @@
...
+      append,
+      create,
+      createNew,
+      truncate,
+      write: true,
+    });
+    try {
+      getNodeFs().writeFileSync(path, data, { flag, signal });
+      if (mode != null) {
+        getNodeFs().chmodSync(path, mode);
```
kt3k, line 126: Ditto. Do we need `chmodSync` call here?

fs/_get_fs_flag.ts
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+// getAccessFlag and getCreationFlag adapted from the original in Rust's std::fs
+// <source path="https://github.com/rust-lang/rust/blob/304441960e7058fe97f09ef00b20739b4dc56d11/library/std/src/sys/unix/fs.rs#L694-L728" />
+
+import { BadResource } from "./unstable_errors.js";
+import { platform } from "node:os";
+import { constants } from "node:fs";
```
kt3k, line 8: This file is included in non-test code, and these imports cause `node:` dependencies included even in Deno programs, which is not a desirable situation. Let's avoid importing these here.

To replace constants, I'd suggest using hard-coded values. To replace `os.platform`, let's use `process.platform === 'win32'` check, which is available as a global variable.

fs/_get_fs_flag.ts
```diff
@@ -0,0 +1,76 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+// getAccessFlag and getCreationFlag adapted from the original in Rust's std::fs
+// <source path="https://github.com/rust-lang/rust/blob/304441960e7058fe97f09ef00b20739b4dc56d11/library/std/src/sys/unix/fs.rs#L694-L728" />
```
kt3k, line 4: I don't think we should borrow these code as they include many unrelated code. Also throwing BadResource for some combinations doesn't seem desirable as Deno API don't behave in that way.

I think `getFsFlags` should be somethig like:

```js
if (options.create) {
  flags |= O_CREAT;
}
if (options.createNew) {
  flags |= O_EXCL;
}
if (options.append) {
  flags |= O_APPEND;
} else if (options.create) {
  flags |= O_TRUNC;
}
return flags;
```

What do you think?

fs/unstable_write_file.ts
jbronder: I'll get this updated.  That should also get rid of the `getNodeStream` function as well.

fs/unstable_write_file.ts
jbronder: I believe the same reasoning stated in lines 64-67 would also apply to the synchronous call.

fs/_get_fs_flag.ts
jbronder: Sound good. I'll update this.

fs/_get_fs_flag.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { BadResource } from "./unstable_errors.js";
+import { platform } from "node:os";
+import { constants } from "node:fs";
+
+const { O_APPEND, O_CREAT, O_EXCL, O_TRUNC, O_RDONLY, O_WRONLY, O_RDWR } =
+  constants;
+
+// deno-lint-ignore ban-types
+type Optional<T extends {}> = { [K in keyof T]?: T[K] };
```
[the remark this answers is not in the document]
jbronder: After looking at Partial. Agreed.

fs/_get_fs_flag.ts
jbronder: The intention of using this implementation of `getFsFlag` was to also apply the same function for the `open(Sync)` APIs when working with `OpenOptions`. The error handling should be consistent with the Node and Deno runtimes and I can understand why this implementation shouldn't be used. 

I can implement your suggestion as part of the flag function specifically for `writeFile(Sync)` and `writeTextFile(Sync)`. In doing so, it might make sense to have two different `getFsFlag` functions one for `writeFile`-specific operations and one for `open`. Something like `getWriteFsFlag` and `getOpenFsFlag` may help simplify the implementations rather than combining into one `getFsFlag` function.

kt3k: > In doing so, it might make sense to have two different getFsFlag functions one for writeFile-specific operations and one for open. Something like getWriteFsFlag and getOpenFsFlag may help simplify the implementations rather than combining into one getFsFlag function.

That sounds reasonable to me

fs/unstable_write_file.ts
```diff
@@ -0,0 +1,132 @@
...
+        const { Readable } = getNodeStream();
+        await getNodeFs().promises.writeFile(path, Readable.fromWeb(data), {
+          flag,
+          signal,
+        });
+      } else {
+        await getNodeFs().promises.writeFile(path, data, { flag, signal });
+      }
+      if (mode != null) {
+        await getNodeFs().promises.chmod(path, mode);
```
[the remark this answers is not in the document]
kt3k: Oh, good findings! Then the current code looks completely reasonable 👍

fs/_get_fs_flag.ts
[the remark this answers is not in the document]
jbronder: I decided to build the type `WriteBooleanOptions` using the `Pick<T, Keys>` utility type instead of a `Partial`.

fs/_get_fs_flag.ts
jbronder: I've left the name of the file as `_get_fs_flag.ts` to group the `getOpenFsFlag` (to be implemented) and `getWriteFsFlag` functions together.

fs/_get_fs_flag.ts
```diff
@@ -0,0 +1,33 @@
...
+>;
+
+/**
+ * Uses the boolean options specified in {@linkcode WriteFileOptions} to
+ * construct the composite flag value to pass to the `flag` option in the
+ * Node.js `writeFile` function.
+ */
+export function getWriteFsFlag(opt: WriteBooleanOptions): number {
+  const { O_APPEND, O_CREAT, O_EXCL, O_TRUNC, O_WRONLY } =
+    getNodeFs().constants;
```
kt3k, line 18: 👍

kt3k (approved):
LGTM. Nice!

PR #6445, opened 2025-02-26
test(fs/unstable): remove windows specific paths and fix ci

kt3k:
These windows specific code paths started failing in canary


kt3k:
Looks like this was fixed without any change on main..

(Could be an issue of VM setup change in gh-actions?)

Created an issue https://github.com/denoland/std/issues/6447

PR #6453, opened 2025-02-27
feat(encoding/unstable): add options argument to hex streaming & performance

BlackAsLight:
Jumping from: https://github.com/denoland/std/pull/6442

Pulling the unstable hex content out into a separate pull request


encoding/unstable_hex_stream.ts
```diff
@@ -2,96 +2,151 @@
...
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
  * @module
  */
 
-import { decodeHex, encodeHex } from "./hex.ts";
 import type { Uint8Array_ } from "./_types.ts";
 export type { Uint8Array_ };
+import {
+  calcMax,
```
kt3k, line 31: I missed this in #6480, but I don't think `calcMax` should be a part of public API. What do you think?

encoding/unstable_hex_stream.ts
```diff
@@ -2,96 +2,151 @@
...
+  /**
+   * Constructs a new instance.
+   *
+   * @param options The options for the hexadecimal stream.
+   */
+  constructor(options: { output?: T } = {}) {
+    const decode = function (): (input: Uint8Array_) => Expect<T> {
+      if (options.output === "bytes") return (x) => x as Expect<T>;
+      const decoder = new TextDecoder();
+      return (x) => decoder.decode(x) as Expect<T>;
```
kt3k, line 77: We can save some memory allocation by moving `const decoder = new TextDecoder();` outside of `decode` arrow function.

Also I think we need to pass `stream: true` option to `decoder.decode` call to allow partial data passed to it

encoding/unstable_hex_stream.ts
BlackAsLight: It's a utility function for the raw encode functions. Without it people would need to be familiar with the ratio between the different bases.

The name though could be better. Essentially you'd use it to calculate the minimum size your buffer needs to be when passing it to encodeRawBase64

encoding/unstable_hex_stream.ts
BlackAsLight: With moving decoder out to the top level scope, I think it would be better done in a separate pull request since this is the last one of the lot for streaming. 

With the stream option, we're only sending it acsii bytes so there won't be any partial data.

kt3k (approved):
LGTM

PR #6457, opened 2025-03-03
feat(encoding/unstable): add format option to encodeBase64 and decodeBase64

BlackAsLight:
Related: https://github.com/denoland/std/pull/6451

1. Introduces an unstable version of `encodeBase64` and `decodeBase64` that merges both Base64 and Base64Url into one with the user providing a string argument to specify which encoding/decoding they want.
3. Introduces `encodeRawBase64` and `decodeRawBase64` which is a more low level version of the previous functions meant to enable users to encode/decode part of a buffer in place.


encoding/unstable_base64.ts
```diff
@@ -0,0 +1,241 @@
...
+ * ) as Uint8Array_;
+ *
+ * const i = buffer.indexOf(",".charCodeAt(0)) + 1;
+ * const o = decodeRawBase64(buffer, i, i, "Base64");
+ *
+ * buffer = buffer.subarray(i, o);
+ * assertEquals(buffer, await Deno.readFile("./deno.lock"));
+ * ```
+ */
+export function decodeRawBase64(
```
kt3k, line 229: When do you think the users would prefer this API over `decodeBase64`?

encoding/unstable_base64.ts
```diff
@@ -0,0 +1,241 @@
...
+ * assertEquals(encodeBase64("Hello World", "Base64Url"), "SGVsbG8gV29ybGQ");
+ * assertEquals(
+ *   encodeBase64(new TextEncoder().encode("Hello World") as Uint8Array_, "Base64Url"),
+ *   "SGVsbG8gV29ybGQ",
+ * );
+ * ```
+ */
+export function encodeBase64(
+  input: string | Uint8Array_ | ArrayBuffer,
+  format: Base64Format,
```
kt3k, line 84: Could this be defaulted to `"Base64"`?

encoding/unstable_base64.ts
BlackAsLight: The only real use case for this API and the `encodeRawBase64` is if you're input and output is a `Uint8Array`. Say you're creating a file and part of it needs to be base64 for whatever consumes it, this API lets you skip the conversion to a string and back and in the same instance use less memory. 

Outside of where the encoding is embedded in other information, I doubt it would find much use.

encoding/unstable_base64.ts
BlackAsLight: It can

kt3k (approved):
LGTM

kt3k:
Note: This conflicts a little bit with the current convention of how to expose the variations of the same algorithm (e.g. `encodeBase64` and `encodeBase64Url`). But this can be a better alternative way to expose such variations as we now have more and more such variations (e.g. base32/base32hex/base32crockford), and some API names look a little strange ( `encodeBase32Crockford` for example)

PR #6458, opened 2025-03-03
feat(cbor): add support for bignums

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6455

This pull request adds native support for encoding bigints that requires +64 bits to encode via the tag item 2 and 3. It also adds automatic decoding of tag items 2 and 3 for converting them back into bigints instead of returning the `CborTag` instance.

- [Cbor Tags](https://www.iana.org/assignments/cbor-tags/cbor-tags.xhtml)
- [RFC8949 BigNums](https://www.rfc-editor.org/rfc/rfc8949.html#name-bignums)


cbor/encode_cbor_test.ts
```diff
@@ -246,6 +246,22 @@ Deno.test("encodeCbor() encoding Dates", () => {
...
 
+Deno.test("encodeCbor() encoding bignums as Uint byte string", () => {
+  const num = 2n ** 64n;
+  assertEquals(
+    encodeCbor(num),
+    new Uint8Array([0b110_00010, 0b010_01001, 1, 0, 0, 0, 0, 0, 0, 0, 0]),
+  );
+});
+
+Deno.test("encodeCbor() rejecting bigints as Int byte string", () => {
```
kt3k, line 257: Looks like this needs to be updated?

kt3k:
Mostly looks good. Left one question

kt3k (approved):
LGTM

PR #6459, opened 2025-03-04
fix(cbor): bug in preallocating space for string encoding

BlackAsLight:
This pull request fixes a bug where the calculation for calculating how much space a string needs to convert into a `Uint8Array` was wrong. Previously it was merely calling the `.length` property on it, but this is only sufficient if the UTF16 encoding and UTF8 encoding have the same length for a given string. This is so when it comes to English text, but using other texts, ones where the UTF8 encoding is shorter will result in left over space in the buffer returned from `encodeCbor` and ones where the UTF8 encoding is longer will result in an out of bounds error.

This pull request fixes the bug by preallocating 3x the size of the string. Which according to this issue, https://github.com/whatwg/encoding/issues/333, is the max a given string could encode into. A subarray at the end is then returned to cut off any excess space leftover in the buffer.


kt3k:
Sounds a reasonable fix to me. Could you add test cases to confirm the fix?

cbor/encode_cbor_test.ts
```diff
@@ -188,6 +188,16 @@ Deno.test("encodeCbor() encoding strings", () => {
   // Can't test the next bracket up due to JavaScript limitations.
 });
 
+Deno.test("encodeCbor() correctly preallocates enough space for strings", () => {
+  const input = "\uD83D\uDCA9";
```
kt3k, line 192: code smell

kt3k (approved):
LGTM. Thanks!

PR #6461, opened 2025-03-04
perf(encoding): improve base64 encode/decode performance

BlackAsLight:
Closes: https://github.com/denoland/std/issues/5944

This pull request brings the performance added in https://github.com/denoland/std/pull/6457 to the stable base64 and base64url functions, but in a non-breaking manner. The only difference a user would find is a difference in error message if providing an invalid encoded string.

### DIFF
--- a/encoding/base64.ts
+++ b/encoding/base64.ts
@@ -22,76 +22,16 @@
  * @module
  */
 
-import { validateBinaryLike } from "./_validate_binary_like.ts";
+import { calcMax, decode, encode } from "./_common64.ts";
+import { detach } from "./_common_detach.ts";
 import type { Uint8Array_ } from "./_types.ts";
 export type { Uint8Array_ };
 
-const base64abc = [
-  "A",
-  "B",
-  "C",
-  "D",
-  "E",
-  "F",
-  "G",
-  "H",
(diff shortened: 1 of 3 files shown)


encoding/base64.ts
```diff
@@ -22,76 +22,16 @@
...
-  "8",
-  "9",
-  "+",
-  "/",
-];
+const padding = "=".charCodeAt(0);
+const alphabet = new TextEncoder()
+  .encode("ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/");
+const rAlphabet = new Uint8Array(128).fill(64); // alphabet.length
+alphabet.forEach((byte, i) => rAlphabet[byte] = i);
```
iuioiua, line 34: This looks like what I create [`BidirectionalMap`](https://jsr.io/@std/data-structures/doc/unstable-bidirectional-map/~/BidirectionalMap) for.

BlackAsLight: I originally had a Map for the rAlphabet variable but found it noticeably slower to index than a Uint8Array

kt3k (approved):
LGTM

PR #6465, opened 2025-03-06
refactor(front_matter): inline tests

timreichen:
In preparation to investigate https://github.com/denoland/std/issues/6464

This PR removes test data abstractions and inlines them. This makes the code more verbose but makes the tests more readable and understandable.


front_matter/any.ts
```diff
@@ -34,7 +34,9 @@ export type { Extract };
  * @param text The text to extract front matter from.
  * @returns The extracted front matter and body content.
  */
-export function extract<T>(text: string): Extract<T> {
+export function extract<T extends Record<string, unknown>>(
+  text: string,
+): Extract<T> {
```
kt3k, line 39: Can we keep the existing type? `T` can be, for example, boolean. This change breaks that usage:

```js
import { extract } from "@std/front-matter/any";

console.log(extract<boolean>(`---
true
---

aaa
`));
```

front_matter/json.ts
```diff
@@ -32,7 +32,9 @@ export type { Extract };
  * @param text The text to extract JSON front matter from.
  * @returns The extracted JSON front matter and body content.
  */
-export function extract<T>(text: string): Extract<T> {
+export function extract<T extends Record<string, unknown>>(
+  text: string,
+): Extract<T> {
```
kt3k, line 37: Similarly to any.ts, can we keep the existing type? `T` can be non Record type, and this change breaks such case:

```ts
import { extract } from "@std/front-matter/json";

console.log(extract<boolean>(`---json
true
---

aaa
`));
```

kt3k (approved):
Now the test cases look easier to follow to me. Thanks for the clean ups! LGTM

PR #6471, opened 2025-03-09
BREAKING(encoding/unstable): merge base32 variations, add format option to encodeBase32 and decodeBase32

BlackAsLight:
Related: https://github.com/denoland/std/pull/6452

1. Rewrites the unstable versions of `Base32`, `Base32Crockford`, and `Base32Hex` into one encode one decode function, providing a string argument to specify which encoding/decoding the user wants.
3. Introduces `encodeRawBase32` and `decodeRawBase32` which is a more low level version of the previous functions meant to enable users to encode/decode part of a buffer in place.


kt3k (approved):
LGTM

PR #6473, opened 2025-03-11
test(toml): inline test data

timreichen:
This PR inlines test data. This makes the tests easier to understand and also prevents dead files (like `toml/testdata/cargoTest.toml`).


iuioiua:
Niiice

kt3k (approved):
LGTM

PR #6477, opened 2025-03-13
fix(encoding): decoding base64 with invalid bytes >= 128

BlackAsLight:
This pull request fixes a simple bug where when decoding base64 the functions wouldn't throw when encountering an invalid byte between 128 - 256. Since the `rAlphabet` varaible only goes up to 128, any bytes not matching this range would return undefined and then later be interpreted as a 0 byte when it was meant to throw an error.

### DIFF
--- a/encoding/_common64.ts
+++ b/encoding/_common64.ts
@@ -114,7 +114,7 @@ export function decode(
 }
 
 function getByte(char: number, alphabet: Uint8Array): number {
-  const byte = alphabet[char]!;
+  const byte = alphabet[char] ?? 64;
   if (byte === 64) { // alphabet.Base64.length
     throw new TypeError(`Invalid Character (${String.fromCharCode(char)})`);
   }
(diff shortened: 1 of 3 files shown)


kt3k:
Could you add test cases to check this fix?

encoding/_common64.ts
```diff
@@ -114,7 +114,7 @@ export function decode(
 }
 
 function getByte(char: number, alphabet: Uint8Array): number {
-  const byte = alphabet[char]!;
+  const byte = alphabet[char] ?? 64;
   if (byte === 64) { // alphabet.Base64.length
     throw new TypeError(`Invalid Character (${String.fromCharCode(char)})`);
```
kt3k, line 119: Not directly related to this change, but can we change this error message to something like:
```
Cannot decode base64 string as it includes invalid character: "${String.fromCharCode(char)}"
```

to align this to our style guide? https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#error-messages

kt3k (approved):
LGTM. Thanks

PR #6480, opened 2025-03-13
feat(encoding/unstable): add encode/decodeRawHex and rewrite underlying code

BlackAsLight:
Related: https://github.com/denoland/std/pull/6453

~1. Introduces an unstable version of `encodeHex` and `decodeHex` that takes an optional redundant format argument merely to match that of the unstable base64 and base32 which have a few variants.~
3. Introduces `encodeRawHex` and `decodeRawHex` which is a more low level version of the previous functions meant to enable users to encode/decode part of a buffer in place.


kt3k:
> Introduces an unstable version of encodeHex and decodeHex that takes an optional redundant format argument merely to match that of the unstable base64 and base32 which have a few variants.

I'd personally prefer not to have option if there's no option. I found this `format` option only confusing..

encoding/unstable_hex.ts
```diff
@@ -0,0 +1,199 @@
...
+ * import { encodeHex, type Uint8Array_ } from "@std/encoding/unstable-hex";
+ *
+ * assertEquals(encodeHex("Hello World"), "48656c6c6f20576f726c64");
+ * assertEquals(
+ *   encodeHex(new TextEncoder().encode("Hello World") as Uint8Array_),
+ *   "48656c6c6f20576f726c64",
+ * );
+ * ```
+ */
+export function encodeHex(
```
kt3k, line 61: Feels like there's no need of unstable version of this as this seems completely compatible with the existing version. What do you think?

BlackAsLight: There is a difference in that a Uint8Array or ArrayBuffer past in will detach the buffer to reuse making the user no longer have access to the original content.

kt3k: Ah, ok. That seems significant difference. Ok, let's land this as unstable

kt3k (approved):
LGTM

PR #6481, opened 2025-03-15
fix(front-matter): handle empty frontMatter data

timreichen:
fixes https://github.com/denoland/std/issues/6464


kt3k:
related https://github.com/denoland/std/pull/6393

front_matter/_formats.ts
```diff
@@ -40,15 +37,15 @@ export const RECOGNIZE_JSON_REGEXP = new RegExp(
 );
 
 export const EXTRACT_YAML_REGEXP = new RegExp(
-  `^${BOM}${YAML_HEADER}${WHITESPACES}${NEWLINE}${WHITESPACES}${FRONT_MATTER}${WHITESPACES}${NEWLINE}${YAML_FOOTER}${WHITESPACES}${NEWLINE}?${BODY}$`,
```
kt3k, line 43: I don't think we should remove the last NEWLINE. This now allow the document like:

```
---yaml
foo: bar
---body
```
which is parsed as `{ frontMatter: "foo: bar", body: "body", attrs: { foo: "bar" } }`. This seems unintentional to me

timreichen: Thanks for catching that. I added additional tests and updated the regexp.

kt3k (approved):
LGTM. Thanks!

PR #6483, opened 2025-03-17
feat(toml): make toml `parse()` generic 

timreichen:


kt3k:
I'm not in favor of this change as other similar parsers (yaml, jsonc, etc) are not generic. Also this doesn't seem making typing more correct.

timreichen:
> I'm not in favor of this change as other similar parsers (yaml, jsonc, etc) are not generic. Also this doesn't seem making typing more correct.

Ok, closing for now.

PR #6484, opened 2025-03-17
fix(assert): make prototype of Object.prototype not equal undefined prototype

### LINKED ISSUE #6334: `@std/assert/equals` considers `Object.create(null)` and `{}` to be equal
**Describe the bug**
- Since https://github.com/denoland/std/pull/6153, `assertEquals` checks object prototypes. However, it makes certain exceptions to keep existing tests within the repo from failing.
- Other assertion libraries have more correct checks: `"node:assert/strict"`, `"ava"`, and `"uvu"` all make a distinction between null-prototype object and plain objects. `"node:assert"` and `"chai"` don't.

**Steps to Reproduce**
The following assertion passes:
```ts
import { assertEquals } from "@std/assert"
assertEquals(Object.create(Object), Object.create(null))
 ```

**Expected behavior**
The assertion fails.

**Environment**
Not applicable.

### PR DESCRIPTION
**Pull request by WWRS**
Fixes #6334

These were previously hardcoded to be equal, and I'm not completely sure why. I hope I'm not missing something.

It's not mentioned in the PR when it was added (#6153). I suspect it was a port of the old system, where constructors were checked rather than prototypes: Undefined constructor could equal Object constructor. But undefined prototype should not equal Object prototype.

### DIFF
--- a/assert/equal_test.ts
+++ b/assert/equal_test.ts
@@ -383,6 +383,18 @@ Deno.test("equal() with constructor and prototype", async (t) => {
   });
 });
 
+Deno.test("equal() considers object with prototype Object.prototype not equal to object with no prototype", () => {
+  const a = Object.create(Object.prototype);
+  const b = Object.create(null);
+  assertFalse(equal(a, b));
+  assertFalse(equal(b, a));
+
+  // ensure the only difference is the prototype
+  Object.setPrototypeOf(b, Object.prototype);
+  assert(equal(a, b));
+  assert(equal(b, a));
+});
+
 Deno.test("equal() with dynamic properties defined on the prototype", async (t) => {
   await t.step("built-in web APIs", async (t) => {
     await t.step("URLPattern", () => {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: assert/equal_test.ts [hunk 1]
```diff
@@ -383,6 +383,18 @@ Deno.test("equal() with constructor and prototype", async (t) => {
   });
 });
 
+Deno.test("equal() considers object with prototype Object.prototype not equal to object with no prototype", () => {
```
WWRS, line 386: Is this too long a name? Suggestions on making it shorter?

kt3k: This looks just fine to me

iuioiua:
This PR hasn't had activity in almost 6 months. Should we close as stale for now?

WWRS:
We would need to disambiguate the behavior mentioned in #6334. Determining this behavior doesn't seem like a priority, so if this PR is not moving toward either being outright accepted or outright rejected, it's stale.

PR #6485, opened 2025-03-17
fix(assert,expect,internal,testing): improve handling of escaped chars in diff_str

### LINKED ISSUE #6326: `assertEquals` error diff separately tokenizes `\` of escaped chars in strings
**Describe the bug**

The error message of `assertEquals` separately tokenizes `\` of escaped characters in strings, e.g `\t` representing a tab literal is treated as two separate tokens despite semantically being a single indivisible unit.

**Steps to Reproduce**

1. Run the following in repl:
    ```ts
    import { assertEquals } from 'jsr:@std/assert@1.0.10'
    const first = 'a\tb'
    const second = 'a\fb'
    assertEquals(first, second)
    ```
2. Observe output, showing diff as `a\[-tb]{+fb}`
    image

**Expected behavior**

Diff to show as `a[-\t]{+\f}b`

**Environment**

- OS: win11/WSL/ubuntu 22
- deno version: 2.1.1
- std version: assert@1.0.10

WWRS:
- Fixes #6326
- Fixes `a\\tb` vs `a\tb`: previously these would be detected as `!==` and cause `assertEquals` to reject, but `diff_str` found no difference
- Combines escaped spaces: `"  "` (double space) is currently treated as one token. This change makes `"\t\f"` also be treated as one token instead of four
  - Adds one test and updates another because the space-combining was previously untested

### DIFF
--- a/internal/diff_str.ts
+++ b/internal/diff_str.ts
@@ -32,7 +32,7 @@ export function unescape(string: string): string {
     );
 }
 
-const WHITESPACE_SYMBOLS = /([^\S\r\n]+|[()[\]{}'"\r\n]|\b)/;
+const WHITESPACE_SYMBOLS = /([^\S\r\n]+|\\[bftvrn]|[()[\]{}'"\r\n]|\b)/;
 
 /**
  * Tokenizes a string into an array of tokens.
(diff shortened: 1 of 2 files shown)


kt3k (approved):
Looks important fix. Thanks for looking into this!

LGTM

PR #6487, opened 2025-03-17
chore: check `test.ts` files that have exports in `_tools/check_mod_exports.ts`

timreichen:
There are some `test.ts` files that are part of an api (for example `front_matter/test.ts`). This PR checks these files for exports to separate test files from api files.


kt3k (approved):
LGTM

PR #6491, opened 2025-03-18
chore: update doc linter, replace hardcoded entry list with dynamically created one

timreichen:
This PR removes the hardcoded entry list. Apparently some files were accidentally excluded by not updating the list.


_tools/utils.ts
```diff
@@ -34,3 +36,37 @@ export function resolveWorkspaceSpecifiers(
...
+    ) {
+      result = false;
+    } else {
+      ts.forEachChild(node, visitNode);
+    }
+  }
+
+  visitNode(sourceFile);
+  return result;
+}
```
iuioiua, line 72: This seems extraneous. What about this?
```ts
export function isTestFile(filePath: string): boolean {
  return filePath.endsWith("test.ts");
}
```

_tools/check_docs.ts
```diff
@@ -23,131 +23,38 @@ import {
   type Location,
   type TsTypeDef,
 } from "@deno/doc";
-import { pooledMap } from "@std/async/pool";
+import { walk } from "../fs/walk.ts";
+import { isTestFile } from "./utils.ts";
+import { greaterOrEqual, parse } from "@std/semver";
+import { join } from "../path/join.ts";
```
iuioiua, line 29: Not that it really matters. More for consistency.
```suggestion
import { walk } from "@std/fs/walk";
import { isTestFile } from "./utils.ts";
import { greaterOrEqual, parse } from "@std/semver";
import { join } from "@std/path/join";
```

_tools/check_docs.ts
```diff
@@ -23,131 +23,38 @@ import {
...
-  "../webgpu/mod.ts",
-  "../yaml/mod.ts",
-] as const;
+const STABLE_VERSION = parse("1.0.0");
+const ROOT = new URL("../", import.meta.url);
+const ENTRY_POINT_URLS = [];
+for await (
+  const { path } of walk(ROOT, { exts: [".json"], match: [/deno.json$/] })
+) {
+  if (isTestFile(path)) continue;
```
iuioiua, line 41: This isn't needed as the parameters of defined in the `walk` call exclude test files anyway.

_tools/check_docs.ts
```diff
@@ -23,131 +23,38 @@ import {
...
+  const { version, exports } = JSON.parse(await Deno.readTextFile(path));
```
iuioiua, line 42: Another option is to dynamically import using the `json` assertion type.

_tools/check_docs.ts
```diff
@@ -23,131 +23,38 @@ import {
...
+  if (!version || !greaterOrEqual(parse(version), STABLE_VERSION)) continue;
+  for (const relativeFilePath of Object.values<string>(exports)) {
+    if (!relativeFilePath.endsWith(".ts")) continue;
```
iuioiua, line 45: Is this needed? Are there any exports which export something besides a `.ts` file?

iuioiua (commented):
Nits. This could be a little over-engineered and contains other changes that don't pertain to the main objective of the PR.

_tools/utils.ts
timreichen: The whole point of this function is to distinguish between `test.ts` which is used for testing (calling `Deno.test()`) and `test.ts` which is part of an api (like in `@std/front-matter/test`.  This cannot be done by filename alone. I figured the best way was to check whether the file has exports or not, if so it must be part of an api.

_tools/check_docs.ts
timreichen: That might be a possibility.
Sidenote: `deno.jsonc` is also a valid with deno, so we might want to support `jsonc` to make this fool-proof in case some std module with use a `deno.jsonc` in the future.

_tools/check_docs.ts
timreichen: Yes, `@std/html` exports a `json`: `"./named-entity-list.json": "./named_entity_list.json"`

_tools/check_docs.ts
timreichen: only `*_test` files. `test.ts` is a special case, as it is used as both test file and api file in std.
Sadly this ambiguous behavior is baked into deno and we have to handle it as such here.

timreichen: You are right.

_tools/check_docs.ts
iuioiua: IMHO, it's a little cleaner. Also, I wouldn't worry about supporting something that isn't currently used. Because it's also possible that `deno.jsonc` files will never be used.

_tools/check_docs.ts
```diff
@@ -23,131 +23,37 @@ import {
...
-  "../yaml/mod.ts",
-] as const;
+const STABLE_VERSION = parse("1.0.0");
+const ROOT = new URL("../", import.meta.url);
+const ENTRY_POINT_URLS = [];
+for await (
+  const { path } of walk(ROOT, { exts: [".json"], match: [/deno.json$/] })
+) {
+  const { version, exports } = JSON.parse(await Deno.readTextFile(path));
+  if (!version || !greaterOrEqual(parse(version), STABLE_VERSION)) continue;
```
kt3k, line 41: I don't think we should skip unstable packages (many of unstable packages are already checked by this script e.g. cache, cbor, random, webgpu)

_tools/check_docs.ts
```diff
@@ -497,8 +403,8 @@ function assertConstructorDocs(
  * Checks a module document for:
  * - Code snippets that execute successfully.
  */
-function assertModuleDoc(document: DocNodeWithJsDoc<DocNodeModuleDoc>) {
-  assertSnippetsWork(document.jsDoc.doc!, document);
+function assertModuleDoc(_document: DocNodeWithJsDoc<DocNodeModuleDoc>) {
+  // assertSnippetsWork(document.jsDoc.doc!, document);
```
kt3k, line 407: Can this be enabled?

_tools/check_docs.ts
```diff
@@ -586,16 +490,10 @@ if (!lintStatus.success) {
...
-
-const iter = pooledMap(
-  navigator.hardwareConcurrency,
-  snippetPromises,
-  (fn) => fn(),
-);
-for await (const _ of iter) {
-  // noop
+for (const url of ENTRY_POINT_URLS) {
+  await checkDocs(url);
```
kt3k, line 494: I think we should keep the current way of checking docs concurrently. This checks items one by one

_tools/utils.ts
kt3k: Maybe `Deno.readTextFileSync(filePath).includes("\nDeno.test")` would suffice? (leading `\n` ensures it's at the beginning of the line)

timreichen: This could lead to unintentional behaviour if `Deno.test` is mentioned in a comment or used in an API. So special treatment would be required again for at least `@std/testing`.

timreichen: This could lead to unintentional behaviour if `Deno.test` was wrapped in a function and therefore not at the beginning of the line, for example (`jsonc/testdata/JSONTestSuite/test.ts`)
Also leads to unintentional behaviour when mentioned in a comment with unlucky line breaks:
```
/* bla bla bla...
Deno.test bla bla bla...
bla bla bla... */
```
So while this might work sometimes, it is not very robust.

_tools/check_docs.ts
timreichen: This was added mainly because `@std/log` is not stable and violates lots of the lint checks.
I wanted to avoid to special treat separate mods, so think it would require a lots of changes in `log` which might get removed from `@std` anyway?

_tools/check_docs.ts
timreichen: Ok for now. But this might unintentionally break things in the future or files might fall outside without throwing, so I think we should generally cover all possible use cases especially for such checks.

_tools/check_docs.ts
timreichen: I disabled that for now, because there are many mods that do not provide an example where the `@module` is declared. These are:
```sh
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/crypto/crypto.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/encoding/base64url.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/cli/parse_args.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/http/etag.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/http/file_server.ts:7
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/http/negotiation.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/http/status.ts:4
[error] @example tag must have a TypeScript code snippet at file:///Users/timreichen/Sites/std/http/user_agent.ts:8
8 errors found
```
I am not sure how to handle that. Does the module need examples in addition to examples for exports?

_tools/check_docs.ts
timreichen: I found the way this was written not very readable and I did some testing, seems like there is little to no impact: 
```sh
file:///std/_tools/check_docs.ts

benchmark   time/iter (avg)        iter/s      (min … max)           p75      p99     p995
----------- ----------------------------- --------------------- --------------------------
pool                  2.1 s           0.5 (   2.1 s …    2.1 s)    2.1 s    2.1 s    2.1 s
loop                  2.1 s           0.5 (   2.1 s …    2.2 s)    2.2 s    2.2 s    2.2 s
```

_tools/check_docs.ts
kt3k: I think we should exclude `@std/log` then.

If we land this change, we are going to miss the docs for unstable package in the future PRs.

_tools/check_docs.ts
kt3k: Ok. It seems fine as is

_tools/check_docs.ts
kt3k: I believe this check is for showing the example in the top page of the package in JSR website (e.g. https://jsr.io/@std/encoding ), and those errored files are not for the top level of the package. I suggest we should do something like:
```suggestion
  if (document.location.filename.endsWith("/mod.ts")) {
    assertSnippetsWork(document.jsDoc.doc!, document);
  }
```

_tools/check_docs.ts
kt3k: Only 4 files in `@std/log` seem having the issues. Adding the check like the below at line 46 makes it working:

```ts
    if (
      filePath.endsWith("log/mod.ts") ||
      filePath.endsWith("log/levels.ts") ||
      filePath.endsWith("log/setup.ts") ||
      filePath.endsWith("log/rotating_file_handler.ts")
    ) {
      continue;
    }
```

_tools/check_docs.ts
timreichen: Done

kt3k (approved):
LGTM

PR #6492, opened 2025-03-18
chore: update tools deps

timreichen:


kt3k (approved):
LGTM. Thanks!

I also updated the way to get docs from the deno-doc api.

PR #6494, opened 2025-03-19
chore: remove `_tools/packages.ts`

timreichen:
Seems like this file isn't used anywhere.


kt3k (approved):
LGTM

kt3k:
Note: This script was a part of migration script from deno.land/std to JSR https://github.com/denoland/std/commit/6dad7609c3b77fe46f46af6091f2404ab6120288

Issue #6490, opened 2025-03-18
std/toml returns hexadecimal integers as `string` not `number`


andy-preston:
**Describe the bug**

The TOML Parser in the standard library parses hexadecimal integers as `string` - it should return all integers as `number`

**Steps to Reproduce**

```typescript
import { assertEquals, assertNotEquals } from "assert";
import { parse } from "jsr:@std/toml";

Deno.test("hexadecimal values are returned as strings", () => {
    const expected = {
        hex1: "0xDEADBEEF",
        hex2: "0xdeadbeef",
        hex3: "0xdead_beef",
    };
    const required = {
        hex1: 0xdeadbeef,
        hex2: 0xdeadbeef,
        hex3: 0xdeadbeef,
    }
    const actual = parse(`hex1 = 0xDEADBEEF
        hex2 = 0xdeadbeef
        hex3 = 0xdead_beef`);
    assertEquals(actual, expected);
    assertNotEquals(actual, required);
});
```
**Expected behavior**

All TOML integers should be returned as `number`

**Environment**

OS: Manjaro (18-March-2025)
Deno Version: 2.2.4
std/toml version: 1.0.2

andy-preston:
If this bug is agreed, I'm more than happy to fix it and do a pull-request. :)

kt3k:
Looks a bug to me. PRs are welcome

re-masashi:
is anyone willing to merge the PR?

kt3k:
fixed by #6496

PR #6496, opened 2025-03-19
fix(toml): handle hexadecimal, octal, and binary numbers

re-masashi:
Fixes the error in #6490. Also handles octal and binary.
The tests for these have also been updated.
Files changed: toml/_parser.ts and toml/parse_test.ts


toml/_parser.ts
```diff
@@ -537,25 +537,70 @@ export const dottedKey = join(or([bareKey, basicString, literalString]), ".");
...
+        break;
+      case "0o":
+        base = 8;
+        break;
+      case "0x":
+        base = 16;
+        break;
+      default:
+        return failure();
+    }
```
babiabeo, line 588: You can just determine the base in the switch above rather than making another one

re-masashi: that's feasible too, although this works as well

re-masashi: should i change the code to do it that way?

babiabeo: Yes, that can avoid duplicate code

re-masashi: alright, I'll be doing it as soon as I can.

kt3k (approved):
LGTM. Thanks!

PR #6497, opened 2025-03-19
refactor(cli/unstable): replace `ProgressBar` `#options` property with actual properties

timreichen:
In preparation for https://github.com/denoland/std/pull/6430


kt3k (approved):
LGTM

PR #6499, opened 2025-03-20
perf(encoding): improve hex encode/decode performance

BlackAsLight:
This pull request brings the performance added in https://github.com/denoland/std/pull/6480 to the stable hex functions, but in a non-breaking manner. The only difference a user would find is a difference in error message if providing an invalid encoded string.


encoding/hex_test.ts
```diff
@@ -21,14 +21,14 @@ const testCases = [
...
   // encoded(hex) / error / msg
-  ["0", RangeError, ""],
-  ["zd4aa", TypeError, "'z'"],
-  ["d4aaz", TypeError, "'z'"],
-  ["30313", RangeError, ""],
-  ["0g", TypeError, "'g'"],
-  ["00gg", TypeError, "'g'"],
-  ["0\x01", TypeError, "'\x01'"],
-  ["ffeed", RangeError, ""],
+  ["0", TypeError, "Invalid Character (0)"],
```
kt3k, line 24: The existing error message feels better to me as it's more explicit about what's wrong. Can we change `decode` function in `_common16.ts` to be aligned with the existing behavior?

BlackAsLight: You mean for instances where the problem is the length and not the actual character? A better error message can be added for those instances for hex, base32 and base64, but a lot of tests would need to be changed which might exceed the scope of this pull request.

kt3k: > You mean for instances where the problem is the length and not the actual character?

Yes

> A better error message can be added for those instances for hex, base32 and base64, but a lot of tests would need to be changed which might exceed the scope of this pull request.

I'd suggest we would address that only for hex for now in this PR to keep the main branch publishable state.

BlackAsLight: The tests in unstable_hex_test.ts would also need to be updated, but I doubt that would be more than a handful.

kt3k (approved):
LGTM.

Note: This change includes the change of priority of errors when the decode input has 2 errors (invalid chars and invalid length). I believe this doesn't affect the users' code in a significant manner.

PR #6500, opened 2025-03-20
feat(text/unstable): add dedent

### LINKED ISSUE #6259: `dedent` function and tagged template function

A `dedent` function is a common need when working with tagged template strings, as they enhance the legibility of strings relative to other code (without which indentations become chaotic).

A `dedent.ts` function in the `text` directory. This file can export an overloaded version of `dedent` such that it works with both string arguments and template strings arrays + quasis.

While there are other libraries for this, I'd prefer only to use the Deno Standard Library wherever possible. This seems like a good candidate for Deno Standard.

WWRS:
resolves #6259

### DIFF
--- a/text/dedent.ts
+++ b/text/dedent.ts
@@ -1,0 +1,68 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/**
+ * Removes indentation from multiline strings.
+ *
+ * - Removes leading newline
+ * - Removes trailing whitespace, including newlines
+ * - Replaces whitespace-only lines with empty lines
+ * - Finds the minimum indentation among remaining lines and removes that much indentation from all of them
+ *
+ * @param input The string or template to remove indentation from.
+ * @returns The string without indentation.
+ *
+ * @example Usage
+ * ```ts
+ * import { assertEquals } from "@std/assert";
+ *
+ * assertEquals(
+ *   dedent(`
+ *     {
+ *       msg: "Hello",
+ *     }
+ *   `),
+ *   `{\n  msg: "Hello",\n}`
+ * );
+ *
+ * assertEquals(
+ *   dedent`line 1
+ *          line 2`,
+ *   "line 1\nline 2"
+ * );
+ * ```
+ */
+export function dedent(input: TemplateStringsArray | string): string {
+  const inputString = typeof input === "string" ? input : input.raw.join("");
+  const ignoreFirstUnindented = !inputString.startsWith("\n");
+  const trimmedInput = inputString.replace(/^\n/, "").trimEnd();
+  const lines = trimmedInput.split("\n");
+
+  let minIndentWidth: number | undefined = undefined;
+  for (let i = 0; i < lines.length; i++) {
+    const indentMatch = lines[i]!.match(/^(\s*)\S/);
+
+    // Skip empty lines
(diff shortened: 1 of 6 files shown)


kt3k:
We have policy to accept any new feature to std as 'unstable' feature first https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100

Can you move the file to `unstable_dedent.ts` and export it as `./unstable-dedent`?

text/dedent_test.ts
```diff
@@ -0,0 +1,71 @@
...
+    `{\n  msg: "Hello",\n}`,
+  );
+});
+
+Deno.test("dedent() handles example 2", () => {
+  assertEquals(
+    dedent`line 1
+           line 2`,
+    "line 1\nline 2",
+  );
```
kt3k, line 21: Is this behavior aligned with proposal-string-dedent? https://github.com/tc39/proposal-string-dedent

text/dedent.ts
```diff
@@ -0,0 +1,99 @@
...
+ *   dedent(`
+ *     {
+ *       msg: "Hello",
+ *     }
+ *   `),
+ *   `{\n  msg: "Hello",\n}`
+ * );
+ * ```
+ */
+export function dedent(input: string): string;
```
kt3k, line 30: Not sure we should have this overload. Does proposal-string-dedent allow this usage? https://github.com/tc39/proposal-string-dedent

text/dedent_test.ts
WWRS: This is aligned with `npm:dedent` (https://www.npmjs.com/package/dedent). In general, I'm following the behavior specified in proposal-string-dedent, but it specifies that an opening newline is required without specifying what happens for strings without opening newlines. We have a couple options for that case:

- If the first line has no indentation, assume it's meant to have the minimum indentation (current)
- Always treat the first line as a normal line, so `dedent` doesn't remove indentation in this test case
- Error if the string doesn't have an opening newline

I think the first option is most likely to match the user's intent. Since proposal-string-dedent is still subject to revision, we don't know what behavior will be chosen (if the proposal is accepted).

text/dedent.ts
WWRS: The only usage in proposal-string-dedent is as a tag template function, but because the proposal converts the template to a string, I think it makes sense to have this overload. This overload is also supported by `npm:dedent` (https://www.npmjs.com/package/dedent).

If we do not add this overload, it will still be possible to use `dedent` on a string, though I think this is less intuitive than the overload:
```ts
const s = "...";
dedent`${s}`
```

WWRS:
Thanks for the review. The two features noted are aligned with `npm:dedent` but not with `proposal-string-dedent` so on both features, I don't have a strong preference on keeping them as they are now or changing them to match `proposal-string-dedent`. Thoughts?

re-masashi (approved):

text/dedent.ts
kt3k: Ok. The overload makes sense to me now

text/dedent_test.ts
kt3k: Thanks for the clarification. Now I think the current form is most inclusive and good. (It seems aligning it to proposal-string-dedent only excludes some usages..)

kt3k (approved):
LGTM. Nice!

PR #6502, opened 2025-03-21
BREAKING(cli/unstable): remove trailing whitespaces from `ProgressBarFormatter` properties

timreichen:


kt3k:
Can you rebase? I think this one can be landed independent of other breaking changes (listed in https://github.com/denoland/std/issues/6501 ) as this should have less effect to the user's code (I guess there should be very small usage of `fmt` option for now)

PR #6504, opened 2025-03-22
refactor(encoding): align error messages

BlackAsLight:
Jumping from this conversation: https://github.com/denoland/std/pull/6499#discussion_r2006977391

This pull request updates the error message when the input has an invalid length to properly decode

### DIFF
--- a/encoding/_common16.ts
+++ b/encoding/_common16.ts
@@ -44,7 +44,9 @@ export function decode(
 ): number {
   if ((buffer.length - o) % 2 === 1) {
     throw new TypeError(
-      `Invalid Character (${String.fromCharCode(buffer[buffer.length - 1]!)})`,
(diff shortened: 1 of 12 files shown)


kt3k:
Thanks for the PR. The error messages look significantly better now to me, but can you also start them with something like `Cannot decode the input as Base64/Base32 ...` to follow the bullet point 4 of our error message style guide? (https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#error-messages) (I believe most error messages in std follow these patterns. ref. https://github.com/denoland/std/pulls?q=is%3Apr+author%3Airbull+is%3Aclosed+align+the+error+messages

kt3k:
> Should the hex error messages also start with this?

That's also preferable in my view

kt3k (approved):
LGTM. Thanks for updating!

PR #6507, opened 2025-03-24
fix(front-matter): handle BOM when recognizing format

timreichen:
This PR fixes a bug where trying recognize a format throws if the data starts with bom.


kt3k (approved):
LGTM

PR #6508, opened 2025-03-24
feat(encoding/unstable): add Uint8Array support to decodeBase64, 32, and hex

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6489

This pull request does two things:
~~1. Removes support for passing `ArrayBuffer` in `encodeBase64`, `encodeBase32`, and `encodeHex` in the unstable modules. I don't know the reasoning behind this being a part of the stable versions, but I also don't see the value in it since anyone who has an `ArrayBuffer` can simply wrap it in a `Uint8Array` before passing it in.~~
2. Adds support for passing in `Uint8Array` in `decodeBase64`, `decodeBase32`, and `decodeHex` in the unstable modules.

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -7,18 +7,18 @@
 export interface ProgressBarFormatter {
   /**
    * A function that returns a formatted version of the duration.
-   * `[mm:ss] `
+   * `[mm:ss]`
    */
   styledTime: () => string;
   /**
    * A function that returns a formatted version of the data received.
-   * `[0.40/97.66 KiB] `
+   * `[0.40/97.66 KiB]`
    * @param fractions The number of decimal places the values should have.
    */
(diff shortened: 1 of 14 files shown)


kt3k:
> Removes support for passing ArrayBuffer in encodeBase64, encodeBase32, and encodeHex in the unstable modules. I don't know the reasoning behind this being a part of the stable versions, but I also don't see the value in it since anyone who has an ArrayBuffer can simply wrap it in a Uint8Array before passing it in.

I'm not in favor of this change. Supporting ArrayBuffer doesn't add much complexity. I think it's handy to have it. Some Web platform APIs (such as `crypto.subtle.digest`) return `ArrayBuffer` instead of `TypedArray`. It works nicely with those APIs
```js
encodeBase64(await crypto.subtle.digest(...));
```

kt3k (approved):
LGTM!

PR #6511, opened 2025-03-25
refactor(ini): remove `IniMap.prototype.#cleanFormatting()`

timreichen:
Seems like this is not needed as `IniMap.prototype.#formatting` properties are not dynamically used.


kt3k (approved):
LGTM

PR #6512, opened 2025-03-25
refactor(ini): simplify `parse()`

timreichen:
This PR moves all parser functionality from `IniMap` into the `parse()` function.
There is a load of unnecessary overhead generated the way `parse()` works in `IniMap`.


ini/parse.ts
```diff
@@ -1,9 +1,64 @@
...
+        yield line;
+        line = "";
+        break;
+      case "\r":
+        yield line;
+        line = "";
+        if (text[i + 1] === "\n") i += 1;
+        break;
+      default:
+        if (line === " " || line === "\t") continue;
```
kt3k, line 54: What does this condition try to do?

Now `parse(" value=true")` throws, but it returns `{ value: true }` on main.

ini/parse.ts
```diff
@@ -81,7 +142,70 @@ export interface ParseOptions {
...
+    if (isComment(line)) continue;
+
+    if (isSection(line, lineNumber)) {
+      sectionName = SECTION_REGEXP.exec(line)?.groups?.name;
+
+      if (!sectionName) {
+        throw new SyntaxError(
+          `Unexpected empty section name at line ${lineNumber}`,
+        );
+      }
```
kt3k, line 175: This seems allowing empty section name now. `parse("[ ]\na=1")` currently throws, but now it returns `{ " ": { a: 1 } }`

ini/parse.ts
```diff
@@ -1,9 +1,63 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
 
-import { IniMap, type ReviverFunction } from "./_ini_map.ts";
+import type { ReviverFunction } from "./_ini_map.ts";
 export type { ReviverFunction };
 
+const SECTION_REGEXP = /^\[(?<name>.+)]$/;
+const KEY_VALUE_REGEXP = /^\s*(?<key>.*?)\s*=\s*(?<value>.*?)\s*$/;
```
kt3k, line 8: The leading and trailing whitespaces seem redundant to me as we use this regexp to trimmed lines

ini/parse.ts
timreichen: I fixed the regexp to allow `\S+` as section name instead of `.+`.

ini/parse.ts
timreichen: You are right, removed for now. I think I will try to optimize this in the spirit of https://github.com/denoland/std/pull/5751 at a later point.

ini/parse.ts
```diff
@@ -1,9 +1,63 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
 
-import { IniMap, type ReviverFunction } from "./_ini_map.ts";
+import type { ReviverFunction } from "./_ini_map.ts";
 export type { ReviverFunction };
 
+const SECTION_REGEXP = /^\[(?<name>\S.*\S|\S+)]$/;
```
kt3k, line 7: Now this doesn't match with `" a "`, for example. `parse("[ a ]\na=1")` returns `{ " a ": { a: 1 } }` on main. It throws in this branch.

I'm not sure the existing check can be done as a single regexp check.

kt3k (approved):
LGTM

PR #6513, opened 2025-03-25
BREAKING(encoding/unstable): Replace `encodeRaw` with `encodeInto` & Remove `decodeRaw`

BlackAsLight:
This pull request makes three breaking changes:

1. Rename `calcMax` to `calcBase64Size`.
   - In my opinion this is a better name than calcMax as it isn't even calculating the max size, but more the minimum required for the previous function to work.
1. Replace `encodeRawBase64` with `encodeBase64Into`.
   - The `encodeRawBase64` was in my opinion too complicated for people to use and had very narrow usage requiring the user to first structure their code in very specific way before passing it to this function. This replacement function now does a lot of that heavy lifting for the user. Giving them a much neater and nicer API to deal with. To accomplish these changes two other unrelated bits of code had to also change:
      1. Several constant variables in `unstable_base64.ts` were moved to `_common64.ts` with the user facing ones being reexported at `unstable_base64.ts`.
      2. `Base64EncoderStream` and `Base64DecoderStream` were made to rely directly on `_common64.ts`, removing it's dependency on `unstable_base64.ts` and in turn, `encodeRawBase64` and `decodeRawBase64`.
1. Remove `decodeRawBase64`.
   - After further consideration, and with the changes made with the previous point, I didn't see a point in this function existing. A `decodeBase64Into` also wouldn't be possible given the current structure how the underlying code works.

These same changes were also applied to base32 and hex.


kt3k:
I'm in favor of the idea. `encodeInto` looks much easier to use than `encodeRaw`.

Some suggestions about the names. How about `calcSizeBase64` and `encodeIntoBase64` instead of `calcBase64Size` and `encodeBase64Into`? The encoding name in the middle feels a bit strange to me. What do you think?

BlackAsLight:
Those names do sound better

timreichen:
I am not a big fan of `calc` in the names since it is never used in any js web API, only css. If we rename them, maybe we could do something similar to `getBoundingClientRect()` or `getComputedStyle()`, for example `getSizeOfBase64()`.

BlackAsLight:
> I am not a big fan of `calc` in the names since it is never used in any js web API, only css.

I don't think that is a far comparison since what web APIs are there that offer similar functionality?

For example, if you wanted to use TextEncoder.encodeInto, to be safe you need to know that the buffer you're encoding into needs to have at least 3x the length of the string you're encoding. I don't know of a function available to abstract this calculation for you. Instead you're required to know how UTF-16 translates to UTF-8.

kt3k (approved):
LGTM

PR #6515, opened 2025-03-25
refactor(ini): remove `IniMap`

timreichen:
Depends on https://github.com/denoland/std/pull/6514 and https://github.com/denoland/std/pull/6512


ini/parse_test.ts
iuioiua, line 1: Maybe these new tests should be in a separate PR so we know nothing's changed before and after this PR is merged.

kt3k (approved):
LGTM. Nice!

PR #6516, opened 2025-03-25
fix(datetime): fix formatting of `fractionalSecond`, fix parsing of `yy` and `yyyy`, test each part type in dateTimeFormatter.format

WWRS:
Previously, larger tests meant these appeared to be covered, but there was no testing of the deep specifications for each part type. Covering the cases revealed a bug with `fractionalSecond`, and the fix is included in this PR.

I also made `formatToParts` require exactly 2 digits when parsing `yy` and exactly 4 digits when parsing `yyyy` instead of 1-2 or 1-4 respectively. The spec requires `yy` to represent exactly 2 digits, but `yyyy` technically represents 4+ digits. In neither case are fewer digits allowed. I think it's unlikely anyone will be trying to parse years with 5+ digits, and parsing `yyyyMMdd` would require a more sophisticated approach if we allowed 5+, so I think only allowing 4 is best.

### DIFF
--- a/datetime/_date_time_formatter.ts
+++ b/datetime/_date_time_formatter.ts
@@ -295,7 +295,7 @@ export class DateTimeFormatter {
           const value = utc
             ? date.getUTCMilliseconds()
             : date.getMilliseconds();
-          string += digits(value, Number(part.value));
+          string += digits(value, 3).slice(0, Number(part.value));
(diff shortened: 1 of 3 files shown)


kt3k (approved):
LGTM. Good catches! Thanks

PR #6517, opened 2025-03-26
fix(ini): replace `isRecord()` with `isPlainObject()`

timreichen:
This PR fixes the section check where a object values like `Date` would pass the check.


kt3k (approved):
LGTM

PR #6521, opened 2025-03-27
chore(cli,http): fix typechecking errors

### LINKED ISSUE #6520: TypeScript errors in some scripts
These 2 items seem to have type errors:
- cli/_tools/compare_with_rust.ts
- http/_mock_conn.ts

WWRS:
resolves #6520

- `Deno.Conn.rid` was removed in https://github.com/denoland/deno/pull/25556
- `Deno.Conn.readable` changed type in https://github.com/denoland/deno/pull/27857
- Not sure when `unicode_width` started failing typecheck. The Rust `unicode_width` has been marked as a `usize` since `compare_with_rust.ts` was added, and `usize` has been marked as a `BigInt` since before that. But TS `unicode_width` has been marked as a `number` since it was added. The test failed before this change and succeeds after.

### DIFF
--- a/cli/_tools/compare_with_rust.ts
+++ b/cli/_tools/compare_with_rust.ts
@@ -49,7 +49,9 @@ Deno.test("fast-check equality with unicode_width Rust crate", async (t) => {
               // deno-lint-ignore no-explicit-any
               (str: any) =>
                 unicodeWidth(str) ===
-                  dylib.symbols.unicode_width(toCString(JSON.stringify(str))),
+                  Number(
+                    dylib.symbols.unicode_width(toCString(JSON.stringify(str))),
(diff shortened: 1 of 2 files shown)


cli/_tools/compare_with_rust.ts
WWRS, line 1: Actually, further up in this file, the `libName`s also seem to be different from the defaults. When I run `cargo build` on Win11, I get a file called `unicode_width_crate.dll` instead of `libunicode_width_crate.dll`. Not sure what happens on the other OSes.

kt3k (approved):
LGTM. Thanks for looking in to them!

I'll mark this commit as `chore` as the fix doesn't affect the user's code.

PR #6522, opened 2025-03-27
feat(collections): option to merge `undefined` in `deepMerge`

### LINKED ISSUE #4752: [feature request(collections)] option to merge `undefined` in `deepMerge`

when merging together deeply nested object (e.g react state), it's important to get 'nonempty' values when merging 'empty' values with 'nonempty' values.

for array, map and set it defaults to merging empty and non-empty values together:

```ts
deepMerge({ a: new Set([1]) }, { a: new Set() })
{ a: Set(1) { 1 } }
```

however, it doesn't work with nullish values. it'd be very useful to be able to output non-nullish value from merging nullish and non-nullish value together.

```ts
deepMerge({ a: 1 }, { a: undefined })
// expected: { a: 1 }
// actual: { a: undefined }
```

in [DeepMergeOptions](https://jsr.io/@std/collections@0.224.1/doc/deep-merge/~/DeepMergeOptions), add `nullish: MergingStrategy` that controls whether `deepMerge` will prefer non-nullish value over nullish ones.

make `nullish: "merge”` as default. this may cause breaking changes.

WWRS:
resolves #4752

Default is `"replace"`, which is the existing behavior.

### DIFF
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -208,7 +208,7 @@ export function deepMerge<
     arrays: "merge";
     sets: "merge";
     maps: "merge";
-    undefined: "replace";
+    undefineds: "replace";
   },
 >(
   record: Readonly<T>,
@@ -225,7 +225,7 @@ function deepMergeInternal<
     arrays: "merge";
     sets: "merge";
     maps: "merge";
-    undefined: "replace";
+    undefineds: "replace";
   },
 >(
   record: Readonly<T>,
@@ -255,7 +255,7 @@ function deepMergeInternal<
 
     if (
       !Object.hasOwn(other, key) ||
-      (other[key] === undefined && options?.undefined === "ignore")
+      (other[key] === undefined && options?.undefineds === "ignore")
     ) {
       result[key] = a;
 
@@ -289,7 +289,7 @@ function mergeObjects(
     arrays: "merge",
     sets: "merge",
     maps: "merge",
-    undefined: "replace",
+    undefineds: "replace",
   },
 ): Readonly<NonNullable<Record<string, unknown> | Iterable<unknown>>> {
   // Recursively merge mergeable objects
@@ -403,7 +403,7 @@ export type DeepMergeOptions = {
    *
    * @default {"replace"}
    */
-  undefined?: "replace" | "ignore";
+  undefineds?: "replace" | "ignore";
 };
 
 /**
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -208,6 +208,7 @@ export function deepMerge<
     arrays: "merge";
     sets: "merge";
     maps: "merge";
+    undefined: "replace";
   },
 >(
   record: Readonly<T>,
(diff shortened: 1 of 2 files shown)


collections/deep_merge.ts
```diff
@@ -387,6 +392,18 @@ export type DeepMergeOptions = {
...
+   * How to handle comparisons between non-`undefined` values and `undefined`.
+   *
+   * - If `"replace"`, the value in `other` is always chosen.
+   * - If `"ignore"`, the value in `other` is only chosen if not `undefined`.
+   *
+   * In both cases, a value of `undefined` is chosen over an omitted value.
+   *
+   * @default {"replace"}
+   */
+  undefined?: "replace" | "ignore";
```
ry, line 406: strange that the key is a keyword `undefined`

WWRS: That's true. Maybe `undefineds` would be better?

collections/deep_merge.ts
```diff
@@ -387,6 +392,18 @@ export type DeepMergeOptions = {
...
+   * How to handle comparisons between non-`undefined` values and `undefined`.
+   *
+   * - If `"replace"`, the value in `other` is always chosen.
+   * - If `"ignore"`, the value in `other` is only chosen if not `undefined`.
+   *
+   * In both cases, a value of `undefined` is chosen over an omitted value.
+   *
+   * @default {"replace"}
+   */
+  undefineds?: "replace" | "ignore";
```
kt3k, line 406: Sorry for late reply, but `undefineds` sounds a bit strange to me. Because this affects only when `undefined` appears as object property, how about `undefinedProp`, `undefinedProperty`, or `propertyUndefined`?

WWRS: Just above, the options are named `arrays`, `sets`, and `maps`, so I think `undefineds` does match with those most closely

kt3k: I see the point, but the usage of the word `undefineds` seems very rare compared to `sets`, `maps`, `arrays`. I'm afraid this might be confusing to the users. (I personally prefer the original `undefined` over `undefineds`. `undefined` is not JavaScript keyword ([ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words)), but just predefined variable name. I personally disagree with @ry's concern..)

fibibot (approved):
New `undefineds: "replace" | "ignore"` option threads cleanly through `deepMergeInternal` at `deep_merge.ts:253`: one short-circuit returns `record[key]` when `other[key] === undefined && options?.undefineds === "ignore"`, otherwise existing behavior runs unchanged. Default is `"replace"`, so existing callers are unaffected. Tests cover scalar-vs-`undefined` both directions, object-vs-`undefined` (with `"ignore"` preserving the object), nested property `undefined`, and `undefined`-vs-omitted both directions — the last case confirms the JSDoc's "a value of `undefined` is chosen over an omitted value" claim. CI is fully green across v1.x / v2.x / canary on all three OSes.

Issue #6470, opened 2025-03-08
`@std/cli`'s promptSelect is horrible with large lists


BlackAsLight:
When providing a list that ends up larger than the screen size available, it can be pretty hard to see the top results and leads to a lot of spam when scrolling back up.

A better user experience would be to figure out how much real estate is actually available in the terminal and never print more than that. When dealing with lists that would overflow the terminal's height, either pages should be introduced so the user can use the left and right keys to iterate between them or the down and up keys should scroll the list so it's only printing a subsection of the list at a given time.

PR #6523, opened 2025-03-27
feat(cli/unstable): promptSelect scrolls when the list is too large

WasixXD:
Solves #6470

Now, only a `visibleLines` amount of options are rendered at once.

I also added the option to control this size and change the character that renders the selected item indicator.


BlackAsLight:
Visible lines seems to have a default value of 3, which seems weird to me. Instead of asking for the lines, I think it would be better to look at `Deno.consoleSize().rows` to see how big the terminal is. A possible max lines may make sense though.

WasixXD:
> Visible lines seems to have a default value of 3, which seems weird to me. Instead of asking for the lines, I think it would be better to look at `Deno.consoleSize().rows` to see how big the terminal is. A possible max lines may make sense though.

But if the user has the terminal in full size it would't be strange to look at that many lines?

BlackAsLight:
> But if the user has the terminal in full size, wouldn't it be strange to look at so many lines at once?

No? They have it in full screen to see lines, and many cli programs take over the whole terminal when being interactive.

WasixXD:
> No? They have it in full screen to see lines, and many cli programs take over the whole terminal when being interactive.

I think that if they want to see more lines, they can just specify it. But i do agree that having 3 as the default is strange;  in comparison, [inquirer select](https://github.com/SBoudrias/Inquirer.js/tree/main/packages/select) has the default as 6

cli/unstable_prompt_select.ts
```diff
@@ -44,8 +55,13 @@ export function promptSelect(
 ): string | null {
   if (!input.isTerminal()) return null;
 
+  const { visibleLines = 3, indicator = "❯" } = options;
```
kt3k, line 58: 3 as default feels too small to me as most terminal has much more lines. How about making default to `values.length` (keeping the existing behavior)?

kt3k:
3 as default visible number seems strange and too small to me too. 6 might make more sense, but that number feels also a bit opinionated. How about setting the default to `values.length` (or `Math.min(values.length, Deno.consoleSize().rows`)`  and letting the user of the API choose the number?

cli/unstable_prompt_select.ts
```diff
@@ -44,8 +55,18 @@ export function promptSelect(
 ): string | null {
   if (!input.isTerminal()) return null;
 
+  const {
+    // Deno.consoleSize().rows - 3 because we need to output the message, the terminal line
+    // and we dont actually use the last line
+    visibleLines = Math.min(Deno.consoleSize().rows - 3, values.length),
```
WasixXD, line 61: Not sure how to make the CI accept the Deno.consoleSize()

BlackAsLight:
Testing out the code, it does look nice, but:
- I can no longer wrap around when hitting the end or go up from the top to jump down to the bottom of the list.
- The code should check the size of the terminal with every print so it does adjust to a change in terminal size.
   - When going from small to big, the size is still small on screen but usable.
   - When going from big to small, I can't see the top half of the list and get a lot of spam in the terminal if I scroll.
- It would also be nice for the list to display some symbol/icon at either end of the list to indicate that there is more content to see off screen as the user might think that what they currently see is all there is.

The code broke when testing it out in `deno repl` mode and I changed the size of my terminal, but idk if that is existing behaviour. It doesn't break if running a file and changing the terminal size

cli/unstable_prompt_select.ts
kt3k: It looks like CI is failing because it doesn't have terminal and `consoleSize` call throws `Inappropriate ioctl for device (os error 25)`

Can you stub `consoleSize` method in test cases like:
```js
stub(Deno, "consoleSize", () => ({ columns: 80, rows: 24 }));
```

WasixXD: Thank you!

WasixXD:
> The code broke when testing it out in `deno repl` mode and I changed the size of my terminal, but idk if that is existing behaviour. It doesn't break if running a file and changing the terminal size

I don't know how to test this one.

From my tests i think the rest is fine, wdyt?

kt3k (approved):
Tested locally with large input array, and it looks working reasonably. Thanks for your contribution!

Issue #122, opened 2019-01-15
Testing module compatibility


43081j:
At the minute we don't really seem to follow node's assertion interfaces or anyone else's.

Is there any reference as to what convention was followed or was it thought up just for deno?

It may be worth rewriting the testing module to adhere to node's assertion module and be entirely compatible.

Basically current implementation:

```ts
assertEqual(actual, expected, msg);
assert(equal(actual, expected), msg);
```

Node implementation:

```ts
assert.equal(actual, expected, msg);
```

So:
* `assertEqual` doesn't seem to be needed, only for setting a nice message (deepEqual and equal would do this themselves)
* `equal` does a deep equality check for us but in node does an `Object.is` alone (as `deepEqual` exists for the former)
* An `AssertionError` is thrown rather than an `Error` in node
* Negated functions exist in node such as `notEqual`

kitsonk:
Personally I would rather take queues from [Chai assert](https://www.chaijs.com/api/assert/) personally.

The behaviour of `assertEqual()` and `equal()` is already established, both in all of `std` but also the core of Deno.  There is a lot of tests that depend upon that deep comparison behaviour.

I don't think we should be pedantic about following anyone else's behaviour.  We can take queues and patterns from others.  The benefit of TypeScript and JSDoc is that the APIs can be self documenting if there is any confusion.

43081j:
I don't think it is about being pedantic and following other behaviours for that reason, but rather that these implementations are well established and known. Any difference in behaviour will lead to confusion as nobody is starting 'new' with deno, they will have used one of the widely used assertion libraries beforehand.

`equal` really shouldn't do a deep equality check and restricting it to that just because it would mean many changes seems a little unfortunate/backwards. It would be worth the effort to have a clean API.

Even if we take chai as an example to work from, it too has different behaviour and is closer to what node implements.

The lack of an assertion exception is a huge one, too, as you can already see repetition in the recent PR for 'error logging'. There should be one source of assertion messages which we can keep very clean.

I think all i listed still applies even if you follow chai, other than the naming convention and strictness.

43081j:
The reason for `equal` being simple (value equality check, usually) is that we can then have a `deepEqual` for when we want recursive equality checks. Leaving `equal` to do a deep check seems fine as long as we then have a `strictEqual` or similar, for when we _don't_ want deep equality (which would be `Object.is` pretty much).

[...]

PR #6524, opened 2025-03-27
feat(fs/unstable): add open, openSync, and FsFile class

jbronder:
This PR adds `open` and `openSync` APIs to the `@std/fs` package which are intended to mirror the `Deno.open` and `Deno.openSync` functions. In support of these APIs, the `OpenOptions` type and `getOpenFsFlag` function are added to this PR.

Also, since the `open(Sync)` APIs return an `FsFile` object, the `FsFile` interface, `NodeFsFile` internal class to convert a Node.js file descriptor into an `FsFile` instance as well as supporting runtime builtin module request functions are also added to this PR, (i.e. `getNodeStream`, `getNodeTty`, and `getNodeUtil` functions). Associated tests of `FsFile` methods are also added to this PR.

Towards #6255.

Testing of the `open(Sync)` APIs are predominately intended to test `OpenOptions`.

Test of the `NodeFsFile` is intended to test the instance methods of these objects.

### Caveats regarding `NodeFsFile` implementation
The `NodeFsFile` does not include the following implementations with reasoning for omissions:
- `seek(Sync)`: These methods rely on having the `SeekMode` enum. At the time of submitting this PR, Node.js cannot process enum types without enabling the experimental flag `--experimental-transform-types` in Node. Running Node.js with TS files only strips types for Node.js >= v23.6.0. For the purposes of passing CI, the `SeekMode` enum and `seek(Sync)` methods have been omitted.

- `lock(Sync)` and `unlock(Sync)`: These methods do not appear to have a direct Node.js equivalent. Based on the comments raised in issues from the Github`nodejs/node` repo: [#49256](https://github.com/nodejs/node/issues/49256), [#54446](https://github.com/nodejs/node/issues/54446), and [#122](https://github.com/nodejs/node/issues/122) a Node.js equivalent may never materialize at all. To briefly summarize comments, the complexity of providing a cross-platform `flock` implementation is driving the decision to not implement advisory file-locking in Node.js. As a result, a `new Error("Method not implemented.")` is suggested to be thrown for these methods.

- `setRaw`: Calling the Node.js version this method, `setRawMode`, appears to require instantiating a `tty.ReadStream` object with the current underlying file descriptor. I'm currently running into `EAGAIN (errno 35)` errors on testing. It's likely that I need to spend more time on this method. In the meantime, a `new Error("Method not implemented.")` is provided as a placeholder.

The `Disposable` interface/protocol has been implemented in the `NodeFsFile` class for users writing in TypeScript with Node.js such as using npm package like `tsx` to run programs. Since the `using` keyword isn't recognized in Node.js yet, test cases have not been developed since Node runs tests in JavaScript and Explicit File Resource Management is currently at Stage 3.


jbronder:
I apologize for the PR that I now realize, in hindsight, was probably a large number of lines of code to review in one PR.  I should've made the commits much smaller and discussed with you a better commit approach before sending out this PR.

If it helps in your review, @kt3k, I can either stub out the `NodeFsFile` methods to throw with `new Error("Method not implemented.")` or omit methods entirely.  The goal would be to get the `open(Sync)` APIs and a minimum implementation of the `NodeFsFile` class reviewed first.  I can then submit forthcoming PRs for `NodeFsFile` method implementations gradually with the relevant tests over time.  Please let me know if there is another approach you'd like to see regarding this PR.

kt3k:
@jbronder Sorry for the delay in review. This PR looks in a good state. No need of splitting in my view. I'm going to review in more details soon.

fs/unstable_types.ts
```diff
@@ -201,3 +201,383 @@ export interface WriteFileOptions {
...
+   * ```ts ignore
+   * import { open } from "@std/fs/unstable-open";
+   * // if "/foo/bar.txt" contains the text "hello world":
+   * using file = await open("/foo/bar.txt");
+   * const buf = new Uint8Array(100);
+   * const numberOfBytesRead = await file.read(buf); // 11 bytes
+   * const text = new TextDecoder().decode(buf);  // "hello world"
+   * ```
+   */
+  read(p: Uint8Array): Promise<number | null>;
```
kt3k, line 321: @ry Do you think we should port this API as well to `@std/fs`?

kt3k: Context: This is legacy API from when Deno have 2 sets of I/O interfaces Reader/Writer vs Web Streams. We deleted Reader/Writer interfaces in many places at Deno 2, but we kept this API for backward compatiblity (to my understanding).

kt3k: Ok. Let's keep them

kt3k:
@jbronder I'll review in more details tomorrow

fs/_node_fs_file.ts
```diff
@@ -0,0 +1,211 @@
...
+    throw new Error("Method not implemented.");
+  }
+
+  // deno-lint-ignore no-unused-vars
+  lockSync(exclusive?: boolean): void {
+    throw new Error("Method not implemented.");
+  }
+
+  async read(p: Uint8Array): Promise<number | null> {
+    const nodeReadFd = this.#nodeUtil.promisify(this.#nodeFs.read);
```
kt3k, line 75: Can we store this function at `this.#nodeReadFs` in the constructor? It feels like too much overhead to call promisify for each read

kt3k: Or maybe store in this function only when it's not initialized yet.

kt3k (approved):
Looks nicely implemented and tested! Thank you for your contribution!

jbronder:
Thank you for looking over my PR and updating the promisified `read` and `write` calls. Good call! I appreciate your time reviewing, @kt3k!

PR #6525, opened 2025-03-28
fix(expect): include diff when `expect.toMatchObject` throws

### LINKED ISSUE #6225: expect toMatchObject should show which object properties caused the assertion to fail
When using expect(...).toMatchObject with complex objects, it's hard to see which property/properties caused the assertion to fail.

**Steps to Reproduce**

```ts
import { it } from '@std/testing/bdd';
import { expect } from '@std/expect';

const expected = {
  field1: 'abc',
  field2: 'def',
  foo: 'bar',
}

it('toMatchObject', () => {
  expect({ 
    field1: 'abc',
    field2: 'def',
    foo: 'baz',
  }).toMatchObject(expected);
});
```

```
$ deno test denobug.test.ts 
running 1 test from ./denobug.test.ts
toMatchObject ... FAILED (2ms)

 ERRORS 

toMatchObject => https://jsr.io/@std/testing/1.0.5/_test_suite.ts:218:10
error: AssertionError: Expected {
  field1: "abc",
  field2: "def",
  foo: "baz",
} to NOT match {
  field1: "abc",
  field2: "def",
  foo: "bar",
}
```

The library could indicate which property did not match, at lease for the top-level properties (in this case, `foo`).

`assertObjectMatch` and Jest do this, for all properties, including nested ones:

image

But obviously, `assertObjectMatch` doesn't support matchers, so there's no easy workaround for this issue.

WWRS:
closes #6225

Does not change the behavior of `expect.toMatchObject`, only the error message.

The behaviors of `expect.toMatchObject` and `assertMatchObject` are slightly different, so one of the tests from the latter does not work. I duplicated the rest of the diff-based tests.

Also makes `buildNotEqualErrorMessage` accept the `formatter` option with a default of `format` which aligns it with `buildEqualErrorMessage`. For both, makes `options` optional.

### DIFF
--- a/expect/_build_message.ts
+++ b/expect/_build_message.ts
@@ -18,9 +18,9 @@ function isString(value: unknown): value is string {
 export function buildEqualErrorMessage<T>(
   actual: T,
   expected: T,
-  options: EqualErrorMessageOptions,
+  options: EqualErrorMessageOptions = {},
 ): string {
-  const { formatter = format, msg } = options ?? {};
+  const { formatter = format, msg } = options;
   const msgPrefix = msg ? `${msg}: ` : "";
   const actualString = formatter(actual);
   const expectedString = formatter(expected);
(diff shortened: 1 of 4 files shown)


kt3k (approved):
The error message looks much better now. Thanks!

PR #6526, opened 2025-03-28
chore(semver): remove `MIN` constant

timreichen:
Seems like this constant is only used in tests.

### DIFF
--- a/semver/_constants.ts
+++ b/semver/_constants.ts
@@ -3,17 +3,6 @@
 import type { Comparator, SemVer } from "./types.ts";
 
 /**
- * The minimum valid SemVer object. Equivalent to `0.0.0`.
- */
-export const MIN: SemVer = {
-  major: 0,
-  minor: 0,
-  patch: 0,
-  prerelease: [],
-  build: [],
-};
-
-/**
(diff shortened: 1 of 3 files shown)


kt3k:
> Seems like this constant is only used in tests.

It's fine to have a private constant for testing in my view. I don't see the point of removing it

timreichen:
> It's fine to have a private constant for testing in my view. I don't see the point of removing it

I would agree in principal, though the same value as `MIN` is already tested in `is_semver_test.ts` a few lines above (`[{ major: 0, minor: 0, patch: 0, build: [], prerelease: [] }]`) and does not matter in `is_range_test.ts`, because it tests the same as the value above except the operator.

kt3k:
> because it tests the same as the value above except the operator.

I think that changes the contents of tested `range` (it now has only one Comparator instead of 2). Can we inline `...MIN` part instead?

timreichen:
> I think that changes the contents of tested `range` (it now has only one Comparator instead of 2). Can we inline `...MIN` part instead?

done

kt3k (approved):
LGTM

PR #6529, opened 2025-03-29
test(toml): replace parser combinator with `parse()` tests

timreichen:
This PR replaces all parser combinator tests with `parse()` tests. This simplifies the api and reduces exports.


kt3k:
I'm against this change. This makes it difficult to work with single parse component

timreichen:
> I'm against this change. This makes it difficult to work with single parse component
> 
> 
> 
> These test cases were initially at _parser_test.ts, for the purpose of testing internal components https://github.com/denoland/std/commit/43ef9695693046465c9386f6e4a9d0249c562337
> 
> 
> 
> We already have enough test cases for testing the public `parse` API. This change makes most test cases duplicate.

I think this would indicate that lots of these tests are doubles and therefore obsolete. I imagine testing parser combinators was implemented before std used codecov. While that made sense then I think we can rely on codecov in order to reduce duplicate tests.

kt3k:
> which means we might need to change tests every time the inner functionality of a combinator changes

Each parser component is tested in isolation. I don't think adding new feature affects unrelated parser components.

> or add new tests if we add combinators for consistency.

This sounds just reasonable to me.

timreichen:
> > which means we might need to change tests every time the inner functionality of a combinator changes
> 
> Each parser component is tested in isolation. I don't think adding new feature affects unrelated parser components.

Not unrelated parsers, but it would not affect any parsers at all if they are not tested separately.

> > or add new tests if we add combinators for consistency.
> 
> This sounds just reasonable to me.

We also don't test selective private functions in other mods like `semver` or `yaml`, so this doesn't seem very reasonable to me. All code is tested through `parse()`, why would we need separate tests?

kt3k:
> All code is tested through parse(), why would we need separate tests?

Granular tests are good for debugging. I think it'd also be good for new contributors to understand the code base.

> We also don't test selective private functions in other mods like semver or yaml, so this doesn't seem very reasonable to me.

Each package has different authors and different commit history. These tests were first written by @FujiHaruka and they look enough reasonable to me.

Issue #2355, opened 2022-06-16
Node built-in https module implementation into deno's std/node library


bytooo:
**Is your feature request related to a problem? Please describe.**
I would like to use the [google-spreadsheet](https://www.npmjs.com/package/google-spreadsheet) module, more specifically the [skypack](https://www.skypack.dev/view/google-spreadsheet) version of it with deno, but it requires the **https node module** and according to the std [manual](https://deno.land/manual/node/std_node) the **Node built-in https module** is not currently implemented.

When running:
```typescript
import { GoogleSpreadsheet } from "https://cdn.skypack.dev/google-spreadsheet";
import credentials from "./mycredfile" assert { type: "json"}
const doc = new GoogleSpreadsheet("doc id");
await doc.useServiceAccountAuth(credentials);
```
I get the following error:
```
error: Uncaught Error: [Package Error] "https" does not exist. (Imported by "gaxios").
throw new Error("[Package Error] \"https\" does not exist. (Imported by \"gaxios\").");
      ^
    at https://cdn.skypack.dev/error/node:https?from=gaxios:14:7
```

**Describe the solution you'd like**
The solution I would like is for the **Node built-in https module** to be implemented into the Deno standard library as a way to provide a pollyfill for it.

**Describe alternatives you've considered**
I have considered looking for other alternatives to [google-spreadsheet](https://www.npmjs.com/package/google-spreadsheet) in the Deno third-party modules but I was not lucky enough to find one.

kt3k:
`std/node` has a basic support of `https` module (At least https.request and https.get are available). I think the above is an issue of skypack's side.

BTW google-spreadsheet module seems available & importable with `esm.sh`. How about trying that version?

The below ran successfully on my machine:
```ts
import * as a from "https://esm.sh/google-spreadsheet";
console.log(a);
```

bytooo:
> `std/node` has a basic support of `https` module (At least https.request and https.get are available). I think the above is an issue of skypack's side.
> 
> 
> 
> BTW google-spreadsheet module seems available & importable with `esm.sh`. How about trying that version?
> 
> 
> 
> The below ran successfully on my machine:
> 
> ```ts
> 
> import * as a from "https://esm.sh/google-spreadsheet";
> 
> console.log(a);
> 
> ```

Thank you so much for your reply, I'll try it and close this issue. I just saw the message in the manual saying that it wasn't yet implemented and created the issue.

PR #6530, opened 2025-03-29
feat(testing/unstable): add `assertInlineSnapshot()`

### LINKED ISSUE #3301: [feature request] Add support for inline snapshots tests

[Jest inline shapshots](https://jestjs.io/docs/snapshot-testing#inline-snapshots) are like file-based snapshots but much more casual. I love using them for small snapshots and I think they'd be a great improvement to Deno. From Jest's docs, they look like:

```ts
it('renders correctly', () => {
  const tree = renderer
    .create(<Link page="https://example.com">Example Site</Link>)
    .toJSON();
  expect(tree).toMatchInlineSnapshot(`
<a
  className="normal"
  href="https://example.com"
  onMouseEnter={[Function]}
  onMouseLeave={[Function]}
>
  Example Site
</a>
`);
});
```

The Deno version might look like:

```ts
Deno.test("renders correctly", async () => {
  const tree = renderer
    .create(<Link page="https://example.com">Example Site</Link>)
    .toJSON();
  await assertInlineSnapshot(`
<a
  className="normal"
  href="https://example.com"
  onMouseEnter={[Function]}
  onMouseLeave={[Function]}
>
  Example Site
</a>
`);
});

```

Right now, I just stick to the file-based snapshots which works fine but I find the locality of the inline snapshots much quicker to iterate on and they feel lightweight. I tend to avoid manual assertions more than a property or two as they're cumbersome to update.

WWRS:
closes #3301

Updating the snapshots is similar to Jest: Jest uses Babel's AST parser to find where to insert the snapshot and then Prettier to format. This instead uses a Deno lint plugin to find where to insert and `deno fmt` to format.

The `expectedSnapshot`s in `_assert_inline_snapshot_test.ts` were automatically generated by `assertInlineSnapshot`!


kt3k:
Thanks for the PR, but this seems only supporting the initial creation of snapshot. In my view the capability of updating the snapshots are essential for snapshot testing tool. Can we also support the updating somehow? (I guess we can do that only by using AST analysis as jest does)

WWRS:
I agree that this limitation of `assertInlineSnapshot` means it's not as nice as Jest's. 

You *can* update a snapshot by manually replacing the existing one with \`CREATE\` and running the test. This does force users to follow the commonly recommended practice of ensuring each updated snapshot should actually have been updated. Also, users always have the option of `assertSnapshot()` for tests that will need to be updated frequently enough that manual updating would be a hassle.

If we wanted to support updating failing snapshots automatically, it's likely that the best solution would involve AST parsing. I don't think Deno exposes its AST parser, and Jest uses Babel which is presumably something we cannot do here.

I'm not sure, but I suspect that even without the ability to update automatically, users might find this feature useful. On the other hand, I do recognize the need to maintain high quality throughout, and shipping an incomplete feature is not great either.

stefnotch:
So if I understand this correctly, there are 4 different options here
- Accept the simple implementation 
- Use a Deno.lint plugin to update the snapshots, since they have an AST inspection and mutation API. 
  - That plugin should ideally ship with Deno by default.
  - Bonus points for getting IDE integration for free. The light bulb would be able to suggest updating the snapshot.
- Make `Deno.assertSnapshot` a built-in feature of Deno, just like `Deno.test` and `Deno.bench`

WWRS:
Also an option would be to expose an AST parsing API as a built-in feature of Deno or as a library. From looking at #2355 it seems like this functionality will not be added to Deno but is in [`deno_swc`](https://github.com/littledivy/deno_swc). Would it be reasonable to include that as a dependency here?

I think adding `assertSnapshot` to Deno is not very likely. All the assertions are in `std/assert` so it doesn't make a ton of sense to have just one or two assertion functions in default Deno.

For `deno_lint`, I don't really know what our options look like there. Really all we need to do is find `assertMatchInlineSnapshot` (and later `expect.toMatchInlineSnapshot`) and manipulate the third param. But like adding to default Deno, it seems unlikely we would want support for just one assertion function in the default linter.

stefnotch:
I was about to write a comment about how the lint plugin option wouldn't work that well, until I realised something.

`Deno.lint.runPlugin` is a function that Deno provides in test mode.
Which exactly fits the bill of what we want! That's amazing!

So I quickly whipped up a prototype which makes use of that to give us inline snapshots!

Entire project is this, run it as per usual with `deno test` and `deno test -A -- --update`
[temp-deno-linter-for-snapshots.zip](https://github.com/user-attachments/files/19595622/temp-deno-linter-for-snapshots.zip)


https://github.com/user-attachments/assets/592996ad-6110-4e5d-ad89-0ccd4a72b2bf




And just the code. I'm okay with you copy-pasting and adapting this straight into Deno, and am releasing this under CC0 and MIT.
```ts
import { serialize, SnapshotOptions } from "@std/testing/snapshot";
import { equal } from "@std/assert/equal";
import { AssertionError } from "@std/assert/assertion-error";

const isUpdateMode = Deno.args.some((arg) =>
  arg === "--update" || arg === "-u"
);

interface SnapshotUpdateRequest {
  fileName: string;
  lineNumber: number;
  columnNumber: number;
  actual: string;
}

// Batch all writes until the very end, and then update all files at once
globalThis.addEventListener("unload", () => {
  updateSnapshots();
});
function updateSnapshots() {
  if (updateRequests.length === 0) {
    return;
  }
  console.log(`Updating ${updateRequests.length} snapshots...`);
  const filesToUpdate = Map.groupBy(updateRequests, (v) => v.fileName);

  for (const [fileName, requests] of filesToUpdate) {
    const fileContents = Deno.readTextFileSync(fileName);
    const pluginRunResults = Deno.lint.runPlugin(
      makeSnapshotUpdater(requests),
      "dummy.ts",
      fileContents,
    );
    const fixes = pluginRunResults.flatMap((v) => v.fix ?? []);
    if (fixes.length !== requests.length) {
      console.error(
        "Something went wrong, not all update requests found their snapshot",
      );
    }
    // Apply the fixes
    fixes.sort((a, b) => a.range[0] - b.range[0]);
    let output = "";
    let lastIndex = 0;
    for (const fix of fixes) {
      output += fileContents.slice(lastIndex, fix.range[0]);
      output += wrapForJs(fix.text ?? "");
      lastIndex = fix.range[1];
    }
    output += fileContents.slice(lastIndex);
    Deno.writeTextFileSync(fileName, output);
  }
}

const updateRequests: SnapshotUpdateRequest[] = [];

export function assertInlineSnapshot<T>(
  actual: T,
  expected: string,
  options?: SnapshotOptions<T>,
) {
  const _serialize = options?.serializer ?? serialize;
  const _actual = _serialize(actual);

  if (equal(_actual, expected)) {
    return;
  }
  if (isUpdateMode) {
    // Uses the V8 stack trace API to get the line number where this function was called
    const oldStackTrace = (Error as any).prepareStackTrace;
    try {
      const stackCatcher = { stack: null as SnapshotUpdateRequest | null };
      (Error as any).prepareStackTrace = (
        _err: unknown,
        stack: unknown[],
      ): SnapshotUpdateRequest | null => {
        const callerStackFrame = stack[0] as any;
        if (callerStackFrame.isEval()) return null;
        return {
          fileName: callerStackFrame.getFileName(),
          lineNumber: callerStackFrame.getLineNumber(),
          columnNumber: callerStackFrame.getColumnNumber(),
          actual: _actual,
        };
      };
      // Capture the stack that comes after this function.
      Error.captureStackTrace(stackCatcher, assertInlineSnapshot);
      // Forcibly access the stack, and note it down
      const request = stackCatcher.stack;
      if (request !== null) {
        const status = Deno.permissions.requestSync({
          name: "write",
          path: request.fileName,
        });
        if (status.state !== "granted") {
          console.error(
            `Please allow writing to ${request.fileName} for snapshot updating.`,
          );
        } else {
          updateRequests.push(request);
        }
      }
    } finally {
      (Error as any).prepareStackTrace = oldStackTrace;
    }
  } else {
    throw new AssertionError("Assertion failed blabla");
  }
}

/// <reference lib="deno.unstable" />
/**
 * Makes a Deno.lint plugin that can find inline snapshots.
 * Also deals with multiple `assertInlineSnapshot` in a single line.
 */
function makeSnapshotUpdater(
  updateRequests: SnapshotUpdateRequest[],
): Deno.lint.Plugin {
  const linesToUpdate = Map.groupBy(updateRequests, (v) => v.lineNumber);

  return {
    name: "snapshot-updater-plugin",
    rules: {
      "update-snapshot": {
        create(context) {
          return {
            'CallExpression[callee.name="assertInlineSnapshot"]'(
              node: Deno.lint.CallExpression,
            ) {
              // Find the update request that corresponds to this snapshot.
              // Successful snapshots don't have an update request.
              const callPosition = toLineAndColumnNumber(
                node.range[0],
                context.sourceCode.text,
              );
              const endColum = callPosition.column +
                (node.range[1] - node.range[0]);
              const lineUpdateRequests = linesToUpdate.get(callPosition.line);
              if (lineUpdateRequests === undefined) {
                return;
              }
              const updateRequest = lineUpdateRequests.find((v) =>
                callPosition.column <= v.columnNumber &&
                v.columnNumber <= endColum
              );
              if (updateRequest === undefined) {
                return;
              }

              context.report({
                node,
                message: "",
                fix(fixer) {
                  return fixer.replaceText(
                    node.arguments[1],
                    updateRequest.actual,
                  );
                },
              });
            },
          };
        },
      },
    },
  };
}

/**
 * Takes an index and returns the 1-based line number and 1-based column number */
function toLineAndColumnNumber(index: number, text: string) {
  const textBefore = text.slice(0, index);
  // TODO: Verify that Chrome's V8 uses the same logic for returning line numbers. What about the other line terminators?
  const lineBreakCount = (textBefore.match(/\n/g) || []).length;

  // Also deals with the first line by making use of the -1 return value
  const lastLineBreak = textBefore.lastIndexOf("\n");
  return {
    line: lineBreakCount + 1,
    column: (index - lastLineBreak),
  };
}

function wrapForJs(str: string) {
  return "`" + escapeStringForJs(str) + "`";
}

function escapeStringForJs(str: string) {
  return str
    .replace(/\\/g, "\\\\")
    .replace(/`/g, "\\`")
    .replace(/\$/g, "\\$");
}
```

stefnotch:
@WWRS Would you be willing to integrate the Deno.lint approach into this PR? I'd really appreciate it, since that'd save me the effort of creating and polishing a pull request for this particular feature.
I'm asking since your PR already has unit tests and other necessary bits.

If yes, here are some quick pointers
- I think we could get away with dropping the first argument for `assertInlineSnapshot`. One already gets the correct file from the stack trace.
- If you have any ideas on how to simplify parts of the code, I'd be delighted to hear them. I couldn't figure out a simpler way of doing the "find the correct `assertInlineSnapshot`". The simple approach of counting wouldn't work here, since the linter doesn't know in which order the inline snapshot functions are called.
- Re-assigning `assertInlineSnapshot` to a new variable is unsupported. So `assertMonochromeInlineSnapshot` wouldn't work. I wonder if there's a good solution for that.
- Wrapping `assertInlineSnapshot` in an extra function would lead to very unexpected results. I'm not sure if one can do much about that.
- I wonder what happens if one types `deno test --watch -A -- --update`

WWRS:
I'm not 100% sold on using the error throwing location as the way to find where to insert the snapshot, it seems a bit obscure and dependent on V8. As noted, we would need to double check how V8 counts line numbers, but also how it counts unicode. (For reference, Jest does use this error throwing approach.)

For finding the right snapshot to update, would the numbering system that's in the current implementation of `toMatchSnapshot` work? It seems like tests are always run from top to bottom within the same file. I'm not sure if this works when running a subset of the tests, but if it breaks here, it might have issues in `toMatchSnapshot` as well.

For whether or not we need a context argument, if we were to remove the argument here, we would probably want to make a similar change in `toMatchSnapshot` to keep the call signatures parallel. I'm not sure how messy that would get, do you think it would be possible? Also, #6541 is worth considering: If we were to add a context-injected `expect` in `bdd` tests, then we could make `toMatchSnapshot` and `toMatchInlineSnapshot` use that `expect` and get its context for free.

I'd like to hear some other opinions on the tradeoffs of the two proposed implementations. As noted above, the magic-word implementation lacks a core capability because it requires screenshots to be manually flagged for updates. Are the missing capabilities of the lint implementation (`assertMonochromeInlineSnapshot` or wrapping in a helper function) also core?

WWRS:
I checked, the existing `assertSnapshot`-style counting system does work. This means we can avoid using errors to find where to insert snapshots.

stefnotch:
Thank you so much for taking a look at this. Those are some very good points, so I did some reading.

### Error Throwing
Yes, the error throwing approach relies on V8 APIs. I wish we had a standard API for this https://github.com/tc39/proposal-error-stacks

For now, I think it's fine, because updating the snapshots relies on filesystem access. So, it's limited to server-side JS runtimes, which generally implement the V8 APIs. Even Bun does.

And yes, we absolutely have to check how Unicode gets counted for the column number. We'll have to verify whether the following behaves as expected.
```js
/* 🐈‍⬛🐈‍⬛🐈‍⬛🐈 */ assertInlineSnapshot(2 + 3, `'5'`)
```
On that note, the linter's ranges also rely on an important detail: They're reported in UTF-16 code-somethings, which matches Javascript string indexing. So at least that part is sensible.

### Numbering system 

`assertSnapshot` relies on a numbering system. It computes a stable ID for each `assertSnapshot` call. That ID only changes when one renames the tests, or starts calling `assertSnapshot` in a different order.
To update a snapshot, it runs the test, and saves the new value with the ID.

(Amusingly, it means that `if(Math.random() < 0.5) { await assertSnapshot(t, 1); } else { await assertSnapshot(t, 2); })` would repeatedly break. But that's a non-deterministic test, so it deserves to break.)


This does not work for `assertInlineSnapshot`. 
To update a snapshot, it has to find the correct place in the source code.

One quick example where the numbering system would not work is
```ts
test("foo", () => {
  if(false) {
    assertInlineSnapshot("never happens");
  }
  assertInlineSnapshot("wait, why am I not getting updated");
});
```
I'm sure one can also think of a counterexample with promises, like calling `assertInlineSnapshot` in a `.then`

### Context argument

Good point, that'd be an odd inconsistency.

So https://github.com/denoland/std/issues/3964 has an open issue about implementing a `expect(foo).toMatchSnapshot()` API.

But clearly the normal snapshotting API needs a test context. So I looked at what Jest and Vitest do.

Vitest passes the context to `expect` when run synchronously, and tells you to use the correct `expect` when you have async tests.
https://vitest.dev/guide/snapshot.html#use-snapshots
```
test('math is easy', async ({ expect }) => {
  expect(2 + 2).toMatchSnapshot()
})
```

Jest instead assumes that it runs within Node.js, and makes use of the `AsyncLocalStorage` API.
https://github.com/jestjs/jest/pull/14139
Deno also implements that API https://docs.deno.com/api/node/async_hooks/~/AsyncLocalStorage
The downside of it is that it doesn't work in browsers. 

I think you're better than I am at judging which of the trade-offs is the best option here.

### `assertMonochromeInlineSnapshot`

There are four options for dealing with the limitation.
1. Only expose it as `expect(foo).toMatchInlineSnapshot(`'bar'`)`. Now the temptation to alias it is much lower.
2. Don't look for a specific work in the syntax tree, only look for a function call at the correct spot. As in, update it as long as there is a `foo(bar, ...)` in the expected position.
3. Add a setting where one can pass in the name to look for, like
```ts
const assertMonochromeInlineSnapshot = createAssertInlineSnapshot<string>({
    serializer: stripAnsiCode,
    name: "assertMonochromeInlineSnapshot" // Users are very likely to forget this, and it would not survive minification or re-assignment.
  });
```
4. Just document the limitation. Inline snapshots are wizardry anyways.

stefnotch:
I just found out that there's a proposal for async context tracking https://github.com/tc39/proposal-async-context

If we had that, then we'd no longer need the first argument for the regular `assertSnapshot` function. And wr could finally perfectly re-implement the `expect().toMatchSnapshot()` function from Jest.

stefnotch:
@WWRS I figured I'd check back in: How is it going? Which trade-offs should we choose? Is there anything I could do to help move this PR forward?

WWRS:
@kt3k This now supports AST analysis, allowing us to update by passing the `--update` flag instead of a magic string.

@stefnotch I am now using the error throwing location instead of the numbering system. For `assertMonochromeInlineSnapshot` I am looking at every function call and grabbing the one that matches. I also added a test to ensure this counts lines and columns the same way V8 does.

Remaining weird stuff:
- I think this fails if someone tries to `await assertInlineSnapshot`. The only case that's likely to happen in is if a user wraps `assertInlineSnapshot` in an async function, which seems rare. But we may need to improve the AST matcher if anyone is actually doing this.
- This fails if someone tries to wrap it in a function where the expected snapshot is not the third variable. Again, should be rare, but if anyone is actually doing this, we can add an option for which index the expected snapshot is at.

testing/_assert_inline_snapshot.ts
```diff
@@ -0,0 +1,401 @@
...
+export function assertInlineSnapshot(
+  context: Deno.TestContext,
+  actual: unknown,
+  expectedSnapshot: string,
+  msgOrOpts?: string | InlineSnapshotOptions<unknown>,
+): void {
+  const options = getOptions(msgOrOpts);
+  const assertInlineSnapshotContext = AssertInlineSnapshotContext.fromContext(
+    context,
+  );
```
stefnotch, line 300: That's a reasonable use of the test context. Am happy with this API.

If we ever want to build the `expect(foo).toMatchInlineSnapshot()` API, we'll have to extract the current file name from the stack trace. But that's a future feature request that I do not absolutely need. As long as I have one API for inline snapshots, I am really happy.

stefnotch (approved):
Thank you so much for all your hard work! I can't wait for this to make its way into Deno

testing/_assert_inline_snapshot.ts
```diff
@@ -0,0 +1,401 @@
...
+        create(context) {
+          const src = context.sourceCode.text;
+          const lineBreaks = [...src.matchAll(/\n|\r\n?/g)].map((m) => m.index);
+          const locationToSnapshot: Record<number, string> = {};
+          for (
+            const [lineColumn, snapshot] of Object.entries(lineColumnToSnapshot)
+          ) {
+            const [lineNumber, columnNumber] = lineColumn.split(":")
+              .map(Number);
+            const location = (lineBreaks[lineNumber! - 2] ?? 0) + columnNumber!;
```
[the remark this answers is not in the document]
WWRS: Ah yeah I'll leave a comment. The line numbers are 1-indexed so we need to do a -1 to get to 0-indexed. Then we need to get the position of the break before this line, so that's the `n-1`th break, which is where the second -1 comes from.

testing/_assert_inline_snapshot.ts
```diff
@@ -0,0 +1,401 @@
...
+    if (snapshotsUpdated === 0) return;
+
+    if (!LINT_SUPPORTED) {
+      throw new Error(
+        "Deno versions before 2.2.0 do not support Deno.lint, which is required to update inline snapshots",
+      );
+    }
+
+    const testFilePath = fromFileUrl(this.#testFileUrl);
+    ensureFileSync(testFilePath);
```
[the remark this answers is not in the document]
WWRS: I probably left this in from `snapshot.ts`. The test file should always exist, and even if it doesn't the read will error. I'll remove this.

testing/_assert_inline_snapshot_test.ts
```diff
@@ -0,0 +1,194 @@
...
+        "--",
+        "--update",
+      ],
+    });
+    await command.output();
+
+    assertEquals(
+      await Deno.readTextFile(countTestFile),
+      `import { assertInlineSnapshot } from "${SNAPSHOT_MODULE_URL}";
+ \n \r \n\r \r\n
```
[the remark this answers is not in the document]
WWRS: After some investigation,

https://github.com/denoland/deno_core/issues/1123

stefnotch (approved):
LGTM

stefnotch:
I believe this is ready for the Deno team to review, right?

On that note, a question for @kt3k : Do you think that this is interesting enough to be noted somewhere in the Deno blog or in a Deno Tweet? While it is a major feature for me, I have no idea how widely used inline snapshots are. Hence this question.

kt3k:
Hey, thank you for your work and your review. I think this PR is in right direction, but I'll review it in more details soon.

I think this feature is worth the tweet. We'll definitely consider to do that. BTW have you seen @ry already tweeted [this](https://x.com/rough__sea/status/1908003160418930710)?

stefnotch:
I was not aware of the ry tweet. I'm absolutely delighted to hear about it.

And thank you for the upcoming review. I look forward to this PR getting polished and landing in the Deno standard library.

kt3k:
Moved `assertInlineSnapshot` and `createAssertInlineSnapshot` to `testing/unstable-snapshot` to publish these first as unstable APIs

testing/unstable_snapshot.ts
```diff
@@ -0,0 +1,411 @@
...
+  /**
+   * Returns an instance of `AssertInlineSnapshotContext`. This will be retrieved from
+   * a cache if an instance was already created for a given test file path.
+   */
+  static fromContext(
+    testContext: Deno.TestContext,
+  ): AssertInlineSnapshotContext {
+    const testFilePath = fromFileUrl(testContext.origin);
+    const { dir, base } = parse(testFilePath);
+    const path = resolve(dir, base);
```
kt3k, line 119: Why do we need to parse and resolve here? Is it possible to use `testContext.origin` directly as context key?

WWRS: Yeah, we should just use `testContext.origin` if that works.

I copied this from the implementation in `AssertSnapshotContext` where it's needed to generate the snapshot file name, but here we don't need that.

testing/unstable_snapshot.ts
```diff
@@ -0,0 +1,411 @@
...
+export function assertInlineSnapshot(
+  context: Deno.TestContext,
+  actual: unknown,
+  expectedSnapshot: string,
+  msgOrOpts?: string | InlineSnapshotOptions<unknown>,
+): void {
+  const options = getOptions(msgOrOpts);
+  const assertInlineSnapshotContext = AssertInlineSnapshotContext.fromContext(
+    context,
+  );
```
kt3k, line 308: I might be missing something in the discussion, but can we get the file name from the CallSite object as @stefnotch's suggested implementation does? I feel that'd be more elegant design. What do you think?

testing/unstable_snapshot.ts
```diff
@@ -0,0 +1,411 @@
...
+export interface InlineSnapshotOptions<T = unknown>
+  extends Pick<SnapshotOptions<T>, "mode" | "msg" | "serializer"> {
+  /**
+   * Whether to format the test file after updating.
+   *
+   * The default is `true`. If multiple snapshots will be created in one test file
+   * and the tests have incompatible `format` options, the snapshots will be written,
+   * the file will not be formatted, and we will throw.
+   */
+  format?: boolean;
```
kt3k, line 34: How about getting `format` option from CLI args (and only from CLI args) (like `deno test -- --update --format`)? I doubt people want to specify this per assertion basis.

testing/unstable_snapshot.ts
```diff
@@ -0,0 +1,411 @@
...
+  serialize,
+} from "./_snapshot_utils.ts";
+
+/**
+ * The options for {@linkcode assertInlineSnapshot}.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ */
+export interface InlineSnapshotOptions<T = unknown>
+  extends Pick<SnapshotOptions<T>, "mode" | "msg" | "serializer"> {
```
kt3k, line 26: How about dropping `mode` from `InlineSnapshotOptions`? That'll be inconsistent with `assertSnapshot`, but I don't think the users would ever specify `mode` via this option (People should specify `mode` only via CLI flag (`--update`)).

kt3k:
Thank you for your work! Left some comments about the design. Other part looks good to me.

testing/unstable_snapshot.ts
WWRS: Should we default to formatting or not formatting? I think generally people will want to format by default.

stefnotch: How much of the formatting can change when inline snapshots are updated? It's just the "single line <=> multi line" formatting, right?

If it is just that, then there's a 3rd option: Have the linter automatically output correctly formatted code. I'm more than happy to look at that in a future PR.

I don't have a strong preference as to whether it should format by default. Having a short flag for "no format" seems like an overkill.

kt3k: I don't have strong preference either.

> Having a short flag for "no format" seems like an overkill.

I agree with this. Let's not have short flag for `--no-format` when we default to format

kt3k (approved):
@WWRS Nice work! LGTM

@stefnotch Thanks for your reviews and initial suggestion of the implementation!

PR #6532, opened 2025-03-30
refactor(toml): add `Scanner` `startsWith()` method

timreichen:
This reduces the usage of slice() and might have a positive impact on performance.


kt3k (approved):
Good idea. LGTM

PR #6535, opened 2025-03-30
chore(toml): make `Scanner` iterable

timreichen:


kt3k:
This seems making the code base trickier to read to me. Scanner is not simply iterable like an array, but contributors need to be very careful when to call `.next()`. This seems obscuring when the pointer moves forward.

timreichen:
I played around a bit more with the toml parser combinator implementation. I think whole implementation can be simplified, so closing this in favour of upcoming PRs.

PR #6537, opened 2025-03-30
fix(http): use relative urls in file server dirlisting

### LINKED ISSUE #6536: http/file_server: Make links use relative urls

I commonly create servers which are responsible for hosting multiple (sub)domains. A pattern I often use is to rewrite the `request.url` property. That way I can visit `https://localhost:8080/https://example.com/path/to/foo` in my browser, and the request will be rewritten to just `https://example.com/path/to/foo`.

The problem with this is is that it breaks `serveDir()`, as any urls in the dir listing will now remove the domain part and redirect to a broken url.

Using relative urls would fix this issue and seems like it would be easy. After all entries in the dir listing are relative to the current directory, so only adding `./` in front of the urls should be enough.

I could try and use the original request in `serveDir()` but this approach seems more robust

jespertheend:
Closes #6536

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -433,7 +433,8 @@ function createBaseHeaders(): Headers {
 }
 
 function dirViewerTemplate(dirname: string, entries: EntryInfo[]): string {
-  const paths = dirname.split("/");
+  const splitDirname = dirname.split("/").filter((path) => Boolean(path));
+  const headerPaths = ["home", ...splitDirname];
 
   return `
     <!DOCTYPE html>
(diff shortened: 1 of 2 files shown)


http/file_server.ts
```diff
@@ -373,7 +370,7 @@ async function serveDirIndex(
       continue;
     }
     const filePath = join(dirPath, entry.name);
-    const fileUrl = encodeURIComponent(posixJoin(dirUrl, entry.name))
+    const fileUrl = encodeURIComponent(entry.name)
```
jespertheend, line 373: Not sure if this should still use some `@std/path` magic here. I considered using `posixJoin(entry.name)` but that seems redundant.

kt3k (approved):
LGTM. Nice! Thanks

PR #6538, opened 2025-03-30
perf(toml): add `Scanner` `match()` method, capturing numbers and bare key patterns using sticky regexp

timreichen:
This allows for more complex regexps match checks.


toml/_parser.ts
```diff
@@ -123,6 +123,14 @@ export class Scanner {
   isCurrentCharEOL() {
     return this.char() === "\n" || this.slice(0, 2) === "\r\n";
   }
+
+  match(regExp: RegExp) {
+    const match = this.#source.slice(this.#position).match(regExp);
```
kt3k, line 128: `#source.slice(this.#position)` this part creates large substring of source input. This seems defeating the purpose of Scanner class. I think we should keep the existing way of matching against small part of source string.

timreichen: I agree that using `slice()` is not very efficient. I updated the PR so it doesn't slice the string. Could you take another look?

toml/_parser.ts
```diff
@@ -121,6 +121,15 @@ export class Scanner {
...
     return this.#source.startsWith(searchString, this.#position);
   }
+
+  match(regExp: RegExp) {
+    regExp = new RegExp(
+      regExp.source,
+      regExp.sticky ? regExp.flags : regExp.flags + "y",
+    );
+    regExp.lastIndex = this.#position;
+    return this.#source.match(regExp);
```
kt3k, line 131: This matches against (potentially large) `#source` text. On the other hand, the current code usually slices only necessary small text before matching/testing against regexp. I'm not sure this doesn't include any perf degradation.

toml/_parser.ts
```diff
@@ -115,6 +115,15 @@ export class Scanner {
   startsWith(searchString: string) {
     return this.#source.startsWith(searchString, this.#position);
   }
+
+  match(regExp: RegExp) {
+    regExp = new RegExp(
```
kt3k, line 120: Can we give `y` flag to the original regexp instance and avoid creating new regexp each time?

timreichen: Yes we could. I just worry about accidental missing sticky flags which might introduce unintended behavior/bugs and headaches in the future.
How about we add a check just to be on the safe side?
```ts
if (!regExp.sticky) {
  throw new Error(`RegExp ${regExp} does not have a sticky 'y' flag`);
}
```

kt3k: > How about we add a check just to be on the safe side?

Sounds a good idea to me.

kt3k (approved):
LGTM

I also did some benchmark test. Now the entire `toml/parse_test.ts` pass slightly faster than the main with this change. Nice work

```
~/denoland/std $ g co -; hyperfine -w 10 "deno test toml/parse_test.ts"
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
Benchmark 1: deno test toml/parse_test.ts
  Time (mean ± σ):      86.1 ms ±  21.9 ms    [User: 61.8 ms, System: 25.5 ms]
  Range (min … max):    64.9 ms … 179.2 ms    33 runs
 
~/denoland/std $ g co -; hyperfine -w 10 "deno test toml/parse_test.ts"
Switched to branch 'toml-add-Scanner-match-method'
Your branch is up to date with 'timreichen/toml-add-Scanner-match-method'.
Benchmark 1: deno test toml/parse_test.ts
  Time (mean ± σ):      83.3 ms ±  16.8 ms    [User: 61.2 ms, System: 25.4 ms]
  Range (min … max):    67.6 ms … 142.8 ms    32 runs
```

PR #6543, opened 2025-03-31
refactor(testing): add instructions to update snapshots

### LINKED ISSUE #6542: Show how to update snapshots when snapshot assertion failed

WWRS:
closes #6542

I chose to use `[TEST FILE PATH]` instead of adding the file path so that the tests in `snapshot_test.ts.snap` wouldn't be affected by where the test is being run from. It would probably be possible to add the relative path using something like
```ts
PlatformPath.relative(Deno.cwd(), context.origin)
```
though this would require additional testing

### DIFF
--- a/datetime/_date_time_formatter.ts
+++ b/datetime/_date_time_formatter.ts
@@ -295,7 +295,7 @@ export class DateTimeFormatter {
           const value = utc
             ? date.getUTCMilliseconds()
             : date.getMilliseconds();
-          string += digits(value, Number(part.value));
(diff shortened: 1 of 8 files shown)


kt3k:
> I chose to use [TEST FILE PATH] instead of adding the file path so that the tests in snapshot_test.ts.snap wouldn't be affected by where the test is being run from.

That sounds good. I also updated the placeholder to `[files]...` to align it to `deno test --help` output

kt3k (approved):
LGTM. Nice! Thanks

PR #6544, opened 2025-03-31
feat(data-structures/unstable): add `BinarySearchTree` methods `ceiling`, `floor`, `higher`, `lower`

### LINKED ISSUE #3069: add method "higher" "lower" "floor" "ceiling" to BinarySearchTree

https://deno.land/std@0.170.0/collections/binary_search_tree.ts?source

add method "higher" "lower" "floor" "ceiling" to BinarySearchTree

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#ceiling(E)

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#floor(E)

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#higher(E)

https://docs.oracle.com/javase/7/docs/api/java/util/TreeSet.html#lower(E)

https://www.npmjs.com/package/@datastructures-js/binary-search-tree#upperbound-ceil

WWRS:
closes #3069

~~May break custom binary trees that reimplement `BinarySearchTree.#findNode()`. I think it's unlikely that anyone is doing that given that `#findNode()` should work on all child data structures of binary search trees.~~ As pointed out, this is not a problem.

The new methods have the same time complexity as `find()`.


lionel-rowe:
> May break custom binary trees that reimplement `BinarySearchTree.#findNode()`

It's not possible to reimplement `#private` methods in derived classes. Implementing a method with the same name in the derived class creates a new method that's only accessible within the derived class, rather than overriding the method in the parent.

```ts
class A {
    #private() {
        return 'A'
    }

    foo() {
        console.log(this.#private())
    }
}

class B extends A {
    #private() {
        return 'B'
    }

    bar() {
        console.log(this.#private())
    }
}

new A().foo() // logs 'A'
new B().foo() // still logs 'A'
new B().bar() // logs 'B'
```

WWRS:
Ah, I see where I got confused. There's a change to an export in `_binary_search_tree_internals` and that gets used by `RedBlackTree`. I thought this meant `RedBlackTree` used some sort of extension of `internals`, but that's just exported to allow `RedBlackTree` access to those private methods. So the type change cannot cause any problems.

kt3k:
BinarySearchTree is part of stable API and we prefer to not mix stable features and unstable features.

Can you create an unstable version of BinarySearchTree in `unstable_binary_search_tree.ts` (probably by subclassing the stable one) and move the new methods there? (and also can you move the test cases of them in `unstable_binary_search_tree_test.ts`?)

WWRS:
@kt3k Moved to `unstable_binary_search_tree.ts` 👍

kt3k (approved):
Thanks for updating!

LGTM

PR #6545, opened 2025-04-01
chore: remove `lint:import-map` task

iuioiua:
IIUC, workspaces functionality automatically takes care of inter-package imports. See #5746 for context.


iuioiua:
Well, CI's passing 🤷🏾‍♂️

kt3k:
We don't test actual publishing in CI (`deno publish` sometimes fail only without `--dry-run`)

iuioiua:
So should we expect all users of workspaces to maintain their package versions via an import map? Genuine question.

PR #6546, opened 2025-04-01
chore: improve lint rules

iuioiua:
These changes improve lint rules by ensuring that future JSR lint rules, which should apply to this codebase, are enforced. It also removes the need to explicitly include the `verbatim-module-syntax` rule.


kt3k (approved):
LGTM

PR #6547, opened 2025-04-01
chore: rename `assertSnippetsWorks` to `assertHasSnippets`

iuioiua:
This is no longer needed since `deno test --doc` now evaluates code snippets in JSDocs.

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -16,7 +16,6 @@ import {
   type DocNodeClass,
   type DocNodeFunction,
   type DocNodeInterface,
-  type DocNodeModuleDoc,
   type JsDoc,
   type JsDocTagDocRequired,
   type JsDocTagParam,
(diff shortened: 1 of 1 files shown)


_tools/check_docs.ts
```diff
@@ -410,14 +409,6 @@ function assertConstructorDocs(
   }
 }
 
-/**
- * Checks a module document for:
- * - Code snippets that execute successfully.
- */
-function assertModuleDoc(document: DocNodeWithJsDoc<DocNodeModuleDoc>) {
-  assertSnippetsWork(document.jsDoc.doc!, document);
```
kt3k, line 418: Doesn't this still ensure that the module doc has some examples?

iuioiua: You're right. `assertSnippetsWork` was named inaccurately, but I've renamed it now.

kt3k (approved):
LGTM 👍

PR #6548, opened 2025-04-02
feat: add name case checker

timreichen:
This PR adds a tool that checks the name cases for types, interfaces, enums, functions, classes, methods, properties, variables, get accessors and set accessors according to https://docs.deno.com/runtime/contributing/style_guide/#naming-convention

**Note**: There seem to be some occurrences that do not follow the style guide, some of them in stable mods. What to do?

PR #6550, opened 2025-04-04
chore: Deno Style Guide lint plugin

iuioiua:
To get the ball rolling on #6551. Inspired by #6548.

### DIFF
--- a/_tools/lint_plugin.ts
+++ b/_tools/lint_plugin.ts
@@ -13,26 +13,26 @@ export default {
       create(context) {
         return {
           MethodDefinition(node) {
-            if (node.accessibility === "private") {
-              // TODO(iuioiua): add fix
-              context.report({
-                node,
-                range: node.range,
-                message: "Use private field instead",
-                hint: "Use #foo() instead of private foo()",
-              });
-            }
+            if (node.accessibility !== "private") return;
+            // TODO(iuioiua): add fix
+            context.report({
+              node,
+              range: node.range,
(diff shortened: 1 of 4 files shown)


iuioiua:
We're going to have to come up with a couple of solutions if we want to implement this and maintain backward compatibility with Deno v1.

_tools/lint_plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+              });
+            }
+          },
+          PropertyDefinition(node) {
+            if (node.accessibility === "private") {
+              // TODO(iuioiua): add fix
+              context.report({
+                node,
+                range: node.range,
+                message: "Use private field instead",
```
timreichen, line 32: I think the messages and hints are kinda similar rn. Maybe we should write the message something like `Class property uses "private" keyword.` to describe the current state and the hint `Use private field "#${name}" instead.` how to fix it.
Same for the method.

_tools/lint_plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+                hint: "Use #foo() instead of private foo()",
+              });
+            }
+          },
+          PropertyDefinition(node) {
+            if (node.accessibility === "private") {
```
timreichen, line 27: Maybe we can do an early return if `node.accessibility !== "private"` to reduce nesting.

kt3k (approved):
LGTM. Nice!

PR #6553, opened 2025-04-06
refactor(assert,cbor,collections,fs,toml,uuid): align error messages to the style guide, add `error-message` Deno Style Guide linter plugin

timreichen:
Part of https://github.com/denoland/std/issues/6551

**Note**: This plugin does not check for 3, 4, 5, 7 and 8 in the [Style Guide](https://docs.deno.com/runtime/contributing/style_guide/) for error messages as they would need language analysis.


kt3k:
Can you exclude the test only error messages from the check target?

timreichen:
> Can you exclude the test only error messages from the check target?

Done. I added single line ignores because most of these tests only cases also should follow the style guide imo but can make the switch at a later point.

kt3k:
> most of these tests only cases also should follow the style guide imo

I disagree with that. That sounds unnecessarily strict rule. Also the style guide says it's rule for `User-facing error messages`

timreichen:
> > most of these tests only cases also should follow the style guide imo
> 
> 
> 
> I disagree with that. That sounds unnecessarily strict rule. Also the style guide says it's rule for `User-facing error messages`

I think it is preferable to have a code base that follows the all the rules whenever possible, also for tests. There are only some exceptions in this particular case of error messages that might be justified to ignore lint rules, like specifically testing lower case errors for example. If the test works the same with a lowercase or uppercase message, there is no reason to not follow the rule imo. Else it is just an abuse of the lint ignore flag, no?

kt3k:
It makes a contribution barrier and makes it unpleasant to contribute to the project to enforce unnecessary rules. Things like `new Error("boom")` should be allowed in testing without deno-lint-ignore directive in my opinion.

We can have access to file name via `context.filename`. I think we can exclude test files using it

timreichen:
> It makes a contribution barrier and makes it unpleasant to contribute to the project to enforce unnecessary rules. Things like `new Error("boom")` should be allowed in testing without deno-lint-ignore directive in my opinion.

I am not sure which rules are considered necessary and which not but we also enforce other lint rules in test files, so I don't know why we would suspend some and enforce others.

I would not call enforcing lint rules a contribution barrier. A warning can easily be ignored with s comment or auto-fixed.

Wouldn't it be rather confusing if a contributor has to handle different rule sets for an api and tests?

iuioiua:
I'd lean towards not being overly strict too.

timreichen:
> I'd lean towards not being overly strict too.

But why hardcode ignoring files when one can just add an ignore file comment? This doesn't make sense to me.

kt3k:
> But why hardcode ignoring files when one can just add an ignore file comment?

I don't think we need to hardcode ignore files, but we can detect user-facing errors by using deno_graph + exports fields of each deno.json file.

_tools/lint_plugin.ts
```diff
@@ -38,5 +43,67 @@ export default {
...
+          NewExpression(node) {
+            if (node.callee.type !== "Identifier") return;
+            const name = node.callee.name;
+            if (!name.endsWith("Error")) return;
+            const argument = node.arguments[0];
+            if (argument?.type !== "Literal") return;
+            const value = argument.value;
+            if (typeof value !== "string") return;
+
+            if (value.match(LEADING_LOWERCASE_REGEXP)) {
```
iuioiua, line 59:
```suggestion
            if (value[0] === value[0].toLowerCase()) {
```

_tools/lint_plugin.ts
```diff
@@ -38,5 +43,67 @@ export default {
...
+                fix(fixer) {
+                  const newValue = argument.raw.at(0) +
+                    value[0].toUpperCase() +
+                    value.slice(1) +
+                    argument.raw.at(-1);
+                  return fixer.replaceText(argument, newValue);
+                },
+              });
+            }
+            if (value.match(TRAILING_PERIOD_REGEXP)) {
```
iuioiua, line 74:
```suggestion
            if (value.endsWith(".")) {
```

_tools/lint_plugin.ts
```diff
@@ -38,5 +43,67 @@ export default {
...
+                  "Remove the period at the end of the error message. See https://docs.deno.com/runtime/contributing/style_guide/#error-messages for more details.",
+                fix(fixer) {
+                  const newValue = argument.raw.at(0) +
+                    value.slice(0, -1) +
+                    argument.raw.at(-1);
+                  return fixer.replaceText(argument, newValue);
+                },
+              });
+            }
+            if (value.match(PERIOD_REGEXP)) {
```
iuioiua, line 88:
```suggestion
            if (value.includes(".")) {
```

_tools/lint_plugin.ts
```diff
@@ -6,6 +6,11 @@
  * {@link https://docs.deno.com/runtime/contributing/style_guide/ | Deno Style Guide}
  */
 
+const PERIOD_REGEXP = /\.\s/;
+const TRAILING_PERIOD_REGEXP = /\.$/;
+const LEADING_LOWERCASE_REGEXP = /^[a-z]/;
+const CONTRACTION_REGEXP = /\S'\S/;
```
iuioiua, line 12: Hint: try avoiding regex as much as possible. They're complicated.

iuioiua (changes requested):

_tools/lint_plugin.ts
kt3k: I wonder if there's any case where `.` character is used for other purpose e.g. `obj.prop`, `1.5`, etc

_tools/lint_plugin.ts
kt3k: I wonder if there's any case where `.` character is used for other purpose e.g. `obj.prop`, `1.5`, etc

_tools/lint_plugin.ts
timreichen: This would return a false positive if it is not a letter char e.g. `"@" === "@".toLowerCase()`

_tools/lint_plugin.ts
timreichen: I implemented it as `value.includes(". ")` with a whitespace.

_tools/lint_plugin.ts
timreichen: We could to `value[0] !== value[0].toUpperCase()` but I don't know if this would bring unintentional behaviors for special chars. WDYT?

iuioiua: Try that 🤷🏾

deno.json
```diff
@@ -43,6 +43,7 @@
     "docs"
   ],
   "lint": {
+    "plugins": ["./_tools/error_message_lint_plugin.ts"],
```
iuioiua, line 46: Needs reverting.

iuioiua (approved):
Just nits. Otherwise, LGTM!

crypto/crypto.ts
```diff
@@ -244,7 +244,7 @@ const stdCrypto: StdCrypto = ((x) => x)({
           return context.digestAndDrop(length).buffer as ArrayBuffer;
         } else {
           throw new TypeError(
-            "data must be a BufferSource or [Async]Iterable<BufferSource>",
+            "Data must be a BufferSource or [Async]Iterable<BufferSource>",
```
kt3k, line 247: I don't think this should be capitalized as this matches the variable name.

kt3k:
It looks like we consistently use `.` in AssertionError message. I think we probably should also exclude `AssertionError` for the period check.

kt3k (approved):
LGTM

PR #6555, opened 2025-04-07
refactor(toml): rewrite `unflat()`

timreichen:


toml/_parser.ts
```diff
@@ -134,20 +134,13 @@ function failure(): Failure {
...
-  const out: Record<string, unknown> = {};
-  if (keys.length === 0) {
-    return cObj as Record<string, unknown>;
-  }
-  if (!cObj) cObj = values;
-  const key: string | undefined = keys[keys.length - 1];
-  if (typeof key === "string") out[key] = cObj;
-  return unflat(keys.slice(0, -1), values, out);
+  values: T,
+): T {
```
kt3k, line 140: I think the return type should be `Record<string, unknown>` instead of `T extends unknown[] | Record<string, unknown>`. Updating

kt3k (approved):
LGTM

PR #6557, opened 2025-04-08
docs(encoding): fixed for calcSize functions

BlackAsLight:
This pull request fixed the docs for calcSizeBase64, 32 and hex as they are incorrectly still mentioning the Raw instead of Into name.


kt3k (approved):
LGTM. Thanks!

Issue #6551, opened 2025-04-04
suggestion: Deno Style Guide lint plugin


iuioiua:
I suggest we create a lint plugin to enforce the rules of the [Deno Style Guide](https://docs.deno.com/runtime/contributing/style_guide/). `std` can be the project used to quickly iterate on this until maturity. Then, we could create a `@deno/style-guide-lint-plugin` package for internal and user Deno projects to use.

The following rules can each correspond to a lint rule:
- [ ] Exported functions: max 2 args, put the rest into an options object
- [ ] Export all interfaces that are used as parameters to an exported member
- [ ] Top-level functions should not use arrow syntax (?)
- [ ] Error Messages
- [ ] Do not depend on external code
- [ ] Prefer # over private keyword (#6550)
- [ ] Naming convention

timreichen:
Partial naming conventions are covered by built in `camelcase`, however it does not work for
- PascalCase functions e.g. `function GenerateKey() {}`
- PascalCase variables e.g. `const KeyPattern = /^[0-9a-f]+$/`
- camelCase classes e.g. `class keyObject {}`
- camelCase type e.g. `type sharedKey = {}`
- camelCase interface e.g. `interface foo {}`
- camelCase enums `enum keyType { publicKey, privateKey }` 

Maybe this could be added to `camelcase`?

Possible addition to the style guide:
params in camelCase e.g. good: `function foo(bar: boolean)  {}` bad: `function foo(Bar: boolean)  {}`
property names in camelCase e.g. good: `const foo = { bar: true }` bad: `const foo = { Bar: true }`
**Note**: This would need triage. There might be exceptions, like classes: `const classes = { ClassA, ClassB } `

timreichen:
There seems to be a problem with lint plugin paths and `workspaces` in `@std`.  I've tried adding plugins myself locally and couldn't make it work. Is this a deno bug?
Error: https://github.com/denoland/std/actions/runs/14261545183/job/39974079927?pr=6550#annotation:5:7

iuioiua:
It's been fixed in canary. See https://github.com/denoland/deno/issues/28747

PR #6558, opened 2025-04-08
chore: add `no-top-level-arrow-syntax` rule to Deno Style Guide linter plugin

iuioiua:
Towards #6551

Blocked by https://github.com/denoland/deno/issues/28799


iuioiua:
This is only currently working on canary

iuioiua:
Does this rule include non-exported arrow functions? The [example snippet](https://docs.deno.com/runtime/contributing/style_guide/#top-level-functions-should-not-use-arrow-syntax) only exemplifies exported arrow functions.

iuioiua:
TODO: It'd be more efficient if we instead entered through `Program` or `ExportNamedDeclaration` instead of `ArrowFunctionExpression`.

kt3k (approved):
LGTM. Nice!

kt3k:
> Does this rule include non-exported arrow functions? The [example snippet](https://docs.deno.com/runtime/contributing/style_guide/#top-level-functions-should-not-use-arrow-syntax) only exemplifies exported arrow functions.

I think this applies to non exported functions as well.

PR #6559, opened 2025-04-08
chore: add `naming-convention` Deno Style Guide linter plugin

timreichen:
Towards https://github.com/denoland/std/issues/6551

### DIFF
--- a/_tools/lint_plugin.ts
+++ b/_tools/lint_plugin.ts
@@ -1,10 +1,26 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
-// @ts-nocheck Deno.lint namespace does not pass type checking in Deno 1.x
 
 /**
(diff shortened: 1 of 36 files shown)


kt3k:
Loosened variable rule as a variable can be a class. The declaration like `const CustomClass = class {};` should be allowed in my view.

Also allowed trailing `_` in names. It seems like a common pattern in std.

kt3k (approved):
LGTM

timreichen:
> Loosened variable rule as a variable can be a class. The declaration like `const CustomClass = class {};` should be allowed in my view.
> 
> 
> 
> Also allowed trailing `_` in names. It seems like a common pattern in std.

Isn't that now how the `camelcase` lint rule works?

kt3k:
Looks similar, but I'm not entirely sure

PR #6561, opened 2025-04-08
chore: add `exported-function-args-maximum` Deno Style Guide linter plugin

timreichen:
Towards https://github.com/denoland/std/issues/6551


_tools/lint_plugin.ts
```diff
@@ -38,5 +38,56 @@ export default {
...
+            switch (params.length) {
+              case 0:
+              case 1:
+              case 2:
+                break;
+              case 3: {
+                const param = params.at(-1)!;
+                switch (param.type) {
+                  case "Identifier":
+                    if (param.name === "options") return;
```
iuioiua, line 59: I don't think the parameter has to be called `options`... Maybe it just has to be an object.

assert/almost_equals.ts
```diff
@@ -29,6 +29,7 @@ import { AssertionError } from "./assertion_error.ts";
  * default is one hundred thousandth of a percent of the expected value.
  * @param msg The optional message to include in the error.
  */
+// deno-lint-ignore deno-style-guide/exported-function-args-maximum
```
iuioiua, line 32: Sidenote: bums me out a little that we don't follow this rule in `@std/assert` 😝

async/_util.ts
```diff
@@ -1,6 +1,7 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
 
+// deno-lint-ignore deno-style-guide/exported-function-args-maximum
```
iuioiua, line 4: Non-blocker: we should have this function just accept an object.

iuioiua (commented):
IMO, we should review all the instances where this rule is ignored. It's a bit of a shame that we ignore this rule in some parts of the codebase. It's a very reasonable rule and not adhering to it could pose a few little backwards compatibility challenges in the future.

toml/_parser.ts
```diff
@@ -134,6 +134,7 @@ function failure(): Failure {
   return { ok: false };
 }
 
+// deno-lint-ignore deno-style-guide/exported-function-args-maximum
 export function unflat(
```
kt3k, line 138: This function is only exported as internal util and not part of public API. Can we exclude this from lint target?

kt3k:
> A max of two arguments seems kind of arbitrary and I think we may find many instances where it doesn't make sense

Note this rule only applies to the public APIs (though this lint rule checks more than that). Internal utilities don't need to follow this.

The rule exists from the beginning of the project, and the maintainers keep checking the compliance to it (for example, see this discussion https://github.com/denoland/std/pull/4651#issuecomment-2089635838 ). I think it's reasonable to keep this rule.

kt3k:
- The rule allows the function in last position. The error on `pooledMap` looks a false positive.
- I think we should exclude files starting with `_`. They are usually internal utils and errors on them should be false positives.
- I think we should exclude `@std/bytes`. The usages of number as the 3rd arg are intentional (discussed in https://github.com/denoland/std/pull/4651 )
- `@std/internal` should be excluded as they are not part of public API
- I think we should also exclude functions starting with `assert`. They generally don't follow this rule.
- `deleteCookie` follows the rule, but it seems getting the error

Only the items below look like truly exceptions to this rule to me:
- `reduceGroups`
- `runningReduce`
- `encodeIntoBase32`
- `encodeIntoBase64`
- `encodeVarint`
- `chown`
- `chownSync`
- `utime`
- `utimeSync`
- `toTransformStream`
- `stub`
- `createCapture`
- `resliceBufferWithPadding`
- `createTextureWithData`

iuioiua:
> * I think we should exclude files starting with `_`. They are usually internal utils and errors on them should be false positives.

> * `@std/internal` should be excluded as they are not part of public API

I don't think this rule should exclusively apply to the public API. The maintainer can benefit a little by having these symbols accept objects by having removing argument position as a pain point. The changes to do this are trivial too.

> * I think we should exclude `@std/bytes`. The usages of number as the 3rd arg are intentional (discussed in [feat(bytes): `@std/bytes@1.0.0-rc.1` #4651](https://github.com/denoland/std/pull/4651) )

We still need to update the style guide.

kt3k:
> > * I think we should exclude files starting with `_`. They are usually internal utils and errors on them should be false positives.
> 
> > * `@std/internal` should be excluded as they are not part of public API
> 
> I don't think this rule should exclusively apply to the public API. The maintainer can benefit a little by having these symbols accept objects by having removing argument position as a pain point. The changes to do this are trivial too.

This rule is very uncommon and restrictive in my view. I guess many contributors would feel uncomfortable if this rule is enforced everywhere. (See @BlackAsLight's [comment](https://github.com/denoland/std/pull/6561#issuecomment-2789319242) for example)

> > * I think we should exclude `@std/bytes`. The usages of number as the 3rd arg are intentional (discussed in [feat(bytes): `@std/bytes@1.0.0-rc.1` #4651](https://github.com/denoland/std/pull/4651) )
> 
> We still need to update the style guide.

That exception is already covered by the 2nd paragraph of the 2nd point.

> An optional parameter that's not in an options object might be acceptable if there is only one, and it seems inconceivable that we would add more optional parameters in the future.

BlackAsLight:
> I don't think this rule should exclusively apply to the public API. The maintainer can benefit a little by having these symbols accept objects by having removing argument position as a pain point. The changes to do this are trivial too.

In this example here, https://github.com/denoland/std/blob/main/encoding/_common64.ts#L54-L60, I'd find it more painful to have to needlessly put it in an object than write it in a particular order.

timreichen:
> > I don't think this rule should exclusively apply to the public API. The maintainer can benefit a little by having these symbols accept objects by having removing argument position as a pain point. The changes to do this are trivial too.
> 
> In this example here, https://github.com/denoland/std/blob/main/encoding/_common64.ts#L54-L60, I'd find it more painful to have to needlessly put it in an object than write it in a particular order.

To be honest I think the rule is great especially for functions like the one you pointed out. I have no idea what the arguments mean without getting into the meat of the function. JSDocs are only required for public apis, so this rule actually helps to understand the internals too.

iuioiua:
> > > I don't think this rule should exclusively apply to the public API. The maintainer can benefit a little by having these symbols accept objects by having removing argument position as a pain point. The changes to do this are trivial too.
> 
> > 
> 
> > In this example here, https://github.com/denoland/std/blob/main/encoding/_common64.ts#L54-L60, I'd find it more painful to have to needlessly put it in an object than write it in a particular order.
> 
> 
> 
> To be honest I think the rule is great especially for functions like the one you pointed out. I have no idea what the arguments mean without getting into the meat of the function. JSDocs are only required for public apis, so this rule actually helps to understand the internals too.

I agree.

_tools/lint_plugin.ts
```diff
@@ -310,5 +310,45 @@ export default {
...
     },
+    // https://docs.deno.com/runtime/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object
+    "exported-function-args-maximum": {
+      create(context) {
+        return {
+          ExportNamedDeclaration(node) {
+            const declaration = node.declaration;
+            if (declaration?.type !== "FunctionDeclaration") return;
+            const params = declaration.params;
+            const id = declaration.id;
```
iuioiua, line 321: Nit
```suggestion
            if (node.declaration?.type !== "FunctionDeclaration") return;
            const { params, id } = node.declaration;
```

iuioiua (approved):
LGTM, barring one nit.

_tools/lint_plugin.ts
```diff
@@ -310,5 +312,58 @@ export default {
         };
       },
     },
+    // https://docs.deno.com/runtime/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object
+    "exported-function-args-maximum": {
+      create(context) {
+        const url = toFileUrl(resolve(context.filename));
```
iuioiua, line 318: Might `import.meta.resolve()` work here?

kt3k: `context.filename` is `main.ts` in test cases, and that throws with `import.meta.resolve` (`main.ts` is not a valid specifier, but a valid file path)

kt3k (approved):
excluded some exceptional packages (e.g. `assert`, `internal`, etc) from the lint target.

LGTM

Issue #655, opened 2020-08-19
Importing @typescript-eslint/parser and std/hash/mod.ts exposes a memory leak


jeroenptrs:
I am running this exact script: `deno cache deps.ts --unstable --lock=lock.json --lock-write` with deps.ts being the following file:

```ts
// deps.ts
export { parse } from "https://jspm.dev/npm:@typescript-eslint/parser@3.7.1";
export * as hash from "https://deno.land/std@0.65.0/hash/mod.ts";
```

This causes the following error after a short amount of time (on macOS catalina):

```sh
<--- Last few GCs --->

[5203:0x72b00000000]    34568 ms: Scavenge 1223.4 (1426.4) -> 1223.4 (1426.4) MB, 3.8 / 0.0 ms  (average mu = 0.991, current mu = 0.972) allocation failure 
[5203:0x72b00000000]    34580 ms: Scavenge 1223.4 (1426.4) -> 1222.0 (1442.4) MB, 12.0 / 0.0 ms  (average mu = 0.991, current mu = 0.972) allocation failure 
[5203:0x72b00000000]    34710 ms: Mark-sweep 1237.4 (1442.4) -> 1235.2 (1447.9) MB, 112.8 / 0.0 ms  (average mu = 0.970, current mu = 0.645) allocation failure scavenge might not succeed


<--- JS stacktrace --->


#
# Fatal javascript OOM in MarkCompactCollector: young object promotion failed
#
```

Note that I've been running either dependency separately as well and have not seen this behaviour, only when I import both it panics.

iuioiua:
Hi @jeroenptrs, I just tested this and didn't run into any issues. Are you still having issues?

iuioiua:
@kt3k, I believe this issue can be closed as it is stale.

PR #6562, opened 2025-04-09
chore: add `no-external-code` rule to Deno Style Guide lint plugin

iuioiua:
Towards #6551


_tools/lint_plugin.ts
```diff
@@ -38,5 +38,32 @@ export default {
...
+        return {
+          ImportDeclaration(node) {
+            const resolvedSpecifier = import.meta.resolve(node.source.value);
+            if (
+              resolvedSpecifier.startsWith("file:") ||
+              resolvedSpecifier.startsWith("jsr:@std") ||
+              resolvedSpecifier.startsWith("jsr:/@std") ||
+              resolvedSpecifier.startsWith("jsr:@deno") ||
+              resolvedSpecifier.startsWith("jsr:/@deno") ||
+              resolvedSpecifier.startsWith("node:")
```
iuioiua, line 53: I've extended this rule so that it makes more sense when used within other Deno codebases to also include `@deno`-scoped and Node imports. If this seems reasonable, I think we should also rewrite [the rule](https://docs.deno.com/runtime/contributing/style_guide/#do-not-depend-on-external-code.) to match. What do we think?

kt3k: It makes sense to include `node:` import as it's builtin to Deno and also it's already used a lot in `@std/fs`. I don't think std should rely on `@deno` namespace, especially in published packages. I think most `@deno`-scoped packages are used in `_tools`. How about excluding `_tools` scripts from lint target?

_tools/lint_plugin.ts
```diff
@@ -59,6 +59,10 @@ export default {
     // https://docs.deno.com/runtime/contributing/style_guide/#do-not-depend-on-external-code.
     "no-external-code": {
       create(context) {
+        if (context.filename.includes("_tools")) {
+          // Tools are allowed to use external code
+          return {};
+        }
```
iuioiua, line 65: I used ignore lint directives instead of plugin logic so that this plugin can be used for other codebases, as per the original intention of #655. Either way, we can address this if that ends up happening.

kt3k: Isn't [this rule](https://docs.deno.com/runtime/contributing/style_guide/#std) specific to std? I guess this rule only makes sense as private rule here

PR #6566, opened 2025-04-10
refactor(datetime): Move behavior of DateTimeFormatter to standalone functions

WWRS:
Improves test coverage.

The functions are almost entirely unchanged, with the exception of some renaming.

These functions should not be used outside tests, and DateTimeFormatter itself is an internal class.

TODO (I will look into these):
- Refactor errors to consistently provide useful feedback
- Add support for more format strings: Goal is parity with Intl.DateTimeFormat
- Add support for time zones (this will improve coverage in this file to almost 100%)


lunadogbot (approved):
Pure refactor of the private `_date_time_formatter.ts`. `DateTimeFormatter` is preserved as a thin wrapper around the new standalone `formatStringToFormatParts`/`formatDate`/`dateStringToDateTimeFormatParts`/`dateTimeFormatPartsToDate`, so `datetime/format.ts` and `datetime/parse.ts` keep working unchanged. Splitting `formatToFormatParts` into a per-symbol lookup (`symbolToFormatPart`) + a single push/advance loop removes ~17 duplicated `formatParts.push(...); index += symbol.length; continue;` blocks. The `sortDateTimeFormatParts` switch to `readonly` + spread-copy fixes a real input-mutation bug — the old code spliced the caller's array.

`case "a"` in `symbolToFormatPart` switches from `{ type: "dayPeriod", value: 1 }` to `{ type: "dayPeriod", value: "short" }`. `formatDate` never reads `part.value` on the `dayPeriod` branch (it derives AM/PM directly from `date.getHours()`), so this is a cosmetic fix, not a behaviour change.

CI is green across the matrix (the one PENDING entry is an unnamed check-suite stub; `license/cla` is success).

- nit: `formatDate`'s `formatParts: FormatPart[]` and `dateStringToDateTimeFormatParts`'s `formatParts: FormatPart[]` params could be `readonly FormatPart[]` to match the new `sortDateTimeFormatParts` signature — neither function mutates the array.

PR #6569, opened 2025-04-11
docs(path): add many more examples to `mod.ts`

johnpyp:
This PR adds many more examples to `path`'s front-page `mod.ts`.

Each example uses `assertEquals` to ensure correctness and testability.


kt3k (approved):
Great improvement for better visibility of the features in JSR module page! Thanks for your contribution

PR #6570, opened 2025-04-11
docs(encoding): add many more examples to `mod.ts`

johnpyp:
This PR adds many more examples to `encoding`'s front-page `mod.ts`.

Each example uses `assertEquals` to ensure correctness and testability.


kt3k (approved):
LGTM. Nice! Thanks

PR #6572, opened 2025-04-11
fix(ini): fix quoted value bug

timreichen:
This PR fixes a bug where the value `value="` was trimmed. This was due to `startsWith()` and `endsWith()` both being true with a value of `"`.


kt3k (approved):
LGTM. Nice!

PR #6573, opened 2025-04-14
BREAKING(text/unstable): fix dedent export path

kt3k:


iuioiua:
Does `deno publish` catch this error?

kt3k:
> Does deno publish catch this error?

No, it doesn't. This path is already published https://jsr.io/@std/text/doc/unstable_dedent

bartlomieju (approved):

PR #6579, opened 2025-04-14
chore: add new internal utilities and simplify `lint:docs` task script

iuioiua:
This PR adds `getPackagesDenoJsons()` and `getEntrypoints()` to `_tools/utils.ts` and changes the internal tools to use these functions, simplifying them. This also includes a significant simplification of `_tools/check_unstable_deps.ts` and making it run ~2x (0.8s) faster.

There are more simplifications that can be made to these internal tools.


kt3k (approved):
LGTM. Nice clean ups!

kt3k:
Note: `lint:circular` is slower now because it now checks all the entrypoints instead of selected ones, but this seems fine to me as this should be easier to maintain and the check itself is more robust (now less chance of human errors).

PR #6581, opened 2025-04-15
chore: add `--doc` flag to `test` task

timreichen:


iuioiua (changes requested):
Perhaps you can just add `--doc` to the `test` task.

timreichen:
> Perhaps you can just add `--doc` to the `test` task.

I tried that but it complains that it is set more than once for workflows.

kt3k (approved):
LGTM. Thanks

PR #6582, opened 2025-04-15
refactor(toml): fix object check in 'merge' util

timreichen:
Pretty sure this check is meant for `record`, not `body`.


kt3k:
'merge' seems used only once for `merge(repeat(pair))` call. There's probably no actual case which is fixed by this change. I changed the commit type to 'refactor' (because it (probably) doesn't fix a user-facing bug)

kt3k (approved):
LGTM

PR #6584, opened 2025-04-15
refactor(toml): remove `Scanner` test

timreichen:
These test cases are covered by other tests.


kt3k:
They look good for demonstrating what `Scanner` does as test cases. I'm not in favor of removing them

See also https://www.thecoder.cafe/p/unit-tests-as-documentation

PR #6585, opened 2025-04-15
fix(toml): fix empty string handling

timreichen:
The current toml implementation throws when trying to parse an empty string.
(Probably) all toml npm implementations allow empty strings and return an empty object, so we should probably too.
```ts
import toml from "npm:toml";
assertEquals(toml.parse(""), {});

import tomljs from "npm:toml-js";
assertEquals(tomljs.parse(""), {});

import * as s from "npm:smol-toml";
assertEquals(s.parse(""), {});

import * as i from "npm:@iarna/toml";
assertEquals(i.parse(""), {});

import * as c from "npm:confbox";
assertEquals(c.parseTOML(""), {});
```

### DIFF
--- a/toml/_parser.ts
+++ b/toml/_parser.ts
@@ -776,8 +776,8 @@ export function toml(
   scanner: Scanner,
 ): ParseResult<Record<string, unknown>> {
   const blocks = repeat(or([block, tableArray, table]))(scanner);
-  if (!blocks.ok) return failure();
   let body = {};
+  if (!blocks.ok) return success(body);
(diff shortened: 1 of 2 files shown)


kt3k (approved):
It looks like empty string should be mapped to `{}` according to the spec maintainer. https://github.com/toml-lang/toml/issues/607

LGTM

I don't think this is a breaking change. I changed the commit type to 'fix'. (A breaking change is an intentional design change, which breaks the users' code. The existing behavior is unintentional bug.)

PR #6587, opened 2025-04-15
fix(toml): fix empty inline table bug

timreichen:
This PR fixes https://github.com/denoland/std/issues/6568

### DIFF
--- a/toml/_parser.ts
+++ b/toml/_parser.ts
@@ -205,6 +205,28 @@ function join<T>(
 ): ParserComponent<T[]> {
   const Separator = character(separator);
   return (scanner: Scanner): ParseResult<T[]> => {
+    const out: T[] = [];
+    const first = parser(scanner);
+    if (!first.ok) return success(out);
+    out.push(first.body);
+    while (!scanner.eof()) {
+      if (!Separator(scanner).ok) break;
+      const result = parser(scanner);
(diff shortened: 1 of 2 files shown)


kt3k:
It feels a bit confusing to have both `join` and `join1` parsers. How about changing the initial condition of `inlineTable` parser to recognize `whitespaces` + `}` pattern?

timreichen:
> It feels a bit confusing to have both `join` and `join1` parsers. How about changing the initial condition of `inlineTable` parser to recognize `whitespaces` + `}` pattern?

I think how it is structured now is right, without special treatment. The x/x1 is borrowed from `many()` and `many1()` in [reparse](https://www.npmjs.com/package/reparse)](https://www.npmjs.com/package/reparse) or [arcsecond](https://www.npmjs.com/package/arcsecond/v/1.0.1). However I could inline the code, so that there are not two different functions with similar names.

kt3k:
> The x/x1 is borrowed from many() and many1() in [reparse](https://www.npmjs.com/package/reparse) or [arcsecond](https://www.npmjs.com/package/arcsecond/v/1.0.1).

Ok. There seem some usages of this naming convention. I added the docs of them

kt3k (approved):
LGTM

PR #6588, opened 2025-04-15
refactor(yaml): add `skipComment()` method

timreichen:


yaml/_loader_state.ts
```diff
@@ -1640,12 +1645,8 @@ export class LoaderState {
           ch = this.next();
         }
 
-        if (ch === SHARP) {
-          do {
-            ch = this.next();
-          } while (ch !== 0 && !isEOL(ch));
-          break;
```
kt3k, line 1647: We now miss this break path. Is this change intentional?

timreichen: This is intentional. A yaml comment ends only on `eol` or `eof`. The yaml string will be invalid on `eof` at that stage and `eol` is handled by the next line `if (isEOL(ch)) break;`.

kt3k (approved):
LGTM

PR #6589, opened 2025-04-15
refactor(yaml): refactor directive handler args

timreichen:
This PR removes the spread args in `yamlDirectiveHandler()` and `tagDirectiveHandler()`.


kt3k:
I prefer the existing implementation. The validation of arg count and actual handlings are defined in single place with the current code. Now they are spread into 2 places with this change

PR #6593, opened 2025-04-16
refactor(yaml): remove `throwError()` method

timreichen:
This PR removes the `throwError()` and throws errors directly. This makes the code easier to read.


kt3k (approved):
> This makes the code easier to read.

Agree.

LGTM

PR #6594, opened 2025-04-16
refactor(yaml): declare-on-use variables

timreichen:
It seems this doesn't really affect performance but is more readable.


kt3k (approved):
LGTM. Thanks!

PR #6596, opened 2025-04-16
test(fmt): reduce unnecessary output during test run

timreichen:
This PR replaces `spy()` call with `stub()` call in order to avoid test cli output pollution.


kt3k:
The existing assertion seems more elegant to me. Do we need to care about console output during test cases?

timreichen:
> The existing assertion seems more elegant to me. Do we need to care about console output during test cases?

Imo yes if we can avoid it.

kt3k:
Updated to use `stub` with spied implementation, keeping the original assertions

kt3k (approved):
LGTM

PR #6605, opened 2025-04-18
fix(toml): fix handling of leading and trailing underscores in number literals

timreichen:
This PR fixes https://github.com/denoland/std/issues/6604

Refs https://toml.io/en/v1.0.0#integer, https://github.com/toml-lang/toml/issues/517

### DIFF
--- a/toml/_parser.ts
+++ b/toml/_parser.ts
@@ -235,6 +235,7 @@ function kv<T>(
 ): ParserComponent<{ [key: string]: unknown }> {
   const Separator = character(separator);
   return (scanner: Scanner): ParseResult<{ [key: string]: unknown }> => {
+    const position = scanner.position;
     const key = keyParser(scanner);
     if (!key.ok) return failure();
     const sep = Separator(scanner);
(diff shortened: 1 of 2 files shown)


kt3k:
Looks better than wrong parsing, but the error message doesn't look great to me. `parse("a = 0x_12")` throws with `Parse error on line 1, column 9: key/value pair doesn't have "="`, but the `=` character does exist.

`smol-toml` throws with more sensible error message:
```
Invalid TOML document: invalid value

1:  a = 0x_12
        ^
```

timreichen:
> Looks better than wrong parsing, but the error message doesn't look great to me. `parse("a = 0x_12")` throws with `Parse error on line 1, column 9: key/value pair doesn't have "="`, but the `=` character does exist.
> 
> `smol-toml` throws with more sensible error message:
> 
> ```
> Invalid TOML document: invalid value
> 
> 1:  a = 0x_12
>         ^
> ```

I think we should go over error handing in general after some more cleaning ups. I've updated the error handling for this particular case as an example. Since this is based on parser combinators, error handling could also be abstracted.

kt3k (approved):
Nice fix! Thanks

PR #6608, opened 2025-04-20
BREAKING(encoding/unstable): base64/32 functions to match proposal-arraybuffer-base64 API

BlackAsLight:
Closes: https://github.com/denoland/std/issues/6601

Looking at the documentation for the proposal on MDN, it has an `omitPadding` option. Testing it out on Firefox, it seems to always add padding, even for base64url. Would it be better to also implement this option here or stick with current behaviour of always adding padding for base64 and always omitting padding for base64url? 
 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/toBase64

### DIFF
--- a/encoding/_common32.ts
+++ b/encoding/_common32.ts
@@ -4,27 +4,35 @@ import type { Uint8Array_ } from "./_types.ts";
 export type { Uint8Array_ };
 
 export const padding = "=".charCodeAt(0);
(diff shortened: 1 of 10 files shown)


kt3k (approved):
LGTM. Nice! Thanks

PR #6610, opened 2025-04-21
fix(http): don't set the Date header in file_server.ts responses (use the default Date header value)

### LINKED ISSUE #6609: http/file_server: Wrong Date header
**Describe the bug**

I noticed in Chrome DevTools that some requests served were not being cached correctly. The browser would often request files (and receive a 304 response) for files I hadn't touched in months.
I'm setting `Cache-Control: public, max-age=2592000` headers on these requests, which should cache responses for up to one month.

The `Date` header should represent the current date of the system. But file_server.ts sets the `Date` header to the last modified date of the file:
https://github.com/denoland/std/blob/390845e7ecc351ee5008fe5184e6256407a52793/http/file_server.ts#L205-L208

Actually I'm not 100% sure what it should be, [the spec is rather vague](https://httpwg.org/specs/rfc9110.html#rfc.section.6.6.1):

> The "Date" header field represents the date and time at which the message was originated

But judging from how nginx handles Date headers, I'm fairly certain it's supposed to be the current date, not the date of the file.

**Steps to Reproduce**

1. Run `deno run -A https://jsr.io/@std/http/1.0.14/file_server.ts`
2. Visit http://0.0.0.0:8000/ in your browser and click a random file
3. Open DevTools and reload
4. Observe the `Date` header in the network panel

**Expected behavior**

The Date is the current date

**Environment**

- OS: macOS 14.4.1 (23E224)
- deno version: 2.1.6
- std version: `http/1.0.14`

jespertheend:
Fixes #6609

`Deno.serve()` already sets the `Date` header on responses. So simply removing these lines should fix this.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -202,11 +202,6 @@ export async function serveFile(
 
   const headers = createBaseHeaders();
 
-  // Set date header if access timestamp is available
-  if (fileInfo.atime) {
-    headers.set(HEADER.Date, fileInfo.atime.toUTCString());
-  }
-
   const etag = fileInfo.mtime
     ? await eTag(fileInfo, { algorithm })
     : await HASHED_DENO_DEPLOYMENT_ID;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks for catching this

Context: See the discussion in https://github.com/denoland/std/issues/6609 for the reasoning

PR #6611, opened 2025-04-22
chore(crypto): update `crypto/_wasm` and `wasmbuild` task

iuioiua:


kt3k:
There are 2 files `crypto/_wasm/lib/deno_std_wasm_crypto.mjs` (this one is new) and `crypto/_wasm/lib/deno_std_wasm_crypto.generated.mjs` which both seem including wasm binary. Are they both necessary, or can we remove one of them?

iuioiua:
@dsherret, am I right in thinking that the currently failing tests are no longer needed and that at most we can just assert the computed hashes?

crypto/testdata/digest_large_inputs.ts
iuioiua, line 1: Is this the right way to do this?

crypto/testdata/digest_many_calls.ts
iuioiua, line 1: Ditto.

iuioiua:
> There are 2 files `crypto/_wasm/lib/deno_std_wasm_crypto.mjs` (this one is new) and `crypto/_wasm/lib/deno_std_wasm_crypto.generated.mjs` which both seem including wasm binary. Are they both necessary, or can we remove one of them?

I've removed the `*.generated.*` files now as they're not needed.

kt3k (approved):
LGTM

imcotton:
Would anytime soon to importing this crypto WASM module as in[^1]? Thus could be used in Cloudflare Workers[^2]?


[^1]: https://deno.com/blog/intro-to-wasm
[^2]: https://github.com/cloudflare/workers-sdk/issues/1366#issuecomment-1172255671

kt3k:
This doesn't change the way to import wasm. We still use string and instantiate from it

> Thus could be used in Cloudflare Workers

Oh, I didn't know this doesn't work in cloudflare workers. Thanks for the input

PR #6613, opened 2025-04-22
chore: improve `lint:docs` script

iuioiua:
These refactors partially prepare the `lint:docs` script for use outside of the Standard Library, as discussed offline with Yoshiya a while ago. There are more improvements to come in upcoming PR(s). Other simplifications and cleanups were made, and the script now executes ~40% (1.3s) faster.


kt3k (approved):
LGTM. Thanks!

PR #6614, opened 2025-04-22
chore: improve `_tools/check_browser_compat.ts` performance

iuioiua:
These changes reduces the execution time of this script by ~70% (60s) and removes the `--check` option, which feels redundant given we can just make the appropriate changes.


kt3k (approved):
LGTM. Thanks!

PR #6615, opened 2025-04-22
feat(datetime): add format_distance_to_now function

fbritoferreira:
This PR introduces the`formatDistanceToNow` function to the `@std/datetime package`.
 The function formats a given Date object as a human-readable relative time string, such as "3 minutes ago" or "1 hour ago".

### DIFF
--- a/datetime/deno.json
+++ b/datetime/deno.json
@@ -9,6 +9,7 @@
     "./format": "./format.ts",
     "./is-leap": "./is_leap.ts",
     "./parse": "./parse.ts",
-    "./week-of-year": "./week_of_year.ts"
+    "./week-of-year": "./week_of_year.ts",
+    "./format-distance-to-now": "./format_distance_to_now.ts"
   }
 }
--- a/datetime/format_distance_to_now.ts
+++ b/datetime/format_distance_to_now.ts
@@ -1,0 +1,30 @@
+import { difference } from "@std/datetime/difference";
+
+/**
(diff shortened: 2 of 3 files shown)


timreichen:
While I think this is kinda neat in general I think it opens a big can of worms concerning localization.
Doesn't this kinda overlap with [Intl.DurationFormat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat)?

iuioiua:
> While I think this is kinda neat in general I think it opens a big can of worms concerning localization. Doesn't this kinda overlap with [Intl.DurationFormat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat)?

I agree. I instead suggest adding an example snippet to [`@std/datetime/difference`](https://jsr.io/@std/datetime/doc/~/difference)'s documentation that uses `Intl.DurationFormat`. E.g. https://github.com/denoland/saaskit/pull/706 and https://github.com/jsr-io/jsr/pull/974#discussion_r2055025724.

iuioiua:
I suggest closing this PR in favour of my suggestion above.

kt3k:
I agree with @timreichen and @iuioiua . `Intl.DurationFormat` covers this, and we generally don't try to compete with web platform APIs.

@fbritoferreira Thanks for your suggestion anyways

Closing without merge

PR #6616, opened 2025-04-23
chore: simplify and move `isTestFile()`

iuioiua:
There are very few edge cases where this will return a false positive. So I think this is more than sufficient.


kt3k (approved):
I'm in favor.

The only exception to the check `filePath.endsWith("test.ts")` is the file `front_matter/test.ts`. This check seems enough for detecting that file

PR #6618, opened 2025-04-23
test(jsonc): move `node-jsonc-parser` out of separate `testdata` folder

iuioiua:
This is one step closer toward having all tests in a single place for this package.


kt3k (approved):
LGTM

Issue #6574, opened 2025-04-14
Add check to ensure exports fields are in kebab case


kt3k:
We published the export path `/unstable_dedent` by mistake https://jsr.io/@std/text@1.0.12/doc/unstable_dedent

We like to catch this in the lint CI step

jbronder:
I'm looking into this. Currently, I'm using `toKebabCase` from `@std/text` like an identity function for checking these fields. Something like: `exportName === toKebabCase(exportName)`.

Using the `toKebabCase` function would separate out export field names such as `ascii85` to `ascii-85` and `base64url` to `base-64-url`.  Would it be safe to assume that we're preserving the former names (i.e. `ascii85`, `base64url`) and not also updating these names to adhere to the kebab case format?

kt3k:
> Using the `toKebabCase` function would separate out export field names such as `ascii85` to `ascii-85` and `base64url` to `base-64-url`. Would it be safe to assume that we're preserving the former names (i.e. `ascii85`, `base64url`) and not also updating these names to adhere to the kebab case format?

Yes, we should keep `ascii85` and `base64url`. I think these names are considered a single word as encoding names. I'm not sure if the existing behavior of `toKebabCase` is correct/sensible.

jbronder:
Fair, I think we might able to use `toKebabCase`, but we'll have to add additional checks for these encoding names, etc. Pragmatically, `toKebabCase` as an identity function does seem to capture the majority (96%) of export name cases. Using `toKebabCase` in it's current implementation creates some exceptional cases (4%) for 2 kinds of fields, namely:
- Export fields with numbers (e.g. files in the`@std/uuid` and `@std/encoding` packages). Numbers and letters would separate.
- Export fields with file extensions (only one so far, `@std/html/named-entity-list.json`). File extensions would separate from their names.

To address this, I would propose adding an array of "ignore" phrases and combine that with the `toKebabCase` function, it'll allow:
- Kebab linting to occur
- Permit existing export names that would "violate" kebab rules to pass linting
- Accommodate the addition of future names that may need to be ignored (new encodings and uuid versions, etc.)

With these kinds of linting changes, better error messaging would accompany the changes as well, including:
- Providing suggestions of export field names consistent with `toKebabCase` to encourage developers to provide proper export names for new exports to pass the linting stage.
- Informing developers on how to add more encoding names, uuid versions, etc. in the future as "ignore" phrases during the linting process.

[...]

PR #6619, opened 2025-04-23
chore: add kebab case linter as `_tools/check_export_names.ts`

jbronder:
This PR addresses issue #6574 and is related to PR #6573. This PR adds a kebab case linter to export fields.

Features of the linter include:
- Providing kebab case naming suggestions to developers when proposing new export field names to the Standard Library.
- Permits names that do not satisfy kebab casing to be ignored during linting.
- Allows for the addition of more names (such as new encoding names, uuid versions, etc.) that need to be ignored during the lint phase by informing developers on how to add "ignore" words to the linter.


_tools/check_import_map.ts
```diff
@@ -3,7 +3,86 @@
...
+    .filter((word) => name.includes(word));
+  if (ignoreWords.length !== 0) {
+    // Sort words by the longest string.
+    ignoreWords.sort((aWord, bWord) => bWord.length - aWord.length);
+    let copyName = name;
+    // Replace words with a single "x" to satisfy a well-formed kebab.
+    for (const word of ignoreWords) {
+      const re = new RegExp(word);
+      copyName = copyName.replace(re, "x");
+    }
```
kt3k, line 34: Looks reasonable to me

iuioiua (changes requested):
I think this should be a rule added to the custom lint plugin.

kt3k:
@iuioiua This checks the fields of deno.json. I guess that can't be covered by lint plugin?

iuioiua:
Oh, sorry, you're right.

kt3k (approved):
Confirmed this catches the error like `unstable_dedent` with sensible error message. Thanks!

```
$ deno -A _tools/check_export_names.ts 
./text/deno.json: export name is not in kebab case or contains an invalid "ignore" word:
	Found export name: "unstable_dedent"
	Suggested export name: "unstable-dedent"
NOTE: If part of a name needs to be preserved (not "kebab'd") because it's related to a new encoding (e.g. "base64") or uuid version (e.g. "v1"), that word can be added to the IGNORE_KEBAB_CASE_WORDS array in "_tools/check_export_names.ts".
```

LGTM

PR #6621, opened 2025-04-23
feat(dotenv): add URL as envPath type

eikooc:
This PR fixes https://github.com/denoland/std/issues/6620, by adding `URL` as a valid type to the `envPath` parameter in the `LoadOptions` interface.

### DIFF
--- a/dotenv/mod_test.ts
+++ b/dotenv/mod_test.ts
@@ -29,6 +29,24 @@ Deno.test("load() handles non-existent .env files", async () => {
   assertEquals({}, loadSync(loadOptions));
 });
 
+Deno.test("load() handles URL as path for .env files", async () => {
+  const conf = loadSync({
+    envPath: new URL(
+      path.toFileUrl(path.join(testdataDir, ".env")),
+      import.meta.url,
+    ),
+  });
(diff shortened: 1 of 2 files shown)


dotenv/mod_test.ts
```diff
@@ -29,6 +29,18 @@ Deno.test("load() handles non-existent .env files", async () => {
   assertEquals({}, loadSync(loadOptions));
 });
 
+Deno.test("load() handles URL as path for .env files", async () => {
+  const conf = loadSync({
+    envPath: new URL(path.join(testdataDir, ".env"), import.meta.url),
```
kt3k, line 34: `new URL(path.join(testdataDir, ".env"), import.meta.url)`

This doesn't work on windows as the paths are `\` separated on windows, but that doesn't work as URL string

You need to call `toFileUrl` from `@std/path`

eikooc: Do you mean like this:

```ts
new URL(
  toFileUrl(path.join(testdataDir, ".env")),
  import.meta.url,
)
```
?

eikooc:
Anything missing @kt3k for this to be merged?

kt3k (approved):
Sorry for the delay in reviewing. This now looks good to me. Thank you for your contribution!

PR #6624, opened 2025-04-25
BREAKING(fs,dotenv): drop Deno v1.x support

iuioiua:
This PR drops Deno v1.x from being used in CI, or rather it just removes the guarantee that the Standard Library will work with Deno v1.x (even though it does 99% of the time). I think this is reasonable to do now as Deno 2 is LTS.

The title is accurate but may seem more significant than these changes are. This just changes the expected error when insufficient permissions are granted for some operations in `@std/fs` and `@std/dotenv`. See [`Deno.errors.NotCapable`](https://docs.deno.com/api/deno/~/Deno.errors.NotCapable).

Please feel free to close if strongly against.


kt3k:
v1 is still significantly used according to our internal check. (Probably many users are still stuck with v1 because of CLI's breaking changes)

Screenshot 2025-04-26 at 22 06 17

I don't think we are ready to do this change.

cc @bartlomieju @ry

.github/workflows/ci.yml
```diff
@@ -19,7 +19,6 @@ jobs:
       fail-fast: false
       matrix:
         deno:
-          - v1.x
```
kt3k, line 22: If we drop v1 CI, we don't what packages work in v1 anymore (any package could randomly break in v1 in the future development)

bartlomieju:
Given @kt3k's usage table I think we should wait more, there's very little code here so I don't think it's a huge maintenance burden.

bartlomieju:
Btw @iuioiua we can keep this PR open for now

kt3k:
BTW I think version upgrade of this PR needs further discussion. Upgrading only `fs`, `dotenv` feels strange to me.

The change to `fs` isn't breaking change to 2.x users. Do we need to do major upgrade for it?

iuioiua:
Closing in favour of another change.

PR #6625, opened 2025-04-26
chore: improve `lint:docs` script

iuioiua:
Now, assertions use `assert()` instead of manually pushing to `diagnostics`.

Step towards having this be a tool that other codebases can use.

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -409,8 +409,6 @@ async function assertDocs(specifiers: string[]) {
       case "interface":
         assertInterfaceDocs(document);
         break;
-      case "variable":
-        break;
     }
   }
 }
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -52,7 +52,7 @@ function assert(
   condition: boolean,
   message: string,
   document: { location: Location },
-) {
+): asserts condition {
   if (!condition) {
     diagnostics.push(new DocumentError(message, document));
   }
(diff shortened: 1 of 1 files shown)


_tools/check_docs.ts
```diff
@@ -77,18 +77,17 @@ function isVoid(returnType: TsTypeDef) {
...
-    );
-  }
+  assert(
+    tag !== undefined,
+    "Symbol must have a @return or @returns tag",
+    document,
+  );
+  assert(
+    // @ts-ignore doc is defined
+    tag.doc !== undefined,
```
kt3k, line 87: This seems throwing uncaught error when tag is undefined:

```
error: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'doc')
    tag.doc !== undefined,
        ^
    at assertHasReturnTag (file:///Users/kt3k/denoland/std/_tools/check_docs.ts:87:9)
    at assertFunctionDocs (file:///Users/kt3k/denoland/std/_tools/check_docs.ts:252:5)
    at assertDocs (file:///Users/kt3k/denoland/std/_tools/check_docs.ts:402:9)
    at eventLoopTick (ext:core/01_core.js:178:7)
    at async checkDocs (file:///Users/kt3k/denoland/std/_tools/check_docs.ts:427:3)
    at async file:///Users/kt3k/denoland/std/_tools/check_docs.ts:452:3
```

_tools/check_docs.ts
```diff
@@ -128,37 +124,29 @@ function assertHasParamTag(
...
-    );
-  }
+  assert(
+    tag !== undefined,
+    `Symbol must have a @param tag for ${param}`,
+    document,
+  );
+  assert(
+    // @ts-ignore doc is defined
+    tag.doc !== undefined,
```
kt3k, line 134: Ditto. This could throw uncaught error

_tools/check_docs.ts
```diff
@@ -128,37 +124,29 @@ function assertHasParamTag(
...
-        ),
-      );
-    }
-    return;
-  }
+  assert(
+    snippets !== null,
+    "@example tag must have a TypeScript code snippet",
+    document,
+  );
```
kt3k, line 149: Similarly to the above, We need to return when snippets === null

kt3k:
Other part looks good to me

kt3k (approved):
LGTM

PR #6629, opened 2025-04-28
feat(assert/unstable): allow asserting predicates against thrown/rejected errors

lionel-rowe:
Closes https://github.com/denoland/std/issues/6592.

This also allows regexes to be used with `assertThrows` and `assertThrows`, for consistency with `assertIsError`.

Clean diffs against cloned versions of the stable files: https://github.com/denoland/std/compare/d470af4397d79b9a25c920b3313fc9a8622d5125...lionel-rowe:std:throws-rejects-predicate

### DIFF
--- a/assert/deno.json
+++ b/assert/deno.json
@@ -30,6 +30,12 @@
     "./equal": "./equal.ts",
     "./fail": "./fail.ts",
     "./unimplemented": "./unimplemented.ts",
+    "./unstable-is-error": "./unstable_is_error.ts",
+    "./unstable-is-error-test": "./unstable_is_error_test.ts",
+    "./unstable-rejects": "./unstable_rejects.ts",
+    "./unstable-rejects-test": "./unstable_rejects_test.ts",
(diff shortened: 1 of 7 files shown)


kt3k:
I'm not in favor of this suggestion. `assertThrows` and `assertRejects` used to accept `errorCallback` (not predicate through) and you were able to do additional assertion about thrown error there. However we deprecated that in https://github.com/denoland/std/pull/2226. (See also the discussion in https://github.com/denoland/std/issues/2220 )

It is now recommended to do additional assertions about thrown/rejected errors using the return value of `assertThrows`/`assertRejects`:

```js
const e0 = assertThrows(...);
// do assertions about e0
const e1 = await assertRejects(...);
// do assertions about e1
```

This structure should be more flexible than the suggested `predicate` parameter.

lionel-rowe:
> using the return value of `assertThrows`/`assertRejects`

D'oh, somehow I missed that the error is now returned. Yeah in that case this seems pretty unnecessary, closing.

PR #6632, opened 2025-04-30
chore: export doc checker

iuioiua:
Towards https://github.com/denoland/fresh/issues/2876 and towards making this tool available for other codebases to use.

### DIFF
--- a/_tools/check_docs.ts
+++ b/_tools/check_docs.ts
@@ -415,7 +415,7 @@ async function assertDocs(specifiers: string[]) {
   }
 }
 
-async function checkDocs(specifiers: string[]) {
+export async function checkDocs(specifiers: string[]) {
   const lintStatus = await new Deno.Command(Deno.execPath(), {
     args: ["doc", "--lint", ...specifiers],
     stdin: "inherit",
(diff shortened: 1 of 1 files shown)


iuioiua:
I think CI just needs to be re-run.

kt3k:
Do you plan to share `_tools/check_docs.ts` using git submodule?

iuioiua:
Yeah, for now. Though, I think we should consider having these modules available under a `@std/tools` at some point. Perhaps when they're a little more mature. They'd be super useful for maintainers of other codebases.

kt3k:
> Though, I think we should consider having these modules available under a @std/tools at some point. Perhaps when they're a little more mature.

I doubt these tool can reach such point anytime soon.

If we publish them as part of `std`. They need to do proper parsing of cli options, validations of options, add test cases of them, etc. If they get 'bug report' for use cases that currently we are not aware, then we need to 'fix' them. These works include far more than what we need for only checking std libs.

If you consider publishing them as standalone CLI tool in JSR, I recommend you build it as a 3rd party library.

kt3k:
sharing as git submodule sounds fine though

kt3k (approved):
LGTM

PR #6634, opened 2025-04-30
fix(html/unstable): add missing range (`unstable_is_valid_custom_element_name`)

### LINKED ISSUE #6633: html: `unstable_is_valid_custom_element_name` missing valid custom element name char range
the `CUSTOM_ELEMENT_NAME_CHARS` regex should also allow the range [#x10000-#xEFFFF] to be compliant with the spec 

[html.spec.whatwg.org/multipage/custom-elements.html#prod-pcenchar](https://html.spec.whatwg.org/multipage/custom-elements.html#prod-pcenchar)

fcrozatier:
closes #6633


kt3k (approved):
Good catch! Thanks

PR #6635, opened 2025-05-01
chore: run some test cases in bun in CI

kt3k:
This PR starts running node compatible test cases in Bun in CI. This makes it easier to accept PRs like https://github.com/denoland/std/pull/6630 which says the change improves the bun compatibility.

In JSR pages, we already mark some of std packages compatible with Bun. So I believe this addition is reasonable

### DIFF
--- a/_tools/node_test_runner/tsconfig_for_bun.json
+++ b/_tools/node_test_runner/tsconfig_for_bun.json
@@ -1,0 +1,15 @@
+{
+  "compilerOptions": {
+    "paths": {
+      "@std/assert": ["./assert/mod.ts"],
+      "@std/assert/greater-or-equal": ["./assert/greater_or_equal.ts"],
+      "@std/path": ["./path/mod.ts"],
+      "@std/internal/format": ["./internal/format.ts"],
+      "@std/internal/styles": ["./internal/styles.ts"],
+      "@std/internal/build-message": ["./internal/build_message.ts"],
+      "@std/internal/diff": ["./internal/diff.ts"],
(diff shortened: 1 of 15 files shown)


_tools/node_test_runner/tsconfig_for_bun.json
```diff
@@ -0,0 +1,15 @@
...
+      "@std/assert": ["./assert/mod.ts"],
+      "@std/assert/greater-or-equal": ["./assert/greater_or_equal.ts"],
+      "@std/path": ["./path/mod.ts"],
+      "@std/internal/format": ["./internal/format.ts"],
+      "@std/internal/styles": ["./internal/styles.ts"],
+      "@std/internal/build-message": ["./internal/build_message.ts"],
+      "@std/internal/diff": ["./internal/diff.ts"],
+      "@std/internal/diff-str": ["./internal/diff_str.ts"],
+      "@std/semver": ["./semver/mod.ts"],
+    }
```
kt3k, line 13: `register` function of `node:module` didn't seem working in Bun (which is used for resolving specifiers like `@std/path` in Node test runner), but instead Bun allows users to remap specifiers via `compilerOptions.paths` option. ref https://bun.sh/docs/runtime/modules#path-re-mapping

> In the spirit of treating TypeScript as a first-class citizen, the Bun runtime will re-map import paths according to the [compilerOptions.paths](https://www.typescriptlang.org/tsconfig#paths) field in tsconfig.json. This is a major divergence from Node.js, which doesn't support any form of import path re-mapping.

bartlomieju (approved):
LGTM, pretty cool!

PR #6638, opened 2025-05-04
fix(toml): should use lowercase value for nan

mateusvmv:
The toml parser is stringifying nan as NaN, and the parser doesn't recognize it a a nan value, since it expects it to be lowercase. This PR fixes that.


kt3k (approved):
LGTM. Thanks for catching this!

kt3k:
Note: The spec says in the [Float section](
https://toml.io/en/v1.0.0#float):

> Special float values can also be expressed. They are always lowercase.

PR #6639, opened 2025-05-05
fix(bytes): allow to concat() readonly arrays of bytes arrays

ngasull:
`concat` currently only allows mutable arrays whereas it doesn't mutate its input.

This PR expands the signature to read-only arrays.

Doesn't refer to any open issue.


kt3k (approved):
LGTM. Thanks!

PR #6640, opened 2025-05-05
ci: use correct Bun version

uki00a:
This is a minor change. This PR changes `matrix.node` to `matrix.bun` to prevent the use of unintended versions of Bun in the future.

https://github.com/denoland/std/blob/6a00d6cd34ca70b3a38932e313151a80531c5d2b/.github/workflows/ci.yml#L105-L126


kt3k (approved):
oh, good catch! Thanks

PR #6643, opened 2025-05-07
chore: remove `no-top-level-arrow-syntax` lint rule

iuioiua:
This rule no longer applies. See https://github.com/denoland/docs/pull/1685


kt3k:
I added back this rule to the style guide https://github.com/denoland/docs/pull/1690

PR #6646, opened 2025-05-09
fix(expect,testing,internal): throw if `expect.hasAssertion` and `expect.assertions` are not checked

### LINKED ISSUE #6518: `expect.hasAssertion` and `expect.assertions` always pass
**Describe the bug**

`expect.hasAssertion` and `expect.assertions` always pass even when there are not the right number of assertions

**Steps to Reproduce**

```ts
import { expect } from "jsr:@std/expect";

Deno.test("should fail", () => {
    expect.hasAssertions();
});

Deno.test("should also fail", () => {
    expect.assertions(2);
    expect(1).toBe(1);
});
```

**Expected behavior**

Both tests fail

**Environment**

- OS: Win11
- deno version: 2.2.5
- std version: 1.0.14

### LINKED ISSUE #6540: `expect.hasAssertion` and `expect.assertions` pollute other assertions tests
**Describe the bug**

If one `expect.hasAssertion` or `expect.assertions` test fails, others may be affected.

**Steps to Reproduce**

Only the first test should fail, but the second also fails: the assertion from the first test leaked into the second.

```ts
test("failing test", () => {
  expect.assertions(0);
  expect(1).toEqual(1);
});

test("expect no assertions", () => {
  expect.assertions(0);
});
```

**Expected behavior**

Assertions tests should not interact across testing contexts

**Environment**

- OS: Win11
- deno version: 2.2.5
- std version: latest

WWRS:
fixes #6518
fixes #6540

- ~~Renamed AssertionState to AssertionCounter~~
- ~~Moved AssertionCounter from `std/internal` to `std/expect`: It's really more of an `expect` feature than an `internal` feature~~
  - ~~This means that `std/testing` now depends on `std/expect`~~
- ~~Removed `AssertionCounter.#state.assertionTriggered` to avoid desynchronizing from `assertionTriggeredCount`~~
- ~~Collected the validation error behavior in `AssertionCounter` rather than duplicating it in `testing`~~
- Ensured `AssertionState` is reset after validation: fixes #6540
- Added a cleanup step when `globalThis` is `unload`ed to ensure all `AssertionState`s are properly validated and reset: fixes #6518

### DIFF
--- a/internal/assertion_state_test.ts
+++ b/internal/assertion_state_test.ts
@@ -1,28 +1,55 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
-import { assertEquals } from "@std/assert";
+import { assertEquals, assertStringIncludes } from "@std/assert";
 import { AssertionState } from "./assertion_state.ts";
 
 Deno.test("AssertionState checkAssertionErrorState pass", () => {
   const assertionState = new AssertionState();
-  assertionState.setAssertionTriggered(true);
-
-  assertEquals(assertionState.checkAssertionErrorState(), false);
+  try {
+    assertionState.setAssertionTriggered(true);
+    assertEquals(assertionState.checkAssertionErrorState(), false);
+  } finally {
+    assertionState.resetAssertionState();
+  }
 });
 
 Deno.test("AssertionState checkAssertionErrorState pass", () => {
   const assertionState = new AssertionState();
-  assertionState.setAssertionTriggered(true);
+  try {
+    assertionState.setAssertionTriggered(true);
 
-  assertEquals(assertionState.checkAssertionErrorState(), false);
+    assertEquals(assertionState.checkAssertionErrorState(), false);
 
-  assertionState.setAssertionCheck(true);
-  assertEquals(assertionState.checkAssertionErrorState(), false);
+    assertionState.setAssertionCheck(true);
+    assertEquals(assertionState.checkAssertionErrorState(), false);
+  } finally {
+    assertionState.resetAssertionState();
+  }
 });
 
 Deno.test("AssertionState checkAssertionErrorState fail", () => {
   const assertionState = new AssertionState();
-  assertionState.setAssertionCheck(true);
+  try {
+    assertionState.setAssertionCheck(true);
+    assertEquals(assertionState.checkAssertionErrorState(), true);
+  } finally {
+    assertionState.resetAssertionState();
+  }
+});
 
-  assertEquals(assertionState.checkAssertionErrorState(), true);
+Deno.test("AssertionState throws if not cleaned up", async () => {
+  const command = new Deno.Command(Deno.execPath(), {
+    args: [
+      "eval",
+      `
+        import { AssertionState } from "@std/internal/assertion-state";
+        const assertionState = new AssertionState();
+        assertionState.setAssertionCount(0);
+      `,
+    ],
+  });
+  const { stderr } = await command.output();
+  const errorMessage = new TextDecoder().decode(stderr);
+  // TODO(WWRS): Test for the expected message when Deno displays it instead of "Uncaught null"
+  assertStringIncludes(errorMessage, "error: Uncaught");
 });
--- a/internal/assertion_state_test.ts
+++ b/internal/assertion_state_test.ts
@@ -2,6 +2,7 @@
 
 import { assertEquals, assertStringIncludes } from "@std/assert";
 import { AssertionState } from "./assertion_state.ts";
+import { stripAnsiCode } from "./styles.ts";
 
 Deno.test("AssertionState checkAssertionErrorState pass", () => {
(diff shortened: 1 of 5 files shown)


internal/deno.json
```diff
@@ -3,7 +3,6 @@
   "version": "1.0.6",
   "exports": {
     ".": "./mod.ts",
-    "./assertion-state": "./assertion_state.ts",
```
kt3k, line 6: It's not possible to remove this path. For example, `@std/testing@1.0.11` depends on `@std/internal@^1.0.6/assertion-state`. If we publish the next version of `@std/internal` with this export removed, that will break `@std/testing@1.0.11` (because the semver range `^1.0.6` will resolve to `1.0.7` at the next `@std/internal` release)

So we generally can only do the additive change to `@std/interanl`

expect/deno.json
```diff
@@ -3,6 +3,7 @@
   "version": "1.0.15",
   "exports": {
     ".": "./mod.ts",
+    "./assertion-counter": "./assertion_counter.ts",
```
kt3k, line 6: This makes `AssertionCounter` a public API of `@std/expect`. I don't think that is a desirable change. Can we keep this in `@std/internal`? (Any API in `@std/internal` is considered private)

kt3k:
Thank you for the PR! This looks like a nice fix, but I left some comments about code organization.

expect/deno.json
kt3k: > Moved AssertionCounter from std/internal to std/expect: It's really more of an expect feature than an internal feature

The reason of putting this in `@std/interanl` is that we want to share this between `testing/bdd` and `expect`, but also we don't want to make this public. In that case, such functionality needs to go to `@std/internal`

WWRS: `AssertionCounter` throws an `AssertionError`, which is only available in `std/assert`. Some options after moving `AssertionCounter` (back) to `std/internal`:

- Leave `AssertionCounter` as it is and have `std/internal` depend on `std/assert`
- Make `AssertionCounter` throw an `Error`
- Also move `AssertionError` to `std/internal` and make `std/assert` re-export it

Thoughts on which is best?

internal/deno.json
WWRS: So in order to support previous versions, we would need to keep `assertion_state.ts` and have it provide the same functions as previously? That would include `checkAssertionErrorState`, `checkAssertionCountSatisfied`, and `resetAssertionState` which have been removed from `AssertionCounter`. Should those get added back in?

expect/deno.json
kt3k: The dependency from `@std/internal` to `@std/assert` doesn't seem great (also probably that doesn't work as it introduces a circular dependency chain of `assert` -> `internal` -> `assert`)

I feel the current implementation does better job than the options above. It only exposes necessary conditions (`.checkAssertionErrorState()` or `.checkAssertionCountSatisfied()`) and `testing/bdd` and `expect` throws `AssertionError` using those conditions. That keeps dependency tree simpler. Can we do the similar with AssertionCounter?

internal/deno.json
kt3k: > So in order to support previous versions, we would need to keep `assertion_state.ts` and have it provide the same functions as previously? That would include `checkAssertionErrorState`, `checkAssertionCountSatisfied`, and `resetAssertionState` which have been removed from `AssertionCounter`. Should those get added back in?

Yes, they need to be kept as is to keep the previous versions of `testing` and `expect` working

WWRS:
@kt3k I've rewritten this PR to modify as little of the behavior as possible.

Note that `_assertions_test.ts` was previously not actually running any tests, so I did rewrite that.

kt3k:
> Note that _assertions_test.ts was previously not actually running any tests, so I did rewrite that.

Thanks for working on this!

internal/assertion_state_test.ts
```diff
@@ -1,28 +1,56 @@
...
+        import { AssertionState } from "@std/internal/assertion-state";
+        const assertionState = new AssertionState();
+        assertionState.setAssertionCount(0);
+      `,
+    ],
+  });
+  const { stderr } = await command.output();
+  const errorMessage = stripAnsiCode(new TextDecoder().decode(stderr));
+  // TODO(WWRS): Test for the expected message when Deno displays it instead of "Uncaught null"
+  assertStringIncludes(errorMessage, "error: Uncaught");
```
kt3k, line 55: Is this supposed to be the message generated in `unload` handler?

kt3k (approved):
Thanks for updating! This now looks good to me 👍 

Nice fixes!

internal/assertion_state_test.ts
WWRS: Yes, I think this is an upstream issue, which I've opened: https://github.com/denoland/deno/issues/29087

PR #6655, opened 2025-05-18
fix(internal): support AssertionState cleanup in node and bun

### LINKED ISSUE #6652: Bug: @std/internal unexpected globalThis.* usage for node.js
**Describe the bug**

https://github.com/poolifier/poolifier/actions/runs/15019258415/job/42204316093:
```
 1) Uncaught error outside test suite:
     Uncaught TypeError: globalThis.addEventListener is not a function
      at new AssertionState (/home/runner/work/poolifier/poolifier/node_modules/.pnpm/@jsr+std__internal@1.0.7/node_modules/@jsr/std__internal/assertion_state.ts:33:16)
      at <anonymous> (/home/runner/work/poolifier/poolifier/node_modules/.pnpm/@jsr+std__internal@1.0.7/node_modules/@jsr/std__internal/assertion_state.ts:212:24)
      at ModuleJob.run (node:internal/modules/esm/module_job:263:25)
      at async ModuleLoader.import (node:internal/modules/esm/loader:540:24)
      at async formattedImport (/home/runner/work/poolifier/poolifier/node_modules/.pnpm/mocha@11.2.2/node_modules/mocha/lib/nodejs/esm-utils.js:9:14)
      at async exports.loadFilesAsync (/home/runner/work/poolifier/poolifier/node_modules/.pnpm/mocha@11.2.2/node_modules/mocha/lib/nodejs/esm-utils.js:100:20)
      at async AsyncFunction.run (/home/runner/work/poolifier/poolifier/node_modules/.pnpm/mocha@11.2.2/node_modules/mocha/lib/nodejs/worker.js:107:5)
```

Or https://github.com/SAP/e-mobility-charging-stations-simulator/actions/runs/15022866262/job/42216243165

**Steps to Reproduce**

Use latest @std/expect with node.js, any supported versions

**Expected behavior**

 Should work with node.js as the previous version

**Environment**

- node version: all LTS
- @std/internal version: 1.0.7

WWRS:
fixes #6652

### DIFF
--- a/internal/assertion_state.ts
+++ b/internal/assertion_state.ts
@@ -27,19 +27,36 @@ export class AssertionState {
       assertionTriggeredCount: 0,
     };
 
+    if (typeof globalThis?.addEventListener === "function") {
+      globalThis.addEventListener("unload", () => {
+        this.#ensureCleanedUp();
+      });
+    } else if (
+      // deno-lint-ignore no-explicit-any
+      typeof (globalThis as any)?.process?.on === "function"
+    ) {
+      // deno-lint-ignore no-explicit-any
+      (globalThis as any).process.on("exit", () => {
+        this.#ensureCleanedUp();
+      });
+    } else {
+      // deno-lint-ignore no-console
+      console.warn("AssertionCounter cleanup step was not registered");
+    }
+  }
+
+  #ensureCleanedUp() {
     // If any checks were registered, after the test suite runs the checks,
     // `resetAssertionState` should also have been called. If it was not,
     // then the test suite did not run the checks.
-    globalThis.addEventListener("unload", () => {
-      if (
-        this.#state.assertionCheck ||
-        this.#state.assertionCount !== undefined
-      ) {
-        throw new Error(
-          "AssertionCounter was not cleaned up: If tests are not otherwise failing, ensure `expect.hasAssertion` and `expect.assertions` are only run in bdd tests",
-        );
-      }
-    });
+    if (
+      this.#state.assertionCheck ||
+      this.#state.assertionCount !== undefined
+    ) {
+      throw new Error(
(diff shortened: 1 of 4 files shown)


internal/assertion_state.ts
```diff
@@ -27,19 +27,36 @@ export class AssertionState {
...
+      // deno-lint-ignore no-explicit-any
+      typeof (globalThis as any)?.process?.on === "function"
+    ) {
+      // deno-lint-ignore no-explicit-any
+      (globalThis as any).process.on("exit", () => {
+        this.#ensureCleanedUp();
+      });
+    } else {
+      // deno-lint-ignore no-console
+      console.warn("AssertionCounter cleanup step was not registered");
```
WWRS, line 44: Is this the right thing to do if we can't find `globalThis.addEventListener` (Deno, browsers) or `process.on` (Node, Bun)? It does seem extremely unlikely this would ever happen.

jerome-benoit: If none are available based on chosen the priority, ressource(s) will be leaked. So you have to throw an explicit error in that case.

jerome-benoit: How can you guarantee that `this.#ensureCleanedUp()` is not unrefing, and the like, in the future? 
Silently (almost) skipping a cleanup is not a good practice. Throwing an error would avoid it, while also ensuring that JS runtime not supporting the desired method are not compatible with @std explicitly.

WWRS: It might be helpful to have the context for this step: #6518

This check only actually matters when `expect.assertions` or `expect.hasAssertion` is called. Previously, we didn't do any such check, which allowed those functions to be used in ways that would register those assertions but not actually do the validation.

kt3k (approved):
LGTM. Nice! Thanks

PR #6656, opened 2025-05-18
Better TOML types

Gouvernathor:


toml/types.ts
```diff
@@ -0,0 +1,4 @@
+export type TOMLTable = {[Key in string]: TOMLValue} & {[Key in string]: TOMLValue|undefined};
```
kt3k, line 1: We'd prefer the capitalization of `TomlTable`. See https://docs.deno.com/runtime/contributing/style_guide/#naming-convention

kt3k: The second part `{[Key in string]: TOMLValue|undefined}` looks strange to me. Is it possible for key to be undefined? Is that spec compliant?

toml/_parser.ts
```diff
@@ -131,18 +132,21 @@ export class Scanner {
 // Utilities
 // -----------------------
 
-function success<T>(body: T): Success<T> {
+function success<T extends TOMLValue|void>(body: T): Success<T> {
```
kt3k, line 135: Why `|void` part necessary?

kt3k:
Run `deno fmt` to format the source code

kt3k:
We currently don't accept a new feature directly, but we'd like to accept it as unstable feature first (See https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#suggesting-a-new-feature )

Let's keep the return type of `parse()` to `Record<string, unknown>`, and create another entrypoint `unstable_parse.ts` which exports unstable `parse` function that returns `TomlValue` type.

toml/_parser.ts
```diff
@@ -229,25 +233,25 @@ export function deepAssign(
...
 > {
-  return (scanner: Scanner) => {
+  return ((scanner: Scanner) => {
     for (const parse of parsers) {
       const result = parse(scanner);
       if (result.ok) return result;
     }
     return failure();
-  };
+  }) as unknown as any;
```
kt3k, line 247: This needs `// deno-lint-ignore no-explicit-any` directive before this line to pass lint check. See the `deno lint` output

bartlomieju:
Closing due to inactivity — this PR has been open for over 10 months with review feedback from @kt3k that was never addressed (naming conventions, lint issues, the requirement to introduce this as an unstable feature first via `unstable_parse.ts`).

If you'd like to revisit this, feel free to open a fresh PR that addresses the review comments. Thanks for the contribution!

PR #6658, opened 2025-05-18
refactor(yaml): move `composeNode()` params in an object

timreichen:
This improves readability on which data is passed to the function.


kt3k (approved):
LGTM

PR #6659, opened 2025-05-19
refactor(http): format file server HTML

iuioiua:
Thanks to https://deno.com/blog/v2.3#improvements-to-deno-fmt


http/file_server_test.ts
```diff
@@ -198,7 +198,12 @@ Deno.test("serveDir() serves directory index with entry to the parent directory"
   assertEquals(res.status, 200);
   assertStringIncludes(
     page,
-    '<a href="../../">home</a>/<a href="../">nested-subdirs</a>/<a href=".">subdir</a>/',
+    `<a href="../../">home</a>
+        /
+          <a href="../">nested-subdirs</a>
+        /
+          <a href=".">subdir</a>
+        /`,
```
lionel-rowe, line 206: This will render as `home / nested-subdirs / subdir /`, not `home/nested-subdirs/subdir/` as before

http/file_server.ts
```diff
@@ -427,11 +427,21 @@ function createBaseHeaders(): Headers {
...
 }
 
+function html(
+  strings: TemplateStringsArray,
+  ...values: unknown[]
+): string {
+  return strings.reduce(
+    (result, str, i) => result + str + (values[i] ?? ""),
+    "",
+  );
```
kt3k, line 437: It looks like MDN recommends implementing this as:
```suggestion
  return String.raw({ raw: strings }, ...values);
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/raw#building_an_identity_tag

http/file_server.ts
```diff
@@ -515,10 +524,11 @@ function dirViewerTemplate(dirname: string, entries: EntryInfo[]): string {
         } else {
           link = "../".repeat(depth);
         }
-        return `<a href="${link}">${escape(path)}</a>`;
+        return html`
+          <a href="${link}">${escape(path)}</a>
+        `;
```
kt3k, line 529: Maybe better to not use `html` here to avoid some rendering changes pointed by @lionel-rowe ?

kt3k (approved):
LGTM. Nice!

PR #6660, opened 2025-05-20
refactor(cli/unstable): remove `previousTime` and `previousValue`

timreichen:
I don't see why they should be provided by the formatter object.
These values can easily be stored and accessed in variables as is demonstrated in the test that is removed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,7 @@ jobs:
         uses: denoland/setup-deno@v2
         with:
           deno-version: ${{ matrix.deno }}
+          cache: true
 
       - name: Run tests
         run: |
@@ -93,6 +94,8 @@ jobs:
 
       - name: Set up Deno
         uses: denoland/setup-deno@v2
+        with:
+          cache: true
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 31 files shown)


BlackAsLight:
I don't agree with this change as these values are needed if one wants the progress bar to display the rate of progress being made. Forcing users to create additional variables to achieve this seems like a bad design choice.

timreichen:
> I don't agree with this change as these values are needed if one wants the progress bar to display the rate of progress being made. Forcing users to create additional variables to achieve this seems like a bad design choice.

Could you give an example how that would look like? I compared it to [cli-progress](https://www.npmjs.com/package/cli-progress#bar-formatting) which doesn't provide previous values in formatter, so I thought this is probably very niche.

BlackAsLight:
> Could you give an example how that would look like? I compared it to cli-progress which doesn't provide previous values in formatter, so I thought this is probably very niche.

A rate like x mb/s. Essentially these values makes it easy for you to communicate the speed at which progress is being made.

```
[===-------] [2 MiB/s]
```

timreichen:
> > Could you give an example how that would look like? I compared it to cli-progress which doesn't provide previous values in formatter, so I thought this is probably very niche.
> 
> A rate like x mb/s. Essentially these values makes it easy for you to communicate the speed at which progress is being made.
> 
> ```
> [===-------] [2 MiB/s]
> ```

Ah, I see. Would it make sense to have the speed as a separate property instead the need to calculate it by hand?
Maybe a delta that can be formatted into a unit?
`[${formatUnit(delta)}/s]`?
I think std would profit by a public `UnitFormatter` anyway.

timreichen:
stale

PR #6661, opened 2025-05-20
fix(dotenv): prevent prototype pollution during parsing

lucacasonato:
If you did 

```
toString=FOO
```

that would overwrite `Object.prototype.toString`.


marvinhagemeister (approved):
LGTM

PR #6662, opened 2025-05-20
refactor(datetime): simplify month difference

timreichen:
This makes the code easier to read and also caches the result of `calculateMonthsDifference()` when either `quarters` or `years` is calculated first.


kt3k (approved):
LGTM

PR #6664, opened 2025-05-21
refactor(unstable/cli): print progress bar when `value` and `max` are changed

timreichen:
This PR decodes buffers in tests and compares them to strings for easier comprehension.
It adds getters and setters for `value` and `max` which call `#print()` when the value changes and also recalculates `unit` and `rate` when `max` changes.


BlackAsLight:
I don't see the point of calling the #print function every time the values are updated. If value is updated a lot then that can lead to a lot of unnecessary prints.

timreichen:
> I don't see the point of calling the #print function every time the values are updated. If value is updated a lot then that can lead to a lot of unnecessary prints. 

I wouldn't say these prints are unnecessary. If there is a lot of progress, I think it is beneficial to see that reflected on the progress bar more than every second, no?

BlackAsLight:
> > I don't see the point of calling the #print function every time the values are updated. If value is updated a lot then that can lead to a lot of unnecessary prints. 
> 
> I wouldn't say these prints are unnecessary. If there is a lot of progress, I think it is beneficial to see that reflected on the progress bar more than every second, no?

No. I very likely could be updating `value` 100x a second. That is a lot of wasted prints.

kt3k:
> I think an inconsistent update interval though could give the effect that the program is having lag spikes and any "delta" values that are introduced or manually calculated via the formatter option can have big jumps up and down.

I see the point, but it feels weird that we can't update frequently to keep the delta calculation correct..

How about splitting the print of progress bar and the calculation of delta of the value?

timreichen:
stale

PR #6667, opened 2025-05-21
fix(text/unstable): make dedent match reference for multiline substitutions

### LINKED ISSUE #6665: `text/unstable-dedent` gives poor results when interpolating multi-line text into tagged template
**Describe the bug**

**Steps to Reproduce**

```ts
import { dedent } from 'jsr:@std/text@1.0.13/unstable-dedent'
import { assertEquals } from 'jsr:@std/assert'
const inner = dedent`
    2
    3
`
assertEquals(inner, '2\n3') // ok
const outer = dedent`
    1
    ${inner}
    4
`
assertEquals(outer, '1\n2\n3\n4') // throws
```

**Diff**

```diff
- Actual
+ Expected

-     1
-     2
+ 1
+ 2
  3
-     4
+ 4
```

This looks to be because the interpolated values aren't differentiated from the template parts when calculating minimum common indentation to strip.

For comparison, the `npm:string-dedent` template function (unofficial [`String.dedent`](https://github.com/tc39/proposal-string-dedent/) polyfill) gives the expected result.

**Environment**

text@1.0.13

WWRS:
fixes #6665

### DIFF
--- a/text/unstable_dedent.ts
+++ b/text/unstable_dedent.ts
@@ -62,12 +62,11 @@ export function dedent(
   input: TemplateStringsArray | string,
   ...values: unknown[]
 ): string {
-  const inputString = typeof input === "string"
-    ? input
-    : String.raw({ raw: input }, ...values);
-  const ignoreFirstUnindented = !inputString.startsWith("\n");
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Nice!

PR #6668, opened 2025-05-21
feat(http/unstable): `unstable-file-server`

### LINKED ISSUE #5681: to-do: make `@std/http/file-server` compatible with `deno serve --parallel`
`deno serve --parallel` was implemented in https://github.com/denoland/deno/pull/24920. `@std/http/file-server` should be made to use this sub-command and take advantage of `--parallel`. Benchmarks would also be great.

iuioiua:
This new module is an updated version of `@std/http/file-server` that uses [`deno serve`](https://docs.deno.com/runtime/reference/cli/serve/), which takes care of the heavy lifting of some argument parsing and can be run in parallel service workers thanks to [`--parallel`](https://docs.deno.com/runtime/reference/cli/serve/#options-parallel). It directly relies on the exported symbols of `@std/http/file-server` and a few kinks need to be figured out.

Closes #5681

### DIFF
--- a/http/unstable_file_server_test.ts
+++ b/http/unstable_file_server_test.ts
@@ -1,0 +1,1069 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+import {
+  assert,
+  assertAlmostEquals,
+  assertEquals,
+  assertMatch,
+  assertStringIncludes,
+} from "@std/assert";
+import { stub } from "@std/testing/mock";
+import {
+  serveDir,
+  type ServeDirOptions,
+  serveFile,
+} from "./unstable_file_server.ts";
+import { eTag } from "./etag.ts";
+import {
+  basename,
+  dirname,
+  fromFileUrl,
+  join,
+  resolve,
+  toFileUrl,
+} from "@std/path";
+import denoConfig from "./deno.json" with { type: "json" };
+import { MINUTE } from "@std/datetime/constants";
+import { lessThan, parse as parseSemver } from "@std/semver";
+import { getAvailablePort } from "@std/net/get-available-port";
+import { concat } from "@std/bytes/concat";
+
+const moduleDir = dirname(fromFileUrl(import.meta.url));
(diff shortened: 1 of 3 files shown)


http/unstable_file_server_test.ts
```diff
@@ -0,0 +1,1069 @@
...
+        `--port=${port}`,
+        "http/unstable_file_server.ts",
+        "--dir-listing",
+        "http/tesdata",
+      ],
+      stdin: "null",
+      stdout: "piped",
+      stderr: "piped",
+    }).spawn();
+    await readUntilMatch(child.stderr, "Listening");
```
iuioiua, line 1059: Surprised to see that the output of `deno serve` prints to `stderr` instead of `stdout`. Is this a bug?

http/unstable_file_server_test.ts
```diff
@@ -0,0 +1,1069 @@
...
+    const response = await fetch(`http://localhost:${port}`);
+    assertEquals(response.status, 200);
```
iuioiua, line 1061: Haven't been able to look deep into this. Any help would be appreciated.

kt3k:
The entrypoint name seems suggesting this will be merged to `file-server` in the future, but I don't think we can do that (this script only work with `deno serve`, not with `deno run`)

Maybe better to give a different name?

iuioiua:
Yes, I should've clarified that my intention was that this was to replace the current `@std/http/file-server`. My reasoning is that it's worth the breakage thanks to `deno serve` doing the heavy lifting of CLI argument parsing and the ability to parallelise requests with `deno serve --parallel`. Though, I know I'm more liberal with breaking changes than most.

Alternatively, we can depend on https://github.com/denoland/deno/issues/29447 to be implemented as to minimise breaking changes and avoiding having to create a new endpoint.

iuioiua:
I'll take another approach if https://github.com/denoland/deno/pull/29812 is merged.

PR #6670, opened 2025-05-22
chore: support up to Deno LTS and refine supported Deno versions logic

### LINKED ISSUE #6669: suggestion: align support with Deno runtime
By only supporting nothing past the Deno LTS version. As Deno 2.1 is the current latest LTS version, this means dropping support for Deno 1. To make it official we'd need to state this fact in the README and in https://docs.deno.com/runtime/fundamentals/stability_and_releases/#long-term-support-(lts).

iuioiua:
This PR changes the codebase such that it supports up to the Deno LTS version. The new `@std/internal/support` endpoint also helps with ensuring functionality is supported in the Deno LTS version.

Closes #6669


internal/deno.json
```diff
@@ -9,6 +9,7 @@
     "./diff": "./diff.ts",
     "./format": "./format.ts",
     "./styles": "./styles.ts",
-    "./types": "./types.ts"
+    "./types": "./types.ts",
+    "./support": "./support.ts"
```
iuioiua, line 13: While mostly used internally, the public endpoint here is created for use by other public APIs.

internal/support.ts
iuioiua, line 1: The idea here is that when code or tests will automatically fail once the Deno LTS version is upgraded in CI. Then, we can go in and update the relevant code, dropping support for functionality that's outside of LTS.

internal/deno.json
kt3k: Sounds like recommending other libs to use this endpoint? I'm not in favor of that policy

kt3k:
I don't think we're ready to make this change for the same reason as https://github.com/denoland/std/pull/6624#issuecomment-2832151328

kt3k:
> These changes are not breaking. They do not force someone to upgrade from Deno v1 to v2
>
> In other words, a person can upgrade their packages after this PR is applied and nothing will break. It'll only change their expected error types and nothing else if they upgrade to Deno 2.

This PR stops CI in v1. The changes after this PR can randomly break Deno v1 and they will be forced to update at that time.

The last Deno v1.x version was released only 8 months ago https://github.com/denoland/deno/releases/tag/v1.46.3 while Node still supports v20 which was released more than 2 years ago https://nodejs.org/en/about/previous-releases I feel it's too early to start breaking it intentionally. Also we (std maintainers/developers) are not so blocked by the requirement of v1.x support as far as I'm aware. I don't see much benefit of making this change.

iuioiua:
For those reasons, I'd say this is safe to do.

PR #6672, opened 2025-05-22
chore: enable dependency caching between GA workflow runs

iuioiua:
It's possible that caching may not be the best idea for some of these cases, but I think it's worth just trying and seeing.


iuioiua:
@crowlKats any ideas about these leaks in Ubuntu?

kt3k (approved):
LGTM

PR #6673, opened 2025-05-22
chore: change badge color

iuioiua:
Looks cooler and suits better, given the Deno logo is now green.


kt3k (approved):
LGTM

> given the Deno logo is now green

BTW, I wonder why they make Deno logo green now..

PR #6675, opened 2025-05-23
fix(cli/unstable): fix resource leak of ProgressBar

kt3k:
This fixes the resource leak of ProgressBar on windows. This fixes the main CI.


dsherret (approved):
LGTM

PR #6676, opened 2025-05-24
refactor(cli/unstable): print initial `ProgressBar`

timreichen:
This PR adds a `print()` call when the progress bar is initialized (`setInterval()` only starts printing after the delay) and removes the `print()` call when it is stopped.

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -295,9 +295,17 @@ export class ProgressBar {
    */
   async stop(): Promise<void> {
     clearInterval(this.#id);
-    await this.#writer.write(this.#clear ? "\r\u001b[K" : "\n")
-      .then(() => this.#writer.close())
-      .then(() => this.#pipePromise)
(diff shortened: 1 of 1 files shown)


timreichen:
> > The line gets cleared which makes the print obsolete.
> 
> 
> 
> Then how about only printing when `this.#clear` is false?
> 
> 
> 
> > The line stays and a newline gets appended. I think it is reasonable to assume that the current print is kept when calling `stop()`. With https://github.com/denoland/std/pull/6664#pullrequestreview-2867473861, the state gets updated more regularly anyway.
> 
> 
> 
> There are still some delay between the last value/max assignment and printing of them (and the last printing can be skipped if we stop in the middle of delay). If we print at `stop`, that can show more correct data.

Ok that makes sense, with https://github.com/denoland/std/pull/6664 (once the issues are solved) the print function is will check whether there is an update and return early if not. I'll update this PR.

kt3k (approved):
LGTM

PR #6678, opened 2025-05-24
refactor(cli/unstable): remove brackets from strings

timreichen:
This PR removes the default addition of the `[]` brackets from formatter elements. This makes it easier to use a custom format without string slicing.


kt3k:
Can you also update `Custom Formatting` example in jsdoc of the class?

PR #6679, opened 2025-05-25
refactor(cli/unstable): string based tests

timreichen:
This PR replaces buffer comparisons with string comparisons.
This makes the test easier to read and understand.

### DIFF
--- a/cli/unstable_progress_bar_test.ts
+++ b/cli/unstable_progress_bar_test.ts
@@ -13,6 +13,8 @@ async function* getData(
   }
 }
 
+const decoder = new TextDecoder();
+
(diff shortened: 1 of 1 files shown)


timreichen:
> Looks like this change caused the flakiness https://github.com/denoland/std/actions/runs/15268755976/job/42939262153?pr=6677

The flakiness might be due to randomness in `getData()`. What if we remove `random()` there?

kt3k:
I feel the flakiness caused by dependency to timers (The 2nd print seems to have happened at completely unreasonable timing). Can we introduce `FakeTime`? like,

```js
using _ = new FakeTime();
```

PR #6686, opened 2025-05-27
refactor(cli/unstable): use `Date.now()` internally and use `FakeTime` for testing

timreichen:
Ref: https://github.com/denoland/std/pull/6684

Seems like the random delay in `setData()` could lead to tests being flaky.
This PR removes the function altogether and adds additional tests.

### DIFF
--- a/fs/unstable_rename_test.ts
+++ b/fs/unstable_rename_test.ts
@@ -115,7 +115,8 @@ Deno.test(
 
 Deno.test(
   "rename() rejects with Error when an existing directory is renamed with an existing regular file path",
-  { ignore: RENAME_HAS_ISSUE },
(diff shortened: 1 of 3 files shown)


kt3k:
Seems still flaky https://github.com/denoland/std/actions/runs/15269321101/job/43006417785?pr=6686

BlackAsLight:
Since your expected results changes based on execution time, you can't have a static expected output. You're going to need something more flexible to confirm the results of the output or it will continue to appear flaky at times.

timreichen:
I updated the `ProgressBar` class so it uses `Date` instead of `performance` internally, so it can be tested with `FakeTime`. This should give more consistent results.

PR #6691, opened 2025-05-28
chore: replace license-checker cli  call with explicit code

timreichen:
**Reasoning**
While a simple cli call is neat in theory, there are several issues that come with this:
- addition of tool config file (`.licenserc.json`): I think we should avoid config files at all costs if possible. It pollutes the repo and makes it harder to maintain, if one is not familiar with the tool and does not know where it is being used.
- not updatable: `deno outdated` does not work for task strings.

Therefore I suggest removing the `.licenserc.json` file in favour of adding a more verbose `check_license.ts` file that does the same task with an inlined config.


kt3k (approved):
LGTM

kt3k:
blocked by https://github.com/denoland/std/pull/6695

PR #6692, opened 2025-05-28
BREAKING(cli/unstable): remove `ProgressBarFormatter.styledTime` and use `DurationFormat`

timreichen:
We should not provide an opinionated time style in the `ProgressBarFormatter` object, we already provide a defaultFormatter which styles the time.
Instead we should point to `Intl.DurationFormat` which allows for all kinds of styles.


BlackAsLight:
This change doesn't format correctly. It puts everything under the seconds section.
```ts
new Intl.DurationFormat(undefined, {
  minutes: "2-digit",
  seconds:"2-digit",
  fractionalDigits: 0
})
  .format({milliseconds: Date.now()});
// "00:1748597655"
```
Screenshot 2025-05-30 at 19 35 27

timreichen:
> This change doesn't format correctly. It puts everything under the seconds section.
> 
> ```ts
> new Intl.DurationFormat(undefined, {
>   minutes: "2-digit",
>   seconds:"2-digit",
>   fractionalDigits: 0
> })
>   .format({milliseconds: Date.now()});
> // "00:1748597655"
> ```
> 
> Screenshot 2025-05-30 at 19 35 27

Good catch!
I think this has to wait until `Temporal` is stable. Then that can be solved with [Temporal.Duration.prototype.round()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Temporal/Duration/round).
Putting this as draft for now.

bartlomieju:
`Temporal` is now stable, do you want to give it another go @timreichen?

timreichen:
> `Temporal` is now stable, do you want to give it another go @timreichen?

I am a bit busy atm. I might have time for another go on this in a few weeks.

If someone wants to take over feel free.

bartlomieju:
CI: all three `test (v1.x, ...)` jobs are failing:

```
TS2339: Property 'DurationFormat' does not exist on type 'typeof Intl'.
TS2694: Namespace 'Temporal' has no exported member 'DurationRoundingOptions'.
```

`Temporal` and `Intl.DurationFormat` don't exist in Deno v1.x. The issue is that `timeFormatter` and `DURATION_ROUNDING_OPTIONS` are module-level constants — they fail at type-check time on import, not just when `defaultFormatter` is called.

Fix options:
1. Move the `Intl.DurationFormat` construction and the `Temporal` type reference inside `defaultFormatter` (lazy init) so the module can still be imported on v1.x
2. Exclude this module from v1.x testing if v1.x support isn't required for `cli/unstable_*`

fibibot (changes requested):
1. CI v1.x is red on all three OSes — root cause is `Temporal.Duration.from(...).round(...)` in `defaultFormatter` (and `Intl.DurationFormat` at module scope). Deno v1.x's V8 doesn't ship either as global. The repo's pattern for optional `Temporal` is in `assert/equal.ts:44` (`const Temporal = (globalThis as any).Temporal ?? Object.create(null)`), but that only works when a non-Temporal code path exists — here the new `defaultFormatter` has no fallback.

2. The simplest fix that keeps the BREAKING removal of `ProgressBarFormatter.styledTime` (the actual goal): keep the `mm:ss` math inline in `defaultFormatter` (lift the four lines from the old `styledTime` getter). The public API change stands; users who want richer styles can construct `Intl.DurationFormat`/`Temporal.Duration` themselves from the exposed `formatter.time`. The PR description's "point to `Intl.DurationFormat`" goal is satisfied by *removing* the opinionated `styledTime` field, not by hard-wiring `Intl.DurationFormat` into the default.

- nit: `largestUnit: "minutes"` rolls hours into minutes (90 min → `90:00`), which matches old behaviour; flagging only because it's not obvious from the format string.

PR #6693, opened 2025-05-28
refactor(cli/unstable): move unit related stuff to separate file

timreichen:
This PR removes `ProgressBarFormatter.styledData` and moves all unit related functions and types inside `_unit.ts`.
**Reasoning**
There are several kinds of ProgressBars. A download progress bar is unit based, but percentage based and task based progress bars do not use units at all.
```sh
[##########----------------------------------------] 20%
```

```sh
[##########----------------------------------------] 20/100
```

The default formatter styles the units already.
Imo we should not special treat the unit based progress bar by providing a opinionated `styledData` and instead provide some kind of unit formatter in `@std/fmt/bytes` which can be pointed to for custom formats, binary as well as decimal.


kt3k:
> There are several kinds of ProgressBars. A download progress bar is unit based, but percentage based and task based progress bars do not use units at all.

I see this point, but I don't think removing `styledData` solves this

timreichen:
> > There are several kinds of ProgressBars. A download progress bar is unit based, but percentage based and task based progress bars do not use units at all.
> 
> I see this point, but I don't think removing `styledData` solves this

Ok, I reverted the removal of `ProgressBarFormatter.styledData` for now.
This is related to https://github.com/denoland/std/pull/6696 and the code overlaps somewhat and could be reduced in a future PR.

kt3k:
Note: This increases the amount of calculation because `unit` and `rate` are calculated at each rendering. However this also makes the output more correct because now we allow to change `max` value of progress (`unit` and `rate` can be changed at each rendering)

PR #6696, opened 2025-05-29
refactor(fmt/bytes): add unit arrays

timreichen:
This PR adds unit arrays which makes it more readable and avoids string manipulation.
Several other small changes like `num = Math.abs(num)` only for negative numbers and pre-calculation of `Math.log(1024)`.

### DIFF
--- a/fmt/bytes.ts
+++ b/fmt/bytes.ts
@@ -71,6 +71,56 @@ export interface FormatOptions {
   maximumFractionDigits?: number;
 }
 
+const BINARY_UNITS = [
+  "B",
+  "kiB",
+  "MiB",
+  "GiB",
+  "TiB",
+  "PiB",
+  "EiB",
+  "ZiB",
+  "YiB",
+];
+
+const DECIMAL_UNITS = [
+  "B",
+  "kB",
+  "MB",
+  "GB",
+  "TB",
+  "PB",
+  "EB",
+  "ZB",
+  "YB",
+];
+
+const BINARY_BIT_UNITS = [
+  "b",
+  "kibit",
+  "Mibit",
+  "Gibit",
+  "Tibit",
+  "Pibit",
+  "Eibit",
+  "Zibit",
+  "Yibit",
+];
+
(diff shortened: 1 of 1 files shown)


fmt/bytes.ts
```diff
@@ -70,6 +65,74 @@ export interface FormatOptions {
    */
   maximumFractionDigits?: number;
 }
+interface Unit {
+  long: string;
```
kt3k, line 69: `long` doesn't seem used?

fmt/bytes.ts
```diff
@@ -70,6 +65,74 @@ export interface FormatOptions {
...
+  short: string;
+  altShort?: string;
```
kt3k, line 71: Why do we need these 2 fields?

fmt/bytes.ts
timreichen: I added long as a clear description and for completeness, similar to [Intl.NumberFormat](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#unit_2).

fmt/bytes.ts
timreichen: The current implementation deviates from the IEC standard names (for example `Kibibyte` uses `KiB` according to IEC but `kiB` in the current implementation). `altShort` is to avoid breaking changes.

kt3k: I think it's strange to include unused values in the source code.

kt3k:
Let's not do this. It seems very confusing to include many of unused values in the source code

timreichen:
> Let's not do this. It seems very confusing to include many of unused values in the source code

This was in preparation for adding the `unitDisplay` as an option. I've updated the PR so it is introduced here to avoid unused values.
One question remains whether `short` and `altShort` should coexist. This implementation deviates from the IEC standard names (for example Kibibyte uses `KiB` according to IEC but `kiB` in the current implementation) which seems like an error imo, but I am not sure if that is the consensus.

PR #6699, opened 2025-05-30
BREAKING(cli/unstable): replace `...` with up and down arrows

timreichen:
After trying the api, `...` seems not too intuitive, nothing indicates if there is more content above.
This PR changes the more content indication `...` to a `↓` arrow and additionally it indicates if moving up is possible with a `↑` arrow. 

```sh
What country are you from?
❯ Brazil
  United States
  Japan
  ↓
```
...
```sh
What country are you from?
  ↑
  United States
  Japan
❯ China
  ↓
```
...
```sh
What country are you from?
  ↑
  China
  Canada
❯ Spain
```


kt3k:
Is this way of indicating 'more content' common in general? What do other tools do for this situation?

cc @WasixXD

WasixXD:
Some tools just loop over (like [Inquirer select](https://github.com/SBoudrias/Inquirer.js/tree/main/packages/select)) and others don't even bother, but I find the arrows a nice fit

kt3k:
I personally prefer `...` for indicating 'more contents' which feels more common to me.

↑ and ↓ in this context feel confusing to me. (I don't remember the use case where ↑ and ↓ indicating 'more contents')

timreichen:
> I personally prefer `...` for indicating 'more contents' which feels more common to me.
> 
> 
> 
> ↑ and ↓ in this context feel confusing to me. (I don't remember the use case where ↑ and ↓ indicating 'more contents')

The arrows have the additional benefit of not just indicating `more contents`, but also the location.
`...` seems fine until one scrolls and content gets out of bounds above, which is not very intuitive to me.

kt3k:
> The arrows have the additional benefit of not just indicating `more contents`, but also the location. `...` seems fine until one scrolls and content gets out of bounds above, which is not very intuitive to me.

I think that situation can be covered by showing `...` above.

kt3k:
Added some examples to demonstrate both patterns (full and partial) of displaying options

PR #6700, opened 2025-05-31
chore(text/unstable): use locale-agnostic Intl.Segmenter

lionel-rowe:
In practice I don't think this can ever make a difference as there are no specific word segmentation rules for `en-US`, but it's more semantically correct.

https://unicode.org/reports/tr35/tr35.html#unknown-or-invalid-identifiers

### DIFF
--- a/text/unstable_slugify.ts
+++ b/text/unstable_slugify.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
 
-const wordSegmenter = new Intl.Segmenter("en-US", { granularity: "word" });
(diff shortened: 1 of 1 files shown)


0f-0b:
This won't work. `Intl.Segmenter.supportedLocalesOf("und")` returns an empty array, which means that `new Intl.Segmenter("und")` will use whatever the host's default locale is and not the intended “undetermined” locale.

lionel-rowe:
> This won't work. Intl.Segmenter.supportedLocalesOf("und") returns an empty array, which means that new Intl.Segmenter("und") will use whatever the host's default locale is and not the intended “undetermined” locale.

Huh, go figure. Tbh I'm not sure (?) if there even currently exist any locales that segment differently in V8, so I'll close this.

PR #6701, opened 2025-05-31
feat(text/unstable): add `toTitleCase()` and `toSentenceCase()`

lionel-rowe:
Closes https://github.com/denoland/std/issues/5424

`toTitleCase` with default options implements [3.13.2 Default Case Conversion `toTitlecase`](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34078).


text/unstable_to_title_case.ts
```diff
@@ -0,0 +1,83 @@
...
+ * import { assertEquals } from "@std/assert/equals";
+ *
+ * assertEquals(toTitleCase("deno is awesome"), "Deno Is Awesome");
+ * ```
+ */
+export function toTitleCase(input: string, options?: TitleCaseOptions): string {
+  const opts = resolveOptions(options);
+  // [3.13.2 Default Case Conversion `toTitlecase`](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34078)
+  // toTitlecase(X): Find the word boundaries in X according to Unicode Standard Annex #29, “Unicode Text Segmentation.”
+  const segments = [...opts.segmenter.segment(input)];
```
kt3k: I wonder this splitting of words matches the user's expectation.

Other `toFooCase` APIs use `splitToWords` utility, and it splits camel case, pascal case, kebab case, constant case, etc into parts, but this doesn't do it. Is that difference intentional?

ex.
```js
toKebabCase("HelloWorld") // => hello-world
toSnakeCase("HelloWorld") // => hello_world
toConstantCase("HelloWorld") // => HELLO_WORLD
toTitleCase("HelloWorld") // => Helloworld // <- this is one word in this case
```

kt3k (commented):
Mostly looks good to me. Nice work!

Left one question

text/unstable_to_title_case.ts
lionel-rowe: > I wonder this splitting of words matches the user's expectation.
> 
> Other `toFooCase` APIs use `splitToWords` utility, and it splits camel case, pascal case, kebab case, constant case, etc into parts, but this doesn't do it. Is that difference intentional?
> 
> ex.
> 
> ```js
> toKebabCase("HelloWorld") // => hello-world
> toSnakeCase("HelloWorld") // => hello_world
> toConstantCase("HelloWorld") // => HELLO_WORLD
> toTitleCase("HelloWorld") // => Helloworld // <- this is one word in this case
> ```

In my informal taxonomy of "dev-first" vs "user-first", all the existing functions are dev-first, whereas `toTitleCase` and `toSentenceCase` are user-first (i.e. suitable for user-facing text). The option-less `toTitleCase` is intended as a spec-compliant implementation of Unicode's [3.13.2 Default Case Conversion `toTitlecase`](https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34078), with `toSentenceCase` (which isn't in any spec) and the various options being variations for some common practical use cases. Namely, `toSentenceCase` is like `toTitlecase` but stops after the first word, and the `force: false` option disables lower-casing of existing chars that are already in uppercase/titlecase.

What you'd probably want for examples like `HelloWorld` is `force: false`, which would leave it untouched, rather than splitting to words. Consider examples like "more videos on YouTube" — it would be surprising and probably unwanted for "YouTube" to become "You Tube".

Or a more involved example (I've changed the `filter` option to an `exclude` option, which now allows combining multiple stop word/filter functions):

https://github.com/denoland/std/blob/a03542b385cb950777ce75bfdccb92ee7bfe7ba5/text/unstable_to_title_case_test.ts#L86-L98

kt3k: Ok. Probably this parsing of the input makes sense for `toTitleCase` and `toSentenceCase`. Can you describe how these 2 APIs handle the input in API docs, with some notes about the difference with other `toXCase` APIs?

text/_title_case_util.ts
```diff
@@ -0,0 +1,69 @@
...
+  locale: boolean | NonNullable<Intl.LocalesArgument>,
+): Intl.LocalesArgument {
+  return locale === true ? undefined : locale === false ? "und" : locale;
+}
+
+export function titleCaseWord(word: string, opts: ResolvedOptions): string {
+  // For each word boundary, find the first cased character F following the word boundary. If F exists, map F to Titlecase_Mapping(F)
+  // https://www.unicode.org/versions/Unicode16.0.0/core-spec/chapter-3/#G34000
+  // A character C is defined to be cased if and only if C has the Lowercase or Uppercase property or has a General_Category value of Titlecase_Letter.
+  const match = /[\p{Lowercase}\p{Uppercase}\p{Titlecase_Letter}]/u.exec(word);
```
0f-0b: This regex can also be written as `/\p{Cased}/u`.

text/_title_case_util.ts
```diff
@@ -0,0 +1,69 @@
...
+  if (match == null) return word;
+
+  const rest = word.slice(match.index + match[0].length);
+  return word.slice(0, match.index) + titleCaseChar(match[0], opts.locale) +
+    (opts.force ? rest.toLocaleLowerCase(opts.locale) : rest);
```
0f-0b: Since Lowercase_Mapping is context-dependent, converting a substring to lowercase is not as simple as taking the substring, converting it to lowercase out of context, and then splicing it back into the original string. As an example, although toLowercase(“Σ”) is “σ”, toTitlecase(“ΩΣ”) is “Ως” instead of “Ωσ”.

text/_title_case_util.ts
```diff
@@ -0,0 +1,69 @@
...
+  const locale = localeOptionToLocalesArgument(opts.locale);
+  const segmenter = new Intl.Segmenter(locale, { granularity: "word" });
+
+  return { ...opts, locale, segmenter };
+}
+
+function localeOptionToLocalesArgument(
+  locale: boolean | NonNullable<Intl.LocalesArgument>,
+): Intl.LocalesArgument {
+  return locale === true ? undefined : locale === false ? "und" : locale;
```
0f-0b: I checked ECMA-402—for `toLocaleLowerCase` and `toLocaleUpperCase` specifically, unavailable locales don't fall back to the default locale but to the root locale, so this time `"und"` actually works.

text/unstable_to_title_case.ts
lionel-rowe: Added a note

text/_title_case_util.ts
lionel-rowe: Fixed by doing the case conversion within the context of the input word. The only context-independent logic is now the mapping of `Titlecase_Letter` chars, but I don't think any of those (or their lower/uppercase variants) can have context-dependent properties.

kt3k (approved):
LGTM

text/_title_case_util.ts
```diff
@@ -0,0 +1,74 @@
...
+export function titleCaseWord(word: string, opts: ResolvedOptions): string {
+  // For each word boundary, find the first cased character F following the word boundary. If F exists, map F to Titlecase_Mapping(F)
+  const match = /[\p{Cased}]/u.exec(word);
+  if (match == null) return word;
+
+  const before = word.slice(0, match.index);
+  const cased = word.slice(match.index);
+  const first = titleCaseChar(cased, opts.locale);
+  const _rest = opts.force ? cased.toLocaleLowerCase(opts.locale) : cased;
+  const rest = _rest.slice(_rest[Symbol.iterator]().next().value?.length ?? 0);
```
0f-0b: This is not quite right. Some characters such as “İ” may turn into multiple characters when lowercased. I'm afraid fixing this would require essentially reimplementing the lowercasing algorithm.

Also I found an edge case related to the character I just mentioned. Without tailorings, `toTitleCase("I\u{307}", { locale: "tr" })` would return `"I"` (the dot is lost), while CLDR correctly returns `"I\u{307}"`. What do you think about this? Should we stick to not using any tailorings for simplicity, drop localization support for now, or somehow adopt CLDR tailorings?

0f-0b: I implemented CLDR tailorings by compiling ICU4X to Wasm on the [`wasm-casemap`](https://github.com/0f-0b/deno_std/tree/wasm-casemap) branch. In addition, I changed all boolean options to default to `false` following best practices. Please feel free to incorporate any of my code into this PR.

lionel-rowe: @0f-0b Thanks, great eye for detail, and the wasm-casemap version looks interesting! But wasm importing of ICU4X and its respective locale data certainly feels like it _ought to_ be overkill for title casing, given that so much of the logic and locale data is already supported by the runtime's built-in ICU support.

It also looks like ICU4X's title casing goes above and beyond the "Default Case Conversion toTitlecase" algorithm, which only operates at the character level (specifically, there's no way it could possibly support Dutch "IJ" as two separate source chars, whereas ICU4X does).

I think it should (?) be possible to implement ICU4X-compatible title casing in pure JS, partially by using a grapheme segmenter to solve the `I\u{307}` problem, which is what I'm trying now. It relies on the invariant that one source grapheme can be uppercased or titlecased to multiple graphemes (`ﬃ -> FFI`, `ﬃ -> Ffi`), but it can never be _lowercased_ to multiple graphemes (`ﬃ -> ﬃ`), which I think (??) always holds true.

PR #6705, opened 2025-06-02
refactor(toml): use RegExp groups for month validation

timreichen:
https://github.com/denoland/std/pull/6704#discussion_r2120340492

### DIFF
--- a/toml/_parser.ts
+++ b/toml/_parser.ts
@@ -680,11 +680,11 @@ export function dateTime(scanner: Scanner): ParseResult<Date> {
   // special case if month is February
   if (groups.month == "02") {
     const days = parseInt(groups.day);
-    const year = parseInt(groups.year);
     // more than 29 days
     if (days > 29) {
       throw new SyntaxError(`Invalid date string "${match}"`);
     }
+    const year = parseInt(groups.year);
     // more than 28 days and is not leap year
     if (days > 28 && !isLeap(year)) {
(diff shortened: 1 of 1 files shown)


toml/_parser.ts
```diff
@@ -2,6 +2,7 @@
 // This module is browser compatible.
 
 import { deepMerge } from "@std/collections/deep-merge";
+import { isLeap } from "@std/datetime";
```
kt3k: `@std/toml` cannot depend on `@std/datetime` as `toml` is stable and `datetime` is unstable. This causes lint CI failure

timreichen: fixed

toml/_parser.ts
```diff
@@ -666,27 +675,27 @@ export function float(scanner: Scanner): ParseResult<number> {
...
-    const year = parseInt(match.substring(0, 4));
-    //more than 29 days
+  const string = match[0];
+  scanner.next(string.length);
+  const groups = match.groups as { year: string; month: string; day: string };
+  // special case if month is February
+  if (groups.month == "02") {
+    const days = parseInt(groups.day);
     if (days > 29) {
       throw new SyntaxError(`Invalid date string "${match}"`);
```
andyvu923, line 691: In these thrown SyntaxErrors should `${match}` be changed to `${string}`?

kt3k (approved):
LGTM 👍

PR #6708, opened 2025-06-03
feat(streams/unstable): add `AbortStream`

BlackAsLight:
This pull request adds an `AbortStream` class which lets you abort a stream via pipeThrough when neither ends of the operation allow you to pass a `AbortSignal`.

```ts

try {
  await (await Deno.open("./deno.json"))
    .readable
    .pipeThrough(new AbortStream(controller.signal))
    .pipeThrough(new CompressionStream("gzip"))
    .pipeTo((await Deno.create("deno.json.gzip")).writable);
} catch {
  await Deno.remove('deno.json.gzip');
}
```


iuioiua:
Great idea!

timreichen:
What is the reason that the `signal` param is optional?

BlackAsLight:
> What is the reason that the `signal` param is optional?

To allow other modules that accept an optional signal to pass it here without having to write extra logic to check if they actually received a signal. 

So if I added the option to pass a signal to my `encodePNG` function, idk if the user actually passed one in or not. If this stream required one to be passed in then I'd need to write extra logic on my side to actually check whether I can pipe it through this AbortStream. 

Essentially it just makes the API neater to use.

kt3k:
> So if I added the option to pass a signal to my encodePNG function, idk if the user actually passed one in or not. If this stream required one to be passed in then I'd need to write extra logic on my side to actually check whether I can pipe it through this AbortStream.

I feel that logic belongs to `encodePNG` in that case, and not `AbortStream`.

It seems a bit confusing to me that `new AbortStream()` is actually no-op stream.

timreichen:
> > So if I added the option to pass a signal to my encodePNG function, idk if the user actually passed one in or not. If this stream required one to be passed in then I'd need to write extra logic on my side to actually check whether I can pipe it through this AbortStream.
> 
> 
> 
> I feel that logic belongs to `encodePNG` in that case, and not `AbortStream`.
> 
> 
> 
> It seems a bit confusing to me that `new AbortStream()` is actually no-op stream.

I had the same thought. I think the signal should be required.

kt3k (approved):
LGTM

Issue #6636, opened 2025-05-01
Run node test runner in darwin and windows in CI


kt3k:
Node test runner currently only run in linux in CI. We should expand them to darwin and windows

PR #6711, opened 2025-06-04
chore: add Darwin and Windows OS's to node test runner

jbronder:
Addresses #6636. This PR adds Windows and MacOS the `matrix.os` variable to expand the CI workflow of the Node test runner.

### DIFF
--- a/fs/unstable_write_file_test.ts
+++ b/fs/unstable_write_file_test.ts
@@ -82,7 +82,8 @@ Deno.test("writeFile() handles 'create' when writing to a file", async () => {
   const encoder = new TextEncoder();
   const data = encoder.encode("Hello");
 
-  // Rejects with NotFound/Error when file does not initally exist.
+  // Rejects with NotFound when file does not initally exist.
+  // TODO: This should throw NotFound on Windows.
   if (platform() === "win32") {
     await assertRejects(async () => {
       await writeFile(testFile, data, { create: false });
@@ -102,7 +103,11 @@ Deno.test("writeFile() handles 'create' when writing to a file", async () => {
 
   // Overwrites the existing file with new content.
   const dataAgain = encoder.encode("Hello, Standard Library");
-  await writeFile(testFile, dataAgain, { create: false });
+  if (platform() === "win32") {
+    await writeFile(testFile, dataAgain);
+  } else {
+    await writeFile(testFile, dataAgain, { create: false });
+  }
   const dataReadAgain = await readFile(testFile);
   const readDataAgain = decoder.decode(dataReadAgain);
   assertEquals(readDataAgain, "Hello, Standard Library");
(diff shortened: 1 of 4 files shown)


jbronder:
Some of the `@std/fs` APIs are throwing when running tests with `deno task test:node` under the `windows-latest`. Should updates to those affected tests be addressed in separate, forthcoming PR(s)?

kt3k:
> Should updates to those affected tests be addressed in separate, forthcoming PR(s)?

We generally don't merge PRs to main which makes CI check red. Can you try to address those issues in this PR?

jbronder:
That makes sense, I'll take a look.

fs/unstable_write_file_test.ts
```diff
@@ -82,10 +82,16 @@ Deno.test("writeFile() handles 'create' when writing to a file", async () => {
...
 
-  // Rejects with NotFound when file does not initally exist.
-  await assertRejects(async () => {
-    await writeFile(testFile, data, { create: false });
-  }, NotFound);
+  // Rejects with NotFound/Error when file does not initally exist.
+  if (platform() === "win32") {
+    await assertRejects(async () => {
+      await writeFile(testFile, data, { create: false });
+    }, Error);
```
kt3k, line 90: Can you also add `TODO` comment above this assertion? Ideally this should also throw `NotFound` on windows

```js
// TODO: This should throw NotFound
```

jbronder: Sure thing!

jbronder:
Quick progress update: I decided to tackle `writeFile(Sync)` APIs. I'm addressing one API set at a time. I did end up making some modifications to the `getWriteFsFlag` function to handle Windows differently from MacOS and Linux. This is documented in the function in comments. I'm imagining that `writeTextFile(Sync)` would be similar.

fibibot (changes requested):
`test-node (latest, windows-latest)` is FAILURE on this head — exactly the matrix entry this PR adds. The PR is stale (last commit 2025-06-13), so CI hasn't re-run since then either.

The Windows branch in `fs/_get_fs_flag.ts:25-32` produces the same final flag as the non-Windows path for every value of `opt.create`: starting with `O_CREAT | O_WRONLY` and XOR'ing off O_CREAT when `!opt.create` yields the same result as starting with `O_WRONLY` and OR'ing in O_CREAT only when `opt.create` is truthy. The "O_CREAT set by default … `{ create: true }` is the default" comment doesn't match what the code actually does. Either drop the Windows branch (it's a no-op) or make the behaviour match the comment.

Separately, each `writeFile`/`writeTextFile` test now carries `TODO: This should throw NotFound on Windows.` and accepts a bare `Error`. That defers the real fix @kt3k asked about. With the flag identical to Linux, O_WRONLY on a missing file should surface as ENOENT/NotFound on Windows too — worth diagnosing why it doesn't before widening the matcher.

- nit: `flag ^= O_CREAT` works only because O_CREAT was unconditionally set above; `flag &= ~O_CREAT` says "clear this bit" without that assumption.

PR #6713, opened 2025-06-05
feat(cli/unstable): add generic for `promptSelect()` and `promptMultipleSelect()`

MAKS11060:
This PR adds generic for
 - `promptSelect()`
 - `promptMultipleSelect()`

If we pass a read-only array, then the output value is already known.

```ts
const result = promptSelect('Select browser:', ["safari", "chrome", "firefox"])
result // 'safari' | 'chrome' | 'firefox' | null

const result2 = promptMultipleSelect('Select browser:', ["safari", "chrome", "firefox"])
result2 // ('safari' | 'chrome' | 'firefox')[] | null
```

### DIFF
--- a/cli/unstable_prompt_multiple_select_test.ts
+++ b/cli/unstable_prompt_multiple_select_test.ts
@@ -660,8 +660,11 @@ Deno.test("promptMultipleSelect() check return types", () => {
     | ("safari" | "chrome" | "firefox")[]
     | null;
 
-  promptMultipleSelect("Please select a browsers:", [...browsers, 'edge']) satisfies
-    | ("safari" | "chrome" | "firefox"| 'edge')[]
+  promptMultipleSelect("Please select a browsers:", [
+    ...browsers,
+    "edge",
+  ]) satisfies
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Nice!

kt3k:
Does anyone have any concern on this change?

tomas-zijdemans:
~ ~Nice! 👏 
I think you can simplify: 
`return [...selectedIndexes].map((it) => values[it] as string) as T[];` 
to: 
`return [...selectedIndexes].map((it) => values[it]) as T[];`~ ~

~ ~Oh, and probably you want to update the doc now that you added generics: `Returns the user's selected value as string.`~ ~

Shoot. I see main has had significant changes after this PR.. 😊 @kt3k: I suggest we close this one.

kt3k:
We switched to different pattern of input parameters to make the API more flexible in https://github.com/denoland/std/commit/7a77939d682cd54d2d8cd6c8de4cec5c492d05c4
 and this change can't be applied now. Thanks for the contribution anyways @MAKS11060

PR #6714, opened 2025-06-05
refactor(streams): throw `RangeError` within `Buffer` methods

iuioiua:
This error type is more fitting for the given error messages. This change is non-breaking as `RangeError` extends `Error`.

### DIFF
--- a/streams/buffer.ts
+++ b/streams/buffer.ts
@@ -322,7 +322,7 @@ export class Buffer {
       return;
     }
     if (n < 0 || n > this.length) {
-      throw new Error(
+      throw new RangeError(
         `Buffer truncation value "${n}" is not between 0 and ${this.length}`,
       );
     }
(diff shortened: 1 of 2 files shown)


kt3k:
> This change is non-breaking as RangeError extends Error.

I don't think this is technically correct. If someone doing error handling like the below, these new errors handled in different path (A) than the current path (B):

```js
try {
  // do many things including buffer.truncate()
  // which could throw both Error and RangeError
} catch (e) {
  if (e instanceof RangeError) {
    // handle range errors <- A
  } else {
    // handle others <- B
  }
}
```

I'm not sure this change is worth it.

Does anyone have any opinions?

iuioiua:
In your example, users would've used `e instanceof Error` which would've evaluated to `true` before and after this change. Exceptions throw error classes that are appropriate for the given error in the rest of the codebase, rather than the generic `Error`. I don't see a sufficient reason to be inconsistent.

kt3k (approved):
Ok. Let's try this

LGTM

PR #6728, opened 2025-06-19
chore: cleanup tasks

iuioiua:


deno.json
```diff
@@ -26,13 +26,11 @@
     "typos": "typos -c ./.github/workflows/typos.toml",
     "build:crypto": "deno task --cwd crypto/_wasm wasmbuild",
     "wasmbuild": "deno run -A jsr:@deno/wasmbuild@0.19.2 --js-ext mjs --inline",
-    "cov:test": "deno task test --coverage --clean --reporter=dot && deno task cov:html",
     "cov": "deno coverage --ignore=\"**/*.generated.mjs,**/_test_utils.ts,.github\"",
     "cov:lcov": "deno task cov --lcov --output=cov.lcov",
-    "cov:html": "deno task cov --html",
```
iuioiua, line 32: `deno test --coverage` in `deno task test` does this.

deno.json
```diff
@@ -7,9 +7,9 @@
     "noUncheckedIndexedAccess": true
   },
   "importMap": "./import_map.json",
+  "unstable": ["webgpu", "unsafe-proto", "fs"],
```
kt3k: We'd like to apply `--unstable-unsafe-proto` only to specific task.

related https://github.com/denoland/std/pull/6342

iuioiua: What's the harm in enabling `--unstable-unsafe-proto` in all tests? Genuine question. Also is this test supposed to pass with `deno task test`? Enabling `--unstable-unsafe-proto` doesn't seem to change test behaviour or results.

https://github.com/denoland/std/blob/2258bf2628a97a03dece8d0235d910bfaf1f501d/assert/object_match_test.ts#L417-L435

kt3k: > What's the harm in enabling --unstable-unsafe-proto in all tests?

With `--unstable-unsafe-proto` enabled, the code can depend on `__proto__` behavior, which doesn't work with default state of Deno. So the future version of std could potentially be not working with default Deno

> Enabling --unstable-unsafe-proto doesn't seem to change test behaviour or results.

That is because we fixed the behavior difference between with and without unsafe-proto in https://github.com/denoland/std/pull/6342/files. Before the fix, it was failing with `--unstable-unsafe-proto`

kt3k: We should test without `--unstable-unsafe-proto` if we target Deno. We should test with `--unstable-unsafe-proto` if we target browser, node, bun, etc. Because we now target both groups of runtime (Deno) vs (Browser, Node, etc), it makes sense to run tests with and without `--unstable-unsafe-proto`.

kt3k (approved):
LGTM

PR #6730, opened 2025-06-20
feat: add callback functionality for binary search tree

dquixote-jr:
I need the functionality to find the k'th node in the tree, which isn't possible with current setup
This augmentation allows people to define whatever functions that they want, in this particular case, it allows me to store the subtree size in a parameter `st_size` and which I can use to find the k'th node in order.

First javascript project PR so this is gonna have some issues, happy to solve them.


data_structures/_test_utils.ts
```diff
@@ -8,4 +8,5 @@ export class MyMath {
 export interface Container {
   id: number;
   values: number[];
+  st_size: number;
```
kt3k: Can you use camelCase?

data_structures/binary_search_tree.ts
```diff
@@ -104,12 +105,18 @@ export class BinarySearchTree<T> implements Iterable<T> {
    * @param compare A custom comparison function to sort the values in the tree.
    * By default, the values are sorted in ascending order.
    */
-  constructor(compare: (a: T, b: T) => number = ascend) {
+  constructor(compare: (a: T, b: T) => number = ascend, callback?: (node: BinarySearchNode<T>) => void) {
```
kt3k: Need to document `callback` param in jsdoc.

kt3k:
> I need the functionality to find the k'th node in the tree, which isn't possible with current setup

This change doesn't seem a good idea for adding that capability to this class. Currently `BinarySearchNode` class is private. This makes it a part of public API. That doesn't seem an ideal change. Can we achieve this without exposing node class public?

For small k's, it seems we can use `lnrValues` (or `rnlValues`) iterator for finding k'th node. What do your tree size and `k` value typically look like?

data_structures/binary_search_tree.ts
dquixote-jr: Added the param.

data_structures/_test_utils.ts
dquixote-jr: done.

dquixote-jr:
> For small k's, it seems we can use lnrValues (or rnlValues) iterator for finding k'th node. What do your tree size and k value typically look like?

The tree can be large, like ~5K. And k can be as large as O(n), so iterating over values with lnr/rnl isn't really a good idea.

> This change doesn't seem a good idea for adding that capability to this class. Currently BinarySearchNode class is private. 

I don't think it's the worst thing to make the BinarySearchNode a public class, as we have already made it a "binary" tree, so having left/right/parent elements is expected of it. A different thing to do would be to expose the BinarySearchNode class, but not give access directly to member objects but rather have setChild/getChild or set/get Left/Right/Parent kind of functions.   
I'm open to any other ideas too, my main use case is to allow people to travel over the tree via nodes somehow via left/right nodes. I don't need them to be able to access the specific internal attributes of the class, just left/right children.

kt3k:
> I don't think it's the worst thing to make the BinarySearchNode a public class, as we have already made it a "binary" tree, so having left/right/parent elements is expected of it. A different thing to do would be to expose the BinarySearchNode class, but not give access directly to member objects but rather have setChild/getChild or set/get Left/Right/Parent kind of functions.

Then maybe it's better to define a new interface which only has `left`, `right`, and `parent` (and maybe `value`) fields, and use it as input type for callback?

Also run `deno task lint` and see the output. We need to fix them to merge this to main

dquixote-jr:
> Then maybe it's better to define a new interface which only has `left`, `right`, and `parent` (and maybe `value`) fields, and use it as input type for callback?

That sounds like a good idea. I've added a new class BSTNode and refactored the setup to use that.

> Also run `deno task lint` and see the output. We need to fix them to merge this to main

Thank you for this. I've got it passing locally. `import type` gave me a headspin but I got it working in the end :)

data_structures/bst_node.ts
```diff
@@ -0,0 +1,113 @@
...
+ * const node = new BSTNode<number>(null, 42);
+ * assertEquals(node.value, 42);
+ * assertEquals(node.left, null);
+ * assertEquals(node.right, null);
+ * assertEquals(node.parent, null);
+ * ```
+ *
+ * @typeparam T The type of the values stored in the binary tree.
+ */
+export class BSTNode<T> {
```
kt3k: We usually avoid using this type of acronym in API names. BinarySearchTree used to be BSTree but we renamed it later https://github.com/denoland/std/pull/2400 I think this should be named `BinarySearchTreeNode`.

Also I think this should be typescript `interface` instead of `class`. We can avoid unnecessary runtime overhead by making this `interface`

data_structures/bst_node.ts
```diff
@@ -0,0 +1,113 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/**
+ * A generic Binary Search Tree (BST) node class.
```
kt3k: Nicely documented 👍

kt3k (commented):
This now looks much better to me. Left some comments

data_structures/bst_node.ts
dquixote-jr: OK, renamed to BinarySearchTreeNode and made into an interface.

dquixote-jr:
Any holdups here?

bartlomieju:
Closing due to inactivity — this PR has been stale for over 7 months. There were unresolved design questions from @kt3k about whether exposing the internal node structure publicly is the right approach for this feature, and the PR has outstanding coverage gaps.

@dquixote-jr — thanks for the contribution and for addressing the initial review feedback. If you'd like to revisit this, feel free to open a fresh PR. It may be worth opening an issue first to align on the API design (callback approach vs. exposing nodes vs. adding a dedicated `kthValue()` method) before writing the implementation.

Issue #6731, opened 2025-06-21
@std/txt longest common prefix


lionel-rowe:
**Is your feature request related to a problem? Please describe.**

Find the longest common prefix of multiple strings.

**Describe the solution you'd like**

```ts
declare function longestCommonPrefix(strs: ArrayLike<string>): string
assertEquals(longestCommonPrefix([]), '')
assertEquals(longestCommonPrefix(['']), '')
assertEquals(longestCommonPrefix(['flower', 'flow', 'flight']), 'fl')
assertEquals(longestCommonPrefix(['flower', 'glow', 'flight']), '')
assertEquals(longestCommonPrefix(['💩', '💫']), '')
```

**Describe alternatives you've considered**

Maybe an equivalent under `collections` for longest common prefixes of lists (array/`ArrayLike`s)?

PR #6738, opened 2025-06-24
fix(text/unstable): `dedent()` correct blank line handling

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6737

### DIFF
--- a/text/unstable_dedent.ts
+++ b/text/unstable_dedent.ts
@@ -1,9 +1,10 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 // This module is browser compatible.
+import { longestCommonPrefix } from "./unstable_longest_common_prefix.ts";
 
 const WHITE_SPACE = String.raw`\t\v\f\ufeff\p{Space_Separator}`;
 const INDENT_REGEXP = new RegExp(
-  String.raw`^([${WHITE_SPACE}]*)[^${WHITE_SPACE}]`,
+  String.raw`^[${WHITE_SPACE}]+`,
   "u",
 );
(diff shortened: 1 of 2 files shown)


WWRS:
Thanks for working on this.

While we're here, I just want to make sure we match ECMAScript's definition of whitespace, which includes many chars other than space and tab: https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-white-space

In addition, I realized that the reference implementation of dedent tries to match the exact whitespace at the starts of lines rather than the number. So `[space][tab]` and `[tab][space]` should be treated as different. Does it make sense to fix that here or to move that to a different PR? (I noticed that #6731 may be useful for that, so a different PR would make sense.)

lionel-rowe:
> While we're here, I just want to make sure we match ECMAScript's definition of whitespace, which includes many chars other than space and tab: https://tc39.es/ecma262/multipage/ecmascript-language-lexical-grammar.html#sec-white-space

@WWRS Good point, I guess the regex should be `/[\t\v\f\ufeff\p{Space_Separator}]/u` (which conveniently still doesn't match line breaks).

> In addition, I realized that the reference implementation of dedent tries to match the exact whitespace at the starts of lines rather than the number. So [space][tab] and [tab][space] should be treated as different. Does it make sense to fix that here or to move that to a different PR? (I noticed that https://github.com/denoland/std/issues/6731 may be useful for that, so a different PR would make sense.)

Good question. That PR's probably closer to landing than this one, so maybe just wait for that to land first.

kt3k (approved):
LGTM. Thanks for the fix! @lionel-rowe 

Thanks for your review too! @WWRS

PR #6742, opened 2025-06-26
feat(cli/unstable): valued prompt items

crowlKats:


kt3k:
Looks reasonable to me. Could you also add an example of using valued items?

cli/unstable_prompt_multiple_select.ts
```diff
@@ -6,6 +6,27 @@ export interface PromptMultipleSelectOptions {
...
+  : PromptEntryWithValue<V>;
+
+/**
+ * A {@linkcode PromptEntry} with an underlying value.
+ *
+ * @typeParam V The value of the underlying Entry.
+ */
+export interface PromptEntryWithValue<V> {
+  /** The title for this entry. */
+  title: string;
```
timreichen, line 31: I think `label` as key would be nicer than `title` like in https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/option#label

cli/unstable_prompt_select.ts
```diff
@@ -12,6 +12,27 @@ export interface PromptSelectOptions {
...
+  : PromptEntryWithValue<V>;
+
+/**
+ * A {@linkcode PromptEntry} with an underlying value.
+ *
+ * @typeParam V The value of the underlying Entry.
+ */
+export interface PromptEntryWithValue<V> {
+  /** The title for this entry. */
+  title: string;
```
timreichen, line 31: Dito

PR #6744, opened 2025-06-27
fix(fs,internal,path): unify `isWindows` implementations

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6743

This also unifies a couple of internal test utils (`disposableStack`, `stubProperty`) as I also used them for testing `checkWindows()`.


fs/_to_file_info.ts
```diff
@@ -1,7 +1,7 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
 import type { FileInfo } from "./unstable_types.ts";
-import { isWindows } from "./_utils.ts";
+import { isWindows } from "../internal/_os.ts";
```
kt3k: This relative import doesn't work when it's published to JSR. (Cross package imports need to start with package name like `@std/internal/...`, but `_os.ts` is not part of public API of `@std/internal`.

internal/_os_test.ts
```diff
@@ -0,0 +1,65 @@
...
+
+// deno-lint-ignore no-explicit-any
+const global = globalThis as any;
+
+Deno.test("checkWindows() returns `true` on Windows on various runtimes", async (t) => {
+  await t.step("Deno", () => {
+    using stack = disposableStack();
+    stack.use(stubProperty(global, "Deno", { build: { os: "windows" } }));
+    stack.use(stubProperty(global, "navigator", undefined));
+    stack.use(stubProperty(global, "process", undefined));
```
kt3k, line 14: Is this better than 3 using statements? (Do we need `DisposableStack` polyfill?)

```suggestion
    using _denoStub = stubProperty(global, "Deno", { build: { os: "windows" } });
    using _navigatorStub = stubProperty(global, "navigator", undefined);
    using _processStub = stubProperty(global, "process", undefined);
```

kt3k (approved):
Now `isWindows` check is a bit complex because of multiple runtime support. It makes sense to introduce a new unified util in `@std/internal`

LGTM

internal/_os_test.ts
lionel-rowe: Not much in it except for importing the polyfill vs creating 3 unused variables instead of one. `DisposableStack` is supported in latest Deno, only partial-polyfilled for testing due to lack of support in Deno 1.x, Node, etc. (the other usage is in `mockLittleEndian` in `random/_test_utils.ts`, which uses `defer` instead of `use`).

kt3k: Ok. Makes sense

PR #6752, opened 2025-07-05
fix(http): allow HEAD method in file-server

### LINKED ISSUE #6750: Allow HEAD mathod in file-server

When using the http file server from the CLI as [documented](https://jsr.io/@std/http/doc/file-server) with the following command:

    deno run --allow-net --allow-read --allow-sys jsr:@std/http/file-server

The HEAD http method returns a 405 method not allowed.

Implement the HEAD method in the file server.

This was discussed in Discord and looks like [serveFile](https://jsr.io/@std/http/doc/file-server/~/serveFile) does support HEAD but [serveDir](https://jsr.io/@std/http/doc/file-server/~/serveDir) does not. Apparently the discrepancy does not make much sense.

krugerLucas:
## Summary

Fixes inconsistency where `serveFile()` supported HEAD requests but `serveDir()` returned 405 Method Not Allowed.

## Problem

- `serveFile()` correctly handles HEAD requests (returns headers without body)
- `serveDir()` rejects HEAD requests with 405 Method Not Allowed  
- This creates inconsistent behavior and breaks HTTP compliance

## Solution

- Changed line 658 in `serveDir()` to allow HEAD requests alongside GET
- HEAD requests now properly delegate to `serveFile()` which already has correct implementation
- Added comprehensive tests for both file and directory HEAD requests

## Changes

- `http/file_server.ts`: Allow HEAD method in serveDir validation
- `http/file_server_test.ts`: Add tests for serveDirHeadRequest scenarios

## Test Plan

- [x] New tests pass: `serveDirHeadRequest` and `serveDirHeadRequestForDirectoryWithIndex`
- [x] Existing tests still pass 
- [x] HEAD returns correct headers without body
- [x] POST/PUT/DELETE still properly rejected with 405

Fixes #6750


kt3k:
This doesn't change `createServeDirResponse` and request body is still returned when the file server returns directory listing (You can check `assert(!res.body)` fails with the url `http://localhost/nested-subdirs/subdir/`). We need to modify `createServeDirResponse` to return null body response when the request method is HEAD (in a similar way as `serveFile`)

wilkmaciej:
@kt3k we probably can close this

kt3k:
closing in favor of https://github.com/denoland/std/pull/7024

thanks for your contribution anyway

PR #6754, opened 2025-07-08
feat(data-structures/unstable): 2d array

crowlKats:


timreichen:
Is this PR related to a problem or has a popular npm package?
This data structure looks like it is intended for very specialized use cases and doesn't look too efficient on a first glance.
It might be worth thinking about implementing efficient `mat2/3/4` (and possibly `vec2/3/4`) if that is the intended usage.

kt3k:
> Is this PR related to a problem or has a popular npm package? This data structure looks like it is intended for very specialized use cases and doesn't look too efficient on a first glance.

This is for implementing (unstable) barcode module in the near future.

I think these methods are enough general and reusable if someone needs to interact with two-dimentional data structure

>  It might be worth thinking about implementing efficient `mat2/3/4` (and possibly `vec2/3/4`) if that is the intended usage.

I think multidimentional version would be useful too. Let's keep iterating.

kt3k (approved):
LGTM

PR #6756, opened 2025-07-09
feat(cli/unstable): Introduce Ansi module

BlackAsLight:

There are many static methods and properties here. This is not an exhaustive list of what is available. Nor is everything guaranteed to work in every terminal. But I have tested each one and they all work in my terminal.

This example creates a very big "Hello World!" in terminals that support these Ansi codes.
```ts
import * as Ansi from "@std/cli/unstable_ansi";

console.log(
  Ansi.doubleHeightTop +
  "Hello World!\n" +
  Ansi.doubleHeightBottom + 
  "Hello World!",
);
```

No tests as of yet have been written for this module as I am not sure how to write unit tests to test its behaviour. The codes are either correct or incorrect and whether they work in the terminal is dependent on whether the terminal supports said code.


kt3k:
I agree with @iuioiua . Class usage like this seems strange to me.

timreichen:
> IMHO, this should be implemented as pure functions and constants, rather than a class. We've also stepped away from using `enum`s in this codebase.

I remember a meeting when it was decided to have classes only in std when they have a changeable state (I think it was a discussion about `CSV` functions vs static class like `JSON`.
).
`import * as Ansi from "..."` will group exports together and works similar to an utility class, except functions are tree-shakable and can also be imported separately if needed.

BlackAsLight:
22 of the 43 codes would just be exporting constant variables. 

To add to the enum bit. It is a const enum so it disappears when the code is transpiled. The numbers behind the enum is what's important for the code and using a const enum to represent each number adds to readability.

BlackAsLight:
I am never attempting a rebase again

cli/unstable_ansi.ts
```diff
@@ -0,0 +1,961 @@
...
+ *
+ * @example Usage
+ * ```ts ignore
+ * import * as Ansi from "@std/cli/unstable_ansi";
+ *
+ * console.log(Ansi.doubleHeightTop + "Hello World!");
+ * console.log(Ansi.doubleHeightBottom + "Hello World!");
+ * ```
+ */
+export const doubleHeightTop = "\x1b#3";
```
iuioiua: Names of constants are usually in capitalised snake case. Ditto elsewhere.

cli/unstable_ansi.ts
```diff
@@ -0,0 +1,961 @@
...
+ * these export function s and variables does not automatically change the
+ * terminal settings. Only once they are passed to stdout or stderr will they
+ * take effect.
+ *
+ * These codes were based off the
+ * [xterm reference](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html).
+ *
+ * @example Basic Usage
+ * ```ts ignore
+ * import * as Ansi from "@std/cli/unstable_ansi";
```
iuioiua: Ditto elsewhere.
```suggestion
 * import * as Ansi from "@std/cli/unstable-ansi";
```

cli/unstable_ansi.ts
```diff
@@ -0,0 +1,961 @@
...
+ * console.log(Ansi.doubleHeightBottom + "Hello World!");
+ * ```
+ *
+ * @module
+ */
+
+/**
+ * Causes content on the current line to enlarge, showing only the top half
+ * of characters with each character taking up two columns. Can be used in
+ * combination with {@linkcode Ansi.doubleHeightBottom} on the next line to
```
iuioiua:
```suggestion
 * combination with {@linkcode doubleHeightBottom} on the next line to
```
AFAIK fixes linking. Ditto elsewhere.

cli/unstable_ansi.ts
```diff
@@ -0,0 +1,961 @@
...
+ */
+export const hardReset = "\x1bc";
+
+/**
+ * This command resets many settings to their initial state without fully
+ * reinitializing the terminal like {@linkcode Ansi.hardReset}. It preserves
+ * things like cursor position and display content, but clears modes,
+ * character sets, etc. Should probably be called when exiting the program.
+ *
+ * @returns string The ANSI escape code for soft resetting the terminal.
```
iuioiua:
```suggestion
 * @returns The ANSI escape code for soft resetting the terminal.
```
No need to declare the return type in JSDocs thanks to TypeScript. Ditto elsewhere.

iuioiua (changes requested):

cli/unstable_ansi.ts
```diff
@@ -0,0 +1,961 @@
...
+ */
+export const enum CursorStyle {
+  Default = 0,
+  BlinkingBlock = 1,
+  SteadyBlock = 2,
+  BlinkingUnderline = 3,
+  SteadyUnderline = 4,
+  BlinkingBar = 5,
+  SteadyBar = 6,
+}
```
kt3k, line 975: Is `const enum` equivalent of the below?

```ts
export const CursorStyle = {
  Default: 0,
  BlinkingBlock: 1,
  SteadyBlock: 2,
  BlinkingUnderline: 3,
  SteadyUnderline: 4,
  BlinkingBar: 5,
  SteadyBar: 6,
} as const;

export type CursorStyle = typeof CursorStyle[keyof typeof CursorStyle];
```

I'm wondering if using `const enum` makes the code more readable.

kt3k:
Mostly looks good to me. Left some question

cli/unstable_ansi.ts
BlackAsLight: I imagine the typings would be the same but the transpiled result is different. 

The const enum would transpiled `setCursorStyle(CursorStyle.BlinkingBlock)` to `setCursorStyle(1)`.

kt3k: Ok. It has some benefit on runtime

kt3k:
Let's try and iterate

PR #6757, opened 2025-07-09
feat(cli): Introduce StaticLine

BlackAsLight:

This class uses the scrollable region Ansi code to allow scrolling content to not overwrite lines at the top and bottom of the terminal. It is meant to provide a foundation for anyone looking to easily have a line at the bottom or top of the terminal.
```ts
import { delay } from "@std/async/delay";
import { StaticLine } from "@std/cli/unstable-static-line";

const id = setInterval(() => console.log(Math.random()), 1000);

const line = new StaticLine();
await line.write("Hello World!");
await delay(1500);
await line.write("How are you?");
await delay(1500);
await line.write("Doing good?");
await delay(1500);
await line.releaseLine();
```

No tests as of yet have been written for this class as I am not sure how to write unit tests to test its behaviour.

### DIFF
--- a/cli/deno.json
+++ b/cli/deno.json
@@ -5,6 +5,7 @@
     ".": "./mod.ts",
     "./parse-args": "./parse_args.ts",
     "./prompt-secret": "./prompt_secret.ts",
+    "./unstable-ansi": "./unstable_ansi.ts",
     "./unstable-progress-bar": "./unstable_progress_bar.ts",
     "./unstable-progress-bar-stream": "./unstable_progress_bar_stream.ts",
     "./unstable-prompt-select": "./unstable_prompt_select.ts",
--- a/cli/deno.json
+++ b/cli/deno.json
@@ -6,6 +6,7 @@
     "./parse-args": "./parse_args.ts",
     "./prompt-secret": "./prompt_secret.ts",
     "./unstable-ansi": "./unstable_ansi.ts",
+    "./unstable-get-cursor": "./unstable_get_cursor.ts",
     "./unstable-progress-bar": "./unstable_progress_bar.ts",
     "./unstable-progress-bar-stream": "./unstable_progress_bar_stream.ts",
     "./unstable-prompt-select": "./unstable_prompt_select.ts",
--- a/cli/unstable_ansi.ts
+++ b/cli/unstable_ansi.ts
@@ -3,8 +3,8 @@
 /**
  * Ansi is a class with static methods and properties that returns various Ansi
  * Escape Sequences. This class is not an exhaustive list of what is possible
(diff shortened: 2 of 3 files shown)


tomas-zijdemans:
~~Perhaps this PR should be closed for now? A way forward could be to get `StaticLine` ready for inclusion, then rewrite `Spinner` to use it. Then we can see if a use case for `askForCursorPositionSync` pops up in the future? 🤷~~

Actually, I think we need this one first to solve other problems

kt3k:
@BlackAsLight sorry for the delay, but can you resolve the conflict?

timreichen:
Why not call it `getCursorPosition()`?

Also I think `Ansi` should not be a class with static methods because no internal state is used and returning a const value does not need a function call.
The file should export consts instead which then can be imported as `import * as Ansi from "..."`

BlackAsLight:
> Also I think `Ansi` should not be a class with static methods because no internal state is used and returning a const value does not need a function call.
> The file should export consts instead which then can be imported as `import * as Ansi from "..."`

https://github.com/denoland/std/pull/6756#pullrequestreview-3003918438

timreichen:
> > Also I think `Ansi` should not be a class with static methods because no internal state is used and returning a const value does not need a function call.
> > The file should export consts instead which then can be imported as `import * as Ansi from "..."`
> 
> https://github.com/denoland/std/pull/6756#pullrequestreview-3003918438

I don't see any issues with just exporting const codes.
Alternatively we could export an object `ANSI_CODE` like we do in `http` for `STATUS_CODE`.

BlackAsLight:
> > > Also I think `Ansi` should not be a class with static methods because no internal state is used and returning a const value does not need a function call.
> > > The file should export consts instead which then can be imported as `import * as Ansi from "..."`
> > 
> > https://github.com/denoland/std/pull/6756#pullrequestreview-3003918438
> 
> I don't see any issues with just exporting const codes.
> Alternatively we could export an object `ANSI_CODE` like we do in `http` for `STATUS_CODE`.

I linked to the other comment in the other pull request because this is the change we've already made there. It just needs to be updated over here.

timreichen:
> > > > Also I think `Ansi` should not be a class with static methods because no internal state is used and returning a const value does not need a function call.
> > > > The file should export consts instead which then can be imported as `import * as Ansi from "..."`
> > > 
> > > https://github.com/denoland/std/pull/6756#pullrequestreview-3003918438
> > 
> > I don't see any issues with just exporting const codes.
> > Alternatively we could export an object `ANSI_CODE` like we do in `http` for `STATUS_CODE`.
> 
> I linked to the other comment in the other pull request because this is the change we've already made there. It just needs to be updated over here. 

Ah, I see now. Sorry for spamming.

PR #6758, opened 2025-07-09
feat(cli/unstable): Introduce StaticLine

BlackAsLight:

https://github.com/denoland/std/pull/6756 should be merged before this.

This class uses the scrollable region Ansi code to allow scrolling content to not overwrite lines at the top and bottom of the terminal. It is meant to provide a foundation for anyone looking to easily have a line at the bottom or top of the terminal.
```ts
import { delay } from "@std/async/delay";
import { StaticLine } from "@std/cli/unstable-static-line";

const id = setInterval(() => console.log(Math.random()), 1000);

const line = new StaticLine();
await line.write("Hello World!");
await delay(1500);
await line.write("How are you?");
await delay(1500);
await line.write("Doing good?");
await delay(1500);
await line.releaseLine();
```

No tests as of yet have been written for this class as I am not sure how to write unit tests to test its behaviour.

### DIFF
--- a/cli/unstable_static_line.ts
+++ b/cli/unstable_static_line.ts
@@ -91,7 +91,7 @@ export class StaticLine {
   async write(line: string): Promise<void> {
     if (this.#released) throw new ReferenceError("Line has been released");
     await write(
-      Ansi.SAVE_CURSOR_POSITION +
+      Ansi.SAVE_CURSOR +
         Ansi.DISABLE_ORIGIN_MODE +
         Ansi.DISABLE_AUTO_WRAP +
         Ansi.setCursorPosition(
@@ -101,7 +101,7 @@ export class StaticLine {
         ) +
         Ansi.ERASE_LINE +
         line +
-        Ansi.RESTORE_CURSOR_POSITION,
+        Ansi.RESTORE_CURSOR,
     );
   }
 
@@ -136,7 +136,7 @@ export class StaticLine {
     const top = linesAtTop.length + 1;
     const bottom = rows - linesAtBottom.length;
     await write(
-      Ansi.SAVE_CURSOR_POSITION +
+      Ansi.SAVE_CURSOR +
         (this.#atTop
           ? (top - index
             ? Ansi.setScrollableRegion(index + 1, top + 1) +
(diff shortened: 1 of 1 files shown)


BlackAsLight:
## Example
```ts
import { delay } from "@std/async/delay";
const id = setInterval(() => console.log(Math.random()), 1000);

for (let i = 0; i < 20; ++i) {
  const line = new StaticLine();
  await line.write(`Line (${i}): Placeholder!`);
  const id = setInterval(
    () => line.write(`Line (${i}): ${Math.random()}`),
    Math.random() * 3000 + 1000,
  );
  setTimeout(() => {
    clearInterval(id);
    line.releaseLine();
  }, Math.random() * 5_000 + 25_000);
  await delay(Math.random() * 1000 + 1000);
}

await delay(30_000);
clearInterval(id);
```

https://github.com/user-attachments/assets/16194467-e094-4305-bc04-9f82f1de03ec

kt3k:
Can you provide some real-world example (cli/library/whatever) in which this type of output is used?

Also does this type of abstraction have prior example?

BlackAsLight:
> Can you provide some real-world example (cli/library/whatever) in which this type of output is used?
> 
> 
> 
> Also does this type of abstraction have prior example?

I don't know of any real world examples, but as I learn more and more about ANSI escape sequences it is becoming apparent just how unreliable terminals can be. 

This type, or some type, of abstraction is needed to "reliably" support multiple spinners and progress bars at once. As request here: https://github.com/denoland/std/issues/5037. The word reliably is doing a lot of heavy lifting here as any other area that starts inserting sequences can screw this all up. 

If we want something truely robust, we'll need to offer some type of complete abstraction away from ansi codes. A framework so to say that isn't merely offering small functions here and there, but that specifies the entire state and keeps track of it all. 

Very little of the terminals state can actually be asked or looked up from the terminal itself. Much of these settings and modes must be remembered in code to know if it's enabled or disabled. With the sync functions this isn't really a problem as you can't have random console logs from other areas screwing everything up by sending their own sequences, but with the spinner and progress bar and any other functionality that allows other code to run, allows for the state to be screwed up.

tomas-zijdemans:
I support this change, it's a good step forwards and allows us to rewrite `Spinner`, `ProgressBar`, and possibly other functions on top of it. I would suggest to: 
- make it internal,  `_static_line.ts` ?
- add tests
- consider using `Deno.stderr.writeSync` instead of `Deno.stderr.write`. Like `Spinner` does it.

PR #6761, opened 2025-07-12
feat(testing/unstable): add `stubProperty` function

lionel-rowe:
Closes https://github.com/denoland/std/issues/6760


kt3k (approved):
Looks reasonable addition to me

LGTM

PR #6763, opened 2025-07-15
fix(http): respect colors and other `Deno.inspect` options when logging `UserAgent`s

lionel-rowe:
Current:

image

After fix:

image


lionel-rowe:
Failing CI unrelated, exit code 143 with 0 failed tests (?)

kt3k:
The CI now seems failing because of the use of Deno specific types (The file has `browser compatible` header and is type-checked with browser type settings). Maybe let's do something like the below?

```js
  [Symbol.for("Deno.customInspect")](
    inspect: (value: unknown, options: unknown) => string,
    options: unknown,
  ): string {
```

PR #6765, opened 2025-07-16
feat(net/unstable): add ip utilities

### LINKED ISSUE #6722: Add  IP utils on @std/net
**Is your feature request related to a problem? 

While adding a feature to Fresh, it was recommended that this functionality—being a desirable, general-purpose process for @std/net—should be addressed by creating an issue in this repository.
This process itself was ported from Hono.

There’s a general-purpose process that would be desirable to add to @std/net.

I’d like to add a set of utility functions.

Implementation into Fresh itself

What I mainly want to prepare is…
- Determine whether an IP is IPv4 or IPv6
- Validation process to check whether a given IP matches a specified static IP or CIDR notation.

WasixXD:
Hopefully closes #6722

This PR adds the `isIPv4(addr: string): boolean` and `isIPv6(addr: string): boolean` functions


http/file_server.ts
```diff
@@ -517,20 +517,22 @@ function dirViewerTemplate(dirname: string, entries: EntryInfo[]): string {
...
-                if (depth == 0) {
-                  link = ".";
-                } else {
-                  link = "../".repeat(depth);
-                }
-                // deno-fmt-ignore
-                return html`<a href="${link}">${escape(path)}</a>`;
-              })
-              .join("/")}/
+            Index of ${
```
iuioiua: It seems these formatting changes were made because your Deno version is out-of-date. Please update Deno and run `deno fmt` to revert these changes.

net/deno.json
```diff
@@ -4,6 +4,7 @@
   "exports": {
     ".": "./mod.ts",
     "./get-available-port": "./get_available_port.ts",
+    "./ip-utils": "./ip_utils.ts",
```
iuioiua:
```suggestion
    "./ip": "./ip.ts",
```
This is sufficient

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+const IPV4_REGEX = /^[0-9]{0,3}\.[0-9]{0,3}\.[0-9]{0,3}\.[0-9]{0,3}$/;
```
iuioiua: How did you get the regex? Genuine question. Do you have a reference that you can link to?

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+export const distinctRemoteAddr = (
+  remoteAddr: string,
+): AddressType | undefined => {
+  if (isIPv4(remoteAddr)) {
+    return "IPv4";
+  }
+  if (isIPv6(remoteAddr)) {
+    return "IPv6";
+  }
+};
```
iuioiua: IMHO I have a feeling this function simply isn't needed and adding it now might be premature. If it were up to me, I'd wait till there's sufficient demand for such a function before adding it to the API.

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Validates whether a given string is a IPv4 address
+ *
+ * @param addr - IPv4 address in a string format (e.g., "192.168.0.1").
+ * @returns a boolean indicating if the string is a IPv4 address
```
iuioiua:
```suggestion
 * @param addr IPv4 address in a string format (e.g., "192.168.0.1").
 * @returns A boolean indicating if the string is a IPv4 address
```
Nits

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * Validates whether a given string is a valid IPv4 address
+ *
+ * @param addr - IPv4 address in a string format (e.g., "192.168.0.1").
+ * @returns a boolean indicating if the string is a valid IPv4 address
+ */
+export function isValidIPv4(addr: string): boolean {
+  if (addr === "localhost") {
```
iuioiua: `localhost` isn't an IP address.

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+  return addr.includes(":");
+}
+
+/**
+ * Validates whether a given string is a valid IPv4 address
+ *
+ * @param addr - IPv4 address in a string format (e.g., "192.168.0.1").
+ * @returns a boolean indicating if the string is a valid IPv4 address
+ */
+export function isValidIPv4(addr: string): boolean {
```
iuioiua: Having both `isValidIPv4()` and `isIPv4()` is confusing. We should likely just have one, and I think it should be named `isIPv4()`. Ditto for the IPv6 function.

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/**
+ * Validates whether a given string is a IPv6 address
+ *
+ * @param addr - IPv6 address in a string format (e.g., "2001:db8::1").
+ * @returns a boolean indicating if the string is a IPv6 address
+ */
+export function isIPv6(addr: string): boolean {
+  return addr.includes(":");
```
iuioiua: This isn't correct

net/ip_utils.ts
```diff
@@ -0,0 +1,137 @@
...
+      const bits = parseInt(bitStr, 0);
+      const zeroes = 32 - bits;
+      const mask = bits === 0 ? 0 : (~0 << zeroes) >>> 0;
+
+      return (ipToInt(address) & mask) === (ipToInt(range) & mask);
+    }
+  }
+
+  return false;
+}
```
iuioiua: Perhaps, this can be implemented in a follow up PR, as to get this one landed sooner.

net/ip_utils_test.ts
```diff
@@ -0,0 +1,96 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+import { assert } from "../assert/assert.ts";
+import { assertEquals } from "../assert/equals.ts";
+import { assertFalse } from "../assert/false.ts";
```
iuioiua:
```suggestion
import { assert, assertEquals, assertFalse } from "@std/assert"; 
```
If you want.

iuioiua (changes requested):
Quick pass

net/ip_utils.ts
WasixXD: Got it from https://github.com/denoland/fresh/pull/3035/files i thought it was clever

0f-0b (changes requested):
The syntax of valid IP addresses is [described in the URL standard](https://url.spec.whatwg.org/#host-writing). Please follow that.

Also, I think it would be a waste to [parse only to throw the result away](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/), so I suggest adding functions that return the parsed address. They can internally share an implementation like this:

<details>

```ts
function ipv4Parser(str: string, buf: Uint8Array): boolean {
  // the actual implementation…
}

function ipv6Parser(str: string, buf: Uint8Array): boolean {
  // maybe reuse `ipv4Parser` for IPv6 addresses in the third form…
}

const buf = new Uint8Array(16);

export function isIPv4(str: string): boolean {
  return ipv4Parser(str, buf);
}

export function isIPv6(str: string): boolean {
  return ipv6Parser(str, buf);
}

export function parseIPv4(str: string): Uint8Array | null {
  return ipv4Parser(str, buf) ? buf.slice(0, 4) : null;
}

export function parseIPv6(str: string): Uint8Array | null {
  return ipv6Parser(str, buf) ? buf.slice(0, 16) : null;
}

export function parseIPv4Into(str: string, buf: Uint8Array): boolean {
  if (buf.length < 4) {
    throw new TypeError("IPv4 output buffer must have a length of at least 4");
  }
  return ipv4Parser(str, buf);
}

export function parseIPv6Into(str: string, buf: Uint8Array): boolean {
  if (buf.length < 16) {
    throw new TypeError("IPv6 output buffer must have a length of at least 16");
  }
  return ipv6Parser(str, buf);
}
```

</details>

net/ip.ts
```diff
@@ -0,0 +1,84 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+/**
+ * Validates whether a given string is a valid IPv4 address
```
iuioiua:
```suggestion
 * Validates whether a given string is a valid IPv4 address.
```
Nit: punctuation

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * @param addr IPv4 address in a string format (e.g., "192.168.0.1").
+ * @returns A boolean indicating if the string is a valid IPv4 address
+ *
+ * @example Check if the address is a IPv4
+ * ```ts no-assert ignore
```
iuioiua:
```suggestion
 * ```ts
```
I don't see a reason why this should be ignored. Ditto for the other snippets.

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+ * import { isIPv4 } from "@std/net/ip"
+ * import { assert, assertFalse } from "@std/assert"
+ *
+ * const correctIp = "192.168.0.1"
+ * const incorrectIp = "192.168.0.256"
+ *
+ * assert(isIPv4(correctIp))
+ * assertFalse(isIPv4(incorrectIp))
+ *
```
iuioiua:
```suggestion
```
Nit

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+  for (const part of parts) {
+    const n = parseInt(part);
+
+    // n >= 256 or is negative
+    if (n >= Math.pow(2, 8) || n < 0) {
+      return false;
+    }
+  }
+
+  return true;
```
iuioiua:
```suggestion
  const octets = addr.split(".");
  return octets.length === 4 && octets.every((octet) => {
    const n = parseInt(octet);
    return n >= 0 && n <= 255;
  });
```
This is how I would've written it. What you had before allowed strings with 4 or more octets, which is invalid.

net/mod.ts
```diff
@@ -17,3 +17,4 @@
  */
 
 export * from "./get_available_port.ts";
+export * from "./ip.ts";
```
iuioiua:
```suggestion
```
Unstable (new) modules aren't usually included in the `mod.ts` file.

net/deno.json
```diff
@@ -4,6 +4,7 @@
   "exports": {
     ".": "./mod.ts",
     "./get-available-port": "./get_available_port.ts",
+    "./ip": "./ip.ts",
```
iuioiua:
```suggestion
    "./unstable-ip": "./unstable_ip.ts",
```
I forgot that new modules need to have the "unstable" prefix.

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  for (const part of parts) {
+    const n = part == "" ? 0 : parseInt(part, 16);
+
+    if (n >= Math.pow(2, 16) || n < 0 || isNaN(n)) {
+      return false;
+    }
+  }
+
+  return true;
```
iuioiua:
```suggestion
  // more than one use of ::
  if (addr.split("::").length > 2) return false;

  const hextets = addr.split(":");

  // insert "" to the sequence when encountering a :: to normalize the hextets
  while (hextets.length < 8) {
    const idx = hextets.indexOf("");
    if (idx === -1) break;
    hextets.splice(idx, 0, "");
  }

  return hextets.length === 8 && hextets.every((hextet) => {
    const n = hextet == "" ? 0 : parseInt(hextet, 16);
    return n >= 0 && n <= 65535;
  });
```

net/ip_test.ts
```diff
@@ -0,0 +1,41 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+import { isIPv4, isIPv6 } from "./ip.ts";
+import { assertEquals } from "@std/assert";
+
+Deno.test("isIPv4() returns boolean for list of IPv4s", () => {
```
iuioiua:
```suggestion
Deno.test("isIPv4()", () => {
```
This should suffice. Ditto for the other test.

iuioiua (changes requested):

iuioiua:
> [parse only to throw the result away](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/)

IMO, the current approach is fine for now.

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * ```
+ */
+export function isIPv4(addr: string): boolean {
+  const parts = addr.split(".");
+
+  if (parts.length <= 0 || parts.length < 4) return false;
+
+  for (const part of parts) {
+    const n = parseInt(part);
```
kt3k: because `parseInt` returns `1` for `"1a"` for example, `isIPv4("1a.1.1.1")` is currently true

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    // n >= 256 or is negative
+    if (n >= Math.pow(2, 8) || n < 0) {
```
kt3k: We need to check if n is NaN. Currently `isIPv4("a.b.c.d")` seems returning true

kt3k:
IPv6 has the form of `x:x:x:x:x:x:d.d.d.d` where 0 <= x <= ffff and 0 <= d <= 255.

See `3.` of section 2.2 of RFC4291 for example https://www.rfc-editor.org/rfc/rfc4291#section-2.2

net/ip.ts
```diff
@@ -0,0 +1,84 @@
...
+ *
+ * const correctIp = "192.168.0.1"
+ * const incorrectIp = "192.168.0.256"
+ *
+ * assert(isIPv4(correctIp))
+ * assertFalse(isIPv4(incorrectIp))
+ *
+ * ```
+ */
+export function isIPv4(addr: string): boolean {
```
kt3k: I wonder if this name conforms with our naming convention. We usually avoid capitalized acronyms (like `IP`). ref https://docs.deno.com/runtime/contributing/style_guide/#naming-convention

iuioiua: I thought about this too initially but ended up just thinking this looks fine.

net/ip.ts
WasixXD: This is very clever, i didn´t know about this `every` function

iuioiua:
> IPv6 has the form of `x:x:x:x:x:x:d.d.d.d` where 0 <= x <= ffff and 0 <= d <= 255.
> 
> See `3.` of section 2.2 of RFC4291 for example https://www.rfc-editor.org/rfc/rfc4291#section-2.2

That section states that this is an additional alternative form for IPv6 addresses. I used to be a network engineer and have never used that form or seen it used. I suggest not implementing for that form until there's sufficient demand for it, which, I predict there never will be.

kt3k:
> > IPv6 has the form of `x:x:x:x:x:x:d.d.d.d` where 0 <= x <= ffff and 0 <= d <= 255.
> > See `3.` of section 2.2 of RFC4291 for example https://www.rfc-editor.org/rfc/rfc4291#section-2.2
> 
> That section states that this is an additional alternative form for IPv6 addresses. I used to be a network engineer and have never used that form or seen it used. I suggest not implementing for that form until there's sufficient demand for it, which, I predict there never will be.

In Node.js, for example, `dns.lookup("example.com", { family: 6, all: true  }, console.log)` returns the ipv6 addresses in that form.

Also URL constructor validates ip address in the hostname, and it recognizes the ipv6 address in the above form correctly (e.g. `new URL("https://[::ffff:127.0.0.1]")` parses to `hostname: "[::ffff:7f00:1]"`.

WasixXD:
I think this one is good to go then, WDYT? @kt3k

PR #6767, opened 2025-07-21
fix(testing): align types for fake timeout/interval signatures

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6766.

The only runtime change is that a custom error is thrown when `setTimeout/setInterval` is called, rather than `TypeError: timer.callback.apply is not a function` being thrown when the callback runs.

### DIFF
--- a/testing/time.ts
+++ b/testing/time.ts
@@ -127,16 +127,26 @@ interface DueNode {
 
 let time: FakeTime | undefined = undefined;
 
-function fakeSetTimeout(
-  // deno-lint-ignore no-explicit-any
-  callback: (...args: any[]) => void,
+function assertIsCallbackFunction(
+  callback: unknown,
+  fnName: "setTimeout" | "setInterval",
+): asserts callback is (...args: unknown[]) => void {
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM

kt3k:
Note: Deno's `setTimeout` seems only accepting functions, but `setTimeout` in lib.dom accepts `string | Function`.

- setTimeout in Deno https://github.com/denoland/deno/blob/fece050b60be934b671694abc5f4fbcb60a4007b/cli/tsc/dts/lib.deno.shared_globals.d.ts#L363-L370
- TimerHandler in lib.dom https://github.com/denoland/deno/blob/fece050b60be934b671694abc5f4fbcb60a4007b/cli/tsc/dts/lib.dom.d.ts#L29417

This should fix the case when testing/time is type-checked with lib.dom.d.ts

kt3k:
I created an issue in Deno about the type difference of `setTimeout` https://github.com/denoland/deno/issues/30166

PR #6771, opened 2025-07-23
feat(encoding): skip whitespace in base64 input

lionel-rowe:
Closes https://github.com/denoland/std/issues/6768

### DIFF
--- a/encoding/_common64.ts
+++ b/encoding/_common64.ts
@@ -137,6 +137,10 @@ export function decodeChunk(
   padding: number,
   retryWs: boolean,
 ) {
+  if (retryWs && buffer.length > 0 && (WHITE_SPACE[buffer.at(-1)!])) {
+    throw new RetriableError();
+  }
+
   const getHextet = (i: number): number => {
     const char = buffer[i]!;
     const hextet = alphabet[char] ?? 64;


encoding/_common64.ts
```diff
@@ -87,17 +91,70 @@ export function encode(
...
+  const indices: number[] = [];
+
+  for (let i = 0; i < length; ++i) {
+    if (WHITE_SPACE[buffer[i]!]) indices.push(i);
+  }
+
+  for (let i = 0; i < indices.length; ++i) {
+    const index = indices[i]!;
+    const start = index + 1;
+    const end = i === indices.length - 1 ? length : indices[i + 1]!;
```
BlackAsLight:
```ts
const end = indices[i + 1] ?? length;
```

encoding/_common64.ts
```diff
@@ -87,17 +91,70 @@ export function encode(
...
   o: number,
   alphabet: Uint8Array,
   padding: number,
 ): number {
+  try {
+    return _decode(buffer, i, o, alphabet, padding, true);
+  } catch (e) {
+    if (!(e instanceof RetriableError)) throw e;
+    return _decode(removeWhiteSpace(buffer), i, o, alphabet, padding, false);
+  }
```
BlackAsLight: Instead of throwing an error and catching it, I think it might be better to give the user the option in Base64Options to strip whitespace beforehand. 
```ts
if (options.ignoreWhiteSpace) {
  buffer = strip(buffer);
}
decode(buffer);
```
This way it's opt in and there isn't a penalty for if the whitespace is at the end.

encoding/_common64.ts
```diff
@@ -87,17 +91,71 @@ export function encode(
...
+}
+
+export function decodeChunk(
+  buffer: Uint8Array_,
+  i: number,
+  o: number,
+  alphabet: Uint8Array,
+  padding: number,
+  retryWs: boolean,
+) {
```
timreichen, line 139: I think this function should have an option object instead of so many args for more clarity.

encoding/_common64.ts
```diff
@@ -87,17 +91,71 @@ export function encode(
...
+  const getHextet = (i: number): number => {
+    const char = buffer[i]!;
+    const hextet = alphabet[char] ?? 64;
+    if (hextet === 64) { // alphabet.Base64.length
```
timreichen: Maybe early return on `!==` to reduce nesting?

encoding/_common64.ts
lionel-rowe: It's almost certainly a micro-optimization that has no practical effect here as this code path is very unlikely to be a bottleneck, but in v8 it's supposed to be [more performant to avoid reading beyond the length of an array](https://v8.dev/blog/elements-kinds#avoid-reading-beyond-the-length-of-the-array).

BlackAsLight: True. I just thought it looked nicer and was easier to follow the logic

encoding/_common64.ts
BlackAsLight: I don't agree. Every value is needed and the function isn't exposed to the user.

encoding/_common64.ts
lionel-rowe: > Instead of throwing an error and catching it, I think it might be better to give the user the option in Base64Options to strip whitespace beforehand.
> 
> ```ts
> if (options.ignoreWhiteSpace) {
>   buffer = strip(buffer);
> }
> decode(buffer);
> ```
> 
> This way it's opt in and there isn't a penalty for if the whitespace is at the end.

My reasoning for the current design is as follows:
- Zero, or essentially zero, performance penalty where no white space encountered (at least in the sync versions, for the stream version it empirically seems to be negligible as it's not a bottleneck)
- Minimal performance penalty for common cases of where whitespace would be encountered (e.g. after 76 chars). Actually I guess whitespace-at-end could be special-cased at negligible perf cost by just checking the last byte (edit: done)
- Consistency with `Uint8Array.fromBase64` with default options

BlackAsLight: I think the option is better because I think the majority of people will know if they have whitespace in their encoding or not, and for the very few that don't know, they can take the performance penalty for assurance.

lionel-rowe: Other than simplicity of implementation, what's the advantage of having it as an option if there's no measurable performance improvement in any plausible case? The tradeoff is complicating the API, as well as inconsistency with `Uint8Array.fromBase64`. Once `Uint8Array.fromBase64` API reaches wide availability, having it as an option would likely incur an additional performance penalty _not_ to handle whitespace:

```ts
function someBase64DecodeFunction(
  base64Data: unknown /* string/u8 arr/stream/whatever */,
  options: { allowWhiteSpace: boolean },
) {
  if (!options.allowWhiteSpace) {
    // check for and throw on presence of whitespace chars, incurring a perf cost in the process
  }
  // some code using `Uint8Array.fromBase64`, which has no option to disallow whitespace
}

BlackAsLight: When `Uint8Array.fromBase64` reaches runtimes other than Firefox. I'd imagine we'd just depreciate this API in favour of using that. Not replace the underlying logic here with that.

encoding/_common64.ts
timreichen: Yes, but it makes it hard to understand what they are for if one is not already familiar with the code.
Alternatively adding tsdoc might also be a solution.

encoding/_common64.ts
lionel-rowe: > When Uint8Array.fromBase64 reaches runtimes other than Firefox. I'd imagine we'd just depreciate this API in favour of using that. Not replace the underlying logic here with that.

There's no stream version of `Uint8Array.fromBase64` though. I was intentionally vague about what `someBase64DecodeFunction` is, cuz yeah it would likely only be the stream version that would be worth keeping.

`unstable-base64` allows `Uint8Array` as input though, so conceivably that'd be worth keeping? (But probably not as it's trivial to just decode as UTF-8 first.)

BlackAsLight:
With this coming to Deno 🔜™️, I don't know if it's still worth it to add this ability compared to just depreciating the methods. 

https://github.com/denoland/deno/pull/30686

lionel-rowe:
> With this coming to Deno 🔜™️, I don't know if it's still worth it to add this ability compared to just depreciating the methods.
> 
> [denoland/deno#30686](https://github.com/denoland/deno/pull/30686)

Fair, I'll close for now assuming there's not much demand for the stream version.

PR #6778, opened 2025-07-29
feat(text/unstable): add `trimBy` functions

lionel-rowe:
Resolves https://github.com/denoland/std/issues/6265 (another attempt at abandoned https://github.com/denoland/std/pull/6286). Similar functionality to PHP [`trim`](https://www.php.net/manual/en/function.trim.php)/[`ltrim`](https://www.php.net/manual/en/function.ltrim.php)/[`rtrim`](https://www.php.net/manual/en/function.rtrim.php).

* No replacement functionality, just removal
* Identical naming to `String#trim` family (not `strip` etc). Is this too confusing vs `String#trim`? Really it's the _same_ as `String#trim`, just with an additional argument that can be configured (the pattern to trim).
* Pattern arg:
   - Iterable of strings interpreted as its constituent members
   - Strips _all_ matches (i.e. `trim('aaxxbb', /[ab]/)` gives `xx`, not `axxb` or `axxbb`)

### DIFF
--- a/text/unstable_trim_by.ts
+++ b/text/unstable_trim_by.ts
@@ -22,6 +22,24 @@ export type TrimPattern =
  * @param pattern The pattern to trim
  * @returns The trimmed input string
  *
+ * @example Remove slashes from start and end of a path
+ * ```ts
+ * import { trimBy } from "@std/text/unstable-trim-by";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const result = trimBy("/path/to/file/", "/");
+ * assertEquals(result, "path/to/file");
+ * ```
+ *
+ * @example Remove leading and trailing line breaks
+ * ```ts
+ * import { trimBy } from "@std/text/unstable-trim-by";
+ * import { assertEquals } from "@std/assert";
+ *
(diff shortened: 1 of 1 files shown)


kt3k:
How about naming these like `trimBy`, `trimStartBy`, and `trimEndBy`? I think that makes it clear that these are different from builtin `String#trim*` APIs, and they take the 2nd args.

lionel-rowe:
> How about naming these like `trimBy`, `trimStartBy`, and `trimEndBy`? I think that makes it clear that these are different from builtin `String#trim*` APIs, and they take the 2nd args.

Done. I also removed the option to use a scalar string as the argument, because IMO the behavior of treating it as chars is kinda confusing (even though that's what the equivalent PHP functions do).

kt3k:

> because IMO the behavior of treating it as chars is kinda confusing

I agree with this, but how about treating `trimBy(foo, bar)` like `trimBy(foo, [bar])` when `bar` is a string? I think that would cover better the original suggestion of `stripPrefix`

kt3k (approved):
LGTM. Thank you for updating! (I added some examples in docs)

PR #6781, opened 2025-08-01
refactor(streams/unstable): toByteStream to make use of autoAllocateChunkSize

BlackAsLight:
I don't know when the `autoAllocateChunkSize` property became a thing on readable streams, but it does make the logic for supporting BYOB streams easier.

This pull request makes use of the property to simplify the code.

### DIFF
--- a/streams/unstable_to_byte_stream.ts
+++ b/streams/unstable_to_byte_stream.ts
@@ -38,6 +38,7 @@ export function toByteStream(
     const reader = readable.getReader();
     return new ReadableStream({
       type: "bytes",
+      autoAllocateChunkSize: 1024,
       async pull(controller) {
         const value = await async function () {
           while (true) {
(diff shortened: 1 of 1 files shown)


streams/unstable_to_byte_stream.ts
```diff
@@ -49,22 +50,23 @@ export function toByteStream(
...
-          }
-        } else controller.enqueue(value);
+        const buffer = new Uint8Array(
+          controller.byobRequest!.view!.buffer,
+          controller.byobRequest!.view!.byteOffset,
+          controller.byobRequest!.view!.byteLength,
+        );
+        const size = buffer.length;
+        if (value.length > size) {
+          buffer.set(value.subarray(0, size));
```
kt3k, line 63: Is it safe to change this to `subarray` (from `slice`)? (Is it possible that value.buffer can be overwritten in the upstream?)

BlackAsLight: As long as the underlying buffer of value is an ArrayBuffer and not a SharedArrayBuffer, we can guarantee it's safe. 

.set is a copy function so it's copying a subsection of value into buffer. Doing it as a slice is just an additional unnecessary copy.

BlackAsLight:
For clarity, the choice of 1024 bytes as a default buffer size was decided arbitrarily. I don't know if it is a good size for a default buffer, but it will only get used when the readable passed in doesn't support byob mode and the consumer doesn't enable byob mode.

PR #6782, opened 2025-08-02
feat(html/unstable): add escapeJs and escapeCss functions

lionel-rowe:
Resolves https://github.com/denoland/std/issues/6748

### DIFF
--- a/html/unstable_escape_js.ts
+++ b/html/unstable_escape_js.ts
@@ -1,0 +1,49 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/** Options for {@linkcode escapeJs} */
+export type EscapeJsOptions = {
+  /**
+   * The number of spaces or tab to use for indentation in the output.
+   * If not specified, no extra whitespace is added.
+   */
+  space?: number | "\t";
+};
+
+/**
+ * Escapes a JavaScript object or other data for safe interpolation inside a `<script>` tag.
+ *
+ * The data must be JSON-serializable (plain object, array, string, number (excluding `NaN`/infinities), boolean, or null).
+ *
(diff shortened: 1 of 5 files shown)


kt3k:
`escapeCss` looks good to me.

html/unstable_escape_js.ts
```diff
@@ -0,0 +1,49 @@
...
+ *   baz: "<!-- ",
+ *   quux: "\u2028\u2029<>",
+ * };
+ * assertEquals(
+ *  escapeJs(input),
+ *  String.raw`{"foo":"\u003c/script>","bar":"\u003cSCRIPT>","baz":"\u003c!-- ","quux":"\u2028\u2029<>"}`,
+ * );
+ * ```
+ */
+export function escapeJs(data: unknown, options: EscapeJsOptions = {}): string {
```
kt3k, line 59: I don't see why data accepts JSON-serializable object. Shouldn't this be `escapeJs(str: string): string`?

lionel-rowe: Makes it a lot more ergonomic to pass arbitrary(-ish) data:

```ts
// might be nested arbitrarily deeply
declare const data: string | number | Record<string, string> | Record<string, Record<string, string>>
const scriptContentToInject = `<script>window.handleData(${escapeJs(data)})</script>`
```

Or (because the output is also JSON-compatible):

```ts
const importMap = { imports: { zod: 'https://esm.sh/v131/zod@3.21.4' } }
const importMapScriptTag = `<script type="importmap">${escapeJs(importMap)}</script>`

kt3k: Ok. Makes sense. Can you add those as usage examples? Otherwise, I guess users might not get how to use it.

kt3k (approved):
LGTM 👍

PR #6783, opened 2025-08-02
refactor(tar): TarStream

BlackAsLight:
This pull request refactors the underlying code of `TarStream` to be easier to read, more memory efficient by skipping using strings as a medium, and offers a real BYOB stream instead of a superficial one, where buffers are taken from the start and passed up.

This pull request passes all existing tests and changes no API usage.

### DIFF
--- a/tar/tar_stream.ts
+++ b/tar/tar_stream.ts
@@ -1,5 +1,55 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
 
+import { toByteStream } from "@std/streams/unstable-to-byte-stream";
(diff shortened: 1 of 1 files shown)


BlackAsLight:
Improvements to `assertValidTarStreamOptions` has also been made.
- mode, uid, and gid weren't checking if negative numbers or NaN was provided.
- mtime was allowing negative timestamps to be provided, but the spec suggests avoiding them due to inconsistencies.
- octals are now being calculated in a way that skips using strings as a medium.

PR #6785, opened 2025-08-06
feat(log): add context management methods to Logger

xseman:
Context logic inspired by log4js-node

https://log4js-node.github.io/log4js-node/api.html#loggers---log4jsgetloggercategory


kt3k:
We discussed about @std/log in https://github.com/denoland/std/issues/6124 and decided to not maintain it anymore. https://github.com/denoland/std/pull/6819

Thanks for your suggestion anyway

PR #6786, opened 2025-08-07
feat(net/unstable): add matchSubnets on @std/net

Octo8080X:
target https://github.com/denoland/std/issues/6722

This PR adds the matchSubnets(addr: string, subnetOrIps: string[]): boolean  functions


timreichen:
Would be great if there were a near 100% test coverage for new apis.

Octo8080X:
@timreichen 
Thanks!
It was a bit difficult, but we achieved 100% coverage!

kt3k (approved):
Looks useful additions to me.

LGTM

net/unstable_ip.ts
```diff
@@ -75,8 +78,273 @@ export function isIPv6(addr: string): boolean {
...
+  }
+
+  for (const subnetOrIp of subnetOrIps) {
+    if (matchSubnet(addr, subnetOrIp)) {
+      return true;
+    }
+  }
+
+  return false;
+}
```
iuioiua, line 123: I don't understand the exact purpose of this function. It's supposed to return `true` when at least one subnet or IP address matches? What if some don't match? I don't the abstraction over `matchIPv4Subnet` and `matchIPv6Subnet`. IMHO, we should remove this function and let the user decide how to use `matchIPv4Subnet` or `matchIPv6Subnet`, whether it be in an `Array.prototype.every` or `Array.prototype.some`, etc.

iuioiua: I guess this can be used for blacklists, but I still think this should be done by the user.

PR #6788, opened 2025-08-09
fix(fs): `expandGlob`/`expandGlobSync` - match non-glob path segments containing escaped glob chars

dsherret:
Closes https://github.com/denoland/std/issues/6787

### DIFF
--- a/fs/expand_glob.ts
+++ b/fs/expand_glob.ts
@@ -534,7 +534,7 @@ export function* expandGlobSync(
 }
 
 const globEscapeChar = Deno.build.os === "windows" ? "`" : `\\`;
-const globMetachars = "*?{}[]";
+const globMetachars = "*?{}[]()|+@!";
 function unescapeGlobSegment(segment: string): string {
   let result = "";
   let lastIndex = 0;


kt3k (approved):
LGTM. Thanks!

dsherret:
@kt3k I updated this PR based on the AI feedback. It was too simplistic before I think

fs/expand_glob.ts
```diff
@@ -532,3 +532,29 @@ export function* expandGlobSync(
   }
   yield* currentMatches;
 }
+
+const globEscapeChar = Deno.build.os === "windows" ? "`" : `\\`;
+const globMetachars = "*?{}[]";
```
kt3k: shouldn't we also include `()|+@!`? (Our glob matcher supports extended glob like `@(foo|bar).txt`

dsherret: Good catch. I updated it. I wonder if we should specifically test this unescapeGlobSegment function.

kt3k (approved):
LGTM!

PR #6791, opened 2025-08-13
chore(assert,expect): bump assert and expect versions

kt3k:
This PR bumps the versions of `@std/expect` and `@std/assert` to ensure the fix #6747 is applied to the users of those packages.

### DIFF
--- a/assert/deno.json
+++ b/assert/deno.json
@@ -1,6 +1,6 @@
 {
   "name": "@std/assert",
-  "version": "1.0.13",
+  "version": "1.0.14",
   "exports": {
     ".": "./mod.ts",
     "./assert": "./assert.ts",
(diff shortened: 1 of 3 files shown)


kt3k:
For more context: The fix https://github.com/denoland/std/pull/6747 is already published as `@std/internal@1.0.10`, but because `@std/assert` and `@std/expect` is not released after that,  there's no obvious way to update `@std/internal` to the latest when the old version of `@std/internal` is stored in their lock file. This PR bumps the versions of these 2 packages, which provides the obvious way to upgrade `@std/internal` to the latest.

marvinhagemeister (approved):

bartlomieju (approved):
LGTM, is there a reason this is done manually instead of automated script?

kt3k:
@bartlomieju The automated script only detects the direct changes in that package. The fix #6747 was done only in `@std/internal` indirectly, and `@std/assert` and `@std/expect` were not updated automatically..

(Maybe we should consider updating the dependent packages as well in such situation..)

PR #6793, opened 2025-08-13
test(tar): improve TarStream's tests

BlackAsLight:
https://github.com/denoland/std/pull/6783 should be merged before this one.

This pull request rewrites the tests for TarStream to not only be more concise, but to also test for more behaviour than before and now offers 100% coverage.


kt3k (approved):
LGTM. Thanks!

Issue #6663, opened 2025-05-21
Pass the official `toml-test` suite


kt3k:
https://github.com/toml-lang/toml-test

https://toml-lang.github.io/toml-test-matrix/output/js-deno-toml.html

There are 30 failures in decoder (parser) tests, 89 failures in encoder (stringify) tests, and 109 failures in invalid input handling.

cosmastech:
Is there a way to get the toml-test suite re-run officially? Would be nice to be able to monitor how much has already been chewed through.

I realize I can run it locally, but I am _very lazy_ 😝

PR #6798, opened 2025-08-15
test(toml): test with the official toml-test suite

4513ECHO:
ref: #6663

Currently there is only tests for `parse()` because we have not had enough *precise* parser to compare the result of `stringify()` with the expected output yet.

Need to be discussed:
- Should we replace current `parse_test.ts`? If so, I think some tests (such as tests for error messages and `__proto__`) are need to be moved.


iuioiua:
Is there anything pending to publish (undraft) this PR?

kt3k:
> Should we replace current parse_test.ts? If so, I think some tests (such as tests for error messages and __proto__) are need to be moved.

I think it's fine to keep the existing tests as is as the entire TOML tests pass very quickly (like < 300ms on my local machine)

4513ECHO:
OK, Thanks!

PR #6800, opened 2025-08-15
feat(async/unstable): add `ensureLastCall` option for `throttle`

lionel-rowe:
Closes https://github.com/denoland/std/issues/6797

### DIFF
--- a/async/unstable_throttle.ts
+++ b/async/unstable_throttle.ts
@@ -19,7 +19,7 @@ export interface ThrottledFunction<T extends Array<unknown>> {
   (...args: T): void;
   /**
    * Clears the throttling state.
-   * {@linkcode ThrottledFunction.lastExecution} will be reset to `NaN` and
+   * {@linkcode ThrottledFunction.lastExecution} will be reset to `-Infinity` and
    * {@linkcode ThrottledFunction.throttling} will be reset to `false`.
    */
   clear(): void;
@@ -33,7 +33,7 @@ export interface ThrottledFunction<T extends Array<unknown>> {
   readonly throttling: boolean;
   /**
    * Returns the timestamp of the last execution of the throttled function.
-   * It is set to `NaN` if it has not been called yet.
+   * It is set to `-Infinity` if it has not been called yet, or reset is called after the last call.
    */
   readonly lastExecution: number;
 }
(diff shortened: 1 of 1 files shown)


async/unstable_throttle.ts
```diff
@@ -93,7 +112,6 @@ export function throttle<T extends Array<any>>(
   };
 
   throttled.flush = () => {
-    lastExecution = NaN;
```
lionel-rowe, line 96: This wasn't doing anything, as `flush()` doesn't read `lastExecution`, and `throttled.clear()` overwrites it anyway.

async/unstable_throttle.ts
```diff
@@ -93,7 +112,6 @@ export function throttle<T extends Array<any>>(
...
     flush?.();
     throttled.clear();
```
lionel-rowe: Should `throttled.clear()` be called here at all though? It's not obvious to me why calling `throttled.flush()` should set `lastExecution` to `NaN`, as opposed to setting it to the current time.

lionel-rowe: Also maybe instead of `NaN` it should be `-Infinity`? That seems like a better value to represent "never" (in the past) as an epoch timestamp, given that `Date.now() > NaN` is `false`

kt3k: I think it's better to not clear at the end of flush. Currently the next call after the flush can happen immediately, which seems against the goal of this util to me.

cc @lowlighter 

> Also maybe instead of NaN it should be -Infinity? That seems like a better value to represent "never" (in the past) as an epoch timestamp, given that Date.now() > NaN is false

`-Inifinity` seems ok to me.

async/unstable_throttle.ts
```diff
@@ -36,53 +46,62 @@ export interface ThrottledFunction<T extends Array<unknown>> {
...
         fn.call(throttled, ...args);
       } finally {
         lastExecution = Date.now();
         flush = null;
       }
     };
     if (throttled.throttling) {
+      if (ensureLast) {
+        timeout = setTimeout(() => flush?.(), timeframe);
+      }
```
kt3k: I think the timer should be cleared right before new one created.
```suggestion
      if (ensureLast) {
        clearTimeout(timeout);
        timeout = setTimeout(() => flush?.(), timeframe);
      }
```

If you add logs like this:
```js
      if (ensureLast) {
        timeout = setTimeout(() => {
          console.log("setTimeout", args);
          flush?.();
        }, timeframe);
      }
```
You can observe unnecessary timers are executed with the current implementation

PR #6801, opened 2025-08-18
fix(crypto): make `timingSafeEqual()` truly timing-safe

babiabeo:
The current `timingSafeEqual()` is not timing-safe and vulnerable to timing based attacks due to its early return on mismatched lengths:
```js
if (a.byteLength !== b.byteLength) return false;
```
This PR refactors the function to use constant-time comparison.

(Im not sure if there is a way to test this)

### DIFF
--- a/crypto/timing_safe_equal.ts
+++ b/crypto/timing_safe_equal.ts
@@ -48,14 +48,20 @@ export function timingSafeEqual(
   a: ArrayBufferView | ArrayBufferLike | DataView,
   b: ArrayBufferView | ArrayBufferLike | DataView,
 ): boolean {
-  if (a.byteLength !== b.byteLength) return false;
   const dataViewA = toDataView(a);
   const dataViewB = toDataView(b);
-  const length = a.byteLength;
+  const lengthA = a.byteLength;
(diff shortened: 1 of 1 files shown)


kt3k:
> The current timingSafeEqual() is not timing-safe and vulnerable

Does this mean the length of compared bytes can be detected by changing the input length?

I think the length of the compared bytes can be still detected as `lengthMax` can be controlled  by the attacker's input length

babiabeo:
> > The current timingSafeEqual() is not timing-safe and vulnerable
> 
> Does this mean the length of compared bytes can be detected by changing the input length?
> 
> I think the length of the compared bytes can be still detected as `lengthMax` can be controlled by the attacker's input length

The main goal isn't to hide the length but to make the comparison constant-time. For example, in a non-timing-safe comparison, when comparing `"abc"` (the input) with `"abcde"` (the secret one), it'll immediately return `false` due to different lengths. The attacker can try other inputs until they find the one taking longer to process.

babiabeo:
Ok after thinking it over, the early return on mismatched length is totally fine since the length is not secret and it doesn’t leak secret information.

I'll close this PR.

Issue #5727, opened 2024-08-19
feat: add promiseState to @std/async


PandaWorker:
Please add this utility to check the Promise status

```ts
export type PromiseState = 'fulfilled' | 'rejected' | 'pending';

/**
 * Return state (fulfilled/rejected/pending) of a promise
 *
 * ```ts
 * import { assertEquals } from "@std/assert";
 * import { promiseState } from "@std/async/promise-state";
 *
 * assertEquals(await promiseState(Promise.resolve("value")), "fulfilled");
 * assertEquals(await promiseState(Promise.reject("error")), "rejected");
 * assertEquals(await promiseState(new Promise(() => {})), "pending");
 * ```
 */
export function promiseState(p: Promise<unknown>): Promise<PromiseState> {
	return Promise.resolve().then(() =>
		Promise.race([p, null]).then(
			v => (v === null ? 'pending' : 'fulfilled'),
			() => 'rejected'
		)
	);
}

```

iuioiua:
Please provide a use case and reasons behind the suggestion.

PandaWorker:
The native implementation of Promise does not provide the property "state", this implementation helps to get the current state of the promise

kt3k:
Can you give us more practical example usages?

Other note: [`npm:promise-status-async`](https://www.npmjs.com/package/promise-status-async) seems providing a very similar API, but it has very few weekly downloads and github activities. I doubt there's much practical usages of it

Issue #3759, opened 2023-11-01
`assertPromiseResolved()`


jespertheend:
**Is your feature request related to a problem? Please describe.**

During tests I often want to verify that a promise has been resolved right at this moment.

**Describe the solution you'd like**

This is a function that I've been using a lot for this:
```js
/**
 * Asserts whether a promise is currently resolved or not. The check is made asynchronously.
 * The call waits for the next event loops and gives the promise a chance to resolve in the current event loop.
 * The reason for this is that there is no way to synchronously check the resolved state of promises in JavaScript.
 * @param {Promise<any>} promise
 * @param {boolean} expected
 */
export async function assertPromiseResolved(promise, expected) {
	let resolved = false;
	(async () => {
		await promise;
		resolved = true;
	})();
	await waitForMicrotasks();
	const msg = expected ? "Expected the promise to be resolved" : "Expected the promise to not be resolved";
	assert(resolved == expected, msg);
}
```

`waitForMicrotasks` is a very basic:
```js
export function waitForMicrotasks() {
	return new Promise(r => setTimeout(r, 0));
}
```
As far as I'm aware this is required, there's no way to figure out the state of promises synchronously.

Is this something that would be suitable for deno_std? I've got a few tests for this as well.

**Describe alternatives you've considered**

Keep using this from my own repository ¯\\\_(ツ)_/¯

lino-levan:
@iuioiua any thoughts?

iuioiua:
What's wrong with asserting the value of the resolved promise? Am I missing something here?

kt3k:
`await waitForMicrotasks()` part in the above example looks inaccurate to me (it looks waiting more than microtasks). The suggested util (`promiseState`) in [this stackoverflow answer](https://stackoverflow.com/a/35820220) looks more general (supporting rejected state) and simpler.

jespertheend:
The `await waitForMicrotasks()` is there because half of the time you need to do that anyway, and when you don't, the extra `setTimeout()` isn't really an issue. I haven't really run across a case where I needed to know whether a promise was resolved right at this moment.

I figured without this, you'd just be confused as to why your promise isn't resolved. Only to realise after debugging for a good few minutes that you need to wait for microtasks.

Some stats: 28 of my tests use `assertPromiseResolve()` at least once, 12 of which fail without the `waitForMicrotasks()` call.
I have also tried
```js
export function waitForCurrentMicroTasks() {
	return new Promise(r => queueMicrotask(r));
[...]

PR #6806, opened 2025-08-28
feat(expect): support async extended matchers

lowlighter:
This allows extended matchers in `expect` to return either `ExtendMatchResult` or `Promise<ExtendMatchResult>` 

This changes provides additional possibilities for custom matchers, like implementing a `expect(value).toBeResolved()` matcher.

Example:
```ts
const { promise, resolve } = Promise.withResolvers<void>();
await expect(promise).not.toBeResolved();
resolve();
await expect(promise).toBeResolved();

// NB: this is different from expect(promise).resolves.toBe()
// Here we test the actual promise state, not its resolved value
```

This is just an example, but it can be used for any custom matcher that requires async operations on the right side
(for example one could image matching agains dynamically fetched files, database query results, etc.)

**Side note:** While this PR doesn't address specifically the following issues, the example (in the diff of the test file) shows how to implement a matcher to test promise states
- #3759
- #5727

*Needs rebase after #6807 for lint issues*

### DIFF
--- a/expect/_extend_test.ts
+++ b/expect/_extend_test.ts
@@ -2,11 +2,12 @@
 
 import { expect } from "./expect.ts";
 import type { Async, Expected, MatcherContext, Tester } from "./_types.ts";
-import { AssertionError, assertThrows } from "@std/assert";
+import { AssertionError, assertRejects, assertThrows } from "@std/assert";
 
 declare module "./_types.ts" {
   interface Expected {
     toEqualBook: (expected: unknown) => ExtendMatchResult;
+    toBeResolved: () => Promise<ExtendMatchResult>;
   }
 }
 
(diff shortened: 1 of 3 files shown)


lowlighter:
Needs rebase after #6807 for lint issues

yannduperis:
I am interested in what this contribution would unlock (thank you). I cannot figure out the status of this merge request, has this been abandoned ?

fibibot (approved):
Async matcher path returns the promise chain at `expect.ts:218` so callers can `await` the assertion, and `emitAssertionTrigger` fires only on pass — same shape as the sync branch immediately below. CI is green; approving.

- nit: the new test at `_extend_test.ts:181` has the same name as the existing sync test at line 102 (`"expect.extend() api test case"`). Renaming it (e.g. `"expect.extend() supports async matchers"`) makes failures unambiguous.
- nit: the `expect.extend` JSDoc at `expect.ts:316–371` only shows sync matchers. A one-line note that matchers may return `Promise<ExtendMatchResult>`, or making one of the example matchers `async`, would surface this new capability for readers who don't read `_types.ts`.
- nit: the `if (context.isNot) { … } else if (!result.pass) { … }` block inside `result.then` at `expect.ts:220–226` duplicates the sync block at lines 230–236 verbatim. A small `check(result)` helper would remove the duplication.

lunadogbot (approved):
Async path at `expect.ts:218–228` returns `result.then(...)` so `applyMatcher` resolves to `Promise<void>`; both the `(value as Promise).then(applyMatcher)` chain at line 246 and direct calls await correctly. `_types.ts` widens `Matcher` to `Promise<ExtendMatchResult>` only — `MatchResult` gets no async variant, so the built-in matcher branch at line 238 can't accidentally land on the async path. A rejected matcher promise propagates unchanged through `result.then`, matching how sync matchers propagate throws. CI green across the matrix.

lowlighter:
It's not abandoned, I'm still waiting on it tbh 😅

PR #6810, opened 2025-09-01
feat(async): support `allowInfinity` option in deadline

lowlighter:
This add a new `allowInfinity?: boolean;` option to deadline to allow `Infinity` value.

```ts
import { parseArgs } from "@std/cli/parse-args";
const { timeout = Infinity } = parseArgs(Deno.args);

await deadline((async () => {
  // Long running task
})(), timeout, { allowInfinity: true })
```

**Why is this needed ?**
- `deadline`/`abortable` doesn't have the same signature, so you can just interchange them inline
  - Or else you need to be verbose and put out your promise, checking the timeout value and decide to call or not `deadline`
- `AbortSignal.timeout()` only support finite values, so using `Infinity` would throw an error *(current behaviour)*
- An `Infinity` deadline makes senses concept-wise, it's just an edge case
- This simplify a lot apps where end-users can customize the timeout themselves (like in the example above)

### DIFF
--- a/async/unstable_deadline.ts
+++ b/async/unstable_deadline.ts
@@ -1,0 +1,58 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+// This module is browser compatible.
+
+import { abortable } from "./abortable.ts";
+
+/** Options for {@linkcode deadline}. */
+export interface DeadlineOptions {
+  /** Signal used to abort the deadline. */
+  signal?: AbortSignal;
+  /** 
+   * Allow the deadline to be infinite. 
+   * 
+   * @experimental **UNSTABLE**: New API, yet to be vetted.
+   */
+  allowInfinity?: boolean;
+}
+
+/**
+ * Create a promise which will be rejected with {@linkcode DOMException} when
+ * a given delay is exceeded.
+ *
+ * Note: Prefer to use {@linkcode AbortSignal.timeout} instead for the APIs
+ * that accept {@linkcode AbortSignal}.
+ * 
+ * @throws {DOMException & { name: "TimeoutError" }} If the provided duration
+ * runs out before resolving.
+ * @throws {DOMException & { name: "AbortError" }} If the optional signal is
+ * aborted with the default `reason` before resolving or timing out.
+ * @throws {AbortSignal["reason"]} If the optional signal is aborted with a
+ * custom `reason` before resolving or timing out.
+ * @typeParam T The type of the provided and returned promise.
+ * @param p The promise to make rejectable.
+ * @param ms Duration in milliseconds for when the promise should time out.
+ * @param options Additional options.
+ * @returns A promise that will reject if the provided duration runs out before resolving.
+ *
+ * @example Usage
+ * ```ts ignore
+ * import { deadline } from "@std/async/deadline";
+ * import { delay } from "@std/async/delay";
+ *
+ * const delayedPromise = delay(1_000);
+ * // Below throws `DOMException` after 10 ms
+ * const result = await deadline(delayedPromise, 10);
+ * ```
+ */
+export async function deadline<T>(
+  p: Promise<T>,
+  ms: number,
+  options: DeadlineOptions = {},
+): Promise<T> {
+  const signals: AbortSignal[] = [];
+  if ((!options?.allowInfinity) || (ms !== Infinity)) {
+    signals.push(AbortSignal.timeout(ms));
+  }
+  if (options.signal) signals.push(options.signal);
+  return await abortable(p, AbortSignal.any(signals));
+}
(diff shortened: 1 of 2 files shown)


lionel-rowe:
If this adds support for `Infinity`, why have an additional option for `allowInfinity`? Why not just always allow it?

Also, should `Infinity` be a special case, or should this just add support for arbitrarily large values? See https://github.com/denoland/std/pull/6775 for an example of how that could work (could move `setArbitraryLengthTimeout` into a common internal utils file, and/or create a corresponding `arbitraryLengthAbortSignalTimeout` class).

lowlighter:
I added it as an option because it could be considered a breaking change, and some people might expect the deadline to eventually be resolved in a finite amount of time. But personally I think it would make sense to just support it without the option

As for the other question, I guess this could be iterated in another PR. But since the spec of AbortSignal.timeout only accepts until Number.MAX_SAFE_INTEGER, it might be tedious to implement maybe ?

lionel-rowe:
> I added it as an option because it could be considered a breaking change, and some people might expect the deadline to eventually be resolved in a finite amount of time.

Changing throwing to non-throwing behavior typically isn't considered breaking (edit: I guess (?) that also means this wouldn't need to be marked unstable, as the interface wouldn't need to change at all).

> As for the other question, I guess this could be iterated in another PR.

It's true the actual use case for using values in `0x1fffffffffffff < n < Infinity` is probably pretty limited, but disallowing numbers within that range while _specifically_ allowing `Infinity` seems like a surprising limitation to me.

<del>

> But since the spec of AbortSignal.timeout only accepts until Number.MAX_SAFE_INTEGER, it might be tedious to implement maybe?

Using similar logic to `setArbitraryLengthTimeout`:~~

```ts
const MAX_ABORT_TIMEOUT = Number.MAX_SAFE_INTEGER

function arbitraryLengthAbortSignalTimeout(delay: number): AbortSignal {
  const ac = new AbortController();
  // ensure non-negative integer (but > MAX_ABORT_TIMEOUT is OK, even if Infinity)
  let currentDelay = delay = Math.trunc(Math.max(delay, 0) || 0);
  const start = Date.now();

  const queueTimeout = () => {
    currentDelay = delay - (Date.now() - start);

    if (currentDelay > MAX_ABORT_TIMEOUT) {
      AbortSignal.timeout(MAX_ABORT_TIMEOUT).onabort = queueTimeout;
    } else {
      AbortSignal.timeout(Math.max(0, currentDelay)).onabort = function() {
        ac.abort(this.reason);
      };
    }
  };

  queueTimeout();
  return ac.signal;
}
```

</del>

lowlighter:
> It's true the actual use case for using values in `0x1fffffffffffff < n < Infinity` is probably pretty limited, but disallowing numbers within that range while _specifically_ allowing `Infinity` seems like a surprising limitation to me.

I mean it depends. 

Most CLI apps that supports a timeout option usually accepts specifying passing either `0`/`-1` as special values to disable timeout completly (e.g. wget). But using these values for `deadline()` might seems counter-intuitive. Maybe this PR should instead be an option like `zeroNoDeadline` (not sure about the naming)

`Infinity` makes sense semantically but as you said it might get confusing for in-between range.

But then maybe it's actually useless to do, I mean, I currrently use `Math.min(Number.MAX_SAFE_INTEGER, timeout)` as a workaround and it's already like 285 616 years... We're probably all be dead by that time 😅

lionel-rowe:
> But then maybe it's actually useless to do, I mean, I currently use Math.min(Number.MAX_SAFE_INTEGER, timeout) as a workaround and it's already like 285 616 years... We're probably all be dead by that time 😅

Oh yeah, I guess implementing this would be thoroughly pointless haha. The equivalent max value for `setTimeout` is only 0xffffffff i.e. ~50 days, which is still conceivably within the lifespan of a long-running program.

IMO either one of `Math.min(Number.MAX_SAFE_INTEGER, ms)` or just returning `new AbortController().signal` (and never aborting it) above `Number.MAX_SAFE_INTEGER` would be equally good. Either case avoids special-casing `Infinity`, and either case would never (practically speaking) lead to wrong behavior.

kt3k:
@lowlighter sorry for the delay in review! I like @lionel-rowe's suggestion. Never aborting when the timeout > `Number.MAX_SAFE_INTEGER` makes sense to me.

I also think this can be landed as `fix` without introducing new option (I think we can just change (fix) the behavior for large `ms` value). Can you just change the existing `deadline`?

lowlighter:
Sorry for the long delay, I've made the requested changes
I'm not sure how to make a revelant test though haha 😅

async/deadline.ts
```diff
@@ -24,7 +24,8 @@ export interface DeadlineOptions {
  * custom `reason` before resolving or timing out.
  * @typeParam T The type of the provided and returned promise.
  * @param p The promise to make rejectable.
- * @param ms Duration in milliseconds for when the promise should time out.
+ * @param ms Duration in milliseconds for when the promise should time out. If
+ * greater than `Number.MAX_SAFE_INTEGER`, the deadline will never expire.
```
kt3k, line 28: 👍

PR #6812, opened 2025-09-04
chore(lint): ensure versioned specifier

dsherret:
Fixes a linting error in canary.


dsherret:
This is already fixed in main.

PR #6821, opened 2025-09-12
fix(assert): add support for checking ArrayBuffer equality

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6814

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -42,8 +42,7 @@ function getKeysDeep(obj: object) {
 }
 
 // deno-lint-ignore no-explicit-any
-const Temporal: any = (globalThis as any).Temporal ??
-  new Proxy({}, { get: () => {} });
+const Temporal = (globalThis as any).Temporal ?? Object.create(null);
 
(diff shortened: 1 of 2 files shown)


assert/equal.ts
```diff
@@ -42,8 +42,7 @@ function getKeysDeep(obj: object) {
 }
 
 // deno-lint-ignore no-explicit-any
-const Temporal: any = (globalThis as any).Temporal ??
-  new Proxy({}, { get: () => {} });
+const Temporal = (globalThis as any).Temporal ?? Object.create(null);
```
lionel-rowe, line 45: Honestly can't remember why I made this a proxy originally, it's just unnecessary complexity. Hope this tiny refactor is OK to include here and doesn't need its own PR

kt3k (approved):
Confirmed this fixes the issue, and also the diff expression in the error message seems reasonable:

```


    [Diff] Actual / Expected


    ArrayBuffer {
-     [Uint8Contents]: <01 02 03>,
+     [Uint8Contents]: <02 02 04>,
      byteLength: 3,
    }

        ^
    at assertEquals
```

LGTM

PR #6822, opened 2025-09-12
fix(async): pass numeric value of `delay` timer to `unrefTimer` when `persistent=false`

lionel-rowe:
Bug was introduced in https://github.com/denoland/std/pull/6775, as `setArbitraryLengthTimeout` returns `{ valueOf(): number }` rather than a primitive `number`, and `Deno.unrefTimer` throws if its argument isn't a `number`.

Previous type checking missed it due to already using a `// @ts-ignore` for the offending line, and the test also missed it due to stubbing `Deno.unrefTimer` with a no-op.

### DIFF
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -65,6 +65,7 @@ export function delay(ms: number, options: DelayOptions = {}): Promise<void> {
         Deno.unrefTimer(+i);
       } catch (error) {
         if (!(error instanceof ReferenceError)) {
+          clearTimeout(+i);
           throw error;
         }
         // deno-lint-ignore no-console
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -12,6 +12,9 @@ export interface DelayOptions {
   persistent?: boolean;
 }
 
+// Make type available in browser environments; we catch the `ReferenceError` below
+declare const Deno: { unrefTimer(id: number): void };
+
 /**
  * Resolve a {@linkcode Promise} after a given amount of milliseconds.
  *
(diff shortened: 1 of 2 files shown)


lionel-rowe:
I think (?) https://github.com/denoland/std/pull/6822/commits/368f865572a90705c30cf8d78d66bd083ac05e18 should reliably fix the flaky test [from this test (v1.x, macOS-latest) CI run](https://github.com/denoland/std/actions/runs/17663317641/job/50200598299):

> ```
> delay() runs to completion with `ms` argument > 0x7fffffff => ./async/delay_test.ts:185:6
> error: Leaks detected:
>   - A timer was started before the test, but completed during the test. Intervals and timers should not complete in a test if they were not started in that test. This is often caused by not calling `clearTimeout`. The operation was started here:
>     at Object.queueUserTimer (ext:core/01_core.js:792:9)
>     at setTimeout (ext:deno_web/02_timers.js:74:15)
>     at queueTimeout (file:///Users/runner/work/std/std/async/delay.ts:90:9)
>     at setArbitraryLengthTimeout (file:///Users/runner/work/std/std/async/delay.ts:93:3)
>     at file:///Users/runner/work/std/std/async/delay.ts:58:15
>     at new Promise (<anonymous>)
>     at delay (file:///Users/runner/work/std/std/async/delay.ts:49:10)
>     at fn (file:///Users/runner/work/std/std/async/delay_test.ts:131:13)
>     at innerWrapped (ext:cli/40_test.js:191:11)
>     at exitSanitizer (ext:cli/40_test.js:107:33)
> ```

PR #6823, opened 2025-09-12
feat(math/unstable): add `math` package with basic math utilities

lionel-rowe:
Closes https://github.com/denoland/std/issues/6803

* `clamp`: Clamp a number within a range. Based on [`Math.clamp` proposal](https://github.com/tc39/proposal-math-clamp), with a notable difference being that current proposal specifies `TypeError`s for `NaN` inputs, whereas my version returns `NaN`s, because throwing `TypeError`s seems rather unexpected given the corresponding behavior of `Math.min` and `Math.max`.
* `modulo`: Floor modulo operation, in contrast with JS's `%` remainder operator. Based on [Python's `%` (floor modulo) operator](https://docs.python.org/3/reference/expressions.html#:~:text=The%20%25%20(modulo)%20operator) and [modulus math proposal](https://github.com/dead-claudia/proposal-positive-modulo).
* `round-to`: Round a number to a specified number of decimal places using various rounding strategies. Basically `Number#toFixed` but returning a number instead of a string.

Probably worth kicking the wheels for things like floating-point rounding issues, in cases where those are avoidable or optimizable.


math/modulo.ts
```diff
@@ -0,0 +1,32 @@
...
+export function modulo(num: number, modulus: number): number {
+  if (!Number.isFinite(num) || Number.isNaN(modulus) || modulus === 0) {
+    return NaN;
+  }
+  if (num === 0) return modulus < 0 ? -0 : 0;
+  if (modulus === Infinity) return num < 0 ? Infinity : num;
+  if (modulus === -Infinity) return num > 0 ? -Infinity : num;
+
+  return (num % modulus + modulus) % modulus;
+}
```
0f-0b, line 31: Here's an alternative implementation for `modulo`.

```ts
export function modulo(num: number, modulus: number): number {
  num %= modulus;
  if (num === 0) {
    num = modulus < 0 ? -0 : 0;
  } else if ((num < 0) !== (modulus < 0)) {
    num += modulus;
  }
  return num;
}
```

In my benchmarks (code below), it was faster than the current implementation in the common case where both inputs are finite and nonzero, although slower when either input is infinite, NaN or 0. All tests still pass.

```ts
// `baseline` and `modulo` are different implementations.
// `cases` is taken from the python parity test.
const sink = new Set<number>();
for (const [a, b] of cases) {
  const group = Deno.inspect([a, b]);

  Deno.bench("baseline", { group }, () => {
    sink.add(baseline(a, b));
  });

  Deno.bench("modulo", { group }, () => {
    sink.add(modulo(a, b));
  });
}
```

math/clamp.ts
```diff
@@ -0,0 +1,27 @@
...
+ * import { clamp } from "@std/math/clamp";
+ * import { assertEquals } from "@std/assert";
+ * assertEquals(clamp(5, [1, 10]), 5);
+ * assertEquals(clamp(-5, [1, 10]), 1);
+ * assertEquals(clamp(15, [1, 10]), 10);
+ * ```
+ */
+export function clamp(num: number, limits: [min: number, max: number]): number {
+  const [min, max] = limits;
+  if (min > max) {
```
marvinhagemeister: Nit: Passing `limits` as an array means that this array needs to be allocated prior to passing it to this function. Given that math functions are often used in high performance scenarios it might be more desirable to avoid the need to allocate at all for determining whether a value is in-between two others.

kt3k (approved):
I often find these utils useful in my personal projects (Especially modulo is necessary for geometry-related programming). I'm in favor of the idea. LGTM

Does anyone have concern/objection about having this (unstable) package?

cc @ry @piscisaureus @bartlomieju

PR #6825, opened 2025-09-13
fix(semver): allow users to set prerelease number during increment

justinmchase:
If a user specifies a prerelease value in the form of `{identifer}.{number}` then both the identifier and the number will be used to set the prerelease metadata.

---

- Fixes https://github.com/denoland/std/issues/6824


kt3k (approved):
LGTM

I also made some changes:
- Added support of longer preleases like `foo.bar`, `foo.bar.1`. These are used as is if they include number, otherwise `.0` is appended
- Reverted the changes in parsing of prerelease and build metadata. I don't think we should trim whitespaces as those characters are invalid in `prerelease` or `build` strings. If you feel strongly about those changes please open new issue or PR.

Issue #6649, opened 2025-05-13
suggestion: convert the doc checker into a custom lint plugin


iuioiua:
The doc checker (https://github.com/denoland/std/blob/main/_tools/check_docs.ts) is currently a script. But this can be converted to a lint plugin thanks to comments being supported in canary in https://github.com/denoland/deno/pull/29189. For someone to implement this, they'd need to create a new lint plugin and use [`npm:comment-parser`](https://www.npmjs.com/package/comment-parser) to parse then process comments.

iuioiua:
Once this is done, I think #6826 can be done.

PR #6832, opened 2025-09-20
chore: docs lint plugin

iuioiua:
First step towards #6649.

### DIFF
--- a/_tools/docs_lint_plugin.ts
+++ b/_tools/docs_lint_plugin.ts
@@ -1,4 +1,5 @@
 // Copyright 2018-2025 the Deno authors. MIT license.
+// @ts-nocheck Deno.lint namespace does not pass type checking in Deno 1.x
 
 function isInternalFile(filename: string): boolean {
   return filename.split(/[\\/]+/).some((part) => part.startsWith("_"));


bytes/concat.ts
```diff
@@ -2,6 +2,7 @@
 // This module is browser compatible.
 
 import type { Uint8Array_ } from "./_types.ts";
+// deno-lint-ignore deno-std-docs/exported-symbol-documented
```
iuioiua, line 5: Alternatively, we could have this rule ignore re-exported types. However, that would assume that re-exported types have documentation. In this codebase that might be true but would be a far reaching assumption in other codebases.

_tools/docs_lint_plugin.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright 2018-2025 the Deno authors. MIT license.
+
+function isInternalFile(filename: string): boolean {
+  return filename.split("/").some((part) => part.startsWith("_"));
```
iuioiua: Will likely need to be refined in the future.

bytes/concat.ts
kt3k: Re-exported items like this one are actually documented in `_` prefixed files. Can we detect such docs by following the imports/exports?

iuioiua: Yes, by using [`@deno/doc`](https://jsr.io/@deno/doc). Though, I haven't tried it yet, and I assume it might come with a bit of a performance cost. IMHO, how it is now is fine enough to get the ball rolling on this lint plugin, and I think we could revisit this later. WDYT?

kt3k: Maybe it's fine to land if this is only informative, not requirement in CI.

> we could have this rule ignore re-exported types

Could you implement this for now? I think this directive would be confusing to contributors

iuioiua:
As discussed offline with @kt3k, I'll close this and instead create a 3rd party package, as adding a doc lint plugin would require a decent amount of work and the Deno team have higher priorities.

PR #6833, opened 2025-09-20
fix(async): convert setTimeout return values to Number (fixing type error when used with Node.js)

iugo:
In Node.js, setTimeout returns a NodeJS.Timeout. When using tsc in a frontend project, this can cause type errors.

code: https://github.com/iugo/deno-std-in-node/tree/async-timeout-type-error

### DIFF
--- a/async/delay.ts
+++ b/async/delay.ts
@@ -89,8 +89,8 @@ function setArbitraryLengthTimeout(
   const queueTimeout = () => {
     currentDelay = delay - (Date.now() - start);
     timeoutId = currentDelay > I32_MAX
-      ? setTimeout(queueTimeout, I32_MAX)
-      : setTimeout(callback, currentDelay);
+      ? Number(setTimeout(queueTimeout, I32_MAX))
+      : Number(setTimeout(callback, currentDelay));
   };
 
   queueTimeout();


kt3k (approved):
Confirmed this fixes type error in your repro setup. Thanks for catching this! LGTM

kt3k:
@iugo BTW I'm a bit confused that `tsc` reports type error from `node_modules/@std/async/delay.ts` in the above repo even with `"skipLibCheck": true` set.. (I thought the option disables the type checking of the code under `node_modules`, but that doesn't seem so in this case..

iugo:
> "skipLibCheck": true

@kt3k Maybe `skipLibCheck: true` only skips `.d.ts` files.

kt3k:
Memo: I looked into your repo's settings a bit more. I found that type checking of `node_modules/@std/async/delay.ts` happens because of `"moduleResolution": "node"` option in tsconfig.json. If I apply the changes below and use `node16` resolution mode, then tsc started using `.d.ts` file instead of type checking `node_modules/@std/async/delay.ts`.

```diff
diff --git a/package.json b/package.json
index cc57d04..513e5a4 100644
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "1.0.0",
   "description": "",
   "main": "index.js",
+  "type": "module",
   "scripts": {
     "init": "tsc --init",
     "test": "tsc --noEmit"
diff --git a/tsconfig.json b/tsconfig.json
index 695c668..ebd079d 100644
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,12 +6,12 @@
     // "outDir": "./dist",
 
     // Custom configurations
-    "moduleResolution": "node",
+    "moduleResolution": "node16",
     "noEmit": true,
 
     // Environment Settings
     // See also https://aka.ms/tsconfig/module
-    "module": "esnext",
+    "module": "node16",
     "target": "es2022",
     // For nodejs:
     "lib": ["dom", "esnext"],
```

KnorpelSenf:
Given that Deno will switch away from using `number` as a return type of `setTimeout`, wouldn't it make more sense to align on `Timer` objects instead? Or else, maybe adjust the return type to use `ReturnType<typeof setTimeout>` in order to support different environments?

PR #6835, opened 2025-09-20
feat(assert/unstable,internal): truncate big diffs

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6631

Default behavior remains unchanged. By setting the opt-in `DIFF_CONTEXT_LENGTH` env var (name subject to bikeshedding), you can enable output like this:

[asciicast](https://asciinema.org/a/MSsmFnYslPZxe0QXgFbkwDjf3)

Maybe this needs marked `UNSTABLE` (~~unstable version of `internal/build-message` with unstable versions of `assert/equals` and `assert/strict-equals` that are identical apart from doc + dependency change~~ edit: probably better to just parameterize stable version and export a customized version from unstable)


BlackAsLight:
This looks pretty nice.

On a somewhat related topic I thought I made an issue suggesting directing the output to a file a while ago, but I can't seem to find it. The point being to allow one to parse the difference in an easier manner than in the terminal.

lionel-rowe:
> I thought I made an issue suggesting directing the output to a file

@BlackAsLight Is that necessary as part of the API though? Why not just ~~pipe~~ redirect stderr to a file?

```sh
$ NO_COLOR=1 deno eval "import { assertEquals } from 'jsr:@std/assert'; assertEquals(1, 2)" 2> errors.txt
$ cat errors.txt
error: Uncaught (in promise) AssertionError: Values are not equal.


    [Diff] Actual / Expected


-   1
+   2

        ^
    at assertEquals (https://jsr.io/@std/assert/1.0.14/equals.ts:64:9)
    at file:///root/source/repos/$deno$eval.mts:1:49
```

BlackAsLight:
> > I thought I made an issue suggesting directing the output to a file
> 
> 
> 
> @BlackAsLight Is that necessary as part of the API though? Why not just pipe stderr to a file?
> 
> 
> 
> ```sh
> 
> $ NO_COLOR=1 deno eval "import { assertEquals } from 'jsr:@std/assert'; assertEquals(1, 2)" 2> errors.txt
> 
> $ cat errors.txt
> 
> error: Uncaught (in promise) AssertionError: Values are not equal.
> 
> 
> 
> 
> 
>     [Diff] Actual / Expected
> 
> 
> 
> 
> 
> -   1
> 
> +   2
> 
> 
> 
>   throw new AssertionError(message);
> 
>         ^
> 
>     at assertEquals (https://jsr.io/@std/assert/1.0.14/equals.ts:64:9)
> 
>     at file:///root/source/repos/$deno$eval.mts:1:49
> 
> ```

I didn't think of that, but that also seems pretty verbose and having it send its contents to a specific file only when it errors seems pretty convenient to me.

kt3k (approved):

Looks reasonable improvement of diffs. LGTM

PR #6840, opened 2025-10-05
feat(msgpack): add encodeStream and decodeStream for streaming

lambdalisue:
I'd like to decode MessagePack from a `ReadableStream<Uint8Array>`, but the current `decode` only supports complete inputs, so I added `decodeStream`. For consistency, I also added `encodeStream`, but it's just a thin wrapper around repeated encode calls, so we can drop it if consistency isn't required.


lambdalisue:
Hum... It seems using `encodeSlice` throws "Cannot safely encode value into messagepack" ~~in some case~~ when try to encode `Error` object. I think it's spec, right?

lambdalisue:
It seems that using @std/msgpack is about twice as slow as using [msgpackr](https://github.com/kriszyp/msgpackr), which already supports streams, so I've decided to stick with msgpackr. Sorry for the noise in this PR.

```
| benchmark         | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| ----------------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |

group bench
| messagepack_rpc   |          8.5 ms |         118.0 | (  6.9 ms …  18.3 ms) |   8.7 ms |  18.3 ms |  18.3 ms |
```

```
| benchmark         | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| ----------------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |

group bench
| messagepack_rpc   |         15.2 ms |          65.6 | ( 12.4 ms …  25.5 ms) |  15.5 ms |  25.5 ms |  25.5 ms |
```

PR #6841, opened 2025-10-05
YAML: Allow to add custom types for parse and stringify

### LINKED ISSUE #5680: YAML: Allow to store dates without time

I want to create a front matter with a code like this:

```yml
---
title: Post title
date: 2024-09-01
---
```

But YAML export the `date` value in this way:

```yml
---
title: Post title
date: 2024-09-01T19:00:00Z
---
```

The only way I found to save only the date (without time) was to convert the value to a string:

```yml
---
title: Post title
date: "2024-09-01"
---
```

But then this value is not parsed as Date.

A way to store dates without including the time. As an idea, maybe detecting if the object is a `Temporal.PlainDate` instance.

oscarotero:
This change adds a new option to add custom types for `parse` and `stringify` functions of YAML, which can fix #5680.

For example: 

```js
import { parse, type ImplicitType } from "@std/yaml";

const smile: ImplicitType = {
  tag: "tag:custom:smile",
  resolve: (data): boolean => data === "=)",
  construct: (): string => "🙂",
  predicate: (data): data is string => data === "🙂",
  kind: "scalar",
  represent: (): string => "=)",
};

const yaml = `
title: =)
tags:
  - =)
  - bar
`

const result = parse(yaml, { extraTypes: [ smile ] }
console.assert(result.title === "🙂");
```

### DIFF
--- a/yaml/_type.ts
+++ b/yaml/_type.ts
@@ -21,13 +21,20 @@ export type RepresentFn<D> = (data: D, style?: StyleVariant) => string;
 
 // deno-lint-ignore no-explicit-any
 export interface Type<K extends KindType, D = any> {
+  /** Tag to identify the type */
   tag: string;
+  /** Kind of type */
   kind: K;
+  /** Cast the type. Used to stringify */
   predicate?: (data: unknown) => data is D;
+  /** Function to represent data. Used to stringify */
   represent?: RepresentFn<D> | Record<string, RepresentFn<D>>;
+  /** Default style for the type. Used to stringify */
   defaultStyle?: StyleVariant;
+  /** Function to test whether data can be resolved by this type. Used to parse */
   // deno-lint-ignore no-explicit-any
   resolve: (data: any) => boolean;
+  /** Function to construct data from string. Used to parse */
   // deno-lint-ignore no-explicit-any
   construct: (data: any) => D;
 }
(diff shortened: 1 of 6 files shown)


kt3k:
We have policy of accepting new feature as unstable first https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100

Can you copy `parse.ts` to `unstable_parse.ts` and make these changes to there? (Also map `./unstable-parse` to `./unstable_parse.ts`, that enables users to import it via `jsr:@std/yaml/unstable-parse`

oscarotero:
@kt3k I moved everything to the unstable files (the `unstable_stringify.ts` already existed so I just had to create `unstable_parse.ts`) and all lint issues caused by private types are fixed. The only thing Ieft is the PR title, but I have no idea how to fix it.

yaml/_type.ts
```diff
@@ -17,17 +20,30 @@ export type StyleVariant =
   | "octal"
   | "hexadecimal";
 
+/**
+ * Function to convert data to a string for YAML serialization.
+ */
 export type RepresentFn<D> = (data: D, style?: StyleVariant) => string;
 
+/**
+ * A type definition for a YAML node.
```
kt3k, line 29: Thank you for documenting this!

kt3k (approved):
LGTM

devingfx:
Why create a whole copy of each modules just to re-export stuff? instead of just make the module `_schema.ts` public...

Let people access this schema module and they can even hack the original schemas... (Just needs a "At your own risk" notice).

I feel this PR a bit overkill for initial goal ...

PR #6842, opened 2025-10-06
feat(http): add support for extra headers in serveFile

moumar:
The headers customization is present in http serveDir but not in serveFile. This PR fixes this.

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -155,6 +155,12 @@ export interface ServeFileOptions {
    * `filePath`.
    */
   fileInfo?: Deno.FileInfo;
+
+  /** Headers to add to each response
+   *
+   * @default {[]}
+   */
+  headers?: string[];
 }
 
 /**
@@ -202,6 +208,15 @@ export async function serveFile(
   }
 
   const headers = createBaseHeaders();
+
+  if (options?.headers) {
+    for (const header of options.headers) {
+      const headerSplit = header.split(":");
+      const name = headerSplit[0]!;
+      const value = headerSplit.slice(1).join(":");
+      headers.append(name, value);
+    }
+  }
 
   const etag = fileInfo.mtime
     ? await eTag(fileInfo, { algorithm })


http/file_server.ts
```diff
@@ -155,6 +155,12 @@ export interface ServeFileOptions {
    * `filePath`.
    */
   fileInfo?: Deno.FileInfo;
+
+  /** Headers to add to each response
+   *
+   * @default {[]}
+   */
+  headers?: string[];
```
iuioiua: I think this should be a [`Headers`](https://developer.mozilla.org/en-US/docs/Web/API/Headers) interface. I also think the same can be said about `serveDir`.

kt3k:
The implementation looks good to me. However can you expose `headers` option only from `unstable_file_server.ts` for now? (We currently have the policy of not accepting new features directly to stable APIs https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#suggesting-a-new-feature )

See how `serveDir` is exposed in `unstable_file_server.ts` for reference.

moumar:
Sounds good to me. Let me update it.

http/unstable_file_server.ts
```diff
@@ -48,16 +49,67 @@ export function serveDir(
   req: Request,
   opts: ServeDirOptions = {},
 ): Promise<Response> {
-  return stableServeDir(req, opts);
+  const headers = opts.headers
```
iuioiua: Why use headers as a string that has to be parsed instead of `Headers`?

moumar: It is a compatibility layer from unstable interface (with Headers) to stable one (with string[]) to avoid rewrite implementation.

iuioiua: IMO, they should both be changed in the unstable API.

moumar: You mean using Headers even in the stable part ? It will break compatibility no ?

kt3k: @iuioiua

> IMO, they should both be changed in the unstable API

If we make these changes, what happens when we stabilize these APIs?

iuioiua: Highlight them as minor breaking changes within the change notes.

kt3k: no such thing like 'minor breaking changes'. Every breaking change requires major version upgrade because `@std` follows semantic versioning now. and I don't think this change isn't worth the breaking change

kt3k: `string[] | Headers` might be fine, but `Headers` doesn't make sense to me.

kt3k: Changed back the type of `headers` to `string[]` for the compatibility with stable `serveDir`. The change of exiting option type should be an independent topic in my view. Please open a new issue for it if you feel strongly about it

PR #6850, opened 2025-10-13
fix(fs,path): `expandGlob`, `expandGlobSync`, and `isGlob` match `@(alternate)` syntax with only one member

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6849, a bug where the `isGlob` function, which is a dependency of `expandGlob` and `expandGlobSync`, failed to recognize the `@(alternate)` extended glob syntax when there was only one member inside the parentheses.


kt3k (approved):
Good catch! Thanks for the fix. LGTM 👍

PR #6853, opened 2025-10-20
fix(http): runtime types for `getCookies`

lionel-rowe:
Fixes https://github.com/denoland/std/issues/6852


kt3k:
Runtime change looks good to me. I'm not in favor of the type change. The change looks too breaking to the ecosystem to me. Also if a user prefers to get 'correct' type when accessing `cookies['key']`, they can enable `noUncheckedIndexedAccess` flag in their "compilerOption" (The same argument applies to the all Record-returning APIs in general)

PR #6855, opened 2025-10-21
docs(random): use PRNG initialization with secure random seed

iugo:
Updated the random number generator to use a cryptographically secure seed.

The previous examples were not convenient for users to understand and use, so they should be adjusted to examples that are more likely to be used in real situations, making it easier for users to understand.

### DIFF
--- a/random/mod.ts
+++ b/random/mod.ts
@@ -22,7 +22,7 @@
  * import { randomIntegerBetween } from "@std/random";
  * import { randomSeeded } from "@std/random";
  *
- * const prng = randomSeeded(BigInt(crypto.getRandomValues(new Uint32Array(1))[0]));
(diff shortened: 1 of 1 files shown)


kt3k:
I think there are legitimate usages of fixed seed numbers (Like, when you want randomly-looking deterministic numbers). Can you append your example to the existing one instead of replacing it?

iugo:
> I think there are legitimate usages of fixed seed numbers (Like, when you want randomly-looking deterministic numbers). Can you append your example to the existing one instead of replacing it?

@kt3k OK. I updated the code changes to add the example instead of replacing it.

kt3k (approved):
LGTM. Thanks for updating!

PR #6856, opened 2025-10-22
fix(assert,streams): handle SharedArrayBuffer being disabled in some browser context

lucsoft:
[Per MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/SharedArrayBuffer#security_requirements) SharedArrayBuffer require secure context and cross-origin isolated. When these contextes are not given then equal and new Buffer fail (as SharedArrayBuffer is undefined) when called in the Web.

### DIFF
--- a/streams/buffer.ts
+++ b/streams/buffer.ts
@@ -135,11 +135,8 @@ export class Buffer {
   constructor(ab?: ArrayBufferLike | ArrayLike<number>) {
     if (ab === undefined) {
       this.#buf = new Uint8Array(0);
-    } else if (ab instanceof SharedArrayBuffer) {
-      // Note: This is necessary to avoid type error
-      this.#buf = new Uint8Array(ab);
     } else {
-      this.#buf = new Uint8Array(ab);
+      this.#buf = new Uint8Array(ab as ArrayBuffer | SharedArrayBuffer);
     }
   }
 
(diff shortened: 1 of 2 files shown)


streams/buffer.ts
```diff
@@ -135,11 +135,8 @@ export class Buffer {
   constructor(ab?: ArrayBufferLike | ArrayLike<number>) {
     if (ab === undefined) {
       this.#buf = new Uint8Array(0);
-    } else if (ab instanceof SharedArrayBuffer) {
-      // Note: This is necessary to avoid type error
-      this.#buf = new Uint8Array(ab);
     } else {
-      this.#buf = new Uint8Array(ab);
+      this.#buf = new Uint8Array(ab as ArrayBuffer | SharedArrayBuffer);
```
lucsoft, line 139: Image

Due to typing constraints, moving it to how it was done in assert/equal.t.s As else i would get in type erorrs that i don't see how i could resolve them without writing a type guard function

BlackAsLight: That's a weird error. The typings seems to think the underlying source can only be `ArrayBuffer` and not `SharedArrayBuffer` because `ArrayLike<number>` is an option. Casting to `ArrayBufferLike` may be a better option than explicitly list both options.

lucsoft: This is what I had before which I changed back to align with the other code. As it seems this cast is more correct then a ArrayBufferLike cast

BlackAsLight: If it was me I'd just cast to any since I know it's correct and it isn't exposed to external users.

PR #6857, opened 2025-10-23
feat(cache): add cleanup callback to `LruCache` and `TtlCache`

### LINKED ISSUE #6851: `LruCache` and `TtlCache` cleanup callback when values ejected from the cache

In some cases, it can be useful to have a cleanup callback for when values are ejected from an `LruCache` or a `TtlCache`. For example, if the cached value contains a blob URL created with `URL.createObjectURL`, the cleanup should call `URL.revokeObjectURL`.

Probably an `options` constructor param with something like  `{ cleanupCallback?: (ejectedKey: K, ejectedValue: V) => void }`.

None

WasixXD:
Closes #6851 

This PR adds a `option: { onEject: (ejectedKey: K, ejectedValue: V) => void) }` param to `LruCache` and `TtlCache` constructor.

### DIFF
--- a/cache/lru_cache.ts
+++ b/cache/lru_cache.ts
@@ -38,6 +38,25 @@ export type { MemoizationCache };
  * // less recent values are removed
  * assert(!cache.has("a"));
  * ```
+ *
+ * @example Adding a onEject function.
+ * ```ts
+ * import { LruCache } from "@std/cache";
+ * import { assertEquals } from "@std/assert";
+ *
+ * const cache = new LruCache<string, string>(3, { onEject: (key, value) => {
+ *  console.log("Revoking: ", key)
+ *  URL.revokeObjectURL(value)
+ * }});
+ *
+ * cache.set(
+ *  "fast-url",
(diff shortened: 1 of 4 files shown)


lionel-rowe:
Hmm, could work, I think it depends less on cleanness of the interface, and more on whether setting/changing/adding (/removing?) eject handlers after creation is a useful feature and/or a footgun. Also what should this do?

```ts
const cache = new LruCache(n)
    .onEject(() => console.log(1))
    .onEject(() => console.log(2))
```

WasixXD:
> Also what should this do?

As for now, this would just print 2

> more on whether setting/changing/adding (/removing?) eject handlers after creation is a useful feature and/or a footgun

I can't think in a scenario where this would be bad

kt3k:
The main idea looks good to me.

> I changed from a options param to a method because i think is cleaner, but I'm not attached to it.

I'd prefer this to be an option instead of a method. Making it an option would make it clearer that it only handles a single callback, not multiple ones.

WasixXD:
> I'd prefer this to be an option instead of a method. Making it an option would make it clearer that it only handles a single callback, not multiple ones.

🫡

PR #6860, opened 2025-10-29
chore: remove `lint:export-names` task

iuioiua:
It seems the [`deno-style-guide/naming-convention` lint rule](https://github.com/denoland/std/blob/a0acfb0084c252ec854fa04a2caf7c043f201375/_tools/lint_plugin.ts#L121) takes care of this sufficiently.


kt3k:
`lint:export-names` seems checking "exports" field of deno.json, preventing errors like https://github.com/denoland/std/issues/6574

Is that covered by style-guide lint rule?

iuioiua:
Oh. No, it doesn't. Whoops, my misunderstanding 🤣

PR #6862, opened 2025-11-03
feat(cli/unstable): custom progress bar refresh interval

KnorpelSenf:
The progress bar only updates every second. For some tasks, it would be nice to specify a different refresh interval.

This pull requests adds a new option to the progress bar which lets developers specify the refresh rate.


kt3k (approved):
The suggestion makes sense to me. LGTM

PR #6868, opened 2025-11-13
fix(fs): "Deno is not defined" when using the module in browser

disjukr:
my env
```
deno --version
deno 2.5.6 (stable, release, aarch64-apple-darwin)
v8 14.0.365.5-rusty
typescript 5.9.2
```

how to reproduce
```sh
echo 'import "jsr:@std/fs@1.0.19";' > test.js
deno bundle --platform browser -o out.js test.js
cat out.js
```

`out.js`
```
// line 12
var EXISTS_ERROR = new Deno.errors.AlreadyExists("dest already exists.");
//                     ^^^^ this makes problem in the web browser
```

### DIFF
--- a/fs/move.ts
+++ b/fs/move.ts
@@ -2,8 +2,6 @@
 import { isSubdir } from "./_is_subdir.ts";
 import { isSamePath } from "./_is_same_path.ts";
 
-const EXISTS_ERROR = new Deno.errors.AlreadyExists("dest already exists.");
-
 /** Options for {@linkcode move} and {@linkcode moveSync}. */
 export interface MoveOptions {
   /**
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

This looks also good for better stack trace. Thanks

kt3k:
BTW do you actually use any feature of `std/fs` in browser? Or is it just included as a dependency?

disjukr:
> BTW do you actually use any feature of std/fs in browser? Or is it just included as a dependency?

it just included as a dependency 😞

timreichen:
My two cents after the merge: I think the error initialization should be put after the `lstat()` check as it is is not needed if the call throws.

PR #6870, opened 2025-11-16
chore: update assertion of NotCapable error

kt3k:
We use Deno version detection for choosing between `NotCapable` and `PermissionDenied` to use in testing. This stopped working with Deno CLI canary as it updated the version to 3.x https://github.com/denoland/deno/pull/31258

This PR stops using version detection, instead just checks `Deno.errors.NotCapable` property for detecting availablility of it.


dsherret (approved):
LGTM

PR #6879, opened 2025-11-26
chore: set `DENO_UNSTABLE_TSGO=1` in canary in CI

iuioiua:


kt3k:
There seems one type error in `@std/toml`. Probably a regression in tsgo side?

iuioiua:
Yeah, I thought so too. @nathanwhit, any ideas?

iuioiua:
@kt3k, take another look at the type check failure in `@std/toml`. Might the failure actually be correct?

lunadogbot (approved):
Conditional `DENO_UNSTABLE_TSGO=1` in the canary matrix and the `env:` block on the lint job both look right — and the earlier `@std/toml` tsgo regression that held this up is no longer failing on canary, so the gate is clean. CI is green across all matrices.

PR #6886, opened 2025-11-28
chore: fmt

kt3k:
`deno fmt` results have been slightly changed in canary and we need these changes. (probably affected by https://github.com/denoland/deno/pull/31427)

### DIFF
--- a/text/unstable_reverse_test.ts
+++ b/text/unstable_reverse_test.ts
@@ -35,7 +35,10 @@ Deno.test("reverse() reverses a string", () => {
 
 // CREDIT: https://github.com/mathiasbynens/esrever/blob/14b34013dad49106ca08c0e65919f1fc8fea5331/README.md
 Deno.test("reverse() handles unicode strings", () => {
-  testUnicode("Lorem ipsum 𝌆 dolor sit ameͨ͆t.", ".teͨ͆ma tis rolod 𝌆 muspi meroL");
+  testUnicode(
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):

text/unstable_reverse_test.ts
```diff
@@ -35,7 +35,10 @@ Deno.test("reverse() reverses a string", () => {
 
 // CREDIT: https://github.com/mathiasbynens/esrever/blob/14b34013dad49106ca08c0e65919f1fc8fea5331/README.md
 Deno.test("reverse() handles unicode strings", () => {
-  testUnicode("Lorem ipsum 𝌆 dolor sit ameͨ͆t.", ".teͨ͆ma tis rolod 𝌆 muspi meroL");
+  testUnicode(
+    "Lorem ipsum 𝌆 dolor sit ameͨ͆t.",
+    ".teͨ͆ma tis rolod 𝌆 muspi meroL",
+  );
```
dsherret, line 41: This is strange. Let me look into it.

dsherret: I looked into it and it seems each string used to be a width of 29 in unicode-width 0.2.0 and now it's 30 in unicode-width 0.2.2: https://github.com/denoland/deno/pull/31427/files#diff-13ee4b2252c9e516a0547f2891aa2105c3ca71c6d7a1e682c69be97998dfc87eR9971

This seems ok because vscode considers this as a width of 31.

PR #6892, opened 2025-12-01
refactor(dotenv): avoid recursive `expand()`

timreichen:
This PR refactors the `expand()` fn so it uses a loop instead of calling it recursively. This makes it much more readable.

### DIFF
--- a/async/deno.json
+++ b/async/deno.json
@@ -14,6 +14,7 @@
     "./unstable-retry": "./unstable_retry.ts",
     "./tee": "./tee.ts",
     "./unstable-throttle": "./unstable_throttle.ts",
-    "./unstable-wait-for": "./unstable_wait_for.ts"
+    "./unstable-wait-for": "./unstable_wait_for.ts",
+    "./unstable-semaphore": "./unstable_semaphore.ts"
   }
 }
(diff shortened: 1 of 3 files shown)


dotenv/parse.ts
```diff
@@ -27,30 +27,29 @@ function expandCharacters(str: string): string {
...
+        inBracketsDefault,
+        notInBrackets,
+        notInBracketsDefault,
+      } = params.at(-1);
+
+      const expandValue = inBrackets ?? notInBrackets;
+      const defaultValue = inBracketsDefault ?? notInBracketsDefault;
+
+      return (
+        variablesMap[expandValue] ?? Deno.env.get(expandValue) ?? defaultValue
```
kt3k, line 47: The existing code seems calling `expand` on `defaultValue`?
```suggestion
        variablesMap[expandValue] ?? Deno.env.get(expandValue) ?? expand(defaultValue, variablesMap)
```

kt3k: ah, nevermind. variables in `defaultValue` seems handled in the future iteration (and such cases are covered in test cases)

kt3k (approved):
LGTM. Thanks!

kt3k:
BTW it looks like `deno run --env-file` does not support default value substitution in variable expansion

PR #6895, opened 2025-12-01
refactor(http): removing unreachable conditions, increasing coverage

achedges:
This change removes a handful of conditional checks that appear to be unreachable and preventing full test line coverage.  I'll leave a comment on each change to explain, but feel free to reject if you'd rather keep these checks in place for sanity, or if I'm just missing something :)

### DIFF
--- a/http/cookie.ts
+++ b/http/cookie.ts
@@ -173,10 +173,7 @@ function validateName(name: string | undefined | null) {
  * See {@link https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1.2.4}.
  * @param path Path value.
  */
-function validatePath(path: string | null) {
-  if (path === null) {
-    return;
-  }
+function validatePath(path: string) {
   for (let i = 0; i < path.length; i++) {
     const c = path.charAt(i);
     if (
@@ -193,10 +190,10 @@ function validatePath(path: string | null) {
 /**
  * Validate Cookie Value.
  * See {@link https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1}.
+ * @param name Cookie name.
  * @param value Cookie value.
  */
-function validateValue(name: string, value: string | null) {
-  if (value === null) return;
+function validateValue(name: string, value: string) {
   for (let i = 0; i < value.length; i++) {
     const c = value.charAt(i);
     if (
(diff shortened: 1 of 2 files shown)


http/cookie.ts
```diff
@@ -173,10 +173,7 @@ function validateName(name: string | undefined | null) {
  * See {@link https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1.2.4}.
  * @param path Path value.
  */
-function validatePath(path: string | null) {
-  if (path === null) {
-    return;
-  }
+function validatePath(path: string) {
```
achedges, line 176: `validatePath()` is only accessible from within the class, and it is only called if `cookie.path` is truthy, so as-is the `path` argument can never be `null`.

http/cookie.ts
```diff
@@ -193,10 +190,10 @@ function validatePath(path: string | null) {
 /**
  * Validate Cookie Value.
  * See {@link https://www.rfc-editor.org/rfc/rfc6265.html#section-4.1}.
+ * @param name Cookie name.
  * @param value Cookie value.
  */
-function validateValue(name: string, value: string | null) {
-  if (value === null) return;
+function validateValue(name: string, value: string) {
```
achedges, line 196: `validateValue()` is only accessible from within the class, and it is only called on a `Cookie` instance, which does not accept a `null` `value`.

http/cookie.ts
```diff
@@ -258,10 +255,10 @@ export function getCookies(headers: Headers): Record<string, string> {
     const c = cookie.split(";");
     for (const kv of c) {
       const [cookieKey, ...cookieVal] = kv.split("=");
-      if (cookieKey === undefined) {
+      if (cookieKey === "") {
```
achedges, line 258: `cookieKey` cannot be `undefined` as the `.split()` function will at least return an empty array.  Instead, an empty string should indicate that `=` was provided as a key.

http/cookie.ts
```diff
@@ -347,13 +344,9 @@ function parseSetCookie(value: string): Cookie | null {
...
     });
 
-  if (!attrs[0]) {
-    return null;
-  }
-
   const cookie: Cookie = {
-    name: attrs[0][0],
-    value: attrs[0][1],
+    name: attrs[0]![0],
```
achedges, line 348: `attrs` should always be at least an empty array.

PR #6897, opened 2025-12-08
feat(uuid): stabilize uuidv7 module

### LINKED ISSUE #6174: @std/uuid: Stabilize `@std/uuid/v7`
The feature is what I want most in `@std/uuid`.
What is required for the stabilization?

### LINKED ISSUE #6772: Stablize `uuid` package's `v7` implementation.
At this point I'm not sure how this package could change. Would like to open a discussion to stabilize it.

lino-levan:
Closes #6772
Closes #6174

I'm a little bit rusty since I haven't contributed in a little bit. Didn't see docs in `CONTRIBUTING.md` file explaining how stabilization works so I took my best shot at it. Let me know if there's anything else that needs to be done.

### DIFF
--- a/uuid/mod.ts
+++ b/uuid/mod.ts
@@ -34,7 +34,7 @@ import { generate as generateV3, validate as validateV3 } from "./v3.ts";
 import { validate as validateV4 } from "./v4.ts";
 import { generate as generateV5, validate as validateV5 } from "./v5.ts";
 import {
-  extractTimestamp as extractTimestampV7,
+  extractTimestamp,
   generate as generateV7,
   validate as validateV7,
 } from "./v7.ts";
@@ -130,5 +130,5 @@ export const v5 = {
 export const v7 = {
   generate: generateV7,
   validate: validateV7,
-  extractTimestampV7,
+  extractTimestamp,
 };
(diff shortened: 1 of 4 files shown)


lino-levan:
I should note what I didn't in my issue. The only change that I could see to this code is the type of the function returning a string. In typescript crypto.createUUID() returns a specific type that this does not. Wondering if it was ever discussed why this is the behavior std chose?

lino-levan:
I suppose at this point consistency is more important anyways and it would be weird to have one api have a different generate/validate typing. Maybe a 2.0 of this package could do that.

kt3k (approved):
LGTM

uuid/mod.ts
```diff
@@ -33,6 +33,11 @@ import { generate as generateV1, validate as validateV1 } from "./v1.ts";
 import { generate as generateV3, validate as validateV3 } from "./v3.ts";
 import { validate as validateV4 } from "./v4.ts";
 import { generate as generateV5, validate as validateV5 } from "./v5.ts";
+import {
+  extractTimestamp,
```
kt3k: a minor point for some consideration

I wonder if it's ok to export this name from here as all other top level methods have version suffixes.. What do you think about the name `extractTimestampV7`?

lino-levan: This makes sense, I think. Other UUID types actually do store timestamps and we don't have functions to extract them right now.

lino-levan: Wait sorry, I misread, we shouldn't do this. It's already on the v7 object that is being exported, we should do V7 on the import, in case in the future we decide to add more.

htunnicliff:
In case anyone is interested in weighing in, I've added a proposal for this in the WebCrypto WG:

- https://github.com/w3c/webcrypto/issues/561

PR #6904, opened 2025-12-20
refactor(yaml): add `resolveTag()` method

timreichen:
This PR moves bloat from `composeNode()` into a separate `resolveTag()` method.
That code is also rewritten to be more readable.


kt3k (approved):
LGTM

PR #6909, opened 2025-12-24
fix(streams/unstable): use relative path for in-package import

kt3k:
It looks like JSR is unhappy with the specifier `@std/streams/unstable-to-byte-stream` in `streams/unstable_capped_delimiter_stream.ts` (probably because it is jsr: specifier referencing in-package path)

This PR tries to avoid the error above by using a relative path.


dsherret (approved):
LGTM

PR #6910, opened 2025-12-24
feat(datetime): add month and day of week constants

a-skua:
## Summary

Add month index constants (JAN-DEC) and day of week constants (SUN-SAT) to `@std/datetime/constants`.

## Motivation

In many languages and cultures (such as Japanese, Chinese, and Korean), months are referred to by their ordinal numbers: January is "1月" (month 1), February is "2月" (month 2), and so on.

However, JavaScript's `Date` object uses 0-based indices for months (0 = January, 11 = December). This mismatch between human intuition and the API frequently leads to off-by-one bugs:

```ts
// Intended: March 1st, 2025
// Actual: April 1st, 2025 (oops!)
new Date(2025, 3, 1)
```

By providing named constants, developers can write clearer, less error-prone code:

```ts
import { MAR } from "@std/datetime/constants";

new Date(2025, MAR, 1)  // unambiguously March
```

```ts
import { SAT, SUN } from "@std/datetime/constants";

const day = date.getDay();
if (day === SAT || day === SUN) {
    console.log("It's the weekend!");
}
```

Changes

- Add month constants: JAN (0) through DEC (11)
- Add day of week constants: SUN (0) through SAT (6)

### DIFF
--- a/datetime/constants.ts
+++ b/datetime/constants.ts
@@ -56,3 +56,212 @@ export const DAY: number = HOUR * 24;
  * ```
  */
 export const WEEK: number = DAY * 7;
+/**
+ * The month index for January.
+ *
+ * @example
+ * ```ts
+ * import { JAN } from "@std/datetime/constants";
+ *
+ * new Date(2025, JAN, 1); // 2025-01-01
+ * ```
+ */
+export const JAN = 0;
+/**
+ * The month index for February.
+ *
+ * @example
+ * ```ts
+ * import { FEB } from "@std/datetime/constants";
+ *
+ * new Date(2025, FEB, 1); // 2025-02-01
+ * ```
+ */
+export const FEB = 1;
+/**
+ * The month index for March.
+ *
+ * @example
+ * ```ts
+ * import { MAR } from "@std/datetime/constants";
+ *
+ * new Date(2025, MAR, 1); // 2025-03-01
+ * ```
+ */
+export const MAR = 2;
+/**
+ * The month index for April.
+ *
+ * @example
+ * ```ts
+ * import { APR } from "@std/datetime/constants";
+ *
+ * new Date(2025, APR, 1); // 2025-04-01
+ * ```
+ */
+export const APR = 3;
+/**
+ * The month index for May.
+ *
+ * @example
+ * ```ts
+ * import { MAY } from "@std/datetime/constants";
+ *
+ * new Date(2025, MAY, 1); // 2025-05-01
+ * ```
+ */
+export const MAY = 4;
+/**
+ * The month index for June.
+ *
+ * @example
+ * ```ts
+ * import { JUN } from "@std/datetime/constants";
+ *
+ * new Date(2025, JUN, 1); // 2025-06-01
+ * ```
+ */
+export const JUN = 5;
+/**
+ * The month index for July.
+ *
+ * @example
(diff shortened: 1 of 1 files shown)


timreichen:
Interesting proposal.
I am not sure if that makes sense for std because I think a chunk of the `datetime` mod will become deprecated after `Temporal` becomes stable. `Temporal` fixed the 0-based month confusion by making months 1 based, which ironically would make these constants create off-by-one bugs when initializing dates:
```ts
new Temporal.PlainDate(2025, MAR, 1); // incorrect month
```

a-skua:
Thank you for pointing that out — I hadn't fully considered the Temporal API implications.

That said, I'm curious about the timeline: even after Temporal stabilizes, when would `Date` actually become deprecated?
I think, `Date` will remain in use for quite a while.

To avoid future confusion, what do you think about providing these constants under a more specific path, such as `@std/datetime/date/constants`? This would make it explicit that these constants are intended for use with the legacy `Date` API, not Temporal.

```diff
-import { MAR } from "@std/datetime/constants";
+import { MAR } from "@std/datetime/date/constants";

new Date(2025, MAR, 1)  // unambiguously March
```

I fully agree with removing these constants once they are no longer needed.

datetime/constants.ts
```diff
@@ -56,3 +56,212 @@ export const DAY: number = HOUR * 24;
...
+ */
+export const FEB = 1;
+/**
+ * The month index for March.
+ *
+ * @example
+ * ```ts
+ * import { MAR } from "@std/datetime/constants";
+ *
+ * MAR; // 2
```
kt3k: Can you make the example more practical one like `new Date(2025, MAR, 1);`?

a-skua: Updated the examples to show practical usage: a8e56589d9c66e3202ee411f5179723d1b279ceb

kt3k:
I don't think `Date` will be deprecated, but `@std/datetime` can be deprecated in the future (We haven't completely decided yet though).

I checked major date libraries such as moment.js, date-fns, day.js, and found that they all still use 0-based month index. I guess maybe these constants still have some utilities. Anyways `@std/datetime` is still unstable package. I would say let's try them.

kt3k (approved):
LGTM

timreichen:
I think the const names should not be abbreviated.

a-skua:
I understand the preference for non-abbreviated names, but I think:

1. In web development, shorter names can be beneficial — especially when dealing with bundle sizes or when these constants are used frequently in code.
2. Within the @std/datetime module, abbreviations like JAN, FEB, SUN, MON are unambiguous. There's no realistic scenario where these could be misinterpreted as something other than month/day-of-week indices.

timreichen:
> I understand the preference for non-abbreviated names, but I think:
> 
> 
> 
> 1. In web development, shorter names can be beneficial — especially when dealing with bundle sizes or when these constants are used frequently in code.
> 
> 2. Within the @std/datetime module, abbreviations like JAN, FEB, SUN, MON are unambiguous. There's no realistic scenario where these could be misinterpreted as something other than month/day-of-week indices.

It is more about consistency in std.
Bundle sizes should not be a concern because bundlers generally are capable of changing declaration names on minification.

a-skua:
Addressed in #6939

PR #6912, opened 2025-12-24
fix(csv): return empty array for empty input with skipFirstRow

### LINKED ISSUE #6809: CSV parsing: header autodetection fails on empty string
```ts
import { parse } from 'jsr:@std/csv';

parse('');  // []
parse(' ', { skipFirstRow: true });  // []
parse('', { skipFirstRow: true });  // fails with the error below
```

error: Uncaught (in promise) TypeError: Cannot parse input: headers must be defined
        throw new TypeError("Cannot parse input: headers must be defined");
              ^
    at parse (https://jsr.io/@std/csv/1.0.6/parse.ts:516:15)

**Environment**

- OS: Ubuntu 25.04
- deno version: 2.4.5
- std version: 1.0.6

murataslan1:
## Description
Fixes #6809

When parsing an empty string with `skipFirstRow: true`, the function was throwing a TypeError because there were no headers to extract. This is inconsistent with parsing an empty string without options, which returns an empty array.

## Problem
```ts
parse('');                          // [] ✓
parse(' ', { skipFirstRow: true }); // [] ✓
parse('', { skipFirstRow: true });  // TypeError: Cannot parse input: headers must be defined ✗
```

## Solution
Return an empty array when there's no data to parse, instead of throwing an error.

## Changes
```diff
if (options.skipFirstRow) {
  const head = r.shift();
  if (head === undefined) {
-   throw new TypeError("Cannot parse input: headers must be defined");
+   return [] as ParseResult<ParseOptions, T>;
  }
  headers = head;
}
```

## After
```ts
parse('');                          // [] ✓
parse('', { skipFirstRow: true });  // [] ✓
```

### DIFF
--- a/csv/parse.ts
+++ b/csv/parse.ts
@@ -513,7 +513,7 @@ export function parse<const T extends ParseOptions>(
     if (options.skipFirstRow) {
       const head = r.shift();
       if (head === undefined) {
-        throw new TypeError("Cannot parse input: headers must be defined");
+        return [] as ParseResult<ParseOptions, T>;
       }
       headers = head;
     }


kt3k:
Thanks for trying to contribute, however I think the issue #6809 is invalid as there is a test case which explicitly states that the empty input with `skipFirstRow=true` should throw https://github.com/denoland/std/blob/6f53e943aa57e30b9994dba20ac1d22db3a027a9/csv/parse_test.ts#L918

I think the existing behavior should be considered intentional, not a bug.

The reason why `parse(' ', { skipFirstRow: true })` returns empty array is that ` ` is parsed as `[" "]` (single column with the name `" "`). So the header info is available here

Thanks for the PR anyways

PR #6913, opened 2025-12-24
fix(text/unstable): only strip single trailing newline in dedent

### LINKED ISSUE #6831: `text/unstable-dedent` strips too many trailing newlines
**Describe the bug**

`text/unstable-dedent` should only strip a single trailing newline, but instead it strips all trailing newlines.

**Steps to Reproduce**

```ts
import { assertEquals } from 'jsr:@std/assert';
import { dedent as stdDedent } from 'jsr:@std/text@1.0.16/unstable-dedent';
import npmDedent from 'npm:string-dedent@3.0.2';

for (const [name, dedent] of Object.entries({
  'jsr:@std': stdDedent,
  'npm:string-dedent': npmDedent,
})) {
  Deno.test(name, async (t) => {
    const v1 = dedent`
      a

    `;

    // throws for jsr:@std, ok for npm:string-dedent
    await t.step('v1', () => assertEquals(v1, 'a\n'));

    const v2 = dedent`
      a\n
    `;

    // throws for jsr:@std, ok for npm:string-dedent
    await t.step('v2', () => assertEquals(v2, 'a\n'));
  });
}
```

**Expected behavior**

Only strip a single trailing newline.

**Environment**

text@1.0.16

murataslan1:
## Description
Fixes #6831

The `dedent` function was using `.trimEnd()` which removes ALL trailing whitespace including multiple newlines. This is inconsistent with `npm:string-dedent` which only removes a single trailing newline.

## Changes
- Replace `.trimEnd()` with `.replace(/\n[\t ]*$/, "")` to only strip a single trailing newline (with any preceding whitespace on that line)
- Updated documentation to reflect the new behavior

## Before
```ts
dedent`
  a

`  // returns 'a' (all trailing newlines stripped)
```

## After
```ts
dedent`
  a

`  // returns 'a\n' (only one trailing newline stripped, matching npm:string-dedent)
```

## Testing
The fix ensures consistency with `npm:string-dedent@3.0.2` behavior.

### DIFF
--- a/text/unstable_dedent_test.ts
+++ b/text/unstable_dedent_test.ts
@@ -35,7 +35,7 @@ Deno.test("dedent() handles empty lines", () => {
   assertEquals(
     dedent(`
       a
-
+        
         b
         \t\t\t
        c


kt3k (commented):
Looks in a good direction. Can you format source code with `deno fmt`? Also can you add a test case to verify the example in #6831 ?

text/unstable_dedent_test.ts
```diff
@@ -22,6 +22,29 @@ Deno.test("dedent() handles example 2", () => {
...
+// Test case for issue #6831: only strip single trailing newline, not all
+Deno.test("dedent() only strips single trailing newline (issue #6831)", async (t) => {
+  await t.step(
+    "preserves content newline when template ends with single newline",
+    () => {
+      // Template: `\n  a\n  ` -> should result in 'a\n' (matching npm:string-dedent)
+      const result = dedent`
+      a
+      `;
+      assertEquals(result, "a");
```
kt3k: This case seems already passing with the existing code i.e. it doesn't seem testing what this PR changes. I'm removing this case

PR #6918, opened 2025-12-28
docs(fs): fix `walk` examples

pasha-bolokhov:
Correct the description of examples of usage of the`match` and `skip` options


kt3k (approved):
Good catch! Thanks

PR #6920, opened 2025-12-30
perf(random): inline random index calculation in shuffle()

tomas-zijdemans:
- Inlined the random index calculation instead of calling randomIntegerBetween
- Cached the PRNG function lookup outside the loop
- Used a temp-variable swap instead of array destructuring
- Added some basic coverage for edge cases

Benchmarks show roughly 1.15-1.25x speedup depending on array size, mainly from reducing per-iteration overhead. The gains are modest but consistent across different input sizes.

### DIFF
--- a/random/shuffle_test.ts
+++ b/random/shuffle_test.ts
@@ -46,15 +46,18 @@ Deno.test("shuffle() shuffles the provided array", () => {
 
 Deno.test("shuffle() with PRNG always returning 0 produces predictable result", () => {
   const items = [1, 2, 3, 4, 5];
-  // prng() = 0 means j = 0 every time, so each element swaps with position 0
-  const shuffled = shuffle(items, { prng: () => 0 });
-  assertEquals(shuffled, [2, 3, 4, 5, 1]);
+  assertEquals(
+    shuffle(items, { prng: () => 0 }),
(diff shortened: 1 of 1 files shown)


random/shuffle_test.ts
```diff
@@ -33,6 +44,29 @@ Deno.test("shuffle() shuffles the provided array", () => {
   assertEquals(shuffle(items, { prng }), [2, 4, 5, 3, 1]);
 });
 
+Deno.test("shuffle() with PRNG always returning 0 produces predictable result", () => {
+  const items = [1, 2, 3, 4, 5];
+  // prng() = 0 means j = 0 every time, so each element swaps with position 0
+  const shuffled = shuffle(items, { prng: () => 0 });
+  assertEquals(shuffled, [2, 3, 4, 5, 1]);
```
kt3k: nit: I think this assertion is unnecessary guarantee about too much details. We might want to change the shuffling algorithms later, but this assertion will prevent it

tomas-zijdemans: Very good point. I will try a new PR to address it.

tomas-zijdemans: https://github.com/denoland/std/pull/6932

PR #6921, opened 2025-12-30
fix(async): improve deadline() validation and performance

tomas-zijdemans:
- Add NaN input validation (throws TypeError)
- Fix docs to clarify >= MAX_SAFE_INTEGER boundary
- Skip AbortSignal.any() overhead for single-signal cases
- Add test for NaN handling

Benchmarks show ~2.7× faster for the common timeout-only case, ~7× faster when no deadline is needed. The dual-signal path is roughly equivalent.


tomas-zijdemans:
I closed this PR, since the NaN check seems overkill and the code got a bit over-engineered. Feel free to merge it if you still like it :D

PR #6922, opened 2025-12-30
refactor(async): improve retry() input validation

tomas-zijdemans:
Switched to using delay() for timeout handling.

Input handling:
- Uses RangeError instead of TypeError for invalid options
- Validates maxAttempts is a positive integer
- Validates multiplier is a finite number >= 1
- Validates jitter is between 0 and 1 (was only checking > 1)
- Adds NaN checks for maxTimeout, minTimeout, and jitter


kt3k:
We have the policy of accepting a new feature to stable APIs as unstable feature first. See https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100

Can you move feature addition part of this change to a different PR, targeting `unstable_retry.ts`?

tomas-zijdemans:
> We have the policy of accepting a new feature to stable APIs as unstable feature first. See https://github.com/denoland/std/blob/main/.github/CONTRIBUTING.md#new-features-in-stable-packages-version--100
> 
> Can you move feature addition part of this change to a different PR, targeting `unstable_retry.ts`?

Sorry, I should have considered that!

I've reverted the new feature, while still keeping delay() instead of setTimeout. I hope this is acceptable (it allows retry to handle very large timeouts correctly)

PR #6923, opened 2025-12-30
refactor(async): optimize delay() for common timeout durations

tomas-zijdemans:
Added an early return for the "normal" path

Fixed unclear comment (delay does not throw, it returns a rejected promise)


kt3k (approved):
LGTM. Thanks!

PR #6924, opened 2025-12-31
refactor(semver): removing unreachable code, bringing coverage to 100%

achedges:
This change refactors a handful of internal functionality to remove unreachable code, and adds a couple of new tests to bring up line coverage.

Comments added inline, I hope this helps!

### DIFF
--- a/semver/parse_range_test.ts
+++ b/semver/parse_range_test.ts
@@ -683,6 +683,7 @@ Deno.test("parseRange() handles wildcards", () => {
   assertEquals(parseRange("<1.*.*"), [
     [{ operator: "<", major: 1, minor: 0, patch: 0 }],
   ]);
+  // FIXME(kt3k): The following case should equal `[{ operator: "<", major: 2, minor: 0, patch: 0 }]`
   assertEquals(parseRange("<=1.*.2"), [
     [{ operator: "<", major: 1, minor: NaN, patch: 0 }],
   ]);


semver/greater_than_range.ts
```diff
@@ -54,7 +54,7 @@ function lessThanComparator(version: SemVer, comparator: Comparator): boolean {
     case undefined:
       return cmp < 0;
     case "!=":
-      return false;
+      return cmp >= 0;
```
achedges, line 57: I'll admit this one had me scratching my head a bit.  It seems that, in the context of this internal utility function, simply inverting the outcome for the `!=` operator (compared to the `=` operator) should suffice for producing the intended effect.

semver/less_than_range.ts
```diff
@@ -56,7 +56,7 @@ function greaterThanComparator(
     case undefined:
       return cmp > 0;
     case "!=":
-      return false;
+      return cmp <= 0;
```
achedges, line 59: See comments on `greater_than_range.ts`

semver/parse_range.ts
```diff
@@ -27,29 +26,8 @@ function parseComparator(comparator: string): Comparator | null {
 
   if (!groups) return null;
 
-  const { operator, prerelease, buildmetadata } =
```
achedges, line 30: Unless I'm missing something here, if the regex match succeeds at all, the resulting groups will contain elements for each semver component.  This code was attempting to manually parse what this regex was already doing.

semver/parse_range.ts
```diff
@@ -347,7 +324,15 @@ function handleEqualOperator(groups: RangeRegExpGroups): Comparator[] {
   }
   const prerelease = parsePrerelease(groups.prerelease ?? "");
   const build = parseBuild(groups.build ?? "");
-  return [{ operator: undefined, major, minor, patch, prerelease, build }];
+
```
achedges, line 327: We can use the `=` string if we have it, the net-effect is the same.

semver/parse_range.ts
```diff
@@ -369,13 +354,8 @@ function parseOperatorRange(string: string): Comparator | Comparator[] | null {
...
     case ">=":
       return handleGreaterOrEqualOperator(groups);
-    case "=":
-    case "":
-      return handleEqualOperator(groups);
     default:
-      throw new Error(
-        `Cannot parse version range: '${groups.operator}' is not a valid operator`,
-      );
+      return handleEqualOperator(groups);
```
achedges, line 358: The existing `default` case should not be possible, so we can use the equals/empty operator to exercise a switch default.

semver/greater_than_range_test.ts
```diff
@@ -204,6 +205,5 @@ Deno.test("greaterThanRange() handles not equals operator", () => {
     prerelease: [],
     build: [],
   }]];
-  // FIXME(kt3k): This demonstrates a bug. This should be false
-  assertEquals(greaterThanRange(version, range), true);
+  assertEquals(greaterThanRange(version, range), false);
```
kt3k, line 208: oh, thanks for fixing this!

semver/less_than_range_test.ts
```diff
@@ -186,6 +186,5 @@ Deno.test("lessThanRange() handles not equals operator", () => {
     prerelease: [],
     build: [],
   }]];
-  // FIXME(kt3k): This demonstrates a bug. This should be false
-  assertEquals(lessThanRange(version, range), true);
+  assertEquals(lessThanRange(version, range), false);
```
kt3k, line 189: 👍

semver/parse_range_test.ts
```diff
@@ -677,6 +683,9 @@ Deno.test("parseRange() handles wildcards", () => {
   assertEquals(parseRange("<1.*.*"), [
     [{ operator: "<", major: 1, minor: 0, patch: 0 }],
   ]);
+  assertEquals(parseRange("<=1.*.2"), [
+    [{ operator: "<", major: 1, minor: NaN, patch: 0 }],
```
kt3k, line 688: I think this should be `[{ operator: "<", major: 2, minor: 0, patch: 0 }]` (though it's the existing behavior)

`npm:semver` works like the below:
```ts
import * as semver from "npm:semver";

console.log(semver.satisfies("1.9.9", '<=1.*.2')) // => prints true
console.log(semver.satisfies("2.0.0", '<=1.*.2')) // => prints false
```

I'm adding `FIXME` comment above for now.

kt3k (approved):
Thank you for working on this! LGTM

semver/parse_range_test.ts
achedges: Got it!  I'll take a look, thanks for the pointer.

PR #6926, opened 2026-01-01
fix(async/unstable): make circuit breaker state getter pure

tomas-zijdemans:
Fixed an issue where accessing .state can mutate internal state and fire onStateChange callback.

Also added a note documenting a known race condition


kt3k (approved):
LGTM

PR #6930, opened 2026-01-04
fix(collections): correct overload ordering

tomas-zijdemans:
The type-guard overload (`el is U`) was declared after the generic boolean overload. I might be wrong, but I think since Typescript matches overload from top to bottom, having the generic one first would means it always wins in the current code ? 🤷 

The below code fails `deno check` on main, but compiles after this fix.

```
import { partition } from "@std/collections/partition";

// Test: type-guard should narrow types
const mixed: (string | number)[] = [1, "a", 2, "b"];
const isString = (x: string | number): x is string => typeof x === "string";

const [strings, numbers] = partition(mixed, isString);

// These assignments will fail at compile time
const _verifyStrings: string[] = strings;
const _verifyNumbers: number[] = numbers;

console.log("strings:", strings);  // ["a", "b"]
console.log("numbers:", numbers);  // [1, 2]
```

**Other changes:**

- Update doc, since both overloads had identical examples (it % 2 === 0), which didn't demonstrate the value of the type-guard overload
- Added tests for type-guard and non-array iterables

### DIFF
--- a/collections/partition_test.ts
+++ b/collections/partition_test.ts
@@ -73,3 +73,42 @@ Deno.test({
     );
   },
 });
+
+Deno.test({
+  name: "partition() handles type guard predicates",
+  fn() {
+    const mixed: (string | number)[] = [1, "a", 2, "b"];
+    const isString = (x: string | number): x is string => typeof x === "string";
+    const [strings, numbers] = partition(mixed, isString);
+
+    assertEquals(strings, ["a", "b"]);
+    assertEquals(numbers, [1, 2]);
+  },
+});
+
+Deno.test({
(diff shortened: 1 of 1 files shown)


kt3k:
> Typescript matches overload from top to bottom, having the generic one first would means it always wins in the current code ? 🤷

I think you are right. Thanks for catching this

ref https://github.com/microsoft/TypeScript/issues/1860#issuecomment-72154737

> Overload ordering absolutely matters. The rule is that the first applicable overload is chosen.

kt3k (approved):
LGTM!

PR #6931, opened 2026-01-04
refactor(collections): optimize zip()

tomas-zijdemans:
zip() is already fast, so this was more just as an experiment. Still, after benchmarking this change, zip seems to run ~3x faster. So might be worth it?

- Replaced .map() with manual loops
- Inlined minOf() calculation
- Added sparse array test


kt3k (approved):
LGTM 👍

PR #6932, opened 2026-01-05
test(random): replace exact-value assertions with property-based checks

tomas-zijdemans:
Attempting to address: https://github.com/denoland/std/pull/6920#discussion_r2660249148


kt3k (approved):
LGTM. Thanks!

PR #6935, opened 2026-01-05
fix(cache/unstable): call onEject for falsy values on delete

tomas-zijdemans:
Should we perhaps rename the ttl and lru files? 🤔  (so from `ttl_cache.ts` to `unstable_ttl_cache.ts` etc.)?

### DIFF
--- a/cache/lru_cache.ts
+++ b/cache/lru_cache.ts
@@ -200,9 +200,8 @@ export class LruCache<K, V> extends Map<K, V>
    * ```
    */
   override delete(key: K): boolean {
-    const value = super.get(key);
-    if (value) {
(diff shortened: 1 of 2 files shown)


kt3k:
> Should we perhaps rename the ttl and lru files? 🤔 (so from ttl_cache.ts to unstable_ttl_cache.ts etc.)?

No need of renames as `@std/cache` is entirely unstable. When a package is unstable (i.e. in 0.x versions), all APIs in there are considered unstable.

`unstable-` prefixes are only necessary when we start adding new unstable APIs in stable (i.e. version 1.0, or above) packages to distinguish stable and unstable APIs clearly.

PR #6936, opened 2026-01-05
perf(collections): optimize minOf/maxOf with fast array paths

tomas-zijdemans:
Optimizes `minOf` and `maxOf` to use the same pattern as `minBy`/`maxBy`:
- Fast path for arrays using index-based iteration
- Manual iterator for non-array iterables

Benchmark results (100K elements):
- Identity selector: ~3% faster (marginal)
- Object property selector: ~20% faster
- BigInt: ~22-25% faster  
- Generator/Iterable: ~25% faster
- Small arrays (10 elements): up to 3x faster

Added tests for iterable inputs, single-element arrays, and NaN-as-first-element.

### DIFF
--- a/collections/max_of_test.ts
+++ b/collections/max_of_test.ts
@@ -84,6 +84,24 @@ Deno.test("maxOf() handles iterable input", () => {
   assertEquals(actual, 8);
 });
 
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Nice!

PR #6939, opened 2026-01-06
refactor(datetime): rename abbreviated month and weekday constants to full names

a-skua:
## Summary
  - Rename month constants from abbreviated to full names (e.g., `JAN` → `JANUARY`)
  - Rename weekday constants from abbreviated to full names (e.g., `SUN` → `SUNDAY`)

  ## Motivation
  Addresses feedback from @timreichen in https://github.com/denoland/std/pull/6910#issuecomment-3694103693

### DIFF
--- a/datetime/constants.ts
+++ b/datetime/constants.ts
@@ -61,45 +61,45 @@ export const WEEK: number = DAY * 7;
  *
  * @example
  * ```ts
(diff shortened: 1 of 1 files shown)


kt3k (approved):
LGTM

Note to other maintainers: These values haven't been published yet. So this change can be landed as 'refactor'

PR #6942, opened 2026-01-07
feat(xml/unstable): add XML parsing and serialization module

tomas-zijdemans:
# New XML parsing and serialization module
What @std/xml **has**:
- Streaming parser, DOM-style parser, serialization
- Browser compatible, position tracking, spec-compliant

What @std/xml **doesn't have**:
- Namespace resolution, DTD/Schema validation, HTML entities
- Custom entities, XPath/selectors, object-to-XML builder

# Benchmark Results

Performance work never really ends, and you often find yourself comparing apples and oranges. Anyway. Here goes.

## The challengers
Library | XML Spec compliant? | Streaming XML parsing? | Error position tracking?
-- | -- | -- | --
SAX | No | Yes | Yes
saxes | Yes | Yes | Yes
fast-xml-parser | No | No | Yes
txml | No | Yes | No
xml2js | No | No | Partial
htmlparser2 | No | Yes| Partial
deno std | Yes | Yes | Yes (configurable)

Error position tracking is nice for debugging, but really hurts performance. So I made it an option that defaults to true for non-streaming and false for streaming (streaming is usually for trusted data sources. Multi-GB feeds or logs where throughput is critical). The results below contain both with and without error position tracking.

## Test data
I used the test files located in `testdata` for non-streaming. I used one 597MB file for the streaming benchmark (google product data), but didn't check that into `testdata`. Other payloads may give different results.

## Small Files (<10KB) — Median Results

Parser | Time (ms) | vs Deno std
-- | -- | --
Deno std (no pos) | 0.009 | 1.2x faster
txml | 0.010 | 1.1x faster
Deno std (+pos) | 0.011 | baseline
saxes | 0.016 | 1.5x slower
htmlparser2 | 0.022 | 2.0x slower
SAX | 0.027 | 2.5x slower
fast-xml-parser | 0.038 | 3.5x slower
xml2js | 0.048 | 4.4x slower

## 1 Large File (301KB) — Median Results

Parser | Time (ms) | vs Deno std
-- | -- | --
Deno std (no pos) | 1.90 | 1.0x same
Deno std (+pos) | 1.90 | baseline
txml | 1.98 | 1.04x slower
saxes | 2.36 | 1.2x slower
htmlparser2 | 4.17 | 2.2x slower
SAX | 7.60 | 4.0x slower
fast-xml-parser | 11.16 | 5.9x slower
xml2js | 14.10 | 7.4x slower

## Streaming (a 597MB file) — Median Results

Parser | Time (s) | vs Deno std
-- | -- | --
Deno std (no pos) | 4.25 | baseline
Deno std (+pos) | 4.68 |  1.1x slower
saxes | 5.46 | 1.3x slower
htmlparser2 | 6.66 | 1.6x slower
SAX | 16.74 | 3.9x slower

### DIFF
--- a/cli/parse_args.ts
+++ b/cli/parse_args.ts
@@ -439,6 +439,11 @@ function isNumber(string: string): boolean {
   return NON_WHITESPACE_REGEXP.test(string) && Number.isFinite(Number(string));
 }
 
+function isConstructorOrProto(obj: NestedMapping, key: string): boolean {
+  return (key === "constructor" && typeof obj[key] === "function") ||
+    key === "__proto__";
+}
+
 function setNested(
   object: NestedMapping,
   keys: string[],
@@ -448,7 +453,12 @@ function setNested(
   keys = [...keys];
   const key = keys.pop()!;
 
-  keys.forEach((key) => object = (object[key] ??= {}) as NestedMapping);
+  for (const k of keys) {
+    if (isConstructorOrProto(object, k)) return;
+    object = (object[k] ??= {}) as NestedMapping;
+  }
+
+  if (isConstructorOrProto(object, key)) return;
 
   if (collect) {
     const v = object[key];
--- a/cli/parse_args_test.ts
+++ b/cli/parse_args_test.ts
@@ -1992,3 +1992,23 @@ Deno.test("parseArgs() accepts readonly array input", () => {
   const args: readonly string[] = ["--foo", "bar"];
   const _parsed = parseArgs(args);
 });
+
+/** ---------------------- PROTOTYPE POLLUTION TESTS ---------------------- */
+// These tests verify the fix for CVE-2020-7598 and CVE-2021-44906.
+
+Deno.test("parseArgs() prevents __proto__ pollution", () => {
+  parseArgs(["--__proto__.polluted", "yes"]);
+  // deno-lint-ignore no-explicit-any
+  assertEquals((Object.prototype as any).polluted, undefined);
+});
+
+Deno.test("parseArgs() prevents constructor.prototype pollution", () => {
+  parseArgs(["--constructor.prototype.isAdmin", "true"]);
+  // deno-lint-ignore no-explicit-any
+  assertEquals((Object.prototype as any).isAdmin, undefined);
+});
+
+Deno.test("parseArgs() allows legitimate nested keys with fix applied", () => {
+  const result = parseArgs(["--foo.bar.baz", "value"]);
+  assertEquals(result.foo, { bar: { baz: "value" } });
+});


crowlKats:
could we get some benchmarks comparing to other parsers?

tomas-zijdemans:
> could we get some benchmarks comparing to other parsers?

Yes, that's a good idea. I'll look into it.

tomas-zijdemans:
> could we get some benchmarks comparing to other parsers?

Updated the description now. Let me know if you would like to benchmark against a specific package

import_map.json
tomas-zijdemans, line 1: Sorry, I have no idea why this formatting is happening 😅

timreichen:
Ref: https://github.com/denoland/deno/issues/24995
There was no reply if `DOMParser` or something similar was to be implemented in deno, so I like this PR in general.
However, it might be worth to check with the  deno core team what their current stance is on this before merging anything.

tomas-zijdemans:
> Ref: [denoland/deno#24995](https://github.com/denoland/deno/issues/24995) There was no reply if `DOMParser` or something similar was to be implemented in deno, so I like this PR in general. However, it might be worth to check with the deno core team what their current stance is on this before merging anything.

Thanks, I was not aware of this discussion. Perhaps we could have it as an unstable module for now? Then we can always kick it out, should the core team decide to implement `DOMParser`

tomas-zijdemans:
Updated again to increase streaming performance and get test coverage to 100%

tomas-zijdemans:
More perf work. Will look into using callbacks instead of arrays of objects

tomas-zijdemans:
More perf gains on non-streaming parsing. Updated benchmark results

crowlKats:
@tomas-zijdemans the perf looks really good! what machine are you running this on for reference?

tomas-zijdemans:
@crowlKats: A MacBook with M1 Max

crowlKats:
could we add https://www.w3.org/XML/Test/ ?

tomas-zijdemans:
> could we add https://www.w3.org/XML/Test/ ?

Hmm, yes that would be good. It says it contains over 2000 files though, not small! 😅

tomas-zijdemans:
> could we add https://www.w3.org/XML/Test/ ?

I've looked into it, the test suite has about 2250 tests (several MB), so it's not practical to include in std. We could add a script that download it and creates a conformance report.

The parser is intentionally lenient, per now it handles the valid tests (for XML 1.0), but does not handle the not-well-formed ones very well (these are mostly already documented in the PR).

It seems like the usual way of dealing with the more "theoretical" edge cases is to add a strict mode that enforces XML conformance . WDYT, is that worth it? Per now, this initial PR is still open - so I don't know if there's appetite to include it in std yet.

BlackAsLight:
> The parser is intentionally lenient, per now it handles the valid tests (for XML 1.0), but does not handle the not-well-formed ones very well (these are mostly already documented in the PR).
> 
> It seems like the usual way of dealing with the more "theoretical" edge cases is to add a strict mode that enforces XML conformance . WDYT, is that worth it? Per now, this initial PR is still open - so I don't know if there's appetite to include it in std yet.

When you say the usual way of dealing with it is a strict mode, do you mean that other implementations are offering two modes? a strict and lenient mode?

Personally I'm not a fan of there being different modes, and given the choice I'd prefer a strict parser that throws if it is given an invalid XML file.

At that same time I can also understand people might want a more lenient XML parser because of its relationship with HTML and HTML having a tendency to be forgiving for unexpected characters. No browser wants to be the browser that breaks someone's website after all.

tomas-zijdemans:
> When you say the usual way of dealing with it is a strict mode, do you mean that other implementations are offering two modes? a strict and lenient mode?

Exactly. I thought this was a tradeoff to either be more conformant, or be more practical by swallowing non-conformant errors that are not worth throwing an error for (e.g. double-dash `--` inside comments is not allowed per the spec, but my parser accepts it and continues).

But looking at saxes more carefully, I realize they are very performant while being "strict by default". So I was most likely wrong.

> 
> Personally I'm not a fan of there being different modes, and given the choice I'd prefer a strict parser that throws if it is given an invalid XML file.
> 
> At that same time I can also understand people might want a more lenient XML parser because of its relationship with HTML and HTML having a tendency to be forgiving for unexpected characters. No browser wants to be the browser that breaks someone's website after all.

Good point. Correctness should matter more (this is Deno after all 😄). So I've changed my mind: The code here should be strict by default. It's simpler, more predictable, and matches what users expect from an XML parser in Deno std.

We are running this code in a couple of Production systems, and it's working really well. So I would argue it's ready in it's current form to be accepted as a unstable feature. I'm happy to keep hacking away at it to drive up the conformance score 👍

PR #6944, opened 2026-01-09
feat(async): stabilize isRetriable option for retry

tomas-zijdemans:
Ref: [discord](https://discord.com/channels/684898665143206084/775393009981849600/1458015186710696026)
Will close this [issue](https://github.com/denoland/std/issues/3369)


async/retry.ts
```diff
@@ -112,10 +121,35 @@ export interface RetryOptions {
  * });
  * ```
  *
+ * @example Only retry on specific error types
```
kt3k, line 124: nice example

PR #6949, opened 2026-01-10
perf(yaml): make `DEPRECATED_BOOLEANS_SYNTAX` a `Set`

timreichen:
This PR makes `DEPRECATED_BOOLEANS_SYNTAX` a `Set` which makes lookups faster.


kt3k (approved):
LGTM

PR #6953, opened 2026-01-13
refactor(assert): improve assertArrayIncludes perf and docs

tomas-zijdemans:
Added a fast path using `Array.prototype.includes`. ~20-80% faster for primitive arrays, which should be the more common case(?).

Also Improved JSDoc with examples and `throws` clause

PS: I chose to not import `isPrimitive` from equal.ts, since it's not exported there. We could move that function to a util maybe.

### DIFF
--- a/assert/array_includes.ts
+++ b/assert/array_includes.ts
@@ -31,7 +31,7 @@ function isPrimitive(value: unknown): boolean {
  * assertArrayIncludes([{ a: 1 }, { b: 2 }], [{ a: 1 }]); // Passes
  * ```
  *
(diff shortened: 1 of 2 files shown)


kt3k:
> Added a fast path using Array.prototype.includes. ~20-80% faster for primitive arrays, which should be the more common case(?).

Not sure about what are more common usages, but the perf gain for primitives looks significant. I'm in favor of the change.

> PS: I chose to not import isPrimitive from equal.ts, since it's not exported there. We could move that function to a util maybe.

Inlining sounds fine to me

PR #6957, opened 2026-01-16
feat(http/unstable): add ServerSentEventParseStream for decoding SSE streams

tomas-zijdemans:
Should resolve: https://github.com/denoland/std/issues/6867

### DIFF
--- a/http/unstable_server_sent_event_stream.ts
+++ b/http/unstable_server_sent_event_stream.ts
@@ -20,6 +20,24 @@ export type ServerSentEventParsedMessage =
     id?: string;
   };
 
+/**
+ * Options for {@linkcode ServerSentEventParseStream}.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ */
+export interface ServerSentEventParseStreamOptions {
+  /**
+   * Whether to ignore comment lines (lines starting with `:`).
+   *
+   * Comments are often used as keep-alive signals and may not be needed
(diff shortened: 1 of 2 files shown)


http/unstable_server_sent_event_stream.ts
```diff
@@ -0,0 +1,206 @@
...
+ * Returns true if a field was added to the message.
+ */
+function parseLine(
+  line: string,
+  message: ServerSentEventParsedMessage,
+): boolean {
+  // Lines starting with colon are comments
+  if (line[0] === ":") {
+    const value = line.slice(1);
+    message.comment = message.comment !== undefined
```
mfulton26, line 56: There could be large or many comments in the stream and in some use cases they may never be used by the consumer. What do you think of having an option to ignore comments to avoid unnecessary processing? (Maybe even having it enabled by default.)

tomas-zijdemans: That could be nice. I added an option for filtering it out of the output.

tomas-zijdemans: They will still be parsed/processed, so we're following the spec. It's only removed from the output for convenience.

tomas-zijdemans: They will still be parsed/processed, so we're following the spec. It's only removed from the output for convenience.

kt3k (approved):
LGTM. Nice addition!

PR #6960, opened 2026-01-18
chore(crypto): exclude benchmark file from linting

tomas-zijdemans:
Running `deno task lint` always gave me an annoying warning, I think it should be OK to exclude it from linting, like we do testdata.


kt3k:
The code under `crypto/_benches` seems manually maintained by us, so I think it still makes sense to lint these scripts.

PR #6961, opened 2026-01-19
fix(collections): stricter enforcement on generics

ycmjason:
Please see the following reproduction:

https://www.typescriptlang.org/play/?noUncheckedIndexedAccess=true#code/KYDwDg9gTgLgBAMwK4DsDGMCWEVzJtAawH0IAbAEwB4AVOUGYFCgZzggCMArYDAGjgBpeiEbM2hYAE8ICODQB8ACgBQcdtwBccAErAAhhRxkptBXzVxJUltqgGjKE0IDaAXQsBKbQAUChWgFBBTgAb0s0HBZ4exYkMngAXjCAXzh9Nj8iQKEFAG5LBGg4JUiUaKtpdjlrFk84TDklawbcTi562PiYF2s3OGT23uk3AvV7GCQoXC6EgpSVFVBIWERUDGxcfGy6BiZWDR5+ODKK4T3xOHtDYykS61l5T3dlS3btPRunU0ULdVrtIIvL5-DlBC4UEgALYcYBQNwhcLqU4xYBxBIDVLpTKgmhBCHQ2Hw-KFYqlKLwFqPWr1Rr3KqYNrcTpo7rDKT9QbcdmjSwTKYzVlzFQLJbgaDwZDoLA4PD+Yj6BK0ERiA7tXgwAQooQq-YSaSPRSqdTvXQOW5mP6VGx2c3fVweFTeOBZAJ43JhCIUq5CpJYjIu3FBEnqIpQEraqk1aR1BpNFqMw4s9E9PqYoZ9MY+ybTH0p+aLbVyZLKi4HPSRKDUaJQRkAcwEqEIKAgAHcUAplDBtDR6okQjACtrW1AcHWFeVW3DMQglNsSOQKEpQnAUPoocBtAByLgZHBbgT6OubuAAZgAjHAUgIXFu1xut25PJ4h97K-YMBOWFPw8lZ-Pl1XdcTx3PcUAPdJj20C8rxvO9gMfZ9X3KeB3w1L8fwVDE-zneVFRgQD7xA3cWH3Q8oLPS9rzgW8iMQl8VCAA

----
image
image

As you can see above, the old typing will sometimes fail to narrow the second generic to fit the provided argument (in this example the second generic is inferred as `"name" | "age"` even the provided array is just `['name']`. The new typing fixes the issue and also make the generics more aligned with the argument type.

I have also included an alternative fix for this but I feel like what I am proposing is better since it aligns the generics to the arguments better.

---

Update: my proposed approach is breaking test when the input object is `any`. I have updated to use the alternative solution and added a test to cover this case.

----

Side note: It is surprising to me that `omit` doesn't suffer from the same issue.

### DIFF
--- a/_tools/node_test_runner/tsconfig_for_bun.json
+++ b/_tools/node_test_runner/tsconfig_for_bun.json
@@ -10,7 +10,8 @@
       "@std/internal/os": ["./internal/os.ts"],
       "@std/internal/styles": ["./internal/styles.ts"],
       "@std/path": ["./path/mod.ts"],
-      "@std/semver": ["./semver/mod.ts"]
+      "@std/semver": ["./semver/mod.ts"],
+      "@std/testing/types": ["./testing/types.ts"]
     }
   }
 }


ycmjason:
would need some help to fix test-bun 😄 i am not sure how to make it aware of `@std/testing`

PR #6963, opened 2026-01-20
feat(http/unstable): implement RFC 9651 Structured Field Values

tomas-zijdemans:
Implements [RFC-9651](https://www.rfc-editor.org/rfc/rfc9651) parsing and serialization. RFC 9651 obsoletes RFC 8941 and is fully backwards compatible, adding support for Date and Display String types.

- Single file implementation (like `http/user_agent.ts`)
- Browser compatible
- Matching RFC 9651 terminology as much as possible
- Has all RFC 9651 types, including dates and display strings that extends RFC 8941
- Official HTTPWG conformance test suite included in `testdata/structured_fields/` (JSON files from https://github.com/httpwg/structured-field-tests)

### DIFF
--- a/http/unstable_structured_fields.ts
+++ b/http/unstable_structured_fields.ts
@@ -331,7 +331,7 @@ export function displayString(
  *
  * assertEquals(item(integer(42), [["q", token("fast")]]), {
  *   value: { type: "integer", value: 42 },
- *   parameters: new Map([["q", { type: "token", value: "fast" }]]),
+ *   parameters: new Map([["q", token("fast")]]),
  * });
  * ```
  */


tomas-zijdemans:
Test coverage is not 100%, this is due to defensive programming. In practise I would argue it's 100%.

Please advise if I need to address the patch test coverage.

kt3k:
We tend to avoid these type of prefixed names like `SfItem`, `SfDate`, etc (especially in public API names). How about removing these prefixes? (I personally think it's fine to make some conflicts with builtins such as `String`, `Date`)

tomas-zijdemans:
> We tend to avoid these type of prefixed names like `SfItem`, `SfDate`, etc (especially in public API names). How about removing these prefixes? (I personally think it's fine to make some conflicts with builtins such as `String`, `Date`)

Thanks for the guidance, I was back and forth on this.

tomas-zijdemans:
Ready for re-review

PR #6971, opened 2026-01-25
feat(async/unstable): allow `AbortableOptions` with optional signal in `abortable`

### LINKED ISSUE #6970: Consider allowing `signal` with `abortable` optional and a no-op

When writing functions with abortable operation in a library that can be consumed by end-users, it make it so the developper must add additional boilerplate code between std and user's convenience by checking whether an abort signal is provided or not:

```ts
export function longComputation(signal?: AbortSignal) {
  // ...
  if (signal)
    await abortable(new Promise(), signal)
  else
    await new Promise()
  // ...
}
```

It'd be nice if we could just forward the end user param back to it:
```ts
export function longComputation(signal?: AbortSignal) {
  // ...
  await abortable(new Promise(), signal)
}
```

This would also make it a nice as most function in `@std/async` already support the abort api, so this would make the writing easier when combining `deadline`/`abortable`/`delay` together

Currently a workaround is to use `deadline(promise, Infinity, { signal })` but it seems a bit hacky

```diff
 export function abortable<T>(
   p: Promise<T> | AsyncIterable<T>,
-   signal: AbortSignal,
+   signal?: AbortSignal,
  ): Promise<T> | AsyncIterable<T> {
+  if (!signal) {
+    return p
+  }
   if (p instanceof Promise) {
     return abortablePromise(p, signal);
   } else {
     return abortableAsyncIterable(p, signal);
   }
 }
```

lowlighter:
Close #6970

Make it so the `abortable()` can be called with both of these manners:
```ts
await abortable(promise, signal)
await abortable(promise, { signal })
```
The second one which is introduced, is close to the options object already supported by `deadline`/`delay`, making it easier for libraries developpers to reuse the same object by unifying these differents APIs, and also avoiding the burden of checking whether their promise must be wrapped by abortable or not depending on whether the signal is provided or not

```ts
async function foo(options: { signal?: AbortSignal } = {}) {
  await delay(100, options)
  await abortable(promise, options)
  await deadline(another_promise, options)
}

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,6 +42,10 @@ jobs:
           deno task test
           deno task test:with-unsafe-proto
 
+      - name: Run tools tests
+        run: deno task test:tools
+        if: matrix.deno == 'canary'
+
       - name: Run timezone-dependent tests
         run: |
           TZ=Australia/Sydney deno test datetime
--- a/_tools/lint_plugin.ts
+++ b/_tools/lint_plugin.ts
@@ -1,5 +1,5 @@
 // Copyright 2018-2026 the Deno authors. MIT license.
-// @ts-nocheck Deno.lint namespace does not pass type checking in Deno 1.x
+// Deno.lint namespace does not pass type checking in Deno 1.x
 
 /**
  * Lint plugin that enforces the
@@ -264,7 +264,8 @@ export default {
             const value = argument.value;
             if (typeof value !== "string") return;
 
-            if (value[0] !== value[0].toUpperCase()) {
+            const char = value[0] as string;
+            if (char !== char.toUpperCase()) {
               context.report({
                 node: argument,
                 message: "Error message starts with a lowercase.",
@@ -272,7 +273,7 @@ export default {
                   "Capitalize the error message. See https://docs.deno.com/runtime/contributing/style_guide/#error-messages for more details.",
                 fix(fixer) {
                   const newValue = argument.raw.at(0) +
-                    value[0].toUpperCase() +
+                    char.toUpperCase() +
                     value.slice(1) +
                     argument.raw.at(-1);
                   return fixer.replaceText(argument, newValue);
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

**Comment by kt3k** [comment]
I feel this overload isn't aligned with our style guide for the function signature https://docs.deno.com/runtime/contributing/style_guide/#exported-functions%3A-max-2-args%2C-put-the-rest-into-an-options-object

(I don't think there's a prior example of an argument which is optional in 2 different ways in std)

Also the ergonomic issue this PR tries to solve feels like too niche to me..

What do people think?

**Comment by tomas-zijdemans** [comment]
Agree with @kt3k . I vote for the simpler `RetryOptions` which is consistent with `delay`, `deadline`, `waitFor` and (hopefully) soon `retry` https://github.com/denoland/std/pull/6968

**Comment by kt3k** [comment]
The situation is a bit tricky here. It already has the signatures of:
```ts
export function abortable<T>(p: Promise<T>, signal: AbortSignal): Promise<T>;
export function abortable<T>(p: AsyncIterable<T>, signal: AbortSignal): AsyncGenerator<T>;
```
(The signal is not optional here because `abortable` only makes sense with abort signal)

and because it's already a stable API, we can't remove positional `signal` argument.

lowlighter:
Yeah the main issue is that the abortable API is already stable, so that's why I suggested to support both signatures to be retro compatible

In my opinion the options object should have been the stabilised api, because as @tomas-zijdemans noted, almost all the other apis of the async package use a signal-compatible and are all in line except for abortable. I was personally confused at first myself why it wasn't possible to pass the same option object because of the same reason

But doing so makes it a breaking change so idk :/

tomas-zijdemans:
If there's no appetite for a breaking change + a new major version, it's probably best to leave it as is (Abortable is a bit different since signal is required here, but an optional feature for `retry` and the others)

kt3k:
Maybe this is good for the consistency with other APIs, and improves the ergonomics. And because we can't make a breaking change for now, this would probably be the best we can do for it.

kt3k (approved):
LGTM

PR #6976, opened 2026-01-29
feat(tar/unstable): add symlink support to TarStream

magurotuna:
Adds `TarStreamSymlink` type and `assertValidLinkname` to support writing symbolic links in `TarStream`. Note that `UntarStream` already supports reading symlinks.

### DIFF
--- a/tar/tar_stream.ts
+++ b/tar/tar_stream.ts
@@ -276,9 +276,6 @@ export class TarStream implements TransformStream<TarStreamInput, Uint8Array_> {
         "Cannot add to the tar archive: The size cannot exceed 64 Gibs",
       );
     }
-    if (input.type === "symlink") {
-      assertValidLinkname(input.linkname);
-    }
 
     // name (100) & prefix (155)
     this.#parsePathInto(input.path, buffer);
@@ -310,11 +307,11 @@ export class TarStream implements TransformStream<TarStreamInput, Uint8Array_> {
       ? 50
       : 53;
     // linkname (100)
+    buffer.fill(0, 157, 257);
     if (input.type === "symlink") {
-      buffer.subarray(157, 257).fill(0);
-      this.#encoder.encodeInto(input.linkname, buffer.subarray(157, 257));
-    } else {
-      buffer.fill(0, 157, 257);
+      checkLinkname(
+        this.#encoder.encodeInto(input.linkname, buffer.subarray(157)).written,
+      );
     }
     // magic (6)
     buffer[257] = 117;
@@ -705,15 +702,19 @@ export function assertValidPath(path: string): void {
  * ```
  */
 export function assertValidLinkname(linkname: string): void {
-  if (linkname.length === 0) {
+  const buffer = new Uint8Array(355);
+  checkLinkname(new TextEncoder().encodeInto(linkname, buffer).written);
+}
+
+function checkLinkname(bytes: number): void {
+  if (bytes === 0) {
     throw new TypeError(
       "Cannot add to the tar archive: Invalid Linkname provided",
     );
   }
-  const encoded = new TextEncoder().encode(linkname);
-  if (encoded.length > 100) {
+  if (bytes > 100) {
     throw new TypeError(
-      `Cannot add to the tar archive: Linkname cannot exceed 100 bytes: The linkname length is ${encoded.length}`,
+      `Cannot add to the tar archive: Linkname cannot exceed 100 bytes: The linkname length is ${bytes}`,
     );
   }
 }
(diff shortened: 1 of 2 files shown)


tar/tar_stream.ts
```diff
@@ -121,7 +146,8 @@ const SLASH_CODE_POINT = "/".charCodeAt(0);
  * The ustar file format is used for creating the tar archive.  While this
  * format is compatible with most tar readers, the format has several
  * limitations, including:
- * - Paths must be at most 256 characters.
+ * - Paths must be at most 256 bytes.
```
magurotuna, line 149: Corrected this text because what the validation logic actually checks is the number of bytes, not the number of characters.
The ustar spec mentions that it only supports ASCII characters (in which case the number of characters == the number of bytes), but it seems like the existing validation for paths doesn't check if a path consists only of ASCII or not. Also, it appears that most tar utility tools (like gnu tar) can handle non-ASCII characters without any issue. So, although it's not strictly compliant with the ustar spec, `TarStream` can process non-ASCII characters.

tar/tar_stream.ts
```diff
@@ -246,6 +276,9 @@ export class TarStream implements TransformStream<TarStreamInput, Uint8Array_> {
         "Cannot add to the tar archive: The size cannot exceed 64 Gibs",
       );
     }
+    if (input.type === "symlink") {
+      assertValidLinkname(input.linkname);
+    }
```
BlackAsLight: The validation shouldn't happen here but instead when you go to write it in the buffer. Doing it here like this causes you to needlessly encode the string twice and `TextEncoder` in Deno has very bad performance. Especially since you're not using `encodeInto` in that functions implementation.

tar/tar_stream.ts
```diff
@@ -641,6 +684,40 @@ export function assertValidPath(path: string): void {
...
+      "Cannot add to the tar archive: Invalid Linkname provided",
+    );
+  }
+  const encoded = new TextEncoder().encode(linkname);
+  if (encoded.length > 100) {
+    throw new TypeError(
+      `Cannot add to the tar archive: Linkname cannot exceed 100 bytes: The linkname length is ${encoded.length}`,
+    );
+  }
+}
```
BlackAsLight, line 720: I think this type of structure would be better as it favours `encodeInto` over `encode`. There is a noticeable difference in performance. It also keeps a similar structure to `parsePath`.
```ts
export function assertValidLinkname(linkname: string): void {
  const buffer = new Uint8Array(355);
  validLinkname(new TextEncoder().encodeInto(linkname, buffer).written);
}

function checkLinkname(bytes: number): void {
  if (bytes === 0) {
    throw new TypeError(
      "Cannot add to the tar archive: Invalid Linkname provided",
    );
  } else if (bytes > 100) {
    throw new TypeError(
      `Cannot add to the tar archive: Linkname cannot exceed 100 bytes: The linkname length is ${encoded.length}`,
    );
  }
}
```

tar/tar_stream.ts
```diff
@@ -271,9 +304,18 @@ export class TarStream implements TransformStream<TarStreamInput, Uint8Array_> {
...
+      ? 50
+      : 53;
     // linkname (100)
-    buffer.fill(0, 157, 257);
+    if (input.type === "symlink") {
+      buffer.subarray(157, 257).fill(0);
+      this.#encoder.encodeInto(input.linkname, buffer.subarray(157, 257));
+    } else {
+      buffer.fill(0, 157, 257);
+    }
```
BlackAsLight, line 315: This logic is filling the range 157 to 257 with zero in both branches. It could just happen before the branch entirely.

Secondly, Going back to the above comment, putting the validation here means only encoding the linkname once instead of twice.
```ts
buffer.fill(0, 157, 257);
if (input.type === "symlink") {
  checkLinkname(this.#encoder.encodeInto(input.linkname, buffer.subarray(157)).written);
}
```

tar/tar_stream.ts
BlackAsLight: The ustar spec, the one linked in the file, makes no mention of requiring any specific encoding for name, linkname, and prefix. It specifies that many of the other fields are to be written in ASCII, but does not for them. As long as they are within the byte limit they are compliant.

I have no problem with this change in wording though.

BlackAsLight (changes requested):

tar/tar_stream.ts
magurotuna: That's right, thanks for the correction. I might have been confused with something else.

magurotuna:
@BlackAsLight Thanks so much for your valuable review! I've fixed it in [822300b](https://github.com/denoland/std/pull/6976/commits/822300bc2a7d29608ad4ee3e9000667f515a9817)

BlackAsLight (approved):

kt3k (approved):
LGTM

PR #6977, opened 2026-01-29
fix(cli): strip ANSI codes from indicator for padding in promptSelect

crowlKats:


kt3k (approved):
Makes sense when the indicator has any ansi styling

LGTM

PR #6981, opened 2026-02-03
feat(xml/unstable): add XML parsing and serialization module

tomas-zijdemans:
# New XML parsing and serialization module

> Sorry, I accidentally closed this [PR](https://github.com/denoland/std/pull/6942). This is a revamp that adds more functionality and handles not-wellformed XML better.

The goal of this PR is to implement XML parsing, especially streaming, to handle large XML files without consuming a lot of memory. I chose `saxes` as the library to beat, since it seems to be the best streaming parser out there.

## Functionality - Deno std vs saxes

**Deno Parser Advantages**
Feature | Notes
-- | --
Dual API (DOM + Streaming) | parse() returns a tree, parseXmlStream() for events. Saxes is streaming-only.
Round-trip Support | Built-in stringify() for serialization. Saxes is parse-only.
Zero-allocation Streaming | Callback API with XmlAttributeIterator avoids per-event object creation

**Saxes Advantages**
Feature | Notes
-- | --
XML 1.1 Support | Full XML 1.1 spec. Deno is XML 1.0 only.

 XML 1.1 support is really the only significant feature saxes has that Deno lacks. And XML 1.1 is rarely used in practice.

## Performance streaming - Deno std vs saxes
All tests are run on a MacBook M1 Max. I used a 597MB XML file with product data in the google product format.

Metric | Deno std | Saxes
-- | -- | --
Heap Delta | 4.14 MB | 21.26 MB
Peak Heap | 34.89 MB | 34.77 MB
Duration | 5.00s | 5.56s

The performance in parsing time is similar, with a gain on the memory side (compared to `SAX` and `htmlparser2` we are a lot faster)

## Performance DOM-style/non-streaming - Deno std vs others
DOM-style streaming is not the main focus of this PR, but running benchmarks against a few dozen files (various types of XML, both small and bigger files), this module performs very well while offering significantly stronger XML conformance than these libs:

| Library | Small Files | Large Files |
|---------|-------------|-------------|
| htmlparser2 | 1.2x slower | 1.7x slower |
| SAX | 1.4x slower | 2.8x slower |
| fast-xml-parser | 2.2x slower | 4.1x slower |
| xml2js | 2.9x slower | 5.3x slower |

## Conformance - Deno std vs saxes

No mainstream npm XML library aims to pass the entire W3C XML test suite (especially DTD/external entity/validation cases). Full support is complex and often disabled or omitted in JS parsers for security reasons (XXE/entity-expansion DoS).

I chose the same approach as saxes: To not support custom entity expansion from DTD, as that poses a security risk (XXE vulnerabilities and Billion-laughs attacks)

The below results are based on the [official test suite](https://www.w3.org/XML/Test/), with the following caveats:
- DTD expansions and validation tests are put in a Skip list
- Only XML 1.0 Fifth version

**With Skip List (1294 tests)**

| Parser | Passed | Total | Rate |
|--------|--------|-------|------|
| Saxes | 1290 | 1294 | 99.7% |
| Deno (sync) | 1294 | 1294 | 100.0% |
| Deno (stream) | 1294 | 1294 | 100.0% |

**Without Skip List (1736 tests)**

| Parser | Passed | Total | Rate |
|--------|--------|-------|------|
| Saxes | 1312 | 1736 | 75.6% |
| Deno (sync) | 1442 | 1736 | 83.1% |
| Deno (stream) | 1417 | 1736 | 81.6% |

### DIFF
--- a/import_map.json
+++ b/import_map.json
@@ -45,7 +45,7 @@
     "@std/ulid": "jsr:@std/ulid@^1.0.0",
     "@std/uuid": "jsr:@std/uuid@^1.1.0",
     "@std/webgpu": "jsr:@std/webgpu@^0.224.9",
-    "@std/xml": "jsr:@std/xml@^0.0.1",
+    "@std/xml": "jsr:@std/xml@^0.1.0",
     "@std/yaml": "jsr:@std/yaml@^1.0.11"
   }
 }


xml/parse_stream.ts
```diff
@@ -0,0 +1,127 @@
...
+ *
+ * @param source The XML text stream to parse. Can be a ReadableStream or any
+ *               AsyncIterable that yields string chunks.
+ * @param callbacks Callback functions invoked for each XML event. All callbacks
+ *                  are optional - only provide the ones you need.
+ * @param options Parsing options.
+ * @returns A promise that resolves when parsing is complete.
+ */
+export async function parseXmlStream(
+  source: ReadableStream<string> | AsyncIterable<string>,
```
kt3k: Does it make sense to only accept `AsyncIterable<string>`? The implementation seems only needing `AsyncIterable` methods

tomas-zijdemans: Yes. That makes perfect sense. I changed it in the latest commit. Thanks!

xml/deno.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "@std/xml",
+  "version": "0.0.1",
```
kt3k: Let's start with 0.1.0
```suggestion
  "version": "0.1.0",
```

In 0.x.y versions, we can still distinguish breaking and non-breaking changes by using minor and patch upgrades

PR #6982, opened 2026-02-04
fix(assert): support `Request` objects in `equal()`

iuioiua:
Previously, this would evaluate to `false` because of internal symbols within `Request` objects being unequal:
```ts
import { equal } from "jsr:@std/assert/equal`;
equal(
  new Request("https://example.com"),
  new Request("https://example.com")
); // false
```

This would have knock-on effects due to `equal()` being the underlying function for comparison within [`assertEquals()`](https://jsr.io/@std/assert/doc/~/assertEquals), [`assertSpyCall()`](https://jsr.io/@std/testing/doc/mock/~/assertSpyCall), etc.

This PR is a quick and dirty fix for handling the comparison of said `Request` objects in `equal()`. I'm sure there's a better way to do this, but at least this improves `equal()` a little.

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -110,6 +110,25 @@ export function equal(a: unknown, b: unknown): boolean {
     if (a instanceof Date && b instanceof Date) {
       return Object.is(a.getTime(), b.getTime());
     }
+    if (a instanceof Request && b instanceof Request) {
+      // `Request.body` properties aren't direct compared as doing so would
+      // consume the body streams. In most cases, this is caught anyway by other
+      // properties being different.
+      return a.bodyUsed === b.bodyUsed &&
+        a.cache === b.cache &&
+        a.credentials === b.credentials &&
+        a.destination === b.destination &&
+        equal(a.headers, b.headers) &&
+        a.integrity === b.integrity &&
+        a.keepalive === b.keepalive &&
+        a.method === b.method &&
+        a.mode === b.mode &&
+        a.redirect === b.redirect &&
+        a.referrer === b.referrer &&
+        a.referrerPolicy === b.referrerPolicy &&
+        equal(a.signal, b.signal) &&
+        a.url === b.url;
+    }
     if (a && typeof a === "object" && b && typeof b === "object") {
       if (!prototypesEqual(a, b)) {
         return false;
--- a/assert/equal_test.ts
+++ b/assert/equal_test.ts
@@ -509,3 +509,33 @@ Deno.test("equal() with ArrayBuffer", async (t) => {
     });
   });
 });
+
+Deno.test("equal() with Requests", async (t) => {
+  await t.step("equal", () => {
+    const reqA = new Request("https://example.com/foo", {
+      method: "POST",
+      headers: { "x-foo": "bar" },
+      body: "hello world",
+    });
+    const reqB = new Request("https://example.com/foo", {
+      method: "POST",
+      headers: { "x-foo": "bar" },
+      body: "hello world",
+    });
+    assert(equal(reqA, reqB));
+  });
+
+  await t.step("not equal", () => {
+    const reqA = new Request("https://example.com/foo", {
+      method: "POST",
+      headers: { "x-foo": "bar" },
+      body: "hello world",
+    });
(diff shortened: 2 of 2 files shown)


tomas-zijdemans:
I might be wrong, but I think input validation is the better way of handling this. We could add:
```
if (a instanceof Request) {
  throw new TypeError("Cannot compare Request instances");
}
if (a instanceof Response) {
  throw new TypeError("Cannot compare Response instances");
}
```
This should be added right after the `WeakMap` and `WeakSet` input validation check.

We could consider doing the same for functions like described [here](https://github.com/denoland/std/issues/6878).

lionel-rowe:
> I might be wrong, but I think input validation is the better way of handling this.

This seems reasonable, given that `Request` objects (along with `Response`, `Blob`, `File`, etc.) can't be reliably compared synchrously unless they're empty. And in the case of `Request` and `Response`, even `body: ""` counts as non-empty for purposes of comparison.

With logic as of current PR:

```ts
// true
equal(
    new Request('null://', { method: 'POST', body: '' }),
    new Request('null://', { method: 'POST', body: 'some content' }),
)
```

tomas-zijdemans:
Here's an alternative PR that solves it with input validation instead: https://github.com/denoland/std/pull/6984.

kt3k:
This change feels confusing to me too as bodies are not compared as pointed by Lionel.

There is special note about `Blob` in doc of `assertEquals`. How about adding similar note about comparing `Request`s?

@iuioiua In what situation, did you find the existing behavior inconvenient? Can you share some examples?

iuioiua:
> This change feels confusing to me too as bodies are not compared as pointed by Lionel.

Again, while not true 100% of the time, it's likely that the bodies are the same if other properties are the same. E.g. if the content-length header, the content-type header and the `bodyUsed` properties are equal, it's more likely than not that the bodies are equal in most cases. I agree, simply adding an explanation or example to the related functions in how to deal `Request` objects (clone requests and compare their bodies separately).

> @iuioiua In what situation, did you find the existing behavior inconvenient? Can you share some examples?

```ts
import { assertSpyCall, stub } from "@std/testing/mock";

using fetchStub = stub(globalThis, "fetch");
await fetch(new Request("https://example.com"));

assertSpyCall(fetchStub, 0, {
  args: [new Request("https://example.com")],
});
```

In my specific use case, the package I'm using passes in a `Request` object to the 1st argument of `fetch()`, which is a valid way to use the Fetch API.

lionel-rowe:
@iuioiua `content-length` would in theory be the highest volatility and thus most likely to prevent false positives. It still wouldn't eliminate them entirely, but more importantly, it isn't visible to JS in the first place:

```ts
new Request('foo:', { method: 'POST', body: 'bar' }).headers.get('content-length') // null
```

Checking `body` for reference equality could work, which would at least enable use cases such as your example in which no body is set. Thus for two non-reference-equal requests, `body: null == body: null` but `body: '' != body: ''` (because the latter is an empty `ReadableStream` when read).

iuioiua:
Yes, of course the content-length header can be incorrect. As I already said, this is not full proof. But it beats being completely unable to check that 2 of these objects aren't equal in practical terms.

lionel-rowe:
> Yes, of course the content-length header can be incorrect.

It's less that it's incorrect, more that it's usually `null` when read from JS. I think in practice this will depend on whether it's user-constructed or received via HTTP, and if the latter, whether it's exposed via CORS, and also whether a request uses chunked encoding or HTTP 2, neither of which set it.

> As I already said, this is not full proof. But it beats being completely unable to check that 2 of these objects aren't equal in practical terms.

IMO a function that promises to check for equality should at minimum never return false positives, except maybe in extreme corner cases. False negatives, while best avoided where possible, are more forgivable when those are based on properties that aren't reference equal and that can't be checked for value equality. In some such cases, such as the case of contentful `Request`s, you could even argue that those aren't false negatives at all (because from the standpoint of synchronous code, two different `ReadableStream`s can never be assumed equal).

tomas-zijdemans:
I propose to close this PR, since we went with the documentation based approach: https://github.com/denoland/std/pull/6989

PR #6984, opened 2026-02-04
fix(assert): reject non-comparable types in equal()

tomas-zijdemans:
This is an alternative solution to: https://github.com/denoland/std/pull/6982

Ref https://github.com/denoland/std/issues/6878, I am unsure if we should also add:
```
if (typeof a === "function") {
  throw new TypeError("Cannot compare function values");
}
```

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -142,7 +142,7 @@ export function equal(a: unknown, b: unknown): boolean {
         throw new TypeError("Cannot compare TransformStream instances");
       }
       if (a instanceof Blob) {
-        throw new TypeError("Cannot compare Blob instances");
+        throw new TypeError("Cannot compare Blob or File instances");
       }
       if (a instanceof Request) {
         throw new TypeError("Cannot compare Request instances");


kt3k:
These are currently compared with object identities and I think that is sort of expected behavior for these objects.

This PR disables comparing with object identify for arbitrarily selected classes. That feels confusing to me.

(Is there any popular assertion libraries which behave in this way?)

tomas-zijdemans:
> These are currently compared with object identities and I think that is sort of expected behavior for these objects.
> 
> This PR disables comparing with object identify for arbitrarily selected classes. That feels confusing to me.
> 
> (Is there any popular assertion libraries which behave in this way?)

Looking at Node and others, you are correct. They do not throw but perform property comparison (in various different ways, so that could also be confusing and lead to false positives). So the approach in this PR would be "non-standard".

We could:
- Add JSDoc noting that Blob/Stream/etc comparison is unreliable
- Do property comparison (for Request, look at `url`, `method`, etc. For Blob, look at `size`, etc.
- Keep this "non-standard" approach, but expand the list of Objects to be more complete

kt3k:
> Add JSDoc noting that Blob/Stream/etc comparison is unreliable

I prefer this approach the best.

> Do property comparison (for Request, look at url, method, etc. For Blob, look at size, etc.

I think these checks can't be perfect (as the main part of these objects can't be compared synchronously). I think the users shouldn't expect these comparisons work as expected.

Also because this API is very widely used across the ecosystem through other assertion APIs, mock APIs, etc, the breaking nature of this change is very concerning.

> Keep this "non-standard" approach, but expand the list of Objects to be more complete

This might have been fine when `std/assert` was still unstable, however the breaking nature of the change is very concerning now.

PR #6985, opened 2026-02-05
feat(cli/unstable): add promptConfirm

crowlKats:
adds a promptConfirm function, which is similar to the built-in confirm, however provides more flexibility with letting users decide what the options are, and what the default value is.

### DIFF
--- a/cli/unstable_prompt_confirm.ts
+++ b/cli/unstable_prompt_confirm.ts
@@ -117,7 +117,7 @@ export function promptConfirm<T>(
   const optionsDisplay = values.map((v) => {
     const isDefault = v.key.toLowerCase() === defaultKey.toLowerCase();
     const key = isDefault ? v.key.toUpperCase() : v.key.toLowerCase();
-    return `${key}${v.label ? ` (${v.label})` : ""}$`;
+    return `${key}${v.label ? ` (${v.label})` : ""}`;
   }).join("/");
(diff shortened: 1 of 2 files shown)


tomas-zijdemans:
Oooh, this is nice!

The trailing `$` on line 120, is that intentional? I might be reading this wrong (I see your tests expect them `"Continue? [y$/N$] "`).

Also, maybe `readLineFromStdinSync` could be reused from `prompt_secret.ts`

crowlKats:
regarding the `$` that was a mistake on my end during experimenting that accidentally stayed in

timreichen:
Could you add a PR description on how this differs from the built in [confirm()](https://docs.deno.com/examples/prompts/) for more clarity?

cli/unstable_prompt_confirm.ts
```diff
@@ -0,0 +1,174 @@
...
+   * @default {"n"}
+   */
+  default?: string;
+
+  /** Clear the current line after the user's input. */
+  clear?: boolean;
+}
+
+/** Default values for the confirm prompt. */
+export const YES_NO_VALUES: PromptConfirmValue<boolean>[] = [
```
kt3k, line 39: If this is the default for `values`, then maybe `values` should be a part of the option bag? (and it also can be omitted?)

PR #6989, opened 2026-02-07
docs(assert): add a note about unreliable equality

tomas-zijdemans:
Ref: https://github.com/denoland/std/pull/6982 and https://github.com/denoland/std/pull/6984

### DIFF
--- a/assert/equal.ts
+++ b/assert/equal.ts
@@ -89,9 +89,11 @@ function sameValueZero(a: unknown, b: unknown) {
 /**
  * Deep equality comparison used in assertions.
  *
- * Note: Types whose state is not exposed as enumerable properties, such as
- * `Blob`, `ReadableStream`, `Request`, and `Response`, may produce unreliable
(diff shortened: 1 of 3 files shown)


lionel-rowe:
This might be better to word more generically, as there are other objects that could suffer from this problem, and also (as of current logic, pending decision on proposed logic of https://github.com/denoland/std/pull/6982) only potential false negatives are a known limitation, whereas false positives would be bugs.

I'd word it maybe something like this:

> This function is based on value equality, but for some cases (such as data that can only be read asynchronously or function properties) value equality is not possible to determine. In such cases, reference equality is used instead, which may cause false negatives for objects such as `Blob`s or `Request`s.

Non-enumerable symbol properties might be the specific implementation detail that makes this problem visible for `Blob` etc, but it's not the think that makes the problem intractable, so it's probably (?) not worth mentioning.

tomas-zijdemans:
That is better indeed. Updated now. Thanks @lionel-rowe!

PR #6993, opened 2026-02-08
chore: add some browser compatible declarations

tomas-zijdemans:
These changes came from running `_tools/check_browser_compat.ts` against main


kt3k (approved):
Nice. Thanks!

PR #6994, opened 2026-02-08
refactor(fmt,html,internal,regexp,text,uuid,xml,yaml): add `CONSTANT_CASE` `RegExp` name lint check

timreichen:
This PR adds a lint check that requires a CONSTANT_CASE  `RegExp` declaration to either end in `_REGEXP` or be  `REGEXP`. This ensures a uniform naming convention across all mods.


tomas-zijdemans:
Nice! 👏 
Would it make sense to rename: 
- `RX_DEC_ENTITY_REGEXP` to `DEC_ENTITY_REGEXP`
- `RX_HEX_ENTITY_REGEXP` to `HEX_ENTITY_REGEXP`
- and maybe also `RX_REGEXP_ESCAPE` to `ESCAPE_REGEXP`

path/windows/constants.ts
```diff
@@ -12,4 +12,4 @@ export const SEPARATOR = "\\" as const;
 /**
  * A regular expression that matches one or more path separators.
  */
-export const SEPARATOR_PATTERN = /[\\/]+/;
+export const SEPARATOR_PATTERN_REGEXP = /[\\/]+/;
```
kt3k: Can you keep this one as an exception as this is already a part of stable API.

path/posix/constants.ts
```diff
@@ -12,4 +12,4 @@ export const SEPARATOR = "/" as const;
 /**
  * A regular expression that matches one or more path separators.
  */
-export const SEPARATOR_PATTERN = /\/+/;
+export const SEPARATOR_PATTERN_REGEXP = /\/+/;
```
kt3k: Ditto: Can you keep this one as an exception as this is already a part of stable API.

timreichen:
Should we use `_Pattern` instead of `_REGEXP` to avoid exceptions?

kt3k:
`_REGEXP` sounds clearer to me. I think it's fine to make some exceptions

timreichen:
@kt3k 
We could add a name lint check for `Set`, `Map` and potentially others as well. Should we?

kt3k (approved):
LGTM. Thanks!

kt3k:
> We could add a name lint check for Set, Map and potentially others as well. Should we?

That feels a bit too restrictive and uncommon to me. RegExp feels exceptional in this category (People tends to include signal of RegExp in variable names, but that's not common for other data types in my view)

PR #6995, opened 2026-02-08
fix(cli): handle empty value in parseArgs

tomas-zijdemans:
Here's a sample script to showcase the bug:

```
import { parseArgs } from "jsr:@std/cli/parse-args";

const tests = [
  ["-f="],
  ["-abc="],
  ["--foo="],
  ["--foo=", "bar"],
];

for (const args of tests) {
  console.log(`parseArgs(${JSON.stringify(args)}):`);
  console.log(parseArgs(args));
  console.log();
}

// Bug 1 (short flags):
//
//   -f=
//     Actual:   { _: [], f: true, "=": true }
//     Expected: { _: [], f: "" }
//
//   -abc=
//     Actual:   { _: [], a: true, b: true, c: true, "=": true }
//     Expected: { _: [], a: true, b: true, c: "" }

// Bug 2 (long flags):
//
//   --foo=
//     Actual:   { _: [], "foo=": true }
//     Expected: { _: [], foo: "" }
//
//   --foo= bar
//     Actual:   { _: [], "foo=": "bar" }
//     Expected: { _: [ "bar" ], foo: "" }
```

### DIFF
--- a/cli/parse_args.ts
+++ b/cli/parse_args.ts
@@ -425,7 +425,7 @@ interface NestedMapping {
 }
 
 const FLAG_REGEXP =
-  /^(?:-(?:(?<doubleDash>-)(?<negated>no-)?)?)(?<key>.+?)(?:=(?<value>.+?))?$/s;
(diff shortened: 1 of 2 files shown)


kt3k (approved):
LGTM. Thanks for catching

PR #6997, opened 2026-02-08
refactor(xml): improve tokenizer edge cases and performance

tomas-zijdemans:
- Conformance score still 100%
- About 9% faster streaming performance (now 1.3x faster than Saxes)

### DIFF
--- a/xml/_parse_sync.ts
+++ b/xml/_parse_sync.ts
@@ -461,7 +461,7 @@ export function parseSync(xml: string, options?: ParseOptions): XmlDocument {
         // Check if we're about to see NDATA
         if (
           pos + 4 < len &&
-          input.slice(pos, pos + 5) === "NDATA"
+          input.startsWith("NDATA", pos)
         ) {
           if (isParameterEntity) {
             error("Parameter entities cannot have NDATA declarations");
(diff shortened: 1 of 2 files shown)


xml/_tokenizer.ts
```diff
@@ -373,6 +373,15 @@ export class XmlTokenizer {
       this.#textStartIdx = -1;
       this.#textPartial = "";
       if (content.length > 0) {
+        // XML 1.0 §2.4: The literal string "]]>" is not allowed in text content.
+        // The per-character scan catches this within a single chunk, but "]]>"
+        // can straddle a chunk boundary. Check the accumulated content as a
+        // safety net (mirrors the sync parser's approach).
+        if (content.includes("]") && content.includes("]]>")) {
```
BlackAsLight: If the first one is false then the second one can't be true and if the second one is true then the first one will also be true. 

I don't see how this code checks to see if the ]] is split across boundaries. Is a single ] also not allowed or is it valid code?

tomas-zijdemans: You are right, the fist check there is pointless

kt3k (approved):
LGTM

PR #7000, opened 2026-02-12
fix(testing): stop beforeAll from swallowing test.only error

tomas-zijdemans:
Resolves: https://github.com/denoland/std/issues/6998


kt3k (approved):
LGTM. Thanks!

PR #7001, opened 2026-02-12
feat(yaml): pass depth to sortKeys callback in stringify

tomas-zijdemans:
Resolves: https://github.com/denoland/std/issues/6898

I'm not sure if this is a fix or a feat 🤷


tomas-zijdemans:
Title updated as requested

PR #7002, opened 2026-02-13
feat(async/unstable): allow setting dynamic timeframe for `throttle`

lionel-rowe:
Closes https://github.com/denoland/std/issues/6796.

If `fn` returns a promise, `lastExecution` is now based on the _asynchronous_ end time of the previous _completed_ call. For example with this function:

```ts
async function fn() {
    const start = performance.now()
    while (performance.now() - start < 50) { /* idle  loop */ }
    await new Promise((res) => setTimeout(res, 100))
}
```

If the throttled function is called once at time 0 and we observe  `lastExecution`:

Observation time | Current implementation | New implementation
-|-|-
0 | -Infinity | -Infinity
100 | ~50 | -Infinity
200 | ~50 | ~150

This will also affect how throttling works with asynchronous functions — the current implementation basically doesn't throttle at all for typical async functions, whereas the new one does.

### DIFF
--- a/async/unstable_throttle.ts
+++ b/async/unstable_throttle.ts
@@ -70,6 +70,26 @@ export interface ThrottledFunction<T extends Array<unknown>> {
  * assert(func.lastExecution > 0);
  * ```
  *
+ * @example With dynamic timeframe
+ *
+ * ```ts no-assert
+ * import { throttle } from "@std/async/unstable-throttle";
+ *
+ * function processUserInput(input: string) {
+ *   // Do some expensive computation with user input that changes on each
+ *   // keypress, which takes a variable amount of time depending on the length
+ *   // or complexity of input.
+ * }
+ *
+ * const processUserInputThrottled = throttle(
+ *   processUserInput,
+ *   // Throttle dynamically, waiting twice as long as the previous execution
+ *   // took to complete before starting the next call.
+ *   (n) => n * 2,
+ *   { ensureLastCall: true },
+ * );
+ * ```
+ *
  * @typeParam T The arguments of the provided function.
  * @param fn The function to throttle.
  * @param timeframe The timeframe in milliseconds in which the function should be called at most once.


async/unstable_throttle.ts
```diff
@@ -73,35 +73,55 @@ export interface ThrottledFunction<T extends Array<unknown>> {
...
  * @param timeframe The timeframe in milliseconds in which the function should be called at most once.
+ * If a callback function is supplied, it will be called with the duration of
+ * the previous execution and should return the
+ * next timeframe to use in milliseconds.
  * @param options Additional options.
  * @returns The throttled function.
  */
 // deno-lint-ignore no-explicit-any
 export function throttle<T extends Array<any>>(
   fn: (this: ThrottledFunction<T>, ...args: T) => void,
```
tomas-zijdemans, line 104: The return type of the function should be updated to match the new async handling, something like `void | PromiseLike<void>`

async/unstable_throttle.ts
tomas-zijdemans, line 153: In clear, I guess the new variable `throttlingAsync` now needs to be set to `false`?

async/unstable_throttle.ts
```diff
@@ -117,9 +137,15 @@ export function throttle<T extends Array<any>>(
...
+    throttling: {
+      get: () => Date.now() - lastExecution <= tf || throttlingAsync,
+    },
     lastExecution: { get: () => lastExecution },
   });
 
   return throttled;
 }
+
+function isPromiseLike(obj: unknown): obj is Promise<unknown> {
```
tomas-zijdemans: `PromiseLike<unknown>` would be more correct than `Promise<unknown>`

async/unstable_throttle.ts
lionel-rowe: I thought about that but actually returning a promise doesn't really make sense, as the wrapped function doesn't always run and so the return value should be treated as meaningless and discarded, i.e. should be `void`/`undefined`.

tomas-zijdemans: True, the return value is discarded. So I guess the benefit of adding `PromiseLike` is to signal/document fn?

async/unstable_throttle.ts
lionel-rowe: > `PromiseLike<unknown>` would be more correct than `Promise<unknown>`

Good point, that catches a subtle bug too as we're relying on `finally` method, which isn't part of `PromiseLike`'s type.

async/unstable_throttle.ts
lionel-rowe: > True, the return value is discarded. So I guess the benefit of adding `PromiseLike` is to signal/document fn?

Sorry, I don't follow. `PromiseLike` is not part of the signature, but promise-likes do now have special handling to enable `lastExecution` to track their time-to-resolve rather than time-to-sync-return.

tomas-zijdemans: > Sorry, I don't follow. `PromiseLike` is not part of the signature, but promise-likes do now have special handling to enable `lastExecution` to track their time-to-resolve rather than time-to-sync-return.

It was more of a cosmetic thing. Totally fine to keep it like this ☺️

kt3k:
@lionel-rowe Can you also add an example with comment about when to use the new option?

PR #7003, opened 2026-02-13
feat(expect/unstable): implement toMatchSnapshot()

bartlomieju:
## Summary

- Implements `expect().toMatchSnapshot()` and `expect().toMatchInlineSnapshot()` in the `@std/expect` module, following [Bun/Jest snapshot testing semantics](https://bun.sh/docs/test/snapshots)
- Adds `expect.setState()` / `expect.getState()` for Jest-compatible test context management
- Supports basic snapshots, hint parameters, property matchers with asymmetric matchers, and snapshot update mode (`-- --update`)
- Snapshot files are stored in `__snapshots__/` alongside test files, using the same format as `@std/testing/snapshot`
- All new APIs are marked `@experimental` since `@std/expect` is a stable module
- Uses `@std/path` for cross-platform path handling (Windows support)

## Changes since review

- Replaced custom path utilities (`fileUrlToPath`, `dirname`, `basename`, `join`) with `@std/path` imports — fixes Windows CI failures and addresses maintenance concerns
- Fixed TypeScript compatibility for Deno v1.x (V8 stack trace API type casts)
- Fixed `getIsUpdate()` strict return type for Deno v2.x type checking
- Extracted shared helpers (`parseSnapshotArgs`, `applyPropertyMatchers`, `throwSnapshotMismatch`) between `toMatchSnapshot` and `toMatchInlineSnapshot`
- Exported `escapeStringForJs` directly instead of re-exporting
- Added `@experimental` JSDoc tags to all new public APIs
- Removed "browser compatible" comment from `expect.ts` and `mod.ts` (snapshot support requires Deno APIs)

## Test plan

- [x] Error cases: missing test name, `.not` guard, missing snapshot, property matcher on non-object
- [x] Snapshot matching: exact match, multi-line objects, arrays, multiple snapshots per test, hint-based keys
- [x] Snapshot mismatch: diff output with update instructions
- [x] Round-trip: write snapshots via `SnapshotContext`, read back and assert match (including special characters like backslashes, backticks, dollar signs, newlines)
- [x] Full existing expect test suite passes (167 tests)
- [x] Existing `@std/testing/snapshot` tests unaffected
- [x] Lint passes (including browser compat, circular deps, export names, docs, unstable deps)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/expect/expect.ts
+++ b/expect/expect.ts
@@ -511,7 +511,7 @@ expect.stringMatching = asymmetricMatchers.stringMatching as (
  * Note: expect.hasAssertions only can use in bdd function test suite, such as `test` or `it`.
  *
  * @example
- * ```ts
+ * ```ts ignore
  *
  * import { test } from "@std/testing/bdd";
  * import { expect } from "@std/expect";
@@ -530,7 +530,7 @@ expect.hasAssertions = hasAssertions as () => void;
  * Note: expect.assertions only can use in bdd function test suite, such as `test` or `it`.
  *
  * @example
- * ```ts
+ * ```ts ignore
  *
  * import { test } from "@std/testing/bdd";
  * import { expect } from "@std/expect";
--- a/testing/bdd.ts
+++ b/testing/bdd.ts
@@ -155,7 +155,7 @@
  * the new test suite it creates. The hooks can be created within it or be added to
  * the options argument for describe.
  *
- * ```ts
+ * ```ts ignore
  * import {
  *   assertEquals,
  *   assertStrictEquals,
@@ -240,7 +240,7 @@
  * callback. The gives you the ability to have test grouping without any extra
  * indentation in front of the grouped tests.
  *
- * ```ts
+ * ```ts ignore
  * import {
  *   assertEquals,
  *   assertStrictEquals,
@@ -323,7 +323,7 @@
  * be useful if you'd like to create deep groupings without all the extra
  * indentation in front of each line.
  *
- * ```ts
+ * ```ts ignore
  * import {
  *   assertEquals,
  *   assertStrictEquals,
@@ -548,7 +548,7 @@ export interface it {
  * Registers an individual test case.
  *
  * @example Usage
- * ```ts
+ * ```ts ignore
  * import { describe, it } from "@std/testing/bdd";
  * import { assertEquals } from "@std/assert";
  *
@@ -646,7 +646,7 @@ export function it<T>(...args: ItArgs<T>) {
  * Only execute this test case.
  *
  * @example Usage
- * ```ts
+ * ```ts ignore
  * import { describe, it } from "@std/testing/bdd";
  * import { assertEquals } from "@std/assert";
  *
@@ -675,7 +675,7 @@ it.only = function itOnly<T>(...args: ItArgs<T>): void {
  * Ignore this test case.
  *
  * @example Usage
- * ```ts
+ * ```ts ignore
  * import { describe, it } from "@std/testing/bdd";
  * import { assertEquals } from "@std/assert";
  *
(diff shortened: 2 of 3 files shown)


expect/_snapshot_state.ts
```diff
@@ -0,0 +1,522 @@
...
+export function registerInlineTeardown(): void {
+  if (inlineTeardownRegistered) return;
+  globalThis.addEventListener("unload", () => {
+    applyInlineUpdates();
+  });
+  inlineTeardownRegistered = true;
+}
+
+/** Re-export escapeStringForJs for use by inline snapshot matcher. */
+export { escapeStringForJs };
```
tomas-zijdemans: Maybe easier to just export the function itself?

expect/_matchers.ts
```diff
@@ -1025,3 +1039,269 @@ export function toThrow<E extends Error = Error>(
...
+      throw new AssertionError(
+        `Snapshot does not match:\n${diffMsg}\n` +
+          "To update snapshots, run:\n" +
+          "    deno test --allow-read --allow-write [files]... -- --update\n",
+      );
+    }
+  }
+}
+
+export function toMatchInlineSnapshot(
```
tomas-zijdemans, line 1206: The function seem to duplicate `toMatchSnapshot`, maybe a shared helper?

expect/_matchers.ts
```diff
@@ -1025,3 +1039,269 @@ export function toThrow<E extends Error = Error>(
     context.customMessage,
   );
 }
+
+export function toMatchSnapshot(
+  context: MatcherContext,
+  // deno-lint-ignore no-explicit-any
```
tomas-zijdemans: There are quite many `no-explicit-any`, could some of them be removed perhaps?

expect/_snapshot_state.ts
```diff
@@ -0,0 +1,522 @@
...
+      return null;
+    };
+    Error.captureStackTrace(obj);
+    return obj.stack;
+  } finally {
+    (Error as V8Error).prepareStackTrace = origPrepareStackTrace;
+  }
+}
+
+// --- Path utilities (minimal, avoids @std/path dependency) ---
```
tomas-zijdemans: I disagree with this decision.
- Other modules that needs path handling imports `std/path` (for example `testing/snapshot.ts`).
- This introduces maintenance of duplicate path code. For example, `path/join.ts` handles Windows:
```
export function join(path: string | URL, ...paths: string[]): string {
  return isWindows ? windowsJoin(path, ...paths) : posixJoin(path, ...paths);
}
```
The equivalent here does not:
```
function join(...parts: string[]): string {
  return parts.join("/");
}
```
The other path methods here have similar problems

tomas-zijdemans:
Since `@std/expect` is stable, I suggest introducing this in **unstable** files (e.g. `unstable_to_match_snapshot.ts`) and adding `@experimental` JSDoc tags.

kt3k:
@tomas-zijdemans 

> Since @std/expect is stable, I suggest introducing this in unstable files (e.g. unstable_to_match_snapshot.ts) and adding @experimental JSDoc tags.

Maybe splitting the file is difficult for this particular case as the API is called via method chain like `expected(...).toMatchSnapshot()`. Maybe we can(/should) allow marking API `unstable` without splitting the file here as an exceptional case? What do you think?

tomas-zijdemans:
> Maybe splitting the file is difficult for this particular case as the API is called via method chain like `expected(...).toMatchSnapshot()`. Maybe we can(/should) allow marking API `unstable` without splitting the file here as an exceptional case? What do you think?

Ah, I didn't think about that.. I agree, in-place JSDoc tags is a pragmatic solution.

PR #7004, opened 2026-02-14
refactor(http): remove unstable header and method modules

tomas-zijdemans:
If we want to keep it, then these should be considered for merging instead: https://github.com/denoland/std/pull/6991 and https://github.com/denoland/std/pull/6992


kt3k:
note: these were originally added in https://github.com/denoland/std/pull/4317 and https://github.com/denoland/std/pull/4320

what do people think?

cc @halvardssm

halvardssm:
What is the reasoning for removing them?

I personally find it quite useful to have such headers and other static properties defined in the standard library as it reduces the amount of accidental typos.

Not against keeping them in sync with the IANA registry.

tomas-zijdemans:
> What is the reasoning for removing them?

Please see the discussion here: https://discord.com/channels/684898665143206084/775393009981849600/1472185311118495796

tomas-zijdemans:
Rebased on main

fibibot (commented):
Deletion is clean — no residual `HEADER.`/`METHOD.` references after the `file_server.ts` substitutions, and both entries are gone from `http/deno.json`. The strategic keep-vs-remove question (raised by halvardssm) is the maintainer's call; deferring on that, hence COMMENT not APPROVE.

- nit: `http/mod.ts:29-31` still has the `## Methods` section in the module docblock describing the removed module — should be deleted alongside.

tomas-zijdemans:
Thanks @fibibot. Fixed.

PR #7010, opened 2026-02-17
feat(collections/unstable): add interleave

tomas-zijdemans:
`interleave` merges multiple arrays into a single flat array by round-robin picking one element from each source in turn. 

Given `[1, 2, 3]` and `["a", "b"]`:
- `zip` produces `[[1,"a"], [2,"b"]]` (3 is dropped)
- `interleave` produces `[1, "a", 2, "b", 3]` (nothing dropped)


kt3k (approved):
LGTM

PR #185, opened 2019-02-10
Add benching

chiefbiiko:
API follows `testing`

Would be cool to enhance this by a high resolution timer from Rust.

PR #7012, opened 2026-02-18
feat(crypto/unstable): add AES-GCM encrypt/decrypt helpers

tomas-zijdemans:
High-level wrappers around Web Crypto's AES-GCM that generate a random 96-bit nonce and produce wire format (`nonce (12 bytes) || ciphertext || tag (16 bytes)`). Inspired by Golang's NewGCMWithRandomNonce.

AES-GCM is the only AEAD cipher in the Web Crypto spec, using it correctly requires the caller to: generate a 12-byte random nonce, call `crypto.subtle.encrypt`, concatenate nonce + ciphertext + tag, and reverse the process on decrypt with length validation. These two helpers solve this common use case.

Future streaming complement: [WICG proposal #185](https://github.com/WICG/proposals/issues/185)

### DIFF
--- a/crypto/_types.ts
+++ b/crypto/_types.ts
@@ -1,0 +1,11 @@
+// Copyright 2018-2026 the Deno authors. MIT license.
+
+/**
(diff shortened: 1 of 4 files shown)


kt3k (approved):
LGTM

PR #7014, opened 2026-02-19
feat(async/unstable): add support for AbortSignal in `pooledMap`

tomas-zijdemans:


async/unstable_pool.ts
```diff
@@ -0,0 +1,184 @@
...
+ *
+ * @typeParam T the input type.
+ * @typeParam R the output type.
+ * @param array The input array for mapping.
+ * @param iteratorFn The function to call for every item of the array.
+ * @param options Options including pool limit and abort signal.
+ * @returns The async iterator with the transformed values.
+ * @throws {RangeError} If `poolLimit` is not a positive integer.
+ */
+export function pooledMap<T, R>(
```
kt3k, line 125: hmm this function signature is incompatible with the existing one. we can't stabilize this version as is. can we somehow invent something else which isn't a breaking change? (maybe we have to use overload?)

tomas-zijdemans: Goog point! I pushed a new commit using overloads

kt3k:
Ok. Now this works, but also it is something new to std (No API in std is overloaded in this way as far as I'm aware, and is something not covered by the styled guide).

I'd like to here the opinions from other community members

fibibot (commented):
CI green across all platforms. Code reads correctly: `signal?.throwIfAborted()` at the top of the `for await` loop prevents new items from starting, and `Promise.allSettled(executing)` in the outer catch lets in-flight items finish before the abort error propagates — matches the documented "currently executing items are allowed to finish" contract.

- nit: in the `TransformStream.transform` catch, `controller.error(e as unknown)`; `e` is already typed `unknown` under strict mode, drop the cast.
- nit: `@throws` JSDoc only lists `RangeError`, but the iterator can also reject with `signal.reason` (abort) or `AggregateError` (`iteratorFn` failures) — worth documenting on the options overload.

tomas-zijdemans:
Thanks @fibibot . The two nits are fixed.

PR #7015, opened 2026-02-19
feat(async/unstable): add `pooledMapSettled`

tomas-zijdemans:
Introduces `pooledMapSettled`. Like the stable `pooledMap`, but does not fail fast.

The relationship to `pooledMap` mirrors `Promise.allSettled` vs `Promise.all`

### DIFF
--- a/async/unstable_pool_settled.ts
+++ b/async/unstable_pool_settled.ts
@@ -75,6 +75,7 @@ export interface PooledMapSettledOptions {
  * @param options Configuration for concurrency and cancellation.
  * @returns An async iterator yielding `PromiseSettledResult<R>` for each item,
  *          in the order items were yielded from the input.
+ * @throws {RangeError} If `poolLimit` is not a positive integer.
  */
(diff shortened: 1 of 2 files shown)


async/unstable_pool_settled.ts
```diff
@@ -0,0 +1,189 @@
+// Copyright 2018-2026 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/** Options for {@linkcode pooledMapSettled}. */
+export interface PooledMapSettledOptions {
+  /**
+   * The maximum count of items being processed concurrently. Must be a
+   * positive integer.
+   */
+  poolLimit: number;
```
kt3k, line 10: hmm, this is not optional argument, but our style guide doesn't allow 4 args for public API.. looks like tricky situation..

tomas-zijdemans: Yep. I can't think of a better solution though 🤷

PR #7019, opened 2026-02-20
feat(data-structures/unstable): add `Deque`

tomas-zijdemans:
Deque (double-ended queue) is a standard data structure in many ecosystems (e.g. Rust’s VecDeque, npm’s Denque with ~73M monthly downloads). 

Potential std consumers: `Channel`, `binary-search-tree`, `untar-stream` and more

## Bench

`Array.push()`/`Array.pop()` is extremely optimized in V8. If you only need a stack (push/pop from one end), use a plain Array. 

`Array.unshift()` and `Array.shift()` is slow because the engine must move all elements. Deque's ring buffer makes these very fast.

The most common real-world use case for a Deque is a queue (push back + pop front) . This is where it shines 🚀 

| Scenario              |   n=100 |    n=1,000 |     n=10,000 |      n=100,000 |
| --------------------- | ------: | ---------: | -----------: | -------------: |
| push (back)           | 1.8× ❌ |   2.2× ❌  |     2.1× ❌  |       2.0× ❌  |
| pop (back)            | 3.5× ❌ |   5.8× ❌  |     3.8× ❌  |       4.0× ❌  |
| unshift (front push)  |  ~same  | **5.6× ✅** | **65.6× ✅** |  **977.0× ✅** |
| shift (front pop)     | 1.4× ❌ | **1.7× ✅** |  **2.5× ✅** |  **733.1× ✅** |
| queue (push→shift)    | 2.0× ❌ | **1.5× ✅** |  **1.8× ✅** |  **620.0× ✅** |
| stack (push→pop)      | 3.8× ❌ |   2.9× ❌  |     3.7× ❌  |       3.4× ❌  |
| indexed read          | 1.4× ✅ |   3.2× ❌  |     9.1× ❌  |       5.4× ❌  |

✅ = Deque faster, ❌ = Array faster. Multiplier is relative speed (e.g. "5.6× ✅" means Deque is 5.6× faster).

Results are from a M1 MacBook pro. 30 warmup + 100 measured runs per cell.

### DIFF
--- a/data_structures/unstable_deque.ts
+++ b/data_structures/unstable_deque.ts
@@ -28,9 +28,9 @@ export type ReadonlyDeque<T> = Pick<
   | typeof Symbol.toStringTag
 >;
 
+/** Round up to the smallest power of two >= n (at least {@linkcode MIN_CAPACITY}). */
 function nextPowerOfTwo(n: number): number {
   if (n <= MIN_CAPACITY) return MIN_CAPACITY;
-  // Bit trick: fill all bits below the highest set bit, then add 1
   n--;
   n |= n >>> 1;
   n |= n >>> 2;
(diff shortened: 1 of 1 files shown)


kt3k:
JS array has `pop`, `push`, `shift`, and `unshift`, and I guess the engines do the similar optimization internally for those methods. How is the performance of this `Deque` compared to a native array?

tomas-zijdemans:
> JS array has `pop`, `push`, `shift`, and `unshift`, and I guess the engines do the similar optimization internally for those methods. How is the performance of this `Deque` compared to a native array?

Added benchmarks in the description

PR #7022, opened 2026-02-23
refactor(async): improve `debounce` readability, docs, and input validation

tomas-zijdemans:
- Renamed local flush variable to `pendingFlush` to avoid shadowing `debounced.flush`
- Replace `typeof timeout === "number"` with `timeout !== null` for readability
- Added more tests

### DIFF
--- a/async/debounce_test.ts
+++ b/async/debounce_test.ts
@@ -1,5 +1,5 @@
 // Copyright 2018-2026 the Deno authors. MIT license.
-import { assertEquals, assertStrictEquals } from "@std/assert";
+import { assertEquals, assertStrictEquals, assertThrows } from "@std/assert";
 import { debounce, type DebouncedFunction } from "./debounce.ts";
 import { delay } from "./delay.ts";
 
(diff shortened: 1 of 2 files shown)


kt3k:
> Replace typeof timeout === "number" with timeout !== null for readability

This is probably also good for Node.js compatibility in the future.

async/debounce_test.ts
```diff
@@ -79,3 +79,54 @@ Deno.test("debounce() handles number and string types", async () => {
...
+  for (const wait of [0.5, -1, NaN, Infinity]) {
+    assertThrows(
+      () => debounce(fn, wait),
+      RangeError,
+      "'wait' must be a positive integer",
+    );
+  }
+});
+
+Deno.test("debounce() re-invocation after flush re-schedules", async () => {
```
kt3k, line 120: Deno's timer can be very unreliable in CI sometimes, and this could be flaky, but let's try

kt3k (approved):
LGTM

async/debounce_test.ts
tomas-zijdemans: Good catch: https://github.com/denoland/std/pull/7034

PR #7024, opened 2026-02-23
fix(http): handle HEAD requests in serveDir

### LINKED ISSUE #6750: Allow HEAD mathod in file-server

wilkmaciej:
## Summary
- `serveDir()` now accepts HEAD requests alongside GET, resolving the inconsistency where `serveFile()` already supported HEAD but `serveDir()` returned 405
- `serveDirIndex()` returns a bodiless response with correct `Content-Length` for HEAD requests on directory listings

Closes #6750
Supersedes #6752

### DIFF
--- a/http/file_server.ts
+++ b/http/file_server.ts
@@ -407,8 +407,10 @@ async function serveDirIndex(
 
   const headers = createBaseHeaders();
   headers.set(HEADER.ContentType, "text/html; charset=UTF-8");
-  const pageSize = new TextEncoder().encode(page).byteLength;
-  headers.set(HEADER.ContentLength, `${pageSize}`);
+  if (req.method === METHOD.Head) {
+    const pageSize = new TextEncoder().encode(page).byteLength;
+    headers.set(HEADER.ContentLength, String(pageSize));
+  }
 
   const status = STATUS_CODE.OK;
   return new Response(req.method === METHOD.Head ? null : page, {


kt3k (dismissed):
LGTM. Nice! Thanks

http/file_server.ts
```diff
@@ -406,9 +407,11 @@ async function serveDirIndex(
 
   const headers = createBaseHeaders();
   headers.set(HEADER.ContentType, "text/html; charset=UTF-8");
+  const pageSize = new TextEncoder().encode(page).byteLength;
+  headers.set(HEADER.ContentLength, `${pageSize}`);
```
kt3k: This seems unnecessary overhead for GET method call. Can this be performed only for HEAD method?

kt3k:
Thanks for the PR. Left one comment

http/file_server.ts
wilkmaciej: i was actually wodering it it should be calculated as from what i know some browsers may use ContentLength for get request to show progress of page load

kt3k: Content-length is still set for GET requests, which is automatically calculated by `Response` object. So this `new TextEncoder().encode(page).byteLength` is duplicated calculation of byte length for GET.

wilkmaciej: oh ok

PR #7027, opened 2026-03-01
feat: `std/concurrency`

tomas-zijdemans:
`@std/async` currently holds both lightweight async helpers and heavier concurrency primitives. As the latter group grows, the module name stops describing its contents well.

The concurrency items are categorically different from the rest of the module, they should get their own package. The concurrency module is about coordinating shared access to resources. This mirrors the `collections` and `data_structures` split that was done. 
 

The split is clean: `Semaphore`, `Channel` (in the works), `CircuitBreaker`, and `Lazy` have no internal dependencies on other `@std/async `files.  These are low-level synchronisation primitives. All four are still unstable, so there should be no deprecation overhead.

"Concurrency primitives" is a well-understood category, and it gives the new package a natural growth path (e.g. `mutex`, `barrier`, `read-write lock`...) without further bloating `@std/async`.

I renamed the files, since the new module itself is unstable.

### DIFF
--- a/.github/workflows/title.yml
+++ b/.github/workflows/title.yml
@@ -41,6 +41,7 @@ jobs:
             cache(/unstable)?
             cbor(/unstable)?
             cli(/unstable)?
+            concurrency(/unstable)?
             collections(/unstable)?
             crypto(/unstable)?
             csv(/unstable)?


kt3k:
Sounds a reasonable move to me.

What do people think?

kt3k:
Note: The core team seems generally against this move

tomas-zijdemans:
No worries! ☺️ There are other ways to split this, as `async` keeps growing. Are you open to other suggestions?
I'll close this PR soon.

tomas-zijdemans:
The core team disagrees, closing.

PR #7028, opened 2026-03-02
feat(data-structures/unstable): add RollingCounter

tomas-zijdemans:
This data structure tracks counts across a fixed number of segments using a ring buffer. The caller controls when the window advances by calling rotate(), making it useful for rate limiting (I'm working on a new module for that), `circuit breaker` ++

### DIFF
--- a/data_structures/deno.json
+++ b/data_structures/deno.json
@@ -10,6 +10,7 @@
     "./comparators": "./comparators.ts",
     "./red-black-tree": "./red_black_tree.ts",
     "./unstable-2d-array": "./unstable_2d_array.ts",
-    "./unstable-rolling-counter": "./unstable_rolling_counter.ts"
+    "./unstable-rolling-counter": "./unstable_rolling_counter.ts",
+    "./unstable-deque": "./unstable_deque.ts"
   }
 }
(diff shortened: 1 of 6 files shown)


kt3k:
Is it common to call this data structure Sliding Window Counter? Is there any prior example library?

I research a bit about rate limiting algorithms, but this data structure seems more similar to one for sliding window log to me (sliding window counter seems using only 2 fixed windows and combining them with weighted interpolation)

tomas-zijdemans:
> Is it common to call this data structure Sliding Window Counter? Is there any prior example library?
> 
> I research a bit about rate limiting algorithms, but this data structure seems more similar to one for sliding window log to me (sliding window counter seems using only 2 fixed windows and combining them with weighted interpolation)

I agree. I was looking at a [Go package](https://pkg.go.dev/go.lepak.sg/playground/slidingwindow). Renaming to `RollingCounter`. It's a ring of counters that rotates.

So usage could look like:
```
// Circuit breaker
const failures = new RollingCounter(10);
const requests = new RollingCounter(10);


// Rate limiter
const permits = new RollingCounter(segmentsPerWindow);
```

kt3k (approved):
LGTM

PR #7029, opened 2026-03-03
ci: update token permission, only give `contents: read`

tomas-zijdemans:


.github/workflows/ci.yml
```diff
@@ -14,6 +14,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
     permissions:
+      contents: write # Required for deployctl upload on push
```
kt3k: deployctl seems only needing `read`? https://github.com/denoland/deployctl?tab=readme-ov-file#action-example

Removing this line
```suggestion
```

kt3k: I guess `contents: write` was used for some past workflows and is not used anymore

PR #7031, opened 2026-03-04
feat(regexp/unstable): add `replaceAllAsync`

lionel-rowe:
Closes https://github.com/denoland/std/issues/7030

### DIFF
--- a/regexp/deno.json
+++ b/regexp/deno.json
@@ -3,6 +3,7 @@
   "version": "1.0.1",
   "exports": {
     ".": "./mod.ts",
-    "./escape": "./escape.ts"
+    "./escape": "./escape.ts",
+    "./unstable-replace-all-async": "./unstable_replace_all_async.ts"
   }
 }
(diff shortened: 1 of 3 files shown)


kt3k:
I'm a bit unsure whether this belongs to `std/regexp` as the original `replaceAll` is an string method. How about `std/text`?

lionel-rowe:
> I'm a bit unsure whether this belongs to `std/regexp` as the original `replaceAll` is an string method. How about `std/text`?

@kt3k Could argue the case either way, personally I'd be more likely to look under `regexp` but maybe that's just me. Also, it's kinda a method on `RegExp` as well, just that it lives under `Symbol.replace` instead:

```ts
assertEquals(
    'abcd'.replaceAll(/[b-c]/g, ''),
    /[b-c]/g[Symbol.replace]('abcd', ''),
)
```

kt3k:
Ok. `std/regexp` now makes sense to me

kt3k (approved):
LGTM

PR #7033, opened 2026-03-05
feat(http/unstable): add RFC 9457 Problem Details response helpers

tomas-zijdemans:
Adds `@std/http/unstable-problem-details`, a module for working with [RFC 9457 – Problem Details for HTTP APIs](https://www.rfc-editor.org/rfc/rfc9457.html).

RFC 9457 defines a standard JSON (and XML) format for machine-readable error responses from HTTP APIs. It is widely adopted across REST ecosystems and replaces the earlier RFC 7807. Having first-class support in `@std/http` gives Deno users a consistent, spec-compliant way to produce and consume structured error payloads without pulling in third-party libraries.


http/unstable_problem_details.ts
```diff
@@ -0,0 +1,367 @@
...
+  headers.set("Content-Type", PROBLEM_JSON_MEDIA_TYPE);
+  return new Response(body, { status, headers });
+}
+
+/** Per RFC 9457 §3.1: ignore standard members whose value type does not match. */
+function normalizeParsedProblemDetails(
+  raw: Record<string, unknown>,
+): Record<string, unknown> {
+  if (typeof raw !== "object" || raw === null || Array.isArray(raw)) {
+    throw new TypeError("Cannot parse Problem Details: expected a JSON object");
```
kt3k: How about including the given type in error message?

tomas-zijdemans: Yes, that's a good catch that better reflects the style guide 👍

tomas-zijdemans:
> Good RFC 9457 implementation — clean generics, solid type validation per §3.1, and thorough tests. A few things:
> 
> 1. **Missing PR description.** Should explain what RFC 9457 is and why it belongs in `@std/http`.

I attempted a new PR title and added PR desciption
> 2. **`parseProblemDetails(response)` consumes the body.** Calling `response.json()` means the body can't be re-read. Worth documenting this explicitly, since callers may not expect it.

Added a comment explaining this in the latest commit.
> 3. **`isProblemDetailsResponse()` only handles JSON.** RFC 9457 also defines `application/problem+xml`. Fine to skip for now, but worth noting as a known limitation in the JSDoc.

Added a comment explaining this in the latest commit.
> 4. **`normalizeParsedProblemDetails` uses `for...in` with `Object.hasOwn` guard.** `Object.keys()` or `Object.entries()` would be more idiomatic and avoid the need for the hasOwn check.

That would work equally well. The `for...in` with `switch` reads as a single linear pass where each key is validated in place. `Object.keys()` would also work and skip inherited properties the same way, but wouldn't meaningfully improve readability or correctness imho. But yes, either is fine so let me know if this is a must-fix.

PR #7036, opened 2026-03-09
feat(async/unstable): use failure-rate threshold in circuit breaker

tomas-zijdemans:
The circuit breaker now trips based on the percentage of failed requests instead of a fixed count. This matters because a hard count like "open after 5 failures" behaves very differently at 10 req/s vs 10,000 req/s. Percentage-based tripping scales naturally with traffic.

Key changes:

- Sliding window backed by `RollingCounter` segments for efficient time-based eviction
- New `minimumThroughput` option prevents tripping on tiny sample sizes during low traffic
- `failureThreshold: 5` replaced by `failureRateThreshold: 0.5` (ratio of failures to total requests)
- Fix an inconsistency so `state` returns the actual state


kt3k (approved):
LGTM

PR #7037, opened 2026-03-09
feat(http/unstable): add RFC 9530 digest fields

tomas-zijdemans:
Adds an unstable `@std/http/unstable-digest-fields` module for creating and verifying RFC 9530 `Content-Digest` and `Repr-Digest` field values.

RFC 9530, an IETF standards-track RFC from Feb 2024, replaces the legacy RFC 3230 `Digest` field with Structured Fields based digest headers for HTTP content and representation integrity. These helpers support creating digest values from strings, bytes, or streams, and verifying `Request` / `Response` bodies without consuming the original body.

This is also useful with HTTP Message Signatures (RFC 9421), where a signed digest field can bind the signature to the message body. The implementation is small and browser-compatible, reusing existing `@std` primitives: `timing-safe-equal`, `unstable-structured-fields`, and `to-bytes`.


tomas-zijdemans:
Updated to use the stable toBytes

tomas-zijdemans:
Parsing has been hardened and docs improved. Ready for review @bartlomieju

fibibot (approved):
CI green. The new unstable module is wired only via `http/deno.json` `./unstable-digest-fields`, not re-exported from `mod.ts`. Symbols have `@experimental`, JSDoc, and per-symbol examples with `@std/assert` assertions.

`parseDigestHeaderValue` runs before any body access — the streaming-body tests at lines 380-413 confirm malformed headers don't drain the request body. Subtle but correct.

- nit: `verifyDigestHeader` reads the whole body via `message.clone().bytes()`. The doc already warns about content-encoded responses; worth also flagging the in-memory cost for large bodies.
- nit: the module-level "Creating a Content-Digest header" example (lines 12-23) builds a `Request` but doesn't assert anything; per-symbol examples below do.

bartlomieju:
One thing worth addressing before this stabilizes: `verifyDigestHeader` reads the entire body into memory via `message.clone().bytes()`. For the primary server-side use case — verifying the `Content-Digest` of an incoming untrusted `Request` — that's a DoS surface, not just a memory-cost concern. An attacker can send a well-formed, correctly-lengthed digest header plus an arbitrarily large body and force full buffering before the comparison runs.

The header parsing and length validation already happen before any body access (nice — the streaming test confirms it), so the fast-fail path is solid. The gap is only on the "valid header, hostile body" case.

Since the module is unstable I don't think this needs to block landing, but I'd like it handled one of two ways:

- Near-term: a doc caveat on `verifyContentDigest` / `verifyReprDigest` stating that callers verifying untrusted requests should bound the body size upstream (e.g. enforce a `Content-Length` limit or wrap the body in a size-capped stream) before calling these helpers.
- Later (optional): an opt-in `maxBodySize` option that rejects with a `TypeError` once the read exceeds the cap, so the protection lives in the helper instead of every call site.

A doc note is probably enough for now — just want the expectation stated explicitly rather than left implicit.

tomas-zijdemans:
Thanks @bartlomieju. 

Let's go for the proper fix:

Add an opt-in `VerifyDigestOptions.maxBodySize`:
- The body is read **incrementally** and verification rejects with a `TypeError` 
  as soon as the running total exceeds the cap — the payload is never fully buffered.
- `maxBodySize` is validated as a non-negative integer (guards `NaN`, which
  would otherwise silently disable the check), before any body is read.
- Default behavior is unchanged (entire body read), so this is backward compatible.
- The docs on both verify helpers now carry an explicit warning recommending
  callers cap the read and/or bound body size upstream (e.g. `Content-Length`)
  when verifying untrusted requests.

PR #7038, opened 2026-03-09
feat(semver): add `minVersion` and `coerce` functions

sxzz:
## Summary

- Add `minVersion(range)` which returns the lowest SemVer that satisfies a given range, ported from node-semver's `minVersion`
- Add `coerce(version, options?)` which extracts a valid SemVer from a loosely-formatted string or number, ported from node-semver's `coerce`. Supports `rtl` and `includePrerelease` options

## Test plan

- [x] `deno test semver/min_version_test.ts` — 26 test cases pass
- [x] `deno test semver/coerce_test.ts` — 20 test cases pass
- [x] `deno check` passes for both files


semver/coerce.ts
```diff
@@ -0,0 +1,121 @@
...
+    ) {
+      if (
+        !match ||
+        next.index + next[0].length !== match.index + match[0].length
+      ) {
+        match = next;
+      }
+      coerceRtlRegex.lastIndex = next.index + next[1]!.length + next[2]!.length;
+    }
+    coerceRtlRegex.lastIndex = -1;
```
tomas-zijdemans: Should it be `= 0` instead of `-1` ?

semver/unstable_min_version.ts
```diff
@@ -0,0 +1,91 @@
...
+          }
+        // falls through
+        case "=":
+        case undefined:
+        case ">=":
+          if (!setMin || greaterThan(compver, setMin)) {
+            setMin = compver;
+          }
+          break;
+        case "<":
```
tomas-zijdemans, line 78: Suggest to add `!=` here as as an ignored case as well, since the type allows it.

semver/coerce.ts
sxzz: This line is ported from node-semver. Now fixed.

tomas-zijdemans:
Looks good 😊

lunadogbot (changes requested):
Two blockers:

1. `@std/semver` is v1.0.8, so per `.github/CONTRIBUTING.md` "New features in stable packages (version >= 1.0.0)" both functions need to land as `semver/unstable_coerce.ts` / `semver/unstable_min_version.ts`, must not be re-exported from `mod.ts`, and need `@experimental **UNSTABLE**: New API, yet to be vetted.` on the JSDoc. The two `mod.ts` re-exports and the `./coerce` / `./min-version` entries in `semver/deno.json` need to go with that move.

2. `lint (ubuntu-latest)` is red on the current head while the matching macOS/Windows lint runs are green. Probably a stale or platform-specific lint rule — please rerun and address whatever it surfaces.

- nit: `min_version.ts:88` — `throw new Error(\`Unexpected operator: ${comparator.operator}\`)` should quote the value per the error-message convention: `` `Unexpected operator: "${comparator.operator}"` ``.

sxzz:
Addressed the review in 59fde1f5:

- Moved `coerce` and `minVersion` to `semver/unstable_coerce.ts` and `semver/unstable_min_version.ts`.
- Removed the stable `mod.ts` re-exports and the stable `./coerce` / `./min-version` package exports.
- Added `./unstable-coerce` and `./unstable-min-version` package exports with unstable JSDoc tags.
- Updated the unexpected operator error message to quote the operator value.

semver/unstable_min_version.ts
```diff
@@ -0,0 +1,95 @@
...
+ *
+ * @param range The range to find the minimum version for.
+ * @returns The minimum version that satisfies the range, or `undefined`.
+ */
+export function minVersion(range: Range): SemVer | undefined {
+  const ZERO: SemVer = {
+    major: 0,
+    minor: 0,
+    patch: 0,
+  };
```
bartlomieju, line 37: Return-shape inconsistency: this `ZERO` omits `prerelease`/`build`, so `minVersion(parseRange("*"))` returns `{major:0,minor:0,patch:0}` while every other path (including `ZERO_PRE` just below, and anything derived from a comparator) returns the full shape with `prerelease: []` / `build: []`. Downstream code doing e.g. `result.prerelease.length` would break only for `*` / `>=0.0.0`-style ranges.

Suggest adding both fields here for consistency with `parse()` output:

```suggestion
  const ZERO: SemVer = {
    major: 0,
    minor: 0,
    patch: 0,
    prerelease: [],
    build: [],
  };
```

Note the tests currently assert the bare shape (`{ major: 0, minor: 0, patch: 0 }`) for `*` and `>=0.0.0`, so they'd need updating alongside this.

semver/unstable_coerce.ts
```diff
@@ -0,0 +1,127 @@
...
+import type { SemVer } from "./types.ts";
+import { parse } from "./parse.ts";
+
+const MAX_SAFE_COMPONENT_LENGTH = 16;
+
+const PRERELEASE_IDENTIFIER = "(?:0|[1-9]\\d*|\\d*[a-zA-Z-][a-zA-Z0-9-]*)";
+const PRERELEASE =
+  `(?:-(?<prerelease>${PRERELEASE_IDENTIFIER}(?:\\.${PRERELEASE_IDENTIFIER})*))`;
+const BUILD_IDENTIFIER = "[0-9A-Za-z-]+";
+const BUILD =
```
bartlomieju, line 12: Minor/maintainability: these identifiers (`PRERELEASE_IDENTIFIER`, `PRERELEASE`, `BUILD_IDENTIFIER`, `BUILD`) duplicate the ones already defined in `_shared.ts` (they're just not exported there). Consider exporting and reusing them from `_shared.ts` to avoid the two definitions drifting apart. Not a blocker.

semver/unstable_coerce.ts
```diff
@@ -0,0 +1,127 @@
...
+ */
+export function coerce(
+  version: string | number,
+  options?: CoerceOptions,
+): SemVer | undefined {
+  if (typeof version === "number") {
+    version = String(version);
+  }
+
+  if (typeof version !== "string") {
```
bartlomieju, line 80: Nit: given the declared `string | number` type, after the `number` branch above converts to a string this guard is unreachable. Harmless as a defensive guard for untyped JS callers, but could be dropped.

bartlomieju (commented):
Took this for a spin locally (tests + doc tests pass, all CI green). Looks close to mergeable — one consistency issue worth fixing, plus two minor nits inline.

The main thing: `minVersion`'s fast-path `ZERO` returns a different object shape than every other return path (it omits `prerelease`/`build`), which I'd fix before merge. Details inline.

PR #7039, opened 2026-03-10
feat(http/unstable): add RFC 9421 message signatures

tomas-zijdemans:
Add HTTP Message Signatures ([RFC 9421](https://datatracker.ietf.org/doc/html/rfc9421)) for signing and verifying HTTP requests and responses. Builds on the structured fields module (RFC 8941) already in std.

Coverage is as good as I can get it. There are a few lines of unreachable code.

### DIFF
--- a/http/unstable_message_signatures.ts
+++ b/http/unstable_message_signatures.ts
@@ -192,7 +192,12 @@ export interface SignatureParams {
   nonce?: string;
   /** Application-specific tag. */
   tag?: string;
-  /** Signature label, defaults to `"sig"`. Must be a valid sf-key (lowercase alphanumeric, `_`, `-`, `.`, `*`). */
+  /**
+   * Signature label, defaults to `"sig"`. Must be a valid sf-key (lowercase
+   * alphanumeric, `_`, `-`, `.`, `*`). Must also be unique among existing
+   * signatures on the target message per RFC 9421 section 4; a `TypeError`
+   * is thrown on collision.
+   */
   label?: string;
 }
 
@@ -825,6 +830,7 @@ export async function signMessage<T extends Request | Response>(
   const label = params.label ?? "sig";
 
   validateSignParams(params);
+  assertUniqueLabel(message.headers, label);
 
   const algorithm = params.algorithm ?? inferAlgorithm(key);
 
@@ -861,6 +867,32 @@ function appendHeader(headers: Headers, name: string, value: string): void {
     headers.set(name, `${existing}, ${value}`);
   } else {
     headers.set(name, value);
+  }
+}
+
+// Per RFC 9421 §4, signature labels MUST be unique within an HTTP message
+// across both the Signature-Input and Signature dictionaries. Appending a
+// signature with a colliding label would produce duplicate dictionary keys
+// that silently corrupt prior signatures on parse (last value wins per
+// RFC 8941).
+function assertUniqueLabel(headers: Headers, label: string): void {
+  for (const name of ["Signature-Input", "Signature"] as const) {
+    const value = headers.get(name);
+    if (value === null) continue;
+    let dict: Dictionary;
+    try {
+      dict = parseDictionary(value);
+    } catch (cause) {
+      throw new TypeError(
+        `Cannot parse existing "${name}" header on message to check label uniqueness`,
+        { cause },
+      );
+    }
+    if (dict.has(label)) {
+      throw new TypeError(
+        `Signature label "${label}" is already present in the "${name}" header on the message; choose a unique label (RFC 9421 section 4)`,
+      );
+    }
   }
 }
 
--- a/http/unstable_message_signatures.ts
+++ b/http/unstable_message_signatures.ts
@@ -1,0 +1,1059 @@
+// Copyright 2018-2026 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/**
+ * Utilities for creating and verifying
+ * {@link https://www.rfc-editor.org/rfc/rfc9421 | RFC 9421} HTTP Message Signatures.
+ *
+ * HTTP Message Signatures provide end-to-end integrity and authenticity for
+ * components of an HTTP message by using detached digital signatures or MACs.
+ * The `Signature-Input` and `Signature` headers are serialized as Structured
+ * Fields Dictionaries ({@link https://www.rfc-editor.org/rfc/rfc9651 | RFC 9651}).
+ *
+ * @example Signing a request
+ * ```ts
+ * import { signMessage } from "@std/http/unstable-message-signatures";
+ *
+ * const key = await crypto.subtle.generateKey("Ed25519", true, ["sign", "verify"]);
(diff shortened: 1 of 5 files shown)


http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1100 @@
...
+  // the Date header). Avoid using ;bs on such fields.
+  const values = headerValue.split(", ");
+  const items: Item[] = values.map((v) => {
+    const bytes = UTF8_ENCODER.encode(v.trim());
+    return sfItem(binary(bytes));
+  });
+  return serializeList(items);
+}
+
+// =============================================================================
```
bartlomieju, line 579: The comment here is good, but this limitation is pretty significant — the Fetch API `Headers.get()` joins multiple values with `", "`, so `;bs` will silently produce wrong results for headers whose single value contains a literal `", "` (e.g. `Date`).

Worth documenting this more prominently in the module-level JSDoc so users know which headers are unsafe with `;bs`.

http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1100 @@
...
+  if (name === "HMAC") return "hmac-sha256";
+  if (name === "RSA-PSS") return "rsa-pss-sha512";
+  if (name === "RSASSA-PKCS1-v1_5") return "rsa-v1_5-sha256";
+  if (name === "ECDSA") {
+    const curve = (alg as EcKeyAlgorithm).namedCurve;
+    if (curve === "P-384") return "ecdsa-p384-sha384";
+    return "ecdsa-p256-sha256";
+  }
+  throw new TypeError(`Cannot infer signature algorithm from key: "${name}"`);
+}
```
bartlomieju, line 330: This silently falls back to `ecdsa-p256-sha256` for any unrecognized ECDSA curve (e.g. P-521). That would produce incorrect signatures. Should throw instead:

```ts suggestion
    if (curve === "P-256") return "ecdsa-p256-sha256";
    if (curve === "P-384") return "ecdsa-p384-sha384";
    throw new TypeError(`Unsupported ECDSA curve: "${curve}"`);
```

http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1100 @@
...
+      return url.protocol.slice(0, -1);
+    case "@request-target":
+      return url.pathname + url.search;
+    case "@path":
+      return url.pathname || "/";
+    case "@query":
+      return url.search || "?";
+    case "@query-param": {
+      if (params.name === undefined) {
+        throw new TypeError(
```
bartlomieju, line 454: Returning `"?"` for absent query strings — confirmed this matches RFC 9421 section 2.2.7. 👍

http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1100 @@
...
+      }
+      if (values.length === 0) {
+        throw new TypeError(
+          `Query parameter "${params.name}" not found in request URL`,
+        );
+      }
+      if (values.length > 1) {
+        throw new TypeError(
+          `Query parameter "${params.name}" occurs multiple times`,
+        );
```
bartlomieju, line 474: RFC 9421 section 2.2.8 doesn't explicitly forbid duplicate query parameter names — it describes resolving each value independently. Throwing here is more restrictive than the RFC requires.

Is this intentional? If so, worth a comment explaining why (e.g. ambiguity avoidance). Otherwise, consider supporting multiple values — the RFC's serialization allows for it.

bartlomieju (changes requested):
Thanks for the thorough implementation! A few things to address:

**`badge.svg` changes don't belong in this PR** — The reformatting of the SVG is unrelated to message signatures. Please drop it from the branch (or submit separately).

tomas-zijdemans:
T%hanks for the feedback!

- Revert unrelated badge.svg change. Ooops! 🫣 
- Fix silent ECDSA curve fallback
- Fix same class of bug for HMAC, RSA-PSS, and RSASSA-PKCS1-v1_5
- Document `;bs` limitation: add warnings in JSDoc
- Add RFC citation for duplicate `@query-param` rejection. The throw is mandated by RFC 9421 section 2.2.8 ("the named query parameter MUST NOT be included")
- Remove redundant `?? undefined` in verifyMessage.
- Add tests. Coverage can't reach 100% becuase of unreachable code

http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1128 @@
...
+  return serializeList([member]);
+}
+
+function resolveBinaryWrapped(headerValue: string): string {
+  // Each field value is individually wrapped as a Byte Sequence.
+  // The Fetch API Headers.get() joins multiple values with ", " and does not
+  // expose getAll(). Splitting on ", " is therefore the best we can do, but
+  // it will mishandle single header values that contain a literal ", " (e.g.
+  // the Date header). Avoid using ;bs on such fields.
+  const values = headerValue.split(", ");
```
bartlomieju, line 571: The `;sf` resolution tries Dictionary → List → Item. Tests cover Dictionary and Item fallback paths, but there's no test for a header that fails Dictionary parse but succeeds as a List (e.g. a header like `"1, 2, 3"` or `"(a b), (c d)"`). That List branch is untested — please add a test case.

http/unstable_message_signatures.ts
```diff
@@ -0,0 +1,1128 @@
...
+ *   signedRequest,
+ *   async (keyId) => lookupPublicKey(keyId),
+ *   { requiredComponents: ["@method", "@authority"] },
+ * );
+ * ```
+ *
+ * @param message The HTTP request or response to verify.
+ * @param keyLookup Resolves a key identifier to a CryptoKey, or `null` if the
+ *   key is not found. When the signature has no `keyid` parameter, the empty
+ *   string `""` is passed.
```
bartlomieju, line 891: If the message already has a signature with the same label (e.g. default `"sig"`), `appendHeader` will produce a Signature-Input dictionary with duplicate keys. Per RFC 8941, the last value wins on parse, so the old signature is silently replaced while its byte sequence lingers in the Signature header.

Consider either:
- Checking for label conflicts and throwing, or
- Documenting in `SignatureParams.label` that the caller is responsible for uniqueness

The multi-signature test uses distinct labels and works correctly, so this only matters for the default-label case with pre-signed messages.

bartlomieju (commented):
Previous round's feedback is fully addressed (ECDSA/HMAC/RSA-PSS throws, `;bs` docs, `@query-param` RFC citation). This is a solid RFC 9421 implementation — well-structured, thorough tests, good use of the existing structured fields module.

A couple of things I'd like addressed before approving:

tomas-zijdemans:
Thanks,

- Added a `;sf` List fallback test using header `"1, 2, 3"`, asserting reserialized output to exercise the untested branch.
- Added `assertUniqueLabel` that throws `TypeError` on label collision in either Signature-Input or Signature header, per RFC 9421 §4.

bartlomieju (approved):

PR #7041, opened 2026-03-10
fix(http): always cancel request body in `serveFile()`

iuioiua:
Since the request body will always be ignored anyway.


kt3k (approved):
LGTM

PR #7042, opened 2026-03-12
BREAKING(crypto): throw TypeError on length mismatch in `timingSafeEqual()`

tomas-zijdemans:
- Length mismatch now throws `TypeError` instead of returning `false`, matching `node:crypto.timingSafeEqual` behaviour ~~(this is a security issue with the current implementation. Every call should take the same amount of time regardless of whether the values match or not)~~
- Switched internal comparison from `DataView.getUint8()` to `Uint8Array` indexing. Twice as fast for typical hash digest sizes (32–64 bytes).
- Removed redundant `DataView` from the parameter union type. `DataView` already satisfies `ArrayBufferView`, so `DataView` inputs still work.
- Replaced` while (++i) `loop with standard `for` loop for readability.
- Added JSDoc note about V8 JIT and SharedArrayBuffer constant-time limitations.
- Added tests for raw `SharedArrayBuffer` inputs and `TypeError` on length mismatch.


iuioiua:
WDYT @kitsonk?

tomas-zijdemans:
Agree with the above. I updated the PR title and description.

I vote for treating the old `return false` as a bug.

Let me know if any further changes are needed ☺️

PR #7044, opened 2026-03-13
perf(cache): optimize _serializeArgList key generation

tomas-zijdemans:
Reduce overhead in `_serializeArgList.ts` by replacing the `assertWeakKey` helper with a direct `WeakRef` check, using a `Set` instead of an array for key tracking, and cutting redundant map lookups. 

Improved the test coverage.


tomas-zijdemans:
Re-added the two tests and the defensive guard

PR #7045, opened 2026-03-13
feat(cache): expose cache property on memoized functions

tomas-zijdemans:
Exposed the cache as a public property so callers can inspect or manually invalidate the cache (e.g. `fn.cache.delete(...)`, `fn.cache.clear()`).

Changed MemoizationCache from type alias to interface to align with the style guide

Removed unused import and dropped the lint suppression.

Replaced the `cache.has(key)` + `cache.get(key)!` two-step lookup with a single `cache.get(key) `call. This avoids a redundant lookup on every cache hit.

Simplified tests a lot, maintaining coverage


tomas-zijdemans:
Thanks, I added a comment and re-added the tests you pointed out.

PR #7046, opened 2026-03-13
feat(cache/unstable): add sliding expiration to `TtlCache`

tomas-zijdemans:
NOTE: Builds on https://github.com/denoland/std/pull/7065

Sliding expiration: entries can now stay alive as long as they're being accessed, with an optional hard deadline. Useful for sessions or rate-limit windows.

### DIFF
--- a/cache/ttl_cache.ts
+++ b/cache/ttl_cache.ts
@@ -4,9 +4,7 @@
 import type { MemoizationCache } from "./memoize.ts";
 
 /**
- * Options for {@linkcode TtlCache.prototype.set} when
- * {@linkcode TtlCacheOptions.slidingExpiration | slidingExpiration} is
- * disabled (the default).
+ * Options for {@linkcode TtlCache.prototype.set}.
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  */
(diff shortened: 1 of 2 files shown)


tomas-zijdemans:
Split out the breaking changes as requested: https://github.com/denoland/std/pull/7065

Once that is merged I will update this PR to build on it.

tomas-zijdemans:
Updated the PR, it now cleanly applies only the new sliding expiration feature

bartlomieju:
The third type parameter `Sliding` feels like it adds more friction than safety. Users have to write the `true` twice — once in the type and once in the constructor:

```ts
const cache = new TtlCache<string, number, true>(100, {
  slidingExpiration: true,
});
```

TypeScript can't infer the class-level generic from the constructor options, so this redundancy is unavoidable with the current design.

A few alternatives worth considering:

1. **Subclass** — `SlidingTtlCache<K, V>` that always has sliding behavior and accepts `absoluteExpiration` in `set()`. Clean separation, no extra type param.

2. **Factory** — `TtlCache.sliding<K, V>(100)` returning a properly-typed instance.

3. **Drop the type param entirely** — accept `absoluteExpiration` at runtime regardless. It's already silently ignored when sliding is off, so the type-level guard isn't buying much beyond catching a niche mistake at compile time.

Any of these would keep the API at two type params, which is much more ergonomic. Since this is still unstable, now's the time to get the shape right.

tomas-zijdemans:
Agree, that makes a lot of sense. I decided for option 3 (dropping the type parameter) with runtime validation.

bartlomieju:
Looks good, nice work! One minor nit for a follow-up:

The `absoluteExpiration` validation in `set()` happens after the entry is already inserted into the map (and the timeout scheduled). If someone passes an invalid value like `NaN` or `-1`, the `RangeError` fires but the entry is already live in the cache. Moving the validation up next to the `ttl` check (before any state mutation) would keep things consistent.

Not blocking — happy to land this as-is and fix in a follow-up.

tomas-zijdemans:
Good catch. Sorry about that! Fix included here: https://github.com/denoland/std/pull/7070

PR #7047, opened 2026-03-13
feat(cache/unstable): improve `LruCache` callback and add `peek()`

tomas-zijdemans:
- `onEject` callback now receives a reason parameter ("evicted", "deleted", or "cleared") so callers can distinguish why an entry was removed. 
- New `peek()` method for reading a value without promoting it. 
- `has()` no longer promotes entries in the eviction order
- Constructor validates `maxSize`, rejecting zero, negative, non-integer, and non-finite values. 
- `maxSize` is now a readonly getter backed by a private field. 
- `onEject` callback is now optional, and `clear()` calls it for every entry
- calling `set`, `delete`, or `clear` inside an `onEject` callback throws instead of corrupting state. 
- `delete()` returns false immediately for non-existent keys and fires `onEject` only after the entry is fully removed. 
- Test coverage improved.

BEHAVIORAL CHANGES:
`has()` no longer promotes entries in the eviction order. Previously, calling `has()` on a key would mark it as recently used and keep it alive longer. Now it only checks whether the key exists. Code that relies on `has()` to prevent eviction should switch to `get()`.

`set()` does not fire `onEject` when overwriting an existing key with a new value. The old value is silently replaced. Callers that need to clean up old values on overwrite should handle that themselves before calling `set()`.

### DIFF
--- a/cache/lru_cache.ts
+++ b/cache/lru_cache.ts
@@ -2,6 +2,7 @@
 // This module is browser compatible.
 
 import type { MemoizationCache } from "./memoize.ts";
+export type { MemoizationCache };
 
 /**
  * The reason an entry was removed from the cache.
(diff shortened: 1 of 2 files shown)


tomas-zijdemans:
I re-added the export and changed to `TypeError`. Thanks! (also updating PR description as requested)

PR #7049, opened 2026-03-13
feat(collections): add `"union"` to `DeepMergeOptions.arrays`

### LINKED ISSUE #7048: Add `"union"` `MergingStrategy` to `DeepMergeOptions.arrays` in `@std/collections/deep-merge`

When I first tried the `deepMerge` `function`, I expected the default behavior to be as follows:
 ```ts
import { deepMerge } from "@std/collections/deep-merge";

const a = { foo: [1, 2] };
const b = { foo: [2, 3] };

const result = deepMerge(a, b, { arrays: "merge" });

const expected = { foo: [1, 2, 3] };

assertEquals(result, expected);
```
Then I realized that the implementation uses `Arrays.prototype.concat()` just like the default strategy by [npm:deepmerge](https://www.npmjs.com/package/deepmerge).

Adding a `"union"` `MergingStrategy` for `arrays` only, preferably through a new exported type `ArraysMergingStrategy`.
Implementation can use the `union` `function` imported from `"./union.ts"`.

---

> [!WARNING]
> The following solution would be a BREAKING CHANGE 

Splitting the `MergingStrategy` type into 3 separate types:
```ts
export type ArraysMergingStrategy = "concat" | "union" | "replace";

export type MapsMergingStrategy  = "union" | "replace";

export type SetsMergingStrategy = "union" | "replace";
```
This way, it is clear what to expect from the strategy. Implementations of maps and sets do not need to change, just the name of the strategy (from `"merge"` to `"union"`).

The default strategy should be `"replace"` for all, since the current fallback if the user entered an unmapped strategy,  is to `return right`. Example below:
 ```ts
import { deepMerge } from "@std/collections/deep-merge";

const a = { foo: new Set(["foo"]) };
const b = { foo: new Set(["bar"]) };

// @ts-expect-error: user passes unmapped strategy
const result = deepMerge(a, b, { sets: "unmapped-strategy" });

const expected = { foo: new Set(["bar"]) };

assertEquals(result, expected);
```

### PR DESCRIPTION
**Pull request by DonBLong**
Adds a `"union"` `MergingStrategy` for `arrays` only, through a new exported type `ArraysMergingStrategy`. Implementation uses the `union` `function` imported from `"./union.ts"`

closes: #7048

### DIFF
--- a/collections/deep_merge.ts
+++ b/collections/deep_merge.ts
@@ -1,6 +1,8 @@
 // Copyright 2018-2026 the Deno authors. MIT license.
 // This module is browser compatible.
 
+import { union } from "./union.ts";
+
 /** Default merging options - cached to avoid object allocation on each call */
 const DEFAULT_OPTIONS: DeepMergeOptions = {
   arrays: "merge",
@@ -303,6 +305,8 @@ function mergeObjects(
     if ((Array.isArray(left)) && (Array.isArray(right))) {
       if (options.arrays === "merge") {
         return left.concat(right);
+      } else if (options.arrays === "union") {
+        return union(left, right);
       }
 
       return right;
@@ -380,6 +384,9 @@ function getKeys<T extends Record<string, unknown>>(record: T): Array<keyof T> {
 /** Merging strategy */
 export type MergingStrategy = "replace" | "merge";
 
+/** Arrays Merging strategy */
+export type ArraysMergingStrategy = MergingStrategy | "union";
+
 /** Options for {@linkcode deepMerge}. */
 export type DeepMergeOptions = {
   /**
@@ -387,7 +394,7 @@ export type DeepMergeOptions = {
    *
    * @default {"merge"}
    */
-  arrays?: MergingStrategy;
+  arrays?: ArraysMergingStrategy;
   /**
    * Merging strategy for maps.
    *
--- a/collections/deep_merge_test.ts
+++ b/collections/deep_merge_test.ts
@@ -247,6 +247,19 @@ Deno.test("deepMerge() handles array merge (merge)", () => {
   );
 });
 
+Deno.test("deepMerge() handles array merge (union)", () => {
+  assertEquals(
+    deepMerge({
+      foo: [1, 2, 3],
+    }, {
+      foo: [3, 4, 5],
+    }, { arrays: "union" }),
+    {
+      foo: [1, 2, 3, 4, 5],
+    },
+  );
+});
+
 Deno.test("deepMerge() handles maps merge (replace)", () => {
   assertEquals(
     deepMerge({

### CODE REVIEW

**Review by bartlomieju — COMMENTED** [review]
The new `unstable_deep_merge.ts` (598 lines) is a near-verbatim copy of the existing `deep_merge.ts` (592 lines). The only meaningful differences are:

1. `import { union } from "./union.ts"` added
2. `ArraysMergingStrategy` type added (`MergingStrategy | "union"`)
3. `DeepMergeOptions.arrays` changed from `MergingStrategy` to `ArraysMergingStrategy`
4. 3 lines in `mergeObjects()` to handle the `"union"` case

Everything else — function signatures, JSDoc, all type utilities, helper functions — is duplicated. Maintaining two near-identical 600-line files is not sustainable.

This should instead modify the existing `deep_merge.ts` directly, since the change is purely additive (adding a new valid string literal to the arrays option). Existing callers are unaffected.

### Additional issues

**JSDoc examples import from wrong path** — All doc examples import from `"@std/collections/deep-merge"` instead of `"@std/collections/unstable-deep-merge"`, so doc tests would either fail or test the wrong module.

**`Merge` type doesn't handle `"union"` for arrays** — The `Merge` type only checks for `{ arrays: "replace" }`. When `arrays: "union"` is passed, the type falls through to `MergeAllArrays<T, U>`, which happens to be correct by coincidence, but the intent isn't expressed.

**Test file also duplicates existing tests** — The test file (494 lines) copies most tests from `deep_merge_test.ts`. Only ~2 tests are actually new (the `"union"` merge test and the default strategy test).

### Suggestion

Instead of a new file, add the `"union"` strategy to the existing `deep_merge.ts`:

1. Add `ArraysMergingStrategy = MergingStrategy | "union"`
2. Change `DeepMergeOptions.arrays` type to `ArraysMergingStrategy`
3. Add the 3-line `union` branch in `mergeObjects()`
4. Add the union-specific tests to the existing test file

If it needs to stay unstable, the new type can be exported from an `unstable_` entrypoint that re-exports and extends the stable module.

**Comment by DonBLong** [comment]
> Maintaining two near-identical 600-line files is not sustainable

My bad. It's my first time contributing to this repo and I have misunderstood the [contributing guides](https://github.com/denoland/std?tab=contributing-ov-file#new-features-in-stable-packages-version--100). I thought "new APIs" meant any addition to a stable API. I also thought that the maintainers would want to compare the diff between the `unstable_` files and the stable ones, which is why I copy-pasted the stable ones instead of extending.

I took your first suggestion, and applied the changes to the original `deep_merge.ts` and `deep_merge_test.ts` and also removed the exported entry `unstable-deep-merge` from deno.json.

I also modified the commit message to reflect that the feature is being added directly to the stable `collections` scope instead of `collections/unstable`.

**Comment by DonBLong** [comment]
> `Merge` type doesn't handle `"union"` for arrays

> [!NOTE] 
>
>**TL;DR**
>
>The reason why I didn't touch the `Merge` type, is because it actually aligns more with the `union` strategy than with the `concat` method, since types cannot concatenate. So, I believe the intent is expressed correctly here, because the `Merge` type basically says that `"replace" = PartialByType<U, Array<unknown>>` which is right side, and any other merging strategy falls through to a union of all types.

---

### About that though

It's funny that you mentioned this type, because that's what confused me about the return of `deepMerge` in the first place. The problem is, **types only know union**. For example:

```ts
const a = { foo: [1, 2, 3] };

const b = { foo: [3, 4, 5] };

const abFoo = a.foo.concat(b.foo);
```

The type of `abFoo` is just a `number[]`, because neither `a.foo` nor `b.foo` are `readonly` so they both have the same type. And even if `b.foo` was `readonly`, `abFoo` would still be just a `number[]` because `number` devours whatever specific numbers `b.foo` would have.

However, according to the `Merge` type, the following is also possible:

```ts
const a = { foo: [1, 2, 3] };

const b = { foo: ["string", 4, 5] };

const abFoo = a.foo.concat(b.foo); // what deepMerge(a, b, { arrays: "merge" }) does internally
```

Because the type of `a.foo` and `b.foo` inside of the `deepMerge` call are **both** immediately converted to `(string | number)[]`. And because I know that `a.foo.concat(b.foo)` would throw the following TypeScript error:

```
Type '(string | number)[]' is not assignable to type 'number[]'
```

I thought that `deepMerge` is probably using the spread syntax to merge the arrays, which would not result in a TypeScript error, but would also not remove duplicates. Or, that it is using `union`, and by `union` I mean either `@std/collections/union` or the new `Set.union` method, which would also not throw a TypeScript error in this case, but would also remove the duplicates from both `a.foo` and `b.foo`:

```ts
const a = { foo: [1, 2, 3] };

const b = { foo: ["string", 3, 4] };

const abFoo = union(a.foo, b.foo);

// OR: const abFoo = Array.from(new Set(a.foo).union(new Set(b.foo)));

assertEquals(abFoo, [1, 2, 3, "string", 4]);
```

### Conclusion

I think the confusion comes from the name `"merge"`, because it doesn't state exactly the type of merge the function is doing. Is it a `union`, an `intersection`, a `difference`, and in case of `arrays` a `concat`? Which why my other solution in #7048 was to rename the option `"merge"` to what it really does, `"concat"` for `arrays` and `"union"` for `Sets` and `Maps`.

lunadogbot (approved):
Implementation matches the prior suggestion: `ArraysMergingStrategy = MergingStrategy | "union"`, widened `DeepMergeOptions.arrays?`, and a three-line `union(left, right)` branch in `mergeObjects()`. `union` is already in the package (no new dep, no circular risk) and its `Set`-based dedupe with first-occurrence ordering carries over here predictably.

On the earlier "Merge type doesn't handle union" concern: the fallthrough to `MergeAllArrays<T, U>` is right — both `"merge"` and `"union"` yield `Array<T[number] | U[number]>` at the type level, so a separate type branch isn't needed.

- nit: line 387 `/** Arrays Merging strategy */` — sentence case for consistency with line 385 (`/** Merging strategy */`).
- nit: `deepMerge`'s JSDoc has `@example`s for `"merge"` (default) and `"replace"` but not `"union"`. Adding one would make the new strategy discoverable from the rendered docs.

fibibot (approved):
Adds `"union"` as a third array merging strategy via a new exported `ArraysMergingStrategy` type that widens `DeepMergeOptions.arrays`. Behavior is correct — `union(left, right)` preserves left-then-right order with Set-based dedup (SameValueZero, so reference equality for objects). CI green.

- nit: `deepMerge`'s JSDoc still claims "either `replace` values, or `merge` them instead" (`collections/deep_merge.ts:17-18`, `:114-115`) — mention `"union"` for arrays.
- nit: `DeepMergeOptions.arrays` JSDoc (`:393`) should list the three values or link to `{@linkcode ArraysMergingStrategy}` so users discover the new option.
- nit: `/** Arrays Merging strategy */` (`:387`) — match the existing `/** Merging strategy */` capitalization.
- nit: add an `@example` to `deepMerge` showcasing `arrays: "union"`, parallel to the existing `"replace"` example.

collections/deep_merge.ts
```diff
@@ -380,14 +384,17 @@ function getKeys<T extends Record<string, unknown>>(record: T): Array<keyof T> {
 /** Merging strategy */
 export type MergingStrategy = "replace" | "merge";
 
+/** Arrays Merging strategy */
```
bartlomieju, line 387: nit (raised by both bots): match the capitalization of `/** Merging strategy */` on the line above.

```suggestion
/** Arrays merging strategy */
```

collections/deep_merge.ts
```diff
@@ -380,14 +384,17 @@ function getKeys<T extends Record<string, unknown>>(record: T): Array<keyof T> {
...
+
 /** Options for {@linkcode deepMerge}. */
 export type DeepMergeOptions = {
   /**
    * Merging strategy for arrays
    *
    * @default {"merge"}
    */
-  arrays?: MergingStrategy;
+  arrays?: ArraysMergingStrategy;
```
bartlomieju, line 397: The `arrays` option's JSDoc above only documents the default. Now that there are three valid values, consider listing them or linking the type so users discover `"union"` from the rendered docs, e.g. add a line like:

```
   * See {@linkcode ArraysMergingStrategy} for the available strategies.
```

collections/deep_merge.ts
```diff
@@ -303,6 +305,8 @@ function mergeObjects(
     if ((Array.isArray(left)) && (Array.isArray(right))) {
       if (options.arrays === "merge") {
         return left.concat(right);
+      } else if (options.arrays === "union") {
+        return union(left, right);
```
bartlomieju, line 309: Semantic note worth documenting: because `union()` funnels both inputs through a single `Set`, `"union"` also dedups *within* each input array, not just across them — e.g. `deepMerge({ a: [1, 1, 2] }, { a: [2, 3] }, { arrays: "union" })` yields `{ a: [1, 2, 3] }`. That differs from `"merge"` (plain `concat`, keeps duplicates). It's correct union semantics, but users coming from `"merge"` may not expect it, so please call it out in the `deepMerge` JSDoc and ideally show it in the `"union"` `@example`. (Dedup uses SameValueZero, so objects compare by reference.)

bartlomieju (commented):
The rework looks great — clean additive change, correct `union` semantics, and CI is green. Two small things still open: the doc nits that @lunadogbot and @fibibot both flagged are not yet applied, and there's a semantic detail about `"union"` worth documenting. I've left inline suggestions; two of the nits land outside the diff so I'm noting them here:

- **`deepMerge` JSDoc doesn't mention `"union"`** (`collections/deep_merge.ts:18` and `:115`): both still read "`replace` values, or `merge` them instead". Please extend to mention that arrays can also be `"union"`-ed.
- **No `@example` for `arrays: "union"`**: there's an example for the `"replace"` strategy but none for the new one. A short parallel example would make it discoverable in the rendered docs (and is a natural place to show the dedup behavior — see the inline note on the implementation).

None of these block correctness; they're just worth closing out since two reviewers independently asked.

DonBLong:
> [!NOTE]
>
> TL;DR
>
> Thank you very much @bartlomieju for the reviews.
>
> After giving this module enough thought, I've decided to close this pull request as well as the issue linked to it.
>
> The reason why I never touched the JSDocs is because any attempt I made to patch it introduced more complexity/confusion rather than clarity. The main cause of this confusion is the way the `DeepMergeOptions` type and the `mergeObjects` function are implemented.
>
> As much as I need this feature, jamming it into this module the way the module is currently structured feels hacky/forced and does not match the standards of the other modules in this library.

---

### Why it is difficult to update the JSDocs

> deepMerge JSDoc doesn't mention "union"

Because the word `merge` is a bit vague/generic, I can't just list the option `union` for arrays, I'd also have to explain that `merge` is actually `concat` so the user knows the difference between both options. However, `merge` for sets is also technically union, and for maps it's a union of keys. So now I feel the need to also explain what `merge` means for both sets and maps. Which will result in converting a 2-line summary describing the options of the function into a paragraph of explanations. And honestly, after reading it out loud, I myself felt deterred from using the function.


### Other Problems

1. The word `values` here is a bit misleading because in case of maps, it's actually referring to the keys of the `Map` rather than the values, since the implementation **replaces** the values from `left` with the values from `right` for all mutual keys.
https://github.com/denoland/std/blob/ad7c87bb1fb192d93443d16cc77e58002de79b71/collections/deep_merge.ts#L15-L16
Implementation here sets `k` to `v` from `right`, so if `left` also has its own value for `k` it will be replaced and not merged. Thus, what is actually being merged is the keys and not the values.
https://github.com/denoland/std/blob/ad7c87bb1fb192d93443d16cc77e58002de79b71/collections/deep_merge.ts#L313-L318
It's also worth pointing out that the JSDoc example doesn't show the case when there are mutual keys between the two maps.
https://github.com/denoland/std/blob/ad7c87bb1fb192d93443d16cc77e58002de79b71/collections/deep_merge.ts#L56-L69


2. The `@default {"merge"}` here is also a bit misleading because it's only applied if the parameter `options` is `undefined`, but, for example, if the user passes an empty object `{}` to the `options` parameter (which is a valid `DeepMergeOptions` value), they will probably be surprised to find that the values have been replaced instead of merged.
https://github.com/denoland/std/blob/ad7c87bb1fb192d93443d16cc77e58002de79b71/collections/deep_merge.ts#L384-L390
Implementation here falls back to returning `right` if the option is not `merge`, when it should have been checking the special case (i.e. when the option is `replace`) instead of checking the default case.
https://github.com/denoland/std/blob/ad7c87bb1fb192d93443d16cc77e58002de79b71/collections/deep_merge.ts#L304-L309

3. Unifying different merging strategies under one option `merge` hinders the scalability of the function, as it makes it difficult to add new different merging strategies for each data structure. If each data structure had it's own merging strategy type, then strategies like `intersect` and `difference` could be easily added to both sets and arrays along with their JSDoc definitions and examples. This would also allow for naming the strategies meaningful names that corresponds to the methods being used such as; `union`, `concat`, `intersect`..etc, which eliminates any confusion about the output of the strategy.

---

I just feel that the right thing to do is open a new issue with these points and try solving them in a new pull request, what I should've done from the beginning, before introducing new strategies like `union` or `intersect`.

PR #7051, opened 2026-03-15
fix(http): handle empty values and multi-line comments in `ServerSentEventStream`

tomas-zijdemans:
**Bug 1:** Empty values silently dropped. Fields like `{ data: "" }` or `{ id: 0 }` produced no output because the code used truthy checks (`if (message.data)`), which treat `""` and `0` as false. Fixed by checking `!== undefined` instead.

**Bug 2:** Multi-line comments threw an error. The serializer rejected comments containing newlines with a `SyntaxError`, but comments are per-line constructs in SSE. This already worked correctly for data, just not for comments. Fixed by splitting on newlines instead of throwing.

See `ServerSentEventParseStream` for a roundtrip example (that requires this fix to work correctly)

Note: `id?: string | number;` in the `ServerSentEventMessage`interface is a spec violation (it should only be `string`). I didn't change it, since that would be a breaking change. @kt3k: please let me know if you'd like me to include that as well.

Spec: https://html.spec.whatwg.org/multipage/server-sent-events.html#parsing-an-event-stream


tomas-zijdemans:
> One question: should `event: ""` still be skipped rather than serialized as `event:\n\n`? An empty event type is meaningless in the SSE spec (it defaults to `"message"` when absent), so emitting an empty `event:` line seems worse than omitting it. The other fields (`data: ""`, `id: 0`, `comment: ""`) make more sense to serialize.

Skipping it should be fine. The only note I would make is that `ServerSentEventParseStream` will parse `event:` as `{ event: "" }`, so skipping empty event on the serialize side would break the serialize→parse roundtrip.

Let me know if you think skipping makes more sense, I'll update accordingly.

PR #7052, opened 2026-03-17
feat(text/unstable): add `truncate`

tomas-zijdemans:
Partially fixes: https://github.com/denoland/std/issues/6456


tomas-zijdemans:
Thanks. I have updated the docs and tightened the input validation as suggested.

Word and grapheme-aware truncation can be cleanly applied on top of this code in future PRs.

PR #7057, opened 2026-03-20
feat(xml/unstable): add XML 1.1 support

tomas-zijdemans:
I went for no version auto-detection from  `<?xml version="...">`. Users explicitly set `xmlVersion: "1.1"`. The reasoning for this is that it's very very common to encounter XML files where the version is not honest, so It would just confuse users. The alternative would be a opaque `strict` or `lenient` mode.

I checked XML 1.1 conformance against the official XML test suite.

The XML module gets improved test coverage with this PR

### DIFF
--- a/xml/_common_test.ts
+++ b/xml/_common_test.ts
@@ -2,21 +2,14 @@
 
 import { assertEquals } from "@std/assert";
 import {
+  isIllegalXmlLiteralChar,
   LINE_ENDING_REGEXP,
   parseName,
-  validateNamespaceBinding as _validateNamespaceBinding,
+  validateNamespaceBinding,
   validatePubidLiteral,
   validateQName,
   validateXmlDeclaration,
 } from "./_common.ts";
-
-/** Wrapper that defaults xml11 to false for existing tests. */
-function validateNamespaceBinding(
-  attrName: string,
-  attrValue: string,
-): string | null {
-  return _validateNamespaceBinding(attrName, attrValue, false);
-}
 
 // =============================================================================
 // validateXmlDeclaration Tests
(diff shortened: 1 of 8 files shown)


xml/_tokenizer.ts
```diff
@@ -329,6 +354,28 @@ export class XmlTokenizer {
...
+        code === 0x0B || code === 0x0C ||
+        (code >= 0x0E && code <= 0x1F) ||
+        (code >= 0x7F && code <= 0x84) ||
+        (code >= 0x86 && code <= 0x9F) ||
+        code === 0xFFFE || code === 0xFFFF;
+    }
+    // XML 1.0: C0 controls except TAB/LF/CR + noncharacters (outside Char production)
+    return (code < 0x20 && code !== 0x09 && code !== 0x0A && code !== 0x0D) ||
+      code === 0xFFFE || code === 0xFFFF;
+  }
```
bartlomieju, line 360: This is identical to `isIllegalLiteralChar` in `_parse_sync.ts` (same ranges, same logic). Extract to `_common.ts` as a shared utility alongside the existing character validation functions.

xml/_parser.ts
```diff
@@ -173,7 +174,11 @@ export class XmlEventParser implements XmlTokenCallbacks {
   /** Pending namespace bindings for current element */
   #pendingNsBindings: Array<[string, string | undefined]> = [];
 
-  constructor(callbacks: XmlEventCallbacks, options: ParseStreamOptions = {}) {
+  constructor(
+    callbacks: XmlEventCallbacks,
+    options: ParseStreamOptions = {},
```
bartlomieju, line 179: The `xml11: boolean` parameter is now threaded through `decodeEntities`, `validateNamespaceBinding`, `normalizeAttributeValue`, `_add`, and this constructor. This works but is becoming a code smell — consider an options/context object or encapsulating version-specific behavior in one place rather than spreading a boolean through every function signature.

xml/_parse_sync.ts
```diff
@@ -683,8 +677,7 @@ export function parseSync(xml: string, options?: ParseOptions): XmlDocument {
         continue;
       }
 
-      // Content model operators: *, +, ?
-      if (code === 42 || code === 43 || code === 63) { // * + ?
+      if (code === CC_STAR || code === CC_PLUS || code === CC_QUESTION) {
```
bartlomieju, line 665: The inline `// Skip '>'` / `// Skip '%'` comments are gone in favor of named constants, which is fine. But semantic context comments like this one (`// Content model operators: *, +, ?`), `// Opening paren - must have whitespace before FIRST paren only`, `// Parameter entity reference`, etc. were also removed — those explain the DTD grammar being parsed and are useful even with named constants.

xml/_common_test.ts
```diff
@@ -2,12 +2,22 @@
...
+  LINE_ENDING_REGEXP,
   parseName,
-  validateNamespaceBinding,
+  validateNamespaceBinding as _validateNamespaceBinding,
+  validatePubidLiteral,
   validateQName,
   validateXmlDeclaration,
 } from "./_common.ts";
 
+/** Wrapper that defaults xml11 to false for existing tests. */
```
bartlomieju: This wrapper (and the matching one in `_entities_test.ts`) avoids touching existing tests, but adds indirection. Since the parameter addition is a simple `xml11: boolean` with a clear default, consider updating the existing call sites directly instead.

tomas-zijdemans:
Thanks for the review. I have updated the code accordingly:
- Deduplicated `isIllegalLiteralChar`
- Clarified naming. Now using `isValidXml10CharRef` and `isValidXml11CharRef`
- Added coverage for XML 1.1 paths
- Re-added the DTD comments
- Removed test wrapper functions

PR #7059, opened 2026-03-21
feat(front-matter/unstable): add `test` to each subpackage, including JSON, TOML, and YAML

uynilo9:
Let people use the `test` function more easily when they want to check a specific format of frontmatter.

### DIFF
--- a/front_matter/deno.json
+++ b/front_matter/deno.json
@@ -8,6 +8,8 @@
     "./test": "./test.ts",
     "./toml": "./toml.ts",
     "./yaml": "./yaml.ts",
+    "./unstable-json": "./unstable_json.ts",
+    "./unstable-toml": "./unstable_toml.ts",
     "./unstable-yaml": "./unstable_yaml.ts",
     "./types": "./types.ts"
   }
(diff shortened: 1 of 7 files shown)


bartlomieju (commented):
A few things to address:

1. `unstable_json.ts` and `unstable_toml.ts` duplicate `extract()` from the stable `json.ts` and `toml.ts`. These files should only contain the new `test()` function — either re-export `extract` from the stable module or just don't include it.
2. The JSON test cases use TOML syntax (`name = 'deno'`) as the front matter body. Since `test()` only validates delimiters this passes, but it's confusing — please use actual JSON content.
3. Import style is inconsistent: `json_test.ts` uses the `@std/front-matter/unstable-json` specifier while `toml_test.ts` uses a relative `./unstable_toml.ts` path. Please use one style consistently.

uynilo9:
Thanks for reviewing. I didn't realise some absurd mistakes I made since I've been little busy these days.

>     2. The JSON test cases use TOML syntax (`name = 'deno'`) as the front matter body. Since `test()` only validates delimiters this passes, but it's confusing — please use actual JSON content.

About the second thing you mention, I use the example directly from `test_test.ts`. Is it also a misused test?

PR #7067, opened 2026-03-27
feat(async): stabilize `Lazy`

tomas-zijdemans:


tomas-zijdemans:
Need to fix a small bug I found

PR #7068, opened 2026-03-27
feat(http): stabilize `cache_control`

tomas-zijdemans:


tomas-zijdemans:
Need to fix a few things I found first

PR #7069, opened 2026-03-27
feat(cache/unstable): add stats to `LruCache`

tomas-zijdemans:
Adding stats to the cache is nice for observability, logging, etc. I plan to add the same to TTL cache.


tomas-zijdemans:

PR #7072, opened 2026-03-30
feat(cache/unstable): add `maxSize` to `TtlCache`

tomas-zijdemans:
- Adds optional `maxSize` option to `TtlCache` that evicts the oldest entry (by insertion order) when capacity is exceeded
- Adds `maxSize` getter (returns `Infinity` when unset)
- Adds `#ejecting` re-entrancy guard on `set()`, `delete()`, and `clear()` during `onEject` callbacks, matching `LruCache`


tomas-zijdemans:

PR #7074, opened 2026-03-30
feat(data-structures/unstable): add `RollingCounter` serialization

tomas-zijdemans:
- Add `toJSON()` / `RollingCounter.from()` for JSON-compatible serialization and restoration
- Add `current` getter for reading the newest segment without iteration
- Add `Symbol.toStringTag` to match module precedent.

### DIFF
--- a/data_structures/unstable_rolling_counter.ts
+++ b/data_structures/unstable_rolling_counter.ts
@@ -193,22 +193,69 @@ export class RollingCounter {
         `Cannot rotate RollingCounter: steps must be a non-negative integer, got ${steps}`,
       );
     }
-    const len = this.#segments.length;
+    const segs = this.#segments;
+    const len = segs.length;
     if (steps >= len) {
       const evicted = this.#total;
-      this.#segments.fill(0);
+      segs.fill(0);
       this.#cursor = (this.#cursor + steps) % len;
       this.#total = 0;
       return evicted;
     }
+
+    let pos = this.#cursor + 1;
+    if (pos >= len) pos = 0;
+
     let evicted = 0;
-    for (let i = 0; i < steps; i++) {
-      this.#cursor = (this.#cursor + 1) % len;
-      evicted += this.#segments[this.#cursor]!;
-      this.#segments[this.#cursor] = 0;
+    const end = pos + steps;
+
+    if (end <= len) {
+      for (let i = pos; i < end; i++) {
+        evicted += segs[i]!;
+        segs[i] = 0;
+      }
+    } else {
+      for (let i = pos; i < len; i++) {
+        evicted += segs[i]!;
+        segs[i] = 0;
+      }
+      const wrap = end - len;
+      for (let i = 0; i < wrap; i++) {
+        evicted += segs[i]!;
+        segs[i] = 0;
+      }
     }
+
+    let newCursor = pos + steps - 1;
+    if (newCursor >= len) newCursor -= len;
+    this.#cursor = newCursor;
     this.#total -= evicted;
     return evicted;
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


data_structures/unstable_rolling_counter.ts
```diff
@@ -129,24 +193,71 @@ export class RollingCounter {
...
+        segs[i] = 0;
+      }
+      const wrap = end - len;
+      for (let i = 0; i < wrap; i++) {
+        evicted += segs[i]!;
+        segs[i] = 0;
+      }
     }
+
+    let newCursor = pos + steps - 1;
```
bartlomieju: Bug: `rotate(0)` sets the cursor to `-1` when `cursor === len - 1`.

When `steps === 0` and `cursor` is at the last position:
- `pos` wraps to `0`
- `newCursor = 0 + 0 - 1 = -1`
- The `>= len` guard doesn't catch it

This corrupts internal state — subsequent `increment()`, `current`, iteration, etc. all index with `-1`.

Simplest fix: early-return at the top of the method:
```ts
if (steps === 0) return 0;
```

tomas-zijdemans: Good catch, thanks! Added the early return and strengthened the test to assert `current` and a subsequent `increment()` after `rotate(0)`.

bartlomieju (commented):
The `rotate()` optimization: could you share the benchmarks that motivated this rewrite? The original 4-line loop was simple and correct — the new version is ~25 lines with wrap-around logic and already introduced a cursor corruption bug (`rotate(0)`). 

Modulo per iteration is cheap on modern engines, so I'd want to see concrete numbers before trading readability for this. What was the benchmark setup (segment count, step sizes) and what workload are you optimizing for?

If the gains are significant for your use case, this might be better as a separate PR with the benchmarks included so the tradeoff is documented.

tomas-zijdemans:
Below is the benchmark.

The speedup is small for regular rotation (the most common use case). I'm seeing an average of about 10% improvement across sizes. The big winner is, the arguably lot less used, bulk rotation seeing 5.5x to 7.5x.

I will revert the perf changes, and then we can see if that's something we want to introduce in a future PR. The readability hit is real, fully agree on that!


```
import { RollingCounter } from "./unstable_rolling_counter.ts";

// --- rotate(1): the common case ---

const small1 = new RollingCounter(5);
Deno.bench({
  group: "rotate(1)",
  name: "5 segments",
  fn: () => {
    small1.increment(1);
    small1.rotate();
  },
});

const med1 = new RollingCounter(60);
Deno.bench({
  group: "rotate(1)",
  name: "60 segments",
  fn: () => {
    med1.increment(1);
    med1.rotate();
  },
});

const large1 = new RollingCounter(1000);
Deno.bench({
  group: "rotate(1)",
  name: "1000 segments",
  fn: () => {
    large1.increment(1);
    large1.rotate();
  },
});

// --- bulk rotate ---

const medHalf = new RollingCounter(60);
Deno.bench({
  group: "rotate(n) bulk",
  name: "60 segments, rotate(30)",
  fn: () => {
    medHalf.increment(1);
    medHalf.rotate(30);
  },
});

const largeHalf = new RollingCounter(1000);
Deno.bench({
  group: "rotate(n) bulk",
  name: "1000 segments, rotate(500)",
  fn: () => {
    largeHalf.increment(1);
    largeHalf.rotate(500);
  },
});

const largeMost = new RollingCounter(1000);
Deno.bench({
  group: "rotate(n) bulk",
  name: "1000 segments, rotate(999)",
  fn: () => {
    largeMost.increment(1);
    largeMost.rotate(999);
  },
});

```

bartlomieju (approved):
LGTM, thanks

PR #7075, opened 2026-03-31
feat(http/unstable): add radix tree router; keep linear scan as routeLinear

esroyo:
Add `routeRadix()`, a radix tree router that provides O(segments) dispatch for static, parametric, and wildcard routes. Routes with complex URLPattern syntax (regex constraints, optional groups, inline wildcards, modifier suffixes) fall back to linear matching while preserving insertion order.

- `routeRadix`: radix tree with fallback to linear for complex patterns
- `routeLinear`: the original linear scan, extracted as its own export
- `route`: re-exported alias for `routeLinear`

The radix router matches routeLinear semantics exactly; insertion order is always respected, even when static and parametric routes overlap at the same tree depth.

Benchmarks show 1.5–9x improvement on static/parametric/wildcard routes, with negligible overhead on complex fallback patterns.

---

```mermaid
---
config:
    themeVariables:
        xyChart:
            plotColorPalette: "#e05c5c, #4a90d9"
---
xychart-beta
    title "route: 🔴 linear vs 🔵 radix - Avg latency (µs)"
    x-axis [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
    y-axis "Avg latency (µs)" 0 --> 7

    line [0.91, 2.04, 6.1, 1.63, 5.54, 1.49, 2.84, 1.08, 3.15, 0.9, 1.45, 1.98]
    line [1.08, 1.38, 1.39, 0.62, 0.61, 1.55, 1.66, 0.91, 1.73, 1.46, 1.62, 2.1]
```

| # | Group | 🔴 linear (µs) | 🔵 radix (µs) |
|---|---|--:|--:|
| 1 | static route — first in small table | 0.91 | 1.08 |
| 2 | static route — last in small table | 2.04 | 1.38 |
| 3 | static route — last in large table | 6.1 | 1.39 |
| 4 | static route — miss (small table) | 1.63 | 0.62 |
| 5 | static route — miss (large table) | 5.54 | 0.61 |
| 6 | parametric route — single param | 1.49 | 1.55 |
| 7 | parametric route — two params (nested) | 2.84 | 1.66 |
| 8 | parametric route — method mismatch | 1.08 | 0.91 |
| 9 | wildcard route | 3.15 | 1.73 |
| 10 | complex — regex constraint | 0.9 | 1.46 |
| 11 | complex — optional group | 1.45 | 1.62 |
| 12 | complex — inline wildcard with suffix | 1.98 | 2.1 |

---

```
    CPU | AMD Ryzen 7 PRO 6850U with Radeon Graphics
Runtime | Deno 2.7.9 (x86_64-unknown-linux-gnu)

file:///../denoland/std/http/unstable_route_bench.ts

| benchmark   | time/iter (avg) |        iter/s |      (min … max)      |      p75 |      p99 |     p995 |
| ----------- | --------------- | ------------- | --------------------- | -------- | -------- | -------- |

group static route — first in small table
| linear      |        921.1 ns |     1,086,000 | (869.7 ns …   1.4 µs) | 921.5 ns |   1.4 µs |   1.4 µs |
| radix       |          1.1 µs |       885,100 | (  1.1 µs …   1.3 µs) |   1.1 µs |   1.3 µs |   1.3 µs |

summary
  linear
     1.23x faster than radix

group static route — last in small table
| linear      |          2.0 µs |       494,200 | (  2.0 µs …   2.6 µs) |   2.0 µs |   2.6 µs |   2.6 µs |
| radix       |          1.3 µs |       751,200 | (  1.3 µs …   1.7 µs) |   1.3 µs |   1.7 µs |   1.7 µs |

summary
  linear
     1.52x slower than radix

group static route — last in large table
| linear      |          5.9 µs |       170,100 | (  5.8 µs …   6.1 µs) |   5.9 µs |   6.1 µs |   6.1 µs |
| radix       |          1.3 µs |       750,600 | (  1.3 µs …   1.6 µs) |   1.3 µs |   1.6 µs |   1.6 µs |

summary
  linear
     4.41x slower than radix

group static route — miss (small table)
| linear      |          1.7 µs |       596,400 | (  1.6 µs …   2.1 µs) |   1.7 µs |   2.1 µs |   2.1 µs |
| radix       |        622.6 ns |     1,606,000 | (600.0 ns … 702.4 ns) | 621.7 ns | 702.4 ns | 702.4 ns |

summary
  linear
     2.69x slower than radix

group static route — miss (large table)
| linear      |          5.6 µs |       179,100 | (  5.4 µs …   5.9 µs) |   5.6 µs |   5.9 µs |   5.9 µs |
| radix       |        606.1 ns |     1,650,000 | (594.0 ns … 641.1 ns) | 610.5 ns | 641.1 ns | 641.1 ns |

summary
  linear
     9.21x slower than radix

group parametric route — single param
| linear      |          1.5 µs |       645,600 | (  1.5 µs …   1.7 µs) |   1.6 µs |   1.7 µs |   1.7 µs |
| radix       |          1.5 µs |       649,500 | (  1.5 µs …   1.9 µs) |   1.5 µs |   1.9 µs |   1.9 µs |

summary
  linear
     1.01x slower than radix

group parametric route — two params (nested)
| linear      |          2.8 µs |       360,900 | (  2.7 µs …   2.9 µs) |   2.8 µs |   2.9 µs |   2.9 µs |
| radix       |          1.6 µs |       614,300 | (  1.6 µs …   1.7 µs) |   1.7 µs |   1.7 µs |   1.7 µs |

summary
  linear
     1.70x slower than radix

group parametric route — method mismatch
| linear      |          1.1 µs |       924,900 | (  1.0 µs …   1.4 µs) |   1.1 µs |   1.4 µs |   1.4 µs |
| radix       |        929.0 ns |     1,076,000 | (900.7 ns … 980.0 ns) | 933.5 ns | 980.0 ns | 980.0 ns |

summary
  linear
     1.16x slower than radix

group wildcard route
| linear      |          3.2 µs |       314,400 | (  3.1 µs …   3.6 µs) |   3.2 µs |   3.6 µs |   3.6 µs |
| radix       |          1.7 µs |       600,900 | (  1.6 µs …   2.0 µs) |   1.7 µs |   2.0 µs |   2.0 µs |

summary
  linear
     1.91x slower than radix

group complex — regex constraint
| linear      |        947.0 ns |     1,056,000 | (908.1 ns …   1.3 µs) | 947.5 ns |   1.3 µs |   1.3 µs |
| radix       |          1.5 µs |       673,300 | (  1.4 µs …   1.7 µs) |   1.5 µs |   1.7 µs |   1.7 µs |

summary
  linear
     1.57x faster than radix

group complex — optional group
| linear      |          1.5 µs |       680,000 | (  1.4 µs …   1.7 µs) |   1.5 µs |   1.7 µs |   1.7 µs |
| radix       |          1.6 µs |       625,800 | (  1.6 µs …   1.8 µs) |   1.6 µs |   1.8 µs |   1.8 µs |

summary
  linear
     1.09x faster than radix

group complex — inline wildcard with suffix
| linear      |          1.9 µs |       529,200 | (  1.8 µs …   2.1 µs) |   1.9 µs |   2.1 µs |   2.1 µs |
| radix       |          2.0 µs |       490,400 | (  2.0 µs …   2.2 µs) |   2.0 µs |   2.2 µs |   2.2 µs |

summary
  linear
     1.08x faster than radix
```

### DIFF
--- a/http/unstable_route.ts
+++ b/http/unstable_route.ts
@@ -181,12 +181,14 @@ function parsePathname(url: string): string {
  *   - Regex-constrained params:           `:id(\d+)`  `:lang(en|fr)`
  *   - Inline wildcards:                   `*.js`  `prefix*`
  *   - Modifier suffixes on params/groups: `:id?`  `:id+`  `:id*`  `(\d+)?`
+ *   - Backslash escapes for literals:     `\+`  `\?`  `\*`  `\:`  `\(`  `\{`
  *
- * This heuristic relies on URLPattern syntax being well-formed: URLPattern's
- * own parser rejects `?`, `+`, and `*` as literal characters in pathnames
- * (e.g. `new URLPattern({ pathname: "/foo+bar" })` throws at construction),
- * so a segment that ends with one of those characters here is always a
- * modifier, never a literal.
+ * Note that URLPattern's parser treats `?`, `+`, and `*` as modifier tokens
+ * in pathnames; literal occurrences (e.g. `/c++`) must be backslash-escaped
+ * (e.g. `/c\+\+`). The `pathname` property preserves the escape sequence,
+ * which the radix tree cannot match against an unescaped request path —
+ * so any segment containing `\` is routed to the linear fallback to keep
+ * URLPattern as the authoritative matcher.
  *
  * The `:id*` "zero-or-more" modifier is caught by the inline-wildcard branch
  * (`includes("*")` with `segment !== "*"`), not by an explicit `endsWith("*")`
@@ -196,6 +198,7 @@ function isComplexSegment(segment: string): boolean {
   if (segment.includes("{") || segment.includes("(")) return true;
   if (segment.includes("*") && segment !== "*") return true;
   if (segment.endsWith("?") || segment.endsWith("+")) return true;
+  if (segment.includes("\\")) return true;
   return false;
 }
 
--- a/http/unstable_route.ts
+++ b/http/unstable_route.ts
@@ -123,18 +123,45 @@ interface IndexedRoute {
 
 interface RouteNode {
   staticChildren: Record<string, RouteNode>;
+  /**
+   * Whether `staticChildren` has at least one entry. Tracked as a flag so
+   * dispatch can skip the substring slice when there are no static children
+   * to look up.
+   */
+  hasStaticChildren: boolean;
   paramChild: RouteNode | null;
   wildcardChild: RouteNode | null;
   routes: IndexedRoute[];
+  /**
+   * Smallest insertion index reachable from this subtree (including its own
+   * `routes`). Used to prune subtrees that cannot improve on the current
+   * best match during dispatch.
+   */
+  minIndex: number;
 }
 
 /**
  * Extract pathname from a URL string without allocating a URL object.
- * Handles both `http://host/path?query` and `http://host/path` forms.
+ *
+ * This is a fast path tailored for `request.url`, which is always a fully
+ * normalized absolute URL with an authority component (e.g. `http://host/p`,
+ * `https://host/p?q`). It is NOT a general-purpose URL parser:
+ *
+ *   - Assumes the URL contains `://` separating scheme and authority. URLs
+ *     without an authority (e.g. `file:/path`, `mailto:x@y`) are not handled.
+ *   - Relies on userinfo containing no literal `/` (per WHATWG URL parsing,
+ *     `/` in userinfo is percent-encoded), so the first `/` after the
+ *     authority delimiter unambiguously starts the path.
+ *
+ * Always call this with a `Request#url` value; do not export.
  */
 function parsePathname(url: string): string {
-  const authorityStart = url.indexOf("//");
-  const pathStart = url.indexOf("/", authorityStart + 2);
+  // Skip past the scheme delimiter `://`. Using indexOf with the full token
+  // avoids the corner case where a scheme contains a colon-prefixed segment
+  // before the authority.
+  const schemeEnd = url.indexOf("://");
+  const authorityStart = schemeEnd === -1 ? 0 : schemeEnd + 3;
+  const pathStart = url.indexOf("/", authorityStart);
   if (pathStart === -1) return "/";
   const qmark = url.indexOf("?", pathStart);
   const hash = url.indexOf("#", pathStart);
@@ -150,9 +177,20 @@ function parsePathname(url: string): string {
  * string, a bare `:param`, or a standalone `*`.
  *
  * Affected syntax:
- *   - Optional / non-capturing groups:  `{.ext}?`  `{foo}`
- *   - Regex-constrained params:         `:id(\d+)`  `:lang(en|fr)`
- *   - Inline wildcards:                 `*.js`  `prefix*`
+ *   - Optional / non-capturing groups:    `{.ext}?`  `{foo}`
+ *   - Regex-constrained params:           `:id(\d+)`  `:lang(en|fr)`
+ *   - Inline wildcards:                   `*.js`  `prefix*`
+ *   - Modifier suffixes on params/groups: `:id?`  `:id+`  `:id*`  `(\d+)?`
+ *
+ * This heuristic relies on URLPattern syntax being well-formed: URLPattern's
+ * own parser rejects `?`, `+`, and `*` as literal characters in pathnames
+ * (e.g. `new URLPattern({ pathname: "/foo+bar" })` throws at construction),
+ * so a segment that ends with one of those characters here is always a
+ * modifier, never a literal.
+ *
+ * The `:id*` "zero-or-more" modifier is caught by the inline-wildcard branch
+ * (`includes("*")` with `segment !== "*"`), not by an explicit `endsWith("*")`
+ * clause. This is intentional — the two checks subsume each other for `*`.
  */
 function isComplexSegment(segment: string): boolean {
   if (segment.includes("{") || segment.includes("(")) return true;
@@ -164,9 +202,11 @@ function isComplexSegment(segment: string): boolean {
 function createNode(): RouteNode {
   return {
     staticChildren: Object.create(null) as Record<string, RouteNode>,
+    hasStaticChildren: false,
     paramChild: null,
     wildcardChild: null,
     routes: [],
+    minIndex: Number.POSITIVE_INFINITY,
   };
 }
 
@@ -179,6 +219,15 @@ function createNode(): RouteNode {
  * Routes with complex URLPattern syntax (regex constraints, optional/non-capturing
  * groups, inline wildcards) fall back to linear matching while preserving
  * insertion order relative to tree-indexed routes.
+ *
+ * The tree is keyed by pathname segments only. Other URLPattern components
+ * (hostname, search, protocol, etc.) are not indexed; routes that constrain
+ * those components are inserted in the tree under their pathname, and the
+ * additional constraints are validated by `pattern.exec()` per request. As a
+ * result, multiple routes sharing the same pathname but differing on
+ * hostname/search/protocol receive no pruning from the tree — every
+ * pathname-matching candidate is tested by the URLPattern matcher in
+ * insertion order, just as `routeLinear` would.
  *
  * @example Usage
  * ```ts ignore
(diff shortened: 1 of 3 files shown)


http/unstable_route.ts
```diff
@@ -96,29 +221,140 @@ export interface Route {
...
-  routeMethod: string | string[] | undefined,
-  requestMethod: string,
-): boolean {
-  if (!routeMethod) return true;
-  if (Array.isArray(routeMethod)) {
-    return routeMethod.some((m) => m.toUpperCase() === requestMethod);
-  }
-  return routeMethod.toUpperCase() === requestMethod;
-}
+export { routeRadix as route };
```
bartlomieju, line 360: This silently changes the semantics of `route()` for existing callers — they go from a simple linear scan to a radix tree with fallback. While behavior *should* be equivalent, if there's any divergence it becomes a regression for current users.

Would it be safer to keep `route` as the linear implementation (preserving current behavior) and offer `routeRadix` as an explicit opt-in? That gives the radix path time to mature without risking surprises for existing consumers.

http/unstable_route.ts
```diff
@@ -96,29 +221,140 @@ export interface Route {
...
-    for (const route of routes) {
-      const match = route.pattern.exec(request.url);
-      if (!match) continue;
-      if (!methodMatches(route.method, request.method)) continue;
-      return route.handler(request, match, info);
+    const pathname = parsePathname(request.url);
+    const segments = parseSegments(pathname);
+    const radixCandidates: IndexedRoute[] = [];
+    collectCandidates(root, segments, 0, radixCandidates);
+    radixCandidates.sort((a, b) => a.index - b.index);
```
bartlomieju, line 320: Every request allocates `radixCandidates` and sorts it, then the merge with `fallbackRoutes` may allocate yet another array. For the common happy path of a single candidate, the sort is wasted work.

This partly explains the ~1.2x slowdown for "static route — first in small table" in the benchmarks. Consider skipping the sort when `radixCandidates.length <= 1`, and/or pre-sorting fallback routes at build time so the merge can be avoided when `radixCandidates` is empty.

http/unstable_route.ts
```diff
@@ -50,11 +50,136 @@ export interface Route {
...
+  paramChild: RouteNode | null;
+  wildcardChild: RouteNode | null;
+  routes: IndexedRoute[];
+}
+
+/**
+ * Extract pathname from a URL string without allocating a URL object.
+ * Handles both `http://host/path?query` and `http://host/path` forms.
+ */
+function parsePathname(url: string): string {
```
bartlomieju, line 135: This custom parser assumes the URL always has an authority (`//`). It works for `http(s)://` URLs from `Request` objects, but would break for schemes like `file:///path` (double `//` is part of the scheme, not authority) or URLs with userinfo containing `/`.

Since this is an internal function only called with `request.url`, it's safe in practice, but a short comment documenting the assumption ("only handles http(s):// URLs from Request objects") would prevent future misuse.

http/unstable_route.ts
```diff
@@ -96,29 +221,140 @@ export interface Route {
...
+  const fallbackRoutes: IndexedRoute[] = [];
+  let insertionCounter = 0;
+
+  function parseSegments(pathname: string): string[] {
+    return pathname.split("/").filter(Boolean);
+  }
+
+  function insert(r: Route): void {
+    const indexed: IndexedRoute = { route: r, index: insertionCounter++ };
+    const segments = parseSegments(r.pattern.pathname);
```
bartlomieju, line 238: The radix tree indexes routes solely by pathname segments, but `URLPattern` can also constrain `hostname`, `search`, `protocol`, etc. Routes like `new URLPattern({ hostname: "api.example.com", pathname: "/data" })` get inserted in the tree under `/data`, so a request to `/data` on a *different* hostname will produce a tree candidate that `pattern.exec()` then rejects.

This is correct (the tree is a pre-filter, `pattern.exec` is authoritative), but it means the tree provides no pruning benefit for multi-component patterns — and may produce *more* candidates than a linear scan would test. Worth documenting this trade-off.

http/unstable_route.ts
```diff
@@ -28,7 +28,7 @@ export type Handler = (
 ) => Response | Promise<Response>;
 
 /**
- * Route configuration for {@linkcode route}.
+ * Route configuration for {@linkcode routeRadix}.
```
bartlomieju, line 31: `Route` is shared by both `routeRadix` and `routeLinear`, so linking it exclusively to `routeRadix` is misleading. Consider linking to both, or reverting to the more general `route`.

http/unstable_route.ts
```diff
@@ -50,11 +50,136 @@ export interface Route {
...
+ *
+ * Affected syntax:
+ *   - Optional / non-capturing groups:  `{.ext}?`  `{foo}`
+ *   - Regex-constrained params:         `:id(\d+)`  `:lang(en|fr)`
+ *   - Inline wildcards:                 `*.js`  `prefix*`
+ */
+function isComplexSegment(segment: string): boolean {
+  if (segment.includes("{") || segment.includes("(")) return true;
+  if (segment.includes("*") && segment !== "*") return true;
+  if (segment.endsWith("?") || segment.endsWith("+")) return true;
```
bartlomieju, line 160: The `endsWith("+")` check correctly catches `:param+` (one-or-more repeating), but it would also flag a hypothetical literal segment like `foo+bar` as complex. In practice URLPattern pathnames won't have unencoded `+` in static segments, so this is fine — but it's worth noting in the doc comment that this heuristic assumes well-formed URLPattern syntax.

Also: does this catch all modifier suffixes? URLPattern also supports `*` as a modifier (e.g., `:id*`), but that's already handled by the `includes("*")` check above. Looks complete to me, just want to confirm that was intentional.

http/unstable_route.ts
esroyo: I absolutely agree :+1:

http/unstable_route.ts
esroyo: You are right. I have changed the approach to avoid the allocations with a more clever pruning (all nodes now track `minIndex`, lowest insertion index in the subtree). It has slightly improved performance indeed :)  d36b5dacaa4a483efc740423152656c51c196503

http/unstable_route.ts
esroyo: Yes. That is now documented explicitly :+1:

Thanks to your comment I found a latent bug with escaped segments (`"/foo\\+bar"`) becoming unreachable :clap:

esroyo:
@bartlomieju Thank you for the insightful comments! I mostly agree with your feedback (except the Array#sort comment perhaps) and have addressed all the points you raised. Thanks so much for your time :heart:

esroyo:
[updated]

```mermaid
---
config:
    themeVariables:
        xyChart:
            plotColorPalette: "#e05c5c, #4a90d9"
---
xychart-beta
    title "AMD Ryzen 7 PRO 6850U — route benchmark"
    x-axis [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
    y-axis "Avg latency (µs)" 0 --> 6

    line [0.92, 1.96, 5.86, 1.63, 5.6, 1.53, 2.72, 1.06, 3.17, 0.93, 1.46, 1.89]
    line [1, 1.12, 1.07, 0.43, 0.43, 1.18, 1.31, 0.54, 1.39, 1.05, 1.44, 1.88]
```

🔴 `linear` &nbsp;&nbsp; 🔵 `radix`

| # | Group | 🔴 linear (µs) | 🔵 radix (µs) |
|---|---|--:|--:|
| 1 | static route — first in small table | 0.92 | 1 |
| 2 | static route — last in small table | 1.96 | 1.12 |
| 3 | static route — last in large table | 5.86 | 1.07 |
| 4 | static route — miss (small table) | 1.63 | 0.43 |
| 5 | static route — miss (large table) | 5.6 | 0.43 |
| 6 | parametric route — single param | 1.53 | 1.18 |
| 7 | parametric route — two params (nested) | 2.72 | 1.31 |
| 8 | parametric route — method mismatch | 1.06 | 0.54 |
| 9 | wildcard route | 3.17 | 1.39 |
| 10 | complex — regex constraint | 0.93 | 1.05 |
| 11 | complex — optional group | 1.46 | 1.44 |
| 12 | complex — inline wildcard with suffix | 1.89 | 1.88 |

lunadogbot (approved):
Bartlomieju's review points are addressed across four commits: `route` keeps linear-scan semantics (`export { routeLinear as route }`), per-request `radixCandidates`/sort replaced with `minIndex` subtree pruning in `RouteNode`, backslash escapes route through the linear fallback, `parsePathname` now documents the `Request#url` scoping, and the `routeRadix` JSDoc explicitly calls out that the tree is keyed by pathname only and other `URLPattern` constraints get validated by `pattern.exec` per request.

The walk's static→param→wildcard recursion order doesn't compromise "first registered wins": `tryRoute` rejects any candidate where `r.index >= state.bestIndex`, and the `param route registered before static route preserves insertion order` step at `http/unstable_route_test.ts:438` exercises exactly that case. Running the full test harness against both `routeRadix` and `routeLinear` (lines 461-462) is the strongest behavioural parity signal.

- nit: `Route`'s `@linkcode route` at `http/unstable_route.ts:31` still names a single function. `Route` is shared by both `routeRadix` and `routeLinear`, so linking to one is still slightly misleading even with `route` aliasing `routeLinear`.

PR #7076, opened 2026-04-01
feat(data-structures/unstable): add `IndexedHeap`

tomas-zijdemans:
I'm working on the cache module and see the need for an Indexed Heap.

The (key, priority) indexed heap with a position map is the standard approach across languages (Go, Rust, Python, Java, C++). The API proposed here is closest to Rust's priority-queue crate.

The existing `BinaryHeap` only supports push/pop; there is no way to remove or re-prioritize an entry without a linear scan. `IndexedHeap` fills that gap by keeping a Map from keys to array positions, at the cost of one extra index entry per element.

### DIFF
--- a/data_structures/unstable_indexed_heap.ts
+++ b/data_structures/unstable_indexed_heap.ts
@@ -1,0 +1,785 @@
+// Copyright 2018-2026 the Deno authors. MIT license.
+// This module is browser compatible.
+
+/** Allows the class to mutate priority internally. */
+interface MutableEntry<K> {
+  readonly key: K;
+  priority: number;
+}
+
+/**
+ * A key-priority pair returned by {@linkcode IndexedHeap} methods.
+ *
+ * Fields are `readonly` to signal that mutating a returned entry has no
+ * effect on the heap.
+ *
+ * @typeParam K The type of the key.
+ */
+export interface HeapEntry<K> {
+  /** The key that identifies this entry in the heap. */
+  readonly key: K;
+  /** The numeric priority of this entry (smaller = higher priority). */
+  readonly priority: number;
+}
+
+/**
+ * Read-only view of an {@linkcode IndexedHeap}. Exposes query and
+ * iteration methods, hiding all methods that modify the heap. Follows
+ * the same pattern as `ReadonlyMap` and `ReadonlySet`.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @typeParam K The type of the keys in the heap.
+ */
+export type ReadonlyIndexedHeap<K> = Pick<
+  IndexedHeap<K>,
+  | "peek"
+  | "peekKey"
+  | "peekPriority"
+  | "has"
+  | "getPriority"
+  | "size"
+  | "isEmpty"
+  | "toArray"
+  | typeof Symbol.iterator
+>;
+
+/** Throws if the priority is NaN, which would silently corrupt the heap. */
+function checkPriority(priority: number): void {
+  if (Number.isNaN(priority)) {
+    throw new RangeError("Cannot set priority: value is NaN");
+  }
+}
+
+/**
+ * A priority queue that supports looking up, removing, and re-prioritizing
+ * entries by key. Each entry is a unique `(key, priority)` pair. The entry
+ * with the smallest priority is always at the front.
+ *
+ * Unlike {@linkcode BinaryHeap}, which only allows popping the top element,
+ * `IndexedHeap` lets you delete or update any entry by its key in
+ * logarithmic time.
+ *
+ * Priorities are plain numbers, always sorted smallest-first. To sort
+ * largest-first instead, negate the priorities.
+ *
+ * | Method                | Time complexity                       |
+ * | --------------------- | ------------------------------------- |
+ * | peek()                | Constant                              |
+ * | peekKey()             | Constant                              |
+ * | peekPriority()        | Constant                              |
+ * | pop()                 | Logarithmic in the number of entries  |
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
Thanks for the PR! The implementation is solid — heap operations are correct, well-optimized, and thoroughly tested. A few points to address:

## Issues

### 1. `pop()` leaks internal `MutableEntry` reference

`peek()` correctly returns a defensive copy (`{ key, priority }`), but `pop()` returns the internal `MutableEntry<K>` object directly, which has a mutable `priority` field. Since the entry is removed from the heap this isn't a correctness bug, but it's inconsistent. Suggest returning `{ key: root.key, priority: root.priority }` from `pop()` too for consistency.

### 2. Destructive `[Symbol.iterator]` is surprising

`[Symbol.iterator]` draining the heap is a significant footgun — passing an `IndexedHeap` to any function that iterates (logging, `Array.from`, `for...of`) silently empties it. I know `BinaryHeap` has the same behavior, but it's worth reconsidering for a new data structure. Options:
- Make `[Symbol.iterator]` non-destructive (snapshot/arbitrary-order iteration) and keep `drain()` for the destructive path
- At minimum, add a prominent warning in the class-level JSDoc

### 3. No `from()` / constructor from iterable

`BinaryHeap` has a `static from()` method. It would be useful to have a way to construct from an iterable of `[key, priority]` pairs — this is a common pattern for collection types.

## Suggestions

- A non-destructive `toArray()` or `entries()` method would be useful for debugging/inspection without draining the heap.
- The stress tests use `Math.random()` without a seed — consider using fixed test data or a seeded PRNG for reproducibility.

tomas-zijdemans:
Thanks for the feedback. Here is what I did:

- `pop()` now returns `{ key: root.key, priority: root.priority }`
- Iterator is now non-destructive. drain() is the explicit destructive sorted path. Note: this intentionally diverges from `BinaryHeap`, which has a destructive iterator. A follow-up to align `BinaryHeap` could make sense.
- `from()` added
- `toArray()`added
- stress tests now use seeded mulberry32 PRNG

bartlomieju (commented):
Nice work on the updates — the non-destructive iterator, `from()`, `toArray()`, and seeded PRNG all look great. A few smaller things:

### `from()` with iterable is O(n log n)

The iterable path calls `push()` per entry, which is O(n log n). Floyd's bottom-up heapify can do this in O(n). Not a blocker at all for an unstable API, but might be worth a `// TODO` or a note for future optimization — especially since the `from(IndexedHeap)` path already takes the fast route by copying the internal array directly.

### `drain()` partial consumption

If a caller breaks out of `for...of heap.drain()` early, the heap is left partially drained. This is standard generator behavior, but it could surprise users. A small JSDoc note like "If iteration is abandoned early, the heap retains the remaining entries" would help.

### `from()` ArrayLike detection

```ts
const entries = "length" in collection
  ? Array.from(collection)
  : collection;
```

This works, but `"length" in collection` is a duck-type check that could match unexpected objects. Since the TypeScript signature already constrains the input, you could simplify to just `Array.from(collection as ArrayLike<...>)` for the non-`IndexedHeap` path — `Array.from` handles both iterables and array-likes natively.

tomas-zijdemans:
Thanks. Changed it now:

- `from()` now uses Floyd's heapify
- Removed the "length" in collection duck-type check and replaced it with a single `Array.from(collection)`
- Added a JSDoc note on `drain()` about what happens if you break out of iteration early 
- Added tests

tomas-zijdemans:
@bartlomieju 

Pushed a minor refactor:
- `update` / `pushOrUpdate` now returns early when the new priority equals the old, and dispatch directly to `#siftUp` or `#siftDown` based on the direction of change. Same for `delete`, using the removed entry's priority vs. the replacement's.
- Renamed the internal `assertValidPriority` to `checkPriority`
- Added missing entries in the complexity table and added a note that mutating during `[Symbol.iterator]` is not supported.

bartlomieju (approved):
LGTM!

data_structures/unstable_indexed_heap.ts
```diff
@@ -0,0 +1,773 @@
...
+      heap.#index.set(key, i);
+    }
+    for (let i = (heap.#data.length >>> 1) - 1; i >= 0; i--) {
+      heap.#siftDown(i);
+    }
+    return heap;
+  }
+
+  /** Bubble the entry at `pos` up toward the root while it is smaller than its parent. */
+  #siftUp(pos: number): number {
```
bartlomieju, line 213: \`Deque.from()\` explicitly validates the collection argument and throws a descriptive \`TypeError\` before proceeding. Here, passing an invalid value (e.g. \`null\`, a number) at the JS level falls through to \`Array.from(collection)\` which throws a generic error without mentioning \`IndexedHeap\`. Consider adding the same kind of guard for consistency:

```ts
if (
  collection === null || collection === undefined ||
  typeof collection !== "object" && typeof collection !== "string" ||
  !(Symbol.iterator in Object(collection) || "length" in Object(collection))
) {
  throw new TypeError(
    "Cannot create an IndexedHeap: the 'collection' parameter is not iterable or array-like",
  );
}
```

bartlomieju:
Also needs to resolve conflicts

data_structures/unstable_indexed_heap.ts
tomas-zijdemans: Good catch!

tomas-zijdemans:
I added the following:
- Copied the runtime guard from `Deque` as suggested in the review
- Changed the Error types to better align with the rest of the data-structures module

PR #7082, opened 2026-04-02
feat(cache): add unified `Cache`

tomas-zijdemans:
This PR replaces `LruCache` and `TtlCache` with a single `Cache<K, V>` that unifies LRU eviction, TTL expiration, stale-while-revalidate, and load-through (`getOrLoad`).

The existing caches extend `Map`, so inherited methods bypass LRU/TTL logic entirely. Both subclass `Map`, making it structurally impossible to compose LRU with TTL.

`Cache<K, V>` uses composition over inheritance: it owns a `Map` for storage and delegates to `IndexedHeap` for deadline-ordered expiration. Mode is determined by options, not class choice, and a discriminated union on `CacheOptions` makes illegal combinations compile-time errors.

Beyond fixing the structural issues, this introduces capabilities that didn't exist before:
- Combined LRU + TTL in a single cache
- Renames the base type to `CacheLike`, a more natural type name
- `getOrLoad(key, loader)` with automatic in-flight deduplication
- Stale-while-revalidate with background refresh, error handling, and configurable soft/hard TTL thresholds
- Per-entry TTL overrides and sliding expiration
- Single heap-driven timer replacing per-entry `setTimeouts`
- Hit/miss/eviction stats, typed ejection reasons, and `Symbol.dispose`

Since `@std/cache` is still experimental (0.2.2), this is the right time to make a clean break before stabilization locks things in.

Bonus: The new `Cache` has significant performance improvements for real-life workflows (e.g. in my testing, write-heavy workloads (set, eviction) are 4-22x faster. hot-key reads are 55x faster thanks to a linked-list)


bartlomieju:
I'll hold on reviewing this one until #7076 lands.

cache/cache.ts
```diff
@@ -0,0 +1,1260 @@
...
+  refresh?: undefined;
+  /** Must be `undefined` for non-SWR caches. */
+  onRefreshError?: undefined;
+}
+
+/**
+ * {@linkcode Cache} options that enable TTL expiration.
+ *
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
```
bartlomieju, line 65: The three option interfaces (`CacheOptionsBase`, `CacheOptionsTtl`, `CacheOptionsSwr`) each duplicate the `maxSize` and `onRemove` fields with full JSDoc. Consider extracting a shared base interface that all three extend, and only declaring the TTL/SWR-specific fields in their respective interfaces. That would cut ~40 lines of duplication and make future changes to the shared fields (e.g. adding a new common option) a single-site edit.

cache/cache.ts
```diff
@@ -0,0 +1,1260 @@
...
+    while (!heap.isEmpty()) {
+      if (heap.peekPriority()! > now) break;
+      const top = heap.pop()!;
+      const entry = this.#data.get(top.key)!;
+      this.#data.delete(top.key);
+      this.#unlink(entry);
+      this.#stats.expirations++;
+      if (!this.#onRemove) continue;
+      this.#removingDepth++;
+      try {
```
bartlomieju, line 623: `this.#data.get(top.key)!` assumes the entry still exists when the timer fires. If a bug or future code path causes the heap and `#data` to go out of sync, this will throw at runtime.

Consider guarding defensively:
```ts
const entry = this.#data.get(top.key);
if (entry === undefined) continue;
```
This also makes the invariant explicit: "the heap may contain stale keys that were already removed."

bartlomieju (changes requested):
This needs a rebase — there are conflicts with main that need resolving.

Two inline comments below.

tomas-zijdemans:
Thanks,
- Extracted a new exported `CacheOptionsShared<K, V>` interface as suggested with `maxSize` and `onRemove`. `CacheOptionsBase`, `CacheOptionsTtl`, and `CacheOptionsSwr` now extend it. 
- Defensive guard added in `#onTimer` as suggested

tomas-zijdemans:
Renamed `MemoizationCache` to `CacheLike` and moved it to the base file (`cache.ts`)

tomas-zijdemans:
Not sure if this should have "BREAKING" in the PR title, since the module is not stable 🤷 . Changing it based on the comment here: https://github.com/denoland/std/pull/7117#issuecomment-4542944235

Ready for review! 😁

PR #7083, opened 2026-04-03
feat(data-structures/unstable): add `includes()` and `removeAt()` to `Deque`

tomas-zijdemans:
- Add `includes()` method (matching `Array.prototype.includes`), and `removeAt(index)` for removing elements by index with negative-index support.
- Add a time-complexity table to the Deque class JSDoc to match `BinaryHeap` docs.
- Extract `#maybeGrow()` and `#removeAtUnchecked()` private helpers to reduce duplication

### DIFF
--- a/data_structures/unstable_deque.ts
+++ b/data_structures/unstable_deque.ts
@@ -753,7 +753,7 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
 
     const mapped: U[] = options?.map
       ? Array.from(unmappedValues, options.map, options.thisArg)
-      : Array.from(unmappedValues as unknown as Iterable<U>);
+      : Array.from(unmappedValues as ArrayLike<U> & Iterable<U>);
 
     const capacity = nextPowerOfTwo(mapped.length);
     result.#buffer = new Array(capacity);
--- a/data_structures/unstable_deque.ts
+++ b/data_structures/unstable_deque.ts
@@ -631,15 +631,11 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
    * @returns An array containing the deque's elements in order.
    */
   toArray(): T[] {
-    const len = this.#length;
-    if (len === 0) return [];
-    const head = this.#head;
-    const end = head + len;
-    if (end <= this.#capacity) {
-      return this.#buffer.slice(head, end) as T[];
+    const result = new Array<T>(this.#length);
+    for (let i = 0; i < this.#length; i++) {
+      result[i] = this.#buffer[(this.#head + i) & this.#mask] as T;
     }
-    return (this.#buffer.slice(head) as T[])
-      .concat(this.#buffer.slice(0, end - this.#capacity) as T[]);
+    return result;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


data_structures/unstable_deque.ts
```diff
@@ -691,10 +783,23 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
    *
    * @returns An iterator yielding elements from front to back.
    */
-  *[Symbol.iterator](): IterableIterator<T> {
-    for (let i = 0; i < this.#length; i++) {
-      yield this.#buffer[(this.#head + i) & this.#mask] as T;
-    }
+  [Symbol.iterator](): IterableIterator<T> {
+    let i = 0;
+    const length = this.#length;
```
bartlomieju: This captures `length`, `head`, `buffer`, and `mask` at iterator creation time. The old generator-based iterator read `this.#length` and `this.#head` live on each `yield`, so mutations during iteration would at least see current state.

The new iterator silently reads stale values if the deque is mutated mid-iteration. This is a behavior change worth documenting — or consider reading from `this` live via closure (at the cost of some of the perf gain).

data_structures/unstable_deque.ts
```diff
@@ -553,11 +631,15 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
    * @returns An array containing the deque's elements in order.
    */
   toArray(): T[] {
-    const result = new Array<T>(this.#length);
-    for (let i = 0; i < this.#length; i++) {
-      result[i] = this.#buffer[(this.#head + i) & this.#mask] as T;
+    const len = this.#length;
+    if (len === 0) return [];
+    const head = this.#head;
+    const end = head + len;
```
bartlomieju: The `as T[]` cast is load-bearing here — `#buffer` is `(T | undefined)[]` but the active region only contains `T` values. Works correctly in practice, just noting the cast isn't purely cosmetic.

data_structures/unstable_deque.ts
```diff
@@ -735,8 +852,17 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
 
   static #copyBuffer<T>(source: Deque<T>, capacity: number): (T | undefined)[] {
     const buffer = new Array<T | undefined>(capacity);
-    for (let i = 0; i < source.#length; i++) {
-      buffer[i] = source.#buffer[(source.#head + i) & source.#mask];
+    const head = source.#head;
+    const len = source.#length;
+    const srcBuf = source.#buffer;
+    const end = head + len;
+    if (end <= source.#capacity) {
```
bartlomieju: Nit: this drops the `& mask` that the original code used. It's correct because the `end <= capacity` guard above guarantees no wrap, but that invariant is non-obvious. A brief comment like `// no wrap: head+i is always in bounds` would help.

data_structures/unstable_deque.ts
```diff
@@ -655,13 +737,23 @@ export class Deque<T> implements Iterable<T>, ReadonlyDeque<T> {
         return result;
       }
       unmappedValues = collection.toArray();
+    } else if (!options?.map && Array.isArray(collection)) {
```
bartlomieju: Minor: when this fast path is taken, `options?.thisArg` is silently ignored. That's fine (no `map` means `thisArg` is meaningless), but the growing number of branches in `from()` is getting hard to follow. Consider adding a comment explaining the fast-path conditions.

bartlomieju (commented):
Thanks for the PR! `includes()` and `removeAt()` are good additions, and the refactoring (`#maybeGrow`, `#removeAtUnchecked`) is clean. The complexity table is a nice touch.

A few high-level points beyond the inline comments:

### PR scope is broad

This bundles four concerns: new features (`includes`, `removeAt`), docs (complexity table), perf optimizations (iterator rewrite, `toArray`/`#copyBuffer`/`from()` changes), and refactoring (`#maybeGrow`, `#removeAtUnchecked`). I'd suggest splitting the perf optimizations into a separate PR — they have different risk profiles and the iterator behavior change in particular deserves focused review.

### Codecov flags 4 uncovered lines

The coverage report shows 3 missing + 1 partial line in `unstable_deque.ts`. Worth checking which lines those are and adding test coverage.

tomas-zijdemans:
Thanks. I've reverted the perf tweaks. Will consider a secondary PR for those.

bartlomieju (approved):

PR #7084, opened 2026-04-04
BREAKING(async/unstable): change `Lazy.peek()` return type, add AbortSignal support

tomas-zijdemans:
- AbortSignal support added to `get()` 
- Fixed a subtle bug where `get()` silently swallowed rejections by attaching a handler directly to the returned promise, suppressing unhandledrejection for uncaught callers.
- `peek()` now returns  a discriminated union matching the `Channel.tryReceive()` convention. This resolves the type-level ambiguity when T includes undefined. 
- Error-reset semantic documented

### DIFF
--- a/async/unstable_lazy.ts
+++ b/async/unstable_lazy.ts
@@ -31,7 +31,7 @@ export interface LazyGetOptions {
  * @example Concurrent deduplication
  *
  * ```ts
- * import { Lazy } from "@std/async/lazy";
+ * import { Lazy } from "@std/async/unstable-lazy";
  * import { assertEquals } from "@std/assert";
  *
  * let initCount = 0;
@@ -49,7 +49,7 @@ export interface LazyGetOptions {
  * @example Composing with retry
  *
  * ```ts ignore
- * import { Lazy } from "@std/async/lazy";
+ * import { Lazy } from "@std/async/unstable-lazy";
  * import { retry } from "@std/async/retry";
  *
  * const db = new Lazy(() =>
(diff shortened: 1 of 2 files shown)


async/unstable_lazy.ts
```diff
@@ -75,37 +94,71 @@ export class Lazy<T> {
...
+   *
+   * @param options Optional settings for this call.
    * @returns The cached or newly initialized value.
    */
-  get(): Promise<T> {
-    if (this.#promise !== undefined) {
-      return this.#promise;
+  get(options?: LazyGetOptions): Promise<T> {
+    const signal = options?.signal;
+    if (signal?.aborted) return Promise.reject(signal.reason);
```
bartlomieju, line 120: Bug: this rejects even when the value is already cached:

```ts
await lazy.get(); // init succeeds, value cached
await lazy.get({ signal: AbortSignal.abort() }); // rejects!
```

Since no waiting is needed when the value is already settled, the signal should be irrelevant. The `this.#settled` fast-path should come before the abort check:

```ts
get(options?: LazyGetOptions): Promise<T> {
    if (this.#settled) return Promise.resolve(this.#value as T);
    const signal = options?.signal;
    if (signal?.aborted) return Promise.reject(signal.reason);
    ...
}
```

The existing test "signal is ignored after successful initialization" doesn't catch this because it uses a non-aborted controller.

tomas-zijdemans: Thanks, good catch. Added the line as suggested.

bartlomieju (approved):
Thanks!

PR #7086, opened 2026-04-05
feat(async/unstable): add AbortSignal and toReadableStream to `Channel`

tomas-zijdemans:
- AbortSignal support for send() and receive()
- Cancellation-aware internal queues
- Improved tryReceive() return type
- toReadableStream() method
- close() overload signatures
- Minor optimizations

### DIFF
--- a/async/unstable_channel.ts
+++ b/async/unstable_channel.ts
@@ -244,6 +244,9 @@ export class Channel<T>
         const onAbort = () => {
           node.cancelled = true;
           rej(signal.reason);
+          while (this.#senders.peekFront()?.cancelled) {
+            this.#senders.popFront();
+          }
         };
         signal.addEventListener("abort", onAbort, { once: true });
         node.res = () => {
@@ -315,6 +318,9 @@ export class Channel<T>
         const onAbort = () => {
           node.cancelled = true;
           rej(signal.reason);
+          while (this.#receivers.peekFront()?.cancelled) {
+            this.#receivers.popFront();
+          }
         };
         signal.addEventListener("abort", onAbort, { once: true });
         node.res = (value: T) => {
--- a/async/unstable_channel.ts
+++ b/async/unstable_channel.ts
@@ -10,14 +10,12 @@ interface SenderNode<T> {
   value: T;
   res: () => void;
   rej: (reason: unknown) => void;
-  cancelled: boolean;
 }
 
 /** Internal node for the FIFO receiver waiting queue. */
 interface ReceiverNode<T> {
   res: (value: T) => void;
   rej: (reason: unknown) => void;
-  cancelled: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
When a signal aborts, the node stays in the `Deque` with `cancelled: true` and is only skipped when `#nextSender()`/`#nextReceiver()` reaches it. This is a pragmatic O(1) abort vs O(n) eager removal tradeoff, but for long-lived channels with frequent cancellations (e.g., timeout-based polling), cancelled nodes accumulate without bound until a successful delivery drains past them. Worth at least documenting this behavior.

tomas-zijdemans:
Good point. Added the front-drain on abort so cancelled nodes are popped eagerly. Covers the timeout-polling pattern.

bartlomieju (commented):
### Cancelled node accumulation is only partially mitigated

The front-drain in `onAbort` only pops cancelled nodes from the front of the deque:

```ts
while (this.#senders.peekFront()?.cancelled) {
  this.#senders.popFront();
}
```

This doesn't help when a non-cancelled node sits at the front:

```ts
const p0 = ch.receive(); // stays at front, never cancelled
for (let i = 0; i < 10000; i++) {
  const c = new AbortController();
  ch.receive({ signal: c.signal });
  c.abort(); // cancelled node stuck behind p0
}
// 10,000 cancelled nodes in the deque until p0 resolves
```

Consider tracking a cancelled count and compacting (or rebuilding) the deque when the ratio exceeds a threshold, e.g. when `cancelledCount > deque.length / 2`. The current front-drain is better than nothing but doesn't bound accumulation.

### `onAbort` listener cleanup is subtly correct but fragile

In both `send()` and `receive()`, the `onAbort` handler calls the raw `rej` from the Promise constructor:

```ts
const onAbort = () => {
  node.cancelled = true;
  rej(signal.reason); // raw rej, not node.rej
  // ...
};
```

Meanwhile `node.rej` is overwritten to also call `signal.removeEventListener("abort", onAbort)`. Since `onAbort` calls raw `rej` instead of `node.rej`, the listener removal wrapped in `node.rej` is bypassed. This works because `{ once: true }` auto-removes the listener after firing, but it's fragile — if someone later changes `once: true` to manual removal, this breaks silently. Consider calling `node.rej(signal.reason)` from `onAbort` instead (guarding against re-entrancy with the `cancelled` flag), or at minimum add a comment explaining the `once: true` dependency.

tomas-zijdemans:
Thanks, that's very good findings! I think I found an elegant and optimal solution:

**Cancelled node accumulation: fixed via eager `Deque.removeFirst()`**

Instead of marking nodes as cancelled and draining from the front, `onAbort` now calls `this.#senders.removeFirst((n) => n === node)` (and likewise for receivers) to remove the exact node from any position in the deque.

This was chosen over threshold-compaction and a linked-list rewrite. `removeFirst` is worst-case O(n) per abort, but in the timeout-polling pattern, Deque's shift-the-shorter-side strategy makes it effectively O(1). It also eliminates the cancelled flag entirely, which simplifies `#nextSender()/#nextReceiver()` to plain `popFront()` and removes the skip logic in `close()`.

Two new tests cover the exact scenario from the review (non-cancelled node at the front, cancelled nodes behind it).

**onAbort listener cleanup: resolved by the same change**

`onAbort` now calls `node.rej(signal.reason)` instead of raw `rej(signal.reason)`, so listener cleanup goes through the wrapper on all three settlement paths.

bartlomieju (approved):

PR #7087, opened 2026-04-07
refactor(http/unstable): improve Cache Control parsing and types

tomas-zijdemans:
- BREAKING: CacheControl is now a merged mapped type instead of a union
- staleIfError added to RequestCacheControl (per RFC 5861)
- Handle backslash escapes inside quoted strings in splitDirectives
- Normalize no-cache="" and private="" to true instead of []
- Hoist digit regex to module scope


bartlomieju (commented):
Nice bug fixes and test coverage. The backslash escape fix, empty field-name normalization, and `staleIfError` addition are all correct. Main concerns are around the type restructuring and Set removal — see inline comments.

http/unstable_cache_control.ts
```diff
@@ -92,7 +86,18 @@ export interface RequestCacheControl extends CacheControlBase {
...
+  noStore?: true;
+  /** When present, the cache must not transform the payload. */
+  noTransform?: true;
+  /**
+   * Maximum age in seconds. In a request: accept responses whose age is no
+   * greater than this. In a response: the response is stale after this many seconds.
+   */
+  maxAge?: number;
+  /** Allow use of stale response if revalidation fails (seconds). */
+  staleIfError?: number;
```
bartlomieju: `noStore`, `noTransform`, `maxAge`, and `staleIfError` are now duplicated across `RequestCacheControl` and `ResponseCacheControl` (including JSDoc). The removed `CacheControlBase` existed precisely to avoid this — if any shared field needs a type or doc change, it now has to be done in two places.

For an API heading toward stabilization, I'd prefer keeping a shared base interface to reduce maintenance burden.

http/unstable_cache_control.ts
```diff
@@ -118,22 +123,16 @@ export interface ResponseCacheControl extends CacheControlBase {
...
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  */
-export type CacheControl = RequestCacheControl | ResponseCacheControl;
-
-/** Union of every directive field across both interfaces, using the widest type
- * for fields that appear in both (e.g. `noCache` becomes `true | string[]`).
- * Avoids per-field type assertions during parsing and formatting. */
-type AllCacheControlFields = {
+export type CacheControl = {
```
bartlomieju, line 134: Good change — the mapped type is more ergonomic than the old union (`RequestCacheControl | ResponseCacheControl`), which required type narrowing to access non-shared fields like `maxStale`.

Worth noting this is a breaking change though: code that pattern-matched on the union or imported `CacheControlBase` will break. The PR title says `fix` but the body says "Breaking type changes" — might want to align these (e.g. `feat` or `refactor!`).

http/unstable_cache_control.ts
```diff
@@ -233,12 +235,10 @@ export function parseCacheControl(value: string | null): CacheControl {
...
 
     // RFC 9111 §4.2.1: when a directive appears more than once, use the first
-    // occurrence. Track seen directive names to skip subsequent duplicates.
-    if (seen.has(name)) continue;
-    seen.add(name);
-
+    // occurrence. Skip if the property is already set on result.
     switch (name) {
       case "max-age":
+        if (result.maxAge !== undefined) continue;
```
bartlomieju: Replacing the 3-line `Set` approach with per-property `if (result.X !== undefined) continue;` guards on every case branch (17 new lines) adds a lot of repetitive boilerplate. The `Set` was uniform and impossible to forget when adding a new directive; with this approach, every new `case` must remember to add the guard.

`Set.has()` on a set of <20 entries is already O(1) in practice, and the allocation is one-time per parse call. Is there a benchmark showing a measurable improvement here? If not, I'd lean toward keeping the `Set` for clarity.

http/unstable_cache_control.ts
```diff
@@ -318,7 +343,6 @@ export function parseCacheControl(value: string | null): CacheControl {
         result.staleIfError = parseNonNegativeInt(rawValue, name);
         break;
       default:
-        // Unknown directives are ignored per RFC 9111 §5.2.3.
         continue;
```
bartlomieju, line 334: The previous code had a useful comment here:
```ts
// Unknown directives are ignored per RFC 9111 §5.2.3.
```
Worth keeping — a bare `continue` in a `default` case doesn't communicate the intent or RFC basis.

http/unstable_cache_control.ts
```diff
@@ -169,7 +170,9 @@ function splitDirectives(value: string): string[] {
   let inQuotes = false;
   for (let i = 0; i < value.length; i++) {
     const c = value.charCodeAt(i);
-    if (c === 34 /* " */) {
+    if (c === 92 /* \ */ && inQuotes) {
```
bartlomieju, line 172: Good fix. Correctly handles `\"` inside quoted strings so it doesn't toggle `inQuotes`. The test with `'no-cache="x-\\"header"'` covers this well.

http/unstable_cache_control.ts
```diff
@@ -247,61 +247,85 @@ export function parseCacheControl(value: string | null): CacheControl {
...
         break;
-      case "no-cache":
-        result.noCache = rawValue === undefined
-          ? true
+      case "no-cache": {
+        if (result.noCache !== undefined) continue;
+        const noCacheFields = rawValue === undefined
+          ? undefined
           : parseFieldNames(rawValue);
+        result.noCache =
```
bartlomieju, line 268: Good fix — `no-cache=""` and `private=""` now correctly normalize to `true` instead of `[]`. An empty field-name list is semantically equivalent to bare `no-cache`/`private`, so this avoids a confusing representation.

http/unstable_cache_control.ts
```diff
@@ -62,16 +64,8 @@ export interface CacheControlBase {
...
-/**
- * Cache-Control directives for requests (e.g. from a client).
- *
- * @experimental **UNSTABLE**: New API, yet to be vetted.
- *
- * @see {@link https://www.rfc-editor.org/rfc/rfc9111#section-5.2.1}
- */
-export interface RequestCacheControl extends CacheControlBase {
+  /** Allow use of stale response if revalidation fails (seconds). */
+  staleIfError?: number;
```
bartlomieju, line 68: Correct per RFC 5861 §4 — `stale-if-error` is valid in both request and response contexts. Good addition.

tomas-zijdemans:
Agreed. Here's what I did:

- Restored CacheControlBase
- Restored Set-based duplicate tracking for readability
- Restored RFC comment in default case

bartlomieju (commented):
The bug fixes here look great — backslash escape handling, empty field-name normalization, and `staleIfError` on `RequestCacheControl` are all correct.

However, it looks like the changes you mentioned in your comment (restoring `CacheControlBase`, `Set`-based dedup, and the RFC comment in `default`) haven't been pushed yet — the diff still shows the old state. Could you push those?

tomas-zijdemans:
Oh my.. 🫣 
Really sorry about that. Pushed now.

bartlomieju (approved):
Previous round's feedback was addressed. The remaining changes are solid:

- **`staleIfError` to `CacheControlBase`**: Correct per RFC 5861 §3-4 (valid in both request and response).
- **`CacheControl` mapped type**: Cleaner than the union — no more type assertions needed to access non-shared fields. The internal `AllCacheControlFields` alias is correctly removed since the public type now serves the same purpose.
- **Backslash escape handling in `splitDirectives`**: Correctly skips any character after `\` while inside quotes. Edge cases (trailing backslash, `\\"` sequences) all handled correctly.
- **`no-cache=""` / `private=""` normalization**: Correct — an empty field-name list is semantically equivalent to bare `no-cache`/`private`.

PR #7088, opened 2026-04-07
feat(collections/unstable): accept `Iterable` in interleave

tomas-zijdemans:
`interleave` now widens the API to `Iterable` without extra cost for array callers. This is in line with the general direction of the collections module. `interleave` is very close to `zip` that does not support `Iterable`, but looking at the module as a whole, I think it's `zip` that needs to change.

Also added a two-array fast path (no inner loop, no per-pair bounds check) and a branch for equal-length inputs where the length guard is unnecessary. This should help with two very common use cases. The general unequal-length path is unchanged.

### DIFF
--- a/collections/unstable_interleave.ts
+++ b/collections/unstable_interleave.ts
@@ -5,7 +5,7 @@
  * Returns all elements from the given iterables in round-robin order.
  * Unlike {@linkcode zip}, which stops at the shortest iterable and returns
  * tuples, `interleave` continues until all input iterables are exhausted and
- * returns a flat array.
+ * returns a flat array. All input iterables are consumed eagerly.
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
@@ -86,18 +86,6 @@ export function interleave<T extends unknown[]>(
     return result;
   }
 
-  // Fast path for equal-length arrays
-  if (minLength === maxLength) {
-    let k = 0;
-    for (let i = 0; i < maxLength; ++i) {
-      for (let j = 0; j < arrayCount; ++j) {
-        result[k++] = arrays[j]![i] as T[number];
-      }
-    }
-    return result;
-  }
-
-  // General case
   let k = 0;
   for (let i = 0; i < maxLength; ++i) {
     for (let j = 0; j < arrayCount; ++j) {
--- a/collections/unstable_interleave.ts
+++ b/collections/unstable_interleave.ts
@@ -2,18 +2,18 @@
 // This module is browser compatible.
 
 /**
- * Merges multiple arrays into a single array by round-robin picking one
+ * Merges multiple iterables into a single array by round-robin picking one
  * element from each source in turn. Unlike {@linkcode zip}, which stops at
  * the shortest array and produces tuples, `interleave` continues through
  * all elements and returns a flat array.
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
- * @typeParam T Tuple of element types, one per input array; result is
+ * @typeParam T Tuple of element types, one per input iterable; result is
  * `T[number][]`.
  *
- * @param arrays The arrays to interleave.
- * @returns A new array containing elements from all input arrays in
+ * @param iterables The iterables to interleave.
+ * @returns A new array containing elements from all input iterables in
  * round-robin order.
  *
  * @example Basic usage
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Widening from `ReadonlyArray` to `Iterable` is a nice improvement. A few concerns:

### The two-array fast path has subtle ordering differences for unequal-length inputs

The general path interleaves in round-robin order: it picks one from each array per round, skipping exhausted arrays. The two-array fast path emits all remaining elements from `a` first, then all remaining from `b`. For equal-length inputs these are equivalent, but for unequal-length inputs the behavior differs from the general path.

Example with 3+ arrays (general path): `interleave([1,2,3], ["a"], [true, false])` produces `[1, "a", true, 2, false, 3]` — continuing round-robin after `"a"` is exhausted.

With the two-array fast path: `interleave([1,2,3], ["a"])` would produce `[1, "a", 2, 3]` — same result as general path here since there are only two arrays. Actually on closer inspection, for two arrays the tail elements only come from one side, so it matches. This is fine.

### Missing test coverage for the fast paths

The new tests only cover iterables (Set, generator). There are no tests specifically exercising:
- The two-array fast path with unequal lengths (to verify it matches general-path semantics)
- The equal-length fast path with 3+ arrays

These fast paths should have explicit tests confirming identical output to the general path.

### `Array.isArray` check vs `Array.from` for all inputs

The approach is sensible — avoid copying arrays that are already arrays. One edge case: a subclass of `Array` would pass `Array.isArray` but might have overridden iteration behavior. This is extremely unlikely in practice, just noting it.

collections/unstable_interleave.ts
```diff
@@ -37,24 +36,69 @@
  *   [1, "a", true, 2, "b", 3],
  * );
  * ```
+ *
+ * @example With iterables
+ * ```ts
+ * import { interleave } from "@std/collections/unstable-interleave";
```
bartlomieju, line 42: Eagerly materializing all iterables into arrays upfront means memory usage is always O(total elements) even if inputs are lazy/streaming. For `interleave` this is unavoidable since the result is an array anyway, so this is fine — but worth noting in the JSDoc that all iterables are fully consumed immediately.

collections/unstable_interleave.ts
```diff
@@ -37,24 +36,69 @@
...
+
   let maxLength = 0;
+  let minLength = Infinity;
   let totalLength = 0;
   for (let i = 0; i < arrayCount; ++i) {
     const len = arrays[i]!.length;
     totalLength += len;
     if (len > maxLength) maxLength = len;
+    if (len < minLength) minLength = len;
   }
```
bartlomieju, line 66: Nit: the two-array fast path avoids the inner `j` loop and the per-element bounds check `i < arrays[j].length`, which is a nice micro-optimization. But the equal-length fast path (below) only skips the bounds check — it still has the same double loop structure as the general case. The performance difference there is likely negligible (a single branch per inner iteration). Consider whether the added code complexity is worth it for that path.

collections/unstable_interleave_test.ts
```diff
@@ -79,3 +79,28 @@ Deno.test({
...
+
+Deno.test({
+  name: "interleave() handles Set iterables",
+  fn() {
+    assertEquals(
+      interleave(new Set([1, 2, 3]), ["a", "b", "c"]),
+      [1, "a", 2, "b", 3, "c"],
+    );
+  },
+});
```
bartlomieju, line 136: Missing: a test for the two-array fast path with unequal-length arrays, and a test for 3+ equal-length arrays (to exercise the equal-length fast path). These are important to verify the fast paths produce identical output to the general case.

tomas-zijdemans:
Agree with all points. Here is what I changed:

- Added test for two-array fast path
- Added tests for 3+ equal-length arrays
- Added JSDoc note about eager consumption
- Dropped the equal-length fast path

bartlomieju (commented):
The `Iterable` widening looks good — clean API improvement. Two minor items:

### `minLength` is computed for all calls but only used by the two-array fast path

Move it inside the `if (arrayCount === 2)` block (or just compute it there as `Math.min(a.length, b.length)`). No need to track it for 3+ inputs.

### Missing test: first array shorter than second

The `"handles second array longer"` test has `[1, 2, 3], ["a"]` (first is longer). Add a test with the first array shorter, e.g. `interleave([1], ["a", "b", "c"])` → `[1, "a", "b", "c"]`, to exercise the second tail loop in the fast path.

tomas-zijdemans:
Good catch. I dropped `minLength` from the shared loop and now compute it inline when needed.

For the test "first array shorter than second" is already available. See line 30 in the test file ☺️

bartlomieju (approved):

PR #7089, opened 2026-04-07
feat(crypto): stabilize AES-GCM encryption

tomas-zijdemans:


bartlomieju (commented):
I'm on a fence about this one. Now that we have solid impl of Web Crypto and `node:crypto` in Deno, I'm thinking that maybe we should deprecate this module.

tomas-zijdemans:
I think this wrapper is still useful even with solid Web Crypto support. The raw `crypto.subtle.encrypt` API requires you to:

- Generate a random nonce yourself every time
- Pack nonce + ciphertext + tag into a single buffer
- Unpack and slice them apart on decrypt
- Remember the right constants (12-byte nonce, 128-bit tag)

Easy to get wrong, so nice to have a helper for it. In my view the module should provide useful helpers on top of Web Crypto, but I might be wrong of course :)

bartlomieju (approved):

PR #7092, opened 2026-04-12
feat(data-structures): stabilize `Deque`

tomas-zijdemans:
I consider `Deque` to be feature-complete and the external API stable. We could add more functions like `some`, `every` or `indexOf`, but these are all convenience helpers that are addable post-stabilization as non-breaking additions, so not a reason to hold.

Follow-up to #7083, which deferred perf changes out of scope. This PR reapplies only the subset that is a pure internal refactor: `toArray`, `includes`, and `#copyBuffer` linearized into a two-pass walk. The iterator rewrite and Deque.from fast path from the original are intentionally omitted; they can be proposed separately if warranted.

### DIFF
--- a/async/unstable_channel.ts
+++ b/async/unstable_channel.ts
@@ -1,7 +1,7 @@
 // Copyright 2018-2026 the Deno authors. MIT license.
 // This module is browser compatible.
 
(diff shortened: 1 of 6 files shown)


bartlomieju (approved):
LGTM

PR #7093, opened 2026-04-12
fix(async/unstable): harden `pooledMapSettled` error handling and abort

tomas-zijdemans:
Harden `pooledMapSettled()` error handling, abort responsiveness, and resource cleanup.

**Bug Fixes**
- Source iterable errors now propagate. Previously silently swallowed; now the iterator rejects with the original error after yielding all in-flight settled results.
- Abort signal interrupts slow/stalled async sources. Each `it.next()` is raced against the signal, so cancellation is immediate instead of waiting for the source to resume.
- Deterministic error priority. Abort reason wins when signal is already aborted at source failure; source error wins otherwise.

**Performance**
- `executing` now uses `Set` instead of `Array`. This eliminates O(n) `indexOf/splice` on every task completion.
- Single abort listener per call instead of creating/tearing down one per `raceWithSignal()` invocation.

**Resource Cleanup**
- Fallback generator path now cancels the reader and releases the lock in `finally` on early consumer `break`.
- `it.return()` called on abort so source iterables can run cleanup.
- `.catch(() => {})` guards on fire-and-forget promises prevent unhandled-rejection noise.

**Docs & Tests**
- Updated `@throws` JSDoc to document both rejection modes.
- Added 9 test cases covering: source error propagation, error-vs-abort priority, abort interrupting slow/stalled sources, and early consumer break (including fallback path).

### DIFF
--- a/async/unstable_pool_settled.ts
+++ b/async/unstable_pool_settled.ts
@@ -29,7 +29,7 @@ export interface PooledMapSettledOptions {
  *
  * If the input iterable itself throws, all currently executing items are
  * allowed to finish and their settled results are yielded, then the iterator
- * closes. The error from the input iterable is not propagated to the consumer.
+ * rejects with the original error from the input iterable.
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM, solid implementation

PR #7094, opened 2026-04-13
BREAKING(cli/unstable): rename `refreshMilliseconds` to `interval`

timreichen:
This PR renames `ProgressBarOptions.refreshMilliseconds` to `ProgressBarOptions.interval`  in order to align with `PollOptions.interval` and `SpinnerOptions.interval`.

### DIFF
--- a/cli/unstable_progress_bar.ts
+++ b/cli/unstable_progress_bar.ts
@@ -94,11 +94,11 @@ export interface ProgressBarOptions {
    */
   keepOpen?: boolean;
   /**
-   * How often the progress bar updates. The progress bar will be updated every
-   * `refreshMilliseconds` milliseconds.
-   * @default {1000}
+   * The time between each frame of the progress bar in milliseconds.
+   *
+   * @default {75}
    */
-  refreshMilliseconds?: number;
+  interval?: number;
 }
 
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
The `@default` tag was changed to `{75}` (matching Spinner) but the destructuring default is still `interval = 1000`. These should agree — 1000ms is probably the right default for a progress bar (once-per-second updates), so the tag should stay `@default {1000}`.

timreichen:
> The `@default` tag was changed to `{75}` (matching Spinner) but the destructuring default is still `interval = 1000`. These should agree — 1000ms is probably the right default for a progress bar (once-per-second updates), so the tag should stay `@default {1000}`.

Good catch, fixed.

bartlomieju (approved):

timreichen:
> The `@default` tag was changed to `{75}` (matching Spinner) but the destructuring default is still `interval = 1000`. These should agree — 1000ms is probably the right default for a progress bar (once-per-second updates), so the tag should stay `@default {1000}`.

On that note I wonder if we should have a lint plugin to compare jsdoc to the actual declared default value so this cannot happen anymore.
This would probably not too hard for literals to implement.

bartlomieju:
Yeah, sounds like a good idea 👍

PR #7095, opened 2026-04-13
BREAKING(cli/unstable): make `ProgressBarFormatter.styledData` a getter

timreichen:
This PR makes `ProgressBarFormatter.styledData` a getter instead of a function. It seems the function call and the `fractions` param don't do anything.


bartlomieju (approved):
Confirmed: `formatUnitFraction(value, max)` hardcodes `.toFixed(2)` and has no `fractions` parameter — the old function signature was dead API surface. Getter is the right call.

PR #7096, opened 2026-04-13
feat(async): stabilize AbortSignal support in debounce()

tomas-zijdemans:
Moving the abort signal support from unstable to the stable version. Deleting the unstable files since they now don't hold any new features, and also are out of sync with the stable version (they can always be added back in later if needed)


async/debounce_test.ts
```diff
@@ -128,3 +129,33 @@ Deno.test("debounce() re-invocation after flush re-schedules", () => {
...
+  let called = 0;
+  const controller = new AbortController();
+  const d = debounce(() => called++, 100, { signal: controller.signal });
+  d();
+  assertEquals(d.pending, true);
+  controller.abort();
+  assertEquals(d.pending, false);
+  await delay(200);
+  assertEquals(called, 0);
+});
```
bartlomieju, line 143: Nit: all existing debounce tests are synchronous (using `flush()` to trigger, never `delay()`). This introduces a real-time dependency that's inconsistent with the rest of the file.

You could make it synchronous — after `controller.abort()`, `d.pending` is already asserted false and the timeout is cleared, so the callback can never fire. But I see the value in belt-and-suspenders confirmation that `setTimeout` was actually cancelled, so I'm fine either way.

bartlomieju (changes requested):
Clean stabilization PR. The implementation is correct and tests cover the important cases. One required change, one nit.

async/debounce_test.ts
tomas-zijdemans: Ooops. that's a clear mistake. My bad. Updated now

bartlomieju (approved):

PR #7097, opened 2026-04-14
fix(async/unstable): circuit breaker stale openedAt and duplicate callbacks

tomas-zijdemans:
- Fix cooldown timer bypassed when wrapped function is slow
- Fix duplicate callbacks on concurrent half-open failures
- Extract `tryCall` helper to simplify predicate error handling
- Improve JSDoc and add test for synchronous throws

### DIFF
--- a/async/unstable_circuit_breaker.ts
+++ b/async/unstable_circuit_breaker.ts
@@ -589,8 +589,7 @@ export class CircuitBreaker<T = unknown> {
 
     try {
       if (this.#isResultFailure(result)) {
-        const syntheticError = new Error("Result classified as failure");
-        this.#handleFailure(syntheticError, currentState.state, currentTime);
+        this.#handleFailure(undefined, currentState.state, currentTime);
         return result;
       }
     } catch (predicateError) {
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM — both bug fixes are real and correctly addressed.

**Stale `openedAt`**: `currentTime` was captured at the top of `execute()` and passed into `#handleFailure` as `openedAt`. If `fn()` is slow, the cooldown starts from the past. Using `Date.now()` at failure time is the right fix.

**Duplicate callbacks**: When `halfOpenMaxConcurrent > 1` and multiple requests fail concurrently, each `#handleFailure` call saw `previousState === "half_open"` and fired `onStateChange`/`onOpen`. Checking the current state to gate callbacks and preserving the existing `openedAt` is correct.

The `tryCall` extraction is clean. One subtlety: it only wraps the predicate, not `#handleFailure` (which the old try/catch did), but that's fine since `#handleFailure` can't throw — all callbacks go through `safeCallback`.

Nit: the `@returns void` tags on `forceOpen`, `forceClose`, `reset` are unnecessary — void return is self-evident from the signature.

PR #7100, opened 2026-04-20
feat(data-structures/unstable): add MultiMap

tomas-zijdemans:
Adds `MultiMap<K, V>` under `@std/data-structures/unstable`. It's a `Map`-like container where each key holds an ordered list of values.

Covers the common "values keyed by group" case (e.g. parsed query strings, HTTP headers, users grouped by role) without hand-rolling `Map<K, V[]>` wrappers at every call site.

I plan to use it in the cache and xml module.


data_structures/unstable_multimap.ts
```diff
@@ -0,0 +1,707 @@
...
+ * @experimental **UNSTABLE**: New API, yet to be vetted.
+ *
+ * @typeParam K The type of the keys in the map.
+ * @typeParam V The type of the values in the map.
+ *
+ * @example Usage
+ * ```ts
+ * import { MultiMap } from "@std/data-structures/unstable-multimap";
+ * import { assertEquals } from "@std/assert";
+ *
```
bartlomieju, line 39:
```suggestion
 * | {@linkcode MultiMap.prototype.get}                          | Linear in the bucket size (copy)|
```

The linear cost is from the `.slice()` copy, not the lookup. Worth noting so readers don't think the lookup itself is slow.

data_structures/unstable_multimap.ts
```diff
@@ -0,0 +1,707 @@
...
+   *
+   * @example Usage
+   * ```ts
+   * import { MultiMap } from "@std/data-structures/unstable-multimap";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const map = new MultiMap([["a", 1], ["a", 2], ["b", 3]]);
+   * assertEquals(map.delete("a"), true);
+   * assertEquals(map.delete("a"), false);
+   *
```
bartlomieju, line 241: Worth a brief comment noting that `Array.prototype.includes` already uses SameValueZero semantics — so this is consistent with the hand-rolled SameValueZero loop in `deleteEntry`. Without the note, readers may wonder why the two methods use different comparison approaches.

data_structures/unstable_multimap.ts
```diff
@@ -0,0 +1,707 @@
...
+   * @experimental **UNSTABLE**: New API, yet to be vetted.
+   *
+   * @returns An iterator of `[key, value]` pairs.
+   *
+   * @example Usage
+   * ```ts
+   * import { MultiMap } from "@std/data-structures/unstable-multimap";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const map = new MultiMap([["a", 1], ["a", 2], ["b", 3]]);
```
bartlomieju, line 405: This split to avoid `.call()` overhead adds ~15 lines and a type cast for a micro-optimization that V8 inlines on hot paths. I'd simplify to just the `.call(thisArg, ...)` branch.

Also: when `thisArg` is explicitly `undefined`, the two branches behave differently in sloppy mode (`this` is `globalThis` with `.call(undefined, ...)` but truly `undefined` with direct call). Moot in strict mode / std, but still a subtle behavioral divergence.

data_structures/unstable_multimap.ts
```diff
@@ -0,0 +1,707 @@
...
+   *
+   * @returns An iterator of keys.
+   *
+   * @example Usage
+   * ```ts
+   * import { MultiMap } from "@std/data-structures/unstable-multimap";
+   * import { assertEquals } from "@std/assert";
+   *
+   * const map = new MultiMap([["a", 1], ["a", 2], ["b", 3]]);
+   *
```
bartlomieju, line 508: `entries()` and `values()` are hand-rolled iterators ("~5x faster than generators"). But `groups()` uses a generator. If the perf argument holds, it should apply here too. If it doesn't matter for `groups()` (fewer yields per key), then generators would be fine for `entries()`/`values()` as well. I'd pick one style for consistency — either hand-roll all three or use generators for all three.

bartlomieju (changes requested):
Nice work — well-documented, well-tested, and covers a real gap. A few things to address:

### `valueCount` getter

`size` counting distinct keys follows `Map` convention, but multimap users very commonly need total value count. Consider adding a `valueCount` getter, cheaply maintained with an internal counter (increment in `add`, decrement in `deleteEntry`, reset in `clear`/`delete`). Without it the only way is to iterate everything.

### `Deno.customInspect`

Not a blocker (other unstable data structures don't have it either), but `[Symbol.for("Deno.customInspect")]` would improve DX — `console.log(multimap)` would show something useful rather than an opaque object.

tomas-zijdemans:
Thanks for the thorough review.

- valueCount: Added a valueCount getter, backed by an internal counter as suggested
- Deno.customInspect: Added one small hook so console.log(map) prints nicely. Opted not to add a Node-side twin (nodejs.util.inspect.custom), keeping this in line with the other data_structures modules. The Deno-side hook alone already fixes the unreadable default output; we can add the Node twin later if there's demand.
- get() complexity note: Clarified the table row to Linear in the bucket size (copy) as suggested
- SameValueZero consistency: Added a short comment on hasEntry() noting that .includes() already matches the SameValueZero loop in deleteEntry()
- Simplify forEach(): Removed the thisArg-undefined branch; one loop with callbackfn.call(thisArg as T, ...) covers both cases.
- Iterator style: Benchmarked hand-rolled vs. generator (1.7×–4.4× faster, never slower). Rewrote groups() as a hand-rolled iterator to match entries()/values(); kept keys() delegating to the native Map iterator.

bartlomieju (approved):
Thanks!

PR #7102, opened 2026-04-22
feat(data-structures/unstable): align RollingCounter with other data structures, add `at()` and `toArray()`

tomas-zijdemans:
This PR is mostly about aligning RollingCounter with Deque, IndexedHeap, MultiMap and other data-structures.

- RollingCounter now implements Iterable<number>.
- [Symbol.toStringTag] is a readonly "RollingCounter" as const property (was a getter).
- @experimental tag added to every public-member JSDoc.
- Class JSDoc gained a complexity table.
- New at(index) with Array.prototype.at semantics (oldest→newest, negatives from newest end, out-of-range → undefined).
- New toArray(); toJSON() delegates to it.
- RollingCounterSnapshot.segments is readonly number[]; field is readonly.
- from() accepts RollingCounterSnapshot, guards null/undefined/non-object, throws TypeError for shape errors, keeps RangeError for per-element numeric errors.
- Tests: two RangeError cases flipped to TypeError; added null/undefined snapshot, toArray, and at tests.

### DIFF
--- a/data_structures/unstable_rolling_counter.ts
+++ b/data_structures/unstable_rolling_counter.ts
@@ -11,7 +11,7 @@
  */
 export interface RollingCounterSnapshot {
   /** Segment counts ordered from oldest to newest. */
-  segments: number[];
+  readonly segments: readonly number[];
 }
 
 /**
@@ -23,6 +23,21 @@ export interface RollingCounterSnapshot {
  * window on your own schedule.
  *
  * The class is iterable and yields segment counts from oldest to newest.
+ *
+ * | Method            | Time complexity                        |
+ * | ----------------- | -------------------------------------- |
+ * | increment(n)      | Constant                               |
+ * | rotate(steps)     | Linear in min(steps, segment count)    |
+ * | current           | Constant                               |
+ * | total             | Constant                               |
+ * | segmentCount      | Constant                               |
+ * | at(index)         | Constant                               |
+ * | clear()           | Linear in the number of segments       |
+ * | toArray()         | Linear in the number of segments       |
+ * | toJSON()          | Linear in the number of segments       |
+ * | [Symbol.iterator] | Linear in the number of segments       |
+ * | RollingCounter()  | Linear in the number of segments       |
+ * | RollingCounter.from(snapshot) | Linear in the number of segments |
  *
  * @experimental **UNSTABLE**: New API, yet to be vetted.
  *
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Overall solid PR. The new `at()` and `toArray()` methods are well-implemented, and the alignment changes (`Iterable`, readonly snapshot, `@experimental` tags, complexity table, `Symbol.toStringTag` as property) are all consistent with how `Deque` and other data structures work in this repo. Tests are thorough.

Two things:

**1. Empty segments should throw `RangeError`, not `TypeError`**

An empty array is the correct type — it's just the wrong length. The constructor throws `RangeError` for the equivalent constraint (`segmentCount < 1`), so `from()` should do the same for consistency. The other two `TypeError` changes (null/non-object snapshot, non-array segments) are correct.

**2. PR scope vs title**

The title says "add `at()` and `toArray()`" but the PR also changes the snapshot interface to readonly, adds `implements Iterable<number>`, switches `Symbol.toStringTag` to a readonly property, adds `@experimental` to all members, adds the complexity table, and reworks `from()` validation. The description documents all of this, but the commit message will be misleading in `git log`. Worth updating the title/commit message to something broader like "align `RollingCounter` with other data structures, add `at()` and `toArray()`".

tomas-zijdemans:
Thanks,
- PR title and commit message changed as requested
- Changed from TypeError to RangeError as requested

lunadogbot (approved):
Author addressed both points from the prior review: `from()` now throws `RangeError` for empty `segments` (consistent with the constructor) and keeps `TypeError` for null/non-object snapshots and non-array `segments`; title is broadened to match the actual scope. The `readonly [Symbol.toStringTag] = "Foo" as const` field matches the pattern used by `Deque`, `IndexedHeap`, and `MultiMap`, so the getter → field switch is alignment, not drift. `at()` is a constant-time read off the ring buffer (`segments[(cursor + 1 + index) % len]`) with the canonical Array-style negative-index handling, and `toArray()` is the existing `toJSON()` body factored out — `toJSON()` correctly delegates to it.

- nit: `at(1.5)` returns `undefined` rather than `at(1)` because the fractional index leaks into the modulo. `Array.prototype.at` floors via `ToIntegerOrInfinity` first. Probably not worth coercing, but if you want strict parity, `index = Math.trunc(index)` at the top of the method would do it.

PR #7104, opened 2026-04-24
fix(path): improve Node.js compatibility, run tests in node and bun

kt3k:
This PR adds tests of `path` in Node and Bun

ref #7103


path/posix/to_file_url.ts
```diff
@@ -26,7 +27,7 @@ export function toFileUrl(path: string): URL {
 
   const url = new URL("file:///");
   url.pathname = encodeWhitespace(
-    path.replace(/%/g, "%25").replace(/\\/g, "%5C"),
+    path.replace(/^\/+/, "/").replace(/%/g, "%25").replace(/\\/g, "%5C"),
```
kt3k, line 30: The below works differently in Deno and Node:
```ts
const u = new URL("file:///");
u.pathname = "//foo";
console.log(u.href);
// prints file:///foo in Deno
// prints file:////foo in Node, Bun, and browsers
```

This change normalizes the differences above (and therefore no breaking changes to deno users)

kt3k: Created an issue in `rust-url` about the above behavior https://github.com/servo/rust-url/issues/1118

lunadogbot (approved):
`path/posix/to_file_url.ts`: collapsing leading `/+` before assigning `url.pathname` is the right fix — Deno already produced `file:///foo` from `pathname = "//foo"`, this aligns Node/Bun to that. Existing `posix.toFileUrl("//localhost/...")` cases in `to_file_url_test.ts` cover it. The `is_glob` worker rewrite is also an improvement — it now `assertEquals` the three results, where the data-URL version only asserted that `postMessage` ran.

- nit: title is `test(path)` but the `to_file_url.ts` change is a runtime behaviour tweak, not test-only. `fix(path)` would be more accurate.

kt3k:
> nit: title is test(path) but the to_file_url.ts change is a runtime behaviour tweak, not test-only. fix(path) would be more accurate.

updated the title following the above message

bartlomieju (approved):
Thanks!

